PR #2666, opened 2015-08-03
Traverse views function to add bk-* classes

mistakevin:
resolves https://github.com/bokeh/bokeh/issues/2527

PR #3177, opened 2015-11-23
Restore jquery-ui as it was before browserify

### LINKED ISSUE #2527: Rendered slider* plots have missing bk-* classes
May happen on more than just sliders.

```
s = Slider(start=0, end=92, value=1, step=1, title="Time", callback=callback)
show(vplot(s))
```

renders as:
screen shot 2015-07-08 at 9 33 59 am

but `script, div = components(s)` outputs code that renders as
screen shot 2015-07-08 at 9 36 36 am

mattpap:
This fixes #2527 and supersedes #2666. This reverts jquery-ui removal, removes prefixer and updates jquery-ui to work with browserify. Also, now we will use jquery-ui's JS and CSS from one version, as opposed to what is going on in master.


bryevdv:
@mattpap If we are going to do this, we should make fork jq-ui under the Bokeh org and install from there as is being pursued with the Numbro PR.

Was there any discussion about this?

mattpap:
I'm not sure this is worthwhile, because it's unlikely we will be ever updating jquery-ui. Furthermore, as we are considering using phosphor widgets, this seems even more unnecessary.

bryevdv:
In that case why do anything at all until we switch to phosphor?

mattpap:
Because we have a bug and we won't switch to phosphor before the upcomming release or two, but it's questionable if we need to invest in the infrastructure that most likely is not going to be used. So, usually I'm for fixing a problem and doing it properly, but in this case it doesn't seem economical.

mattpap:
Ping @bryevdv.

bryevdv:
Weird I left a comment and it seems to have disappeared. It's just a big amount of churn for one bug. I don't have a good feel for how serious it is, though. If you think it is serious enough that we can't just wait as-is until the Phosphor replacement is ready, I will defer to your judgment.

mattpap:
@bryevdv, you commented on the issue (#2527). Given this issue have been with bokeh.widgets for over a year and not many people complained, then it's not that big of a deal. This PR is the effect of us struggling to get PR #2666 working properly. Note also that I wasn't sure if we will be pursing phosphor widgets at the time of resolving this issue. Perhaps @mistakevin and @birdsarah could chime in.

As to repo churn, I created [1], it's a bit smaller.

https://github.com/bokeh/bokeh/compare/mattpap/restore-jquery-ui-small?expand=1

bryevdv:
I think we are in agreement that it is not that big of a deal, so we can just leave things as they currently are and make changes when real work to integrate phosphor widgets and layout begins.

birdsarah:
I'm fine with not doing this and pushing forward - getting modularized js that havoc's suggestin and phosphor working seem like bigger wins. I'm much more excited about using whatever slider widget I want than bokeh's built in one.

bryevdv:
OK, I will close this PR tomorrow unless there is further comment.

PR #3183, opened 2015-11-24
Patches with hole (test implementation)

birdsarah:
This doesn't have any tests and is a first attempt. The basics seem to work, but would appreciate eyes and feedback from @bokeh/dev 

See a demonstration here: https://rawgit.com/bokeh/bokeh/scratch/patches_with_hole/examples/plotting/file/patches_with_holes.html

(note the second plot has tap for hit testing)

Details:
* patches can have a hole
* patches with hole can still be multipatches (ie separated with a NaN to make seperate but connected entities)
* basic hit testing works - but it captures the hit anywhere in the exterior ring (i.e. it doesn't respect the cut out)
* if you do a stroked patch with hole, it fills the non-cut out behavior as white - that seems correct to me.
* if you were to have an interior ring with a NaN in the middle e.g. `[[1,2, NaN, 4],[2,3,4,1]]` or  [[1,2, NaN, 4],[2,3,NaN,1]] things would probably break down but that data doesn't make sense, so I think that's ok.


bokehjs/src/coffee/renderer/glyph/patches.coffee
```diff
@@ -81,40 +128,19 @@ class PatchesView extends Glyph.View
...
-            ctx.closePath()
-            ctx.stroke()
-            ctx.beginPath()
-            continue
-          else
-            ctx.lineTo(sx[j], sy[j])
+        if _.isNumber(sx[0])
+          @_render_polygon(ctx, sx, sy, 'stroke')
+        else
+          @_render_polygon_with_hole(ctx, sx, sy, 'stroke')
```
almarklein: So when you render a hole, you provide tuples?

birdsarah:
Hi @almarklein,

tuples may work, I haven't tried, but the expected way is a list of lists. I have modeled it on the GeoJSON spec: http://geojson.org/geojson-spec.html#id4 - "The first element in the array represents the exterior ring. Any subsequent elements represent interior rings (or holes)."

So, here's what a data source for patches might look like:

type | xs 
------------ | -------------  
current patches |  `[3, 3.1, 4, 3.9]`
current patches with two distinct items | `[5, 6, 5.5, np.NaN, 5, 6, 5.5]`
new patches with hole | `[[1, 1.1, 2, 1.9], [1.8, 1.7, 1.2, 1.2]]`
new patches with hole with distinct items | `[[1, 1.1, 2, 1.9], [1.8, 1.7, 1.2, 1.2], np.NaN, [3, 3.1, 4, 3.9]]`

imagine the ys having appropriately matching items.

see lower chart of https://rawgit.com/bokeh/bokeh/scratch/patches_with_hole/examples/plotting/file/patches_with_holes.html for how this all renders.

birdsarah:
@bokeh/dev - looking to do the full implementation of this week - with tests etc. Would really appreciate feedback. If not, I'll assume it's on the right track.

bokehjs/src/coffee/mapper/linear_mapper.coffee
```diff
@@ -15,9 +16,17 @@ class LinearMapper extends HasProperties
 
   v_map_to_target: (xs) ->
     [scale, offset] = @get('mapper_state')
-    result = new Float64Array(xs.length)
-    for x, idx in xs
-      result[idx] = scale * x + offset
+    if not _.isNumber(xs[0])
```
bryevdv: what will this do if `x[0]` is `NaN`? Would `_.isArray(xs[0])` work?

birdsarah: Nope, sadly. This was one of my big learnings working this up: `_.isArray()` returns `false` for Float64Array's so `_.isArray()` is often not good.

birdsarah: To answer your original question:

http://underscorejs.org/#isNumber
isNumber_.isNumber(object)
Returns true if object is a Number (including NaN). 

So, in this case: `if not _.isNumber(xs[0])` is my work around for `_.isFloat64Array(xs[0])` (which doesn't existing in underscore).

birdsarah:
Superceded by: #3399

PR #3204, opened 2015-11-28
Warnings cleanup

bryevdv:
This PR cleans up a number of runtime warnings, depends on #3190


bryevdv:
will close and re-open against master to see tests run

PR #3206, opened 2015-11-29
add hover_glyph to specify hover policy

### LINKED ISSUE #1476: GlyphRenderer should have hover_glyph property

bryevdv:
issues: closes #1476

This PR enables styling inspected (hovered) glyphs by specifying a `hover_glyph` property on a `GlyphRenderer` (or as a convenience, using `hover_fill_alpha`, etc. with the `bokeh.plotting` interface). This eliminates the need to write any custom callbacks to do highlighting on hover. 

Note: with this implementation, the hovered glyphs are always rendered ***on top of*** the regular glyphs, not "instead of" the regular glyphs. Thus, typically a hover glyph would want to have a high alpha to obscure the glyph below it. 

An example you can try out below:
```
from bokeh.models import ColumnDataSource, Circle, HoverTool

output_file("hover.html")

x, y = data.ix['2010-10-06'].index.to_series(), data.ix['2010-10-06']['glucose']

# Basic plot setup
p = figure(width=600, height=300, x_axis_type="datetime", tools="", toolbar_location=None, title='Hover over points')
p.line(x, y, line_dash="4 4", line_width=1, color='gray')

# Add a circle, that is visible only when selected
# source = ColumnDataSource({'x': x, 'y': y})
# invisible_circle = Circle(x='x', y='y', fill_color='gray', fill_alpha=0.5, line_color="white", size=20)
# visible_circle = Circle(x='x', y='y', fill_color='firebrick', fill_alpha=0.9, line_color="blue", size=20)
# cr = p.add_glyph(source, invisible_circle, hover_glyph=visible_circle)

cr = p.circle(x, y, fill_color="grey", alpha=0.1, line_color=None, size=20, 
              hover_fill_color="firebrick", hover_alpha=0.5, hover_line_color="white")

p.add_tools(HoverTool(tooltips=None, renderers=[cr], mode='hline'))

show(p)
```

This PR currently WIP, still needed:
* [x] documentation support
* [x] factoring/cleanup
* [x] appropriate tests


bryevdv:
OK since hover highlight does not need callback now, I added a new example for that. Hover over points to show neighbors connected:
screen shot 2015-11-30 at 9 33 43 pm

In order to reword the selections/inspections highlight section, I also fixed a longstanding omission in `bokeh.plotting` glyph functions, which was the inability to style `selection_glyph` properties. This was a pretty small change in `plotting_helpers.py` but it probably bears a little scrutiny.

bryevdv:
I'm not really sure what sort of tests can be added. The main JS changes were in the middle of glyph renderer, which probably needs to be split up in order to be tested effectively. Barring that, I would say this is ready. @bokeh/dev

bokeh/plotting_helpers.py
```diff
@@ -125,6 +125,7 @@ def _process_sequence_literals(glyphclass, kwargs, source):
         kwargs[var] = var
 
 def _make_glyph(glyphclass, kws, extra):
+        if extra is None: return None
```
fpliger: What's the need for this? I fell I'm missing context here. In case extra is `None` I'd expect you still want to call `glyphclass(**kws)` and just skip `kws.update(extra)`

bryevdv: It's just to make these two lines simpler: 

https://github.com/bokeh/bokeh/pull/3206/files#diff-3e6325a902133d8fd2e886fe13214143R464

If there are no `selection_` or `hover_` params, then we want to set those glyphs to `None` not to copies of the main glyph.

bryevdv: I don't think this is great code, BTW, but it's already not great code. Some day all of this should be more finely factored...

birdsarah:
this is marked as ready, but you'd mentioned adding tests in your initial checklist

bryevdv:
@birdsarah see my comment above :) I'm not sure how this can be effectively tested at the moment but would like ideas.

birdsarah:
@bryevdv - sorry somethings going wrong with my reading github comments on my email. I'll take a look.

bokeh/plotting_helpers.py
fpliger: :) got it.

I think the code is ok. I was more concerned about the intention. I see that it's convenient and that's fine as _make_glyph is a private factory and is used in "conscious" way. I'd me negative about this if _make_glyph was more exposed.

sphinx/source/docs/user_guide/source_examples/styling_glyph_selections_plotting_params.py
```diff
@@ -0,0 +1,18 @@
...
+renderer = plot.circle([1, 2, 3, 4, 5], [2, 5, 8, 2, 7], size=50,
+
+                       # set visual properties for selected glyphs
+                       selection_color="firebrick",
+
+                       # set visual properties for non-selected glyphs
+                       nonselection_fill_alpha=0.2,
+                       nonselection_fill_color="blue",
+                       nonselection_line_color="firebrick",
+                       nonselection_line_alpha=1.0)
```
mattpap: That's off topic, but I wonder how far would you go with this type of indentation? I think that at this point it already looks ugly and I would prefer just simple by-one indentation.

mattpap:
@bryevdv, I see only added/modified examples in the documentation. Would you mind adding one in `plotting/file` or `glyphs`? It would be much easier to test by inspection having at least one in an usual location.

mattpap:
I went ahead and added one.

birdsarah:
This doesn't update for text props, I'd like to be able to do something like this:
```python
    text = plot.text(
        x=[1, 2], y=[1, 1],
        text=['Hovered', 'Hovered'],
        text_alpha=0.1, hover_text_alpha=0.9,
    )
```

mattpap:
One problem I see is that my CPU gets pretty angry when I use hover tool in this PR. This obviously won't be fixed here.

bryevdv:
@mattpap thanks for adding the example. I had not noticed a CPU spike, but I can do at least a quick pass on profiling in case it is something very simple and stupid to improve.

bryevdv:
> This doesn't update for text props, I'd like to be able to do something like this:

I doesn't work for text because text glyphs don't yet support any hit testing. I would expect it to work immediately if a `_hit_point` method was added to the text glyph. I think that is outside the scope of this PR, though.

bryevdv:
So drilling down into things a little, I don't actually see anything jumping out at me too much, but I did notice this:

screen shot 2015-12-02 at 5 47 34 pm

I think the axis layout update needs to be examined. 

 I also noticed this:

screen shot 2015-12-02 at 5 38 27 pm

It seems to be invalidating the dom layout all the time? am I misreading that? @birdsarah are we doing something to cause alot of extra unnecessary work?

bryevdv:
@mattpap @birdsarah any objections to merging this as-is?

mattpap:
> any objections to merging this as-is?

No. LGTM.

bryevdv:
OK I will merge in an hour unless I hear something from @birdsarah

birdsarah:
> It seems to be invalidating the dom layout all the time? am I misreading that? @birdsarah are we doing something to cause alot of extra unnecessary work?

Probably - we are creating and destroying dom elements all the time with hover. It's why I added `tooltips=None` feature before as it greatly improves performance. Two thoughts:

1) I'm not sure if in the example you're profiling you have tooltips on, if not then there maybe something wrong with the implementation
2) We should handle tooltips smarter anyway and keep updating a single dom element rather than creating and destroying one all the time which I believe will be more efficient - should also make it easier to style - currently it's a pain trying to grab the hovertool in developer tools.

birdsarah:
> I would expect it to work immediately if a _hit_point method was added to the text glyph. 

Makes sense - thanks!

bryevdv:
@birdsarah great I've started a branch to play around with some optimizations, I'll push some smaller individual PRs later.

PR #3024, opened 2015-10-24
Replace legacy charts with new ones

fpliger:
Replaces old legacy charts left with new ones based on the CompositeGlyphs and more advanced builders design.

PR #3208, opened 2015-11-29
Replace legacy donut chart

rothnic:
This pull request implements the remaining chart on top of #3024, so it can be merged as soon as possible, while this is finished up. This should be completed by Dec 2nd.

Still left to do:

- Update examples (file/notebook)
- Simplify input options/infer default inputs when some are omitted
- Update documentation
- (beyond legacy chart's capability) Implement some kind of legend
- (beyond legacy chart's capability) Expand input options
    - wedge/text attributes
    - coloring within categorical hierarchy
    - coloring by level


rothnic:
Current example uses the autompg data for columns 'cyl' and 'origin'. First level in this screenshot has the cylinders and second level has the origin values that are associated with each cyl value from the first level.

image

rothnic:
I'll merge master in and get the notebook example updated tonight. Currently looks like this:

[screenshot](https://cloud.githubusercontent.com/assets/452052/11486854/42b44474-9781-11e5-9e18-b0ee9301865c.png)

bryevdv:
Merge master should fix TravisCI

PR #3213, opened 2015-12-01
conda list returns too much without -f

bryevdv:
Travis install script currently reporting an error:
```
DEPS_FLAG='['\''pytest-2.8.1-py34_0'\'', '\''pytest-cov-1.8.1-py34_0'\'', '\''pytest-html-1.6-py34_0'\'', '\''pytest-selenium-1.0b1-py34_0'\'', '\''pytest-variables-1.2-py34_0'\'']'
+'[' '!' '['\''pytest-2.8.1-py34_0'\'',' ''\''pytest-cov-1.8.1-py34_0'\'',' ''\''pytest-html-1.6-py34_0'\'',' ''\''pytest-selenium-1.0b1-py34_0'\'',' ''\''pytest-variables-1.2-py34_0'\'']' == '*pytest*' ']'
scripts/travis_install: line 54: [: too many arguments
```

Due to `conda list foo` being very permissive by default and actually searching for `*foo*`


bryevdv:
@damianavila I thought this was a small simply fix for the error in `travis_install` but maybe fixing it uncovered some other problem? Just the py2 notebook examples are failing here (I have restarted several times). It seems they are trying to pull resources from CDN? But I don't understand why py3 notebooks would not fail in the same way. Either way I guess we need to check resource usage in the tests.

bryevdv:
Actually, other PRs are failing in the same way it seems.

damianavila:
LGTM, the resources thing is unrelated...

PR #3216, opened 2015-12-01
Add missing call to output_file in quickstart.rst

vahtras:
The example in the section "Linked panning and brushing"
in the Quickstart documentation appear to be missing a call to
bokeh.plotting.output_file. Without this call, running the example
produces no output.


bryevdv:
Awesome, thanks for the catch @vahtras

bryevdv:
There is a known ongoing issue with our CI machinery, I am going to go ahead and merge this now.

PR #3218, opened 2015-12-01
try forcing inline resources for notebooks

bryevdv:
Just a test

### DIFF
--- a/tests/travis/test
+++ b/tests/travis/test
@@ -552,6 +552,9 @@ def start_jupyter_notebook():
     #First create the customjs to run the notebook
     create_custom_js()
 
+    env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


bryevdv:
@damianavila @mattpap this seems to fix the problem (tests still running but notebooks appear to be passing now). Is this the right solution? FWIW I can't run notebooks at all with "relative" resources, it tries to load resources from a relative URL unsurprisingly.

damianavila:
LGTM as a temporary patch... we need to explore the problem further...

bryevdv:
OK, tests pass and report looks good (modulo known issues). Merging as a hotfix.

mattpap:
Apparently this is the right workaround, because everything else is either broken or sometimes works by coincidence (`relative` resources can be made usable).

PR #3225, opened 2015-12-02
Add bokeh logo and a link to the gapminder demo

birdsarah:
Just makes it look a bit nicer

gapminder-720


mattpap:
Skipping CI.

mattpap:
That's a great example that I wasn't aware of.

PR #3230, opened 2015-12-03
Add more missing calls to output_file in plotting

vahtras:
Examples in subsection "Rectangles and Ovals" lack
calls to output_file


bryevdv:
Thanks for the catch @vahtras !

PR #3241, opened 2015-12-05
Added Overrides to rid the RuntimeWarning from properies

brianpanneton:
All of these seemed like intended overrides. Might want to double check though.


brianpanneton:
Found and issue with this patch

PR #3251, opened 2015-12-06
Bokeh Command docs

bryevdv:
This is a PR to work up documentation for the Bokeh command and in particular the `bokeh serve` command. 

Test docs have been deployed to:

http://bokeh.pydata.org/en/server_branch/


bryevdv:
@bokeh/dev please review, staged docs deploy has been updated above. There's more to say but I'd like this to go in a dev build tomorrow morning so the docs are deployed for real.

bryevdv:
@bokeh/core  OK There's certainly more to say and add, but I'd like at least this much to go into the next dev build. Feedback on the docs still welcome.

Issue #1683, opened 2015-01-12
Reset orignal plot's size


imaluengo:
Currently the `reset` tool of the plots only resets the point of view and the zoom of the data. But, if you resize the plot the only way of recovering its original size is by refreshing the web (in standard web deployment) or by replotting (in ipython notebook).

Is not really a problem, but it would be nice to be able to resize the plot with the `reset` or other tool.

bryevdv:
This will be done in #3254

PR #3254, opened 2015-12-07
Add support for undo and redo tools

mattpap:
This works currently with pan, wheel zoom and box zoom tools, thought wheel zoom is too granular. Tool buttons are disabled if can't undo/redo. Reset tool clears the history. Future developments could include a dropdown menus showing undo/redo history (needs better toolbar implementation).

I was too lazy to base64-encode new icons, so I switched from manually encoded icons to using `data-uri()` function in less. This way icons were moved from JS to CSS. ~~Given the current implementation, this required to use `content` CSS property, which may cause compatibility issues (IE). If so, we can always switch to using `background-image` later, but before release~~.


bryevdv:
@mattpap have you coordinated with @birdsarah? She is working on `+/-` tool buttons for zoom in/zoom out and it seems like these efforts need some coordination. 

How does CSS `data_uri` work with standalone documents? Is the file encoded by `less` and made the `content`?

mattpap:
> How does CSS data_uri work with standalone documents?

The same as before, just the process is automated and you will be warned if you run into IE compatibility issues (32k limit).

mattpap:
> it seems like these efforts need some coordination

Depends what will be merged first. Adding state manipulation code to new tools is usually just one line.

bryevdv:
> The same as before, just the process is automated and you will be warned if you run into IE compatibility issues (32k limit).

great, let's just keep them small.

bryevdv:
> Is there any PR with her work?

Not yet AFAIK. ping @birdsarah

mattpap:
> great, let's just keep them small.

Currently all icons are under 2k, thought it seems quite a lot given their simplicity (but that's a separate issue).

mattpap:
btw. Undo/redo icons are scrap from reset icon, so they look so-so. Someone with more arts and crafts abilities should look into this later.

bryevdv:
> Undo/redo icons are scrap from reset icon, so they look so-so. Someone with more arts and crafts abilities should look into this later.

Please contact Victoria about better icons.

birdsarah:
- I got caught up on other PRs so haven't started zoom in/out so no worries on my end.

- I have just started testing this locally, icons do not display on firefox (but do on safari and chrome)

birdsarah:
Tap & polyselect tool aren't showing in safari either - I'm getting the error: 

Unexpected CSS token: : mixins.less:354
Unexpected CSS token: ; continuum.less:402

birdsarah:
I've tried this out. I found that plotting/file/color_scatter.py was a good example to test it on because it has a lot of tools on it.

I found it a little un-intuitive (and I knew what to expect). The underlying problems here aren't with this PR, but they affect this PR.

See the screenshot below:
screen shot 2015-12-07 at 11 32 02 am

All the items highlighted in red affect range. The ones on the left change it, the three on the right reset it. People already get confused that selections don't get reset when you hit the reset button, i think this is only going to increase that confusion. I understand why the tools are in those groups - it makes a lot of sense behind the scenes, but i'm not sure it's a meaningful distinction for users.

The undo/redo work like reset - that is only on range changes.

But if we look at the tool bar

mattpap:
> The undo/redo work like reset - that is only on range changes.

This is how it works currently. I didn't investigate if this can be extended further.

mattpap:
@birdsarah, don't concern your self too much with the state of the art of the toolbar. It's a mess. I'm going to clean this up, making toolbar a first class citizen.

mattpap:
Looks like `content` doesn't work at all in FF, at last as expected by a fairly sane person. I'm going to use `background-image` instead.

bryevdv:
I'm really not opposed to having reset also reset selections, just FYI. Resetting ranges what just what I happened to have in mind when I made the tool.

mattpap:
@bryevdv, @birdsarah, I switch to `background-image`, so now it works it FF/Chrome. I also fixed missing icons.

mattpap:
> Unexpected CSS token: : mixins.less:354

Where this comes from? Less won't parse under a CSS parser.

mattpap:
@birdsarah, would you mind starting an issue with your thought about the toolbar layout?

birdsarah:
> @birdsarah, would you mind starting an issue with your thought about the toolbar layout?

Sure.

----

>> Unexpected CSS token: : mixins.less:354

> Where this comes from? Less won't parse under a CSS parser.

That comes from Safari console, when using relative resources - screenshot:

screen shot 2015-12-07 at 12 15 09 pm

----

>>  The undo/redo work like reset - that is only on range changes.

>This is how it works currently. I didn't investigate if this can be extended further.

Yes, I think that's how it should work. Sorry, that was a straggling comment - I pressed "comment" to early.

birdsarah:
Re: the console log - it may not be from your PR.

mattpap:
@bokeh/dev any final thoughts on this?

bryevdv:
I will say I am a little dubious about the utility of a "redo" button, or at least I am not sure that "undo an undo" is the same as redo. But I am fairly neutral on this otherwise. Would just having the `ResetTool` reset more (i.e., canvas size and selections) be a simpler interim measure? I'm only worried about introducing new models that we might decided we don't need later.

mattpap:
> "undo an undo" is the same as redo.

Yes. This works as in any text editor or an editor of other kind.

mattpap:
>  I'm only worried about introducing new models that we might decided we don't need later.

That's a risk of software development business. If you are concerned about bokeh.js size, then the true solution is to generate custom bokeh.js (at least) for static HTML plots (which I will be working on in near future).

bryevdv:
> > "undo an undo" is the same as redo.

> Yes. This works as in any text editor or an editor of other kind.

Not exactly? In a text editor:

* I delete a line
* I hit redo and delete a line
* I hit redo and delete a line 
* ...

This seems to be: 

* I pan
* I hit undo, and that "un-pans"
* I hit redo, and that restores the pan
* I hit redo again, what happens?

birdsarah:
> I hit redo again, what happens?

I thought it was disabled if it had no more events to come off the stack. This aspect made sense to me when I was using it.

What made less sense is that the stack only has a subset of actions in it:

* I pan
* I select
* I zoom
* Undo - undo zoom
* Undo - undo pan

@mattpap asked me to move my comments about the toolbar into a separate issue, and I have, but I think that this is a tricky addition either:
* given the state of the toolbar.
* or that reset, undo, redo don't reverse selections.

Either one would ease the pain of this addition in my mind. 

On the flip side, undo/redo are not being added as default tools and as a result I'm ok with them living in obscurity.

birdsarah:
From the PR's point of view, there are no tests and that's a problem for merging. It would be easy to write some selenium tests for undo and redo.

mattpap:
I added support for undo/redo selections and resize. Note that now reset tool is out of sync, because it resets pan/zoom only.

bryevdv:
> I added support for undo/redo selections and resize. Note that now reset tool is out of sync, because it resets pan/zoom only. 

I'm really ok having it reset more things. Would you like to add that to this PR? If not, LGTM to merge

mattpap:
> I'm really ok having it reset more things. Would you like to add that to this PR?

I can do it here.

What about making it configurable as well (e.g. allow `pan`, `zoom`, `selection`, `resize` or a combination of those)? I have some stashed code for this. Thought maybe not a great idea at this point, because this would have to be configured on each tool (undo, redo, reset) individually.

bryevdv:
I do ultimately want that but lets leave configuration for another PR

bryevdv:
closes: #1683

mattpap:
@bryevdv, reset tool updated. I expect tests will still fail, they fail locally, but I can't make any sense of it with all the warnings and other garbage logged to the screen.

bryevdv:
> but I can't make any sense of it with all the warnings and other garbage logged to the screen.

Are you talking about the JS unit tests? That's the only place I know of now that has extra output (from running server tests)

mattpap:
Yes, I should have mentioned that. Failure is related to defaults, which I didn't change to this extent (as the failure indicates). Maybe this came from `master` merge. I'm not sure.

mattpap:
@bryevdv, I figured it out, I actually added a bokehjs-only property to tools. Unfortunately, presentation of  of the output from defaults' tests is beyond my perception. Dumping bare dictionaries in one big blob of text is not the way to go.

bryevdv:
@mattpap you should have seen it when there were 150 default mismatches :) But, almost all of those were simple values, not the dicts that are left now. FWIW the last 12 are all matching, there is just an issue with the test that shows them differing when there really aren't. That test output will be gone soon, and any new errors would probably be much easier to read.

mattpap:
Test failures seem unrelated. Perhaps another `master` merge would help. Anyway, this has been in ready state for a while.

bryevdv:
OK, merged master let's see if that helps the tests.

PR #3261, opened 2015-12-07
Rename Legend.{orientation->location} and implement more locations

mattpap:
Now we have total of ten locations: four corners, five centered and absolute locations. See `glyphs/legends`. This PR won't implement automatic location at this point. `Legend.orientation` is deprecated, but `@deprecated` decorator doesn't seem to work.

TODO:
- [x] be explicit about screen coordinates
- [x] update note in the documentation (absolute location)
- [x] make sure deprecation warning is shown

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -44,7 +44,8 @@ class Legend(Annotation):
         default="top_right", help="""
     The location where the legend should draw itself. It's either one of
     ``bokeh.enums.LegendLocation``'s enumerated values, or a ``(x, y)``
-    tuple indicating an absolute location (from the bottom-left corner).
+    tuple indicating an absolute location absolute location in screen
+    coordinates (pixels from the bottom-left corner).
     """)
 
     border_props = Include(LineProps, help="""
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -40,7 +40,8 @@ class Legend(Annotation):
     def orientation(self, location):
         self.location = location
 
-    location = Either(Enum(LegendLocation), Tuple(Float, Float), help="""
+    location = Either(Enum(LegendLocation), Tuple(Float, Float),
+        default="top_right", help="""
     The location where the legend should draw itself. It's either one of
     ``bokeh.enums.LegendLocation``'s enumerated values, or a ``(x, y)``
     tuple indicating an absolute location (from the bottom-left corner).
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -40,7 +40,7 @@ class Legend(Annotation):
     def orientation(self, location):
         self.location = location
 
-    location = Enum(LegendLocation, help="""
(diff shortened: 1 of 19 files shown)


bokeh/models/annotations.py
```diff
@@ -28,8 +28,23 @@ class Legend(Annotation):
...
+    @orientation.setter
+    @deprecated("Bokeh 0.11", "Legend.location")
+    def orientation(self, location):
+        self.location = location
+
+    location = Either(Enum(LegendLocation), Tuple(Float, Float),
+        default="top_right", help="""
+    The location where the legend should draw itself. It's either one of
+    ``bokeh.enums.LegendLocation``'s enumerated values, or a ``(x, y)``
+    tuple indicating an absolute location (from the bottom-left corner).
```
bryevdv: absolute position in data or screen coordinates?

bryevdv:
I think there was some problem with the deprecated decorator and properties, unfortunately.

bokeh/models/annotations.py
mattpap: Does `data` coordinates make any sense in this context? I assumed no, so it's implicitly `screen` coordinates.

mattpap: Or do you want me to make the docstring (and sphinx documentation) explicit about this?

bryevdv: It would be trivial to use the default mappers to map data coordinates to screen coordinates, and if someone wants to orient a legend wrt some data locations it doesn't seem unreasonable (tho they'd probably want to disable panning/zooming in this case)

bryevdv: Yes I am just asking to be explicit, sorry.

sphinx/source/docs/user_guide/styling.rst
```diff
@@ -621,18 +621,29 @@ for every element of the list:
...
+``"bottom_center"``
+
+``"bottom_left"``
+
+``"left_center"``
+
+``"center"``
+
+or a ``(x, y)`` tuple indicating an absolute location (measured from
+the bottom-left corner).
```
bryevdv: It's probably worth adding another example plot that shows the absolute location usage.

mattpap:
> I think there was some problem with the deprecated decorator and properties, unfortunately.

I may have a solution to this issue. WIP

sphinx/source/docs/user_guide/styling.rst
mattpap: Currently `glyphs/legends` contains one example. The note is wrong, saying we don't support absolute coordinates, so this needs some improvements in general.

bryevdv: It would be best to try and make all the necessary docs/examples support updates in this PR.

mattpap:
I made bokeh's deprecation warnings always visible. It's not perfect with decorators, but that's something to improve later (and clean up this mess with every deprecated functionality using its own deprecation method).

bryevdv:
Looks great, merging

Issue #3236, opened 2015-12-04
CI Build breakdown


bryevdv:
Some suggestions:
* JS, docs, and flake tests only need to run once, suggest one py3 test for all three
* Consider python + integration tests could be merged into one build, to share "install" cost. 
* Longest tests (both "examples" builds) should be run first together, so that their start times are not unnecessarily staggered. The current order adds ~5-10 minutes to every full run.

bryevdv:
I'm also wondering about the value of caching. My informal observation is that it seems to be detecting changes and packing/uploading almost all the time, and that takes quite some time. More or less than without, I'm not sure.

damianavila:
> I'm also wondering about the value of caching. My informal observation is that it seems to be detecting changes and packing/uploading almost all the time, and that takes quite some time. More or less than without, I'm not sure.

That's a good observation... I am thinking on that at well... we should time this somehow... I would be open to remove it to check if performs better...

PR #3265, opened 2015-12-07
try some other build matrix configurations

bryevdv:
issues: ref #3236 

ping @damianavila 

This results in the following build matrix:

screen shot 2015-12-07 at 2 50 50 pm

This a little different from the changes suggested in the referenced issue. Basically:

* Python+JS unit tests everywhere, but together to save on install times. (The tests suites themselves are both very fast compared to install time)
* Integration tests only once, on Py 3.4 
* flake_docs only once, on py2.7 (since that is where they are deployed from)
* examples tests everywhere (except disabled for py3.5 temporarily for now)

***NOTE***: the 3.5 examples tests are unfortunately disabled (for now) because we simply cannot afford three "long" tests with only five workers, with the volume of PRs we have. It results in an immediate logjam.  Looking into adding resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,23 +23,28 @@ env:
         - secure: "v9VkMz6cjFBZhlf6iiXSr0o2Eo/HKMMOseAxE6H+sr1vbPg19edGU+7+aAQrn09lE4oKU5JwZSzQ2KalJgrNTGrRrGyEtWGAcuS6oP17zg9LsutLGptNwyHBnCvB5aUPW7JSNs+kor7CmvZ8sVnUh810YSREjt2ZsjEG3ysZjVc="
         - secure: "OI9ErePcqSewA8ZheZq5YtNyeLA8UM1UGskcMEAmPMf9AisYasJaPoY8NeqLGDyMcg7TcXbFQBZv7jPEngZ3RZMsDajfC3p63WOBDS7PfurKTVWTXZ28Z6vzwa3riFIca3D0eVP6bt8KZrYMo7HvWQG3vzmTSxRjx4e2wHxciKo="
     matrix:
-      - GROUP=js_tests
-      - GROUP=python_tests
-      - GROUP=integration_tests
-      - GROUP=examples_flake_docs
+      - GROUP=unit
+      - GROUP=integration
+      - GROUP=examples
+      - GROUP=flake_docs
 
 matrix:
     exclude:
+        # integration tests only py 3.4
         - python: 2.7
-          env: GROUP=js_tests
-        - python: 2.7
-          env: GROUP=integration_tests
+          env: GROUP=integration
         - python: 3.5
-          env: GROUP=js_tests
+          env: GROUP=integration
+        #
+        # flake_docs only py 2.7
+        - python: 3.7
+          env: GROUP=flake_docs
         - python: 3.5
-          env: GROUP=integration_tests
+          env: GROUP=flake_docs
+        #
+        # TEMPORARILY DISABLE EXAMPLES for py 3.5
         - python: 3.5
-          env: examples_flake_docs
+          env: GROUP=examples
 
 branches:
   only:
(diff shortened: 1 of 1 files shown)


bryevdv:
@bokeh/core please review/weigh in on this change. I think this is a fairly reasonable breakdown for now, that will increase test throughput. The run that above would have finished in ~35 minutes if there had not been a random timeout failure. I think as soon as two of the five workers are available on a given  build thats a fairly stable approximate time. We are currently seeing elapsed test times in excess of one hour sometimes.

birdsarah:
Thoughts:
- the Integration tests are fast (5 min) compared to the examples tests (30 min), at least at the moment, if we had lots of them I can imagine this not being true
- I see examples and integration tests as the same (I also have some things in the works that I think would run them under one umbrella) - or at least serving the same purpose
My medium-term suggestion to get our build time better would be to move all integration tests out into a seperate testing cadence (that may or may not be in a separate repo) and have them run once or twice a day, allowing us to quickly pick up broken changes. Especially as I would like to see integration (inc. examples) run against a suite of browsers, which will slow things down further.
- If that's too extreme, I would suggest strategically defining a core set of integration tests that can quickly identify broken things. I think we could probably do this in 20 selenium tests or less which would only take 5 minutes or so to run. These run in main bokeh repo, and then the full set of bokeh/integration tests.

If we go with the above then it would look something like:
* 2.7 - unit, integration (core), flake_docs
* 3.4 - unit, integration (core)
* 3.5 - unit, integration (core)

And then all examples / and integration running on a schedule against master and against multiple browsers. (I'm not sure how to use travisci to run a scheduled set of tests).

In the short-term. I would also be fine with only running the examples on one version of python.

----
More background on my opinion here: I don't love the examples tests. For me, tests should have an assert statement and then fail and tell you something useful. When the examples tests fail we have to go and debug why that failure happened. Running through a whole bunch of examples, and checking they all work as expected, is:
a) moving out of testing into QA ( acknowledging this is a very grey and fuzzy spectrum )
b) The tests we currently do on examples don't test that the examples work *as expected*. They take a screenshot. This is not the same thing. And, I think they cause us a fair amount of grief while still having quite a lot of things slip through. Examples that spring to mind: when my patches PR made all the glyphs disappear on tap; notebook tests just screenshotting a mathjax error and passing instead of running.

---

OK The above is probably not what you wanted in terms of feedback.....

In the short-term, I think my only change from what you've done would be to include the integration tests in the examples run. That'll only add a couple of minutes to the very long examples runs and will get us one more build shorter than what you've got.  In addition, you could do integration tests on py3.5 while examples are getting fixed. So that's:

* py2.7 - unit | integration + examples | flake_docs (3 builds)
* py3.4 - unit | integration + examples (2 builds)
* py3.5 - unit | integration (2 builds)

bryevdv:
Well that leaves the same amount of tests (7) as before. :) My view on the current integration test is that running them across python versions is completely redundant and wasteful. They only test BokehJS capabilities, in-browser. What do they test that might vary across python versions?

----

(Will respond to other discussion later)

birdsarah:
> They only test BokehJS capabilities, in-browser. What do they test that might vary across python versions?

That's not quite true. They test the end-to-end of building a plot in python and then making sure it's doing what we expect in the browser. 

Can I ask the same question about the examples: what do they test that might vary across python versions?

bryevdv:
> Can I ask the same question about the examples: what do they test that might vary across python versions?

The testing of examples serves multiple purposes. They are user facing, so one of the reasons to test the examples is to make sure the *examples work* as opposed to making sure that Bokeh works, per se. There have definitely been accidental breakages in the past when someone added something that was not py2/py3 compatible. That's why they are tested across versions. 

Regarding their utility for *testing Bokeh*, I think what we have currently is overkill. I'd personally like to narrow the image diff tests to a small set of basic and reasonably comprehensive targeted tests that can actually reliable be passed/failed on the diff alone. I'd like for the full examples test run to be nightly and generate a full report, because inspecting that has been useful more times than I can count.

bryevdv:
BTW When I say "examples tests" I mean testing that basically function (run without error) and generate a report. If there's a better/more unified way to do that with selenium, I have no strong opinions about the implementation.

bryevdv:
Anyway this is not the place to discuss all that, this is just to clear the immediate logjam and be able to move forward. Unless there is some objections I will merge this very early tomorrow morning

damianavila:
>Anyway this is not the place to discuss all that, this is just to clear the immediate logjam and be able to move forward. Unless there is some objections I will merge this very early tomorrow morning

I am OK to merge this one as is as a temporary solution... there are better ways to set up the matrix without having to change the code to much, just specifying some things at the `include` level, but I can do that later by myself... this seems OK to me for now...

bryevdv:
@damianavila OK great, I tried to actually simplify the code a little bit, but if it can be simplified even more that's great.

PR #3268, opened 2015-12-07
Remove pandas deprecation

fpliger:
Removes pandas deprecation when using some charts due to `df.sort` rename to `df. sort_values`


bryevdv:
LGTM

PR #3269, opened 2015-12-08
Production Bokeh

quasiben:
- [x] example supervisor setup
- [x] example nginx setup


mattpap:
@quasiben, can we get a word of explanation where this comes from and why this should be in this repository? It seems to me a misdirected PR, but maybe I miss some discussion on other channels.

bryevdv:
This is to support documentation providing guidance for deploying the new bokeh server in common scenarios.

bryevdv:
@quasiben can you do one last thing? Can you close this and re-open as branch off the main repo? Then I can take over and move things around a bit and add the Sphinx documentation in the same PR.

quasiben:
closing in favor of #3273

PR #3273, opened 2015-12-08
Production Bokeh

quasiben:
Requires Nginx > 1.3


mattpap:
I think this doesn't have to be at the top-level. Maybe lets put it in `scripts/`?

quasiben:
@mattpap everything is now in `scripts`

mattpap:
> everything is now in scripts

OK, thought this was more a question to @bryevdv about the structure of bokeh's repository.

bryevdv:
@mattpap yah absolutely I planned to move it around that's why I had ben make a new PR off the main repo so I could take over

scripts/prod_deploy/nginx.conf
```diff
@@ -0,0 +1,18 @@
+server {
+    listen 80;
+    server_name 54.174.175.215;
```
bkreider: To catch any hostnames or IP addresses this should be: 
```
server {
    listen       80  default_server;
    server_name  _;
```
http://nginx.org/en/docs/http/server_names.html

bkreider:
I haven't ever tuned Nginx for reverse proxying websockets but we should test and look at buffering:
https://www.nginx.com/resources/admin-guide/reverse-proxy/

>If buffering is disabled, the response is sent to the client synchronously while it is receiving it from the proxied server. This behavior may be desirable for fast interactive clients that need to start receiving the response as soon as possible.

I did a quick google search which ended up having a lot of noise but i found this (for some servlet app):
http://guac-dev.org/doc/gug/proxying-guacamole.html

> Do not forget to specify "proxy_buffering off".

>Most proxies, including Nginx, will buffer all data sent over the connection, waiting until the connection is closed before sending that data to the client. As Guacamole's HTTP tunnel relies on streaming data to the client over an open connection, excessive buffering will effectively block Guacamole connections, rendering Guacamole useless.
```

File: scripts/prod_deploy/nginx.conf (hunk 1 above)
Comment (bkreider) [inline scripts/prod_deploy/nginx.conf]: Oh, and there is a gotcha when setting the Host header, you need this if you ever don't run on port 80 or 443.  We got bitten by this in Wakari.  I don't know why the examples fail to mention this.  Most people must always run on 80 or 443.

```
```
If you don't do that and say nginx is running on port 81, your app will generate urls like this:  `http://<hostname>/myapp` instead of `http://<hostname>:<port>/myapp.

birdsarah:
I'm not sure this should be in bokeh's core repo. 

There will be multiple ways to deploy bokeh depending on your needs. This is just one of them. It certainly serves as great reference. For a start, it's debian-centric.

Some thoughts:
* If we are deploying bokeh server for the bokeh docs site then we could have a seperate repo with all the config for that site in it and point to it as a real live example.
* We could put this into examples.

bryevdv:
I'm planning to make things more general, i.e. "here's an nginx stanza you can set up for your system" not "here's an nginx stanza here's how to set it up on one specific machine in detail" mostly this will all go in the users guide

birdsarah:
For my own knowledge only: I'm assuming that Apache setup will also be possible, we just won't be providing a reference?

bryevdv:
I would like to have at least some minimal Apache info for the release.

bryevdv:
Will write up the `demo.bokehplots.com` deployment, closing this PR

quasiben:
Work has now moved to the following repo: https://github.com/bokeh/demo.bokehplots.com/

PR #3278, opened 2015-12-08
Keep the output_file the same name as the example file

birdsarah:


bryevdv:
Travis failure was spurious, other examples test pass and change is trivial by inspection, so I will go ahead and merge.

PR #3280, opened 2015-12-09
Feature/add image diff testing to selenium

birdsarah:


birdsarah:
Here's the latest test report: https://s3.amazonaws.com/bokeh-travis/95986164/report.html#

The test is registering as failing, but you can see it hasn't really, the difference, i believe is in resolution or some other small difference caused by running a slightly different browser on a slightly different platform.

I see three options:
* when calculating whether the screenshots are the same allow for a small amount of variance. This is probably the easiest solution but will require some tweaking over time, to get the right balance between low-noise and not missing actual important differences.
* do the work to use browserstack to get the screenshots (which we'd planned to do anyway) and see if that eliminates the difference 
* figure out how the source of the difference between travisci ff and the screenshot that i took on my computer and fix is - this seems the least fruitful as we want different people on different machines to be able to take screenshots and travis could change their settings too.

ping @bryevdv @mattpap  @mistakevin

mattpap:
> when calculating whether the screenshots are the same allow for a small amount of variance

You could use `pdiff`, as we do in the main diff testing machinery.

birdsarah:
> You could use `pdiff`, as we do in the main diff testing machinery.

2 things:

1) I am not sure that using pdiff will change the problem. The difference between these and the examples tests is that these tests add the base_screenshot to the repo. The examples tests, I believe, compare a base that was previously generated by travis. As a result the small difference is likely to be gone on the examples tests

2) These are tests that I expect people to run locally and it seems to me that `pdiff` has a higher barrier for people to run locally. I may be wrong on this, feel free to correct me.

birdsarah:
I will try none-the-less with pdiff on a seperate branch just to check my understanding is correct.

mattpap:
`pdiff` is conda installable, so it shouldn't be a problem.

birdsarah:
who built our version - can they build os-x and windows versions?

mattpap:
@birdsarah, eh, obviously only linux-64 packages (that's not unexpected). We should use anaconda's builders for this purpose. We shouldn't build packages locally, but use and tests our infrastructure. Maybe @damianavila could chime in with some advice.

damianavila:
>Maybe @damianavila could chime in with some advice.

In the short term you can probably conda convert `pdiff`` to other platforms... or build it from the other platforms... @mattpap, the repo lives here:https://github.com/bokeh/pdiff, but I don't see any conda recipe... do you know where is lives?

mattpap:
> he repo lives here:https://github.com/bokeh/pdiff

This is actually my port of pdiff to scala ;)

mattpap:
> probably `conda convert pdiff`

I would say this works only for pure python packages and pdiff is compiled. That's why I'm suggesting using anaconda.org, because it should allow to compile it on different platforms.

mattpap:
btw. http://pdiff.sourceforge.net/

damianavila:
> I would say this works only for pure python packages and pdiff is compiled. That's why I'm suggesting using anaconda.org, because it should allow to compile it on different platforms.

That's correct, I think... thanks for the clarification.

bryevdv:
Since we are pausing this until after 0.11 I will close for now to clean up PR list

PR #3283, opened 2015-12-09
fix broken docstring automation

bryevdv:


bryevdv:
Observed enough Travis CI to see broken import fixed, tests able to run, merging now.

PR #3312, opened 2015-12-10
Add data labels to charts data sources to support hover tools, and fix charts defaults

### LINKED ISSUE #2965: custom Bar chart hover tooltips
version: 0.10.0

It is not possible to display custom values as tooltip on mouse hover.

    from bokeh.charts import Bar, show

    data = {'ID': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'Total': [0.0, 20.0, 0.0, 6.0, 8.0, 0.0, 8.0, 5.0, 8.0, 1.0, 2.0]}

    p = Bar(data, label='ID', values='Total', width=400, tools='hover')

    hover = p.select(dict(type=HoverTool))
    hover.tooltips = [('Value of ID',' $x'),('Value of Total',' @Total')]

    show(p)

image

### LINKED ISSUE #3296: Charts defaults are broken
Specially colors are broken after new defaults PR. Bar chart, for instance, has white set as default color

rothnic:
The primary purpose of this was to provide relevant fields associated with non-aggregated or aggregated data.

Each source to glyph renderers will now have the columns provided to the chart available, plus a column called "chart_index". 

So `Chart(data, x='displ', y='mpg', color='cyl', marker='cyl', line_color='origin')` would take all the unique column attributes  ('cyl', 'cyl', 'origin') => ('cyl', 'origin'), and add the values to each data source. So you would end up with 3 extra columns besides just the data that is required for the glyph. You'd have something like this:

```python
{'chart_index': [{'cyl': 8, 'origin': 1}], 'cyl': [8], 'origin': [1], ... <other values required for glyph>}
``` 

Previously, the chart was internally using the name that pandas.DataFrame.groupby produces, `(name, data_group)`, where name is either a single value if grouped by one column, or a tuple of values if grouped by more than one, which would be `(8, 1)` for this example. 

Instead of using the tuple of unique values for that group of data, what you see above as 'chart_index' is carried around as representing the label of that subset of data. This makes it easier to reference both the column and unique values internally for charts. One note is that if you do not pass any column names into charts, the chart_index will be 'None' as a special value used for that case, meaning that you have not grouped the data, so there is no index to label the data by.

Examples were modified on various charts to include the hover tools, and this was made easier by offering a hover kwarg to charts. This kwarg can accept what is passed to tooltip on HoverTool without the user having to import and set the values. It also can take more simple inputs and try to create the tooltip value based on the chart inputs.

All of the following are equivalent, but you could show fewer values as well through either the second or third example:

```python
Chart(...., hover=True)

Chart(..., hover=['cyl', 'origin'])

Chart(..., hover=[('cyl', '@cyl'), ('origin', '@origin')])
```

This work resolves #2965.

Also, while completing this work I came across issues from the recent properties rework. Specifically, the default coloring of some glyphs wasn't correctly being used. 

This work resolves #3296.


fpliger:
quick comment: I think you need to change:

This work resolves issue #2965.

to

Issues: fixes: #2965

to make github close the issue automagically..

rothnic:
They say they are supposed to support anything [here](https://help.github.com/articles/closing-issues-via-commit-messages/), but i think it can be sensitive sometimes. Think I should not of put "issue" there.

fpliger:
Ah, cool.. Good to see the full list! Thanks :)

bokeh/charts/builder.py
```diff
@@ -216,6 +216,10 @@ class Builder(HasProps):
         is (Column, Ascending).
         """)
 
+    source = Instance(ColumnDataSource)
+
+    hover = Either(List(Tuple(String, String)), List(String), Bool, default=None)
```
fpliger: I'm afraid calling it hover may create confusion with the `HoverTool` itself. What about calling it `tooltips` or `hover_tooltips` to not get confused?

fpliger: Actually `hover_spec` is appropriate too since you already use it elsewhere..

bokeh/charts/utils.py
```diff
@@ -643,9 +643,22 @@ def add_wedge_spacing(df, spacing):
...
+        if isinstance(hover_spec, bool):
+            tooltips = [(col, '@' + col) for col in chart_cols]
+        elif isinstance(hover_spec[0], tuple):
+            tooltips = hover_spec
+        else:
+            tooltips = [(col, '@' + col) for col in hover_spec]
+
+        chart.add_tools(HoverTool(tooltips=tooltips))
+
+    elif use_hover:
```
fpliger: I'm not sure I get the reason for use_hover.... what is it really doing?

rothnic: I should probably split this in two functions, or refactor it a bit. I had an existing function to reduce duplication between charts that used hover, but it was more tailored towards charts with aggregation.

bokeh/charts/builder.py
rothnic: Ok, I'll think about it. Initial thought is it is broader than the kwarg tooltip input to hover, but i like tooltips as the name either way.

bokeh/charts/utils.py
fpliger: yeap.. a split is probably easier to read and document. tnx :)

bokeh/charts/attributes.py
```diff
@@ -107,6 +110,9 @@ def __init__(self, columns=None, df=None, iterable=None, default=None,
 
         super(AttrSpec, self).__init__(**properties)
 
+        if self.default is None and self.iterable is not None:
+            self.default = next(copy(iter(self.iterable)))
+
```
fpliger: You don't need to copy here, do you? This breaks in 2.7.. if you remove this tests should pass..

fpliger:
LGTM after fixing the broken tests and examples (do not work on py2.7)

As a minor comment I'd add "HOVER" or something like that to the title of the examples with hover tool. Those need attention.

rothnic:
@fpliger, I am doing something similar elsewhere in the attribute spec so I make sure I have an iterable and make sure I'm not consuming it. I could probably refactor and simplify to handle the default differently, but would probably want to do that separate from this PR. I think the issue was the order of operations, but it still is strange that I can't reproduce that exact issue locally. If this doesn't pass, I'll see what I can do to reproduce the issue.

rothnic:
The ipython notebook heatmap stalled out in the test. I ran it locally and most of it worked correctly, but some of it gave an error about underscore not being available. I merged master in to get things up to date and to force the tests to run again.

rothnic:
@fpliger: I don't think that the remaining failed tests are related to charts.

fpliger:
@rothnic travis is happy now. Will merge it soon if no one objects...

PR #3327, opened 2015-12-11
fix erroneous dispatch method

bryevdv:


mattpap:
It looks like this code was never run before. What triggered it this time? Maybe a test based on this could have been added?

bryevdv:
I just happened to notice it. I want to automate this somehow and add a test for that.

PR #3330, opened 2015-12-14
Periodic forcing and server examples cleanup

### LINKED ISSUE #3335: line_animate_widget.py is broken
Testing example plotting/server/line_animate_widget.py on master and the buttons don't cause stop and start to work.

I imagine we could actually do this really nicely now by adding and removing periodic callback.

bryevdv:
issues: fixes  #3335 

This PR adds a `bokeh.utils.forcing` module that has decorators very useful with server periodic callbacks. This PR also cleans up all the server examples to use those forcing functions where appropriate and in general make the example structure, imports, etc. consistent.

TODO:
* [x] add module docs to sphinx
* [x] add tests for module
* [x] apply forcing function everywhere else appropriate 

---

some  currently problematic examples are
* selection histogram: sometimes does not update, "ping pong"?
* glyphs.py `width_units` etc seemingly ignored
* timeout stocks, needs to be simplified, timeout callback issue?


bryevdv:
@bokeh/dev would appreciate comments on naming, API, etc. We definitely need something like this, but perhaps it could use tweaks.

bokeh/util/tests/test_forcing.py
```diff
@@ -0,0 +1,61 @@
...
+    def foo(val):
+        results.append(val)
+    return foo
+
+w = 0.3
+A = 3
+phi = 0.1
+offset = 2
+
+def test_sine():
```
birdsarah: can you be more explicit in your test names e.g. test_forcing_sine_function_returns_x_when_y.

I find it helps when in the future you're not just looking at a test failure of `test_repeat`.

birdsarah:
From testing so far:
* animated has a bunch of jitter on it, especially when it first loads, I am not seeing this in master
* line_animate_widget is very jumpy - not smooth at all

bokeh/util/testing.py
```diff
@@ -1,59 +1,59 @@
-""" Functions to help with testing Bokeh and reporting issues.
-"""
+''' Functions to help with testing Bokeh and reporting issues.
+'''
```
mattpap: I'm curious why changing this?

bryevdv: Strong preference.

bryevdv:
@birdsarah 

> animated has a bunch of jitter on it, especially when it first loads, I am not seeing this in master

I've seen the initial jitter on master since forever, I think it is a general issue. I am very surprised you haven't seen it elsewhere?

> line_animate_widget is very jumpy - not smooth at all

Also weird, it's very smooth on my machine. I also think that's probably not related to this PR, tho.

bryevdv:
Here's what it looks like on my machine with a 20ms interval and M=8 (it actually looks better, GIF encoding limited to 20fps)
ezgif com-video-to-gif-2

bokeh/util/tests/test_forcing.py
bryevdv: so my convention is to have `test_<modulename>` and so adding `forcing` to `test_sine` inside `test_forcing` seems redundant to me. It's testing `forcing.sine` by virtue of the fact that it is in `test_forcing.py`

For the other, I'd prefer `_sequence_values`, which is what I will add.

birdsarah:
i'll take another look. my experience with the server has been historical jitters, but not since new server went it.

birdsarah:
Here is a movie of my problems with periodic_callback (https://www.dropbox.com/s/6xiqsnnyvw6bsjk/jittery%20on%20periodic%20callback%20branch%20-%20small.mov?dl=0). In it I switch back and forth to master and you can see the difference (in particular at about 1m 15 in you see it really freaking out on periodic_callback)

You're not experiencing this problem with this PR, is there anything I can do to isolate it. Do I need to clean out a cache somewhere when switching branches?

birdsarah:
After I merged latest master I am not seeing the jitters.....

bryevdv:
Any reason not to merge? Does the API seem ok?

birdsarah:
to be honest there's a lot of different things going on in this PR and a lot of changes that don't change functionality. So I feel like there's a good change I've missed something as I've skipped through things that look like just formatting changes.

i certainly like `repeat`, `sine`, `cosine` etc. - what its done to geojson_points is great!

`forcing` feels wrong though. For me --force is usually something I do when it's got potentially bad consequences but I know what I'm doing.

To me the idiom is like paramaterizing test functions: http://pytest.org/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

(will finish this thought just posting for now)

birdsarah:
So I don't love `bokeh.utils.parametrize` because it's a mouthful and easy to misspell. 

But that's my best suggestion at the moment.

Also, I'm assuming these could be stacked - just checking my understanding 

````python
@bounce(['blue', 'pink', 'red'])
@repeat([0, 1])
def update(i, color):
    # alternate between original/updated and inject color into the string
    source.geojson = [original, updated][i] % color
````

**Comment by bryevdv** [comment]
@birdsarah FWIW this is what made me associate it with forcing functions:

https://en.wikipedia.org/wiki/Forcing_function_(differential_equations)

But that's a pretty narrow usage to draw on. I'm certainly open to new suggestions. `driver` came to mind but that has too many other connotations. 

I'm not sure if they would stack as written, I think they probably would. Slightly related I was also considering a "by keyword" option:
```
@bounce(color=['blue', 'pink', 'red'])
@repeat(i=[0, 1])
def update(i, color):
    # alternate between original/updated and inject color into the string
    source.geojson = [original, updated][i] % color
```

**Comment by bryevdv** [comment]
I'm also not sure about `bokeh.utils` I feel like at least up until now most of those are "internal" or "for developers"

**Comment by birdsarah** [comment]
Thanks for the reference - we all have our different heritages coloring things :D 

I'm not sure what the right name is. Happy to defer. Just my two cents about the negative implications of `force`

Personally I'm not bothered about the keyword option.

I completely agree about `bokeh.utils`. Whilst these are generic functions, their only use, for Bokeh, is the server - what about `bokeh.server.utils`

**Comment by birdsarah** [comment]
naming thoughts: `generators`, `series`, `sequences`, `spawn`, `queue`, `cue`, `feed`

File: bokeh/util/forcing.py [hunk 3]
```diff
@@ -0,0 +1,145 @@
+''' Provide a set of decorators useful for repeatedly driving an
+a function parameter.
```
almarklein: "... driving an a function ..."

bryevdv:
@birdsarah none of those really convey what I have in my mind, which is to motivate something, in the literal sense of "cause to move".  The only other word that's come to mind is `driving`. I'm fine with that, but then the question is where does it go

bryevdv:
@birdsarah `bokeh.server.utils.driving.reapeat` is way too many dots. Also these could  be useful in notebooks without the server. So I propose `bokeh.driving` or `bokeh.util.driving` I'd like to merge this tomorrow, so input appreciated @bokeh/core

bryevdv:
Ok new idea! I've wanted to move some things into a `bokeh.core` package for a while to tidy up the top level. So I intend to make `bokeh.core.driving` (later also `easing` e.g.) unless there is objection. Ping @birdsarah

bryevdv:
I've not heard back any feedback from anyone, so this is the version (`bokeh.driving`) that I intend to merge (very soon)

birdsarah:
sorry for the delayed response. `bokeh.driving` or `bokeh.core.driving` are both great.

bokeh/driving.py
```diff
@@ -0,0 +1,146 @@
...
+        @bounce([0, 1, 2])
+        def update(i):
+            print(i)
+
+    If this function is repeatedly called, it will print the following
+    sequence on standard out:
+
+    .. code-block:: none
+
+        0 1 2 2 1 1 0 0 1 2 2 1 ...
```
pzwang: Shouldn't this be:

    0 1 2 2 1 0 0 1 2 2 1 0

i.e. why does the sequence `1 1` ever appear?

pzwang:
I'm -1 on "force" and more supporting of the "driver" concept.  Somehow "driving" sounds more confusing to me.  But I think that ultimately these are just "data animators", so maybe `bokeh.animate.*` would be the right place for them?

I like the stacking and keyword values.

bryevdv:
Happy to add keyword values in a follow on PR. The stacking is potentially useful, but also a possibly confusing: all the functions advance "at once", you don't get a product space of values or anything like that. Could certainly add some notes in docs about that. 

I don't think these are quite as useful on the JS side as the "easing" functions for transiting between state, but I could imagine they might have some use there. I agree `bokeh.server` is not the right place. 

I suppose `bokeh.animate.*` could work the only slightly odd thing is that if/when there are JS counterparts there will have to be models in `bokeh.models` as well.

bryevdv:
@pzwang I think I like `bokeh.animate.drivers` or `bokeh.animate.driving` I am going to merge this PR as-is to unblock the other examples PRs, and then I will make another PR to change the location and add the kwargs option too. If you have a preference between `bokeh.animate.drivers` and `bokeh.animate.driving` please let me know otherwise I will just pick one.

PR #3334, opened 2015-12-14
hotfix to only render toolbar once

bryevdv:
Fixes a bug that caused the toolbar to be unresponsive on streaming updates. 

@birdsarah @mattpap @brendancol any comments welcome, I am sure this is a simplistic fix but it does seem to be an improvement.

### DIFF
--- a/bokehjs/src/coffee/common/tool_manager.coffee
+++ b/bokehjs/src/coffee/common/tool_manager.coffee
@@ -16,9 +16,12 @@ class ToolManagerView extends Backbone.View
   initialize: (options) ->
     super(options)
     @listenTo(@model, 'change', @render)
-    @render()
+    @have_rendered = false
 
   render: () ->
+    if @have_rendered
+      return
+    @have_rendered = true
     @$el.html(@template(@model.attributes))
     @$el.addClass("bk-sidebar")
     @$el.addClass("bk-toolbar-active")


bryevdv:
@damianavila this would be good to get in the dev build before the webinar

birdsarah:
I've tested locally on a few normal plots and on the server plots. The fix works as expected. It's so cool being able zoom in and pan on the animated plot.

bokehjs/src/coffee/common/tool_manager.coffee
```diff
@@ -16,9 +16,12 @@ class ToolManagerView extends Backbone.View
   initialize: (options) ->
     super(options)
     @listenTo(@model, 'change', @render)
```
mattpap: This PR makes this line pointless.

mattpap:
I think this PR doesn't fix the underlying rendering issue. Needs more thought (I don't mind doing this myself). -1 for now.

bryevdv:
@mattpap OK, the dev build for the webinar is this afternoon...

mattpap:
> OK, the dev build for the webinar is this afternoon...

I see, so lets merge this for now.

mattpap:
Obviously this will need a followup.

bryevdv:
Agreed needs followup, the only reason I'm OK for a quick fix is that to my knowledge no one has ever asked or demonstrated updating the toolbar tools after a plot is drawn.

PR #3354, opened 2015-12-16
add chdoig's scikit-learn clustering demo

bryevdv:
Not for this PR, but some "nice to haves" that could improve this app if there is time:

* spinner when computing/busy
* deactivate widgets when computing/busy
* auto layout (ping @havocp)
* visual style tweaks (ping @SaraSchnadt )


bryevdv:
Skipping tests since this is adding a single example-only.

PR #3364, opened 2015-12-17
Tile wrap around

brendancol:
Mercator tile sources now have a `wrap_around` flag. 

`wrap_around = True` (default):
screen shot 2015-12-16 at 10 30 35 pm

`wrap_around = False`:
 screen shot 2015-12-16 at 10 34 13 pm

The addition of wrapping to other layers (e.g. `DynamicImageLayer`, `GeoJSONLayer`) will come in subsequent PRs. The colored layer seen on the US is a `DynamicImageLayer`. 

This example is available at `examples/plotting/file/dynamic_map.py`


bokeh/models/tiles.py
```diff
@@ -58,6 +58,10 @@ class MercatorTileSource(TileSource):
     resolution (plot_units / pixels) of minimum zoom level of tileset projection. None to auto-compute.
     """)
 
+    wrap_around_180 = Bool(default=True, help="""
```
bryevdv: Maybe just `wrap_around`? With the scale mentioned in the help? I'm not aware of any other property names with digits in them. Obviously it works but it's a little incongruous to me.

brendancol: yes i agree on that.  change made and pushed

birdsarah:
what happens to axis labesl when you use wrap around?

brendancol:
The axises are `Range1d` so they will display the actual coordinates (i.e. not 'wrapped').

bokeh/models/tiles.py
```diff
@@ -58,6 +58,10 @@ class MercatorTileSource(TileSource):
     resolution (plot_units / pixels) of minimum zoom level of tileset projection. None to auto-compute.
     """)
 
+    wrap_around = Bool(default=True, help="""
+    flag to indicate whether to load tiles outside of world bounds.
```
birdsarah: Two comments on this docstring:
1) You're not outside of world bounds :D you've just looped round
2) Can you add a note about the limitation regarding the axes, so at least its documented somewhere that displaying the coordinates is not expected to work.

brendancol: Good call. make quick update

birdsarah:
Tested this locally by updating `plotting/file/dynamic_map.py` as follows

````python
tile_source = STAMEN_TONER
tile_source.wrap_around = False
p.add_tile(tile_source)
````

And it worked as expected.

bryevdv:
I can believe we might want to expand or change this some later, but as all the changes are self-contained I will go ahead and merge.

PR #3366, opened 2015-12-17
Add Interface for Chord Graphs

### LINKED ISSUE #844: add chord chart

merqurio:
issues: closes #844

Hello Bokeh team :grin:, finally found some time to do this!

I want to create a PR to add Chord Graphs interface to Bokeh. (issue #844)
chord graph example

I have already shown that creating chord graphs with Bokeh is [possible](https://www.continuum.io/blog/developer-blog/redrawing-brain-bokeh), now I want to make it available to everyone with an easy to use interface. My hope is also that making this PR I can get some help from some of the __Bokeh gurus__ to resolve some of the features I've tried to do but I got lost.

#### TODO:
- [x] Add the code to generate the chord with the chart interface
- [x] Show labels in the chart
- [x] Expose customisations for the Chart (labels etc..) 

## Future commits
- Add hoverTool to arcs
- Add tapTool to arcs
- WebGL rendering for arcs
- WebGL rendering for beziers

I'll be working on this actively and I take the compromise of maintining it but any help is **very** welcomed :grimacing:, not a super expert, yet !

Thanks !!

### DIFF
--- a/bokeh/charts/builders/chord_builder.py
+++ b/bokeh/charts/builders/chord_builder.py
@@ -1,0 +1,294 @@
+"""This is the Bokeh charts interface. It gives you a high level API
+to build complex plot is a simple way.
+
+This is the Chord class which lets you build your Chord charts
+just passing the arguments to the Chart class and calling the proper
+functions.
+"""
+# -----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2016, Continuum Analytics, Inc. All rights reserved.
+#
+# Powered by the Bokeh Development Team.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+
+# -----------------------------------------------------------------------------
+# Imports
+# -----------------------------------------------------------------------------
+from __future__ import absolute_import
+
+import numpy as np
+import pandas as pd
+from math import cos, sin, pi
+from bokeh.charts.properties import Dimension
+from bokeh.charts.builder import create_and_build, Builder
+from bokeh.charts.attributes import MarkerAttr, ColorAttr, CatAttr
+from bokeh.charts.utils import color_in_equal_space, help
+from bokeh.models import Range1d
+from bokeh.models.glyphs import Arc, Bezier, Text
+from bokeh.models.renderers import GlyphRenderer
+from bokeh.models.sources import ColumnDataSource
+from bokeh.core.properties import Instance, Bool, String, Array, Float, Any, Seq, Either, Int
+
+
+# -----------------------------------------------------------------------------
+# Classes and functions
+# -----------------------------------------------------------------------------
+
+class Area:
+    """ It represents an arc area. It will create a list of available points through the arc representing that area and
+    then those points will be used as start and end for the beziers lines.
+    """
+
+    def __init__(self, n_conn, start_point, end_point):
+        # Number of connections in that arc area
+        self.n_conn = n_conn
+        # The start point of the arc representing the area
+        self.start_point = start_point
+        self.end_point = end_point
+        # Equally spaced points between start point and end point
+        free_points_angles = np.linspace(start_point, end_point, n_conn)
+        # A list of available X,Y in the chart to consume by each bezier's start and end point
+        self.free_points = [[cos(angle), sin(angle)] for angle in free_points_angles]
+        assert self.n_conn == len(self.free_points)
+
+
+class ChordBuilder(Builder):
+    """ This is the Chord builder and it is in charge of plotting
+    Chord graphs in an easy and intuitive way.
+
+    Essentially, we provide a way to ingest the data, make the proper
+    calculations and push the references into a source object.
+    We additionally make calculations for the ranges. And finally add
+    the needed glyphs (markers) taking the references from the source.
+
+    """
+
+    default_attributes = {'color': ColorAttr(),
+                          'marker': MarkerAttr(),
+                          'stack': CatAttr()}
+
+    dimensions = ['values']
+
+    values = Dimension('values')
+
+    arcs_data = Instance(ColumnDataSource)
+    text_data = Instance(ColumnDataSource)
+    connection_data = Instance(ColumnDataSource)
+
+    origin = String()
+    destination = String()
+    value = Any()
+    square_matrix = Bool()
+    label = Seq(Any())
+    matrix = Array(Array(Either(Float(), Int())))
+
+    def set_ranges(self):
+        rng = 1.1 if not self.label else 1.8
+        self.x_range = Range1d(-rng, rng)
+        self.y_range = Range1d(-rng, rng)
+
+    def setup(self):
+
+        # Process only if not a square_matrix
+        if not self.square_matrix:
+            source = self.values._data[self.origin]
+            target = self.values._data[self.destination]
+            union = source.append(target).unique()
+            N = union.shape[0]
+            m = pd.DataFrame(np.zeros((N, N)), columns=union, index=union)
+
+            if not self.label:
+                self.label = list(union)
+
+            if self.value is None:
+                for _, row in self.values._data.iterrows():
+                    m[row[self.origin]][row[self.destination]] += 1
+                self.matrix = m.get_values()
+
+            if self.value is not None:
+
+                if isinstance(self.value, int) or isinstance(self.value, float):
+                    for _, row in self.values._data.iterrows():
+                        m[row[self.origin]][row[self.destination]] = self.value
+                    self.matrix = m.get_values()
+
+                elif isinstance(self.value, str):
+                    for _, row in self.values._data.iterrows():
+                        m[row[self.origin]][row[self.destination]] = row[self.value]
+                    self.matrix = m.get_values().T
+        else:
+            # It's already a square matrix
+            self.matrix = self._data.df.get_values()
+
+        if self.label:
+            assert len(self.label) == self.matrix.shape[0]
+
+    def process_data(self):
+
+        weights_of_areas = (self.matrix.sum(axis=0) + self.matrix.sum(axis=1)) - self.matrix.diagonal()
+        areas_in_radians = (weights_of_areas / weights_of_areas.sum()) * (2 * pi)
+
+        # We add a zero in the begging for the cumulative sum
+        points = np.zeros((areas_in_radians.shape[0] + 1))
+        points[1:] = areas_in_radians
+        points = points.cumsum()
+
+        colors = [color_in_equal_space(area / areas_in_radians.shape[0]) for area in range(areas_in_radians.shape[0])]
+
+        arcs_data = pd.DataFrame({
+            'start_angle': points[:-1],
+            'end_angle': points[1:],
+            'line_color': colors
+        })
+
+        self.arcs_data = ColumnDataSource(arcs_data)
+
+        # Text
+        if self.label:
+            text_radius = 1.1
+            angles = (points[:-1]+points[1:])/2.0
+
+            text_positions = pd.DataFrame({
+                'angles': angles,
+                'text_x': np.cos(angles) * text_radius,
+                'text_y': np.sin(angles) * text_radius,
+                'text': list(self.label)
+            })
+
+            self.text_data = ColumnDataSource(text_positions)
+
+        # Lines
+
+        all_areas = []
+        for i in range(areas_in_radians.shape[0]):
+            all_areas.append(Area(weights_of_areas[i], points[:-1][i], points[1:][i]))
+
+        all_connections = []
+        for j, region1 in enumerate(self.matrix):
+            # Get the connections origin region
+            source = all_areas[j]
+            color = colors[j]
+            weight = weights_of_areas[j]
+
+            for k, region2 in enumerate(region1):
+                # Get the connection destination region
+                target = all_areas[k]
+                for _ in range(int(region2)):
+                    p1 = source.free_points.pop()
+                    p2 = target.free_points.pop()
+                    # Get both regions free points and create a connection with the data
+                    all_connections.append(p1 + p2 + [color, weight])
+
+        connections_df = pd.DataFrame(all_connections, dtype=str)
+        connections_df.columns = ["start_x", "start_y", "end_x", "end_y", "colors", "weight"]
+        connections_df["cx0"] = connections_df.start_x.astype("float64")/2
+        connections_df["cy0"] = connections_df.start_y.astype("float64")/2
+        connections_df["cx1"] = connections_df.end_x.astype("float64")/2
+        connections_df["cy1"] = connections_df.end_y.astype("float64")/2
+        connections_df.weight = (connections_df.weight.astype("float64")/connections_df.weight.astype("float64").sum()) * 3000
+
+        self.connection_data = ColumnDataSource(connections_df)
+
+    def yield_renderers(self):
+        """Use the marker glyphs to display the arcs and beziers.
+        Takes reference points from data loaded at the ColumnDataSource.
+        """
+        beziers = Bezier(x0='start_x',
+                         y0='start_y',
+                         x1='end_x',
+                         y1='end_y',
+                         cx0='cx0',
+                         cy0='cy0',
+                         cx1='cx1',
+                         cy1='cy1',
+                         line_alpha='weight',
+                         line_color='colors')
+
+        yield GlyphRenderer(data_source=self.connection_data, glyph=beziers)
+
+        arcs = Arc(x=0,
+                   y=0,
+                   radius=1,
+                   line_width=10,
+                   start_angle='start_angle',
+                   end_angle='end_angle',
+                   line_color='line_color')
+
+        yield GlyphRenderer(data_source=self.arcs_data, glyph=arcs)
+
+        if self.label:
+
+            text_props = {
+                "text_color": "#000000",
+                "text_font_size": "8pt",
+                "text_align": "left",
+                "text_baseline": "middle"
+            }
+
+            labels = Text(x='text_x',
+                          y='text_y',
+                          text='text',
+                          angle='angles',
+                          **text_props
+                          )
+
+            yield GlyphRenderer(data_source=self.text_data, glyph=labels)
+
+
+@help(ChordBuilder)
+def Chord(data, source=None, target=None, value=None, square_matrix=False, label=None, **kws):
+    """ Create a chord chart using :class:`ChordBuilder <bokeh.charts.builders.chord_builder.ChordBuilder>`
+    to render a chord graph from a variety of value forms.
+
+    This chart displays the inter-relationships between data in a matrix.
+
+    The data can be generated by the chart interface. Given a source
+
+    Reference: `Chord diagram on Wikipedia <https://en.wikipedia.org/wiki/Chord_diagram>`_
+
+    Args:
+        data (:ref:`userguide_charts_data_types`): the data source for the chart.
+        source (list(str) or str, optional): Data source to use as origin of the connection to a destination.
+        target (list(str) or str, optional): Data source to use as destination of a connection.
+        value (list(num) or num, optional): The value the connection should have.
+        square_matrix (bool, optional): If square matrix, avoid any calculations during the setup.
+        label (list(str or num) optional): The labels to be put in the areas.
+
+    Returns:
+        :class:`Chart`: includes glyph renderers that generate the chord
+
+    Examples:
+
+    .. bokeh-plot::
+        :source-position: above
+
+            import pandas as pd
+            from bokeh.charts import output_file, Chord
+            from bokeh.io import show
+            from bokeh.sampledata.les_mis import data
+
+            nodes = data['nodes']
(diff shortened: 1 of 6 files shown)


fpliger:
Yessssss! Thank you @merqurio for stepping ahead with this! It's really cool.

Please let us know where you need help with.. we are happy to help!

merqurio:
For sure with WebGL. I'm learning CoffeeScript little by little but the metaprogramming solution you use for the tools  it's making it hard for me :grin:. 

I hope I have the basic interface working before the weekend.

Thanks Fabio, always there !

bryevdv:
spurious test failure (we seem to get alot lately), I restarted the test

bryevdv:
@merqurio My apologies we decided to clean up the top level package for this release and it's introduced some merge conflicts for you. I hope it's not too bad to resolve them, but if here's anything I can do to help please let me know.

merqurio:
Done ! no problem, thanks @bryevdv. hope to finish this today :grimacing:

merqurio:
If I yield to `bokeh.charts.builder.Builder.yield_renderers` an `Arc` with a `ColumnDataSource`, would it be wrong ? I've been looking to the other implementations and they usually yield everything as individual glyphs. or maybe I'm getting it wrong. :confused: 

Thanks!

rothnic:
The primary approach has been to yield glyph renderers, which reference a glyph and a column data source.

merqurio:
Perfect ! Thanks @rothnic

bryevdv:
This is still marked `WIP` and the freeze is tomorrow. Let's target this PR for `0.11.1`

merqurio:
Understanding how you implement the chart interface took me longer than expected :disappointed_relieved:

bryevdv:
@merqurio No worries, we plan on having a `0.11.1` fairly soon and the list of things for `0.11` is already plenty long. This will be a great addition when it lands!

merqurio:
Hi, I'm getting this error, and I don't have cleat where should I declare this attributes, where should I ?

```
AttributeError: unexpected attribute 'start_angle' to ArcGlyph, possible attributes are bottom_buffer, color, dash, data, fill_alpha, fill_color, glyphs, label, left_buffer, line_alpha, line_color, operations, renderers, right_buffer, source, top_buffer, values, width, x or y
```

Thanks !

bryevdv:
@merqurio are those private implementation attributes? Then you can prefix them with an underscore, e.g. `_start_angle`, `_end_angle` if they should be "public" then you will need to add class attribute properties, similar to https://github.com/bokeh/bokeh/blob/master/bokeh/models/glyphs.py#L123

merqurio:
Can please someone test this ? The test code I use is [here](https://gist.github.com/merqurio/ab61bb4c7d08cccf4e58). :grin:

Maggie-M:
Looks good to me.  Just to confirm, this is what the chart is expected to look like?

screen shot 2016-01-13 at 3 30 06 pm

bokeh/charts/builders/chord_builder.py
```diff
@@ -0,0 +1,182 @@
+"""This is the Bokeh charts interface. It gives you a high level API
+to build complex plot is a simple way.
+
+This is the Scatter class which lets you build your Scatter charts
```
fpliger: Scatter -> Chord

bokeh/charts/builders/chord_builder.py
```diff
@@ -0,0 +1,182 @@
...
+from bokeh.core.properties import Instance
+
+
+# -----------------------------------------------------------------------------
+# Classes and functions
+# -----------------------------------------------------------------------------
+
+
+def Chord(data=None, agg=None, **kws):
+    """ Create a scatter chart using :class:`ChordBuilder <bokeh.charts.builders.chord_builder.ChordBuilder>`
```
fpliger: scatter -> Chord

bokeh/charts/builders/chord_builder.py
```diff
@@ -0,0 +1,182 @@
...
+    to render the geometry from values.
+
+    Args:
+        data (:ref:`userguide_charts_data_types`):  correlation table
+        labels (str or list(str), optional): the column  of label to use for the x and y
```
fpliger: label?

bokeh/charts/builders/chord_builder.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+    Returns:
+        :class:`Chord`:
+
+    Examples:
+
+    .. bokeh-plot::
+        :source-position: above
+
```
fpliger: Example code is missing here... we could use this https://gist.github.com/merqurio/ab61bb4c7d08cccf4e58, right? Or maybe a simplified version?

rothnic:
I took a look at the example data, and this is quite different compared to the rest of the chart interface. A general approach has been to at least support generating the derived dataset from typical table-like data.

I would consider the example data the derived data that should be able to be passed in directly, but we should also provide some easier avenue to creating the chart without a lot of up front manipulation.

fpliger:
Left a few comments. My last comment is that it's design differs a bit from the new charts design.. I think it looks good but would make an issue for a follow up effort to align this implementation with the rest...

In gereral I think it's in good shape

bryevdv:
> I took a look at the example data, and this is quite different compared to the rest of the chart interface. A general approach has been to at least support generating the derived dataset from typical table-like data.
 
I'm not sure about this particular case, but I do want to state that this is only appropriate for a subset of charts. It simply does not make sense in all cases. There are lots of charts, things like streamlines and quiver plots of vector fields, contour and "pcolor" for arrays, etc. that  will never look like tables, and it would not make sense to try and shoehorn things into that mold. It's probably worth codifying the set of "tabular" charts as the "statistical charts interface" or something.

fpliger:
> There are lots of charts, things like streamlines and quiver plots of vector fields, contour and "pcolor" for arrays, etc. that will never look like tables, and it would not make sense to try and shoehorn things into that mold. It's probably worth codifying the set of "tabular" charts as the "statistical charts interface" or something.

Yup... agree we should not be too strict on making everything look like a table. (`ColumnDataSource' is an example that shows that this one size fits all approach makes things hard sometimes). We should probably push more in the direction of this diversity of charts combined with diversity of inputs..

As a side note: That's also an argument for making data manipulation the more decoupled as possible..

bryevdv:
Yup. Have planned to introduce some new data source types to complement `ColumnDataSource` just have not had the time yet.

rothnic:
There is nothing tying Chart or Builder to table-like data. Currently, all other charts happen to support it through the use of the ChartDataSource which just provides some convenience manipulation of that kind of data, but you could provide an array of blobs and use that however you want as long as you yield the renderers.

I'm not saying that there shouldn't be charts that utilize something else, just that it should be carefully considered because most of the world's data is stored in that format, so by supporting producing the derived data from that format, you have made it easier to use to a large population of people. 

Sure, anyone can write some function to transform their data into the format used in the example, but then what if that chart just doesn't work for them when it is displayed. They will try another chart. If that chart has it's own custom format, you transform again. Then so on. The point of the standardization isn't to be a pain, it is to easily move between the charts with minimal manipulation of the data.

With some standardization on format, you get a lot of power. I don't think it is shoehorning at all, as it is relational data, which would easily fit into a relational data store. I can think of a relatively simple SQL query that would extract this kind of information from data I work with, so it shouldn't be hard to do the same with pandas. I think it is fine as-is for the PR, but I'll provide an example afterwards for deriving this dataset with pandas.

merqurio:
I wasn't expecting that many feedback :grin: 

@Maggie-M Looks good, but color are missing, did you pull all the changes ?

@fpliger Just made all the modifications you mention. Thanks for the feedback! I really appreciate it.

@rothnic I took the example data and modifications directly from one of the examples in the [gallery](http://bokeh.pydata.org/en/latest/docs/gallery/les_mis.html). I totally understand your point of passing a more structured data and to do the math in our side. I'm open to whatever the Bokeh team decides. Also, please, I'm not an expert in Bokeh, and I've been a bit lost at the time of implementing this as part of the chart interface, any feedback on that would be greatly appreciated.

rothnic:
### Current Sample Data

There are two "tables" in the sample data. Nodes and links. The links have a source id and a target id, where each id exists in nodes. The current approach so far is to have a single table-like input. So, you'd just be doing two left joins of the nodes into the links table to get what I'd consider consistent with the rest of the charts. 

It looks like the example added to the docstring creates an ndarray, which equates to a table, but I think a better, more simple and consistent example would be to use a dict. You would do some manipulation, then end up with something like `data={'source': [...], 'target': [...], 'value': [...]}`, then you could either pick those up by name, or use kwargs of Chord(data, source='source', target='target', value='value') to identify the columns containing the data.

### Why it makes sense to use a table for input to this chart.

I wanted to follow up on my reasoning, just to show a concrete example of the type of data I work with and how you might aggregate the data in a similar way, then describe how it would be useful for more than one chart type. 

I work with website activity data where we often want to understand how users are moving through the site, how many times this occurs, and how much revenue is associated with those. So we'll focus on three simplified tables in our data.

1. **page_view_facts**: columns of referring_page_key, page_key, clicks, revenue, visitor_key, page_view_id, date_key
2. **page_dimension**: url, domain, page_type, page_category, page_key
3. **visitor_dimension**: first_page_view_id, visitor_key

To produce the data for analysis, I would do the following:

``` SQL
SELECT ref_pages.page_type as referring_page_type,
    ref_pages.page_category as referring_page_category,
    pages.page_type as target_page_type,
    pages.page_category as page_category,
    IF(page_view_id=first_page_view_id, 1, 0) as new_visitor,
    SUM(clicks) as clicks,
    SUM(revenue) as revenue
FROM page_view_facts
INNER JOIN page_dimension as ref_pages ON ref_pages.page_key=page_view_facts.referring_page_key
INNER JOIN page_dimension as pages USING(page_key)
INNER JOIN visitor_dimension USING(visitor_key)
WHERE YEAR(date_key)=2015
GROUP BY 1, 2, 3, 4
```

At this point this data has multiple columns that could be used as the `value`, `source`, and `target` in the example data. Per each flow of a visitor from page type to page type, or page category to page category, I might want to know how many were new users (this would require further aggregating), how many clicks occurred as result of the navigation, or revenue. If I only use page types, I'd be removing the page category dimensions or vice versa, and I'd need to aggregate to that new level to produce the chord graph. This is how all other charts work is you could either pass in three columns (source, target, value), or identify them by name in an existing table. `Chord(df, source='referring_page_type', target='page_type', values='revenue', agg='sum')`

So at this point I have a chord graph, but maybe I don't think it worked well for my data, or maybe I want to augment my chord graph with a bar chart. I can easily reuse this dataset to have a bar chart that provides the avg number of clicks per page type, or something similar to the chord graph. `Bar(df, label='referring_page_type', values='revenue', stack='page_type', agg='sum')`

I don't want to be nitpicky, but I just think there is a big benefit of standardization of the interface where possible (this will become more apparent with interactive charts).  And I also wanted to thank @merqurio for the awesome chart. I'll definitely have to reference how you did this because I'd really love to do a sankey diagram for the same type of analysis that I described above.

merqurio:
Okay @rothnic, now I totally see your point, and I think that implementing that would be awesome. I don't usually use it for those type of data BUT I would love to. I can try to implement it but let's imagine that your data, like in my real world scenario, is a matrix of [150,150], how would you manage it ?

I think that maybe depending on the arguments passed to the object, we could manage to just plot it, or to do the calculations from a data source and then plot it. What do you think @fpliger & @bryevdv ?

fpliger:
@rothnic argument does make sense and I'm +1 on going down this road here.. and I do think that interface standardization does make a lot of sense. (I still don't think it always makes sense to use this API... specially if we think about Geo or Graphs/Networks related charts)

@merqurio  Regarding your question, can you make a small example?

rothnic:
@merqurio: I think this is similar to the bar chart in that there are cases where it might or might not need aggregating as in my example. In general, a matrix is an efficient way to store the relationships between two dimension, but it would be difficult to handle where you have more than two dimensions in the data.

So, the matrix is fine to use internally to chart once you interpret the inputs, but the most flexible format is one of records as in the case I described. Imagine you take a simplified version of the example that I provided, the user would need to transform their data into the table format, then you do whatever you want internally to produce the renderers. The use could have looked like this:

`Chord(data, source='referring_page_type', target='page_type', values='clicks')`

Where data looks like this: 

referring_page_type, page_type, clicks
index, vendor, 10
vendor, index, 20
index, blog, 10
blog, index, 5

Which could be quickly created with this:

``` python
data = {'referring_page_type': ['index', 'vendor', 'index', 'blog'],
            'page_type': ['vendor', 'index', 'blog', 'index'],
            'clicks': [10, 20, 10, 5]
}

# in a dataframe
df = pd.DataFrame(data)

# internally to the chart, you'd aggregate based on the inputs provided
# inputs provided are `source`, `target`, `values`, `agg`
# what you want to do is equivalent of df.groupby(['referring_page_type', 'page_type']).clicks.sum()
agg_data = getattr(getattr(df.groupby([source, target]), values), agg)()
```

In this case, you would have the input as I described, and you'd internally create a pandas dataframe. In this case there are no duplicate records (which could have existing with more columns as in the previous example), but you could still consistently apply a default sum on the column that is provided as `values`. Because the grouping would apply to the two dimensions provided, `source`, and `target`, the sum would apply only to a single row. If you had the input of my initial example, the sum would actually do something because you might have rows where the referring_page_type, and page_type are identical.

So, I'd lean more towards accepting the inputs as the table format because it doesn't require reconfiguration outside the chart to change what you are looking at, but I'm not against supporting the matrix format some way as well. I consider that almost as a "specification" kind of input and I've considered supporting very specific specifications for explicit control of the chart for things like Bar, when the table input interface doesn't do what you want.

With my original example, I can swap between these two cases without having to reformat the input data:

``` python
Chord(data, source='referring_page_type', target='page_type', values='clicks')

Chord(data, source='referring_page_category', target='page_category', values='clicks')
```

Internally, we'd be aggregating to the level of `source` and `target` in both cases based on `values`.

rothnic:
Also, here is how you could use the example data to create a table.

``` python

nodes = data['nodes']
links = data['links']

nodes_df = pd.DataFrame(nodes)
links_df = pd.DataFrame(links)

source_data = links_df.merge(nodes_df[['name']], how='left', left_on='source', 
                        right_index=True)
source_data = source_data.rename(columns={'name': 'source_name'})

source_data = source_data.merge(nodes_df[['name']], how='left', left_on='target', 
                        right_index=True)
source_data = source_data.rename(columns={'name': 'target_name'})

In[37]: source_data.head()
Out[37]: 
   source  target  value      source_name      target_name
0       1       0      1         Napoleon           Myriel
1       2       0      8  Mlle.Baptistine           Myriel
2       3       0     10     Mme.Magloire           Myriel
3       3       2      6     Mme.Magloire  Mlle.Baptistine
4       4       0      1     CountessdeLo           Myriel
```

This joins the two "tables" in the example data via two separate left joins.

merqurio:
Great feedback :+1:  !!! Expect a commit with all this Tuesday/wensday next week! :grin:

merqurio:
I'm back sorry for the absence. :+1:

merqurio:
Should I take `bokeh/charts/builders/bar_builder.py` as an example of how to organize all the code ?

alexsavio:
Hi, 
This is indeed quite interesting for some of my analyses. My MD boss would like it for sure!

This is nice to see the results of connectivity analyses in different types of acquisitions/brains/patient groups, etc...

I guess WebGL is a bit far from my reach for now, but after the soon-to-be @merqurio push I could help, just let me know what can I do. I am missing the labels on the circle, for example.

Keep up the good work! :D

merqurio:
Yeahp, I already worked on that, right now I'm looking a nice way of including everything commented by @rothnic et al. I'll finish it soon.

bryevdv:
@merqurio do you expect to be able to continue work on this? Perhaps there is a short-term stopping point, and then issues could be made for any remaining tasks?

merqurio:
I'm on it; It's been busy time for me but I just started vacations yesterday for a week, I hope to finish all the modifications you asked soon.

Right now I'm adding the preprocess of the data to the visualization this way:

**Data input:**
- [x] `matrix`(square matrix) and `labels`
- [x] Table with: `source`, `target`, `value`
    - `source` and  `target` used as labels
    - `value` as the width of the arc and num of connections
- [ ] Table with: `source`, `target` and a constant `value`
- [ ] Table with `source`, `target` no values
    - Value calculated from aggregation of source and target
    - `source`, `target` used as labels

I'll rebase, squash and push all

bryevdv:
@merqurio I understand completely, just checking in :)

merqurio:
Ok, I got all the preprocessing done. :v: 

TODO:
- [ ] Resolve bug on number of beziers _¿ Was there form the beginning ?_
- [ ] Add Tooltips to arcs with the name of the label of the source
- [ ] Custom callback on click to make gray and semi-transparent non-selected arcs and beziers
    - Should priority on click go to source or target ? In my opinion source priors.

bokeh/charts/builders/chord_builder.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+
+    default_attributes = {'color': ColorAttr(),
+                          'marker': MarkerAttr(),
+                          'stack': CatAttr()}
+
+    # TODO: Determine what should be included here (origin, destination and value?), maybe up to the construction used?
+    dimensions = ['values']
+
+    values = Dimension('values')
```
merqurio: If someone can tell me what to do with this, that would be great :grin:

bryevdv: is it defined on a parent class? Then you want to use `Override`

fpliger: No, it shouldn't be defined on the parent class

merqurio: Ok, then I remove it ? I mean @havocp says [here](https://github.com/bokeh/bokeh/blob/master/bokeh/charts/builder.py#L137) explicitly that it **must** be overridden. Thanks for your time guys !

bryevdv:
@merqurio there seem to be two problems with the docstring, the first is that Sphinx requires a blank line between options an contend (in this case, the inline script), so try applying this patch:
```
diff --git a/bokeh/charts/builders/chord_builder.py b/bokeh/charts/builders/chord_builder.py
index b952967..9a1c114 100644
--- a/bokeh/charts/builders/chord_builder.py
+++ b/bokeh/charts/builders/chord_builder.py
@@ -221,6 +221,7 @@ def Chord(data, source=None, target=None, value=None, square_matrix=False, **kws
 
     .. bokeh-plot::
         :source-position: above
+        
         import numpy as np
         import pandas as pd
         from bokeh.charts import output_file, show, Chord
```
Next, that example script does not actually work for me, I get 
```
ValueError: cannot label index with a null key
```
Though, the `Chord` section in the reference guide does otherwise show up fine.

bokeh/charts/builders/chord_builder.py
bryevdv: I will confess I don't see the connection, that link is for an attribute called 'dimension' and this line is a property called `values` but I am not an expert on the charts code.

bryevdv:
moved milestone and tags to a related issue

merqurio:
Right now, the Python part is **almost** done, just labels are missing. IMO, the labels should appear when hovering the arcs, but that's not possible right now so I'm going to implement that a text glyph renders near the arc until a better solution comes up. I'll also expand the docs.

Do you agree @bryevdv if the hovering part is done in another PR ? Thanks again for your patience. :v:

bryevdv:
I think hover in a separate PR is fine. It might be a good idea to merge current master into this PR. In any case please mark `ready` when you'd like more testing / closer review

merqurio:
This should be ready to be tested @bryevdv. I would appreciate feedback on the code too, if someone thinks something could be improved ! ✌️

bryevdv:
@merqurio I will try to take a closer look tonight. Is there a nice example we can add to `examples/charts/file` ?

merqurio:
Sure ! I'll create a nice example and add it, I just rebased and squashed everything too, so the PR is just a clean single commit.

bryevdv:
@merqurio there seems to be a real test failure (failed import)

merqurio:
yahp, I think I just found it while creating the example file. BTW, this is a preview of it:

screen shot 2016-04-20 at 18 11 15

bryevdv:
Looks great! I think we could make the labels a little fancier so the ones on the left are not upside down, but let's get this merged when test are green, and make issues for incremental improvements like that. Thanks!

merqurio:
Are those failures related to the PR ? Doesn't seem so.. 😧

bryevdv:
I think those failures are spurious, I've restarted them

bryevdv:
OK Travis is green, merging now. Thanks for the contribution @merqurio !

merqurio:
Thanks to you @bryevdv. This was my first PR to a big project ✌️ 🎉 🎈

damianavila:
Congrats @merqurio, really great PR!

merqurio:
Thanks @damianavila !! I really want to keep contributing to bokeh

Issue #3302, opened 2015-12-09
validation warning for dynamic_map.py


Maggie-M:
ERROR:/Users/mmari/Desktop/continuum/bokeh/bokeh/validation/check.pyc:W-1001 (NO_GLYPH_RENDERERS): Plot has no glyph renderers: Figure, ViewModel:Plot, ref _id: 2d5c737d-8d7f-4b49-96d1-d75c6fb49d4a

brendancol:
Relax NO_GLYPH_RENDERERS validation #3223: https://github.com/bokeh/bokeh/issues/3223

bryevdv:
This does seem like a duplicate?

Issue #3223, opened 2015-12-02
Relax NO_GLYPH_RENDERERS validation


mattpap:
Currently `Plot` requires at least one `GlyphRenderer`, but with new additions, i.e. tile renderers, you can get a valid plot without glyph renderers. We have to add one more abstract base class and update `Plot`'s validation. See issue #3189 for details.

damianavila:
This was closed but I am not sure if the validation was "relaxed" as proposed by the first message, can @bryevdv or @mattpap validate what the status of this one? Thanks!

bryevdv:
Yes this was resolved (tile/dynamic map no longer throw validation errors)

damianavila:
>Yes this was resolved (tile/dynamic map no longer throw validation errors)

Thanks!

PR #3368, opened 2015-12-17
NO_{GLYPH->DATA}_RENDERER warning

brendancol:
This PR addresses two current issues:
https://github.com/bokeh/bokeh/issues/3223
https://github.com/bokeh/bokeh/issues/3302

In the past, any valid plot required a `GlyphRenderer`.  That is no longer case with the addition of `TileRenderer` and `DynamicImageRenderer`.  The PR adds a new abstract base class `DataRenderer` to `bokeh/models/renderers.py` which inherits from `Renderer` and is now the parent class of `GlyphRenderer`, `TileRenderer`, and `DynamicImageRenderer`.  This parent class can now be validated in `bokeh/models/plots.py`. Initial testing looks fine and this removes the warning displayed by the `dynamic_map.py` example.

There are also a couple flake8 fixes...


bokeh/models/renderers.py
```diff
@@ -26,7 +26,14 @@ class Renderer(Model):
     generally useful to instantiate on its own.
 
     """
-class TileRenderer(Renderer):
+
+@abstract
+class DataRenderer(Renderer):
+    """ A base class for data renderer types (e.g. GlyphRenderer). ``DataRenderer`` is not
+    generally useful to instantiate on its own.
```
mattpap: No. Abstract classes can not be instantiated, but I see where this comes from. Please change `Renderer`'s docstring as well. "An abstract base class for ..." should suffice. Also, `GlyphRenderer` should be enclosed with double backticks.

mattpap:
Looks good. Tests failed due to phantomjs crash.

Great you improved the quality of the source code involved, however you should have approached this differently for three reasons. First, it's harder than necessary for a reviewer to separate meaningful changes from refactoring. Secondly, if the PR stalls or there are lots of merges in the meantime, you will have a lot of really boring work resolving conflicts. Lastly, you are at risk of stalling a PR because we will be arguing about stylistic issues. My suggestion, at least separate refactoring from important changes in different commits. Preferably, any large scale refactoring should go through a separate PR. In this particular case, first option would suffice.

brendancol:
@mattpap great explanation. moving flake8 fixes to separate PR.

mattpap:
@brendancol, that wasn't necessary in this PR, I was just talking about future PRs. But OK, lets proceed.

brendancol:
Thanks to BeyondCompare it was no problem.

bryevdv:
LGTM, tested locally, will merge when travis passes and report inspected.

PR #3370, opened 2015-12-18
Better approach to notebook comms:

### LINKED ISSUE #1732: push_notebook memory leaks

bryevdv:
issues: fixes #1732 

On notebook show, create new comms and coordinate comms_target in `doc_js`. Store comms as `last_comms` on state. Close any previous `last_comms`.

On push_notebook, use `last_comms` to send message to last showndocument.


bryevdv:
@havocp I could use help with the two `TODO`s, I think they are both one-liners just need a nudge in the right direction.

bryevdv:
handling of `last_comms` can probably be tightened up too.

bokeh/io.py
```diff
@@ -424,6 +430,38 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+
+    if not document:
+        warnings.warn("No document to push")
+        return
+
+    if not state.last_comms:
+        warnings.warn("Cannot find a last shown plot to update")
+        return
+
+    # TODO prepare actual patch
```
havocp: Some ideas:

1) each time we send the doc, save its JSON. When we want to patch it, compute new json and diff vs old json. Will have to port the JSON diff algorithm we have over in document.coffee to Python. Downside: have to port the diff code. 

2) each time we send the document, connect to on_change and start recording all events. when we want to patch, replay them all as patch-doc. Downside: event record could grow infinitely in theory.

3) resend entire doc each time. Downside: sends entire doc to change one property.

There could be issues when coffee makes changes during document rendering; those changes won't be synced back. So the events may not always apply correctly, depending on what they are, because the Python side won't have an accurate copy of the document.

bokeh/io.py
```diff
@@ -424,6 +430,38 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+    if not document:
+        document = state.document
+
+    if not document:
+        warnings.warn("No document to push")
+        return
+
+    if not state.last_comms:
+        warnings.warn("Cannot find a last shown plot to update")
+        return
```
havocp: this warning should probably say "call output_notebook() and show() before push_notebook()" or otherwise explain how to fix it

bryevdv:
OK this is working! I will clean up and add some tests and such tomorrow. Any additional suggestions welcome.

bokeh/_templates/doc_js.js
```diff
@@ -12,7 +12,14 @@ Bokeh.Collections.register_models({
   var websocket_url = null;
 {%- endif %}
 
+{% if comms_target -%}
+  var comms_target = "{{ comms_target }}";
+{%- else %}
```
havocp: Might be nice to call this notebook_comms_target or somehow be clear it's for the notebook case.

bokeh/io.py
```diff
@@ -424,6 +431,38 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
     _push_to_server(session_id=session_id, url=url, app_path=app_path,
                     document=document, io_loop=io_loop)
 
+def push_notebook(document=None, state=None):
+    ''' Update the notebook with the data for the current document.
```
havocp: this won't work or even make sense if you're using notebook with output_server right ? it should probably throw an exception saying "use push() instead" or maybe just do the right thing ...

bokeh/io.py
```diff
@@ -424,6 +431,38 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+    if not document:
+        warnings.warn("No document to push")
+        return
+
+    if not state.last_comms:
+        warnings.warn("Cannot find a last shown plot to update")
+        return
+
+    import json
+    msg = state.document._compute_patch_since_json(state.last_json)
```
havocp: the passed-in doc is ignored I think? if we switch to a different doc I think a push-doc message is probably a lot better than a patch message.

bokehjs/src/coffee/common/embed.coffee
```diff
@@ -8,6 +8,18 @@ HasProperties = require "./has_properties"
 {pull_session} = require "./client"
 {Promise} = require "es6-promise"
 
+_handle_notebook_comms = (msg) ->
+  logger.debug("handling notebook comms")
+  doc = _.values(@)[0] # should only be one
```
havocp: hmm. it may be that the comms target should be in the render item (like sessionid) so there can be a different one for each document in principle. Also it's analogous to sessionid here perhaps, we could handle it in a very similar way.

bryevdv: Adding comms creation to the middle of `_standalone_docs_json_and_render_items` seems a bit strange but I don't see anywhere else to do it. Also, `notebook_div` currently only always sends one doc, it does:
```
item = render_items[0]

div = _div_for_render_item(item)
```

havocp: I know it's always one now, just trying to reduce the number of ways the code works.
I was imagining maybe in notebook_div do `render_items[0]['notebooktarget'] = comms_target` after calling `_standalone_docs_json_and_render_items`, and then embed.coffee would handle that case in the loop over the items. then you can pass each _handle_notebook_comms the single specific document and div that it will care about.
If you allow PUSH-DOC this may be especially useful because you'll need the div in order to replace the doc entirely.

bryevdv: OK routing through `render_items` should be done now

bryevdv:
@havocp OK this now tunnels comms through individual `render_items` and also I believe it implements `PUSH-DOC` (not quite sure how to test)

bryevdv:
OK I think this is done for now. It has one current limitation, which is that `push_notebook` always uses the `Comms` from the *last shown* thing. This is a problem if someone wants to have multiple interactors  in a single notebook. I think this can be resolved, but my "quick fix" that I just tried was not sufficient, so I would like to leave it for later.

bokeh/core/properties.py
```diff
@@ -711,6 +711,38 @@ def __setattr__(self, name, value):
             raise AttributeError("unexpected attribute '%s' to %s, %s attributes are %s" %
                 (name, self.__class__.__name__, text, nice_join(matches)))
 
+    # this is like _value_record_references but expects to find refs
+    # instead of models, and takes a doc to look up the refs in
+    def _json_record_references(doc, v, result, recurse):
+        if v is None: return
+        if isinstance(v, dict) and set(v.keys()) == set(['id', 'type']):
```
havocp: there's an optional `subtype` key (see `def ref`) I guess, so it should be that the set is a superset of `['id', 'type']`

bokeh/core/properties.py
```diff
@@ -711,6 +711,38 @@ def __setattr__(self, name, value):
...
+            for elem in v:
+                HasProps._json_record_references(doc, elem, result, recurse)
+        elif isinstance(v, dict):
+            for k, elem in v.items():
+                HasProps._json_record_references(doc, elem, result, recurse)
+
+    # add all references from 'v' to 'result', if recurse
+    # is true then descend into refs, if false only
+    # descend into non-refs
+    def _value_record_references(v, result, recurse):
```
havocp: should be implemented in terms of `_visit_value_and_its_immediate_references()` I think, or `collect_models` (the latter handles deep recursion) ...

also, is the `recurse` parameter ever used?

bokeh/document.py
```diff
@@ -531,6 +532,100 @@ def _initialize_references_json(cls, references_json, references):
...
+        removed = from_keys - to_keys
+        added = to_keys - from_keys
+        shared = from_keys & to_keys
+
+        events = []
+        for key in removed:
+          # we don't really have a "remove" event - not sure this ever happens even -
+          # but treat it as equivalent to setting to null
+          #events.append(Document._event_for_attribute_change(from_obj, key, None, value_refs))
+          raise RuntimeError("'RemoveAttribute' events not supported")
```
havocp: delete the comment?

It looks like this is in fact an assertion failure (i.e. by using the API it should be impossible to trigger) because `Property.__delete__` removes the application-set value but it doesn't remove the attribute itself since the default remains.
So we could say `raise RuntimeError("bug: should not be possible to delete attribute %s" % key)` or something

bokeh/document.py
```diff
@@ -531,6 +532,100 @@ def _initialize_references_json(cls, references_json, references):
...
+        for r in to_json['roots']['root_ids']:
+          to_roots[r] = to_references[r]
+          to_root_ids.append(r)
+
+        from_root_ids.sort()
+        to_root_ids.sort()
+
+        if set(from_root_ids) - set(to_root_ids):
+          # this would arise if someone does add_root/remove_root during
+          # document deserialization, hopefully they won't ever do so.
```
havocp: this comment is accurate in coffee but not here. In python as you're using this, it would happen if someone does add/remove root between show()'ing in a notebook for the first time and a later push_notebook().

bokeh/document.py
fpliger: +1

sphinx/source/docs/user_guide/notebook.rst
```diff
@@ -23,19 +23,16 @@ are required. You can see an example below:
...
+It is possible to drive updates to Bokeh plots using Jupyter
 notebook widgets, known as interactors. The key doing this is the
-|push_notebook| method on |ColumnDataSource|. This method allows you to
-update plot data sources in the notebook, so that the plot is made to
-update. Typically, |push_notebook| is used in the update callback for the
-interactor. An example is shown below:
+|push_notebook| function. This function allows you to update document
+data and properties in the notebook, so that any plots, etc are made to
+update. A common use of |push_notebook| is in an update callback for
+interactors. An example is shown below:
```
fpliger: we should add a note that `push_notebook` only affects the last executed plot

bokeh/io.py
```diff
@@ -424,6 +431,67 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+    if not state.last_comms:
+        warnings.warn("Cannot find a last shown plot to update. Call output_notebook() and show() before push_notebook()")
+        return
+
+    import json
+    to_json = document.to_json()
+    if state.last_pushed_doc and state.last_pushed_doc != document:
+        msg = dict(doc=to_json)
+    else:
+        msg = Document._compute_patch_between_json(state.last_pushed_json, to_json, document)
```
havocp: I think these hand-constructed messages don't have the msgtype or msgid, which is fine since they don't go over the websocket, except how will the coffee tell whether it's a PUSH-DOC or a PATCH-DOC?

bokeh/io.py
```diff
@@ -424,6 +431,67 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+        warnings.warn("No document to push")
+        return
+
+    if not state.last_comms:
+        warnings.warn("Cannot find a last shown plot to update. Call output_notebook() and show() before push_notebook()")
+        return
+
+    import json
+    to_json = document.to_json()
+    if state.last_pushed_doc and state.last_pushed_doc != document:
```
havocp: I don't think last_pushed_doc can be false, if last_comms is true as checked above - we rely on that for using `last_pushed_json` unconditionally below, as well.

bokehjs/src/coffee/common/embed.coffee
```diff
@@ -8,6 +8,18 @@ HasProperties = require "./has_properties"
 {pull_session} = require "./client"
 {Promise} = require "es6-promise"
 
+_handle_notebook_comms = (msg) ->
+  logger.debug("handling notebook comms")
+  # @ is bound to the doc
+  @apply_json_patch_string(msg.content.data)
```
havocp: this assumes you have the PATCH-DOC type json not the PUSH-DOC type json. I think probably the json should include msgtype?

However, It looks like we don't have coffee code to handle PUSH-DOC right now, because we never send a PUSH-DOC from server to client on the websocket. That said. I think it's a one-liner:
`@replace_with_json(msg.content.data['doc'])`

havocp: to test this you have to create a new `Document()` in the notebook and pass it to `push_notebook` I guess

bokehjs/src/coffee/common/embed.coffee
```diff
@@ -163,4 +179,4 @@ embed_items = (docs_json, render_items, websocket_url) ->
 
 module.exports =
   embed_items: embed_items
-  inject_css: inject_css
+  inject_css: inject_css
```
havocp: your editor seems to hate trailing newline but emacs/vi/traditional editors really want it to be there :-)

bokehjs/src/coffee/common/embed.coffee
```diff
@@ -8,6 +8,18 @@ HasProperties = require "./has_properties"
...
 {Promise} = require "es6-promise"
 
+_handle_notebook_comms = (msg) ->
+  logger.debug("handling notebook comms")
+  # @ is bound to the doc
+  @apply_json_patch_string(msg.content.data)
+
+_init_comms = (target, doc) ->
+  logger.info("Initializing Jupyter comms for target #{target} and doc #{doc}")
+  comm_manager = Jupyter.notebook.kernel.comm_manager
```
damianavila: Just a comment, eventually the `Jupyter` namepace would/could be eliminated (not sure if it will happen or when) and you will need to require Jupyter from namespace...
```
define([
    'base/js/namespace'
], function(Jupyter) {```
```
As I said, the elimination is proposed (in fact you currently can load Jupyter like I show above) but not sure when will happen and if it will happen at the end...

damianavila:
@bryevdv , I just did a quick overview of the PR and  I like what I see... I just left a comment for the future.

bokehjs/src/coffee/common/embed.coffee
bryevdv: But that does not work now? Just wondering if there is any path that leads to "future proof".  `define` is synchronous at least, right? Maybe something like this could work:
```
if Jupyter?
    comm_manager = Jupyter.notebook.kernel.comm_manager
    comm_manager.register_target(target, (comm, msg) ->
        comm.on_msg(_.bind(_handle_notebook_comms, doc))
    )
else 
    define([ 'base/js/namespace'], (Jupyter) ->
        comm_manager = Jupyter.notebook.kernel.comm_manager
        comm_manager.register_target(target, (comm, msg) ->
            comm.on_msg(_.bind(_handle_notebook_comms, doc))
        )
```

damianavila: define works just fine right now... so if you want to be future proof, you should just use the else branch, I think...

bryevdv: I get this:
```
TypeError: define is not a function. (In 'define', 'define' is undefined)
```

bryevdv: needs to be `window.require` not `define`

bokeh/document.py
bryevdv: Is the solution to remove the entire check then? Or does this case need extra code to handle it?

bokeh/io.py
bryevdv: I'll just check the keys for now. Later I'd like to send real protocol messages / make NM comms just another kind of session.

bokehjs/src/coffee/common/embed.coffee
bryevdv: spoke too soon, that's not working reliably. I imagine because `require` is asynchronous, so sometimes the python tries to connect before the JS comms is done or even starts to get registered. (se below) So I'm just going to put it back the way it was. Please implore them not to remove important things from `Jupyter`. It's of no benefit that I can see to do do, and I have no idea how any one would ever be able to reliably make a python connection work if they can't at least assume the JS target has been already been registered. 

screen shot 2015-12-21 at 3 26 38 pm

bokeh/document.py
havocp: I would either make the comment/error more accurate or handle the case. Easiest of course is to just change what the error says ("current limitation: push_notebook can't push a change in root models" or something).

bokehjs/src/coffee/common/embed.coffee
damianavila: I will try to research on this later... but I certainly do not expect the removal of Jupyter namespace anytime soon... even more if there are cases that actually need the namespace.

bokeh/document.py
bryevdv: Well it's pretty easy to get into this situation in the notebook (and it does break). If there is a reasonable path to make it work we should do that.

bokeh/core/properties.py
bryevdv: It's called with `True` in `_event_for_attribute_change`

havocp: I saw that but didn't see where true or false changed the behavior of this method

bryevdv: Oh right, no the function is implicitly recursive in any case.

birdsarah:
Tested locally and all seems to be working, some small non-blocking notes:
- I seem to have to run the output_notebook cell twice to get the resources to load (seems to be underscore not being loaded)
- The console is pretty noisy - listing the set_data, render times and "handling notebook comms" for every change - I tried explicitly setting `BOKEH_LOG_LEVEL=info`
- The interact numba example can get pretty slow e.g. if I run the first part, then the second, then go back and run the first again, the first is slower the second time round - although still functioning - I have no idea whether this is anything to do with this PR, just wanted to note it.

Question: I was adding Resources(mode='server') and using those as I couldn't figure out how else to get the local resources to load - help gratefully appreciated.

bryevdv:
> I seem to have to run the output_notebook cell twice to get the resources to load (seems to be underscore not being loaded)

OK that's weird I definitely have never had to do that. 

> The console is pretty noisy - listing the set_data, render times and "handling notebook comms" for every change - I tried explicitly setting BOKEH_LOG_LEVEL=info

Again not sure what's going for you I tried just now with both `info` and `debug` and they both worked just as expected (and without specifying it defaulted to `info`)

> The interact numba example can get pretty slow e.g. if I run the first part, then the second, then go back and run the first again, the first is slower the second time round - although still functioning - I have no idea whether this is anything to do with this PR, just wanted to note it.

Given the current "last plot" limitation I am probably going to delete the later parts

For resources I use `BOKEH_RESOURCES=inline`

sphinx/source/docs/user_guide/notebook.rst
bryevdv: done

bokeh/core/properties.py
bryevdv: removed. I am not sure how to re-implement this with. It's also not ever currently getting called? The `value_record` function has
```
if v['id'] not in result:
                result[v['id']] = v
``` 
which should crash if it was called `v` should be a `Model` there

bokeh/document.py
bryevdv: So actually this situation does not trigger that exception. I am not sure what is up. I do this:

* create p
* show it
* create a new p
* push notebook
I get an error, but not that value error, and if I print the length of `from_root_ids` and `to_root_ids` they are both 1.

havocp: Looks like my coffeescript you ported this from is broken and only sees removed roots not added roots, so it silently gets confused with added roots. The conditional both here and in coffee needs to detect any difference in the two root sets. Maybe `if from - to: raise "can't remove roots"` and `if to - from: raise "can't add roots"`

bryevdv: I'm actually starting to wonder if the "record events" approach might not be cleaner.

havocp: I think it would be less code but it has the issue that there's no limit on the size of the event record, and no way to know if the event record will ever be used. So it's possible to keep around a really giant list of events each of which can be large due to `references`. To fix that we'd have to write an event compression thing which could be done with this "compute diff" code we're writing here... (compress events by converting doc to json, apply all the events, compute diff, compute new events from diff...)

bryevdv: Couldn't we merge/collapse patch events?

havocp: the algorithm to do that is where things will get complicated, is all. one algorithm for it is this same diff code... another is n-squared (compare each new event to all previous)... maybe could index past events by model and attribute modified 

I just had a simpler idea. Keep a list of changed models and attributes.

havocp: oops, hit the button before I finished. Add a document listener that tracks a set of changed models and attributes, a set of added roots, set of removed roots. Don't have to track attribute values (we'll get the latest values on demand.) Whenever a root is removed, remove any no-longer-in doc models from the set of changed models. Then when you `push_notebook`, generate RootAddedEvent, RootRemovedEvent, and a bunch of ModelChangedEvent from your recorded set of changes. So instead of having to compute the diff, we track events to record the diff, but we don't record raw events, only a resulting diff.

Anyway. I think you basically have the patch working as it is now, but maybe that would have been a better approach?

mrocklin:
Trying this within a small application.  Getting the following error within the browser (sorry for the wonky error message.)  Not sure how to debug further.

```
Error: Could not open comm↵    at new Error (nativ…in.js?v=40e10638fcf65fc1c057bff31d165e9d:12785:33", message: "Could not open comm", error_stack: 
```

mrocklin:
Also getting errors within the notebook that look like the following

    TypeError: Bokeh.embed.embed_items is not a function

My operations are in the following order:

1.  Call output_notebook
2.  Create figure
3.  Call show(figure)

mrocklin:
OK, that was apparently a version mismatch.  

New question:  when I show the same figure in multiple cells of my notebook a call to `push_notebook` seems to only affect the most recent rendering of that figure.  Is there a way around this?

bryevdv:
> Is there a way around this? 

Not yet. It's unclear what a "right" answer is. E.g. for instructional notebooks that build something up step by step you do only want the last shown figure to affected. We will probably add some explicit handle that can be passed to `push_notebook` to explicitly select something to update, but that is work to be done.

mrocklin:
Verified that I've left my push_notebook streaming scripts up for a decent while now and haven't experienced a memory leak

sphinx/source/docs/user_guide/notebook.rst
mrocklin: Is there a way to update multiple plots in the notebook?  (this is somewhat important to me)

bryevdv: It's important to everyone I am sure. :) I am going to try and get this done today.

bryevdv:
@mrocklin I think I have a solution for you can you try the PR now? You can download a notebook here: https://notebooks.anaconda.org/bryanv/notebook-comms It does note render well due to missing `Jupyter` JS module but you should be able to download it.

bryevdv:
The basic idea is this: `show` returns an opaque handle that can be passed to `push_notebook`:
```
p1 = figure(**opts)
r1 = p1.circle([1,2,3], [4,5,6], size=20)

p2 = figure(**opts)
r2 = p2.circle([1,2,3], [4,5,6], size=20)

t = show(hplot(p1, p2))

r1.glyph.fill_color = "white"
push_notebook(handle=t)
```
witbhout any handle passed, `show_notebook` still updates whatever was last shown.

bryevdv:
OK the last commit gets static renderered notebooks rendering https://notebooks.anaconda.org/bryanv/color_scatterplot

PR #3373, opened 2015-12-19
Register modules from plugins in the main bokeh bundle

### LINKED ISSUE #3339: Adding User Defined Layouts

mattpap:
This allows bokeh to know about plugins' modules, to allow custom models to require such modules. Previously this resulted in module not found exception. The new approach is to register plugins' modules in the main bundle (so there is one set of modules, instead of 1 + n), resulting in one `require()` being used by all involved parties (main bundle, plugins and custom models).

fixes #3339

### DIFF
--- a/bokehjs/gulp/tasks/scripts.coffee
+++ b/bokehjs/gulp/tasks/scripts.coffee
@@ -167,7 +167,7 @@ gulp.task "scripts:build", (cb) ->
       .pipe change (content) ->
         "(function() { var define = undefined; return #{content} })()"
       .pipe change (content) ->
-        "bokehRequire = #{content}"
+        "Bokeh = #{content}"
       .pipe(insert.append(license))
       .pipe(sourcemaps.write('./'))
       .pipe(gulp.dest(paths.buildDir.js))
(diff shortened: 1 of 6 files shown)


mattpap:
@brendancol, please test if this works for you. I modified `glyphs/custom` to show that custom models can extend widgets.

mattpap:
@damianavila, why did you remove metadata (labels, milestone) from this PR?

damianavila:
@mattpap yes... sorry about that, I am trying to follow the BEP... since the PR links the original issue, it is the issue where you have to include the metadata...
I kind of agree/accept if you setup "temporary" milestones on PR for easy filtering, and in fact, I usually do not remove those until the release, but labels should be definitely in the corresponding issue. 
If you want to setup the milestone here again, just go ahead... additionally, we need to improve the BEP to include this temporary filtering action... and come up with something better to deal with issues/PRs :wink:

bryevdv:
@mattpap LGTM and tested that the example works. I think this will work for @brendancol's needs but it's a solid PR so I am going to merge now in any case.

PR #3374, opened 2015-12-19
Reorganize top level package 

bryevdv:
ping @bokeh/dev 

My intent with this PR is to try and move things that average users will never have any reason to use directly out of the top level and into `bokeh.core` or `bokeh.utils`, etc. There are a few other things that could potentially be moved out (`_version.py`, `application`, `sphinxext`) as well. It also makes `bokeh.plotting` a package, consistent with the other APIs. From my perspective this makes the structure of the library easier to document. Opinions welcome. 

This obviously makes a number of small changes to many files, and shuffles other files around. I'm fine to keep this up to date and merge after `0.11`. OTOH in might be nice to get this out of the way in a release that will already have other changes.


bryevdv:
There are also some things in `bokeh.util` that depend on things in `bokeh`. It might make a nice distinction to move anything that depends on `bokeh` into `bokeh.core` and `bokeh.util` can be left for truly general utils that only depend on system libs.

mattpap:
>  I'm fine to keep this up to date and merge after 0.11

Maintaining this PR for next couple weeks is an utter waste of resources. Lets review and merge this as soon as possible.

pzwang:
I'm not against this in principle, and I haven't looked into any of the detailed changes in the commits.

The thing that makes me gun shy about this is that we already get so many people complaining about how existing code and examples out there in the wild don't work, because of the amount of API churn.  We also see people complaining that googling for these API breakages is difficult or impossible, and they give up on using the library.  I know a bunch of deprecation warnings are included in this commit, but I would just encourage that we put extra thought and care into reasoning through all the symbols and sub-sub-packages that are moving around...

FWIW I do agree with @mattpap that if we're going to do this, we need to do it quickly and ASAP.

bryevdv:
@pzwang no tests changed (except for internal imports) and the only examples changes were two or three in `glyphs` that I would not actually expect users to ever do (e.g. looping over the legend enum values). As I mentioned this is to get all the code that users *don't* use out of the way, which is why it basically didn't break anything in examples.

bryevdv:
@pzwang each of the commits is very self-contained and describes exacly what it did

bryevdv:
I will add that "power" users might use `bokeh.core.properties` but that's a good reason to move it now, since user models is still a new (not yet fully released) feature. If we don't move it now then `bokeh.properties` and friends should probably never be moved (but I'd really like to move them)

We can also add dummy modules that just make transitive imports and throw deprecation warnings for any of these that seem worrisome.

bryevdv:
Btw if I were to suggest any deprecation modules the list would be: `browserlib`, `templates`, `properties`, and `mixins`.

bryevdv:
I went ahead and added those modules with deprecation warnings.

bryevdv:
ping @pzwang @mattpap OK to merge?

bryevdv:
ping @birdsarah too since you have several open PRs

pzwang:
My initial concerns are resolved by your previous comment.  However, I haven't reviewed the code (but don't hold off merge on my account).

birdsarah:
Fine by me, although I haven't reviewed it (I should be able do that tomorrow). Most of my open PRs need work, so that's ok. If you wanted to merge #3369 before this one, that would be nice, but not that important.

bryevdv:
OK there seems to be general agreement to merge this and soon so I will merge as soon as Travis passes.

bryevdv:
It occurred that it wouldn't hurt for `plotting_helpers` to have a deprecation module too.

PR #3376, opened 2015-12-20
inspected is a non-serializable attr at least for now

bryevdv:
This was causing problems with `us_marriages_divorce` it seemed especially highlighted by line inspections. Some day might be nice to offer serializations of inspections as an option but this keeps the status quo working for now.


bryevdv:
ping @havocp @mattpap just FYI

PR #3380, opened 2015-12-21
Add simple examples using excel inputs

fpliger:


fpliger:
superseeding this PR and continuuing work on a clean one: #3507

PR #3387, opened 2015-12-21
Movies example

hsparra:
Recreate most of the movie reviews Shiny example using Bokeh.


bryevdv:
@hsparra I renamed the file to `main.py` which is the convention for "directory" apps, so now you can do, e.g. 
```
bokeh serve movies --show
```
I noticed in the data lots of big movies with zero dollars, is there maybe an issue with the data?

bryevdv:
Merging as is now to get in dev build tomorrow. Current appearance:

screen shot 2015-12-28 at 8 41 55 pm

PR #3391, opened 2015-12-21
Cleanup charts

fpliger:
Clean charts from:

* annoying console warnings
* flake warning


fpliger:
this should be ready now... I reverted the updated of `df.sort` to the new `df.sort_values` on the examples since I think it'll break for many users (it's better to have a short warning..). I could also do:

```python
            try:
                df = df.sort_values(by=columns, ascending=self.ascending)
            except AttributeError:
                df = df.sort(columns=columns, ascending=self.ascending)
```

but the drawback is that it'd make the examples more complex and I'm not sure it's worth it..

fpliger:
@bokeh/core feedback is welcome.. ^^^ 

I'll merge it later today if there are no objections

Issue #3270, opened 2015-12-08
general charts examples improvements


fpliger:
Collection of general improvements mentioned on merged PRs:
- [x] Rename `xyvalues` dataset in examples (It looks unusual, we should call these something else, everywhere. Even just generic `data` would be better.)
- [ ] simplify single examples
- [x] replace `OrderectDict`s with `dict`s
- [x] find a better data examples for `step`
- [x] `examples/charts/file/timeseries.py` overlaps with `stocks` example? (if so we should remove these)

jbcrail:
I've started work on the first and third tasks.

jbcrail:
Would this plot be a good replacement for the `step` example?
![postage_rates](https://cloud.githubusercontent.com/assets/6038/11959577/20438a86-a89c-11e5-9013-aa94c58bbe67.png)

bryevdv:
Yes that seems better to me.

fpliger:
> Yes that seems better to me.

+1

jbcrail:
Add to existing #3395, or create new PR?

bryevdv:
Seems to fall under a general charts improvement, here is fine.

jbcrail:
Regarding task 5, `examples/charts/file/stocks_timeseries.py` creates one plot that overlaps with the four in `timeseries.py`, so I'll remove this example.

Also, the `stock_timeseries` example downloads MSFT data, but never plots it. Should I remove the data or add it to the plot?

PR #3395, opened 2015-12-21
Improve general chart examples

jbcrail:
This only addresses the first and third tasks for #3270.


jbcrail:
I didn't rename `xyvalues` in `examples/charts/file/hover_span.py`, since it is exported by the `bokeh.sampledata.degrees` module.

Should I change the `bokeh.sampledata.degrees` module to export `data` instead? I hesitated because I wanted to initially limit the PR's scope to the `examples` directory.

bryevdv:
The sample data exists to support the examples so I'm fine with changing the name in `sampledata.degrees` to some thing less odd.

examples/app/fourier_animated.py
```diff
@@ -154,7 +152,7 @@ def create_centric_plot(foos, title='', r = 1, y_range=(-2, 2), period = pi/2, c
 cf2 = lambda x: (4*np.cos(3*x))/(3*pi)
 cf3 = lambda x: (4*np.cos(5*x))/(5*pi)
 cf4 = lambda x: (4*np.cos(7*x))/(7*pi)
-fourier = OrderedDict(
+fourier = dict(
```
bryevdv: This one might matter, I would double check that the example runs as expected. Or just leave it as-is for now, I am hoping to find time to try and pare down/simplify both fourier examples separately.

bryevdv:
@jbcrail for the tooltips the order matters. But you can get rid of the ordered dict, because the tooltips spec now accepts a sequence of tuples (so just assign the current argument to `OrderedDict` directly, don't pass it to a `dict`.

examples/charts/file/palettes.py
```diff
@@ -9,20 +7,20 @@
...
+                          ("Oranges9", Oranges9), ("PRGn9", PRGn9),
+                          ("PiYG9", PiYG9), ("PuBu9", PuBu9),
+                          ("PuBuGn9", PuBuGn9), ("PuOr9", PuOr9),
+                          ("PuRd9", PuRd9), ("Purples9", Purples9),
+                          ("RdBu9", RdBu9), ("RdGy9", RdGy9),
+                          ("RdPu9", RdPu9), ("RdYlBu9", RdYlBu9),
+                          ("RdYlGn9", RdYlGn9), ("Reds9", Reds9),
+                          ("Spectral9", Spectral9), ("YlGn9", YlGn9),
+                          ("YlGnBu9", YlGnBu9), ("YlOrBr9", YlOrBr9),
+                          ("YlOrRd9", YlOrRd9)])
```
bryevdv: order might matter here

examples/plotting/file/brewer.py
```diff
@@ -17,7 +15,7 @@
 df = df.set_index(['x'])
 
 def stacked(df, categories):
-    areas = OrderedDict()
+    areas = dict()
```
bryevdv: order probably matters here

examples/plotting/file/burtin.py
```diff
@@ -27,7 +26,7 @@
 Diplococcus pneumoniae,          0.005,      11,           10,       positive
 """
 
-drug_color = OrderedDict([
+drug_color = dict([
```
bryevdv: this is to plot the legend in the desired order

examples/plotting/server/fourier_animated.py
```diff
@@ -143,7 +141,7 @@ def create_centric_plot(foos, title='', r = 1, y_range=(-2, 2), period = pi/2, c
 cf2 = lambda x: (4*np.cos(3*x))/(3*pi)
 cf3 = lambda x: (4*np.cos(5*x))/(5*pi)
 cf4 = lambda x: (4*np.cos(7*x))/(7*pi)
-fourier = OrderedDict(
+fourier = dict(
```
bryevdv: same comment as about for other fourier example

examples/app/fourier_animated.py
jbcrail: I'll revert the change. I tested the examples and both worked correctly. But this could've just been a coincidence of key/value ordering.

jbcrail:
This now addresses all tasks in #3270 except the second one (simplify single examples).

bryevdv:
LGTM is this still WIP or is it ready?

jbcrail:
The PR is ready. I don't have any additional changes.

bryevdv:
Awesome! Just FYI there are status tags you can update to indicate this (I'll change this on)

jbcrail:
I wasn't aware that non-team members could update the labels. Thanks for the info!

examples/plotting/file/brewer.py
```diff
@@ -2,6 +2,8 @@
 import numpy as np
 import pandas as pd
 
+from collections import OrderedDict
```
fpliger: Why?

jbcrail: It must have slipped in during one of the several merges with master.

Issue #2321, opened 2015-05-20
Support a "Patch" with a hole in it / GIS support


birdsarah:
This is commonly needed in Geographic Information Systems (GIS), there may be other uses too.

Briefly mentioned this to @pzwang and @bryevdv and it was thought to be a good idea.

It's worth noting the GeoJSON spec for this exact problem:
## Polygon

http://geojson.org/geojson-spec.html#id4

Coordinates of a Polygon are an array of LinearRing coordinate arrays. The first element in the array represents the exterior ring. Any subsequent elements represent interior rings (or holes).

``` json
No holes:

{ "type": "Polygon",
    "coordinates": [
      [ [100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0] ]
      ]
   }

With holes:

{ "type": "Polygon",
    "coordinates": [
      [ [100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0] ],
      [ [100.2, 0.2], [100.8, 0.2], [100.8, 0.8], [100.2, 0.8], [100.2, 0.2] ]
      ]
   }
```
## MultiPolygon

http://geojson.org/geojson-spec.html#id7

Coordinates of a MultiPolygon are an array of Polygon coordinate arrays:

``` json
{ "type": "MultiPolygon",
    "coordinates": [
      [[[102.0, 2.0], [103.0, 2.0], [103.0, 3.0], [102.0, 3.0], [102.0, 2.0]]],
      [[[100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]],
       [[100.2, 0.2], [100.8, 0.2], [100.8, 0.8], [100.2, 0.8], [100.2, 0.2]]]
   ]
}
```

---

I'd be delighted to try and implement, but need some direction in terms of implementation.

Some thoughts:

I had initially imagined modifying Patch. However, Patch currently supports MultiPolygon's by separating the list of xs or ys with a NaN (https://github.com/bokeh/bokeh/blob/master/bokeh/models/glyphs.py#L604). If we were trying to follow the GeoJSON logic, and maintain easy backward compatibility if seems like we'd need some other type of separator to signify when it was an inner ring, and then the NaN would signify the start of a new Patch.....this feels like it's getting unwieldy.

[...]

PR #3399, opened 2015-12-22
Patches with hole

birdsarah:
Fixes: #2321


birdsarah:
test implementation to solicit feedback: #3183

birdsarah:
ping @bokeh/dev, this is ready for review.

birdsarah:
Details:

 -   patches can have a hole
 -   patches with hole can still be multipatches (ie separated with a NaN to make seperate but connected entities)
 -   basic hit testing works - but it captures the hit anywhere in the exterior ring (i.e. it doesn't respect the cut out)
 -   if you do a stroked patch with hole, it fills the non-cut out behavior as white - that seems correct to me.
 -   if you were to have an interior ring with a NaN in the middle e.g. [[1,2, NaN, 4],[2,3,4,1]] or [[1,2, NaN, 4],[2,3,NaN,1]] things would probably break down but that data doesn't make sense, so I think that's ok.

birdsarah:
One question: should I have also re-implemented `v_map_from_target` in LinearMapper (given that I updated `v_map_to_target`)?

pzwang:
Thank you thank you thank you!!

Can you elaborate on what this means?

    if you do a stroked patch with hole, it fills the non-cut out behavior 
    as white - that seems correct to me.

bokehjs/src/coffee/common/canvas.coffee
```diff
@@ -140,37 +140,34 @@ class Canvas extends LayoutBox.Model
...
-  # transform underlying screen coordinates to view coordinates
+  # Transform: underlying screen coordinates -> view coordinates
   sx_to_vx: (x) ->
-    return x
+    return @vx_to_sx(x)
 
   sy_to_vy: (y) ->
-    # Note: +1 to account for 1px canvas dilation
-    return @get('height') - (y + 1)
+    return @vy_to_sy(y)
```
bryevdv: I don't understand these. `sy_to_vy` means transform "screen" to "view" coordinates. But you are immediately calling `vy_to_sy` which means transform "view" to "screen" coordinates.

birdsarah:
@pzwang I mean

screen shot 2015-12-28 at 8 38 42 am

bokehjs/src/coffee/common/canvas.coffee
birdsarah: Yeah, there was a lot of duplicated code so I de-duped it.

This may not have been helpful?

birdsarah:
Actually, my comment was slightly wrong - it's white if it happens to be the first glyph, otherwise it follows the default behavior of being set to what the previous item's fill is (which I'm not sure seems right).

bryevdv:
> Actually, my comment was slightly wrong - it's white if it happens to be the first glyph, otherwise it follows the default behavior of being set to what the previous item's fill is (which I'm not sure seems right - but that is outside of this scope of this PR).

No, the is inconsistent with the interpretation and use of all everywhere else. it should not fill at all.

birdsarah:
> No, the is inconsistent with the interpretation and use of all everywhere else. it should not fill at all. 

Really? See issue: https://github.com/bokeh/bokeh/issues/3421

bryevdv:
> To clarify, the current behavior of Bokeh is: if fill_color=[None, 'pink'] the first glyph will fill white. This PR doesn't change that. 

OK, that's what was missing. I thought you were proposing a special case.

birdsarah:
I also thought I was proposing a special case as I hadn't understood the default behavior properly !!!!  I have updated the example to make the behavior clear.

screen shot 2015-12-28 at 10 03 39 am

birdsarah:
OK. So hit testing doesn't work on patches with holes, but still works on normal patches (including discontinuous). I'd like to put this up for merge and then work on hit testing later as an incremental fix.

bryevdv:
@birdsarah do you mean it does not work at all, or that it only works partially (e.g. ignoring the holes and only hit testing the outer right as you described earlier)?

birdsarah:
I mean it does not work at all. What it was doing before, actually brought up 4 copies of the hover tool, which seemed worse than nothing at all.

bokehjs/src/coffee/common/canvas.coffee
bryevdv: well my main concern is that it's now not comprehendible. What that code says to me is "map from screen to view by mapping from view to screen". I realize the actual transformations may be identical but in that case I'd factor it out into a third method that both other methods call. But these are also very low level, constantly used methods so adding more levels of indirection might have non-negligible performance costs overall.

bokehjs/test/test_common/test_canvas.coffee
```diff
@@ -0,0 +1,67 @@
...
+    expect(@cm.vy_to_sy(view_y)).to.eql(@height - view_y - 1)
+
+  it "v_vx_to_sx should return what it is passed", ->
+    xx = new Float64Array([72.72, 87.87, 224.24, 209.09])
+    expect(@cm.v_vx_to_sx(xx)).to.eql(xx)
+
+  it "v_vy_to_sy should make y relative to canvas height less 1 px", ->
+    view_yy = new Float64Array([10.0, 20.0, 30.0])
+    screen_yy = new Float64Array([@cm.vy_to_sy(view_yy[0]), @cm.vy_to_sy(view_yy[1]), @cm.vy_to_sy(view_yy[2])])
+    expect(@cm.v_vy_to_sy(view_yy)).to.eql(screen_yy)
```
bryevdv: The different 1px behavior x vs y seems like a bug.

birdsarah: This is not something I introduce it is a long standing part of canvas.coffee , I just added tests for it while I was changing nearby code.

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/common/canvas.coffee#L138
vs
https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/common/canvas.coffee#L135

bryevdv: Yes I'm just commenting, since I would have expected to see the dilation equally. We could probably investigate it later.

bokehjs/src/coffee/renderer/glyph/patches.coffee
```diff
@@ -5,12 +5,30 @@ hittest = require "../../common/hittest"
...
+
+    return _.flatten(flattened)
+
   _build_discontinuous_object: (nanned_qs) ->
     # _s is @xs, @ys, @sxs, @sys
     # an object of n 1-d arrays in either data or screen units
     #
     # Each 1-d array gets broken to an array of arrays split
-    # on any NaNs
+    # on any Nas
```
almarklein: ?

birdsarah: :)

birdsarah:
@bryevdv - this is ready.

Two things you requested are working:
* hit testing including not hitting on holes
* my canvas canges shouldn't be confusing anymore

I changed the data structure for declaring patches with holes, I think it makes it clearer overall.

bokehjs/src/coffee/mapper/linear_mapper.coffee
```diff
@@ -15,10 +17,7 @@ class LinearMapper extends Model
 
   v_map_to_target: (xs) ->
     [scale, offset] = @get('mapper_state')
-    result = new Float64Array(xs.length)
-    for x, idx in xs
-      result[idx] = scale * x + offset
-    return result
```
bryevdv: I'm very concerned about this. Not generating typed arrays whenever we can could have performance implications for *everything*. I would need to see some benchmarks before endorsing this as-is.

birdsarah:
I would say the following about this PR though:

1) there is no validation for potential bad data. For example, this is fine: 
```python
xs =[
  [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ] ], # single patch with a hole
  [1, 2, 3, NaN, 3, 4, 5], # discontinuous: filled patch and filled patch
  [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ], NaN, 3, 4, 5], # discontinuous: patch with a hole and filled patch
]
```

but this isn't

```python
xs =[
  [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ] ], # single patch with a hole
  [1, 2, 3, NaN, 3, 4, 5], # discontinuous: filled patch and filled patch
  [ 1, 2, [ [ 1, 2, 3 ], [ 3, 4, 5 ] ], NaN, 3, 4, 5], # this doesn't make any sense - but we don't check the data
]
```

FWIW this will render.

2) This PR does pave the way for full GeoJSON support which was my main motivation for getting this done.

3) This PR doesn't add any documentation - I would appreciate some guidance on this (or even a push from someone else) - where exactly under the patches glyph docstrings should i note it (at the top or under both xs and ys), and how to be concise (just point to example?). Should this be added to the user guide?

4) this doesn't have a snap feature - I can't remember why I set it up like that

5) this removes a number of Float64 Arrays - not sure if it'll be a performance hit, but am guessing that we can add them back, in certain places, in if we need them. 

All in all, I think this is a good foundation, and allows us to build proper GeoJSON support. I worry whether the data structure is confusing, although i do think it's the right one - my preferred way of making it less confusing would be API breaking:

what it is now:

    [ [ 1, 2, 3, NaN, 3, 4, 5 ] ], # discontinuous: filled patch and filled patch
    [ [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ], NaN, 3, 4, 5 ] ], # discontinuous: patch with a hole and filled patch

what I think would be less confusing

    [ [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ], NaN, [ [ 3, 4, 5 ] ] ] ], # discontinuous: patch with a hole and filled patch

or

    [ [ [ [ 1, 2, 3 ], [ 3, 4, 5 ] ], NaN, [ 3, 4, 5 ] ] ], # discontinuous: patch with a hole and filled patch

The reason this data structure is problematic is because you have different numbers of items that make up a polygon, a polygon with a hole will always be one item, whereas without a hole it can be 3+. I should add that on the flip side this is already true to some extent with patches because you can have a discontinuous list that is say [sqare, nan, circle, nan, triangle].

bokehjs/src/coffee/util/util.coffee
```diff
@@ -51,7 +51,35 @@ get_indices = (data_source) ->
...
     []
 
+map_vector_that_may_contain_patches_with_holes = (vector, map) ->
+  # Possible inputs are:
+  # [1, 2, 3]
+  # [1, 2, NaN, 3, 4]
+  # [[[1, 2], [3, 4]]]
+  # [1, 2, NaN, [[1, 2], [3, 4]]]
+  # Note: At the top level items are either a single value or an array of arrays
+  result = []
```
birdsarah: @bryevdv - regarding your comment elsewhere about Float64Arrays. The place to add them back is here. This function will get messy, but at least it's not duplicated anymore so hopefully contains the mess.

birdsarah: I'm working on this now.

bokehjs/src/coffee/mapper/linear_mapper.coffee
birdsarah: see below notes - i'm working on this.

bryevdv:
Should we just move to a completely different data structure for patches? All of those look brittle/confusing/hard to construct. Not surprising, we are trying to put ragged oddly dimensioned non-columnar data into columns. On top of that `NaN` and JSON is also very problematic. Should `Patches` and `MultiLine` simply be driven (always) by (only) GeoJSON data sources? We could do python-side conversions of old non-hole patches formats to GeoJSON for compatibility for some length of time. 

I'm just spitballing here. But it's hard to look at these nested nested nested lists and think it's something we will be happy to live with.

birdsarah:
> Should we just move to a completely different data structure for patches? All of those look brittle/confusing/hard to construct. 

There may be two different things. The data structure of patch with hole matches GeoJSON (http://geojson.org/geojson-spec.html#id4), which may look odd to you, but makes sense to me. So I wouldn't want to change that. Or more specifically, I'd want to be able to use that kind of data structure.

This is currently non-breaking so those using existing patches won't have any issues. We've had requests from people for this feature. Can we mark it as experimental (i.e. the API may change) and get feedback from users?

birdsarah:
I would agree that the `NaN` is a source of problems. We have discussed how we would like to handle something like a GeoJSON datasource in the future in a more elegant way that allows you to swap in and out a variety of data structures and they just serve themselves up as iterable to the GlyphRenderer.

I think that is out of scope for this PR.

bryevdv:
My concern is that people are now extremely sensitive to breaking changes, so adding something that we almost certainly know we won't want to keep as-is seems imprudent. 

GeoJSON supports patches with holes right? And GeoJSON is what people actually want? Then we should bite the bullet. I'd suggest:

* `Patches` be made to *only* work with a standard, bonafide GeoJSON format for multi-polygons as input 
* current "column format" be automatically concerned to the GeoJSON format on the python side (but only up to existing capability, i.e. no holes. You have to use the G true GeoJSON data format format to get holes)

And the same thing for `MultiLine`. Doing this means the linear mappers can be *greatly* simplified, too, because there could be a dedicated separate mapper just for GeoJSON inputs.

birdsarah:
> current "column format" be automatically concerned to the GeoJSON format on the python side (but only up to existing capability, i.e. no holes. You have to use the G true GeoJSON data format format to get holes)

Can you clarify this.  Do you mean "*coerced* to the GeoJSON format". Also it would really help if you could provide some if-this-then-coerce-to-that with written out lists as I think the potential for miscommunication is high.

birdsarah:
I think i've found a different mistake (checking now)

bryevdv:
> Also it would really help if you could provide some if-this-then-coerce-to-that with written out lists as I think the potential for miscommunication is high.

Sure. The only thing `Patches` supports now is "simple patches" and "disconnected patches", i.e.:

    x = [[1, 2, 3, 4], [1,2,3,4, nan, 5,6,7], ...]   
    y = [[1, 2, 3, 4], [1,2,3,4, nan, 5,6,7], ...]  

So, convert that much, and only that much, to GeoJSON. Not a GeoJSON expert but seems pretty trivial? The above would turn into a `Polygon` and `MultiPolygon`

We don't currently have any home-baked format for patches with holes and I am saying specifically that we should never add or support one. (To get patches with holes, must use GeoJSON format)

birdsarah:
Now i'm more confused...is there going to be a new glyph?

birdsarah:
I should say that I no longer think this is going in 0.11.1. I still haven't finished the remake of the happiness demo.

bryevdv:
> Now i'm more confused...is there going to be a new glyph? 

No, I am suggesting `Patches` be changed (in BokehJS) to only accept collections of GeoJSON `Polygon` and `MultiPolygon`. And that for purposes of easing migration, we convert the existing format to that automatically before/during serialization (with a deprecation warning)

birdsarah:
so if you use patches you have to use a GeoJSON datasource?

birdsarah:
(or will have to in the future)

bryevdv:
> so if you use patches you have to use a GeoJSON datasource?

Yes my impression is that pretty much everyone would rather use the popular standard format instead of the non-standard one we invented anyway? We can certainly have public functions that help create GeoJSON out of other formats.

birdsarah:
> Yes my impression is that pretty much everyone would rather use the popular standard format instead of the non-standard one we invented anyway? We can certainly have public functions that help create GeoJSON out of other formats. 

ok, that makes sense to me.

> I should say that I no longer think this is going in 0.11.1.

I think this is still true.

birdsarah:
I believe Float64Arrays are back everywhere they can be @bryevdv

birdsarah:
Here's a gif of it in action

patches with hole small

birdsarah:
@bryevdv I've realized that this could be merged as is. It's non-breaking. We can then build the GeoJSON conversion and deprecation warnings on top. I still have a number of questions about that, but this moves us forward.

birdsarah:
I'm going to open a new issue to lay out a plan for the GeoJSON-based Patches and we can decide if that needs to be part of this PR seperately.

birdsarah:
I've marked this as WIP because my proposal at #3708 would make this PR obsolete.

bryevdv:
> I've realized that this could be merged as is. It's non-breaking.  

It's not that its non-breaking. It's that if we merge this and people start using expanded home-grown format, then were are stuck *supporting it* and that is actually the main thing I am trying to avoid. Converting the existing "list of lists (possibly with nans)" format to GeoJSON doesn't seem like it would be too bad. If we merge this we will also have to convert the new expanded format as well.

birdsarah:
@bryevdv your last comment seems redundant / out of context given my last comment that I have proposal #3708 that would make this PR obsolete. Can you clarify?

bryevdv:
I couldn't tell if you had changed positions about merging as-is (so I was responding to it), or merely making a new observation.

birdsarah:
Yes, I've changed my position. 

I have a hunch that Patch, Patches, Line, MultiLine solve existing problems for people and their data structure is relatively simple, and fairly natural to understand. My current* position is that:

1) they stay as they are and we make new GeoJSON-centric glyphs `Polygon`, `MultiPolygon`, `LineString`, `MultiLineString` 
2) Of Patch, Patches, Line, MultiLine, I would guess that Patches is the least useful, but I wouldn't bet on it - we have loads of users we don't know about
3) I 100% agree with you on supporting a complex new data structure carefully
4) I think the change to Patches is too big and hence my proposal for four new geotypes
5) I am really stuck on this coordinate vs x y seperately thing - I don't have my head around it properly and would like to discuss.

* I say current because I don't have strong feelings about this and am learning a lot in this discussion.

birdsarah:
Totally seperate topic. I can imagine scenarios where you don't want to exclude the inner when you do a hover - should we add an option for this - or make another issue?

birdsarah:
I'm having this feeling like I've been mis-understanding what you're saying all along.

Are we saying that Patches can be "driven" by either a GeoJSONDataSource or a ColumnDataSource. If they're being driven by a GeoJSONDataSource then they understand how to do patches with holes. If they're being driven by a ColumnDataSource they understand how to do filled patches only?

bryevdv:
Closing per BEP (keeping branch for now)

birdsarah:
I have a patch of master (as of Apr 13, 2016) with this feature on in case anyone needs it. https://github.com/bokeh/bokeh/tree/birdsarah/patches_with_holes

PR #3400, opened 2015-12-22
Remove caching mechanism

damianavila:
Let's see how we perform without caching...


damianavila:
This is ready @bokeh/core... since @birdsarah fixes the flask dependency in another PR, this could be merge after her PR w/o further modifications...

bryevdv:
LGTM

PR #3410, opened 2015-12-23
Docfix/2448 TextInput widget callback description

### LINKED ISSUE #2448: Document when the callback occurs on the  TextInput widget.
Was with @brevdev working on a plot and had a notion that the TextInput callback happened onchange (i.e. everytime you enter a character) actually it works when you press enter or leave a cell.

Where should we put this kind of documentation? It's not going to be auto-generated.

canavandl:
issues: fixes #2448

Add callback attribute to TextInput child class (override from InputWidget parent), in order to specify when the callback is executed.

### DIFF
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -21,10 +21,6 @@ class InputWidget(Widget):
     Widget's label.
     """)
 
-    callback = Instance(Callback, help="""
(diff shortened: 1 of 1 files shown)


canavandl:
Hold before merge - bokeh apparently some machinery about overriding properties that I'm conflicting with.

PR #3411, opened 2015-12-23
Update dev_guide/setup.rst to align develop deps in line with scripts/devdeps.py

canavandl:
Update dev_guide/setup.rst to match dependencies required (and tested for) in scripts/devdeps.py

Note: I removed pdiff and boto from the listed deps in setup.rst because they're only used by Travis CI and not the developer. They're not tested for in devdeps.py because they're not required to develop/test locally. Removing them may be a point of debate.

### DIFF
--- a/sphinx/source/docs/dev_guide/setup.rst
+++ b/sphinx/source/docs/dev_guide/setup.rst
@@ -194,13 +194,11 @@ Testing dependencies include the following additional libraries:
 
 * beautiful-soup
 * colorama
-* pdiff
-* boto
-* nose
-* mock
-* coverage
-* websocket-client
+* pytest
+* pytest-cov
 * pytest-selenium >= 1.0
+* mock
+* websocket-client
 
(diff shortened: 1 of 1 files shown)


canavandl:
@damianavila - does the project still require pillow (PIL)? I see it in the meta.yaml file but didn't see it imported anywhere.

damianavila:
@canavandl pillow was used under the hood by some scipy function in AR... since AR was removed, I think we can delete that dependency...

bryevdv:
It's in the test dependencies, it might be  for the image diff machinery. @mattpap ?

mattpap:
`pdiff` and `boto` are used by `tests/travis/test`. They are mostly used on travis ci, but they can be useful locally as well. However, they are so rarely used locally, that, I think, there is no need to mention them, at least at the top-level.

bryevdv:
OK, I am going to go ahead and merge this as-is, then. I plan to re-arrange some of the dev guide and this is better to get in before.

PR #3417, opened 2015-12-25
Add detailed tutorial on individual charts components

rothnic:
@fpliger and I discussed that it would be good to go through and show how charts works by putting together some ipython notebooks. This serves three purposes:

1. Matures the component interfaces by using them on their own
2. Documents the components in an interactive way, useful for others creating new charts
3. Helps identify tests that should be added

At this point I've described some of the reasoning behind the data formats, which is a common question. Next I've outlined how the ChartDataSource, and AttrSpec works, then how they work together. While doing this, I've found small updates needed so things work as expected outside of the Builder. Some of these changes should help in providing more of the intended flexibility to manually override the attribute functionality in the case it isn't doing exactly what is wanted.

Questions:

1. What should this be called and where should it go?
2. Any other near term topics of high priority to be covered? (Plan to do one more on creating a custom builder)


rothnic:
Here is what the current notebooks look like:

[Deep Dive - Data](https://notebooks.anaconda.org/nroth/deep-dive)
[Deep Dive - Attributes](https://notebooks.anaconda.org/nroth/deep-dive-attributes)

rothnic:
@fpliger, might be of some interest to you. I added a notebook on composite glyphs, which tries to walk through the original intent behind having those as stand-alone units, that they should be directly useable.

I think this notebook highlights some issues with Chart and some things Chart could do better for this use case, but the idea is that you could do something like this eventually:

``` python

# 3 bars assigned to 2 labels, the two associated with a will stack
show(Chart(stack(BarGlyph('a', 1), BarGlyph('a', 2), BarGlyph('b', 1))))

# 2 bars at labels of income and profits, each using values from the associated column in df
show(Chart(df, BarGlyph('income', 'income'), BarGlyph('profits', 'profits')))

```

[Deep Dive - Composite Glyphs](https://notebooks.anaconda.org/nroth/deep-dive-composite-glyphs)

fpliger:
thanks @rothnic . I took a quick look a this PR a couple of days ago and definitely has useful info! You think it's ready?

datnamer:
Woah composite glyphs was exactly what I was looking for when I commented in #3632

fpliger:
>> A huge +1 from me to have some flexible intermediate (and hopefully statistical) abstraction just below charts but above everything else.

@datnamer sorry, I don't understand your comment.. can you be more specific? 

CompositeGlyphs are already a key component of the `Charts` design and charts already have a flexible intermediete (statistical) abstraction layer (that is maturing and evolving).. That's why I feel I'm missing something...

datnamer:
Misunderstanding on my part.

bryevdv:
So I'd like to merge this PR soon, we need to decide, do the notebooks 

* go in `bokeh-notebooks`, or

* under `examples/interactive` (pending examples re-org)

fpliger:
I must confess I'm keen on  `examples/interactive` ... it's more like a `tutorials` or `deep dives` . 

In general I think `bokeh-notebooks` would be a better fit in this case.

rothnic:
Yeah after the discussion about moving some of the examples/notebooks, I'm not sure where they should go. You'd kind of think to put them in bokeh-notebooks, but at the same time if it was just implemented in the sphinx documentation in user guides or something you wouldn't think twice. I guess that is what is difficult for me originally about #3664.

rothnic:
@bryevdv: I'll have the changes to charts ready tomorrow and I can move the notebooks if needed. I think I still want to make sure I have tests for the use cases outlined under the composite glyphs notebook.

bryevdv:
@bokeh/dev what about `examples/howtos` instead of `interactions` or `interactive` ? That would be sufficiently broad to cover things I have in mind, as well as connote a level of additional instruction or context.

fpliger:
yeap.. as mentioned elsewhere I'm +1 on `examples\howtos`

bryevdv:
I just merged the other PR, so at this point I would suggest `howto/charts` is the best place for the charts notebooks

PR #3418, opened 2015-12-26
Update toggle.coffee

gregnordin:
Change button active state based on directly on button state rather than the presence or absence of a CSS class.


bryevdv:
Thanks for the PR!

PR #3424, opened 2015-12-28
remove crossfilter example and associated bokeh code

brendancol:
Removing Crossfilter example.  Updated version will arrive in future PR.


bryevdv:
ping @mattpap I think this might have very minimal impact on your PR, I'm merging it so users don't see a broken example (we can add back a new working version that is a pure example after the freeze)

PR #3427, opened 2015-12-28
fixed json encoder import error

brendancol:
Fix for import error: (dateutil __init__ does not provide the submodule you have to import it explicitly)


bryevdv:
hrm why is it showing the whole file diffed?

brendancol:
looks like that file may have had different line endings or `^M` character carriage returns

PR #3439, opened 2015-12-29
Attempt to make widgets look a bit better

fpliger:


bryevdv:
ping @havocp @pzwang 

A very minimal PR to make the existing VBox/HBox layout less terrible will be good. My only question is about what interface to explore. We should only add attributes we expect to keep in the future (if we replace the implementations with PhosphorJS ones, e.g) If `padding` and `children_padding` make sense in that situation, great, let's keep them. Otherwise, just adding some reasonable default padding without any configurable interface is probably the best short term plan.

fpliger:
> We should only add attributes we expect to keep in the future (if we replace the implementations with PhosphorJS ones, e.g) 

That's a hard task since the difference between VBox/HBox and a potential PhosphorJS layout is huge... 
Wouldn't it be better to make "the best we can" on the current (VBox/HBox) and then just create a *new* layout instead of replacing the existing with the new, fancy one? This would be a smoother transition both for use and users..

fpliger:
Another option would be to not touch VBox/HBox at all and add a custom model that shows how to extend it... but this means that we are [now] forcing users to write JS/CS/whatever to make a decently nice layout. (and I don't think we should)

bryevdv:
I think we are definitely on different pages regarding what constitutes a smooth transition for users. There will *always* be a use-case for simple vbox/hbox layouts, so getting rid of them is not really an option. Replacing/renaming them is also becoming an increasingly bad option. I agree 100% with @havocp that making them better under the covers so all users just automagically get less crappy layout one day, without having to touch any code at all, is the best route to take. What I am looking for here is a small incremental improvement (just enough so that hbox'ed sliders aren't literally on top of one another, e.g) that is largely or completely invisible.

fpliger:
> I think we are definitely on different pages regarding what constitutes a smooth transition for users. There will always be a use-case for simple vbox/hbox layouts, so getting rid of them is not really an option. 

That's what I'm saying (probably bad english doesn't help)... HBox/Vbox are simple cases and phosphor layouts are a much rich and better option that can live in another new layout class. That what I meant with "instead of replacing the existing with the new"... I wasn't proposing any renaming/replacement


>  I agree 100% with @havocp that making them better under the covers so all users just automagically get less crappy layout one day, without having to touch any code at all, is the best route to take.

I agree here too. I just don't think we should do this by making the existing VBox and HBox become a Phosphor Layout automatigally... Or if we do we should eventually create a "SimpleBox" or somehthing to support the simpler case..

bryevdv:
> That's what I'm saying (probably bad english doesn't help)... HBox/Vbox are simple cases and phosphor layouts are a much rich and better option that can live in another new layout class. That what I meant with "instead of replacing the existing with the new"... I wasn't proposing any renaming/replacement

OK my apologies. But I do want to stress that even for simple HBox/VBox, Phosphor seems like it could provide immediate improvements if they were re-implemented using it. And then we can add additional new phosphor-based layouts too.

> I agree here too. I just don't think we should do this by making the existing VBox and HBox become a Phosphor Layout automatigally... Or if we do we should eventually create a "SimpleBox" or somehthing to support the simpler case..

The reason I think we should is because the existing simple implementations performs drastically differently across platforms. My interest in implementing VBox/HBox in Phosphor is in getting quality and consistency across platforms without have to duplicate all the effort that has gone into phosphor for just that.

fpliger:
@bryevdv ok. I think we have similar opinions on this.. the "problem" I see now is trying to find the minimal improvements for the release (because of the short time).

> OK my apologies. But I do want to stress that even for simple HBox/VBox, Phosphor seems like it could provide immediate improvements if they were re-implemented using it. And then we can add additional new phosphor-based layouts too.

> The reason I think we should is because the existing simple implementations performs drastically differently across platforms. My interest in implementing VBox/HBox in Phosphor is in getting quality and consistency across platforms without have to duplicate all the effort that has gone into phosphor for just that.

Right. That's why I think you are correct that we shouldn't change HBox/VBox too much now (new attrs, etc..) .. we could discuss this better if we had more time, but we don't. So deferring this important discussion to after the release seems the best option..

Will try to remove the new attrs and just add spacers to HBox/VBox that auto margin.. This should be better then what we have now. 

Also, I think that width attr on widgets would be helpful too

havocp:
Adding some default padding is probably safest. We don't really know what the future API will or should be.

fpliger:
This is how OHLC would like with a default "auto" marging between it's children

screen shot 2015-12-29 at 11 53 28 am

and this is how it looks like setting the HBox width to 1000:

screen shot 2015-12-29 at 11 53 56 am


Would this be a "good" default? I prefer auto then fixed spacing but it may be a personal bias :)

bryevdv:
I would prefer the first one more if it left justified everything instead of right justified, but honestly either one is an improvement. It would be good to survey a few other apps to see how they look as well.

fpliger:
> I would prefer the first one more if it left justified everything instead of right justified, but honestly either one is an improvement. 

@bryevdv both are possible with the changes in the PR. THe first one is the "default" when you don't specify any width for the HBox (so it expands on the full page) the second one is with `HBox(..., width=1000)`

> It would be good to survey a few other apps to see how they look as well.

yeap. working on it..

brendancol:
@fpliger I'm excited for this.  any opinion on adding a more general `css_properties` to apply arbitrary css to the element...something like:
```
css_properties = Dict(String, String, default=None)
```

```
  render: () ->

    ...

    css_props = @mget("css_properties")
    if css_props? then @$el.css(css_props)

    ...

```

fpliger:
> @fpliger I'm excited for this. any opinion on adding a more general css_properties to apply arbitrary css to the element...something like...

I think we have discussed this in the past with @mattpap and @bryevdv .. IIRC the problem here was choosing to add the "right thing" or the "right spelling". I'm open to such improvements ... if we get a team consensus otherwise we may risk to regret having to remove it later. Maybe we can add this with a big warning about this "experimental" feature.

bryevdv:
I'm not sure about this PR, this is purely a stopgap until we do real work after `0.11`. As @havocp notes we don't know yet what the right interface, which to means we should really only do whatever is possible without modifying the interface at all (otherwise risk adding something, only to immediately remove/break it)

fpliger:
Yup.. that is how it is ATM. Does not change the interface at all. We can keep this PR as is (with just HBox/VBox changes) and add any other thing (if any at all) on other PRs

gregnordin:
As a user I'm all for adding some sort of default margins to VBox/HBox. As Bryan mentioned, I would prefer left justified auto margins, but either way is an improvement. For the future, I am really looking forward to add Phosphor.

fpliger:
quick thought: #3447 could be used to get the same effect but it's probably worth it to also have a better default... so should be better to also go ahead with this one

bokehjs/src/coffee/widget/vbox.coffee
```diff
@@ -23,9 +24,20 @@ class VBoxView extends ContinuumView
...
     height = @mget("height")
-    if height? then @$el.css(height: height + "px")
+    if height?
+      @$el.css(height: height + "px")
+      t = (children.length-2)*2
+      spacer_height = height/(children.length*2)
+    else
+      spacer_height = 20
+
+    spacer = '<div class="bk-vbox-spacer" style="height: ' + spacer_height + 'px;"></div>';
```
mattpap: This should use string interpolation, or better `$('<div>').addClass('bk-vbox-spacer').css({height: spacer_height})`.

bokehjs/src/coffee/widget/vbox.coffee
```diff
@@ -23,9 +24,20 @@ class VBoxView extends ContinuumView
     width = @mget("width")
     if width? then @$el.css(width: width + "px")
     height = @mget("height")
-    if height? then @$el.css(height: height + "px")
+    if height?
+      @$el.css(height: height + "px")
+      t = (children.length-2)*2
```
mattpap: Leftover `t`.

PR #3443, opened 2015-12-29
turn off sending defaults to BokehJS

bryevdv:
Defaults should now all be in sync. We will only send theme values and user-set values to BokehJS.

Let's see how Travis, the report, and some hand testing look.


bryevdv:
OK, report looks good, all server examples working as expected. Merging so we can easily get more testing.

PR #3447, opened 2015-12-30
Support custom css injection and css classes on widgets

fpliger:


bokeh/resources.py
```diff
@@ -439,7 +447,7 @@ def render_css(self):
         return CSS_RESOURCES.render(css_raw=self.css_raw, css_files=self.css_files)
 
 
-class Resources(JSResources, CSSResources):
+class Resources(CSSResources, JSResources):
```
mattpap: Why? Alphabetic order?

bokeh/resources.py
```diff
@@ -425,10 +425,18 @@ class CSSResources(BaseResources):
...
+        super(CSSResources, self).__init__(mode='inline', version=version,
+            root_dir=root_dir, minified=minified, log_level=log_level,
+            root_url=root_url, path_versioner=path_versioner)
+        self.custom_css_files = [] if custom_css_files is None else custom_css_files
+
     @property
     def css_files(self):
         files, _ = self._resolve('css')
-        return files
+        return files + self.custom_css_files
```
mattpap: You should use reuse `_resolve()` instead. If uses wants `inline` resources, everything should be inlined. Furthermore, notebook and server support is required. Also we need to decide what to do in CDN case.

bokehjs/src/coffee/util/dom_util.coffee
```diff
@@ -8,5 +8,11 @@ waitForElement = (el, fn) ->
       fn()
   interval = setInterval(handler, 50)
 
+# Add css_classes to an element
+add_css_classes = (el, css_classes) ->
+  for css_class in css_classes
+    el.addClass(css_class)
```
mattpap: You can add multiple classes with `addClass()`, no need for extra layer of indirection.

bokehjs/src/coffee/widget/slider.coffee
```diff
@@ -16,6 +17,7 @@ class SliderView extends ContinuumView
     html = @template(@model.attributes)
     @$el.html(html)
     @render()
+    dom_util.add_css_classes(@$el, @model.css_classes())
```
mattpap: This line should be in `render()`, otherwise `change` event will remove custom classes (see line 14).

mattpap: However, as `css_classes` is a property of `Widget`, this logic should have been implemented in `WidgetView`. There is no such class, so you would have to implement one, make widget views inherit from it, rename rendering functions, etc.

bokehjs/src/coffee/widget/vbox.coffee
```diff
@@ -35,11 +40,15 @@ class VBox extends BaseBox.Model
   defaults: ->
     return _.extend {}, super(), {
       children: []
+      css_classes: []
```
mattpap: You don't have to override this. Unless inheritance chain is broken, then it should be fixed instead.

bokeh/models/widgets/widget.py
```diff
@@ -7,8 +7,11 @@
...
 
 from ...core.properties import abstract
+from ...core.properties import String, List
 from ..component import Component
 
 @abstract
 class Widget(Component):
     """ A base class for all interactive widget types. """
+
+    css_classes = List(String)
```
mattpap: It would be convenient to allow space separated strings as well. Use `.accepts(String, lambda str: [ s.trim() for s in str.split("\s+") ])`.

mattpap: Also, not really a `String`. There should be `CSSClass` property descriptor implemented on top of `Regex`.

mattpap:
I doubt this will be really useful in this form, because (cross-compatible) CSS is rarely enough to style things properly (you have to change HTML as well). You can play with `content`, but that's rarely cross-compatible. One would have to have access to the template as well, but that is very close to just implementing custom models (which, btw., will have templating capability added at some point). Also, ad-hoc styling with raw CSS, without (at least) guidelines or an "API", will make things very fragile and version dependent (not that we didn't do this already with raw JS), but with CSS is even worse (in my opinion).

bokeh/resources.py
fpliger: No, this is just a leftover... shouldn't be changed and I'll fix that.

bokeh/resources.py
fpliger: I had mixed feelings yesterday about this  so it's here for now.. If user wants `inline` and wants to inject it's own css that in being server from another (his own?) CDN or whatever, why we should not support it?

bryevdv:
As much as I'd like to offer some improvement in this release, I think this is too rushed. There are probably better ways to inject resources through the `Application` machinery, so I am worried about doing things to stable interfaces like `output_file` that will be inconsistent with future work. I'd support #3439 for some extra default padding and a quick `0.11.1` after some deliberation.

bokehjs/src/coffee/widget/vbox.coffee
fpliger: No... chain is fine, this is a leftover and will be fixed

bokehjs/src/coffee/widget/slider.coffee
fpliger: Right... I was not sure about both things. tnx

bokeh/models/widgets/widget.py
fpliger: yeap.. need better thinking here.

fpliger:
@mattpap I disagree about this being useful or not. Our widgets look and feel relly sucks and users only have really limited and not "easy to pick" solutions for that... It may not be the best approach as it's really not thought or carefully designed at all due to the time but we should have some support for this kind of feature somehow. If you have specific ideas in mind please share them somewhere and we can pursue/discuss that path. As I said, this is really a last minute try and don't have much thought at all..

I agree with @bryevdv that it's too rushed and it's probably not a good idea to change a stable API with changes that are very likely to change anyway... This was an attempt and I think we should try to support something like this (talking about feature that let users be able to specify custom style and add classes to widgets and not the API/implementation) but thi should not be forced "from nothing" in the freeze day. I'm +1 about leaving this as just some exploration work and have a better shared solution for `0.11.1`

PR #3449, opened 2015-12-30
Fix unclosed file in resources _inline()

humm:
Raises ResourceWarning in python 3.4.


bryevdv:
Thanks for the PR @humm !

PR #3452, opened 2015-12-30
Fix FontSizeSpec DeprecationWarning

humm:


humm:
Raise a DeprecationWarning. 
Could have been done with a simple list too. I don't know what is the canonical way.

bryevdv:
Now I am wondering why I am not seeing these warnings. @humm what version of python are you using?

humm:
Well, I am using 3.4.4 from pyenv under OS X, but I am using it from jupyter, and loading bokeh with a hosts of other packages, some of which I suspect are forcing these warning to not be ignored (DeprecationWarning are ignored by default, ResourceWarning are ignored unless you built python in debug mode (which is not my case), and you can change that with https://docs.python.org/3.4/library/warnings.html#warnings.filterwarnings).

bryevdv:
Ah Jupyter, that makes sense. Although I did think we recently forced turning on deprecation warnings. @mattpap ?

mattpap:
Yes, deprecation warnings are forced visible, if only they use `bokeh.deprecation` machinery. If they use custom code, as it happens a lot and in the case of `FontSizeSpec` specifically, then no luck. I think we can make fixing this a pre-release task, despite freeze (this should be a non-invasive change).

bryevdv:
@mattpap yes we can consider that, I will go ahead and merge this fix now, though, since it is appropriate in any case.

bryevdv:
Thanks @humm !

PR #3474, opened 2015-12-31
Airports Example: fixed panning/zooming and refactored to use figure

brendancol:
fixed panning/zooming and refactored to use figure instead of plot object for simplier code example


examples/glyphs/airports_map.py
```diff
@@ -42,12 +38,4 @@
...
 
-doc = Document()
-doc.add_root(p)
-
-if __name__ == "__main__":
-    filename = "airports_map.html"
-    with open(filename, "w") as f:
-        f.write(file_html(doc, INLINE, "Bokeh Airports Example"))
-    print("Wrote %s" % filename)
-    view(filename)
```
mattpap: `examples/glyphs` use the verbose API, so this should be restored. Otherwise you can move it to `examples/plotting/file`. Actually, it used to use a mixture of both, so just do the later.

brendancol:
ah good call...moving now..

examples/plotting/file/airports_map.py
```diff
@@ -22,13 +17,14 @@
 tile_source = WMTSTileSource(**tile_options)
 
 # set to roughly extent of points
-x_range = Range1d(start=airports['x'].min() - 10000, end=airports['x'].max() + 10000)
-y_range = Range1d(start=airports['y'].min() - 10000, end=airports['y'].max() + 10000)
+x_range = Range1d(start=airports['x'].min() - 10000, end=airports['x'].max() + 10000, bounds=None)
+y_range = Range1d(start=airports['y'].min() - 10000, end=airports['y'].max() + 10000, bounds=None)
```
bryevdv: Just to be clear, this is a workaround, not a fix? The plot should work with auto-bounds turned on? (Or alternatively: some plots should have a way to disable/express incompatibility with auto-bounds) I'd say a workaround is OK for the release, but let's make an issue to invetigate further  ping@birdsarah

bryevdv:
restarted test (spurious failure I think)

bryevdv:
Merging, leaving associated issue open to investigate root cause further.

examples/plotting/file/airports_map.py
birdsarah: I guess my question would be why would you want to pan far beyond the edge of the data? (Apart from the fact that this is a demo about the tile renderer) I have been assuming that having something that prevents you panning beyond the edges of your data is useful. This does raise an interesting idea for a feature though to express bounds as a multiple. Anyway, it's fixed for now. Happy to open a new issue to discuss.

bryevdv: I believe this is just a workaround to a bug. With this tile renderer the plot tools behave badly in conjunction with bounds.

PR #3485, opened 2015-12-31
Remove legacy charts from example notebooks

jbcrail:
I removed examples using Blaze and lists as a data source. I also upgraded from the legacy charts.

I renamed one file to stick with the naming convention used by the other notebooks.

Fixes part of #3390


bryevdv:
Thanks @jbcrail !

PR #3486, opened 2016-01-01
Bring GridToolManager up to date

### LINKED ISSUE #3469: Can't turn wheel zoom tool back on
ping @birdsarah 

Seems especially true in grid plots but maybe other too. After turning it off, it cannot be re-activated.

If this is a one-line fix it could be entertained for `0.11`

bryevdv:
issues: fixes #3469 

Missed some fixes and changes applied to ToolManager. This fixes two small issues 1) the inability to swtich wheel zoom back on mentioned in #3469 and also 2) wheel zoom should be off by default.


birdsarah:
Thanks @bryevdv!

PR #3489, opened 2016-01-01
fix 'over-render' due to hover inspections

### LINKED ISSUE #3467: Hover tool performance seems degraded
ping @birdsarah I wonder if this is related to the "over render()" you noticed.

A *very* short and self-contained fix coud potentially be entertained. 

```
examples/plotting/file/texas.py:
 hover tool performance seems degraded / laggy
```

bryevdv:
issues: fixes #3467 

@birdsarah this was a pretty simple one: we added a `request_render` on `inspect` events. But the code that triggered `inspect` events was mistakenly looking for `null` to indicate no hits, not an "empty" selection object. So it was triggering every hit-test, even if there was no result. 

There's a slight complication in that we do need to trigger the *first* empty inspection, so e.g., that hover tools turn off when they are supposed to. 

This stops all the over-rendering. I will say it did not improved the `texas.py` hover as much as I thought maybe it should. There is still a slight lag, but maybe there always was, and I am just imagining that it was better. In any case, I do think this should be merged. I tested it behaving correctly with every hover and hover glyph example, including server ones.

### DIFF
--- a/bokehjs/src/coffee/common/selection_manager.coffee
+++ b/bokehjs/src/coffee/common/selection_manager.coffee
@@ -11,6 +11,8 @@ class SelectionManager extends HasProperties
     super(attrs, options)
     @selectors = {}
     @inspectors = {}
+    @empty = hittest.create_hit_test_result()
+    @last_inspection_was_empty = false
 
   serializable_in_document: () -> false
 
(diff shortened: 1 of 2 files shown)


bryevdv:
OK, it's still slow because it is re-rendering on every successful hit test. But it doesn't need to do that either, unless there is a hover glyph. Will fix that as well.

bryevdv:
OK, performance is back to what I expect!

PR #3494, opened 2016-01-01
Notebooks working with server

bryevdv:
Currently gets server notebooks working minimally, with caveats:

* users must all extra ws origins (noted in notebooks)

* users must explicitly push session to "default" (so multiple notebooks clobber)

If this can be improved upon *very* simply, it should be. 

Associated issue: #3461


bryevdv:
Sorry for stupid commit history, forget I had already pushed before a local rebase.

bryevdv:
Marking "notebook + server" a known issue, closing this PR (will make a better one after release)

PR #3497, opened 2016-01-01
Docs/Examples tweaks for 0.11 (3)

bryevdv:


bryevdv:
@jbcrail FYI for future reference right before release we typically have single PRs to collect docs edits, etc. in order to not cause a CI logjam right before release. I will go ahead and merge this PR.

PR #3499, opened 2016-01-01
Update resources.py to not leak paths

### LINKED ISSUE #3488: bokeh.resources.INLINE.render_js() unnecessarily exposes minor info about server in HTML (0.11.0.dev7)
Hello, I am using bokeh.resources.INLINE.render_js() to embed bokeh's required JS resources.
I've noticed quite a few comments of the form:
````
/* BEGIN /path/to/bokeh_js_file.js */
\<js code here\>
/* END /path/to/bokeh_js_file.js */ 
````

This path exposes things like the username of the computer, path to python libraries, whether or not a virtual env is used, etc... None of this is a major security concern per se, but I don't think it should be available in the HTML for the world to see. It looks like this is done by the _inline method BaseResources (https://github.com/bokeh/bokeh/blob/master/bokeh/resources.py#L305), so I assume the same problem occurs with bokeh.resources.INLINE.render_css().

Can we have this disabled by default?

### PR DESCRIPTION
**Pull request by bryevdv**
issues: closes #3488

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -16,7 +16,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 import copy
-from os.path import join, relpath
+from os.path import basename, join, relpath
 import re
 
 from . import __version__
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bokeh/resources.py [hunk 1]
```diff
@@ -303,13 +303,13 @@ def _resolve(self, kind):
         return (files, raw)
 
     def _inline(self, path):
-        begin = "/* BEGIN %s */" % path
+        begin = "/* BEGIN %s */"
```
mattpap: You should remove `%s` as well (the same below). Preferably use `os.path.basename(path)`, so that we know what component we're dealing with.

bryevdv: good catch, thanks

mattpap:
LGTM

PR #3503, opened 2016-01-02
Fix spelling and other issues in docs

jbcrail:
Fixes part of #3497.


bryevdv:
@jbcrail FYI for release we usually try to all collaborate on one docs/examples PR so that the CI machinery does not go into a logjam, I will open a new one soon.

PR #3504, opened 2016-01-02
Fix selection histogram layout

fpliger:


examples/plotting/server/selection_histogram.py
```diff
@@ -58,7 +58,15 @@
 vh1 = pv.quad(left=0, bottom=vedges[:-1], top=vedges[1:], right=vzeros, alpha=0.5, **LINE_ARGS)
 vh2 = pv.quad(left=0, bottom=vedges[:-1], top=vedges[1:], right=vzeros, alpha=0.1, **LINE_ARGS)
 
-layout = vplot(hplot(p, pv), hplot(ph, Paragraph()))
+# NOTE: Version 0.11 has introduced auto spacing by default on VBox/HBox/vplot/hplot
+#       so for now we must tweak some spacing and borders to have it closely
+#       aligned.
```
bryevdv: Would say:
```
An improved auto layout algorithm will be coming in 0.11.1 but for now, settings by 
hand to get plots more closely aligned.
```

PR #3507, opened 2016-01-02
Add examples with excel files

fpliger:
issues: none


fpliger:
setting it to ready since I'm not sure I'll be able to make the changes I wanted  early today.. so others can look at it and provide their first impressions

damianavila:
this one needs xlrd as a testing dependency in conda.recipe/meta

PR #3525, opened 2016-01-04
Random Tiles Example: Temporary fix for bounds issue on map

brendancol:


mattpap:
@brendancol, is this related to some issue? Otherwise, what change in behavior should we observe?

examples/app/random_tiles/main.py
```diff
@@ -24,7 +25,9 @@ class RandomTileSource extends MercatorTileSource
 
 def create_plot():
     axis_range = [-10000000, 10000000]
-    p = Figure(tools='pan,wheel_zoom', x_range=axis_range, y_range=axis_range, plot_height=800, plot_width=800)
+    x_range = Range1d(start=axis_range[0], end=axis_range[1], bounds=None)
+    y_range = Range1d(start=axis_range[0], end=axis_range[1], bounds=None)
+    p = Figure(tools='pan,wheel_zoom', x_range=x_range, y_range=y_range, plot_height=800, plot_width=800)
```
bryevdv: wouldn't `p.x_range.bounds = None`, etc.  work with the original version?

bryevdv: well, I think so but not worth holding up this merge.

PR #3532, opened 2016-01-04
Add support for inline autoload and restore support for inline resources in notebook

mattpap:
Inline resources and autoload is a slight non-sense, but technically it works and it's simpler to restore inline resource support in notebook this way, rather than adding a big if statement. This PR implements raw css support to make inline resources work, thus completes implementation of autoload.


damianavila:
@mattpap this one will fail after merging 3547, there is a conflict, otherwise, LGTM, tested OK. I am merging 3547 now, can you fix the conflict here so I can merge this one as well?

mattpap:
You shouldn't have merged that one, I just fixed the problem here.

damianavila:
Uppss... I misunderstood you... testing this on top of master (and not on top of Bryan's PR). Sorry for the confusion.

damianavila:
If I try BOKEH_RESOURCES=INLINE jupyter notebook on any notebook with this PR I get the `Javascript error adding output!` error...

damianavila:
Just for the record, it should be `inline` lowercase, not uppercase...

PR #3534, opened 2016-01-04
Improve initialization of plot size

### LINKED ISSUE #3468: embed_responsive_with_height.py starts out wrong size
ping @birdsarah @almarklein 

After resizing the window, the plot "pops" to a much larger size then is responsive

```
examples/embed/embed_responsive_with_height.py:
 starts out small and "pops" to full size on browser window resize
```

almarklein:
issues: fixes #3468

The problem described in #3468 was related to a resize event being handled before the plot was fully initialized. This is why there was a `this.el.parentNode is null` error in the console, causing the initial sizing to fail. The solution is to test for this situation and schedule a new event in a future event-loop iteration.

### DIFF
--- a/bokehjs/src/coffee/common/plot.coffee
+++ b/bokehjs/src/coffee/common/plot.coffee
@@ -455,6 +455,16 @@ class PlotView extends ContinuumView
     # user-configurable in the future, as it may not be the right number
     # if people set a large border on their plots, for example.
 
+    # We need the parent node, because the el node itself collapses to zero
+    # height. It might not be available though, if the initial resize
+    # happened before Bokeh got properly initialized. If that happens, we
+    # will try again, but only once.
+    if not @.el.parentNode and not @_re_resized
+      that = this
+      setTimeout( (-> that.resize_width_height(use_width, use_height, maintain_ar)), 10)
+      @_re_resized = true
+      return
+
     avail_width = @.el.clientWidth
     avail_height = @.el.parentNode.clientHeight - 50  # -50 for x ticks
     min_size = @mget('min_size')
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/common/plot.coffee
```diff
@@ -455,6 +455,16 @@ class PlotView extends ContinuumView
     # user-configurable in the future, as it may not be the right number
     # if people set a large border on their plots, for example.
 
+    # We need the parent node, because the el node itself collapses to zero
+    # height. It might not be available though, if the initial resize
+    # happened before Bokeh got properly initialized. If that happens, we
+    # will try again, but only once.
+    if not @.el.parentNode and not @_re_resized
+      that = this
+      setTimeout( (-> that.resize_width_height(use_width, use_height, maintain_ar)), 10)
```
mattpap: You should use fat arrow (=>) instead `that = this` JS idiom.

mattpap:
>  before the plot was fully initialized

I would say simply before added to the DOM. Otherwise you can't take measurements.

bokehjs/src/coffee/common/plot.coffee
```diff
@@ -455,6 +455,16 @@ class PlotView extends ContinuumView
     # user-configurable in the future, as it may not be the right number
     # if people set a large border on their plots, for example.
 
+    # We need the parent node, because the el node itself collapses to zero
+    # height. It might not be available though, if the initial resize
+    # happened before Bokeh got properly initialized. If that happens, we
+    # will try again, but only once.
```
mattpap: Can it be unsuccessful after one retry? If so, what then?

almarklein:
> Can it be unsuccessful after one retry? If so, what then?

@mattpap I think it should be successful in one try, as everything should be settled in the next event loop iteration. But indeed we cannot know how that works out in the future. I updated to try multiple times as needed with increasingly large intervals.

mattpap:
LGTM

mattpap:
Is this intended for 0.11?

bryevdv:
> Is this intended for 0.11?

No

PR #3545, opened 2016-01-05
Hotfix set range defaults to ``None``

birdsarah:


damianavila:
Previos CI run worked except from integration, after the fix, integration worked as well, so merging now...

PR #3550, opened 2016-01-05
Small fix in scatter.ipynb - allow making selection

almarklein:
There was one plot in this example that specified a `nonselection_fill_color`, but did not allow a way to make a selection. I assumed the intention was to demonstrate how scatter glyphs behave under a selection, so I added the lasso tool.


damianavila:
LGTM, self-contained and tested. Merging.

PR #3552, opened 2016-01-05
Use self.mode instead of mode in BaseResources.__init__()

mattpap:
This won't allow `BOKEH_RESOURCES=INLINE` anymore.

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -212,28 +212,28 @@ class BaseResources(object):
         self.components = components if components is not None \
             else ["bokeh", "bokeh-widgets", "bokeh-compiler"]
 
-        self.mode = settings.resources(mode)
-        self.root_dir = settings.rootdir(root_dir)
-        self.version = settings.version(version)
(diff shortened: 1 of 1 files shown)


bokeh/resources.py
```diff
@@ -212,28 +212,28 @@ def __init__(self, mode='inline', version=None, root_dir=None,
...
-        self.version = settings.version(version)
-        self.minified = settings.minified(minified)
-        self.log_level = settings.log_level(log_level)
-        self.path_versioner = path_versioner
+        self.mode = settings.resources(mode);           del mode
+        self.root_dir = settings.rootdir(root_dir);     del root_dir
+        self.version = settings.version(version);       del version
+        self.minified = settings.minified(minified);    del minified
+        self.log_level = settings.log_level(log_level); del log_level
+        self.path_versioner = path_versioner;           del path_versioner
```
bryevdv: are the `del`s just insurance against accidental use of function arguments?

mattpap: Yes.

PR #3558, opened 2016-01-06
Use correct order of div and script in notebook_div template

### LINKED ISSUE #3540: graphs are not displayed when using runipy
I am trying to plot graphs though bokeh .It runs fine when i run it in jupyter and shows proper output.But when i run ipynb file using runipy though terminal it runs fine but output is  not displayed(i.e graphs not displayed)

mattpap:
fixes #3540

### DIFF
--- a/bokeh/core/_templates/notebook_div.html
+++ b/bokeh/core/_templates/notebook_div.html
@@ -8,7 +8,7 @@ Renders a ``div`` to display a Bokeh model into a Jupyter Notebook.
 :type plot_div: str
 
 #}
+{{ plot_div|indent(4) }}
 <script type="text/javascript">
   {{ plot_script|indent(2) }}
 </script>
-{{ plot_div|indent(4) }}


mattpap:
It's interesting to note that before this PR notebooks worked but didn't work after nbcoverting them to static HTML. This is because in a notebook, HTML is added dynamically and JS is evaluated asynchronously, so even that HTML was added after the script, it happened to be added before it was needed. This is not the case in static HTML, where the script is evaluated synchronously, so the div must be added before it in the HTML structure. I'm not sure if notebooks can fail without this PR, but I can imagine this happening. For now I didn't observe any failures however.

damianavila:
code LGTM, currently testing...

mattpap:
To add to our confidence, this template is now exactly what we have in `file.html`. I will make sure for the next release, that there is only one template to get the job done.

bryevdv:
OK, thanks for the extra information. Yes, for the future we could look to ways to reduce the number of injection code paths (ideally to just one)

damianavila:
From Bryan at FD:

>I'd be ok with 3558 with the following assurances: run cells one at at time by hand run cells with "run all", save executed notebook and reload, test uploaded notebook to anaconda.org, test nbconvert by hand

I have tested all of this successfully... so I am inclined to merge it, please @bokeh/core raise you voice if you want to stop this, otherwise I will merge in a little bit

PR #3560, opened 2016-01-06
Fix interactive excel example

fpliger:
changing the plots on the right box


mattpap:
I wanted to try this out, but `xlwings` is Windows/OSX only, at least according to pip.

fpliger:
@mattpap yes, unfortunately it is.. :(

damianavila:
OK, merging this one because it is self contained and testes by @fpliger

PR #3568, opened 2016-01-07
Use split not contains for browser compatibility

birdsarah:


birdsarah:
Hi @bokeh/core, I've made this PR against a branch `docs_updates/0.11.0`. My proposal is this:

- we keep this branch docs_updates/0.11.0 and we use it to make ongoing improvements to the docs site that is applicable to 0.11.0
- everything in this branch will also be in master
- but not everything in master will be in this branch

My hope is that this will allow us to make ongoing improvements to things like the user guide while development continues.

birdsarah:
I should add that I'm happy to take ownership of PRs from this branch into master.

In addition, I would imagine that we would be regularly deploying the docs manually from this branch, and each time overriding the tag (I'm not sure how to do this, but I believe you are @damianavila @bryevdv) to 0.11.0.

When 0.11.1 comes out, we would remove this branch and make a new one at the point of the release commit called `docs_updates/0.11.1` etc.

damianavila:
The proposal sounds good to me... anyone can re-deploy but you can ping me and I can do it when you need it...

birdsarah:
I would be more than happy to be another person who knows how to do it, so there are more available bodies.

bryevdv:
+1

birdsarah:
I have tested the change on Safari/Chrome/FF (OSX).

bryevdv:
@birdsarah so this PR is ready to merge into the branch? Should I go ahead and do that or do you want to?

PR #3572, opened 2016-01-07
Changed range update cb_obj to range object from model object and updated examples

### LINKED ISSUE #3173: cb_obj for range callback is plot not range
See: https://github.com/bokeh/bokeh/blob/2819b349ae47232c72a275d4e7714adf329c7235/bokehjs/src/coffee/common/plot.coffee#L207

The callback is called with `@model` instead of with `rng` which, for example, does not have the property `start` so you can't do `cb_obj.get('start')`.

canavandl:
issues: fixes #3173 

To keep with the pattern that the plot object that triggers the callback is injected into the js closure as the cb_obj, this PR makes the range update callback inject the range and not model.

I changed the examples I was aware of. Let me know if there are others.


bryevdv:
need to add a 0.11.1 release notes and add this change under migration notes.

bryevdv:
restarting test in case it was just a spurious travis ci thing

canavandl:
ping @bryevdv 

Added release notes and mention of change under migration notes.

PR #3574, opened 2016-01-07
Create info command

### LINKED ISSUE #3506: "bokeh static" command
Should be a command to report the location of static assets (e.g, to help with configuring `nginx`, etc)

Maggie-M:
issues: closes #3506


mattpap:
@bryevdv, can this be renamed to `info` (or similar)? I would prefer `bokeh static` to be reserved for a static file server.

bryevdv:
In this case I suggest in this case let's generalize `bokeh info`, e.g.:

    $ bokeh info
    bokeh version: 0.11.0
    python version: 3.5
    static paths: << path to static files >>
   ...

And then specifically:

    $ bokeh info --static
    << path to static files >>

mattpap:
Yes. I wanted to suggest this as well.

PR #3588, opened 2016-01-08
quick add of demo apps to gallery

bryevdv:


birdsarah:
I shouldn't have accepted this. Making a new PR with some changes on the same branch.

PR #3594, opened 2016-01-09
support for generating apps from Jupyter Notebooks

### LINKED ISSUE #3593: Add option to parse Jupyter notebooks as input to bokeh command
Should only require extracting the python source lines from `.ipynb` files and feeding them as-is to existing machinery.

bryevdv:
issues: closes #3593

Needs: 
* [x] small refactor to share code
* [ ] more interesting demo notebook than `sliders.ipynb`
* [ ] docs support


havocp:
Is it useful to support main.ipynb in an app directory ?

If writing a notebook to be used in this way, should the notebook contain output_notebook or not? may have to make it a no-op rather than warn about it?

bryevdv:
> Is it useful to support main.ipynb in an app directory ?

I think so

> If writing a notebook to be used in this way, should the notebook contain output_notebook or not? may have to make it a no-op rather than warn about it? 

I would say ignore it and `show` as well (possibly a `DEBUG` statement mentioning they are being discarded) since I imagine people might want to prototype in an actual running notebook . Another nice feature (future) suggested a long time ago by @hhuuggoo would be to monkeypatch Jupyter's `interact` and auto-generate a widget layout similar to what would be in the notebook.

bryevdv:
I'm actually somewhat of the mind that we should just quietly ignore all of those things in general (i.e., `DEBUG` rather than `WARN`) Are there instances where it is not obvious that is the "right thing"?

havocp:
I was just thinking people might cargo cult them in there and confuse themselves expecting them to do something. maybe downgrade all the warnings to debug logging ?

havocp:
right which is what you suggested :-) I think debug is better than totally silent, but debug makes sense. could consider adding push/pull session while we are at it

bryevdv:
I'd consider a little less stern sounding INFO to be ok too.

bryevdv:
@havocp OK I made them log info. The machinery will have to be made a little more sophisticated to log `push_session` and `pull_session` since they are not in `bokeh.io` (or could make transitive imports there). 

Let me know what you think about the rest, I factored most of the code into a `CodeHandler` that accepts just source (and code runner only accepts source now as well). File reading is pushed off to the subclasses.

bokeh/application/handlers/code.py
```diff
@@ -0,0 +1,102 @@
...
+
+import os
+
+from bokeh.io import set_curdoc, curdoc
+
+from .code_runner import _CodeRunner
+from .handler import Handler
+
+class CodeHandler(Handler):
+    """ Run source code which modifies a Document i
```
bryevdv: typo fixed locally, not pushed yet

bokeh/application/handlers/code_runner.py
```diff
@@ -2,23 +2,22 @@
...
 from types import ModuleType
-import codecs
 import os
 import sys
 import traceback
 import uuid
 
 class _CodeRunner(object):
-    """Load and run a Python file."""
+    """ Compiile and run a Python source code."""
```
bryevdv: typo fixed locally, not used yet

bokeh/application/handlers/tests/test_notebook.py
```diff
@@ -0,0 +1,41 @@
...
+        result = {}
+        def load(filename):
+            handler = NotebookHandler(filename=filename)
+            handler.modify_document(doc)
+            result['handler'] = handler
+            result['filename'] = filename
+        _with_script_contents(source, load)
+
+        assert result['handler']._runner.path == result['filename']
+        assert result['handler']._runner.source == "\n# coding: utf-8\n"
```
bryevdv: not sure if this test could be better, or if there is a better test. Just trying to make sure it's using the file that it's supposed to, and calling `nbformat.read` on it as well.

bokeh/application/handlers/tests/test_script.py
```diff
@@ -20,68 +20,19 @@ def with_file_object(f):
...
-            handler.modify_document(doc)
-        _with_script_contents("raise RuntimeError('nope')", load)
+            result['filename'] = filename
+        _with_script_contents(source, load)
 
-        handler = result['handler']
-        assert handler.error is not None
-        assert 'nope' in handler.error
+        assert result['handler']._runner.path == result['filename']
+        assert result['handler']._runner.source == source
```
bryevdv: Just trying to make sure it uses the right file, not re-test `CodeHandler` and `CodeRunner` but if there are other tests that are useful please suggest.

bokeh/application/handlers/directory.py
```diff
@@ -14,12 +14,14 @@ def __init__(self, *args, **kwargs):
...
 
         src_path = kwargs['filename']
-        mainpy = os.path.join(src_path, 'main.py')
-        if not os.path.exists(mainpy):
-            raise ValueError("No 'main.py' in %s" % (src_path))
+        main = os.path.join(src_path, 'main.py')
+        if not os.path.exists(main):
+            main = os.path.join(src_path, 'main.ipynb')
+            if not os.path.exists(main):
+                raise ValueError("No 'main.py' or 'main.ipynb' in %s" % (src_path))
```
havocp: worth complaining if the app has both main.py and main.ipynb? "main.ipynb ignored" kind of message

bokeh/command/util.py
```diff
@@ -38,7 +38,10 @@ def build_single_handler_application(path):
     if os.path.isdir(path):
         handler = DirectoryHandler(filename=path)
     else:
-        handler = ScriptHandler(filename=path)
+        if path.endswith(".ipynb"):
+            handler = NotebookHandler(filename=path)
+        else:
```
havocp: might be good to insist the script ends in ".py"

bokeh/application/handlers/tests/test_script.py
havocp: testing that some "typical" notebook (which would include output_notebook and show) works end-to-end wouldn't hurt, I don't think ... though it could probably be a CodeHandler test instead, the only notebook specific thing right now would be having the output_notebook and show in there, but CodeHandler also allows that.

fpliger:
@bokeh/core what about merging this PR and adding better docs and example on following PRs?

PR #3618, opened 2016-01-11
Improve how webgl canvas is blitted into the visible canvas

### LINKED ISSUE #3605: WebGL AA is broken on Firefox on Windows
The borders in the webgl rendering of both lines and circles look terrible. It looks like something is wrong with the anti-aliasing. The problem is visible on Firefox on Windows. It looks fine on Edge and Chrome. 

~~This looked good the last time I checked.~~ I *think* this looked good the last time I checked. Maybe I broke it with the introduction of a fix for IE.

bokehaa02
bokehaa01

almarklein:
issues: fixes #3605

Previously, we blitted without interpolation. Apparently this causes problems for firefox. Now we do it with interpolation, but align the image such that the pixel correspondence is one-to-one, and the resulting visualization is crisp.

### DIFF
--- a/bokehjs/src/coffee/common/plot.coffee
+++ b/bokehjs/src/coffee/common/plot.coffee
@@ -417,16 +417,20 @@ class PlotView extends ContinuumView
     @_render_levels(ctx, ['image', 'underlay', 'glyph', 'annotation'], frame_box)
 
     if ctx.glcanvas
-      # Blit gl canvas into the 2D canvas. We need to turn off interpolation, otherwise
-      # all gl-rendered content is blurry. The 0.1 offset is to force the samples
-      # to be *inside* the pixels, otherwise round-off errors can sometimes cause
-      # missing/double scanlines (seen in Chrome).
-      for prefix in ['image', 'mozImage', 'webkitImage','msImage']
-         ctx[prefix + 'SmoothingEnabled'] = false
+      # Blit gl canvas into the 2D canvas. We set up offsets so the pixel
+      # mapping is one-on-one and we do not get any blurring due to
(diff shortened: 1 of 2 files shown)


almarklein:
I verified this on Windows. Would be good if others can test this on Linux and OSX.  To do so run this code:
   
    import numpy as np
    
    N = 1000
    x = np.random.normal(0, np.pi, N)
    y = np.sin(x) + np.random.normal(0, 0.2, N)
    
    p = figure(webgl=True)
    p.circle(x, y, radius=0.5)
    output_file("scatter_test.html", title="scatter test")
    show(p)

... and inspect the result. You should see a bunch of circles, and they should not look jaggy nor blurry:
bokehaa03

almarklein:
You can use this static HTML file to check. It shows the webgl canvas below the actual plot. Both plots should be equally sharp. https://dl.dropboxusercontent.com/u/1463853/bokeh/scatter_test.html

bryevdv:
LGTM (as far as I can tell) on OSC Chrome. Note, it seemingly has some problem masking the data. If I zoom in the circles start disappearing when part of them still would be overlapping the screen area.

almarklein:
> If I zoom in the circles start disappearing when part of them still would be overlapping the screen area. 

Strange, I don't see that. Does the circle disappear when its center goes off the viewport? Maybe its a driver bug/feature. OpenGL does drop vertices like that, you'd expect that for sprites the pointsize is taken into account, but maybe not all drivers do that.

bryevdv:
> Does the circle disappear when its center goes off the viewport? 

Can't tell exactly but that at least roughly describes what I see.

bryevdv:
Perhaps we can look into it, but separately.

Issue #3606, opened 2016-01-11
Bokeh 11 Barchart : Need a way to keep the original x labels order / to disable the lexical order sort


FabriceSalvaire:
I am using Bokeh to plot a kind of precomputed barchart/histogram where I give as input the x labels and their corresponding counts:

```
from bokeh.charts import Bar, output_file, show
data = {
    'labels': ('EN', 'PD', 'AD'),
    'counts': (10, 20, 30),
}
p = Bar(data, values='counts', label='labels')
output_file("bar.html")
show(p)
```

Unfortunately Bokeh 11 sort the x labels in lexical order, while I want to keep the given order. It was not the case for Bokeh 10.

Other minor issue: I have to clear the x and y label which are in English and irrelevant in my case "sum(counts)".

Maggie-M:
moved to [bkcharts repo](https://github.com/bokeh/bkcharts/issues/47)

Issue #2924, opened 2015-09-29
Turn off Bar Plot sorting by category axis labels


tvaught:
Bar Chart data is being sorted against my wishes. The bars should probably retain the ordering of the column that is assigned to the label attribute (the DataFrame ordering).  Perhaps a `sort` attribute could be added in the event sorting is desired.

![image](https://cloud.githubusercontent.com/assets/110926/10176201/f7b45e4e-66b8-11e5-8908-5f4d63c3d43b.png)

tonyfast:
I think this is related to #2935

bryevdv:
@rothnic any suggestions on a spelling for this? I imagine the implementation will not be so difficult?

tvaught:
How about a `sort_labels=False|'ascending'|'descending' ` argument to the `Bar` call?  Not sure why `labels` was chosen as a change from `cat` (or, better yet, `categories`, but I'll trust that there was a good reason.

euri10:
I faced the same issue and "hacked" it with while it's being investigated
```python
plt.x_range = FactorRange(factors=df['label'].tolist())

**Comment by rbinnun2** [comment]
Thanks euri10, setting the x_range does the trick.

btw, the current sorting is causing a FutureWarning: 

'''
C:\Program Files\Anaconda\lib\site-packages\bokeh\charts\_attributes.py:78: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df = df.sort(columns=columns)
'''

**Comment by spawnrider** [comment]
Just had the same warning that  @rbinnun2 :+1: 
```
d:\dev\python34\lib\site-packages\bokeh\charts\_attributes.py:78: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df = df.sort(columns=columns)
```

rothnic:
There are a number of places that you would want or not want to sort with charts. On one hand, you could say just pre-sort the data, on another you'd like to specify it through the chart interface in some way because you may want to assign colors in one order, then maybe the same data is used for the labels and you want that sorted in a different way. So it would be difficult to have a data structure passed in that communicates both cases. 

So I do think we want some specification on the input, and we have that built-in to the underlying components, but it needs to be exposed through each chart interface that is more pertinent to the specific chart.

I'll also look at fixing the warning.

ahothan:
I have the same issue for ordering bars in each group:
https://github.com/bokeh/bokeh/issues/2935

[...]

PR #3623, opened 2016-01-11
Added unsorted_label parameter to chart builder to don't sort labels

FabriceSalvaire:
This pull request solves the issue #3606.

I don't know if it is the right way to achieve this but it works.

Example to test:

    from bokeh.charts import Bar, output_file, show
    data = {
        'labels': ('EN', 'PD', 'AD'),
        'counts': (10, 20, 30),
    }
    p = Bar(data, values='counts', label='labels', unsorted_label=True)
    output_file("bar.html")
    show(p)

### DIFF
--- a/bokeh/charts/builder.py
+++ b/bokeh/charts/builder.py
@@ -230,6 +230,10 @@ class Builder(HasProps):
         using the valid input for the `HoverTool` tooltips kwarg.
         """)
 
+    unsorted_label = Bool(default=False, help="""
+        A boolean flag to tell to don't sort the labels.
+        """)
+
     def __init__(self, *args, **kws):
         """Common arguments to be used by all the inherited classes.
 
@@ -340,6 +344,9 @@ class Builder(HasProps):
                     attributes[attr_name].iterable = custom_palette
                 else:
                     attributes[attr_name] = self.default_attributes[attr_name]._clone()
+
+        if kws.get('unsorted_label', False):
+            attributes['label'].sort = False
 
         # make sure all have access to data source
         for attr_name in attr_names:


bryevdv:
I think I'd like this better if were something like `label_sort=None`, with the reason that it would potentially be nice to supply a function to do the sorting (*ideally* that could be deferred to the browser). Adding an `unsorted_label` bool flag would easily be inconsistent with that kind of future work (i.e., of we add this now and `label_sort` later, what meaningful interpretation could be given to the combination `label_sort=mysort, unsorted_label=True`?)

fpliger:
We have been using `AttrSpec`s for all charts attributes so I think we should stick with these... That's actually how you control ordering for other attributes like `color`, `group`, `stack`, etc.. I'd actually expect `label` to accept a a `CatAttr` as argument (just like the other attributes) but it doesn't. So I'd fix this behaviour instead of adding another special argument for this special attribute...

fpliger:
@FabriceSalvaire actually it does work out of the box.. you just need to explicitly pass both `label` and `values` arguments. So in your case you can do with just:

```Python
from bokeh.charts import Bar, output_file, show
from bokeh.charts.attributes import  CatAttr
data = {
    'labels': ('EN', 'PD', 'AD'),
    'counts': (10, 20, 30),
}
label = CatAttr(df=df, columns='labels', sort=False)
p = Bar(data, values='counts', label=label)
output_file("bar.html")
show(p)
```

FabriceSalvaire:
I felt this solution, then it is just a documentation issue. I agree we cannot add a parameter for each particular case.

Actually the chart API is fine for usual cases, but I think it would be nice to have an intermediate level between a low and high level API. I don't know how to use the low level API to make complex custom plots and  for the chart API the user guide presents a kind of processing on top of Panda data frame.

For example there are many ways to plot an histogram:
* from precomputed data
* using weighted counts
* using fixed / variable width binning
* with vertical bar 
* or like a polyline or with x/y error bar, in this case we interpret the histogram as a 1D function
* etc.

I believe there is no limitation in the BokehJS side, it is more a matter of examples.

fpliger:
@FabriceSalvaire yes... totally agree it's a documentation issue. We already have this marked and are working on it. There is an open PR (#3417) that is tackles part of this work of generating documentation content and another PR to integrate mode content directly in the docs should be landing soon.. Any suggestions or contributions on this side are very very welcome. :)

Regarding the general concept of charte being high level and target for the usual use cases... that is. It is actually not an easy problem to try and cover all different kinds of inputs (and data shapes) and offer deep level of customization (like ordering labels, grouping, stacks, markers, colors, etc.. ) without blowing an API. My (very opinionated in this case :) ) opinion is that the new `Charts` design does do a good job about offerening a considerable vast level of expressiveness, maintaining still an fairly limited API. That said, we aim to make it more simple and at the same time powerful on every iteration..

Finally, regarding mid level API, plotting should be exactly what you are talking about. :)

fpliger:
closed as feature is already build in and we just need to work on better docs

b-trav:
@fpliger - your code snippet above does not work for me.

If I execute it, I get "NameError: name 'df' is not defined"

If I replace df=df with data=data, I get:

"ValueError: expected an instance of type ColumnDataSource, got {'counts': (10, 20, 30), 'labels': ('EN', 'PD', 'AD')} of type dict"

b-trav:
For a code snippet that does work check out @fpliger's earlier answer in this post:

https://github.com/bokeh/bokeh/issues/2924

PR #3630, opened 2016-01-11
Make annulus glyph work on IE too

### LINKED ISSUE #3543: Annulus glyph does not render correctly in IE and Edge
Instead of a "donut" it shows an opaque circle that seems to correspond to the inner circle of the donut. Can be seen in e.g. `plotting/notebooks/glyphs.ipynb` and `plotting/file/glyphs.py`.

Confirmed on Win7, Win10 / IE 11, Edge

This might very well be a bug/feature of the IE canvas engine. Can we work around it?

almarklein:
issues: fixes #3543

IE was confused about what part of the annulus should be filled, which granted, is not an easy task :)  By drawing the glyph in two parts, IE knows how to do it too.

To test, check e.g. `examples/glyphs/glyphs.py`


bryevdv:
merge master to try and get tests passing

almarklein:
ah, thanks. Travis is green.

bryevdv:
Unfortunately it looks like this can leave artifacts:

screen shot 2016-01-14 at 10 30 21 am

(from the the test report at ttps://s3.amazonaws.com/bokeh-travis/0.11.0-c312cf8/plotting/file/glyphs.png)

almarklein:
Arg, I had such artifacts before, but they are gone on the browsers that I tested with the current implementation. Mmm, I'll try a bit further ...

bryevdv:
Well, perhaps it's a phantomJS issue then, if we can demonstrate that it actually works across the real major platform browsers I think it would be OK to proceed.

almarklein:
Here's a link to a static plot with the new implementation. You can zoom out to get more overlap to validate if that works out correctly: https://dl.dropboxusercontent.com/u/1463853/bokeh/annulus.html

I verified that it works correctly on Win10 with FF, Chrome, Edge, IE.

bryevdv:
Still visible on OS X / safari:

screen shot 2016-01-15 at 5 10 49 pm

Starting to wonder if  browser detection (as much as that sucks) is the way to go.

almarklein:
Mmm.... I tested on Linux with Chrome and FF. And also on a OSX VM with a (rather old version of) Safari. All look fine. So it looks like it breaks for the newer version of Safari.

How hard is it to detect either Safari or IE?

almarklein:
@bryevdv I added browser detection for IE. Can you try again (same link)?

bryevdv:
Yup looks perfect on OSX/safari now

almarklein:
Great. Ready to go then, from my end.

PR #3637, opened 2016-01-12
0.11.0 Documentation fixes and flake8

canavandl:
issues: fixes 3014

Three things:
- ~~Refactor examples and chart builders to use pandas.sort_values instead of deprecated pandas.sort~~
- Change BoxPlot docstring reference section (remove broken Wikipedia image link and replace with link)
- Remove extra line return in Donut docstring

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ branches:
 
 before_install:
     - export PATH="$HOME/miniconda/bin:$PATH"
+    - export PATH="$HOME/build/bokeh/bokeh/bokehjs/node_modules/phantomjs/lib/phantom/bin/phantomjs:$PATH"
     - echo $TRAVIS_BUILD_NUMBER > __travis_build_number__.txt
     # To actually run the examples you need to include the [ci enable examples] in your commit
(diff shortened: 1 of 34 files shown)


bryevdv:
I think the pandas change forces users to have  the very most recent pandas, which is too new for many people.  I think we will just have to live with the warnings for awhile.

birdsarah:
if this is really 0.11.0 fixes - should it go into `docs_updates/0.11.0` and get deployed?

canavandl:
that's fine

birdsarah:
i'd love to get @bryevdv's thoughts on this as there's a bunch of non docs commits in here too.

bryevdv:
`docs_updates/0.11.0` will be merged in to master fairly soon, I'm not sure it makes a difference either way unless the docs changes here are critical to re-deploy.

canavandl:
These don't really change any doc content, only styling (and silences a ton of the sphinx build warns). They can probably be merged separately

bryevdv:
@birdsarah @canavandl so this should just get merged to master?

birdsarah:
Yes, master.

PR #3640, opened 2016-01-12
Add new waterfall chart

fpliger:


bryevdv:
Can you attach a screenshot?

fpliger:
screen shot 2016-01-12 at 6 41 09 pm

@bryevdv ^^^  .. This still mostly playing (thus WIP)... probably can avoid adding a new chart Glyph.. and avoid some of the things done on the builder.

Also, it'd be nice to have annotations but there is some work to do on that side first.

bryevdv:
Closing per BEP (leaving branch)

PR #3645, opened 2016-01-13
add height and width to list of deprecated Chart attributes

### LINKED ISSUE #3578: Sizing plots, charts: Inconsistency between doc an api
Hi I encountered the following issue while migrating from v0.10 to v0.11

In v0.10:
one could size a ```Chart``` or a ```figure``` with the parameters ```width``` and ```height```.

In v0.11:
a ```figure``` will still size according to ```width``` and ```height``` and also according to ```plot_width``` and ```plot_height```)
a Chart won't size anymore according to those, even though the documentation still mentions these attributes (http://bokeh.pydata.org/en/latest/docs/reference/charts.html#core-classes). It will size according to ```plot_width``` and ```plot_height``` (as mentioned here: http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#dimensions)

Questions:
* Is it a doc issue? or height/width in Charts still serve a (different?) role
* Should this be added to the migration guide?
* Corollary question: what is the purpose of having two different set of parameters for what seems to be the same purpose?

fpliger:
Issues: fix #3578


bryevdv:
Fixed the conflict. LGTM

PR #3646, opened 2016-01-13
Fix bad path to bokeh logo static

canavandl:
Potential solution for broken bokeh master:

The url arg in the ImageURL glyph test example in examples/glyphs/glyph.py is wrong and gives a 404 warn. It's unclear why the tests seem to pass for py27 and not py34 though

I'm unable to reproduce the failed test w/ py34 locally, so if the hotfix doesn't work I'll add a print statement on the stdout of glyph.py to see what response TravisCI is getting that I'm not.


canavandl:
The tests passed and the code change definitely won't break anything, so I'm going to merge.

Issue #3651, opened 2016-01-13
Unifying Bokeh and BokehJS project structure


canavandl:
Proposed
* [x] Remove ar (AbstractRendering) #3663
* [x] Remove HasParent class (move display_defaults into defaults)
* [x] Make BokehJS class structure match Bokeh class structure
* [x] Make BokehJS project structure match (python file glyphs.py -> glyphs/ dir on bokehjs)
* [x] Move "Spec" handling code "down" from `GlyphRenderer`
* [x] Align JS and python model properties (private JS props explicitly stored)

canavandl:
Bokeh class structure:

* HasProps
  * Model
    * Component
      * Plot
      * Widget
        * specific widgets
    * Glyph
      * specific glyphs...
      * Marker
        * specific markers...
    * Renderer (is just an abstract base class)
      * Annotation
        * specific annotations...
      * GuideRenderer
        * Axis
        * Grid
    * TileSource
      * specific tiles...
    * Tool
      * specific tools...

BokehJS class structure:

* HasProperties
  * Plot (not subclass of Component)
  * Component
    * Widget
      * specific widgets
  * Glyph
    * specific glyphs...
    * Marker
      * specific markers...
  * Annotation
    * specific annotations...
  * GuideRenderer
    * Axis
    * Grid
  * TileRenderer
    * specific tile stuff...
  * Tool
    * ButtonTool
      * ActionTool
        * specific tools
      * GestureTool
        * specific tools
      * InspectTool
        * specific tools

Proposal -

1. Add `Model` as abstract base class of `HasProperties` to BokehJS
2. Remove `Renderer` abstract base class in Bokeh (have subclasses just inherit from `Model`)
  * To me, `Renderer` doesn't add much clarity over `Model` and only causes the code to diverge (i.e. is an Annotation a model or a renderer?)
3. Is it worth unifying the `Plot` and `Component` relationship?

bryevdv:
Some possible observations: 

* rename JS `HasProperties` to `HasProps`
* Renderer` could be used to identify anything that draws to the canvas
* `Component` could be used to identify anything that could be layed-out (as DOM elements)
* There's more to the data sources hierarchy?
* There's also things like tickers and tick formatters

bryevdv:
I would say the `HasProperties` -> `HasProps` rename, and adding `Model` base to BokehJS are not problematic or controversial, and could happen immediately as their own PR. In general we should do this work incrementally in any case.

[...]

PR #3672, opened 2016-01-17
Fix base of BokehJS class hierarchy 

bryevdv:
This PR adds `models/model` and `models/component` and has public facing BokehJS models inherit from one or the other. 

This implements a small part of #3651

### DIFF
--- a/bokehjs/src/coffee/api.coffee
+++ b/bokehjs/src/coffee/api.coffee
@@ -19,9 +19,6 @@ module.exports = (Bokeh) ->
   Bokeh.Random        = require("./common/random")
   Bokeh.SVGColors     = require("./common/svg_colors")
 
-  # models
-  Component = require('./models/component')
-
   # mappers
   Bokeh.LinearMapper      = require("./mapper/linear_mapper")
   Bokeh.LogMapper         = require("./mapper/log_mapper")
--- a/bokehjs/src/coffee/api.coffee
+++ b/bokehjs/src/coffee/api.coffee
@@ -19,6 +19,9 @@ module.exports = (Bokeh) ->
   Bokeh.Random        = require("./common/random")
   Bokeh.SVGColors     = require("./common/svg_colors")
 
+  # models
+  Component = require('./models/component')
+
   # mappers
   Bokeh.LinearMapper      = require("./mapper/linear_mapper")
   Bokeh.LogMapper         = require("./mapper/log_mapper")
--- a/bokehjs/test/test_common/test_document.coffee
+++ b/bokehjs/test/test_common/test_document.coffee
@@ -2,7 +2,7 @@ _ = require "underscore"
 {expect} = require "chai"
 utils = require "../utils"
 
-HasProperties = utils.require "common/has_properties"
+Model = utils.require "models/model"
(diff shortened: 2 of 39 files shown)


bryevdv:
Weird this builds locally for me after a `gulp clean` will look into Travis failure when I get to the office.

bokehjs/src/coffee/api.coffee
```diff
@@ -19,6 +19,9 @@ module.exports = (Bokeh) ->
   Bokeh.Random        = require("./common/random")
   Bokeh.SVGColors     = require("./common/svg_colors")
 
+  # models
+  Component = require('./models/component')
```
mattpap: This is unnecessary. `Component` is unlikely to be used by the user.

bryevdv: Agreed, I will remove it

mattpap:
It fails for me locally as well.

mattpap:
You should import from `../models/model` (lower case) in `source/data_source.coffee`.

bryevdv:
Yah, trying to reproduce on clean checkout

mattpap:
No need to do a clean checkout, just `rm -rf bokehjs/build`. I don't use `gulp clean`, but it looks like it doesn't clean enough.

bryevdv:
Why was that every working for me? Grr... also do you know any way to get better error messages when there are errors like that?  and definitely seems like `gulp clean` does not work correctly

bryevdv:
Oh, it looks like `gulp clean` only works at the top level:

    BUILD_DIR = if typeof argv.buildDir == "string" then argv.buildDir else "./build"

mattpap:
> Why was that every working for me?

It depends on the exact workflow. Maybe you just did `gulp dev-build`, but used later used minified version of bokeh (invalid at this point). Or caching, or ...

mattpap:
> Oh, it looks like gulp clean only works at the top level:

So, actually `gulp clean` works for me. As to the top-level thing, there is silent assumption that gulp is run from within `bokehjs/` directory.

bryevdv:
I do `gulp build` from further in the tree all the time, I had assumed `gulp clean` worked too, since it "worked"

bryevdv:
OK this PR seems pretty self-contained I think I'd rather merge this and new start new work on top of it. @mattpap is there anything else you would think should go in this PR before merging?

bokehjs/test/test_common/test_document.coffee
```diff
@@ -455,7 +455,7 @@ describe "Document", ->
     # double-check different results if we do include_defaults
     parsed_with_defaults = JSON.parse(d.to_json_string(include_defaults=true))
     attrs = parsed_with_defaults['roots']['references'][0]['attributes']
-    expect('tags' of attrs).to.be.equal true
+    #expect('tags' of attrs).to.be.equal true
```
mattpap: Why? I presume will be fixed in a further PR?

mattpap:
LGTM

mattpap:
> I do gulp build from further in the tree all the time (...)

I didn't realize `gulp` allows to do this. I just tried this and `gulp clean` sill works for me from e.g. `bokehjs/src` or deeper.

bryevdv:
OK well it was just a speculation looking at the code, maybe I just thought I had `gulp clean` but really didn't. Anyway, merging now.

Issue #3688, opened 2016-01-20
Docs ToDos


birdsarah:
Cleaning up from: https://github.com/bokeh/bokeh/issues/3404

Missing content problems (this list needs re-checking):
- [x] "to be addeds" on dev_guide/notes
- [x] /user_guide/server.html - placeholder section titles
- [x] /user_guide/cli.html - remove mention of things not yet implemented!
- [ ] squished images http://bokeh.pydata.org/en/dev/docs/user_guide/notebook.html
- [ ] empty pages (or nearly empty):
  - [ ] http://bokeh.pydata.org/en/latest/docs/reference/server.html
  - [ ] http://localhost:5009/en/latest/docs/user_guide/tutorials.html

New content: 
- [ ] add an example for custom tiles to the user_guide geo page

Misc:
- [ ] left collapse docs menu
- [ ] subtle "top" to scroll to top would be helpful
- [ ] ~~automate the ALL_VERSIONS used for dropdown~~ (right now it adds the previosu version, but should be the latest one of any minor version)
- [ ] search bar on search page to match
- [ ] display searched for text in search bar
- [ ] use responsive plots
- [ ] userguide/tutorials is undewhelming (and wierd after learning more)
- [ ] too much white space on right side of gallery pages (due to no left menu)
- [ ] gallery design

Questions:
- [ ] what do we want the search facets to be (can be changed after final deploy)
- [ ] reference/core just has a bunch of stuff: "Enumeration....." is that right?
#### Search

Can we have the first tab just be a search of latest? (we may just have to make our own tabs and multiple search engines for this)

This doesn't seem to be possible - in particular:
- cannot seem to change the first tab
- cannot index latest

---
## Comments not yet distilled

I think the refguide needs to be more expanded navigation too

---

not shown here, but I think in a page full of functions/methods, it is really hard to tell where one stops and the next starts. I'd almost like the entire description indented a bit underneath, e.g. "autoload_server"

Here's another example of what I was mentioning: do the descriptions describe the thing above, or the thing below? Everything bleeds together. My first inclinations is that indenting all the descriptions would clearly (at least to me) indicate that they are "under" the thing above. Also BTW some the elements have the "before" pseudo elements (like "method") but others like these do not. 

<img width="480" alt="screen shot 2015-12-31 at 5 20 08 pm" src="https://cloud.githubusercontent.com/assets/1796208/12091068/23236b1a-b2c2-11e5-9997-b3de77c4ef76.png">

---

[...]

PR #3691, opened 2016-01-20
Auto-update releases link in docs nav

### LINKED ISSUE #3686: Auto-update the "releases" link in conf.py
From sphinx/README.txt:

> In ``conf.py`` the links that appear in the docs nav appear. On a new release, the 
releases link should be changed to the latest release notes (we should find a way to automate this)

From @bryevdv: There's a script that updates the versions to the new version for release, I'd suggest it could also note the previous version and insert it in conf.py where appropriate.

birdsarah:
Fixes #3686

### DIFF
--- a/sphinx/README.txt
+++ b/sphinx/README.txt
@@ -15,10 +15,11 @@ To build the gallery, ggplot and seaborn are also needed:
 Manual actions for new releases
 ===============================
 
-In ``conf.py`` the links that appear in the docs nav appear. On a new release, the 
-releases link should be changed to the latest release notes (we should find a way to automate this)
-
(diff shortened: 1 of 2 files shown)


bryevdv:
I think `ALL_VERSIONS` could be automated too, but that could be another PR.

birdsarah:
> I think ALL_VERSIONS could be automated too

It is somewhat curated - but if the rule is only pick the last point release for last 3 versions, then yes.

> but that could be another PR.

+1

birdsarah:
(added ALL_VERSIONS automation as a todo on #3688)

PR #3697, opened 2016-01-20
Add Renderer base class to *_renderer models

canavandl:
ref issue #3651 

Simply add Renderer base class to *_renderer models to mirror python API


mattpap:
LGTM, nothing to review here.

bryevdv:
What about `Annotation`?

canavandl:
Good eye. Thanks @bryevdv . Should be good now.

PR #3703, opened 2016-01-21
Task/add component base class

canavandl:
ref issue #3651 

supersedes #3624 

Add Component base class to BokehJS
Make Layout subclass of Component (not Widget) and refactor into distinct directory
Rename bokehjs/src/coffee/models to bokehjs/src/coffee/model (in line with singular naming convention)


bokehjs/src/coffee/common/models.coffee
```diff
@@ -12,11 +12,20 @@ module.exports = {
...
+  VBox:                     require '../layout/vbox'
+  VBoxForm:                 require '../layout/vboxform'
+
   CategoricalMapper:        require '../mapper/categorical_mapper'
   LinearColorMapper:        require '../mapper/linear_color_mapper'
   LinearMapper:             require '../mapper/linear_mapper'
   LogMapper:                require '../mapper/log_mapper'
 
+  Model:                    require '../model/model'
+  Component:                require '../model/component'
```
mattpap: Those are abstract classes and shouldn't be necessary here.

canavandl: You are right. I'll remove them

bokehjs/src/coffee/callback/customjs.coffee
```diff
@@ -1,5 +1,5 @@
 _ = require "underscore"
-Model = require "../models/model"
+Model = require "../model/model"
```
mattpap: I hope this is temporary change and we will consistently use plural form across bokeh/bokehjs.

canavandl: AFAIK we seem to use singular directory names in BokehJS. In Bokeh, most of what are directories in BokehJS (i.e. renderer/glyph/) is a single file with a plural name (glyphs.py) in Bokeh.

I was just trying to continue that pattern.

bokeh/tests/test_layouts.py
```diff
@@ -0,0 +1,37 @@
...
+        from bokeh.models.widgets.layouts import Layout
+        self.layoutCls = Layout
+
+    def test_expected_props(self):
+        expected_properties = set(['width', 'height'])
+        actual_properties = get_prop_set(self.layoutCls)
+        self.assertTrue(expected_properties.issubset(actual_properties))
+
+    def test_props_defaults(self):
+        layout = self.layoutCls()
```
mattpap: `Layout` is an abstract class, so you shouldn't be initializing it directly. Currently we have a phony `@abstract` decorator, but it doesn't do any verification whether you initialize an abstract class or not. You should be testing concrete subclasses of `Layout` here.

bokeh/tests/test_layouts.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import absolute_import
+
+import unittest
+import inspect
+
+
+def get_prop_set(class_object):
```
mattpap: I'm pretty sure this or similar method is implemented on `HasProps`, `Model` or `Viewable`.

bokehjs/src/coffee/callback/customjs.coffee
mattpap: It's good to stay consistent rather than make unrelated changes. However, at the very end of this sequence of changes, we should switch to plural form. I presume it should be added as a task to the meta-issue (@bryevdv, your comment would be appreciated).

bryevdv: It's all the model subclasses are already in `bokeh.models` (plural) on the python side. It's arguable whether `bokeh.model.Model` should have been `bokeh.models.Model` but I couldn't make it work, and it doesn't seem unreasonable for the overall base class to be separate.

bryevdv:
@canavandl the `Renderer` PR caused a conflict. Speaking of which, after this PR let's do the directory re-org so others can work on BokehJS with less risk of really severe conflicts.

bokeh/models/component.py
```diff
@@ -7,7 +7,9 @@
 
 @abstract
 class Component(Model):
-    """ A base class for all embeddable models, i.e. plots and widgets. """
+    """ A base class for all embeddable models, i.e. plots, layouts and widgets.
```
bryevdv: I guess "embeddable" is not the right work here anymore. Maybe "for all DOM-level components, i.e. plots, layouts, and widgets"

bokehjs/src/coffee/callback/customjs.coffee
bryevdv: And if we don't put `model.coffee` in `src/coffee/models` then it should just go in `src/coffee`, not `src/coffee/model`

bokeh/models/component.py
canavandl: I like that - it's more precise wording.

canavandl:
I'll deal w/ the merge conflict rn

bryevdv:
I think we should resolve the location of `model.coffee` here. I'm fine with either of these:

* `src\coffee\model.coffee`  (matches python a little closer, but puts coffee file in top level
* `src\coffee\models\model.coffee` (slight mismatch with python.. what I had actually wanted but could not make work)

bryevdv:
@mattpap do you have a preference?

mattpap:
I would prefer 1-1 mapping (as much as possible). Having `*.coffee` files at the top-level isn't that bad, especially that we should use relative `require()` imports anyway.

bryevdv:
OK, let's do that then. @canavandl I think after that change this will be good to go.

canavandl:
That's fine. Do you have an opinion of where to put component.coffee (and plot.coffee if/when we refactor it out of src/coffee/common/)?

bryevdv:
Would just leave it in common for now

canavandl:
ping @bryevdv 

should be ready for review/merge

PR #3707, opened 2016-01-22
Remove inline style from embed_multiple.py

### LINKED ISSUE #3705: inline CSS from embed_multiple.py jinja template overrides Bokeh plot icon CSS
screen shot 2016-01-21 at 11 26 41 pm

Need to remove inline CSS of jinja template in embed_multiple.py

canavandl:
issue: fixes #3705 

Remove inline style of div tag that overrides bokeh button icon style. Plots are now vertically aligned (as consecutive divs) and not horizontal. It's be possible to wrap them in an HBox, but that seems to detract from demonstration of embedding/jinja templating of multiple components.

### DIFF
--- a/examples/embed/embed_multiple.py
+++ b/examples/embed/embed_multiple.py
@@ -46,15 +46,14 @@ template = Template('''<!DOCTYPE html>
     <head>
         <meta charset="utf-8">
         <title>Bokeh Scatter Plots</title>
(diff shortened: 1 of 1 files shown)


bryevdv:
Travis is being dumb, the py27 unit tests job is not starting. LGTM, merging

PR #3711, opened 2016-01-22
Allow to align plot title to left and right (center worked before)

mattpap:


bryevdv:
It's not clear to me that the alignment should be against the entire canvas, I think the left/right edges of the frame (i.e. aligned with the edges of the plot region) might be more commonly desired? That said, we could merge this as an immediate improvement. Any thoughts @birdsarah ? +1 to merge as-is, and fix up more later (i.e. when titles become label annotations)

birdsarah:
oh boy, this is soooo exciting!

I am +100 to merge. I've tested locally

Actually from a design point of view I would often want it all the way left aligned. For example, see: http://cecp.mit.edu/pages/economic-growth.html where I manually aligned the html on the left like this. Checkout http://www.economist.com/blogs/graphicdetail for a whole slew of examples that left align like this.

So any further improvements, shouldn't lose this.

One day soon we'll be able to do this:
screen shot 2016-01-22 at 6 21 24 pm

birdsarah:
It would be nice if `major_label_text_align` worked. But not going to hold up merging this PR.

PR #3712, opened 2016-01-22
Use make_id() instead of str(uuid.uuid4())

mattpap:
This restores support for custom ID generators and simple IDs in particular, which is useful for debugging.


bryevdv:
I was actually about to make an issue to discuss making simple id's the default. I don't think we need UUID's inside documents.

PR #3717, opened 2016-01-23
Fix layout update for non-visible axes

### LINKED ISSUE #3626: Support swapping of axes/ranges without re-creating entire plot
This example server app simplifies what I am doing to support interactive charts. Based on the data selected, new ranges and axes can be required. 

https://gist.github.com/rothnic/18371807da60b288166f

This app renderers fine the first time, but on selector change to 'c' the axis will be changed to a CategoricalAxis. The initial error exists on the bokehjs side in axis.coffee, in the _computed_bounds method when get('ranges') results in [undefined, undefined].

An alternative approach I took was to leave all axis types in renderers and just setting visible the one I want to show. This also resulted in errors on the bokehjs side, but I was able to fix the error by adding `@_last_size = 0` below this line: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/renderer/guide/axis.coffee#L363

rothnic:
This resolves #3626 based on the research outlined in the issue. In short, there were two methods of supporting swapping axes out after the chart has been rendered. This approach is probably the least desired option because we should have a simple method to add axes as, as needed, but it is the most simple. Currently, the only alternative approach to this problem is to recreate the entire plot.

Whether more work will be done or not to fix this issue upstream in the rendering logic, this simple change should be there, otherwise layout constraints will be solved for non visible axes, which kiwi will error out on when arguments provided are 0, -0.

### DIFF
--- a/bokehjs/src/coffee/renderer/guide/axis.coffee
+++ b/bokehjs/src/coffee/renderer/guide/axis.coffee
@@ -361,6 +361,7 @@ class Axis extends GuideRenderer.Model
   update_layout: (view, solver) ->
     if not @get('visible')
       size = 0
+      @_last_size = 0
     else
       size = (@_tick_extent(view) + @_tick_label_extent(view) +
               @_axis_label_extent(view))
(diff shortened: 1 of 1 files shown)


rothnic:
Tests passed locally for me, so as long as there aren't any surprises, this should be ready.

bryevdv:
OK I will probably be revisiting some of this in a general PR to clean up layout but not before `0.11.1` so merging this now.

PR #3718, opened 2016-01-23
Fix #3698 attribute name in doctemplate / helpers.py

### LINKED ISSUE #3698: Doc: data_source is not an expected attribute to Text
````
p = figure(x_range=self.xr, y_range=self.yr, title=index, tools='hover, crosshair', **plot_formats)
p.text(data_source=source, x='x', y='y', text=['test'], text_color="black", text_font_size="10pt")
````
results in 
````
AttributeError: unexpected attribute 'data_source' to Text, possible attributes are angle, angle_units, name, tags, text, text_align, text_alpha, text_baseline, text_color, text_font, text_font_size, text_font_style, visible, x, x_offset, y or y_offset
````

whereas docs mentioned data_source as a possible parameter (http://bokeh.pydata.org/en/latest/docs/reference/plotting.html)

aavanian:
Change data_source to source which is the attribute expected for glyph construction

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -418,7 +418,7 @@ Keyword Args:
 Other Parameters:
     alpha (float) : an alias to set all alpha keyword args at once
     color (Color) : an alias to set all color keyword args at once
-    data_source (ColumnDataSource) : a user supplied data source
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks @aavanian

PR #3719, opened 2016-01-23
Don't generate source maps for *.min.js files

### LINKED ISSUE #3676: Minified files should not try to load .map files
Users see this, all the time:

screen shot 2016-01-18 at 6 12 56 pm

ping @mattpap @bokeh/core Is this feasible?

mattpap:
This fixes #3676.


canavandl:
LGTM

PR #3727, opened 2016-01-24
Update elements.csv 

Luke-in-the-sky:
First row in elements.csv was one column short: the data has an 'id' column, but the header row is missing the 'id' key.

Adding a comma at the beginning of the whole file does the trick

### DIFF
--- a/bokeh/sampledata/elements.csv
+++ b/bokeh/sampledata/elements.csv
@@ -66,7 +66,7 @@ atomic number,symbol,name,atomic mass,CPK,electronic configuration,electronegati
 65,Tb,Terbium,158.92535,#30FFC7,[Xe] 4f9 6s2,1.2,,92.3 (+3),,566.0,-50.0,solid,metallic,1629,3503,8.22,lanthanoid,1843,-,6
 66,Dy,Dysprosium,162.500,#1FFFC7,[Xe] 4f10 6s2,1.22,,107 (+2),,573.0,-50.0,solid,metallic,1685,2840,8.55,lanthanoid,1886,-,6
 67,Ho,Holmium,164.93032,#00FF9C,[Xe] 4f11 6s2,1.23,,90.1 (+3),,581.0,-50.0,solid,metallic,1747,2973,8.8,lanthanoid,1878,-,6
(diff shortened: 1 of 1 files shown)


bryevdv:
Him, I guess I am curious why this was not breaking things in the past.

bryevdv:
In any case, is there are reason not to name the first column "id" ?

Luke-in-the-sky:
Sure, that would be nice too :) I just thought of introducing the smallest change possible, that's why I limited the edit to introducing a single comma. 
Actually, IMHO, the whole 'id' column should go, it's just redundant info. Do you think that would be too much of a change if a whole column where to be removed?

bryevdv:
I think that would be fine, just need to verify that the (two? elements.py and periodic.py) examples that use it display correctly.

bryevdv:
@Luke-in-the-sky Just FYI we are planning to release `0.11.1` early next week. If you'd like this PR to be included then it will need to be finished/merged by Friday or so.

mattpap:
btw. Erbium is missing CPK (according to wikipedia should be `#00E675`).

Luke-in-the-sky:
@bryevdv Sorry, I'm new to this. Could you tell me a bit more of what needs to be done when you say 'need to be finished/merged'? Thanks

fpliger:
@Luke-in-the-sky thanks for taking care of this..

'need to be finished/merged' means that all the changes you have in mind have been done, commited and pushed. You can add the `ready` label to the PR (or just comment that it's ready if you prefer ) and we will know that it's ready for a review. We will take care of merging and closing it if everything looks good.

bryevdv:
@Luke-in-the-sky I got the impression you wanted to remove the first column, so I have been waiting on that to merge this PR.

Luke-in-the-sky:
Thanks for explaining: it's done now

bryevdv:
@Luke-in-the-sky awesome thanks! If I can trouble you for one last edit as long as we are working on this file can you update the CPK value for erbium as @mattpap described:

> btw. Erbium is missing CPK (according to wikipedia should be #00E675).

Then we will get this merged!

Luke-in-the-sky:
sorry, forgot that one. `ready` now

bryevdv:
Thanks @Luke-in-the-sky !

PR #3731, opened 2016-01-25
support horizontal legend orientation

### LINKED ISSUE #3832: Support horizontal legend orientations

bryevdv:
issues: closes #3832

A few notes:

* chose `legend_orientation` because `orientation` was only very recently deprecated (as a former synonym for `location`) Other suggestions welcome. 

* thoughts on spacing welcome, could possibly use tweaking (script to test with attached below)


bryevdv:
```

from numpy import pi, sin, cos
import numpy as np

from bokeh.models.glyphs import Line, Circle
from bokeh.models import (
    Plot, DataRange1d, LinearAxis, ColumnDataSource,
    PanTool, WheelZoomTool, PreviewSaveTool, Legend,
)

x = np.linspace(-2*pi, 2*pi, 1000)
y = sin(x)
y2 = cos(x)

source = ColumnDataSource(data=dict(x=x, y=y, y2=y2))

xdr = DataRange1d()
ydr = DataRange1d()

plot = Plot(x_range=xdr, y_range=ydr, min_border=50, plot_width=800)

line_glyph = Line(x="x", y="y", line_color="blue")
line = plot.add_glyph(source, line_glyph)
circle = Circle(x="x", y="y2", line_color="red")
circle = plot.add_glyph(source, circle)

plot.add_layout(LinearAxis(), 'above')
plot.add_layout(LinearAxis(), 'below')
plot.add_layout(LinearAxis(), 'left')
plot.add_layout(LinearAxis(), 'right')

pan = PanTool()
wheel_zoom = WheelZoomTool()
preview_save = PreviewSaveTool()

plot.add_tools(pan, wheel_zoom, preview_save)

from bokeh.core.enums import LegendLocation

for location in LegendLocation:
    legend = Legend(legends=[(location, [line]), ("other", [circle])], location=location, legend_orientation="horizontal")
    plot.add_layout(legend)

legend = Legend(legends=[("x=100px, y=150px", [line]), ("other", [circle])], location=(100, 150))
plot.add_layout(legend)

doc = Document()
doc.add_root(plot)

if __name__ == "__main__":
    filename = "legends.html"
        f.write(file_html(doc, INLINE, "Legends Example"))
    print("Wrote %s" % filename)
    view(filename)
```

mattpap:
> Other suggestions welcome.

Just use `orientation` and remove deprecated behavior. I know it's early, but if we want to adhere to deprecation cycle, this PR should be rescheduled for 0.12. There should be no `XYZ.xyz_attribute` properties in our API, unless `attribute` has multiple meanings within `XYZ`. I'm aware of only axes requiring this. There is `Plot.plot_{width,height}`, but that's because of confusion in `figure()`, which I believe should have been resolved differently.

bryevdv:
That would work too.

bryevdv:
That said, I want to use this in the layout work I'm starting soon, so I think we will have to make a long running `0.12` branch soon and merge this into that.

bryevdv:
I tweaked it so that it now takes the individual label widths into account. `label_width` is really a minimum label width (which has always been true actually, but is more evident here, perhaps)

screen shot 2016-02-05 at 7 45 31 am

bryevdv:
@bokeh/core this is ready for review

bryevdv:
This is going to conflict with some other work unless I merge now, please open new issues for anything too address. @bokeh/core

PR #3733, opened 2016-01-25
Remove always_active and make InspectTool.active configurable

### LINKED ISSUE #3604: HoverTool.always_active is not implemented

mattpap:
This fixes #3604. Originally I wanted to just implement `always_active`, but then realized that there is already `active` property used internally by inspect tools, so I just exposed this one and removed `always_active`. Which is a good thing also because its name was confusing. This potentially could be extended to gesture tools as well.


bokehjs/src/coffee/tool/inspectors/inspect_tool.coffee
```diff
@@ -28,7 +28,8 @@ class InspectTool extends Tool.Model
   event_type: "move"
 
   nonserializable_attribute_names: () ->
-    super().concat(['event_type', 'inner_only'])
+    attrs = _.without(super(), 'active')
+    return attrs.concat(['event_type', 'inner_only'])
```
mattpap: I'm not sure what's our policy on this is. It's definitively not pretty.

mattpap: On the other hand, this could be viewed as an override.

bryevdv: That (`Override`) seems better to me. 

FWIW the `always_active` was originally the means to enable pan and wheel zoom tools on at the same time, before events were more carefully handled separately by gesture type. It can definitely go now.

mattpap: > That (Override) seems better to me.

So I was only saying it could be viewed as an override. `Override` (property descriptor) doesn't belong here, because it's bokeh-side, whereas nonserializable properties are bokehjs-side. I'm not sure if we are on the same page and this code fine fine to proceed with?

bryevdv: Well, I was referring to the python side. Specifically, only one tool from each gesture tool can be active at a time, so it doesn't seem like a single default value will do. I thought you were proposing `Override` to selectively turn on only some of the tools by default. There are combinations that don't make sense, though, turning on pan and box zoom e.g. is impossibly at the same time, since they are both drag tools. What is the right thing to do in that case?

mattpap: So, this is why I limited my self to only inspect tools, because that's the simple case. With other kinds of tools, it's more complex and would require, I presume, cumbersome logic. A tool manager is a solution, but staying with `active` property for all tools would work too, just with cumbersome logic. My first approximation to encoding active toolset is the following descriptor `Tuple(GestureTool, Set(InspectTool))`.

bryevdv: I missed that this was only on `InspectTool`. Sorry. Given that LGTM

mattpap:
git is a charm, no manual conflict resolution with `master`.

PR #3735, opened 2016-01-25
Task/3651 unify directory structure

canavandl:
ref issue: #3651 

Super WIP!!! The coffee import paths are broken all over. (I'll kill the TravisCI builds until they're fixed)

Currently just a demonstration of the proposed directory structure for feedback.


canavandl:
Worth breaking `src/coffee/models/guides/` into `axes` and `grids` directories?

canavandl:
Worth flattening nested `tools/actions/ tools/gestures/ tools/inspectors` into single dir?

mattpap:
I would say not yet.

bokehjs/src/coffee/models/widgets/defaults.coffee
```diff
@@ -0,0 +1,204 @@
+
+all_defaults = {
```
mattpap: You shouldn't have added this.

mattpap: I presume this happened because `.gitignore` wasn't updated in time.

canavandl:
I agree ^ 

There's a lot more code in the coffee/models/tools directories (and a bunch of base classes i.e. tool.coffee, action_tool.coffee, gesture_tool.coffee). It seems cleaner to have that little bit of nesting.

bokehjs/src/coffee/models/widgets/defaults.coffee
canavandl: very possibly - i'll look into it.

mattpap:
There is a little issue with this PR. It should be merged without major delays, so that we can work on bokehjs. However, it's scheduled for 0.11.1, but that's a highly breaking change for all custom JS code.  Or maybe not, because I doubt that many users actually imported anything in JS callbacks. Any thoughts @bokeh/dev?

bokehjs/src/coffee/models/widgets/defaults.coffee
mattpap: And you should update defaults generating code as well.

mattpap: Actually I should have suggested in the issue you do `git clean -dfx` before any work.

mattpap:
> (I'll kill the TravisCI builds until they're fixed)

You could use `[no ci]` in commits. That should prevent travis ci jobs from starting.

canavandl:
thanks ^. I did not know that.

bryevdv:
> because I doubt that many users actually imported anything in JS callbacks.

AFAIK we've never demonstrated this anywhere, and in fact it never occurred to me that it would be possible (makes senses I just had never thought of it).

canavandl:
Comparison of bokeh/bokehjs directory structures.

```bash
$ ls -1 bokehjs/src/coffee/models
annotations/
axes/
callbacks/
component.coffee
formatters/
glyphs/
grids/
layouts/
mappers/
markers/
ranges/
renderers/
sources/
tickers/
tiles/
tools/
widgets
```

```bash
$ ls -1 bokeh/models
__init__.py
annotations.py
axes.py
callbacks.py
component.py
formatters.py
glyphs.py
grids.py
images.py
layouts.py
map_plots.py
mappers.py
markers.py
plots.py
ranges.py
renderers.py
sources.py
tests/
tickers.py
tiles.py
tools.py
widgets/
```

canavandl:
Probably should refactor plots into models/plots/ dir

mattpap:
@bryevdv, there one example: `examples/models/custom.py`, but that's already affected by layout changes. In general, I think, there should be no issue with this, just wanted to make sure the team is aware of the possibility.

canavandl:
ping @mattpap. Do you know the importance of src/coffee/api.coffee?

@bryevdv and I were unsure about it, and I removed it and the tests still pass (after removing the require line from src/coffee/main.coffee)

bryevdv:
I just realized I misunderstood about `api.coffee` I was thinking of the old api *directory*. That said maybe we can delete most of what's inside it, thoughts @mattpap?

mattpap:
`api.coffee` is a leftover from the old bokehjs JS-only API. bokehjs itself doesn't use it and I doubt people are using it. Everything that we use from `Bokeh` global is defined in `main.coffee`. If we want to further split bokehjs and/or make bokehjs' contents configurable, `api.coffee` has to go (at least in the current form).

mattpap:
Oh, and widgets' equivalent of `api.coffee` was already removed.

canavandl:
Ok - so I went ahead and removed api.coffee because it was causing some path error. This is ready for review @mattpap @bryevdv

bokehjs/gulp/tasks/generate_defaults.py
```diff
@@ -104,8 +104,5 @@ def output_defaults_module(filename, defaults):
 
 output_defaults_module(filename = os.path.join(dest_dir, 'common/defaults.coffee'),
                        defaults = all_json)
-output_defaults_module(filename = os.path.join(dest_dir, 'widget/defaults.coffee'),
+output_defaults_module(filename = os.path.join(dest_dir, 'models/widgets/defaults.coffee'),
```
bryevdv: I'd honestly rather create both of these somewhere else entirely but that can happen in a different PR

bryevdv:
This looks good at a glance, I will checkout/build a little later, and also try to go over a bit more carefully. It would be great to get this merged today so that other BokehJS work can proceed, but it would also be good to get a few m ore eyes on it / trying it out  @bokeh/dev

bryevdv:
Looking at the report at https://s3.amazonaws.com/bokeh-travis/0.11.1dev2-9169795/report.html it looks like all the file examples look great, but not seeing plots in notebook maybe? @canavandl can you try out notebook example by hand?

bokehjs/gulp/tasks/generate_defaults.py
canavandl: I agree - the locations weren't great initially and the new proposed locations aren't any better. I'd prefer to update their locations in a separate PR. I also want to rename the modules that register all of the models:

`common/models.coffee` (for models)
`models/widgets/main.coffee` (for widgets)

mattpap:
I tried notebooks locally and they work. Most likely it's our notebook testing infrastructure that doesn't work.

bokehjs/gulp/tasks/generate_defaults.py
bryevdv: Sure, another PR is fine. I'm not sure about those locations, maybe just all of them at the top level is preferable, named for the modules they govern:
```
bokeh.coffee
bokeh-widgets.coffee
bokeh-compiler.coffee
```
etc.

mattpap: > common/models.coffee (for models)

This is already named this way. Or you just didn't propose a new name?

canavandl: I didn't propose a new name. I like what @bryevdv suggests tho

bryevdv: Alternatively all of those could go in a `build` subdirectory.

mattpap: At this point I'm not sure who is proposing what, because we speak both about defaults and models' paths.

mattpap: > bokeh-widgets.coffee

This would be confusing after compilation (whatever this refers to - name doesn't suggest at all), because it would generate `bokeh-widgets.js` and we already generate that.

bryevdv: > This would be confusing after compilation (whatever this refers to - name doesn't suggest at all), because it would generate bokeh-widgets.js and we already generate that.

Ah that's a good point.... I guess my main suggestion is that the "build" coffee files (that do nothing but collect `require` statements for a built  final module) get a consistent naming convention, and (possibly) a separate "place" for them all to go together.

bryevdv: How about:
```
src/coffee/
   |
   +---build/
          |
          +---bk-requires.coffee
          +---bk-widgets-requires.coffee
          +---bk-compiler-requires.coffee             

```

mattpap: @bryevdv, the files you are referring to are actually "required" in `common/base`, so putting them outside source tree is cumbersome (or at least needs reconfiguration of browserify). One thing I want to put on the table, is that I'm going to auto-generate models' paths, so this renaming task may be a no-op, because I may be able to generate them on-the-fly when generating the final bundle.

mattpap: Also, compiler doesn't need such file, because it doesn't contain any models.

mattpap: On the other hand auto-generated defaults could be put in build (or maybe in test), because we only use them in tests. However, using `build/` directory, will require changes to module resolution code.

mattpap:
I went through the diff and I don't see anything wrong, so LGTM.

bryevdv:
OK, let's merge so that other work can proceed, and get a dev build out this afternoon for some testing.

mattpap:
@bryevdv, at least we don't have to worry about typos in imports, because browserify verifies everything.

bokehjs/gulp/tasks/generate_defaults.py
bryevdv: @mattpap OK, will defer to your upcoming work on this then.

Issue #3638, opened 2016-01-12
Provide method for turning server autoadd off


rothnic:
If you try to set `curstate().autoadd = False` you get an error that you cannot set that attribute. There is an existing note in the code that this still needs to be added.

The driving need behind this is that there is logic within bokeh.charts.builder.create_and_build to check for this value and only set the chart as root if autoadd==True.

https://github.com/bokeh/bokeh/blob/master/bokeh/core/state.py#L88

PR #3736, opened 2016-01-25
Fix/add autoadd setter

brendancol:
Related issues:
https://github.com/bokeh/bokeh/issues/3638
https://github.com/bokeh/bokeh/issues/3546

example to see in action:
```python
from bokeh.models.widgets import HBox
from bokeh.io import curdoc, curstate
from bokeh.charts import Scatter
curstate().autoadd = False
s = Scatter(df, x='mpg', y='hp', color='cyl', marker='origin',
                          title="Auto MPG", xlabel="Miles Per Gallon",
                                            ylabel="Horsepower")
b = HBox()
b.children = [s]
doc = curdoc().add_root(b)
```
screen shot 2016-01-25 at 11 57 04 am


bryevdv:
@brendancol the setter was added to `State` what's the rationale for the last commit?

brendancol:
So the `builder.py` change was a false fix for trouble in my PR example.  PR example updated, curstate -> curdoc change reverted...

bryevdv:
@brendancol can you elaborate? Are you saying setting `autoadd` to `False` before creating the chart does not solve the problem?

brendancol:
No, it DOES solve the problem.  All should be good here

bryevdv:
Leaving #3546 open pending a more comprehensive (automatic) solution.

PR #3738, opened 2016-01-25
Add support for bokeh static command

mattpap:
Run `bin/bokeh static` and you will get bokehjs served under `http://localhost:5006/static/`. Address (host), port and log level are configurable. This uses the "full" bokeh server, but configures only `/static/` url, so no apps, websockets, etc.


bryevdv:
Was failing for a dumb reason. Should pass now then we can merge.

bokeh/command/subcommands/static.py
```diff
@@ -0,0 +1,60 @@
...
+
+import logging
+log = logging.getLogger(__name__)
+
+from bokeh.server.server import Server
+from bokeh.util.string import nice_join
+
+from ..subcommand import Subcommand
+
+LOGLEVELS = ('debug', 'info', 'warning', 'error', 'critical')
```
bryevdv: This should be factored out somewhere.

bokeh/command/subcommands/static.py
```diff
@@ -0,0 +1,60 @@
...
+
+class Static(Subcommand):
+    ''' Subcommand to launch the Bokeh static server. '''
+
+    name = "static"
+
+    help = "Run a Bokeh static server (hosting bokehjs' static files)"
```
bryevdv: suggest `help= "Serve BokehJS static JavaSript and CSS assets"`

bokeh/command/subcommands/static.py
```diff
@@ -0,0 +1,60 @@
...
+        )),
+
+        ('--log-level', dict(
+            metavar='LOG-LEVEL',
+            action  = 'store',
+            default = 'info',
+            choices = LOGLEVELS,
+            help    = "One of: %s" % nice_join(LOGLEVELS),
+        )),
+    )
```
bryevdv: This should be codified in a test (inching towards locking down API and CLI specifications)

bryevdv:
@mattpap a few small comments

mattpap:
@bryevdv, I think your comments were addressed. I resolved conflicts. This is ready.

bryevdv:
test failures seem to be real, not sure whats going on

bryevdv:
LGTM, merging

PR #3741, opened 2016-01-25
Task use Firefox on Saucelabs not Travis

birdsarah:
Run our browser tests on saucelabs.


scripts/travis_install
```diff
@@ -48,9 +48,10 @@ conda install --yes nodejs
 echo "node version $(node -v)"
 echo "npm version $(npm -v)"
 
+conda install --yes boto
```
mattpap: `boto` for Python 3? Is that true?

.gitignore
```diff
@@ -14,6 +14,7 @@
 .sw[nop]
 *.tmp
 *.coverage.*
+pytest-report.html
```
mattpap: Could this be configured to be a hidden file (dot prefixed)?

scripts/travis_install
birdsarah: Yes! Has been for a while now.

birdsarah: The report (run on py3.4 integration) successfully uploaded to s3: https://s3.amazonaws.com/bokeh-travis/104764781/report.html

.gitignore
birdsarah: yes, but it's useful to see locally and if it's a hidden file it'll be harder for people to find. I could go either way on this - just noting.

birdsarah:
@mattpap - thanks for the early eyes - I've opened this chunk as a standalone PR #3744 - lets carry on the conversation there about the pytest-report.html and whether/how it should be in .gitignore (the report name is completely configurable).

birdsarah:
Previously our selenium tests were running an a version of firefox installed on travis. Not only did this not give us control over which exact version of travis, but it did not provide options for future expansion.

By integrating with SauceLabs we'll be able to test against multiple browser / OS combinations including mobile devices. SauceLabs offers free testing to open source projects on their OpenSauce plan (http://saucelabs.com/opensauce/).

In addition, SauceLabs gives a video of every test that ran so it's easy to see what happened: https://saucelabs.com/tests/467e2c5f99a54163a5a25ac351a4a600

This PR integrates pytest-selenium's html report and uploads them to S3. These nicely integrate the returned details from saucelabs. Here is an all passing test report: https://s3.amazonaws.com/bokeh-travis/104764781/report.html

And here is one with a failing test, the video from saucelabs along with the traceback is included in the report: https://s3.amazonaws.com/bokeh-travis/104811980/report.html

screen shot 2016-01-25 at 9 32 07 pm

There is still lots more to do on testing and using the full capabilities to try and reduce our manual burden of testing, but for now this is a one-for-one switch between Travis Firefox and Saucelabs Firefox.

There is one downside to this PR, the integration tests now take much longer because they are being run on Saucelabs, which then connects back to the localhost server running on travis. They are still shorter than the examples runs for what it's worth. Regardless, I think this is an important step forward and will hopefully just push us to running our integration and cross browser tests on a different schedule.

birdsarah:
@mattpap - I removed the default for the html report and only run it on the integration test in `.travis.yml` so there's no need for the gitignore.

Just like with the examples tests, the url for the s3 upload is printed just after the test run results.

bryevdv:
* how much longer do the integration tests take?
* Is there any Saucelabs integration with GH (can we just launch tests there directly from commits w/o going through Travis first?)
* Can we set up Sauelabs tests on a schedule (say "super full tests" 2x/day for master or for "all PRs" overnight)

birdsarah:
> how much longer do the integration tests take?

~9 min (~5 min -> ~14 min) - I'm thinking of setting-up another PR with browserstack once we get an account to see if that changes the performance any.

> Is there any Saucelabs integration with GH (can we just launch tests there directly from commits w/o going through Travis first?)

No. Saucelabs doesn't provide a test runner. It is pytest-selenium that's hooking into the Saucelabs API and pushing a test up-to-it and then getting the results back.

> Can we set up Sauelabs tests on a schedule (say "super full tests" 2x/day for master or for "all PRs" overnight)

In principle yes, the question is how. I see two options:
1) continue using travisci: Have a cron-job that triggers a commit with a message in it that will cause the relevant tests to run. Given that we already have the machinery to select which tests are run based on tags in commits like [no ci] [no examples] etc this should be easy.
2) have our own testing server. Make a server that, given a schedule, pulls code from git and runs the tests. this seems like it would be pretty easy to set-up.

Related: There was an issue about this https://github.com/travis-ci/travis-ci/issues/582 which referenced the service https://nightli.es/. The service seems to be down, but the code appears to on github https://github.com/coderanger/nightlies. Maybe we could run our own instance.

birdsarah:
TestingBot - which also integrates with pytest-selenium may be an option too - investigating.

birdsarah:
Also see: http://traviscron.pythonanywhere.com/ referenced from the travis-ci issue.

birdsarah:
CircleCI also seems like it would handle things for us: https://circleci.com/docs/nightly-builds

birdsarah:
https://docs.travis-ci.com/user/apps/#Tools - TronCI and TravisCron both linked from here.

birdsarah:
superceded by #3869

PR #3744, opened 2016-01-26
Output pytest report as html and upload to s3

birdsarah:
This is an incremental piece of doing browser testing on saucelabs and imagediff testing.

@mattpap started commenting on the WIP PR #3741 and I realized that this bit is totally standalone which'll keep the diff smaller on the rest of #3741.

You can see a successful upload to s3 here: https://s3.amazonaws.com/bokeh-travis/104764781/report.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 .sw[nop]
 *.tmp
 *.coverage.*
+pytest-report.html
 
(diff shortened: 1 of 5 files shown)


birdsarah:
#3741 turned out to be easier than planned - closing this.

PR #3752, opened 2016-01-26
Generate defaults for test task, not scripts:coffee

mattpap:
Also move some code around to make `tasks/scripts.coffee` more readable.


bryevdv:
restarted (apparently) spurious test failure

PR #3754, opened 2016-01-26
remove Chart from preload

### LINKED ISSUE #3726: Pandas required to use the server!
In the happiness demo I am using `plotting.Figure` to make my plots.

When I setup a clean environment with only `bokeh` and `django` as the requirements, I got the following error. I don't use the charts interface at all, so I'm guessing this is a bug.

```sh
Failed to push document: RuntimeError('The bokeh.charts interface requires Pandas (http://pandas.pydata.org) to be installed.',)
/envs/happiness/lib/python3.5/site-packages/bokeh/client/connection.py in push_doc, line 231
```

bryevdv:
issues: fixes #3726

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -51,7 +51,6 @@ class Viewable(MetaHasProps):
     @classmethod
     def _preload_models(cls):
         from . import models; models
-        from .charts import Chart; Chart
 
     @classmethod
     def get_class(cls, view_model_name):
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -51,6 +51,13 @@ class Viewable(MetaHasProps):
     @classmethod
     def _preload_models(cls):
         from . import models; models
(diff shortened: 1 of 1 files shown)


bryevdv:
@brendancol can you try this change out with your older charts based version of xfilter?

@birdsarah can you verify it works when you expected it to?

@mattpap any comments/verification welcome but AFAICT nothing in `bokeh.charts` is a `Model`

bryevdv:
n/m can confirm this does not work as-is. When trying "Happiness" get reverse lookup fail for "Figure" which I am guessing is accidentally brought in by `Chart`

bryevdv:
new approach seems to work, comments welcome.

bryevdv:
ping @mattpap OK I am going to merge this as it seems to work, if you have other/better ideas we can put them in later

birdsarah:
Late to the party, but this is working for me - thanks!

PR #3757, opened 2016-01-27
Task use firefox on Browserstack not travis

birdsarah:
Related to #3741, testing on browserstack instead of saucelabs.


birdsarah:
Here's a link to a report with failed tests run through browserstack - pytest-selenium does yet support embedding video in report - we could submit a pr for it though: https://s3.amazonaws.com/bokeh-travis/105044759/report.html

birdsarah:
This demonstrates how to integrate with browserstack. Browserstack appears to be no faster than SauceLabs so sticking with SauceLabs for now and pausing this PR.

PR #3758, opened 2016-01-27
Fix bad layouts imports in examples

canavandl:
supersedes #3756 

Todo:
deprecate `bokeh.models.widgets.layouts` import path

### DIFF
--- a/bokeh/models/__init__.py
+++ b/bokeh/models/__init__.py
@@ -21,4 +21,14 @@ from .sources import *
 from .tickers import *
 from .tiles import *
 from .tools import *
-from .widgets import *
+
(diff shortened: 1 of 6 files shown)


canavandl:
ping @bryevdv 

I'm not confident in my deprecations of layouts imports from bokeh.models.widgets. I feel like I had to do a hacky thing so that specific widgets could be imported from `bokeh.models` without loading the layouts that have to be available in bokeh.models.widgets.

bryevdv:
@canavandl That sucks, but I don't see any other option. Just leave comments in all three places (the two deprecations, and the crappy `__init__` to change as soon as the deprecation is carried out.

Issue #2935, opened 2015-10-01
Ordering of grouped bars in bokeh.charts Bar


tobyhodges:
I've been updating my scripts to take advantage of the improved charts interface in 0.10 and can't figure out how to control the order of grouped bars. I've checked the documentation/tutorials but couldn't find any information on this, sorry if I've missed it. Can someone point me in the right direction please?

Here is an example of the kind of data that I'm working with, where I'm plotting numbers for consecutive years. As you can see from the plot, I can't get the grouped bars to be arranged in the same order as the years are listed in the legend. 

```
from bokeh.charts import Bar, output_file, show
import pandas as pd

data = pd.DataFrame({'Group' :['A','A','A','A','B','B','B','B','C','C','C','C','D','D','D','D'],
                     'Year'  :['2009','2010','2011','2012','2009','2010','2011','2012','2009','2010','2011','2012','2009','2010','2011','2012'],
                     'Height':[1,2,3,4,2,2,2,2,3,0,3,0,4,3,2,1]})
fig = Bar(data, label='Group', values='Height', group='Year', legend='top_right')
output_file('chartsBarTest.html')
show(fig)
```

![chartstest](https://cloud.githubusercontent.com/assets/9694524/10221647/ef9bcd58-684f-11e5-9712-6316a93831d0.png)

damianavila:
Duplicate of #2924, let's continue the discussion there...

fpliger:
This is actually a different issue then #2924 .. the order of the category axis labels is already supported but honestly the order of the grouping is not... even explicitly specifying `group=CatAttr(columns=['Year'], sort=True)` and `color=ColorAttr(columns=['Year'], sort=False)` 

ping @rothnic in case I'm missing any internals..

fpliger:
pinging @ahothan here since you've mentioned you are having the same issue. Please feel free to provide any example to demonstrate your use case. This would probably be very helpful..

ahothan:
@fpliger here is a small example that shows the problem.

from bokeh.charts import Bar
from bokeh.charts import show
from bokeh.charts import output_notebook
from pandas import DataFrame
dc = {
    'count': {0: 772, 1: 2243, 2: 101, 3: 1833, 4: 2058, 5: 100, 6: 2236, 7: 1836, 8: 100, 9: 4387, 10: 93}, 
[...]

PR #3778, opened 2016-01-28
Fix issue with bar chart grouping/stacking order

rothnic:
Fixes: issue #2935

This fixes two issues:

1) an issue where the order of grouped elements wasn't being maintained during the grouping operations
2) an issue where the stack/group flags weren't getting set correctly within the bar builder

This also adds a test to check the final order of the glyphs as they would be rendered.

### DIFF
--- a/bokeh/charts/builders/bar_builder.py
+++ b/bokeh/charts/builders/bar_builder.py
@@ -5,17 +5,17 @@ This is the Bar class which lets you build your Bar charts just passing
 the arguments to the Chart class and calling the proper functions.
 It also add a new chained stacked method.
 """
-#-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


bryevdv:
Are there any associated GH issues to close?

bokeh/charts/builders/bar_builder.py
```diff
@@ -60,7 +61,7 @@ class BarBuilder(Builder):
     values = Dimension('values')
 
     dimensions = ['values']
-    #req_dimensions = [['values']]
+    # req_dimensions = [['values']]
```
fpliger: is there any reason we should keep this commented code here? If so we should add more context..

fpliger:
LGTM

PR #3784, opened 2016-01-29
responsive is False by default, no need to set

birdsarah:


mattpap:
Looks like there have been some S3 connectivity issues recently. Anyway, LGTM.

birdsarah:
merging as accepted and tests now passing.

PR #3790, opened 2016-01-30
Don't use spaces in JSON separators to reduce data size

mattpap:
This reduces size of choropleth.html by ~10% (1MB).

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -80,10 +80,18 @@ class BokehJSONEncoder(json.JSONEncoder):
         else:
             return self.transform_python_types(obj)
 
-def serialize_json(obj, encoder=BokehJSONEncoder, **kwargs):
+def serialize_json(obj, encoder=BokehJSONEncoder, indent=None, **kwargs):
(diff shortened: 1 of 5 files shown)


bryevdv:
Hooray for easy wins

bryevdv:
Looks like the "generate defaults" test is also passing in `separators`, resulting in a dupe keyword:

https://travis-ci.org/bokeh/bokeh/jobs/105940760#L1187

mattpap:
> Hooray for easy wins

I also did a preliminary analysis of binary JSON protocols and it doesn't look promising in general and looks completely useless in static HTML case. But that's preliminary work, I will followup with this later.

bryevdv:
@mattpap there is no need to investigate binary protocols for the web socket. I have demonstrated that it is possible to send numpy `.asarray` bytes output directly into JS typed array buffer views from web socket messages. No encoding whatsoever, so that is the way to go for sure. For standalone docs, maybe just base64. The benefit there is not so much size but real NaN/Inf

bryevdv:
LGTM

PR #3791, opened 2016-01-30
Add filename detection magic to save()

### LINKED ISSUE #3885: Proposal: remove need for output_file() and display in "native" window
I would like to discuss two related ideas:

1) Remove the need to require specifying an `output_file()`. Having to choose a name (and path) for an html file sits in the way of doing quick interactive plotting IMO. It feels like something of an export mechanism. 

1a) One simple solution is to automatically create an html file in a temp folder. We'd have to do a bit of management to clean up old html docs. 

1b) Alternatively, we could make the Python process serve the html file. This would require substantially more changes and design. This would lean towards making apps, not sure if this is an approach we want to support?.

2) The second proposal is to show the generated web page in a "native" window instead of adding a tab to a browser. I think this would give Bokeh a more interactive/native feel, similar to MPL. Of course, the "native" window would really be a web-browser in disguise, like [XUL](https://en.wikipedia.org/wiki/XUL) or [nwjs](http://nwjs.io/) or [Electron](http://electron.atom.io/). Flexx already has functionality to do this based on XUL (which is available anywhere where Firefox is installed), and Chrome. A similar approach should be possible based on Internet Explorer.

### LINKED ISSUE #4118: add implicit filename support for save()

mattpap:
issues: closes #4118, closes #3885

This uses inspect module and stack frame analysis to figure out what filename to use if not given explicitly through arguments or `output_file()`. I know this is ugly magic, but it should be pretty useful. I modified two examples in `plotting/file` to show this off.


bryevdv:
Just to be explicit, this turns `foo.py` into `foo.html` automatically?

mattpap:
It also preserves the full path, so `foo/bar/baz.py` -> `foo/bar/baz.html`.

almarklein:
What happens if I run this in an interactive session (i.e. there is no script filename to use as a reference)?

mattpap:
@almarklein, it should fail as if file name wasn't provided by any means (and method added in this PR is considered the measure of last resort). Perhaps we could detect interactive session and use whatever fits (e.g. "unnamed.html", maybe indexed to allow saving consecutive plots easier).

almarklein:
I can see two approaches to this.

1) Try to come up with a sensible name and place it next to the script. This is what this PR does so far IIUC. 

2) Not place an html file anywhere. Except that we place one in a temp folder. This case would only work in conjunction with show(), of course. This is basically a user saying "I want to see the plot, but I don't care about the generated html file".

Maybe a hybrid?

3) Do as in (1), unless we're in interactive mode, in which case we use (2).

mattpap:
I would say that `save()` could work as in this PR and `show()` as in 2). Though this may be a semantic change to the API.

bryevdv:
I'd suggest only changing the choropleth example, and not the periodic table one (I show that one off, it would be pretty annoying to have it stop auto-opening a browser). I think it would be good to add a comment in the choropleth example that states why `output_file` is not needed and that the filename is being auto-inferred. Also needs documentation support (both docstrings and user guide) and if possible, some tests.

bryevdv:
moved milestone and tags to related issue

almarklein:
Thinking about this some more, putting the html in a temp folder hidden from the user brings complexities. Since we hide the html from the user, we would be responsible for cleaning things up. But its hard to tell what filer are still in use and which are not. A user may have a tab open in her browser for an html file generated a week ago.

As for generating an html file in interactive mode brings complexities: you'd want unique names, because if a user creates two plots in an interactive session, they should not overwrite. But this can quickly lead to a lot of plots, because in an interactive session, you tend to re-run code a lot.

Therefore I now think it's better to keep it simple. Put the html right next to the script (as this PR does) for both `save()` and `show()`, and when there is no script, we raise an error.

bryevdv:
agreed

bryevdv:
LGTM but would prefer the change to revert the periodic example since I would like that one to continue to auto-open in a browser when run.

bokeh/io.py
```diff
@@ -311,14 +311,14 @@ def _show_with_state(obj, state, browser, new):
     elif state.server_enabled:
         _show_server_with_state(obj, state, new, controller)
 
-    if state.file:
+    else:
         _show_file_with_state(obj, state, new, controller)
```
almarklein: I think this is correct, because previously, if no filename was given, it would just ... do nothing.

almarklein:
I just pushed two commits to fix `_detect_filename()` and to apply the same trickery to `show()`.

bokeh/io.py
almarklein: Though the tests seem to explicitly want this function to be called ... @bryevdv @mattpap thoughts? Note that `_show_file_with_state()` is now able to find a filename if not given via `state`.

almarklein:
This is about done. A few questions:

* I removed `output_file()` from one example in the userguide, and verified that the example still works. Should we remove it from more examples to make them more concise?
* The docstring of `plot()` and `save()` are updated. Shall I also mention this behavior in the userguide?
* I think it makes sense to apply a similar logic to the title of the HTML file, i.e. use the title of the plot object if no title is given, instead of "Bokeh plot". WDYT?

almarklein:
ping

bryevdv:
I don't have strong feelings on this PR, it's not a feature I would ever use. If you and @mattpap feel it is ready, I'd say go ahead, or if you want more feedback, ping @bokeh/dev more widely

almarklein:
Ok, @bokeh/dev the main remaining point is that the user now separately specifies a title for the plot and for the html page. It makes sense that this is possible, because a page can contain multiple plots, but in most cases a page is just one plot.

My question is: when the user does not specify `output_file()` (which is what this PR allows), should the page title take on the title of the plot, or "Bokeh plot" as it does now?

mattpap:
> should the page title take on the title of the plot, or "Bokeh plot" as it does now?

I would vote for the former. However, I would postpone this and rethink support for multiple plots along the way (we have a tiny regression in this area).

bokeh/io.py
```diff
@@ -372,30 +377,57 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
         state = _state
 
     filename, resources, title = _get_save_args(state, filename, resources, title)
-
     _save_helper(obj, filename, resources, title, validate)
+    return os.path.abspath(filename)
+
+travis_filename_counter = 0
```
mattpap: @almarklein, this is some leftover?

almarklein:
> this is some leftover?

Good catch. Removed it. 

Made  #4287 for the document title thing. This is ready from my end.

almarklein:
@mattpap  Ping, I think we can merge this. I made a new issue for the auto-title thing.

mattpap:
LGTM

PR #3792, opened 2016-01-31
attempt to have a 'bokeh.exe' launcher on Windows

### LINKED ISSUE #3571: missing bokeh entry point (otherwise bokeh.bat) for Windows
hi,

on my windows, I have a "bokeh" file instead of a "bokeh.exe".

Wouldn't it be normal to get an entry point ? something like (but I guess slightly different)
````
    entry_points={
        'console_scripts': [
            'bokeh = bokeh.command.bootstrap:_main',
        ],
    },
````

stonebig:
issues: fixes #3571

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -581,6 +581,12 @@ setup(
         'bokeh.util.tests',
     ],
     package_data={'bokeh': package_data},
+    # use setuptools functionalities
+    entry_points={
+        'console_scripts': [
+            'bokeh = bokeh.command.bootstrap:_main',
+        ],
+    },
     author='Continuum Analytics',
     author_email='info@continuum.io',
     url='http://github.com/bokeh/bokeh',


bryevdv:
ping @damianavila @ilanschnell can you comment on this proposed fix for #3571 

Thanks @stonebig this is too late in cycle to go in the upcoming `0.11.1` release but we will have a `0.11.2` soon enough, and this will help us get started testing.

damianavila:
Regretabbly. this is not building... you are referencing a _main function which does not exist... additionally that main function should not accept argument to be used by setuptools entry_points.

I had an alternative branch to solve this one... so going to close this one as superseded by #3874

@stonebig thanks for pushing for this :+1:

stonebig:
no problem. thanks for the better fix.

PR #3794, opened 2016-02-01
Core refactor

bryevdv:
OK, there's lots more I want to do, but this PR is at a nice stopping place and was getting large (though most of the changes are just `require` fixups). The primary work was to make some of the more important files match the project structure of the python library. However, a some small bits of work were done along the way:

* factor some simple things out of `HasProps` and delete unused cruft from `HasProps`
* split `properties.coffee` into `properties.coffee` and `property_mixins.coffee`
* light cleanup of `properties.coffee`
* added some new property types on the JS side
* added ~100 new tests

Next steps in other PRs:

* split "context setting" parts off off mixin props
* make mixin props just contribute normal properties
* specify/initialize all (bokeh) properties in a uniform way 
* remove all nonserializable attribute names 
 - make some visible to python (e.g., `Frame`)
 - make others plain object attrs
* remove Backbone `Collection` dependence 
* fix up remaining bogus defaults mismatches

Once (most of) those are done, we will be in a good position to 1) add lots of new annotations 2) cleanup/refactor/extend canvas layout in preparation for whole document layout, 3) add the namespace concept


bryevdv:
ping @mattpap @bokeh/core The commits are all self-contained, would consider splitting into separate PRs if desired.

bryevdv:
I marked this as `0.12` but could potentially go in a `0.11.2`

bryevdv:
The slightly more strict `Property` already caught a few bugs

bryevdv:
@mattpap there is some issue with loading minified JSin the notebook, it says this:
```
SyntaxError: Unexpected token '>'
```
Any ideas off the top of your head? There are small issues with two MPL examples otherwise everything else seems to be working.

mattpap:
I had seen this error many times in bokeh-scala, but this was an issue with my custom build. I tested this locally and notebooks work for me.

bryevdv:
@mattpap Hrmm ok I will try to clean/clear everything out and try from scratch.

bryevdv:
@mattpap I have fully cleaned the repo and re-installed all the npm modules. I still get that with the minified, inline resources. Any suggestions for other things to investigate?

bryevdv:
@mattpap this SO link seems to indicate that it could be due to missing `.map` files:

http://stackoverflow.com/questions/18561556/syntax-error-unexpected-token

Could the recent change have anything to do here?

mattpap:
We don't generate nor "import" source maps in `*.min.js` files, so that can't be the issue.

mattpap:
Knowing we speak about `INLINE` resources, I can reproduce this.

bryevdv:
Sorry, I should have mentioned that earlier. That's how I typically test directly from master

bryevdv:
I've traced it to this, but it has not been much help (the row/col numbers are off)

screen shot 2016-02-01 at 5 25 24 pm

mattpap:
This may be a problem with uglifyjs (minification). I locally added support for `inline-dev` resources and notebooks work. I also removed comments from minified JS files (under `inline` resources) and still fails. So, the problem is in JS minified code.

mattpap:
On the other hand `sever(-dev)` resources worked, so that can't be the case.

mattpap:
The same example in `plotting/file` (I'm working on `color_scatter` btw.) works with minified inline resources. My next guess is IPython notebook doing something fishy.

bryevdv:
FWIW I pasted the JS I copied from the JS console into JS Fiddle and get the same error.

bryevdv:
I guess the next question, is this broken on master too? I've been assuming not, but have not verified.

bryevdv:
Seems to be failing on master too. Will have to investigate to see if `0.11` works with inline resources. I have to step out for now.

mattpap:
Am I missing something or do we "test" two notebooks on travis ci. By "test" I mean skip.

bryevdv:
I see several on the S3 report, but none of them render correctly.

mattpap:
Yeah, there are more, just hard to locate.

mattpap:
Thanks to the fact that we keep full report history, you can just go through commits and see when a particular example stops to fail. That will give us indication when notebooks were broken.

bryevdv:
@mattpap IIRC notebooks hyve been broken in the report a long time.You surmised that this was to do with testing machinery, but maybe they are actually this failure mode.

mattpap:
For 0.11 I tested everything locally and it worked.

mattpap:
This is the first time I see this particular problem locally outside bokeh-scala (and I didn't see this problem in boekh-scala since I switched to node.js for building bokehjs).

bryevdv:
> For 0.11 I tested everything locally and it worked.

OK, well maybe I can try a bisection later too.

mattpap:
This piece of garbage: `replace(></3>/g,"&gt;")`. It's not part of `bokeh*.js`. There is one occurrence, however similar, but correct, code (HTML escaping, probably from eco) is included multiple times. I don't know yet why this is there.

bryevdv:
OK, I believe this is ready now. I'd like to merge it after the `0.11.1` release.

bryevdv:
Ok, one last change: (almost) all the tests names/locations are aligned with the actual project filenames under `src/coffee`.

Issue #2589, opened 2015-07-20
Only create CSS rules to bokeh classes to avoid styling external elements


leafty:
When including bokeh css in an html page to embed bokeh plots, it includes general styling rules. For example, we have:
```css
label {
  display: inline-block;
  margin-bottom: 5px;
  font-weight: bold;
}
```
Bokeh should not do this and should always apply rules to its own elements (like `.bk-canvas`).

birdsarah:
Thanks @leafty, I agree. I have a couple of other css tickets open. Referencing them here for when we do an overhaul (#2374 #2504 #2451 #2606).  Of course pull requests are definitely welcome!

mistakevin:
We could wrap the entirety of the bokeh css in a unique class
```css
.bokeh-style { 
  /* all the bokeh css */
}
```
and then in the generated `<div class="plotdiv" id="b02b5032-014a-42b3-ac59-9af54a0fb74e">...</div>` just add a `.bokeh-style` class. That should keep the style from bleeding out.

@bryevdv @birdsarah @bokeh/dev

mattpap:
There is a rule already that all bokeh's CSS classes need `bk-` prefix. I know there are still some classes that weren't fixed, but that should be done instead of adding a wrapper class.

mistakevin:
Isn't keeping vanilla vendor libraries more conventional? If we did, we could remove bootstrap from checkin to git and pull it in with npm. And how does bootstrap's javascript play with the `bk-*` prefixes?

Those concerns aside, I'm still seeing issues with bootstraps global styles bleeding over <img width="397" alt="screen shot 2016-01-06 at 9 18 10 am" src="https://cloud.githubusercontent.com/assets/4411115/12145905/7fe4caa2-b456-11e5-9b2a-a3ff933eafd3.png">

mattpap:
> Isn't keeping vanilla vendor libraries more conventional, or, at least, a goal?

Definitively would be more convenient.

>  If we did, we could remove bootstrap from check-in to git and pull it in with npm.

Ultimately, we should remove bootstrap altogether. It's not a library but a framework and that causes issues.

> And how does bootstrap's javascript play with the bk-* prefixes?

BS' JavaScript is also prefixed.

> Those concerns aside, I'm seeing issues with bokeh's tag styles bleeding over.

Those styles aren't handled by prefixing and shouldn't have been included in the first place.

mattpap:
[...]

PR #3797, opened 2016-02-01
Containing all bokeh style to .bk-plot

mistakevin:
This will help better contain all of bokeh's style inside the `.bk-plot` class. There is bleeding regarding global html tags that the `.bk-*` naming convention just won't fix.

See here the `<label></label>` tag is inheriting bokeh's style
screen shot 2016-02-01 at 9 48 39 am

This PR will contain all styles to inside the `.bk-plot` class

screen shot 2016-02-01 at 9 48 59 am

screen shot 2016-02-01 at 9 50 53 am

See https://github.com/bokeh/bokeh/issues/2589


mistakevin:
@bokeh/dev

mistakevin:
@mattpap I chose `.bk-plot` because assumed all plots were already wrapped in this class. If we add a new class, are we going to auto generate the templates with the new class? If so, where would you suggest putting that in? The plot style will break if it is not wrapped, so it can't be optional.

mattpap:
`bokeh.embed` will have to be changed accordingly, so that `bk-bokeh` is added along `plotdiv`. Then `plotdiv` should be deprecated as it doesn't have prefix.

birdsarah:
The integration test failures on this PR are real - the selenium tests use css selectors to find elements and I'm guessing that this change has broken them.

Also, isn't `bk-bokeh` a little cumbersome? I thought `bk` was an abbreviation of `bokeh` so it seems a bit redundant. I would say just `bokeh` or `bk` is fine - my preference would be for `bk` as it makes the sub levels like `bk-plot` make more sense.

mattpap:
Well, we can call it `bk-root` or whatever. The point is to have `bk-` prefix consistently everywhere. Note that `bk-` prefix should be configurable if it causes conflicts with some environment (not that we are even close to making this possible).

mattpap:
> The integration test failures on this PR are real

Perhaps that's true, but they fail in very unhelpful way: `TypeError: string indices must be integers`.

mattpap:
Also, reviewing this PR, I don't see anything that could interfere with selenium.

mattpap:
OK, most likely we do embedding improperly in some/all cases. I recall not seeing `plotdiv`, so `bk-bokeh` isn't going to be there as well and then no CSS will be applied.

birdsarah:
> Perhaps that's true, but they fail in a very unhelpful way: TypeError: string indices must be integers.

Agreed - which is why I'm getting the better HTML reporting set-up. So you have an easy link to a video on failure.

mattpap:
@birdsarah, in this case it's more about a robust implementation (e.g. JSON schema validation) than nicer error reporting.

birdsarah:
@mattpap, can you clarify? When you write:

>  it's more about a robust implementation (e.g. JSON schema validation) than nicer error reporting

It reads to me like you're suggesting fixing the guts of selenium. This does not seem like a reasonable suggestion.

mattpap:
@birdsarah, I thought this was our code. I don't expect we will be fixing selenium, thought this piece of code in the tracebace should be rewritten from scratch.

birdsarah:
> @birdsarah, I thought this was our code. I don't expect we will be fixing selenium, thought this piece of code in the tracebace should be rewritten from scratch.

And just confirming, that given that the code in the traceback isn't ours, you're ok with me working on trying to give more helpful feedback given selenium's limitations.

mattpap:
@birdsarah, obviously yes.

birdsarah:
@mistakevin would you like me to push fixes to the selenium tests to this branch?

mistakevin:
@birdsarah Yes!

birdsarah:
I was mistaken, the selenium failure, was not a result of this work, but of the selenium window being too small and the box_select not being on-page to click.

This is fixed in PR #3896. 

When running this locally, I am having a problem with the sliders app. It appears that something is now missing from bokeh-widgets.css.

Here's what it looks like:

screen shot 2016-02-23 at 6 47 03 pm

This is what master looks like:

screen shot 2016-02-23 at 6 47 35 pm

I may have compiled something wrong, I just ran the regular `python setup.py develop --build-js` and then did `../../bin/bokeh serve --show sliders.py` from `examples/apps`

bokehjs/src/less/bokeh-widgets.less
```diff
@@ -1,3 +1,5 @@
-@import "vendor.less";
-@import "overrides.less";
-@import "widgets.less";
+.bk-plot {
```
birdsarah: I believe this is the source of my sliders problem, the nesting isn't working as expected. My `bokeh-widgets.css` output in `bokehjs/build/css`

starts:
```css
.bk-plot {
  /* makes the font 33% larger relative to the icon container */
  /* Font Awesome uses the Unicode Private Use Area (PUA) to ensure screen
   readers do not read off random characters that represent icons */
  /*!
 * jQuery UI CSS Framework 1.11.2
 * http://jqueryui.com
 *
 * Copyright 2014 jQuery Foundation and other contributors
 * Released under the MIT license.
 * http://jquery.org/license
 *
 * http://api.jqueryui.com/category/theming/
 */

/* Layout helpers
----------------------------------*/
.bk-ui-helper-hidden {
	display: none;
}
.bk-ui-helper-hidden-accessible {
```

Note the nesting - this doesn't work in css only in less or sass. Not sure quite what the problem is.

mistakevin:
@birdsarah Good catch. Looks like we were explicitly bringing in the CSS files as `inline`. Treating them as LESS seemed to do the trick.

bokehjs/src/vendor/slick-grid-2.1.0/slick-default-theme.css
```diff
@@ -96,7 +96,6 @@ classes should alter those!
 
 .bk-slick-row.loading {
   opacity: 0.5;
-  filter: alpha(opacity = 50);
```
mistakevin: This is for pre-IE8. Hoping we can live without it, as LESS does not support it.

bryevdv: We don't support even IE8, it does't have typed arrays.

birdsarah:
Have re-tested this, seems to be working fine.

canavandl:
I'm unable to get the tool icons to load using this branch:

screen shot 2016-02-24 at 2 38 46 pm

canavandl:
I think it's an issue of bringing in the the css as less (the icon pngs are generated css from a sphinx static folder)

canavandl:
The icon problems appears to be a defect in the `relative-dev` and `absolute-dev` BOKEH_RESOURCES settings that is in master and not related to this PR. 

So this LGTM and i'll open a hotfix PR for the resources issue.

mistakevin:
@mattpap After merging this, I just realized I accidently reverted back to `.bk-plot` after we had decided to go with something else. If this is an issue, I'll be happy to write-up a small PR changing it.

cc. @birdsarah @canavandl @bryevdv

PR #3803, opened 2016-02-02
Publish bokehjs separately in a notebook to avoid parsing issues in jquery

mattpap:
Publishing separately as `application/json` allows to use a different code path in ipython notebook, which doesn't require jquery (just a simple eval) and thus allows to load bokehjs without issues. I also added simple notification when bokehjs is actually loaded and improved styling of bokehjs' notebook banner.


mattpap:
@bryevdv, this is fixes notebook issues with #3794.

canavandl:
Awesome

The code looks good, the resources loaded as expected and the sauce report LGTM. I'm going to go ahead and merge then kick off another dev build. thanks @mattpap

PR #3809, opened 2016-02-03
release updates

bryevdv:


damianavila:
@bryevdv you only need to change the ALL_VERSIONS thing... anything else should be automated in the `deploy` script

PR #3810, opened 2016-02-03
Apply patch from 3769 issue

### LINKED ISSUE #3769: reset tool resets selection incorrectly
It sets to `[]` in some cases, which is not a valid selection object. 

This patch fixes
```
diff --git a/bokehjs/src/coffee/models/plots/plot.coffee b/bokehjs/src/coffee/models/plots/plot.coffee
index 67a399a..8eb91b2 100644
--- a/bokehjs/src/coffee/models/plots/plot.coffee
+++ b/bokehjs/src/coffee/models/plots/plot.coffee
@@ -286,12 +286,16 @@ class PlotView extends ContinuumView
 
   update_selection: (selection) ->
     for renderer in @mget("renderers")
-      if renderer instanceof GlyphRenderer.Model
-        selected = selection[renderer.id] or []
-        renderer.get('data_source').set("selected", selected)
-
+      if not renderer instanceof GlyphRenderer.Model
+        continue
+      ds = renderer.get('data_source')
+      if selection?
+        if renderer.id in selection
+          ds.set("selected", selection[renderer.id])
+      else
+        ds.get('selection_manager').clear()
   reset_selection: () ->
-    @update_selection({})
+    @update_selection(null)
 
   _update_single_range: (rng, range_info, is_panning) ->
     # Is this a reversed range?
```

birdsarah:
Fixes #3769

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot.coffee
+++ b/bokehjs/src/coffee/models/plots/plot.coffee
@@ -286,12 +286,17 @@ class PlotView extends ContinuumView
 
   update_selection: (selection) ->
     for renderer in @mget("renderers")
-      if renderer instanceof GlyphRenderer.Model
-        selected = selection[renderer.id] or []
-        renderer.get('data_source').set("selected", selected)
+      if not renderer instanceof GlyphRenderer.Model
+        continue
+      ds = renderer.get('data_source')
+      if selection?
+        if renderer.id in selection
+          ds.set("selected", selection[renderer.id])
+      else
+        ds.get('selection_manager').clear()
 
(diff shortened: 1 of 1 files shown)


birdsarah:
I haven't tested this, I just made a PR from @bryevdv's patch on #3769 - which applied cleanly.

The code looks good to me.

bryevdv:
(I had tested it)

bryevdv:
Ugh there is a bug `if not renderer instanceof GlyphRenderer.Model` needs to be `if not (renderer instanceof GlyphRenderer.Model)`

PR #3812, opened 2016-02-03
use import_required for nbformat and nbconvert

electronwill:
Use import_required for nbformat and nbconvert and move the import
into __init__ so that not having Jupyter Notebook installed doesn't
cause errors when attempting unrelated uses of Bokeh.


electronwill:
@bryevdv please review.

bryevdv:
LGTM

Issue #3774, opened 2016-01-28
MPL shifting pandas boxplot


chbrandt:
bokeh.__version__ : '0.11.0'
pandas.__version__ : '0.17.1'

When using `bokeh.mpl` to display (in a notebook) the plot generated by `pandas.DataFrame.boxplot()` the resulting plot has its data shifted rightwards.
The following code snippets should reproduce it:
```
import numpy as np
import pandas as pd
signal = np.random.normal(0.5,0.1,900) 
instr = np.random.poisson(0.5,100)
sample = np.concatenate((signal,instr),axis=0)
np.random.shuffle(sample)
df = pd.DataFrame({'sample':sample})
nbins = 10
df['quantil'] = pd.qcut(df['sample'],nbins)

from bokeh.io import show,output_file
output_file('boxplot_mpl.html')

import matplotlib.pyplot as plt
df.boxplot(column='sample',by='quantil')
plt.ylim([0,1.1])
plt.xticks(rotation=90)

from bokeh import mpl
show(mpl.to_bokeh())
```
Here is the (wrong) figure:
![bxp_shift_wrong](https://cloud.githubusercontent.com/assets/482051/12674144/8fd04602-c681-11e5-9f6b-3aabdb8b62fe.png)

If instead we use (in place of the last two lines) matplotlib directly
```
plt.show()
```
we should see the correct plot:
![bxp_shift_right](https://cloud.githubusercontent.com/assets/482051/12674157/a934e4ae-c681-11e5-8f7a-fa6c8857de26.png)
 

There is notebook I made available, if it helps. At the very end of [this notebook](http://nbviewer.jupyter.org/github/chbrandt/pynotes/blob/master/issues/bokeh/boxplot_report.ipynb) we can see the difference/error.

[]

chbrandt:
Related/same issues #2277 , #3130

bryevdv:
@chbrandt is there any chance that you could take a look into this issue? There's now more issues/work than the current core dev team can really deal with directly, so I am trying to make a point to solicit new contributions. Any problem is almost certainly contained in one of the few files in `bokeh/core/compat`

chbrandt:
@bryevdv  Sure.

Issue #3130, opened 2015-11-16
Bokeh does not plot pandas boxplot correctly using mpl.to_bokeh()


mmcdan:
Hello, I am trying to render a pandas boxplot using mpl.to_bokeh(). It looks like there is an off-by-one error somewhere in the groups. The data itself looks fine. Screenshots and code below:

Matplotlib:
![image](https://cloud.githubusercontent.com/assets/8503476/11175184/3b6cf52e-8be4-11e5-955c-0faa3aa59e9f.png)

Bokeh 0.10.0:
![image](https://cloud.githubusercontent.com/assets/8503476/11175206/74251838-8be4-11e5-82bd-03ede3e11ed8.png)

Code:
```python
from bokeh.plotting import output_file, show
from bokeh.sampledata.iris import flowers
import matplotlib
matplotlib.use('TKAgg')
from bokeh import mpl

flowers.boxplot(column='petal_length', by='species')
output_file("pd_boxplot.html")
show(mpl.to_bokeh())
```

Issue #2277, opened 2015-05-12
Matplotlib to bokeh conversion discards category labels


CatherineH:
I'm trying to get seaborn violin plots working in bokeh. I'm following [this example](http://bokeh.pydata.org/en/latest/docs/gallery/violin.html?highlight=violin) . My category labels seem to get discarded on conversion to bokeh. Here's some minimal code:
```python
import numpy as np
import matplotlib.pyplot as plt
import seaborn as sns
from bokeh import mpl
from bokeh.plotting import show

# generate some random data
data = 1 + np.random.randn(20, 6)
my_names = ["kirk","spock","bones","scotty","uhura","sulu"]

# Use Seaborn and Matplotlib normally
my_mpl = plt.figure()
#when seabord is called directly, names works for assigning categories
sns.violinplot(data, color="Set3", names=my_names)
my_mpl.savefig("mpl_test.png")
my_bokeh = mpl.to_bokeh(my_mpl)
print type(my_bokeh)
# Convert to interactive Bokeh plot with one command
show(my_bokeh)
```

This is the image that matplotlib produced:
![mpl_test](https://cloud.githubusercontent.com/assets/1953328/7593659/c7295b08-f8a9-11e4-9779-c760c5b833e7.png)
and the image after converting to bokeh:
![mpl_test2](https://cloud.githubusercontent.com/assets/1953328/7593675/dbfdae12-f8a9-11e4-9c8c-212deb3c537e.png)

philippjfr:
Just as a quick update to this issue. It seems to get the categorical axis labels now but they are drawn in the wrong place. Here's a quick example:

```python
import seaborn as sns
from bokeh import mpl
from bokeh.io import output_notebook, show

output_notebook()
iris = sns.load_dataset("iris")
sns.violinplot(iris.sepal_length, iris.species)
show(mpl.to_bokeh())
```

![image](https://cloud.githubusercontent.com/assets/1550771/9976360/b3da1452-5ed9-11e5-839d-fd1bb8b590e3.png)

chbrandt:
@philippjfr As of versions bokeh-0.11.0 and seaborn-0.7.0 the plot works fine.

PR #3814, opened 2016-02-04
Fix boxplot shifting when using 'mpl' from matplotlib and seaborn.

chbrandt:
Fix issues #3774, #3130, #2277

Issue 2277 was already working (as commented there). This fix took care of that.

[notebook](http://nbviewer.jupyter.org/github/chbrandt/pynotes/blob/4c8eab1508105ddaf32d24178cd10634d65cebaf/issues/bokeh/boxplot_report.ipynb) updated with examples from those issues.


bryevdv:
This is awesome, thanks @chbrandt ! We are in the middle of the `0.11.1` release, but as soon as that is done we will get this merged and in a dev build as well.

damianavila:
Code LGTM and I tested this successfully, so :+1:

bryevdv:
Thanks @chbrandt !

PR #3829, opened 2016-02-04
Add Stamen toner background tile provider

brendancol:
This PR adds an additional tile provider to `tile_providers.py` so you can do the "map sandwich" with `STAMEN_TONER_BACKGROUND` below and `STAMEN_TONER_LABEL` above.


bokeh/tile_providers.py
```diff
@@ -14,6 +14,17 @@
     )
 )
 
+#: Provide toner background tiles from http://tile.stamen.com
+STAMEN_TONER_BACKGROUND = WMTSTileSource(
+    url='http://tile.stamen.com/toner-background/{Z}/{X}/{Y}.png',
```
birdsarah: The docs at [maps.stamen.com](http://maps.stamen.com) seem to indicate the url is: http://tile.stamen.com/toner/{z}/{x}/{y}.png

brendancol: yes that is another service which is already included in `tile_providers.py`.  this is version which does not have labels baked in.

birdsarah: Would be good to add that information to the comment so that it's in the docs.

Would also be good to add (presumably in the docstring at the top of this file) a link to docs for stamen that we can monitor for changes to their api.

birdsarah:
Tested this locally - working great.

bokeh/tile_providers.py
brendancol: - [ ] add docstring w/link to stamen
- [ ] add to docs

bryevdv:
Marking this WIP pending docs tasks.

bryevdv:
I was actually thinking it would be appropriate to add documentation to the User's Guide.

PR #3846, opened 2016-02-09
Real properties for BokehJS

bryevdv:
ping @bokeh/core 

This PR is not quite done but I'm pushing branch now to be able to see test report, and because  I would like to get some eyes on it earlier rather than later. 

Real properties in BokehJS will facilitate:

* getting rid of `nonserializable_attribute` junk
* automatic visual configuration GUIs
* mixin props machinery now lives at proper place in class hierarch
* automatic matching of properties between python and JS under test
* better to test

A few other things I did along the way to simplify things in order to be able to get through everything:

* removed all support for (Backbone model) property setters


bryevdv:
This basic idea is on models you can do any combination of this:
```
  mixins: ['line']

  props: ->
      render_mode:    [ p.RenderMode,   'canvas'  ]
      x_range_name:   [ p.String,       'default' ]
      y_range_name:   [ p.String,       'default' ]
      location:       [ p.Number,       null      ]
      location_units: [ p.SpatialUnits, 'data'    ]
      dimension:      [ p.Dimension,    'width'   ]
    }

  defaults: ->
      # overrides
      line_color: 'black'

      # internal
      for_hover: false
    }
```

havocp:
As you're typing in "defaults" and "props" fields remember that both could be auto-generated :-) with the hard parts already in test_defaults.coffee and generate_defaults.py - the more you make the JS look just like the Python the easier it will be.

I was debating the best way to do this - one approach is similar to how it is now for the tests, where there's one big generated file and props/defaults could be looked up by class name. Another idea would be to generate one file per class (perhaps those files contain mixin classes, so the generated part of each class is mixed in to the hand-coded part?). I didn't have a chance to play with this at all so not sure.

Anyway this and your earlier PR to clean up JS props stuff are great to see.

bryevdv:
> As you're typing in "defaults" and "props" fields remember that both could be auto-generated :-)

It's definitely on my mind, either a rigorous 1-1 check or generation from a single source. But one step at a time :) I'm not sure the JS side can be as easily expressive as the python side (instances of particular types, lists of particular types, etc) so there are more `p.Any` on the JS side now, but maybe those can be improved more and more easily than I am thinking,

bryevdv:
OK I think we are down to one issue (or close to that): how to deal with optional exclusive attributes, e.g circle radius vs size

bryevdv:
@bokeh/core I will test more extensively tonight and try to add more tests, but I think this may be ready for close inspection.

bryevdv:
known issues (I think some of these are related)

* twin axis, circles on extra range not plotting

* fourier_animated not plotting

* continuous notebook not working

* widget weird sizes for OHLVC

screen shot 2016-02-10 at 12 56 17 pm

canavandl:
Add props to layouts seemed fix the ohlc example

screen shot 2016-02-10 at 5 49 16 pm

bryevdv:
@bokeh/core I have not tested *everything* but have tested many things, and the few remaining issues have been resolved. I plan on merging this tomorrow if there is no objection, and fixing any other issues in follow-on PRs.

@mattpap any comments welcome especially.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
```diff
@@ -31,20 +34,20 @@ class AnnularWedgeView extends Glyph.View
...
       ctx.lineTo(sinner_radius[i], 0)
-      ctx.arc(0, 0, sinner_radius[i], 0, -angle[i], not direction[i])
+      ctx.arc(0, 0, sinner_radius[i], 0, -angle[i], not direction)
       ctx.closePath()
 
       ctx.rotate(-angle[i]-start_angle[i])
       ctx.translate(-sx[i], -sy[i])
 
-      if @visuals.fill.do_fill
+      if @visuals.fill.do
```
mattpap: Surprising this works. `do` is a keyword in CoffeeScript. I suggest renaming this to something else, e.g.  `doit`.

bokehjs/src/coffee/model.coffee
```diff
@@ -1,13 +1,13 @@
...
 
 class Model extends HasProps
   type: "Model"
 
-  defaults: ->
-    return _.extend {}, super(), {
-      tags: []
-      name: null
-    }
+  props: () -> _.extend super(), {
```
mattpap: You have to decide whether you use `{}` as the first argument or not, and do it consistently. The reason we use `{}` is to make sure `_.extend` doesn't modify `super()` in-place. As long as you can assure that `props()` always returns a fresh instance of an object, then `{}` is unnecessary. However, if at any point `props()` returns a reference to, lets say, a class member, then it will modify the global state and lead to a subtle bug.

examples/plotting/server/fourier_animated.py
```diff
@@ -62,7 +62,7 @@ def create_plot(foos, title='', r = 1, y_range=None, period = pi/2, cfoos=None):
         if i==len(foos)-1:
             # if it's the last foo let's draw a circle on the head of the curve
             sources['floating_point'] = CDS({'x':[shift], 'y': [cy]})
-            p.line('line_x', 'line_y', color="palette[i]", line_width=2, source=sources['lines'])
+            p.line('line_x', 'line_y', color=palette[i], line_width=2, source=sources['lines'])
```
mattpap: So `color` was interpreted as a field?

bokehjs/test/models/glyphs/glyph.coffee
```diff
@@ -8,37 +8,37 @@ view_for = (model) ->
...
+  #     expect(view.distances.inner_radius.units).to.be.equal 'data'
+  #     obj.set({ 'inner_radius' : { 'value' : 1, 'units' : 'screen' } })
+  #     expect(view.distances.inner_radius.units).to.be.equal 'screen'
+
+  #   it "should update units when an angle value is set", ->
+  #     obj = Collections('AnnularWedge').create()
+  #     view = view_for(obj)
+  #     expect(view.angles.start_angle.units).to.be.equal 'rad'
+  #     obj.set({ 'start_angle' : { 'value' : 60, 'units' : 'deg' } })
+  #     expect(view.angles.start_angle.units).to.be.equal 'deg'
```
mattpap: This should be either removed or a comment should be added.

bryevdv: Oh right, it got late last night.  I will move these to the `Renderer` class where they belong now.

bokehjs/src/coffee/model.coffee
bryevdv: This was just a mistake, great catch thanks

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
bryevdv: will do(it)

examples/plotting/server/fourier_animated.py
bryevdv: Yes, and the error was silently ignored, it seems

bokehjs/test/models/glyphs/glyph.coffee
bryevdv: Actually I think these are redundant and covered by the properties tests now, so I will just delete. Going to make a separate issue to list places that most need more testing in BokehJS

bryevdv:
@mattpap I've addressed your previous comments

canavandl:
I think this LGTM. I've spot-tested a ton of html/notebook/server examples and they seem to be working.

mattpap:
LGTM

bryevdv:
OK will merge once travis is green

Issue #3850, opened 2016-02-09
Problem with formatting on docstring for `output_file`


birdsarah:
http://bokeh.pydata.org/en/0.11.1/docs/reference/core.html#bokeh.core.state.State.output_file

<img width="868" alt="screen shot 2016-02-08 at 8 48 28 pm" src="https://cloud.githubusercontent.com/assets/1796208/12907867/5f1bd50e-cea5-11e5-9fbe-48e4e6cfb441.png">

PR #3851, opened 2016-02-09
Fix docstring

birdsarah:
Fixes: #3850


bryevdv:
spurious test failure, merging

PR #3854, opened 2016-02-10
Simplify shapefile (.shp) w/ Bokeh Server + GeoPandas

brendancol:
This example demonstrates consuming a shapefile and displaying it via a `GeoJSONDataSource`. Couple of things to note (which are also in README.md):

- installing geopandas can be a pain. Fortunately there is a conda channel with geopandas:
`conda install -c ioos geopandas` (this should be copied the bokeh channel)
- to get example shapefile...`import bokeh.sampledata; bokeh.sampledata.download()`

remaining items:
- [ ] move geopandas to bokeh channel: `anaconda copy ioos/geopandas/0.1.1  --to-owner bokeh`

### DIFF
--- a/bokeh/sampledata/__init__.py
+++ b/bokeh/sampledata/__init__.py
@@ -112,12 +112,17 @@ def _getfile(base_url, file_name, data_dir, progress=True):
     real_name, ext = splitext(file_name)
 
     if ext == '.zip':
+
         if not splitext(real_name)[1]:
             real_name += ".csv"
 
         print("Unpacking: %s" % real_name)
 
-        with ZipFile(file_path, 'r') as zip_file:
-            zip_file.extract(real_name, data_dir)
+        if splitext(real_name)[1] == '.shp':
+            with ZipFile(file_path, 'r') as zip_file:
+                zip_file.extractall(data_dir)
+        else:
+            with ZipFile(file_path, 'r') as zip_file:
+                zip_file.extract(real_name, data_dir)
 
         remove(file_path)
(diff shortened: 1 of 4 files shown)


birdsarah:
waaahhhhh :D :D :D 

I know this is a WIP but I couldn't help myself and got it running locally - this is soooo cool.

brendancol:
Yes so first thing I need is to remove the hardcoded shapefile path and add my shapefile to sampledata...coming soon.

birdsarah:
@brendancol - my jumping in too early two cents:

1) mention the geopandas install in the README - when you `bokeh serve` something the import error doesn't work as you'd expect
2) i'm concerned that the geopandas install is non trivial - i had to use my "geo" env that if you recall we had some fun setting up previously.

Would definitely recommend testing this on clean env and clean machine and documented what is tested to work. I think it's worth it though - this is a really nice example of pulling a bunch of libraries together to do something non-trivial.

Also fwiw - I don't think you need the save button feature.

brendancol:
@birdsarah 

> 1) mention the geopandas install in the README - when you bokeh serve something the import error doesn't work as you'd expect

Yes definitely need that in the README.  Regarding this import try/except, Im think maybe of adding a `Paragraph` instance to curdoc to display message in the browser

>  i'm concerned that the geopandas install is non trivial - i had to use my "geo" env that if you recall we had some fun setting up previously.

I talked to @hsparra about the install and he's going to serve as an impartial user to test the pain/lack-there-of for installing Geopandas.

> Also fwiw - I don't think you need the save button feature.

So not preserving shared boundaries is a bit lame, but I wanted this example to demonstration both reading and writing of shapefiles.

brendancol:
https://github.com/geopandas/geopandas/issues/237

hsparra:
Regarding this, I was able to run

conda install -c ioos geopandas

on a clean environment and the install went smoothly. I was then able tot
do "import geopandas" with no errors.

bryevdv:
Resolved to locate in separate project

PR #3863, opened 2016-02-11
Mess with using canvas2svg

birdsarah:
the main tweak that seems to need to be made is removing a transform. here's the diff

```diff
7c7
< <g transform="scale(2,2) translate(0.5,0.5) scale(2,2) translate(0.5,0.5)">
---
> <g transform="">
16c16
<     <g clip-path="url(#bhvdFFmELTIC)">
---
>     <g>

```

Depending on what program is rendering the svg, the clip path seems to matter or not. github's svg renderer seems to care. See the svg output here: https://github.com/bokeh/bokeh/blob/canvas2svg/examples/models/glyph1_tweaked.svg

I've included the svgs in this PR so you can see.

The original svg (`glyph1_raw.svg`) was made by running `window.c2s.getSerializedSvg(true)` at the browser console.

Obviously we wouldn't attach the c2s object to the window - just hacking about to see what's possible.


birdsarah:
superseded by #3867

Issue #538, opened 2014-04-18
Headless static (svg, png) image generation 


abastardi:
Enable generation of PNG images directly on the server, similar to the Vega.js headless mode (https://github.com/trifacta/vega/wiki/Headless-Mode).

scherrey:
I see that .8 has shipped but this issue is still open. Is there an update as to the disposition of this capability?

bryevdv:
@scherrey Unfortunately not. Here is a message I sent recently the Bokeh mailing list on 2015/03/04:

----

Programmatic generation of static images is definitely something we would very much like to support. Unfortunately, it is also a surprisingly difficult proposition. The first observation is that browsers canvas implementation only generate raster images, and don't help with SVG or PDF. Worse, browser security models mean that getting the data out is also very problematic. You can follow this history of this topic in a few issues:

https://github.com/bokeh/bokeh/issues/538
https://github.com/bokeh/bokeh/issues/1589

MPL support is currently provided by a third party library, `mplexporter`. It is very useful in a large number of cases, but it is by no means complete, which explains errors you have seen. There is a MPL enhancement proposal to make a "serious" complete JSON serialization layer, much better than mplexorter. In principle this would work fantastically, allowing us to leverage all the MPL work for static image generation. But realistically, there are no indications that this work will materialize or even really get started any time in the near future (best guess: years, if at all). The most recent plan was to try and use the current BokehJS patched with node-canvas to generate images inside Node. But BokehJS is a large and complicated library and it is unclear that Node support is currently up to the task. Experiments and tests are ongoing, but not encouraging. 

If that plan proves unworkable, I am afraid that really only leaves the very undesirable proposition of creating and maintaining our own "duplicate" rendering code path in python that mirrors the BokehJS rendering code. This is obviously bad from an effort/maintenance viewpoint, but it's also disappointing because it means static image generation won't "just work" for other language bindings without including some python code or a python tool. However, given that we are probably going to make a move to WebGL for browser rendering, this may truly be the "best worst" possibility. But obviously it is a nontrivial undertaking that will take a nontrivial amount of time and effort. 

I wish I had some better news to give you. It is my earnest hope that we will have a solution in 2015 but I can't provide any better estimates at this time.

mfripp:
[...]

PR #3867, opened 2016-02-11
Ability to save as svg

birdsarah:
Partially fixes: #538 

To test:
1. Run any file example
2. Add `?svg=1` to the end of the html file
3. The canvases will go blank
4. Click the `preview/save` button - it will download the svg of the plot

To Do:
- [x] display either the original canvas or the svg'd canvas in the whitespace (not sure why this is happening)

Notes:
* canvas2svg apparently works with node-canvas, so if the other work getting node-canvas integrated works, then we should be able to programatically get out png or svg
* the addition of canvas2svg increases bokeh.js from 1649KB to 1691KB (an increase of 42KB)


birdsarah:
This is ready for review and as far as I can tell is working.

Functionality notes:
* There are some glyphs that don't work due to the limitations of canvas2svg
* Interaction works although it's very slow - on each render the new svg file is rendered to the screen, as a result, you can interact with the plot and save an interacted version
* There are intermittent grey lines at the borders that come and go and are not wanted

For code review:
* This was done quickly, happy to hear what the "right" way of setting an svg mode and testing if it's on is

birdsarah:
Given that this isn't perfect, but is a highly desired feature - we could maybe add it as "experimental" support or something like that.

birdsarah:
To see which glyphs are (not) supported: https://rawgit.com/birdsarah/bokeh-miscellany/master/test_svg_on_glyphs.html?svg=1

To see normal bokeh performance: https://rawgit.com/birdsarah/bokeh-miscellany/master/test_svg_on_glyphs.html

birdsarah:
Marking this as WIP because @bryevdv has much better ideas of how it should be structured and is doing some work to support that.

bryevdv:
superseded by #3867 (reopen if I am mistaken)

birdsarah:
:person_frowning:  you superceded this with itself?

humm:
So I am testing the feature (great work! absolutely fantastic!), and it seems to work only when output_file is called in `inline` mode, whereas the default seems to be `cdn` (I have *no idea* what the `inline` and `cdn` are, just figured out that `inline` worked). Is that by design?

birdsarah:
@humm - see: http://bokeh.pydata.org/en/latest/docs/reference/io.html#bokeh.io.output_file `cdn` and `inline` refer to where bokehjs gets loaded from. `cdn` would definitely not be expected to work because this branch is not in a published release of bokeh. I was going to point you to an explainer in the docs (http://bokeh.pydata.org/en/0.11.0/docs/user_guide/quickstart.html#resources) and discovered the docs are wrong (the default is now `cdn`) - so thanks for bringing this up!!!! :D

guillochon:
This is great and I hope it makes it into the release soon!

bryevdv:
@birdsarah I am gong to close this PR but keep the branch for reference (in anticipation of a future version of this after the layout work)

Horta:
That is a great feature!

Issue #3827, opened 2016-02-04
Cross browser automated testing


hsparra:

birdsarah:
This isn't quite finished yet, #3869 put in place the basic machinery.

PR #3869, opened 2016-02-12
Test with firefox on saucelabs

birdsarah:
Partially fixes: #3827 

Supercedes: #3741


birdsarah:
integration test passed at b463cab

birdsarah:
Our selenium tests are currently running an a version of firefox installed on travis. This has the following problems:
1) we don't have control over which exact version of firefox we are running
2) we can't expand to cross-browser testing

By integrating with SauceLabs we'll be able to test against multiple browser / OS combinations including mobile devices. SauceLabs offers free testing to open source projects on their OpenSauce plan (http://saucelabs.com/opensauce/).

In addition, SauceLabs gives a video of every test that ran so it's easy to see what happened: https://saucelabs.com/tests/467e2c5f99a54163a5a25ac351a4a600

This PR also integrates pytest-selenium's html report and uploads them to S3. These nicely integrate the returned details from saucelabs. Here is an all passing test report: https://s3.amazonaws.com/bokeh-travis/104764781/report.html

And here is one with a failing test, the video from saucelabs along with the traceback is included in the report: https://s3.amazonaws.com/bokeh-travis/104811980/report.html

screen shot 2016-01-25 at 9 32 07 pm

There is still lots more to do on testing and using the full capabilities to try and reduce our manual burden of testing, but for now this is a one-for-one switch between Travis Firefox and Saucelabs Firefox.

@bryevdv previously had a concern about the time increase, this has been eliminated by using pytest-xdist which runs tests in parallel (I plan to implement examples tests in pytest so that we can use them with pytest-xdist and speed them up).

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+            upload = False
+
+        if upload is True:
+            with open(report_file, "r") as f:
+                html = f.read()
+            filename = join(build_id, "report.html")
+            key = S3Key(bucket, filename)
+            key.set_metadata("Content-Type", "text/html")
+            key.set_contents_from_string(html, policy="public-read")
+            print("\n%s Access report at: %s" % ("---", join(s3, filename)))
```
bryevdv: This seems like it woudl be clearer to me:
```
        try:
            conn = boto.connect_s3()
            bucket = conn.get_bucket(s3_bucket)

            with open(report_file, "r") as f:
                html = f.read()

            filename = join(build_id, "report.html")
            key = S3Key(bucket, filename)
            key.set_metadata("Content-Type", "text/html")
            key.set_contents_from_string(html, policy="public-read")
            print("\n%s Access report at: %s" % ("---", join(s3, filename)))

        except NoAuthHandlerFound:
            print("Upload was requested but could not connect to S3.")

```

bryevdv: would a check / print of the `exitstatus` be useful?

bryevdv:
A few minor comments otherwise LGTM

birdsarah:
Thanks, will fix!

tests/conftest.py
birdsarah: The report will be generated regardless of exitstatus, other parts of `pytest` that care to handle `exitstatus` will do so, this doesn't override pytests existing sessionfinish behavior, it just augments it (at least that's my understanding)

birdsarah:
@bryevdv, this is ready again. May be worth noting that #3896 continues to improve on this.

PR #3872, opened 2016-02-12
more bokehjs cleanup

bryevdv:
this was the start of my canvas layout PR but I think it makes more sense as a smaller PR of its own. It:

* adds test selection, e.g. `gulp test:core`

* moves more things into `core/utils` that make sense there


bryevdv:
@bokeh/core this is ready for review

bryevdv:
@bokeh/core have not heard back any objections, but I think everything here should be fairly un-controversial, so merging now

Issue #3871, opened 2016-02-12
MultiSelect: on_change method does not work properly


petermountanos:
I am having a problem with the MultiSelect widget in v0.11.1, particularly with the on_change method. It seems there's also [other people ](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/0OGyHe9tpGE)having this issue as well.
    
```python
from __future__ import print_function
from bokeh.models import MultiSelect, Select, CustomJS, VBox
from bokeh.io import curdoc


def update(attrname, old, new):
    print("""MultiSelect\nOptions: %s\nSelected:%s\n\n
             Select\nOptions:%s\nSelected: %s"""
          % (mscolors.options, mscolors.value, colors.options, colors.value))

# set up select widget
colors = Select(title="colors", options=["Red", "Green", "Blue"])
colors.on_change('value', update)
colors.callback = CustomJS(code="console.log('Select: ' + cb_obj.get('value'))")

# set up multiselect widget
mscolors = MultiSelect(title="Colors", options=["Red", "Green", "Blue"])
mscolors.on_change('value', update)
mscolors.callback = CustomJS(code="console.log('MultiSelect: ' + cb_obj.get('value'))")

curdoc().add_root(VBox(colors, mscolors))
````
Run using ```bokeh server --show main.py```.

As identified by @birdsarah changing the selected value on the UI doesn't actually update the HTML selected flags. Therefore, the ```on_change``` callback function can't properly listen for a change. However, if you check the JS console, the ```CustomJS``` callback clearly works. I've tried to debug the issue myself, but I can't seem to figure it out.

When changing the value of the Select widget, I'll get output in my Python shell like: 
```
Select
Options:['Red', 'Green', 'Blue']
Selected: Green

MultiSelect
Options: ['Red', 'Green', 'Blue']
Selected:[]
```
But when I change the value of the MultiSelect widget, I get no output in the Python shell. Like mentioned above, I get the expected output to the JS console. Thanks.

===Version Information===
Bokeh version: 0.11.1
Python version: 2.7.11-CPython
Platform: Darwin-15.3.0-x86_64-i386-64bit
Browser: Chrome

bryevdv:
Guessing this is the problem:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/multiselect.coffee#L40

[...]

PR #3873, opened 2016-02-12
Issue #3871: MultiSelect - Remove 'silent': true

petermountanos:
Remove the {‘silent’: true} option in the MultiSelect coffee script to
debug issue #3871 in the Bokeh master. It seems to have fixed the issue.

### DIFF
--- a/bokehjs/src/coffee/models/widgets/multiselect.coffee
+++ b/bokehjs/src/coffee/models/widgets/multiselect.coffee
@@ -37,7 +37,7 @@ class MultiSelectView extends BokehView
     )
 
   change_input: () ->
-    @mset('value', @$('select').val(), {'silent': true})
+    @mset('value', @$('select').val())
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @petermountanos ! I'd like to do a quick check to make sure there is not eventing issues (seems not), and having the PR ready makes that alot easier to do.

petermountanos:
Great, thanks! I didn't come across any issues, but would be great if you can check it out.

petermountanos:
CI tests failed, specifically the Python: 3.4, GROUP=examples batch. It seems the failure was caused by a timeout

bryevdv:
restarted, we get an annoying number of spurious failures on TravisCI

bryevdv:
Confirmed no ping-ponging, merging now. Thanks @petermountanos !

PR #3874, opened 2016-02-13
Win entry point

damianavila:
Supersede: #3792

Right now the package lives in the dev2 channel, please test it if you have a win machine:

`conda install -c bokeh/channel/dev2 bokeh=0.11.1dev3`

ping @stonebig

### DIFF
--- a/bokeh/__main__.py
+++ b/bokeh/__main__.py
@@ -1,5 +1,13 @@
 import sys
 
-from bokeh.command.bootstrap import main
+from bokeh.command.bootstrap import main as _main
 
-main(sys.argv)
+def main(args=None):
+    """The main"""
+    if args is None:
+        args = sys.argv[1:]
+
+    _main(sys.argv)
+
+if __name__ == "__main__":
+    main()
--- a/setup.py
+++ b/setup.py
@@ -581,6 +581,11 @@ setup(
         'bokeh.util.tests',
     ],
     package_data={'bokeh': package_data},
+    entry_points={
+        'console_scripts': [
+            'bokeh = bokeh.__main__:main',
+        ],
+    },
     author='Continuum Analytics',
     author_email='info@continuum.io',
     url='http://github.com/bokeh/bokeh',


almarklein:
LGTM. Just for my understanding, would simply `entry_points={'console_scripts': ['bokeh= bokeh.__main__'], },` not work?

damianavila:
>Just for my understanding, would simply entry_points={'console_scripts': ['bokeh= bokeh.__main__'], }, not work?

Nop, entrypoints need to reference a function without args... this is why I wrapped `_main(args)` into main() on `__main__`

bryevdv:
I think it might work (and without the wrapper)? I would find it strange if every python package with entry points had to jump through that kind of boilerplate hoop.

damianavila:
It is requesting funcs with no args:

http://pythonhosted.org/setuptools/setuptools.html?highlight=console#automatic-script-creation

So I don't think it will work w/o the wrapper...

bryevdv:
In that case I don't see any point for the default `None` arg. Why not just:
```
import sys

from bokeh.command.bootstrap import main

def run():
    main(sys.argv[1:])

if __name__ == "__main__":
   run()
```
But I'm also wondering about the `[1:]` that was added. Neither your version or mine preserves any code path that executes main without slicing `argv`, but the original certainly did. Was the original in error?

bokeh/__main__.py
```diff
@@ -1,5 +1,13 @@
...
 
-from bokeh.command.bootstrap import main
+from bokeh.command.bootstrap import main as _main
 
-main(sys.argv)
\ No newline at end of file
+def main(args=None):
+    """The main"""
+    if args is None:
+        args = sys.argv[1:]
```
bryevdv: actually wait, `args` is not used at all

bryevdv:
Actually `args` is not used at all, I think this is what is needed:
```
import sys

from bokeh.command.bootstrap import main

def run():
    main(sys.argv)

if __name__ == "__main__":
   run()
```

bryevdv:
Also we are using `distutils.core.setup` not `setuptools`? And we have an actually checked-in (not generated) `bin/bokeh`?

bryevdv:
(We can talk about using `setuptools`, I just don't understand how this could work as-is without it)

damianavila:
> Actually args is not used at all, I think this is what is needed:

Yep... I should not write code on late Friday after a full working day :wink:

My idea was to actually use `args`... in any case, you simplification looks ok to me but I have to test it yet

>Also we are using distutils.core.setup not setuptools? And we have an actually checked-in (not generated) bin/bokeh?

Yes, we are using distutils... and with distutils a bat bokeh file is created... even with no entry point reference... but for the exe we need setuptools... regrettably.
Let me iterate on this again. I will ping you when it is ready.

damianavila:
Just for your info... I have being testing a lot of thing and I can generate the exe... and tracking down the issue it seems something related with conda convert... I have to keep investigating...

almarklein:
Closing, I based my noarch branch on this branch.

PR #3121, opened 2015-11-13
Use flexx.pyscript to allow Python callbacks

almarklein:
issues: implements part of #2788 
replaces #2799

Done in Bokeh:
* Modifies the CustomJS class to allow using it as a decorator for a callback function.
* Adds an example
* Adds docs (note that the result plot for the example in the docs does not work)

Done in Flexx:
* PyScript is now more Pythonic. Probably enough for most code to work as-is. Users don't need to know about JavaScript's quircks in most cases.
* PyScript now works on all Python versions 3.3+ (including pypy)
* The above is tested with CI
* Released, so it can be pip-installed.

To discuss:
- [x] I propose that flexx is an optional dependency. When trying to use this feature when flexx is not installed, we raise an informative exception,
- [x] Flexx can be pip-installed. Should we make it conda installable?

PR #3877, opened 2016-02-14
PyScript now also works on py27

### LINKED ISSUE #3675: Defining a JS callback using Python for Python 2.x
Since Flexx is currently  only available for Python 3.x, The CustomJS as Python functionality is only available on Python 3.x. Tasks needed to fix this:

* [x] Make Flexx' code work on Python 2.7 (https://github.com/zoofIO/flexx/pull/79)
* [x] Make Flexx' commonast module produce correct code using the ast module of Python 2.7
* [x] Make Flexx' test suite pass on Python 2.7
* [x] Update code in Bokeh that warns on 2.x
* [x] Update Bokeh's docs
* [x] Update pyscript to use JS-specific namespace (https://github.com/zoofIO/flexx/issues/76)
* [x] Update pyscript to use bound methods (avoid `that = this` stuff)
* [x] Flexx new release
* [ ] ~~Create conda package for Flexx~~
* [x] Change how legacy py is supported, to an approach that does not require translation, so we can make a noarch package.
* [x] Flexx release again (v0.3.1)
* [ ] Create noarch conda package for Flexx in the conda channel

almarklein:
issues: closes #3675, continues #3121

Wait with merging this until we've got a new Flexx release and corresponding conda package.

### DIFF
--- a/bokeh/models/callbacks.py
+++ b/bokeh/models/callbacks.py
@@ -38,12 +38,8 @@ class CustomJS(Callback):
         try:
             from flexx.pyscript import py2js
         except ImportError:
-            if sys.version_info < (3, ):
-                raise RuntimeError('Using Python functions for CustomJS '
-                                    'is currently not supported on Python 2.x')
-            else:
-                raise RuntimeError('To use Python functions for CustomJS, you need Flexx '
-                                   '("conda install -c bokeh flexx" or "pip install flexx")')
+            raise RuntimeError('To use Python functions for CustomJS, you need Flexx '
+                                '("conda install -c bokeh flexx" or "pip install flexx")')
         # Collect default values
         default_values = func.__defaults__  # Python 2.6+
         default_names = func.__code__.co_varnames[:len(default_values)]
--- a/bokeh/models/callbacks.py
+++ b/bokeh/models/callbacks.py
@@ -44,6 +44,7 @@ class CustomJS(Callback):
         default_values = func.__defaults__  # Python 2.6+
         default_names = func.__code__.co_varnames[:len(default_values)]
         args = dict(zip(default_names, default_values))
+        args.pop('window', None)  # Clear window, so we use the global window object
         # Get JS code, we could rip out the function def, or just
         # call the function. We do the latter.
         code = py2js(func, 'cb') + 'cb(%s);\n' % ', '.join(default_names)
--- a/bokeh/models/callbacks.py
+++ b/bokeh/models/callbacks.py
@@ -5,10 +5,12 @@ from __future__ import absolute_import
 import sys
 from types import FunctionType
 from ..model import Model
+from ..util.dependencies import import_required
 
 from ..core.properties import abstract
 from ..core.properties import Dict, Instance, String, Enum
 from ..core.enums import ScriptingLanguage
+
 
 @abstract
 class Callback(Model):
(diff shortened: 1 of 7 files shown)


almarklein:
This includes Travis' suggestion to use a namespace for functionality specific to JS (e.g. `window.Math.pow`). The purpose is to avoid both pyflakes warnings and user confusion (zoofIO/flexx#76). I've gone with the name `window` , because this is historically already a name to indicate the JS root object. If anyone has an opinion/suggestion for another base-name (as opposed to 'window'), now would be a good time to speak up :)

As for getting a stub `window` object into the module namespace (to fool pyflakes), there are a few options:
```python
from flexx.pyscript import window  # this works, but feels a bit awkward perhaps, because we don't import anything else from flexx
from bokeh.models import CustomJS, window  # we'd have to add a stub name to the module
window = None  # This is what I went with now
```
Opinions?

bryevdv:
I don't think we should make a fake `Model`. My first inclination was to suggest a new module just for this:
```
from bokeh.javascript import window
```
But it seems like that would be completely useless apart from and without `flexx` installed, so maybe `bokeh` itself is not the correct place, unless it is made very specific to `CustomJS`. Given that I see two options personally:

* `from flexx.pyscript import window`  (OK)

* `window = CustomJS.window` (also OK)

Any other ideas?

mattpap:
Can't we configure pyflakes instead of fooling? All of the suggestions seem cumbersome in the current form. If `window` is supposed to be just a dummy variable then it should be implicit and no import should be required (as in JS). Otherwise, e.g. for auto-completion purpose, `window` should be an opaque object imported either directly from `flexx` or from `bokeh.util.js`.

bryevdv:
We could but so would anyone else who wanted not have annoying flakes warnings. Also the implicitly available thing seems more confusing. Where did it come from? Python users expect to have to import things, so this gives them one to hold on to, even if it's just a placebo.

almarklein:
Mmm... I think I maybe like @mattpap's idea of adding an ignore for `window.` in flake8 (I think this is possible, right?). Importing a name into the Python module that is actually never used in Python might be similarly confusing as a missing name. Since its a name that's only used/available in JS, maybe it makes sense not having the name in Python-land.

By the way, there is also `undefined`, for which there is a similar story (though `undefined` is an actual object in Flexx, that's also used to indicate unset signal values.

Another option for having the name in the module (if the flake8-ignore is not getting enough support), is to make it very clear that the name is a stub and only used in JS:
``` python
window = 'only used in JS'  # or
window = None  # only used in JS
```

bryevdv:
Well like I said the problem with "fixing up" flake8 is that *everyone* who wants to not have lint warnings also has to do that. I don't think foisting that task off to users is a good idea, personally.

almarklein:
Ok. I think my preference goes to do this in the examples:
```
window = None  # name only used in JS - defined here to silence pyflakes
```
This way, there is nothing magical about it and its clear what this is about. I'll update the comment if everyone agrees.

mattpap:
> Importing a name into the Python module that is actually never used in Python might be similarly confusing as a missing name.

This is a good thought. Mixing Python and Python-like language will be indeed confusing.

Is pyflakes going to complain about `import window` in a callback (I presume so)? E.g.:
```
def callback():
    import window
    window.alert()
```
Another option is to put `window` as an argument to a callback.

bryevdv:
> Another option is to put `window` as an argument to a callback.

I'd prefer that out of all the things suggested so far

almarklein:
Nice idea. I implemented it, and I think it works great. Travis is green (flexx is available in the bokeh channel). Ready from my end.

sphinx/source/docs/user_guide/interaction.rst
```diff
@@ -290,13 +290,9 @@ for users to define client-side interactions without having to learn
 JavaScript. To use this functionality you need the Flexx library
 (install with ``conda install -c bokeh flexx`` or ``pip install flexx``).
```
bryevdv: I think a note somewhere here about the default `window` argument is needed.

bokeh/models/callbacks.py
```diff
@@ -38,16 +38,13 @@ def from_py_func(cls, func):
         try:
             from flexx.pyscript import py2js
         except ImportError:
-            if sys.version_info < (3, ):
```
bryevdv: Let's be consistent and use this everywhere to manage warning about imports that are required for optional features: https://github.com/bokeh/bokeh/blob/master/bokeh/util/dependencies.py#L23

bryevdv:
@almarklein I left two small comments, then I think this PR will be perfect.

almarklein:
done

PR #3882, opened 2016-02-16
Add KML --> GeoJSONDataSource Example

brendancol:
This is a short example which parses a kml file of world countries into a `GeoJSONDataSource`. 

Note:
- There are dependencies on `fastkml (pip install fastkml)` and `geojson (pip install geojson)`
- To get source data you will need to:
```python
import bokeh.sampledata
bokeh.sampledata.download()
```

screen shot 2016-02-15 at 9 14 11 pm

### DIFF
--- a/bokeh/sampledata/__init__.py
+++ b/bokeh/sampledata/__init__.py
@@ -74,6 +74,7 @@ def download(progress=True):
         (s3, 'world_cities.zip'),
         (s3, 'airports.json'),
         (s3, 'movies.db.zip'),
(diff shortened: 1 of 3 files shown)


birdsarah:
looks like you need to add fastkml in as a test dependency for examples to pass, alaso suggest merging the lastest testing work from master.

bryevdv:
I actually think this example might be better in a narrative section of the user's guide. Thought's, @bokeh/dev ?

bryevdv:
Resolved to place in separate project.

PR #3887, opened 2016-02-17
[WIP] Phosphor layouts for HBox and Vbox

dwillmer:
Please don't merge yet, still a few things to sort out:

- [x] Get the HBox and VBox View classes built and rendering with phosphor BoxPanels.
- [x] Remove the additional divs from the hierarchy that Backbone.js introduces.
- [ ] In the HBox-inside-VBox demo, the Hbox isn't getting the mutation callback, so we're using a timeout.
- [ ] Build the phosphor Widgets on the plotwidget/continuumview classes, instead of the box views.
- [ ] Default initial sizes - ideally take some params passed in from python - currently a few hard-coded minHeight/widths.
- [ ] Remove the hard-coded phosphor bundle location in the js template.
- [ ] Completely untested for other layouts - needs comprehensive validation.


dwillmer:
I've been testing with

```bash
python setup.py develop
BOKEH_RESOURCES=inline python examples/charts/file/boxplot_multi.py
```

We'll gradually start testing with the other examples as we validate/improve this approach.

bryevdv:
This is being superseded by the global solver work. But I will keep the branch so that Phosphor integration examples can be derived from it.

Issue #3635, opened 2016-01-12
Timeseries: Downsampling on server, chart pan and zoom slow even with LOD, and WebGL rendering problem


nothinkelse:
Hi. I have came to bokeh mainly for "high-performance interactivity over very large or streaming datasets" as stated in description. I started learning with bokeh just 2 days ago. Thus I started already with the currently newest version 0.11.0. 

I have a time series where events occur with frequency around 1 second. Thus for one week it is 7x24x3600=604 800 points. Thus downsampling data automatically on server and (in case of zoom) only sending the visible data to web browser is a very interesting feature. 

1) Is there any documentaiton on how to achieve downsampling of data on server with the new tornado bokeh server? (anything? I am willing to dig into it. I need to solve this issue for my project anyway)

2) Additional idea: would it be possible to write my own down-sampling function? E.g. on each zoom/pan my function with arguments such as current view window x,y,xwidth,ywidth would be called and my function would return a data to be plotted? Otherwise a situation might occur when the "default downsampling method" provided by bokeh might not be what people need.

I also like very much the LOD and WebGL functions in web browser. 

3) However even with LOD I still experience slow zooming. Is is possible to set how much the LOD should downsample the data? Or is it possible that LOD is not simplifying the data enough by default?
Here is an example:
      
      import numpy as np
      import matplotlib.pyplot as plt
      days=10
      pointsPerSecond=1
      numpoints=days*24*3600
      
      xx=np.arange(numpoints)
      randomPoints=np.random.rand(numpoints);
      randomPoints=randomPoints-0.5;
      timeseries=np.cumsum(randomPoints);
      print timeseries.nbytes/1024/1024, randomPoints.nbytes/1024/1024
      from bokeh.plotting import figure, show, output_notebook,output_server
      
      p = figure(title="Server Plot",webgl=False)
      p.line(x=xx,y=timeseries)
      output_server("test3")
      
      show(p)

![image](https://cloud.githubusercontent.com/assets/6547340/12264674/971a1308-b938-11e5-8f39-0531b30bedc3.png)

And it looks like this when zooming with the mouse scroll. Is it possible that even after LOD there are too many points? Zooming for this example is slow, choppy. (Running in chrome Version 47.0.2526.106 m, windows, Bokeh 0.11.0):

![image](https://cloud.githubusercontent.com/assets/6547340/12264689/bc387206-b938-11e5-8ac4-060b732e43b1.png)

4) When trying this with WebGL the rendering looks wrong:

[...]

PR #3888, opened 2016-02-17
A series of WebGL fixes (regression, precision, data update, miter limit)

### LINKED ISSUE #3464: maps_cities.py data disappears after pan/zoom
ping @almarklein 

Other WebGL examples work fine. 

Could consider a *very* short self-contained fix for `0.11`

```
examples/glyphs/maps_cities.py:
 data disappears after pan/zoom
```

### LINKED ISSUE #3795: datetime datapoints to pile up with webgl enabled
In Bokeh 0.11 there is some kind of truncation error in the rendering of a datetime axis when webgl=True. It causes the datapoints to pile up at discrete times.

Consider this minimal example:

```python
import numpy as np
import bokeh.plotting as bk
from datetime import datetime,timedelta

no_of_datapoints=10

base = datetime(2016, 2, 1, 9, 36, 0)
date_list = [base - timedelta(minutes=x) for x in range(0, no_of_datapoints)]
datapoints=np.arange(no_of_datapoints)

p=bk.figure(webgl=True,x_axis_type="datetime")
p.scatter(x=date_list,y=datapoints)
bk.show(p)
```

With webgl=True the diagram gets piled up datapoints:
qvm85ci

With webgl=False the diagram is rendered correctly 
wxt6mwa

See also this [stackoverflow discussion](http://stackoverflow.com/questions/35113105/visualizing-timeseries-with-bokeh-datapoints-overlap/35126714)

### LINKED ISSUE #3800: Bokeh 0.11: Enables webgl causes inconsistent update of scatter points
This problem might be related to this issue: https://github.com/bokeh/bokeh/issues/3795

Lets consider this minimal example:
```python
from bokeh.plotting import figure, show, ColumnDataSource
from bokeh.models import Select
from bokeh.models.widgets import RadioGroup
from bokeh.io import vform
import numpy as np

N = 100
x1 = np.random.normal(0.0, 1.0,N)
x2 = x1 * (1.0 + np.random.normal(0.0, 0.1,N))
y1 = 5.0*x1 + np.random.normal(0, 0.1, N)
y2 = 50.0*x2 + np.random.normal(0, 1, N)

source = ColumnDataSource({'x' : x1, 'y' : y1})

TOOLS = "box_select,box_zoom,lasso_select,help,reset,pan"
p = figure(tools=TOOLS, plot_width=400, plot_height=400, webgl=True)

x_radio_group = RadioGroup(labels=["x1", "x2"], active=0)
y_radio_group = RadioGroup(labels=["y1", "y2"], active=0)
controls = vform(x_radio_group, y_radio_group)

def update(attr, old, new):
    source.data["y"] = y1 if y_radio_group.active==0 else y2
    source.data["x"] = x1 if x_radio_group.active==0 else x2
    
    # Updating the source in one step does not change this behavior
    #data = dict(x=x1 if x_radio_group.active==0 else x2, y=y1 if y_radio_group.active==0 else y2)
    #source.data = data

source.on_change('selected', update)

def select_x(value):
    update('x-change', None, None)

def select_y(value):
    update('y-change', None, None)

x_radio_group.on_click(select_x)
y_radio_group.on_click(select_y)
```

You can reproduce this problem in the following way (using the bokeh server):

- Fire up the bokeh server (e.g. bokeh server file --show)
- Select a subset of the data points (using box select or lasso)
- Switch between y1 and y2 radio buttons
- Notice that even though the y-axis scaled as expected, the scatter points did not!

Now if you repeat this exercise with webgl=False, you should see that both the y-axis and the scatter points are updated as expected.

almarklein:
issues:
* fixes #3795
* fixes #3800 
* fixes part of #3635 
* might have fixed #3464

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/bokehgl.coffee
+++ b/bokehjs/src/coffee/models/glyphs/bokehgl.coffee
@@ -1,3 +1,4 @@
+_ = require "underscore"
 gloo2 = require "gloo2"
 color = require "../../core/util/color"
 color2rgba = color.color2rgba
@@ -36,14 +37,22 @@ fill_array_with_vec = (n, m, val) ->
         a[i*m+j] = val[j]
     return a
 
+visual_prop_is_singular = (visual, propname) ->
+    # This touches the internals of the visual, so we limit use in this function
+    # See renderer.coffee:cache_select() for similar code
+    return not _.isUndefined(visual[propname].spec.value)
+
 attach_float = (prog, vbo, att_name, n, visual, name) ->
     # Attach a float attribute to the program. Use singleton value if we can,
     # otherwise use VBO to apply array.
-    vbo.used = true
-    if visual[name].spec.value?
-      prog.set_attribute(att_name, 'float', null, visual[name].spec.value)
+    if not visual.doit
       vbo.used = false
+      prog.set_attribute(att_name, 'float', null, 0)
+    else if visual_prop_is_singular(visual, name)
+      vbo.used = false
+      prog.set_attribute(att_name, 'float', null, visual[name].value())
     else
+      vbo.used = true
       a = new Float32Array(visual.cache[name + '_array'])
       vbo.set_size(n*4)
       vbo.set_data(0, a)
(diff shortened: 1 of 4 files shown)


bryevdv:
@almar is this still WIP or is it ready?

almarklein:
@bryevdv still wip. I am going through the webgl issues ... more of this tomorrow

bryevdv:
NP just checking :)

almarklein:
@bryevdv its ready now

bokehjs/src/coffee/models/glyphs/bokehgl.coffee
```diff
@@ -48,30 +48,34 @@ attach_float = (prog, vbo, att_name, n, visual, name) ->
...
     # then use this single color for all vertices (no VBO). Otherwise we
     # create an array and upload that to the VBO, which we attahce to the prog.
     m = 4
     vbo.used = true
-    if visual.color.spec.value? and visual.alpha.spec.value?
-      rgba = color2rgba(visual.color.spec.value, visual.alpha.spec.value)
+    colorname = prefix + '_color'
+    alphaname = prefix + '_alpha'
+    if visual[colorname].spec.value? and visual[alphaname].spec.value?
+      rgba = color2rgba(visual[colorname].spec.value, visual[alphaname].spec.value)
```
bryevdv: could be e.g. `visual[colorname].value()` I think?

bokehjs/src/coffee/models/glyphs/bokehgl.coffee
```diff
@@ -48,30 +48,34 @@ attach_float = (prog, vbo, att_name, n, visual, name) ->
...
+        prog.set_attribute(att_name, 'vec4', null, [0,0,0,0])
+        return
       # Get array of alphas
-      if visual.alpha.spec.value?
-        alphas = fill_array_with_float(n, visual.alpha.spec.value)
+      if visual[alphaname].spec.value?
+        alphas = fill_array_with_float(n, visual[alphaname].spec.value)
       else
-        alphas = visual.cache.alpha_array
+        alphas = visual.cache[alphaname+'_array']
```
bryevdv: same comment as above, also `visual[alphaname].array()` when appropriate

bryevdv:
two small comments otherwise LGTM

bokehjs/src/coffee/models/glyphs/bokehgl.coffee
almarklein: mmm, I get:
```js
>> visual.line_color.value()
X  Error: attempted to retrieve property value for property without value specification
// and
>> visual.line_color.array()
X  TypeError: source is undefined
// but ...
>> line.cache.line_color_array
   Array [ "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", "#0f0", 140 meer… ]
```
Am I doing something wrong?

bokehjs/src/coffee/models/glyphs/line.coffee
```diff
@@ -51,7 +51,7 @@ class LineView extends Glyph.View
       x: this.renderer.plot_view.canvas.vx_to_sx(geometry.vx)
       y: this.renderer.plot_view.canvas.vy_to_sy(geometry.vy)
     shortest = 9999
-    threshold = Math.max(2, @visuals.line.width.value() / 2)
+    threshold = Math.max(2, @visuals.line.line_width.value() / 2)
```
almarklein: @bryevdv I came across this, and assumed it slipped through during earlier refactoring (because everything is `line_xxx` now). Can you confirm this is supposed to be `line_width`?

bokehjs/src/coffee/models/glyphs/bokehgl.coffee
bryevdv: The first one means the that the `spec` is configured for a data source column, or possibly is `null`. Are you relying on the color being `null` or `undefined` to not draw? If so I'd recommend testing the `doit` attribute to turn things on or off like the other renderers do. 

For the second I guess your usage actually differs a bit from the other renderers, `array` actually takes the data source to read from as an argument. Maybe this will have to do for now.

almarklein:
@bryevdv ok, I changed how the visuals is used. Should be better now.

Also found that dashes are ugly due to the recent miter fixes. Looking into it now.

almarklein:
Ok, fixed. Travis is also green, now that the flexx conda package is in the bokeh channel.

almarklein:
Euhm. That last commit applies to another PR. Sorry about that. But this is ready, anyway.

bokehjs/src/coffee/models/glyphs/line.coffee
bryevdv: Yes it's a little verbose but easier to construct generically.

Issue #3889, opened 2016-02-18
Speed up examples tests by using pytest-xdist


birdsarah:
pytest-xdist allows us to run tests in parallel, we should be able to use it to speed up the examples tests.

Issue #3826, opened 2016-02-04
Visual Diff Tests


hsparra:
- For a subset of integration tests have
  - Authoritative images
  - Do visual diff testing with the images

PR #3896, opened 2016-02-20
Speed up examples tests

birdsarah:
Fixes: #3889 

Convert examples tests to parametrized py.tests:
 - this means that running each example is recorded as one test
 - crucially means that we can use xdist to parallelize the test run


birdsarah:
Parallelizing with xdist speeds examples tests up to ~9.5 minutes.

Still todo:
 - [ ] notebook tests and timeouts - why?
 - [ ] upload images & pdiff results to s3

birdsarah:
This now moves the examples tests to pytest. 

This:
- allows tests to be parallelized (which speeds them up) (partial: #3889)
- partially fixes the examples tests messing up the local jupyter environment (partial: #3903)
- writes the report locally, so that you can see an example report locally too
- also breaks out the pdiff work so can be re-used by the other selenium tests (partial: #3826)
- runs and uploads on both py2.7 and py3.4

Things to note:
- reference images are downloaded for local tests so that dev can see local examples diff report
- diff image filename contains the version the image was diffed with (this seems like it'll be useful)

I haven't moved over the following code https://github.com/bokeh/bokeh/blob/master/tests/travis/test#L173-L176
```python
if args.upload and __version__.endswith("-dirty"):
    write("Can't upload reference images when working directory is dirty.")
    write("Make sure that __version__ doesn't contain -dirty suffix.")
    sys.exit(1)
```
is it needed

Questions: 
- does/should py2 vs py3 make a difference (now we're running both)? 
- should we delete / clean-up old images - there are a lot now!
- Is this the best organization for the ref images - I'm confused by the get_version_by_git and s3 setup. I can imagine it's extra confusing for people who don't have access to inspect the s3 bucket. I'm wondering what the pros and cons of checking the images in is.

birdsarah:
documenting failures:
twice on py34 (when py27 ran):
- compat/pandas_dataframe.py
- compat/ggplot_density.py
lots on py2 & lots on py3:
- https://travis-ci.org/bokeh/bokeh/jobs/110712882
- https://travis-ci.org/bokeh/bokeh/jobs/110712886

birdsarah:
Going with 3 cores seems to stop failures - it may be that it was just too much for the travis machines.

birdsarah:
Worth noting that this builds on top of #3869 - would be good to merge that first - will also make the diff more manageable.

birdsarah:
marking this as wip again as the code organization is a bit messy, so want to clean-up, to make bits re-usable.

bryevdv:
* does/should py2 vs py3 make a difference (now we're running both)?

Are you asking if we should run both, or if not, which one?

* should we delete / clean-up old images - there are a lot now!

I think so

* Is this the best organization for the ref images - I'm confused by the get_version_by_git and s3 setup. I can imagine it's extra confusing for people who don't have access to inspect the s3 bucket. I'm wondering what the pros and cons of checking the images in is.

I think for the (eventual) purpose written integration test plots, that there is just one fixed set of canonical reference images. I'm not sure about any of the other specifics about how things are organized so maybe concrete proposals would help clarify.

birdsarah:
Have also got the notebook tests running again. For example: https://s3.amazonaws.com/bokeh-travis/111138234/examples.html#howto/charts/histograms

birdsarah:
>> does/should py2 vs py3 make a difference (now we're running both)?

> Are you asking if we should run both, or if not, which one?

Sorry, that was a very poorly worded question! This PR switches to running the pdiff for both py2 and py3 because the old constraints of no py3 compatibility are gone. I am wondering if there should be any difference in processing output for the two; I would suggest no.

---

>>    Is this the best organization for the ref images - I'm confused by the get_version_by_git and s3 setup. I can imagine it's extra confusing for people who don't have access to inspect the s3 bucket. I'm wondering what the pros and cons of checking the images in is.

> I think for the (eventual) purpose written integration test plots, that there is just one fixed set of canonical reference images. I'm not sure about any of the other specifics about how things are organized so maybe concrete proposals would help clarify. 

Purpose written integration tests with canonical ref images makes sense. As does leaving the examples as they are - we could tweak the storage system of get_version_by_git but there's probably not much value in it, so leaving for now.

birdsarah:
py2 tests seem to complain about signal handler when using xdist

py3 tests, it's ggplot_density and pandas, both in compat, that fail when trying to run to fast.

birdsarah:
I've removed the `all_notebooks` command line option, I don't think we're using it - apologies if I'm off on this one.

birdsarah:
I think this is ready (still waiting on two test runs). 

What this PR accomplishes in the end:
* converts examples tests to pytest with parametrization, meaning that each example is now a test that passes or fails
* some cosmetic improvements to the examples report
* breaks out re-usable bits of code that can be shared with other tests:
 * image diff testing (#3826)
 * phantomjs screenshot
 * s3 upload 
 * pretty writing functions
 * running a jupyter notebook
 * running a bokeh server
* uses pytest-xdist to speed up py3 examples tests to ~13 minutes
* does not use pytest-xdist to speed up py2 examples tests as was consistently failing, complaining about `ValueError: signal only works in main thread` - i suspect this may be fixable
* both py2 and py3 examples tests are now image-diffed and uploaded to s3
* fixes examples notebook tests - so they actually run again
* partially fixes the examples tests messing up the local jupyter environment (partial: #3903)
* can write the examples report locally, so that you can see an example report locally too

This wasn't quite as awesome on the test speed-up as I'd hoped, but it's still an improvement, but had lots of other wins, especially progressing the image diff testing

birdsarah:
`ValueError: signal only works in main thread` also happens in py3, when `n=4` - may be masking a root cause.

birdsarah:
We will need to pick a trade-off between the speed-up we get with xdist and any additional instability.

We could also use py.test's `-x` flag which just bails from testing as soon as a single test fails. This would let us get to fails faster, freeing up resources, although we'd still need to re-start the tests. We could start with n=2 on py2 and n=3 on py3 and a -x flag for both and see how we feel about what's annoying us.

tests/plugins/utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def info(msg=None):
+    msg = " " + msg if msg is not None else ""
+    write("%s%s" % ("[INFO]", msg))
+
+
+def ok(msg=None):
+    msg = " " + msg if msg is not None else ""
+    write("%s%s" % (green("[OK]"), msg))
```
bryevdv: Maybe these could go in `bokeh.util.terminal` (doesn't have to be now) it might be nice to add colorized output to server output, e.g.  `setup.py` also defines these but I think they have to stay directly inside `setup.py`

bryevdv:
It looks like that our own "timeout" code uses `signal`. Maybe need to use `pytest-timout` with "thread" timeout method?

https://bitbucket.org/pytest-dev/pytest-timeout

> This is the surest and most portable method. It is also the default on systems not supporting the signal method.

tests/examples/test_examples.py
```diff
@@ -0,0 +1,231 @@
...
+    env['BOKEH_BROWSER'] = 'none'
+
+    class Timeout(Exception):
+        pass
+
+    def alarm_handler(sig, frame):
+        raise Timeout
+
+    signal.signal(signal.SIGALRM, alarm_handler)
+    signal.alarm(10)
```
bryevdv: we are using `signal` here

tests/examples/test_examples.py
```diff
@@ -0,0 +1,231 @@
...
+            return proc.wait()
+        except KeyboardInterrupt:
+            proc.kill()
+            raise
+    except Timeout:
+        warn("Timeout - Example timed out when attempting to run")
+        proc.kill()
+        return 0
+    finally:
+        signal.alarm(0)
```
bryevdv: and here

birdsarah:
@bryevdv, nice find on pytest timeout - thanks!

birdsarah:
> we could start with n=2 on py2 and n=3 on py3 and a -x flag

Having implemented this, I am happy with it already. When the py3 examples tests failed in a test run, they did so in 6 minutes. Whilst we'd rather they didn't fail, at least they weren't sucking up resources. 

With a very small sample size n=2 and n=3 seems to be working will, both test runs completing in under 15 minutes.

I am having a quick try with replacing our signal handler with pytest-timeout, but then this should be good to merge.

Although it looks big, it may be worth remembering this is mostly just an elaborate cut and paste from what was `tests/travis` to the new `tests/examples`. I believe I have included all functionality with the following exceptions that I'm aware of:

1) I have removed the all_notebooks command line option - it seemed like extra code that no-one was using
2) I haven't implemented the "don't upload if dirty" warning from: https://github.com/bokeh/bokeh/pull/3896/files#diff-ee664e71c8b0b1c6eff386de2a06dae4L173

Any other omissions might be unintentional.

birdsarah:
A quick play with pytest-timeout wasn't successful, I'd like to propose merging as is, we can still work on it.

bryevdv:
it looks like it didn't recognize `--max-fail` does something else need to be installed to enable `pytest-timeout` is it a separate package?

birdsarah:
> it looks like it didn't recognize --max-fail does something else need to be installed to enable pytest-timeout is it a separate package

doh! only do one thing at once!

birdsarah:
thanks @bryevdv - i got confused by the reporting on travis (the error message is before the red fail notification!)

birdsarah:
In development of this PR, I had limited the amount of data in charts/files/timeseries.py - see https://github.com/bokeh/bokeh/commit/9b5f7751b65fec86709ee100bff29b554f33822c

This seemed to make it less prone to failure - I think it just takes longer to render so is more likely to fail. 

I undid this change, but just had the test failure again: https://travis-ci.org/bokeh/bokeh/jobs/111360578#L1907

Should I keep the data limiting?

Other than this small detail, which could also be addressed later, @bryevdv this is ready.

birdsarah:
Sorry, also worth noting that I have not ported over tests/travis/nbexecuter.py https://github.com/bokeh/bokeh/pull/3896/files#diff-92a7119ce5bcc2bdfe080a2fe0ca6e4f as the notebooks were being run on the jupyter server.

bryevdv:
How/where do you see a single report with all the image diffs? I looked at https://travis-ci.org/bokeh/bokeh/jobs/111383890 but did not see a link to one

bryevdv:
Apart from that question this LGTM

bryevdv:
OK well, now some conflicts to resolve first.

birdsarah:
> How/where do you see a single report with all the image diffs? I looked at https://travis-ci.org/bokeh/bokeh/jobs/111383890 but did not see a link to one

https://s3.amazonaws.com/bokeh-travis/111383890/examples.html

You could find it by searching for "examples.html" or you could just know that it's `https://s3.amazonaws.com/bokeh-travis/<jobid>/examples.html` the job id is the last number in the link you posted.

I moved where the report uploads to because:
a) it makes it easier to find them you don't have to search the logs, you can just use the above formula
b) the previous system wouldn't work any more because the py2 report would override the py3 report as they were trying to upload to the same location - using the job-id disambiguates.

bryevdv:
Can we make the link more prominent, and also preceded by something that can be searched for uniquely? In the old output, I could search for "report" and it went directly to the URL to cut and paste.

birdsarah:
> Can we make the link more prominent, and also preceded by something that can be searched for uniquely?

Yes.

You can uniquely search for `examples.html`.

birdsarah:
Just noticed another small problem with this PR - it doesn't gracefully handle external contributors - like the saucelabs issue - they can't upload the examples report to S3 - this shouldn't fail the test.

bryevdv:
Something easier type (no punctuation for instance) I do this all the time with the old format  and it was already annoyingly too much work to get to the report.

birdsarah:
Now would be a good time to tell me exactly what you want. I don't think I find myself as frustrated by this as you, so not sure what would be easy. I was just about to set a prefix "EXAMPLES REPORT SUCCESSFULLY UPLOADED"

I haven't even been doing the search thing, I've just been copying and pasting the jobid from the url.

birdsarah:
This should be ready again, I'm going to run a test from an external contributor to make sure the upload is behaving as expected.

birdsarah:
See: https://travis-ci.org/bokeh/bokeh/jobs/111886435#L2119

For how uploads are now handled on an external contributor PR. It's not very subtle, but is clear!

tests/plugins/utils.py
fpliger: +1 but think it's better on a different PR..

fpliger:
Really nice PR. LGTM

bryevdv:
Great, merging.

PR #3901, opened 2016-02-20
fix hist bin autosize with unique values

### LINKED ISSUE #3659: Histogram bin size auto select
Dear authors,

In the special case where there is only one unique value in the histogram input data, the automatic bin size function of 'Histogram' create a  **`bins` should be a positive integer error**.

Here a case to reproduce the error:

```
from bokeh.charts import Histogram, output_file, show, vplot, hplot, Bar, defaults

arr = [5,5,5,5,5]

output_file("Hist.html")
p1 = Histogram(arr, title="Histogram", xlabel="Distances", ylabel="Counts")
p1.x_range = Range1d(-10, 10) 
         
p = vplot(hplot(p1))
show(p)
```

Whereas when you change the array to following array it works perfectly:
```
arr = [5,5,5,5,5,6]
```

Please fix the autoselection process for this special case.

Thank you.

fpliger:
issues: fixes #3659

### DIFF
--- a/bokeh/charts/builders/histogram_builder.py
+++ b/bokeh/charts/builders/histogram_builder.py
@@ -152,7 +152,7 @@ class HistogramBuilder(BarBuilder):
         y_max = max([comp_glyph.y_max for comp_glyph in self.comp_glyphs])
         y_min = min([comp_glyph.y_min for comp_glyph in self.comp_glyphs])
 
-        x_buffer = ((x_max + x_min)/2.0)*0.1
+        x_buffer = ((x_max + x_min)/2.0)*0.05
 
         self.x_range = Range1d(start=x_min - x_buffer, end=x_max + x_buffer)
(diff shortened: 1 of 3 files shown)


bryevdv:
This works (so merging not) but the bar is awfully narrow on the default range (and looks just like a vertical like with the explicit `(-10, 10)` range in the original issue).  Not sure what to do about it though. If there are ideas to improve this please make another issue to discuss. 

screen shot 2016-02-21 at 4 37 50 pm

fpliger:
@bryevdv yeah, I've noticed... (actually I've also made a change to reduce the default range otherwise it was looking even worse) but making it look better would mean tweaking with the bin width and I wanted to check we are not doing anything formally wrong. Still a bit stretched on time and this looks like a very specific case so I'd think it should go into its own issue.

Issue #2622, opened 2015-07-27
List of color tuples broken?


jmhessel:
I think that bokeh doesn't properly support lists of colors at the moment.

````                                         
from bokeh.plotting import figure, show, output_file

output_file("broken.html")
xs = [1,2,3]
ys = [1,2,3]
col = [(255, 0, 0), (0, 255, 0), (0, 0, 255)]

p = figure(title = "Broken")
p.circle(xs, ys, radius=.25, color=col)
show(p)
````

When I run this code, I get 3 circles with no color. The documentation seems to suggest that colors can be passed in this way.

bokeh.__version__ is 0.9.2

**Comment by bryevdv** [comment]
Just to be clear, the problem is only with lists of color **tuples**. There are many examples with lists of color strings, e.g. `"blue"` or `"#ff0000"`.

**Comment by canavandl** [comment]
When you set the color property as a RGB-style tuple, Bokeh (correctly) sets the fill/line_color attrs of the dataspec map as ColorSpec instances that get serialized to CSS-style RGB strings in JSON, like:
```javascript
...
 {'attributes': {'doc': None,
   'fill_alpha': {'value': 1.0},
   'fill_color': {'value': 'rgb(5, 3, 77)'},
   'id': '2e4667b9-f6ad-4c7d-baa6-f20e054bc92f',
   'line_alpha': {'value': 1.0},
   'line_color': {'value': 'rgb(5, 3, 77)'},
   'radius': {'units': 'data', 'value': 0.25},
   'tags': [],
   'x': {'field': 'x'},
   'y': {'field': 'y'}},
  'id': '2e4667b9-f6ad-4c7d-baa6-f20e054bc92f',
  'type': 'Circle'},
...
```
However, when setting a glyph property as a sequence (i.e. x, y, or color), the RGB-style list gets set as a named column in the plot's ColumnDataSource and the fill/line_color attrs get set as the keys. This is a problem because the 3-element lists don't correspond to CSS colors when the plots are rendered on the client.

```javascript
...
{'attributes': {'callback': None,
   'column_names': ['y', 'x', 'line_color', 'fill_color'],
   'data': {'fill_color': [[255, 0, 0], [0, 255, 0], [0, 0, 255]],
    'line_color': [[255, 0, 0], [0, 255, 0], [0, 0, 255]],
    'x': [1, 2, 3],
    'y': [1, 2, 3]},
   'doc': None,
[...]

PR #3904, opened 2016-02-21
Canavandl/2622 color tuple seq

canavandl:
issues: fixes: #2622

Bokeh doesn't support lists of color tuples as color arguments. As discussed in #2622, there's not an obvious path forward to resolve this, as it's unclear to how distinguish between a 3-element field that's intended to be a color tuple or a multi-list.

This PR adds an explicit warning to the 'styling visual attributes' docs sections about this edge case and suggests a couple of work-arounds.


birdsarah:
LGTM - built locally, and looks fine.

kkumawat25:
How to convert a list of tuples from pallete to RGB

bryevdv:
@kkumawat25 GitHub is not the appropriate venue for support/usage questions. Please use the project Discourse:

https://discourse.bokeh.org/

PR #3909, opened 2016-02-22
Take shallow copy of DataFrame in `ChartDataSource`

### LINKED ISSUE #3899: TimeSeries modifies a pd.DataFrame inplace
From the ML - https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/yF77ulMnE6Q
```python
df1 = pd.DataFrame(data=np.random.rand(5,3), columns = ('a', 'b' ,'c'), index = pd.DatetimeIndex(start='01-01-2015',periods=5, freq='d'))   
df2 = df1.copy()
assert df2.equals(df1)
TimeSeries(df2)
assert df2.equals(df1)
```
results in:
```python-traceback
AssertionError                            Traceback (most recent call last)
<ipython-input-32-8de7d5ef3815> in <module>()
      3 assert df2.equals(df1)
      4 TimeSeries(df2)
----> 5 assert df2.equals(df1)

AssertionError: 
```

The reason being that the call to TimeSeries has appended an index column:
```
In [36]: df2.columns 
Out [36]:  Index(['a', 'b', 'c', 'index'], dtype='object')
```

dhirschfeld:
Avoids modifying the input DataFrame inplace
Fixes #3899

### DIFF
--- a/bokeh/charts/data_source.py
+++ b/bokeh/charts/data_source.py
@@ -151,7 +151,7 @@ class DataGroup(object):
             data (:class:`pandas.DataFrame`): the subset of data associated with the group
             attr_specs dict(str, :class:`AttrSpec`): mapping between attribute name and
             the associated :class:`AttrSpec`.
-            
+
         """
         self.label = label
         self.data = data.reset_index()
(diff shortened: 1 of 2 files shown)


canavandl:
The test would be appreciated. I don't think you need to call ds.setup_derived_columns() as the method is called in the ChartDataSource init. I would leave the comment explaining the reason for the test though, so:

### ChartDataSource initializer shouldn't modify input DataFrame
assert  pd_data.equals(test_data.pd_data)

dhirschfeld:
I *think* that should be good to go pending CI checks...

canavandl:
LGTM. Thanks @dhirschfeld for contributing!

Issue #3622, opened 2016-01-11
Wrong BOKEH_RESOURCES argument in Developing Examples section for server demos


canavandl:
From http://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html#bokeh-dev:

"""
For Bokeh server examples, add BOKEH_DEV=true to the server invocation:

``` bash
BOKEH_DEV=true bokeh serve example-server.py
```

"""

Needs to set BOKEH_RESOURCES=server-dev (not absolute-dev) because the served plots aren't able to resolve the full paths of the js/css resources and need them as a URL.

rothnic:
I'd like to add after working with server and bokehjs, it is kind of difficult at the moment. I'm going to list my understanding here that I had to find out by poking around, so feel free to correct me as needed. This information could be useful in the developers guide.

* Source maps don't work with server, get warnings about the maps referencing files not available, which may be associated with the following issues
* When using BOKEH_RESOURCES=server-dev, this tells server to use the bokeh/bokeh/server/static folder, which is only updated from setup.py (which is called by gulp install), which seems to defeat the purpose of the development mode
* --debugjs isn't currently accepted, which was previously used to turn on some of this functionality
* if server has to use the static folder, there is no gulp task that assists with building any changes and updating that folder while doing development with server and bokehjs
* the coffeescript source maps are not being generated ( i think this would appear as comments within the compiled js files)

This may need to go in another issue, but I see this as the development documentation for server is no longer valid.

bryevdv:
previous text is now gone, nothing to update

PR #3910, opened 2016-02-22
Add correct server dev instruction to docs

canavandl:
fixes: issues #3622 

Docfix to note the correct BOKEH_RESOURCES value (server-dev not absolute-dev, as is set by BOKEH_DEV=true) to load local JS/CSS resources.

### DIFF
--- a/sphinx/source/docs/dev_guide/setup.rst
+++ b/sphinx/source/docs/dev_guide/setup.rst
@@ -285,8 +285,8 @@ Bokeh to use relative paths to development resources, issue:
 
     BOKEH_RESOURCES=relative-dev python example.py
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Hmm. I'm not sure. I never have to set anything at all, because I always use `python setup.py develop` and the server always uses and servers the BokehJS that is "installed". Maybe that is simpler advice, and we can just get rid of `server-dev` entirely

canavandl:
Let me check it again. I had issues with that a month ago (when I opened the issue) - but let me see if that's still true.

canavandl:
Yes - I think you're right. I'm going to closed this PR as rejected, continue the conversation in #3622, then likely open a different docfix PR eventually.

PR #3913, opened 2016-02-22
Some canvas and layout improvements

bryevdv:
This PR does some preliminary work towards the general layout system. I'd like to merge this and begin a new PR for integrating @havocp 's algorithm. This PR:

* adds some debugging instrumentation to track the number of added  constraints
* moves the constraint solver to the document, instead of having one per plot
* cleans up `canvas.coffee`, moves some util functions there to `core/utils`
* stops canvas from excessivee and unnecessary re-rendering


bryevdv:
Hrmm, when I run the examples locally, they work...

bryevdv:
Oh, yikes...  PhantomJS < 2.0 does not support Function.prototype.bind.apply` apparently: https://github.com/ariya/phantomjs/issues/10522

@mattpap @birdsarah, thoughts? Here is where I was using it, maybe there is an alternative implementation? 

https://github.com/bokeh/bokeh/pull/3913/files#diff-cea1759b174a6af0c5a026df7ee93ad2R12

bryevdv:
This appears to be a polyfill that might work: https://github.com/kdimatteo/bind-polyfill/blob/master/index.js

Not sure where exactly that would go, though.

bryevdv:
@bokeh/dev ready for review

PR #3918, opened 2016-02-23
Spike of switching out selenium screenshot for phantomjs for examples tests

birdsarah:


birdsarah:
I don't think we should do this for now, am leaving as paused but would like to keep this branch for a while.

PR #3920, opened 2016-02-23
Fix typo in docs: toolbar_position -> toolbar_location

ukrutt:
The docs say:

> Try running the code and changing the `toolbar_position` value.

(http://bokeh.pydata.org/en/0.11.1/docs/user_guide/tools.html#positioning-the-toolbar)

This is incorrect, the parameter name is `toolbar_location`.


bryevdv:
So, it is! Thanks for the catch, and the PR!

ukrutt:
Happy to help :-)

PR #3926, opened 2016-02-24
Disable saucelabs tests for external contributors

birdsarah:
The saucelabs tests require access to secure environment variables, external
contributors don't have access to them. For an explanation see:
https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests

This disables the integration tests for external contributor PRs. There are
ways to finess this later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,13 @@ before_install:
     # test run, but we can manually use travis' saucelabs start/stop commands
     - if [[ -z "${SAUCE_USERNAME}" ]] || [[ -z "${SAUCE_ACCESS_KEY}" ]]; then echo "External committer, not using saucelabs"; else export SAUCELABS=True; fi
     - echo $SAUCELABS
-    - if [[ -z "${SAUCELABS}" ]]; then "source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh)"; else "export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start"; fi
+    - if [[ -z "${SAUCELABS}" ]]; 
+      then
+        source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh); 
+      else 
+        export DISPLAY=:99.0;
+        sh -e /etc/init.d/xvfb start; 
+      fi
 
 install:
     - scripts/travis_install
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ before_install:
     # test run, but we can manually use travis' saucelabs start/stop commands
     - if [[ -z "${SAUCE_USERNAME}" ]] || [[ -z "${SAUCE_ACCESS_KEY}" ]]; then export SAUCELABS=False; else export SAUCELABS=True; fi
     - echo $SAUCELABS
-    - if [[ -z "$SAUCELABS" == "True" ]]; then "source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh)"; fi
+    - if [[ -z "${SAUCELABS}" == "True" ]]; then "source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh)"; fi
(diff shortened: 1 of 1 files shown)


birdsarah:
this did not have the desired outcome. i thought it would only turn off integration tests for external contributors, but is turning off for all. closing this for now and will do something better with fixtures. I have opened an issue: #3927

birdsarah:
ping @damianavila @fpliger @brendancol, I am struggling with bash syntax here. I'm sure you can see what I'm trying failing to accomplish. Any thoughts gratefully received.

.travis.yml
```diff
@@ -54,7 +54,10 @@ before_install:
     - export TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
     # We don't use the travis saucelabs add on because it installs on every
     # test run, but we can manually use travis' saucelabs start/stop commands
-    - source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh)
+    - if [[ -z "${SAUCE_USERNAME}" ]] && [[ -z "${SAUCE_ACCESS_KEY}" ]]; then export SAUCELABS=True; else export SAUCELABS=False; fi
+    - echo $SAUCELABS
+    - if [ -z $SAUCELABS = True ]; then source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh); fi
```
damianavila: I guess this should be `==`

.travis.yml
```diff
@@ -54,7 +54,10 @@ before_install:
...
+    - if [ -z $SAUCELABS = False ]; then export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start; fi
```
damianavila: same here...

.travis.yml
mattpap: This conditional statement doesn't look like bash to me. Not enough quotes at least.

birdsarah:
@mattpap @damianavila - thank you so much for chipping in.

Here are all the variations of that conditional statement that I've tried in a series of commits:

```sh
if [ -z $SAUCELABS == "True" ];
if [[ -z "$SAUCELABS" == "True" ]];
if [[ -z "$SAUCELABS" = "True" ]];
if [[ -z "${SAUCELABS}" = True ]];
if [[ -z "${SAUCELABS}" == True ]]; 
if [[ -z "${SAUCELABS}" == "True" ]];
if [ -z $SAUCELABS = True ];
```

There are more possible combinations of `"` `[` vs `[[` and `=` vs `==` but may be too many to try manually.

Any thoughts on the right one?

birdsarah:
@mattpap @damianavila - I finally got it - thanks for the nudge. 

I did not understand the significance & meaning of `-z`. Which, for my future reference: http://www.tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html
```
[ -z STRING ]	True of the length if "STRING" is zero.
[ -n STRING ] or [ STRING ]	True if the length of "STRING" is non-zero.
```
= and == are roughly equivalent apparently, but behave differently depending on whether they're in [] or [[]] (http://tldp.org/LDP/abs/html/comparison-ops.html)

```
The == comparison operator behaves differently within a double-brackets test than within single brackets.

[[ $a == z* ]]   # True if $a starts with an "z" (pattern matching).
[[ $a == "z*" ]] # True if $a is equal to z* (literal matching).

[ $a == z* ]     # File globbing and word splitting take place.
[ "$a" == "z*" ] # True if $a is equal to z* (literal matching).
```

and for the $thing vs ${thing} - http://www.tldp.org/LDP/abs/html/parameter-substitution.html#PARAMSUBREF

```
${parameter}

    Same as $parameter, i.e., value of the variable parameter. In certain contexts, only the less ambiguous ${parameter} form works.
```

ready to go cry now.

thanks for the nudges, it spurred me on.

I am going to close this PR and open a new clean one!

mattpap:
@birdsarah, bash is really messed up. In general you should always use quotes to make sure globs at al. don't kick in. So you will use `if [[ "$X" == "X" ]]`. `-z` is helpful and could be translated to `not` in Python (but not a logical one, just `__nonzero__`/`__bool__` in `not ""` or `not []`).

> There are more possible combinations of (...)

There is also `if test ...`. Just for fun. I think it's just alternative to `[[ ]]` (but an external program, or maybe `test` allows arithmetic as well).

> and for the $thing vs ${thing}

`${thing}` can be also used (with some extra syntax) as a replacement for sed, e.g. `${original_string//word/$new_word}"` will replace all `word` with `new_word`.

birdsarah:
forgot to say thanks for this @mattpap

PR #3938, opened 2016-02-25
Changed .bk-plot to .bk-root for style wrapper

mistakevin:
This is to help the namespace clash with the other `.bk-plot` class in use.

Follow-up to https://github.com/bokeh/bokeh/pull/3797

### DIFF
--- a/bokeh/core/_templates/plot_div.html
+++ b/bokeh/core/_templates/plot_div.html
@@ -6,4 +6,4 @@ Renders a basic plot div, that can be used in conjunction with PLOT_JS.
 :type elementid: str
 
 #}
-<div class="plotdiv bk" id="{{ elementid }}"></div>
+<div class="plotdiv bokeh" id="{{ elementid }}"></div>
--- a/bokeh/core/_templates/plot_div.html
+++ b/bokeh/core/_templates/plot_div.html
@@ -6,4 +6,4 @@ Renders a basic plot div, that can be used in conjunction with PLOT_JS.
 :type elementid: str
 
 #}
-<div class="plotdiv bokeh" id="{{ elementid }}"></div>
+<div class="plotdiv bk-root" id="{{ elementid }}"></div>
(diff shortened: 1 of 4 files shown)


mattpap:
I think `bk` may be too short for good cooperation with web sites. For example, code highlighters use very short labels for colorful syntactic components, so this could be easily confused.

mistakevin:
Good point. maybe `.bk-style`?

canavandl:
i'm +1 on bk-style

bryevdv:
Then why not just .bokeh (shorter, no dash)

mistakevin:
Changed to `.bokeh`

mistakevin:
@birdsarah Do you know why these tests would be failing?

mattpap:
> Then why not just .bokeh (shorter, no dash)

Because it's better to be consistent and have all CSS classes prefixed with `bk-`, even if this doesn't look as appealing as pure `bokeh`.

mattpap:
Also I don't want to do `git grep bokeh` to learn what code depends on this CSS class (e.g. investigating selenium failures).

mattpap:
`bk-bokeh`, `bk-root`, `bk-toplevel` (maybe others) would work for me.

bryevdv:
Sorry I misunderstood, I thought it was a proposal to use `bokeh-` everywhere instead of `bk-`. Trying to do too many things... if it's just this spot `bk-root` seem ok.

bryevdv:
Ithought this was failing because it was on a fork, I was mistaken, there is a real failure

bryevdv:
@mistakevin I am not sure there is any way to "re-open" this it may have to be a new PR

mistakevin:
You should be able to revert it from your end. It'll create a new PR and you can merge that.

bryevdv:
i did already revert it, I meant a new PR to finish/continue whatever is needed to make it actually work and pass the tests.

mistakevin:
Oh yeah. I'll get one in

bryevdv:
Thanks!

PR #3945, opened 2016-02-26
Demo more prominant report location

birdsarah:
Try and make it easy to find the report location on a travis log


birdsarah:
ping @bryevdv, what do you think about adding report upload locations to end of poor man logger, as shown in this PR. That way you can just go to the end of the log, and it's printed there - no more searching needed?

bryevdv:
Sounds good

PR #3948, opened 2016-02-26
concepts file changes [ci disable examples]

electronwill:
[ci disable examples]

- add 'screen units' and 'data-space units' to glossary
- 'to server' -> 'to serve'
- small rewrite about high level and low level interfaces

### DIFF
--- a/sphinx/source/docs/user_guide/concepts.rst
+++ b/sphinx/source/docs/user_guide/concepts.rst
@@ -32,6 +32,13 @@ Charts
     interface to quickly construct these kinds of plots. See
     :ref:`userguide_charts` for examples and usage.
 
+Data-space Units
+    Units used to specify glyph height or width relative to the data in a graph. 
+    Bokeh supports both data-space units and screen units. For example, in a 400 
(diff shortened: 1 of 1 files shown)


bryevdv:
I'm a little hesitant about putting this here. I was trying to keep this particular glossary very small, and restricted to very core *things*, that could be linked to linked to (which is the case for almost every entry, except "document" does not have a good landing yet). Basically so that it could act as descriptive table of contents. Units are important but they are definitely not central abstractions like everything else not the list, and there is nothing to link to from either. I'd prefer to consider other locations for this information.

bryevdv:
Basically everything else in the list is an extremely high level concept, and then we'd be adding two extremely in-the-weeds details.

electronwill:
That does make sense, and it's a good idea to keep the glossary consistent with regards to whether it handles details or broad concepts. I'll update this PR and https://github.com/bokeh/bokeh/pull/3950 to move the screen units and data-space units information to the Plotting page.

electronwill:
Ready for review.

PR #3950, opened 2016-02-26
fixes to 'plotting' [ci disable examples]

electronwill:
[ci disable examples]

- link 'screen units' and 'data-space units' to glossary
- make "reference documentation linked above" a link
- per Bryan in bokeh flowdock channel 2016-02-17, infinite pan/zoom
  is now default, although max and min range may still be set.

### DIFF
--- a/sphinx/source/docs/user_guide/plotting.rst
+++ b/sphinx/source/docs/user_guide/plotting.rst
@@ -49,9 +49,12 @@ example plots for all of them by clicking on entries in the list below:
     * |x|
 
 All the markers have the same set of properties: ``x``, ``y``, ``size`` (in
-:ref:`screen units <userguide_glossary>`), and ``angle`` (radians by default).
-Additionally, |circle| has a ``radius`` property that can be used to specify
-:ref:`data-space units <userguide_glossary>`.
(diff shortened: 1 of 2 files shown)


bryevdv:
This depends on on another PR, in general that should be noted, or the two PRs combined (probably would have been good in this case given how small the PRs are)

electronwill:
Ah, I see. In the future I'll just make a single PR in such a case.

electronwill:
Ready for review.

bryevdv:
Well so, being picky now, I do think they might merit a small section on their own, maybe "A note on Units" or something. That's because there are several things that can make use of either screen or data units. I guess the question is where... I'm inclined to say maybe the "Styling Visual Attributes" chapter is the right place. That's where we explain lots of other details like line joins and color conventions.

electronwill:
Ah, good-- I do like that better than just putting the info in these two places. I'll take a more detailed look soon and update the PR.

bryevdv:
ping @electronwill

electronwill:
Sorry for the lag-- I got pulled away onto another project this week. I think we have it finished now, so I'll get back on this tomorrow and will probably have it ready quite quickly.

electronwill:
OK, ready for review.

PR #3953, opened 2016-02-26
Other libraries in docs

### LINKED ISSUE #3951: Add links to more external libraries
Bryan suggested that I make a PR to add information about HoloViews's new Bokeh interface to the Bokeh website, so I've created this issue to reference from the PR.

jbednar:
issues: fixes #3951

Now that the HoloViews library has a Bokeh backend, it will be of interest to some Bokeh users, and so I'd like to add a link to it from within the Bokeh docs.  In this proposed PR, I've added that by generalizing the discussion of "Leveraging Other Libraries".  So that HoloViews isn't needed when building the Bokeh site, I included a static screenshot of using Bokeh with HoloViews.

I also added what is essentially a stub, mentioning that Bokeh works well with other JavaScript libraries.  Given the title of the page, it seems appropriate to include that information, but someone who knows more about interaction with such libraries might be able to expand this section to be more useful.


sphinx/source/docs/user_guide/compat.rst
```diff
@@ -25,11 +66,22 @@ the ``bokeh.mpl.to_bokeh`` function:
...
+significantly in the future, based on plans (`MEP 25`_) for Matplotlib
+to adopt a native JSON ingest/export functionality of its own. This
+will allow Bokeh and other systems to interact more robustly with
+Matplotlib, potentially in both directions (i.e., with Matplotlib able
+to display Bokeh plots).
 
+You can see examples of Bokeh rendering Matplotlib plots generated
+from each of these libraries in the `examples/compat`_ directory. This
+link points to master; don't forget to make sure you're looking at
+examples that match with the version of bokeh you're using.
```
bryevdv: There's a newish `bokeh-tree` sphinx extension that will generate correctly versioned links into the GH source tree"

https://github.com/bokeh/bokeh/blob/master/bokeh/sphinxext/bokeh_github.py#L111

As long as we are editing this, might as well fix this up and get rid of the message about master.

sphinx/source/docs/user_guide.rst
```diff
@@ -84,3 +85,4 @@ structure.
 .. _Matplotlib: http://matplotlib.org
 .. _Pandas: http://pandas.pydata.org
 .. _Seaborn: http://web.stanford.edu/~mwaskom/software/seaborn
+.. _HoloViews: http://holoviews.org
```
bryevdv: I do try to to keep these in alpha order.

bryevdv:
@jbednar two small comments. The test failure is because this is from a fork, we are still working out our sauce labs integration.

birdsarah:
@jbednar, the PR from a fork issue has been resolved in master, if you could merge and push, that would be great.

jbednar:
Ok, the two comments should be addressed.  Ready to merge?

bryevdv:
docs only, skipping tests.

jbednar:
The image link on http://bokeh.pydata.org/en/dev/docs/user_guide/compat.html doesn't work.  The file is in github (https://github.com/bokeh/bokeh/blob/master/sphinx/source/_images/hv_sample.png), so presumably it needs a different path so that it can be found?

bryevdv:
It should be copied during the deploy if it is referenced correctly. See 

https://raw.githubusercontent.com/bokeh/bokeh/master/sphinx/source/docs/user_guide/setup.rst

for an example (I think you have extra `..` that is the problem)

jbednar:
I think you're right; I've pushed that change to my fork, but would I now need to make a new PR just for those two characters?

bryevdv:
Yup

jbednar:
Sigh. :-)

PR #3955, opened 2016-02-27
Noarch conda package and entry points

### LINKED ISSUE #3571: missing bokeh entry point (otherwise bokeh.bat) for Windows

almarklein:
### Issues
* supersedes #3874 (and #3792)
* fixes #3571
* related to #1827
* related to conda/conda-build#796 and conda/conda-build#797

### Relevant parts of the code

* the conda.recipe dir
* scripts/build_upload.sh
* scripts/travis_install
* scripts/jenkins_install -> outdated, remove?
* binstar.yaml -> outdated, remove?

### Tasks
* [x] add entry point to setup.py
* [x] add entry point to conda recipe
* [x] make conda recipe a noarch package (get it with `conda install bokeh -c pyzo`)
* [x] test the noarch conda package (done on Linux and Windows and OS X)
* [x] update travis_install
* [x] update build_upload.sh
* [x] decide on specifying `futures` as a dependency or not
* [x] remove `jenkins_install` and `binstar.yaml`?
* [x] Some fixes in conda-build: https://github.com/conda/conda-build/pull/798 and https://github.com/conda/conda-build/pull/799


almarklein:
There is one caveat of the noarch approach for us. We use `- futures [py27]`, but this line gets evaluated at build-time, not install time (I checked). This means that we have to choose to either always install `futures` as a dependency, or have Python 2 user install it manually if they want to use the server.

bryevdv:
Is there a (stubbed out?) `futures` package available for py3 that is harmless to install?

bryevdv:
alternatively, we vendor the `futures` library. I'm not normally a huge fan but it might make sense in this specific case. It think the benefits of a no arch pkg would outweigh the downsides of vending it in.

almarklein:
> Is there a (stubbed out?) futures package available for py3 that is harmless to install?

Maybe just ... `futures`? is `root/lib` not always before `root/../site-packes`? Apparently futures *is* available as a conda package on Python 3.

bryevdv:
> remove jenkins_install and binstar.yaml?

`jenkins_install` can 100% be removed, and I believe `binstar.yml` too, as long as `build_and_upload.sh` doesn't need it (I don't think so)

scripts/build_upload.sh
```diff
@@ -96,23 +96,12 @@ echo "Removing first bokeh build at $first_build_loc"
...
 
 # upload conda pkgs to anaconda.org
-platforms=(osx-64 linux-64 win-64 linux-32 win-32)
-for plat in "${platforms[@]}"
-do
-    echo Uploading: $plat
-    anaconda -t $bintoken upload -u bokeh $plat/bokeh*.tar.bz2 -c $channel --force --no-progress
-done
+echo Uploading noarch pkg
+anaconda -t $bintoken upload -u bokeh noarch/bokeh*.tar.bz2 -c $channel --force --no-progress
```
almarklein: @damianavila I think this is correct, but maybe you want to test it?

almarklein:
This is about ready. I made `futures` a dependency, it does not seem to have a negative effect on py3k (except for the extra download).

conda.recipe/meta.yaml
```diff
@@ -15,7 +18,7 @@ requirements:
     - python
     - distribute
     - setuptools
-    - nodejs >=4.1
+    - nodejs >=4.1  [unix]
```
damianavila: there is a nodejs available for win-64 here: https://anaconda.org/javascript/nodejs/files... just to let you know...

conda.recipe/meta.yaml
```diff
@@ -54,15 +58,15 @@ test:
     - mock
     - pandas
     - pytest >=2.9.0
-    - pytest-cov ==1.8.1
-    - pytest-selenium >=1.2.1
+    - pytest-cov ==1.8.1       [unix]
```
damianavila: Are we using `unix` here because win packages are not available in our Bokeh space? If that's the case, we should made and upload the packages...

scripts/build_upload.sh
damianavila: It seems correct...

scripts/travis_install
```diff
@@ -25,7 +25,7 @@ conda install --yes $DEPS_TRAVIS
 
 conda config --add channels bokeh
 
-CONDA_PY="${TRAVIS_PYTHON_VERSION/./}" conda build --quiet --no-test conda.recipe
+CONDA_PY="${TRAVIS_PYTHON_VERSION/./}" conda build --skip-existing --quiet --no-test conda.recipe
```
damianavila: why the `--skip-existing`?

damianavila:
Left some tiny comment, still TravisCI is not passing...

conda.recipe/meta.yaml
almarklein: Yes, otherwise it wont build. The recipe does not actually run tests, so these dependencies are currently not necessary. So ok for now. If we do make the recipe run tests we should probably add a pytest-conv windows package.

scripts/travis_install
almarklein: This was to work around a bug in conda. But I fixed that, and I think the fix made it to the latest release. Will try removing it again.

almarklein:
Travis fails because there is now a readme in the examples (recently added?), and the noarch build does not know what to do :/  will look into this.

bryevdv:
All the apps examples got readmes

bryevdv:
BTW If conda build can't deal with "extra" READMEs I would call that a bug in conda.

bryevdv:
I've pinged the conda FD channel

kalefranz:
Found the problem with conda-build.  See https://github.com/conda/conda-build/pull/838

bryevdv:
@almarklein there should be a new conda-build pacakge in a couple of days

almarklein:
Ok, we should probably just restart Travis by then.

bryevdv:
@bokeh/dev new `conda-build` just released, restarting tests

bryevdv:
OK whatever that `futures` package is it does not seem to be compatible with python 3:
https://travis-ci.org/bokeh/bokeh/jobs/116336061#L1313-L1316

bryevdv:
OK two possible options:

* vendor in `futures` and somehow only use it for python 2.7

* make a new meta-package `futures-compat` (or whatever) that depends on `futures` for py2 and does not depend on `futures` for py2, then have the Bokeh no-arch build depend on this

ping @kalefranz @ilanschell

almarklein:
My vote's for  a `futures-compat` package. We only need to create it once, and should be maintenance free after that.

ilanschnell:
Another possibility would be (but this would require changes to `conda-build`) to create to create two bokeh noarch packages: `bokeh py2` and `bokeh py3`

The `python_noach` packages are actually more than just `noarch`, they also don't depend on a specific version of Python.  From a conda perspective, all that matters is that they are placed in the `noarch` directory (visible on all platforms).

I could very easily make two (`py2` and `py3`) `futures-compat` packages manually, and place them in http://repo.continuum.io/pkgs/free/noarch/.  This would be a good proof of concept, and we could then think about adding the ability to create noarch packages which still depend on a specific version of Python to `conda-build`.

bryevdv:
+1 on `futures-compat`

ilanschnell:
Ok, so I just added the "self made" `futures-compat` packages to the default (noarch) channel.  Works like a charm:
```
$ conda create -n foo2 python=2.7 futures-compat
... installs futures-3.0.3-py27_0 (and the empty futures-compat)
$ conda create -n foo2 python=3.5 futures-compat
... does NOT install futures (only the empty futures-compat)
```

bryevdv:
@almarklein @bokeh/dev OK tests passing! image diff report looks reasonable. I propose merging this tomorrow, getting a dev build out, and soliciting the ML for help testing installation. (But if possible I would like to get #4037 merged as well, before the dev build)

scripts/travis_install
bryevdv: pinging to check on this

almarklein:
Awesome!

bryevdv:
OK, merging now

damianavila:
We should have a devel build soon to test this out on production...

damianavila:
> We should have a devel build soon to test this out on production...

OK... @bryevdv I read your comment before commenting... :wink:

PR #3958, opened 2016-02-28
Delayed slider callback

### LINKED ISSUE #3957: Delayed Update Slider
Sometimes, an image render or other expensive task performed in a callback can take too long to complete as a user drags the slider to the desired value.  In these cases it would be preferable for the slider to not execute the callback till the user mouses-up...

justacec:
This PR implements just a few small changes to allow the end user to specify that the slider callback should only be executed when he or she mouses up.  This is needed for situations when the callback performs a time expensive task and executing the callback at every step would be counter-productive.  The change is implemented by adding a boolean property to the Slider class which is then interrogated in the SliverView _render function.  Essentially, I moved the callback from the "slide" property of the Slider to the "stop" property which gives the desired effect.

Example: 

    offset_slider = Slider(start=-5, end=5, value=0, step=.1,
                       title="Offset", callback=callback, delay_callback=True)

One issue with this is the number above the slider does not update while dragging.  It might be nice to allow that number to change while dragging, but to only execute the callback when the user mouses-up.

issues: fixes #3957


justacec:
Ohh jezz...  This is wrong...  Delete this pull request.  I will resubmit.  It was tainted with other changes from a different branch

justacec:
Ok, I might have fixed this.

justacec:
This branch actually has no conflicts, it is just a by-product of the broken way in which I generated the pull request.  How do I clean up that error and get the CI to validate the pull request?

mattpap:
@justacec, looking at individual commits, I think, you should remove the first and the last.

bokeh/models/widgets/inputs.py
```diff
@@ -152,6 +152,10 @@ class Slider(InputWidget):
     A callback to run in the browser whenever the current Slider value changes.
     """)
 
+    delay_callback = Bool(default = False, help="""
```
mattpap: Preferably `Bool(False, ...)` or `Bool(default=False, ...)` (no spaces around `=`).

bokehjs/src/coffee/models/widgets/slider.coffee
```diff
@@ -22,15 +22,26 @@ class SliderView extends ContinuumView
...
+    else
+      @$('.slider').slider({
+        orientation: @mget('orientation')
+        animate: "fast",
+        slide: _.throttle(@slide, 200),
+        value: @mget('value')
+        min: min,
+        max: max,
+        step: step,
+      })
```
mattpap: This code shouldn't be duplicated. Do as follows:
```
opts = {...}
if not @mget('delay_callback')
 opts.stop = @slide
else
 opts.slide _.throttle(@slide, 200)
@$('.slider').slider(opts)
```

bokeh/models/widgets/inputs.py
justacec: Corrected.   (Not sure why the spaces cause issues though.  I think they help for readability.)

bokehjs/src/coffee/models/widgets/slider.coffee
justacec: Corrected

justacec:
@mattpap I would love to remove those spurious commits but I am not sure what the clean way to do so is.  What is your recommended method?  I know that a lot of sites suggest rebasing, but I understand that has nasty repercussions which could affect this PR.

bokeh/models/widgets/inputs.py
mattpap: > Not sure why the spaces cause issues though. I think they help for readability.

This is the style we currently use and we would prefer the code base to stay consistent. If we decide to write spaces around `=`, we will change all instances and then ask submitters to stay consistent with the new style. In general you should follow the style and other rules of the surrounding code. We have fairly lax attitude here, but there are projects where every single mismatch will be required to be fixed.

justacec: No problem.  I can sympathize with the desire to maintain a consistent style. I will try to watch it.

mattpap:
> What is your recommended method? I know that a lot of sites suggest rebasing (...)

Our approach is to allow changing history (rebasing, ammending, cherry-picking) of unpublished work and of published work when asked to do so. The former case is rather obvious and unrelated here. It's bad to change published commits if there's a possibility someone created derivative work. So you never rebase `master` or equivalent. In this case it's unlikely, so you are free to fix the branch. You can still mess up by rebasing too much, but then we would ask you to rebase again. You should proceed with `git rebase -i 35a9c55^` (or equivalent) and follow instructions (pretty much remove lines with commits to be removed). If the commit range appears to be wrong, you can always cancel a rebase by removing all lines (that's explained as well in the instructions). An alternative is to close this PR and open a new one with a fixed branch, but that's just a lot more work to gain nothing (at least in this case).

justacec:
Got it.  I will do that tonight when I get home.

justacec:
@mattpap Ok, I think that I cleaned this up.  Learned a big lesson... Don't merge the master branch and then rebase, rebase and then merge master...

mattpap:
LGTM.

justacec:
Let's get this moving??

mattpap:
> Let's get this moving??

That was one LGTM, we need another one. Well, nothing official in this matter yet, but my preference is to have at least two people signing off before merging.

tobyhodges:
@justacec for clarification, my understanding is that LGTM = looks good to me

bryevdv:
I'd suggest a slight change, if we are going to have a throttle rate, let's just make that the configurable thing. I.e., instead of `delay_callback = Bool` have `callback_throttle = Float` is the throttle rate in `ms` and ` value of `None` means no throttling.

justacec:
@bryevdv I am not sure that I agree (arm could be twisted).  As an end user, it is not obvious to me that specifying a throttle time of `None` would mean that it would effectively delay the callback execution until I raised the mouse button.  I would attach the meaning of `None` to a delay of zero.  That being said, I think offering a re-configurable delay time would be a good addition.  I tend to favor giving end users more freedoms.

bryevdv:
Sorry I meant None means there is no throttling, that is, reproducing the existing behavior where there are updates continuously.

justacec:
@bryevdv Ok, I see.  Your original suggestion was that we remove the delay_callback parameter and mode it all in the setting of the callback_throttle.  Is this suggestion still on the table or are we to adding an additional parameter to support the adjustment of the delay in addition to the bool option to delay the callback.

As an alternative, one could set the default value of callback_throttle to Infinity (and possibly beyond) which would mean that the callback would not be executed until the mouse up.  That might be an intuitive way to address this capability, add another capability, and keep it inside of a single variable.  The question is whether Bokeh knows how to serialize infinity.

bryevdv:
@justacec yes I was proposing a single property, but I did not realize there were three cases. So perhaps this:
````
callback_policy = SliderCallbackPolicy  # enum with "continuous", "throttle", "mouseup"

callback_throttle = Float # used when policy is "throttle"

```

justacec:
Ahh...  I like it.  If @mattpap is good with this method I can make the necessary changes tonight.  Till then I suggest this be reflagged as WIP.

justacec:
I have added the code necessary to implement @bryevdv 's suggestion.  However, durring testing of that code, I found a bug in a recent commit.  It is being tracked via issue #3989.  Till that is repaired, this is on hold (albeit finished).

bryevdv:
@justacec agreed let's get that other PR finished and merged first. I don't see the new code, does it still need to be pushed?

justacec:
@bryevdv yea.  I have not pushed it yet because I wanted to test it out before I submitted it.  I could go ahead and push it up though. Hopefully, the other PR gets merged in soon.  I would like to get this one finished as soon as possible (because this one is simple and I like seeing todo items get checked off  :)  )

bryevdv:
CSS PR merged, give it a try now.

justacec:
I merged and then compile and tried the color_slider.py example.  Still no sliders.  I will have to play with this a bit more to figure out what is going on.  Maybe some cacheing?

justacec:
Ok, I think this is in pretty good shape.  There are some small tweaks that I am going to do.  However, the throttling does not seem to work.  No matter what I set the timeout time is, the execution never slows down.  I think that either I am simply doing it wrong, or something is broken elsewhere.  Or I just miss-understand the throttle concept.  Run the new example that I added and watch the log as you drag the second slider.  If I understand things correctly, if I drag that slider to the left and the right rapidly, the callback should not execute more than once in a 2 second period (because I set the callback_throttle to 2000).  I think that it is being more responsive than that.

justacec:
Additionally, I think that one possibility could be to move the policy logic to the slide function which would allow for the label to be updated with the current slider position (for precise fine tuning before you lift the mouse).  The only thing would be how to indicate to the end user that the callback has not fired yet and the number is just an indicator of where the slider is.  Possibly a different color (lower saturation?) while the callback had not been issued?  

Thoughts @bryevdv

justacec:
@bryevdv @mattpap I think this is ready to go.  I have added all the functionality that I can think would be relevant right now.  Everything seems to work well and responds the way it was intended.  The throttle function never worked in the slider before because the function was being regenerated (and therefore it always thought that it had not previously run) at each render call.  To fix this I have moved the throttle call to the init routine so it is defined only once and then repeatedly executed later.

I have also added some additional functionality to allow the input field which shows the slider position to always update as the user drags the slider and the be in a subdued color while dragging.  This is useful for the case of the mouseup slider.  Previously, the @slide function was not being executed until the mouse was released.  This meant that the input feedback field was not being updated till then and the user could not know exactly which value would be selected when the mouse was released.  

The only other thing that I could think of for this would be to have the color change to be subdued when the current value of the slider is not the most recent value which was send to the callback.  This would only be the case for the `mouseup` and the `throttle` modes.  I would say that can be done in a later PR.  I think the sliders have received enough attention for the time being.

Checkout the `slider_callback_policy.py` example.

Thoughts?

justacec:
@bryevdv @mattpap Just a reminder that this PR is ready for review

bryevdv:
@justacec I have a presentation to prepare and give later today but I will try to look at it after that.

bokeh/models/widgets/inputs.py
```diff
@@ -152,6 +152,20 @@ class Slider(InputWidget):
     A callback to run in the browser whenever the current Slider value changes.
     """)
 
+    callback_throttle = Float(default=200, help="""
+    Number of microseconds to pause between callback calls as the slider is moved.
+    """)
+
+    callback_policy = Enum("continious", "throttle", "mouseup", default="throttle", help="""
```
bryevdv: spelling (important for an enum), should be "continuous"

bryevdv:
@justacec I checked out branch and the operation looked great. There's only a couple of tiny thing. First there is a misspelling in the enum value for "continuous". Also if we want to make things as consistent a possible, the best thing to do would be to define  an actual documentable `Enum` type as is done here:

https://github.com/bokeh/bokeh/blob/master/bokeh/core/enums.py

Last thing is the para, it shows up like this for me: 
screen shot 2016-03-08 at 2 12 27 pm

Do you actually see the HTML formatting?

justacec:
Oh yea, I also did not see the formatting.  I have no idea on why it was
not formatting.  The misspelling: good catch.  I will update that quickly.
I will also change the enum over.  Then things should be good to go.

Concerning the html.  Let me know why the HTML was not rendering if you
can.  I was at a loss.

mattpap:
`Paragraph` widget doesn't allow HTML.

justacec:
What method should I use to inject HTML such that it is rendered?

justacec:
Ok, everything fixed with the exception of the HTML in the paragraph...  That is still a todo?  Suggestions?  I did not see any other widget that might fill that need.  I will keep looking to see how I might inject formatted HTML into the output.

justacec:
Ok, I have simply removed the HTML tags.  I think it would look better with them, but there does not seem to be an easy way to have Bokeh programmatically output a block of HTML.  Maybe a future feature (div widget).  The PreText widget looks ok, but it has the scroll bars always visible, even when not needed.  It seems as though this was done on purpose.  If not, I suggest that the css be altered to be auto to get rid of the bars if they are not needed.

I believe this to be completed now.  ;)

@bryevdv Ready for re-review

bryevdv:
restarted spurious test failure

PR #3959, opened 2016-02-29
Computed transforms

### LINKED ISSUE #3825: Computed Transforms

justacec:
This pull request implements client side computed transforms.  In general, a new family of classes based off the `Transform` class are established.  Examples of classes in the family are:

 1. *Jitter*: A class which adds a random uniform fluctuation to the input data values.  The width of the variation is specified in the `interval` property of the class
 2. *LinearInterp*: Linearly interpolate between a set of provided points.
 3. *StepInterp*: Linearly interpolate between a set of provided points and then floor the result to the input control points of the transformation.

These transforms are intended for use in the backend models and not necessarily for the end user.

#### TODO

- [x] Implement the `Jitter` class
- [x] Implement the `LinearInterp` class
- [x] Implement the `StepInterp` class
- [x] Change the name of `LinearInterp` to `LinearInterolation` 
- [x] Change the name `StepInterp` to `StepInterpolation`
- [x] Adjust the `Jitter` class to support different distributions (Gaussian and Uniform)
- [x] Compose a series of unit tests to evaluate the deterministic transforms
- [x] Add new examples to highlight the use of the transforms
- [x] Remove the `LinearColorMapperTransform` temporary mapper
- [x] Investigate whether we need to change the Image glyph back to having a `LinearColorMapper` class for its color mapper or if we should leave it as `ColorMapper`
- [x] ~~Get some official documentation put in (Is this really needed?)~~
- [x] Transition the source sorting code from LinearInterpolation to the base Interpolation class for re-use
- [x] Extend the sorting function to support passing the `x` and `y` values as arrays for the interpolation routines
- [x] Remove the Log interpolation routine shells
- [x] Add the different modes for the StepInterpolator
- [x] Update the documentation in the class definitions
- [x] Add a good Jitter example
- [x] Add a good Interpolation example
- [x] Fix the Jitter JS unit tests
- [x] Remove the palette initializer from the interpolator base class definition.

issues: fixes #3825


justacec:
@bryevdv I have some questions about your interpolation transformations that you would like to have implemented.  For the LinearInterpolation, I assume that we would need to pass a series of points, not just x values as control points.  The interpolation would then "transform" input x' values into y' values via a simple linear mapping inside the domain of the x control points.  If this is the correct interpretation, I think that the class would need to have a parameter called `control_points` which would be a `ColumnDataSource` with two columns, one for the x and one for the y.  Then the input `x` values to the `compute` method in the JS would linearly interpolate based on those control points.  Alternatively, the user could specify the x and the y values of the control points in separate parameters.  What is your preference?

the real question comes in with the `LogarithmicInterpolation`.  Can you give me a good mathematical example for this.  Again, I assume that we would need to have a set of control points with x and y values.  These values can help define a transform, but I am not sure how.  

If I can understand this a bit more, I think I can code this up pretty quickly now that I have the framework in place.

justacec:
I have uploaded an example output (from the examples directory, specifically the examples/plotting/file/transforms.py file) for this early framework using the Jitter to my DropBox.  The URL is:

[https://www.dropbox.com/s/kku5qvmkwgsdgio/JitterExample.html?dl=0](https://www.dropbox.com/s/kku5qvmkwgsdgio/JitterExample.html?dl=0)

The screenshot of the example is the following:

screen shot 2016-02-28 at 9 54 53 pm

bryevdv:
@justacec this is great! I'm in the middle of traveling home tonight I'll respond to your questions first thing tomorrow

fpliger:
@justacec here are a few thoughts/suggestions shared with @bryevdv .

With all the items listed this PR can get really big and potentially harder to manage. So here are a list of proposals to try to help:

* let's to try to keep this PR focused on the limited effort to add the minimal `tranforms` needed and then extend it with more `transforms` in following additional PRs. 
* better to keep the `ColorMappers` work out of this PR (basically on a quick follow up PR since it's mostly written already) so it's easy to understand/discuss the interface/how to add it, etc...
* let's focus on a limited set of `transforms` to be part of a minimal set. The suggestion here is `Jitter`, `LinearInterpolation`, `StepInterpolation`
* a very important thing here is agreeing on the right interface for it and you raise a good point about having `x` and `y` points to interpolate. At the same time there *are* use cases when the `x` component can be omitted and we "consider" those values being as regular integer values in the interval starting from 0, that covers the length specified by `y`. That said, as interface for Linear/Step/Log Interpolators I'd propose the following:

```Python

class Interpolator(Transform):
    y = Either(String, Seq(Float)) # not nullable :)
    x = Either(String, Seq(Float))
    data_source = Instance(ColumnDataSource)

# possible uses:
interp = Interpolator(y=[1, 2, 3])
interp = Interpolator(x=[-10, 2, 13], y=[10, 30, 40])

cds = ColumnDataSource(data=....)
interp = Interpolator(y='column_y', data_source=cds)
interp = Interpolator(y='column_y', x='column_x', data_source=cds)
interp = Interpolator(y='column_y', x=[-10, 20, 33, ...], data_source=cds)
```

This last use case adds some more validation checks... In general there are a few things to be validated with this interface.

Since this is a really important addition that could be used in a lot of places this interface should be well discussed to try and get it right in the first attempt. ping @bokeh/dev

justacec:
I completely agree on all points.  I made a slight modification of the
ColorMapper just to use as a development tool.  It is the intent to back
this change out for the final merge and keep this completely separate from
the ColorMapper's.  I will likely work on the linear interpolator tonight.

birdsarah:
@justacec, there's been some testing improvements added to master, can you merge master into this pr?

justacec:
Sure thing.  I will do that tonight.

examples/plotting/file/transforms.py
```diff
@@ -0,0 +1,33 @@
...
+image = p.image(image=[d], x=0, y=0, dw=10, dh=10, color_mapper = cm)
+
+callback = CustomJS(args=dict(image = image, jitter = jitter), code="""
+        jitter.set('interval', cb_obj.get('value'))
+        image.attributes.data_source.forceTrigger()
+    """)
+
+slider = Slider(start=0, end=4, value=jitter.interval, step=.1, title="Jitter Width", callback=callback)
+
+output_file("image.html", title="image.py example")
```
birdsarah: This is what's causing the examples tests to fail, the html must be the same as the file name. In this case
```python
output_file("transform.html", title="image.py example")
```
also image.py example is also not the best title for this example

justacec: Ohh, yea.  I will fix that up.  I just copied over the image example to start twiddling with.

bryevdv:
BTW @justacec would you be interested in being added to @bokeh/dev? That would allow you to create branches for PRs off the main repo. That plays a little better with our CI testing machinery, and  can also sometimes make collaboration easier if its helpful for multiple people to work on a branch together.

justacec:
Sure.  I would not have issue with that.

bryevdv:
@justacec do you still have the ability to work on this? If not, could you re-submit this on a PR on a branch on the main repo so that someone else can pick up the work?

justacec:
@bryevdv Thanks for the poke.  I have gone through and added a lot of this in.  The only thing that I lack right now (with respect to the initial concept) is the log interpolation.  I would like to put this in but I need a little more background on the "Log" interpolation and what that means.  I have done some reviewing online and have found some difficulty in locating a consistent definition of logarithmic interpolation.  Can you describe it in words?

If I can get my head around what is intended for this transform, I should be able to code it up rather quickly.

Concerning recasting this PR off the main branch.  I would be more than happy to, but I am not sure I have the rights to do that.  Also if one was to do that would all of the revision history be lost into a single squashed commit?  Additionally, I have never done that, so you would need to walk me though the steps (would be nice to learn how to do this).

fpliger:
@justacec @bryevdv my understanding from conversation is that we would move ahead with this PR without `Log` interpolation and add it on a following focused PR. Is this correct? If so... shall we mark this PR as `ready`?

justacec:
@fpliger You are correct that @bryevdv and I spoke and have decided to not include the log transforms.  The only thing that I have left for this is to include unit tests in the JS code side of the house and possibly include an example script highlighting the effect.  There is not much checking to be done on the Python side since the classes do very little work.  That being said, we could go ahead and mark this as ready and do the unit tests in a separate PR (can be done against the main repo so others can work on it at the same time, but I do not that I have the perms to create a branch off the main repo yet).  When it comes to the example, I am not sure that is necessary since the end users will use these transforms very little.

bryevdv:
@justacec can you merge master in again? I think some recent CI testing changes will have the test reports behaving better.

justacec:
@bryevdv will do

bokeh/models/transforms.py
```diff
@@ -0,0 +1,107 @@
...
+    Dependant coordinate denoting the value of a point at a location.
+    """)
+
+    data = Instance(ColumnDataSource, help="""
+    Data which defines the source for the named columns if a string is passed to either the ``x`` or ``y`` parameters.
+    """)
+
+    # Define an initialization routine to do some cross checking of input values
+    def __init__(self, palette=None, **kwargs):
+        super(Interpolator, self).__init__(**kwargs)
```
bryevdv: I don't think this is needed?

justacec: Ahh, now I see where you were looking.  I agree.  That must have been a hold over for something else.  Will remove this evening when I get home from work.

bokehjs/src/coffee/models/transforms/interpolator.coffee
```diff
@@ -0,0 +1,85 @@
...
+class Interpolator extends Transform.Model
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
+    @_x_sorted = []
+    @_y_sorted = []
+    @_sorted_dirty = true
+
+    @bind 'change', () ->
+      console.log 'Changed the configuration'
```
bryevdv: should use the `logger` facility from `{logger} = require "../../core/logging"`

bokehjs/src/coffee/models/transforms/interpolator.coffee
```diff
@@ -0,0 +1,85 @@
...
+      @_sorted_dirty = true
+
+  props: ->
+    return _.extend {}, super(), {
+      x: [ p.Any, '']
+      y: [ p.Any, '']
+      data: [ p.Any, null]
+    }
```
bryevdv: FYI in master there is now a cleaner/simpler way to do this, see e.g.:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/ranges/data_range1d.coffee#L11-L20

Also I think the defaults for `x` and `y` should be `null` and, simply omitting a default will make `null` the default
Lastly (picky!) every other place in the lib we do line the prop defs up for readibility.

bokehjs/src/coffee/models/transforms/interpolator.coffee
```diff
@@ -0,0 +1,85 @@
...
+    tsx = []
+    tsy = []
+
+    # Populate the tsx and tsy variables correctly depending on the method by which the user populated the interpolation
+    # data.
+    if typeof(@get('x')) == 'string'
+      data = @get('data')
+
+      tsx = data.get_column(@get('x'))
+      tsy = data.get_column(@get('y'))
```
bryevdv: need logging/error raised if column is missing

bokehjs/src/coffee/models/transforms/jitter.coffee
```diff
@@ -0,0 +1,49 @@
...
+      r1 = Math.random()
+      r2 = Math.random()
+      r2 = (2*r2-1)*Math.sqrt(2*(1/Math.E))
+      break if -4*r1*r1*Math.log(r1) >= r2*r2
+    rn = r2/r1
+
+    # Transform the standard normal to meet the characteristics that we want (mu, sigma)
+    rn = mu + sigma*rn
+
+    return rn
```
bryevdv: This function could go here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/util/math.coffee

bokehjs/src/coffee/models/transforms/jitter.coffee
```diff
@@ -0,0 +1,49 @@
...
+
+class Jitter extends Transform.Model
+  initialize: (attrs, options) ->
+    super(attrs, options)
+
+  props: ->
+    return _.extend {}, super(), {
+      mean:         [ p.Number, 0        ]
+      width:        [ p.Number, 1        ]
+      distribution: [ p.String, 'uniform']
```
bryevdv: is also possible to define enums (which will validate) on JS side:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/enums.coffee
https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/properties.coffee#L153

bokehjs/src/coffee/models/transforms/transform.coffee
```diff
@@ -0,0 +1,13 @@
+_ = require "underscore"
+Model = require "../../model"
+
+class Transform extends Model
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
```
bryevdv: not needed if it just calls `super`

bokehjs/src/coffee/models/transforms/step_interpolator.coffee
```diff
@@ -0,0 +1,54 @@
+_ = require "underscore"
+Transform = require "./transform"
+Interpolator = require "./interpolator"
+p = require "../../core/properties"
+
+
+class StepInterpolator extends Interpolator.Model
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
```
bryevdv: not needed if it just calls `super`

bokehjs/src/coffee/models/transforms/linear_interpolator.coffee
```diff
@@ -0,0 +1,37 @@
+_ = require "underscore"
+Transform = require "./transform"
+Interpolator = require "./interpolator"
+
+class LinearInterpolator extends Interpolator.Model
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
```
bryevdv: not needed

bryevdv:
@justacec I left a few more comments, they are mostly very small. This is a large PR, so iteration is hopefully expected! But I think it is very close, and will be excited to get this merged in. I do suggest merging master in, @mattpap recently added a simpler/better way of specifying properties and all models should use that going forward.

bokehjs/src/coffee/models/transforms/step_interpolator.coffee
```diff
@@ -0,0 +1,54 @@
...
+
+  props: ->
+    return _.extend {}, super(), {
+      mode: [ p.String, "after"]
```
bryevdv: same comment about JS side enums

bokehjs/src/coffee/models/transforms/interpolator.coffee
```diff
@@ -0,0 +1,85 @@
...
+    else
+      tsx = @get('x')
+      tsy = @get('y')
```
bryevdv: I'm not sure if we need to require both `x` and `y` to be in the data source or not. One of coud be a literal sequence, and the other in the source? If we don't allow that, we just need to detect and warn explicitly about that case

bryevdv: Ah, I see you are checking that the types match above.

bryevdv: I think some of this logic could be simplified a bit with a slight refactoring along these lines:

```
# _get_values has the logic to check for literal sequence, or column in source
tsx = @_get_values(@get('x'))
tsy = @_get_values(@get('y'))

if tsx.length != tsy.length
  # raise error

if tsx.length == 0 
  return
```

justacec:
@bryevdv Thanks for the feedback.  I will take a look at all of these comments tonight.  I will give myself a suspense of Sunday COB to have them corrected or have follow-up discussions.

bryevdv:
@justacec great, and thanks for all your hard work and patience this is going to be a great feature that will enable even more great features!

justacec:
Well, we all use and benefit from great free software like this.  Sometimes
it just makes sense to just give back sometimes.
 #LunchesShouldntAlwaysBeFree

bryevdv:
@justacec FYI another option just FYI if you are busy, is to close this PR, and re-open as a branch on the main repo (you should have an invitation to @bokeh/dev that would allow this). Then someone else could work on these last few remaining tasks.

bokeh/models/transforms.py
justacec: Ok, I have removed the palette=None initializer from this.

bokehjs/src/coffee/models/transforms/interpolator.coffee
justacec: I will remove this as it was not meant to stay and was just some debugging stuff for me.

bokehjs/src/coffee/models/transforms/interpolator.coffee
justacec: Ok, I have adjusted this and have double checked the other transforms in the JS transforms dir to convert those as well.

bokehjs/src/coffee/models/transforms/interpolator.coffee
justacec: Added.

bokehjs/src/coffee/models/transforms/jitter.coffee
justacec: I shifted this over and imported the math.coffee code in as math.  Not sure if that is a good idea since it is so close to "Math" which is the core JS math routines.  I think I am going to rename that bokeh_math.

bokehjs/src/coffee/models/transforms/jitter.coffee
justacec: Corrected.

As a side note.  I think it would have been nice if I could have specified this as:

    [p.Enum(enums.DistributionType), 'uniform']

It seems a bit dangerous for people to need to edit the important properties.coffee file to add an enum type.  I suppose I could have just put:

    [p.enum_prop('Distribution', enums.DistributionType), 'uniform']

I think the former is cleaner looking, but it still requires the user to pull in two files instead of just the properties...  Who knows which is better...

bokehjs/src/coffee/models/transforms/step_interpolator.coffee
justacec: Corrected

bokehjs/src/coffee/models/transforms/linear_interpolator.coffee
justacec: Corrected

bokehjs/src/coffee/models/transforms/step_interpolator.coffee
justacec: Corrected

bokehjs/src/coffee/models/transforms/interpolator.coffee
justacec: @bryevdv Corrected, I think.

Concerning the question about allow for a string in one parameter and then an array in the other...  I had thought about that as well but was concerned that it would increase the possibility of a length mismatch between the x and the y.  This could also occur if the user changed the source or the array.  At each change, the dirty flag would be raised and the system would then throw an error because one or the other would not be the same length as the other (possibly).  This is because you can not specify all of these in a single call.

However, if the user specifies two arrays, there is no way to change the length of the arrays without throwing an error (if everything works the way I expected it too).

This might mean that the correct thing to do would be to simply have the interpolators return a NA style value in the event that the defining datasets for the interpolation are in a failed state.  

Thoughts?

justacec:
@bryevdv I need to do some more checking on these changes that I have just gone though and made.  I will need to do that in a bit though.  I just wanted you to know I currently do not feel confident that these changes are compete and rather represent a WIP to address your comments.

fpliger:
@justacec what are the main things that are missing? I don't want to push you in any way if you don't have the time atm but rather check if there is something we can do to help if someone else have some time...

justacec:
I honk that I am basically there.  I pushed several commits this morning which I hope addressed all of Brendan's comments.  I noticed that the TravisCI is broken right now so I can't check the status of the unit tests right now.  I just need to verify everything works as intended.  Should be done very quickly now.

Justace

bryevdv:
I restarted Travis, hopefully just an intermittent problem.

bokehjs/src/coffee/models/transforms/step_interpolator.coffee
```diff
@@ -0,0 +1,50 @@
...
+      diffs = (Math.abs(tx - x) for tx in @_x_sorted)
+      mdiff = _.min(diffs)
+      ind = _.findIndex(diffs, (num) ->
+        return mdiff == num
+      )
+
+    if ind != -1
+      ret = @_y_sorted[ind]
+    else
+      ret = NULL
```
mattpap: `null`?

justacec: WIP :)

justacec:
@bryevdv Getting there.  Just have to fix this annoying "add" is not recognized as a function in the JS unit test for Jitter.  I am not sure why it is not working.

justacec:
@bryevdv @mattpap Ok, I believe that I have addressed the issues that have been outstanding.  The only thing that I could think of here would be the fact that the examples which I put together do not render well.  I am most likely mis-using the layout tools.  I would go in and fix them, but I know that layout is in flux right now.  So, two things which could be updated for completeness:

 - Fix the example layouts
 - Add documentation

What are your thoughts on these?  (different PR?)

bryevdv:
Improving any layout can definitely wait for another PR, as for docs I leave that decision to you, if you'd rather make a separate PR that is fine with me. I will test out this branch this afternoon, please just mark it `status: ready`when you consider it ready for final review / merge.

Thanks!

justacec:
@bryevdv Lets leave the docs out for now and make that a separate PR.  Concerning the layout, I figured out that issue this morning and submitted a change already.

bryevdv:
Fantastic, I will take it for a spin in a few hours!

justacec:
@bryevdv Just take a look at the examples that I placed in the examples/models directory.  The interpolator example has one limitation that you need to manually change an option to get the curve to appear.

bryevdv:
Hi @justacec I just took a look at those examples. I must be honest and say they are a little more complicated than our typical examples of this type. Normally in these exampels we try to very clearly demonstrate one "thing" or technique, and I wonder if all the the widgets and interactions (which are great!) might be more of a distraction in this case. So I have a few suggestions:

Jitter Example
---------
I think this would be improved by getting rid of all the widgets. first of all there is a layout issue if the browser is too narrow:

screen shot 2016-04-15 at 2 19 35 pm

But mostly I think it would be better if there were just three fixed columns: one with no jitter, one with uniform jitter, and one with normal jitter. You could use the new `Label` annotation to label all three columns of points.

Additionally I think it looks quite a bit more distinct with `size=5` and `N=1000` 

Interpolators Example
-------
This example confused me at first and for quite a while. With only two control points to start by default, it is very unclear what is going on. Also, step mode has no effect when the interpolation  is linear (naturally) but I think this might confuse people into thinking it is "broken" since the control is still active. I also think it might also be more clear if users can *compare* by seeing all the modes at once. So here is my suggestion:

Have only one widget, a dropdown with the following options:
* Linear
* Step (before)
* Step (center)
* Step (after)
Have a Div that instructs users to select one to see something. Have a fixed set of control points (but more of them, maybe 8-10 or more)

----

I think all the user interactions are cool but I think the examples should focus on the single important topic at hand without distraction. We could add more sophisticated demos like this in other places in the docs. What do you think?

bokehjs/src/coffee/models/transforms/interpolator.coffee
bryevdv: Well one reason I suggested is that there is no enforcement on the python side. It is easily possible (as things are now) to give one sequence, and one string column name. But in that case no error would happen  until things hit the browser. I think it should either work to have the combination, or if that's to be not allowed, we need to add a model validation so that the user is warned immediately from python that there is a problem.

bryevdv:
Apart from the examples comments, I have one comment about the sequence/column name question, and everything else looks fantastic.

justacec:
@bryevdv All points taken.  I can see where the uses could get confused as you suggested when interacting with the examples.  I will add in all of your suggestions.  Most likely not tonight, but I can do it quickly.  I am not sure I understand what you mean by a comment on the sequence/column name question.  Can you expand on that comment?

bryevdv:
@justacec yes sorry I meant this: https://github.com/bokeh/bokeh/pull/3959/files#r59930186

I don't mind pitching in BTW :)  I will try to post some updated versions of the examples for you to consider tonight if I can.

bokeh/models/transforms.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class Jitter(Transform):
+    ''' Apply either a uniform or normally sampled random jitter to data.
+
+    '''
+
+
+    mean = Float(default=0, help="""
+    The central value for the random sample
```
bryevdv: I don't think it makes sense for this to be exposed? I've never seen any use of jitter that was not symmetric about the data points.

bryevdv:
@justacec here is an updated jitter example for you to consider:
```
import numpy as np

from bokeh.io import vplot
from bokeh.models import CustomJS, Button
from bokeh.models.transforms import Jitter

N = 1000

    x=np.ones(N), xn=2*np.ones(N), xu=3*np.ones(N), y=np.random.random(N)*10
))

normal = Jitter(width=0.2, distribution="normal")
uniform = Jitter(width=0.2, distribution="uniform")

p = figure(x_range=(0, 4), y_range=(0,10))
p.circle(x='x',  y='y', color='firebrick', source=source, size=5, alpha=0.5)
p.circle(x='xn', y='y', color='olive',     source=source, size=5, alpha=0.5)
p.circle(x='xu', y='y', color='navy',      source=source, size=5, alpha=0.5)

callback=CustomJS(args=dict(source=source, normal=normal, uniform=uniform), code="""
    data=source.get('data')
    for (i=0; i < data['y'].length; i++) {
        data['xn'][i] = normal.compute(data['x'][i]+1)
    }
    for (i=0; i < data['y'].length; i++) {
        data['xu'][i] = uniform.compute(data['x'][i]+2)
    }
    source.trigger('change')
""")

button = Button(label='Press to apply Jitter!', callback=callback)

output_file("transform_jitter.html", title="Example Jitter Transform")

show(vplot(button, p))
```
screen shot 2016-04-16 at 1 17 02 pm

If you merged master, you could add a `Label` annotation above each column, but I think this would also be fine as-is.

bryevdv:
BTW the next step (in another PR) is to allow transforms to be able to be specified in "DataSpecs" and then they will just get applied automatically. It's a little awkward to have to prompt users to press a button for now, but that's OK.

bryevdv:
And here is an updated example for interp:
```
import numpy as np

from bokeh.io import vplot, hplot
from bokeh.core.properties import NumberSpec
from bokeh.models import CustomJS, TextInput, Select, Div
from bokeh.models.transforms import LinearInterpolator, StepInterpolator

N = 300

controls = ColumnDataSource(data=dict(x=[1, 2, 3, 4, 5], y=[2, 8, 7, 3, 5]))
source = ColumnDataSource(data=dict(x=[], y=[]))

linear = LinearInterpolator(x='x', y='y', data=controls)
step = StepInterpolator(x='x', y='y', data=controls)

p = figure(x_range=(0, 6), y_range=(0, 10))
p.circle(x='x', y='y', source=controls, size=15, alpha=0.5, color="firebrick")
p.circle(x='x', y='y', source=source, size=2, alpha=0.2, color="navy")

callback = CustomJS(args=dict(source=source, linear=linear, step=step), code="""
    var mode = cb_obj.get('value');
    var data = source.get('data');
    var dx = 6 / %d;

    if (mode == 'None') {
        data['x'] = [];
        data['y'] = [];
    }
    else {
        if (mode == 'Linear') { interp = linear; }
        else if (mode == 'Step (before)') { interp = step; step.set('mode', 'before'); }
        else if (mode == 'Step (center)') { interp = step; step.set('mode', 'center'); }
        else if (mode == 'Step (after)')  { interp = step; step.set('mode', 'after');  }

        for (i=0; i < %d; i++) {
            data['x'][i] = i * dx
            data['y'][i] = interp.compute(data['x'][i])
        }
    }

    source.trigger('change')

""" % (N, N))

mode = Select(
    title='Interpolation Mode',
    value='None',
    options=['None', 'Linear', 'Step (before)', 'Step (center)', 'Step (after)'],
    callback=callback)

output_file("transform_interpolator.html", title="Example Transforms")

show(vplot(hplot(mode), p))
```

However, I think there may be a small bug with before/after step inter. Take a look at the screen shot for "after":

screen shot 2016-04-16 at 1 44 06 pm

Look at the values between `x=[0,1]`, they are at `y=0`, but they should be at `y=1`, correct?

bryevdv:
Same comment for linear:

screen shot 2016-04-16 at 1 46 09 pm

I think in general it should always be the same consistent policy: interpolating before the first control point, or after the last control point, always just returns the first or last control point value respectively. The only other possibly sensible thing I can think would be to return `nan` outside the control points.

bryevdv:
OK, so I added two new versions of the examples for you to consider. Then there is just these two issues:
* consistent interpolation outside control points
* remove `mean` property (should always be zero I think)

justacec:
@bryevdv Ok, going to work on this tonight (god willing)

Concerning the remove the `mean` request.  I would suggest keeping it just in case users want that feature for some reason that we do not know yet.  I can think of at least one now.  By keeping it in and defaulting it to zero the users have the ultimate freedom.

Concerning the "interpolation" outside the control space, we should return `nan`.  Interpolation in that space is simply undefined.

justacec:
@bryevdv Ok, I think I got it all in. 

Concerning the DataSpec.  Yes, that sounds cool!  I will wait for this to be agreed to be merged and then I will look into the adding the necessary elements to handle the DataSpec.

bryevdv:
@justacec The reason I suggested returning the start/end values for interpolates outside `[start, end]` is that I think that is the behavior that you will find you need for the color mapper uses that started this. :) For color mapping it is very common to want to "clamp" values outside the range to the top or bottom color of the palette, which would fall out naturally with that behavior I think.

bryevdv:
Can you describe the use case for `mean` that you have in mind?

bryevdv:
Regarding the "how can we use line here" I have some ideas, but let's leave that to another PR. Basically I think you could make an interpolator a property of the `Line` glyph, but there are some implications to consider, namely if we have to add additional "synthetic" points to make the visual look right, how does that intersect with hit-testing/selection.

justacec:
@bryevdv on the subject of the interpolation boundary: Good call.  My inner physicist is struggling with the concept of returning clamped values for a prediction outside the well defined range.  I do agree with your assessment that color clamping would most likely the way a color mapper would be implemented.

Got it!  new parameter.  "clip" which defaults to True.  When True it would return nan outside the range.  When False, it would return the clamp values.  Now, an extension for this could be an enum called clip_mode which would default to "truncate" but then also allow "nearest" which would cover the case that you suggested.  What are your thoughts here.

justacec:
@bryevdv on the subject of the mean: Specifically I was thinking about using it to simulate a dodge effect when you have two different classes of points which are degenerate.  By specifying a mean shift you can bias on set of data one way and then the other set of data the other way.  Outside of specific examples, I think that the the amount of code that is present to implement the bias is low and gives flexibility to the end user.  By removing that feature we would be deciding for anybody else that they can not use it.  Since it is already coded and does not introduce any secondary problems, I lobby to keep the feature and default it to zero.

bryevdv:
Re: `mean`, you have convinced me!

re: clipping, I was about to suggest making it a policy, so that sounds good. If there's just two values for the enum, in this case I think a `Bool` property is fine. `clip` seems OK, not sure if there is a better name, or `strict_bounds` might be more explicit but also longer.

justacec:
@bryevdv Ok, I will put in the clipping mode as a bool tonight sometime.

justacec:
@bryevdv @mattpap this is ready for review.

bryevdv:
@justacec I was just coming here to merge this after a long weekend, and of course there is a new conflict. But it is a tiny one:
```
bryan@0199-bryanv (git:transforms|MERGING) ~/work/bokeh $ git diff
diff --cc bokeh/core/enums.py
index 94ce1bd,cb0fbda..0000000
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@@ -173,6 -170,5 +173,11 @@@ StepMode = enumeration("before", "after
  #: Specify different callback policies for the slider widget
  SliderCallbackPolicy = enumeration("continuous", "throttle", "mouseup")
  
++<<<<<<< HEAD
 +#: Specify a distribution to use for the Jitter class
 +JitterRandomDistribution = enumeration("uniform", "normal")
 +
++=======
+ #: Specify sorting directions
 -SortDirection = enumeration("ascending", "descending")
++SortDirection = enumeration("ascending", "descending")
++>>>>>>> master
```
I'm actually not quite sure what's going on there, that looks very simple and straightforward, I'm not sure what has Git confused. As soon as you merge master and push and Travis is green I will merge. 

Next time I'd definitely recommend a branch off the main repo, as it can help make collaboration more efficient in some cases like this (e.g. I would fix this merge myself tight quick, if I could)

mattpap:
> I'm actually not quite sure what's going on there

The problem is with newline character on last line (or lack of it on `master`). When #4169 is merged, this won't be a problem anymore (newline character will be required at the end of each source file).

justacec:
@bryevdv Ok.  Got the master merged in

bryevdv:
Thank you @justacec for your patience and great contribution!

Issue #3968, opened 2016-03-01
Document usage of a reverse tunnel to an instance of a standalone server


machinelearningdeveloper:
This was already discussed on the Bokeh mailing list.

Per @bryevdv, insert new documentation at about the following location:

https://github.com/bokeh/bokeh/blame/master/sphinx/source/docs/user_guide/server.rst#L496

machinelearningdeveloper:
I will issue a PR against this issue.

bryevdv:
Brilliant, thanks!

PR #3969, opened 2016-03-01
Document use of SSH tunnels with Bokeh. #3968.

machinelearningdeveloper:
SSH can be used to create tunnels between a remote host with a running instance of the Bokeh server and a local machine, and to access the remote Bokeh server as if it were running on the local machine.  The documentation attached to this PR describes how to create such tunnels.


bryevdv:
Thanks @machinelearningdeveloper !

damianavila:
Closed by #3969

PR #3974, opened 2016-03-01
Add cross browser capabilities

birdsarah:
Fixes: #3827 

This adds a command line argument --cross-browser, that causes tests to run against Firefox, Chrome, and Internet Explorer. It leaves the default as Firefox.

It also switches the default platform to run against to Windows 10 because:
* windows is needed to run internet explorer
* probably good to have our integration tests running against windows as no regular bokeh devs use windows


birdsarah:
superseded by #3977

PR #3975, opened 2016-03-01
Visual diff test

birdsarah:
Fixes: #3826

Adds a `screentest` fixture with an `is_valid` method:
* adds a --set-new-base-screenshot command line parameter for setting a baseline
* uses same image-diff testing as examples
    
Still TODO:
- [x] incorporate screenshots and diff into pytest html report
- [x] upload screenshots and diff into
- [x] prevent screenshot tests running on wrong configuration

Future work:
* hook in to perceptualdiff's tolerance feature - on a per-test basis?


birdsarah:
Here is an example output when a screenshot test fails:
https://s3.amazonaws.com/bokeh-travis/112997379/tests/pytest-report.html

screen shot 2016-03-01 at 3 42 47 pm

birdsarah:
This is ready.

Includes:
* an initial screenshot diff test
* screenshot class that, hopefully, makes it easy to add further screenshot tests
* checks so that screenshots can only be generated and tested via saucelabs driver (when using firefox and linux) to ensure a uniform screenshot test across developers

PR #3977, opened 2016-03-02
Add cross browser capabilities

### LINKED ISSUE #3827: Cross browser automated testing

birdsarah:
Fixes #3827 

Builds on #3975 (which should be merged first)

Supersedes #3974 

cross browser tests need to be explicitly written to be cross browser so the author can take care with any selenium nuances. This implements a marker that marks a test to be run as a cross browser test, rather than providing a global option. 

This should make it easier to get started with cross-browser tests and provides ongoing flexibility to write targeted tests (e.g. just something for touch).

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -53,9 +53,9 @@ test:
     - flake8
     - mock
     - pandas
-    - pytest
+    - pytest >=2.9.0
     - pytest-cov ==1.8.1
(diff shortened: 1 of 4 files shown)


birdsarah:
@bokeh/dev - will merge this after a successful test run. (doing it with a little urgency as pytest-1.2.1 is causing issues on other people's test runs - have talked to @bryevdv about it)

PR #3990, opened 2016-03-03
Revert "Some canvas and layout improvements"

birdsarah:
Reverts bokeh/bokeh#3913

This PR appears to have broken crossfilter. @bryevdv plans to add small bits back incrementally.


mattpap:
Is this a non-trivial bug? This change interferes with my work greatly.

bryevdv:
Let me try using this branch to identify the problem by working forward. Then maybe we can apply a simple fix to current master. @birdsarah hold off on merging

birdsarah:
> Is this a non-trivial bug? This change interferes with my work greatly.

@mattpap, I believe it's something to do with the crossfilter.

bryevdv:
It's completely not rendering whole-plot replacements, and I was not able to figure out a fix working backwards from master.

PR #4010, opened 2016-03-09
support --args to pass command line args to bokeh apps

### LINKED ISSUE #2026: Allow bokeh server scripts to take command-line arguments
I'm writing a script that I'd like to take a command-line argument. I can't figure out a way to do that without resorting to terrible hacks (which I'm not entirely sure would work anyway). Has this come up before? Any suggestions for workarounds in the meantime?

bryevdv:
issues: closes #2026

### DIFF
--- a/bokeh/application/handlers/code.py
+++ b/bokeh/application/handlers/code.py
@@ -22,10 +22,20 @@ class CodeHandler(Handler):
     _io_functions = ['output_server', 'output_notebook', 'output_file',
                      'show', 'save', 'push', 'reset_output']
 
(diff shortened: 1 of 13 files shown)


bryevdv:
@bokeh/dev this is ready for review

bryevdv:
OK @bokeh/dev, I am going to go ahead and merge this. There was no feedback, but hopefully it should be a fairly uncontroversial addition. 

I do think it seemed to require a little to much tinkering/plumbing than I expected, perhaps there are some useful refactors that could help make things like this a little more self-contained.

PR #4011, opened 2016-03-09
clarify that default mode is 'cdn' rather than 'inline'

electronwill:
Clarify that default mode for output_file() is 'cdn' rather than
'inline'.


bryevdv:
Error is spurious, merging. 

@birdsarah, FYI the py3 examples test seems to be failing fairly reliably on this:
```
[gw0] FAILED tests/examples/test_examples.py::test_file_examples[/home/travis/build/bokeh/bokeh/examples/compat/pandas_dataframe.py] --- Logging error ---
```

PR #4013, opened 2016-03-10
fix notebook and toolbar css handling

### LINKED ISSUE #4012: Jupyter Notebook CSS is tied to the notebook structure
We have a [dashboard server](https://github.com/jupyter-incubator/dashboards_server) that executes notebook files and displays the output. I see some notebook-specific styling to clean up a few things in [continuum.less](https://github.com/bokeh/bokeh/blob/819663e2edc47aa4373b230412279aef6ef01ee0/bokehjs/src/less/continuum.less#L19). The `#notebook` selector bolts the styling to the notebook structure, but we need this styling to override some of our own table styling and would rather not mimic the notebook structure just to get this styling applied.

Is there some way this styling can be made less specific so that plots generated after loading `output_notebook` can get this styling? This may also be helpful in the future as Jupyter Notebook evolves into Jupyter Lab.

You can see the extra borders appearing around the Bokeh plot in this screenshot:

image

bryevdv:
issues: fixes #4012 

This fixes the visible table borders in notebooks and also problems occurring with toolbar lines.

### DIFF
--- a/bokehjs/src/less/continuum.less
+++ b/bokehjs/src/less/continuum.less
@@ -22,8 +22,6 @@
     border: none;
     td {
       border: none;
-      padding: 0;
-      margin: 0;
     }
   }
 }
--- a/bokehjs/src/less/continuum.less
+++ b/bokehjs/src/less/continuum.less
@@ -16,7 +16,7 @@
 .tableheader {
   background-color: #C0C0C0;
 }
-#notebook .bk-plot-wrapper table {
+.bk-plot-wrapper table {
   border: none !important;
   tr {
     border: none !important;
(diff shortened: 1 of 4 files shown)


bokehjs/src/less/continuum.less
```diff
@@ -16,7 +16,7 @@
 .tableheader {
   background-color: #C0C0C0;
 }
-#notebook .bk-plot-wrapper table {
```
mattpap: So why this was bound to notebook structure before? Removing this implies that the following styles will be applied always. Is this intended?

mattpap: Anyway, removing fixed ID-based styling is mandatory as this is something a **library** must not use.

bryevdv: I'm not sure why it was added originally, but at least preliminary testing indicates that it does not seem needed (and I do think those styles should be universally applied?) and as you indicate it really needs to go in any case.

bryevdv:
sorry I forgot to link the issue, removing issue tags

bryevdv:
@bokeh/dev I will merge this tomorrow unless there is objection. ping @birdsarah

bryevdv:
@birdsarah this error seems to be something that needs to be fixed up somehow?

https://travis-ci.org/bokeh/bokeh/jobs/114975767

bryevdv:
@birdsarah OK I removed the `padding` and `margin` that were previously applied only to `#notebook`. Now all the non-notebook examples match up exactly, and the notebook examples have a slight offset in the diff. I'm not actually sure this isn't a good thing. At least, I can't think of why they ought to be different. Anyway, every thing is green and report looks good by inspection and local testing, so I am going to go ahead and merge this.

PR #4018, opened 2016-03-10
Fix up formatting of calendars example

birdsarah:
Not sure when this got a bit borked, but this is what master was looking like for me:

screen shot 2016-03-10 at 12 33 22 pm

and here's after this PR:

screen shot 2016-03-10 at 12 32 45 pm


mattpap:
1. Wrapped tooltips is an issue on its own. I will look into this in hover management PR.
2. Plot titles issue looks like a bug introduced recently.

bryevdv:
Will merge this as soon as tests pass

bryevdv:
(Agreed there is more work to do but these tweaks do no harm)

PR #4021, opened 2016-03-11
HTML injection widget

### LINKED ISSUE #4008: Allow for the injection of raw HTML code
Currently, the Widget library contains a `Paragraph` and a `PreText` widget allowing the user to put basic text on the rendered page.  Neither of these widgets allows for the inclusion of formatted text using HTML markup.  A widget should be added to support the inclusion of raw HTML.  The widget can be a simple div named div tag.

justacec:
This is a simple PR which creates a new `Div` widget.  The widget renders to an HTML <div> tag.  Be default, the text property of the object will render as HTML code.  However, the boolean `render_as_text` property can be set to `True` and the text property will be rendered as text with not HTML interpretation.  

The code to use would look something like this:

    para = Div(text = "<H1>Slider Values:</H1><P>Slider 1: 0<P>Slider 2: 0<P>Slider 3: 0", width = 200, height = 150, render_as_text = False)

To demonstrate this new feature, I modified the `slider_callback_policy.py` example in the `examples/plotting/file` directory to use the `Div` widget with HTML rendering (I think it looks much better now)

A screen shot of this is the following:

screen shot 2016-03-10 at 9 53 46 pm

Caution: I simply copied the `Paragraph` widget and changed the name to `Div` and added some small logic to alter the `text` function call to an `html` function call.  It may be possible to combine the features of the `Paragraph` and the `Div` widgets into a single widget.

Fixes #4008


justacec:
@birdsarah This is ready for some initial review and comment.

bokeh/models/widgets/markups.py
```diff
@@ -28,6 +28,26 @@ class Paragraph(Markup):
...
+    width = Int(500, help="""
+    The width of the block in pixels.
+    """)
+
+    height = Int(400, help="""
+    The height of the block in pixels.
+    """)
+
+    render_as_text = Bool(False, help="""
+    Should the text be rendered as raw text (False), or should the text be interprited as an HTML string (True)
```
birdsarah: Specify the defaults in all of the help strings.

bokeh/models/widgets/markups.py
```diff
@@ -28,6 +28,26 @@ class Paragraph(Markup):
...
+class Div(Markup):
+    """ A block (div) of text.
+
+    """
+
+    text = String(default="", help="""
+    The contents of the widget.
+    """)
+
+    width = Int(500, help="""
```
birdsarah: I would suggest a default width of 600 as that's the default width of a plot

bokeh/models/widgets/markups.py
```diff
@@ -28,6 +28,26 @@ class Paragraph(Markup):
...
+    The width of the block in pixels.
+    """)
+
+    height = Int(400, help="""
```
birdsarah: This seems large, but I don't feel strongly about it.

examples/plotting/file/slider_callback_policy.py
```diff
@@ -20,10 +20,10 @@
       };
     }
 
-    para.set('text', "Slider Values\\n\\n  Slider 1: {0}\\n  Slider 2: {1}\\n  Slider 3: {2}".format(s1, s2, s3))
+    para.set('text', "<H1>Slider Values</H1><P>Slider 1: {0}<P>Slider 2: {1}<P>Slider 3: {2}".format(s1, s2, s3))
```
birdsarah: It's custom to use lowercase for html tags.

examples/plotting/file/slider_callback_policy.py
```diff
@@ -1,5 +1,5 @@
 from bokeh.io import vform, output_file, show
-from bokeh.models import CustomJS, Slider, Paragraph, PreText
+from bokeh.models import CustomJS, Slider, Div, PreText
```
birdsarah: Why have you kept the PreText import

examples/plotting/file/slider_callback_policy.py
```diff
@@ -20,10 +20,10 @@
       };
     }
 
-    para.set('text', "Slider Values\\n\\n  Slider 1: {0}\\n  Slider 2: {1}\\n  Slider 3: {2}".format(s1, s2, s3))
+    para.set('text', "<H1>Slider Values</H1><P>Slider 1: {0}<P>Slider 2: {1}<P>Slider 3: {2}".format(s1, s2, s3))
 """)
 
-para = PreText(text = "Slider Values:\n\n  Slider 1: 0\n  Slider 2: 0\n  Slider 3: 0", width = 200, height = 150)
+para = Div(text = "<H1>Slider Values:</H1><P>Slider 1: 0<P>Slider 2: 0<P>Slider 3: 0", width = 200, height = 150, render_as_text = False)
```
birdsarah: Shouldn't be spaces for arguments e.g. `Div(text="<h1>Slider Values:", width=200)` etc

birdsarah:
@justacec my sincere apologies for the delay I had a bit of a fail with spam folder and my github notifications.

This looks good, I've made some minor comments.

Two bigger things:
1) The example should be under `models` not `plotting/file`
2) The unit test failures are real. Python and JS properties should match each other.

birdsarah:
I just realized you didn't add the example! So you can ignore that comment :D

bokeh/models/widgets/markups.py
justacec: I went ahead and changed that.  Based on this comment, should the default width of the Paragraph class also be changed to 600.  I was using that as a model.

bokeh/models/widgets/markups.py
justacec: Large = SLOC?
Large = Height?

This could be dramatically dropped down if these defaults are pulled into the Markup class and the Paragraph and the Div are simple subclasses with the whole intent on just changing the tag type.  This was what I was getting at in my original comments.  There is a lot of code duplication.  I was just trying to get something present which could be cleaned up later.

examples/plotting/file/slider_callback_policy.py
justacec: Corrected

examples/plotting/file/slider_callback_policy.py
justacec: Removed.   Just missed it.

examples/plotting/file/slider_callback_policy.py
justacec: Corrected.  @mattpap has hit me up on this as well.  Just a habit from my personal style.

bokeh/models/widgets/markups.py
justacec: Added

justacec:
@birdsarah Thanks for getting back to me.  I just assumed there were many many other more important sub-projects being worked on.  I believe that I have addressed all of your comments with the exception of the "is this too big" style comment.  As I suggested in there, it may be possible to dramatically reduce the amount of duplication if we decide to suck all of the matching properties into a base class (possibly the Markup class) and then just subclass the Paragraph and the Div tag from that.  We could also just make the div tag the real workhorse and then make the paragraph an extension which ensure that the render_as_text property is set to true.

In the end there are a couple of different directions here.  Which would you like for me to move toward?

birdsarah:
Hi @justacec 

> Large = SLOC?
Large = Height?

> This could be dramatically dropped down if these defaults are pulled into the Markup class and the Paragraph and the Div are simple subclasses with the whole intent on just changing the tag type. This was what I was getting at in my original comments. There is a lot of code duplication. I was just trying to get something present which could be cleaned up later.

I just meant the height value - 400px is big.

I am +1 on de-duping code, I am also fine with getting something in that can be cleaned up later.

It's up to you how you want to proceed.

Even with no additions, I would say this is ready to go.

ping @bokeh/dev for a second review.

bryevdv:
Should the be marked `ready` ?

fpliger:
LGTM ...

bryevdv:
I agree with @birdsarah that we can merge this as-is, but there is also room for some re-factoring code-sharing. @justacec can you indicate whether you'd like to do the `Markup` factor in this PR? If not that's fine we will just need another issue to capture that task, then we can merge this.

justacec:
@fpliger @birdsarah @bryevdv I have now updated this by moving the duplicated code into the Markup widget as a base class and then removed the duplicated code in the derived classes.  I have also merged in master so this is current.  This is ready for final review.  After that review I would say it is good for merge.

justacec:
I am assu

bryevdv:
@justacec the js unit test is failing because there is a new way to specify properties, here is an example from `Range1d`: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/ranges/range1d.coffee#L9-L13

```
  @define {
      start:  [ p.Number, 0 ]
      end:    [ p.Number, 1 ]
      bounds: [ p.Any       ] 
    }
```
If you change to that, I think everything will pass.

justacec:
@bryevdv Ok, updated the props structures as suggested.  Looks like things are going to be good with this change.  Lets see if this CI builds clean.

birdsarah:
This is looking good. I don't want to hold-up merging, but we should add an issue to put this (and any other widgets we may have missed) into the user guide: http://bokeh.pydata.org/en/latest/docs/user_guide/interaction.html#textinput

bryevdv:
@birdsarah good point. I will merge this now, and make an issue to add it to the users guide. 

Thanks @justacec !

Issue #3993, opened 2016-03-04
Updating a Span (line annotation) duplicates it


philippjfr:
When updating an existing ``Span`` (line annotation) it is rendered again and again without deleting the old line.

Here is an example that reproduces the issue (adapted from [this](http://bokeh.pydata.org/en/latest/docs/user_guide/plotting.html#spans) example in the documentation):

```python
from datetime import datetime as dt

from bokeh.sampledata.daylight import daylight_warsaw_2013
from bokeh.plotting import figure, show, output_file
from bokeh.models import Span

output_notebook()

p = figure(x_axis_type="datetime", y_axis_type="datetime")

p.line(daylight_warsaw_2013.Date, daylight_warsaw_2013.Sunset,
       line_dash='solid', line_width=2, legend="Sunset")
p.line(daylight_warsaw_2013.Date, daylight_warsaw_2013.Sunrise,
       line_dash='dotted', line_width=2, legend="Sunrise")

daylight_savings_start = Span(location=(dt(2013, 3, 31, 2, 0, 0) - dt(1970, 1, 1)).total_seconds()*1000,
                              dimension='height', line_color='green',
                              line_dash='dashed', line_width=3)
daylight_savings_end = Span(location=(dt(2013, 10, 27, 3, 0, 0) - dt(1970, 1, 1)).total_seconds()*1000,
                            dimension='height', line_color='red',
                            line_dash='dashed', line_width=3)
p.renderers.extend([daylight_savings_start, daylight_savings_end])

p.title = "2013 Sunrise and Sunset times in Warsaw"
p.yaxis.axis_label = 'Time of Day'

show(p)
```

When updating the line position like this:

```python
from bokeh.io import push_notebook
daylight_savings_start.set(location=(dt(2013, 1, 28, 2, 0, 0) - dt(1970, 1, 1)).total_seconds()*1000)
push_notebook() 
```

It will duplicate the line like this:

![image](https://cloud.githubusercontent.com/assets/1550771/13532095/91cc1542-e221-11e5-9f48-1f51046fa3d0.png)

bryevdv:
I'm at a loss, the render path does not have any loops or repetitions:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/annotations/span.coffee#L62-L75

That could be responsible for this. ping @canavandl

mattpap:
`Span` in canvas mode doesn't request render of the plot, so it doesn't remove previously painted stuff.

[...]

PR #4028, opened 2016-03-14
Canavandl/3992 refactor annotations

### LINKED ISSUE #3992: BoxAnnotation does not update consistently
When trying to update a BoxAnnotation it only updates the first time you push the new data, any subsequent changes get ignored unless you change some other plot attributes as well. I've also looked whether the events are getting sent and that seems to be the case, so it's probably an issue with triggering an update on the Coffeescript end.

Here is an example:

```python

output_notebook()

#reduce data size

low_box = BoxAnnotation(plot=p, top=80, fill_alpha=0.1, fill_color='red')
mid_box = BoxAnnotation(plot=p, bottom=80, top=180, fill_alpha=0.1, fill_color='green')
high_box = BoxAnnotation(plot=p, bottom=180, fill_alpha=0.1, fill_color='red')

p.renderers.extend([low_box, mid_box, high_box])

p.title = "Glucose Range"
p.xaxis.axis_label = 'Time'
p.yaxis.axis_label = 'Value'

show(p)
```

If we now adjust the the ``top`` of the low_box like this:

```python
low_box.set(top=100)
push_notebook() 
```

nothing happens. If we also set a new title it does update:

```python
p.title = 'Update test'
low_box.set(top=100)
push_notebook()
```

canavandl:
issues: fixes #3992 #3993 

Refactor annotations to support DataSources (and result in vectorized implementation)

- [x] BoxAnnotation
- [x] Span
- [x] PolyAnnotation


bokeh/models/annotations.py
```diff
@@ -111,8 +112,9 @@ class BoxAnnotation(Annotation):
 
     """
 
-    left = Either(Auto, NumberSpec(), default=None, help="""
-    The x-coordinates of the left edge of the box annotation.
+    left = NumberSpec(default=None, help="""
+    The x-coordinates of the left edge of the box annotation. If value is set
+    to `None`, the box annotation will continue to the edge of the plot area.
     """)
```
bryevdv: This will break code. @mattpap added a transform option of some sort, that could be used to turn "auto" into `None` which might be one option.

canavandl: Ok - I can't currently remember my reasoning for removing that. I'll add it back (it shouldn't impact my code)

canavandl: So i'm not sure the original code ever worked as intended. The dimension parameters didn't work as specs (there wasn't any sort of datasource lookup) and there was no JS-side code to handle any sort of "auto" values.

The annotation examples in the user guide also still work (using default `None` value and not Auto) the same with the new implementation.

Do you want to add Auto and do the transforms or just have the dimensions be NumberSpecs?

canavandl:
todo:
* [ ] document new plot.add_annotation method
* [ ] refactor examples to use add_annotation method
* [ ] document new poly annotation

for discussion:
* breaking `adding annotations` into stand-alone document section (probably wise)

canavandl:
I'm not super pleased w/ the `xs` and `ys` attrs being NumberSpecs. I want them to be `specs` and was trying to mimic the API for patches and multi_lines, but it causes validation errors when trying to instantiate a PolyAnnotation obj using lists, like:

```python
> PolyAnnotation(xs=[1,2,4], ys=[1,5,3])
ValueError: expected an element of either String, Dict(String, Either(String, Float)) or Float, got [1, 2, 4]
```

Should we consider adding a `SeqSpec` (or NumberSeqSpec?) property type to handle these sorts of situations? It would also help patches/multi_lines

canavandl:
User guide examples of PolyAnnotation:

screen shot 2016-03-18 at 4 14 38 pm

bryevdv:
Maybe poll the mailing list for some input about `PolyAnnotation` My feeling is that the most critical annotation to vectorize is label (because you might want to attach it to glyphs) and that vectorization on the other annotations is a "(maybe) nice to have" but not a "must have"

canavandl:
Todo: Also need to add annotation-level clipping for css render modes.

bryevdv:
I realize there was not really any feedback on the ML. I guess my suggestion would be this: just these annotations represent single boxes, polys, spans for now (so no "spec" and no "source"). If we find a demand for being able to "vectorize" them, then it would certainly be possible to add that later in a non-breaking way (maybe a "sequence spec" as you suggested)

bryevdv:
@canavandl if you agree, please close as `rejected`

canavandl:
I agree. I'll reject/supercede this PR with another, as there were several annotation improvements in addition to vectorization (exposing PolyAnnotations on the python side, updating properties, etc.)

I'll cherry-pick those commits into a new PR shortly.

mattpap:
>  If we find a demand for being able to "vectorize" (...)

I think we shouldn't create another set of glyphs here. If somethings needs to be vectorized, I think it should belong to glyphs. Annotations should have other properties, like binding to edges of canvas (e.g. attribution), or being able to figure out optimal position on-screen (e.g. legends; not yet).

bryevdv:
> If somethings needs to be vectorized, I think it should belong to glyphs.

I *almost* agree. I do think attaching text labels to many glyphs is such a common thing that it definitely merits "vectorization" And although it is similar to `text`, I also think labels (as annotations) are different from "text scatter" (i.e. do want to support screen coordinates, don't care about hit-testing, at least not in data space with an index)

PR #4029, opened 2016-03-14
Fix output_notebook(resources) default docstring

parente:
Mentioned in issue #3639


bryevdv:
HI @parente thanks for the PR! Can I suggest just `default: CDN` as the text? It is a slight fib, but it's indistinguishable to general users, and I think, less confusing. It's only the way it is to defer importing `bokeh.resources` as long as possible, for obscure an wholly uninteresting reasons.

parente:
> Can I suggest just default: CDN as the text? 

All set.

PR #4033, opened 2016-03-16
Make docserver.py run on both windows and unix by using threading instead of os.fork.

### LINKED ISSUE #4027: docserver.py is not windows compatible
The `docserver.py` script is not windows compatible since it uses `os.fork`. Is there an alternative way of running a doc server on windows?

gonike:
issues: closes #4027

See the discussion in [this issue](https://github.com/bokeh/bokeh/issues/4027).

I exchanged the use of `os.fork()` in `sphinx/docserver.py` against using the `threading` module, which should work regardless of the os. I also added the possibility to shutdown the server.

My implementation is mainly due to [this GH answer](http://stackoverflow.com/a/17325148/2207840).

Caution: My experience with web programming in python is near-zero, so there is a real chance that what I did is plain nonsense. I tried the new version on both Win 7 and Arch linux and it seemed to work on both systems.

Happy to provide further relevant information!

### DIFF
--- a/sphinx/docserver.py
+++ b/sphinx/docserver.py
@@ -2,15 +2,21 @@ from __future__ import print_function
 
 import flask
 import os
+import threading
+import time
+import webbrowser
 
 from tornado.wsgi import WSGIContainer
 from tornado.httpserver import HTTPServer
 from tornado.ioloop import IOLoop
 
+
+
 _basedir = os.path.join("..", os.path.dirname(__file__))
 
 app = flask.Flask(__name__, static_path="/unused")
 PORT=5009
+http_server = HTTPServer(WSGIContainer(app))
 
 """this is a simple server to facilitate developing the docs.  by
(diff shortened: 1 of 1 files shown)


gonike:
I see that some python 2.7 checks have failed. Does the code have to be python 2.7 compliant?

bryevdv:
@gonike Yes Bokeh needs to work equally on python 2.7 and python 3.4+ That said, we currently have issues with spurious test failures sometimes, I suspect that is the case here. I will restart the test and we can see if it passes.

bryevdv:
Also, thanks for the PR!

bryevdv:
@gonike one small change, `input` actually waits for `<enter>`, so not just any key will work. May I suggest changing the wording int `u()` to `"Press <ENTER> to exit..."`

bryevdv:
Tests now passing, and tested locally. Merging now, thanks @gonike !

PR #4034, opened 2016-03-16
Fix typo: supervisctl -> supervisorctl

luizirber:
On the server documentation, there is a typo in the calls to `supervisorctl`.


bryevdv:
Thanks for the catch @luizirber !

PR #4037, opened 2016-03-17
Implementation of TypeScript API

mattpap:
- [x] add support for getters/setters (`Object.defineProperty`)
- [x] fix all clashes between unrelated `this.prop` and `this.get("prop")`
- [x] implement CoffeeScript/TypeScript mid-level API (bokeh.plotting)
 - [x] finalize edge cases
 - [x] add support for legends
 - [x] add support for tools
 - [ ] ~~add remaining API functions (io, hold, etc.)~~
- [x] TypeScript typings for low- and mid-level APIs
 - [x] add support for included properties (mixins with prefix)
 - [ ] make `attrs` in `new Model(attrs)` type-safe without too much code duplication
 - [ ] figure out how to remove `string` from `Color` union type
- [x] make `gulp test` work again
 - [x] fix unit tests
 - [ ] ~~add support for on-the-fly typescript compilation~~
- [x] add examples:
 - [x] low-level (`bokehjs/examples/anscombe`)
 - [x] mid-level (`bokehjs/examples/burtin`)

Note: to compile everything (scripts, typings and examples), issue `gulp examples`.


sccolbert:
If the intent of these files is to define the typings for the coffeescript implementations, they should be written as `d.ts` files instead of normal `.ts` files.

Here's a snippet from the builtin typescript api file for how to declare an instantiable class:

`lib.core.d.ts`
```typescript
interface Error {
    name: string;
    message: string;
}

interface ErrorConstructor {
    new (message?: string): Error;
    (message?: string): Error;
    prototype: Error;
}

declare var Error: ErrorConstructor;
```

mattpap:
@bryevdv, new commits require attention. Currently we use `set()` and `get()` to manage properties. This is cumbersome when defining an API, so I added support for setters and getters to models. In this PR you can do `grid_plot.children` as well as `grid_plot.get("children")`. It would be too easy if it was the end of the story. Currently we set a lot of `this.something` attributes which often clash with actual properties (e.g. `this.palette = do_something(this.get("palette"))`). I attempted to clean all this up, but we will need to verify this (I'm sure I missed plenty of cases; dynamically typed programming languages are horrible in such an event).

mattpap:
@sccolbert, I originally intended to implement a bridge layer on top of those interfaces, but as I managed to add setters/getters to bokeh's models, now I can just use definition files (as suggested).

bryevdv:
These changes make me intensely happy. Thanks for figuring this out @mattpap I'd say make a best effort to ensure every `@foo` that needs changing is changed, and then merge. We will do extensive testing before the next release.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -730,15 +724,37 @@ class Plot extends Component.Model
...
+
+    renderer = new GlyphRenderer.Model({data_source: source, glyph: glyph})
+    @add_renderers(renderer)
+
+    return renderer
+
+  add_tools: (tools...) ->
+    for tool in tools
+      tool.plot = this
+      @set("tools", @get("tools").concat([tool]))
```
almarklein: if you place this out of the loop, you can extend the Array just once

bokeh/plotting/helpers.py
```diff
@@ -291,110 +291,6 @@ def _process_tools_arg(plot, tools):
...
-
-    for typename, group in itertools.groupby(sorted([ tool.__class__.__name__ for tool in plot.tools ])):
-        if len(list(group)) > 1:
-            repeated_tools.append(typename)
-
-    if repeated_tools:
-        warnings.warn("%s are being repeated" % ",".join(repeated_tools))
-
-    return plot
-
```
bryevdv: Heh, I did this in an other open PR...

bryevdv: Here: https://github.com/bokeh/bokeh/pull/4046 I think it makes more sense in that PR anyway, if you want to back out the commit to make this one smaller

bokehjs/src/coffee/models/plots/plot.coffee
almarklein: mmm, also, tools seem to need to have the plot object at instantiation (at least with current master).

bokeh/plotting/helpers.py
mattpap: I will revert this when I'm done with the API.

mattpap:
@bryevdv, @almarklein, so the plotting API is finally in a usable state. `burtin` example fully works (to build it issue `gulp examples`).

mattpap:
@bokeh/dev, if you have spare time, please tests this PR (specifically the "fix clashes between `this.prop` and `this.get("prop")`" part).

mattpap:
@almarklein, looking at how tools are implemented, it's unlikely I will make it work soon. Even if we set `plot` early, then a tool has to listen to changes to `renderers` and not precompute and cache stuff in the constructor.

almarklein:
> premature caching (non-lazy caching)

I encountered a case of premature caching in the ranges. Had to trigger an update manually (https://github.com/dwillmer/phosphor-bokeh-demo/blob/master/data/bokeh/bokehhelper.js#L77-L81)

mattpap:
@almarklein, this is why I have to set some properties (e.g. `x_range`) before calling super constructor on Figure, and the remainder can be set later. Unfortunately this doesn't work with `tool` because tools need back reference to `Plot` (`Figure`) and this back reference must be fully initialized.

mattpap:
@almarklein, I made tools work when added after `Plot` initialization.

bokeh/plotting/helpers.py
mattpap: After all I will leave it because `master` merged cleanly.

almarklein:
@mattpap you rock. I just tried to implement the stocks plot in TS using this API. It worked in one go. Awesome!

bryevdv:
One selenium tests failing, appears to be toolbar height slightly different, also missing logo: https://s3.amazonaws.com/bokeh-travis/118104359/tests/pytest-report.html

almarklein:
Would this be the way to add a legend? `new Bokeh.Line({x:{field:'t'}, y:{field:'foo'}, legend: 'foo');`, i.e. no extra action required on the plot? Or are legends this still a WIP?

bryevdv:
Not sure what's going on with examples tests, a bit strange: [gw1] FAILED tests/examples/test_examples.py::test_server_examples[/home/travis/build/bokeh/bokeh/examples/models/buttons_server.py] Traceback (most recent call last):
  File "/home/travis/miniconda/lib/python2.7/logging/__init__.py", line 874, in emit
    stream.write(fs % msg)
  File "/home/travis/miniconda/lib/python2.7/site-packages/_pytest/capture.py", line 232, in write
    self.buffer.write(obj)
ValueError: I/O operation on closed file
Logged from file file_server.py, line 111
  File "/home/travis/miniconda/lib/python2.7/logging/__init__.py", line 874, in emit
    stream.write(fs % msg)
  File "/home/travis/miniconda/lib/python2.7/site-packages/_pytest/capture.py", line 232, in write
    self.buffer.write(obj)

bokeh/models/widgets/buttons.py
```diff
@@ -26,7 +26,7 @@ class AbstractButton(Widget):
     An optional image appearing to the left of button's text.
     """)
 
-    type = Enum(ButtonType, help="""
+    button_type = Enum(ButtonType, help="""
```
bryevdv: This is a breaking API change, we need to be sure to document it in the release notes

bryevdv:
> add remaining API functions (io, hold, etc.)

`hold` hasn't been in Bokeh since like 0.5 :)

mattpap:
> Would this be the way to add a legend?

Yes, I will add it later today.

bokeh/models/widgets/buttons.py
mattpap: On Python-level I can add deprecation code to ease the transition. It's just bokehjs where this has to be `!= "type"`.

mattpap:
> hold hasn't been in Bokeh since like 0.5 :)

Yeah I know, but some form (definitively less stateful than the original) may be useful in JS API, because all updates to properties and thus all rendering is agile, so user may want to hold off rendering until he created all glyphs.

bryevdv:
@mattpap just merged a possible fix for the pytest file I/O issue to master

bokehjs/src/coffee/api/typings/models/sources.d.ts
```diff
@@ -5,10 +5,13 @@ declare namespace Bokeh {
...
 
+    type Data = {[key: string]: ArrayLike<any>};
+
     export var ColumnDataSource: { new(attributes?: KeyVal, options?: KeyVal): ColumnDataSource };
     export interface ColumnDataSource extends DataSource {
-        data: {[key: string]: ArrayLike<any>};
-        stream(new_data: Object, rollover: Number): void;
+        data: Data;
+
+        stream(new_data: Data, rollover: number): void;
```
almarklein: This feels a bit overly strict to me; defining a type for an arg of one function. But maybe that's just me being too used to Python's duck typing...

bokehjs/src/coffee/api/typings/plotting.d.ts
```diff
@@ -1,6 +1,8 @@
 declare namespace Bokeh.Plotting {
     function show(figure: Figure, target?: HTMLElement): void;
 
+    function color(r: number, g: number, b: number): string;
+
```
almarklein: Do we have this function on the Python side? If not, maybe `rgb2hex` would be more clear.

almarklein:
Running some examples. Some fail (verified that they work on 0.11.1):

* plotting/file/color_sliders.py -> no sliders (also python_callback.py, slider.py)
* plotting/file/glyphs.py -> stalls my browser
* plotting/file/grid.py -> tools look like ugly buttons instead of icons (also in line_compare.py, linked_brushing.py, linked_panning.py)
* plotting/file/image.py -> blanc: `this.properties.rows is undefined`, image_rgba.py seems to work though
* plotting/file/line_select.py -> picking fails: `TypeError: d0.glyph.glyph_view is undefined`
* plotting/file/clustering.py -> plots stack vertically instead of in a 2x2 grid.

mattpap:
@almarklein, I added support for labels.

bokehjs/src/coffee/api/typings/models/sources.d.ts
mattpap: It's actually a member and a function, and there might be other in future. This is static typing, i.e. you want to be as precise as possible. Note that this typing isn't that specific, because it still uses `any`. This could have been replaced naively with `number | string`, but that might be insufficient. Unfortunately typescript doesn't have tools to express the correct type, because you need an open (typeclass-based, like in Haskell) type system.

bokehjs/src/coffee/api/typings/plotting.d.ts
mattpap: Yeah, that's a temporary convenience function to make an example working.

bokehjs/src/coffee/api/typings/models/sources.d.ts
mattpap: As to `number` vs `Number` I'm not sure if it really matters, but `number` seems more natural.

mattpap:
> plotting/file/line_select.py -> picking fails: TypeError: d0.glyph.glyph_view is undefined

@bryevdv, that's your change.  I'm not sure how to get `visuals` from a model. Looking at the code, separation of rendering and data (view and model) is broken here.

mattpap:
> plotting/file/image.py -> blanc: this.properties.rows is undefined

Too much copy-paste. `Image` glyph doesn't have `rows`.

mattpap:
> plotting/file/glyphs.py -> stalls my browser (prob also in master)

It renders after some half a minute. That example needs real attention.

mattpap:
@birdsarah, can you look into examples' test failures? It doesn't make any sense to me.

bryevdv:
> EDIT: the easiest would be reverting removal of glyph_view.

Probably, unfortunately. It was removed in this commit:

https://github.com/bokeh/bokeh/commit/f004054b757c9f84969e2ec353d39b1986626e12

Because it's not used anywhere in the actual codebase, it seemed like a bad idea for views to set themselves on their models in this way. I'd forgotten that this one example depended on it, though.

almarklein:
>  I'd forgotten that this one example depended on it, though. 

And if I am not mistaken, we use similar code in the Pepsi project for line selection.

almarklein:
>  do you just want to put it back in this PR? We can make an issue to explore improving it later

Later is fine by me. I'll make an issue.

birdsarah:
@mattpap, the examples test fail are just arbitrary fails - the server examples are currently particularly flaky - we put in a commit a couple of days ago (https://github.com/bokeh/bokeh/commit/1a272fea9ddaff3d8f5cc2c1d6a65be2eb528e01) that may reduce the number of arbitrary fails. Try merging master to see if it improves. Otherwise, you just need to restart them manually.

bryevdv:
I've restarted the two tests. If they fail again, I will merge in master and we can see how that goes. 

@mattpap once tests are passing, how much more needs to go in this PR vs could be done in smaller follow-on PRs?

bryevdv:
@birdsarah it looks like master (and the hopeful server test fix) was already merged

bryevdv:
@mattpap the current failure on `button_server.py` is real, it is a result of the `button.type` -> `button.button_type` change

birdsarah:
@mattpap, I've run the button_server test locally and I get the error: 

    AttributeError: unexpected attribute 'type' to Button, possible attributes are button_type, callback, clicks, disabled, icon, label, name or tags

birdsarah:
@mattpap - I completely forgot, but what I should have done,and you can to, is look at the `examples.log`. The link is at the end of the test run (but not in the poor man logger). @bryevdv and I just discussed that we should print out the `examples.log` on failure to help with debugging.

bryevdv:
@mattpap if the change to `button_type` is made, then get JS console error:
```
Error: String property 'icon' given invalid value: Icon(681e9ed3-f0ba-4dc3-93bf-75a43f6cdba2)
```

bryevdv:
@mattpap, I would like *if it is possible and reasonable* (i.e. ready "enough") to merge this tomorrow, with #3955 and get a dev build out.

mattpap:
> Error: String property 'icon' given invalid value: (...)

This has been broken for awhile. `AbstractButton.icon`'s type was set to `String` (should be `Instance`).

mattpap:
> plotting/file/glyphs.py -> stalls my browser (prob also in master)

This issue is persistent on `master`. The reason for slowness is (at least) that `_doc_attached` is called multiple times per plot. The number of times varies between examples (usually between 5 and 12 (!)). This particular example (`glyphs.py`) is very slow because it embeds one plot per glyph.

mattpap:
NPM fails like crazy on travis ci today.

bryevdv:
There seems to be a recent somewhat persistent problem with phantomJS

bryevdv:
On phone, hit wrong button, apologies

mattpap:
@bryevdv, yeah, I was restarting tests multiple times to get tests pass for efd35bf.

bryevdv:
`_doc_attached` could be called multiple times for different objects. If it is repeatedly calling on the same object that's a bug

mattpap:
@bryevdv, in particular it's "canvas attached to document" (which is logged from `Canvas._doc_attached()`). Even in a trivial example (1 plot, 1 glyph renderer), it shows up in the debug log 5 times.

bryevdv:
That's unexpected

mattpap:
I'm pretty confident this PR is ready for merging at this point. I fixed all the pointed out issues that were introduced in this PR and the remainder is persistent in `master`.  The API needs further work, which I will continue in more fine-grained PRs. Specifically:
* better embedding story (currently there is `Bokeh.Plotting.show(), but it could be more flexible)
* `attrs` in `new model(attrs)` is currently untyped
* add support for `fig.glyph(field_name, ...)` where `field_name` is `string` (this is tricky to do right in typed setting; in Scala I could get this cheaply by using symbols (`'field_name` syntax), but that's not support in TS)
* make callbacks accept plain JS functions (we don't use the serialization protocol in the API, so we can pass whatever we want)
* expose methods on models (currently I exposed only attributes and just few methods)

Going further, we should have TS typings for views and other internal interfaces as well. This way we will be able to prove that, at least on interface level, our changes make sense. In this workflow we would start the work by changing typings first, making sure there are no errors, and then we would implement the logic in coffeescript. Until we would switch to TS full-scale (I doubt this will happen soon, because that needs an incredible amount of work and myself I'm not convinced to TS besides the static typing part, but I will leave this for a separate discussion).

birdsarah:
> @bryevdv, in particular it's "canvas attached to document" (which is logged from Canvas._doc_attached()). Even in a trivial example (1 plot, 1 glyph renderer), it shows up in the debug log 5 times.

I was battling with that in the layout example. I didn't find a way to reduce it. There was a point in the middle of a night when I understood why it was happening although it's escaping me now.

birdsarah:
Some small notes: 
1. What about having the new examples be under the examples directory in a bokehjs folder. It makes sense to me to have them where all the other examples are so that they are readily discoverable for users.
2. Can you add a README about how to run the examples. How do you get from a .ts file to a .js file.

Also, I might be missing something as it's a very large PR, but I can't see any documentation or tests.

almarklein:
> make callbacks accept plain JS functions (we don't use the serialization protocol in the API, so we can pass whatever we want)

I'm guessing we need this short term for the demo?

@birdsarah the new examples are build using `gulp examples`. We should document this with a readme. But I think this makes the root examples directory rather unsuited.

bryevdv:
@birdsarah I plan do docu the crap out of this, but it will have to be after this week at the earliest. But we need to get this into a dev build for some others to use ASAP.

bryevdv:
@mattpap @birdsarah please both add issues to cover any follow on work. Unless there is objection, I plan to merge this in the next hour or so.

bryevdv:
ping @sccolbert FYI

mattpap:
>  It makes sense to me to have them where all the other examples are (...)

That makes sense, but there are reasons to put it under `bokehjs/`. First, bokehjs is still kind of a separate library from bokeh. Second, and more importantly, it was just convenient to do it this way due to the existing build system. We obviously can do it even if examples are moved up, but that requires a little more thought that I could spent during this work.

mattpap:
> I'm guessing we need this short term for the demo?

Possibly yes, but my main concern for the demo is to make auto ranges and selection/inspection tools work without hacks (as in `stocks` example).

sccolbert:
looking...

bokehjs/examples/anscombe/anscombe.ts
```diff
@@ -0,0 +1,77 @@
+module Anscombe {
```
sccolbert: The recommendation is to use `namespace` here instead of `module`. The `module` keyword is deprecated:
http://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html

bokehjs/examples/anscombe/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "noImplicitAny": true,
+    "noEmitOnError": true,
+    "module": "commonjs",
+    "moduleResolution": "node",
```
sccolbert: Are we really using node module resolution for Bokeh? It looks like we just have a global "Bokeh" namespace, which is fine, but doesn't require this compiler flag.

sccolbert:
I briefly reviewed the TypeScript portion and made comments on some trivial issue I saw. I only commented a single instance of each issue, but they were repeated among all examples. Aside from that, :shipit:

bokehjs/examples/tap/tap.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const p = plt.figure({title: "Tappy Scatter", tools: tools})
+
+    const circles = p.circle({field: "x"}, {field: "y"}, {source: source, radius: radii,
+                             fill_color: colors, fill_alpha: 0.6, line_color: null})
+
+    p.text({field: "x"}, {field: "y"}, indices, {source: source, alpha: 0.5,
+           text_font_size: "5pt", text_baseline: "middle", text_align: "center"})
+
+    const tap = p.tools.filter((t) => t instanceof Bokeh.TapTool)[0] as Bokeh.TapTool
```
mattpap: @sccolbert, do you know how to get rid of `as Bokeh.TapTool` in this scenario? In Scala I would do `p.tools.collect { case t: Bokeh.TapTool => t }` and I would get a sequence of `Bokeh.TapTool`s. I presume I could achieve this by using a direct for loop and type condition there, but that isn't a big improvement over the existing code.

sccolbert: Not really, presumably `p.tools` is something like `Array<BaseTool>` and you are searching for a more derived type. The definition of the `Array.filter` method is this:

```typescript
    filter(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any): T[];
```

so there is no way to map to a more derived type.

That said, filtering an entire array to find the first matching element is not a very efficient way of doing things. You are traversing the entire array, plus allocating a new array, just to immediately discard all of that work.

You could roll your own `find` function which would look something like this:

```typescript
interface IClass<T> {
  new (...args: any[]): T;
}

function findFirstOfType<T, U extends T>(values: T[], kind: IClass<U>): U {
  for (let v of values) {
    if (v instanceof kind) {
      return v;
    }
  }
  return void 0;
}

findFirstOfType(p.tools, Bokeh.TapTool);
```

bryevdv:
@mattpap did you want to make any more small changes?

mattpap:
@bryevdv, not in this PR.

mattpap:
> Also, I might be missing something as it's a very large PR, but I can't see any documentation

Documentation is often overrated, mostly because keeping it in sync with the code is real pain. See [1] as nice example (we switched from AMD to commonjs at least a year ago). Thought some minimal explanation regarding how to get started with TypeScript and bokehjs would be helpful. Also, I'm not convinced the way things currently are will be the final approach.

>  or tests.

`bokehjs/examples` are tests for this PR. There could have been some unit tests added for setter/getter infrastructure, but that requires further work to move remaining property handling code to class-level, so I will be testing this when it's done.

[1] https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/dev_guide/notes.rst#bokehjs-amd-module-template-for-a-model

waseem18:
@mattpap I'm currently going through the code in this PR. I find it a bit difficult to understand few things.  As you've mentioned  `Thought some minimal explanation regarding how to get started with TypeScript and bokehjs would be helpful.` It would be really helpful if the code is documented.

PR #4039, opened 2016-03-17
Fix that a bar chart with one "item" does not show a label

### LINKED ISSUE #3661: Bar Plot plotting one value
Hi,

I have discovered following "Bug".
If I have an array looking like this:

```
types = ["car","car","car","car"]
```

and then trying to plot a barplot to look how many and which items in the array exist like this:

```
output_file("Types.html")
typePlot = Bar(types, title="types", tools=TOOLS, xlabel="Types", ylabel="Frequency")
tP = vplot(hplot(typePlot))
show(tP)
```

I get the plot which i expected but if there is only one type of value in the array it does not write the name of the type under the bar as it does with normally two or more different types of values in the barplot.
I hope you can fix this.

Regards,

Marko

almarklein:
Fixes #3661 

There was code to explicitly hide the x-axis, ticks and labels when there is but one item in the bar chart. But when the label is explicitly specified, it should be shown regardless.

### DIFF
--- a/bokeh/charts/builders/bar_builder.py
+++ b/bokeh/charts/builders/bar_builder.py
@@ -320,7 +320,7 @@ def Bar(data, label=None, values=None, color=None, stack=None, group=None, agg="
     chart = create_and_build(BarBuilder, data, **kw)
 
     # hide x labels if there is a single value, implying stacking only
-    if len(chart.x_range.factors) == 1:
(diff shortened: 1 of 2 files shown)


bryevdv:
Can we add a test case for this somehow?

almarklein:
@bryevdv test added

bryevdv:
Great, LGTM

PR #4044, opened 2016-03-19
Fix default bokeh dir on Windows

aspeskin:
The previous version attempted to create a top level directory on Windows.

This resolves: RuntimeError: could not create bokeh config directory at ...


aspeskin:
Getting ready for PyData at Strata and experienced this error when downloading sampledata on a corporate windows laptop.  @ahmadia helped with this fix.

bryevdv:
Nice catch, thanks @aspeskin

PR #4055, opened 2016-03-23
Working on new layout

birdsarah:
There's still a lot going wrong here, but it's moving, this is not expecting code review, mostly for @bryevdv and @havocp 

`box.coffee` is mostly taken from @havocp and I still don't feel like I have control over it.

Here's a video of where we're at. 

new_layout_scratch

Still a hot mess and the coloring is helping me to see what's going wrong.

Plots seem to be being drawn multiple times as you can see by the borders flickering darker.

Note: Title has been removed for now - obviously we need to add back in, but one thing at a time


birdsarah:
Here's another video where you'd be forgiven for thinking things work. @havocp, in this simple example it looks like they get ~60/40 of the available space - why don't the two plots get equal size - 50/50?

new_layout_scratch_3

birdsarah:
Check out this bit of awesome crazy when the resize tool is on:

new_layout_scratch_4

You can't see if great in the gif, but it seems like LOADS of copies of plot are being rendered at the same time.

birdsarah:
The lots of plots rendering causing the border to darken is present on master. I'd love some background on what causes that and how to stop it. I'm wondering whether there are stale versions of variables roaming around the solver.

birdsarah:
closing this. with a more direct copy and paste from havoc's original things are looking better.

Issue #4059, opened 2016-03-23
CSS in docs broken


birdsarah:
docs look like this:

<img width="950" alt="screen shot 2016-03-23 at 3 31 11 pm" src="https://cloud.githubusercontent.com/assets/1796208/14004572/531c0612-f118-11e5-8ee5-11bdc1c132c5.png">

from @bryevdv: FWIW bisection points to here: https://github.com/bokeh/bokeh/commit/564060ccab6b7b438c4c9c0cb9845a9a74351a48

PR #4060, opened 2016-03-23
Add the new bk-root wrapper

birdsarah:
Fixes: #4059


bryevdv:
ping @mattpap this is somewhat related to your other recent issue about preview tool I think

Should the `bk-root` class name be added back to the BB view directly? I don't quite understand why it was removed. Also ping @mistakevin

mattpap:
Most likely there is connection between those issues. Actually I had an issue with creating the API as well, because I have to maintain `bk-root` all the time. I will look in this.

mattpap:
As this PR is a hotfix, it's fine. We can revert this when proper solution is found.

bryevdv:
OK, will merge when tests pass.

PR #4064, opened 2016-03-24
Add stock example using ts api

almarklein:
Adds a stock example making use of the new ts api. Based against #4037

### DIFF
--- a/bokehjs/examples/stocks/stocks.ts
+++ b/bokehjs/examples/stocks/stocks.ts
@@ -1,3 +1,5 @@
+import plt = Bokeh.Plotting;
+
 console.log(`Bokeh ${Bokeh.version}`);
 Bokeh.set_log_level("debug");
 
@@ -5,15 +7,16 @@ Bokeh.set_log_level("debug");
 function make_plot(title:string, source:Bokeh.ColumnDataSource) {
     
     // Create plot
-    var xr = new Bokeh.DataRange1d({});
-    var yr = new Bokeh.DataRange1d({});
-    var plot = new Bokeh.Plot({        
+    const xr = new Bokeh.DataRange1d({});
+    const yr = new Bokeh.DataRange1d({});
+    const plot = new plt.Figure({
+        tools: "pan,wheel_zoom,save,reset",        
         x_range: xr,
         y_range: yr,         
         title: title,
         plot_width: 400,
         plot_height: 400,
-        border_fill_color: "#eeeeff",
+        background_fill_color: "#eeeeff",
     });
     
     // Add axis and grid
@@ -25,8 +28,8 @@ function make_plot(title:string, source:Bokeh.ColumnDataSource) {
     plot.add_layout(new Bokeh.Grid({ticker: yaxis.ticker, dimension: 1}));
     
     // Add a line for each entry in the source
-    var colors = ['#aa0000','#00aa00', '#0000aa', '#aaaa00', '#aa00aa', '#00aaaa'];
-    var i = -1;
+    const colors = ['#aa0000','#00aa00', '#0000aa', '#aaaa00', '#aa00aa', '#00aaaa'];
+    let i = -1;
     for (let key in source.data) {
         if (key != 't') {
             i += 1;
@@ -36,8 +39,8 @@ function make_plot(title:string, source:Bokeh.ColumnDataSource) {
         }
     }
     
-    // Add tools
-    plot.add_tools(new Bokeh.PanTool(), new Bokeh.WheelZoomTool());
+    // Add tools - not needed if we use Figure
+    //plot.add_tools(new Bokeh.PanTool(), new Bokeh.WheelZoomTool(), new Bokeh.ResetTool(), new Bokeh.PreviewSaveTool());
     
     // Tweaks to work around bootstrap problems 
     xr.renderers = plot.renderers;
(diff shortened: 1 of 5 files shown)


bokehjs/examples/stocks/stocks.ts
```diff
@@ -0,0 +1,73 @@
...
+            plot.add_glyph(line, source);
+        }
+    }
+    
+    // Add tools
+    plot.add_tools(new Bokeh.PanTool(), new Bokeh.WheelZoomTool());
+    
+    // Tweaks to work around bootstrap problems 
+    xr.renderers = plot.renderers;
+    yr.renderers = plot.renderers;
```
almarklein: @mattpap This was still needed, is that expected?

almarklein:
Not sure if we want this in master, but its more or less what we'd need in the phosphor+bokeh demo.

bokehjs/examples/stocks/stocks.ts
mattpap: That may be the case. There are some rough edges (minor differences between python and js).

bokehjs/examples/stocks/stocks.ts
```diff
@@ -0,0 +1,73 @@
...
+    source.stream(d, 50);
+}
+setInterval(new_data, period*1000);
+
+// Create plot and attach to DOM
+const plot = make_plot('Simple stocks demo', source);
+const doc = new Bokeh.Document();
+doc.add_root(plot);
+const div = document.getElementById("plot");
+Bokeh.embed.add_document_static(div, doc);
```
mattpap: You can now use `show(fig, target_element)`.

bokehjs/examples/stocks/stocks.ts
```diff
@@ -0,0 +1,73 @@
+console.log(`Bokeh ${Bokeh.version}`);
+Bokeh.set_log_level("debug");
+
+
+function make_plot(title:string, source:Bokeh.ColumnDataSource) {
+    
+    // Create plot
+    var xr = new Bokeh.DataRange1d({});
+    var yr = new Bokeh.DataRange1d({});
+    var plot = new Bokeh.Plot({
```
mattpap: I strongly suggest against `var`. Preferably use always `const` and `let` for truly modifiable (!= mutable, because an array can be declared `const` but you can still mutate its internal state) variables (e.g. `i`).

bokehjs/examples/stocks/stocks.ts
```diff
@@ -0,0 +1,73 @@
+console.log(`Bokeh ${Bokeh.version}`);
+Bokeh.set_log_level("debug");
+
+
+function make_plot(title:string, source:Bokeh.ColumnDataSource) {
```
mattpap: I was hope this annoying prefixing can be removed at some point. Thought I didn't find a way for TS to do this.

bokehjs/examples/stocks/stocks.ts
almarklein: `plt.show()` only works for Figure, not Plot. Intended?

almarklein:
@mattpap updated

almarklein:
BTW the tap example is broken for me.

bokehjs/examples/stocks/stocks.ts
```diff
@@ -0,0 +1,74 @@
...
+    plot.add_layout(new Bokeh.Grid({ticker: yaxis.ticker, dimension: 1}));
+    
+    // Add a line for each entry in the source
+    const colors = ['#aa0000','#00aa00', '#0000aa', '#aaaa00', '#aa00aa', '#00aaaa'];
+    let i = -1;
+    for (let key in source.data) {
+        if (key != 't') {
+            i += 1;
+            let line = new Bokeh.Line({x:{field:'t'}, y:{field:key}, legend: key, 
+                                       line_color: colors[i%6], line_width: 2});
```
almarklein: @mattpap would be nice to be able to `plot.line(...)` here, but as I understand it this API does not support setting line_color, etc.

mattpap: This is how plotting API works in python, thought we could make glyph functions work on `Plot` (they don't need anything specific from `Figure`).

almarklein: Sorry, the example uses a Figure now. I meant that I do not seem to be able to pass line_color et al. to the `line()` method.

mattpap: You can, see the `legends` example.  Do you get some errors from tsc?

bokehjs/examples/stocks/stocks.ts
mattpap: So this is why `tap` example fails. Initialization of range (and tools) need some work to make it robust.

mattpap:
> BTW the tap example is broken for me.

Yes, the same goes for (new example) `legends`. Initialization of ranges and tools must be improved.

mattpap:
As to `plt.show()`, it now works with `Component`, so `Plot` and layouts will work.

bokehjs/examples/stocks/stocks.ts
almarklein: mmm, I thought I tried it that way, but anyway, it works now. With `plot.line()`  the legend is shown, but with `new Bokeh.Line()` it is not. Is that expected?

almarklein:
> > BTW the tap example is broken for me.

> Yes, the same goes for (new example) `legends`. Initialization of ranges and tools must be improved.

Do you want (me) to fix up the examples, or do you plan to fix range-initialization?

Also, can we merge this into your branch?

bokehjs/examples/stocks/stocks.ts
mattpap: `legend` is only available on glyph functions (e.g. `plot.line()`). `Bokeh.Line` is a glyph constructor and it accepts only actual properties of a glyph. The are more cases like this. Further more, in the plotting API you can be more flexible about values than it's with models. E.g. figure.x_range = [0, 10], but `plot.x_range = new Bokeh.Range1d({start: 0, end: 10})`.

mattpap:
> Do you want (me) to fix up the examples

I fixed what I could. Comments in the other PR. Feel free to fix any remaining.

> you plan to fix range-initialization?

This is my immediate concern (as well as tool initialization, because selections don't work with the API).

Issue #2006, opened 2015-03-02
Unable to Stack Multiple Twin Axis on Same Side


amy-ding:
When I try to stack multiple axis on the same side, the image compresses, resulting in an x-axis with width 0.  

The below works fine, with a regular multi-line plot:
```
output_notebook()
plot = figure(title = 'test')
x = range(0, 10)
y = range(0, 10)
y2 = range(11, 20)
y3 = range(12, 20)
plot.line(x, y, legend = 'y', y_axis_label = 'y')
plot.extra_y_ranges = {}
plot.extra_y_ranges['y2'] = Range1d(start = 10, end = 20)
plot.extra_y_ranges['y3'] = Range1d(start = 10, end = 20)
plot.add_layout(LinearAxis(y_range_name = 'y2', axis_label = 'y2'), 'right')
plot.add_layout(LinearAxis(y_range_name = 'y3', axis_label = 'y3'), 'left')

plot.line(x, y2, legend = 'y2', color = 'red', y_axis_label = 'y2', y_range_name = 'y2')
plot.line(x, y3, legend = 'y3', color = 'green', y_axis_label = 'y3', y_range_name = 'y3')

plot.extra_y_ranges
```

However, when I change the plot.add_layout() to both be on the right side, you get an image like this:

![twin_axis](https://cloud.githubusercontent.com/assets/11274379/6444635/e16a2ae8-c0cc-11e4-949d-3aa9fafd2c41.png)

I'm using 0.7.1, but I don't think there were significant changes wrt the axis plotting, in 0.8. This is in follow-up to https://groups.google.com/a/continuum.io/d/msg/bokeh/S45Yy5VNczE/cpNe1jge_I8J

bryevdv:
When I try this on current master it works as expected: 
![screen shot 2015-03-02 at 11 53 31 am](https://cloud.githubusercontent.com/assets/1078448/6446776/d73600b4-c0d2-11e4-8e65-85bfddc53cca.png)
Can you provide more information about your system, versions, etc? Also if possible can you try installing the dev build from last Friday?

silverorange05:
Right, it works when you specify one to be on the left, and one to be on the right.  Try substituting in this:

plot.add_layout(LinearAxis(y_range_name = 'y2', axis_label = 'y2'), 'right')
plot.add_layout(LinearAxis(y_range_name = 'y3', axis_label = 'y3'), 'right')

[...]

PR #4067, opened 2016-03-25
WIP: More layout work

birdsarah:
@bryevdv @havocp 

(probably not going to have time to work on this till after strata)

Made some progress on the layout. Still to do:

- [ ] add titles back - I think we should do this properly with title being added via add_layout - it will be less hacky and easier to reason about constraints.
- [ ] toolbar (sigh - this could be hard)
- [ ] shouldn't go off the window size but the outer div size so
- [ ] does it work with embed, components?
- [ ] does it work with server?
- [ ] does it work in notebook?
- [ ] tests (there are currently none)
- [ ] ability to turn it off if you want a fixed size!
- [x] implement so widgets line up nicely too
- [ ] there's no customizations - apart from having a fixed box, what else do we want?
- [ ] what happens when Rows are used with components?
- [ ] implement layouts_legacy
- [ ] twin axes (left and right: #2006)

Optional:
- it would be nice to specify %ages e.g. `Column(children=[row1, row2])` default assumes you want  an equal split of space (so 50/50 in the case of two rows). But maybe you don't: you could do `Column(children=[(row1, 33%), (row2, 66%)])`. This is marked as optional as it should be straight forward to work around this with a custom template.

Questions:
- [ ] ~~@havocp - Box.Model (coffee/layouts/box.coffee) is an exact copy/paste from your layout-demo branch except that I canned all the whitespace stuff: it didn't seem to be doing anything and I was trying to get a smaller surface area to reason about - what am I missing - what should it be doing?~~
- [ ] @bryevdv - how do you feel about layouts_legacy for all the old layouts - similar to the chart archiving we did, and then the new layouts are just in layouts?

gif


bryevdv:
@birdsarah should this PR be closed now?

bryevdv:
is/will be supersede by other work

PR #4072, opened 2016-03-27
Hotfix: print out examples log on fail

birdsarah:
WIP


birdsarah:
Note to self - need to take into account slave nodes I think

damianavila:
200 files? I guess you need to merge master in :wink:

Issue #4070, opened 2016-03-25
Plotting with subsets (row-wise) of ColumnDataSources (like filtered data)


clairetang6:
Currently, Bokeh only allows glyphs to take full columns of data from a ColumnDataSource. This means that you must have additional ColumnDataSources for any plots that only use a subset of the data. This causes data to be replicated and makes linked selection more complicated for users, who would have to keep track of indices themselves. 

It would be nice if Bokeh allowed glyphs to take subsets of columns as data. An application using multiple plots, each showing some subset (e.g. filtered by the value of some column, like the example @dennisobrien showed at the SF meetup) of the data (or the full set), would then be able to completely share the data in the same way that Bokeh allows plots to share the full ColumnDataSource now. Linked selection would work intuitively with selected data points (rows) showing up in whichever plots contain that point. Holding shift before selecting on any plot would add to the selection. 

One possibility for an API would be to include another parameter that picks the subset of data for a renderer to use by the indices (which could default to all). 

``` python
data_source = ColumnDataSource(df)
fig = figure()
r = fig.circle(x = ‘x’, y = ‘y’, source=data_source, indices=[0, 1, 2]) 
r = fig.circle(x = ‘x’, y = ‘y’, source=data_source, indices=np.arange(n)[df.weather == “sunny”])
r = fig.circle(x = ‘x’, y = ‘y’, source=data_source, indices=[i for i, val in enumerate(df.weather) if val == “sunny”])
```

Another possibility could be to make the data source “filterable” in a pandas-like way. 

``` python
r = fig.circle(x = ‘x’, y = ‘y’, source=data_source[ data_source.weather == “sunny” ]) 
```

I tried to look through the issues and google group to find some related posts. 
https://github.com/bokeh/bokeh/issues/2323
https://github.com/bokeh/bokeh/issues/2710
https://groups.google.com/a/continuum.io/forum/#!searchin/bokeh/data$20source%7Csort:date/bokeh/_DIOQGlC6ds/EbZL0KRfBwAJ
https://groups.google.com/a/continuum.io/forum/#!searchin/bokeh/data$20source%7Csort:date/bokeh/79eSMaDV39I/OTfZqZHIug0J

Is this something that would be in Bokeh’s scope?

clairetang6:
[...]

PR #4074, opened 2016-03-28
ColumnDataSource subsets

clairetang6:
issues: #4117

[Plotting with subsets (row-wise) of ColumnDataSources (like filtered data)](https://github.com/bokeh/bokeh/issues/4070)

This is definitely a work-in-progress, but I got some basic linked selection working. The approach that I took was to create a new model, ColumnData (which is basically what the ColumnDataSource is now), which holds the data and column names. The ColumnData can then be associated with multiple ColumnDataSources. Then I made ColumnDataSource aware of which indices of the data it would use so that you can use subsets of data held in the ColumnData. 

Here's an example of what it looks like:
cds_subsets

```python
from bokeh.plotting import Figure, output_file, show
from bokeh.models.sources import ColumnDataSource, ColumnData
from bokeh.io import gridplot

output_file("cds_subsets.html", mode="relative-dev", root_dir="/users/claire/documents/bokeh/bokehjs/build/")

data = {'x': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'y': [10, 5, 2, 6, 1, 4, 8, 2, 7, 3]}

#create one ColumnData object to be associated with three different subsets
column_data = ColumnData(data)
data1 = ColumnDataSource(column_data=column_data)
data2 = ColumnDataSource(column_data=column_data, indices=[i for i, val in enumerate(data['y']) if val < 8 ])
data3 = ColumnDataSource(column_data=column_data, indices=[i for i, val in enumerate(data['y']) if val > 2 ])

#creating a ColumnDataSource directly from the data defaults to include all the points
data4 = ColumnDataSource(data)

plot_size_and_tools = {'plot_height': 200, 'plot_width': 200, 
                        'tools':['lasso_select', 'box_select', 'reset'],
                        'x_range':[0, 11], 'y_range':[0, 11]}

fig1 = Figure(title="Full dataset", **plot_size_and_tools)
fig1.circle(x='x', y='y', source=data1, color='black')

fig2 = Figure(title="Subset 1", **plot_size_and_tools)
fig2.circle(x='x', y='y', source=data2, color='green')

fig3 = Figure(title="Subset 2", **plot_size_and_tools)
fig3.circle(x='x', y='y', source=data3, color='red')

fig4 = Figure(title="CDS not linked", **plot_size_and_tools)
fig4.circle(x='x', y='y', source=data4, color='black')

show(gridplot([[fig1, fig2, fig3, fig4]]))
```

This is still missing a lot, but I think it could be a really useful feature! 

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Hi @clairetang6 thanks for the PR I definitely think this is a good feature, but I also think it merits some discussion about implementation. I am traveling tomorrow but I will try to write something more substantial in the next few days. A few (very brief) comments for now:
* I don't think we can't break `.data` or any other API on CDS, unfortunately. They are heavily used and frequently demonstrated and would cause much outcry if they were to stop working. We might work around this with `@property` getter that delegates to `column_data`, e.g.,  or we could switch to inheritance. Possibly other ideas as well.

* I think the best conceptual framework might be a `filter` property, which could be `None` (no filtering), a `Seq(Int)` which explicitly lists the subset indices, *or* a new type which is a function that returns a `Seq(Int)` Then people could supply arbitrary JS snippets to filter with, or we could provide some "canned" common types of filter functions. I think this opens the door to the "pandas like syntax" you mention in the related issue, we can make `__getitem__` on CDS that returns an appropriately constructed filter function.

clairetang6:
Hi @bryevdv, as I was looking through the code and thinking about how to get this working, I did think that it would be good to keep the CDS working as is. I've never written code that a lot of other people use, but it does make sense how important that is! I like the idea of allowing the filter to be a function too. Also, there's no rush for writing up more feedback. Thanks for the comments so far.

bryevdv:
Moved milestone and tags to associated issue.

bryevdv:
Hi @clairetang6 I'd like to go ahead and close this PR if that is OK? I definitely want to pursue this feature but along the "filter" lines described above. Also there is already a conflict now with master, and we still have to get through a couple of other disruptive tasks and features yet.

clairetang6:
Hi @bryevdv, I have been merging in master and fixing the conflicts, but haven't changed any of the implementation. I just pushed those changes up so you could see them, but I'm fine with closing this too. 

I think one of the main issues to think about for the implementation is how to separate out the filter from the data. That way, when you have multiple plots showing different views of the same data, you don't need to duplicate the data. Instead you would just have different filter properties. To keep the original .data attribute on the CDS working might be a little bit tricky. Getting data would be easy if you could just get it from another object, but setting data might be more complicated because it might not be clear if you are setting the data for the subset or setting the data for the whole dataset and how that affects the other plots that are using the same data source. 

I'd still be interesting in working on this whether in this PR or another one!

bryevdv:
@clairetang6 That's great to here! I think you raise some good points there, and actually to me that points to the need to really discuss this feature more up-front. It involves some pretty entrenched API that we have to be careful about not breaking, and there are probably also lots of use-cases that should be considered (some possibly to be rejected but those should be explicit decisions). I'll go ahead and close this particular PR, would you want to create a working document on the Wiki to start collecting use-cases and requirements? This ideally could turn into a BEP

clairetang6:
Hi @bryevdv, I created a document on the wiki for this feature. https://github.com/bokeh/bokeh/wiki/Filterable-Data-Source

After some more consideration about not breaking the CDS, I thought maybe we could just add a new type of data source and keep the CDS as is.

ericdfournier:
Just to comment on the value of this proposal/functionality to the community. I am currently working on an interactive visualization which is attempting to automatically link the selection or points within a scatter plot to the selection of patches in an adjacent map pane (and visa versa). Based upon a lot of trial and error, it seems as if the proposal submitted by @clairetang6 would provide the requisite functionality with minimal need for breaking changes to the rest of the API. (Thanks to both for all of your work on this project).

bryevdv:
There is an open PR scheduled for `0.12.6` https://github.com/bokeh/bokeh/pull/5982

ericdfournier:
Cheers!

PR #4078, opened 2016-03-29
Remove support for backbone collections

mattpap:
This is based on top of `ts_api` branch. When #4037 is merged there should be one commit left with actual related changes. Backbone collections weren't used for a long time, so just removing the dead weight here.


mattpap:
btw. @bryevdv, what's the purpose of `prefix` in `base.Config`? I didn't find any application for this across the repository.

bryevdv:
I am almost certain it had to do with the old flask server

bryevdv:
Also this is awesome, thanks.

mattpap:
@bokeh/dev, @birdsarah, can some one help explaining this: `ValueError: signal only works in main thread`?

bryevdv:
I only have vague knowledge, I thought that test code was changed not to use signal or something, but maybe not?

mattpap:
Reducing the number of threads in the `-m examples` case "fixed" the problem. I don't know why this only fails in py3 and that it shows now (I didn't encounter this issue before).

mattpap:
@bokeh/dev, please review this.

canavandl:
LGTM - it makes js test writing/reading a lot nicer.

bryevdv:
OK, merging this, just FYI for @birdsarah about the change of numthreads to 1 for now.

birdsarah:
> @bokeh/dev, @birdsarah, can some one help explaining this: ValueError: signal only works in main thread?

> OK, merging this, just FYI for @birdsarah about the change of numthreads to 1 for now. 

I'm a  bit bummed about this - changing to n=1 has slowed down all tests for everyone and I'm not sure actually fixed anything.

@mattpap, I'm sorry I missed this question at the time. The reason for the `ValueError` seems to be that it is that it's xdist (which parallelizes the tests) failing un-gracefully. When I was working on the testing re-org I tried to improve this to no avail. The `ValueError` only occurs when the tests fail. You can look at the examples.log and the examples.html to see why your test actually failed.

I tried quickly the other day to make the output of examples.log more available: https://github.com/bokeh/bokeh/pull/4073 but failed. 

My recollection is that it is just a coincidence that all the examples tests started passing when you switched to n=1. Although I could be wrong. @bryevdv if you're still restarting examples test runs then switching to n=1 hasn't changed anything except slowing down the test run.

bryevdv:
I am restarting tests constantly for a variety of reasons right now (mostly the PhantomJS thing but others too, not sure the breakdown)

PR #4080, opened 2016-03-29
Don't attach document multiple times when traversing the graph

mattpap:
This fixes the problem with `_doc_attach()` being called multiple times, specifically `Canvas._doc_attached()`, which caused substantial performance degradation.

### DIFF
--- a/bokehjs/src/coffee/core/has_props.coffee
+++ b/bokehjs/src/coffee/core/has_props.coffee
@@ -373,23 +373,19 @@ class HasProps extends Backbone.Model
         throw new Error("models must be owned by only a single document")
       else
         @document._notify_attach(@)
-        return
+    else
+      @document = doc
+      @document._notify_attach(@)
 
-    first_attach = @document == null
-    @document = doc
+      for ref in @_immediate_references()
+        ref.attach_document(@document)
 
-    if doc != null
-      doc._notify_attach(@)
-      if first_attach
-        for c in @_immediate_references()
-          c.attach_document(doc)
+      # TODO (bev) is there are way to get rid of this?
+      for name, prop of @properties
+        prop.update()
 
-    # TODO (bev) is there are way to get rid of this?
-    for name, prop of @properties
-      prop.update()
-
-    if @_doc_attached?
-      @_doc_attached()
+      if @_doc_attached?
+        @_doc_attached()
 
   detach_document: () ->
     if @document != null
--- a/bokehjs/src/coffee/core/has_props.coffee
+++ b/bokehjs/src/coffee/core/has_props.coffee
@@ -368,10 +368,16 @@ class HasProps extends Backbone.Model
     _.values(result)
 
   attach_document: (doc) ->
-    if @document != null and @document != doc
-      throw new Error("Models must be owned by only a single document")
+    if @document != null
+      if @document != doc
+        throw new Error("models must be owned by only a single document")
+      else
+        @document._notify_attach(@)
+        return
+
     first_attach = @document == null
     @document = doc
+
     if doc != null
       doc._notify_attach(@)
       if first_attach


mattpap:
The remainder of the logic in `HasProps.attach_document()` needs some more though, because, I think, at least `first_attach` is unnecessary.

almarklein:
So `_doc_attach()` was called too often because the error was raised at times where it should not?r

mattpap:
No. The reason was that this code has to do with object graphs and it was attaching the document to an object multiple times during traversal, causing a computational blow up. Consider a `Plot`. It has property `renderers`, which can contain instances of `GuideRenderer` model, which has back reference to the plot. The original code didn't allow for infinite recursion to happen, but it did allow calling `_doc_attach()` multiple times.

bryevdv:
@almarklein its the `return` that was added that is the fix. LGTM

mattpap:
As we are questioning the logic of this function, @bryevdv, what's the purpose of this:
```
# TODO (bev) is there are way to get rid of this?
for name, prop of @properties
  prop.update()
```
code?

mattpap:
> Can't the first if block be removed entirely (...)

I don't think so, because you have to detach document first, before you can attach a new one.

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -373,23 +373,19 @@ class HasProps extends Backbone.Model
         throw new Error("models must be owned by only a single document")
       else
         @document._notify_attach(@)
```
bryevdv: My issue/question is about this line. If `@document == doc` haven't we already attached? Why are we calling `_notify_attach` again?

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -368,22 +368,24 @@ class HasProps extends Backbone.Model
...
+
+      for ref in @_immediate_references()
+        ref.attach_document(@document)
 
-    # TODO (bev) is there are way to get rid of this?
-    for name, prop of @properties
-      prop.update()
+      # TODO (bev) is there are way to get rid of this?
+      for name, prop of @properties
+        prop.update()
```
bryevdv: IIRC this was needed to ensure that updates from applying theme values actually caused changes to happen.

bryevdv: Added here: https://github.com/bokeh/bokeh/commit/f004054b757c9f84969e2ec353d39b1986626e12

bokehjs/src/coffee/core/has_props.coffee
mattpap: @bryevdv, `_notify_attach()` increments a reference counter, so it has to be called on every attach, even if document is attached to a given object. Given the current design, this is necessary, because otherwise when detaching a document, models would be detached too early (e.g. back references from tools or guides to plot). Originally I just added a return statement, but then tests failed and had to add this line.

mattpap: Saying this, I think, we may need to revisit this logic in general in future, but for now this PR fixes the performance issue and appears not to break bokehjs.

bokehjs/src/coffee/core/has_props.coffee
mattpap: @bryevdv, I see. This explanation raises my attention, so I will look into this.

bokehjs/src/coffee/core/has_props.coffee
bryevdv: Ah I see, ok, thanks for the explanation.

mattpap:
@bryevdv, should I presume this is LGTM?

bryevdv:
Yes LGTM

Issue #4082, opened 2016-03-30
Expose Geometries to TapTool Callbacks


brendancol:
TapTool currently does not expose the geometry of the click event to the callback.

brendancol:
merged

PR #4083, opened 2016-03-30
exposed geometries as part of TapTool callback data

brendancol:
This PR adds a `cb_data` object to the TapTool `callback.execute` call which contains the geometries from `plot_model.get('tool_events')`

https://github.com/bokeh/bokeh/issues/4082


bryevdv:
LGTM

canavandl:
LGTM

PR #4084, opened 2016-03-30
Add Ellipse glyph

### LINKED ISSUE #3848: Oval with height=width does not produce a circle
Hi,
I'm using bokeh 0.11.0 in a jupyter notebook and when I try to draw an oval with height and width equal, I don't get a circle.

```python
output_notebook()

p.x_range = Range1d(0, 2)
p.y_range = Range1d(0, 2)

p.oval(x=1.0, y=1.0, width=1.0, height=1.0)
show(p)
```

humm:
- [x] fixes #3848
- [x] added test in `tests/glyphs/Ellipse.py`
- [x] .rst files were edited to add the new documentation (and to edit the Oval one)

The Ellipse glyph is implemented with the CanvasRenderingContext2D.ellipse()
function, which is currently marked as experimental, and unsupported on some
mobile platform. A more robust implementation would need to use Bezier curves
until more platforms support the function.

### DIFF
--- a/sphinx/source/docs/reference/bokehjs.rst
+++ b/sphinx/source/docs/reference/bokehjs.rst
@@ -291,6 +291,29 @@ together with a crossbar (+) at the given coordinates.
     :alt: Click to view JSFiddle
     :target: http://jsfiddle.net/bokeh/tx3Z4/
 
+.. _bokehjs_ellipse:
+
+``ellipse``
+~~~~~~~~
+
+The ellipse glyph displays ellipses centered on the given coordinates with the
+given dimensions and angle.
+
+* ``x``, ``y`` - center point coordinates
+* ``width``
+* ``height``
+* ``angle``
+
+  * default: 0
+
+* :ref:`bokehjs_line_properties`
+* :ref:`bokehjs_fill_properties`
+
+.. image:: ../../_images/bokehjs_glyphs/ellipse.png
+    :height: 300px
+    :alt: Click to view JSFiddle
+    :target: http://jsfiddle.net/bokeh/h6eXz/
+
 .. _bokehjs_image:
 
 ``image``
@@ -707,4 +730,3 @@ the given coordinates.
     :height: 300px
     :alt: Click to view JSFiddle
     :target: http://jsfiddle.net/bokeh/3hXa8/
-
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -594,8 +594,6 @@ class Oval(Glyph):
     The %s values for the ovals.
     """)
 
-
-
 class Patch(Glyph):
     """ Render a single patch.
 
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -186,6 +186,43 @@ class Bezier(Glyph):
     The %s values for the Bézier curves.
     """)
 
+class Ellipse(Glyph):
+    u""" Render ellipses. """
+
+    __example__ = "tests/glyphs/Ellipse.py"
+
+    # a canonical order for positional args that can be used for any
+    # functions derived from this class
+    _args = ('x', 'y', 'width', 'height', 'angle')
+
+    x = NumberSpec(help="""
+    The x-coordinates of the centers of the ellipses.
+    """)
+
+    y = NumberSpec(help="""
+    The y-coordinates of the centers of the ellipses.
+    """)
+
+    width = DistanceSpec(help="""
+    The widths of each ellipse.
+    """)
+
+    height = DistanceSpec(help="""
+    The heights of each ellipse.
+    """)
+
+    angle = AngleSpec(default=0.0, help="""
+    The angle the ellipses are rotated from horizontal. [rad]
+    """)
+
+    line_props = Include(LineProps, use_prefix=False, help="""
+    The %s values for the ovals.
+    """)
+
+    fill_props = Include(FillProps, use_prefix=False, help="""
+    The %s values for the ovals.
+    """)
+
 class Gear(Glyph):
     """ Render gears.
 
@@ -519,7 +556,8 @@ class Oval(Glyph):
 
     .. note::
         This glyph renders ovals using Bézier curves, which are similar,
-        but not identical to ellipses.
+        but not identical to ellipses. In particular, widths equal to heights
+        will not render circles. Use the ``Ellipse`` glyph for that.
     """
 
     __example__ = "tests/glyphs/Oval.py"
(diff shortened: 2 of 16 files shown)


humm:
Someone will need to adjust `sphinx/source/docs/reference/bokehjs.rst`, and in particular the target for the jsfiddle. I don't know if I can, should and how to do that.

humm:
And do tell me if you prefer using Bezier curves rather than `ctx.ellipse()` to support more and older platforms.

bryevdv:
AFAIK it's impossible to get true ellipses with Beziers, and that's what `Oval` already uses, I don't think that is advisable. Do you have information about what platforms do not support `ctx.ellipse` offhand? 

I would also not worry about the current JS docs, they are all about to get a complete overhaul and the old docs there are out of date.

bryevdv:
@mattpap any comments from you welcome, especially re: new TS parts. This LGTM at a quick glance, I will try to try it out tomorrow. Probably worth noting that if the aspect ratio is not 1-1 then ellipses with equal width/height will not look circular, either.

humm:
For the support of `ctx.ellipse`, I consulted this page: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/ellipse. On the desktop, only Firefox is missing, but support will come with 48, which is a few months away (August 2016). On the mobile side, support is more spotty, but with Safari and Chrome, 70% of the users are supported.

bryevdv:
Oh that's interesting, I did not realize it was not supported across the board already. @bokeh/dev thoughts on this?

mattpap:
>  I did not realize it was not supported across the board already.

Not having full support at least on desktop is a problem, I think. Our goal is to relieve the user from all the cross-browser compatibility issues (unless he wants to use the JS/TS API), so this is an incompatible change. However, if accompanied with a big fat warning, maybe we could proceed.

mattpap:
> especially re: new TS parts.

All looks good.

humm:
As a stop-gap measure, I could implement ellipse drawing in javascript without using `ctx.ellipse` (for instance, this: https://jsfiddle.net/fcyb/whvnamvh/ : the real ellipse is in red, 10 pixels right and below the approximation.), and choose the drawing code depending on browser support.

Instead of using `lineTo`, I could also do a good-enough approximation with Bezier curves, employing 8 controls points. 

This adds code, that could be eventually phased out when browser support is good enough.

bryevdv:
That fiddle is not rendering for me. I guess in light of the state of support I would be OK with an approx, but in this case I think it should be a "fixup" to implement `ctx.ellipse` if it is missing, like these:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/util/canvas.coffee

humm:
Ok, I'll look at adding a fixup for `ctx.ellipse`.

bryevdv:
I guess "polyfill" is the fancy jargon JS devs like to use. :)

bryevdv:
@humm Just FYI this branch now has conflicts with master, and will need master merged in and conflicts resolved. Do you still intend to pursue this work?

humm:
@bryevdv Yes. I have been busy in the last few days, but I will have time to write the "polyfill" Friday, and I will rebase the PR on the current master at this time.

bryevdv:
@humm great! not intending to hurry FYI just have to check in on status from time to time :)

humm:
@bryevdv No problem. Well, probably won't be able to finish today, but the JS cubic bezier curves equivalent of `ctx.ellipse` is done: https://jsfiddle.net/fcyb/yb4nzu25/

sphinx/source/docs/reference/bokehjs.rst
```diff
@@ -291,6 +291,29 @@ together with a crossbar (+) at the given coordinates.
...
+  * default: 0
+
+* :ref:`bokehjs_line_properties`
+* :ref:`bokehjs_fill_properties`
+
+.. image:: ../../_images/bokehjs_glyphs/ellipse.png
+    :height: 300px
+    :alt: Click to view JSFiddle
+    :target: http://jsfiddle.net/bokeh/h6eXz/
+
```
bryevdv: I would like to commend you on your thoroughness! This file is somewhat stale but the plan is to automate a good portion of the JS docs around properties, etc. just like with python version.

humm:
Ok, so the polyfill is now implemented, and everything is properly rebased. I tested Ellipses under Chrome and Firefox, and they render the same, upon a purely visual inspection. It will be properly worth it to discuss the need for `Oval` in the future, and whether it should be deprecated once `Ellipse` is available.

Also, this is unrelated, but the `angle` argument produces unintuitive results. Rotating an rect or ellipse by 90 degree does not follow the aspect ratio. In the following image, the rectangles and ellipses have a height and width of 2 and 40 respectively, but they do not appear to be 40 units tall, and 2 units wide, or 2 units tall, and 40 units wide. Is this expected? Did I miss an explanation somewhere?
madejkaajkaajkidnbp4f3cdqm5h 83uaaaaasuvork5cyii

produced with:

```
import math
from bokeh.plotting import figure, output_server, show

output_server()
plot = figure(width=300, height=300)
plot.rect(x=[10, 20, 30], y=[1, 2, 3], width=40, height=2, angle=math.pi/2,
             color="#386CB0", fill_color=None, line_width=1)
plot.ellipse(x=[10, 20, 30], y=[1, 2, 3], width=40, height=2, angle=math.pi/2,
             color="#386CB0", fill_color=None, line_width=1)

show(plot)
```

sphinx/source/docs/reference/bokehjs.rst
humm: Thank you. Of course, the fiddle link will need to be updated (and the fiddle created), it points to the oval one right now.

bryevdv:
@humm I expect this has to do with the same aspect issue that has always plagued circles, etc if the data space aspect is not 1-to-1. The code measures what 2 and 40 mean in x any y dimensions respectively, and these are different because of the non-uniform aspect. Then things are rotated, with the effective outcome that the scale measurement of the x dimension is what is actually used for the y dimension, and vice versa.

bryevdv:
BTW is this still `WIP` or should be marked `ready`?

sphinx/source/docs/reference/bokehjs.rst
bryevdv: All the fiddles are going away too. Codepen has a POST api so we can actually check the examples in to be maintained in the repo/docs, make links that load them, instead of manually updating things like jsfiddle requires. I hope to get to it in the next few weeks.

humm:
This is ready for me.

bryevdv:
Thanks @humm !

PR #4085, opened 2016-03-31
More TypeScript API

### LINKED ISSUE #4086: DataTable `getFormatter not a function` bug
Getting an error when using DataTable:

**Reproduction:**
`python examples/models/data_tables.html`

**Error:**
`Uncaught TypeError: getFormatter(...) is not a function`
`line 1476: slick.grid.js`

Looks like this commit is where it broke:

```
commit 90c6e1bfa5fc962fa151298b6cfc32d8eeca4375
Merge: fa1a222 5ea2fb7
Author: Mateusz Paprocki <mattpap@gmail.com>
Date:   Tue Mar 29 23:55:04 2016 +0200

    Merge pull request #4037 from bokeh/mattpap/ts_api

    Implementation of TypeScript API
```

mattpap:
This is continuation of PR #4037.

fixes #4086


mattpap:
@almarklein, auto ranges now work as expected.

bokehjs/examples/stocks/stocks.ts
```diff
@@ -59,6 +51,8 @@ module Stocks {
     const period = 0.5;  // seconds
     function new_data() {
         t += period;
+        // XXX: explicit type annotation required due to TypeScript#6041.
+        // Remove when TypeScript 2.0 is available.
         const d: Bokeh.Data = {
```
mattpap: @almarklein, so I see where the initial confusion about strict static typing came from. As indicated in the comment, this type annotation is required due to a bug in TypeScript. When TypeScript 2.0 ships, user won't be concerned of typing details in this case, but will still get an error if he uses the wrong type.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -352,11 +352,16 @@ class PlotView extends Renderer.View
     @update_range(null)
 
   build_levels: () ->
+    renderer_models = @mget("renderers")
+    for tool_model in @mget("tools")
+      synthetic = tool_model.get("synthetic_renderers") ? []
+      renderer_models = renderer_models.concat(synthetic)
```
mattpap: @bryevdv, this needs your attention. My goal is (at some point) disallow setting `plot_model.renderers` and separate renderers added by the user from "synthetic" renderers added by other models (e.g. `HoverTool` with its `ttmodels`). This is a proof of concept and I only configured `HoverTool` to work like this, but it seems to work smoothly. If we don't proceed this way, we won't be able to fill in missing dependencies between properties of different models (e.g. `HoverTool` depends on `Plot.renderers`, but sets `Plot.renderes` when configuring `ttmodels`, thus giving an infinite loop). The alternative is to use `silent: true`, but I doubt this will be reliable.

bokehjs/examples/hover/hover.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const circles = p.circle({field: "x"}, {field: "y"}, {source: source, radius: radii,
+                             fill_color: colors, fill_alpha: 0.6, line_color: null})
+
+    p.text({field: "x"}, {field: "y"}, indices, {source: source, alpha: 0.5,
+           text_font_size: "5pt", text_baseline: "middle", text_align: "center"})
+
+    const hover = p.select_one(Bokeh.HoverTool)
+    hover.tooltips = (source, info) => {
+        const ds = source as Bokeh.ColumnDataSource
```
mattpap: This will be a common problem. Although the API says we use generic `DataSource`, we depend on `data` attribute, so pretty much we expect a `ColumnDataSource`. Do we actually support or plan to support non-columnar data sources? Just to note, `RemoteSource` and thus `AjaxDataSource` are column data sources. Maybe we should squash together `DataSource` and `ColumnDataSource`?

almarklein: How does stuff like patches work, do they fit in a ColumnDataSource?  I see that GeoJSONDataSource is a ColumnDataSource, so I'm guessing they do? Do we ever plan to put images in a DataSource?

almarklein:
@mattpap how can I get a ref to the view of a plot. I can make a plot appear using `Plotting.show()`, but I also need the view so we can do resizing etc. (The old `make_model()` does not work anymore, because Plot is then missing a document)

mattpap:
@almarklein, use `Bokeh.index`. It contains references to top-level views (those directly embedded into the DOM). However, if you have a layout and then plots as children then you will get just reference to the layout and you will have to explore layout's view to get the plots. Definitively I can make `plot.show()` return the constructed top-level view. As to subviews, I need to give this a thought.

bokehjs/examples/hover/hover.ts
mattpap: > How does stuff like patches work, do they fit in a ColumnDataSource?

Yes, just instead of storing arrays of values, we store arrays of arrays (2D object) of numbers values. Similarly, in bi-categorical plots you have to figure out how to encode labels and values as columns in a data source, instead of just providing a dictionary of key (tuple), value pairs.

almarklein:
@mattpap The view seems to be added later to `Bokeh.index`. It's not there right after `Plotting.show()`. Also `Plot.id` is not defined in `.d.ts` in current master.

>  I can make `plot.show()` return the constructed top-level view

+1

bokehjs/src/coffee/models/plots/plot.coffee
bryevdv: I'll comment more on this tomorrow.

bryevdv: This seems OK, I am not too fond of calling them `synthetic_renderers`. They are just normal renderers that happen to get contributed in a different place (i.e., from a tool).

bokehjs/examples/hover/hover.ts
bryevdv: There will almost definitely be other kinds of `DataSource` in particular I imagine a data source for graph/network data. It might also make sense to make column data source really just be 1D columns, and move arrays to an `ArrayDataSource` or something.

bryevdv:
@mattpap I fixed the small conflict in tap tool. Apart from naming comment on "synthetic" renderers, LGTM

bokehjs/src/coffee/models/plots/plot.coffee
mattpap: @bryevdv, so what naming convention would you suggest? I just picked the first word that came to my mind.

mattpap:
> I fixed the small conflict in tap tool.

This was a recent addition or did I miss something earlier?

mattpap:
I will continue with more TS API additions in another PR.

bokehjs/src/coffee/models/plots/plot.coffee
bryevdv: Is there a reason tools can't also have a `renderers` property? 

```
additional = tool_model.get("renderers") ? []
renderer_models = renderer_models.concat(additional)
```

bokehjs/examples/hover/hover.ts
mattpap: So we will have to give this some thought in future, because currently bokehjs is very tightly bound to column data sources. E.g. we use integer index a lot, but that may not be idiomatic/sufficient with other types of data sources. I supposed it would be `tuple(int, int)` in array case and I'm not sure what in graph case. Thought I may be overthinking this.

bryevdv:
> This was a recent addition or did I miss something earlier? 

Yes brendan's tap tool geometry PR merged today

bokehjs/src/coffee/models/plots/plot.coffee
mattpap: I will be doing further work in this area, so we don't have to decide now (this is internal logic anyway).

mattpap: Some tools already have `renderers` property for a different purpose. So perhaps `computed_renderers`, but that's used as well (at least since this PR). I was considering just adding some property with those renderers to tools, but for some reason (which I can't recall right now) I thought it will be better to set them on a plot. I will give this more thought this week.

mattpap:
Tests failed due to #4099. Restarted.

bokehjs/src/coffee/models/plots/plot.coffee
bryevdv: Ah that's a good point. `sub_renderers`? It's also an internal detail, we shouldn't let it hold up this PR

bryevdv:
OK, LGTM to merge whenever

PR #4091, opened 2016-03-31
Point to the correct location when you look for the noarch package

### LINKED ISSUE #4090: Build not succeeding because path problem looking for the noarch package
We just need to change some paths to look correctly.

damianavila:
- [X] issues: fixes #4090


damianavila:
OK, merging now to see how it breaks :wink:

Issue #359, opened 2014-02-16
remove setuptools


hhuuggoo:
setuptools is wierd

we should replace setup.py develop with something like this:

```
from __future__ import print_function
import sys
if len(sys.argv)>1 and sys.argv[1] == 'develop':
   from os.path import dirname, abspath, join
   site_packages = site.getsitepackages()[0]
   fname = join(site_packages, "bokeh.pth")
   path = abspath(dirname(__file__))
   with open(fname, "w+") as f:
       f.write(path)
   print("develop mode, wrote path (%s) to (%s)" % (path, fname))
   sys.exit()
```

PR #4093, opened 2016-04-02
Improve docs for python -m bokeh command

### LINKED ISSUE #4092: Unclear how to run bokeh command when exe not available
*Original title: bokeh serve: entry point with `if __name__ = '__main__':`*

I work with a standard python installation (not anaconda) on Windows 10, bokeh is installed with pip.

The standard way to start the bokeh server with 'bokeh serve' is not working:
The script 'serve' is not executed. The line 

```python
#!C:\Python34\python.exe
```

is not recognized. Windows asks, which app should be used to open the file...

The following modification of the file 'serve' worked for me:

```python
#!C:\Python34\python.exe
if __name__ = '__main__:
    import sys
    from bokeh.command.bootstrap import mainmain(sys.argv)_
```

On windows the script can now be invoked with:

```python
python -m bokeh serve
```

On Linux the standard way and the way described above should both work (though I can not test).

Additonally the server can be started from any python script, it is no longer necessary to start the server manually from a terminal:

```python
import subprocess
import time    # optionally, might be useful to give the server time to initialize
args = ['python', '-m', 'bokeh', 'serve']
p = subprocess.Popen(args) 

# optional delay for x seconds to give the server time to initialize
time.sleep(1)
```

I suggest 

1. to change the file 'serve' and use 'if __name__ ...' as shown above. 

2. to add the additional way to start the server with 'python -m bokeh serve' to the documentation.

almarklein:
Issues: fixes #4092 

The ability to use `python -m bokeh` was added when we implemented `__main__.py` sometime ago. It is mentioned in the docs [here](http://bokeh.pydata.org/en/latest/docs/user_guide/server.html) but there are two places where it makes sense mentioning it too (which is what this PR does).

### DIFF
--- a/sphinx/source/docs/dev_guide/setup.rst
+++ b/sphinx/source/docs/dev_guide/setup.rst
@@ -251,13 +251,13 @@ If you build Bokeh on a Windows machine in a Conda environment with either
 ``setup.py install`` or ``setup.py develop``, running ``bokeh serve`` will
 not work correctly. The .exe will not be available within the Conda
 environment, which means you will use the version available in the base
-install, if it is available. Instead, you can make sure you use the version
-within the environment by explicitly running the ``bokeh serve`` python script
-in the root of the bokeh repository, similar to the following example:
+install, if it is available. Instead, you can make sure you use the Python
+version within the environment by making use of Python's ``-m`` flag,
+as in the following example:
 
 .. code-block:: sh
 
-    python bokeh serve path\to\<yourapp>.py
+    python -m bokeh serve path\to\<yourapp>.py
 
 Developing Examples
 -------------------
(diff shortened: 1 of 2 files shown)


bryevdv:
Is there nothing we can do to actually *improve* things? Is it impossible to install a new `.exe` in the right environment? A comment like this buried in the docs is not actually going to stop many people from trying to just run it as a normal command and then complaining about problems.

almarklein:
The exe is installed when using anaconda, but not anymore with pip, since we got rid of setuptools. Other examples are when installing with `develop`. 

Having `try: import setuptools\nexcept ImportError: pass` will make the `entry_points` work again when setuptools is available (i.e. anyone using pip). We did have something like this, I cannot remember exactly why we got rid of it.

almarklein:
Mmm, it seems we got rid of setuptools in #359 and #365, which is ages ago. I now see that we also set "scripts" in addition to "entry_points", but indeed, these don't work as well on Windows, because they do not come with an exe. My proposal to try-import setuptools still stands.

bryevdv:
@almarklein please make a separate issue to discuss improvements. Will go ahead and merge this PR now as-is

PR #4094, opened 2016-04-02
Allow getting ref to view again in line click/select event

### LINKED ISSUE #4069: Line picking is broken because glyph.glyph_view is no more
Failing example `plotting/file/line_select.py` -> picking fails, and console reports `TypeError: d0.glyph.glyph_view is undefined`. This is code used in a client project.

The `glyph_view` attribute was removed in https://github.com/bokeh/bokeh/commit/f004054b757c9f84969e2ec353d39b1986626e12, because it breaks the model-view separation. We should either put it back, or find a better way to achieve line picking.

almarklein:
fixes #4069

This PR proposes a fix to #4069 by adding a `get_view()` function to the "event object" when a line is clicked or selected. By allowing the user to get a ref to the view via the event object, we avoid having to put a reference of the view on the model. Could not give the event object a direct reference to the view, because that results in a recursion error in `HasProps._value_record_references()`. Thus the workaround using a function.

### DIFF
--- a/bokehjs/src/coffee/common/hittest.coffee
+++ b/bokehjs/src/coffee/common/hittest.coffee
@@ -17,11 +17,11 @@ point_in_poly = (x, y, px, py) ->
 
 create_hit_test_result = ->
   result = {
+    # the glyph that was picked
+    get_view: () -> null,  # function, to avoid infinite recursion
+    get_model: () -> null,  # also a function, for consistency
     # 0d is only valid for line and patch glyphs
     '0d': {
-      # the glyph that was picked
-      glyph: null,
-      get_view: () -> null,  # this is a function, because setting the view causes inf. recursion
       # array with the [smallest] index of the segment of the line that was hit
       indices: []
     }
--- a/bokehjs/src/coffee/common/hittest.coffee
+++ b/bokehjs/src/coffee/common/hittest.coffee
@@ -21,6 +21,7 @@ create_hit_test_result = ->
     '0d': {
       # the glyph that was picked
       glyph: null,
+      get_view: () -> null,  # this is a function, because setting the view causes inf. recursion
       # array with the [smallest] index of the segment of the line that was hit
       indices: []
     }
--- a/bokehjs/src/coffee/common/selector.coffee
+++ b/bokehjs/src/coffee/common/selector.coffee
@@ -10,8 +10,9 @@ class Selector extends HasProps
     @set('timestamp', new Date(), {silent: silent})
     @set('final', final, {silent: silent})
     if append
-      indices['0d'].indices =  _.union(@get('indices')['0d'].indices, indices['0d'].indices)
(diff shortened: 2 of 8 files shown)


bokehjs/src/coffee/common/hittest.coffee
```diff
@@ -21,6 +21,7 @@ create_hit_test_result = ->
     '0d': {
       # the glyph that was picked
       glyph: null,
+      get_view: () -> null,  # this is a function, because setting the view causes inf. recursion
```
bryevdv: Why on `0d` and not one level up?

bokehjs/src/coffee/common/selector.coffee
```diff
@@ -10,8 +10,10 @@ class Selector extends HasProps
     @set('timestamp', new Date(), {silent: silent})
     @set('final', final, {silent: silent})
     if append
+      if @get('indices').get_model()
+        indices.get_model = @get('indices').get_model
+        indices.get_view = @get('indices').get_view
       indices['0d'].indices =  _.union(@get('indices')['0d'].indices, indices['0d'].indices)
-      indices['0d'].glyph =  @get('indices')['0d'].glyph or indices['0d'].glyph
       indices['1d'].indices =  _.union(@get('indices')['1d'].indices, indices['1d'].indices)
       indices['2d'].indices =  _.union(@get('indices')['2d'].indices, indices['2d'].indices)
```
almarklein: @bryevdv what actually happens here? Does this combine the indices from multiple "selections" on the same glyph, or of multiple glyphs? If the former, my earlier comment about selecting multiple lines can be ignored.

almarklein:
> Why on 0d and not one level up?

Because that's where we put it earlier. But you are right, putting it one level up makes more sense. I also made it `.get_model()` instead of `.glyph` to make it more consistent.

But this "event system" is in rather a messy state. The `get_model()` and `get_view()` are only supported for lines and I don't feel great about them being functions ...

bokehjs/src/coffee/common/selector.coffee
bryevdv: I spoke in haste. This combines selections on a single *data source*, which might be driving several glyphs. The dimensionality reflects this: selection on "scatters" are just a 1d list of indices. What about selections on multilines? Well, you have a 1d list of indices that could represent that the k'th line has been "hit" *at all*, but you might want more information, i.e. which of the individual points of the k'th were specifically hit? "0d" is for a single line, where you just want to know if the line has been hit at all, this tells you that, and then "glyph" will tell you which glyph was actually the one the mouse hit... honestly all this should be re-thought, but data sources should be simplified at the same time, so not sure when that can happen.

almarklein:
Superseded by #4098

bryevdv:
@almarklein reminder to set status on closed PRs

PR #4095, opened 2016-04-03
add canonical link to sphinx template

### LINKED ISSUE #3217: Build documentation with canonical links to "latest"
For example, the first search result here: https://www.google.com/search?q=bokeh+model currently points to http://bokeh.pydata.org/en/0.9.3/docs/reference/models.html instead of the latest version of the documentation.

If we inserted a canonical links to http://bokeh.pydata.org/en/latest/docs/reference/models.html in all versions of the documentation, Google would point to http://bokeh.pydata.org/en/latest/docs/reference/models.html whenever it would normally return a reference to the specific version of the documentation.  

This is extremely helpful if our URL structure is not changing but the interfaces/names of our functions/variables are.

bryevdv:
issues: fixes #3217

### DIFF
--- a/sphinx/source/bokeh_theme/layout.html
+++ b/sphinx/source/bokeh_theme/layout.html
@@ -32,6 +32,10 @@
   –––––––––––––––––––––––––––––––––––––––––––––––––– -->
   <link rel="icon" type="image/png" href="{{ pathto('_static/' + theme_favicon, 1) }}">
 
+  <!-- Canonical
+  –––––––––––––––––––––––––––––––––––––––––––––––––– -->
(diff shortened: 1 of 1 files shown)


bryevdv:
Yields, for example:

screen shot 2016-04-03 at 1 09 40 am

bryevdv:
ping @birdsarah @bokeh/dev for comments. As far as I understand this is the correct usage of `canonical`

bryevdv:
OK I am going to merge this and we will see. It's possible to get the full benefit we would also need to re-deploy old versions of the docs with this change.

Issue #3514, opened 2016-01-04
rename bokeh.client.connection 


bryevdv:
Suggest `bokeh.client._connection` as it is intended to be private.

waseem18:
Hello @bryevdv 

I would like to work on this task.

PR #4097, opened 2016-04-04
Renamed connection.py to _connection.py  Issue #3514

waseem18:
This PR solves the task in the issue #3514 .

1. Renamed `connection.py` to `_connection.py`.
2. `connection.py` was being imported in `session.py`. Changed `from .connection import ClientConnection` to `from ._connection import ClientConnection`.


bryevdv:
Hrm, there's another place where it is imported:
```
bryan@0199-bryanv (git:master) ~/work/bokeh/bokeh $ grip "from \.connection"
./client/session.py:
   10 : from .connection import ClientConnection
./server/tornado.py:
   24 : from .connection import ServerConnection
```
I'd actually like to understand why we aren't seeing a test failure right now.

PR #4098, opened 2016-04-04
Allow getting ref to view again in line click/select event

### LINKED ISSUE #4069: Line picking is broken because glyph.glyph_view is no more

almarklein:
Fixes #4069. Supersedes #4094. Related to an earlier PR #3191.

This PR proposes a fix to #4069 by adding a get_view() function to the "event object" when a line is clicked or selected. By allowing the user to get a ref to the view via the event object, we avoid having to put a reference of the view on the model. Could not give the event object a direct reference to the view, because that results in a recursion error in HasProps._value_record_references(). Thus the workaround using a function.

This whole selection system needs thought. This is just a quick fix to get line selection working again (we use it in a client project) until its solved in a proper way.


bryevdv:
test failures appear to be real failures

almarklein:
@bryevdv tests pass, ready from my end.

bryevdv:
Sure LGTM I thikn we will have to revisit all of this soon tho in a more comprehensive way.

Issue #4104, opened 2016-04-04
Crossfilter Example: type is not a function error


brendancol:
When running crossfilter example patched for TypeScript API, I get the following error:

`Bokeh: Failed to repull session TypeError: type is not a function`

bryevdv:
This was resolved.

PR #4105, opened 2016-04-04
Crossfilter example fixes

brendancol:
This PR addresses some bugs in the crossfilter bokeh server example introduced by changes in bokeh core:

TypeScript API updates:
- [x] https://github.com/bokeh/bokeh/issues/4103

Cannot repull session: type is not a function error:
- [x] https://github.com/bokeh/bokeh/issues/4104

- [ ] JS Error on after invoking server callbacks:
`Bokeh: Error handling message: TypeError: Cannot read property 'solver' of null, [object MessageEvent]`
This happen after changing a dropdown and recreating the Figure object

- [ ] move models and coffee to top-level directory instead of `models`


examples/app/crossfilter/models/styleable_box.coffee
```diff
@@ -10,6 +11,7 @@ class StyleableBoxView extends BokehView
     class: "bk-vbox"
 
   initialize: (options) ->
+    debugger
```
canavandl: remaining `debugger` statement

examples/app/crossfilter/models/styleable_box.coffee
```diff
@@ -39,15 +39,10 @@ class StyleableBox extends BaseBox.Model
...
-      children: []
-      css_properties: {}
-      orientation: 'vertical'
-    }
-
-  children: () ->
-    return @get('children')
+  @define {
+    css_properties: [p.Dict, null]
+    orientation:    [p.String, 'vertical']
```
mattpap: You could use `p.Orientation` enum property. Unless this orientation has more choices.

brendancol: Ah nice, yes I'll switch an use that...

mattpap:
> Cannot repull session: type is not a function error:

Error reporting has to be really improved here. I took me awhile to figure out which `type` this refers to (https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/has_props.coffee#L89).

bryevdv:
removed milestone, those are for issues. restarted tests which seem to be spurious failures. @brendancol is there more work to do?

bryevdv:
@brendancol the problem is something with `StyleableBox`. If I change that to `HBox` then things "render" (very badly but everything is there). The issues is that the class definition is not succeeding somehow, to the `type` that BokehJS tries to instantiate to create one of them is `undefined`.

Should also add with the `HBox` to get it to "render" then it is actually back to the state that I saw: first render ok, no updates (console error about solver)

bryevdv:
This is the problem:
```
css_properties: [p.Dict,        null      ]
```
There is no `p.Dict`. that is the `type` that is `undefined`

To help find this, I added this conditional debugger:
```
    for name, [type, default_value] of props
      if not type?
        debugger
      @properties[name] = new type({obj: @, attr: name, default_value: default_value})
```

bryevdv:
OK so now it has the state I saw before. First render is ok, widgets fail to update, with console error:
```
Bokeh: Error handling message: TypeError: null is not an object (evaluating 'this.document.solver'), [object MessageEvent]
```
@birdsarah does this ring any bells?

mattpap:
> There is no p.Dict. that is the type that is undefined

I'm not sure what going on in this PR, but I already fixed this in c45139b.

bryevdv:
I'm not sure either, I definitely saw your change before. As I said it's just now back to what I originally saw ~2 weeks ago.

bryevdv:
bad merge from master maybe? In any case @mattpap is there anything better to do that to add a `try`/`catch` around that chunk of code to log better information? I'm not sure a bad property type can be detected any earlier than there.

mattpap:
I added a meaningful error message in #4141 for this.

bryevdv:
I'm also slightly worried about the `from .helpers import load_component`. A change recently added the application directory to `sys.path` so some imports like `import foo` from the top level work. But I am not sure how well this intersects with packages in subdirs, and in particular I think the behavior across py2 vs p3 might be different. I want to demonstrate practices that are always know to work, so for now we might want to move all *python* to just the top level (with no `__init__.py`)

birdsarah:
> ```Bokeh: Error handling message: TypeError: null is not an object (evaluating 'this.document.solver'), [object MessageEvent]```

> @birdsarah does this ring any bells?

I discussed this a little with @brendancol - it's probably to do with how the pieces are initialized and the document hasn't been added. That's my guess.

bryevdv:
@birdsarah @brendancol the fixes that are hear should go in, regardless. I propose we merge this as-is, and re-evalulate if there is any additional work after the layout work lands. Thought?

mattpap:
> Bokeh: Error handling message: TypeError: null is not an object (evaluating 'this.document.solver')

I fixed a few of those recently (in some PR, don't remember), so it may be fixed by now.

bryevdv:
OK, merging this now, we will certainly visit any other issues before release.

PR #4106, opened 2016-04-04
Canavandl/rebased label annotation

### LINKED ISSUE #3822: Improved Annotations and Legends

canavandl:
supercedes #3956 (in which I added a bunch of unrelated LaTeX stuff)

- [x] issues: fixes #3822 
- [x] tests added / passed
- [ ] ~~~release document entry (if new feature or API change)~~~


canavandl:
Example:

```python
from bokeh.models import ColumnDataSource, Range1d, Label

output_file("label.html", title="label.py example")

source = ColumnDataSource(data=dict(height=[66, 71, 72, 68, 58, 62],
                                    weight=[165, 189, 220, 141, 260, 174],
                                    names=['Mark', 'Amir', 'Matt', 'Greg',
                                           'Owen', 'Juan']))

p = figure(title='Dist. of 10th Grade Students at Lee High',
           x_range=Range1d(140, 275))
p.scatter(x='weight', y='height', size=8, source=source)
p.xaxis[0].axis_label = 'Weight (lbs)'
p.yaxis[0].axis_label = 'Height (in)'

label = Label(x='weight', y='height', text='names', level='glyph',
              x_offset=5, y_offset=-5, source=source)
p.add_annotation(label)

show(p)
```

screen shot 2016-04-04 at 1 48 53 pm

canavandl:
ping @bokeh/dev for review

bokeh/models/plots.py
```diff
@@ -239,6 +240,31 @@ def add_glyph(self, source_or_glyph, glyph=None, **kw):
...
+
+        Args:
+            annotation (Annotation) : instance of annotation to add to plot
+        Returns:
+            Annotation
+
+        '''
+        if not isinstance(annotation, Annotation):
+            raise ValueError("'annotation' argument to add_annotation must be \
+                             Annotation subclass")
```
bryevdv: personally would prefer single lines even if longer

bokeh/models/plots.py
```diff
@@ -239,6 +240,31 @@ def add_glyph(self, source_or_glyph, glyph=None, **kw):
...
+
+        if annotation.plot is not None:
+            raise ValueError("annotation %s to be added already has 'plot' \
+                             attribute set" % annotation)
```
bryevdv: ditto

bokeh/models/plots.py
```diff
@@ -239,6 +240,31 @@ def add_glyph(self, source_or_glyph, glyph=None, **kw):
...
+
+        if hasattr(annotation, 'source') and annotation.source is None:
+            annotation.source = ColumnDataSource()
```
bryevdv: Does this need to be handled here? We can add a validation to check for missing source/columns if field names are given, and then just send the `None` on to BokehJS deal with. Alternatively we can make the default for `source` on this model be a lambda that returns a new CDS

bryevdv: Thinking a bit more, the lambda func default that returns a blank CDS (on `Label`) is probably the simplest way to get rid of this test.

bryevdv:
A few small comments otherwise looks good.

bokehjs/src/coffee/models/annotations/label.coffee
```diff
@@ -0,0 +1,184 @@
...
+    if ctx.textBaseline == 'top'
+      y_shift = 0.0
+    if ctx.textBaseline == 'middle'
+      y_shift = -0.5 * height
+    if ctx.textBaseline == 'bottom'
+      y_shift = -1.0 * height
+    if ctx.textBaseline == 'alphabetic'
+      y_shift = -0.8 * height
+    if ctx.textBaseline == 'hanging'
+      y_shift = -0.17 * height
```
birdsarah: Where do these numbers come from - I understand `middle` being 0.5, but why 0.17 or 0.8?

birdsarah: Perhaps you can add a comment that points to a reference.

birdsarah:
Two things:
1) Can we have an example that shows off css mode and canvas mode - suggest just adding some labels to existing exampels
2) As there are no tests for the BokehJS side, can I suggest, at least a selenium screenshot test - I can help you write it.

bokeh/models/annotations.py
```diff
@@ -183,6 +186,100 @@ class BoxAnnotation(Annotation):
...
+    Specifies whether the text is rendered as a canvas element or as an
+    css element overlaid on the canvas. The default mode is "canvas".
+
+    .. warning::
+        Not all visual styling properties are supported if the render_mode is
+        set to "css". The border_line_dash property isn't fully supported and
+        border_line_dash_offset isn't supported at all. Setting text_alpha will
+        modify the opacity of the entire background box and border in addition
+        to the text. Finally, clipping Label annotations inside of the plot
+        area isn't supported in "css" mode.
```
birdsarah: It may also be worth noting that css labels will not be present in the output from the "save" tool. I wouldn't put this is in the warning as I'd argue it's a feature not a bug.

birdsarah:
The label doesn't appear to update when the server updates the data source. The label appears, but doesn't move as the x and y update

screen shot 2016-04-04 at 9 45 59 pm

birdsarah:
Really excited to see this coming along - I intend to use it for layout work.

FWIW I think it's ok to merge without server working.

bokehjs/src/coffee/models/annotations/label.coffee
canavandl: I'll add a link to the w3c spec around textBaseline. The actual baseline are font-specific, so these are guessestimates.

bokeh/models/annotations.py
canavandl: added

canavandl:
I forgot to add a change listener on the data source. Works w/ bokeh serve now:

screen shot 2016-04-05 at 12 00 24 pm

ping @birdsarah do you want me to include the fourier_animated.py patch?

birdsarah:
whoop whoop - no i don't think we should change the fourier_animated example - it was just a test.

canavandl:
Ok - I've added a selenium test and a css-label example (retrospectively, I should have put the citation in the off-axis whitespace):

screen shot 2016-04-06 at 10 41 29 pm

Ready for review/merge

bokeh/models/tests/test_annotations.py
```diff
@@ -122,6 +135,43 @@ def test_BoxAnnotation():
...
+    assert label.x_offset == 0
+    assert label.y_offset == 0
+    assert label.render_mode == 'canvas'
+    assert label.x_range_name == 'default'
+    assert label.y_range_name == 'default'
+    assert isinstance(label.source, ColumnDataSource)
+    assert label.source.data == {}
+    yield check_text, label
+    yield check_background, label, None, 1.0
+    yield check_border, label, None, 1.0, 0.0
```
bryevdv: There's an actual unit test failure: https://travis-ci.org/bokeh/bokeh/jobs/121332793#L2231-L2231

I think the `0.0` here at the end is the problem

canavandl: ah - shoot. yes, it is

bryevdv:
There's an integration test (taptool generates alert) that is suddenly failing on multiple PRs

PR #4107, opened 2016-04-04
Add deprecation for {AbstractButton,ButtonGroup}.{->button_}type

### LINKED ISSUE #4101: Head breaks apps that use widgets with type parameter
After the Typescript merge any application or bokeh code that uses a widget with the **type** parameter, such as a Button, will error with an AttributeError.

Recommend: Use a deprecation warning instead or "magically" convert it behind the scenes when running so it does not error.

To Recreate: Run any app or Bokeh code with a widget that contains the type parameter that run under 0.11.1. Alternatively, run the following Bokeh code (taken from Bokeh docs) against head:

```
from bokeh.io import output_file, show, vform

output_file("button.html")

button = Button(label="Foo", type="success")

show(vform(button))
```

mattpap:
fixes #4101

### DIFF
--- a/bokeh/models/widgets/buttons.py
+++ b/bokeh/models/widgets/buttons.py
@@ -3,7 +3,9 @@
 """
 from __future__ import absolute_import
 
-from ...core.properties import abstract
+import warnings
+
+from ...core.properties import abstract, HasProps
 from ...core.properties import Bool, Int, String, Enum, Instance, List, Tuple, Override
 from ...core.enums import ButtonType
 from ..callbacks import Callback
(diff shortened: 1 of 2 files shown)


bryevdv:
What about `button.kind` instead of `button.button_type`?

I don't have strong feelings about this, except that this is the only opportunity to change this either way.

mattpap:
I was about to ask if we are OK with the new name. One thing to consider is that `button_type` (of type btw. `ButtonType`) is used also in clickable groups, so `kind` (and even `type` before) may be insufficient to determine by the user what this property refers to.

bryevdv:
That's a good point, I think in this one case the redundancy is not unreasonable.

mattpap:
I was thinking about `theme`, but that would be confusing with actual themes.

mattpap:
Or maybe `style`, but we may want to reserve this one for theming as well.

hsparra:
Regarding `button.kind`, I actually like it if different widgets use the same property name if it is doing something similar. Like `type`, I know I am choosing the type of the specific widget. If there were different types of text fields, for example, then having one property name, such as `kind` is often easier for me than having to know, `button_type`, `text_type`, `super_cool_widget_type`, etc. Just my 2 cents. Now if the parameter/property is doing very different things then I would have different names.

bryevdv:
As I said I don't have super strong feelings, it would be good to get this merged soon, as we need to make a dev build with the latest TS API work.

mattpap:
The original issue is fixed. We can always revisit the naming convention before next release.

bryevdv:
Agreed, merging now.

PR #4109, opened 2016-04-05
Fix function naming in embed.coffee

mattpap:
This makes server work again. Why this wasn't caught earlier? Because server code in bokehjs catches all errors, even if they are unrelated to user input. Logging programming errors isn't the right approach. Wrong code should fail as soon as possible. This also raises a question if our testing framework could do a better job in such situation.

### DIFF
--- a/bokehjs/examples/anscombe/anscombe.ts
+++ b/bokehjs/examples/anscombe/anscombe.ts
@@ -73,5 +73,5 @@ namespace Anscombe {
     doc.add_root(grid);
 
     const div = document.getElementById("plot");
-    Bokeh.embed.add_document_static(div, doc);
+    Bokeh.embed.add_document_standalone(doc, div);
 }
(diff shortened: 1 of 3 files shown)


bryevdv:
> This makes server work again 

I run server examples off master fairly often, can you point to when it was broken for my benefit?

mattpap:
@bryevdv, https://github.com/bokeh/bokeh/pull/4085/commits/24c723d50047f255768874f8cb387809cadd829b, i.e. last my commit in #4085.

mattpap:
> This also raises a question if our testing framework could do a better job in such situation.

I think that it would be an improvement if examples' tests scanned through JS console logs and searched for common error indicators, e.g. `/error/i`, `undefined is not a function`, `TypeError`, etc. Any thoughts?

bryevdv:
I think that's a great idea

bryevdv:
OK, will go ahead and merge this one. I will make an issue with the testing idea

PR #4120, opened 2016-04-05
Fix resizing of embedded plots

### LINKED ISSUE #4110: Plot resizing is broken
In particular the example `embed_responsive_width_height.py`. This also breaks our phosphor+Bokeh demo.
In v0.11.1, the DOM looks like:
```
<div class='bk-plot'>  <-- this is what @el refers to
  <div class='bk-plot-wrapper'>
       <table> ... etc.
```

In master, the DOM looks like:
```
<div class='bk-root'>
    <div>  <-- this is what @el refers to
        <div class='bk-plot-wrapper'>
             <table> ... etc.
```

To resize, the plot queries the `clientWidth`and `clientHeight` of the parent node of its element. But this now fails, because of the extra layer.

As for a solution, I tried two things that both seem to work:
* Use the client size of `@el.parentNode.parentNode`. Or maybe safer: traverse upwards until meeting a node that is `bk-root`.
* Use client size of `@el.parentNode`, but set the CSS `height`of the `bk-root` to 100%.

Tasks:
* [x] @birdsarah @mattpap is that `<div>` intended? --> Yes
* [ ] Maybe not crucial for this issue, but somewhat related: there are still references to bk-plot in the code, was that supposed to be removed, or should the classless div above be a `bk-plot`
* [x] Fix the issue.
* [x] Selenium test to avoid having another regression like this (working on it)
* [x] Shortcut to allow (Phosphor) to set the width height to specific values, to avoid a DOM measurement.
* [x] Typing for `plot.resize_width_height()`

almarklein:
fixes #4110


bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -556,19 +556,32 @@ class PlotView extends Renderer.View
...
-    avail_height = @.el.parentNode.clientHeight - 50  # -50 for x ticks
+    # Check that what we found is a bk-root. If not, this is probably a subplot, which we 
+    # can not currently make responsive in a good way
+    if not node.classList.contains('bk-root')
+       Bokeh.logger.warn('subplots cannot be responsive')
+       return
+
+    # Get the available width
+    avail_width = node.parentNode.clientWidth
+    avail_height = node.parentNode.clientHeight - 50  # -50 for x ticks
```
mattpap: @almarklein, unrelated, but what this "-50 for x ticks" stand for?

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -556,19 +556,32 @@ class PlotView extends Renderer.View
...
     # if people set a large border on their plots, for example.
 
-    # We need the parent node, because the el node itself collapses to zero
-    # height. It might not be available though, if the initial resize
+    # Try to find bk-root. We will query the parent of that node for its size
+    node = @.el
+    for i in [0..2]  # Use for-loop; if we ever change DOM structure, it should still work
+      if node is null or node.classList.contains('bk-root')
+         break
+      node = node.parentNode
```
mattpap: Maybe we should just stick to jquery, e.g. `@$el.parents(".bk-root")[0]`?

bokehjs/src/coffee/models/plots/plot.coffee
mattpap: Or, where this magic constant comes from?

almarklein: Need some extra room for the ticks on the x-axis.

almarklein: Is there a better way to get how much space we need for that?

bokehjs/src/coffee/models/plots/plot.coffee
almarklein: That makes sense. I am not very familiar with jquery so I tend to write more vanilla js. What does that do exactly? I deliberately only search 3 levels up, so that subplots are easily ruled out.

almarklein:
> unrelated, but what this "-50 for x ticks" stand for?

Responding again, because that diff was outdated: that's extra room for ticks on the x-axis. A bit arbitrary, do you know of a better way to get how much space we need for that?

bryevdv:
(0n phone sorry to be brief) axes models have computed properties that return gown much space they need.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -548,29 +548,49 @@ class PlotView extends Renderer.View
...
+    # can not currently make responsive in a good way
+    if not node.classList.contains('bk-root')
+       Bokeh.logger.warn('subplots cannot be responsive')
+       return
+    
+    # Apply size, we need some extra space at the bottom. The sides are fine.
+    bottom_offset = if @model.below[0] then @model.below[0]._last_size else null
+    bottom_offset = Math.round(bottom_offset or 30) + 10
+    @_resize_width_height(use_width, use_height, maintain_ar,
+                          node.parentNode.clientWidth, node.parentNode.clientHeight - bottom_offset)
```
almarklein: @bryevdv does this seem reasonable? I could not find a public property on the axis, but maybe I missed it?

bryevdv: Computed property, not one of the "declared" Python/bokehjs ones. You'll have to look at the functions on the model and see what computed properties they correspond to

bryevdv:
@almarklein sorry I was mistaken, no computed propertied anymore, just the "private" methods that implement the measurements:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/axes/axis.coffee#L400-L401

there's also `@_size_constraint.value()` which should have the requested padding, but note if an axis requests less padding than `min_padding` then the border takes `min_padding` it's probably better to ask the plot itself what it's border "gutter" region sizes are. pinging @birdsarah since this is involved with her layout work probably.

almarklein:
@bryevdv Wait, this does not make any sense. Ticks and labels are drawn inside the canvas. The extra space is needed, but its to account for padding around canvas and for the toolbar. Made changes accordingly.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -548,29 +548,56 @@ class PlotView extends Renderer.View
...
       setTimeout( (=> this.resize_width_height(use_width, use_height, maintain_ar)), 2**@_re_resized)
       @_re_resized += 1
       return
 
-    avail_width = @.el.clientWidth
-    avail_height = @.el.parentNode.clientHeight - 50  # -50 for x ticks
+    # Check that what we found is a bk-root. If not, this is probably a subplot, which we 
+    # can not currently make responsive in a good way
+    if not node.classList.contains('bk-root')
+       Bokeh.logger.warn('subplots cannot be responsive')
```
mattpap: You can't use `Bokeh` global in library code. There should be global `logger` in this module already.

bokehjs/src/coffee/api/typings/models/plots.d.ts
```diff
@@ -3,6 +3,11 @@ declare namespace Bokeh {
         plot?: Plot;
     }
 
+    export interface PlotView extends View<Plot> {
+        model: Plot;
+        resize_width_height(use_width:boolean, use_height:boolean, maintain_ar?:boolean, width?:number, height?:number): void;
+    }
+
```
almarklein: @mattpap is this correct? Gulp does not error on it, but I am not very familiar with the working of these files.

tests/integration/interaction/test_responsive.py
```diff
@@ -156,3 +162,132 @@ def test_responsive_plot_starts_at_correct_size(output_file_url, selenium):
     # Canvas width should be just under 1000
     assert canvas.size['width'] > 900
     assert canvas.size['width'] < 1000
+
+
+def test_responsive_resizes_width_and_height(output_file_url, selenium):
```
almarklein: @birdsarah Thanks for this mini "framework", it was real easy to add tests for it using your examples!

bokehjs/src/coffee/api/typings/models/plots.d.ts
mattpap: `model: Plot` is unnecessary, it's inherited from `View` (`mode: ModelType`). Somehow strange that TS compiler doesn't mention this. Also, please use ` ` after `:`.

mattpap:
I've been experimenting with this PR and, at least in FF, I get all sorts of behavior related to scroll bars. Sometimes there is one, the other two and never one. Especially horizontal scroll bar is persistent. This is unrelated to whether `margin: 0` is set on `body` or not. Also, I think we should actually take margins, borders, perhaps padding into account when computing size.

almarklein:
Ah yes, horizontal scrollbars do appear. I just pushed a commit to also offset in width to fix that. Scrollbars may still appear briefly *during* resizing, because the canvas may be temporarily too large when the window/container is made smaller. If there are still scrollbars after resizing is done, its a bug and perhaps we need to use larger margins. I tested with FF, Chrome and Edge, and all seem ok.

mattpap:
Still one pixel off in y axis.

tests/integration/interaction/test_responsive.py
birdsarah: Thanks for the shoutout @almarklein. And thanks for taking the plunge and writing a test.

tests/integration/interaction/test_responsive.py
```diff
@@ -156,3 +162,132 @@ def test_responsive_plot_starts_at_correct_size(output_file_url, selenium):
...
+    css_resources = resources.render_css()
+    
+    script, div = components({'red': red})
+    
+    html = template.render(js_resources=js_resources,
+                        css_resources=css_resources,
+                        plot_script=script,
+                        plot_div=div)
+    
+    filename = output_file_url.split('/', 3)[-1]  # strip "http://xxxx:yyyy/"
```
birdsarah: It would be easy to expose the `file_path` as a fixture too so that you can just use `file_path` like `output_file_url` and not have to manually produce it.

If this PR is urgent, then we can just change it later. If not I could push the change to the branch.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -163,8 +164,7 @@ class PlotView extends Renderer.View
     if @mget('responsive')
       throttled_resize = _.throttle(@resize, 100)
       $(window).on("resize", throttled_resize)
-      # Just need to wait a small delay so container has a width
-      _.delay(@resize, 10)
+      @resize()
```
almarklein: Good catch. I was going to complain that Sarah probably added this for a reason, but we cover this with our checking-whether-its-in-the-dom-and-recalling-if-not thing.

almarklein:
Nice work @mattpap !  I still see scrollbars during resizing, though none after the resize. Do you mean you get a vertical scrollbar? It does look like something is sticking out on the right though. Not sure if that matters
bokeh01
.

tests/integration/interaction/test_responsive.py
almarklein: I's sort of urgent, but does not seem fully ready yet, so go ahead :)

bryevdv:
Apart from the conflict, what is remaining on this PR? Is there anything after this PR before cutting a dev build that dave can use?

mattpap:
@bryevdv, we will use a development version of bokeh (a submodule in the demo), because waiting for dev builds takes too much time and we need additional small changes anyway.

mattpap:
>  I still see scrollbars during resizing, though none after the resize.

This may be inevitable, I'm not sure right now. Thought as long as they aren't visible after, then we are OK.

>  It does look like something is sticking out on the right though.

@almarklein, I looks like you didn't recompile fully (including less) or maybe there is some caching issue. Or maybe it's cross-browser compatibility thing. I attempted to deal with this particular issue in commits 24e59cf and 5d7470d.

almarklein:
@mattpap apparently `gulp watch` does not recompile (all) style files. It seems fixed indeed.

> Thought as long as they aren't visible after, then we are OK.

I agree. The scrollbars seem to appear much less often in Chrome and Edge, BTW.

mattpap:
> apparently gulp watch does not recompile (all) style files.

Yeah, it rarely works as expected. I use it only for actual incremental work and use `build` or preferably `dev-build` (no expensive minification) after a pull or other major changes.

mattpap:
I regret at attempting to fix this issue.

bryevdv:
How can I help?

mattpap:
I looks like I made it work (up to this one stupid pixel). How many things I broke? Better not to know.

mattpap:
This PR contains another take on setting `bk-root`. Now it's done dynamically during embedding (one place instead of multiple). However, I have to set it on `body`. I would like to set it on `view.$el`, but that won't work with CSS (at least without wrapping, which I think we don't want to do).

almarklein:
> up to this one stupid pixel

Can we not increase the offset by 1 in that case?

How did things break? Because the classless div was removed? Was it really necessary to remove it?

bryevdv:
> However, I have to set it on body

Isn't this at least potentially problematic for anything else on the page?

mattpap:
> Isn't this at least potentially problematic for anything else on the page?

This was a horribly stupid idea defeating the purpose of `bk-root`. I have to revert this or find a reasonable one. The best solution would be to add it on the main element of the embedded view, but I'm not able to find a sane solution to get CSS working in such scenario.

birdsarah:
Hi. I have a much simpler solution:

https://github.com/bokeh/bokeh/pull/4133

birdsarah:
It may be worth noting that this isn't a hack. The responsiveness broke when the plot_div template changed 

```diff
-<div class="plotdiv bk-plot" id="{{ elementid }}"></div> 
+<div class="bk-root">
+    <div class="plotdiv" id="{{ elementid }}"></div>
+</div> 
```
So it seems sensible to me to now check up one more parentNode for the height.

almarklein:
@birdsarah That was more or less what we had, we added a bit more smartness to better determine the required amount of offset in width/ height to avoid showing scrollbars. Then Matheusz made it smarter still by measuring exactly how much the offset should be, but also changed the DOM structure, which apparently causes trouble.

Frankly, I'm not sure I get why the change to the DOM would be better. Also @mattpap, you mentioned its still 1px off, but the scrollbars are gone, right? What 1px do you mean?

mattpap:
@birdsarah, see the second commit in this PR, it does what you did but in a more generic way.

birdsarah:
Can I suggest fixing one thing at a time. The one line fixes the responsiveness, and is a clear response to a bug introduced in the past

The other things you've done may be improvements but seem to be causing a lot of headaches. Can I also suggest that we don't have good enough test coverage to feel confident in messing with fundamental things like the plot_div template etc.

mattpap:
> What 1px do you mean?

@almarklein, see the n-1 commit. I added `availHeight -= 1`. Without this you get an extra pixel from nowhere and the annoying vertical scrollbar.

tests/integration/integration_tests_plugin.py
```diff
@@ -30,6 +30,7 @@ def output_file_url(request, file_server):
     filename = request.function.__name__ + '.html'
     file_obj = request.fspath.dirpath().join(filename)
     file_path = file_obj.strpath
+    url = file_path.replace('\\', '/')  # Windows-proof
```
birdsarah: This is not the place to do this.

The place would be to fix handling of windows file_paths in `tests/plugins/file_server.py` `where_is` method

https://github.com/bokeh/bokeh/blob/almarklein/4110-resize/tests/plugins/file_server.py#L115

mattpap:
>  The one line fixes the responsiveness, (...)

I kinda fixes the problem, until we change the HTML in a another commit and it doesn't work again.

tests/integration/integration_tests_plugin.py
birdsarah: mmmmm.....I think I've changed my mind on the above comment.

mattpap:
@almarklein, can you resubmit your original idea in another PR, including changes needed for the demo, excluding all crap I added? Lets have something working for the demo and lets revisit other issues (magic constants, `bk-root` later)

birdsarah:
> see the second commit in this PR, it does what you did but in a more generic way.

I guess I should note my bias that as I'm working on layout (which includes responsiveness) - where subplots do work, tick extents are taken into account etc, then I am not inclined to invest heavily in this version of responsiveness.

birdsarah:
> It kinda fixes the problem, until we change the HTML in a another commit and it doesn't work again.

ok - i'll take a look at the new PR. Changing the templates makes me very nervous due to our lack of test coverage. Especially doing when there's pressure for a deadline.

bokehjs/src/coffee/api/typings/models/plots.d.ts
```diff
@@ -3,6 +3,11 @@ declare namespace Bokeh {
         plot?: Plot;
     }
 
+    export interface PlotView extends View<Plot> {
+        resize_width_height(use_width: boolean, use_height: boolean, maintain_ar?: boolean,
+                            width?: number, height?: number): void;
+    }
+
```
mattpap: @almarklein, btw., during experimentation phase you don't actually need static APIs, so you can easily proceed with something like:
```
let view: Bokeh.View<Plot> = make_plot(...);
(view as any).resize_width_height(...);
```

almarklein: Ah, thanks for the tip. The view was set to `any` anyways (until recently), but I thought I'd add some typing already so we wont forget this function.

almarklein:
#4136

bokehjs/src/coffee/api/typings/models/plots.d.ts
mattpap: Sure, just pointing out if we need to iterate at faster pace.

PR #4121, opened 2016-04-06
Remove obsolete CSS classes and bokehjs/src/templates

mattpap:


bryevdv:
LGTM I am going to make an issue and then merge, going forward we should not have any more "issue PRs"

PR #4123, opened 2016-04-06
Update to phantomjs=2.1.7 (now phantomjs-prebuilt) hosted by GH

### LINKED ISSUE #4099: Fix phantomjs download failures
Our travis CI tests are pretty much useless right now due to a trivial failure. Based on quick research we just have to provide our own download URL that can be configured through an env variable (see http://npmjs.com/package/phantomjs for details). I think we should just have to upload phantomjs binary (in correct version) to our S3 and change `.travis.yml` accordingly. Note that `npm install` is called in `conda.recipe/build.sh`, but I think changing `.travis.yml` is the right thing to do. If there other other ideas, please speak up, or preferably submit a PR.

canavandl:


bryevdv:
@birdsarah we could try putting `n=2` back in this PR, and just use this PR to run a bunch of times to see what if any kind of failures we are still seeing

canavandl:
Tests have passed twice w/ py.test examples -n1 flag. Now testings -n2 flag.

bryevdv:
@birdsarah this signal error:
```
E       ValueError: signal only works in main thread
```
seems pretty reliable with `n=2`. It certainly makes sense to me why this couldn't ever happen with `n=` (there is only the main thread)

bryevdv:
I thought we had discussed at some point a solution for that test runner code that did not involve `signal`? That's the crux of the issue. In any case I think we should switch back to `n=1` in this PR and merge it, taking up `n=2` in another PR.

bryevdv:
sometimes being able to edit directly in GH is the most amazing thing. OK will run tests a few more times with `n=1`

bryevdv:
passed immediately with `n=1`, merging

PR #4128, opened 2016-04-06
fix charts legend sorting

### LINKED ISSUE #4057: Request: reorderable legends with chart API
I have plots like:

screen shot 2016-03-22 at 11 14 52

Unfortunately, the legend entries are in the opposite order of the content of the plot.  I can't flip around my columns because of assumptions I make in creating the area chart (negative data should be on the left of any input data frame.)  Code for that is at https://github.com/msarahan/bokeh/commit/c1dda0f2861355ab2189f6b7e8b2d1da2751b7f9

If legend entries were reorderable somehow, so that I could either flip the collection around, or arbitrarily order it by passing a list of keys, that would be very helpful.

fpliger:
issues: fixes #4057


bryevdv:
@fpliger do you have bandwidth to add some documentation for this feature?

fpliger:
@bryevdv good point. I better find the bandwidth before merging this.... thanks :)

bokeh/charts/builder.py
```diff
@@ -547,6 +552,33 @@ def generate_help(cls):
...
+                    item_order = [x[0] if isinstance(x, tuple) else x
+                        for x in attributes[attr].items]
+
+                    item_order = [str(x) if not isinstance(x, string_types)
+                        else x for x in item_order]
+
+                    def foo(leg):
+                        return item_order.index(leg[0])
+
+                    return list(sorted(legends, key=foo, reverse=(not asc)))
```
bryevdv: `sort_legend` can be a list, but won't this exit/return on the first one? I don't see how more than one of the tuples can have effect. Is the intent of multiple tuples to do sub-sorting within groups? I don't quite follow the intent.

fpliger: ooops... definitely. I've overseen it when encapsulating this in a function. Will fix it. tnx

fpliger: @bryevdv thinking a bit more about this I'm not really sure it makes sense since all attributes would refer to the same legends so wouldn't be sub-sorting but re-sorting instead which would not be correct. I can't find a use case when you'd use more then one attribute in `sort_legend`. It would make more sense if `sort_legend` was a `Dict` or a `Tuple(String, Bool)`...

bryevdv: Right I'm not suggesting implementing sub-sorting is the thing to do, only that the current interface suggests something more general than is actually implemented, so maybe the property needs to be tightened up. `Tuple(String, Bool)`  (or maybe better with enum: `Tuple(String, SortDirection)`) seems reasonable to me.

bryevdv: Tho in that case, two properties `legend_sort_field` and `legend_sort_direction` might be simpler for users.

fpliger: +1. I like the latter more too since it's more informative...  `Tuple(String, SortDirection)`

fpliger: Oh.. we commented about the same time. Breaking down to 2 props may be simpler.. I'm ok with both options. @bryevdv or others have a preference?

bokeh/charts/builder.py
```diff
@@ -547,6 +573,47 @@ def generate_help(cls):
...
+    @property
+    def sort_legend(self):
+        warnings.warn("Chart property 'sort_legend' was deprecated in 0.12 \
+            and will be removed in the future.")
+        return [(self.legend_sort_field, self.legend_sort_direction)]
+
+    @sort_legend.setter
+    def sort_legend(self, value):
+        warnings.warn("Chart property 'sort_legend' was deprecated in 0.12 \
+            and will be removed in the future.")
```
bryevdv: need to point users to the replacement. Also not certain, but worth seeing if https://github.com/bokeh/bokeh/blob/master/bokeh/util/deprecate.py works with properties.

PR #4132, opened 2016-04-07
Hotfix - Fix path to phantomjs executable

### LINKED ISSUE #4099: Fix phantomjs download failures

canavandl:
issues: fixes #4099

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ branches:
 
 before_install:
     - export PATH="$HOME/miniconda/bin:$PATH"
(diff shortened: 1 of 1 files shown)


canavandl:
That seemed to fix it. The server examples haven't worked recently, but the notebook and file examples are now being captured.

Here's the py2 examples dump: https://s3.amazonaws.com/bokeh-travis/121469880/examples.html

PR #4136, opened 2016-04-07
Fix resizing of embedded plots

### LINKED ISSUE #4110: Plot resizing is broken

almarklein:
fixes #4110, supersedes #4120

### DIFF
--- a/bokehjs/src/coffee/api/typings/models/plots.d.ts
+++ b/bokehjs/src/coffee/api/typings/models/plots.d.ts
@@ -9,6 +9,7 @@ declare namespace Bokeh {
         add_layout(obj: Model, place?: Place): void;
         add_glyph(glyph: Glyph, source?: DataSource, attrs?: ModelOpts): GlyphRenderer;
         add_tools(...tools: Array<Tool>): void;
+        resize_width_height(use_width:boolean, use_height:boolean, maintain_ar?:boolean, width?:number, height?:number): void;
     }
     export interface IPlot extends IBasePlot {
         x_range?: Range;
--- a/bokehjs/src/coffee/api/typings/models/plots.d.ts
+++ b/bokehjs/src/coffee/api/typings/models/plots.d.ts
@@ -3,13 +3,17 @@ declare namespace Bokeh {
         plot?: Plot;
     }
 
+    export interface PlotView extends View<Plot> {
+        model: Plot;
+        resize_width_height(use_width:boolean, use_height:boolean, maintain_ar?:boolean, width?:number, height?:number): void;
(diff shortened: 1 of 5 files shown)


almarklein:
cc @mattpap @birdsarah 

@mattpap you want me to put back some of the improved styling that you did? Or do you want to do that?

mattpap:
> You want me to put back some of the improved styling that you did? Or do you want to do that?

No and no. I need to rethink those.

almarklein:
I was planning to add back in some of the improvements that @mattpap did, but I'm out of time now. I propose we merge this now, so that Dave can proceed with a working plot.

mattpap:
Sure, lets merge. As to the other modifications, I think we will have to wait for @birdsarah's PR(s). We have to reduce the number of embedding possibilities and perhaps always use a layout (even a trivial one) and then embed a plot.

bryevdv:
> Sure, lets merge. 

OK, looks good as far as I can tell, though I am not an expert in this area. I do like the smaller scope of this PR.

birdsarah:
> We have to reduce the number of embedding possibilities and perhaps always use a layout (even a trivial one) and then embed a plot.

I think that's the direction I'm going in. I'm writing something up at the moment.

Issue #3054, opened 2015-11-02
Toolbar Improvements: Active is programmable


birdsarah:
@brendancol just hit the example where he wanted to hide the toolbar (`toolbar_location=None`) but wanted zooming on.

We recently just turned zoomwheel (#3003) off by default, but the only way to turn it on is by clicking the toolbar icon.

We could add an argument to WheelZoomTool `active=True` for example, but it seems like we'd want something that worked across all tools and was consistent.

The challenge is that you should only have one tool active in a group at any one time. For WheelZoom this doesn't matter. It lives in the `scroll` group and there's only one tool in the `scroll` group, but for `pan` events in particular it's a problem.

I'm not entirely sure what a clean / obvious spelling is. We could have `active=True` on the invidivual models assuming that if you're using the Model interface you should know what you're doing and then have `figure` and `chart` have something like active_tools='list,of,tools' and it makes checks on it?

Worth noting that you can programmatically click the WheelZoomTool with if your toolbar is visible (but not if you've set `toolbar_location=None`)
````html
<script>
    Bokeh.$('.bk-button-bar-list[type="scroll"] button').click()
</script>
````

Update: if you have set `toolbar_location=None` you can reach into the guts of Bokeh to turn your WheelZoom on as follows:

````js
for (plot_index in Bokeh.index) {
 var plot_tools = Bokeh.index[plot_index].tools;
 for (tool_index in plot_tools) { 
   var tool = plot_tools[tool_index];
   if (tool.mget('tool_name') == 'Wheel Zoom') {
     tool.mset('active', true);
   }
 }
}
````

dwf:
+1, it'd be really nice to be able to turn this on by default from Python.

danaki:
I develop a Flask application and I need to click on wheel zoom every time I reload the chart. Please consider to make a setting in bokeh.charts.defaults. Would be great to configure the set of available tools by default too.

esvhd:
+1 This would be very useful to do from Python from my side too. I was just searching for ways to do that.

fingon:
+1 - while I just started looking hard at Bokeh, I instantly missed this (and found someone else did too; why wouldn't I want to use scrollwheel to interact with a canvas?).

I would like to select the 'selected tools' by default as well. I think probably clean way to do this would be to add second list to figure, e.g. select_tools=[] or equivalent with similar syntax as tools=, except it would select those by default.

datascientette:
[...]

PR #4138, opened 2016-04-08
Declarative property management

### LINKED ISSUE #4141: Declarative Property management
clean up usage of properties in BokehJS, including 

-  all public (serializable) properties use `@define`
- all private (nonserializable) properties use `@internal`
- all overrides use `@override`
- all cases of `defaults` are removed
- `defaults`, `nonserializable_property_names` etc. are removed
- `set(prop)` and `get(prop)` throw exception when `prop not in @props`

mattpap:
issues: closes #4141

- [x] all public (serializable) properties use `@define`
- [x] all private (nonserializable) properties use `@internal`
- [x] all overrides use `@override`
- [x] all cases of `defaults` are removed
- [x] `defaults`, `nonserializable_property_names` etc. are removed
- [x] `set(prop)` and `get(prop)` throw exception when `prop not in @props`
- [x] computed properties participate in override analysis
- [x] make sure bokehjs works after all those changes


bryevdv:
@mattpap I have made an issue and linked per BEP-1

bokehjs/src/coffee/models/sources/geojson_data_source.coffee
```diff
@@ -18,7 +18,7 @@ class GeoJSONDataSource extends ColumnDataSource.Model
   initialize: (options) ->
     super(options)
     @geojson_to_column_data() # this just validates the initial geojson value
-    @register_property('data', @geojson_to_column_data, true)
+    @define_computed_property('data', @geojson_to_column_data, true)
```
mattpap: @bokeh/dev, can someone explain why `GeoJSONDataSource` inherits from `ColumnDataSource`, yet overrides its `data` property with a computed `data` property? @bryevdv, your comment above indicates you were confused about this as well.

mattpap: Because of this, I can't throw an exception in `define_computed_property()`, just log a warning.

mattpap:
@bokeh/dev, this needs review.

bokeh/models/tools.py
```diff
@@ -186,11 +186,12 @@ class TapTool(Tool):
...
+    pass
 
-    active = Bool(True, help="""
-    Whether the tool is intially active or not. If set to ``False``, the user
-    will have to click tool's button to active it.
-    """)
+    #active = Bool(True, help="""
+    #Whether the tool is intially active or not. If set to ``False``, the user
+    #will have to click tool's button to active it.
+    #""")
```
mattpap: @bryevdv, I'm not sure what to do about this. I think this could be moved to `Tool`. It won't be bullet proof, due to know issues (need for an actual tool manager).

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -266,40 +287,30 @@ class HasProps extends Backbone.Model
...
-    if not @constructor._nonserializable_names_cache?
-      names = {}
-      for n in @nonserializable_attribute_names()
-        names[n] = true
-      @constructor._nonserializable_names_cache = names
-    @constructor._nonserializable_names_cache
-
   attribute_is_serializable: (attr) ->
-    (attr not of @_get_nonserializable_dict()) and (attr of @attributes)
+    if attr == "id"
```
mattpap: This is (and other similar instances) is ugly. I may need to to  add `id` to `@props`.

mattpap:
Currently our examples' tests are all green, yet most don't work. There are failures indicated in JS console, but the new testing infrastructure does nothing about this.

mattpap:
I presume this is also why integration really fail. Currently they fail with a bogus `index error`, but that's because presumably nothing renders, so there is no point in proceeding with actual selenium tests.

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -16,22 +16,22 @@ class GlyphRendererView extends Renderer.View
 
     selection_glyph = @mget("selection_glyph")
     if not selection_glyph?
-      attrs = _.extend {}, glyph_attrs, @model.selection_defaults
-      selection_glyph = new (@mget("glyph")).constructor(attrs)
+      #attrs = _.extend {}, glyph_attrs, @model.selection_defaults
```
mattpap: Applying arbitrary properties to arbitrary glyphs.

mattpap:
@bryevdv, what `reference_point` stands for in `models/glyphs/glyph.coffee`? There is `get_reference_point()` method that always returns `undefined` (which is turned into `0` at call site).

bokehjs/src/coffee/models/sources/geojson_data_source.coffee
bryevdv: ping @birdsarah

bokeh/models/tools.py
bryevdv: My only comment is what about Button Actions (e.g. preview save tool)? they are just there, to be clicked, not active or not active .

bokehjs/src/coffee/core/has_props.coffee
bryevdv: The downside of objects maintaining their own identity is that it is redundant, and and could potentially go out of sync with the index if ids-to-models (which would be a bug / due to a bug, of course). But we seem to actually be doing it anyway, and just making a big effort to say that we aren't, so I'd say it's worth exploring.

bryevdv:
In the recent examples tests I see alot of "no diff" @birdsarah is this because the images match exactly, or because there are no images to match? I had assumed it was the former...

bryevdv:
@mattpap the intent behind `reference_point` was to give more control over the legend, if you wanted to pick a *specific* index in your column to be used to represent the data in the legend. But this was never fully developed, and AFAIK not exposed to python, which means we could rip it out and re-visit better legend control more thoroughly at a different time.

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
bryevdv: I guess there is an implicit assumption (not enforced) that they are all the *same* glyph type (up to an equivalence class of "marker", those should be freely interchangeable)

mattpap:
@bryevdv, I had to fix in this PR several errors introduced in `ts_apiN` PRs. I had to make some local modifications to the new testing infrastructure to make work for me.

bokeh/models/tools.py
mattpap: Yeah, I will have to introduce another abstract class to encode this.

bokeh/model.py
```diff
@@ -255,7 +255,21 @@ def _to_json_like(self, include_defaults):
...
+        subtype = getattr(self.__class__, "__subtype__", None)
+        if subtype is not None and subtype != self.__class__.__view_model__:
+            attrs = {}
+            for attr, value in all_attrs.items():
+                if attr in self.__class__.__dict__:
+                    continue
+                else:
+                    attrs[attr] = value
+        else:
+            attrs = all_attrs
```
mattpap: @bryevdv, this change is a big one. Pretty much all charts examples failed in this PR. We were serializing all properties, even if there was nothing in bokehjs. I'm still curious why defaults tests worked. Also, this may not be the best location for this code, but looking on how much logic is there in `properties*()` methods, I preferred not to touch them in this PR.

bokehjs/gulp/eco.coffee
```diff
@@ -1,14 +1,59 @@
...
+          .replace(/&/g, '&amp;')
+          .replace(/</g, '&lt;')
+          .replace(/>/g, '&gt;')
+          .replace(/\x22/g, '&quot;');
+      };
+      (function() {
+    #{indent(script, 4)}
+      }).call(__obj);
+      return __out.join('');
+    }
```
mattpap: Previously this code modified `_obj` with `escape` and `safe` function. So, if you passed `@model.attributes` to a template, you got two new model attributes for free.

bokeh/model.py
bryevdv: A concrete example would help me here.

mattpap: E.g. `Chart` which is `Plot` and is implemented as `Plot` in bokehjs. `Chart` adds `legend`, which previously was serialized. This resulted in an equivalent of `Plot.set("legend", ...)`, which in this PR is not allowed, because `Plot` doesn't define `legend`.

mattpap:
> `py.test: error: unrecognized arguments: --upload`

Integration tests fail with this, yet the succeeded two commits ago.

bokeh/model.py
bryevdv: I see. I think this might break round-tripping of subtypes then, I'm not sure if that is a problem or not, but I do believe it used to work.

mattpap: It used to work and it still works. The only difference is that we won't be sending garbage over the wire. When use `__subtype__`, then you pretty much say that all new properties are private to the python side. This works somewhat like nonserializable (now internal) properties in bokehjs.

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
mattpap: That's a different thing. Here the problem is that e.g. `decimated_defaults` contain both fill and line defaults, when glyph can just support non or one of them (e.g. `Line`).

bryevdv: Ah, yes I see.

bokehjs/src/coffee/models/annotations/box_annotation.coffee
```diff
@@ -105,22 +105,18 @@ class BoxAnnotation extends Annotation.Model
...
+  @override {
+    fill_color: '#fff9ba'
+    fill_alpha: 0.4
+    line_color: '#cccccc'
+    line_alpha: 0.3
+  }
+
+  @internal {
+    silent_update: [ p.Boolean, false ]
+  }
```
bryevdv: Well, I'm removing this in #4146 but I will handle the conflict.

bokehjs/src/coffee/models/annotations/poly_annotation.coffee
```diff
@@ -59,23 +59,19 @@ class PolyAnnotation extends Annotation.Model
...
+    fill_color: "#fff9ba"
+    fill_alpha: 0.4
+    line_color: "#cccccc"
+    line_alpha: 0.3
+    line_alpha: 0.3
+  }
+
+  @internal {
+    silent_update: [ p.Boolean, false ]
+  }
```
bryevdv: also removed in #4146 just FYI

bokehjs/src/coffee/models/axes/axis.coffee
```diff
@@ -314,54 +314,51 @@ class Axis extends GuideRenderer.Model
...
     @add_dependencies('tick_coords', this, ['computed_bounds', 'layout_location'])
 
-    @register_property('ranges', @_ranges, true)
-    @register_property('normals', (() -> @_normals), true)
-    @register_property('dimension', (() -> @_dim), true)
-    @register_property('offsets', @_offsets, true)
+    @define_computed_property('ranges', @_ranges, true)
+    @define_computed_property('normals', (() -> @_normals), true)
+    @define_computed_property('dimension', (() -> @_dim), true)
+    @define_computed_property('offsets', @_offsets, true)
```
bryevdv: Thanks for the better name, FWIW I would like to eventually get rid of computed properties entirely.

bokehjs/src/coffee/models/tools/button_tool.coffee
```diff
@@ -17,7 +18,7 @@ class ButtonToolButtonView extends Backbone.View
 
   initialize: (options) ->
     super(options)
-    @$el.html(@template(@model.attrs_and_props()))
+    @$el.html(@template({model: @model}))
```
bryevdv: nice

bryevdv:
did a somewhat quick pass over things and LGTM

mattpap:
I tested as much as I could locally. This PR is actually much better tested than `ts_apiN` were and multiple bugs introduced there were fixed here. There are two minor issues to resolve, reexpose `Tool*.active` and restore `*_defaults` in `GlyphRenderer`, both which I will handle in separate PRs.

bryevdv:
@mattpap regarding tools and `active` please check out https://github.com/bokeh/bokeh/issues/3054 a well

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
mattpap: @bryevdv, I restored the support. Now I apply fill and line props individually if supported.

bokehjs/src/coffee/models/sources/geojson_data_source.coffee
birdsarah: @mattpap 

> @bokeh/dev, can someone explain why GeoJSONDataSource inherits from ColumnDataSource, yet overrides its data property with a computed data property? @bryevdv, your comment above indicates you were confused about this as well.

It's not ideal. We had a bigger plan to support GeoJSON (and other non-columnar data sources) properly but everything in BokehJS is currently very structured around the CDS so this compromise of sending the GeoJSON down the pipe and then turning it into columnar data on the BokehJS side is what we came up with.

I've been trying to dig up the long discussion I had with bryan about it, but can't find it at the moment.

But there's this which describes some of the plans: https://github.com/bokeh/bokeh/wiki/Geo-Ramblings#renderers--non-columnar-data-sources

birdsarah:
> In the recent examples tests I see alot of "no diff" @birdsarah is this because the images match exactly, or because there are no images to match? I had assumed it was the former...

@bryevdv - it's the former. I added it into the template to make it explicit that the two pictures didn't have a diff.

Issue #177, opened 2013-11-16
compound glyphs


bryevdv:
Seems like this would be useful for making combining lines and arrowhead simpler. Also for line plots that also have markers, if the legend asks a compound glyph (for the lines and markers together) to render, then it will correctly get both the line and the marker in the legend.

birdsarah:
Just had another desire for a compound glyph after thinking about #3206.

I want to make a box with some text (for example) and then when I hover/select the box, then the styling for all the things is updated so they're acting as one with respect to hit testing.

bryevdv:
Yes I have also been thinking about this wrt annotations. I want to do some preliminary work to line up the JS and python project structures before embarking on some things, but reference-able "anchor points" for glyphs would be useful in many areas.

bryevdv:
Arrows/error bars are implemented, there's not really enough left here to be actionable.

jasonsross:
could someone point me to where arrows have been implemented? I see arrow annotations, but those don't seem like robust plotting tools as opposed just putting an arrowhead line_cap on segments allowing quiver plots. Right now you have what you call a quiver plot, but how is it quiver if they are not arrows?

bryevdv:
Only arrow annotations have been implemented. As I said the original issue was never detailed or specific enough to implement anything general. It's something we could consider if someone is interested enough to develop a fully fleshed out proposal.

bryevdv:
>  Do you just mean you want a clear explanation of the request or a fully fleshed out implementation?

Not implementation, but things like:

* clear and specific requirements and goals (definition of "success")
* brief discussion about possible implementation approaches or any pitfalls
* concrete API suggestions 
* complete examples of usage that would be expected to work when the feature is done

PR #4142, opened 2016-04-08
Canavandl/arrow annotation

canavandl:
Add support for arrow annotation

- [x] ref #177
- [x] tests added / passed
- [] ~~~release document entry (if new feature or API change)~~~


canavandl:
example:

```python
from bokeh.models import Arrow
# output to static HTML file
output_file("arrow.html")

arrow1 = Arrow(tail_x=2, tail_y=2, head_x=3, head_y=5,
               tail_border_line_color='black',
               tail_body_fill_color='green'
               )

arrow2 = Arrow(tail_x=5, tail_y=7, head_x=2, head_y=6, head_border_line_color=None)

arrow3 = Arrow(tail_x=100, tail_y=100, head_x=150, head_y=300,
               tail_units='screen', head_units='screen')

p.add_annotation(arrow1)
p.add_annotation(arrow2)
p.add_annotation(arrow3)

show(p)
```

screen shot 2016-04-08 at 5 11 57 pm

bryevdv:
Are you switching the styles based on various colors being `None`? I would suggest explicit enum properties for arrow style. There's lots :) 

http://mathematica.stackexchange.com/questions/7052/how-can-i-customize-arrowhead-shape-to-match-common-latex-styles

A style of `None` means don't draw that arrow head at all.

looking at the link above makes me think half-filled or dual-color arrowheads would be pretty neat :)

canavandl:
Added `head_style` and `tail_style` properties of `ArrowStyle` enum ["open", "closed"]. None will not draw arrow head.

@bryevdv which other arrow head styles do you want to support before merging?

bryevdv:
Additional reference: http://www.graphviz.org/doc/info/attrs.html#k:arrowType

canavandl:
screen shot 2016-04-11 at 11 11 34 am

canavandl:
styling.rst plot:

screen shot 2016-04-11 at 12 52 22 pm

canavandl:
Now ArrowStyle auto-docs as:

screen shot 2016-04-11 at 1 25 57 pm

bryevdv:
Awesome!

bokeh/models/annotations.py
```diff
@@ -107,6 +108,89 @@ class Legend(Annotation):
...
+    The y-coordinates to locate the arrow heads.
+    """)
+
+    head_size = Float(default=25, help="""
+    The size, in pixels, of the arrow head.
+    """)
+
+    tail_units = Enum(SpatialUnits, default='data', help="""
+    The unit type for the head_x and head_y attributes. Interpreted as "data
+    space" units by default.
```
bryevdv: I wonder if there are any use cases where a user might want to specify units for `x` and `y` independently? Also the grouping of props here is a bit weird I would put all the "head" props and all the "tail" props together.

bokeh/models/annotations.py
```diff
@@ -107,6 +108,89 @@ class Legend(Annotation):
...
+    """)
+
+    head_units = Enum(SpatialUnits, default='data', help="""
+    The unit type for the head_x and head_y attributes. Interpreted as "data
+    space" units by default.
+    """)
+
+    tail_style = Enum(ArrowStyle, default=None, help="""
+    The style of the head's arrow tail.
```
bryevdv: This sentence is a bit awkward. Maybe "A style for the arrow head at the head of the arrow. Now I actually wonder if head and tail for the coordinates is too confusing. What about start/end? Then: "A style for the arrow head at the start of the arrow"

bokeh/models/annotations.py
```diff
@@ -107,6 +108,89 @@ class Legend(Annotation):
...
+    head_style = Enum(ArrowStyle, default='open', help="""
+    The style of the head's arrow head.
+    """)
+
+    body_props = Include(LineProps, use_prefix=True, help="""
+    The %s values for the arrow body.
+    """)
+
+    tail_border_props = Include(LineProps, use_prefix=True, help="""
+    The %s values for the arrow tail outline.
```
bryevdv: so `tail_border_line_color?` I think `tail_line_color` would suffice

bokeh/models/annotations.py
```diff
@@ -107,6 +108,89 @@ class Legend(Annotation):
...
+    head_border_props = Include(LineProps, use_prefix=True, help="""
+    The %s values for the arrow head outline.
+    """)
+
+    head_body_props = Include(FillProps, use_prefix=True, help="""
+    The %s values for the arrow head. If `fill_color` is set to None, the arrow
+    head will be an open-backed wedge.
+    """)
+
+    head_body_fill_color = Override(default="black")
```
bryevdv: same comment for all of these I think "body" is superfluous

bokeh/models/annotations.py
```diff
@@ -107,6 +108,89 @@ class Legend(Annotation):
...
+    The style of the head's arrow tail.
+    """)
+
+    head_style = Enum(ArrowStyle, default='open', help="""
+    The style of the head's arrow head.
+    """)
+
+    body_props = Include(LineProps, use_prefix=True, help="""
+    The %s values for the arrow body.
+    """)
```
bryevdv: I think `use_prefix = False` maybe? Then: `arrow.line_color = "black"` which seems fine?

bokeh/models/annotations.py
bryevdv: Not saying we necessarily should change to allow it but worth a quick thought.

bryevdv:
OK a few second thoughts about some naming, the implementation/tests/docs looks great.

bokeh/models/annotations.py
canavandl: > I wonder if there are any use cases where a user might want to specify units for x and y independently? Also the grouping of props here is a bit weird I would put all the "head" props and all the "tail" props together.

I thought about this and couldn't think of a good use case for it. I'm willing to merge as is and revisit if/when somebody requests it.

bryevdv: Well it's the sort of thing that's hard to change later without breaking anything, but maybe not impossible. And maybe not needed at all, so fine with keeping it as-is.

bokeh/models/annotations.py
bryevdv: Alternatively graphviz uses this language: http://www.graphviz.org/doc/info/attrs.html#d:arrowhead

bryevdv: "Style of the arrowhead at the head/tail node of an arrow" in which case we could keep head and tail as is. The main confusion for me was seeing "head" twice

canavandl: I used head/tail because it was more clear (to me) where the head was than for start/end. But having added support for double-sided arrows, that seems like a less difficult distinction.

ggplot uses x/y/xend/yend (where the xend/yend is the arrow head):

```r
b + geom_curve(
  aes(x = x1, y = y1, xend = x2, yend = y2),
  data = df,
  arrow = arrow(length = unit(0.03, "npc"))
)
```
http://docs.ggplot2.org/current/geom_segment.html

I'm willing to move to start_x/y and end_x/y

bryevdv: hmm I kind of like `x` and `y` at the beginning. `xstart`, `xend` /  `ystart`, `send` ?

bryevdv: or `x_head` if we keep head/tail

bokeh/models/annotations.py
canavandl: how would you set the fill and line props to use the "head_" prefix but not clobber each other as class variables? i.e.:

```python
...
head_props = Include(LineProps, use_prefix=True, help="""
     The %s values for the arrow head outline.

head_props = Include(FillProps, use_prefix=True, help="""
    The %s values for the arrow head. If `fill_color` is set to None, the arrow head will be an open-backed wedge.""")
```

bryevdv: that will add `head_line_XYZ` and `head_fill_XYX`

canavandl: ah - I didn't grok what the `Include` method did.

canavandl: It appears that doing the above causes the FillProps attr to override the LineProps attr (as both are set to the same class var name `head_props`.

bryevdv: What do glyphs do?

canavandl: I'm not sure any glyphs set multiple line or fill props with the same prefix, so they don't have the naming collision (glyphs IIRC don't use prefixes). This is why I was doing the border/body prefixing.

bryevdv: ping @mattpap

bryevdv: So first, I expect this will be a widely used interface, so we do have to "get it right" I see three options:
* add all the properties by hand (don't use `Include`)
* extend `Include` to accept a specified prefix in addition to inferring it
* split out a smaller `ArrowHead` class that `Arrow` can contain (then `Include` can be used on `ArrowHead` exactly as in glyph)

Under the last scenario, you'd have e.g., `arrow.head.line_color` and `arrow.tail.style`

I think I might be slightly partial to the last, actually, but I would accept any of the above.

bryevdv: @bokeh/dev thoughts or preferences here?

canavandl:
Three changes:
* Go to a naming system of x_start/end and y_start/end from head/tail_x and head/tail_y
* `body_line_props` to simply `line_props` (so arrow.line_width not arrow.body_line_width)
* Manually add the start/end head `FillProps` (so arrow.start_fill_color not arrow.start_body_fill_color)

Pending agreement on the property naming conventions, this should be ready to review/merge.

canavandl:
ping @bryevdv for review. Specifically:
* the new naming conventions of `Arrow` coords (x_start/y_end)
* implementation of arrow_head.coffee

bryevdv:
@canavandl this looks great my only comment offhand is that a classed named `Normal`, `Open`, `Vee` are probably a little too generic. Maybe `NormalHead` etc?

canavandl:
@bryevdv I made the suggested change (e.g. Open -> OpenHead)

bryevdv:
LGTM @bokeh/dev any other thoughts or comments before merging?

PR #4146, opened 2016-04-09
overlay annotations respond to backbone changes

### LINKED ISSUE #3992: BoxAnnotation does not update consistently

bryevdv:
issues: fixes #3992

Noticed that `PolyAnnotation` almosts had the same issue. Also realized that the `update` functions only purpose was to be for silent updates (normal updates that trigger `change` can just use standard `.set`) so removed that internal property and simplified associated logic)

### DIFF
--- a/bokehjs/src/coffee/models/annotations/box_annotation.coffee
+++ b/bokehjs/src/coffee/models/annotations/box_annotation.coffee
@@ -12,10 +12,14 @@ class BoxAnnotationView extends Renderer.View
     @$el.hide()
 
   bind_bokeh_events: () ->
+    # need to respond to either normal BB change events or silent
+    # "data only updates" that tools might want to use
     if @mget('render_mode') == 'css'
       # dispatch CSS update immediately
+      @listenTo(@model, 'change', @render)
       @listenTo(@model, 'data_update', @render)
     else
+      @listenTo(@model, 'change', @plot_view.request_render)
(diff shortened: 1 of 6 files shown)


bryevdv:
@philippjfr @canavandl @mattpap this is ready for review

canavandl:
LGTM

mattpap:
@bryevdv, I resolved merge conflicts after #4138.

PR #4147, opened 2016-04-09
update in streaming lists in place

### LINKED ISSUE #4077: Avoid copies in ColumnDataSource.stream
When trying to optimize performance when using `ColumnDataSource.stream` I believe I've become bottlenecked by the fact that `.stream` makes local copies when updating local columns.  When we look at the `_stream` function

```python
    def _stream(self, doc, source, new_data, rollover=None):
        old = self._saved_copy()

        import numpy as np

        for k, v in new_data.items():
            if isinstance(self[k], np.ndarray):
                data = np.append(self[k], new_data[k])
            else:
                data = self[k] + new_data[k]
            if rollover and len(data) > rollover:
                data = data[-rollover:]

            super(PropertyValueDict, self).__setitem__(k, data)

        from ..document import ColumnsStreamedEvent

        self._notify_owners(old,
                            hint=ColumnsStreamedEvent(doc, source, new_data, rollover))
```

We see a few cases of copying:

```python
old = self._saved_copy()
data = self[k] + new_data[k]
data = data[-rollover:]
```

While this might be fine for smallish lists.  Copies of larger lists quickly become expensive, especially when we do this on a tight inner loop.

```python
In [1]: L = list(range(1000000))

In [2]: %time _ = L.copy()
CPU times: user 10.6 ms, sys: 3.74 ms, total: 14.4 ms
Wall time: 14.2 ms
```

One could consider operating in place instead

```
L = self[k]
L.extend(new_data[k])
del L[:-rollover]
```

The del from front isn't ideal, but it tends to complete in around 1ms.  

Unfortunately, things stop rendering.  I suspect this is because it's tricky to identify what has actually changed and send out the proper events.

### Motivation

My applications runs quite quickly when I use `stream(..., 1000)` but horribly slowly when I use `stream(..., 20000)`.  Stream is unfortunately not linear in the size of the newly added dataset.

### Benchmark script

I used the following bokeh server script to help in benchmarking

```python
#!/usr/bin/env python

from __future__ import print_function, division, absolute_import

from bokeh.plotting import curdoc, figure, ColumnDataSource
import random

doc = curdoc()

source = ColumnDataSource({'x': [], 'y': [], 'width': [], 'height': []})
fig = figure()
fig.rect(source=source, x='x', y='y', width='width', height='height',
        alpha=0.01)

k = 100

def f():
    data = {'x': [random.random() for i in range(k)],
            'y': [random.random() for i in range(k)],
            'width': [random.random() for i in range(k)],
            'height': [random.random() for i in range(k)]}
    source.stream(data, 1000000)

doc.add_periodic_callback(f, 100)

doc.add_root(fig)
```

bryevdv:
issues: closes #4077

@mrocklin can you compare/benchmark and see if this PR offers improvement?

@havocp the only thing I was slightly wary of was using a `hint` to short circuit the `unchanged` computation. It seems ok, but if you have any thoughts, please share.

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -420,7 +420,7 @@ class BasicProperty(Property):
         return default
 
     def _real_set(self, obj, old, value, hint=None):
-        unchanged = self.descriptor.matches(value, old)
+        unchanged = self.descriptor.matches(value, old) and (hint is None)
         if unchanged:
             return
 
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -420,6 +420,8 @@ class BasicProperty(Property):
         return default
 
     def _real_set(self, obj, old, value, hint=None):
+        # Currently as of Bokeh 0.11.1, all hinted events modify in place. However this may 
+        # need refining later if this assumption changes. 
         unchanged = self.descriptor.matches(value, old) and (hint is None)
         if unchanged:
             return
--- a/bokeh/core/property_containers.py
+++ b/bokeh/core/property_containers.py
@@ -169,18 +169,15 @@ class PropertyValueDict(PropertyValueContainer, dict):
         for k, v in new_data.items():
             if isinstance(self[k], np.ndarray):
                 data = np.append(self[k], new_data[k])
+                if rollover and len(data) > rollover:
+                    data = data[-rollover:]
+                super(PropertyValueDict, self).__setitem__(k, data)
             else:
-                data = self[k] + new_data[k]
-            if rollover and len(data) > rollover:
-                data = data[-rollover:]
-
-            super(PropertyValueDict, self).__setitem__(k, data)
+                L = self[k]
+                L.extend(new_data[k])
+                del L[:-rollover]
 
         from ..document import ColumnsStreamedEvent
 
         self._notify_owners(old,
                             hint=ColumnsStreamedEvent(doc, source, new_data, rollover))
-
-
-
-
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -100,14 +100,21 @@ class Model(with_metaclass(Viewable, HasProps, CallbackManager)):
         return self._document
 
     def trigger(self, attr, old, new, hint=None):
-        dirty = { 'count' : 0 }
-        def mark_dirty(obj):
-            dirty['count'] += 1
-        if self._document is not None:
-            self._visit_value_and_its_immediate_references(new, mark_dirty)
-            self._visit_value_and_its_immediate_references(old, mark_dirty)
-            if dirty['count'] > 0:
-                self._document._invalidate_all_models()
+        # The explicit assumption here is that hinted events do not
+        # need to go through all the same invalidation steps. Currently
+        # as of Bokeh 0.11.1 the only hinted event is ColumnsStreamedEvent.
+        # This may need to be further refined in the future, if the
+        # assumption does not hold for future hinted events (e.g. the hint
+        # could specify explicitly whether to do normal invalidation or not)
+        if not hint:
+            dirty = { 'count' : 0 }
+            def mark_dirty(obj):
+                dirty['count'] += 1
+            if self._document is not None:
+                self._visit_value_and_its_immediate_references(new, mark_dirty)
+                self._visit_value_and_its_immediate_references(old, mark_dirty)
+                if dirty['count'] > 0:
+                    self._document._invalidate_all_models()
         # chain up to invoke callbacks
         super(Model, self).trigger(attr, old, new, hint)
 
--- a/bokeh/tests/test_document.py
+++ b/bokeh/tests/test_document.py
@@ -321,7 +321,8 @@ class TestDocument(unittest.TestCase):
         assert event.hint.data == dict(a=[11, 12], b=[21, 22])
         assert event.hint.rollover == 200
         assert event.attr == 'data'
-        assert event.old == dict(a=[10], b=[20])
+        # old == new because stream events update in-place
+        assert event.old == dict(a=[10, 11, 12], b=[20, 21, 22])
         assert event.new == dict(a=[10, 11, 12], b=[20, 21, 22])
         assert len(curdoc_from_listener) == 1
         assert curdoc_from_listener[0] is d
--- a/examples/app/ohlc/main.py
+++ b/examples/app/ohlc/main.py
@@ -16,6 +16,7 @@ source = ColumnDataSource(dict(
 p = Figure(plot_height=600, tools="xpan,xwheel_zoom,xbox_zoom,reset", x_axis_type=None)
 p.x_range.follow = "end"
 p.x_range.follow_interval = 100
+p.x_range.range_padding = 0
 
 p.line(x='time', y='average', alpha=0.2, line_width=3, color='navy', source=source)
 p.line(x='time', y='ma', alpha=0.8, line_width=2, color='orange', source=source)


mrocklin:
I ran the minimal benchmark script included in #4077 and found that server-side computational load still increased over time, signaling that maybe this PR doesn't entirely resolve the original issue.  Copying the script here for convenience:

```python
#!/usr/bin/env python

from __future__ import print_function, division, absolute_import

from bokeh.plotting import curdoc, figure, ColumnDataSource
import random

doc = curdoc()

source = ColumnDataSource({'x': [], 'y': [], 'width': [], 'height': []})
fig = figure()
fig.rect(source=source, x='x', y='y', width='width', height='height',
        alpha=0.01)

k = 10

def f():
    data = {'x': [random.random() for i in range(k)],
            'y': [random.random() for i in range(k)],
            'width': [random.random() for i in range(k)],
            'height': [random.random() for i in range(k)]}
    source.stream(data, 1000000)

doc.add_periodic_callback(f, 100)

doc.add_root(fig)
```

I run `bokeh serve script.py --show` and then track CPU usage over time.  It starts low but builds over time.

bryevdv:
OK there is currently one failing unit test. Because we are updating the lists in place, the value of `old` in the document notifications is not "correct", at least not in the sense it normally is. It might be acceptable to document and drop the expectations on `old` in the streaming case, but I would like to verify that the change leads to a positive improvement in update speed or CPU utilization. @mrocklin please advise on this wrt to your use-case when you can.

bryevdv:
@mrocklin I guess I would still expect the CPU load to increase over time, since AFAIK there is no way to the rollover without some copying (the `del L[:-rollover]` has to?) I would expect to shave of some overhead, but not all of the increases. I'm not sure there is anything that can be done about that? Please correct me if I am mistaken about anything here.

bryevdv:
Hrm, that is a bigger CPU loa than I would have expected, though.

bryevdv:
For 100000 rectangles, the render eventually takes considerably more than 100ms, so my first guess is that this benchmark is building up huge back pressure which is not something we do anything about yet (because tornado does not have hooks, AFAIK)

mrocklin:
The above example starts running hot on my machine after a minute, when the size of the list is in the thousands.  Even if it were a million I wouldn't expect the `del` to cost that much.  They're decently cheap.

```python
In [1]: L = list(range(1000000))

In [2]: %time L = L + [1, 2, 3, 4, 5]
CPU times: user 22.2 ms, sys: 3.78 ms, total: 25.9 ms
Wall time: 26 ms

In [3]: %time del L[:5]
CPU times: user 1.34 ms, sys: 160 µs, total: 1.5 ms
Wall time: 1.51 ms
```

(although of course, microbenchmarks lie).  I suspect that there is still copying or something odd going on.

mrocklin:
Where would the backpressure be?  I thought that the bokeh server intentionally did things to avoid building up back pressure with periodic callbacks.

mrocklin:
Regardless, I think that I'll be happy if we can make that example run without burning up the CPU.  I suspect that even if we were to reduce the rectangle count to 1e5 or maybe 1e4 it would still burn hot.

How hard is it to run the Python profiler on the Bokeh server code?  That might point to some interesting spots.

bryevdv:
The backpressure (I'm supposing) is from the fact that the callbacktrys to run roughly every 100ms, but render times were taking twice that or more. 

We definitely need to do some profiling, regardless of how hard it is to do.

bryevdv:
`cProfile` seems easy enough to use, at least, though I am not sure how helpful it is. There are other profiling tools we might want to look into. Anyway, ~30 seconds with `rollover=100000`:

```
Interrupted, shutting down
         21742443 function calls (17589348 primitive calls) in 34.040 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    781/1    0.021    0.000   34.044   34.044 {built-in method exec}
        1    0.000    0.000   34.044   34.044 bokeh:3(<module>)
        1    0.000    0.000   33.237   33.237 bootstrap.py:14(main)
        1    0.000    0.000   33.231   33.231 serve.py:438(invoke)
        1    0.000    0.000   33.226   33.226 server.py:127(start)
        1    0.000    0.000   33.226   33.226 tornado.py:257(start)
        1    0.028    0.028   33.225   33.225 ioloop.py:746(start)
     1324    0.008    0.000   25.340    0.019 kqueue.py:65(poll)
     1394   25.332    0.018   25.332    0.018 {method 'control' of 'select.kqueue' objects}
     1381    0.005    0.000    7.839    0.006 stack_context.py:271(null_wrapper)
     1343    0.003    0.000    7.788    0.006 ioloop.py:594(_run_callback)
 3746/537    0.022    0.000    7.646    0.014 gen.py:257(wrapper)
 1797/743    0.009    0.000    7.609    0.010 gen.py:938(__init__)
 1809/753    0.021    0.000    7.604    0.010 gen.py:990(run)
4194/1518    0.003    0.000    7.587    0.005 {method 'send' of 'generator' objects}
      320    0.002    0.000    7.526    0.024 tornado.py:68(on_done)
      320    0.002    0.000    7.522    0.024 tornado.py:55(invoke)
      320    0.001    0.000    7.495    0.023 session.py:151(wrapped_callback)
     1481    0.006    0.000    7.489    0.005 session.py:31(_needs_document_lock_wrapper)
 1088/448    0.009    0.000    7.447    0.017 document.py:486(_with_self_as_curdoc)
      320    0.001    0.000    7.446    0.023 session.py:145(with_document_locked)
      320    0.001    0.000    7.445    0.023 document.py:497(wrapper)
      320    0.001    0.000    7.438    0.023 document.py:498(invoke)
      320    0.005    0.000    7.437    0.023 foo.py:17(f)
      320    0.008    0.000    7.425    0.023 sources.py:225(stream)
      320    0.008    0.000    7.412    0.023 property_containers.py:164(_stream)
      342    0.001    0.000    7.401    0.022 property_containers.py:41(_notify_owners)
      342    0.002    0.000    7.400    0.022 properties.py:463(_notify_mutated)
      507    0.002    0.000    7.390    0.015 properties.py:422(_real_set)
      495    0.001    0.000    7.387    0.015 properties.py:395(_trigger)
      495    0.004    0.000    7.386    0.015 model.py:102(trigger)
4114963/959    3.765    0.000    7.093    0.007 model.py:195(_visit_value_and_its_immediate_references)
16552833/16552831    3.354    0.000    3.371    0.000 {built-in method isinstance}
```

mrocklin:
Do you have the same result, but sorted by tottime rather than cumtime?

bryevdv:
Sure but also you can just run, e.g. `python -m cProfile -s tottime ~/work/bokeh/bin/bokeh serve --show foo.py`

```
Interrupted, shutting down
         26206600 function calls (21162815 primitive calls) in 37.477 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1355   26.737    0.020   26.737    0.020 {method 'control' of 'select.kqueue' objects}
5005217/1005    4.873    0.000    9.070    0.009 model.py:195(_visit_value_and_its_immediate_references)
20115801/20115799    4.223    0.000    4.241    0.000 {built-in method isinstance}
        7    0.090    0.013    0.090    0.013 {built-in method waitpid}
      663    0.083    0.000    0.083    0.000 {built-in method loads}
     1589    0.051    0.000    0.051    0.000 encoder.py:197(iterencode)
    55/52    0.047    0.001    0.156    0.003 {built-in method load_dynamic}
     8316    0.035    0.000    0.035    0.000 {built-in method stat}
2323/2304    0.034    0.000    0.165    0.000 {built-in method __build_class__}
     4310    0.032    0.000    0.118    0.000 <frozen importlib._bootstrap>:2016(find_spec)
        1    0.029    0.029   36.671   36.671 ioloop.py:746(start)
       96    0.027    0.000    0.027    0.000 {method 'read' of '_io.BufferedReader' objects}
 3842/541    0.024    0.000    9.682    0.018 gen.py:257(wrapper)
      557    0.022    0.000    0.025    0.000 sre_compile.py:255(_optimize_charset)
     1240    0.022    0.000    0.022    0.000 {method 'send' of '_socket.socket' objects}
 1789/727    0.022    0.000    9.638    0.013 gen.py:990(run)
    781/1    0.019    0.000   37.480   37.480 {built-in method exec}
    20078    0.018    0.000    0.020    0.000 sre_parse.py:197(__next)
21668/21598    0.016    0.000    0.017    0.000 {built-in method hasattr}
      663    0.016    0.000    0.022    0.000 <frozen importlib._bootstrap>:1621(get_data)
  934/237    0.015    0.000    0.050    0.000 sre_parse.py:448(_parse)
    23392    0.015    0.000    0.020    0.000 <frozen importlib._bootstrap>:52(<listcomp>)
    23392    0.014    0.000    0.038    0.000 <frozen importlib._bootstrap>:50(_path_join)
     4311    0.012    0.000    0.046    0.000 gen.py:1051(handle_yield)
     1240    0.012    0.000    0.062    0.000 iostream.py:360(write)
      584    0.011    0.000    0.011    0.000 decoder.py:349(raw_decode)
 1777/717    0.010    0.000    9.645    0.013 gen.py:938(__init__)
 1167/461    0.010    0.000    9.472    0.021 document.py:486(_with_self_as_curdoc)
     6047    0.010    0.000    0.018    0.000 abc.py:178(__instancecheck__)
     2466    0.009    0.000    0.169    0.000 message.py:163(send)
      353    0.009    0.000    9.448    0.027 sources.py:225(stream)
     1240    0.009    0.000    0.043    0.000 iostream.py:827(_handle_write)
     1293    0.009    0.000   26.746    0.021 kqueue.py:65(poll)
      353    0.008    0.000    9.435    0.027 property_containers.py:164(_stream)
    21608    0.008    0.000    0.008    0.000 {method 'format' of 'str' objects}
6819/6460    0.007    0.000    0.024    0.000 concurrent.py:313(_set_done)
     5786    0.007    0.000    0.035    0.000 functools.py:726(wrapper)
     2361    0.007    0.000    0.007    0.000 {method 'update' of 'dict' objects}
     1233    0.007    0.000    0.071    0.000 websocket.py:653(_write_frame)
6804/6450    0.007    0.000    0.030    0.000 concurrent.py:264(set_result)
    16623    0.007    0.000    0.008    0.000 {built-in method getattr}
     6826    0.007    0.000    0.007    0.000 concurrent.py:163(__init__)
29313/29221    0.007    0.000    0.009    0.000 {method 'join' of 'str' objects}
     5786    0.007    0.000    0.016    0.000 gen.py:1200(convert_yielded)
      823    0.007    0.000    0.123    0.000 <frozen importlib._bootstrap>:1902(_get_spec)
77855/77021    0.007    0.000    0.007    0.000 {built-in method len}
     5786    0.007    0.000    0.011    0.000 functools.py:688(dispatch)
     6869    0.007    0.000    0.007    0.000 _weakrefset.py:70(__contains__)
      665    0.007    0.000    0.007    0.000 {method 'read' of '_io.FileIO' objects}
```

bryevdv:
Here is `_visit_value_and_its_immediate_references` (which also accounts for all the `isinstance` calls):

https://github.com/bokeh/bokeh/blob/056433ebcdf5cde778241d2bb135301b801357c9/bokeh/model.py#L195-L208

bryevdv:
Thinking out loud, I really think `.stream` shoud be able to bypass most of the "auto-trigger" machinery for causing change updates, but perhaps it is not currently.

mrocklin:
We can treat the `kqueue` stuff as tornado just waiting for something to happen.  The `isinstance` call craziness is interesting and maybe worth investigating.  

Generally though my guess is that when running this computation the CPU wasn't actually running that hot.  Is this accurate?  It might be worth letting it run longer, priming the dataset with 10000 rectangles to start, or tweaking the parameters a bit so that it fills up faster.

I can try to dive in here sometime next week after webinar things are done.

bryevdv:
The code path eventually ends up through here:

https://github.com/bokeh/bokeh/blob/master/bokeh/model.py#L102-L112

It passes the `hint` for the streaming along all the way, but it still does all that invalidating and visiting. Maybe that can be skipped if there is a `hint`. I will try it. Any other thoughts @havocp ?

bryevdv:
OK initial test indicates that makes a huge improvement (CPU for server stays at ~2%) with 100k rectangles. 

So, just need to make sure to do this the best, most robust way possible. Maybe the `hint` could flag whether it is safe to skip all of that extra work in `trigger` (a little bit safer than just assuming it can be skipped any time there is a hint?)

bryevdv:
@mrocklin if yo uwant to play around in the mean time, here is a patch:
```
diff --git a/bokeh/model.py b/bokeh/model.py
index 5a3ba79..d03bac4 100644
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -100,14 +100,15 @@ class Model(with_metaclass(Viewable, HasProps, CallbackManager)):
         return self._document
 
     def trigger(self, attr, old, new, hint=None):
-        dirty = { 'count' : 0 }
-        def mark_dirty(obj):
-            dirty['count'] += 1
-        if self._document is not None:
-            self._visit_value_and_its_immediate_references(new, mark_dirty)
-            self._visit_value_and_its_immediate_references(old, mark_dirty)
-            if dirty['count'] > 0:
-                self._document._invalidate_all_models()
+        if not hint:
+            dirty = { 'count' : 0 }
+            def mark_dirty(obj):
+                dirty['count'] += 1
+            if self._document is not None:
+                self._visit_value_and_its_immediate_references(new, mark_dirty)
+                self._visit_value_and_its_immediate_references(old, mark_dirty)
+                if dirty['count'] > 0:
+                    self._document._invalidate_all_models()
         # chain up to invoke callbacks
         super(Model, self).trigger(attr, old, new, hint)
 
```

bryevdv:
With 1k rects and 100ms update, server is ~2% and Chrome is ~12% on OSX late 2013 MBP. I do think 10s of thousands of things is going to tax almost any browser, and gets into the realm of downsampling yourself, or looking into datashader.

mrocklin:
Makes sense.  I'm curious though, what happens to the server at 10k rects?

bryevdv:
It's just a rule of thumb, in my head. But my experience is you start bumping up against limitation of HTML Canvas implementations. Although it's also possible that the JSON overhead is also a significant contributor. Binary array protocol will help with that, but Canvas performance just is whatever it is.

bryevdv:
For reference, here is `tottime` with 100k rects / 100ms with the patch:
```
Interrupted, shutting down
         1333506 function calls (1288373 primitive calls) in 53.409 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1880   51.428    0.027   51.428    0.027 {method 'control' of 'select.kqueue' objects}
        7    0.093    0.013    0.093    0.013 {built-in method waitpid}
      663    0.086    0.000    0.086    0.000 {built-in method loads}
     2253    0.086    0.000    0.086    0.000 encoder.py:197(iterencode)
    55/52    0.045    0.001    0.161    0.003 {built-in method load_dynamic}
        1    0.041    0.041   52.605   52.605 ioloop.py:746(start)
120273/120271    0.038    0.000    0.061    0.000 {built-in method isinstance}
     8287    0.035    0.000    0.035    0.000 {built-in method stat}
 5373/745    0.035    0.000    0.862    0.001 gen.py:257(wrapper)
     1749    0.035    0.000    0.035    0.000 {method 'send' of '_socket.socket' objects}
2323/2304    0.033    0.000    0.159    0.000 {built-in method __build_class__}
 2457/978    0.032    0.000    0.800    0.001 gen.py:990(run)
     4310    0.031    0.000    0.118    0.000 <frozen importlib._bootstrap>:2016(find_spec)
       96    0.023    0.000    0.023    0.000 {method 'read' of '_io.BufferedReader' objects}
      556    0.023    0.000    0.025    0.000 sre_compile.py:255(_optimize_charset)
      815    0.020    0.000    0.020    0.000 decoder.py:349(raw_decode)
24688/24618    0.019    0.000    0.020    0.000 {built-in method hasattr}
     1749    0.019    0.000    0.096    0.000 iostream.py:360(write)
    781/1    0.019    0.000   53.412   53.412 {built-in method exec}
     6027    0.018    0.000    0.069    0.000 gen.py:1051(handle_yield)
    20065    0.018    0.000    0.020    0.000 sre_parse.py:197(__next)
      663    0.016    0.000    0.023    0.000 <frozen importlib._bootstrap>:1621(get_data)
  932/236    0.015    0.000    0.050    0.000 sre_parse.py:448(_parse)
    23392    0.014    0.000    0.020    0.000 <frozen importlib._bootstrap>:52(<listcomp>)
     3492    0.014    0.000    0.265    0.000 message.py:163(send)
    23392    0.014    0.000    0.037    0.000 <frozen importlib._bootstrap>:50(_path_join)
 2455/977    0.014    0.000    0.825    0.001 gen.py:938(__init__)
      506    0.013    0.000    0.591    0.001 sources.py:225(stream)
     8108    0.013    0.000    0.023    0.000 abc.py:178(__instancecheck__)
     1800    0.013    0.000   51.441    0.029 kqueue.py:65(poll)
```
Still see python hovering in singe digit percentage CPU. But safari is at like 60% or more. Which could either be JSON overhead, canvas performance, maybe some other unecessary copying or work, some combination of all of the above.

mrocklin:
That's awesome.

bryevdv:
OK @mrocklin I have pushed the change that prevents the normal invalidation work from happening on hinted events (e.g. column streaming). I think this is probably about the best we can do on an immediate time frame, though perhaps we can take a deeper dive again later. I would appreciate if you could test with the current PR on your app, and report back at least qualitatively. 

@havocp I added a comment to the last commit explaining the rationale for skipping invalidation on hinted events. I think this is OK and reasonable for now, but your comment would be valuable. 

@bokeh/dev ready for review.

bryevdv:
OK, I added one more comment explaining some logic around hinted events. AFAICT this PR seems reasonable, so I will merge it soon unless there is comment.

mrocklin:
I'll have time to give this a go tomorrow if you want to wait.

bryevdv:
@mrocklin I do want your feedback in any case, but I am going to go ahead and merge this so it can be in the dev build today. I think the changes make sense and do offer some improvement, but your feedback can help us see what next areas to look at might be.

Issue #4065, opened 2016-03-24
The Box (zoom or select) tool should be configurable to respect aspect


bryevdv:
propose `match_aspect = Bool` property on `BoxZoomTool` that, if set restricts the aspect of the box overlay to match the plot frame.

PR #4148, opened 2016-04-09
add option to have box zoom respect plot frame aspect

bryevdv:
issues: #4065

@bokeh/dev please review 

There are a few slightly rough edges. In particular, there seems to be an off by pixel error somewhere, if the aspect controlled box hits up against plot range ends (left or bottom).

But, I think this is important and useful enough as-is to merge and make issues to fix any small things later.

### DIFF
--- a/bokehjs/src/coffee/models/tools/gestures/box_zoom_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/gestures/box_zoom_tool.coffee
@@ -6,6 +6,87 @@ p = require "../../../core/properties"
 
 class BoxZoomToolView extends GestureTool.View
 
+  _match_aspect: (basepoint, curpoint, frame) ->
+
+    # aspect ratio of plot frame
+    hend = frame.get('h_range').get('end')
+    hstart = frame.get('h_range').get('start')
+    vend = frame.get('v_range').get('end')
+    vstart = frame.get('v_range').get('start')
+    w = hend - hstart
+    h = vend - vstart
+    a = w/h
+
+    # if curpoint[0] > hend
+    #   curpoint[0] = hend
+    # if curpoint[0] < hstart
+    #   curpoint[0] = hstart
+    #
+    # if curpoint[1] > vend
+    #   curpoint[1] = vend
+    # if curpoint[1] < vstart
+    #   curpoint[1] = vstart
+
+    # current aspect of cursor-defined box
+    vw = Math.abs(basepoint[0]-curpoint[0])
+    vh = Math.abs(basepoint[1]-curpoint[1])
+    if vh == 0
+      va = 0
+    else
+      va = vw/vh
+
+    if va >= a
+      [xmod, ymod] = [1, va/a]
+    else
+      [xmod, ymod] = [a/va, 1]
+
+    # OK the code blocks below merit some explanation. They do:
+    #
+    # compute left/right, pin to frame if necessary
+    # compute top/bottom (based on new left/right), pin to frame if necessary
+    # recompute left/right (based on top/bottom), in case top/bottom were pinned
+
+    # basepoint[0] is left
+    if ( basepoint[0] <= curpoint[0] )
+      left = basepoint[0]
+      right = basepoint[0] + vw * xmod
+      if right > hend
+        right = hend
+    # basepoint[0] is right
+    else
+      right = basepoint[0]
+      left = basepoint[0] - vw * xmod
+      if left < hstart
+        left = hstart
+
+    vw = Math.abs(right - left)
+
+    # basepoint[1] is bottom
+    if ( basepoint[1] <= curpoint[1] )
+      bottom = basepoint[1]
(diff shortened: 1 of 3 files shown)


bryevdv:
Should mention: run `examples/models/tile_source.py` to test.

bryevdv:
ping @pzwang @jbednar FYI

bokehjs/src/coffee/models/tools/gestures/box_zoom_tool.coffee
```diff
@@ -6,6 +6,87 @@ p = require "../../../core/properties"
...
+
+    # if curpoint[0] > hend
+    #   curpoint[0] = hend
+    # if curpoint[0] < hstart
+    #   curpoint[0] = hstart
+    #
+    # if curpoint[1] > vend
+    #   curpoint[1] = vend
+    # if curpoint[1] < vstart
+    #   curpoint[1] = vstart
```
almarklein: forgot to clean this up?

almarklein:
code LGTM, but when I try it, the selection triangle is drawn in the upper left of the canvas. Tried in FF, Chrome and Edge, all the same issue.

bryevdv:
selection triangle?

bryevdv:
OK I gave up trying to build Bokeh on Windows after 20 minutes of `npm` install failing in myriad ways. @almarklein can you explain more what you mean or give a screen shot? I really have no idea what selection "triangle" could refer to in this case. Also are you making sure to use inline/local BokehJS?

almarklein:
I meant selection rectangle :)  Also, it seemed I needed an `npm install`, I got no warnings or errors, and the JS code did contain your new code so I assumed all was well. It works on FF, Chrome and Edge. Sorry for the noise.

That said, I wonder if things might work more intuitively if you made the selection the smallest rectangle that fits instead of the largest. As it is now, you can drag to make a pretty large rectangle, and moving e.g. upward, it will stay huge, until it flips to the other side...

bryevdv:
That would mean drawing a rectangle that doesn't track the mouse cursor anymore, which I think is more confusing. (imaging: drag up 2 and over 10, the smallest rectangle with fixed aspect is 2x2 but that is nowhere near the mouse) Right now the rectangle always tracks the mouse except it is allowed to diverge when the box abuts the frame border.

bryevdv:
@bokeh/dev any other thoughts on this question?

almarklein:
mmm, yeah, that makes sense. Maybe, ideally, you'd also want to show the rectangle actually being drawn, in a thin line. But let's not hold this back this PR now.

bryevdv:
That also seems potentially confusing, to me at least. But agreed to leave it to another issue/PR. I will merge this in a few hours unless there is other comments.

jbednar:
The user interface for this seems to work well, to me; thanks!

However, is there any less-clunky way to enable it?  I'm used to doing 
`figure(tools='pan,wheel_zoom,box_zoom,reset',...`, and as far as I can tell I now have to muck around specifically with the BoxZoomTool object:

```python
from bokeh.models.tools import BoxZoomTool
fig.add_tools(BoxZoomTool(match_aspect=True))
```

even though I really don't care about the box zoom tool in particular -- all I want to do is be able to declare that my plot has a fixed aspect ratio, and I'd want *everything* to respect that henceforth.  With the above code, the box zoom tool respects the aspect ratio, but the wheel zoom doesn't -- if I happen to do a scroll when over the X or Y axis, the aspect ratio is destroyed despite the above setting.  Isn't there some way we can pass a simple option to the figure that would be respected by each of the various tools?

almarklein:
>  Isn't there some way we can pass a simple option to the figure that would be respected by each of the various tools?

I think that makes a lot of sense. Perhaps an `aspect` property on the plot, that is `None`/`null` by default, meaning a free aspect ratio, or a float to specify the ratio explicitly, which James would set to `1`, but could also be set to  `2`, e.g. to view anisotropic image data.

bryevdv:
@almarklein I think @jbednar is asking for something different. He is asking not to be able to set the aspect ratio, but to be able to set a policy easily on the plot, for tools to automatically match the aspect ratio (whatever it might happen to be).

jbednar:
> He is asking not to be able to set the aspect ratio, but to be able to set a policy easily on the plot, for tools to automatically match the aspect ratio (whatever it might happen to be).

True, I was ambiguous, but actually @almarklein's suggestion would be the most natural solution for my geography-based use cases.  Right now, I have to be very careful when specifying the initial extent of the plot, always calculating the right length to be proportional to the width that I choose so that the result is not distorted.  Having an `aspect` parameter would let me avoid this annoying and very error-prone step, which is part of what I really do want: just to declare that this particular plot should have a specific, fixed aspect ratio, both initially and throughout all user interaction.

almarklein:
Yeah, in my experience, when you want a *fixed* aspect ratio, you also want a *specific* aspect ratio in most cases.

PR #4161, opened 2016-04-11
Enable xul browser to show plots standalone (not in browser)

### LINKED ISSUE #3885: Proposal: remove need for output_file() and display in "native" window

almarklein:
issues: would close #3885 

This is a quick demonstration of what I mean in point 2 of #3885. You can show a plot using `show(p, 'xul')` or `view(.., 'xul')`, and it will end up in a native looking window with a Bokeh icon in the task bar. To test, run the `plotting/file/les_mis.py` example. You need flexx installed (`conda install -b bokeh flexx`).

@bryevdv if you're still -1 on this, we can just close it. It was simple enough to demo it quickly so that I'm sure you know what I mean :)

### DIFF
--- a/bokeh/util/browser.py
+++ b/bokeh/util/browser.py
@@ -5,6 +5,9 @@ import webbrowser
 
 from ..settings import settings
 
+# todo: need this locally
+ICON = 'http://bokeh.pydata.org/en/latest/_static/images/favicon.ico'
+
 def get_browser_controller(browser=None):
     browser = settings.browser(browser)
 
@@ -15,12 +18,20 @@ def get_browser_controller(browser=None):
                     pass
 
             controller = DummyWebBrowser()
+        elif browser.lower() in ('xul', ):
+            from flexx.webruntime import launch
+            class XulWebBrowser(object):
+                def open(self, url, new=0, autoraise=True):
+                    self.c = launch(url,  browser, title='Bokeh plot', icon=ICON)
+
+            controller = XulWebBrowser()
         else:
             controller = webbrowser.get(browser)
     else:
         controller = webbrowser
 
     return controller
+
 
 def view(location, browser=None, new="same", autoraise=True):
         ''' Opens a browser to view the specified location.
--- a/examples/plotting/file/les_mis.py
+++ b/examples/plotting/file/les_mis.py
@@ -64,4 +64,4 @@ p.select_one(HoverTool).tooltips = [
 
 output_file("les_mis.html", title="les_mis.py example")
 
-show(p) # show the plot
+show(p, 'xul') # show the plot


bryevdv:
adding a dependency like Flexx makes it -2 :) 

But, are there hooks we could add that would enable this to be "added" externally?. i.e. what about allowing the ability to "register" a new browser type, along with any function needed to execute things on that browser? Then this kind of thing could be added optionally by anyone that wanted it.

bryevdv:
@almarklein no more "issue PRs", all PRs should link to a regular issue

almarklein:
> adding a dependency like Flexx makes it -2 :) 

Only an optional dependency, of course, just as with the Python-to-JS translation.

A hook could work, though I'm not sure if it's worth the effort, since it would probably take more lines to create a `CustomWebBrowser` class than it would to create a new show function:
 
``` python
def show_xul(*args, **kwargs):
    from flexx.webruntime import launch
    launch(save(*args, **kwargs), 'xul')  # in #3791 `save()` returns the filename

# show(p)
show_xul(p)
```

I guess this is a question of whether we want to support this from Bokeh or not. My take is that it could be useful, e.g. to create more desktop-like apps. If not, it's easy enough for users to use a snippet like the above, so I would not bother with hooks in that case.

almarklein:
BTW: another possible use-case is to avoid spamming new browser tabs during interactive work, because the XUL process terminates when the Python process terminates.

bryevdv:
Let's try to get some additional opinions.

bryevdv:
ping @bokeh/dev any thoughts on this? Given that the desired effect is possible with a couple of lines as demonstrated above, I am -1 on adding more to the core library

bryevdv:
@almarklein I'm going to go ahead and close this since it seems like something that people can do on their own with a few lines of code.

almarklein:
Agreed

PR #4166, opened 2016-04-13
Use setuptools if available in setup.py

### LINKED ISSUE #4119: Try use setuptools in setup.py (so entry points get installed on Windows)
In our `setup.py`, setuptools is only a dependency for our "nightly" build. We have tried hard to not depend on setuptools for the rest. For all the evil that setuptools represents, it does ensure that entry points get installed correctly (i.e. with a .exe) on Windows. 

In our `setup.py` we have `scripts` entry that distutils picks up, installing a script with a shebang (which does not really work on Windows). And we also have an `entry_points` section, that gets ignored by distutils, but picked up by setuptools. Since we don't import setuptools for regular installs, I think that entry does exactly nothing.

It is easy (and quite common, I believe) to add a few lines to `setup.py` so that when setuptools is available, "develop" mode is enabled, and `entry_points` get handled. Since setuptools is a dependency of pip, it means our Windows users using Pip will have a working bokeh command:

```python
try:
    import setuptools
except ImportError:
    pass  #  no "develop" and no entry_points
```

I would propose adding these lines to our script, and possibly remove our own implementation of "develop". It would need proper testing though, since we've tried hard not to depend on setuptools, and I am not sure if our own "tricks" intervene with the tricks of setuptools.

### LINKED ISSUE #4178: Bokeh thinks site_packages is in conda root on Windows
In `setup.py` we have logic that tries to establish the location of `site_packages`, apparently this fails on Windows. In effect, we fail at removing any existing bokeh installs when using `develop`, and the `bokeh.pth` is written in the root of the conda env.

almarklein:
fixes  #4119, fixes #4178

* Try to import setuptools, ignore if not available. If installing with pip (in which case setuptools is available) this will make the entry_points being picked up, and thus the .exe being created on Windows.
* Change how the `packages` is constructed (automatic instead of manually).
* Fix `site_packages` detection on Windows
* Able to cleanup existing bokeh installs that were installed via setuptools, (e.g. `bokeh-0.11.1 blabla`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -200,25 +200,33 @@ def getsitepackages():
 
 
 def check_remove_bokeh_install(site_packages):
+    old_bokeh_files = []
     for d in os.listdir(site_packages):
         bokeh_path = join(site_packages, d)
         if not (d == 'bokeh' or d.startswith('bokeh-')):
             continue
-        prompt = "Found existing bokeh install: %s\nRemove it? [y|N] " % bokeh_path
-        val = input(prompt)
-        if val == "y":
-            print("Removing old bokeh install...", end=" ")
+        old_bokeh_files.append(bokeh_path)
+
+    if len(old_bokeh_files) == 0:
+        return
+
+    print("Found old bokeh files:")
+    for path in old_bokeh_files:
+        print(" - %s" % path)
+    val = input("Remove %s? [y|N] " % ("it" if len(old_bokeh_files)==1 else "them",))
+    if val == "y":
+        print("Removing old bokeh files...", end=" ")
+        for path in old_bokeh_files:
             try:
-                if isdir(bokeh_path): shutil.rmtree(bokeh_path)
-                else: os.remove(bokeh_path)
-                print("Done")
+                if isdir(path): shutil.rmtree(path)
+                else: os.remove(path)
             except (IOError, OSError) as e:
-                print(e)
-                print("\nUnable to remove old bokeh at %s, exiting" % bokeh_path)
+                print("\nUnable to remove old bokeh file at %s, exiting (reason: %s)" % (path, e))
                 sys.exit(-1)
-        else:
-            print("Not removing old bokeh install")
-            sys.exit(1)
+        print("Done")
+    else:
+        print("Not removing old bokeh install")
+        sys.exit(1)
 
 
 def remove_bokeh_pth(path_file):
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ except ImportError:
 
 if 'nightly' in sys.argv:
     from setuptools import setup
+
     sys.argv.remove('nightly')
 
     with open('__conda_version__.txt', 'r') as f:
@@ -406,6 +407,16 @@ def parse_jsargs():
 
     return jsbuild
 
+
+def package_tree(pkgroot):
+    """ Get list of packages by walking the directory structure and
+    including all dirs that have an __init__.py or are named test.
+    """
+    subdirs = [os.path.relpath(i[0], ROOT).replace(os.path.sep, '.')
+               for i in os.walk(os.path.join(ROOT, pkgroot))
+               if '__init__.py' in i[2] or os.path.split(i[0])[1] == 'tests']
+    return subdirs
+
 # -----------------------------------------------------------------------------
 # Main script
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -401,6 +407,16 @@ def parse_jsargs():
...
     return jsbuild
 
+
+def package_tree(pkgroot):
+    """ Get list of packages by walking the directory structure and
+    including all dirs that have an __init__.py or are named test.
+    """
+    subdirs = [os.path.relpath(i[0], ROOT).replace(os.path.sep, '.')
+               for i in os.walk(os.path.join(ROOT, pkgroot))
+               if '__init__.py' in i[2] or os.path.split(i[0])[1] == 'tests']
```
bryevdv: I think we should just put `__init__.py` in all the `test directories. There was some talk about it having issues with `py.test` but lots of them already do, and it seems to have no ill effect at all. ping @birdsarah

setup.py
```diff
@@ -20,6 +20,11 @@
 from shutil import copy
 
 try:
+    import setuptools
+except ImportError:
+    pass  # setuptools allows for "develop" and entry_points, but it's not essential
```
bryevdv: So is this only needed really on windows, and for `setup.py install` ?

almarklein: yes, this is enough for setuptools to latch itself to distutils (in the dirty way that it does this), so that `setup()` will know about entry points et. al.

bryevdv: Sorry my question is: should we do this unconditionally all the time, or restrict to only when we know we need it (i.e. installs on windows)

almarklein: Good point. Since we implement `develop` ourselves, it probably makes sense to restrict it yes.

almarklein:
@bryevdv points addressed and fixed two other issues that I encountered.

bryevdv:
LGTM will test on OS X tomorrow. ping @damianavila

bryevdv:
I think there might be issues, I get this doing a `develop`
```
Found existing bokeh install: /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.11.0dev4_33_gac8fda9-py3.4.egg-info
Remove it? [y|N] y
Removing old bokeh install... Unable to remove old bokeh at /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.11.0dev4_33_gac8fda9-py3.4.egg-info, exiting
bryan@0199-bryanv (git:almarklein/setuptools) ~/work/bokeh $ rm /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.11.0dev4_33_gac8fda9-py3.4.egg-info
```

bryevdv:
If I run again after removing by hand, it happens again immediately.

bryevdv:
OK the second time was another old egg info file, but why is the `rm` in the script failing? I can remove them by hand no issue.

bryevdv:
Yah another `develop` after an `install` yields this:
```
bryan@0199-bryanv (git:almarklein/setuptools) ~/work/bokeh $ python setup.py develop --install-js

Found existing bokeh install: /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh
Remove it? [y|N] y
Removing old bokeh install... Done
Found existing bokeh install: /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.12.0dev6_175_g216cd0d-py3.4.egg-info
Remove it? [y|N] y
Removing old bokeh install... Unable to remove old bokeh at /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.12.0dev6_175_g216cd0d-py3.4.egg-info, exiting
```

I'd say there are two things:

* would be better to collect the list of files to delete first, and present a single list, and `[y/N]` prompt to delete them all
* need to make the deletion actually work again

bryevdv:
It's failing because `rmtree` only expects directories, and you are now also passing it egg files.

bryevdv:
@almarklein the changes I suggested were small, so I just implemented them quickly. Now see, e.g.:
```
Found old bokeh files:
 - /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh
 - /Users/bryan/anaconda/lib/python3.4/site-packages/bokeh-0.12.0dev6_176_ge16b9c5_dirty-py3.4.egg-info
Remove them? [y|N] y
```
this LGTM whenever you want to merge

damianavila:
This one LGTM... I really like to **only** use setuptools for win and I really love the auto detection of the packages tree...

bryevdv:
OK I will merge when CI is green.

almarklein:
Thanks for the fixes @bryevdv

PR #4167, opened 2016-04-14
Fix previewSaveTool' modal view

### LINKED ISSUE #4058: PreviewSaveTool's modal is broken after adding .bk-root
Modal is added to the body, not under `.bk-root`, so styles aren't applied.

### LINKED ISSUE #4081: Preview/Save Widget on wrong layer
On Docs side when selecting Preview/Save Widget od plot preview is hidden behind text (probably wrong z-index).
Link:
http://bokeh.pydata.org/en/latest/docs/user_guide/quickstart.html

almarklein:
Fixes #4058 and fixes #4081, by attaching the tool's element to the DOM under the plot' element. 

Would perhaps be a bit nicer if we could put it directly under `bk-root` instead of under the plot, but the plot is not attached to the DOM at the time that the tool initializes, and fiddling with timeouts seems too much fuss. 

cc @mattpap

### DIFF
--- a/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
@@ -27,7 +27,7 @@ class PreviewSaveToolView extends ActionTool.View
     @$('.bk-bs-modal-body img').attr("src", canvas.toDataURL());
     @$el.modal('show')
     # Bootstrap creates backdrop DOM element in the body, we move it to
-    # somwhere inder bk-root so that css works it should.
+    # somwhere under bk-root so that css works as it should.
     bd = document.getElementsByClassName('bk-bs-modal-backdrop')[0]
     @plot_view.el.appendChild(bd)
 
--- a/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
@@ -11,6 +11,8 @@ class PreviewSaveToolView extends ActionTool.View
 
   initialize: (options) ->
     super(options)
+    # Attach to DOM (should be somewhere under bk-root to make css work)
+    @plot_view.el.appendChild(@el)
     @render()
 
   render: () ->
--- a/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/actions/preview_save_tool.coffee
@@ -26,6 +26,10 @@ class PreviewSaveToolView extends ActionTool.View
     canvas = @plot_view.canvas_view.canvas[0]
     @$('.bk-bs-modal-body img').attr("src", canvas.toDataURL());
     @$el.modal('show')
+    # Bootstrap creates backdrop DOM element in the body, we move it to
+    # somwhere inder bk-root so that css works it should.
+    bd = document.getElementsByClassName('bk-bs-modal-backdrop')[0]
+    @plot_view.el.appendChild(bd)
 
 class PreviewSaveTool extends ActionTool.Model
   default_view: PreviewSaveToolView


mattpap:
I doubt this is sufficient, because modal's backdrop should attach to `body` anyway.

almarklein:
mmm, good point. I had ignored the backdrop. So do you think the DOM elements for the tools should be split in two pieces, or should the CSS be changed to make it work again?

almarklein:
@mattpap I am confused. In `d:\dev\pylib\bokeh\bokehjs\src\vendor\bootstrap-3.1.1\less\modals.less` the styling is defined. I though that the `bk-` prefix was for Bokeh, but this is vendored code?

Am I correct that `.bk-bs-modal` is made into `.bk-root .bk-bs-modal` when the `.less` is converted to `.css`?

mattpap:
All code under `src/vendor` was manually prefixed.

> Am I correct that `.bk-bs-modal` is made into `.bk-root .bk-bs-modal` when the

Yes.

almarklein:
> I doubt this is sufficient, because modal's backdrop should attach to body anyway.

Wait... I cannot find any reference to backdrop except for in the `.less`. It seems that the tool has no DOM elements for this. FWIW, the tool works as it did before, except that the background is not grayed out.

almarklein:
This should do the trick (tested to work on Windows FF/Chrome/Edge).

bryevdv:
@mattpap @birdsarah I will defer to you two to evaluate this PR

bryevdv:
OK have not seen any feedback so I checked out the branch and tested it. Works great in a file and server app. In the notebook I see:

screen shot 2016-04-20 at 11 20 44 am

almarklein:
z-ordering? For some reason Bokeh plots in the notebook seems broken on my system somehow, so I cannot reproduce at this point.

bryevdv:
are you using inline resources?

almarklein:
I tried several resource modes, but keep getting the same problem. When the `output_notebook()` cell runs, it just says `Loading BokehJS ...`. When a plot is `show()`n, it says in th JS console: `  Error: Class 78677579-106d-4f24-a44f-aa2b625cb81a not found in registry `.

bryevdv:
I tried manually increasing the `z-index` in the browser dev tools and could not get it to render correctly above the notebook cell text.

bryevdv:
@bokeh/dev any suggestions?

birdsarah:
Total guess - nested z-indexes causing the problem.

Can I suggest a somewhat drastic alternative fix - get rid of the modal and just have it so when you click on the button it downloads the image (opens the browser's native save dialog).

Here's how to do that:
```
  do: () ->
    canvas = @plot_view.canvas_view.canvas[0]
    downloadLink = document.createElement("a")
    downloadLink.download = "bokeh.png"
    downloadLink.innerHTML = "Download png"
    downloadLink.href = canvas.toDataURL()
    downloadLink.onclick = (event) -> document.body.removeChild(event.target)
    downloadLink.style.display = "none"
    document.body.appendChild(downloadLink)
    downloadLink.click()
```

bryevdv:
I would be happy to ditch the modal. This is essentially what #4228 does, but the implementation there is slightly different from what you have above, and does not work on Safari. Do you know if the above code works on safari? If so, perhaps a comment on #4228 would be helpful.

birdsarah:
missed #4228 - thanks.

bryevdv:
If we are agreed to ditch the modal (and jetison some 3rd party library code for it?) can we just close this PR unmerged?

mattpap:
> If we are agreed to ditch the modal (and jetison some 3rd party library code for it?)

Including my work on tools, where I will try to remove need for bootstrap to render dropdowns, we could finally remove bootstrap from core bokehjs.

bryevdv:
> Including my work on tools, where I will try to remove need for bootstrap to render dropdowns, we could finally remove bootstrap from core bokehjs. 

That would be amazing. I will close this PR unmarked tomorrow unless there is comment or objection.

PR #4169, opened 2016-04-14
Add custom code quality analytics

### LINKED ISSUE #4168: Apply basic code quality rules to all source files
Currently we only inspect `*.py` files, but leave `*.coffee`, `*.ts`, `*.rst`, .., ..., ... alone.

mattpap:
fixes #4168

- [x] add custom code quality checks
- [x] exclude `bokeh/server/static/{js,css}`
- [ ] ~~use `.gitignore` to exclude files, e.g. `bokehjs/examples/*/*.js`~~
- [x] fix all errors
- [ ] ~~clean up `setup.cfg`~~


bryevdv:
@Maggie-M can  probably help with the "fix all errors" part once the script is working fully.

mattpap:
> use .gitignore to exclude files, e.g. `bokehjs/examples/*/*.js`

That would be fun, but for now it's too much work.

mattpap:
I had to disable max_line_length test, because there are too many failures to fix now. Additionally, there are cases where long lines are acceptable, e.g. inline base64-encoded images in less files (thought those should be replaced with proper tooling that less gives us).

mattpap:
@bokeh/dev, please review this today.

almarklein:
LGTM

However, we are already using flake8, for which we currently ignore [all errors and warnings except a few](https://github.com/bokeh/bokeh/blob/master/setup.cfg#L14-L15). Why not simply include more tests there, or remove the global excludes and exclude all the tests that we don't care about?

>  had to disable max_line_length test, because there are too many failures to fix now.

We have the max line length set to 160 for now.

mattpap:
@almarklein, flake8 works only (as far I can tell) with python files. This is the reason for this PR. Ultimately we will have one set of rules across all files and then use domain specific tools (like flake8).

bryevdv:
@mattpap it's checking (and failing) `bokeh/__conda_version__.py` which should probably be excluded:

https://travis-ci.org/bokeh/bokeh/jobs/134439611#L2571

bryevdv:
I made the change, will merge when green.

bryevdv:
@mattpap OK just adding the file to the exclusions did not work, it seems the exclusions have to be directories?

birdsarah:
Only had the opportunity for a quick look as I'm at pycon, but looks good. Thanks for doing this.

bryevdv:
OK, added support for file exclusions. I also added a main check at the end so it can just be run, my local pytest is completely borked and I can't spend any more time trying to fix it today.

bryevdv:
Tests appear to be passing now. Will merge when green. The file exclusion stuff could probably be removed when/if you come back to add `.gitignore` support, but it should do for now. As you said, the PR should be merged sooner rather than later.

PR #4171, opened 2016-04-14
Don't render twice after selection

### LINKED ISSUE #4170: bokehjs renders twice after selection
This is because `GlyphRenderer` requests to render on data source's data changes and `select` event, which notify about the same thing happening.

mattpap:
fixes #4170

### DIFF
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -55,7 +55,12 @@ class GlyphRendererView extends Renderer.View
     @listenTo(@model, 'change', @request_render)
     @listenTo(@mget('data_source'), 'change', @set_data)
     @listenTo(@mget('data_source'), 'stream', @set_data)
-    @listenTo(@mget('data_source'), 'select', @request_render)
+
+    # XXX: you might be tempted to enable the next line, but that
+    # will duplicate rendering effort, because `request_render()`
+    # is already triggered by `set_data()` under `change` event.
+    #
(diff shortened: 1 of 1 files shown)


bryevdv:
I'm a little confused. The whole point of `request_render` is that it *doesn't* actually trigger an immediate  render, merely requests that one be triggered. With the idea that multiple request within one render throttling interval still only result in one render. So, the question really is: Is the "request" part somehow broken and `request_render` is immediately, synchronously triggering a full render? Or is there just some timing coincidence that puts the two requests in different throttling intervals?

birdsarah:
Appreciate the helpful note in the code here.

This seems like something we could use a test for. It's easy to say that, but I know I don't know how to write it. If it was in python I'd use mock and test the number of calls to render, in bokehjs I don't know. I've used sinon to do a little mocking, but I've never figured out how to instantiate & test views in a way that isn't creating more headaches in the test than in the code. This is a testing challenge I'd love to be solved for bokehjs. I've mentioned to @bryevdv before. Do you have any thoughts on this? I know it's a fairly heavy thing to put on essentially a one-line PR, but it seems like a nicely isolated thing, which is why I thought I'd mention it here.

bryevdv:
Another thought. Can this: 
```
@listenTo(@mget('data_source'), 'change', @set_data)
```
Be made mode specific, so that changes to selections don't trigger `set_data`? That seems like a definite improvement, since it seem currently lots of extra `set_data` work is done on every selection. However, my question about why repeated calls to `request_render` are triggering multiple renders still stands.

bryevdv:
superseded by #4442

PR #4174, opened 2016-04-14
Add serve subcommand to make.bat in Sphinx docs

### LINKED ISSUE #4173: There is no make.bat serve to view locally build docs on Windows
In the Sphynx docs, the Makefile provides a `serve` subcommand, but the `make.bat` does not. Normally I'd just view the HTML file directly, but because some resources are referenced using a double-slashes-prefix (e.g. `//bokehplots.com/...`) the docs don't really work when viewed that way.

almarklein:
Fixes #4173

### DIFF
--- a/sphinx/make.bat
+++ b/sphinx/make.bat
@@ -194,4 +194,11 @@ results in %BUILDDIR%/doctest/output.txt.
 	goto end
 )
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Spurious failure, merging now.

PR #4177, opened 2016-04-14
Add 'div' widget to interaction.rst, create example

### LINKED ISSUE #4122: Add new div widget example to user guide
Widget was added in #4021

Maggie-M:
issues: fixes #4122 

Though a lot of the text could use some work.


bryevdv:
Spurious test failures, but this is docs only, so merging now

PR #4181, opened 2016-04-15
Set defaults to none so that the HTML will render with auto unless sp…

### LINKED ISSUE #4180: Set default width and height for Markup widgets to None
Currently, the Markup widgets use a default width and height of several hundred pixels.  I recommend that we change that to None by default which will not transmit any width or height cues to the browser.  The user can still use the properties if he or she wants, they would just by default be set to auto in the HTML.  I think this would be the expected behavior from the end users perspective.

justacec:
This is a simple PR to change the default settings for the width and the height of the various Markup widgets (Paragraph, Div) to "None".  This change will result in generated HTML with no specification for width or height.  If the user wants to have a specific size, for layout purposes, he or she can specify them when they instantiate, or update the object model.

- [x] issues: fixes #4180 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/markup.coffee
+++ b/bokehjs/src/coffee/models/widgets/markup.coffee
@@ -24,8 +24,8 @@ class Markup extends Widget.Model
 
   @define {
     text: [ p.String, '' ]
-    width: [ p.Number, 500]
-    height: [ p.Number, 400]
+    width: [ p.Number, none]
+    height: [ p.Number, none]
   }
 
 module.exports =
--- a/bokeh/models/widgets/markups.py
+++ b/bokeh/models/widgets/markups.py
@@ -15,11 +15,11 @@ class Markup(Widget):
     The contents of the widget.
     """)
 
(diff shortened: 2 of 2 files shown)


bokehjs/src/coffee/models/widgets/markup.coffee
```diff
@@ -24,8 +24,8 @@ class Markup extends Widget.Model
 
   @define {
     text: [ p.String, '' ]
-    width: [ p.Number, 500]
-    height: [ p.Number, 400]
+    width: [ p.Number, none]
+    height: [ p.Number, none]
```
bryevdv: There is no `none` on the JS side, you can just omit the second element completely and it will default to `null` (which is what is correct)

bryevdv:
@birdsarah This should probably be closed, in favor of the ongoing work you are doing?

birdsarah:
Yes, they just got set to None in my latest PR (still a WIP)

birdsarah:
#4308

PR #4189, opened 2016-04-19
Force JS execution after page/DOM fully loads

### LINKED ISSUE #4188: Bokeh + Flask Causes Race Condition
The discussion started in the mailing list:

https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/lzj1-P5VKTs

Essentially, when using autoload_static() in Flask appears to cause a race condition where the javascript is executed before the page/DOM is fully loaded. This results in the following error:

Bokeh: ERROR: autoload.js configured with elementid '3b44507e-4010-416a-836c-fa4fd18ac117' but no matching script tag was found. 

One proposed solution is to wrap the javascript inside of some sort of "onload" function but I am no JS expert.

seanlaw:

- [x] issues: fixes #4188 
- [x] tests added / passed

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -25,94 +25,96 @@ calls it with the rendered model.
 
 
 #}
-window.onload = (function(global) {
-  function now() {
-    return new Date();
-  }
-
-  if (typeof (window._bokeh_onload_callbacks) === "undefined") {
-    window._bokeh_onload_callbacks = [];
-  }
-
-  function run_callbacks() {
-    window._bokeh_onload_callbacks.forEach(function(callback) { callback() });
-    delete window._bokeh_onload_callbacks
-    console.info("Bokeh: all callbacks have finished");
-  }
-
-  function load_libs(js_urls, callback) {
-    window._bokeh_onload_callbacks.push(callback);
-    if (window._bokeh_is_loading > 0) {
-      console.log("Bokeh: BokehJS is being loaded, scheduling callback at", now());
-      return null;
+document.addEventListener("DOMContentLoaded", function(event) { 
+  (function(global) {
+    function now() {
+      return new Date();
     }
-    if (js_urls == null || js_urls.length === 0) {
-      run_callbacks();
-      return null;
+
+    if (typeof (window._bokeh_onload_callbacks) === "undefined") {
+      window._bokeh_onload_callbacks = [];
     }
-    console.log("Bokeh: BokehJS not loaded, scheduling load and callback at", now());
-    window._bokeh_is_loading = js_urls.length;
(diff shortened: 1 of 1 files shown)


bryevdv:
@bokeh/dev This seems a reasonable and straightforward solution to the problem in the linked issue, but maybe there are other considerations or approaches, comments appreciated.

mattpap:
This seems either unnecessary or misplaced, but I need to give this some thought.

mattpap:
In any case, this shouldn't use `window.onload`, but `addEventListender` instead.

bryevdv:
For context it appears that at least on some browsers, the script executes
before the DOM is done loading. This cause the code that looks up its own
script tag to fail. Maybe we accidentally go away with something for a
while, but it seems true that the script needs to execute only after the
DOM loads. So I think we have to do something to ensure that. Agreed
`addEventListener` is probably better, I forgot about that.

Sean, here is a reference for that:
http://www.w3schools.com/jsref/event_onload.asp

seanlaw:
Cool. I found (and tested) another solution using `addEventListener` that simply wraps the script in this way:

document.addEventListener("DOMContentLoaded", function(event) { 
    //DOM is loaded and ready
    //Run Bokeh anonymous function here
});

Bryan, you'll have to guide me through what to do next. Do I simply create a new branch again but with the new fix and submit a PR again? Should I wait for more comments before solving the problem multiple different ways?

bryevdv:
@seanlaw No, GH is really nice in this regard. If you just add/commit the new changes to this branch on your local checkout, and then push again, this PR will be automatically updated.

bryevdv:
Also FYI the one test failure is spurious and unrelated.

seanlaw:
Made the new change. See if that works or is better?

bryevdv:
@seanlaw Thanks! Yah I noticed the annoying diff, but I can see it is just the small change. GH diffs are usually fairly good but sometimes they can appear messy under some circumstances. 

ping @birdsarah @havocp @mattpap comments on this?

bryevdv:
Verbal conversation with @birdsarah, we can't think of any reasons not to pursue this approach. @mattpap if you still have concerns please open another issue to discuss them. I will merge this went Travis turns green. Thanks for the contribution @seanlaw !

bryevdv:
as a last check I built the docs and inspected the gallery, which uses `autoload_static` throughout, and everything works as expected.

seanlaw:
Amazing! Thanks for pushing it through. I'm guessing this makes me an intro-level bokeh contributor rather than a bokeh developer? :)

bryevdv:
Unfortunately I think this broke notebook support (I was not aware that notebook `show` used this same code path, it did not used to). I will open a new issue to investigate

PR #4192, opened 2016-04-20
Improve Bokeh Layout

### LINKED ISSUE #4164: Improving Bokeh's Layout
## Background

The over-arching motivation is that when you put multiple plots and widgets on a page it should look "nice" - things should line up, whitespace should be sensible. This is somewhat vague but is the big picture.

Furthermore, layouts should, at the least, have the capability to be responsive (that is the elements resize based on the browser size) - a norm in the modern web. I would suggest responsive should be the default, although it should be easy to turn off.

The issue tracker has a host of issues which relate to layout in a variety of different ways from wanting titles to be properly layoutable objects, to wanting to constrain aspect ratios, to bugs with the current responsive behavior, vbox/hbox, min_border being broken to name a few. Issues I know about are listed below.

To get this work started, @havocp created a layout algorithm and example, which can be seen here: https://cdn.rawgit.com/birdsarah/bokeh-miscellany/master/layout-demo/index.html

At the heart of Havoc's work is a top-down approach where constraints are pulled up to the parent/document/root level and then everything is solved at once. This allows, for example, for the left-axis of one plot to align with the left-axis of another plot - even if they have widely differing axes labels.

I have since spent some time prototyping this to work with Bokeh plots, widgets etc. https://github.com/bokeh/bokeh/pull/4055, https://github.com/bokeh/bokeh/pull/4067, https://github.com/bokeh/bokeh/pull/4137

## Proposed work

With this background in place, I propose the following work.

#### Introducing Box, Row, and Column

Box, Row and Column are introduced. A row is a horizontal box. A column is a vertical box. Rows and Columns can hold plots, widgets or more rows and columns. Allowing for a highly composable layout where things just line up.

These would likely replace HBox/VBox/hplot/vplot.

#### Re-factor solver related code to be consistent.
Reasoning about / debugging the solver is hard (or at least I find it to be). This is further compounded by lots of little pieces of code that use it but in different ways.

Havoc's layout-algorithm works from a top-down approach where layoutable models have children and pass up their constraints through a `get_constraints` method and a `get_constrained_variables` method.

- Constraints are gathered by the document on add_root [See example here](https://github.com/bokeh/bokeh/blob/birdsarah/scratch_new_layout_8_5_make_plot_layout/bokehjs/src/coffee/document.coffee#L132)

- Models that have layout. Will have a `get_constrained_variables` and a `get_constraints` method on their Model.

- Some models need to be updated during render / re-render. As a result, models like this will have an `update_constraints` method on their View.
 
#### LayoutDom and LayoutCanvas
Currently we have a LayoutBox which helps objects lay themselves out on the canvas and a much more adhoc set of things that layout items in the dom.

- LayoutBox -> LayoutCanvas - There end up being too many references to boxes floating around. Re-naming to LayoutCanvas makes it clear what objects are laying themselves out on the canvas.

- LayoutDom - Similar to LayoutCanvas this has a bunch of constraints and solver variables. It allows things to lay themselves out on the webpage. In particular the plot canvas and widgets use it. 

#### Plot becomes a Box, Toolbar becomes a widget.

We need to do this without breaking the API, that's a given, I'm talking in conceptual terms here. What we currently call a plot is currently a somewhat messy html template with <canvas> at the heart of it and then space round each side for a toolbar. Conceptually we would like a PlotCanvas to which we can add a toolbar above|below|left|right. 

The toolbar can easily just be a widget - an html element that needs to be laid out.

This allows simplification of the plot template (no more table layout). It also means that the toolbar can be put anywhere on the page.

The plot decides whether it's a horizontal box or a vertical box depending on whether the toolbar goes on top or bottom.

It also means that by default a plot would fill to be the size of the page

#### Title becomes an annotation

Full discussion on #4307

~~Title will become a first class citizen using a LayoutCanvas and @lukec's new labal annotation. This should mean that it can be added above|below|left|right and get any orientation etc it wants.~~

#### All the current resize code on Plot is removed

The current resize code on Plot is removed in favor of the document-level solver handling the size. 

## Results of prototyping

I have prototyped all the above except for the Title (which I see as very low-risk). The result, aside from having aligned, responsive, nested layouts, is, I believe:
- simpler code overall
- less types of things (they're either Box, LayoutDom, or LayoutCanvas)
- easier to reason about layouts 

Although I've worked out a bunch of things, and I feel good about this direction, there's still a fairly long list of things that will be mini hurdles along the way: griplot, components, server, the resize tool, deprecation

I didn't try implementing aspect ratios, and I would propose that this can be a second step after the initial round of work, but I can't imagine it being difficult.

## API Questions / Proposals

* We add a `add_title` method to Plot which makes a Title and uses add_layout to add it onto the plot. If your using charts or figure nothing will change and we'll call this method for you.

* We deprecate HBox/VBox/vplot/hplot by putting them under layouts_legacy, in much the same way we did charts.

* This new world makes everything fluid by default. Propose that we keep this and if you want a fixed size you specify the size with plot_height, plot_width, widget_height, widget_width. (I haven't prototyped this but it should work). This means many users may not see a change if they've specified a plot_height/width. Others who haven't will see all their plots grow.

* I can imagine many more configuration options that users might want for the layout. But I think this bare minimum gives us layout, and moves us forward without affecting users much, and then we can add on new options like fixed aspect ratio, after the fact.

## Concerns

This is going to be a large PR. I'm not sure what to do about that. I can't think of a way to make this change more atomic without making it take a lot longer to build.

I believe the resize tool could be very tricky. I'd like to leave the door open to turn if off for non-fixed size plots at least in the short-term. 

---

Current open issues that are relevant to this work, grouped by theme:

The issues are listed here because they:
* will get fixed as a result of this work
* will get partially fixed as a result of this work
* should be taken into account when doing this work

* Broken layouts
 * https://github.com/bokeh/bokeh/issues/3303 - Layout can allow plots to overlap
 * https://github.com/bokeh/bokeh/issues/2504 - Datatables spills out on server rendering (on FF)
 * https://github.com/bokeh/bokeh/issues/1744 - Hover panel hides under neighboring gridplot figures
 * https://github.com/bokeh/bokeh/issues/2891 - Support multiple None entries per row in GridPlot
 * https://github.com/bokeh/bokeh/issues/2350 - HBox/hplot not working in IPython Notebook
 * https://github.com/bokeh/bokeh/issues/2284 - HBox not working properly with plots
 * https://github.com/bokeh/bokeh/issues/1256 - VBox doesn't work properly under IE
 * https://github.com/bokeh/bokeh/issues/4337 - Plot Blank if Categorical Label too long 

* HBox/VBox/hplot/vplot/... and the documnet
 * https://github.com/bokeh/bokeh/issues/3650 - Discussion: Moving to document-unaware layouts (HBox, VBox, Tabs)
 * https://github.com/bokeh/bokeh/issues/3074 - curdoc, vplot, hplot role with new Document API
 * ~~https://github.com/bokeh/bokeh/issues/3563 - Setting min_border does not work~~

* Styling (sizing) of widgets
 * https://github.com/bokeh/bokeh/issues/3346 - Slider size from python code to css
 * https://github.com/bokeh/bokeh/issues/2374 - Improvements to aid styling of widgets
 * https://github.com/bokeh/bokeh/issues/2344 - VBox: width has no effect (Firefox)
 * https://github.com/bokeh/bokeh/issues/1710 - App drop-downs under Firefox are the wrong size
 * https://github.com/bokeh/bokeh/issues/2451 - css - hplot of Paragraph & TextInput

* Titles
 * https://github.com/bokeh/bokeh/issues/1449 - GridPlot ignores `title` property
 * https://github.com/bokeh/bokeh/issues/149 - Allow plot titles to be located on any edge of the plot (not just the top)
 * https://github.com/bokeh/bokeh/issues/2081 - title_text_align plot property behaves strangely
 * https://github.com/bokeh/bokeh/issues/1085 - Orientation on titles

* GridPlot
 * https://github.com/bokeh/bokeh/issues/3985 - Bokeh Logo with GridPlot
 * https://github.com/bokeh/bokeh/issues/1448 - GridPlot could allow 1d child sequence together with (n, m) tuple
 * https://github.com/bokeh/bokeh/issues/2398 - HelpTool doesn't work with GridPlot

* Aspect Ratio
 * https://github.com/bokeh/bokeh/issues/4065 - The Box (zoom or select) tool should be configurable to respect aspect
 * https://github.com/bokeh/bokeh/issues/1196 - Reset and resize tools do not work with GMapPlot
 * https://github.com/bokeh/bokeh/issues/474 - Easily control aspect ratio
 * https://github.com/bokeh/bokeh/issues/2861 - GMap improvements
 * https://github.com/bokeh/bokeh/issues/3148 - Geo ToDo
 * https://github.com/bokeh/bokeh/issues/3012 - Additional Features / Fixes for TileRenderer / TileSource

* Responsive
 * https://github.com/bokeh/bokeh/issues/3783 - Reset should call the responsive resize.
 * https://github.com/bokeh/bokeh/issues/3061 - Setting responsive for figures in tabs results in buggy tabs
 * https://github.com/bokeh/bokeh/issues/2920 - Make plots responsive only in x or y 

* Toolbar
 * https://github.com/bokeh/bokeh/issues/3819 - Toolbar Improvements: Make tools independent of bar
 * https://github.com/bokeh/bokeh/issues/3818 - Toolbar Improvements: Regroup Tools
 * https://github.com/bokeh/bokeh/issues/3804 - support for multiple tap tools
 * https://github.com/bokeh/bokeh/issues/3436 - Tools UX enhancement
 * https://github.com/bokeh/bokeh/issues/3340 - Investigate better implementation for toolbar rendering
 * https://github.com/bokeh/bokeh/issues/3250 - Tool icons should have the same dimensions

* Misc
 * https://github.com/bokeh/bokeh/issues/2865 - Off-canvas legends
 * https://github.com/bokeh/bokeh/issues/113 - Axis should have option to have a fixed orientation
 * https://github.com/bokeh/bokeh/issues/2006 - Unable to Stack Multiple Twin Axis on Same Side
 * https://github.com/bokeh/bokeh/issues/3893 - Does components() in bokeh.embed work for widgets?
 * ~~https://github.com/bokeh/bokeh/issues/2759  - Resolve misleading property names in~~

birdsarah:
Issues: fixes #4164

Things to add back that have been removed early:
- [ ] Title (should be a proper renderer - see deleted commend from bev bokeh/models/plots.py ~L434)
- [ ] Padding

Things not to forget:
- [ ] Clean-up reference to panel on frame. Needed because of loop that adds axes to the sides of frame and each other in add_constraints in plot.
- [ ] Responsive flag
- [ ] Resize tool


birdsarah:
ping @mattpap - here's the first commit!

birdsarah:
Closing this PR. #4203 is proving more productive than initially anticipated. Further PRs will follow off that branch.

PR #4198, opened 2016-04-20
Histogram deals better with ill-defined ranges (#3660)

### LINKED ISSUE #3660: Histogram empty plot with negative values
We observed that plotting of an unique negative value vector does print an empty Histogram. This is quite an artificial test case, however we would like to report it.

```
from bokeh.charts import Histogram, output_file, show, vplot, hplot, Bar, defaults

arr = [-21, -21, -21, -21, -21]

output_file("Hist.html")
p1 = Histogram(arr, title="Histogram", xlabel="Distances", ylabel="Counts", bins=2)
p1.x_range = Range1d(-25, 25) 
         
p = vplot(hplot(p1))
show(p)
```

almarklein:
Fixes #3660 

I played with different approaches to deal with this problem, but this seemed to be the least intrusive. The problem can be considered an upsteam problem of pandas: Pandas selects bin bounds that are 1% wider than the actual min/max range, to ensure inclusion of all values. When the data is no range (e.g. `[3, 3, 3]`), the bin width seems to be 1% of the (mean) *value* of the data. This breaks down for values that are zero or smaller than zero though. This PR fixes this by specifying bin bounds when the data std is zero.

### DIFF
--- a/bokeh/charts/stats.py
+++ b/bokeh/charts/stats.py
@@ -334,7 +334,18 @@ class Bins(Stat):
         self.bin_column = self.column + bin_str
         bin_models = []
 
-        binned, bin_bounds = pd.cut(self.bin_stat.get_data(), self.bin_stat.bin_count,
+        data = self.bin_stat.get_data()
+        bins = self.bin_stat.bin_count
+
+        # Choose bin bounds when data range is ill-defined; pd.cut()
+        # does not handle this well for values that are <= 0
+        if data.size < 2:
+            raise ValueError('Histogram data must have at least two elements.')
+        if data.ndim == 1 and not data.std():
+            margin = 0.01 * abs(float(data[0])) or 0.01
+            bins = np.linspace(data[0] - margin, data[0] + margin, bins+1)
+
+        binned, bin_bounds = pd.cut(data, bins,
                                     retbins=True, include_lowest=True, precision=0)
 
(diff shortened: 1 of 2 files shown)


bokeh/charts/stats.py
```diff
@@ -334,7 +334,18 @@ def calculate(self):
...
 
-        binned, bin_bounds = pd.cut(self.bin_stat.get_data(), self.bin_stat.bin_count,
+        data = self.bin_stat.get_data()
+        bins = self.bin_stat.bin_count
+
+        # Choose bin bounds when data range is ill-defined; pd.cut()
+        # does not handle this well for values that are <= 0
+        if data.size < 2:
+            raise ValueError('Histogram data must have at least two elements.')
+        if data.ndim == 1 and not data.std():
```
bryevdv: I think `data.std() == 0` would express the intent a bit more clearly

bryevdv:
Looks pretty good, needs to be documented somewhere, preferably with an inline plot in the docs.

bryevdv:
@almarklein as a concrete suggestion, a one or two line note in the docstring would be sufficient.

almarklein:
@bryevdv docstring updated. Since this is a fix to get the histogram producing more or less useful output for ill-defined data, I think an example is not necessary.

bryevdv:
LGTM

PR #4202, opened 2016-04-20
make reset tool behaviour configurable

### LINKED ISSUE #4201: Make reset tool configurable as to what is reset
Request for client work that reset tool not also reset the plot canvas dimensions.  Having it reset the canvas dimensions was added as a request to numerous user requests. Propose to make a configurable property to control whether canvas dimensions are also reset.

bryevdv:
issues: closes #4201 

@mattpap This is done except for whatever changes are needed to update the TypeScript API. Please make those ASAP

ping @dwillmer

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -128,14 +128,16 @@ class ResetTool(Tool):
     resets the data bounds of the plot to their values when the plot was
     initially created.
 
-    .. note::
-        This tool does not also reset the plot canvas size, if the plot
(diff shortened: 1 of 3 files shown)


mattpap:
For now it's OK, but before release we should consider if there shouldn't be a wider reset policy.

mattpap:
Also we could not reset size if plot is responsive. However, this wouldn't help with the demo, because plots there aren't responsive, just resized manually.

Issue #3563, opened 2016-01-06
Setting min_border does not work


philippjfr:
Setting the convenience attribute ``min_border`` does not appear to be working in 0.11. This small example should reproduce the issue.

```python
from bokeh.plotting import figure, output_notebook, show

output_notebook()

# create a new plot with a title
p = figure(plot_width=400, plot_height=400, min_border=0)
p.border_fill_color = "whitesmoke"
p.min_border = 0

p.circle([1,2,3,4,5], [2,5,8,2,7], size=10)

show(p)
```

Neither setting it in the constructor nor setting the attribute seems to make a difference. Only explicitly setting ``min_border_left/right/top/bottom`` actually applies it.

hafen:
I'm experiencing the same issue (putting 0.11 into rbokeh).  None of the border options seem to be working for me.

bryevdv:
@hafen Can you provide the JSON you are trying? `min_border_left` is working in the user guide generated plot here: http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#border

It is conceivable `min_border` has issues, our properties system had to undergo some changes recently, and that was kind of an odd duck. Will investigate.

mmcdan:
Would like to confirm that for 0.11.0, ```min_border=0``` does not work for me, but ```min_border_left=0, min_border_right=0, min_border_top=0, min_border_bottom=0``` does work.

mattpap:
Looking at `common/plot.coffee`, there is simply no logic for `min_border`. Otherwise the property is present and serialize properly. I can submit a fix tomorrow.

mattpap:
It's interesting to note that this originally used to work thanks to differences between property values in bokeh and bokehjs. As this gap was recently closed, as long as we want to keep numerical defaults for all five properties, there is no way to distinguish which of `min_border` and `min_border_*` was set intentionally and should take precedence. Currently the simplest solution would be to set the default value of `min_border` to `None`. Or we could introduce virtual properties. Alternatively, we could get rid of `min_border_*` altogether and set `min_border`'s type to `Instance(Rect).accept(Int, lambda v: Rect(v, v, v, v)` where `Rect` is a model with `top`, `right`, `bottom`, `left` properties (thought to do it right, we would need to introduce type parameterized models).

mattpap:
Actually setting `min_border` to `None` doesn't make sense. It's `min_border_*` which would need a change, but that defeats having default values at all.

mtinti:
[...]

Issue #2759, opened 2015-08-20
Resolve misleading property names in `Bokeh.index[<item_id>].renderers` and `Bokeh.index[<item_id>].model.renderers`


brendancol:
BokehJS: On of these are renderers, and the other are renderer_views.  Confusing names should be resolved

bryevdv:
It's specifically `PlotView` that stores all the renderer views as `plot_view.renderers`

birdsarah:
Hi, this is a change I've made as part of the layout work - as it was confusing me too!

birdsarah:
Done in this commit: https://github.com/bokeh/bokeh/pull/4192/commits/b811f2eab65adc61115d5b3daec565a913ad3b99

May be a while before it's merged.

PR #4203, opened 2016-04-20
Document is responsible for gathering all Solver Constraints

birdsarah:
Fixes: #3563, #2759, #4203 

Work towards: #4164 

This implements just the re-working of Solver to be top-down from the document.

Added in the fix for #3563 as it's also easy and stand-alone.


bokeh/models/plots.py
```diff
@@ -56,12 +57,8 @@ def _select_helper(args, kwargs):
         selector = kwargs
     return selector
 
-class LayoutBox(Model):
-    ''' Represents an **on-canvas** layout.
-
-    '''
 
-class Plot(Component):
+class Plot(LayoutDOM):
```
mattpap: Say that this is a `DOM` layout doesn't seem necessary to the user. `Layout` would suffice.

bokehjs/src/coffee/core/layout/layout_canvas.coffee
```diff
@@ -0,0 +1,49 @@
+_ = require "underscore"
+
+{Variable, EQ, GE, Strength}  = require "./solver"
+Model = require "../../model"
+p = require "../properties"
+
+class LayoutCanvas extends Model
+  type: 'LayoutCanvas'
+
+  @internal {
```
mattpap: I presume (based on other changes) this is an internal class. If so, it should be marked accordingly.

bokeh/models/plots.py
birdsarah: @mattpap, I know what you mean, but I think it's helpful to distinguish between `LayoutDOM` and `LayoutCanvas`. I think it helps to hint they're not in a class hierarchy (that is `LayoutCanvas` is not a subclass of `Layout`). Am happy for suggestions other than `DOM` as the suffix.

bokehjs/src/coffee/core/layout/layout_canvas.coffee
birdsarah: I'm not entirely sure what an internal class is, but I'm guessing yes - it's not something we are going to expose to users. How do I mark it as such?

birdsarah: Perhaps related is that I'd like `LayoutDOM` (or whatever it's to be called) to also be private and not exposed to users and to be put into `/core/layout` in the same place as `LayoutCanvas` but it seemed to be necessary to make it public....I think because of the check in `io.py` - if you have any suggestions, in this area, that would be great.

bokeh/models/plots.py
birdsarah: Perhaps related is that I'd like LayoutDOM (or whatever it's to be called) to also be private and not exposed to users and to be put into /core/layout in the same place as LayoutCanvas but it seemed to be necessary to make it public....I think because of the check in io.py - if you have any suggestions, in this area, that would be great.

(note I previously put this note in the wrong place)

birdsarah:
@bryevdv, further to our conversations, I was able to make this work as an isolated PR. I guess I just needed fresh eyes. So this PR is against master and future layout PRs will continue to build off it.

birdsarah:
Changing back to WIP as am adding a few more things into this PR.

birdsarah:
This should be ready @bryevdv @mattpap 

Nothing in the functionality of bokeh should have changed. The purpose of this PR is to refactor all solver work to go through a centralized solver on the document so that it can now coordinate constraints across plots in the document.

birdsarah:
There appears to be a problem with server plots. Looking into it.

birdsarah:
OK. Back to ready again. Wasn't a server thing. HBox VBox were not passing constraints up to Document.

bokehjs/src/coffee/core/layout/layout_canvas.coffee
mattpap: This is done by adding `serializable_in_document: () -> false`. However, if you need document attached to instances of this class, then this isn't going to work. Actually, ignore this for now.

bryevdv:
@birdsarah ran some examples locally and things looked mostly good, but there are definitely a few failures in the tests:

https://travis-ci.org/bokeh/bokeh/jobs/127422968

Looks like `glyphs.py`, `calendar.py` and `legend.py` and possibly a few others (ignore the notebook comm failures, that's fixed in another PR)

Also some kind of boto debugging has been turned on, exploding the log size. It won't show the whole thing anymore, so I can't get to the report link at the bottom.

bokeh/models/layouts.py
```diff
@@ -28,8 +28,24 @@ class Layout(Component):
...
 
+    disabled = Bool(False, help="""
+    Whether the widget will be disabled when rendered. If ``True``,
+    the widget will be greyed-out, and not respond to UI events.
+    """)
+
+    # TODO: (mp) Not yet, because it breaks plotting/notebook examples.
+    # Rename to _repr_html_ if we decide to enable this by default.
+    def __repr_html__(self):
+        return notebook_div(self)
```
bryevdv: Let's just delete this?

bokeh/models/plots.py
```diff
@@ -420,17 +417,15 @@ def _check_colon_in_category_label(self):
...
     """)
 
     right = List(Instance(Renderer), help="""
     A list of renderers to occupy the area to the right of the plot.
     """)
 
-    # TODO (bev) LayoutBox here is a temporary workaround to the fact that
-    # plot titles are not proper renderers
-    above = List(Either(Instance(Renderer), Instance(LayoutBox)), help="""
+    above = List(Instance(Renderer), help="""
```
bryevdv: awesome

bryevdv:
I did a quick pass through the diff, many things look much nicer/cleaner. All the tests are very nice as well. I will take another look later today. I'd say as as soon as all the non-notebook failures in the examples tests are working this can get merged. Does that sound good?

bryevdv:
It looks like there was just one more failure that I missed above, `unemployment.py`:

https://travis-ci.org/bokeh/bokeh/jobs/127422968#L2135
https://travis-ci.org/bokeh/bokeh/jobs/127422968#L2385
https://travis-ci.org/bokeh/bokeh/jobs/127422968#L2801
https://travis-ci.org/bokeh/bokeh/jobs/127422968#L3366

birdsarah:
Thanks for the notes @bryevdv, and for the thorough look through on the examples. Will check them out and update.

bokeh/models/layouts.py
birdsarah: I can, I was just keeping everything the same as it was before.

birdsarah:
> Also some kind of boto debugging has been turned on, exploding the log size. It won't show the whole thing anymore, so I can't get to the report link at the bottom. 

Yes! I'm not sure when that happened, but I believe it's in master, so I'm not going to change that in this PR. Already a lot going on.

bokeh/models/layouts.py
mattpap: > Let's just delete this?

This code is there to show that we can speak IPython's display protocol directly, just our examples have to be updated for this change. We can remove this though, but this PR is no place for that.

birdsarah:
@bryevdv. Have fixed: legends, calendar, unemployment, datatable, horizon, models/glyphs examples which are those that I perceived to be broken from looking at examples report with fresh eyes (fresh being realizing that actual is on the left!)

bryevdv:
If the repair looks good and server examples run well I say LGTM to merge.

birdsarah:
I haven't run all the server examples. I'll do that when I get a moment if I don't hear from anyone else.

bryevdv:
Resolved small merge conflict in math utils file after arrow PR merge.

bryevdv:
Notes on the apps:

crossfilter
-------------
Does not update if controls are used, but this is the case before. Shows console error:

screen shot 2016-05-04 at 4 20 52 am

export_csv
--------------
Updates work great, but column containing slider is cropped / partially under the plot:

screen shot 2016-05-04 at 4 22 29 am

Can actually move the slider further right, obscuring it completely, 

random_tiles
-----------------
Fails with console error, fairly sure this is unrelated and is from recent TS API work

screen shot 2016-05-04 at 4 25 57 am


simple_hdf5
---------------
Unrelated error where the hdf5 file can't seem to be loaded after the first interaction.

---

Everything else works great.

birdsarah:
Let's make separate issues I think.

1) The slider obscuring is going to fall nicely out of imminent work, so not much point in fixing it in its current state.
2) hdf5 and random appear to not be this PR
3) crossfilter - am I understanding that it was broken, but I broke it more. Lets open another issue and I'm happy to support any layout problems with it.

bryevdv:
Not sure about more, I think maybe it's now broken differently. :) OK, merging now.

bryevdv:
It says "Fixes 4203" at the top, but this PR is 4203, is that supposed to be a different issue?

birdsarah:
Doh! Fixed.

Issue #4206, opened 2016-04-21
Add example for CustomJS callback for exporting ColumnDataSource to csv


jbrambleDC:
as discussed with @bryevdv  in [https://groups.google.com/a/continuum.io/forum/#!msg/bokeh/O6eccziL1lM/KMqLESxHAQAJ](url) I will be adding a PR for an example of a bokeh server app that creates a data table of some data and allows a user to click a download button and then download a .csv of the ColumnDataSource associated with that DataTable in their browser.

jbrambleDC:
this is addressed by #4207

PR #4207, opened 2016-04-21
adds server app example for exporting ColumnDataSource data to csv

jbrambleDC:
Opening this PR at the reccomendation of @bryevdv . This addresses #4206 . This is a simple bokeh server app that can be ran with `bokeh serve --show csv_export.py`. Upon launching the app you will see a data table render. The sliders allow one to change the max salary of this data. The download button is configured with a CustomJS callback to export the data to csv in the users browser.  I suspect we do not need to tests given that this is an example server app and not a a feature being added to the Bokeh library.

I have currently only tested this on chrome. It should work on Firefox and IE as well. 

NOTE: this is my first PR to bokeh, so please bear with me. 

Also, I completely understand if we want to avoid allowing people to add their own sample data .csvs to this repo in Pull Requests. It could pollute the repo. We can discuss using some of the Bokeh sample data to accomplish this same task if needed. 

the customJS is defined as follows:

```python
js_callback = """
var data = source.get('data');
var filetext = 'name,income,years_experience\\n';
for (i=0; i < data['name'].length; i++) {
    var currRow = [data['name'][i].toString(),
                   data['salary'][i].toString(),
                   data['years_experience'][i].toString().concat('\\n')];

    var joined = currRow.join();
    filetext = filetext.concat(joined);
}

var filename = 'data_result.csv';
var blob = new Blob([filetext], { type: 'text/csv;charset=utf-8;' });

//addresses IE
if (navigator.msSaveBlob) {
    navigator.msSaveBlob(blob, filename);
}

else {
    var link = document.createElement("a");
    if (link.download !== undefined) {
        var url = URL.createObjectURL(blob);
        link.setAttribute("href", url);
        link.setAttribute("download", filename);
        link.style.visibility = 'hidden';
        document.body.appendChild(link);
        link.click();
        document.body.removeChild(link);
    }       
}"""

button.callback = CustomJS(args=dict(source=source), code=js_callback)
```


bryevdv:
@jbrambleDC This is awesome! A few quick observations:

* We are trying to keep all the apps in one place for now, so I would suggest moving to  `examples/apps/export_csv` or similar
* the last line of the data file seems to be corrupt
* we might want to consider making a smaller `.csv` file for the repository (or else we can upload it to s3 and make it part of `bokeh.sampledata`
* A paragraph or div widget hat updates with the number of rows would be a nice addition, if you wanted to get fancy you could make a "stats" box like the stocks app example has
* It might be possible to write the callback in python! http://bokeh.pydata.org/en/latest/docs/user_guide/interaction.html#customjs-with-a-python-function 

Finally on OS X this is working great on chrome, but does not seem to be on OS X, will require at least a little investigation which I can help with at some point soon.

bryevdv:
> NOTE: this is my first PR to bokeh, so please bear with me.

Also, we are always very happy to have new contributions and new contributors, and try to provide as much help as possible. Thanks!

jbrambleDC:
@bryevdv  I will move this example to the specified location.

- Interesting, that the last line is corrupt. I cant reproduce that on OS X, even when downloading all of the data. I've opened it with vim and with excel 2016, and have seen no issues.
- I'm open to a smaller csv, or uploading to s3. let me know what you need from me to make that happen.
-  I would be super excited to write the callback in Python. I think when possible we should always write Python. However, looking at the PyScript docs I can't find any way to reference DOM objects or create elements in the document. My apologies if I am missing something.

I assume your last comment means that this is not working on safari. or not working on IE? I would very much appreciate help with trouble shooting that.

jbrambleDC:
Tested this on Safari, and I am indeed failing to get it download. To be honest, I dont have a ton of knowledge on how safari operates as a browser. clicking the button merely refreshes the page.

bryevdv:
@jbrambleDC by "corrupt" I just mean the name on the last line is the empty string. It does have the right number of commas, so I expect would work. I see that in Vim but also here on GH if I go to the raw view and scroll to the bottom.

If we do keep the file in repo, I would suggest keeping it under ~250 lines. I think that would be ok and not detract from the point of the demo, but if you think that is too short we can either put it up on S3 and add to `bokeh.sampledata` or else look at existing sample data (maybe the "autompg" data set would also work really well?

Regarding writing in python, it's a little weird, the python is basically converted to JS in a very simple way, so I think things like:
```
 var link = document.createElement("a");
```
would just become:
```
link = document.createElement("a")
```
@almarklein can help answer questions about this, tho I will also so it's merely a "nice to have, maybe" not a super high priority to worry about.

jbrambleDC:
Made your suggested changes. I will try to figure out the safari issue. Ideally someone in the Bokeh community could also try running this on Internet Explorer as well. I will also look into writing the callback in PyScript

jbrambleDC:
@bryevdv I have looked into writing the callback in Python. My preference is to punt on that, after learning that Python callbacks converted to JS are currently only supported in Python 3. Would be interested in hearing your opinions on that.

My personal preference is to avoid things only supported in python 3 in favor of being inclusive towards the community. 

I think the final open issue on this is to figure out how to get this working on Safari (if possible), or noting as much in the app if not possible.

bryevdv:
Jordan they now also work with Python 2 is there a place the docs need to be updated?

jbrambleDC:
yes @bryevdv from: http://bokeh.pydata.org/en/latest/docs/user_guide/interaction.html#customjs-with-a-python-function

There is a note stating `This functionality is currently only supported on Python 3.x., however Python 2.7 support is planned.`

bryevdv:
@jbrambleDC ah, the py2 support is more recent than the latest release, but the docs have been updated in master:

https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/interaction.rst#customjs-with-a-python-function

 I believe the download issue for safari  may be the same one reported in https://github.com/bokeh/bokeh/pull/4228, if so I am not sure there is anything we can do about it unfortunately.

jbrambleDC:
great news @bryevdv does this mean that I will need to build a dev release of bokeh to get PyScript working for python 2.7? Otherwise I will give it a crack. 

How should we address the Safari Issue? Create div in the app? or just leave a comment in the code mentioning that it doesnt work in Safari?

Hopefully I can get these things addressed and we can merge this week.

jbrambleDC:
@bryevdv I think everything is good to go here. Unless you want to do something further to address the Safari issue. Im going to punt on the PyScript implementation for now. I think it will be good however, eventually to show people how to do these kind of document interactions with JS callbacks and Python implementations.

Let me know if you need anything before merge!

bryevdv:
@jbrambleDC waht about adding a new custom template? If you add a `templates/index.html` to the directory, then you could put some instructions and comments (e.g. about safari) directly on the page. t would look something like:
```
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <title>{{ title if title else "Bokeh Plot" }}</title>
        {{ bokeh_css }}
        {{ bokeh_js }}
    </head>
    <body>
        <div>
            HERE IS SOME CUSTOM STUFF
        </div>
        {{ plot_div|indent(8) }}
        {{ plot_script|indent(8) }}
    </body>
</html>
```

bryevdv:
OK @jbrambleDC thank you, this a extremely close. Just a couple of last points:

* for a "directory" app, the python file needs to be `main.py` by convention, then you can run e.g. `bokeh serve --show export_csv` from the directory above. 

* `button.type` has been deprecated, use `button.button_type` instead:
```
2016-04-29 14:58:01,582 Starting Bokeh server with process id: 63803
/Users/bryan/work/bokeh/bokeh/models/widgets/buttons.py:38: UserWarning: 
            Property 'type' was deprecated in Bokeh 0.12.0
            and will be removed. Use 'button_type' instead.
            
  """)
```
* Based on some conversation in #4228, I think this code performs a bit better:
```
else {
    var link = document.createElement("a");
    link = document.createElement('a')
    link.href = URL.createObjectURL(blob);
    link.download = name
    link.target = "_blank";
    link.style.visibility = 'hidden';
    link.dispatchEvent(new MouseEvent('click'))
    link.click();
}
```
Using this, on Safari a new tab is opened wth the plain text CSV contents, which is better than it doing nothing at all. On Chrome and FF on OS X the download works as expected. 
* The custom template PR *is* newer than this one, I would recommend merging in master first, `git merge master` so that the effect and appearance of the template can be evaluated. If I merge master in locally, it looks like this:
screen shot 2016-04-29 at 3 02 44 pm
Maybe a little additional explanation and instruction, and some minimal styling? With the other JS code above we could change the message to indicate that on Safari, a new tab will be opened (that can be saved manually) instead of the download dialog.

bryevdv:
Thanks for your patience this will be a great demo example!

jbrambleDC:
@bryevdv Ive made all of the requested changes except the customJS. I changed it to:

```
else {
    var link = document.createElement("a");
    link = document.createElement('a')
    link.href = URL.createObjectURL(blob);
    link.download = filename
    link.target = "_blank";
    link.style.visibility = 'hidden';
    link.dispatchEvent(new MouseEvent('click'))
    link.click();
}
```

It seems to open up the tab in Safari however it now does nothing in Chrom Version 49.0.2623.112 (64-bit) on OSX.

jbrambleDC:
Never mind, @bryevdv I've rewritten the JS and added some instructions into the template. Forgive me if my HTML is not the worlds best. I think this can ship!

Also, Thanks a ton for your feedback, patience, assistance, and professionalism!

bryevdv:
This looks great, I will plan to try and merge this sometime tomorrow!

bryevdv:
Thanks for the PR @jbrambleDC !

PR #4212, opened 2016-04-21
More TS API improvements

mattpap:
- [x] always create legend, so that we can style it before adding glyphs
- [x] toolbar is now a `<div>` and its CSS is fixed
- [x] toolbar is rendered immediately, instead of in `render()`
- [x] fix `Legend`'s typings


mattpap:
> FAIL: Image dimensions do not match

@birdsarah, how do I recover from this?

> PhatomJS Error: : Could not open comm

@bryevdv, this might be related to your recent changes.

bokehjs/src/coffee/models/axes/axis.coffee
```diff
@@ -283,14 +283,20 @@ class AxisView extends Renderer.View
...
+    if isNaN(x) or isNaN(y)
+      return
+
     if angle
-      ctx.translate(sx+nx*standoff+nx*xoff, sy+ny*standoff+ny*yoff)
+      ctx.translate(x, y)
       ctx.rotate(angle)
       ctx.fillText(label, 0, 0)
       ctx.rotate(-angle)
-      ctx.translate(-sx-nx*standoff+nx*xoff, -sy-ny*standoff+ny*yoff)
```
mattpap: @bryevdv, I assumed this is a bug. This should just be `-x, -y`, but it's not. Or do I miss something here?

mattpap: Also, it it a true optimization to split zero and non-zero angle cases here? I have to verify that, but I would expect `rotate(0)` to be a no-op and `fillText(label, non_zero, non_zero)` to actually perform translation behind the scenes.

bryevdv:
My recent change only wrapped the `comm` attempt in a `try`/`catch`

bokehjs/src/coffee/models/axes/axis.coffee
bryevdv: It used to be prudent to avoid setting anything on the CTX necessarily, I'd have to go dig up the benchmarks I found, is it still the case, I don't know.

bryevdv: @mattpap yes it looks like it could be `-x` and `-y`

mattpap:
> My recent change only wrapped the comm attempt in a try/catch

@bryevdv, that issue resolved by itself. We should look into future tests run to see if this persists.

@birdsarah, I still didn't get any feedback regarding integration tests. Not that other work depends on this PR.

bryevdv:
@mattap re: image dimensions are you asking how to update the rest? (Because the dimensions are expected to change?)

mattpap:
@bryevdv, at this point I'm not sure if it's a bug I introduced in this PR (based on manual testing it doesn't seem so) or this is a problem with testing infrastructure. We aren't under pressure for merging this (thought it would be helpful), but I have pending changes and I would like to know what is going on here.

bokehjs/src/coffee/models/axes/axis.coffee
mattpap: @bryevdv, I'm also curious if you benchmarked, e.g., `ctx.translate(x, y); ctx.translate(-x, -y)` vs. `ctx.save(); ctx.restore()`?

bryevdv:
@mattpap I din't see every GH notification sometimes (and I definitely don't see every GH ping responded to!), I would suggest if something is pressing that multiple communication channels (email, gitter) also be used.

mattpap:
@birdsarah, I need the integration test issue resolved at some point.

birdsarah:
@mattpap 

1) I would have thought the explanation is pretty clear - the images are different sizes
2) If you look at the test report: https://s3.amazonaws.com/bokeh-travis/124979554/tests/pytest-report.html you can SEE the difference between the checked in image and your test image - one has a horizontal line under the toolbar, one doesn't - I am guessing this is something to do with your css changes.

Please at least try reading the test reports.

mattpap:
@birdsarah 

> 1) I would have thought the explanation is pretty clear - the images are different sizes

That's not an explanation. An explanation would be to be informed where this difference comes from. It would be helpful to know even that this is an off by one pixel difference.

A different question is why only integration tests fail and not everything else. Examples should fail as well, but they happily pass and only manual inspection reveals the (subtle; but it depends on perception) difference. I would expect to get the same kind of feedback from different but related test groups, i.e. if one fails hard, the other do as well.

> If you look at the test report:

I can't make anything out of that report. It's unhelpful. Examples' report is actually better in this case, yet getting to it isn't as easy as it used to be (some of its functionality is broken as well). The reason is vast amount of unrelated logs being recorded, which makes travis ci fail to display the entire log. The link to the report is at the end of it, thus requiring me to open raw log and search in it. To you this is perhaps a minor thing, but to me, a developer requiring a helpful tool for his daily work, it's a major pain when those little annoyances accumulate (and there are plenty of them, not only in the testing infrastructure, but in build tools, etc. as well).

> Please at least try reading the test reports.

I will comment on this in a more appropriate place.

birdsarah:
- "That's not an explanation. An explanation would be to be informed where this difference comes from." - Fair enough, that's just the raw output from pdiff, it's not something I wrote, we could maybe test the exit status and write a more detailed help statement. Open an issue?

- On a related note, I squashed the log in the image diff tests so the html test report says "no log output captured" we could probably do better than that. Another issue?

I can't tell from  the rest of your comments, if you think your grievances are are things that I have introduced into Bokeh. I don't think they are, and I agree with most of them....so more issues? 

Why only integration tests fail and not everything else:
- lack of unit tests (although its hard (impossible?) to unit test css)
- examples tests pass, even when there's diffs - it's been like that for as long as I've been working with bokeh (which is not as long as you) I agree that this is sub-optimal which is why I was excited to create image diff tests with reference images checked into the library that fail on a diff. 

"I can't make anything out of that report. It's unhelpful." - Please open an issue with suggestions of how to make it better. It reads fine to me, so its hard for me to know how to improve it.

Examples report: "Yet getting to it isn't as easy as it used to be" - what did it used to be? Sounds like we need an issue for this too

"The reason is vast amount of unrelated logs being recorded, which makes travis ci fail to display the entire log." - Can you elaborate / open an issue

bryevdv:
Hi.
 
I think we are all on the same page in wanting Bokeh to be better, for users and for developers. I think we all want more testing, and more automated testing. Is there any disagreement here? I can't think of any two people who have one more to improve the testing of Bokeh than the two of you. Then we just have to figure out constructive ways to work together and accommodate one another, as well as  be accountable to one another, as we work toward those goals (which will surely take longer than we'd like, such is the reality of limited resources).

Here is my super-abridged take on the friction in this PR (***please*** correct me if I am mischaracterizing anything)

* @mattpap is frustrated about certain aspects of the current testing machinery
* @birdsarah could potentially help, but is frustrated by terse requests / missing details

Presumably, no one here *wants* or *intends* to frustrate anyone else, so let's decide not to do that. In my experience, that means thinking about what I can do differently, to *help the person I want help from*. 

One of the really valuable observations that came out of the estimation meeting is that our GH issues can definitely at times be underspecified, lack enough information, or overlap too much with other issues. So let's use this as an opportunity to practices being better there. I'm happy to shatter my "under 600 issues" dream, for a raft of detailed issues that really focus constructive discussion and development. So:

@mattpap will you commit creating detailed, fine-grained issues for specific enhancements that you would like to see, explaining why they would be helpful to you or others, answering any questions for more detail?

@birdsarah will you commit to discussing those issues, providing sincere feedback about feasibility, difficulty, possible implementation issues, and asking questions for more details if necessary?

If we can do those things, then we can plan and prioritize these things as a team. 

Thank you both for all you've done to make Bokeh better.

birdsarah:
@bryevdv - thank you for your note. 

> will you commit to discussing those issues, providing sincere feedback about feasibility, difficulty, possible implementation issues, and asking questions for more details if necessary?

Absolutely.

mattpap:
Despite unrelated discussions in this PR, I would like to actually fix the problem there is with integration tests. At this point we know that there is a slight difference in how toolbar is rendered. There used to be more space (1~2px) above the toolbar. The difference comes most likely from the fact that toolbar isn't embedded into a table directly anymore, but has its own div and thus there are no interaction between table's borders and toolbar borders (but maybe there is some thing else going on in similar fashion). This difference is acceptable, especially that whole table structure is going away soon.

Examples tests don't fail hard and give valuable feedback on what is actually different between `master` and this PR. The reason they are helpful is because despite plot size change, they take more space into account, so plot size difference (up to some degree) is incorporated in a diff. On the other hand, integration tests fail hard. This raises a question I asked at the beginning of this PR, how do I recover from this? Obviously I can restore original plot size, which I don't have intention to, because I think the new size is actually better (less space wasted). What do I do next? I tried to follow integration tests' code, but I don't see where `base` images come from. I presume `base` images are stored on S3, the same as images produced by examples. At this point my observation is that I'm not allowed to change plot size in integration tests. So are there any actions for me to perform to get around this? Is this a bug or a design oversight? I would expect image diff images to fail softly and always produce a diff.

I have a long list of things that don't work or don't work for me in the new testing infrastructure (some inherited from the old one), but that's something for longer term development perspective. The problem in the PR is something I need to get fixed soon.

mattpap:
Now I see that both failing tests' base images are committed into the repository.

bryevdv:
@mattpap do you have enough to go on now?

mattpap:
@bryevdv, it depends. Just pushing this forward yes, but that would leave the issue unresolved. As I see it, with integration tests we have to update those images every time something changes, even if it's unrelated. Currently there are only two base images to consider, but ultimately there will be lots of them and this approach is not going to function. Note also that putting generated stuff in the repository is against our policy. I would like examples and integration tests use the same soft approach with uploads to S3 and then we can experiment with and implement new, possibly better, approach with hard failures.

bryevdv:
@mattpap I agree that putting them in the repo is not scalable in the long term. Rather than S3, I would suggest making a separate GH repo to hold reference images for image tests.That would be much more accessible to anyone (everyone) that does not interact with S3 regularly. I also agree that now that there is some infrastructure, the actual policies about what and when to test using it need some thought and guidance. There *are* things that should *never change*. At least certainly not accidentally or without deliberation. Can you collect your own ideas for features, polices, improvements, etc. into a discussion issue? You, me, @birdsarah and anyone else interested can weigh in and then we can create a epic with sub-tasks for improving testing even more.

birdsarah:
I very recently updated the dev docs with information on how to run the screenshot tests, including how to create new base screenshots.

https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/dev_guide/testing.rst#screenshot-tests

It's very straightforward....once you have it setup.

I can also submit the updated screenshots if you like.

birdsarah:
(On the testing note, for background, some good discussions were had here: https://github.com/bokeh/bokeh/pull/3896. Other related PRs: https://github.com/bokeh/bokeh/pull/3869, https://github.com/bokeh/bokeh/pull/3975)

birdsarah:
@mattpap I added the base screenshots and I also fixed the merge conflict as both merge conflicts were related to my recent PR so it seemed easy for me to do.

mattpap:
@birdsarah, thanks.

mattpap:
So, I think we should set `toolbar_location=None` for all integration tests that don't explicitly test toolbar's behavior. This is not only to reduce unnecessary work as in this PR, but also to simply make reasoning about test failures easier, because failure in toolbar rendering, shouldn't stop unrelated tests from succeeding (or failing, but for an unrelated reason). Of course we need a test that actually tests the toolbar and for the remainder of tests we need tool activation API.

PR #4216, opened 2016-04-22
Adding built assets to npm

### LINKED ISSUE #4210: Built JS and CSS not included in NPM package
We need to include the built `bokeh.js` and `bokeh.css` with the npm package. Then add
`"main": "build/js/bokeh.js"` to our `package.json`

mistakevin:
- [x] issues: fixes #4210 

Steps to publish are documented in #4210

### DIFF
--- a/bokehjs/.npmignore
+++ b/bokehjs/.npmignore
@@ -1,0 +1,8 @@
+/node_modules
+*~
+*.sw[opn]
+*.bak
+*#*
+*.#*
+*.DS_Store
+npm-debug.log
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -100,5 +100,6 @@
   "scripts": {
     "build": "gulp build",
     "dev-build": "gulp dev-build"
-  }
+  },
+  "main": "build/js/bokeh.js"
 }


bryevdv:
What about the CSS and `bokeh-widgets.js`?

mistakevin:
You pull those in manually from your `node_modules` folder using the `"browser"` object in your `package.json`. The `"main"` just points to the library you want to bring in with a simple `require(<package name>)` call.

bryevdv:
@damianavila any comments?

bryevdv:
@bokeh/dev I will merge this later today unless there is objection. @mistakevin is there some way we could test this ahead of time?

bryevdv:
@mistakevin actually I am still a bit confused. There is no mention of `bokeh-widgets.js` in the "browser" section:
```
  "browser": {
    "bootstrap/button": "./src/vendor/bootstrap-3.1.1/js/button.js",
    "bootstrap/dropdown": "./src/vendor/bootstrap-3.1.1/js/dropdown.js",
    "bootstrap/modal": "./src/vendor/bootstrap-3.1.1/js/modal.js",
    "bootstrap/tab": "./src/vendor/bootstrap-3.1.1/js/tab.js",
    "slick_grid/slick.core": "./src/vendor/slick-grid-2.1.0/slick.core.js",
    "slick_grid/slick.grid": "./src/vendor/slick-grid-2.1.0/slick.grid.js",
    "slick_grid/plugins/slick.rowselectionmodel": "./src/vendor/slick-grid-2.1.0/plugins/slick.rowselectionmodel.js",
    "slick_grid/plugins/slick.checkboxselectcolumn": "./src/vendor/slick-grid-2.1.0/plugins/slick.checkboxselectcolumn.js",
    "jqrangeslider/jQRangeSlider": "./src/vendor/jqrangeslider-5.7.0/jQRangeSlider.js",
    "jqrangeslider/jQDateRangeSlider": "./src/vendor/jqrangeslider-5.7.0/jQDateRangeSlider.js",
    "jquery_event_drag": "./src/vendor/jquery-event-2.2/jquery.event.drag.js",
    "jquery_event_drop": "./src/vendor/jquery-event-2.2/jquery.event.drop.js",
    "gear_utils": "./src/vendor/gear-utils/gear-utils.js",
    "kiwi": "./src/vendor/kiwi/kiwi.js",
    "gloo2": "./src/vendor/gloo/gloo2.js"
  }
```

To be clear, we split what used to just be one library (`bokeh.js`) into two smaller pieces (`bokeh.js` and `bokeh-widget.js`) What is the "proper" npm way to handle this? Is it possible to have a single npm project that is comprised of multiple js files?

mistakevin:
Yes. I meant if a user needed to bring in `bokeh-widgets.js` into their projects, they put it in THEIR `packages.json`. Many libraries that have been broken up into many smaller modules do this. Take Zurb Foundation for instance

screen shot 2016-04-26 at 3 58 36 pm

Instead of bringing in all of `foundation.js`, you can bring in the individual modules as you ned them. Then in your code, you can simply `require` the library.

So in my `package.json`, I would have 

```json
{
  "dependencies": {
    "bokehjs": "0.12",
  },
  "browser": {
    "bokeh-widgets": "./node_modules/bokehjs/build/js/bokeh-widgets.js",
  },
}
```

and in my code I would have

```javascript
window.Bokeh = require('bokehjs');
require('bokeh-widgets');
```

bryevdv:
OK thanks for the explanation. The part I was missing was that all of our `build` dir gets packaged automatically (I believe, from what you have said) I thought perhaps we had to explicitly specify. Will merge this later tonight.

mistakevin:
Yeah, the `.npmignore` is basically the same as `.gitignore`, but without the `build` directory. `.npmignore` will be used now, instead of `.gitignore`

damianavila:
Sorry to be late here... just a simple question... don't we need, eventually, to ship bokeh-widgets as a standalone package? Or this `browser` pattern is wide extended enough in the JS community?

Additionally, do we have any docs to tell JS people how to use it? It could be enough to have something like the description above, but from the PR, I don't see any docs changes.

mistakevin:
Adding something to the docs would sound like a great idea.

I don't know how entangled `bokeh.js` and `bokeh-widgets.js` are. Do you always have to have `bokeh-widgets.js`? In some modules there is, for instance, a `bokeh.full.js` vs `bokeh.core.js` and `bokeh.widgets.js` that you can bring in modularly.

PR #4220, opened 2016-04-22
store computed final props on a private attribute

### LINKED ISSUE #4219: New strict JS property updates broke DatetimeTickFormatter
The `formats` property is actually a bit problematic. The values sent from python are actually diffs applied on top of a standard base set of formats stored in BokehJS. The "final computed" formats were stored in `@formats` but recent properties changes made `@formats` equivalent to `@get('formats')` however the "final computed" formats actually contain other information and fail validation with errors like: 
```
2016-04-22 15:41:45,748 error handling message Message 'PATCH-DOC' (revision 1): DeserializationError('String expected str, got [6]',)
2016-04-22 15:41:45,749   message header {'msgtype': 'PATCH-DOC', 'msgid': '2DCC546DE9464C298093EC714AF6DE29'} content {'references': [], 'events': [{'attr': 'formats', 'new': {'minsec': [[6], [':%M:%S']], 'microseconds': [[8], ['%fus']], 'milliseconds': [[5, 7], ['%3Nms', '%S.%3Ns']], 'months': [[5, 7], ['%b%y', '%m/%Y']], 'days': [[5, 5], ['%m/%d', '%a%d']], 'years': [[3, 4, 4], [{}, '%Y', {}]], 'seconds': [[3], ['%Ss']], 'hourmin': [[5], ['%H:%M']], 'hours': [[3, 5], ['%Hh', '%H:%M']], 'minutes': [[3, 3], [':%M', '%Mm']]}, 'kind': 'ModelChanged', 'model': {'type': 'DatetimeTickFormatter', 'id': 'e492fb4b-edef-48b9-8846-abd1960e6b1a'}}, {'attr': 'formats', 'new': {'minsec': [[6], [':%M:%S']], 'microseconds': [[8], ['%fus']], 'milliseconds': [[5, 7], ['%3Nms', '%S.%3Ns']], 'months': [[5, 7], ['%b%y', '%m/%Y']], 'days': [[5, 5], ['%m/%d', '%a%d']], 'years': [[3, 4, 4], [{}, '%Y', {}]], 'seconds': [[3], ['%Ss']], 'hourmin': [[5], ['%H:%M']], 'hours': [[3, 5], ['%Hh', '%H:%M']], 'minutes': [[3, 3], [':%M', '%Mm']]}, 'kind': 'ModelChanged', 'model': {'type': 'DatetimeTickFormatter', 'id': '02ed10d1-f233-43d5-bc22-0585f1249521'}}]}
```

bryevdv:
- [x] issues: fixes #4219
- [x] tests added / passed

This PR makes the "internal" formats dict on a datetime tick formatter actually be internal. This PR also removes some unused "function" formats that were not compatible with serialization. 

Note that it is less than ideal that the `formats` bokeh property is always set from the JS side, it migh tbe better to set "base formats" and overlay the "diff formats" from python. This PR does not address that.

### DIFF
--- a/bokehjs/src/coffee/models/formatters/datetime_tick_formatter.coffee
+++ b/bokehjs/src/coffee/models/formatters/datetime_tick_formatter.coffee
@@ -14,22 +14,6 @@ _us = (t) ->
   # microsecond / millisecond tick.
   return Math.round(((t / 1000) % 1) * 1000000)
 
-_two_digit_year = (t) ->
-  # Round to the nearest Jan 1, roughly.
-  dt = new Date(t)
-  year = dt.getFullYear()
-  if dt.getMonth() >= 7
-      year += 1
-  return SPrintf.sprintf("'%02d", (year % 100))
-
-_four_digit_year = (t) ->
-  # Round to the nearest Jan 1, roughly.
-  dt = new Date(t)
-  year = dt.getFullYear()
-  if dt.getMonth() >= 7
-      year += 1
-  return SPrintf.sprintf("%d", year)
-
 _array = (t) ->
   return tz(t, "%Y %m %d %H %M %S").split(/\s+/).map( (e) -> return parseInt(e, 10) );
 
@@ -56,8 +40,8 @@ class DatetimeTickFormatter extends TickFormatter.Model
   type: 'DatetimeTickFormatter'
 
   @define {
-      formats: [ p.Any, {} ] # TODO (bev)
-    }
+    formats: [ p.Any, {} ] # TODO (bev)
+  }
 
   # Labels of time units, from finest to coarsest.
   format_order: [
@@ -77,10 +61,7 @@ class DatetimeTickFormatter extends TickFormatter.Model
     'hours':        ['%Hh', '%H:%M']
     'days':         ['%m/%d', '%a%d']
     'months':       ['%m/%Y', '%b%y']
-    # TODO: the functions here currently aren't serializable
-    # and so can't be set from the python client. Although,
-    # I'm not at all sure what purpose these functions have.
-    'years':        ['%Y', _two_digit_year, _four_digit_year]
+    'years':        ['%Y']
   }
 
   # Whether or not to strip the leading zeros on tick labels.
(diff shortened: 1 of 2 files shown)


bryevdv:
ping @mattpap

bokehjs/src/coffee/models/formatters/datetime_tick_formatter.coffee
```diff
@@ -77,26 +61,26 @@ class DatetimeTickFormatter extends TickFormatter.Model
...
-    'years':        ['%Y', _two_digit_year, _four_digit_year]
+    'years':        ['%Y']
   }
 
   # Whether or not to strip the leading zeros on tick labels.
   strip_leading_zeros: true
 
   initialize: (attrs, options) ->
     super(attrs, options)
+    @formats = _.extend({}, @_formats, @get("formats"))
```
mattpap: I presume this is intentional, but let me reiterate. This can be viewed as `@set("formats", _.extend(...))`, so you are modifying user's input and (when using server) pushing changes to the client. I would say a computed property should have been defined here (e.g. `computed_formats`). In this case it may be harmless to leave it as-is. However, in general I think we shouldn't modify user's input like this. This is specifically visible in `Plot.renderers` where we mix all kinds of stuff and in the end, it requires special serialization code to make it work (among other issues).

bryevdv: What's actually different about this case is the the user's input is not expected it be *all* formats. Rather, the users inputs are assumed to be just a delta that is overlaid on top of the base defaults. I think it might currently work to just set the defaults in python as is done with tool overlays now. I will try this first.

bryevdv:
@mattpap I made it work the way other models currently work. Setting just one field of dict triggers the right sort of update (it did not in the past, which is why it was the way it was). For example, adding this
```
plot.xaxis[0].formatter.formats['months'] = ['%m/%Y FOO']
```
to the weather app yields 
screen shot 2016-04-22 at 5 15 24 pm

I think we can agree this is better?

bryevdv:
OK, I think this should be pretty uncontroversial, and an improvement that makes this case more consistent with other similar cases.

PR #4223, opened 2016-04-23
send date without TZ as-is, do not assume gmt

### LINKED ISSUE #4139: Bokeh Datatable renders dates one day off.
I am using Bokeh 0.11.0

When I produce a data table with dates, the rendered dates are one day off from the source of the data.  Taking the example from the bokeh documentation page (http://bokeh.pydata.org/en/0.11.1/docs/user_guide/interaction.html). 

capture

If you look at the actual dates inside data, it should be 04/01/2014 to 03/102/014 (not 02/28/2014 to 03/09/2014).

capture2

bryevdv:
- [x] issues: fixes #4139
- [x] tests added / passed

This PR changes the JSON encoder to not use `calendar.timegm`, which converts to local timestap, assuming that the datetimes provided are already GMT, causing them to be offset incorrectly.

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -6,10 +6,10 @@ from __future__ import absolute_import
 import logging
 log = logging.getLogger(__name__)
 
-import calendar
 import datetime as dt
 import decimal
 import json
+import time
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


bryevdv:
With this code:
```
from datetime import date
from random import randint

from bokeh.io import output_file, show, vform

data = dict(
    )

columns = [
    ]

show(vform(data_table))
```
yields 
screen shot 2016-04-23 at 11 32 27 am

bryevdv:
ping @bokeh/dev for comment and discussion

bryevdv:
OK I am going to go ahead and merge this, I think there are still more review and improvements to make around our datetime handling, but sending `date`/`datetime` as-is (absent any tz info) should hopefully be an uncontroversial change.

PR #4225, opened 2016-04-24
Fix #4224: make extra_x_ranges accept CategoricalAxis / FactorRange

### LINKED ISSUE #4224: extra_x_ranges should allow Range not only Range1d
bokeh 0.1.1

While `extra_y_ranges` allows Categorical or Linear axis, `extra_x_ranges` only allows for `Range1d`:

```python

plot = figure()
plot.extra_x_ranges['x2'] = FactorRange('foo', 'bar')
plot.extra_y_ranges['y2'] = FactorRange('bar', 'baz')
show(plot)
```

Acceptance criteria:
* Fix the issue
* Add unit tests

rs2:
- [x] issues: fixes #4224 
- [x] tests added / passed

### DIFF
--- a/bokeh/tests/test_objects.py
+++ b/bokeh/tests/test_objects.py
@@ -8,9 +8,6 @@ from bokeh.model import Model, _ModelInDocument
 from bokeh.document import Document
 from bokeh.core.property_containers import PropertyValueList, PropertyValueDict
 from bokeh.util.future import with_metaclass
-from bokeh.models import Range1d, FactorRange
-from bokeh.plotting import figure, show
-
 
 def large_plot(n):
     from bokeh.models import (Plot, LinearAxis, Grid, GlyphRenderer,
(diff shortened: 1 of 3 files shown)


bokeh/tests/test_objects.py
```diff
@@ -639,5 +642,42 @@ def mutate(x):
...
 
+
+class BaseTwinAxis(object):
+    """Base class for testing extra ranges"""
+
+    def verify_axis(self, axis_name):
+        plot = figure()  # no need for setUp()
+        range_obj = getattr(plot, 'extra_{}_ranges'.format(axis_name))
+        range_obj['foo_range'] = self.get_range_instance()
+        self.assertIsNone(show(plot))
```
bryevdv: First, tests! This is great, a few suggestions tho. I think this might be a more appropriate place for these tests:

https://github.com/bokeh/bokeh/blob/master/bokeh/models/tests/test_plots.py

Then I'd mention `show` tries to open a browser, I am a little concerned that it might cause some issue in some testing configuration. The validation happens immediately on the set attr, so I think it can simply be omitted. Also as matter of conceptual purity, I think it would be good to avoid imports from `bokeh.plotting` in this tests under `bokeh.models`. May I suggest:
```
def verify_axis(self, axis_name):
     plot = Plot()  # no need for setUp()
     range_obj = getattr(plot, 'extra_{}_ranges'.format(axis_name))
     range_obj['foo_range'] = self.get_range_instance()
```

bryevdv:
LGTM should be able to merge later today!

bryevdv:
Thanks @rs2 !

PR #4228, opened 2016-04-25
Add support for SaveTool (immediate PNG file download)

### LINKED ISSUE #219: preview save tool should auto download-prompt
since apparently this is possible, and then just rename the tool to "save"

mattpap:
This works for most standards compliant browsers (FF, Chrome, Opera, Edge) and IE 10+. Doesn't work in Safari. This gives us >95% coverage of web browser market (as of March 2016). For Safari we could add fallback to "preview/save" or open a new tab with with an image (which could be a universal fallback).

(almost) fixes #219.


bryevdv:
How does this fail on Safari? Perhaps there is some workaround.

mattpap:
I can't test it in Safari (no version for Windows anymore and I don't have on OSX VM yet). It should just not work without out any visible errors.

mattpap:
As I said in the initial commit, for Safari we need to either fallback to preview/save or open a new tab with an image.

bryevdv:
So when I try it on safari, it navigates the current tab to the `data:image/png;base64,<<lots of data>>` URL, replacing the plot with just the image. Opening a new tab would be acceptable, but I think the current behavior that navigates away from the current URL is not reasonable.

bryevdv:
As an aside, the preview save tool modal is not working on Safari, either. It shows up in a div at the bottom of the page.

mattpap:
@bryevdv, that behavior actually makes sense given this implementation. Setting `target="_blank"`, either globally or just in Safari should fix this.

mattpap:
> It shows up in a div at the bottom of the page.

That's broken after adding `bk-root`, see https://github.com/bokeh/bokeh/issues/4058 and related issues/PRs.

bryevdv:
FYI possibly relevant discussion here https://github.com/eligrey/FileSaver.js/issues/12#issuecomment-47247096 maybe new webkit support coming soon...

bokehjs/src/coffee/models/tools/actions/save_tool.coffee
```diff
@@ -0,0 +1,29 @@
...
+    name = "bokeh_plot.png"
+
+    if canvas.msToBlob?
+      blob = canvas.msToBlob()
+      window.navigator.msSaveBlob(blob, name)
+    else
+      link = document.createElement('a')
+      link.href = canvas.toDataURL('image/png')
+      link.download = name
+      link.dispatchEvent(new MouseEvent('click'))
```
birdsarah: What about adding an on_click callback that removes the link element so that this cleans up after itself.

birdsarah:
I missed this PR, and wrote up a code snippet on #4167.

My snippet works on chrome/firefox, and does what @mattpap notes as the fallback for safari - it opens the png in a new browser window that the user then has to manually save. I wouldn't describe it as a great user experience, but as @mattpap noted, it's not a large portion of users.

However, from some quick testing, my snippet doesn't work on IE, but @mattpap's code in this PR does.

I'm not a fan of the preview/save tool. I think this PR is a great improvement. I'm fine with a fallback of "open a new tab with with an image" as suggested by @mattpap, because I think it's the same number of clicks for those users who have to have a fallback, and a usability improvement for a large chunk of users.

Of course, I'd like to see some tests for this PR. But, testing file downloads is hard.

bokehjs/src/coffee/models/tools/actions/save_tool.coffee
birdsarah: Edit: ignore

Tested, and this doesn't seem to leave any cruft behind, so nothing to clean up.

bryevdv:
Here is my proposal:

Set `target="_blank"` globally, and remove and replace the old save tool entirely.

mattpap:
> Set target="_blank" globally, and remove and replace the old save tool entirely.

I will verify this actually works next week.

mattpap:
We could also consider adding a new tool or an option to save tool for saving data from a data source as CSV, etc. (if possible). I'm pretty sure this would be a valued tool in the area of reproducible research.

bryevdv:
I tested `target="_blank"` in the CSV export PR on OS X  Safari / FF / Chrome

bryevdv:
> We could also consider adding a new tool or an option to save tool for saving data from a data source as CSV, etc. (if possible). I'm pretty sure this would be a valued tool in the area of reproducible research. 

This is at least demonstrated in #4207 but agreed a built-in tool could make sense as well

birdsarah:
> I will verify this actually works next week.

@mattpap - I used SauceLab's live testing to test on different windows versions. If you were wanting to test on OSX you could do the same - I can run you through it if you haven't done it before.

darribas:
On `bokeh` 0.12.3 I add the `savetool` to a plot and try to save on Chrome+macOS. The save icon appears as one of the tools as expected but, when I click on it, nothing really happens. Am I doing something wrong/missing anything?

PR #4237, opened 2016-04-27
Maggie m/add chart methods

### LINKED ISSUE #3908: Add xaxis, yaxis accessors to Chart class
From a discussion on the mailing list.

Chart objects do not have an easy, obvious way to get access to the xaxis and yaxis.  For comparison, Figure has a number of convenient property accessors.

- xaxis - returns a list of x axes "above" and "below" if they exist
- yaxis - returns a list of y axes "left" and "right" if they exist
- axis - returns a list of all axes
- legend - returns a list of all legends
- xgrid - returns a list of x grids
- ygrid - returns a list of y grids
- grid - returns a list of all grids

Chart has a few class variables with these names already.
- legend
- xgrid
- ygrid

First question:  What new methods should be added to Chart?
xaxis and yaxis are clear.  But what about legend and grid?  Do the names need to be changed to prevent collision with the class variables?

Second question: What should be returned?  A list, or just the Axis object?
Chart objects only have a one xaxis and one yaxis, so returning a list might not make sense here.

Maggie-M:
issues: partially addresses #3908


bryevdv:
With this code:
```
data = dict(stamp=[
                .33, .33, .34, .37, .37, .37, .37, .39, .41, .42,
                .44, .44, .44, .45, .46, .49, .49],
            postcard=[
                .20, .20, .21, .23, .23, .23, .23, .24, .26, .27,
                .28, .28, .29, .32, .33, .34, .35],
            )
line = Step(data, y=['stamp', 'postcard'],
            dash=['stamp', 'postcard'],
            color=['stamp', 'postcard'],
            title="U.S. Postage Rates (1999-2015)", ylabel='Rate per ounce', legend=True)
output_file("test.html")
line.circle([5], [0.3], size=20)
show(line)
```
I get the following output:
screen shot 2016-05-02 at 5 21 04 am

So I think this is ready/working? It might be nice to add a better example, but I am not sure what one would be, so I would be ok with putting off adding an example into another issue/PR.

@bokeh/dev thoughts?

bryevdv:
So the original issue also asked about `xgrid`, `xaxis`, etc, unfortunately I think that will take a little more care/thought. I'd like to just merge this as-is and deal with the rest in a separate PR. @bokeh/dev thoughts?

fpliger:
> So the original issue also asked about xgrid, xaxis, etc, unfortunately I think that will take a little more care/thought,, given the unfortunate name conflicts. I'd like to just merge this as-is and deal with the rest in a separate PR. @bokeh/dev thoughts?

+1... just to be more explicit, that would mean `grids`, `axes`, `legend` and `grid`. Right?

bryevdv:
Yes although after your reminder about deprecated properties I think we can get it all in this PR.

bryevdv:
I did some investigation, I am not sure that the clashing properties ever could have worked as properties, because the grids, legends, etc get added early. Setting `xgrid` to `False` would happen too late and have no effect. Accordingly, I have just removed them. Thoughts, @bokeh/dev  ?

fpliger:
> I did some investigation, I am not sure that the clashing properties ever could have worked as properties, because the grids, legends, etc get added early. Setting xgrid to False would happen too late and have no effect. Accordingly, I have just removed them. Thoughts, @bokeh/dev ?

I'd suggest changing the API to make it match plotting as much as possible on `0.12`.. that's actually something that will break the API if we do change and it's definitely better to do it now. We could change those props to turn the way they are in `Figure`, capiture them in the constructor and handle the if needed and then just expose the setters/getters. The `xgrid`/``ygrid` is special because we are overriding the name and those will impact the call of  `create_grids` method but we could do the same if we actually do the same logic behind  a `setter` property. 

I do agree though that should have a dedicated PR.

bryevdv:
@fpliger This PR makes `Chart` and `Figure` *exactly* the same. My comment was that this:

* doing this "breaks" the previous sense of `Chart.xgrid`, `Chart.ygrid` and `Chart.legend`

HOWEVER

* there's no way those attributes could have worked in the past (so less reason not to remove / change them)

bryevdv:
FWIW I think it's better to make this change to make `Figure` and `Chart` align more atomically in one PR.

fpliger:
> 
@fpliger This PR makes Chart and Figure exactly the same. My comment was that this:

> doing this "breaks" the previous sense of Chart.xgrid, Chart.ygrid and Chart.legend
>HOWEVER

>there's no way those attributes could have worked in the past (so less reason not to remove / change them)

Ok. that's correct. Sorry, your comment  + removing the `xgrid`, `ygrid` and `legend` made me thing the opposite 😄 

So the plan is to go ahead and align them with `Figure` API, without deprecation cycle (since they didn't work as properties anyway). Right?

bryevdv:
That would be my vote. We could add a "no-op" setter with a warning, that would be the most conservative approach.

fpliger:
> That would be my vote. We could add a "no-op" setter with a warning, that would be the most conservative approach.

+1

bryevdv:
Added the setters with warnings:
```
In [11]: from bokeh.charts import Scatter, output_file, show

In [12]: from bokeh.sampledata.autompg import autompg as df

In [13]: p = Scatter(df, x='mpg', y='hp', title="HP vs MPG",
            xlabel="Miles Per Gallon", ylabel="Horsepower")

In [14]: p.xgrid = False
/Users/bryan/work/bokeh/bokeh/charts/chart.py:104: UserWarning: Non-functional 'xgrid' setter has been removed; use 'xgrid' keyword argument to Chart instead
  warnings.warn("Non-functional 'xgrid' setter has been removed; use 'xgrid' keyword argument to Chart instead")

In [15]: p.xgrid
Out[15]: [<bokeh.models.grids.Grid at 0x108e7b518>]
```
Will merge when green.

PR #4239, opened 2016-04-27
Add a prominent banner to previous and development docs

### LINKED ISSUE #4100: Add prominent banner to old versions of the docs
An example here from Django:

https://docs.djangoproject.com/en/1.7/ref/templates/builtins/

Would suggest something even taller / brighter. Text similar to:
```
This page refers to a past release of Bokeh. For the latest released docs, navigate HERE
```

bryevdv:
issues: fixes #4100

### DIFF
--- a/sphinx/source/bokeh_theme/static/js/main.js
+++ b/sphinx/source/bokeh_theme/static/js/main.js
@@ -12,6 +12,21 @@ $(document).ready(function() {
     });
   });
 
+  $.get("/en/latest/version.txt", function(data) {
+    v = $.trim(data)
+    style = "font-family: sans-serif;font-size: 16px;width: 100%;text-align: center;background-color: pink;color: firebrick;padding: 5px;padding-top:12px; padding-bottom:12px;"
+    loc = window.location.pathname;
+    dev = /^\/en\/dev/
+    if (dev.exec(loc)) {
+      content = $('<div>The documentation on this page refers to a DEVELOPMENT VERSION. For the latest release, go to <a href="http://bokeh.pydata.org/en/latest/">http://bokeh.pydata.org/en/latest/</a></div>').attr('style', style);
+      $('html:first').prepend(content);
+    }
+    else if (v != window._BOKEH_CURRENT_VERSION) {
+      content = $('<div>The documentation on this page refers to a PREVIOUS VERSION. For the latest release, go to <a href="http://bokeh.pydata.org/en/latest/">http://bokeh.pydata.org/en/latest/</a></div>').attr('style', style);
+      $('html:first').prepend(content);
+    }
+  })
+
   var loc = window.location.pathname;
 
   if (loc.split('docs/').length == 1 ) {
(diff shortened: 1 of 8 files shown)


bryevdv:
I also took this PT to silence some sphinx warnings and errors that had crept in.

bryevdv:
@birdsarah @bokeh/dev any comments welcome, otherwise I will merge late today or early tomorrow.

bryevdv:
Tested `fab latest`

sphinx/source/bokeh_theme/static/js/main.js
```diff
@@ -12,6 +12,21 @@ $(document).ready(function() {
...
+    style = "font-family: sans-serif;font-size: 16px;width: 100%;text-align: center;background-color: pink;color: firebrick;padding: 5px;padding-top:12px; padding-bottom:12px;"
+    loc = window.location.pathname;
+    dev = /^\/en\/dev/
+    if (dev.exec(loc)) {
+      content = $('<div>The documentation on this page refers to a DEVELOPMENT VERSION. For the latest release, go to <a href="http://bokeh.pydata.org/en/latest/">http://bokeh.pydata.org/en/latest/</a></div>').attr('style', style);
+      $('html:first').prepend(content);
+    }
+    else if (v != window._BOKEH_CURRENT_VERSION) {
+      content = $('<div>The documentation on this page refers to a PREVIOUS VERSION. For the latest release, go to <a href="http://bokeh.pydata.org/en/latest/">http://bokeh.pydata.org/en/latest/</a></div>').attr('style', style);
+      $('html:first').prepend(content);
```
birdsarah: I've actually never used the `:first` pseudo-selector before and had to look it up. According to https://developer.mozilla.org/en-US/docs/Web/CSS/:first it's not supported on firefox. 

What about just  `$('body').prepend(content);`?

birdsarah:
There seem to be unrelated changes in here, like the ellipse stuff.

This is big an improvement, thanks.

birdsarah:
Random thought: should we open a follow-on issue to have it try to take you to the same page on the latest version? e.g. if you're on http://bokeh.pydata.org/en/0.11.0/docs/reference/models/formatters.html it points you to http://bokeh.pydata.org/en/0.11.1/docs/reference/models/formatters.html

sphinx/source/bokeh_theme/static/js/main.js
bryevdv: I did just try it and it worked fine at least on the FF I have installed

bryevdv:
I thought about that, but it also occurred to me that the old page may no longer exist in the new docs so agree that a better 404 is needed first

bryevdv:
> There seem to be unrelated changes in here, like the ellipse stuff.

Yes, there were 5-6 sphinx warnings and errors that crept in, the changes to fix them were tiny, so I just put them in here since I was already working in the docs.

bryevdv:
@birdsarah made the change you recommended and tested (see http://bokeh.pydata.org/en/dev/)  will merge when green unless there are other comments.

bryevdv:
Another possible issue to make: we could put in a check that the docs build has no errors or warnings, so that the little things don't collect.

birdsarah:
> Yes, there were 5-6 sphinx warnings and errors that crept in, the changes to fix them were tiny, so I just put them in here since I was already working in the docs. 

GREAT - I noticed these the other day.

> Another possible issue to make: we could put in a check that the docs build has no errors or warnings, so that the little things don't collect. 

+1

PR #4245, opened 2016-04-28
Add GridSpec and plotting/file/graphs example

### LINKED ISSUE #4244: Add example with drawing networkx graphs

mattpap:
- [x] preliminary implementation of `GridSpec` (matplotlib inspired)
- [x] `plotting/file/graphs`

`GridSpec` allows to create a grid plot without dealing with lists of lists manually. Instead you use `row, col` indexes or slices. The proper implementation would allow plots spanning multiple positions (again using slices) (outside of scope of this PR, because grid plot doesn't support spans yet).

fixes #4244


examples/plotting/file/graphs.py
```diff
@@ -0,0 +1,55 @@
...
+        xs.append([x0, x1])
+        ys.append([y0, y1])
+    f = figure(width=300, height=300,
+               x_axis_type=None, y_axis_type=None,
+               outline_line_color=None,
+               tools=[], toolbar_location=None)
+    f.multi_line(xs, ys, line_color="black")
+    f.circle(vx, vy, size=16, line_color="black", fill_color=node_color)
+    f.text(vx, vy, text=labels, text_color=text_color,
+           text_font_size="10px", text_align="center", text_baseline="middle")
```
mattpap: When working on this example, a few things bothered me and should be considered for future development. I presume some are known and there might be even issues open for them, but I wanted to list issues that are specific to this particular piece of code in one place. This example btw. took me way too much time to create than I would expect given the higher-level API. I know we could have graph chart implemented, but still we won't have all of them (or at least not now), so we should be able to do more with less typing and thinking using plotting API.

1. Constructing grid plots should more flexible. That's partially fixed in this PR.
2. `figure()` should allow `bare` mode (and maybe more), so that I don't have to disable each feature individually. Pretty much we should have different sets of defaults ready to be used. This is at least partially different from theming (disabling toolbar vs. disabling (changing color of) outline) and very different from usual notion of (property) defaults in bokeh.
3. There should be a way to compose glyphs, so that I could create e.g. encircled text. This may or may not have something to do with annotations. Most likely not. Additionally, e.g. hit testing would work only for entire group.
4. Even besides composite glyphs, centering text as we do it is cumbersome (but maybe that's the right way given issues with computing text dimensions on canvas).
5. There should a way to say that I want certain e.g. line color for a group of glyphs (this is similar yet not equivalent to (3)). I imagine this being configured via context managers.
6. Glyph functions should be more flexible about their inputs (mostly coordinates), so that I don't have to spend half of a function for rearranging data. Accepting points as `(x, y)` pairs is a minimum. At this point I'm not sure how the API will look like (lack of function overriding makes it a bit difficult, given existing API flexibility).

bryevdv:
I get an exception when I run `graph.py` (py 3)
```
bryan@0199-bryanv (git:mattpap/4244_graphs) ~/work/bokeh/examples/plotting/file $ BOKEH_RESOURCES=inline python graphs.py 
  File "graphs.py", line 43, in <module>
    colorings = [ [ color_map[zeta] for zeta in solution ] for solution in solutions ]
  File "graphs.py", line 43, in <listcomp>
    colorings = [ [ color_map[zeta] for zeta in solution ] for solution in solutions ]
  File "graphs.py", line 43, in <listcomp>
    colorings = [ [ color_map[zeta] for zeta in solution ] for solution in solutions ]
KeyError: 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4
```
Running with `python -i`, can observe:
```
>>> color_map
{1: red, -1/2 - sqrt(3)*I/2: green, -1/2 + sqrt(3)*I/2: blue}
```

bryevdv:
If I change the colormap line to:
```
colorings = [ [ color_map.get(zeta, "grey") for zeta in solution ] for solution in solutions ]
```
Then I get 

screen shot 2016-05-02 at 10 57 41 am

I think people will be very excited to see an example like this.

bryevdv:
@mattpap regarding all the observations. I agree alot of those make sense, we should make individual issues for them. Some could possibly be handled by improvements to theming, others might require new API, but seems like could be incremental

mattpap:
> but seems like could be incremental

Of course.

mattpap:
> `1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4`

The problem is in the solution. Running `expand(_, complex=True)` on that number gives `3/4`, which is a nonsense. @bryevdv, can you show me `solutions` (output from `solve()`)? Also, what version of `sympy` do you have installed?

bryevdv:
```
>>> solutions
[(-1/2 - sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, 1, 1, -1/2 - sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, 1, -1/2 - sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, 1), (-1/2 + sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, 1, 1, -1/2 + sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, 1, -1/2 + sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, 1), (1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, -1/2 + sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, -1/2 + sqrt(3)*I/2, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, -1/2 + sqrt(3)*I/2), (1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2), (1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2, -1/2 - sqrt(3)*I/2, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 + sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, 1/4 - sqrt(6 - 6*sqrt(3)*I)/4 + sqrt(3)*I/4, -1/2 - sqrt(3)*I/2), (1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, -1/2 + sqrt(3)*I/2, -1/2 + sqrt(3)*I/2, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, -1/2 + sqrt(3)*I/2, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, 1/4 - sqrt(6 + 6*sqrt(3)*I)/4 - sqrt(3)*I/4, 1/4 - sqrt(3)*I/4 + sqrt(6 + 6*sqrt(3)*I)/4, -1/2 + sqrt(3)*I/2)]
```
and
```
bryan@0199-bryanv (git:mattpap/4244_graphs) ~/work/bokeh/examples/plotting/file $ conda list sympy
# packages in environment at /Users/bryan/anaconda:
#
sympy                     0.7.6.1                  py34_0  
```

bryevdv:
I updated sympy and now it works.

bryevdv:
My only real comments are: is `bokeh.io` the right place for `GridSpec`, and is the name `GridSpec` too similar to "Spec" properties?

mattpap:
> My only real comments are: is bokeh.io the right place for GridSpec,

I implemented it close to `gridplot()`, which raises a question whether it shouldn't be relocated. But yeah. Any suggestions?

> and is the name GridSpec too similar to "Spec" properties?

This is after matplotlib's naming convention. It could get confusing if `GridPlot.children` was changed to support `GridSpec`. Any suggestions (e.g. `GridBuilder`, could be confusing with charts)?

bryevdv:
If it is only useful with `GridPlot` then I would say put it with `GridPlot`. Perhaps split out `bokeh.models.grid_plot` which has `GridPlot` and `GridSpec` ?

After thinking on it some, I don't really have a better name either. The only thing that comes to mind is  `GridProvider`.

bryevdv:
Unless this could somehow be considered or made part of a grid layout in which case there could be `bokeh.layouts.grid_layout`

mattpap:
I moved `GridSpec` to `bokeh.layout`.

birdsarah:
I won't have a chance to look at this properly while I'm on vacation, but one very quick reaction is that it seems confusing to have gridplot and GridSpec. I'm guessing GridSpec is more featureful so why not just combine them?

PR #4250, opened 2016-04-28
add new Histogram stat to expose np.histogram

### LINKED ISSUE #3576: Bokeh Histogram Attribute density doesn't work
The att density (described below) doesn't seem to be working.

screen shot 2016-01-07 at 5 19 32 pm

fpliger:
issues: fixes #3576

Also includes support for list of scalars as bins, to allow customized non uniform bins

### DIFF
--- a/bokeh/charts/builders/histogram_builder.py
+++ b/bokeh/charts/builders/histogram_builder.py
@@ -46,7 +46,7 @@ def Histogram(data, values=None, label=None, color=None, agg="count",
     This chart implements functionality to provide convenience in optimal
     selection of bin count, but also for segmenting and comparing segments of
     the variable by a categorical variable.
-    
+
     Args:
       data (:ref:`userguide_charts_data_types`): the data source for the chart.
         Must consist of at least 2 values. If all values are equal, the result
@@ -116,14 +116,15 @@ class HistogramBuilder(BarBuilder):
     use Freedman-Diaconis rule)
     """)
 
-    density = Bool(True, help="""
+    density = Bool(False, help="""
     Whether to normalize the histogram. (default: True)
 
     If True, the result is the value of the probability *density* function
     at the bin, normalized such that the *integral* over the range is 1. If
     False, the result will contain the number of samples in each bin.
 
-    For more info check ``numpy.histogram`` function documentation.
+    For more info check :class:`~bokeh.charts.glyphs.HistogramGlyph`
+    documentation.
 
     """)
 
@@ -138,7 +139,8 @@ class HistogramBuilder(BarBuilder):
 
     def get_extra_args(self):
         """Build kwargs that are unique to the histogram builder."""
-        return dict(bin_count=self.bins)
+
+        return dict(bin_count=self.bins, density=self.density)
 
     def _apply_inferred_index(self):
         # ignore this for now, unless histogram later adds handling of indexed data
--- a/bokeh/charts/builders/histogram_builder.py
+++ b/bokeh/charts/builders/histogram_builder.py
@@ -21,7 +21,7 @@ the arguments to the Chart class and calling the proper functions.
 from __future__ import absolute_import
 
 from ...models import Range1d
-from ...core.properties import Bool, Int
+from ...core.properties import Bool, Int, Either, Float, List
 
 from ..builder import create_and_build
 from .bar_builder import BarBuilder
@@ -61,7 +61,7 @@ def Histogram(data, values=None, label=None, color=None, agg="count",
       agg (str, optional): how to aggregate the bins. Defaults to "count".
       bins (int, optional): the number of bins to use. Defaults to None to auto select.
       density (bool, optional): whether to normalize the histogram. Defaults to False.
-      
+
       **kw:
 
     In addition to the parameters specific to this chart,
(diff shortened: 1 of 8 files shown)


bokeh/charts/stats.py
```diff
@@ -387,6 +387,59 @@ def sort(self, ascending=True):
...
+        data = self.bin_stat.get_data()
+        bins = self.bin_stat.bin_count
+
+        # Choose bin bounds when data range is ill-defined; pd.cut()
+        # does not handle this well for values that are <= 0
+        if data.size < 2:
+            raise ValueError('Histogram data must have at least two elements.')
+        if data.ndim == 1 and data.std() == 0:
+            margin = 0.01 * abs(float(data[0])) or 0.01
+            bins = np.linspace(data[0] - margin, data[0] + margin, bins+1)
```
almarklein: There is a ref to `pd.cut` in the above comment. Also, it could be that `np.histogram` behaves differently in the presence of ill-defined data. We should probably just add tests for this.

almarklein: > We should probably just add tests for this.

I can add them to this PR, if you don't mind.

fpliger: @almarklein that' would be great, thanks!

almarklein: Done. I verified with the newly added test and with a simple example plot, that the above trickery can be removed, because `np.histogram` deals with it ok. It can even handle data with just one value. So I'd propose to only leave a check for `if data.size ==0: raise ...`.

almarklein: Maybe leave a comment that `pd.cut()` does need extra precaution, in case we switch it again in the future :)

fpliger: Yup, thanks!

fpliger:
I'm tagging it back to WIP since a user request just dropped on gitter asking about support for bins argument as a list of scalars (like in `np.histogram`). It should be quick enough... Will try to add it later tonight or tomorrow.

bokeh/charts/glyphs.py
```diff
@@ -912,13 +913,29 @@ class HistogramGlyph(AggregateGlyph):
...
     bin_width = Float()
-    bin_count = Float(help="""Provide a manually specified number of bins to use.""")
+    bin_count = Either(List(Float), Int, default=None, help="""
+    If bins is an int, it defines the number of equal-width bins in the
+    given range (10, by default). If bins is a sequence, it defines the
+    bin edges, including the rightmost edge, allowing for non-uniform
+    bin widths.
+
+    (default: None, use Freedman-Diaconis rule)
+    """)
```
bryevdv: `bin_count` does not seems like a good name anymore

examples/charts/file/histogram_multi.py
```diff
@@ -8,18 +8,20 @@
...
-hist3 = Histogram(df, values='hp', title="df, values='hp'")
+hist3 = Histogram(df, values='hp', title="df, values='hp'", density=True)
 
 hist4 = Histogram(df, values='hp', color='cyl',
                   title="df, values='hp', color='cyl'", legend='top_right')
 
 hist5 = Histogram(df, values='mpg', bins=50,
                   title="df, values='mpg', bins=50")
+hist6 = Histogram(df, values='mpg', bins=[10, 20, 30, 40],
+                  title="df, values='mpg', bins=[10, 20, 30, 40")
```
bryevdv: Suggest making the example use unequal bins since that is a primary use-case.

bokeh/charts/glyphs.py
bryevdv: Also what interpretation does (did) a `Float` have for `bin_count`?

fpliger: @bryevdv right.. At some point I was renaming `bin_count` to `bins` that is really the best name (just like the `np.histogram` argument) to then realize that there is some work to be done to avoide clashing names (HistogramGlyph has already a property `bins` that actually could be just turned to a `_bins` attribute but then it'd also scaled down to `Stats` that has the same problem... My take was to rename those internal props with some more thinking in a following PR (since it won't break exposed Charts API because `bins` is the name we expose right now). 

That's said I'm ok with doing it in this PR (was trying to switch to a more "small and fast" PRs mindset :) )

examples/charts/file/histogram_multi.py
fpliger: Sure, will do :)

fpliger:
Ok, comments should be addressed...

bokeh/charts/builders/histogram_builder.py
```diff
@@ -111,19 +113,24 @@ class HistogramBuilder(BarBuilder):
 
     """
 
-    bins = Int(default=None, help="""
-    Number of bins to use for the histogram. (default: None,
-    use Freedman-Diaconis rule)
+    bins = Either(List(Float), Int, default=None, help="""
+    If bins is an int, it defines the number of equal-width bins in the
+    given range (10, by default). If bins is a sequence, it defines the
```
almarklein: it says 10 is default, but the default is None, selecting an optimal number of bins.

bokeh/charts/stats.py
```diff
@@ -220,7 +221,14 @@ class BinStats(Stat):
 
     Bin counts using: https://en.wikipedia.org/wiki/Freedman%E2%80%93Diaconis_rule
     """
-    bin_count = Either(Int, Float)
+    bins = Either(Int, Float, List(Float), default=None, help="""
+    If bins is an int, it defines the number of equal-width bins in the
+    given range (10, by default). If bins is a sequence, it defines the
```
almarklein: dito

fpliger:
thanks @almarklein great catch. Should be fixed now

bryevdv:
LGTM and tested locally.

PR #4264, opened 2016-05-02
only wrap autoload_static in an onload handler

### LINKED ISSUE #4254: Notebook rendering is broken
It appears #4189 broke notebook rendering. Perhaps this is because the DOM is already loaded? I was not aware that notebook `show` currently went though the `autoload_js` template. A fix might be to split off the `onload` part into a wrapper that is only applied selectively.

bryevdv:
issues: fixes #4254

This PR changes the embedding code to only wrap the JS for `autoload_static` in a document `onload` handler, and reverts the notebook embedding code to the way it was previously. 

I have verified that the notebook functions with this PR. 

@seanlaw if possible, can you check whether `autoload_static` functions as expected for you? (Looking at the output in an ipython console, it looks good, but independent verification is certainly welcome)

@mattpap I do want to take a deeper dive into all of our embedding/templating soon, but I'd like to get your opinion if this is acceptable as an immediate fix.

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -25,96 +25,94 @@ calls it with the rendered model.
 
 
 #}
-document.addEventListener("DOMContentLoaded", function(event) { 
-  (function(global) {
-    function now() {
-      return new Date();
+(function(global) {
+  function now() {
+    return new Date();
+  }
+
+  if (typeof (window._bokeh_onload_callbacks) === "undefined") {
+    window._bokeh_onload_callbacks = [];
+  }
+
+  function run_callbacks() {
+    window._bokeh_onload_callbacks.forEach(function(callback) { callback() });
+    delete window._bokeh_onload_callbacks
+    console.info("Bokeh: all callbacks have finished");
+  }
+
+  function load_libs(js_urls, callback) {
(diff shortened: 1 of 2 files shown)


mattpap:
@bryevdv, yes, this looks better than the original change.

bokeh/core/_templates/autoload_js.js
```diff
@@ -25,96 +25,94 @@ calls it with the rendered model.
...
+  if (window._bokeh_is_loading === 0) {
+    console.log("Bokeh: BokehJS loaded, going straight to plotting");
+    run_inline_js();
+  } else {
+    load_libs(js_urls, function() {
+      console.log("Bokeh: BokehJS plotting callback run at", now());
+      run_inline_js();
+    });
+  }
+}(this));
```
mattpap: I presume this is just shift left and removal of the wrapper? I just can't read this diff.

bryevdv: Yes, I'm not sure why it diffed so oddly

bryevdv:
@seanlaw it occurred to me I can just build the docs and check the scripts there to validate that the `onload` is present in `autoload_static` since that is what the gallery uses. It is, so I am going to go ahead and merge this. Thanks for all the help on this issue!

Issue #4248, opened 2016-04-28
Line plot scaling issue when WebGL is used 


papahabla:
When enabling WebGL in the following code, the y axis labels are correct, but the scaling of the line plot y data is incorrect. If `webgl=False` or `scatter` is used in place of `line`, the data is presented correctly. 

```python
import numpy as np
from bokeh.plotting import figure, show
from bokeh.io import output_notebook
from bokeh.io import push_notebook

output_notebook()

fs = 48000
t = .1
x = np.random.standard_normal(size=(1, int(np.round(t*fs)))) * 0.25
print(x)
print('MAX ABS: {}'.format(np.max(np.abs(x))))

p = figure(title='Time Signal', plot_height=300, plot_width=800, webgl=True)
r = p.line(np.linspace(0, t, x.shape[1]), x[0, :])
show(p)
```

OS: Mac 10.10.5
Browser: Chrome 50.0.2661.86 (64-bit)
Bokeh: 0.11.1

bryevdv:
ping @almarklein

almarklein:
@papahabla can you please explain "incorrect", maybe via a screenshot?

papahabla:
@almarklein 

Below are images of the above described OS, Browser, Bokeh. 

**Chrome - webgl=False**
![test_bokeh_gl_false](https://cloud.githubusercontent.com/assets/10580407/14992332/80f7d92c-111a-11e6-893c-3711de16fc59.png)


**Chrome - webgl=True**
![test_bokeh_gl_true](https://cloud.githubusercontent.com/assets/10580407/14992385/b903d654-111a-11e6-842e-7b74df70df84.png)


Also, when using Safari with `webgl=True', the plot lines do not draw.

**Safari 9.1 - webgl=True**

![test_bokeh_safari_gl_true](https://cloud.githubusercontent.com/assets/10580407/14992574/79baae0e-111b-11e6-9446-874d766103a7.png)

almarklein:
Ok, this is a known issue in 0.11; sharp corners result in long miters which are not truncated. This was fixed in #3888 (will be in 0.12). WebGL is know not to work (well) in Safari, unfortunately.

almarklein:
Ok, this is a known issue in 0.11; sharp corners result in long miters which are not truncated. This was fixed in #3888 (will be in 0.12). WebGL is know not to work (well) in Safari, unfortunately.

almarklein:
@papahabla Thanks for the report though!

Issue #4240, opened 2016-04-27
Color Issue w/ IE11 and WebGL


crashMOGWAI:
First off, props to everyone on this library! 

The below snippet works on IE10 & QtWeb as expected. However, once I updated to IE11 the RGB coloring on selection of the plot is messed up. The expected behavior returns when simplifying the plotting, so it seems to be pointing to `ColumnDataSource` but everything traces through fine..so I'm scratching my head. Anyone else see this issue? Any guess as to why this is happening? 

Excuse the hacked code, but its as simple as I could manage while replicating the issue:

```
from bokeh.plotting import ColumnDataSource,figure, show, output_file
import pandas as pd

#init data
FILE = (['foo','foo','foo','bar','bar','bar','baz','baz'])
A = ([5,5,6,6,8,8,5,5])
B = ([1,1,2,2,3,3,1,1])
c1 = (100, 50, 10)
c2 = (200, 100, 0)
c3 = (120, 95, 50)
#build and name index using data
index = pd.MultiIndex.from_arrays([FILE,A,B], names=['iFILE','iA','iB'])

init = pd.DataFrame({'X': [ 0, 1, 2, 3, 4, 5, 6, 7],
                   'Y': [ 1, 2, 3, 4, 5, 6, 7, 8],
                   'HUE': [c1,c1,c2,c2,c3,c3,c1,c1],
                   'dFILE':FILE,
                   'dA': A,
                   'dB': B}, index=index)
                   
#group data                   
df = {} 
group = init.groupby(level=('iA','iB'))
for name, group in group:
    uniA = name[0]
    uniB = name[1]
    k = int(uniA), int(uniB)  
    df.update({k:group})                   
                   
#configure plot
TOOLS='resize,crosshair,pan,wheel_zoom,tap,hover'
p = figure(tools=TOOLS, 
           webgl=True)
for perPt in df:
    dataSource = ColumnDataSource(df[perPt])
    color1 = df[perPt]['HUE'].head(1)[0]
    color2 = df[perPt]['HUE'].head(1)[0]
    p.circle(x='X', y='Y', source=dataSource,
             size=10, line_width=4,
             fill_color=None, line_color='lightgrey',        
             # set visual properties for selected glyphs
             selection_alpha=1,
             selection_fill_color= color1,
             selection_line_color= 'yellow',                      
             # set visual properties for non-selected glyphs
             nonselection_alpha=1,
             nonselection_fill_color=color1,
             nonselection_line_color=color2)
[...]

Issue #2683, opened 2015-08-05
Our Webgl support does not work very well on Safari


almarklein:
On Safari, our webgl examples do not always work. Not sure if this is a bug of Safari or on our side. Need more datapoints.

Test pages:

* To test that webgl works: https://get.webgl.org/
* To test Bokeh webgl examples: https://dl.dropboxusercontent.com/u/1463853/bokeh/index.html 

Collecting what we know:

* On my OSX 10.6.8 (snow leopard) VM everything works as expected.
* On Sarah's machine (Mavericks) even the simple webgl example crashes Safari. (http://pastebin.com/NMg325hE)
* On Sarah's machine with intel card selected the iris example works, the others not (Program has unset variables error)
* On Sarahs' ipod (Yosemite) webgl works. Bokeh demos not sure (but might be a bit heavy ;))
* On Bryan's machine (Yosemite), webl works, but our Bokeh examples do not.
* On OSX and Chrome, everything works fine (which suggests its not purely a driver issue).

Conclusion so far: looks like something goes wrong between Safari and the driver, and the way we use webgl triggers that (in some situations).

birdsarah:
Clarifications:
* I have not tested bokeh on the ipod. Only webgl support. Sorry for confusion.
* On intel, same errors were visible for all plots. But only iris worked - something different between circle and marker?

almarklein:
Thanks for the clarifications. The circle and marker are the same, except for handling of `angle` (circle has no angle) and `radius` (circle allows radius in screen units). But in these examples we use none of that...

jerlich:
On 10.10.5 with safari Version 9.0.1 (10601.2.7.2) bokeh webgl examples do not work. They don't show any data points. 

On Chrome it works.

The demo at get.webgl.org works.

almarklein:
Documenting a log by @birdsarah 

LINE 10k
```
[Log] Bokeh: setting log level to: 'info' (line10k.html, line 20)
[...]

PR #4269, opened 2016-05-03
Make webgl work again

### LINKED ISSUE #4265: WebGL broken because public props do not exist anymore on the glyph
Previously, glyphs that support webgl notified the webgl part that a visuale (e.g. color) had changed, using `set_visuals_changed()`. ~~It appears that this is no longer done, probably because it got removed with the properties refactoring.~~ Looking into it.

almarklein:
* Fixes #4265 
* Partly address #4240
* Partly address #4279
* Leaving as WIP to include a fix for #4248 if necessary

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/bokehgl.coffee
+++ b/bokehjs/src/coffee/models/glyphs/bokehgl.coffee
@@ -65,7 +65,7 @@ attach_color = (prog, vbo, att_name, n, visual, prefix) ->
     m = 4
     colorname = prefix + '_color'
     alphaname = prefix + '_alpha'
-    
+
     if not visual.doit
       # Don't draw (draw transparent)
       vbo.used = false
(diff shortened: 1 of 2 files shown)


bryevdv:
LGTM (examples work on non-Safari) so merging now. 

@almarklein is there an issue to keep investigating safari and webGL? Alternatively, can we simply ignore `webgl=True` on safari, if we know we can't fix it?

almarklein:
Yes we have #2683, this is still open. I remember putting some test html docs up somwhere to "crowdsource" some experiences. I'd rather not ignore `webgl=1` even if we know webgl wont work, because we might create the illusion that webgl is running while its not. We could produce a warning in the console when webgl is used on Safari, but that would require browser detection :(

bryevdv:
Sorry, I did mean ignore in the sense of "but issue a console warning"

PR #4274, opened 2016-05-03
TapTool improvements and new bokehjs example

### LINKED ISSUE #4261: There should be an actual click tool
Currently we have tap tool which acts as a hybrid of select and inspect tool. It modifies selection on associated data sources and executes a callback. I would like to have a tool (or a mode) where I could execute a callback only. One option is to add (e.g.) `ClickTool` or add `select = Bool(True)` to `TapTool`. Perhaps we can get both sides by in one change by adding the later and making `click` tool alias to set `select=False`.

mattpap:
- [x] add click behavior to `TapTool`
- [x] add `HitTestResult` class in place of raw object data structure
- [x] add `bokehjs/examples/donut`
- [x] make `ImageURL` work again for unspecified width and height
- [x] don't fetch undefined urls in `ImageURL`
- [x] expose `sprintf()` in JS API
- [x] ~~add data attribution~~ (#4284)

fixes #4261


bokehjs/test/test_common/test_selection_manager.coffee
```diff
@@ -77,9 +77,6 @@ describe "SelectionManager", ->
 
   sm = new SelectionManager()
 
-  it "should have a reference to an empty selection", ->
-    expect(sm.empty).to.deep.equal empty_selection
```
mattpap: @birdsarah, in the context of our yesterday's discussion, here's a nice example of a test that does nothing, yet exposes internal implementation of the class and gets in the way of modifying it. Additionally, `empty` should be a private member of `SelectionManager`.

birdsarah: Sure. I don't disagree.

bryevdv:
LGTM, I do think the interaction on the example is pretty non-intuitive, particularly the pausing. I had to look at the code to make sure it was behaving as expected.

bokehjs/test/test_common/test_selection_manager.coffee
birdsarah: It ocurrs to me that perhaps we're talking past each other because we see the problem differently. 

My understanding of your concern - **please correct me** - is the build up of unhelpful tests that cause a maintenance burden without adding value. 

My concern is that we don't have sufficient tests or a culture of writing tests. The result is frequent regressions, high manual testing burden, and a very high barrier for new contributors (because you can't know without lots experience whether you might be breaking something). For me, I am ok with building up some unhelpful tests if overall we have can improve against the goals of regressions, manual testing, test-driven development (rather than experience-driven) (sorry to overload a term).

I know that lots of unhelpful or flaky tests can erode confidence also, but tests are the only way I know to improve the goals I've mentioned.

I'd be really interested to know if you think I've characterized your concerns correctly, and for you to elaborate on those concerns. We could also move this comment to a more appropriate issue.

mattpap:
@bryevdv, yes it's very intuitive, but for now serves the purpose it was intended for. I didn't have much time to polish it. In longer term perspective it should support drill down functionality.

bokehjs/test/test_common/test_selection_manager.coffee
mattpap: >  (...) build up of unhelpful tests that cause a maintenance burden without adding value.

Correct.

> (...) we don't have sufficient tests or a culture of writing tests

Correct.

mattpap: > We could also move this comment to a more appropriate issue

@birdsarah, what you wrote is mostly correct. I will comment and expand in a more appropriate place.

PR #4276, opened 2016-05-03
Remove problem styles for SASS/Autoprefixer support

### LINKED ISSUE #4275: Remove problem styles for SASS/Autoprefixer support
* When processing `bokeh.css` with sass, there is a bad line that sass breaks on here https://github.com/bokeh/bokeh/blob/master/bokehjs/src/less/continuum.less#L371
```bash
Error in plugin 'sass'
Message:
    static_src/scss/bokeh/bokeh-0.11.0.css
Error: style declaration must contain a value
        on line 3200 of static_src/scss/bokeh/bokeh-0.11.0.css
>>   top: ;
   ------^
```

* For a compiled css file with bokeh included, tools like [autoprefixer](https://github.com/postcss/autoprefixer) will yell at you for having non-final spec style here https://github.com/bokeh/bokeh/blob/master/bokehjs/src/less/flexbox.less#L19
```bash
gulp-autoprefixer:
  autoprefixer: /vagrant/frontend/static_src/scss/app.css:24997:5: You should write display: flex by final spec instead of display: box
  autoprefixer: /vagrant/frontend/static_src/scss/app.css:25013:5: You should write display: flex by final spec instead of display: box
```

mistakevin:
- [x] issues: fixes #4275

### DIFF
--- a/bokehjs/src/less/continuum.less
+++ b/bokehjs/src/less/continuum.less
@@ -367,9 +367,6 @@
         clear: both;
       }
       &.bk-bs-dropdown {
-        &:before {
-          top: ;
-        }
         &:after {
             content: " \2014";
             float: none;
--- a/bokehjs/src/less/flexbox.less
+++ b/bokehjs/src/less/flexbox.less
@@ -16,11 +16,6 @@
   /* IE 10 */
   display: -ms-flexbox;
 
-  /* old */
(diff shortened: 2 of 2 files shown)


bryevdv:
AFAICT this is ok, so I will merge it.

Issue #149, opened 2013-11-01
Allow plot titles to be located on any edge of the plot (not just the top)


pzwang:

birdsarah:
@bryevdv - do you have a S/M/L estimate on this? - currently title placement is horrid.

bryevdv:
small to medium

mattpap:
Shouldn't we get layout working for entire plots, so that issue is trivially fixed?

bryevdv:
This use-case will help get layout working.

bryevdv:
Edit: to elaborate a bit more: Layout *is* working. It is just not applied in as many places as it could be (For instance, to position title labels)

birdsarah:
Done with #4392.

Issue #4267, opened 2016-05-03
User guide adding annotations uses p.renderers.extend instead of p.add_layout


birdsarah:
http://bokeh.pydata.org/en/0.11.1/docs/user_guide/plotting.html#adding-annotations

The source examples:
* https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/source_examples/plotting_span.py
* https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/source_examples/plotting_box_annotation.py

both use 

````
p.renderers.extend([daylight_savings_start, daylight_savings_end])
```

where as https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/source_examples/plotting_label.py 

uses the plot api 

```
p.add_layout(labels)
p.add_layout(citation)
```

I think we should encourage the use of p.add_layout, and update the user guide accordingly.

birdsarah:
Note I updated this issue from `add_annotation` to `add_layout` as I'm removing the `add_annotation` method in a PR I'm working on.

bryevdv:
Should get a deprecation?

birdsarah:
No. See note here: https://github.com/bokeh/bokeh/commit/585d572ad29558909477036939fd18aa7acdf65e

birdsarah:
TL;DR - add_annotation has never been released.

birdsarah:
I was looking at master examples of docs when I spotted the `add_annotation` method and I was looking at 0.11.1 when I spotted the `p.renderers.extend` which originally gave me the impression that it had been released, but I checked and it hasn't.

bryevdv:
Ok agree then. I thought it was in 0.11.1

Issue #2865, opened 2015-09-16
Off-canvas legends


birdsarah:
We get asked a lot about moving legends off the plot.

At the moment there's no way to do that.

Looking at matplotlib & ggplot2:

* matplotlib: perhaps not as quick & easy as it should be: http://stackoverflow.com/questions/4700614/how-to-put-the-legend-out-of-the-plot#4701285
* ggplot2: the default left, right, top, bottom are off canvas but you can specify a position to have it on canvas too. http://docs.ggplot2.org/current/theme.html (about half way down the page)

As an additional requirement, I can also imagine, as a web person, wanting broader control over where a legend goes on your page.

For Bokeh, I think there's short-term and long-term solutions.  In the short-term maybe it would be nice to have the option to get a legend as plot object that you could at least pass to a `gridplot` or `components` method and then put it where you want. (maybe this is already possible and I haven't tried yet)

VelizarVESSELINOV:
:+1: 
An old hope:
http://stackoverflow.com/questions/26254619/position-of-the-legend-in-a-bokeh-plot

bryevdv:
There are to different use-cases, I just want to make sure to draw out the distinction:

*  A legend outside the main plot area, but still on the same canvas. 

* A legend on a completely different canvas, positionable as a DOM element.

Both of these are waiting on some work to clean up, update, and expand the low level canvas layout code, which will be a priority after `0.11`

VelizarVESSELINOV:
For me ```A legend outside the main plot area, but still on the same canvas.``` is covering very well my current use-cases.

bryevdv:
Happily, that's probably then one that will happen first. :)

kynan:
Are there plans to support this for the charts API? Or is there a way of doing that already? I failed adapting the [example from the user guide](http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#outside-the-plot-area).

birdsarah:
@kynan - I would suggest opening a new issue. It is possible, but it would be very hacky.

kannanse:
@birdsarah - Can you show an example of doing it the hacky way? Has #4941 been implemented??

I tried chart.legend.location = (0.0,-30.0), but seems to remove the legend completely.

antoniomolram:
as @kannanse says, it would be really helpful to have the feature as he said. It is much more clean as to create the Legend on a separate structure! ;)

PR #4282, opened 2016-05-04
Make title an annotation

birdsarah:
Work towards: #4164 

Fixes: #149, #1085, #2081, #2865, #4267

Includes a generic fix for putting any annotation (including legends) in a side panel - although annotations other than legend and title may need tweaks to work properly.

ToDo:
- [ ] Figure out the API
- [ ] properly calculate extent of a rotated title (wait for @canavandl's rotating label PR to get merged)


bokeh/models/axes.py
```diff
@@ -26,10 +26,6 @@ class Axis(GuideRenderer):
     Ability to hide the entire axis from the plot.
     """)
 
-    location = Either(Auto, Enum(Location), help="""
-    Where should labels and ticks be located in relation to the axis rule.
-    """)
-
```
bryevdv: OK but we should mark the property deprecated and add a no-op (python) property getter and setter so that trying to use `.location` doesn't just fall over with an exception

birdsarah:
Things are coming along nicely, off-plot annotations have basic support, and a new title annotation has been added. Here's the updated models/iris.py example.

screen shot 2016-05-06 at 3 38 46 am

Off-plot legends have basic support although I can think of lots of little pieces of clean-up I'd do.

birdsarah:
So. I know this won't go in as it is, but I'd appreciate some eyes and some help.

I'm happy enough with all the foundational work up to where I added the screenshot in the timeline above. Although I sure there are things to change (as Bryan has already noted).

But the last commit is where I need help from @bokeh/core. 

@bryevdv and I discussed what the API changes would look like when title became an annotation. We agreed that title comes out of Plot, but plotting API should not change. But I'm not sure how to actually accomplish this with properties and getters/setters intermediating between an actual object. Detailed feedback & solutions would be much appreciated.

For now I haven't put in all the deprecation warnings that this PR will need as first I'd like to focus on what the new API is going to be, then I can go back and add any deprecations.

I am VERY unfamiliar with the properties system python and js side so all thoughts are gratefully appreciated.

Finally, I didn't see this coming, but I think we might need to make changes to annotations so that changes to them are triggered by server. I can't figure out how to get the new titles to updated in, for example, examples/models/taylor_server.py

birdsarah:
pinging @havocp @canavandl about the annotation-server connection question above.

birdsarah:
Just had an idea, that doesn't work, but wanted to note it in case it sparks someone's thinking. 
I was thinking that Plot.title needs to be a title object, which should then make it easy to update via server. The problems with that:
* you can add an arbitrary number of titles at the moment and put them wherever you want
* it would make Plot.title and Figure.title have different meanings because the title on figure is really the text string.

:ram: :gem: :gem: :gem:

pzwang:
I don't know what those emoji mean (sheep and diamonds? :-) 

To address the problem of "multiple title objects", you can have the Plot.title refer to the first or the "primary" one. That is about 99% of use cases, and it should be easy to use and explain.  If people want to add multiple annotations to the Plot, then they can use the selector mechanism to get them out.

To disambiguate with Figure.title, we could think about making that be Figure.title_text or something...?  I think it's important to have consistency in the API here, because this is the sort of thing that can confuse the crap out of people.

birdsarah:
:ram: :gem: :gem: :gem: = rambling 

Thanks @pzwang. I like the idea, but I'm not stuck on how it will work in implementation. Figure inherits from Plot and Figure.title currently means the title string. Perhaps we could have a special plot.add_title method that adds the special main title, and sets plot.title to be the title object. We could even have it not be on the main plot.renderers list and be handled specially in plot.coffee. This would let users add plot.title whenever they like and not have it interfere with adding a top axis (a current unfortunate side effect of the newly flexible system - is that axes aren't guaranteed to be stuck to the plot - it depends on the order you add_layouts)

Other thoughts: 
* implement title_padding_x for each side and then title_standoff can be an alias to the new title_padding_inner (this may be better done as part of a future PR with all the layout and alignment stuff)
* implement title_location ('left', 'center', 'right'). Left is the default and the current behavior. `center` will be a short-hand for text_align='center' and position is centered and `right` will be a short-hand for text_align='right' and position is right. You can do all this manually now, but this will be an important time saver, as I want to align my titles in all three positions depending on what I'm up to.

canavandl:
@birdsarah I'll make a quick PR to add a change listener on the label annotation (similar to https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/annotations/span.coffee#L15)

bryevdv:
Properties can have `transform` method that accepts a different type and adapts it to the actual property type. So I guess my initial thoughts at what would be "ideal" would be:
```
class Plot

    # this is the ideal?
    # title = List(Instance(Label))

    # instead
    # title = TitleCollection()
```

Where `TitleCollection` behaves like a `List(Instance(Label))` but has a `transform` that will turn a single string into `[Label("the text")]` 

So you could set the `.title` as a string and get the current behavior, or read and write the property as an actual list of labels to get the new behaviour.

What *woudl* break is reading `.title` and expecting to get a string back. I think that's worth considering.

birdsarah:
> or read and write the property as an actual list of labels to get the new behavior
> What would break is reading .title and expecting to get a string back. I think that's worth considering. 

Yes, absolutely that's my concern. I think it's not only reading title and expecting to get a string back expecting to be able to update the string.

---

Although I said I'm happy with the foundational work, I'm not in love with the current way of spelling `Title`s in this PR. Label's allow for a datasource for text to be a field. I think it's more intuitive for title's to always accept a single string. So, right now there's a python helper that lets you do `Title("My title")` but what it's really doing is changing that into `Title(text=["My title"])`. I'm starting to think that Title shouldn't inherit from Label, but that they should both inherit from a parent `TextAnnotation` class which holds all the rendering and then Label handles label data coming from a source. In fact, what if, we had Label and Labels (or MultiLabel?) annotations. Labels accepts a source and fields (and can't be put on a panel, cos that would be really confusing) and Label accepts values. 

---
I'm currently thinking:

```python
class Plot

    title = Instance(Title)
```
I think title we fix-up the Title API so that `plot.title.text` gives you back the string and you can set the string by saying `plot.title.text = "My title"` (@canavandl I'm hoping you can give me some pointers on making that work).  

Then on Figure, can we have Figure._title which is that actual title object and Figure.title aliases to Figure.title.text. I'm guessing we'll have to alias all the title_text properties too. I'm not exactly sure of the mechanism to do this aliasing elegantly both python and javascript side. ping @bryevdv @mattpap 

I think you add a title by doing `plot.add_title(Title, 'position')`. By doing this you set plot.title, and, as with other layouts it gets added to the list of renderers and the appropriate side renderer. 

Plot.title will be guaranteed to be added last in your list of side renderers so that it is always rendered outside axes and legends and anything else you might have added to side panels. Having played around with the examples in the last commit, I think this will be a useful contract.

bryevdv:
> I think it's not only reading title and expecting to get a string back expecting to be able to update the string.
 
Updating the string would work in what I described:
```
p.title = "foo" 

# transformed into p.title == [Label("foo")]

**Comment by bryevdv** [comment]
It's also fine to make a new `Title` type that is simpler.

**Comment by birdsarah** [comment]
OK. @bryevdv has enlightened me seperately on some inner workings of the properties system that I had no clue about.

So,

```python
Plot
    title = TitleProperty
    title_location = [left, right, above, below] (default: above)
```

.title will be an instance of `Title`, but you'll be able to set it using p.title = 'My title' thanks to a transform method on `TitleProperty` (similar to https://github.com/bokeh/bokeh/blob/master/bokeh/core/properties.py#L1463-L1472)

(There are probably more discussions to have around the remaining title properties but I can make some forward progress for now).

My current thinking is that title is not added to the list of side renderers, but bokehjs checks for it and adds it last to the list of sidepanels at run time based on title_location.

I've made strikethroughs on comments above which are no longer part of my thinking to help keep it clear.

bryevdv:
Along those lines I am not sure title needs `x` and `y`? maybe just text props and a property for alignment within its box (left, center, right) ?

birdsarah:
> Along those lines I am not sure title needs x and y? maybe just text props and a property for alignment within its box (left, center, right) ? 

I disagree. I think there's often times when you want to tweak something visually. I think alignment for left center right should be conveniences around it.

bryevdv:
I guess my concern is how to make them meaningful and consistent in combination. How about a `location` that is either "left" "right" "center", or a percentage (measured access the containing box along the direction of text) And if you need to control the position in the "cross" direction, that could be a percentage along the cross direction

In this case we'd need to pick a subset of text props because I think the baseline and justification props that are already there would be very confusing and should be omitted.

birdsarah:
> I guess my concern is how to make them meaningful and consistent in combination.

Agreed

> And if you need to control the position in the "cross" direction, that could be a percentage along the cross direction

My experience from web development is that both %ages and absolute values are necessary. %ages are usually what you want - especially given that the new layouts will be responsive. But there are always exceptions. Speaking of which, I think that we should be supporting %ages in more locations.

> In this case we'd need to pick a subset of text props because I think the baseline and justification props that are already there would be very confusing and should be omitted. 

I may be in a minority. As a power user of Bokeh, my instinct is that these restrictions will come back to be annoying. I think that all the text properties will already be hidden from plotting users because they will be buried under the title object, so not generally something they're going to be worried about.

I'm now thinking that we should focus conveniences on the plotting and charts API and keep the model more flexible. So we could have `Figure.title_align` (left, center, right) which will set both the x/y and the text_align property in one convenient swoop without handcuffing Title's flexibility.

This may sound crazy, but I'd also be slightly more amenable to restrictions on `Title` if `Label` accepted a string for text, because that would make working with Label easier when Title was too restrictive.

birdsarah:
FWIW, I reckon that baseline is a seldom used property - I don't recall ever seeing a question about it. So I would be fine with getting rid of it. 

That leaves me down to fighting for justification, which I definitely do find useful.

bryevdv:
I think have the `text_align` standard `TextProps` property and separate `title_align` is incredibly confusing as well as easily inconsistent. Set one to left and the other to right, what happens? At *best* you have to draw outside the layout box, but then what's the point of the layout box? These things have to be correctly coordinated into the subset of combinations that are actually meaningful, which means in my mind we absolutely can't expose them all as-is together. 

The `TextProps` align and baseline attributes reference the `x` and `y` positions, but whole point of the layout work is to take burden of those kinds of details away. It's not even clear to me what user-supplied `x` and `y` would even be relative to, exactly. What are specific use cases? Are there examples we can see of other libraries providing this level of complexity for titles, and if so are there examples one anyone actually using it?

Keep the 99% dead simple. If someone really needs this let them lay out an actual `Label` in the "above" area.

bryevdv:
Alternatively:
```
def Title(value, **kw): 
    return Label(text=[value], **kw)

p.title = Title("foo")
```

birdsarah:
`x` and `y` are relative to the panel. The use case is that sometimes you want to line things up differently.

You seem convinced about Title so I'm fine with going along with it, but I will fix up Label to make it more usable, which will meet my needs.

How are we going to handle a subset of text properties on title? Create a new `TitleTextProps`?

For clarity can you list out all the properties you expect (and don't expect) on Title. And while we're at it, what about border and background (which are available on Layout).

birdsarah:
I'm good with the simple function to get us started. I would like Title to pick nice typographic defaults - although we could do that in the simple function to get started.

We'd also need the `p.title_location` (which would default to above), but that's fine.

birdsarah:
Wait, but if we did the simple function how could we do `p.title = "My Title"`.

My brain hurts.

bryevdv:
We can't do that :) I think `Title` (object) is better, the function was just a compromise idea. 

I would add all the text props except `text_align` and `text_baseline`, and just adding them "by hand" would be fine by me.

birdsarah:
> I would add all the text props except text_align and text_baseline, and just adding them "by hand" would be fine by me. 

Are you saying that instead of doing:

```python
class Title
    text = Include(TextProps)
```

I do:

```python
class Title
    text_font = String()
    text_font_size = FontSizeSpec()
    ....
```

bryevdv:
Yes, seems ok to me. Another class is OK but just doing it by hand (since there's only one) is my personal pref.

birdsarah:
I haven't fully thought about it, but it seems like it breaks consistency. You're setting all your usual text properties and you expect to be able to set the next one, and then all of a sudden you can't.

I'm not sure I care, just thinking out loud.

bryevdv:
To be clear about my meanings, when I say "inconsistent" I specifically mean: there are two variables that can be set independently, and there are some combinations of values that are nonsense. I think "surprising" users (what I think you are describing) is also bad but I think it's less bad than not designing out real true potential inconsistencies.

bokeh/models/annotations.py
```diff
@@ -338,6 +339,27 @@ class Label(Annotation):
...
+    text_font_style = Override(default='bold')
+
+    def __init__(self, *args, **kwargs):
+        if len(args) == 1:
+            kwargs['text'] = [args[0]]
+        elif len(args) > 1:
+            raise ValueError("Too many positional arguments, expecting title text only")
+        super(Title, self).__init__(**kwargs)
+
+
```
bryevdv: I think this should not subclass from `Label`, so that it can just have `text = String` and there are no lurking surprises

birdsarah: Given all the other properties suggestions, Title will not be able to subclass from Label.

bokeh/plotting/figure.py
```diff
@@ -49,6 +66,91 @@ def __init__(self, *arg, **kw):
...
+        if title:
+            title = Title(
+                title,
+                text_font=title_text_font,
+                text_font_size=title_text_font_size,
+                text_font_style=title_text_font_style,
+                text_color=title_text_color,
+                text_alpha=title_text_alpha,
+                text_align=title_text_align,
+                text_baseline=title_text_baseline
```
bryevdv: suggest just putting all the `kw.pop`'s inline here instead of creating all these variables that are not used anywhere else

bokeh/plotting/figure.py
```diff
@@ -49,6 +66,91 @@ def __init__(self, *arg, **kw):
...
+
+    @property
+    def title(self):
+        return self._title.text[0]
+
+    @title.setter
+    def title(self, title_text):
+        title = Title(title_text)
+        self.add_layout(title, 'above')
+        self._title = title
```
bryevdv: needs to remove any previous?

birdsarah:
> To be clear about my meanings, when I say "inconsistent" I specifically mean: there are two variables that can be set independently, and there are some combinations of values that are nonsense. I think "surprising" users (what I think you are describing) is also bad but I think it's less bad than not designing out real true potential inconsistencies. 

I'm 100% agreed with your previous statement "I guess my concern is how to make them meaningful and consistent in combination."

I am fine with describing my concern as "surprising".

A question about properties. Is it possible to have a property that "dictates defaults". I'm not quite sure how to explain.

I'm imagining you can manually setup a title like this:

```python
title = Title(x=22, y=1, text='my title', text_align='right')
```
super customized

Or you can use something like this:

```python 
p = figure()
p.title = 'my title'
p.title_align = 'center'
```

This would auto-magically take care of the title's `x` and `text_align` and make a title

```python
title = Title(x=<middle of canvas>, y=0, text='my title', text_align='center')
```

The challenge of course is that `<middle_of_canvas>` isn't known till runtime (see more below).

But you could also do somthing like this (who knows why)

```python 
p = figure()
p.title = 'my title'
p.title_align = 'center'
p.title.text_align = 'right'
```

This would make a title

```python
title = Title(x=<middle of canvas>, y=0, text='my title', text_align='right')
```

Notice the manually provided value of text_align has taken precedence.

Still the same issue as above where `<middle of canvas>` isn't yet known.

I was thinking that `x` should accept a `Number` or `left`, `center`, `right`. But I can see a number of issues, especially when title is on a different side......

OK, I still have more thinking to do.

bokeh/plotting/figure.py
birdsarah: We're not going to be implementing like this, so no worries.

pzwang:
Hey guys, can we do a quick voice chat about this?  I think we could iterate faster that way, and there's enough comments going back and forth that cause me to raise my eyebrows, that I'd like to understand the issues and help try to drive this topic to a conclusion that everyone is happy with.

I haven't read deeply every comment in the thread thus far, but FWIW, I am particular sensitive to issues about internal consistency that then lead to extremely unintuitive behaviors or increased cognitive burden for users.  The Charts and Plotting APIs need to optimize for usability (albeit for different classes of users).  The object-level API is where internal consistency is the primary virtue.

birdsarah:
@pzwang @bryevdv and I are having a call at tomorrow. You're welcome to join. Will send you an invite.

bryevdv:
@birdsarah there is no currently no property delegation or aggregation like that. This is my current proposal:

```
class Title(Annotation):

    def __init__(*args, **kw): 
        if len(args) > 1: 
             raise ValueError()

        if len(args) = 1:
            if 'text' in kw: 
                raise ValueError()
            kw['text'] = args[0]       
         
         super(Title, self).__init__(**kw)

    text = String()

    position = Either(TitlePosition, Percentage, Int)

    text_props = Include(TextProps)

    text_baseline = Override('center')

```
So then: 

* This works as expected: `Title("foo")`
* This also works as expected: `Title(text="foo")`
* Keep all text props, including `align` and `baseline` which determine "anchor point" for `position`
* `position` sets location of anchor point along the direction of text [1]:
 *  `left`, `center`, `right`
 * a percentage
 * a fixed pixel offset
* Title anchor points always centered in the transverse direction (*please* let's keep this simple)

AFAICT there are no inconsistencies or surprises here. However, it *does* mean that it's possible to make the Title draw text outside the layout box. There's no way around this unless we *only* expose a  `left`, `center`, `right` policy that internally drives all those other settings to prevent drawing outside. 

[1] I recognize there are other possibilities here this ignores (vertical text) but AFAIK text rendering on canvas is very primitive. Let the browsers implement more things and we can catch up to that when (if) its ready

bryevdv:
Sorry, hit wrong button.

birdsarah:
superseded by #4305 and future title PRs

PR #4283, opened 2016-05-04
Fix angle attr for Label annotation

### LINKED ISSUE #4278: Label doesn't accept angle property
The following code adds the label, but doesn't rotate it.

```python
misc = Label(x=1, y=1, text=["Some stuff"], angle=np.pi/3)
plot.add_layout(misc)
```
Both the python and bokehjs side have an angle property and the bokehjs seems to attempt to use it (https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/annotations/label.coffee#L105)

canavandl:
- [x] issues: fixes #4278 
- [x] tests added / passed

Now:

screen shot 2016-05-04 at 6 23 00 pm

There is one thing to note (and this is mentioned in the Label `angle` attr docstring):

>The center of rotation for canvas and css render_modes is different. For `render_mode="canvas"` the label is rotated from the top-left corner of the annotation, while for `render_mode="css"` the annotation is rotated around it's center.

Getting rotated label locations to match for css and canvas modes seems very difficult and not worth resolving, so I'm ok with adding a warning.


canavandl:
ping @birdsarah

bryevdv:
FYI the merge conflict is in the screenshot:
```
CONFLICT (content): Merge conflict in tests/integration/annotations/screenshots/base__test_label__test_label.png
```

birdsarah:
Rockin!

I'll check out the code in a moment.

For the merge conflict, I added new screenshots to @mattpap's TS API PR. So you just need to add a new screenshot.

He made a good point which is that we should have toolbars on tests if we don't need them - these are meant to be highly isolated tests that only test the specific feature. I definitely agree. Screenshot tests are horrible for false negatives so we have to be really careful. 

So, although it's a little out-of-scope, I think we should update the integration test in that vain:
1) don't use the plotting api - use Plot, 
2) use Range1d (it's more specific and has less auto-magic)
3) use toolbar_location=None on Plot.

bokehjs/src/coffee/models/annotations/label.coffee
```diff
@@ -143,6 +144,7 @@ class LabelView extends Renderer.View
         'font-size': "#{@_text_font_size[i]}"
         'font-family': "#{@mget('text_font')}"
         'background-color': "#{@visuals.background_fill.color_value()}"
+        'transform': "rotate(#{rad_to_deg(@_angle[i])}deg)"
```
birdsarah: I'm a little confused - is angle always in radians?

I've been hunting around and I think the answer is yes. You can provide it in degrees python side (via AngleSpec), but then it comes down the pipe in radians? 

I just want to confirm this.

bokehjs/src/coffee/models/annotations/label.coffee
```diff
@@ -102,8 +103,8 @@ class LabelView extends Renderer.View
     for i in [0...@_text.length]
       ctx.save()
 
-      ctx.rotate(@mget('angle'))
       ctx.translate(@sx[i] + @_x_offset[i], @sy[i] - @_y_offset[i])
+      ctx.rotate(@_angle[i])
```
birdsarah: Definitely good to move this after the translate. I'm pretty sure I'll be using translate for my side panels work, so this makes a lot more sense.

birdsarah:
Code looks good. A few questions and mostly a note about the screenshot test.

bokehjs/src/coffee/models/annotations/label.coffee
mattpap: > I'm a little confused - is angle always in radians?

See https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/properties.coffee#L224

mattpap: I'm not sure why convert to degrees here. `rotate()` transform supports radians and other (see https://developer.mozilla.org/en-US/docs/Web/CSS/angle).

canavandl: I missed that when I read the spec. Thanks for pointing that out @mattpap

canavandl:
superceded by #4297

Issue #2590, opened 2015-07-20
Ongoing WebGL related dev


almarklein:
This is a meta-issue to keep tack of the tasks related to support for WebGL in Bokeh.

Tasks:

* [x] Set up system by which Bokeh can let glyphs draw themselves in WebGL and have the result blended with the normal canvas.
* [x] Create example to better show off WebGL: @bryevdv suggested a variant of a [sklearn clustering example](http://scikit-learn.org/stable/auto_examples/cluster/plot_cluster_comparison.html) (#2604)
* [x] Make WebGL support work on Safari
* [x] See if we can make plots share a gl context to allow 16+ plots on one page (#2607)
* [x] See if we can still support selections with > 65k points by drawing in multiple passes
* [ ] Use memory buffers instead of normal arrays for data throughout BokehJS (#2204)
* [ ] Track down data processing bottlenecks in BokehJS that prevent taking full advantage of WebGL
* [ ] Maybe support some nonlinear transformations in WebGL?

Supported glyphs:

* [x] line (#2681)
* [ ] images
* [ ] maps
* [ ] ...
* markers:
  * [x] circle
  * [x] square
  * [x] cross
  * [x] x
  * [x] circle x
  * [x] square cross
  * [x] diamond
  * [x] diamond cross
  * [x] triangle
  * [x] inverted triangle

bryevdv:
I think this can be closed. Things remaining have their own issues.

PR #4292, opened 2016-05-06
Add WebGL support for more kinds of markers

almarklein:
Implements part of #2590

Markers that we already have:
* circle
* square

Markers that can be done:
* [x] triangle and inverted_triangle
* [x] diamond
* [x] cross, circle_cross, square_cross, diamond_cross
* [x] x, circle_x, square_x
* [x] asterix

Markers that we can/should (probably) not do:
* annulus - can make the shape, but cannot parametrize at this point and API has no default ratio, so skipping for now
* oval, ellipse - same as annulus, but is also more difficult to make shape (can be done though, I think)
* rect
* bezier stuff, etc.

Other tasks in this PR:
* [x] WebGL was broken again (related to canvas element)
* [x] It appears that all WebGL markers are drawn a few pixels off the correct position
* [x] A (visual) test for the above
* [x] Update docs on WebGL
* [x] Investigate WebGL blurryness

### DIFF
--- a/sphinx/source/docs/user_guide/webgl.rst
+++ b/sphinx/source/docs/user_guide/webgl.rst
@@ -32,10 +32,11 @@ Support
 -------
 
 Only a subset of Bokeh's objects are capable of rendering in WebGL.
-Currently this is limited to the line, circle marker, and square marker. We plan
-to extend the support to more markers, lines, and other objects such
-as maps. You can safely combine multiple glyphs in a plot, of which
-some are rendered in WebGL, and some are not.
+Currently supported are the circle and line glyphs, and all markers
+supported by ``scatter()`` (asterisk, circle, square, diamond, triangle,
+inverted_triangle, cross, circle_cross, square_cross, diamond_cross,
+x, square_x, circle_x). You can safely combine multiple glyphs in a
+plot, even if some are rendered in WebGL, and some are not.
 
 The performance improvements when using WebGL varies per situation. Due
 to overhead in some places of BokehJS, we can currently not benefit
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -39,6 +39,9 @@ class CanvasView extends BokehView
 
     logger.debug("CanvasView initialized")
 
+  get_canvas_element: () ->
+    return @$('canvas.bk-canvas')[0]
+
   get_ctx: () ->
     canvas_el = @$('canvas.bk-canvas')
     ctx = canvas_el[0].getContext('2d')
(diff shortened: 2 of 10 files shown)


almarklein:
Managed to make markers without a fill (cross, x) work too. This means that all markers supported by `scatter()` are supported in WebGL!

almarklein:
Ready from my end. To check this out, best look at the new `plotting/file/marker_compare.py` example.

bokehjs/src/coffee/models/markers/asterisk.coffee
```diff
@@ -1,8 +1,12 @@
 _ = require "underscore"
 Marker = require "./marker"
+bokehgl = require "../glyphs/bokehgl"
 
 class AsteriskView extends Marker.View
 
+  _init_gl: (gl) ->
+    @glglyph = new bokehgl.AsteriskGLGlyph(gl, this)
+
```
bryevdv: If `bokehglk.coffee` just exported the standard type names, without the "GL" tacked in the middle, then I think this repeated boilerplate could be reduced to a single method on `Glyph` base class that just does something like:
```
_init_gl: (gl) ->
  @glglyph = new bokehgl[@type](gl, @)
```
would need to be a little more sophisticated, check / log that GL glyph actually exists, but the point is that that would happin in one place.

sphinx/make2.bat
```diff
@@ -0,0 +1,204 @@
...
+)
+
+if "%1" == "serve" (
+	python docserver.py
+	if errorlevel 1 exit /b 1
+	goto end
+)
+
+
+:end
```
bryevdv: what is this?

sphinx/source/docs/user_guide/webgl.rst
```diff
@@ -32,10 +32,11 @@ Support
...
 Only a subset of Bokeh's objects are capable of rendering in WebGL.
-Currently this is limited to the line, circle marker, and square marker. We plan
-to extend the support to more markers, lines, and other objects such
-as maps. You can safely combine multiple glyphs in a plot, of which
-some are rendered in WebGL, and some are not.
+Currently supported are the circle and line glyphs, and all markers
+supported by ``scatter()`` (asterisk, circle, square, diamond, triangle,
+inverted_triangle, cross, circle_cross, square_cross, diamond_cross,
+x, square_x, circle_x). You can safely combine multiple glyphs in a
+plot, of which some are rendered in WebGL, and some are not.
```
bryevdv: suggest "even when" instead of "of which"

almarklein:
@bryevdv comments addressed. Good point of detecting whether there is a GL glyph available based on name. The `make2.bat` was a leftover from a previous PR that was sitting untracked in my local repo until I accidentally `git add` -ed it. Shall I rebase to clear that from history?

bryevdv:
@almarklein Yes let's kick the tires on PR cleanup rebasing.

bryevdv:
Well now there's conflict to resolve anyway :)

almarklein:
Rebased to 6 commits.

bryevdv:
@bokeh/dev look how nice that looks :)

bryevdv:
I'm tempted to try a squash merge for fun too, but will make a regular merge :)

bryevdv:
@almarklein was there not another issue that was actually completely closed by this?

almarklein:
Maybe you mean #4293? The blurriness should have been improved, but there might still be "unnecessary" blurriness by not taking retina display into account.

PR #4297, opened 2016-05-09
Add anglespec attribute to Label annotation

canavandl:
supercedes: #4283 

- [x] issues: fixes #xxxx
- [x] tests added / passed

There is one thing to note (and this is mentioned in the Label `angle` attr docstring):

>The center of rotation for canvas and css render_modes is different. For `render_mode="canvas"` the label is rotated from the top-left corner of the annotation, while for `render_mode="css"` the annotation is rotated around it's center.

Getting rotated label locations to match for css and canvas modes seems very difficult and not worth resolving, so I'm ok with adding a warning.

todo:
* Need new screenshot but currently struggling to build bokehjs locally.


canavandl:
Here's the selenium base image using the `Plot` api:

screen shot 2016-05-09 at 11 53 51 am

birdsarah:
I would consider adding linear axes to your plot as well. I think it helps for human verification. But I'm very ambivalent.

canavandl:
base image: 

screen shot 2016-05-09 at 4 23 31 pm

bokeh/models/annotations.py
```diff
@@ -271,9 +271,12 @@ class Label(Annotation):
...
     angle = AngleSpec(default=0, help="""
     The angles to rotate the text, as measured from the horizontal.
-
+    
     .. warning::
-        The `angle` property is not supported for `render_mode="css"`
+        The center of rotation for canvas and css render_modes is different.
+        For `render_mode="canvas"` the label is rotated from the top-left
+        corner of the annotation, while for `render_mode="css"` the annotation
+        is rotated around it's center.
```
bryevdv: Is this actually true? I would expect the canvas rotation to be around whatever the baseline/align anchor point is

canavandl: Ah - I believe that's probably correct. I'll test and update the docstring accordingly.

birdsarah:
lgtm

Issue #2504, opened 2015-07-03
Datatables spills out on server rendering (on FF)


birdsarah:
Check out the data table on examples/glyphs/data_tables.py vs examples/glyphs/data_tables_server.py

Something about the server rendering causes slickgrid to set the columns on the data table too wide.

birdsarah:
The first place i would look is the server embed's inject_css method. slick grid could be looking at the width of the element before the bokeh css has kicked in.

birdsarah:
On Ubuntu, FF is affected, but Chrome isn't.

birdsarah:
Not just data tables. Also affects button widget in examples/glyphs/widgets_server.py

(again only in FF)

birdsarah:
Verify as part of #4392

birdsarah:
Fixed in master, probably due to #4343.

Issue #3346, opened 2015-12-15
Slider size from python code to css


Pegase92:
Good afternoon,

Will it be possible to have a new attribute in the Slider object class that may define the size (i.e. the width) , instead of being defined in the bokeh.min.css ? This will be heplfull.

Many thanks

Benoit

brendancol:
I was thinking it would be nice to have a `css_properties` dictionary on some widget base class which would allow users to supply arbitrary css properties to be place at the element level.  User Defined Widgets / Layouts are coming soon, but the code for such an extension would look like this:

```
from bokeh.properties import Dict, String
from bokeh.models.widgets import BaseBox

class StyleableBox(BaseBox):
    __implementation__ = '''
        _ = require "underscore"
        build_views = require "common/build_views"
        ContinuumView = require "common/continuum_view"
        BaseBox = require "widget/basebox"

        class StyleableBoxView extends ContinuumView
          tag: "div"
          attributes:
            class: "bk-vbox"

          initialize: (options) ->
            super(options)
            @views = {}
            @render()
            @listenTo(@model, 'change', @render)

          render: () ->
            children = @model.children()
            build_views(@views, children)
            for own key, val of @views
              val.$el.detach()
            @$el.empty()

            css_props = @mget("css_properties")
            if css_props? then @$el.css(css_props)

            for child in children
              @$el.append(@views[child.id].$el)
            return @

        class StyleableBox extends BaseBox.Model
          type: "StyleableBox"
          default_view: StyleableBoxView

          defaults: ->
            return _.extend {}, super(), {
              children: []
              css_properties: {}
            }

          children: () ->
            return @get('children')

        module.exports =
          Model: StyleableBox
          View: StyleableBoxView
        '''
    css_properties = Dict(String, String, default=None)
```

Any love for adding a couple of element level css properties as a `Dict(String, String)`?

birdsarah:
I'm working on layout which changes this, but i recently added this feature on a branch I was working on.

Here's the commit: https://github.com/bokeh/bokeh/commit/1b050a47a09b5c70137b17ace202027bbf9ad340

birdsarah:
Confirm as part of #4392.

birdsarah:
This is still buggy:

`r_slider = Slider(start=0, end=255, value=init_color_value, step=1, width=600, title="R")`

In notebook:
* If `show(Row(r_slider))` - works and responds to specified width of slider
[...]

Issue #2451, opened 2015-06-25
css - hplot of Paragraph & TextInput


birdsarah:
See: http://nbviewer.jupyter.org/github/birdsarah/bokeh-miscellany/blob/master/electa/Followers-2.ipynb

Maybe I've coded this wrong, but I wanted to use Paragraph as a label for the text input, but it's come out fairly misaligned.

damianavila:
Can you try this outside the notebook? maybe there is some clashing with the css...

birdsarah:
Sorry - also looks bad, just different bad :)

bryevdv:
IMO we should get out of the `Paragraph`, `HBox`, and `VBox` business. I'd like to find ways to make it easier for users to use real HTML templates and layouts, and the new `components` is a great step in that direction. It would be fruitful to see if the same treatment can be applied to the `autoload` functions as well.

birdsarah:
@bryevdv - yes - that sounds great.

bryevdv:
I think all of this is covered by the current ongoing layout work. Please re-open if I am mistaken.

birdsarah:
Re-opening until it's actually fixed!

birdsarah:
Confirm as part of #4392

bryevdv:
I had to fudge some data and make a few other small changes (`CustomJS` instead of `Callback`, use a simple axis type). This is what it looks like on master:
<img width="808" alt="screen shot 2016-06-11 at 9 00 02 pm" src="https://cloud.githubusercontent.com/assets/1078448/15988462/abcda1ec-3017-11e6-8f5f-70ef30dbbf7b.png">

bryevdv:
FWIW it looks the same outside the notebook, and also with `Row` and `Column`

birdsarah:
Agreed - this is not fixed.

bryevdv:
Well I stupidly did not save the updated test example code, and we need to get out a release with the tornado fix ASAP, so moving to `0.12.7`

bryevdv:
Unfortunately the original test case is no longer available, so I am going to close this as `norepro` there is a good chance it is covered by one of the many other overlapping issues.

Issue #2344, opened 2015-05-28
AppVBox: width has no effect (Firefox)


bald:
When the parameter `width` is specified for AppVBox, this has no effect on the output when opened with Firefox (38). Chrome displays fine!

    def layout(app):
        return AppVBox(app=app, children=['ticker1', 'plot'], width = 100)

[Bokeh version 0.9]

birdsarah:
Fixed in #4343

PR #4303, opened 2016-05-09
Add box layout with widgets supported

birdsarah:
DO NOT REVIEW :D 

Continues work on #4164 

Fixes: #2504, #3346, #2344, #1710, #2451

The goal of this PR is to add the initial box layout with a `Row` and `Column`. For this PR, only widgets can be gracefully laid out in a `Row`, and `Column`.

This work builds off #4282, which should be merged first.

This PR also updates the default styling of widgets to try and give a more unified default across plots and widgets.

There may be widgets that still need small tweaks so they play nicely with the new layout.

- [ ] test DataTables
- [ ] test DataTables on server
- [ ] test on server
- [ ] test with components
- [ ] specifying width & height for a widget fixes its width/height
- [ ] fix expanding to fill width/height of window


birdsarah:
Before:

screen shot 2016-05-09 at 1 44 24 pm


This is the before screengrab, which I'm hoping to make look nice with layout.

bokeh/models/annotations.py
```diff
@@ -338,6 +339,27 @@ class Label(Annotation):
...
+    text_font_size = Override(default={'value' : '14pt'})
+
+    text_font_style = Override(default='bold')
+
+    def __init__(self, *args, **kwargs):
+        if len(args) == 1:
+            kwargs['text'] = [args[0]]
+        elif len(args) > 1:
+            raise ValueError("Too many positional arguments, expecting title text only")
+        super(Title, self).__init__(**kwargs)
```
bryevdv: So you can do `Title("foo")` but if you do `Title(text="foo")` it will break in weird ways. My first reaction is to prefer to not subclass from `Label` for this reason. Tho possibly this would be acceptable:
```
def __init__(self, text, **kwargs): 
     # make sure 'text' key not in kwargs
     # make sure text param is a string
     kwargs['text'] = [text]
     super(Title, self).__init__(**kwargs)
```

bokeh/models/plots.py
```diff
@@ -572,6 +533,16 @@ def border_fill(self, color):
             """)
         self.border_fill_color = color
 
+    @property
+    def title(self):
+        return self._title.text[0]
```
bryevdv: This is another reason not to subclass from `Label`, if `Title.text` is just a `String` then this sort of thing would not be needed

bokeh/models/annotations.py
birdsarah: Bryan - this PR has no commits at the moment. It's chained off the Title PR - which still has updates to be made as per our discussions. I'm not sure you want to be commenting on it.

birdsarah:
DO NOT REVIEW :D 

(There's not point in reviewing just yet - there are a lot of moving parts at the moment  -will post another note when it's ready.)

birdsarah:
superseded #4308

Issue #2367, opened 2015-06-05
Different sizes not used for selection & nonselection glyphs


birdsarah:
See the example here:

http://nbviewer.ipython.org/github/birdsarah/bokeh-miscellany/blob/master/notebooks/Different%20size%20for%20selected%20%26%20not%20selected%20demo.ipynb

Simplified for clarity:

``` python
    nonselection = Circle(x='x', y='y', fill_color='gray', size=5)
    selection = Circle(x='x', y='y', fill_color='red', size=25)
    plot.add_glyph(source, nonselection, selection_glyph=selection, nonselection_glyph=nonselection)
```

The size that is used is whatever is passed as the second argument to add_glyph.

bryevdv:
Current behavior is as designed and documented, i.e. only "visual" (line and fill) props are considered for selection/nonselection/hover policies. Changing to `feature` to reflect that it would be new feature development.

bryevdv:
An alternative short-term help would be to issue a validation warning

mattpap:
This is a recurring problem when creating more complex visualizations. Recently I was bit by this in PR #4312, where I wanted to make pie chart's selected slice to stand out, but I can't change selection annular wedge's x, y coordinates, because those are bound to the main glyph.

My proposal is to change definition of `GlyphRenderer` class to:
```
class GlyphRenderer(DataRenderer):
    glyph = Instance(Glyph)
    selection_glyph = Instance(Either(Glyph, Visuals))
    nonselection_glyph = Instance(Either(Glyph, Visuals))
    hover_glyph = Instance(Either(Glyph, Visuals))
```
where `Visuals` is a base class that would be implemented by `<GlyphName>Visuals`. Using `Visuals` instead of a `Glyph` would retain the current, data sharing, behavior. However, using a `Glyph` would mean `GlyphRenderer` would perform all data mapping independently of `glyph`. I think glyphs could implement `Visuals` to reduce code duplication.

bryevdv:
noting that this would fall out of #10403 as I envision it at least

PR #4312, opened 2016-05-11
JS/TS charts API

### LINKED ISSUE #4311: Create JS/TS charts API

mattpap:
This is really preliminary stuff.

- [x] pie chart
- [x] bar chart
- [ ] line chart
- [x] hover centered on slices, not glyph
- [x] hover glyph sometimes doesn't deactivate (tooltip hides properly)
- [ ] hover can affect only certain properties (e.g. can't change `AnnularWedge`'s radius) (#2367)
- [x] tooltip attachment in the vertical dimension
- [x] `Quad` supports categorical axes
- [x] allow to anchor tooltip to `Quad` glyph's sides

Those are related:
- [ ] fix `range_padding` when constructing plots in JS
- [ ] `DataRange1d` should consider only visible part of `AnnularWedge`
- [ ] `AnnularWedge.bounds()` always returns `[0, 0, 0, 0]`

fixes #4311


bryevdv:
ping @sccolbert

bokehjs/src/coffee/api/typings/charts.d.ts
```diff
@@ -0,0 +1,38 @@
+declare namespace Bokeh.Charts {
+    //var Chart: { new(attributes?: IChart, options?: ModelOpts): Chart };
+    //export interface IChart extends IBasePlot {
+
+
+
+    interface IPie {
```
sccolbert: `IPieOpts` a better name for this?

bokehjs/src/coffee/api/charts.coffee
```diff
@@ -0,0 +1,138 @@
...
+
+pie = (data, opts) ->
+  labels = _.clone(data.labels)
+  values = _.clone(data.values)
+
+  i = 0
+  while i < values.length
+    if values[i] <= 0
+      labels.splice(i, 1)
+      values.splice(i, 1)
```
sccolbert: This is `O(n^2)`. It can be done in linear time with a normal for-loop which filters the values pair-wise. You can also achieve the shallow-copy at the same time, making the whole lot more efficient.

bokehjs/src/coffee/api/charts.coffee
```diff
@@ -0,0 +1,138 @@
...
+  return array.reduce(((a, b) => a + b), 0)
+
+cumsum = (array) ->
+  result = []
+  array.reduce(((a, b, i) -> result[i] = a + b), 0)
+  return result
+
+num2hexcolor = (num) -> sprintf("#%06x", num)
+
+pie = (data, opts) ->
```
sccolbert: API docs for how to use this function?

bryevdv: FYI Chris we currently have zero experience documenting JS libraries. If you have suggestions for best or common tools or formats, especially that could integrate well into our existing docs site, they would be appreciated.

sccolbert: JSDoc is the de facto standard: http://usejsdoc.org/

I'm sure we could hook into some output formatter to translate to RST/HTML/whatever for the doc site.

But I would say the first thing is JSDOC format for the API docs, just so people reading the source know how to call the function.

bryevdv:
@mattpap is there any status update to this?

mattpap:
@bryevdv, no. It would be nice to get some feedback on existing work.

bryevdv:
@mattpap sorry, I did not realize you were ready for feedback, I will take a look now.

bokeh/models/annotations.py
```diff
@@ -439,9 +439,10 @@ class Tooltip(Annotation):
     """
     level = Override(default="overlay")
 
-    side = Either(Auto, Enum(Side), default="auto", help="""
+    attachment = Enum("horizontal", "vertical", "left", "right", "above", "below", help="""
```
bryevdv: should there be something like `"auto"` as well? I'm imagining the case (to be handled later) that we might want to automatically dodge multiple tooltips

bryevdv: For now `"auto"` could just map to one of the others, but it leaves the door open for the future to do something a bit smarter.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
```diff
@@ -107,6 +107,14 @@ class AnnularWedgeView extends Glyph.View
   draw_legend: (ctx, x0, x1, y0, y1) ->
     @_generic_area_legend(ctx, x0, x1, y0, y1)
 
+  _scxy: (i) ->
+    r = (@sinner_radius[i] + @souter_radius[i])/2
+    a = (@_start_angle[i]  + @_end_angle[i])   /2
+    return {x: @sx[i] + r*Math.cos(a), y: @sy[i] + r*Math.sin(a)}
+
+  scx: (i) -> @_scxy(i).x
+  scy: (i) -> @_scxy(i).y
```
bryevdv: unfortunate duplication of computation, but unavoidable with the current API I think (my fault)

bryevdv:
@mattpap I left a few small comments. There is also the comment from chris above. Otherwise I think it is generally good, I definitely like the example code. Needs at least some minimal documentation on the arguments and options.

bokeh/models/annotations.py
mattpap: > that we might want to automatically dodge multiple tooltips

This functionality doesn't belong to an annotation, rather to a tool or perhaps a manager on top of that. Still `auto` in this case may make sense, but to automatically choose between horizontal and vertical dimensions.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
mattpap: > or caching

NO, NO, NO. No, caching. Unless this is the only way to make things work and we gave it an actual thought.

I added `get_anchor_point()` method, which should be implemented accordingly. It currently uses `scx()` and `scy()`, but that should be changed later. I didn't want to do too much unrelated rewriting in this PR.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
```diff
@@ -107,6 +107,14 @@ class AnnularWedgeView extends Glyph.View
   draw_legend: (ctx, x0, x1, y0, y1) ->
     @_generic_area_legend(ctx, x0, x1, y0, y1)
 
+  _scxy: (i) ->
+    r = (@sinner_radius[i] + @souter_radius[i])/2
+    a = (@_start_angle[i]  + @_end_angle[i])   /2
+    return {x: @sx[i] + r*Math.cos(a), y: @sy[i] + r*Math.sin(a)}
```
mattpap: @bryevdv, I presume I forgot to consider `direction` here?

bokehjs/src/coffee/models/glyphs/quad.coffee
```diff
@@ -41,6 +67,23 @@ class QuadView extends Glyph.View
     result['1d'].indices = hits
     return result
 
+  get_anchor_point: (anchor, i, spt) ->
+    left = Math.min(@sleft[i], @sright[i])
+    right = Math.max(@sright[i], @sleft[i])
+    top = Math.min(@stop[i], @sbottom[i])     # screen coordinates !!!
```
mattpap: We need proper abstractions to deal with coordinate systems (among other things). I shouldn't be concerned with whether top is the minimum or maximum of a set of values in screen coordinates.

bryevdv: That's the idea with view vs screen coords but it seems there is some leakage.

bokeh/models/annotations.py
bryevdv: I don't think the functionality belong here. But it could be appropriate for an annotation to express its preference for  "feel free to dodge me" or "don't dodge me" to the thing that goes.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
bryevdv: Right, I wasn't a suggesting caching here, just commenting in general.

bokehjs/src/coffee/models/glyphs/annular_wedge.coffee
bryevdv: Seems so

mattpap:
@bryevdv, @birdsarah, btw. JS API is completely broken after merging in `master`. The problem is that currently `this.document` is used earlier than before, causing `attach_document()` fail in `add_panel()` (called from `add_layout()`).

bokehjs/src/coffee/models/glyphs/quad.coffee
mattpap: I understand the idea, which is fine. The problem is that there isn't enough tooling around those concepts. We have methods for converting between those two types of coordinate systems, but that's not enough abstraction to not be bothered with low-level details day-to-day development.

mattpap:
@bryevdv, @birdsarah, any ideas ^^^?

birdsarah:
Sorry I missed the earlier comment, I've been in a bit of a hole. Can you point me to the attach_document and add_panel calls you're referring to. 

Random stab in the dark: but have you tried putting things that need the document in _doc_attached which gets called after the document in notify_attach?

bryevdv:
@mattpap how do you want to collaborate on this? Do you want me to make a PR off this one? Or should I push changes directly here? Some issues I see:
* `add_tools`, etc on `PlotCanvas` for now (easy quick fix can push immediately)
* update to new `title` API
* implement `GridPlot` in terms of `Row` and `Column`

bryevdv:
@mattpap I don't know about TypeScript to fix the `title` issue, but I can answer questions to help with whatever changes are needed on the TS side. I will push a commit for a fix for the first bullet.

mattpap:
@bryevdv, can I get an explanation (or some references) why there is both `Plot` and `PlotCanvas`? As I see it currently, I will have to change all APIs from `plot.something` to `plot.plot_canvas.something`, which is cumbersome. Additionally, I have to change all APIs to start with a document (to get layout happy), which brings asymmetry between Python and JS APIs. We do we have to maintain layout at model level, where it's a property of the view system (I presume).

bryevdv:
@mattpap There is an open issue to discuss, document and design `Plot` and `PlotCanvas` (or whatever it is called eventually) I think some of the methods from `PlotCanvas` can and probably should be lifted in to `Plot`. My basic conception of `Plot` is that is should be a fairly lightweight and simple container for renderers (soothing like `select` and the methods I forwarded) should actually be moved into `Plot` and `PlotCanvas` can be made to reference what it needs. Your input and help with design improvement here would be extremely welcome. (Or put another way: please don't feel that you can't change thing in this area, you certainly can). If you'd like to have a call/discussion about things, or iterate on a document we can do that too.

mattpap:
> please don't feel that you can't change thing in this area, you certainly can.

I know that. The problem is that I don't know the motivation and the goals she wanted to achieve with this change, so it's far from obvious what should be the next step. For now I will just patch things up, so that I can get to the main problem (document).

bryevdv:
@mattpap FYI I am working today on a separate PR, to merge into this, one that does the work to simplify `PlotCanvas`.

mattpap:
@bryevdv, status of examples:
- [ ] anscombe
  - [ ] no gird lines
  - [ ] messedup toolbar
- [ ] burtin
  - [ ] `invalid Expression argument: undefined`
- [ ] charts
  - [ ] first pie chart is of wrong size
  - [ ] text labels are offset (not sure if related)
- [ ] hover
  - [ ] `invalid Expression argument: undefined`
- [ ] legends
  - [ ] `invalid Expression argument: undefined`
- [ ] linked
  - [ ] messedup toolbar
- [ ] stocks
  - [ ] `invalid Expression argument: undefined`
- [ ] tap
  - [ ] `invalid Expression argument: undefined`

bryevdv:
@mattpap thanks for the update, happily looks like most caused by the same two issues.

mattpap:
> `invalid Expression argument: undefined`

I fixed this in the JS API, but this way I broke everything else. Initialization order is broken in general.

mattpap:
One simple hack later and now everything at least renders. Still the hack shouldn't be there in the first place and there are plenty of glitches yet to be fixed.

mattpap:
In general toolbar behaves badly. I'm not sure if this is intrinsic problem or just a problem with how I ported new code to the API. Toolbar either clashes with axes or is too close to the plot. On a side note, why did toolbar get new styling of tooltips?

sccolbert:
I second the question of custom tooltips. In phosphor, we've just been setting the `title` attribute on the node to get the native tooltips.

birdsarah:
> On a side note, why did toolbar get new styling of tooltips?

Trying to unify styling throughout bokeh. Tooltips hopefully now better match look-and-feel of legend and the outline of the plot. Trying to get all greys, whites, border-radius' etc consistent. It is also my intention to change the default hover from blue-yellow to a matching white background, black text, gray border. (we may want to offer some themes in case some people have got attached to the blue-yellow). I'm not saying I've accomplished these goals, but I keep chipping away at them.

bryevdv:
> Toolbar either clashes with axes or is too close to the plot. 

toolbar defaults to the right now, so the commonest case of axis to the left, title on top, should be fine. In general I much prefer this more compact presentation. If you don't want the toolbar snug to the plot (i.e. you want it outside any axes) you can set `toolbar_sticky=False`

birdsarah:
> I second the question of custom tooltips. In phosphor, we've just been setting the title attribute on the node to get the native tooltips.

@sccolbert - are you referring to @mattpap's questions about styling changes on the tooltips? Just in case: the tooltips were already styled, just different from everything else - like everything in Bokeh! I'm not opposed to setting the title attribute to get native tooltips. But I'd be more interested in finding a better solution than both of these that improves UX for non-mouse users (https://www.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/).

bryevdv:
@mattpap There is a trivial one-line conflict in the Coffee side. On the TS API side, I made these changes and everything was basically working:
```
 $ git diff
diff --git a/bokehjs/examples/anscombe/anscombe.ts b/bokehjs/examples/anscombe/anscombe.ts
index 211c974..a6253c1 100644
--- a/bokehjs/examples/anscombe/anscombe.ts
+++ b/bokehjs/examples/anscombe/anscombe.ts
@@ -40,12 +40,12 @@ namespace Anscombe {
         const plot = new Bokeh.Plot({
             x_range: xdr,
             y_range: ydr,
-            title: title,
             plot_width: 400,
             plot_height: 400,
             border_fill_color: "white",
             background_fill_color: "#e9e0db",
         });
+        plot.title.text = title
         const xaxis = new Bokeh.LinearAxis({axis_line_color: null});
         const yaxis = new Bokeh.LinearAxis({axis_line_color: null});
         plot.add_layout(xaxis, "below");
diff --git a/bokehjs/examples/burtin/burtin.ts b/bokehjs/examples/burtin/burtin.ts
index ede934a..1473c6a 100644
--- a/bokehjs/examples/burtin/burtin.ts
+++ b/bokehjs/examples/burtin/burtin.ts
@@ -80,7 +80,6 @@ namespace Burtin {
     const small_angle = big_angle / 7
 
     const p = plt.figure({
-        title: "",
         plot_width: width, plot_height: height,
         x_axis_type: null, y_axis_type: null,
         x_range: [-420, 420], y_range: [-420, 420],
diff --git a/bokehjs/examples/donut/donut.ts b/bokehjs/examples/donut/donut.ts
index 0f856b6..4e73a83 100644
--- a/bokehjs/examples/donut/donut.ts
+++ b/bokehjs/examples/donut/donut.ts
@@ -116,7 +116,7 @@ namespace WebBrowserMarketShare {
     })
 
     function render(item: MonthlyShares) {
-        fig.title = `${item.month} ${item.year}`
+        fig.title.text = `${item.month} ${item.year}`
 
         const source = new Bokeh.ColumnDataSource({
             data: {
@@ -163,7 +163,7 @@ namespace WebBrowserMarketShare {
                 const i = cds.inspected['1d'].indices[0]
                 const name = cds.data["names"][i]
                 const share = Bokeh.sprintf("%.02f%%", cds.data["shares"][i])
-                fig.title = `${name}: ${share}`
+                fig.title.text = `${name}: ${share}`
             }
             paused = !paused
         },
diff --git a/bokehjs/examples/stocks/stocks.ts b/bokehjs/examples/stocks/stocks.ts
index e372b04..21e6b59 100644
--- a/bokehjs/examples/stocks/stocks.ts
+++ b/bokehjs/examples/stocks/stocks.ts
@@ -8,11 +8,11 @@ namespace Stocks {
         // Create plot
         const plot = plt.figure({
             tools: "pan,wheel_zoom,save,reset",
-            title: title,
             plot_width: 400,
             plot_height: 400,
             background_fill_color: "#eeeeff",
         });
+        plot.title.text = title
 
         // Add axis and grid
         const xaxis = new Bokeh.DatetimeAxis({axis_line_color: null, axis_label: 'time'});
diff --git a/bokehjs/src/coffee/api/plotting.coffee b/bokehjs/src/coffee/api/plotting.coffee
index d38d0b5..4aae5d6 100644
--- a/bokehjs/src/coffee/api/plotting.coffee
+++ b/bokehjs/src/coffee/api/plotting.coffee
@@ -365,7 +365,11 @@ class Figure extends models.Plot
     @_legend.legends = legends
 
 figure = (attributes={}, options={}) ->
-  new Figure(attributes, options)
+  title_text = attributes['title'] ? null
+  delete attributes['title']
+  figure = new Figure(attributes, options)
+  figure.title.text = title_text
+  figure
 
 show = (obj, target) ->
   multiple = _.isArray(obj)
diff --git a/bokehjs/src/coffee/api/typings/models/plots.d.ts b/bokehjs/src/coffee/api/typings/models/plots.d.ts
index c2fec6d..129c49c 100644
--- a/bokehjs/src/coffee/api/typings/models/plots.d.ts
+++ b/bokehjs/src/coffee/api/typings/models/plots.d.ts
@@ -22,7 +22,7 @@ declare namespace Bokeh {
         plot_width?: Int;
         plot_height?: Int;
 
-        title?: Title | string;
+        title?: Title;
         title_location?: Location;
 
         // {{{ outline = include[LineProps]
index d7091ab..326a535 100644
@@ -1,4 +1,5 @@
 _ = require "underscore"
+$ = require "jquery"
 
 Canvas = require "../canvas/canvas"
 CartesianFrame = require "../canvas/cartesian_frame"
```
The missing `$` require will be fixed in another PR and building shows these issues:
```
[14:29:41] examples/legends/legends.ts(16,27): error TS2345: Argument of type '{ title: string; x_range: DataRange1d; y_range: DataRange1d; }' is not assignable to parameter of type 'IFigure'.
  Types of property 'title' are incompatible.
    Type 'string' is not assignable to type 'Title'.
[14:29:41] examples/legends/legends.ts(22,27): error TS2345: Argument of type '{ title: string; x_range: DataRange1d; y_range: DataRange1d; }' is not assignable to parameter of type 'IFigure'.
  Types of property 'title' are incompatible.
    Type 'string' is not assignable to type 'Title'.
[14:29:41] Finished 'examples:legends' after 2.15 s
[14:29:43] examples/tap/tap.ts(33,26): error TS2345: Argument of type '{ title: string; tools: string; }' is not assignable to parameter of type 'IFigure'.
  Types of property 'title' are incompatible.
    Type 'string' is not assignable to type 'Title'.
[14:29:43] examples/tap/tap.ts(43,21): error TS7006: Parameter 'ds' implicitly has an 'any' type.
[14:29:43] Finished 'examples:tap' after 3.53 s
[14:29:43] examples/hover/hover.ts(33,26): error TS2345: Argument of type '{ title: string; tools: string; }' is not assignable to parameter of type 'IFigure'.
  Types of property 'title' are incompatible.
    Type 'string' is not assignable to type 'Title'.
[14:29:43] examples/hover/hover.ts(42,23): error TS7006: Parameter 'source' implicitly has an 'any' type.
[14:29:43] examples/hover/hover.ts(42,31): error TS7006: Parameter 'info' implicitly has an 'any' type.
```
But as I said, everythign still looks pretty good. Do you want me to push these changes?

bryevdv:
As a reference, the discussion @birdsarah and I had was that passing `title` as a kward to `figure` or `Chart` works, and they just convert that to setting `self.title.text` so I followed a similar approach above.

bryevdv:
actually that `$` require *is* in master, did the auto-merge delete it?

mattpap:
@bryevdv, so do I understand correctly that toolbar being rendered on top of axes is by design?

birdsarah:
yes that is `toolbar_sticky=True` (default) you can set `toolbar_sticky=False` for the old behavior where it lives off the canvas.

bryevdv:
OK, the last potentially disruptive PR (apart from this one) has been merged. The only other non-examples/docs PRs I can imagine considering:
* add viridis colormap
* children update PR (based off this one)

bryevdv:
It's been a long slog, I appreciate everyone's hard work. We will change some branch and release processes to make sure we don't get in a long-release situation like this again (my fault for merging some work too early)

mattpap:
> yes that is toolbar_sticky=True (default)

To me this is broken and unacceptable.

bryevdv:
Both of these are infinitely better than the old style:
screen shot 2016-06-16 at 5 03 24 pm
screen shot 2016-06-16 at 5 03 10 pm

Making the default right is a concession to the fact that axes are *usually* not on the right, titles are *usually* on the top, and they *may* be long. It will produce the most attractive result under the widest set of normal use-cases. In cases where it doesn't it's perfectly possible to: 1) move the toolbar elsewhere, or 2) make it not sticky

mattpap:
@bryevdv, try left toolbar with left axis.

bryevdv:
>  try left toolbar with left axis.

I don't have to, I know that looks bad. Set `toolbar_sticky=False`

bryevdv:
@mattpap this seems easy enough to warn about: If `toolbar_sticky` is `True` and an axis side and toolbar location match, print a warning. Would that satisfy you for now?

mattpap:
@bryevdv, no. If those don't work together, then this situation shouldn't be allowed, either by throwing a hard error or setting sticky = False automatically when locations match.

bryevdv:
We definitely have a difference in perspective. I tilt towards giving users more capability, even if that sometimes means not every configuration is guaranteed to be useful to the same degree. Combined with a an immediate and detailed warning, I am completely satisfied. Let's open for a wider discussion, then. @bokeh/dev @birdsarah @pzwang

pzwang:
>  If those don't work together, then this situation shouldn't be allowed, either by throwing a hard error or setting sticky = False automatically when locations match.

It's not *always* that those don't work together.  Perhaps I want the axis turned on, and ticks and tick labels turned off, just so I can display an axis title?  Perhaps I want the axis ticks and no tick labels, and still have the toolbar on that side of the plot?  What if I just have a few tool icons visible, and there is enough room to display the axis title and the toolbar on the same side, and I have `tick_out=0`?  etc.

If we could unconditionally declare with certainty that a visual configuration is invalid, then @mattpap is right that it should be disallowed.  One of the challenges of a compositionally-oriented library like Bokeh is that there will always be many scenarios where the user may want to compose something unexpected, and we should bias towards allowing those things.

bryevdv:
@mattpap As @pzwang mentions, there *are* cases where they are on the same side that *do* make sense: internal ticks (only) or possibly some day internal tick values. the the overlap is no problem. So doing something unilaterally like a hard error or adjusting without asking is too limiting, for no reason. I will add the warning.

bryevdv:
@mattpap @birdsarah @pzwang https://github.com/bokeh/bokeh/pull/4551

mattpap:
@bryevdv, there are plenty of js test failures due to plot <-> plot_canvas modifications.

bryevdv:
Is there anything I can do to assist?

mattpap:
@bryevdv, perhaps you could ask @birdsarah to fix those. After all she knows those tests the best.

mattpap:
Besides tests, remaining issues are:

- [ ] no grid lines in anscombe
- [ ] colored toolbar and missing top outline in burtin
- [ ] first pie chart is of wrong size in charts
- [ ] no hover in hover

birdsarah:
happy to help on testing. may need some collaboration as i don't know the ts api at all.

for example a test failure like the missing "add_layout" function. Seems like we need to add an add_layout function somewhere.

bryevdv:
Just a heads up #4549 adds palettes so it will probably result in one more conflict to resolve. I will merge it later tonight.

bryevdv:
I resolved the conflict in `palettes.coffee`

bokehjs/src/coffee/models/tools/toolbar_base.coffee
```diff
@@ -95,7 +95,8 @@ class ToolbarBase extends LayoutDOM.Model
     # Get the constraints from widget
     constraints = super()
     # Set the fixed size of toolbar
-    constraints.push(EQ(@_sizeable, -30))
+    if @_sizeable?
```
birdsarah: When would toolbar not have a sizeable dimension?

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -239,7 +314,8 @@ class Plot extends LayoutDOM.Model
       # ALL BELOW ARE FOR PLOT CANVAS
       plot_width:        [ p.Number,   600                    ]
       plot_height:       [ p.Number,   600                    ]
-      title:             [ p.Instance, () -> new Title.Model({text: ""})]
+
+      title:             [ p.Any, () -> new Title.Model({text: ""})] # TODO: p.Either(p.Instance(Title), p.String)
```
birdsarah: This is inconsistent with the python API, can you share some context.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -620,194 +624,66 @@ class PlotCanvas extends LayoutDOM.Model
...
+    @right_panel.attach_document(@document)
 
-  @internal {
-      # This is set by parent plot for convenient access
-      toolbar: [ p.Instance ]
-  }
+    logger.debug("PlotCanvas attached to document")
 
   @override {
-    outline_line_color: '#e5e5e5'
```
birdsarah: These overrides do not seem to have made it up to plot, am I missing something?

bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
```diff
@@ -156,16 +156,13 @@ class GMapPlotCanvasView extends PlotCanvas.View
...
-    ctx.fillStyle = @mget('border_fill_color')
+    ctx.fillStyle = @model.plot.border_fill_color
     ctx.fill()
 
 class GMapPlotCanvas extends PlotCanvas.Model
   type: 'GMapPlotCanvas'
   default_view: GMapPlotCanvasView
 
-  @define {
-      map_options: [ p.Any ]
```
birdsarah: Where did these get moved to?

bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -121,7 +121,7 @@ class LayoutDOMView extends BokehView
 class LayoutDOM extends Model
   type: "LayoutDOM"
 
-  constructor: (attrs, options) ->
+  initialize: (attrs, options) ->
```
birdsarah: Interesting! I had a lot of back and forth on this. I tried to always use initialize where I could. 

But sometimes it seemed to be necessary to use constructor. I may have been wrong in this case or this was an artefact when I was juggling too many balls in the layout PR, but just wanted to check in on this.

If I'm recalling correctly, which I may not be, it was because constraints started getting added to the solver before the model was fully initialized...but I'm feeling pretty hazy. Anyway, I expect it would just blow up if this was meant to be constructor.

birdsarah:
I had a look over all the coffeescript in detail as it's stuff I know. Looks great. Thanks for the heavy lifting cleaning up the Plot/PlotCanvas API issue. Lots of great clean-up in document and hasprops. 

Had a few small points. Will work on the failing JS tests over the weekend. 

1) Do you want me to add a note or anything when I start working on them?
2) Would you like a PR against this branch or me to push to this branch?

bokehjs/src/coffee/models/plots/plot.coffee
bryevdv: I expect it's just timing, still needs to be updated

bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
bryevdv: I moved them to `GMapPlot`

birdsarah: Oh never mind. They were already there, which is why they didn't show up in the diff. Sorry for the noise.

birdsarah:
Apologies for the noise with a commit and a revert. I was working too quickly.

I know what the problem is with the linked toolbar, but don't immediately know how to fix it. The problem is that the toolbar_location property has not been correctly set on the toolbar. This is causing the css in the rendered plot to have the wrong class. Here's a before and after where I manually change it in the developer tools. 

screen shot 2016-06-18 at 4 57 58 pm
screen shot 2016-06-18 at 4 58 11 pm

You can see that layout has correctly made the space for it. The JS line where the class setting happens is: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/tools/toolbar_base.coffee#L32

birdsarah:
I believe that is now the correct fix.

birdsarah:
@mattpap, I have fixed burtin, and the gridplot toolbar issue affecting anscombe and linked (this was not due to toolbar_sticky but a missing initialization step in gridplot code). 

I am seeing gridlines on anscombe, so just wanted to touch base on that. They were made fainter to match the gray used throughout bokeh. This now means in the defaults with the salmon background color of anscombe they are very hard to pick out. I have changed the background color on anscombe so hopefully you can see them. I'm not proposing we keep that change, just wanted to highlight that there isn't a grid line problem.

@bryevdv The fix on burtin highlights a bug in layout. I think we should open a separate issue for it.

birdsarah:
screen shot 2016-06-18 at 5 21 45 pm

birdsarah:
also just to note that I'm definitely not happy with the gridplot toolbar positioning - it needs better alignment - the open issue for that is here: https://github.com/bokeh/bokeh/issues/4398

birdsarah:
@bryevdv worth noting that the tweaks that have been made to the examples here (specifically burtin and anscombe) are tweaks that we should make to main python examples. They were not made during layout PR so that we could see only the api breaks from the layout PR. But those examples will look better for a small refresh due to new defaults.

birdsarah:
@mattpap, in your note on outstanding issues you commented that "first pie chart is of wrong size in charts". 

I had a look at this and have uncovered a layout bug in master as a result. However, advance warning, you might not be happy about the intended behavior. The expected behavior is that the left axis of each plot should now align - that was the key purpose of the layout work. The first pie has correctly shrunk, but the others have not.

bryevdv:
Yes if the issue is independent please fix in a separate PR

birdsarah:
The charts problem is indeed indicative of a bug in layout. I have fixed it in commit 16c2746 on this PR using the same stop-gap measure, that I have on python side.

screen shot 2016-06-19 at 4 33 29 am

This fixes it for the defaults, but not for all layouts. The issue tracking this is here: https://github.com/bokeh/bokeh/issues/4579

bryevdv:
OK, @birdsarah @mattpap tests are green. Shall we merge this to get it out of the way?

bryevdv:
FYI the charts line hover (python) example also has no hover tooltips on this branch

bryevdv:
There are other small issues:

* hover tooltip on pie charts flickers annoyingly 
* burtin toolbar obscures plot outline
* anscombe woudl be better w/o toolbar entirely (no tools in it)

The hover tool is the only serious issue and I would recommend looking at it separately

bryevdv:
The problem with hover is that the tooltip `clear` is getting called between the `add` and the drawing

birdsarah:
> burtin toolbar obscures plot outline

The easy fix for this will be to set `toolbar_sticky=False` which will push it off canvas slightly and complete the border. I think that will look just fine.

birdsarah:
> hover tooltip on pie charts flickers annoyingly 

The easy fix for this, I believe, will be to use the mouse follow mode - whatever that's called

birdsarah:
> FYI the charts line hover (python) example also has no hover tooltips on this branch

That's wierd. I remember line hit testing got broken on master, but I thought it was fixed, we should check we haven't lost something in merge.

birdsarah:
> OK, @birdsarah @mattpap tests are green. Shall we merge this to get it out of the way? 

Fine with me, but this is really @mattpap's beast. The things that appear to be left to fix are minor tweaks that can def be done in follow-on PRs. Especially as they're mostly not about changing the API, just tweaking the examples.

I have at this point looked carefully at all the JS stuff, but haven't looked at the typescript at all really.

bryevdv:
@mattpap @birdsarah I have fixed the hover tool for all examples, JS and python. I am still not sure what is the final root cause. The update to `.data` in `add` was not triggering the backbone event (perhaps object identity, and a copy is needed?) I added manual trigger, and it works great.

bryevdv:
All the issues above are addressed, tests passing. @mattpap I am merging this now, let's put any cleanup in follow on PRs.

Issue #4313, opened 2016-05-11
Save button broken on master


birdsarah:
I get a console message:

```
TypeError: this.plot_view.canvas_view.canvas is undefined
```

brittainhard:
@birdsarah Potential fix by changing [this line]
(https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/tools/actions/save_tool.coffee#L8) to

```
canvas = @plot_view.canvas_view.ctx.canvas
```

This grabs the actual DOM canvas element, and lets you call functions like toDataUrl.

PR #4317, opened 2016-05-13
Save Tool Fix

brittainhard:
Fixes #[4313](https://github.com/bokeh/bokeh/issues/4313) @birdsarah let me know what you think.


birdsarah:
Lets add a test so this doesn't happen again.

brittainhard:
@birdsarah would this be a selenium test?

birdsarah:
no, a js unit test should do just fine. You want to test for the property that save tool needs in order to work properly. the property that was removed by someone and no-one noticed because there wasn't a test.

from what i understand it's very hard to do a selenium test for a file download as you move out of the browser and into the desktop.

brittainhard:
Awesome. I don't think there are any tests for tools, but i'll have one up soon.

brittainhard:
Awesome. I'll have this soon.

birdsarah:
Hi @brittainhard, I took a look and here's a few thoughts.

I was expecting a test that would test the regression and then pass when you made the fix. Having said that, that would be a fairly silly test. 

I think what the point is, is that that the save tool needs a method on plot_view that returns the actual canvas element. So suggest a small refactor in the spirit of continuous improvement:  a method on plot_view that returns the canvas element (think carefully about naming that as `canvas` is generally the canvas model) - that can have a test in the plot tests.

The tests for save tool would then be a little more tricky, you'd need to do some mocking to test that the result from the new method that you created is correctly used by the save tool - I would say you can skip this for this PR. Although if you want to play with mocking and sinon I would say that that's something we need to get more experience of as a team. A couple of my recent PRs (still open) have a bunch of tests that use sinon, but i still have a lot to learn. 

It would be fine to miss off that second part for this PR.

brittainhard:
@birdsarah do you mean you want a second test that would expect `@plot_view.canvas_view.canvas` to return `undefined`? I can add that easily.

I should say that it seems to me that any test which does not directly test the tools themselves will not effectively solve this regression problem. 

If we were to add a method to @plot_view to grab the canvas, regardless of whether we refactored the code, then wouldn't the regression rest be the same? That is, we would be testing for that particular method on plot_view which the tools happen to use. If you would also like me to refactor the tools to use this method directly, and then test this method, I can certainly do that.

I would also be willing to dive in to get some real testing done on the tools which, as far as I know, have no unit tests at all.

brittainhard:
@birdsarah does this look good enough for a merge?

bryevdv:
LGTM ping @birdsarah

bokehjs/src/coffee/models/tools/actions/save_tool.coffee
```diff
@@ -5,7 +5,7 @@ ActionTool = require "./action_tool"
 class SaveToolView extends ActionTool.View
 
   do: () ->
-    canvas = @plot_view.canvas_view.canvas[0]
+    canvas = @plot_view.canvas_view.ctx.canvas
```
birdsarah: You should use this new method you made!!

bokehjs/test/models/plots/plot.coffee
```diff
@@ -190,6 +190,35 @@ describe "Plot.View render", ->
...
+    })
+    @test_plot.document = @test_doc
+    @test_plot._doc_attached()
+    @test_plot_view = new @test_plot.default_view({ 'model': @test_plot })
+
+  it "should exist because get_canvas_element depends on it", ->
+    expect(@test_plot_view.canvas_view.ctx).to.exist
+
+  it "should exist to grab the canvas DOM element using canvas_view.ctx", ->
+    expect(@test_plot_view.get_canvas_element).to.exist
```
birdsarah: You haven't actually tested the method you created.

You should setup a canvas, attach it to the plot view and then assert that the return value of `get_canvas_element()` is the canvas_view.ctx.canvas.

bokehjs/test/models/plots/plot.coffee
```diff
@@ -190,6 +190,35 @@ describe "Plot.View render", ->
...
+    @test_plot = new Plot({
+      x_range: new Range1d({start: 0, end: 1})
+      y_range: new Range1d({start: 0, end: 1})
+    })
+    @test_plot.document = @test_doc
+    @test_plot._doc_attached()
+    @test_plot_view = new @test_plot.default_view({ 'model': @test_plot })
+
+  it "should exist because get_canvas_element depends on it", ->
+    expect(@test_plot_view.canvas_view.ctx).to.exist
```
birdsarah: Does exist actually test for anything - I can't make it tell me anything useful. If I delete something that I was expecting to exist then it still passes. I'm not sure, I just want you to be sure that you've seen these tests failing and passing in a predictable way.

birdsarah:
This isn't ready. See comments above.

brittainhard:
@birdsarah @bryevdv I have no idea how to set up a DOM element canvas in a mocha test. The `get_canvas_element` needs access to an actual HTML canvas element, not a JavaScript object. If there is a way to set up an actual mock DOM canvas in these tests, please let me know so I can finish this.

bryevdv:
@brittainhard I'm guessing it needs to be a selenium test (@birdsarah correct me if I am mistaken) IN which I don't have much experience yet. @canavandl can you offer some assistance here, @birdsarah is swamped this week.

birdsarah:
> @birdsarah @bryevdv I have no idea how to set up a DOM element canvas in a mocha test. The get_canvas_element needs access to an actual HTML canvas element, not a JavaScript object. If there is a way to set up an actual mock DOM canvas in these tests, please let me know so I can finish this.

FWIW, the way to do this would have been to extend the canvas stub that you've already been using - by calling utils.stub_canvas(). If you're not familiar with sinon though that's a fair amount of learning curve - it took me a while to get those stubs working. We can work on that later.

birdsarah:
LGTM.

bryevdv:
I'll squash-merge this when travis is green

Issue #2240, opened 2015-05-02
Tick format language


Jdash99:
Charts does not show up when changing the tick format language to other than english ("en").

For example, inside ipython-notebook:

``` python
    import numpy as np
    import pandas as pd
    from bokeh.models import NumeralTickFormatter
    from bokeh.plotting import figure, show

    bokeh.plotting.output_notebook()

    df = pd.DataFrame(np.random.randint(0, 900000000, (10,1)), columns=['my_int'])
    p = figure(plot_width=700, plot_height=280, y_range=[0,1000000000])

    for index, record in df.iterrows():
    p.rect([index], [record['my_int']/2], 0.8, [record['my_int']], 
               fill_color="red", line_color="black")

    p.yaxis.formatter=NumeralTickFormatter(language="es")
    #p.yaxis.formatter=NumeralTickFormatter(language="en") -->Works fine
    show(p)
```

Kernel info:

``` python
    Python 3.4.3 |Anaconda 2.2.0 (32-bit)| (default, Mar  6 2015, 12:08:17) [MSC v.1600 32 bit (Intel)]
    Type "copyright", "credits" or "license" for more information.

    IPython 3.0.0 -- An enhanced Interactive Python
```

bryevdv:
@mattpap do you have any quick comments on this, do you remember if there were any issues already known?

colinfang:
`language='en-gb'` doesn't work either.

Btw, I think it is better if we add a sentence saying the format follows `numeral.js` and place a link to http://numeraljs.com/ in the [reference guide]( http://bokeh.pydata.org/en/latest/docs/reference/models.html#bokeh.models.formatters.NumeralTickFormatter)

bryevdv:
FWIW with `0.12.6` this code:
```
import numpy as np
import pandas as pd
from bokeh.models import NumeralTickFormatter
from bokeh.plotting import figure
from bokeh.io import output_file, show


df = pd.DataFrame(np.random.randint(0, 900000000, (10,1)), columns=['my_int'])
p = figure(plot_width=700, plot_height=280, y_range=[0,1000000000])

for index, record in df.iterrows():
    p.rect([index], [record['my_int']/2], 0.8, [record['my_int']],
           fill_color="red", line_color="black")

p.yaxis.formatter=NumeralTickFormatter(language="es")

output_file("foo.html")
show(p)
```
does produce what appears to be a correct plot. Tho I don't know if there should be any changes to the ticks (they look the same to me). 

[...]

Issue #2204, opened 2015-04-22
Look into use of DataViews and ArrayBuffers for more efficient data send/recv


pzwang:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView

almarklein:
Here's some "binary json" variations:

* msgpack - http://msgpack.org/
* bson - http://bsonspec.org/
* ubson - http://ubjson.org/

I'd be happy to look into this once the WebGL implementation is up.

almarklein:
cc @bryevdv we already had an issue for this. Will start looking into this next week.

almarklein:
I looked into compression. Not necessary per see, but it would be nice to reduce the size of transfers and exported documents. Unfortunately, you'd need a 3d party library, and these are all pretty big. Examples:

* [Pako](https://github.com/nodeca/pako) is a zlib implementation in JS which is reportedly very fast. Comes in at a 5.5k sloc count though.
* [js-deflate](https://github.com/dankogai/js-deflate) provides delate/inflate (i.e. zlib) compression at about 2k sloc.
* [lz-string](http://pieroxy.net/blog/pages/lz-string/index.html) is a string compression lib (no typed arrays iiuc) provides LZW-ish compression at 485 sloc. Not an "official" format though.
[lz4.js](https://github.com/pierrec/node-lz4) provides lz4 compression at 6k sloc.
* [lzma](https://github.com/nmrugg/LZMA-JS) is an LZMA (de)compressor at 2.6k sloc.
* [here](https://github.com/antimatter15/bzip2.js) is a bz2 decompressor in 245 sloc. 
* [minilzo](https://github.com/abraidwood/minilzo-js) implements LZO decompression in 560 sloc, based on minilzo.c which is written by the O in LZO.

almarklein:
Ok, I think I've done enough googling and reading through format specs for now. I need some help moving further. In my first post of this issue I put together an overview and I propose three options. cc @bryevdv @pzwang

bryevdv:
[...]

PR #4322, opened 2016-05-14
Using ArrayBuffers instead of arrays

almarklein:
Issues: related to #2204 and #2590

Currently, we serialize arrays to a list of scalars on the Python side, which is then interpreted as a JS Array. When the example contains larger datasets, this bloats the generated HTML document, but JS Arrays are also know to be inefficient (memory wise and speed wise). 

This PR is a start at a possible approach to improve this, to get some discussion going on this topic. This only covers the bit for exporting a plot to an HTML document (i.e. does not include server/websocket stuff yet), and how to unpack it at the JS side. Also, the rest of BokehJS would need some changed to correctly deal with the data being an ArrayBuffer rather than an Array.

For reference, the scatter10k example exports to 222kb instead of 1340kb (6x smaller) with this change (though the difference will be much smaller for examples with little data).

(Higher level discussions should probably occur in #2240 or in a new issue.)

### DIFF
--- a/bokeh/util/serialization.py
+++ b/bokeh/util/serialization.py
@@ -4,6 +4,7 @@ Bokeh objects.
 """
 from __future__ import absolute_import
 
+from base64 import encodebytes
 from six import iterkeys
 
 from .dependencies import import_optional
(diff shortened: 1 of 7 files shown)


bryevdv:
@almarklein I'd like to close this PR for now, and take it up again later.

almarklein:
Sure. This needs (much) more discussion.

PR #4323, opened 2016-05-15
Move Plot to PlotCanvas

birdsarah:
Part of #4164

Plot will become a Box that will hold a Toolbar (widget) and a PlotCanvas.
This commit just empties plot.coffee into plot_canvas.coffee

It should have no effect on Bokeh, but is important to get in early otherwise
I'm going to have unmanageable merge conflicts.


birdsarah:
@bokeh/dev - this is v. small but v. important if anyone's able to check it out and merge.

bryevdv:
LGTM

PR #4326, opened 2016-05-15
Fix handling of host wildcard: --host '*'

### LINKED ISSUE #4325: Bug --host '*' wildcard result in HTTP "403: Forbidden" error
According to the [documentation](http://bokeh.pydata.org/en/latest/docs/user_guide/cli.html#network-configuration) using `bokeh serve --host '*'` should accept connections on any IP address and port (possibly in combination with the `--address 0.0.0.0` flag).

However when I try for instance by adding an alternative IP address to the local host network interface:

```
sudo ip address add 192.168.99.37/24 dev wlan0
```

then launching `bokeh` with:

```
bokeh serve --address 0.0.0.0 --port 5006 --host '*'
```

and trying to access http://192.168.99.37:5006/ result in the following HTTP 403 page:

```
403: Forbidden
```

This bug is annoying when deploying services with docker orchestration tools such as docker-compose or kubernetes it is not always possible to know the public ip address that will be used to access the service ahead of time and configuring a public DNS is too cumbersome when in development mode.

ogrisel:
Fix #4325. Apparently the `match_host` function in `bokeh/server/tornado.py` was not designed to handle a port number while the `_create_hosts_whitelist` function in `bokeh/server/server.py` always append a port number (`:80`) if none is provided in the CLI therefore setting the whitelist to `*:80` which would previously not match anything.

This PR changes two things:

- Setting `--host '*'` will result in accepting requests on any port (instead of just ":80")
- It's now also possible to use the wild card with a specific port, e.g.: `--host *:8080` or with any port with a specific hostname or ip `--host hostname:*`  explicitly.

The behavior that `:80` is appended to non-wildcard host names is preserved to follow what was previously documented.

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -102,10 +102,24 @@ If no host values are specified, then by default the Bokeh server will
 accept requests from ``localhost:<port>`` where ``<port>`` is the port
 that the server is configured to listen on (by default: {DEFAULT_PORT}).
 
-If an asterix ``*`` is used in the host value (for example ``--host *``) then
-it will be treated as a wildcard.  As a warning, using permissive host values
-like ``*`` may be insecure and open your application to HTTP host header
-attacks.
+If an asterix ``*`` is used in the host value then it will be treated as a
+wildcard:
+
+.. code-block:: sh
+
+    bokeh serve app_script.py --address=0.0.0.0 --host='*'
+
+Using the wildcard can be helpful when testing applications that are deployed
+with cloud orchestration tools and when the public endpoint is not known ahead
+of time: for instance if the public IP is dynamically allocated during the
+deployment process and no public DNS has been configured for the testing
+environment.
+
+As a warning, using permissive host values like ``*`` may be insecure and open
+your application to HTTP host header attacks. Production deployments should
+always set the ``--host`` flag to use the DNS name of the public endpoint such
+as a TLS-enabled load balancer or reverse proxy that serves the application to
+the end users.
 
 Also note that the host whitelist applies to all request handlers,
 including any extra ones added to extend the Bokeh server.
--- a/bokeh/server/tornado.py
+++ b/bokeh/server/tornado.py
@@ -31,13 +31,19 @@ from .views.static_handler import StaticHandler
 def match_host(host, pattern):
     """ Match host against pattern
 
-    >>> match_host('192.168.0.1', '192.168.0.1')
+    >>> match_host('192.168.0.1:80', '192.168.0.1:80')
     True
+    >>> match_host('192.168.0.1:80', '192.168.0.1')
+    True
+    >>> match_host('192.168.0.1:80', '192.168.0.1:8080')
+    False
     >>> match_host('192.168.0.1', '192.168.0.2')
     False
     >>> match_host('192.168.0.1', '192.168.*.*')
     True
     >>> match_host('alice', 'alice')
+    True
+    >>> match_host('alice:80', 'alice')
     True
     >>> match_host('alice', 'bob')
     False
@@ -45,7 +51,31 @@ def match_host(host, pattern):
     False
     >>> match_host('alice', '*')
     True
+    >>> match_host('alice', '*:*')
+    True
+    >>> match_host('alice:80', '*')
+    True
+    >>> match_host('alice:80', '*:80')
+    True
+    >>> match_host('alice:8080', '*:80')
+    False
+
     """
+    if ':' in host:
+        host, host_port = host.rsplit(':', 1)
+    else:
+        host_port = None
+
+    if ':' in pattern:
+        pattern, pattern_port = pattern.rsplit(':', 1)
+        if pattern_port == '*':
+            pattern_port = None
+    else:
+        pattern_port = None
+
+    if pattern_port is not None and host_port != pattern_port:
+        return False
+
     host = host.split('.')
     pattern = pattern.split('.')
 
--- a/bokeh/server/server.py
+++ b/bokeh/server/server.py
@@ -24,6 +24,15 @@ def _create_hosts_whitelist(host_list, port):
 
     hosts = []
     for host in host_list:
+        if '*' in host:
+            log.warning('Host wildcard %r can expose the application to HTTP '
+                        'host header attacks. Host wildcard should only be '
+                        'used for testing purpose.', host)
+        if host == '*':
+            # do not append the :80 port suffix in that case: any port is
+            # accepted
+            hosts.append(host)
+            continue
         parts = host.split(':')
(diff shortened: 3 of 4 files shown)


bokeh/command/subcommands/serve.py
```diff
@@ -102,7 +102,7 @@
 accept requests from ``localhost:<port>`` where ``<port>`` is the port
 that the server is configured to listen on (by default: {DEFAULT_PORT}).
 
-If an asterix ``*`` is used in the host value (for example ``--host *``) then
+If an asterix ``*`` is used in the host value (for example ``--host '*'``) then
```
ogrisel: It's necessary to quote `*` in shell commands otherwise file globing will kick-in.

bokeh/server/tornado.py
```diff
@@ -70,7 +100,8 @@ def check_whitelist(request_host, whitelist):
 
     if request_host in whitelist:
         return True
-
+    print([(request_host, host, match_host(request_host, host))
+           for host in whitelist])
```
ogrisel: Left-over debug print... Will amend my commit.

bryevdv:
The port that the bokeh server should listen on is given by `--port`. The `--host` merely says what html HOST values (which have no port associated) should be accepted. I don't think `--host` should accept a port (because it has no effect and that is confusing/misleading). I think it would be better to raise an informative runtime error stating that no port is not applicable to the host whitelist.

ogrisel:
@bryevdv are you sure? This seems to contradict what is written in the [documentation](http://bokeh.pydata.org/en/latest/docs/user_guide/cli.html#network-configuration) for the `--host` flag. I am ok to do the change but instead of raising a runtime error, I think we should raise a deprecation warning to avoid breaking deployments by users that would have followed the documentation.

bryevdv:
Well I think the wording in the documentation is very confusing but also let me look into things more carefully to make sure I am not mispeaking above. I'm out of the country the next week so it may not be until the following week.

ogrisel:
FYI the `_create_hosts_whitelist` function accept take as an agument the port passed with the `--port` CLI flag and is then used to generate the `tornado_kwargs['hosts']` and then `tornado_kwargs['extra_websocket_origins']` flags. So changing this can have quite an impact. Ok for waiting. I can work with the current state of my branch in the mean time.

ogrisel:
BTW @bryevdv if you have the permissions, would you mind starting the following travis jobs:

https://travis-ci.org/bokeh/bokeh/jobs/130361694

I thing this timeout failure is unrelated to the change in this PR.

bryevdv:
OK so the HTML request host can have a port,  I was confusing myself. That said, I am trying to think of the use cases where a wildcard port would be useful. I can see these cases:

* behind a proxy --- in this case you'd configure `--host` to be the same a the public face proxy host and port (no wildcard port needed)

* users connecting directly to bokeh server -- in the case the only post it is *possible* to connect on is the one given by `--port` (no wildcard port needed)

The original motivation to add wildcard host was for an API that did automated deployments across clusters, the wildcard saved the users some work of having to coordinate a possibly unknown or hard to obtain head node hostname. 

Of course, this may merely be a failure of my imagination. Can you describe your use-case in more detail so we can try to understand when a wildcard port is needed?

ogrisel:
> The original motivation to add wildcard host was for an API that did automated deployments across clusters, the wildcard saved the users some work of having to coordinate a possibly unknown or hard to obtain head node hostname.

This is my use case: I use docker to deploy a bokeh app (part of @mrocklin's dask/distributed to monitor the state of the compute cluster). [docker networking](https://docs.docker.com/v1.8/articles/networking/) use bridge network interface and the contained app is not aware of the externally visible IP address that will forward HTTP traffic to that interface.

I don't need a wild card port explicitly as long as `--host '*'` is not translated to `--host '*:80'` as it used to be the case in bokeh master but it was natural implement support for 'host:*' in `match_host`. I am ok with simplifying the code of `match_host` to ignore the port information. Could you please give me the relevant docstring's doctest you want me to support (so as to avoid any confusion).

bokeh/server/tornado.py
```diff
@@ -31,21 +31,51 @@
...
     """
+    if ':' in host:
+        host, host_port = host.rsplit(':', 1)
+    else:
+        host_port = None
+
+    if ':' in pattern:
+        pattern, pattern_port = pattern.rsplit(':', 1)
+        if pattern_port == '*':
+            pattern_port = None
```
ogrisel: @bryevdv I think I could just remove the 2 above lines to match your expectations, right?

bryevdv:
@ogrisel I'm not necessarily opposed to adding wildcard ports, I just wanted to make sure I understood the use case, because I think we we need to expand and clarify the docs with some "scenarios" that make it more obvious why or when you'd use a given option in a given way. 

We are not actually running doctest for anything at the moment. Most of Bokeh is either declarative models, or command line interfaces, neither of which is really helped with doctests.  It does look like `match_host` would benefit from them, though.

bryevdv:
Would you be interested in adding a small section to the`server.rst` docs that talk about these kind of automated deployments and why wildcard hosts are useful in these scenarios? As well as possible an admonishment that they should probably not be used in other scenarios?

bryevdv:
restarted CI again, travis is being especially fidgety today

bryevdv:
Another thing that might be good is an `INFO` or `WARN` log if a wildcard `--host` of any sort is given, right now only see information if you crank the level to debug.

bryevdv:
also, thanks for the PR! :)

ogrisel:
@bryevdv I amended the commit to address your comments (log a warning and expand the documentation a little).

Also for your information, the doctest in tornado.py pass on my box. It can be run with:

```sh
py.test --doctest-modules bokeh/server/tornado.py
```

bryevdv:
@ogrisel Looks good and thanks! If I can ask for one more small change, since we have not integrated doctests into our CI machinery (and I don't think this PR is the best place to do that), can you add a few additional lines of unit tests here:

https://github.com/ogrisel/bokeh/blob/088fe1a61a13bb16fb4fa97a07151e0c6aaa47a1/bokeh/server/tests/test_tornado.py#L43

Then I will plan to merge this when I return from travel. Thanks!

ogrisel:
@bryevdv done!

ogrisel:
The travis failure is again a random timeout when running the examples and does not seem related to this PR.

birdsarah:
Restarted. Sorry for that, we have a number of testing kinks.

bryevdv:
Tests, code, and docs look good, merging.  Thanks @ogrisel !

ogrisel:
:beers:

bryevdv:
> 🍻 

That's a great idea, next time we are at the same conference, the first round's on me :)

PR #4332, opened 2016-05-17
Fix step plot fail on Windows with datetime data

### LINKED ISSUE #3616: Charts/Timeseries (STEP) chart is broken on Windows
Renders non sense lines

almarklein:
Fixes #3616


bryevdv:
Is there any sort of test we could add?

almarklein:
@bryevdv done

bryevdv:
awesome will merge when green

PR #4334, opened 2016-05-17
Deprecate old layouts

### LINKED ISSUE #3650: Discussion: Moving to document-unaware layouts (HBox, VBox, Tabs)
Issue:

"Dumb" layout widgets (HBox, VBox, Tabs) don't work with the `bokeh serve` command because they don't get added to Document. (note: by "dumb" I mean the layouts that are unaware of the document state)

```python
from bokeh.plotting import figure, output_file, show, curdoc
from bokeh.io import hplot
from bokeh.models.widgets import HBox, Panel, Tabs

# output to static HTML file
output_file("layout.html")

p1 = figure(plot_width=400, plot_height=400)
p1.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], size=20, color="navy", alpha=0.5)

# add a patch renderer with an alpha an line width
p2 = figure(plot_width=400, plot_height=400)
p2.patch([1, 2, 3, 4, 5], [6, 7, 8, 7, 3], alpha=0.5, line_width=2)

HBox(p1, p2)
### results in implicit vplot(p1, p2) as `p1` and `p2` are added sequentially to curdoc, not the HBox
```

In order to get the "dumb" models to work, it's necessary to:
* Remove the children roots (here `p1` and `p2` from the document
* Add the layout (containing the children) to the document
(this is what the convenience hplot/vplot classes do)

Confusingly, using "dumb" layouts as a script:

```python
show(Tabs(tabs=[Panel(child=p1), Panel(child=p2)]) ### Works fine

show(HBox(p1, p2) ### raises validation error
ERROR:/Users/lcanavan/dev/bokeh/bokeh/core/validation/check.py:W-1004 (BOTH_CHILD_AND_ROOT): Models should not be a document root if they are in a layout box: Figure, ViewModel:Plot, ref _id: 1009, Figure, ViewModel:Plot, ref _id: 1033
```

This is an inconsistent interface as both "dumb" models should have the same behaviour.

The convenience layouts hplot and vplot work as expected for scripts and bokeh serve, but their naming isn't very good (they can hold more than just plots). There is no similar 'tab' convenience layout.

Proposal -

* Deprecate hplot and vplot
* the `show` command should accept 'dumb' layouts and modify the document accordingly (remove the child roots from the document and add the layout)
* the ``bokeh serve`` interface should include explicitly adding 'dumb' layouts as roots and modify the document accordingly 
* However, if there are no roots added to the document, bokeh could make a "best guess" of the components to include (and issue a warn)

Semi-related: this would solve the error raised when
```python
layout = hplot(p1, p2)
show(p1)
```
(Because the `hplot` command adds both plots to the curdoc)

Unresolved -
* What to do with Figure vs figure?

birdsarah:
This is part of the #4164 series of chained PRs, it is opened against a branch so that the changes just to deprecate the old layouts can be reviewed.

DO NOT MERGE THIS!

The main PR is: #4343


birdsarah:
ping @bryevdv - this is ready for review (but don't merge as its part of the chained prs)

This PR does two main things:
1) It deprecates HBox, VBox, VForm replacing them with Row & Column
2) It provides a mechanism for objects to be added to documents on show so that we don't need to maintain row & Row.

This is a breaking change, users who have apps that don't add anything to document or call show will be affected. However, that was only a relatively minor set of our examples - you can see the difference.

We can revert the breaking change, but that will mean still dealing with figure/Figure etc for a while. Or we can explain the upgrade - which is not that bad.

bokeh/io.py
```diff
@@ -612,32 +615,35 @@ def gridplot(plot_arrangement, **kwargs):
...
-    layout = HBox(children=list(children), **kwargs)
-    curdoc().add_root(layout)
-    _push_or_save(layout)
+    warnings.warn(
+        """
+        The new Row is responsive by default, it resizes based on the space available. If you would
+        like to keep using a fixed size row like hplot you can set responsive=False
+        on Row. This has been automatically set on hplot.
+        """)
+    layout = Row(children=list(children), responsive=False, **kwargs)
```
bryevdv: Message is out of date? `responsive` IS set to `False` by default

bryevdv: All of the messages below look to be inconsistent in this way, too.

examples/plotting/server/line_animate.py
```diff
@@ -24,6 +24,6 @@ def update(step):
 
 curdoc().add_periodic_callback(update, 50)
 
-session.show() # open the document in a browser
+session.show(p) # open the document in a browser
```
bryevdv: I'm OK with requiring an argument to `show` I'm guessing it makes things much cleaner and simpler. But in that case lets print an informative message/exception if it is called with no arguments, that instructs about what is needed to be done now. 

Alternatively, we still don't have to keep `figure`/`Figure` to do something with "no args", we probably could just default to showing `curdoc().roots` (and still warning/deprecating)

bryevdv:
@birdsarah I left comments, otherwise LGTM

bryevdv:
@birdsarah I left comments, otherwise LGTM

However, just FYI attention will be needed eventually on the notebook:

screen shot 2016-05-19 at 1 43 48 pm

bryevdv:
I also get this if I show a plot a second time:

screen shot 2016-05-19 at 1 45 20 pm

bryevdv:
(If you'd like me to collect this info in a different place, let me know)

bokeh/io.py
birdsarah: I guess I'm not communicating very well!!!

Row is responsive True. To reduce surprise layouts I set `responsive=False` when creating a Row for someone using hplot, etc. What I was trying to say is:
1) hplot is deprecated
2) Row has different defaults
3) if you want to use Row but have it behave like hplot then set `responsive=False`
4) this has been done for you in hplot

examples/plotting/server/line_animate.py
birdsarah: To be clear the message would only be necessary if you called it with no args and curdoc().roots is empty. But a message in that case sounds great.

birdsarah:
Thanks Bryan - yes - I've been working on notebook support, I think I may need a consult from @damianavila - but there's a chance that once I solve one problem, it will resolve itself. Can I suggest that we collect misc info on #4164 for the time being as its the one consistent place for this issue.

bryevdv:
@birdsarah this can be closed?

birdsarah:
superseded: #4343

PR #4331, opened 2016-05-17
Refactor WebGL code and provide as bokeh-gl.js

almarklein:
Fixes #4321

This basically splits the `bokehgl.coffee` code into a few smaller files. No actual changes to that code, other than the gulp files.

Issue #4321, opened 2016-05-14
Put WebGL functionality in separate bokeh-gl.js?


almarklein:
Currently, we have a `bokehgl.coffee` which kept growing as features were added. With over 1400 lines, its time to split it up into a few separate files. Further we could build a separate `bokeh-gl.js` (similar to `bokeh-widgets.js` to reduce the size of the main `bokeh.js`, though this needs some discussion to establish when WebGL support should be included.

Tasks:
- [x] Refactor into smaller chunks (#4341)
- [ ] Decide on building a separate `bokeh-gl.js`
- [ ] Build a separate `bokeh-gl.js` (#4331)?

almarklein:
> There is a file size cost to splitting things up. Could we make a completely  separate and optional .js to reduce normal loads?

@bryevdv would that extra cost be due to the overhead for a "module"? What's that cost, proximately?

What do you mean with separate, more like a third party lib? If that means I can split that code up without affecting the total file size, that might be good.

How would optional actually work? At what point is decided a user/dev that webgl functionality is included in a plot?

bryevdv:
I mean like we have already split bokeh-widgets.js off we could split off bokeh-gl.js potentially too. It not great that everyone who is never using webgl has to pay for 1400 lines of code all the time. 

Bryan

mattpap:
By splitting off webgl stuff we should get at least 60 kB reduction in size of `bokeh.min.js`.

PR #4341, opened 2016-05-19
Refactor WebGL into smaller chunks

almarklein:
Implements first part of #4321. This is essentially #4331 without the part that builds a separate `bokeh-gl.js`.

This does not change any code, but only refactors it into multiple files. I know that each file introduces some overhead, but the single `bokehgl.coffee` has become rather huge, making it difficult to maintain.


almarklein:
@bryevdv I made this into a separate PR so it can be merged quickly; the refactoring is holding back progress on other WebGL work (like blurriness/pixel_ratio stuff). We can discuss `bokeh-gl.js` in #4321.

bryevdv:
LGTM, please merge when green

bokehjs/src/coffee/models/glyphs/webgl/markers.coffee
```diff
@@ -0,0 +1,524 @@
...
+    {
+        vec2 P = gl_PointCoord.xy - vec2(0.5, 0.5);
+        P = vec2(v_rotation.x*P.x - v_rotation.y*P.y,
+                 v_rotation.y*P.x + v_rotation.x*P.y);
+        float point_size = SQRT_2*v_size  + 2.0 * (v_linewidth + 1.5*u_antialias);
+        float distance = marker(P*point_size, v_size);
+        gl_FragColor = outline(distance, v_linewidth, u_antialias, v_fg_color, v_bg_color);
+        //gl_FragColor.rgb *= gl_FragColor.a;  // pre-multiply alpha
+    }
+    """
```
mattpap: @almarklein, perhaps the next step should be to move those code fragments to individual files, so that proper syntax highlighting and language specific tools could be used.

almarklein: That sounds like a plan. How easy/difficult is it to make gulp include these files into the `bokeh.js` / `bokeh-gl.js` ?

mattpap: It should be fairly easy. There are gulp extensions for templating source files. The only pain point is mocha, which bypasses the build and compiles sources by itself. We have to change that, because I already had issues with getting typescript to work in bokehjs tests (it doesn't currently).

PR #4343, opened 2016-05-20
Improve Bokeh Layout

birdsarah:
Fixes: #4164 (and a list of other things which I need to document here).

---

##### Tasks / bugs (in priority order)
- [x] Decide what to do with breaking API on VBoxForm (previously you could put plots in it, but you can't put plots in WidgetBox) - we could just do some work in the deprecation method so that it doesn't break people's work, but we don't compromise WidgetBox.
- [x] Titles (almost done - just a little deprecation)
- [x] Crosshair tool with server
- [x] Tabs - work as standalone, but can't be put in a layout (see `examples/models/glyphs`)
- [x] Add layout function

##### Could do after first
- [ ] ~~Helptool is not deduped on ToolbarBox~~ #4385
- [ ] ~~Make the resize event customizable - may not want it to be on window~~ #4386 
- [ ] Restore public resize method from plot - even if it just calls document.resize for now.

Tests
 - [ ] models/layouts/box.coffee
 - [ ] models/tools/toolbar_* - especially _init_tools and _merge_tools

Docs
 - [ ] Document off-canvas annotations
 - [ ] Note that off-canvas legends require (x, y)
 - [ ] Document responsiveness
 - [ ] Changes for use of embed with box responsive (you need to make sure the parent has a size)

##### Do after first 

- [ ] BoxAR is only known to work for a single plot in a single box.
- [ ] Legend is too tall (and maybe too wide) if more than two items
- [ ] Ensure that we have the hooks we need for phosphor (or similar) widgets having removed the method from Plot.
- [ ] Better document / investigate all the _ar modes which haven't been a focus.
- [ ] Mixed mode - start with limited support for a fixed box of widgets next to a resizable plot.

##### Done (or not doing)
- [x] Two plots in a Row is having a rough time in initialization - you have to manually resize for things to get ironed out.
- [x] put toolbar back into view 
- [x] don't resize based on window size
- [x] Restore aspect ratio capability for box responsiveness (was previously on plot.coffee resize_width_height method) - inc. responsive_width_height example 
- [x] Notebook support
- [x] In a grid of plots, toolbars can be set to any combination of `None`, `left`, `above` without a problem. But `right` or `below` will break layout - they do work on a single plot though. I'm not happy with how the constraints are being done for right & below either (including the gorpy hand-coded offset so they're not behind the plot) - the layout bug may sort itself out if the constraints improve.
- [x] tests - models/plots/plot.coffee
- [x] GridPlot is completely broken & needs re-implementing.
- [x] Make WidgetColumn and WidgetRow. Fixing the left toolbar problem will probably fix mixed layout with a fixed widgetrow/column.
- [x] Widget boxes at top of plot e.g. ohlc - deprecations, WidgetRow and left toolbar problem will probably fix this.
- [x] ~~Do more in deprecations. Construct Columns or rows for people if they have mixed plots and widgets.~~ assuming we're ok with the small breaking change, this is now looking ok - widgets are auto-wrapped
- [x] Reset tool should trigger a resize event
- [x] ~~Resize tool should only be usable with responsive=fixed mode and should trigger a resize event on mouse_up~~ it's actually ok, with the reset tool resize for its limited use.
- [x] Allow rows & columns to accept None as a spacer (makes grid plot work again)
- [x] Spacer - have grid plot try and to the right thing in a fixed layout when passed None
- [x] Spacer - make it behave in a box layout
- [x] Fixed gridplot with left toolbar - column needs moving over
- [x] GMapPlot canvas

#### Known bugs/not-yet-features (planning to live with for a while)
* If responsive mode is `height_ar`, toolbar box `above` or `below` doesn't work
* `box_ar` mode doesn't work in grid plots
* zooming/panning so the axis labels increase in size does not trigger other plots to re-align - not sure whether we want this - seems expensive (maybe could do at the end of a gesture)


birdsarah:
Some basics are now working. The below screenshot is fully responsive.  You can run this yourself with the `mega_plot.py` that's in top level of bokeh on this branch.

screen shot 2016-05-20 at 7 54 03 am

An early start with reviewing would be great, but, for now there's no need to say x & y doesn't work as I'm already working on that laundry list. I'll send out a note, when I think things are "working" and need as many eye balls as I can get.

bokeh/core/enums.py
```diff
@@ -179,3 +179,5 @@ def enumeration(*values, **kwargs):
 #: Specify sorting directions
 SortDirection = enumeration("ascending", "descending")
 
+#: Responsive types
+Responsive = enumeration("box", "width_ar", "height_ar", "box_ar", "fixed")
```
bryevdv: What does "ar" stand for or signify?

bokeh/io.py
```diff
@@ -612,32 +615,29 @@ def gridplot(plot_arrangement, **kwargs):
...
-    ''' Generate a layout that arranges several subplots horizontally.
-
-    '''
-    _remove_roots(children)
-    layout = HBox(children=list(children), **kwargs)
-    curdoc().add_root(layout)
-    _push_or_save(layout)
+    warnings.warn(
+        """
+        The new Row is responsive by default, it resizes based on the space available.
```
bryevdv: suggest:  "Switching to 'Row' will make elements responsive by default (they will resize based on the space available)."

bryevdv: (same suggestion for others)

bokeh/models/layouts.py
```diff
@@ -33,6 +35,32 @@ class LayoutDOM(Model):
...
+    the aspect ratio (plot_width/plot_height) is maintained.
+
+    `height_ar` mode constrains only the height. The items being laid out will resize to
+    fit the height and will take up whatever width they may need. For a Plot,
+    the aspect ratio (plot_width/plot_height) is maintained.
+
+    `box_ar` mode constrains the width and height, but maintains the plot aspect ratio
+    for a plot inside the box.
+
+    """)
```
bryevdv: I think "constrains" might be too much information. Reading "`width_ar` mode constrains only the width." Made me think the width would be the dimension that was **not** responsive. Maybe just "In ``width_ar`` mode only the width varies responsively") Also single back ticks have no effect, if they are to highlight in sphinx docs, need to be double back ticks

bokeh/core/enums.py
birdsarah: It signifies that aspect_ratio is preserved during responsiveness.

bryevdv: Saw below, maybe just `width_aspect` is more self-evident in isolation, though

bokeh/models/plots.py
```diff
@@ -462,8 +476,20 @@ def _check_colon_in_category_label(self):
...
+    """)
+
+    toolbar_location = Enum(Location, help="""
+    Where the toolbar will be located. If set to None, no toolbar
+    will be attached to the plot.
+    """)
+
+    toolbar_sticky = Bool(default=True, help="""
+    Stick the toolbar to the edge of the plot. Default: True. If False,
+    the toolbar will be outside of the axes, titles etc.
```
bryevdv: what happens if it is `True` and there *are* axes, titles, etc?

bokeh/models/plots.py
```diff
@@ -552,6 +568,42 @@ def border_fill(self, color):
             """)
         self.border_fill_color = color
 
+    @property
+    def logo(self):
+        warnings.warn(
+            """
+            Plot property 'logo' was deprecated in Bokeh 0.12.0 and will be removed.
+            User 'toolbar.logo' instead.
```
bryevdv: "User" should be "Use", also suggest `Plot.toolbar.logo` to be more explicit

bryevdv: same comments below

bryevdv:
@almarklein @birdsarah something about the webgl scaling work  (I assume) is off:

screen shot 2016-05-23 at 9 36 07 pm

background only takes up 1/4 of the area it should

bryevdv:
Changing the data set results in this:

screen shot 2016-05-23 at 9 37 23 pm

was working fine earlier today.

bryevdv:
OK, the problem does not happen on master, so maybe an issue with the merge?

birdsarah:
This is ready for some review. Screenshots are super helpful, but as the page gets super long, its my intention to delete screenshots from the thread as they are resolved. I'll keep a local copy. 

1) Need to add titles back - waiting on @canavandl Title annotation PR
2) Layout still has some quirks that need to be worked through:
- fixed grid isn't right
- box_ar, can't be nested (this can probably be addressed in a follow-on)
- width_ar struggles with being in a complex layout
3) GridPlot temporarily doesn't support `None` - working on that now.

Some features:

1) There's a new layout object ToolbarBox - it can accept tools from multiple plots and merge them into single buttons - it is layoutable so it could be embedded seperately from plots
2) The toolbar's been updated and has a new "sticky" default - we may want to make the stickyness smarter in the future, or add some logic to figure
3) Gridplot is now all python side - its a combination of a layout and ToolbarBox
4) There are 5 responsive modes: fixed, box, width_ar, height_ar, box_ar _ar modes maintain the aspect ratio of the plot
5) box modes will not work in the notebook because the notebook output cell doesn't have a height to "read" from.
6) there is no more auto-magic adding things to documents meaning we should be able to use figure with server (haven't tested)
7) embedding works, but users may need to make a tweak to their code - see how i needed to update the embedding example

I have a series of scripts I've been using for testing, I'll post links to them here.

birdsarah:
I need to document the api breaks, they are minor except for one.

The one is that the js resize method is gone so https://github.com/bokeh/bokeh/blob/master/examples/embed/embed_responsive_width_height.py would not longer work.

ping @almarklein on this.

almarklein:
@birdsarah I think we do need a method *like* that in order to embed Bokeh plots in apps (e.g. with Phosphor). It's used in the Bokeh-Phosphor demo at least.

birdsarah:
@almar, I should clarify, that layout feature is available just not through javascript. You just set `Plot(...., 
responsive='box_ar')`

birdsarah:
Just noting that @almarklein and I have started discussing the hooks necessary for phosphor widgets. I'll make some tweaks now - making the resize eventHandler customizable. But follow-on work may be necessary.

birdsarah:
Testing notes from @bryevdv 

```sh
examples/app
------------

* ohlc

    layout proportions, missing "top" widgets

* selection_histogram

    layout proportions (needs gridplot?)

* sliders.py

    right edge of widgets clipped just barely 

    happens elsehwere, just needs a bit more padding?

* stocks

    pretty off in a few different ways (sizes, some widget overlap)


plotting/server
---------------

* geojson_points

    blank page, no console error

* hover
    
    blank page, console: 

    TypeError: undefined is not an object (evaluating 'this.plot.plot_canvas().add_renderers')

* line_animate_widget.py

    "top" widgets covered up

* selection_histogram.py

    same as app version


plotting/file
-------------

* color_sliders.py

    just minor "right edge" widget clipping

* linked_brushing.py
* linked_panning.py

    reeeallly tall plots :)

* line_compare.py

    layout seems off


models
------

* buttons_server.py

    "button_type" issue (not a layout problem)

* data_tables_server.py

    table rendering to the left, even tho in a vbox

* population_server.py
* taylor_server.py
* transform_interpolator.py

    "top" widgets covered up

* transform_jitter.py

    sticky toolbar covers labels (not a layout problem, per se)

* iris_splom.py

    occasionally on resize one or two plots end up slightly the wrong size

* grid.py

    panning to regions with larger axis labels - axis gutter adjusts but plots below don't 
    also adjust

* trail.py 

    blank page, console:

    Error: array item 1 must be a variable

* calendars.py

    blank page, console:

    Error: unsatisfiable constraint

* custom.py

    might be working as expected, custom div very big tho


charts/file
-----------

    All "multi" ones look great!
```

birdsarah:
@bryevdv, there are four items in the todo list at the top of this pr that when all done should fix a bunch of the layout issues you've noted. I'll ping you when that's done.

birdsarah:
Bug with crosshair tool and server causes `examples/plotting/server/hover` to fail, and causes `examples/app/slider` to load intermittently.

The offending line is: https://github.com/bokeh/bokeh/blob/birdsarah/4164_rework_layout/bokehjs/src/coffee/models/tools/inspectors/crosshair_tool.coffee#L81
```coffeescript
@plot.plot_canvas().add_renderers(@spans.width)
```
Crosshairs are trying to be created before plot_canvas has been created by plot.

birdsarah:
Tabs inside a layout doesn't work - it's trying to render plot_canvas before they're ready - similar to crosshairs issue.

birdsarah:
ping @bryevdv 

This is starting to shape up. Titles are back, but I need to clean-up the api - deprecate properties etc.

At the top of this list is a prioritized todo list. Can you check through this and see if you have any comments. 

I'm struggling with the crosshair issue and would appreciate it if you have time to look into it.

bryevdv:
@birdsarah I will take a look now

bryevdv:
@birdsarah here is a "quick fix" diff you can apply if you like, I think we should visit the `Plot`, `PlotCanvas` breakdown, and actually all of the entire gorpy `build_levels` cruft, but not in this PR. 
```
index a1e0ed9..1c2f07a 100644
@@ -111,6 +111,10 @@ class PlotCanvasView extends Renderer.View
     @build_levels()
     @bind_bokeh_events()
 
+    # TODO (bev) ugly - this second one is to handle views added in the first
+    # build_levels, that may have added renderers before build_levels
+    @build_levels()
+
     @update_dataranges()
 
     @unpause()
@@ -566,7 +570,7 @@ class PlotCanvas extends LayoutDOM.Model
         @min_border_left = @min_border
       if not @min_border_right?
         @min_border_right = @min_border
-    
+
     logger.debug("Plot initialized")
 
   _doc_attached: () ->
diff --git a/bokehjs/src/coffee/models/tools/inspectors/crosshair_tool.coffee b/bokehjs/src/coffee/models/tools/inspectors/crosshair_tool.coffee
index 067b6e5..70cdba6 100644
--- a/bokehjs/src/coffee/models/tools/inspectors/crosshair_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/inspectors/crosshair_tool.coffee
@@ -6,6 +6,11 @@ p = require "../../../core/properties"
 
 class CrosshairToolView extends InspectTool.View
 
+  initialize: (options) ->
+    super(options)
+    @model.plot.plot_canvas().add_renderers(@model.spans.width)
+    @model.plot.plot_canvas().add_renderers(@model.spans.height)
+
   _move: (e) ->
     if not @mget('active')
       return
@@ -78,9 +83,6 @@ class CrosshairTool extends InspectTool.Model
       })
     }
 
-    @plot.plot_canvas().add_renderers(@spans.width)
-    @plot.plot_canvas().add_renderers(@spans.height)
-
 module.exports =
   Model: CrosshairTool
   View: CrosshairToolView
```

bryevdv:
With this change, the notebook examples I tried (which happened to have crosshair) work. Will note that "run all" on `glyphs.ipynb` seems to run super slow tho

bryevdv:
Seeing this in notebook comms basic usage notebook:

screen shot 2016-05-26 at 10 57 33 am

mattpap:
@bryevdv, the answer to fixing crosshair tool is using "synthetic" renderers that I added recently. I have somewhere a commit or stash laying around fixing precisely this issue, but you can follow hover tool's implementation.

bryevdv:
@mattpap thanks, I'll take a further look

bryevdv:
@mattpap that worked like a charm, thanks

@birdsarah the crosshair issue should be resolved now

bryevdv:
FYI, with `burtin.py`:

screen shot 2016-05-26 at 2 18 16 pm


shading toolbar maybe desired but maybe not

birdsarah:
Re: burtin - that's a good example of where we want `toolbar_sticky=False` - what's happened is that the canvas panel has expanded to accomodate the toolbar. And the canvas panel was colored - but that's not what we want in this case.

birdsarah:
Amazing work on the crosshair - thanks!

bokeh/core/enums.py
almarklein: To me `width_aspect` sounds like width-aspect. Something like `width_fixed_aspect_ratio` would be clear but way too long. `width_far` or `width_ar` is a bit cryptic imo. 

What about just 'width', 'height', 'both' for the fixed ar ones? Then 'fixed' can stay, and for what is now 'box' we can use 'free' or 'full', or keep it 'box'.

bokeh/core/properties.py
```diff
@@ -1855,3 +1855,29 @@ def transform(self, value):
...
+    def __init__(self, default=None, help=None):
+        types = (Enum(enums.Responsive), Bool)
+        super(Responsive, self).__init__(*types, default=default, help=help)
+
+    def transform(self, value):
+        if value is True:
+            value = 'width_ar'
+        if value is False:
+            value = 'fixed'
+        return value
```
almarklein: Could be `return {True: 'width_ar', False: 'fixed'}.get(value, value)`

bokeh/models/layouts.py
almarklein: The word "only" bothers me, because both width and height are adjusted, really. Maybe something like "In width_ar mode the width is adjusted while the aspect ratio is maintained.".

bryevdv:
@bokeh/dev OK, merging. Still things to do and fix but let's do them incrementally in follow-on PRs.

PR #4346, opened 2016-05-20
WebGL supports HDPI and HDPI is updated on browser zoom

### LINKED ISSUE #4293: WebGL blurry on OS X, not visible in Safari
Hi all,

I'm noticing that WebGL output appears to be very blurry in Chrome/Firefox on OS X, and completely invisible in Safari. With WebGL disabled, the glyphs look crisp and clear:

screenshot 2016-05-07 10 11 21

With WebGL enabled, it becomes a blurry mess:

screenshot 2016-05-07 10 11 04

...and on Safari none of the glyphs are visible! I've noticed the same behavior with line plots too, very blurry:

screenshot 2016-05-07 10 21 46

You can see the blurry version here: https://sne.space/sne/SN1987A-blurry/, the non-blurry version here: https://sne.space/sne/SN1987A/.

Any idea on how to improve the WebGL rendering? And how to make the plots work in Safari?

Thanks!

### LINKED ISSUE #4345: When using the browser zoom the plot content becomes blurry
To reproduce:
* Show a plot

almarklein:
Issues: fixes #4345, fixes #4293.

This PR:
* makes WebGL aware of the pixel_ratio for HDPI support. 
* improves/simplifies the blitting of the gl canvas into the 2d canvas. 
* tweaks some WebGL markers to look better when zoomed in.
* makes plot update when the browser zoom is changed.

Consequently, our WebGL glyphs have never looked so nice! Maybe even nicer than the 2D glyphs (due to better aa). Should also look great on retina displays now.

### DIFF
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -47,15 +47,23 @@ class CanvasView extends BokehView
     ctx = canvas_el[0].getContext('2d')
     return ctx
 
+  need_render: () ->
+    # Public method so the plot can determine if size/dpi changed
+    width = @mget('width')
+    height = @mget('height')
+    dpr = window.devicePixelRatio  # this changes when browser zoom changes
+    return not _.isEqual(@last_dims, [width, height, dpr])
+
   render: (force=false) ->
 
     width = @mget('width')
     height = @mget('height')
+    dpr = window.devicePixelRatio
 
     # only render the canvas when the canvas dimensions change unless force==true
-    if not _.isEqual(@last_dims, [width, height]) or force
+    if not _.isEqual(@last_dims, [width, height, dpr]) or force
 
-      ratio = get_scale_ratio(@ctx, @mget('use_hidpi'))
+      @_pixel_ratio = ratio = get_scale_ratio(@ctx, @mget('use_hidpi'))
 
       logger.debug("Rendering CanvasView [force=#{force}] with width: #{width}, height: #{height}, ratio: #{ratio}")
 
@@ -69,12 +77,9 @@ class CanvasView extends BokehView
       @$('div.bk-canvas-overlays').attr('style', "z-index:75; position:absolute; top:0; left:0; width:#{width}px; height:#{height}px;")
       @$('div.bk-canvas-events').attr('style', "z-index:100; position:absolute; top:0; left:0; width:#{width}px; height:#{height}px;")
 
-      @last_dims = [width, height]
+      @last_dims = [width, height, dpr]
 
-    else
-      ratio = get_scale_ratio(@ctx, @mget('use_hidpi'))
-
-    return ratio
+    return @_pixel_ratio
 
   set_dims: (dims, trigger=true) ->
     @requested_width = dims[0]
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -69,12 +69,12 @@ class CanvasView extends BokehView
       @$('div.bk-canvas-overlays').attr('style', "z-index:75; position:absolute; top:0; left:0; width:#{width}px; height:#{height}px;")
       @$('div.bk-canvas-events').attr('style', "z-index:100; position:absolute; top:0; left:0; width:#{width}px; height:#{height}px;")
 
-      @ctx.scale(ratio, ratio)
-      @ctx.translate(0.5, 0.5)
-
       @last_dims = [width, height]
 
-    return
+    else
+      ratio = get_scale_ratio(@ctx, @mget('use_hidpi'))
+
+    return ratio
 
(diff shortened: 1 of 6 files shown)


almarklein:
@birdsarah this might touch to some code related to your work with resizing etc.

birdsarah:
@almarklein, yes it will and a i'm sure there'll be a merge conflict on canvas and plot_canvas. Not a problem as it's only a little code. Thanks for the heads-up and the small-sized PR :D

One observation is that with tests it would be easier to know that I hadn't screwed up fixing the merge conflict. I'm sure this'll get merged before #4343 so once I've merged master into my PR I'll ping you to make sure this hasn't gone wrong.

birdsarah:
I should also have said that this is pretty cool. Can I clarify something as it relates to layout. Will the canvas resize or will it stay the same size and the pixels will update? I'm assuming it stays the same size, but just wanted to check. Also with the new layout there maybe some cases where we don't need to trigger a re-render as one will have been triggered anyway, so we may want to look at this.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -170,6 +170,12 @@ class PlotCanvasView extends Renderer.View
       # Just need to wait a small delay so container has a width
       _.delay(@resize, 10)
 
+    # Re-render when dpi changes (modern browsers emit a resize event on zoom)
+    check_resize_need_render = () ->
+      if @canvas_view.need_render()
+         @request_render()
+    $(window).on("resize", check_resize_need_render.bind(this))
```
birdsarah: I've started [using jquery](https://github.com/bokeh/bokeh/pull/4343/files#diff-f9194b68ef1a3a51bbb8f5b03193b403R91) proxy for this kind of thing: `$(window).on("resize", $.proxy(@resize, @))`

My default is an assumption that jquery's thought about this for me. As I haven't thought about it, do you have any opinions - then we can be consistent.

bokehjs/src/coffee/models/canvas/canvas.coffee
```diff
@@ -47,15 +47,23 @@ class CanvasView extends BokehView
     ctx = canvas_el[0].getContext('2d')
     return ctx
 
+  need_render: () ->
+    # Public method so the plot can determine if size/dpi changed
+    width = @mget('width')
+    height = @mget('height')
```
birdsarah: In the new layout system, these may not be the right variables to check. I'm still ironing out the details of that. Is there a functional test so we can know if we got this right?

bokehjs/src/coffee/models/canvas/canvas.coffee
```diff
@@ -47,15 +47,23 @@ class CanvasView extends BokehView
...
+    dpr = window.devicePixelRatio  # this changes when browser zoom changes
+    return not _.isEqual(@last_dims, [width, height, dpr])
```
birdsarah: I'm a bit confused, I thought `canvas.set_dims` does not update `model.width` and `model.height` - facts to the contrary gratefully appreciated. If that's true, doesn't that mean that any time an object is responsive `model.width` and `model.height` will have gotten out of sync with `@last_dims` - even under current responsive code?

bokehjs/src/coffee/models/plots/plot_canvas.coffee
almarklein: I don't know jquery that well, but I'll update to `$.proxy()` to be consistent :)

bokehjs/src/coffee/models/canvas/canvas.coffee
almarklein: Yes: change your browser zoom from 100% to say 150%. The text (and glyphs) should remain crisp. If it only switches after manipulating the plot, it did not work.

Testing this with Selenium is probably a bit difficult? Maybe we could test that the `pixel_ratio` is indeed changed or something. Selenium does seem [capable of zooming](http://stackoverflow.com/questions/15024756).

almarklein:
@birdsarah thanks for your comments

>  Will the canvas resize or will it stay the same size and the pixels will update?

The canvas `style.width` and `style.height` remain the same, so layout-wise, there is no resize. The `canvas.width` and `canvas.height` can be changed though, as to match the canvas pixels to the screen pixels.

> I thought `canvas.set_dims` does not update `model.width` and `model.height`

Not sure, good point. I was assuming that the existing code worked correctly. I will do some more testing with responsive plots.

Would it make sense to split this PR up and: 1) merge the WebGL part into master; 2) merge master into your branch; 2) let me do a PR for the zoom stuff against your branch?

birdsarah:
@almarklein, that merge strategy sounds great, and then we can check it's still working. fwiw, i was planning to use width & height more - i'm just nailing down the details today in fact. i'm not sure if this'll affect this yet - sorry for being a bit vague - my head is in a strange corner of bokeh right now :)

almarklein:
Superseded by #4348 (not deleting branch yet, because it has some stuff I'd like to contribute later)

PR #4348, opened 2016-05-21
Make hidpi work for webgl

### LINKED ISSUE #4293: WebGL blurry on OS X, not visible in Safari

almarklein:
Issues: fixes #4293. This is like #4346 but without the browser-zoom update stuff.

This PR:
* makes WebGL aware of the pixel_ratio for HDPI support. 
* improves/simplifies the blitting of the gl canvas into the 2d canvas. 
* tweaks some WebGL markers to look better when zoomed in.
* ~~makes plot update when the browser zoom is changed.~~

Consequently, our WebGL glyphs have never looked so nice! Maybe even nicer than the 2D glyphs (due to better aa). Should also look great on retina displays now.

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -448,14 +448,20 @@ class PlotCanvasView extends Renderer.View
         @canvas.get("canvas_height") != height)
       @canvas_view.set_dims([width, height], trigger=false)
 
-    @canvas_view.render(force_canvas)
-
     for k, v of @renderer_views
       if not @range_update_timestamp? or v.set_data_timestamp > @range_update_timestamp
         @update_dataranges()
         break
 
     ctx = @canvas_view.ctx
+    ctx.save()  # Save default state
+
+    # Set hidpi-transform
+    ratio = @canvas_view.render(force_canvas)
+    ctx.pixel_ratio = ratio  # we need this in WebGL
+    ctx.scale(ratio, ratio)
+    ctx.translate(0.5, 0.5)
+
     title = @mget('title')
     if title
       @visuals.title_text.set_value(ctx)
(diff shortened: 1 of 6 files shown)


almarklein:
@birdsarah the amount of code that conflicts with your works should be much less for this PR.

Once this is merged in master, you can merge master into your epic PR when you're up for it. Ping me when it's done, and I'll contribute the zoom-update mechanics to your PR.

birdsarah:
@almarklein - thank you so much - I really appreciate it.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -448,14 +448,20 @@ class PlotCanvasView extends Renderer.View
...
     for k, v of @renderer_views
       if not @range_update_timestamp? or v.set_data_timestamp > @range_update_timestamp
         @update_dataranges()
         break
 
     ctx = @canvas_view.ctx
+    ctx.save()  # Save default state
+
+    # Set hidpi-transform
+    ratio = @canvas_view.render(force_canvas)
```
bryevdv: It's unusual for `render` to return anything. But more than that, this ratio is only conditionally returned on one side of branch inside `render`, which is both hard to reason about and potentially fragile. I think this ratio should be communicated in a different manner. E.g., an internal property set on canvas init, a method that can be called, or just calling `get_scale_ratio(@ctx, @ctx.get('use_hidpi'))` here explicitly.

almarklein: `Canvas.render` always returns ratio, I think you missed an indent :)  but I can see your concern for render returning something. Maybe we should just rename that method, because it does not render anything, it makes sure the canvas is setup correctly.

That said, are you ok if we fix this in the PR that I'll make against @birdsarah 's branch? It requires changes to that method anyway, and I'd like to minimize the code changes to that code here to avoid conflicts.

bryevdv: Well I was thinking `ratio` is only computed in one branch, but I see the line computing it is duplicated exactly at the top of the other branch.  That should definitely be fixed. If another PR is easier, I guess that's OK.

almarklein: Right, I had fixed that in the next commit (which eventually did not make it in this PR). I will clean this up / add a clear todo. Sorry for the little sloppiness here.

bryevdv:
OK would say merge this whenever you like then.

PR #3175, opened 2015-11-23
Add bounds to ranges (limits pan and zoom)

birdsarah:
Fixes #2942

PR #4349, opened 2016-05-21
Ranges can have minimum range to constrain min zoom level

### LINKED ISSUE #572: Support for pan/zoom range limits
Should be able to configure plots with hard limits for panning and zooming.

almarklein:
fixes #572

Implemented by allowing a third parameter in the `bounds` argument.

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1561,6 +1561,36 @@ class Datetime(PropertyDescriptor):
         return value
         # Handled by serialization in protocol.py for now
 
+class TimeDelta(PropertyDescriptor):
+    """ TimeDelta type property.
+
+    """
+
+    def __init__(self, default=datetime.timedelta(), help=None):
+        super(TimeDelta, self).__init__(default=default, help=help)
+
+    def validate(self, value):
+        super(TimeDelta, self).validate(value)
+
+        timedelta_types = (datetime.timedelta,)
+        try:
+            import numpy as np
+            timedelta_types += (np.timedelta64,)
+        except ImportError:
+            pass
+
+        if (isinstance(value, timedelta_types)):
+            return
+
+        if pd and isinstance(value, (pd.Timedelta)):
+            return
+
+        raise ValueError("Expected a timedelta instance, got %r" % value)
+
+    def transform(self, value):
+        value = super(TimeDelta, self).transform(value)
+        return value
+        # Handled by serialization in protocol.py for now
 
 class RelativeDelta(Dict):
     """ RelativeDelta type property for time deltas.
--- a/bokeh/models/ranges.py
+++ b/bokeh/models/ranges.py
@@ -69,7 +69,7 @@ class Range1d(Range):
 
     min_interval = Either(Float, TimeDelta, Int, default=None, help="""
     The level that the range is allowed to zoom in, expressed as the
-    minimum visible interval. If set to ``None`` (defaut), the minimum
+    minimum visible interval. If set to ``None`` (default), the minimum
     interval is not bound. Can be a timedelta. """)
 
     max_interval = Either(Float, TimeDelta, Int, default=None, help="""
@@ -146,7 +146,7 @@ class DataRange1d(DataRange):
 
     min_interval = Float(default=None, help="""
     The level that the range is allowed to zoom in, expressed as the
-    minimum visible interval. If set to ``None`` (defaut), the minimum
+    minimum visible interval. If set to ``None`` (default), the minimum
     interval is not bound.""")
 
     max_interval = Float(default=None, help="""
@@ -263,7 +263,7 @@ class FactorRange(Range):
 
     min_interval = Int(default=None, help="""
     The level that the range is allowed to zoom in, expressed as the
-    minimum number of visible categories. If set to ``None`` (defaut),
+    minimum number of visible categories. If set to ``None`` (default),
     the minimum interval is not bound.""")
 
     max_interval = Int(default=None, help="""
--- a/bokeh/models/ranges.py
+++ b/bokeh/models/ranges.py
@@ -9,7 +9,8 @@ from ..model import Model
 from ..core.enums import StartEnd
 from ..core.properties import abstract
 from ..core.properties import (
-    Auto, Bool, Int, Float, String, Datetime, Instance, List, Either, Enum, MinMaxBounds
+    Auto, Bool, Int, Float, String, Datetime, TimeDelta, Instance, List,
+    Either, Enum, MinMaxBounds,
 )
 from .callbacks import Callback
 from .renderers import Renderer
@@ -65,6 +66,16 @@ class Range1d(Range):
         Range1d(0, 1, bounds='auto')  # Auto-bounded to 0 and 1 (Default behavior)
         Range1d(start=0, end=1, bounds=(0, None))  # Maximum is unbounded, minimum bounded to 0
     """)
+
+    min_interval = Either(Float, TimeDelta, Int, default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum visible interval. If set to ``None`` (defaut), the minimum
+    interval is not bound. Can be a timedelta. """)
+
+    max_interval = Either(Float, TimeDelta, Int, default=None, help="""
+    The level that the range is allowed to zoom out, expressed as the
+    maximum visible interval. Can be a timedelta. Note that ``bounds`` can
+    impose an implicit constraint on the maximum interval as well. """)
 
     def __init__(self, *args, **kwargs):
         if args and ('start' in kwargs or 'end' in kwargs):
@@ -132,6 +143,16 @@ class DataRange1d(DataRange):
     If you only want to constrain one end of the plot, you can set min or max to
     ``None`` e.g. ``DataRange1d(bounds=(None, 12))``
     """)
+
+    min_interval = Float(default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum visible interval. If set to ``None`` (defaut), the minimum
+    interval is not bound.""")
+
+    max_interval = Float(default=None, help="""
+    The level that the range is allowed to zoom out, expressed as the
+    maximum visible interval. Note that ``bounds`` can impose an
+    implicit constraint on the maximum interval as well.""")
 
     flipped = Bool(default=False, help="""
     Whether the range should be "flipped" from its normal direction when
@@ -239,6 +260,16 @@ class FactorRange(Range):
         The plot will display the chart with only the factors ["apples", "peaches", "bananas"] (in that order)
         and the plot will not pan left of apples or right of bananas.
     """)
+
+    min_interval = Int(default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum number of visible categories. If set to ``None`` (defaut),
+    the minimum interval is not bound.""")
+
+    max_interval = Int(default=None, help="""
+    The level that the range is allowed to zoom out, expressed as the
+    maximum number of visible categories. Note that ``bounds`` can
+    impose an implicit constraint on the maximum interval as well.""")
 
     def __init__(self, *args, **kwargs):
         if args and "factors" in kwargs:
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -246,46 +246,105 @@ class PlotCanvasView extends Renderer.View
   reset_selection: () ->
     @update_selection(null)
 
-  _update_single_range: (rng, range_info, is_panning) ->
-    # Is this a reversed range?
-    reversed = if rng.get('start') > rng.get('end') then true else false
+  _update_ranges_together: (range_info_iter) ->
+    # Get weight needed to scale the diff of the range to honor interval limits
+    weight = 1.0
+    for [rng, range_info] in range_info_iter
+      weight = Math.min(weight, @_get_weight_to_constrain_interval(rng, range_info))
+    # Apply shared weight to all ranges
+    if weight < 1
+      for [rng, range_info] in range_info_iter
+        range_info['start'] = weight * range_info['start'] + (1-weight) * rng.get('start')
+        range_info['end'] = weight * range_info['end'] + (1-weight) * rng.get('end')
 
-    # Prevent range from going outside limits
-    # Also ensure that range keeps the same delta when panning
+  _update_ranges_individually: (range_info_iter, is_panning, is_scrolling) ->
+    hit_bound = false
+    for [rng, range_info] in range_info_iter
+      # Is this a reversed range?
+      reversed = if rng.get('start') > rng.get('end') then true else false
 
-    if rng.get('bounds')?
-      min = rng.get('bounds')[0]
-      max = rng.get('bounds')[1]
+      # Limit range interval first. Note that for scroll events,
+      # the interval has already been limited for all ranges simultaneously
+      if not is_scrolling
+        weight = @_get_weight_to_constrain_interval(rng, range_info)
+        if weight < 1
+            range_info['start'] = weight * range_info['start'] + (1-weight) * rng.get('start')
+            range_info['end'] = weight * range_info['end'] + (1-weight) * rng.get('end')
 
-      if reversed
-        if min?
-          if min >= range_info['end']
-            range_info['end'] = min
-            if is_panning?
-              range_info['start'] = rng.get('start')
-        if max?
-          if max <= range_info['start']
-            range_info['start'] = max
-            if is_panning?
-              range_info['end'] = rng.get('end')
-      else
-        if min?
-          if min >= range_info['start']
-            range_info['start'] = min
-            if is_panning?
-              range_info['end'] = rng.get('end')
-        if max?
-          if max <= range_info['end']
-            range_info['end'] = max
-            if is_panning?
-              range_info['start'] = rng.get('start')
+      # Prevent range from going outside limits
+      # Also ensure that range keeps the same delta when panning/scrolling
+      if rng.get('bounds')?
+        min = rng.get('bounds')[0]
+        max = rng.get('bounds')[1]
+        new_interval = Math.abs(range_info['end'] - range_info['start'])
 
-    if rng.get('start') != range_info['start'] or rng.get('end') != range_info['end']
+        if reversed
+          if min?
+            if min >= range_info['end']
+              hit_bound = true
+              range_info['end'] = min
+              if is_panning? or is_scrolling?
+                range_info['start'] = min + new_interval
+          if max?
+            if max <= range_info['start']
+              hit_bound = true
+              range_info['start'] = max
+              if is_panning? or is_scrolling?
+                range_info['end'] = max - new_interval
+        else
+          if min?
+            if min >= range_info['start']
+              hit_bound = true
+              range_info['start'] = min
+              if is_panning? or is_scrolling?
+                range_info['end'] = min + new_interval
+          if max?
+            if max <= range_info['end']
+              hit_bound = true
+              range_info['end'] = max
+              if is_panning? or is_scrolling?
+                range_info['start'] = max - new_interval
+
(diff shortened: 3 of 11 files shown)


almarklein:
Needs test, but I'd first like a green light on whether it makes sense to implement it in this way (API-wise).

almarklein:
Actually, although a maximum range can be enforced via the min and max bounds, I think it makes sense to allow setting a max range independently, e.g. having no bounds in x and y, but still a bound on zoom level.

We could do either `Range1D( ... bounds=(1, 10, 1, 5))` or `Range1D(..., bounds=(1, 10), range_bounds=(1, 5)`. I think I lean towards the latter.

birdsarah:
Thanks for taking this on. There was a lot of debate about the api for bounds so I think @bryevdv might have some strong opinions.

Some thoughts:
* I think we should do min_zoom and max_zoom at the same time
* I am -1 on adding parameters to the bounds argument, I think it's going to be very confusing remembering what is what. And, as you said, you may well want to set zoom independently from range bounds.
* I think we do want this to apply for categorical axes - what's the spelling for that? My initial thought is something like n units in view so if it was 2 - you'd see a minimum of two categories on the plot when you zoomed in.
* I think it would be cool if you could pass in a [timedelta object](https://docs.python.org/3/library/datetime.html?highlight=time%20delta#datetime.timedelta)

My initial thought is `Range1d(...., min_zoom=<interval>, max_zoom=<interval>)` but my recollection was that @bryevdv was -1 on `min_` `max_` pre/sufffixing things - I believe because it becomes  verbose. I think `Range1d(..., zoom_bounds=(<min_interval>, <max_interval>))` is fine too.

birdsarah:
> We could do either Range1D( ... bounds=(1, 10, 1, 5)) or Range1D(..., bounds=(1, 10), range_bounds=(1, 5). I think I lean towards the latter.

Some related thoughts above, but I wanted to say, I'm not a fan of `range_bounds`. I find the following kind of confusing:

```python
range = Range1d(1, 5)
range.bounds = (1, 4)
range.range_bounds = (1, 5)
```

birdsarah:
`interval_bounds` ??

birdsarah:
In case it's useful here's the discussion on range bounds: https://github.com/bokeh/bokeh/pull/3175 I was wrong, the bulk of the conversation was about that fact the `start` and `end` or `upper` and `lower` were not appropriate in the case of bounds (as they imply ordering).

It also has the following references in case useful:

---
d3 defines a rangeExtent and scaleExtent that would be used in this context.

https://github.com/mbostock/d3/wiki/Ordinal-Scales#ordinal_rangeExtent
`ordinal.rangeExtent()`
Returns a two-element array representing the extent of the scale's range, i.e., the smallest and largest values.

https://github.com/mbostock/d3/wiki/zoom-behavior#scaleExtent
`zoom.scaleExtent([extent])`
Specifies the zoom scale's allowed range as a two-element array, [minimum, maximum]. If not specified, returns the current scale extent, which defaults to [0, Infinity].

---

I also had a poke around Grammar of Graphics and didn't see anything conclusive, but it's a big book so may have missed it.

---

ggplot2 - http://docs.ggplot2.org/current/xylim.html
has the concept of xlim and ylim
xlim(15, 20) 
for discrete data you can do: xlim("a", "b", "c")

---

vega has a concept of domainMin/Max and rangeMin/Max domainMin/Max is for quantitative scales only, but rangeMin/Max appears to be for all (I haven't seen an example with this in action though so not 100%)

https://github.com/vega/vega/wiki/Scales

almarklein:
I like `zoom_bounds=(<min_interval>, <max_interval>))`, because the bounds apply to zooming, and are based on a given interval. That's what I've implemented now (but yet only for `Range1D`, until we reach agreement on the API). I also added timedelta.

To be honest, the API for the `Range` seems a bit asymmetric to me: we have three "limits" that are each represented by two values:
* initial range (now `min` and `max` scalars)
* absolute boundaries (now `bounds` tuple)
* zoom boundaries (to be added)

Why should we treat the initial range different than the other limits? Though it might be too late to change this API ...

birdsarah:
> Why should we treat the initial range different than the other limits? Though it might be too late to change this API ...

We already convert `Range1d(1, 2)` to `Range1d(start=1, end=2)` as a convenience, I don't see why we couldn't have `Range1d(1, 2)` and `Range1d(start=1, end=2)` go to `Range1d((1, 2))`. And just have start & end be convenience getters/setters as opposed to properties.

Getting the underlying API right before 1.0 I think is a good idea. I'm not sure whether we'd ever want to get rid of start and end though as it's just a lot nicer doing this:
```python
def update():
    if plot.range.start < 12:
        plot.range.start = 0
```
vs:

```python
def update():
   if plot.range.bounds[0] < 12:
       plot.range.bounds = (0, plot.range.bounds[1])
```

Having said that. Actually it may be that we want `bounds` tuple to be the odd one out. I was wrong in my earlier recollection. @bryevdv is not averse to `min_` `max_`  it's just that they are very problematic for a bounds argument where the meaning can change by context.

I don't think that same confusion is true for `zoom_bounds`. What about `min_interval` and `max_interval`? That leaves bounds as the odd-one-out for very good reasons (documented on #3175)

bryevdv:
> We already convert Range1d(1, 2) to Range1d(start=1, end=2) as a convenience, I don't see why we couldn't have Range1d(1, 2) and Range1d(start=1, end=2) go to Range1d((1, 2)). And just have start & end be convenience getters/setters as opposed to properties.

Protocol change, will create work for all language binding maintainers. I personally like `start` and `end` for the example code reason given 

I think this proposal is off, though. The only thing missing right now is to prevent from "zooming in" too far, AFAIU, and that does not need two number, just one, a `min_interval` or something. We also have to consider how this would work e.g. with log scales.

almarklein:
> zooming out is already controllable by setting the existing bounds, and I think that is enough. 

I am starting to think that the fact that zooming out can be constrained by the bounds may be a *bad thing* ... 

What about this proposal: The `bounds` property defines the boundaries for the *center* of the view, instead of the edges of the view. This would make the `bounds` and `zoom_bounds` orthogonal (and free of conflicts). Plus I think in this situation it *would* make sense to make `'auto'` the default bounds.

If we end up with two zoom boundaries, I'd prefer to keep it a tuple, because `zoom_bounds` feels similar to `bounds`; having `min` and `max` being the "odd one", is acceptable, I guess.

bryevdv:
I understand the thinking, but it seems very unusual. Can you point to any other systems that specify things in this way? Let's also ask a more concrete question: I'm a user and I want to keep the plot in between `(0,10)`. What do I set? `bounds = (0,10)` is pretty trivial to explain and understand. With a scheme described... `bounds=(1,8), zoom_bounds=(2,0)` ? It seems like we would be making the simplest, most commonly desired use-case really hard and difficult to explain. 

I agree with @birdsarah and suggest `min_interval` and `max_interval` which would apply wherever the plot is centered. It's possible you want tot plot to be contrained in between 0 and 10, but you also don't want a view range bigger than 5 anywhere inside that range. This captures that use case, and keeps the common use case simple.

bokehjs/src/coffee/models/ranges/data_range1d.coffee
```diff
@@ -17,6 +17,8 @@ class DataRange1d extends DataRange.Model
       follow_interval: [ p.Number        ]
       default_span:    [ p.Number, 2     ]
       bounds:          [ p.Any           ] # TODO (bev)
+      min_interval: [ p.Any ]
+      max_interval: [ p.Any ]
```
almarklein: `min_interval` and `max_interval` can be anything that `min` and `max` can be, except that they can also be zero. What type should I use here?

almarklein:
Done from my end, except for the one question above. Better do a squash merge, because this was going back and forth.

bokeh/models/ranges.py
```diff
@@ -240,6 +259,15 @@ class FactorRange(Range):
...
 
+    min_interval = Int(default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum visible categories. If set to ``None`` (defaut), the minimum
+    interval is not bound.""")
+
+    max_interval = Int(default=None, help="""
+    The level that the range is allowed to zoom out, expressed as the
+    maximum visible categories.""")
+
```
bryevdv: What is the meaning/interpretation of these wrt factor ranges?

almarklein: The number of visible categories. I will add "number" to the docstring.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -338,6 +338,22 @@ class PlotCanvasView extends Renderer.View
...
+      sign = if reversed then -1 else 1
+
+      if min_interval > 0
+        if current_interval < min_interval
+          range_info['start'] = center - sign * 0.5 * min_interval
+          range_info['end'] = center + sign * 0.5 * min_interval
+      if max_interval > 0
+        if current_interval > max_interval
+          range_info['start'] = center - sign * 0.5 * max_interval
+          range_info['end'] = center + sign * 0.5 * max_interval
```
bryevdv: I this the right thing to do? I would probably favor just ignoring the update altogether if the new interval would be to big or too small. This is because e.g. the zoom tool keeps the data coordinates of the cursor constant, but doing the above could violate that, unless I'm mistaken. That would be surprising and undesirable, at least to me.

bokeh/core/properties.py
```diff
@@ -1561,6 +1561,36 @@ def transform(self, value):
...
+
+        if pd and isinstance(value, (pd.Timedelta)):
+            return
+
+        raise ValueError("Expected a timedelta instance, got %r" % value)
+
+    def transform(self, value):
+        value = super(TimeDelta, self).transform(value)
+        return value
+        # Handled by serialization in protocol.py for now
```
bryevdv: is this supposed to be in this PR?

bryevdv: Ah, to support time deltas for intervals, ok..

bokehjs/src/coffee/models/plots/plot_canvas.coffee
bryevdv: I also think `new_interval` might be more accurately descriptive than `current_interval`, it had me confused for a bit.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -338,6 +338,22 @@ class PlotCanvasView extends Renderer.View
...
+      sign = if reversed then -1 else 1
+
+      if min_interval > 0
+        if new_interval < min_interval
+          range_info['start'] = center - sign * 0.5 * min_interval
+          range_info['end'] = center + sign * 0.5 * min_interval
+      if max_interval > 0
+        if new_interval > max_interval
+          range_info['start'] = center - sign * 0.5 * max_interval
+          range_info['end'] = center + sign * 0.5 * max_interval
```
almarklein: @bryevdv I updated the variable name to `new_range`. Actually, when the interval is too big/small, it pins the range relative to the *current* center; there wont be any shifting. 

I don't think canceling the event is a good idea: imagine using the box zoom tool and zooming in too far: nothing would happen. Also scroll wheel is in steps so you would not reach the real minimum/maximum.

One thing that I am not sure about is that the constraint is per axis, so zooming can still take place in the other dimension unless it is constrained too. This is IMO desired behavior for the box zoom, but perhaps not for scroll zoom. Not sure what to do though. Triaging between these two cases seems awkward. Maybe this is acceptable as users would typically constrain both axis?

bryevdv: Setting to a smaller interval around the current center is not the correct behavior in general.  Consider a scroll wheel zoom with the mouse at the far right edge of the plot. The range should shrink largely from the left side only, keeping the right range point exactly as it is. That's not what this code would do tho, it would "jump" do that the current center is still the center and the far right edge (currently under the mouse) would suddenly be shifted out of the plot.

almarklein: Good point. I think it should be relatively easy to fix though.

What's your take on having the axis being constrained individually? If you're fine about that, I'll make above change and this should be about done.

bryevdv: I would personally refuse to update the range if *either* range constraint would be violated. But if you think otherwise perhaps it's worth a wider discussion.

birdsarah: > I would personally refuse to update the range if either range constraint would be violated.

I would suggest starting with constraining individually as that's always been our behavior to date. I think it would be nice if we could offer what you're suggesting as a convenience wrapper / extra mode.  I can see how why what your suggesting would be useful in most cases, but I think it's a really low-level place to enforce that and it might not suit some people. We already have where you can be over an axis and zoom independently. This feels like it might clash with that.

bokehjs/src/coffee/models/ranges/data_range1d.coffee
birdsarah: Given that bounds is `p.Any`, I think it's reasonable to have `p.Any` until we make a property that describes all range properties.

almarklein:
@bryevdv ok, I had a go at this. It was a bit of a rabitwhole, but I think I've got it working. I made it such that a weight factor can be calculated that expresses how much of the change in range (with respect to the current range) can be applied without violating the bounds/min_interval/max_interval. Then this weight can be either calculated and applied for each range individually, or for all ranges together.

This keeps the aspect ratio equal for scrolling, and as far as I can tell, all other behavior is the same (but could use some testing).

almarklein:
Wait. I am not happy with this. I wrote something down, then realized something that I need to try.

almarklein:
Ok, I think I've got it now. I had a proposal written down to slightly change how  `bounds` are enforced, so that it becomes untangled from the interval limits. Then I realized that by making the implicit max-interval imposed by the bounds explicit, the constraints can be applied independently, keeping the code relatively simple and the behavior straightforward. This is good for a round of review.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -304,46 +304,94 @@ class PlotCanvasView extends Renderer.View
...
+          if max?
+            if max <= range_info['end']
+              range_info['end'] = max
+              if is_panning? or is_scrolling?
+                range_info['start'] = max - new_interval
+
+      if rng.get('start') != range_info['start'] or rng.get('end') != range_info['end']
+        rng.have_updated_interactively = true
+        rng.set(range_info)
+        rng.get('callback')?.execute(rng)
```
almarklein: This particular function was changed to do the iteration inside the function instead of outside, adds the interval constraint, and changes how the `range_info['end']` is set when `range_info['start']` is constrained, so it works well for zoom operations. Other than that, I did not change this code.

bryevdv:
This PR still suffers from the not keeping the mouse centered on the data point during a scroll zoom. I hate to be so insistent but I am absolutely -1 on that. If then bounds will be exceeded we should simply cancel the current operation. Bounds are exactly that, bounds. There is no guarantee that they can be attained, only a guarantee that they may not be exceeded. 

Additionally, the behavior in the "multiple ranges" example is also bizarre, just beginning a scroll zoom scrunches the whole plot to the left, and then it's impossible to zoom back, have to reset.

almarklein:
> This PR still suffers from the not keeping the mouse centered on the data point during a scroll zoom.

If the zooming would be canceled if a bound is exceeded (which I tried, at some point), the user would have the zoom-out being stopped, but after panning a bit it would work again, which leads to a "jaggy" feeling IMO. Also, it would be nearly impossible to reach the maximum zoom level, because 1) scroll-zooming occurs in steps 2) you'd have to go at it from the exact center of the bounds.

I went with translating the view as necessary to honor the boundaries, which indeed breaks the mouse-centering. IMO this feels better though; the view sliding away is a gentle indication that a boundary is reached, while it still allows me to zoom out further (until the other bound kicks in and the zoom is at it absolute maximum).

I could put up an exported HTML doc with the two approaches side by side so we can gather some experiences and opinions.

almarklein:
> the behavior in the "multiple ranges" example is also bizarre,

This is also true for master. The current bounds are a bit awkward. Will fix this in this PR.

bryevdv:
> IMO this feels better though 

I guess we need to open up a wider discussion then. I feel very strongly in the opposite. Apart from breaking existing behavior, it is supremely annoying -- the *whole point* of a scroll zoom is that it zooms on the location you are centered on.

almarklein:
Bryan and I had a discussion about the behavior of the scroll-zoom tool when it encounters a bound. In short, a "gliding" approach allows one to zoom out, while the view glides away, but this would break the scroll-zoom's most important feature that it keeps its focus point on the same location. The two approaches can be seen [here for reference](https://dl.dropboxusercontent.com/u/1463853/bokeh/prevent_user_zooming_or_panning_too_far2.html). 

The current resolution is to introduce two policies to replace the "auto" bounds that we currently have. In "tight" mode the bounds will be as in the current "auto" mode, and in "loose" mode the bounds would be set 100% wider than necessary. This should avoid hitting any bounds in an annoying way in most cases altogether. This will be a new PR.

It would be relatively easy to enable the "gliding" approach, should we ever want to. If we do, it would likely be an optional feature. Since it's not exactly trivial to come up with an API for this, we leave this for when its needed (if ever).

This PR is ready from my end.

bokeh/models/ranges.py
```diff
@@ -133,6 +144,16 @@ class DataRange1d(DataRange):
     ``None`` e.g. ``DataRange1d(bounds=(None, 12))``
     """)
 
+    min_interval = Float(default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum visible interval. If set to ``None`` (defaut), the minimum
```
bryevdv: typo, missing `l` in `default`

bokeh/models/ranges.py
```diff
@@ -240,6 +261,16 @@ class FactorRange(Range):
         and the plot will not pan left of apples or right of bananas.
     """)
 
+    min_interval = Int(default=None, help="""
+    The level that the range is allowed to zoom in, expressed as the
+    minimum number of visible categories. If set to ``None`` (defaut),
```
bryevdv: `default` typo

bryevdv:
A few minor comments otherwise LGTM

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -246,46 +246,105 @@ class PlotCanvasView extends Renderer.View
...
+    if weight < 1
+      for [rng, range_info] in range_info_iter
+        range_info['start'] = weight * range_info['start'] + (1-weight) * rng.get('start')
+        range_info['end'] = weight * range_info['end'] + (1-weight) * rng.get('end')
+
+  _update_ranges_individually: (range_info_iter, is_panning, is_scrolling) ->
+    hit_bound = false
+    for [rng, range_info] in range_info_iter
+      # Is this a reversed range?
+      reversed = if rng.get('start') > rng.get('end') then true else false
```
bryevdv: I know this was copied from the original but can't this just be: `reversed = (rng.get('start') > rng.get('end'))` ?

almarklein:
Done.

cgrimal:
Hi guys, thanks for the great work! I'm really interested in using this feature (to constrain the pan tool especially) but I don't know how. Could you point me in the right direction? Also, in which Bokeh version will it be included?
Many thanks

almarklein:
@cgrimal It's in 0.12.0 and above (at the time of writing, the latest dev release is 0.12.1dev3).

Here is a minimal example:
```python
from bokeh.io import show

x_range = Range1d(0, 3, bounds=(-1, 3.5), min_interval=1.5)
y_range = Range1d(0, 3, bounds=(-0.5, 4), min_interval=1.5)

p = figure(tools='pan, box_zoom, wheel_zoom, reset',
           x_range=x_range, y_range=y_range)
p.line(x=[1,2,3,4,5], y=[1,4,3,5,2])

show(p)
```

cgrimal:
Thanks @almarklein ! Works perfectly :D
Have a good day.

Issue #1848, opened 2015-02-09
tools not working on Gallery - Windows 7 + Chrome 40.0.2214.111 m


timelyportfolio:
On my Windows 7 machine with Chrome 40.0.2214.111 m, the tools are not working.  They work fine on Firefox and even IE. 

![bokeh_tools](https://cloud.githubusercontent.com/assets/837910/6113085/3bece40c-b05a-11e4-902b-d2d7454aaa8b.gif)

Unfortunately, no errors in the console either.

bryevdv:
Just the resize and wheel zoom tool, or others as well?

timelyportfolio:
All tools except for pan/zoom do not work.  They did at some point in the past.  I discovered when I was testing the rbokeh implementation.  Tools do not work there in the RStudio Viewer (webkit based) either, but I considered this an edge case.

timelyportfolio:
Just wanted to update.  I went back to commit 34f85b32ca03c4b02a4b7385c32065e1b1173b51 and still have the issue.

bryevdv:
FYI @pzwang reported problems on OS X with version 40. However everything is working for me with Version 42.0.2298.0 dev (64-bit) (on OS X). Would it be possible to see if the problem persists with a newer Chrome?

timelyportfolio:
does not work (same issue) with Version 41.0.2272.43 beta-m (64-bit) and Bokeh 0.7.1

Since no console error, I can't determine what is happening.  Buttons seem to accept click but behavior is not changed.  Stays on pan.

Shift+scroll does not work either.

bryevdv:
@timelyportfolio right, no more need for `shift+scroll` the wheel zoom and pan can coexist and be active together now without interference. I'll have to fire up a VM to look at this, will probably be this afternoon sometime.

scizmeli:
Hello, any advances on this bug?

mridsole:
Also getting this problem with Chrome 44.0.2403.130 m

bryevdv:
Question we should have asked earlier: are those experiencing this issue using touchscreen enabled laptops that can have both mouse and touch events?

bryevdv:
ref: #3905

dhirschfeld:
I can confirm that [disabling touch](http://bokeh.pydata.org/en/0.10.0/docs/faq.html#are-the-tools-not-working-in-chrome-cromium) makes the tools work.
[...]

Issue #4352, opened 2016-05-23
Add a test for the toolbar click regression 


birdsarah:
See #1848 and its fix #4353 

There's already a test that detects whether buttons can be selected or deselected. But to test the regression we'l'l need to have that test not only run against standard firefox, as all our tests do now, but against a browser with the configuration of both touch and mouse functionality - which is where users are having problems.

bryevdv:
closing as stale

Issue #2997, opened 2015-10-15
Touch problems on all but the simplest cases


birdsarah:
Touch support was first introduced in 0.7.0, it worked for simple cases, but appears to have not worked for other cases even back then.

Since then, touch support has degraded further and even some simple cases don't work. I am working on a separate PR to get touch support back to the levels described here, but there's still more work to be done as the performance described here is not sufficient.

For reference see the following html and notebook (https://github.com/birdsarah/bokeh-miscellany/tree/master/touch)
* a standalone plot - https://rawgit.com/birdsarah/bokeh-miscellany/master/touch/scatter-070-single.html
* a set of vplots - https://rawgit.com/birdsarah/bokeh-miscellany/master/touch/scatter-070-vplot.html
* a set of hplots - https://rawgit.com/birdsarah/bokeh-miscellany/master/touch/scatter-070-hplot.html
* a notebook of plots - http://nbviewer.ipython.org/github/birdsarah/bokeh-miscellany/blob/master/touch/scatter-070.ipynb

You may have to load them locally to see the bounding box clearly. I have found the nbviewer link works best and highlights all the issues

Behavior observed (Note this is only for touch - mouse events are working just fine):

Single plot:

* at default page zoom:
 * pan tool - works as expected
 * box zoom and select - basically work as expected although there's a subtle offset
 * wheel zoom - works as expected - can be a little fidgety when the box tools are also on
* when page (not plot) is zoomed in:
 * pan tool - works as expected
 * box zoom and select - don't work properly - there's a large offset between where u select and where the box is
 * wheel zoom - sort of works, but it's clearly centered at an offset 

HBox & VBox

Generally, the first plot works, but the other plots - to the right or down the page - don't work - the bounding box is no-where near expected - similar to zoomed in on a single plot.

Specific observations:
* pan tool - on last VBox plot (4th in 4-plot VBox) pan is constrained in the y direction 
* pan tool - on last HBox plot (4th in 4-plot HBox) pan is constrained in the x direction
* pan tool - doesn't always behave as expected if the full plot isn't on the page
* box select & zoom - first plot behaves, others have a large offset
* wheel - just not quite right on everything but the first plot - a little hard to characterize

All of the above are made worse when zoomed into page.

Note: Although HBox and VBox are no longer in Bokeh, similar offset behavior has been observed with latest vplot and hplot

Notebooks: 
similar problems to above - offsets on bounding boxes, pan not working if plot not fully in view, etc.

[...]

PR #4353, opened 2016-05-23
Fix toolbar for chrome

birdsarah:
Fixes: #1848 

This is effectively a revert of commit 9e57a12

Couldn't do an actual clean revert due to file having been subsequently moved.

Ping @bryevdv - the original commit is yours. I have done a quick test and I found that, on my phone, I could still click toolbar buttons - but this was not a thorough test. 

We have lots of woes in touch support land (#2997) so I would propose putting this change back in to get desktop users working again and then we can focus on touch later.

I haven't added a test, as I have a lot going on at the moment, and it would be a somewhat involved selenium tests to detect the regression. There's already a test that detects whether buttons can be selected or deselected. But to test the regression we'l'l need to have that test not only run against standard firefox, as all our tests do now, but against a browser with the joint touch functionality. I've added an issue: #4352

### DIFF
--- a/bokehjs/src/coffee/models/tools/button_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/button_tool.coffee
@@ -10,11 +10,7 @@ class ButtonToolButtonView extends Backbone.View
   template: button_tool_template
 
   events: () ->
-    # TODO (bev) this seems to work OK but maybe there is a better way
-    if 'ontouchstart' of document
-      return { 'touchstart .bk-toolbar-button': '_clicked' }
-    else
-      return { 'click .bk-toolbar-button': '_clicked' }
+    return { 'click .bk-toolbar-button': '_clicked' }
 
   initialize: (options) ->
     super(options)


bryevdv:
I wonder if we could just attach both event handlers unconditionally

birdsarah:
I feel like there are two things here.
1) Fixing a bug that comes up reasonably often.
2) Handling touch properly. 

The first one is quick, the second one isn't.

bryevdv:
Yes but it we can fix the bug and also keep the current level of touch functionality (such as it is) we should prefer to do that.

birdsarah:
> Yes but it we can fix the bug and also keep the current level of touch functionality (such as it is) we should prefer to do that. 

That's my point - I don't believe we have a current description of what toolbar support is for touch devices, or was historically and we currently expect it to be. If we did, then sure, it would just be a matter of testing it against this PR.

bryevdv:
All of the things on these vines have generally always worked for me: https://vine.co/bokehplots

So, pan, pinch zoom, lasso and box select

birdsarah:
**This doesn't change the touch events  that interact with the plot.** HammerJS handles that just fine. 

This fix is only for activating and deactivating the toolbar buttons. It fixes this problem: "Are the tools not working in Chrome/Cromium?" from http://bokehplots.com/pages/faqs.html

When I tested the fix with my phone, I was still able to select toolbar buttons on my device.

We should probably use a hammer event to activate those buttons as then we could feel more confident that they were working for everyone, but this gets us working for those desktop users.

birdsarah:
(and for what its worth don't forget that i'm almost certain that they only ever worked for a single plot on a page: https://github.com/bokeh/bokeh/issues/2997)

bryevdv:
OK let's merge then.

PR #4360, opened 2016-05-24
Merges the hipdi work into the layout PR

almarklein:
Works!


almarklein:
@birdsarah

birdsarah:
@almarklein - there are two defaults files from the bokehjs tests that have been accidentally added to this PR. Suggest squashing after removing so we don't end up with a large addition and removal.

birdsarah:
Fixed it.

almarklein:
Sorry about that! Thanks for the fix.

birdsarah:
Sorry for the delay @almarklein.

PR #4378, opened 2016-05-26
Adjust hidpi on browser zoom (against layout branch)

### LINKED ISSUE #4345: When using the browser zoom the plot content becomes blurry

### LINKED ISSUE #4367: webgl issue on new layout with multiple plots - some plots are blank.
See the following two screengrabs, the only difference between them is the use of webgl=True.

Here is code to reproduce.

```python
import numpy as np

from bokeh.models import GridPlot, ToolbarBox, Column, Row

N = 400
colors = ["#%02x%02x%02x" % (int(r), int(g), 150) for r, g in zip(50 + 2 * x, 30 + 2 * y)]

TOOLS = "crosshair,pan,wheel_zoom,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"

def p():
    p = figure(tools=TOOLS, webgl=True)
    p.scatter(x, y, radius=radii, fill_color=colors, fill_alpha=0.6, line_color=None)
    return p

layout = GridPlot(
    children=[
        [p(), p(), p(), p()],
        [p(), p(), p(), p()],
        [p(), p(), p(), p()],
        [p(), p(), p(), p()],
    ],
    toolbar_location='above',
    responsive='box'
)

show(layout)
```
The same issue exists for a smaller grid of say just 4 plots.

`webgl=True`
==========

screen shot 2016-05-25 at 11 15 43 am

`webgl=False`
==========

screen shot 2016-05-25 at 11 16 04 am

almarklein:
Fixes #4345 and fixes #4367.

This makes the hidpi pixel ratio adapt when the browser zoom changes. This turned out to be easier than the original PR that I made against master earlier (#4346), because we already get an update now when the browser zoom changes.

I tested that all the HIDPI stuff works in embedded plots as well. This means that the document does an update when it gets a `window.resize` event, regardless whether the size did not actually change, and regardless whether the plot is embedded or not. In case someone ever tries to add efficiency here, please keep in mind this use-case.

cc @birdsarah

### DIFF
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -45,14 +45,16 @@ class CanvasView extends BokehView
     ctx = canvas_el[0].getContext('2d')
     return ctx
 
-  render: (force=false) ->
+  prepare_canvas: (force=false) ->
+    # Ensure canvas has the correct size, taking HIDPI into account
+    # returns true if the canvas was updated
 
     width = @model._width._value
     height = @model._height._value
+    dpr = window.devicePixelRatio
 
     # only render the canvas when the canvas dimensions change unless force==true
-    if not _.isEqual(@last_dims, [width, height]) or force
-
+    if not _.isEqual(@last_dims, [width, height, dpr]) or force
 
       @$el.css({
         width: width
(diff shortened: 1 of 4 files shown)


almarklein:
This did not make it into Sarah's branch. Rebased it against master, see #4405.

PR #4405, opened 2016-05-28
Adjust hidpi on browser zoom

### LINKED ISSUE #4345: When using the browser zoom the plot content becomes blurry

### LINKED ISSUE #4367: webgl issue on new layout with multiple plots - some plots are blank.

almarklein:
Supersedes #4378. Fixes #4345 and fixes #4367.

This makes the hidpi pixel ratio adapt when the browser zoom changes. This turned out to be easier than the original PR that I made against master earlier (#4346), because we already get an update now when the browser zoom changes.

I tested that all the HIDPI stuff works in embedded plots as well. This means that the document does an update when it gets a `window.resize` event, regardless whether the size did not actually change, and regardless whether the plot is embedded or not. In case someone ever tries to add efficiency here, please keep in mind this use-case.

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -404,7 +404,10 @@ class PlotCanvasView extends Renderer.View
 
     ctx = @canvas_view.ctx
 
-    # Get hidpi ratio
+    # Prepare the canvas and get pixel ratio.
+    # Note that this may cause a resize of the canvas, which means that
+    # this should be considered the main rendering entry point; any previous
+    # calls to ctx.save() may be undone (due to the canvas resize).
     @canvas_view.prepare_canvas(force_canvas)
     ctx.pixel_ratio = ratio = @canvas_view.pixel_ratio  # Also store on cts for WebGL
 
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -405,8 +405,8 @@ class PlotCanvasView extends Renderer.View
     ctx = @canvas_view.ctx
 
     # Get hidpi ratio
-    ratio = @canvas_view.render(force_canvas)
-    ctx.pixel_ratio = ratio  # we need this in WebGL
+    @canvas_view.prepare_canvas(force_canvas)
+    ctx.pixel_ratio = ratio = @canvas_view.pixel_ratio  # Also store on cts for WebGL
 
     # Set hidpi-transform
     ctx.save()  # Save default state, do *after* getting ratio, cause setting canvas.width resets transforms
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -47,13 +47,12 @@ class CanvasView extends BokehView
 
   prepare_canvas: (force=false) ->
     # Ensure canvas has the correct size, taking HIDPI into account
-    # returns true if the canvas was updated
 
     width = @model._width._value
     height = @model._height._value
     dpr = window.devicePixelRatio
 
-    # only render the canvas when the canvas dimensions change unless force==true
+    # only resize the canvas when the canvas dimensions change unless force==true
     if not _.isEqual(@last_dims, [width, height, dpr]) or force
 
       @$el.css({
@@ -61,7 +60,7 @@ class CanvasView extends BokehView
         height:height
       })
 
-      # Scale the canvas
+      # Scale the canvas (this resets the context's state)
       @pixel_ratio = ratio = get_scale_ratio(@ctx, @mget('use_hidpi'))
       canvas_el = @$('.bk-canvas')
       canvas_el.css({
@@ -73,7 +72,6 @@ class CanvasView extends BokehView
 
       logger.debug("Rendering CanvasView [force=#{force}] with width: #{width}, height: #{height}, ratio: #{ratio}")
       @last_dims = [width, height, dpr]
-      return true
 
   set_dims: (dims, trigger=true) ->
     @requested_width = dims[0]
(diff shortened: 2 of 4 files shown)


bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -402,7 +402,12 @@ class PlotCanvasView extends Renderer.View
     # reach the frame in time (sometimes) so force an update here for now
     @model.get('frame')._update_mappers()
 
+    # Get context and restore to its initial state. The restoring
+    # Would stricly speaking not be necessary, but in case there is an error
+    # during drawing, restore() won't be called enough. This guards for that.
     ctx = @canvas_view.ctx
+    for i in [0..3]
+      ctx.restore()
```
bryevdv: I'm a little confused by this, can you elaborate more? This seems like it will unconditionally restore, regardless of whether there was an error or not. If not, in which case it will unexpectedly blow away any context `push` states? Or are you assuming / relying on the context state to otherwise be "zeroed" before getting to the code that might error out?

almarklein: The idea is that this will ensure that the context's state stack is empty. So that we have control (in each iteration) of the exact transformations, even if there where errors in the last draw iteration.

The reason that we now do the `scale(ratio, ratio)` into the draw iteration instead of only when the canvas is resized, it because when blitting the WebGL canvas, we need an unscaled canvas (this was causing the blurring). However, doing the scaling on each iteration introduced this vulnerability of blowing everything up when there is an error during the draw iteration. This fixes that vulnerability.

bryevdv: I understand the idea,  my question is this: how are we assured that previous intentional ctx `save` operations that might have happened prior to / outside this area of code aren't blown away. Let's say that code that calls this render does a `ctx.save`, and then calls this render, and there are no errors. Then this code calls `ctx.restore` three times, unconditionally? I don't see how that earlier `ctx.save` does not get blown away, even though it should not be.

bryevdv: If there are errors, do they raise exceptions? Then `try` / `catch` would seem to be a better option to keep `save` / `restore` pairs happening "in the same place"

almarklein: Right. Now I understand your concern. However, I think there has always been an implicit assumption that `plot.render()` was the "rendering entry point" and that there were no calls to `save()` before it, because `plot.render()` calls `@canvas_view.prepare_canvas()` (previously `@view_canvas.render()`) which may resize the canvas. And when a canvas is resized, the state of a context is reset. So we probably should add a comment about this assumption.

An alternative might be to always set the size of the canvas in `@canvas_view.prepare_canvas()` but I read somewhere that this might be unreliable in some browsers if the size is not changed. This silly for-loop is my version of `ctx.reset()`. I agree it does not look pretty, but it seemed the easiest way to get the job done, without having to change any other code.

Using `try` / `catch` could work, though I don't know what that would do to tracebacks, since there is no `finally` (right?).

bryevdv: AFAIK JS does support `finally`: http://www.w3schools.com/jsref/jsref_try_catch.asp So that would be my suggestion to try

almarklein:
Fair enough. `try-finally`'s added.

bryevdv:
OK, well this could be better, but is it really the case that all the places you added `try` blocks can fail? My assumption was that it was only webgl related paths that might have issues, and so we'd only need `try` blocks selectively around those areas. For instance, can the `Grid` render fail in the presence of webgl? If not, it already lines up `save`/`restore`

almarklein:
I think probably most glyphs can potentially fail during rendering. This lining up of `save` and `restore` is not a WebGL specific issue. Though it's always hard to tell when JS fails with an error or breaks something and continues ...

Personally, I prefer the "reset" approach over all these try-finallies, for one because it also works in case of any new glyphs being added that might do `ctx.save()`. Though a better place for the "reset" would be the `prepare_canvas()` method (which already does a reset in case it resizes the canvas).

bryevdv:
I've never once encountered an HTML5 canvas rendering operation fail. That's why I assumed it was webgl specific. Is there some what to reproduce or demonstrate the failure condition?

almarklein:
Maybe I am just overly careful, but the plot blowing up looks really bad. It could well be that all `ctx` methods are safe, but there might be an error somewhere *during* the drawing algorithm, like trying to access an attribute of a nonexisting atribute. I tested by just adding `throw "foo"` in a render method.

bryevdv:
OK I'm still not sure I understand the concern. Things like undefined attribute access are *bugs in our code* and we need to fix them, not paper over them. If that's the intent of this, I am -1 on it. I had thought there was something intrinsically unstable about webgl on some platforms that we simply had no better way to protect against at all. Since this particular aspect requires more discussion, I'd suggest this: a AP that *only* has the basic webgl fixes/improvements, and a separate issue to discuss when or where this sort of exception handling is appropriate.

almarklein:
Ok, got rid of the context reset stuff and the try-finallies. Will create an issue to explain my case more clearly when this has merged.

PR #4409, opened 2016-05-31
Use correct "property names"

### LINKED ISSUE #4356: Line hover broken in master
To reproduce run examples/howto/us_marriages_divorces.py on master.

screen shot 2016-05-23 at 2 18 52 am

Let's definitely write a selenium test for this. There's already a hover test can use as an example. We don't need the selenium test to be a screenshot test as we can just check for the presence of the hover tooltip in the DOM with the correctly populated data.

bryevdv:
issues: fixes #4356 

Needed update glyph.x -> glyph._x after properties PR


bryevdv:
Still trying to get last spurious test fail to pass. When/how did `pandas_dataframe.py` get added back?

mattpap:
> Still trying to get last spurious test fail to pass.

Tests pass or am I missing something?

bryevdv:
It is now, finally.

PR #4410, opened 2016-05-31
update layout mode names

### LINKED ISSUE #4396: API - Name the responsive modes.
There are currently 5 modes: `fixed`, `box`, `width_ar`, `height_ar`, `box_ar`

@havocp @almarklein @bryevdv have all mentioned concerns with these being non-obvious.

To recap what these modes do:
* fixed - the plot is just the size of the specified width & height
* box - the plot fills the available size in it's box - causing the plot to stretch & squeeze however
*  width_ar - the plot fills the available width but maintains the aspect ratio (calculated from the specified width and height)
* height_ar - the plot fills the available height - but maintains the aspect ratio (calculated from the specified width and height)
* box_ar - the plot fills the available height or width while maintaining the aspect ratio (calculated from the specified width and height)

I would greatly appreciate naming suggestions from @bokeh/dev.

Secondly these modes are passed into the `responsive` parameter of any Layout object. Is that the right parameter name. In 0.11 we have `responsive=True/False` on Plot. I have kept and expanded this, but we don't have to keep this. We could have, for example, `mode` on the layout objects Row, Column. I think responsive is a good parameter, but just wanted to bring it up.

bryevdv:
issues: fixes #4396

changes:

box_ar    -> scale
width_ar  -> width_scale
height_ar -> height_scale

### DIFF
--- a/bokeh/charts/tests/test_chart.py
+++ b/bokeh/charts/tests/test_chart.py
@@ -53,7 +53,7 @@ class TestChart(unittest.TestCase):
         self.assertEqual(self.chart.title.text, "new_title")
 
     def test_responsive(self):
-        self.assertEqual(self.chart.responsive, 'width_ar')
+        self.assertEqual(self.chart.responsive, 'width_scale')
 
     def check_chart_elements(self, expected_tools):
         self.assertIsInstance(self.chart.left[0], LinearAxis)
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@ -180,4 +180,4 @@ JitterRandomDistribution = enumeration("uniform", "normal")
 SortDirection = enumeration("ascending", "descending")
 
 #: Responsive types
-Responsive = enumeration("box", "width_ar", "height_ar", "box_ar", "fixed")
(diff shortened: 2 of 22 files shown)


bryevdv:
@bokeh/dev there was also a suggestion to replace `box` with `stretch` I'm slightly in favor of this, but it's also not a change that can be automatically done easily, so I think I'm ok with keeping `box` unless there are strong objections.

almarklein:
I'd prefer "stretch" as it conveys the idea pretty accurately, whereas "box" is more abstract.

mattpap:
>  there was also a suggestion to replace `box` with `stretch`

That's a good idea. `box` seems too abstract in this context.

bryevdv:
OK, it's a pain, but I will do it. First let's make sure we have agreement, I think @birdsarah still wanted to express some thoughts.

birdsarah:
As noted on the issue, I'm not happy with the names currently proposed, and am trying to consolidate my thoughts. 

Of all the bits of clean-up that need to be done, this seems like one not to rush.

bryevdv:
There's conflicts now, I will close this PR and open a new one. I will summarize my understanding of the current proposal in the issue.

PR #4417, opened 2016-06-01
Put webgl examples in own directory

### LINKED ISSUE #4279: Tests for WebGL
Our WebGL functionality does not really have tests and I have been fixing it up after refactorings in bokehjs a few times now. I'd love to have a few tests for it to avoid this in the future.

* Can we run tests on selenium/saucelabs with webgl enabled?
* Do we actually test for image similarity yet?

cc @birdsarah

almarklein:
Closes #4279


bryevdv:
@almarklein I made a comment elsewhere but it might have been missed. We already have an "interactive tester" that runs examples and collects comments as you go so that all of the comments are available at the end. Please add a webgl category to this existing script, instead of a new, separate script. 

https://github.com/bokeh/bokeh/blob/master/scripts/interactive_tester.py

bryevdv:
Right now IIRC that script maps categories to directories, so we can either collect all the webgl examples in one directory (maybe not a bad idea) or the script can be gently extended to also map names to explicit lists of examples.

bryevdv:
Isn't all the browser stuff already handled by 

https://github.com/bokeh/bokeh/blob/master/bokeh/util/browser.py

and setting `BOKEH_BROWSER` ? That is what I use to quickly run the interactive tester with different browsers.

almarklein:
@bryevdv I looked at the interactive tester but I understood it as running *all* examples, which seemed too much for the webgl-specific testing. Which is why I made the script. I guess I did not look close enough to see that it runs one category at a time.

I now put the webgl specific examples in their own directory (`examples/plotting/webgl`) and added that category in the interactive tester. Having the examples together in this way already helps a lot to do a quick webgl test.

> Isn't all the browser stuff already handled by  [...] setting `BOKEH_BROWSER` ?

Unfortunately, on Windows that does not work at all. But I guess its ok to just copy the url manually in case multiple browsers need testing.

(rebased this branch to remove merge conflict and to pop out that script I added and removed)

bryevdv:
What about just `examples/webgl` ? I know all the example happen to be `plotting` ones, but putting it in it's own directory at the top would make it so other examples are not precluded. Also might make it easier for people to find.

scripts/interactive_tester.py
```diff
@@ -16,6 +16,7 @@
     'plotting-file'    : '../../examples/plotting/file',
     'plotting-notebook': '../../examples/plotting/notebook',
     'server'           : '../../examples/plotting/server',
+    'webgl'           : '../../examples/plotting/webgl',
```
bryevdv: being picky :)  alignment is one space off

almarklein:
Moved one level up

bryevdv:
@almarklein I am fine to merge this as-is, but if you'd like, now that the examples are all in one directory, a `README.md` would be nice. It could be a simple text table, or have screenshots like, e.g. https://github.com/bokeh/bokeh/tree/master/examples/app  If you'd like to defer that until later, then please feel free to merge this PR whenever.

almarklein:
Good point. I added a readme and added a mention of the webgl in the readme of the root examples dir.

PR #4243, opened 2016-04-28
Integrate ComputedTransforms into the DataSpec System

justacec:
This PR will integrate the new Computed Transforms into the DataSpec system.  This will allow the user to apply transforms to the data provided to various plotting functions.

As an example usage, the following line will apply a random jitter to the x coordinate of a set of circles

    bk.circle(x={'field': 'xvalues', 'transform': Jitter(width=0.3)}, y='yvalues', source=mydata)

- [ ] issues: fixes #4205 
- [ ] tests added / passed
- [ ] release document entry (if new feature or API change)

This is no where near ready but I wanted to put the PR in to begin the necessary work.

PR #4420, opened 2016-06-01
Initial work to integrate the new ComputedTransforms into the DataSpe…

### LINKED ISSUE #4205: Add the ComputedTransforms to the DataSpec System
Currently, to enable a Transform to be performed, the user must actively click on a control to force the calculation.  Integrating this into the DataSpec system will allow for the transformation to be performed on the fly without the need for a user to force the calculation.

justacec:
The following is a duplicate message from closed PR: #4243 (Closed and reopened to get the PR correct)

This PR will integrate the new Computed Transforms into the DataSpec system.  This will allow the user to apply transforms to the data provided to various plotting functions.

As an example usage, the following line will apply a random jitter to the x coordinate of a set of circles

    bk.circle(x={'field': 'xvalues', 'transform': Jitter(width=0.3)}, y='yvalues', source=mydata)

- [x] issues: fixes #4205 
- [x] tests added / passed

This is no where near ready but I wanted to put the PR in to begin the necessary work.


bokeh/core/properties.py
```diff
@@ -1600,7 +1600,7 @@ def set_from_json(self, obj, json, models=None):
 
 class DataSpec(Either):
     def __init__(self, typ, default, help=None):
-        super(DataSpec, self).__init__(String, Dict(String, Either(String, typ)), typ, default=default, help=help)
+        super(DataSpec, self).__init__(String, Dict(String, Either(String, Instance(HasProps), typ)), typ, default=default, help=help)
```
justacec: @bryevdv You had suggested that this could be made more specific.  Specifically, `Instance(Transform)`.  I had wanted to do that as well but I ran into a error when I executed the code.  The error was: 

    NameError: name 'Transform' is not defined

I believe this is because of a circular import.

bryevdv:
@justace yes probably, in this case you can pass a string name to `Instance` too, see e.g.

https://github.com/bokeh/bokeh/blob/master/bokeh/models/layouts.py#L112

bokeh/core/properties.py
bryevdv: @justace yes probably, in this case you can pass a string name to `Instance` too, see e.g.

https://github.com/bokeh/bokeh/blob/master/bokeh/models/layouts.py#L112

bryevdv:
@bokeh/dev any thoughts? I'd like to merge this for `0.12` There is certainly more ease/convenience that can be added, but I think this is useful and valuable enough that it should go in now. It has JS tests for the js transform functionality, and for reference, this code:
```

import numpy as np

from bokeh.models import ColumnDataSource, Jitter

p = figure(plot_width=500, plot_height=400, x_range=(0,3), y_range=(0,10))

y1 = np.random.random(2500) * 10
y2 = np.random.normal(size=2500)*2 + 5

p.circle(x={'value': 1, 'transform': Jitter(width=0.4)}, y=y1,
         color="navy", alpha=0.3)

p.circle(x={'value': 2, 'transform': Jitter(width=0.4)}, y=y2,
         color="firebrick", alpha=0.3)

output_file("jitter.html")

show(p)
```
Generates this output:
screen shot 2016-06-11 at 7 25 54 pm

PR #4422, opened 2016-06-02
Updated package requirements for building docs

### LINKED ISSUE #4020: Problem with sphinx.ext.napoleon when building docs
I'm trying to build the docs for the first time right now, and I receive an 

    Extension error:
    Could not import extension sphinx.ext.napoleon (exception: No module named 'sphinx.ext.napoleon')

when doing `make all` in the sphinx directory. Could this be related to the discussion [here](https://github.com/bokeh/bokeh/issues/2984)? I'm quite new, have never worked with sphinx before, and am a bit unsure about how to proceed. Sphinx version is 1.2.3.

gonike:
Fixes #4020, among others.
Removed `sphinxcontrib-httpdomain`, `ggplot` and `seaborn` from the list of packages required to build the docs. Added `flask` (which is required to run the docserver). I wasn't exactly sure about `ggplot` and `seaborn`, but I could successfully build the docs without them and had the impression that everything was where it belonged. Are these still required anywhere?

### DIFF
--- a/sphinx/source/docs/dev_guide/documentation.rst
+++ b/sphinx/source/docs/dev_guide/documentation.rst
@@ -7,16 +7,14 @@ Requirements
 ------------
 
 We use Sphinx_ to generate our HTML documentation. You will need the following
-packages installed in order to build Bokeh documentation:
+packages installed in order to build and view the Bokeh documentation:
 
 * docutils
 * sphinx
-* sphinxcontrib-httpdomain
-* seaborn
 * pygments
 * yaml
 * pyyaml
-* ggplot
+* flask
 
 These can be installed using ``conda`` or ``pip`` or from source. In
 addition to the package requirements, you will also need to have the sample


bryevdv:
HI @gonike, technically speaking `Flask` is not required, anyone could also just run e.g. `python -m http.serve` on the docs dir. I'm not opposed to adding it though, since `make serve` is probably the easiest thing for everyone to do. Also ggplot and seaborn are needed for a couple of the example in the gallery.

gonike:
Hey @bryevdv, would you be fine with me adding a corresponding note (`for building some examples in the gallery`, `for conveniently viewing the generated html files`) to ` ggplot`, `seaborn`, and `flask` then? (Considering that they are not strictly required to build the docs.)

Regarding the build failure I'm not sure if that would be a good idea. Right now I can build the docs from current dev version without having built the library itself (instead using the conda-provided lib version), which results in some plots not appearing. If these would make the build fail, I'd have to build the library in order to contribute any small change in the docs.

However, I have the impression that at the moment I don't get any notification of failed plots at all during the build process! I checked all the messages, and everything I get is a bunch of warnings due to "duplicate labels". (And some other warnings, but they don't seem to be related to the plots missing due to novel features.)

bryevdv:
@gonike A note would also be perfectly acceptable. Regarding "failing" I'm really referring to our build machinery. That could be a non-zero return status at the end, it doesn't mean we have to stop the docs build immediately on any problem. If the failures are not being logged that is an issue, though. Maybe i forgot to add logging to the `bokeh-plot` directive, but it would be a good idea for it to at least warn on plot render exceptions.

bryevdv:
Hi @gonike I am going to go ahead and close this PR, another PR to update these docs a bit more extensively came out of the Pycon marathon of sprinting. Thanks for this PR let us know if theres anything else you'd like to work on that we can help with!

PR #4423, opened 2016-06-02
Changed docserver shutdown instruction.

gonike:
As commented on my previous PR #4033, changed message during docserver startup from 

    Press any key to exit...

to

    Press <ENTER> to exit...


bryevdv:
Thanks for the catch!

bryevdv:
OK not sure what happened with the first test run, things appear to be going fine now. One test is expected to fail since you are not in @bokeh/dev but I will merge once everything else is green.

PR #4430, opened 2016-06-03
Initial fix for touch events with multiple plots

Zyell:
Fixes Issue #2997 

only tested on desktop with dual touch mouse

### DIFF
--- a/bokehjs/src/coffee/common/ui_events.coffee
+++ b/bokehjs/src/coffee/common/ui_events.coffee
@@ -108,8 +108,8 @@ class UIEvents extends Backbone.Model
       x = e.srcEvent.pageX
       y = e.srcEvent.pageY
     else
-      x = e.center.x
-      y = e.center.y
+      x = e.pointers[0].pageX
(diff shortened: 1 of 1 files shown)


birdsarah:
LGTM.

Michael and I tested this on his touchscreen laptop, we were able to reproduce the before (described in 2997) and fix it.

I then tested the fix on:
* my old ipod
* a new samsung phone (android)
* a new android phone

My old ipod exhibited the problem and then with the fix was resolved.

Interestingly, the samsung and android did not suffer from the original problem, but the fix did not degrade or alter performance as far as I could tell.

I had a look at hammer.js and see that pointers is something that comes from hammer, and it looks like we can rely on: http://hammerjs.github.io/api/#event-object

Although we should probably add a test for this, we're not yet set up for automated mobile testing so I think that putting this in is an improvement, even if we could do better in the future.

bryevdv:
LGTM

bryevdv:
OK, merging. Thanks @Zyell !

PR #4434, opened 2016-06-03
fix broken defaults test and fix defaults mismatch

### LINKED ISSUE #4432: JS Property mismatches
* JS Plot object has leftover text mixin for title. This is causing "invalid attribute"  messages being sent to the server:

> 2016-06-02 14:58:37,499 JSON had attr 'title_text_alpha'  on obj <bokeh.plotting.figure.Figure object at 0x10d09c080>, which is a client-only or invalid attribute that shouldn't have been sent

* unit test for properties is somehow not catching these and should be fixed.

bryevdv:
issues: fixes #4432

This PR fixes a really dumb error in `defaults.coffee` that caused the property matching test to always pass. As a result, ~50 mismatches accrued. I have fixes the test, and all the mismatches. Some review of a few of the cases will be helpful.

ping @canavandl please check out the fixes for `Title` for now I special cased some things in the test but would l like to improve that if possible

ping @birdsarah @mattpap


sphinx/source/docs/user_guide/source_examples/styling_arrow_annotations.py
```diff
@@ -13,6 +13,6 @@
 
 for i, style in enumerate(ARROW_HEADS):
     p.add_layout(Arrow(x_start=0.2, y_start=i, x_end=0.2, y_end=i, end=style()))
-    p.add_layout(Label(x=0.2, x_offset=20, y=i, text=[style.__name__], text_baseline='middle', text_align='left'))
+    p.add_layout(Label(x=0.2, x_offset=20, y=i, text=style.__name__, text_baseline='middle', text_align='left'))
```
bryevdv: I also added this small docs example fix. Arrow `text` is no longer a `DataSpec`

bokehjs/test/common/defaults.coffee
```diff
@@ -62,9 +62,17 @@ check_matching_defaults = (name, python_defaults, coffee_defaults) ->
       continue
 
     # special case for date time tickers, class hierarchy and attributes are handled differently
-    if name = "DatetimeTicker" and k = "tickers"
+    if name == "DatetimeTicker" and k == "tickers"
```
bryevdv: I think I owe everyone a beer for this one...

bokehjs/test/common/defaults.coffee
```diff
@@ -62,9 +62,17 @@ check_matching_defaults = (name, python_defaults, coffee_defaults) ->
...
       continue
 
+    # special case for Title derived text properties
+    if name == "Title" and (k == "text_align" or k == "text_baseline")
+      continue
```
bryevdv: ping @canavandl this is what I had to add to get `Title` matching, but maybe there is a better way.

bokehjs/test/common/defaults.coffee
```diff
@@ -62,9 +62,17 @@ check_matching_defaults = (name, python_defaults, coffee_defaults) ->
...
+
+    # special case for selections that have a method added to them
+    if k == 'selected'
+      v['0d'] = _.omit(v['0d'], 'get_view')
+
```
bryevdv: ping @almarklein if there is something better we can do, please let me know. Without this, the `get_view` key shows up in JS but not python, so there is a mismatch and error

bokehjs/src/coffee/models/renderers/renderer.coffee
```diff
@@ -286,19 +286,6 @@ class RendererView extends BokehView
...
-    _coords = this.prototype._coords.concat(coords)
-    this.prototype._coords = _coords
-
-    result = {}
-    for [x, y] in coords
-      result[x] = [ p.NumberSpec ]
-      result[y] = [ p.NumberSpec ]
-
-    @define(result)
-
```
bryevdv: @mattpap I moved this to be able to bring the class hierarchy clkoser into accord. `Glyph` shloud inherit from `Model` (not `Renderer`) and this change allows that to happen on the JS side.

bokehjs/src/coffee/model.coffee
```diff
@@ -5,6 +5,19 @@ p = require "./core/properties"
...
+    _coords = this.prototype._coords.concat(coords)
+    this.prototype._coords = _coords
+
+    result = {}
+    for [x, y] in coords
+      result[x] = [ p.NumberSpec ]
+      result[y] = [ p.NumberSpec ]
+
+    @define(result)
+
```
bryevdv: see note below on `renderer.coffee`

bokehjs/src/coffee/models/glyphs/glyph.coffee
```diff
@@ -166,7 +166,7 @@ class GlyphView extends Renderer.View
       @visuals.line.set_vectorize(ctx, reference_point)
       ctx.stroke()
 
-class Glyph extends Renderer.Model
+class Glyph extends Model
```
bryevdv: This makes things match the python side

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -261,14 +261,17 @@ class Plot extends LayoutDOM.Model
       webgl:             [ p.Bool,     false                  ]
       hidpi:             [ p.Bool,     true                   ]
 
-      min_border:        [ p.Number,   null                   ]
+      min_border:        [ p.Number,   5                      ]
```
bryevdv: @birdsarah just checking this seems ok?

bokeh/models/annotations.py
```diff
@@ -114,11 +119,17 @@ class Legend(Annotation):
         Dict(String, List(Instance(GlyphRenderer))), lambda d: list(d.items())
     )
 
+def _DEDAULT_ARROW():
```
canavandl: spelling

bokeh/models/annotations.py
```diff
@@ -149,15 +160,15 @@ class Arrow(Annotation):
     space" units by default.
     """)
 
-    end = Instance(ArrowHead, default=lambda: OpenHead(), help="""
+    end = Instance(ArrowHead, default=_DEDAULT_ARROW, help="""
```
canavandl: spelling

bryevdv:
remarking WIP, just need to fix up some unit tests. Comments already welcome, however.

bokeh/models/annotations.py
```diff
@@ -149,15 +159,15 @@ class Arrow(Annotation):
     space" units by default.
     """)
 
-    end = Instance(ArrowHead, default=lambda: OpenHead(), help="""
+    end = Instance('.models.arrow_heads.ArrowHead', default=_DEDAULT_ARROW, help="""
```
canavandl: spelling (_dedault_arrow)

bokehjs/test/common/defaults.coffee
canavandl: I'm sure not using the 'text' mixin on the `Title` coffee model was the best idea. It might be better to explicitly enumerate them instead.

bryevdv: I was thinking similar

bokehjs/test/common/defaults.coffee
almarklein: I was not even aware that the selection objects exist in Python-land too. But even if they do, `get_view()` is probably a rather JS specific feature. So adding it as an exception here, might make the most sense, for now.

bryevdv: Yes, definitely, and you can read/set the selection programmatically from python code. Truthfully, it probably needs to be turned into a "real" model object, then this method would not show up.

bryevdv: Will just keep this exception for now.

bryevdv:
ping @mattpap just wanted to let you know, plan to merge this very soon, I don' think it should cause you much or any conflict.

bokehjs/src/coffee/models/plots/plot.coffee
birdsarah: yes.

bokehjs/test/common/defaults.coffee
```diff
@@ -178,12 +186,12 @@ describe "Defaults", ->
...
-        #console.log(python_defaults)
-        #console.log('coffee defaults:')
-        #console.log(coffee_defaults)
-        console.log(_.difference(_.keys(python_defaults), _.keys(coffee_defaults)))
+        # console.log(name)
+        # console.log('python defaults:')
+        # console.log(python_defaults)
+        # console.log('coffee defaults:')
+        # console.log(coffee_defaults)
+        # console.log(_.difference(_.keys(python_defaults), _.keys(coffee_defaults)))
```
birdsarah: You commented this out, but this reporting was extremely helpful to me, and I couldn't find another way of getting correct reporting.

birdsarah:
Have had a quick look, there are lots of things going on around me, so I may have missed some things, but it looks like you're cleaning up a bunch of defaults changes that I made in the layout PR amongst other things - thanks ever so much for that!

I haven't run it locally, but LGTM.

bokehjs/test/common/defaults.coffee
bryevdv: I'm ok if you want to uncomment those two lines

bryevdv:
OK I am going to uncomment those two log lines in the test and then merge this. Ping @mattpap for a heads up

PR #4436, opened 2016-06-03
Chord diagram colors

### LINKED ISSUE #4435: Chord Diagram only has one color when using Python 2.7
Running the chord example produces the following output when using Python 2.7:

image

The bug is caused during the generation of colors in chord builder by the difference in integer division between Python 2 and Python 3.

clairetang6:
Color generation in the chord builder is on line 138. 

```python
colors = [color_in_equal_space(area / areas_in_radians.shape[0]) for area in range(areas_in_radians.shape[0])]
```

In Python 2.7, ```colors``` becomes a list of all the same color because ```area / area_in_radians.shape[0]``` is always 0. 

To fix this, I added ```division``` to the ```from __future__``` import statement. 

- [x] issues: fixes #4435
- [x] tests added / passed
- [ NA ] release document entry (if new feature or API change)

### DIFF
--- a/bokeh/charts/builders/chord_builder.py
+++ b/bokeh/charts/builders/chord_builder.py
@@ -16,7 +16,7 @@ functions.
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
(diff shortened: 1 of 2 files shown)


birdsarah:
LGTM - thanks @clairetang6

PR #4437, opened 2016-06-04
added encoder for timedeltas AND a test to confirm

### LINKED ISSUE #2583: Can't serialize timedelta column
If you have a data frame with a column of timedeltas, then make a ColumnDataSource out of it, the ColumnDataSource will be created:

````python
In [9]: source.data['delta']
Out[9]:
[Timedelta('0 days 00:33:00'),
 Timedelta('0 days 00:35:00'),
 Timedelta('0 days 03:01:00')]
````

But if you use that source in a plot, even if you don't use the column, when it comes time to serialize (show/save/embed etc) the plot, it fails:
````python
Timedelta('0 days 00:33:00') is not JSON serializable
````
Maybe we can provide some validation on ColumnDataSource creation? Or, at least provide a more helpful message on failure, as it's not immediately obvious what went wrong.

### PR DESCRIPTION
**Pull request by chalmerlowe**
- [X] issues: fixes #2583
- [X] tests added / passed

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -42,6 +42,10 @@ class BokehJSONEncoder(json.JSONEncoder):
         # datetime is a subclass of date.
         elif isinstance(obj, dt.datetime):
             return time.mktime(obj.timetuple()) * 1000. + obj.microsecond / 1000.
+        # Timedelta
+        # timedelta is class in the datetime library
+        elif isinstance(obj, dt.timedelta):
+            return dict(days=obj.days, seconds=obj.seconds, microseconds=obj.microseconds)
         # Date
         elif isinstance(obj, dt.date):
             return time.mktime(obj.timetuple()) * 1000.
--- a/bokeh/core/tests/test_json_encoder.py
+++ b/bokeh/core/tests/test_json_encoder.py
@@ -53,7 +53,6 @@ class TestBokehJSONEncoder(unittest.TestCase):
         ts = pd.tslib.Timestamp('April 28, 1948')
         self.assertEqual(self.encoder.default(ts), -684115200000)
 
-
 class TestSerializeJson(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by bryevdv** [comment]
Thanks @chalmerlowe ! I only see one issue, which is a whitespace/line ending one reported by our code linter:
```
E         File contains trailing whitespace: bokeh/core/tests/test_json_encoder.py, line 141.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: bokeh/core/tests/test_json_encoder.py, line 141.'])
```
If you delete the trailing whitespace and re-push to this branch, everything should be ready to go!

chalmerlowe:
Took out some whitespace on line 141 to resolve a check by the linter.

bryevdv:
LGTM (the one failing test is expected) any other comments @bokeh/dev ?

almarklein:
I added this too in https://github.com/bokeh/bokeh/pull/4349/files, except that I encoded it as an integer (total ms) just like the datetime is, which makes timedeltas Just Work when used as interval bounds.

almarklein:
I would not mind if this was merged in --- I can fix the conflict --- but I suspect serializing as an int would be more convenient than a dict.

bryevdv:
@almarklein Ah yes sorry I didn't realize this was similar to what is in #4349 If yo udon't mind fixing up the conflict I'd say yes let's merge this since it has a nice test. Agree for interval checking the `ms` encoding will mbe more useful, but the test can be updated as well. If you agree, please go ahead merge

bryevdv:
I'm actually curious about the `dict` encoding used here and previously for `relativedelta`. I'm not sure offhand where that format is handled on the JS side but it must be.

almarklein:
Merging now. Failing test is a glitch. @chalmerlowe Thanks for your contribution! Note that the serialization will be changed to a float (in ms) in an upcoming PR. If this is a problem for your use-case, please let us know.

PR #4442, opened 2016-06-05
Fix bug in throttling of render

### LINKED ISSUE #4439: Multiple renders ocurring
May be related to #4170.

May be related to performance #4395 

May be relevant to #4401 

Bokeh is rendering more times than it needs to seemingly all the time - on selection, on resize, on slider change. You always get two renders.

Also, wierdly as soon as you select / click on the slider (before moving it) you get two renders.

clairetang6:
This PR fixes a bug where multiple calls to ```request_render``` in the same throttling interval will cause two renders when it should only cause one.

The fix is to make sure that any calls to ```request_render``` don't generate a new render (through setTimeout) when one is already pending. 

I didn't add any tests for this. I would need some guidance on how that should be done if anyone has any ideas. 

- [x] issues: fixes #4439
- [ ] tests added / passed
- [ NA ] release document entry (if new feature or API change)

### DIFF
--- a/bokehjs/src/coffee/core/util/throttle.coffee
+++ b/bokehjs/src/coffee/core/util/throttle.coffee
@@ -35,7 +35,7 @@ throttle = (func, wait) ->
       clearTimeout(timeout)
       pending = true
       delay_animation(later)
-    else if (!timeout)
+    else if (!timeout and !pending)
       timeout = setTimeout((-> delay_animation(later)), remaining)
     return result
 


birdsarah:
I believe this is also the more general fix, compared to #4170

birdsarah:
ping @bryevdv

bryevdv:
LGTM change makes sense. Tested wth `color_scatter.py`. Qualitatively lasso select seems qualitatively more responsive but I have no hard measurements, however `trace` level logs confirm there are no extra or unexpected renders on selection. 

I'm going to go ahead and merge this because it is an important fix. However, some kind of testing would also be appreciated. I'm also not entirely sure how to best test this, so it might require some joint investigation. @clairetang6 can you make an issue to investigate adding a test for this function? Also, thanks!!

Issue #4425, opened 2016-06-02
Improve server index page 


bryevdv:
As a follow on to #4415 improve the basic `RootHandler` appearance and operation. Possible ideas:

* Make page content temptable 
* Provide default page with brief paragraph, project links, etc
* Improve styling of default list, CSS or bootstrap `list-group`
* Possibly use doc title in list in some way
* Possibly add a means for docs to provide a longer description
* Add docs for all of the above, and for disabling the index page

PR #4443, opened 2016-06-05
Add styling to bokeh apps index page and add docs

boylea:
Solves issue #4425 

This makes the server app index presentable

Here is a screenshot of the index page for two running applications:

screen shot 2016-06-04 at 6 33 40 pm

- [ ] issues: fixes #xxxx
- [ ] tests added / passed

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -67,6 +67,14 @@ Note that if multiple scripts or directories are provided, they
 all receive the same set of command line arguments (if any) given by
 ``--args``.
 
+You can see an index of all running applications at the server root:
+
+.. code-block:: none
+
+    http://localhost:5006/
+
+This index can be disabled with the ``--disable-index`` option
+
 Network Configuration
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -462,6 +462,11 @@ class Serve(Subcommand):
             action = 'store_true',
             help    = 'Do not use the default index on the root path',
         )),
+
+        ('--disable-index-redirect', dict(
+            action = 'store_true',
+            help    = 'Do not redirect to running app from root path',
+        )),
     )
 
 
@@ -535,6 +540,7 @@ class Serve(Subcommand):
                 "the `bokeh secret` command can be used to generate a new key.")
 
         server_kwargs['use_index'] = not args.disable_index
+        server_kwargs['redirect_root'] = not args.disable_index_redirect
 
         server = Server(applications, **server_kwargs)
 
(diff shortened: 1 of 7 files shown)


birdsarah:
LGTM - Thanks for adding this @boylea and for adding the docs.

ping @danielfrg @bryevdv  - the docs that @boylea has added, should probably have gone in as part of #4415.

boylea:
The integration tests are failing. I'm having trouble getting them to run at all locally, but will look into it tonight

birdsarah:
That's because you're an external contributor (so you don't have access to SauceLabs). I thought I'd fixed this problem, but apparently not.

Basically this means that the tests are passing.

birdsarah:
@bryevdv - I'm not sure this 100% covers all the things you had in mind for #4425.

danielfrg:
This is great. I notice that this PR removes the redirect when there is only one app running. Is this the desired behaviour? Usually I have just one app running and in that case I thought it was nice to just open the app without the need to see the index.

bryevdv:
I would say we could at least make that configurable e.g --index-redirect  I don't have strong opinions about whether the default is true or false.

danielfrg:
I think defaults are important here.

In my opinion the default for one app should be redirect instead of the index page. I believe this because like I said I usually have one app running, but thats me... now I am wondering if people really run more than one app when they call the command `bokeh serve ...`. If thats the most common use case then the index page as default is probably better, with the `--index-redirect` flag.

In any case this index is much better than my ugly list and the `404` error before that :)

Personally I would remove the "To disable this index page..." message from the index page, i think docs are good enough, i think it looks weird to tell you to remove the index from the index page. Not really that important.

quasiben:
@boylea the new styling is really great!  

My $0.02: I don't mind the message on disabling the index page and I have a slight preference for the redirect.

bryevdv:
My preference is also for a default to redirect. My rationale is this: If there is only one app, then the only possible useful action a user can take from the "list" index page is to click the one link to view the one app. But then why make them do this superfluous action at all? I can appreciate some cases might favor uniformity, but I would suggest `disable-index-redirect` to turn off a default redirect in that case.

bryevdv:
@boylea I think the preference is to keep the redirect as a default with an option to turn it off. Is that something you have the bandwidth to work on? As an alternative, just restoring the redirect and adding an issue to add `--disable-index-redirect` later would be fine.

boylea:
@bryevdv reverted to using redirect. Let me know if you'd rather I squish that change into the first commit to avoid the back and forth history. 

I don't have the bandwidth to do much more for the next couple days. I could get to it Thursday/Friday so it's up to you if want to wait for me to get to it or just do it yourself

bryevdv:
@boylea thanks! I have been experimenting with squash merges, so no need to fix anything up I don't think. We are planning to freeze for `0.12` Monday. So if you have change to work on the policy flag, great! If now, we will merge this and add an issue. Thanks so much for the PR!

boylea:
Added the option for disabling the redirect.

PR #4452, opened 2016-06-06
Fixes to legend rendering

### LINKED ISSUE #4390: Legend is too tall (and maybe too wide) if more than two items
See `examples/charts/file/scatter_multi.py` - at two or three legend items, the legend height is correct, but as number of legend items increases, the legend height increases inappropriately (but seemingly in proportion to the increase)

almarklein:
Fixes #4390

There were several things wrong with the current implementation of the legend rendering:
* It followed a rather peculiar version of the box model:
  * the `legend_padding` was used as padding *and* as margin
  * the `legend_spacing` was used as spacing *and* as padding
* the legend items were not rendered at their correct position (which was the main cause of #4390)
* The text height was not correctly taken into account
* the `label_standoff` was not used, instead the `legend_spacing` was used as standoff.

It now should follow the standard box model. I added a `legend_margin` property to complete it. Further I reduced the default value of the `label_width`, because some labels are just real short (and the actual text length is taken into account anyway), an of the `label_standoff` because it was huge.

cc @birdsarah

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -74,7 +74,7 @@ class Legend(Annotation):
 
     label_text_font_size = Override(default={ 'value' : '10pt' })
 
-    label_standoff = Int(15, help="""
+    label_standoff = Int(5, help="""
     The distance (in pixels) to separate the label from its associated glyph.
     """)
 
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -79,11 +79,11 @@ class Legend(Annotation):
     """)
 
     label_height = Int(20, help="""
-    The height (in pixels) of the area that legend labels should occupy.
+    The minimum height (in pixels) of the area that legend labels should occupy.
     """)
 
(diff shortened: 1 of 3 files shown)


birdsarah:
Had a look at: https://s3.amazonaws.com/bokeh-travis/135698545/examples.html#models/legends and it looks great. 

Looks like you have some test clean-up to do. And also appears that we have a reporting problem on the integration test -  the pytest-html report (https://s3.amazonaws.com/bokeh-travis/135698548/tests/pytest-report.html) is showing everything as passing, but the travis report clearly shows a failing test.

almarklein:
@birdsarah where do you see that we have an integration test failing? I only saw the Python-only legend tests, which I fixed, and now Travis is green.

bryevdv:
OK, charts legends that looked bad in prelim testing look much better with this PR. Merging now.

mattpap:
@almarklein, TS API wasn't updated. Please fix this in a followup PR.

PR #4453, opened 2016-06-06
4344/Better handle pd.DatetimeIndex in bokeh.charts 

### LINKED ISSUE #4344: TimeSeries not correclty represent x-axis when applied to DataFrame with datetime index
Bokeh version: 0.11.1

I'm trying to use TimeSeries to plot a Pandas dataframe with a DateTime index but it is not working :  I get time in milliseconds in the x-axis (as well as in the hover tool).
A partial fix is to create a Date column from the index and to specify explicitly x to use that column: it fixes the x display but the hover tool is still not working.

.

canavandl:
- [x] issues: fixes #4344
- [x] tests added / passed

### DIFF
--- a/bokeh/charts/builder.py
+++ b/bokeh/charts/builder.py
@@ -18,7 +18,9 @@ types on top of it.
 
 from __future__ import absolute_import
 
+import numpy as np
 import warnings
+
 from six import string_types
 from .attributes import AttrSpec, ColorAttr, CatAttr
 from .chart import Chart
@@ -670,12 +672,12 @@ class XYBuilder(Builder):
         """
         dim_ref = getattr(self, dim)
         values = dim_ref.data
-        dtype = dim_ref.dtype.name
+        dtype = dim_ref.dtype
 
         sort = self.sort_dim.get(dim)
 
         # object data or single value
-        if dtype == 'object':
+        if dtype.name == 'object':
             factors = values.drop_duplicates()
             if sort:
                 # TODO (fpliger):   this handles pandas API change so users do not experience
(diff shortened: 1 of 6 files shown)


bokeh/charts/builder.py
```diff
@@ -688,7 +690,7 @@ def _get_range(self, dim, start, end):
 
             setattr(self, dim + 'scale', 'categorical')
             return FactorRange(factors=factors.tolist())
-        elif 'datetime' in dtype:
```
canavandl: Per http://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html, this is a better check for datetime dtypes (handles array protocol types such as `dtype='<M8[ns]'`)

bokeh/charts/data_source.py
```diff
@@ -167,9 +167,7 @@ def get_values(self, selection):
             :class:`pandas.DataFrame`
 
         """
-        if selection in special_columns:
-            return special_columns[selection](self.data)
-        elif isinstance(selection, str):
+        if isinstance(selection, str):
             return self.data[selection]
         elif isinstance(selection, list) and len(selection) == 1:
```
canavandl: So this code never worked. If selection was a list it would raise an exception `TypeError: unhashable type: list` on L170 (due to check the existence of a list in a dictionary)

bokeh/charts/data_source.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, label, data, attr_specs):
 
         """
         self.label = label
-        self.data = data.reset_index()
```
canavandl: This conflicts with L171 (resetting the index then trying to read the index value).

bokeh/charts/builder.py
```diff
@@ -688,7 +690,7 @@ def _get_range(self, dim, start, end):
...
+        elif np.issubdtype(dtype, np.datetime64):
```
canavandl: Per http://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html, this is a better check for datetime dtypes (handles array protocol types such as dtype='<M8[ns]')

canavandl:
thanks @Maggie-M

canavandl:
Todo: fix up examples

bryevdv:
fixed another whitespace issue, will merge when green

PR #4457, opened 2016-06-08
explicitly document installing into the bokehjs directory

andreamcinnes:


bryevdv:
Thanks @andreamcinnes !

PR #4471, opened 2016-06-10
Move toolbar to right of plots

### LINKED ISSUE #4399: Make toolbar right by default?
With the new toolbar and title set-up, there's an increased chance that the title and toolbar might clash with each other. 

If that clash does happen, a user can set `toolbar_sticky=False`.

None-the-less, I wanted to propose moving the toolbar default position to the right side of a plot.

See examples/charts/file/scatter_multi.py for an example where a default right toolbar would be more pleasing.

Maggie-M:
issues: fixes #4399

screen shot 2016-06-10 at 10 04 11 am

The new toolbar location leaves the top of the y-axis flush with the top of the browser window

screen shot 2016-06-10 at 10 05 18 am

### DIFF
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -450,7 +450,7 @@ class Plot(LayoutDOM):
         The toolbar is automatically created with the plot.
     """)
 
-    toolbar_location = Enum(Location, default="above", help="""
+    toolbar_location = Enum(Location, default="right", help="""
     Where the toolbar will be located. If set to None, no toolbar
     will be attached to the plot.
     """)
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -84,7 +84,7 @@ class ToolbarBox(LayoutDOM):
 
     """
 
-    toolbar_location = Enum(Location, default='above', help="""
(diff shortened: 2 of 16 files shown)


bryevdv:
ping @bokeh/dev @birdsarah thoughts? This seems OK and allows a default sticky toolbar to exist and not overlap titles which are usually at the top. 

Only issue:  seems like we might want to increase the min border a bit? that "10" tick label has basically no space above it and abuts the window edge directly. Or maybe there are other ideas.

Maggie-M:
Changed the toolbar places where it was set explicitly on the left in examples because it would overlap the y-axis

screen shot 2016-06-10 at 10 28 59 am

bryevdv:
@birdsarah I guess the options are update the image, or update the test to put the mostly empty toolbar on top explicitly?

bryevdv:
OK, tests green now. I am +1 on merging this, tho it should definitely be called out in the release notes. @bokeh/dev any comments?

bryevdv:
@bokeh/dev I'll merge this later today unless there's comment or objection. We can always revert it...

birdsarah:
Happy to see this went through :D

Issue #4403, opened 2016-05-28
Better defaults for resize event


birdsarah:
Currently we listen to window resize. A better, default would probably be listening to the `.bk-root` element and then resize for that root.

Related: #4386

damianavila:
@almarklein @birdsarah is this solved by #4473?

birdsarah:
No.

Issue #4386, opened 2016-05-27
Resize-event should be customizable


birdsarah:

almarklein:
@birdsarah Do you have an idea of how this would work from a user perpective? Div's don't emit `resize` events automatically, so if a user set the resize-reference to another element, he/she is also responsible for triggering a resize event on that element, I suppose?

Listening for resize events on bk-root seems tricky; who/what code is responsible for making bk-root trigger a resize event?

damianavila:
@almarklein @birdsarah is this solved by #4473?

almarklein:
I am not sure, as I don't understand the precise meaning. If it means that 3th party code can embed a Bokeh plot(s) and make it resize, then such code can do that by calling `document.resize()`.

Though maybe we want to be able to unsubscribe the windows.resize event. Though in that case we'd need a dedicated handler for `window.resize` to keep track of browser zoom.

birdsarah:
No. This discussion is still open.

birdsarah:
No. This discussion is still open.

thomn:
I would really need this too. Here's a video showing the problem : https://dl.dropboxusercontent.com/u/13563926/Capture%20d%27%C3%A9cran%20vid%C3%A9o%20de%2011-07-2016%2013%3A48%3A33.webm

Issue #3783, opened 2016-01-29
Reset should call the responsive resize.


birdsarah:
Bryan noticed on the happiness demo that resize causes the plot to shrink.

It actually causes the plot to shrink or grow depending on whether responsive view is making plot bigger or smaller. 

I'm guessing that reset is resetting the plot_width and plot_height (which I don't think it used to do) and then the responsive resize code is not being called.

birdsarah:
You can see this same phenomenon in `howto/charts/excel.ipynb`

birdsarah:
I see #1683 was marked as fixed via #3254 - I'm guessing that's the source of this bug. ping @mattpap

jbednar:
I'm not sure if it's the same problem as you're seeing, but the "reset" tool doesn't seem to be generating a callback for our (admittedly highly customized) callback setup in https://github.com/bokeh/datashader/blob/master/examples/nyc_taxi.ipynb).  I.e., our Python callback function is called on any zoom or pan, and the reset tool correctly resets the ranges, but it doesn't seem to generate a call to our Python code.  After the reset, I have to then zoom or pan a little to get the image redrawn.  Again, this could be our fault, as part of our own JavaScript code we're using, but I don't *think* so, because we haven't done anything special for the reset tool, at least not deliberately.  So I think our problem has the same source as what you are reporting here, and hopefully the same fix.

philippjfr:
>our Python callback function is called on any zoom or pan, and the reset tool correctly resets the ranges, but reset doesn't seem to generate a call to our Python function.

Note that the python callback is executed through a regular ``CustomJS`` callback. Reset events definitely do not trigger callbacks attached to the ``x_range`` or ``y_range`` currently.

birdsarah:
Confirm under #4392

jbednar:
> Confirm under #4392

Is that a statement, i.e. you confirm that it still applies under 4392, or a question, asking us to confirm that it still applies under 4392, or...?

birdsarah:
It's a ToDo. I believe that it's done, but I want to check. #4392 is the layout clean-up PR.

birdsarah:
There are other issues with the resize and reset tool since the layout came in, but the original bug reported on this issue is now fixed. That is, if you now hit reset on a responsive plot, it does not weirdly resize to an original "fixed" size.

jbednar:
Thanks!

PR #4473, opened 2016-06-10
Improving resizing and initialization

### LINKED ISSUE #4401: Trigger event when bokeh finishes rendering
At the moment, we do a rather un-graceful trigger of a window resize at three delayed time intervals to try and get the page in the right shape when it a responsive layout loads.

However, if your browser took more than 100ms to load the event you're going to end up with some ugly looking layout. e.g. when loading: https://cdn.rawgit.com/birdsarah/bokeh-miscellany/master/new_layout_tests/words_and_plots.html

bad layout

(thanks @almarklein)

Having an event at the end of bokeh render has been requested in the past: #2150 but with the responsive layout I'm putting this into the bug category as anything that will get us to a better way to trigger resize after render would be good. 

The current resize code is here:

https://github.com/bokeh/bokeh/blob/birdsarah/4164_rework_layout/bokehjs/src/coffee/models/layouts/layout_dom.coffee#L35-L39

```coffeescript
    if @model._is_root is true
      resize = () -> $(window).trigger('resize')
      _.delay(resize, 50)
      _.delay(resize, 100)
      _.delay(resize, 200)
```

almarklein:
Issues: fixes #4401

Other related issues that we might fix in this PR:
* #4386, #4403
* #3783

This is a WIP. I spend a lot of time trying to figure out how the events propagate and which part of the code sets solver constraints, etc. For some examples (e.g. `howto/dashboard.py`) work fine now, but I just saw that `howto/words_and_plots.py` is not initialized entirely correct yet.

Any comments and suggestions are very welcome.
@birdsarah @bryevdv


bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -96,6 +96,11 @@ class PlotCanvasView extends Renderer.View
 
     @throttled_render = throttle(@render, 15) # TODO (bev) configurable
 
+    # Keep track of which plots of the canvas are not yet rendered
+    if not @model.document._unrendered_plots?
+      @model.document._unrendered_plots = {}  # poor man's set
+    @model.document._unrendered_plots[@id] = true
+
```
almarklein: I am not proud of this ...

birdsarah:
@almarklein - I noticed this comment from you. 

>  # Ideally the solver would  settle in one pass (can that be done?),  but it currently needs two passes to get it right.

I don't know why the solver doesn't get it right on first render, but for things other than widgets (which we need to render in order to know how big they are), can we not solve twice render once? Currently rendering and solving is all linked together but that isn't necessary.

birdsarah:
>  For some examples (e.g. howto/dashboard.py) work fine now, but I just saw that howto/words_and_plots.py is not initialized entirely correct yet.

@almarklein - that is not the result of my testing. I have tested `words_and_plots.py`, `dashboard.py`, and my own `large_lorenz.py` (which tests a slow to load unresponsive page). All work great!

Maybe you have a cacheing issue that's making you think that they're not yet working.

----
Once I'm past initialization, (after loading a page and playing with it for a bit), a small resize triggers the following (reported) events.

```console
Bokeh: Suggest width on Document -- 990
Bokeh: Suggest height on Document -- 471
Bokeh: Rendering CanvasView [force=false] with width: 990, height: 471, ratio: 2
Bokeh: Suggest width on Document -- 990
Bokeh: Suggest height on Document -- 471
Bokeh: Circle GlyphRenderer (1033): render finished in 7ms
```
This is (as you know! - am just writing out for my own thoughts) because of: 

```coffeescript
  resize: () ->
    # Notes on resizing (xx:yy means event yy on object xx):
    # window:event -> document.resize() -> solver:resize
    #   -> LayoutDOM.render()
    #   -> PlotCanvas.resize() -> solver:update_layout

    # Ideally the solver would  settle in one pass (can that be done?),
    # but it currently needs two passes to get it right.
    @_resize()
    @_resize()
```

I don't *think* it is necessary to be calling the document resize twice forever, I believe it is only necessary on initialization. And my experiments with removing it are pointing me in that direction also. In fact many of my tests are initializing fine with just one resize - even at initialization. But widgets are a disaster without the double initialization. The following seems to be working well.  What do you think?

```coffeescript
  resize: () ->
    # Notes on resizing (xx:yy means event yy on object xx):
    # window:event -> document.resize() -> solver:resize
    #   -> LayoutDOM.render()
    #   -> PlotCanvas.resize() -> solver:update_layout

    # Ideally the solver would  settle in one pass (can that be done?),
    # but it currently needs two passes to get it right.
    @_resize()
    if not @_first_resize_complete?
      @_resize()
      @_first_resize_complete = true
```


----

Separate issues (maybe not to fix in this PR but noting as I found while testing). 

1) Multiple renders do not seem to be occurring any more, but on `plotting/file/iris.py` which is a fixed example - window resize is causing a re-render of just the circle renderer....not sure what would be triggering that.

2) I also noticed this odd canvas initialization on first load - when mode is `box`:
screen shot 2016-06-13 at 10 53 18 pm

This only happens on box mode (not `fixed`, `width_ar`). It happens in Chrome and Firefox if the developer tools are not present, but only happens in Firefox if the developer tools are present. This makes me think we have an additional race condition - where the constraints from the width and height updates are not getting properly registered in time.

It also only happens when the plot size is larger than the default, it doesn't happen when the plot is smaller than the default.

Here are some examples to show it failing in all directions:

screen shot 2016-06-13 at 11 33 13 pm

screen shot 2016-06-13 at 11 33 40 pm

At 900 x 900:
screen shot 2016-06-13 at 11 34 00 pm

birdsarah:
All that said, I would be happy to merge this as-is, as it is a definite improvement on where we currently are.

birdsarah:
I found the fix for the box issue. I have a PR against your branch with my suggestions in it.

birdsarah:
I would suggest we deal with the other issues you mentioned in another PR. This is a particularly tricky issue and I think it'll be good to have this in in one commit that doesn't do anything else.

almarklein:
> ... can we not solve twice render once? Currently rendering and solving is all linked together but that isn't necessary.

Yes please! That's what I tried to work towards. But apparently there is code in the rendering that influences the layout. The one line that I suspected the most (and which I commented on) I could not remove.

> > howto/words_and_plots.py is not initialized entirely correct yet.

> that is not the result of my testing

I'm sure it fails for me (did gulp clean, CTRL-F5). Its *almost* correct, but the bottom margin of the first plot is initially slightly too small. Happens on Firefox and Chrome. 

> I don't think it is necessary to be calling the document resize twice forever, I believe it is only necessary on initialization.

It is for me/Windows, really. On Chrome the resize will sometimes lead to incorrect layout otherwise. And without it on Edge, the words-and-plots example initialized completely off.

almarklein:
cc @mattpap

birdsarah:
Forgot to say, all good on the windows stuff. Hopefully we can better get to the bottom of it later.

birdsarah:
Looks like you have a JS unit test failure that needs updating given your new code.

birdsarah:
@almarklein - not sure why some of the JS tests are failing on an "unknown edit variable" but others aren't. Maybe some kind of testing anomaly but may be indicative of a race-condition....I'm not sure. I can look into it if you like.

almarklein:
I think I've seen that error earlier when working on this PR ... also suspicious that it fails to run the iris example.

birdsarah:
Yeah....I remember seeing it from when I was working on layout too...i've just forgotten why off the top of my head.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -470,11 +467,33 @@ class PlotCanvasView extends Renderer.View
...
-    @canvas_view.set_dims([width, height], true)
+    @canvas_view.set_dims([width, height], true)  # this indirectly calls @request_render
+
+    # Prepare the canvas size, taking HIDPI into account. Note that this may cause
+    # a resize of the canvas, which means that any previous calls to ctx.save() may be undone.
+    @canvas_view.prepare_canvas()
+
+    try
+      @update_constraints()
+    catch silent_error
```
birdsarah: I have long wished that kiwi.js threw errors up more gracefully so that we could handle them cleanly. @bryevdv and I had been talking about making some improvements to kiwi.js - this could be one of them.

birdsarah:
@almarklein I took the liberty of fixing up the broken example given that I was the one who broke it!

almarklein:
Thanks! One failing tests: `test_box_responsive_resizes_width_and_height_with_fixed_aspect_ratio`. Is that new? I restarted to see whether its a glitch.

birdsarah:
So, the test is still failing with both of our fixes in. I'm not sure why, but I do think it's just a test issue. But it is a complicated test. There's one above it that's much simpler that just checks that it hasn't breached its height. Perhaps we could add just a similar short one and remove this v. complicated test. @almarklein this test was yours originally, what do you think?

birdsarah:
working on another idea!

birdsarah:
@almarklein, I've updated the test to be testing the correct mode `scale_both`!!

I'm now convinced however that it's picking up a real bug:

```python
from bokeh.io import show
from bokeh.models import Plot, ColumnDataSource, Rect, DataRange1d, Row

source = ColumnDataSource(dict(x=[1, 2], y=[1, 1]))
plot = Plot(
    plot_height=400, plot_width=800,
    x_range=DataRange1d(), y_range=DataRange1d(),
    responsive='scale_both'
)
plot.add_glyph(source, Rect(x='x', y='y', width=0.9, height=0.9))

show(plot)
```

When I run this, if I make the window small. The plot seems to be the right size, but the window is much bigger than it should be.

You can also see this on `examples/embed/embed_responsive_width_height.py`

birdsarah:
My short-term proposal is to xfail the integration test and to open a new issue to investigate the bug in scale_both - this PR is a major win for the major modes (fixed, scale_width, stretch_both) - and `scale_both` mode support is still very limited.

bryevdv:
I'm OK with that, thoughts @almarklein @mattpap ?

birdsarah:
I should add - I'd like to fix the bug before release, but I'm currently testing all my work by locally merging this branch so I think it would benefit everyone to merge it.

birdsarah:
I'm merging this to free up other work. Happy to keep working on it today. The issue for the scale_both problem is: #4537

PR #4475, opened 2016-06-11
remove dupe props in PlotCanvas, cleanup

bryevdv:


bryevdv:
@mattpap  OK I think this diff sucks because I had to merge master into this to avoid future conflicts, but master is not merged into your JS/TS branch yet. Not saying this is necessarily what should get merged, but I'd like to get your opinion on this. We can cherry-pick commits to a another PR if that ends up making more sense. My commits start with "remove broken/useless resolve_ref cruft"

With this branch, everything seems to work with the exception of tile renderer stuff (ping @brendancol) All the dupe props on `PlotCanvas` are removed.

mattpap:
Why this PR doesn't trigger travis CI tests?

mattpap:
> everything seems to work with the exception of ...

It breaks all bokehjs examples in a new way. Easy to fix but still.

mattpap:
>  Easy to fix but still.

Actually only partially, because `plot` isn't set on `Axis` (fails in Axis._ranges).

bryevdv:
> Why this PR doesn't trigger travis CI tests?

Because it's a PR to a branch different than master

Regarding other issues, I definitely didn't expect to be able to finish this alone, but I was hoping I could clear out some of the tedious lifting.

mattpap:
> Because it's a PR to a branch different than master.

Good to know. I was expecting something to be wrong here.

bryevdv:
If you think this PR is a profitable way to proceed, then it could be merged into your PR. Alternatively we can make a new PR of this branch against master just for running tests. There's definitely still some things wrong so your expectation is founded.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -64,21 +66,46 @@ class Plot extends LayoutDOM.Model
   initialize: (options) ->
     super(options)
 
+    for xr in _.values(@extra_x_ranges).concat(@x_range)
+      plots = xr.get('plots')
```
birdsarah: As a side note - `plots` doesn't seem like a great name for this property. It makes this code super confusing.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -64,21 +66,46 @@ class Plot extends LayoutDOM.Model
...
+      if _.isArray(plots)
+        plots = plots.concat(@)
```
birdsarah: Why do we attach the plot views?

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -88,32 +115,66 @@ class Plot extends LayoutDOM.Model
     if not @height?
       @height = @plot_height
 
+  Object.defineProperty(this.prototype, "plot_canvas", { get: () -> @_plot_canvas })
```
birdsarah: Why can't we just set @plot_canvas in line 100?

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -69,6 +63,14 @@ class PlotCanvasView extends Renderer.View
     super(options)
     @pause()
 
+    # TODO (bev) this sucks a bit
```
birdsarah: Can you elaborate on this comment - so we know what to try and clean-up in the future?

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -69,6 +63,14 @@ class PlotCanvasView extends Renderer.View
...
+    @visuals = {}
+
+    for spec in @model.plot.mixins
```
birdsarah: If I'm not mistaken this is duplicating code from Renderer.

Can I suggest that we just break out a method and reuse it in both places.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -813,49 +682,48 @@ class PlotCanvas extends LayoutDOM.Model
...
-    constraints = []
 
-    # Add the constraints that always apply for a plot
-    min_border_top    = @get('min_border_top')
-    min_border_bottom = @get('min_border_bottom')
-    min_border_left   = @get('min_border_left')
-    min_border_right  = @get('min_border_right')
-    frame             = @get('frame')
-    canvas            = @get('canvas')
+    min_border_top    = @plot.min_border_top
```
birdsarah: I would have thought these constraints now belong on Plot given that you're dealing with min_border stuff there.

birdsarah: Cancel that - I forgot they relate to the panels......bleugh...this somehow feels murky.

bokehjs/test/models/axes/axis.coffee
```diff
@@ -16,10 +16,11 @@ Toolbar = utils.require("models/tools/toolbar").Model
 describe "Axis.Model", ->
 
   it "should have a SidePanel after add_panel is called", ->
-    p = new PlotCanvas({
+    pp = new Plot({
```
birdsarah: This seems like an odd re-working of this test.

I think it highlights that the separation isn't quite right between Plot and PlotCanvas when it comes to the side panels. The ownership is now spread over plot and PlotCanvas. Seems to me like the logical extension of this move is that SidePanels are on Plot.

bokehjs/test/models/plots/plot_canvas.coffee
```diff
@@ -25,11 +25,13 @@ Toolbar = utils.require("models/tools/toolbar").Model
 describe "PlotCanvas.Model", ->
 
   it "should set the responsive mode to box by default", ->
-    p = new PlotCanvas({x_range: new DataRange1d(), y_range: new DataRange1d()})
+    pp = new Plot({x_range: new DataRange1d(), y_range: new DataRange1d()})
+    p = pp.plot_canvas
```
birdsarah: We're unit testing plot canvas here. PlotCanvas should be initiated and tested.

(Same for all below, or move the test)

bokehjs/test/models/plots/plot_canvas.coffee
```diff
@@ -187,11 +190,12 @@ describe "PlotCanvas.View render", ->
     utils.stub_solver()
 
     @test_doc = new Document()
-    @test_plot = new PlotCanvas({
+    plot = new Plot({
       x_range: new Range1d({start: 0, end: 1})
       y_range: new Range1d({start: 0, end: 1})
       toolbar: new Toolbar()
     })
+    @test_plot = plot.plot_canvas
```
birdsarah: this is just test hacking and is going to leave a confusing mess behind.

birdsarah:
Made some comments inline.

Overall:

I think the SidePanel's should move up to Plot and be handled there. If Plot is where x_ranges, and titles, and annotations are specified, then side panels that handle them should be on Plot.

I think this move is a great opportunity to add tests for parts of PlotCanvas (formerly Plot) that were previously un-tested. I'm specifically thinking of the API methods `add_renderers`, `add_glyph`, `add_layout`, `add_tools`.

Let's make sure after this re-jig that all the tests make sense and help document & test functionality rather than just tweaking things so they pass.

bokehjs/src/coffee/models/plots/plot.coffee
mattpap: The way you implemented this suggests that you wanted to make `_plot_canvas` private and immutable from outside (as much it is possible in JS). What you see here is the correct implementation of that intent. Of course we can initialize `plot_canvas` directly, but that has different semantics. Which do you actually want?

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -88,32 +115,66 @@ class Plot extends LayoutDOM.Model
...
+      else
+        # XXX: this part should be unnecessary, but you can't configure tool.plot
+        # after construting a tool. When this limitation is lifted, remove this code.
+        attrs = _.clone(tool.attributes)
+        attrs.plot = this
+        new tool.constructor(attrs)
 
-  plot_canvas: () ->
-    @_plot_canvas
+    @set(@toolbar.tools, @get("toolbar").tools.concat(new_tools))
```
mattpap: Addint a layer of indirection to access `tools`, breaks semantics of `select()`/`select_one()`. With the existing implementation it's not possible to select tools. Whatever we do, we want not to break symmetry between this and Python APIs.

PR #4478, opened 2016-06-12
Bryanv/4363 notebook unminified

### LINKED ISSUE #4363: Bokeh Server w/ multiple processes

bryevdv:
- [x] issues: fixes #4363

ping @mattpap OK this sucks for sure, but it works to let un-minified inline resources work in the notebook. I'd like to propose merging this for now, until you figure out what to do longer term for `bokeh-compiler.js`

### DIFF
--- a/bokehjs/gulp/tasks/scripts.coffee
+++ b/bokehjs/gulp/tasks/scripts.coffee
@@ -6,6 +6,7 @@ gulp = require "gulp"
 gutil = util = require "gulp-util"
 rename = require "gulp-rename"
 transform = require "vinyl-transform"
+replace = require "gulp-replace"
 uglify = require "gulp-uglify"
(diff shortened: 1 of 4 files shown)


bryevdv:
@mattpap I have to look at #3639 again, I think this has to be merged. There is zero chance of debugging without being able to use unminified resources.

bryevdv:
I will merge in the next hour or so unless there's comment.

mattpap:
This isn't great, but we can live with this. The `replace` line will be removed when compiler is removed.

bryevdv:
Agreed

PR #4479, opened 2016-06-12
simplify erroring test

### LINKED ISSUE #4195: restore pandas_dataframe test
Was temporarily removed in #4193, should be restored when the root cause can be addressed.

bryevdv:
issues: fixes #4195

Maybe this will help. Will run CI tests on it repeatedly.


bryevdv:
OK failed once out of ~8 times. Interestingly other `compat` examples also failed (differently) that one time.

Issue #4285, opened 2016-05-05
Testing infrastructure bug/wish list


mattpap:
NOTE: this will be updated and refined gradually.
- [x] don't display unrelated (especially `DEBUG`) logs in the main log
- [ ] unify test output from different test runners (if or as much as possible) (especially `gulp test`)
- [x] restore original logic around javascript errors/warning/messages
- [ ] images in examples' report should have fixed size (as originally), so that we don't have to wait for all images to load to be able to use the report. Otherwise the page jumps and one can't focus on an example.
- [ ] integration tests should be aware of failures in javascript and not proceed unless bokehjs and plots loaded properly
- [x] some examples' images are no different, yet there is full black diff image generated and uploaded
- [x] LABEL WHICH IMAGE IS EXPECTED, AND WHICH IS ACTUAL

birdsarah:
If you'd like me to not comment on this issue, while your thinking evolves, then do let me know.

> integration tests should be aware of failures in javascript and not proceed unless bokehjs and plots loaded properly

+1 - Do we need a BokehJS hook for this?

In addition, the selenium tests currently fail in a completely unintelligible way if the python part of the test fails. Need to handle this much more gracefully.

> some examples' images are no different, yet there is full black diff image generated and uploaded

Yes - I've noticed this too - not sure what the cause is - maybe there's a diff that we can't see - that doesn't make any sense though.

> images in examples' report should have fixed size (as originally), so that we don't have to wait for all images to load to be able to use the report. Otherwise the page jumps and one can't focus on an example.

I understand your reasoning, but I really didn't like the squashed images. I found it made it harder to parse the examples report as a whole. Happy to think of compromise solutions - like fixed height boxes with scrolling images inside them. Having a grid to of fixed-height boxes to make the whole report more compact.

I would also add (as you said, there's probably more in my brain, but I just thought of this):
* I find it confusing that in the examples report, the actual/test-generated image is on the left and the expected/reference is on the right. For whatever reason I intuitively expected actual images to be on the right. In the integration screenshot tests, the images are on the right. We should be consistent, and maybe we should add extra information in both reports so its super clear (especially if you're new) which is the reference image and which is the test image.

birdsarah:
> don't display unrelated (especially DEBUG) logs in the main log

I did a quick thing on this in my latest PR - e.g. https://travis-ci.org/bokeh/bokeh/jobs/128202446; but I still get a bunch of boto stuff at the end.

[...]

PR #4482, opened 2016-06-12
quiet boto log level

bryevdv:
Addresses one bullet of #4285


bryevdv:
Much better: 

screen shot 2016-06-12 at 3 33 33 pm

PR #4483, opened 2016-06-12
Adding Callback Instances to Group class

### LINKED ISSUE #2715: Can't add callbacks to CheckboxGroups
When I run

    cb = Callback(code="console.log(cb_obj);")
    checkbox_group = CheckboxGroup(labels=["1", "2", "3"], active=[0, 1], callback=cb)

I get an error: 

    AttributeError: unexpected attribute 'callback' to CheckboxGroup, possible attributes are active, disabled, inline, labels, name, session or tags

Is this expected behavior? If we can't add callbacks to CheckboxGroups, what is the point of having them?

guillochon:

- [x] issues: fixes #2715
- [ ] tests added / passed

Have not been able to test this PR, but just followed the example for Ranges: https://github.com/bokeh/bokeh/blob/master/bokeh/models/ranges.py#L26

### DIFF
--- a/bokeh/models/widgets/groups.py
+++ b/bokeh/models/widgets/groups.py
@@ -33,6 +33,10 @@ class AbstractGroup(Widget):
         """
         self.on_change('active', lambda attr, old, new: handler(new))
 
+    callback = Instance(Callback, help="""
+    A callback to run in the browser whenever a button group is manipulated.
+    """)
+
 @abstract
 class ButtonGroup(AbstractGroup, ButtonLike):
     """ Abstract base class for groups with items rendered as buttons.
@@ -50,10 +54,6 @@ class Group(AbstractGroup):
     inline = Bool(False, help="""
     Should items be arrange vertically (``False``) or horizontally
     in-line (``True``).
-    """)
-
-    callback = Instance(Callback, help="""
-    A callback to run in the browser whenever a radio button group is manipulated.
     """)
 
 class CheckboxGroup(Group):
--- a/bokeh/models/widgets/groups.py
+++ b/bokeh/models/widgets/groups.py
@@ -52,6 +52,10 @@ class Group(AbstractGroup):
     in-line (``True``).
     """)
 
+    callback = Instance(Callback, help="""
+    A callback to run in the browser whenever a radio button group is manipulated.
+    """)
+
 class CheckboxGroup(Group):
     """ A group of check boxes.
 
@@ -59,10 +63,6 @@ class CheckboxGroup(Group):
 
     active = List(Int, help="""
     The list of indices of selected check boxes.
-    """)
-
-    callback = Instance(Callback, help="""
-    A callback to run in the browser whenever a checkbox group is manipulated.
     """)
 
 class RadioGroup(Group):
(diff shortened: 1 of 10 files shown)


bokeh/models/widgets/groups.py
```diff
@@ -50,6 +51,10 @@ class Group(AbstractGroup):
     Should items be arrange vertically (``False``) or horizontally
     in-line (``True``).
     """)
+    
+    callback = Instance(Callback, help="""
+    A callback to run in the browser whenever a group is manipulated.
+    """)
```
bryevdv: To be precise about I this, I think for now it would be better to add this just to the two subclasses I linked in the issue, not all of the `AbstractGroup` subclasses seem to have the callback machinery. But those two do.

bokeh/models/widgets/groups.py
```diff
@@ -4,9 +4,10 @@
 from __future__ import absolute_import
 
 from ...core.properties import abstract
-from ...core.properties import Bool, Int, String, List
+from ...core.properties import Bool, Int, String, List, Instance
 from .widget import Widget
 from .buttons import ButtonLike
+from .callbacks import Callback
```
bryevdv: This is one level up: `..callbacks`

bryevdv:
@guillochon if you can update the PR from the small comments here and also https://github.com/bokeh/bokeh/issues/2715#issuecomment-225460883 then I will build and construct a test

guillochon:
OK, I believe the proper changes have been made, ready to be tested.

bryevdv:
@guillochon thanks! There is one small test issue left, we added a fairly strict linter, and it does not allow trailing whitespace:

https://travis-ci.org/bokeh/bokeh/jobs/137263865#L2566-L2570

With that fixed, all the tests that are expected to pass will be passing. Thanks!

bryevdv:
OK checked locally, and it's working! However, if I can trouble you for one more change, the `CheckBoxGroup` also has the JS code to call the callback, but the callback is not declared in JS or python. This is actually bug and results in an error when a button is pushed. 

Can you add the delcaration to the python class, and also the `p.Instance` to:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/checkbox_group.coffee

Then I believe the cb will be supported on everything except `RadioGroup`. If you wanted to really round this out, then you could add the same one line to execute js callbacks here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/radio_group.coffee#L41

Then the python declaration could go back on the base class as you had it originally. I realize you may not have lots of time, so just let me know. If need be we can merge this as-is and someone can do the rest in a follow up. Thanks for the contribution!

bryevdv:
Brilliant. Will merge when tests are green, thanks!

bryevdv:
@guillochon last problem (I promise!) The python `callback` declaration needs to go up one more level to `ButtonGroup`. That's what `CheckboxButtonGroup` and `RadioButtonGroup` inherit from. right now they are missing on those classes, and this was caught by some of our automated JS/python checks (thanks @havocp !)
```
CheckboxButtonGroup: python is missing some properties found in CoffeeScript
    CheckboxButtonGroup.callback: coffee defaults to null but python has no such property
CheckboxButtonGroup
[]
RadioButtonGroup: python is missing some properties found in CoffeeScript
    RadioButtonGroup.callback: coffee defaults to null but python has no such property
RadioButtonGroup
[]
```
I'd also be ok with moving the declaration all the way up to `AbstractGroup`, don't have a strong feeling about that.

bryevdv:
All expected to pass tests are passing, merging. Thanks @guillochon !

PR #4484, opened 2016-06-12
update all the responsive modes

### LINKED ISSUE #4396: API - Name the responsive modes.

bryevdv:
issues: fixes #4396

ping @birdsarah could use some help verifying this set of change is complete, etc.

### DIFF
--- a/bokeh/charts/glyphs.py
+++ b/bokeh/charts/glyphs.py
@@ -770,7 +770,7 @@ class QuartileGlyph(Interval):
 class BoxGlyph(AggregateGlyph):
     """Summarizes the distribution with a collection of glyphs.
 
-    A box glyph produces one "stretch_both" for a given array of vales. The box
+    A box glyph produces one "box" for a given array of vales. The box
     is made up of multiple other child composite glyphs (intervals,
     scatter) and directly produces glyph renderers for the whiskers,
     as well.
--- a/bokeh/charts/glyphs.py
+++ b/bokeh/charts/glyphs.py
@@ -770,7 +770,7 @@ class QuartileGlyph(Interval):
 class BoxGlyph(AggregateGlyph):
     """Summarizes the distribution with a collection of glyphs.
 
-    A box glyph produces one "box" for a given array of vales. The box
+    A box glyph produces one "stretch_both" for a given array of vales. The box
     is made up of multiple other child composite glyphs (intervals,
     scatter) and directly produces glyph renderers for the whiskers,
     as well.
--- a/bokeh/charts/tests/test_chart.py
+++ b/bokeh/charts/tests/test_chart.py
@@ -53,7 +53,7 @@ class TestChart(unittest.TestCase):
         self.assertEqual(self.chart.title.text, "new_title")
 
     def test_responsive(self):
-        self.assertEqual(self.chart.responsive, 'width_ar')
+        self.assertEqual(self.chart.responsive, 'scale_width')
 
     def check_chart_elements(self, expected_tools):
         self.assertIsInstance(self.chart.left[0], LinearAxis)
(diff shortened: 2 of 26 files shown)


bryevdv:
E.g. not sure if this also needs updating:
```
$ grin bk-layout-box 
./less/continuum.less:
  343 : .bk-toolbar-box.bk-layout-box .bk-toolbar-right {
```

bryevdv:
OK, I am now pretty confident that all the changes that are in the PR are ones that should be there. Could use help with the converse: has anything been missed?

mattpap:
Looking at this, I would change `responsive` to something more appropriate, e.g. `sizing_mode`.

bryevdv:
I do think that the name "responsive" seems a bit more like a boolean flag. I don't have super strong feelings about that except that it might be hard to change correctly with confidence. Thoughts @bokeh/dev ?

mattpap:
> except that it might be hard to change correctly with confidence

That's actually a good reason to change it now to something unique, because later on it will be even harder to distinguish from other uses.

bryevdv:
@birdsarah any thoughts?

birdsarah:
I like `sizing_mode`, or `responsive_mode` when writing about the new layout stuff, I have been using the word mode a lot so this feels right. I think `sizing` is probably less jargon-y than `responsive` so happy to go with that.

birdsarah:
I like `sizing_mode`, or `responsive_mode` when writing about the new layout stuff, I have been using the word mode a lot so this feels right. I think `sizing` is probably less jargon-y than `responsive` so happy to go with that.

bryevdv:
OK I will update to `sizing_mode` this afternoon and get this merged.

birdsarah:
If we go with sizing_mode we'll need to deprecate `responsive` and also decide if we want to keep a True/False shortcut (I would vote for no in the short-term) until it becomes clear that there are preferred modes.

birdsarah:
Thanks for tackling this Bryan and for everyone else contributing - it needed fresh eyes.

bryevdv:
Yes, I would say no to `True`/`False` for `sizing_mode`, but keep the deprecation on `responsive`

bryevdv:
OK, changing "responsive" is going to be a bit of a task. I'd like to merge this, and do that separately.

bryevdv:
@bokeh/dev any objections to the plan for separate PRs?

birdsarah:
No!

birdsarah:
I never really liked stretch, and just had a mini-revelation: what about `fill_both` instead of `stretch_both`. To my mind, `fill` seems intentional, but `stretch` sounds bad. 

The correlating `fill_width` and `fill_height` would also still be correct names if we ever decide to implement them.

bryevdv:
after more discussion in issue, decided to keep "stretch". Merging.

Issue #4204, opened 2016-04-21
Bug Report? DatePicker widget errors on changing dates


jbrambleDC:
I am running a bokeh server app on 0.11.1. I am creating a DatePicker widget as follows:
```python
beginning = DatePicker(title="Begin Date", min_date=datetime(2014,11,1),
                       max_date=datetime.now(),
                       value=datetime(datetime.now().year,1,1)
                       )
```

I am running the app as follows:
`bokeh serve --show usage_app.py`

The DatePicker loads properly and displays as expected in the app. I have tested the default value, and it properly sets a default value of 2016-01-01, I have tried it using strings and datetime objects, both of which work. The issue is on changing dates. When I select a new date or try to change the month on the calendar display it immediately errors out and changes to a different, more primitive display.

<img width="321" alt="screen shot 2016-04-20 at 8 33 46 pm" src="https://cloud.githubusercontent.com/assets/6984748/14694543/4efce17a-0737-11e6-9275-1661b1c88b15.png">

In order to be exhaustive, I have tried declaring the object with and without mind and max dates. and I still reproduce the same error.

Here is the stack trace:

```
ERROR:bokeh.server.protocol.server_handler:error handling message Message 'PATCH-DOC' (revision 1): ValueError('expected a date, string or timestamp, got {}',)
DEBUG:bokeh.server.protocol.server_handler:  message header {u'msgid': u'6FA6441007BD454A9A19EB7098C41857', u'msgtype': u'PATCH-DOC'} content {u'references': [], u'events': [{u'new': {}, u'kind': u'ModelChanged', u'model': {u'type': u'DatePicker', u'id': u'538ca18f-efd5-4907-b770-c891f375f13a'}, u'attr': u'value'}]}
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/bokeh/server/protocol/server_handler.py", line 38, in handle
    work = yield handler(message, connection)
  File "/Library/Python/2.7/site-packages/tornado/gen.py", line 1008, in run
    value = future.result()
  File "/Library/Python/2.7/site-packages/tornado/concurrent.py", line 232, in result
    raise_exc_info(self._exc_info)
  File "/Library/Python/2.7/site-packages/tornado/gen.py", line 1017, in run
    yielded = self.gen.send(value)
  File "/usr/local/lib/python2.7/site-packages/bokeh/server/session.py", line 42, in _needs_document_lock_wrapper
[...]

PR #4489, opened 2016-06-13
functioning but poorly styled datepicker

bryevdv:
partially addresses #4204 

This PR provides a functioning (but unstyled) date picker. It is attached to an `<input>` and the input shows the currently selected date. If someone else can get the styling working, great. Otherwise at least it functions correctly. 

Note: I removed the code from the "render" method because it seems to be getting called alot, unnecessarily, and there is no reason to recreate the objects every time (also doing so cleared the input so that the date did not show)

### DIFF
--- a/bokehjs/src/coffee/models/widgets/date_picker.coffee
+++ b/bokehjs/src/coffee/models/widgets/date_picker.coffee
@@ -11,23 +11,19 @@ class DatePickerView extends Widget.View
 
   initialize: (options) ->
     super(options)
-    @render()
-
-  render: () ->
-    super()
-    @$el.empty()
-    $label = $('<label>').text(@mget("title"))
-    $datepicker = $("<div>").datepicker({
+    @label = $('<label>').text(@mget("title"))
+    @input = $('<input type="text">')
+    @datepicker = @input.datepicker({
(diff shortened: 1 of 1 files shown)


bryevdv:
My hunch is that to get the styling working we would have to fork/vendor jquery-ui. AFAICT there is no control over the css classes used on the DOM items created internally by the date picker function.

jbrambleDC:
This looks good to me. Just for clarification does the styling issue itself boil down to jquery-ui itself? forking could be a bit drastic, no? I see the jquery-ui project itself does not support github issues it seems. 

However I did notice this PR open related to DatePicker: https://github.com/jquery/jquery-ui/pull/1698

seems the purpose of this PR is to add change events and callbacks to the datepicker class, so may in all likelihood be related.

bryevdv:
@jbrambleDC I think it has to do with our CSS prefixing. 

@mattpap @birdsarah @mistakevin is everything is always under a `bk-root` element now, could that provide an avenue to stop prefixing altogether?

mistakevin:
Everything I've used seems to be namespaced under `.bk-root` in the latest dev build. I have not used every widget though, and there was an issue with the "save" modal which has been taken care of with an inline download option (which is great)

PR #4490, opened 2016-06-14
Canavandl/4372 log color mapper

### LINKED ISSUE #4372: LogColorMapper?
Bokeh provides a ColorMapper base class and a [LinearColorMapper](http://bokeh.pydata.org/en/0.11.1/docs/reference/models/mappers.html), but no logarithmic color mapper as far as I can see.  Can a logarithmic mapper be provided?  matplotlib provides LogNorm and SymLogNorm, and it would be very useful if similar functionality were available in Bokeh.

canavandl:
- [x] issues: fixes #4372 
- [x] tests added / passed (needs more)

Todo:
Currently doesn't handle instances where elements in image have value in the range (0, 1). Those pixels end up being black. Need to research what mpl does.

example code:
```python
import numpy as np

from bokeh.models import Range1d, LinearColorMapper, LogColorMapper

output_file('image.html')

N = 100

# scale only shows the spike.

image = Z1

mapper = LogColorMapper(palette="Spectral11")

p.image(image=[image], color_mapper=mapper,
        dh=[1.0], dw=[1.0], x=[0], y=[0])

show(p)
```

yields (sorry for the bad palette choice): 

screen shot 2016-06-14 at 1 15 02 pm

compared to LinearColorMapper:

screen shot 2016-06-13 at 7 25 38 pm


bokeh/models/glyphs.py
```diff
@@ -353,7 +353,7 @@ def __init__(self, **kwargs):
     images to have a gap between them, when they should appear flush.
     """)
 
-    color_mapper = Instance(LinearColorMapper, help="""
+    color_mapper = Instance(ColorMapper, help="""
```
jbednar: Looks like this was a mistake before; good catch.

bokeh/models/mappers.py
```diff
@@ -50,17 +50,47 @@ class LinearColorMapper(ColorMapper):
     this are clamped to ``high``.
     """)
 
-    # TODO: (jc) what is the color code for transparent?
-    # TODO: (bev) better docstring
-    reserve_color = Color("#ffffff", help="""
-    Used by Abstract Rendering.
```
jbednar: I think we can safely remove anything to do with abstract rendering, but maybe that should be a separate PR?

bryevdv: Agreed. And although it was added for AR it's potentially a useful feature in general. In any case removing properties should require a deprecation cycle at this point.

jbednar:
> Currently doesn't handle instances where elements in image have value in the range (0, 1). Those pixels end up being black. Need to research what mpl does.

datashader uses log1p to avoid the issue, i.e. adding 1 to the value before taking the log.  mpl's code is a little tricky to figure out.  In matplotlib/colors.py, the LogNorm class does:

```python
            resdat = result.data
            mask = result.mask
            if mask is np.ma.nomask:
                mask = (resdat <= 0)
            else:
                mask |= resdat <= 0
            cbook._putmask(resdat, mask, 1)
            np.log(resdat, resdat)
            resdat -= np.log(vmin)
            resdat /= (np.log(vmax) - np.log(vmin))
            result = np.ma.array(resdat, mask=mask, copy=False)
```

jbednar:
I'm not sure we need all the flexibility of changing the log base that's in there now.  Having lots of options probably actually hampers communication, since most people just want to say "logarithmic mapping -- check", not keep track of lots of details or have to communicate them to viewers of the data.  So I'd put in only one option for now unless there is pressure later to have multiple options and configurability for this feature.

jbednar:
It's a bit discouraging that adding a new and relatively simple subclass requires changes to 10 different files. :-(  3 of those are for tests, which is thus just mirroring the fact that changes were spread over multiple files elsewhere.  And of course any project that is split across languages is going to need code at both ends, plus potentially glue between them.  Still, one would hope that one wouldn't have to update tiny little bits spread over *so* many different locations!

bryevdv:
Well some of the work was to factor out code from a single class to a base class that could be shared. That's a one time cost

jbednar:
Maybe it would have been good to separate that into two PRs, then, one for the refactoring (no change to behavior) and one for the new functionality.

bokeh/models/mappers.py
canavandl: @bryevdv - do you want me to add it back or deprecate? I have no strong opinion either way.

bryevdv:
trailing whitespace: 
```
E         File contains trailing whitespace: bokeh/models/mappers.py, line 73.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: bokeh/models/mappers.py, line 73.'])
```
I think we will all need to get in the habit of running at least some of the tests locally before pushing.

bokeh/models/mappers.py
bryevdv: let's just keep it for now

jbednar:
Looks good to me now.

bryevdv:
Yup, will merge when green.

bryevdv:
> Maybe it would have been good to separate that into two PRs, then, one for the refactoring (no change to behavior) and one for the new functionality.

I agree that is probably a good ideal to strive for. I'm still learning the practice of patience :)

bryevdv:
spurious failures, I think, restarted the two tests

bryevdv:
OK it's the same test consistently failing... no idea what tho

bokeh/models/mappers.py
```diff
@@ -5,10 +5,9 @@
 
 from ..model import Model
 from ..core.properties import abstract
-from ..core.properties import Float, Color, Enum, Seq, Instance
+from ..core.properties import Float, Color, Enum, Seq
 from ..core.enums import Palette
 from .. import palettes
-from .transforms import Jitter
```
bryevdv: This is causing the test problem. The example does `from bokeh.models import Jitter` which was only "working: because this import was transitively available in `bokeh.models` @Maggie-M is going to make another issue/PR to fix, then you can merge that into this PR and it will pass.

bryevdv:
left line comment about test failure. Will have a fix PR to merge into this soon.

mattpap:
@canavandl, @bryevdv, according to MDN, `Math.log1p()` isn't available in IE (any version). A polyfill is needed.

jbednar:
@mattpap, thanks!

Issue #2268, opened 2015-05-11
DateRangeSlider not rendering


mvaz:
Hi all,

I'm trying to add a DateRangeSlider to a bokeh application.

``` Python
import logging
logging.basicConfig(level=logging.DEBUG)

from bokeh.properties import Instance
from bokeh.models import HBox, VBoxForm, Slider, DateRangeSlider
from bokeh.server.app import bokeh_app
from bokeh.server.utils.plugins import object_page

class SlidersApp(HBox):
    extra_generated_classes = [["SlidersApp", "SlidersApp", "HBox"]]

    inputs = Instance(VBoxForm)

    slider = Instance(Slider)
    dateRange = Instance(DateRangeSlider)

    @classmethod
    def create(cls):
        obj = cls(width=500)

        obj.slider = Slider(title="amplitude", name='slider', value=1.0, start=-5.0, end=5.0)

        start = '2010-01-01'
        end = '2015-01-01'
        obj.dateRange = DateRangeSlider(title="Period:", name="period", value=(start,end), bounds=(start,end), range=(dict(days=1), None))

        obj.inputs = VBoxForm(children=[obj.slider, obj.dateRange])

        obj.children.append(obj.inputs)

        return obj

@bokeh_app.route("/bokeh/sliders/")
@object_page("sliders")
def make_object():
    app = SlidersApp.create()
    return app
```

However, when rendering the obj above none of the components appear on the browser (Chrome v42, Safari v8.06, Firefox v34). In the error console, Safari complains of the following:

```
[Log] Bokeh: setting prefix to http://127.0.0.1:5007/ (bokeh.js, line 61288)
[Log] Bokeh: setting log level to: 'info' (bokeh.js, line 15080)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (jquery.min.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (util.js.map, line 0)
```

If it is not in the  obj.dateRange is added to the children, then the other slider draws as expected.

Any help appreciated.
Cheers, Miguel

bryevdv:
Hi @mvaz those errors are harmless and expected, the `.map` files are only used for development purposes. We have not found a way to silence those lines yet, though. what version of Bokeh are you using?

mvaz:
Hi @bryevdv, I'm currently using 0.8.2. Should I try with the devel branch?
Thought the info would be helpful... I don't see any other error messages anywhere.

[...]

PR #4491, opened 2016-06-14
working date slider for discussion

bryevdv:
addresses #2268 

Opening this PR for discussion of a better solution. The problem is the build. The jq date slider has "dynamic" dependencies that are just expected to be present. There are no `require` calls so our build did not pick them up. This PR has "working" date range slider, but I am sure there is a better way. 

ping @mattpap


bryevdv:
Here is a sample server app to test, note is has some layout issues:
```
from bokeh.models import DateRangeSlider, HBox, WidgetBox
from bokeh.io import curdoc

from datetime import datetime

start = '2010-01-01'
end = '2015-01-01'

beginning = DateRangeSlider(
    title="Period:", name="period", value=(start,end),
    bounds=(start,end), range=(dict(days=1), None))

def cb(attr, old, new):
  print(new)

beginning.on_change('value', cb)

curdoc().add_root(HBox(children=[WidgetBox(beginning)]))
```

jbrambleDC:
@bryevdv would this be used in lieu of the DatePicker? What are current plans for fixing the styling of the DatePicker?

bryevdv:
This PR as-is is not going to get merged. I am closing as rejected but will keep the branch for reference.

PR #4492, opened 2016-06-14
Tweaks to almar's resize PR

birdsarah:
See comments on #4473


bryevdv:
ping @almarklein

almarklein:
Looks good. But on Windows it appears the double-resize is really necessary. Will merge this now and then make it render twice...

PR #4493, opened 2016-06-14
Don't listenTo resize events if fixed.

### LINKED ISSUE #4488: Plots are rerendered on resize even if responsive="fixed"
This is a big performance regression. Nicely visible in e.g. `plotting/file/color_scatter`, where it spins CPU when I try to resize chrome's developer tools' panel. Going further, even if not `"fixed"`, then if no actual resize happened, there should be no rerendering.

birdsarah:
- [x] issues: fixes #4488
- [x] tests added / passed


bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -38,9 +38,14 @@ class LayoutDOMView extends BokehView
...
 
+    if @model.responsive is 'fixed'
+      # Render isn't triggered by resize - so trigger manually
+      @render()
+
   bind_bokeh_events: () ->
     @listenTo(@model, 'change', @render)
-    @listenTo(@model.document.solver(), 'resize', @render)
+    if @model.responsive isnt 'fixed'
+      @listenTo(@model.document.solver(), 'resize', @render)
```
mattpap: If `responsive` is changed, do we get this event handler rewired to get the expected behavior?

bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -38,9 +38,14 @@ class LayoutDOMView extends BokehView
       _.delay(resize, 100)
       _.delay(resize, 200)
 
+    if @model.responsive is 'fixed'
```
mattpap: Please use `==` and `!=` respectively, not those fancy operators.

birdsarah: I'd rather not. The correct javascript is `===`. Coffeescript compiles both `==` and `is` to `===`. At least this way I'm not polluting my muscle memory.

mattpap: I already know you would rather not, because I examined your recent contributions. Using two operators for one thing is one too many. We default to `==` and that's what you should use. When you use `is`, you actually increase confusion.

birdsarah: Sorry to have confused you.

This is why we need a style guide.

birdsarah:
@mattpap, i'm hoping to merge this within the next hour. It's a definite improvement. If it doesn't meet your performance needs or you're unhappy with the code I'm happy to make a follow-on PR.

bryevdv:
+1 on merging this

PR #4504, opened 2016-06-15
Add FuncTickFormatter for arbitrary python/javascript/coffeescript-based tick formatting

### LINKED ISSUE #4462: Adding a FuncFormatter for ticks based on PyScript
Further to the discussion in https://github.com/bokeh/bokeh/issues/1668 on more general ticking support I was hoping to start discussion around a more flexible solution by allowing simple Python functions to be transpiled with PyScript. This would allow for more complex ticking behavior to be expressed without having to implement custom models in most cases. Also pyscript is already used for CustomJS so it doesn't add another dependency.

In our case we just want to format a simple second counter in minutes without converting these relative time values to faked datetimes. This formatter can be expressed as a simple Python function:

```python
def formatter(seconds):
    m, s = divmod(seconds, 60)
    s = ('%d' % s) if s > 9 else ('0%s' % s)
    return "%d:%s" % (m, s)
```

which I've had to modify only slightly to transpile to Javascript correctly. @canavandl suggested implementing a CustomModelFactory, that will take a Python function and create a model from the transpiled function, would be doable. I was hoping this would be general enough to be included as part of bokeh.

Of course the formatter could also accept any JS function so it's not completely reliant on pyscript.

canavandl:
- [x] issues: fixes #4462
- [x] tests added / passed (needs tests)

example code:

```python
from bokeh.plotting import output_file, show
from bokeh.models import Plot, Range1d, LinearAxis, FuncTickFormatter

output_file('format.html')

def add_prefix(x):
    return "python_%d" % (x,)

left_formatter = FuncTickFormatter.from_py_func(add_prefix)

below_formatter = FuncTickFormatter(code="""
function (x) {
    return "javascript_" + x
};
""")

plot = Plot(plot_width=400, plot_height=400,
            x_range=Range1d(0,10), y_range=Range1d(0,10))
plot.add_layout(LinearAxis(formatter=left_formatter), 'left')
plot.add_layout(LinearAxis(formatter=below_formatter), 'below')

show(plot)

```

yields:
screen shot 2016-06-15 at 12 49 55 am


bryevdv:
The absolute most useful example we could make with this is nice lat/lon ticks for tile renderer plots.

jbednar:
This looks great to me, apart from the test failure.  @bryevdv, we're hoping to use it to choose to format a value stored in seconds but labelled in minutes, which seems very straightforward to do once this PR is applied.  @brendancol, for lat/lon, I guess one would just convert the values along each of the four edges separately, which would at least give some idea how the values change in between (depending on the projection)?

bokeh/models/formatters.py
```diff
@@ -200,6 +205,50 @@ class CategoricalTickFormatter(TickFormatter):
...
 
+class FuncTickFormatter(TickFormatter):
+    """ Format a function based on a python function that is transpiled to
+    javascript via PyScript
+    """
+
+    @classmethod
+    def from_py_func(cls, func):
+        """ Create a CustomJS instance from a Python function. The
+        function is translated to Python using PyScript.
```
jbednar: Translated to JavaScript, you mean (not Python)?

bokeh/models/formatters.py
```diff
@@ -200,6 +205,50 @@ class CategoricalTickFormatter(TickFormatter):
...
+        wrapped_code = "function (%s) {%sreturn formatter(%s)};" % (arg[0], code, arg[0])
+
+        return cls(code=wrapped_code, lang='javascript')
+
+    code = String(default="", help="""
+    An anonymous JavaScript/CoffeeScript function to reformat a single tick to
+    the desired format.
+
+    .. warning::
+        The anonymous can have only a single argument
```
jbednar: The anonymous function can?

mattpap:
@canavandl, make sure to update TypeScript typings. If unsure, ping me with questions.

bokeh/models/formatters.py
bryevdv: Also is "anonymous" needed? Will it inform or confuse more users?

jbednar: Probably better to just replace "anonymous" with "function" here.   Saying "anonymous" in the first sentence seems ok to me, though.

jbednar:
Looks good, thanks!

bokeh/models/formatters.py
```diff
@@ -200,6 +205,62 @@ class CategoricalTickFormatter(TickFormatter):
...
+    code = String(default="", help="""
+    An anonymous JavaScript or CoffeeScript function expression to reformat a
+    single tick to the desired format. E.g.:
+
+    code = '''
+    function (tick) {
+        // Convert from minutes to seconds and add units str
+        return tick * 60 + " seconds"
+    };
+    '''
```
bryevdv: probably need a `.. code-block:: python` here

PR #4511, opened 2016-06-15
Change Title.title_padding to Title.offset and Title.title_align to title.align

### LINKED ISSUE #4381: API issue: title_standoff, title_padding, title_offset
We used to have a property `title_standoff`, we now have a property `title_padding`. 

They don't do the same thing. In the case of an above title (which used to be the only option), `title_standoff` would move the title away from the plot, but `title_padding` moves the title along the axis of the text (to the right in the case of an above title).

To add to the confusion that's not what I would expect padding to do. I would expect it to pad all sides of a title. I would call it title_offset, especially as it's called x_offset and y_offset in the Label implementation.

I do think for title we need something like plot's min_border. where you can set it all or pick a side to add border/padding to.

ping @canavandl

birdsarah:
Fixes #4381

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -557,12 +557,12 @@ class Title(TextAnnotation):
     The text value to render.
     """)
 
-    title_align = Enum(TextAlign, default='left', help="""
+    align = Enum(TextAlign, default='left', help="""
     Location to align the title text.
     """
     )
 
-    title_padding = Float(default=0, help="""
+    offset = Float(default=0, help="""
     Offset value to apply to the title alignment coordinate.
     """)
 
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -649,13 +649,13 @@ class Plot(LayoutDOM):
 
     @property
     def title_standoff(self):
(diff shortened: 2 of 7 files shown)


almarklein:
LGTM. Though I am a bit confused about the use of prefixes in properties like these. I saw them on `legend` too (e.g. I added `legend_margin`), but here you remove that prefix. Do we have some sort of rule of thumb as to when we use a prefix and when not?

birdsarah:
> I saw them on legend too (e.g. I added legend_margin), but here you remove that prefix. 

Yes, I saw that too. It would be my preference to remove them.

> Do we have some sort of rule of thumb as to when we use a prefix and when not?

My assumption has been no, and that it's just arrived like this organically. But I could be wrong and there may be some magical doc somewhere that I really need to read!

almarklein:
I somehow assumed it was used in more places, but Legend and Title are the only classes where it seems to be used (at least within annotations.py). So after this PR, legend would be the only one. I agree that these prefixes should probably be removed.

bryevdv:
Well the old names were `title_text_font`, etc. because they were properties of *Plot*. But now that there is an actual `Title` object, the prefix is redundant.

bryevdv:
restarted what seemed to be a spurious test fail

Issue #4392, opened 2016-05-27
Layout - the clean-up


birdsarah:
PR #4343 got epic, so we want to merge as soon as possible. This issue collects a series of other issues that can be tackled as stand-alone PRs which clean-up either known bugs or follow-on features related to layout.

Note - this is a zenhub epic - please install zenhub plugin (free) to be able to see all the related issues: https://www.zenhub.io/

birdsarah:
MIGRATION NOTE

The following examples had to be changed in order to run:

```diff
examples/app/fourier_animated.py
+curdoc().add_root(layout)

examples/app/selection_histogram.py
+curdoc().add_root(layout)

examples/app/timeout.py
+curdoc().add(p)

examples/plotting/server/animated.py
-session.show()
+session.show(p) 

examples/plotting/server/geojson_points.py
-session.show()
+session.show(p) 

examples/plotting/server/line_animate.py
-session.show()
+session.show(p) 

examples/embed/embed_responsive_width_height.py
+ complete overhaul the js method is gone and you now use sizing_modes to achieve the same
```

For server there were two types of fixes due to auto-magic going away (but allowing use of figure or Figure).

For the embed_responsive_width_height. This is a complete change. If anyone was using the JS `get_size_for_available_space` method (introduced in 0.11) - then they're going to have to change their code. I don't think many people are using this, and I consulted with @almarklein to make all these changes and to make sure good alternatives are available. He may be able to help write a migration note for the removal of this method.

ping @bryevdv

bryevdv:
It looks like almost all the issues in this Epic are closed, I'm going to close this one, let's continue going forward in the the "Focus" epid for layout

PR #4512, opened 2016-06-15
A few fixes to the typings

almarklein:
I think this can be considered part of the layout work #4392

* Add new legend prop introduced in #4452 
* Remove `resize_width_height`
* Add `Document.resize` which should now be used in place of `resizie_width_height`

cc @mattpap

### DIFF
--- a/bokehjs/src/coffee/api/typings/document.d.ts
+++ b/bokehjs/src/coffee/api/typings/document.d.ts
@@ -12,6 +12,7 @@ declare namespace Bokeh {
         roots(): Array<Model>;
         add_root(model: Model): void;
         remove_root(model: Model): void;
+        resize(): void;
 
         title(): string;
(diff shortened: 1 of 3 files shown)


mattpap:
LGTM

almarklein:
BTW: I think there are still several discrepancies between the ts typings and the real API, even more so after the recent layout work. How hard would it be to write a script that at least tests that all classes and all public methods and properties are typed (only check that they are typed, without checking the actual types)?

mattpap:
@almarklein, the best approach would be to hook-in into TypeScript compiler, but that's a lot of work, I presume. Otherwise we are left with parsing `*.ts` files on our own. Additional issue is that some interfaces are broken into multiple (e.g. `Plot`), so we would have to also accommodate for that.

PR #4522, opened 2016-06-15
Collection of title fixes

birdsarah:
Fixes:

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1914,9 +1914,9 @@ class TitleProp(Either):
 
     def transform(self, value):
         if isinstance(value, str):
-            if value == "":
-                return None
-            else:
-                from bokeh.models.annotations import Title
-                value = Title(text=value)
+            from bokeh.models.annotations import Title
+            warn("""Setting Plot property 'title' using a string was deprecated in 0.12.0,
+            and will be removed. The title is now an object on Plot (which holds all of it's
+            styling properties). Please use Plot.title.text instead.""")
+            value = Title(text=value)
         return value
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -37,6 +37,9 @@ DEP_MSG_0_12_0 = """
     Plot property '%s' was deprecated in 0.12.0 and will be removed. Use '%s' instead.
     """
 
+# We create an empty title by default
+DEFAULT_TITLE = lambda: Title(text="")
+
 
 class Plot(LayoutDOM):
     """ Model representing a plot, containing glyphs, guides, annotations.
@@ -420,8 +423,8 @@ class Plot(LayoutDOM):
     Whether to use HiDPI mode when available.
     """)
 
-    title = TitleProp(help="""
-    A title for the plot. Can be a text string or a Title annotation.
+    title = TitleProp(default=DEFAULT_TITLE, help="""
+    A title for the plot. Can be a text string or a Title annotation. Default is Title(text="").
     """)
 
     title_location = Enum(Location, default="above", help="""
--- a/bokeh/models/tests/test_plots.py
+++ b/bokeh/models/tests/test_plots.py
@@ -156,6 +156,11 @@ def test_plot_raises_error_if_toolbar_and_tools_are_set():
         Plot(tools=[PanTool()], toolbar=Toolbar())
 
 
+def test_plot_with_no_title_specified_creates_an_empty_title():
+    plot = Plot()
+    assert plot.title.text == ""
+
+
 ############
 #
 # Test GridPlot


birdsarah:
ping @bryevdv - this is functionally done. I have some test clean-up to do, but you can start reviewing.

birdsarah:
@bryevdv - see the current visual difference generated by creating a default empty title:

https://s3.amazonaws.com/bokeh-travis/137919455/tests/pytest-report.html

It's live-able with, but we should open an issue to do better - and decide what better is.

birdsarah:
If you're ok with it, I'll update the screenshots - yay for visual testing!

birdsarah:
@bryevdv, I don't know how to fix the defaults mismatch, and would appreciate guidance: https://travis-ci.org/bokeh/bokeh/jobs/137946083

Here's the error from my local run:

```console
--------------------------------------------------------------------------- Captured stderr call ----------------------------------------------------------------------------
Bokeh: Bokeh does not support Polygons with holes in, only exterior ring used.
Bokeh: MultiPoint not supported in Bokeh
Bokeh: Bokeh does not support Polygons with holes in, only exterior ring used.
GMapPlot: defaults are out of sync between Python and CoffeeScript
    GMapPlot.title: coffee defaults to null but python defaults to {"attributes":{"background_fill_alpha":{"value":1},"background_fill_color":{"value":null},"border_line_alpha":{"value":1},"border_line_cap":"butt","border_line_color":{"value":null},"border_line_dash":[],"border_line_dash_offset":0,"border_line_join":"miter","border_line_width":{"value":1},"level":"annotation","name":null,"plot":null,"render_mode":"canvas","tags":[],"text":"","text_alpha":{"value":1},"text_color":{"value":"#444444"},"text_font":"helvetica","text_font_size":{"value":"10pt"},"text_font_style":"bold","title_align":"left","title_padding":0},"type":"Title"}
Bokeh: LogTickFormatter not configured with a ticker, using default base of 10 (labels will be incorrect if ticker base is not 10)
Python/Coffee matching defaults problems: 1
Bokeh: unrecognized side: 'null'
```

bryevdv:
@birdsarah you need to construct a default title on the JS side too

bryevdv:
compare: https://github.com/bokeh/bokeh/blob/birdsarah/title_fixes/bokehjs/src/coffee/models/tools/gestures/poly_select_tool.coffee#L67-L88

birdsarah:
cool, will do.

birdsarah:
@bryevdv - done. It's interesting that only GMapPlot (the subclass) failed and not Plot. I wonder if that means our test_defaults is still not catching some things.

birdsarah:
@bryevdv - are you good with the visual changes? https://s3.amazonaws.com/bokeh-travis/137919455/tests/pytest-report.html

bryevdv:
If we must, but couldn't the `Title` report it's dimension as 0 if `text` is empty or `null`?

birdsarah:
Yes, but I'm a bit loathed to fiddle around in the Title coffeescript at this late stage.

bryevdv:
It's a single `if` here? 

https://github.com/bokeh/bokeh/blob/29dbca514f67ec0e2b8fef4938378759b4a6324c/bokehjs/src/coffee/models/annotations/title.coffee#L54-L57

I think it's worth a try but it's your call.

birdsarah:
will - do! I got scared of the title rendering - but may be inappropriately!

birdsarah:
That worked @bryevdv - thanks! Sorry for being a pessimist.

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -14,6 +15,11 @@ PlotCanvas = require("./plot_canvas").Model
 class PlotView extends LayoutDOM.View
   className: "bk-plot-layout"
 
+  bind_bokeh_events: () ->
+    super()
+    title_msg = "Title object cannot be replaced. Try changing properties on title to update it after initialization."
+    @listenTo(@model, 'change:title', () => logger.warn(title_msg))
```
mattpap: This needs an explanation. Is it just a temporary issue?

bryevdv: Yes, this is temporary, and may be resolved by #4459 if it can get in in time, I need to try re-doing that PR on top of yours, though.

bokehjs/src/coffee/models/annotations/title.coffee
```diff
@@ -52,9 +52,13 @@ class TitleView extends TextAnnotation.View
       @_css_text(ctx, @mget('text'), sx, sy, angle)
 
   _get_size: () ->
-    ctx = @plot_view.canvas_view.ctx
-    @visuals.text.set_value(ctx)
-    return ctx.measureText(@mget('text')).ascent + 10
+    text = @model.text
+    if text == ""
```
bryevdv: This should probably also check for `null` to handle the `title.text = None` case

bokehjs/src/coffee/models/plots/plot.coffee
birdsarah: Just to note that this is not new to title, its them same fundamental problem as not being able to swap out an axis on the fly.

birdsarah: Also, there is a full user-facing message in bokeh/core/properties.py

Issue #4384, opened 2016-05-27
New layout API - bokeh.layouts - layout, row, column, gridplot


birdsarah:
In bokeh.layouts we currently have layout and gridplot.

The main goals of the layout function are:
1. make sure everything has the same responsive mode
2. provide a convenient spelling for building layouts.

Proposal:
* we add functions row and column that also auto apply the same responsive mode to all the children
* layout, row, column should accept arbitrarily nested children

birdsarah:
ping @bryevdv - the layout proposal is documented here.

birdsarah:
New layout functions were added in #4253.

Separate issue #4538 opened for allowing deeply nested input.

Issue #4388, opened 2016-05-27
Missing layout tests


birdsarah:
No tests were written for:
- models/layouts/box.coffee
- models/tools/toolbar_\* - especially _init_tools and _merge_tools

box.coffee is mostly @havocp's layout algorithm. If we don't get round to writing tests, we should consider having it as a vendored file to discourge people from editing it.

The toolbarbox code was copy-pasted from the old gridplot.

In addition the python-side `layouts.layout` and `layouts.gridplot` are untested - but that module still needs some work / discussion so waiting for it to settle before testing.

birdsarah:
This is definitely a very high priority but does not need to make it into 0.12 as it doesn't affect users.

bryevdv:
I am going to hijack this issue to add tests for things marked "needs test" in #8085

bryevdv:
The new baselines tests cover every example in minute detail

mattpap:
I will consider this completed. Tests for certain code paths, like toolbar and gridplot, were added in various PRs, and other are not relevant anymore.

PR #4523, opened 2016-06-15
Add new layout functions for row & column

birdsarah:
Partially fixes: #4384 

The other part of #4384 is for layout functions to accept arbitrarily nested lists. But this can wait till later. These functions should make it quick enough to spell out layouts.

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -10,7 +10,8 @@ from __future__ import absolute_import
 from .core.enums import Location, Responsive
 from .models.tools import ToolbarBox
 from .models.plots import Plot
-from .models.layouts import LayoutDOM, Row, Column, Spacer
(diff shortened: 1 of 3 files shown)


birdsarah:
Note this doesn't have tests, but I do have an open issue saying that we need them: #4388.

birdsarah:
Forgot to add a widgetbox function.

birdsarah:
ping @bryevdv - ok to merge?

birdsarah:
i need to merge this so can then fix up the sizing_mode pr

bryevdv:
looks OK at a quick glance

birdsarah:
I'm going to be updating these functions in the sizing_mode PR as they should behave differently when  we have responsive and sizing_mode, so we can take a closer look in that PR.

PR #4524, opened 2016-06-15
Undeprecate HBox and VBox and add docstrings

### LINKED ISSUE #4496: Undeprecate VBox and HBox
I would like names `VBox` and `HBox` to stay, because they are more meaningful to from UI point of view than the new counterparts.

canavandl:
- [ ] issues: fixes #4496 
- [ ] tests added / passed

Undeprecate HBox and VBox and add docstrings explaining behavior and return value (instances of Row and Column).

### DIFF
--- a/bokeh/models/layouts.py
+++ b/bokeh/models/layouts.py
@@ -202,23 +202,33 @@ class Column(Box):
     """
 
 
+def HBox(*args, **kwargs):
+    """ Lay out child components in a single horizontal row.
+
+    Children can be specified as positional arguments, as a single argument
+    that is a sequence, or using the ``children`` keyword argument.
+
+    Returns a Row instance.
+    """
+    return Row(*args, **kwargs)
+
+
+def VBox(*args, **kwargs):
+    """ Lay out child components in a single vertical row.
+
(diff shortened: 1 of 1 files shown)


birdsarah:
Interesting. I was imagining that if we're keeping them as synonyms then we should re-implement the classes, but they just inherit from Row/Column and don't do anything else. 

@mattpap, how were you thinking the undeprecation would be implemented? I'm fine with what luke's done if it meets your needs.

canavandl:
@birdsarah 's recommendation is probably a better solution. I'll make the change in the AM

bryevdv:
That would require subclassing or model subtyping which is even more obscure. Offhand I think my reference is for the simple functions.

bryevdv:
@mattpap do you have thoughts? we need to come to a decision.

bryevdv:
ping @mattpap we are intending to have an RC today. Do you have any input? AFAICT it meets your request but it would be good for you to confirm this.

mattpap:
@bryevdv, looks good.

bryevdv:
Thanks, merging now

Issue #4514, opened 2016-06-15
Split out responsive and sizing_mode as per discussion on #4484


birdsarah:
ping @bryevdv - not sure if you opened an issue for this.

Apologies if this is a dupe.

damianavila:
A link to the PR where this is solved would be nice, thanks!

birdsarah:
#4527

damianavila:
Thank you @birdsarah

PR #4527, opened 2016-06-16
make sizing_mode distinct from responsive

bryevdv:
issues: closes #4484


bryevdv:
ping @birdsarah

bryevdv:
@birdsarah I forgot to update the JS :) After doing so tho, I get three defaults mismatches. Not sure why they were not failing before? Anyway, what are the correct values on the JS side?
```
GMapPlot.toolbar: defaults are out of sync between Python and CoffeeScript
    GMapPlot.toolbar.sizing_mode: coffee defaults to null but python defaults to "stretch_both"
Bokeh: LogTickFormatter not configured with a ticker, using default base of 10 (labels will be incorrect if ticker base is not 10)
Spacer: defaults are out of sync between Python and CoffeeScript
    Spacer.sizing_mode: coffee defaults to null but python defaults to "stretch_both"
Spacer
[]
Toolbar: defaults are out of sync between Python and CoffeeScript
    Toolbar.sizing_mode: coffee defaults to null but python defaults to "stretch_both"
Toolbar
[]
ToolbarBox: defaults are out of sync between Python and CoffeeScript
    ToolbarBox.sizing_mode: coffee defaults to null but python defaults to "stretch_both"
ToolbarBox
[]
Python/Coffee matching defaults problems: 3
    1) match between Python and CoffeeScript
```

bryevdv:
Oh, I suspect it's because the `sizing_mode` property is now 
```
sizing_mode = Enum(SizingMode,...)
```
which makes the "first" of the enum the default unless specified I think. The default on bnoth sides should be "fixed"?

birdsarah:
>  The default on bnoth sides should be "fixed"?

Yes

bryevdv:
@birdsarah including spacer?

bryevdv:
OK, would appreciate eyes on this :)

birdsarah:
First reaction is yes - just for consistency. At the end of the day, every layout should have all matching responsive modes.

birdsarah:
Am just grabbing some food and then will take a look.

bryevdv:
I just assumed the point of a spacer was always to expand as much as possible

birdsarah:
That's not quite how layouts work....I'd like to write a full explanation....but I'm working on a bunch of different things right now. You can run this code and see that it does work (because the spacers will be type fixed).

```python
import numpy as np
from bokeh.io import show

N = 50

y = np.sin(x)

TOOLS = "pan,wheel_zoom,box_zoom,reset,save,crosshair"

l = figure(title="line", tools=TOOLS, plot_width=300, plot_height=100)
l.line(x,y)

aw = figure(title="annular wedge", title_location='right', tools=TOOLS, plot_width=100, plot_height=300)
aw.annular_wedge(x, y, 10, 20, 0.6, 4.1, inner_radius_units="screen", outer_radius_units="screen")

bez = figure(title="bezier", tools=TOOLS, plot_width=300, plot_height=300)
bez.bezier(x, y, x+0.4, y, x+0.1, y+0.2, x-0.1, y-0.2)

q = figure(title="quad", tools=TOOLS, plot_width=300, plot_height=300)
q.quad(x, x-0.2, y, y-0.2)

# specify "empty" grid cells with None
p = gridplot(children=[[None, l, None], [q, bez, aw]], responsive='fixed')
show(p)
```

bokeh/layouts.py
```diff
@@ -7,23 +7,23 @@
...
         children List(List(Instance(LayoutDOM))): An list of lists containing any of the
-        following: Plot, Widget, WidgetBox, Row, Column, ToolbarBox, Spacer. All tems in the grid
-        are then assigned the responsive mode of the layout.
+        following: Plot, Widget, WidgetBox, Row, Column, ToolbarBox, Spacer. All items
+        in the grid are then assigned the sizing mode of the layout.
 
-        responsive Enum(``box``, ``fixed``, ``scale_width``, ``scale_height``, ``scale_both``) :  How
-        the grid will respond to the html page. Default is ``box``.
+        sizing_mode ``"fixed"``, ``"scale_width"``, ``"scale_height"``, ``"scale_both"``, and
+        ``"stretch_both"``. Default is ``"stretch_both"``
```
birdsarah: I don't like the remove of the explanation of what sizing_mode does. This is one of the primary ways I expect people to use layout and I think more detail is needed. It may not have been a good sentence, but we definitely want a sentence or two.

bokeh/layouts.py
```diff
@@ -80,8 +80,8 @@ def gridplot(children=None, toolbar_location='left', responsive='fixed', toolbar
         toolbar will be located, with respect to the grid. If set to None,
         no toolbar will be attached to the grid.
 
-        responsive Enum(``box``, ``fixed``, ``scale_width``, ``scale_height``, ``scale_both``) :  How
-        the grid will respond to the html page. Default is ``fixed``.
+        sizing_mode ``"fixed"``, ``"scale_width"``, ``"scale_height"``, ``"scale_both"``, and
+        ``"stretch_both"``. Default is ``"stretch_both"``
```
birdsarah: as above

bokeh/models/widgets/widget.py
```diff
@@ -14,4 +14,4 @@
 class Widget(LayoutDOM):
     """ A base class for all interactive widget types. """
 
-    responsive = Override(default='fixed')
+    sizing_mode = Override(default='fixed')
```
birdsarah: This Override isn't necessary anymore - it was at some point in the past, but then everything went to fixed.

bokeh/plotting/tests/test_helpers.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import
+
+import pytest
+
+import bokeh.plotting.helpers as ph
```
birdsarah: No need to change this, but why not `from bokeh.plotting.helpers import _convert_responsive`?

bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -207,26 +207,26 @@ class LayoutDOM extends Model
...
 
   @define {
-      height:   [ p.Number, null ]
-      width:    [ p.Number, null ]
-      disabled: [ p.Bool, false ]
-      responsive: [ p.Responsive, null ]
+      height:      [ p.Number              ]
+      width:       [ p.Number              ]
+      disabled:    [ p.Bool,       false   ]
+      sizing_mode: [ p.SizingMode, "fixed" ]
```
birdsarah: hmmmm.....at some point during layout I decided to not set a default responsive/sizing_mode on the LayoutDOM but to Override it lower down the chain. This is why there's an override on Widget.

Was the change to move it up to to LayoutDOM deliberate?

If I recall correctly, which I may not be, the reason was because setting the default to LayoutDOM isn't really correct. There are a number of our LayoutDOM items that want to be in box/stretch_both mode for good reasons. In particular PlotCanvas and Toolbar spring to mind.

bokeh/models/widgets/widget.py
birdsarah: Cancel the above - see note below.

bokeh/models/layouts.py
```diff
@@ -40,10 +40,10 @@ class LayoutDOM(Model):
     the widget will be greyed-out, and not respond to UI events.
     """)
 
-    responsive = Responsive(help="""
-    The type of responsiveness for the item being displayed. Possible values are
-    ``"fixed"`` (or ``False``), ``"scale_width"`` (or ``True``),
-    ``"scale_height"``, ``"scale_both"``, ``"stretch_both"``.
+    sizing_mode = Enum(SizingMode, default="fixed", help="""
```
birdsarah: See note below. I think default should be `None` on LayoutDOM.

birdsarah: I know I said it should be fixed on everything - but I mispoke - I didn't mean at this high level.

birdsarah:
Made a number of comments, but it's not that much. I can help fix this in the morning.

birdsarah:
I misspoke about the sizing_mode default and hadn't understood all the implications until I read through all your code - sorry about that.

bokeh/plotting/tests/test_helpers.py
bryevdv: most test files test lots of functions in a single module under test, so I like to import the module in question with a short name, and use that throughout, rather than try to do individual imports from the module in question.

bokehjs/src/coffee/models/layouts/layout_dom.coffee
bryevdv: I changed it here to fix those defaults inconsistencies that cropped up. Also I don't understand what a `None` / `null` value is supposed to mean? I feel like it should not be allowed at all

bokeh/models/layouts.py
bryevdv: Is `None` really an acceptable value for an enum though?

birdsarah: We've talked this through and we're agreed on fixed for LayoutDOM default.

bokeh/models/widgets/widget.py
birdsarah: Double cancel :D

bokeh/plotting/tests/test_helpers.py
birdsarah: Thanks. Was just a learning moment question.

birdsarah:
@bryevdv are agreed on merging this, so we can get a dev build out.

damianavila:
@bryevdv this PR is referencing another PR #4484 as issue... I guess you reference it wrong, is that the case?

damianavila:
The correct reference seems to be #4514 (see last Sarah's comment there).

PR #4531, opened 2016-06-16
remove ResizeTool from set of default tools

### LINKED ISSUE #4520: Resize Tool Should Not Be In Defaults
There are too many cases in which the resize tool causes issues with the plot, toolbar, and widgets. Because of this, it would be best to remove it from the default toolbar in order to prevent the average user from encountering bugs. If they really want the resize tool, they can add it back it.

Also, there has been talk of removing the resize tool completely.

bryevdv:
issues: fixes #4520


mattpap:
@bryevdv, please update JS API as well (`api/plotting.coffee`).

bryevdv:
@mattpap thanks for the catch!

Issue #4530, opened 2016-06-16
Link to CSS colors in docs is broken, link address changed


cbcoutinho:
I was looking for all available colors in bokeh today when I came across the docs [here](http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#specifying-colors), where the following tries to go to w3schools.com; however, it is broken:

Specifically, this is broken:
   any of the [147 named CSS colors](http://www.w3schools.com/cssref/css_colornames.asp), e.g 'green', 'indigo'

I found the correct link to w3schools.com - it looks like they changed the link on their side.

http://www.w3schools.com/colors/colors_names.asp

bryevdv:
Thanks for the catch @cbcoutinho would you be able to submit a PR to fix this? This is the file to update:

https://github.com/bokeh/bokeh/blob/9491d517a84edc528ac84970cb338b5c24ac1d66/sphinx/source/docs/user_guide/styling.rst

cbcoutinho:
Ok I edited the file associated with the error. I'm not sure if I did the PR correctly - let me know if I need to change anything.

bryevdv:
PR was perfect, thanks

PR #4534, opened 2016-06-16
Update link to w3schools.com

cbcoutinho:
This PR fixes the hyperlink in [this file](https://github.com/bokeh/bokeh/blob/9491d517a84edc528ac84970cb338b5c24ac1d66/sphinx/source/docs/includes/colors.txt)

Associated with issue #4530


bryevdv:
Awesome thanks @cbcoutinho ! I am halting the tests and merging immediately since this is doc-only change.

PR #4542, opened 2016-06-16
fixes for recent changes

### LINKED ISSUE #4541: clustering app example needs updates for recent changes
Fails because `theme.yaml` tries to set `title_text_font_size` on `Plot` This bypasses the (python) property that deprecates this former `Plot` property, and tries to set a (Bokeh) property with that name directly on the plot. This fails, because of the work to make `Title` its own model.

Will fix up the `theme.yaml` and note this problem in migration guide. Since we barely demonstrated and not discussed the theming, hopefully this will not bite many people at all.

bryevdv:
issues: fixes #4541


bryevdv:
should note problem with setting plot title text properties in `theme.yaml`

PR #4546, opened 2016-06-16
Fix scale_both

### LINKED ISSUE #4537: scale_both isn't quite working right
After #4473, `scale_both` mode isn't quite working right. 

The plot is the correct size, but there's extra whitespace to the right or bottom of the plot.

The selenium test for this mode is currently set to xfail.

almarklein:
Fixes #4537

Still having some trouble to run Selenium locally, so trying like this first.


birdsarah:
Have been working on this with almar, have run the tests locally and its working, so am merging to kick off a dev build

PR #4549, opened 2016-06-16
Add viridis and related color palettes

### LINKED ISSUE #4548: Add viridis and other new color maps
Color maps from BIDS: https://github.com/BIDS/colormap/blob/master/colormaps.py

canavandl:
- [x] issues: fixes #4548

Add viridis, plasma, magma and inferno colormaps from matplotlib.

Discussion:
* Am i correctly citing their work? I had to derive the hexcolors from their rgb values and didn't need the mpl glue code, so none of their original code (which is [here](https://github.com/BIDS/colormap/blob/master/colormaps.py)) is actually included.


canavandl:
example code
```python
import numpy as np

from bokeh.models import Range1d, LinearColorMapper, LogColorMapper
# from bokeh.color_map import viridis

output_file('image.html')

N = 100

# scale only shows the spike.

image = Z1

mapper = LogColorMapper(palette="Viridis256")

p.image(image=[image],
        color_mapper=mapper,
        # palette='Spectral11',
        dh=[1.0],
        dw=[1.0],
        x=[0],
        y=[0])

show(p)
```

yields:
screen shot 2016-06-16 at 6 23 16 pm

bryevdv:
brilliant

canavandl:
One sec - for some reason Magma256 is returning the Viridis256 map. Let me investigate.

canavandl:
Magma256:

screen shot 2016-06-16 at 6 39 52 pm

Plasma256:

screen shot 2016-06-16 at 6 40 31 pm

Inferno256:

screen shot 2016-06-16 at 6 42 36 pm

bryevdv:
@canavandl I thought we had talked about including the "generator" function too? I.e. if you wanted a 64 color palette, you could call `viridis(64)` or something similar? How much effort would that be?

canavandl:
It should be super easy - i just had to think on it for a minute.

PR #4551, opened 2016-06-17
validation warning when sticky toolbars might overlap

### LINKED ISSUE #4550: Warn when sticky toolbars might visually overlap other components
Currently with `toolbar_sticky == True` the toolbar can potentially overlap with annotations on the sides of plots, e.g. axes. The default configuration is designed to avoid this possibility, and some cases of sticky toolbars and axes on the same side can be desirable (e.g., internal-only ticks). However, sticky toolbars and axes on the same side will often produce less than desirable results, and so to help users diagnose this outcome, we will issue a validation warning in cases when visual overlap is possible. 

Note: `toolbar_sticky` will like be renamed/changed.

bryevdv:
issues: closes #4550

### DIFF
--- a/bokeh/core/validation/check.py
+++ b/bokeh/core/validation/check.py
@@ -25,7 +25,7 @@ def check_integrity(models):
         logger.error("E-%d (%s): %s: %s" % msg)
 
     for msg in sorted(messages['warning']):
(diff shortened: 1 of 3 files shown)


bryevdv:
MIGRATION: note that `toolbar_sticky` is now the default

PR #4552, opened 2016-06-17
implement a scheme to be able to select active gesture tools programatically

### LINKED ISSUE #3054: Toolbar Improvements: Active is programmable

bryevdv:
- [x] issues: fixes #3054
- [x] tests added / passed

### DIFF
--- a/bokeh/charts/chart.py
+++ b/bokeh/charts/chart.py
@@ -30,7 +30,7 @@ from ..models import (
     Plot, HoverTool, FactorRange
 )
 from ..plotting import DEFAULT_TOOLS
-from ..plotting.helpers import _process_tools_arg, _glyph_function
+from ..plotting.helpers import _process_tools_arg, _glyph_function, _process_active_tools
 from ..core.properties import Auto, Either, Enum, String, Override
 from ..util.deprecate import deprecated
 from ..util._plot_arg_helpers import _convert_responsive
@@ -128,6 +128,10 @@ class Chart(Plot):
             kwargs['sizing_mode'] = _convert_responsive(kwargs['responsive'])
             del kwargs['responsive']
 
+        self._active_drag = kwargs.pop('active_drag', 'auto')
+        self._active_scroll = kwargs.pop('active_scroll', 'auto')
+        self._active_tap = kwargs.pop('active_tap', 'auto')
+
         title_text = kwargs.pop("title", None)
 
         super(Chart, self).__init__(*args, **kwargs)
@@ -156,7 +160,7 @@ class Chart(Plot):
         self._tooltips = []
 
         if hasattr(self, '_tools'):
-            self.create_tools(self._tools)
+            self.create_tools(self._tools, self._active_drag, self._active_scroll, self._active_tap)
 
     def add_renderers(self, builder, renderers):
         self.renderers += renderers
@@ -194,7 +198,7 @@ class Chart(Plot):
         if ygrid:
             self.make_grid(1, self._yaxis.ticker)
 
-    def create_tools(self, tools):
+    def create_tools(self, tools, active_drag, active_scroll, active_tap):
         """Create tools if given tools=True input.
 
         Only adds tools if given boolean and does not already have
@@ -209,8 +213,9 @@ class Chart(Plot):
 
         if len(self.toolbar.tools) == 0:
             # if no tools customization let's create the default tools
-            tool_objs = _process_tools_arg(self, tools)
+            tool_objs, tool_map = _process_tools_arg(self, tools)
             self.add_tools(*tool_objs)
+            _process_active_tools(self.toolbar, tool_map, self._active_drag, self._active_scroll, self._active_tap)
 
     def start_plot(self):
         """Add the axis, grids and tools
@@ -219,7 +224,7 @@ class Chart(Plot):
         self.create_grids(self._xgrid, self._ygrid)
 
         if self.toolbar.tools:
-            self.create_tools(self._tools)
+            self.create_tools(self._tools, self._active_drag, self._active_scroll, self._active_tap)
 
         if len(self._tooltips) > 0:
             self.add_tools(HoverTool(tooltips=self._tooltips))
--- a/bokeh/core/compat/bokeh_renderer.py
+++ b/bokeh/core/compat/bokeh_renderer.py
@@ -59,7 +59,7 @@ class BokehRenderer(Renderer):
     def close_figure(self, fig):
         "Complete the plot: add tools."
         # Add tools
-        tool_objs = _process_tools_arg(self.plot, DEFAULT_TOOLS)
+        tool_objs, tools_map = _process_tools_arg(self.plot, DEFAULT_TOOLS)
         self.plot.add_tools(*tool_objs)
 
         # Simple or Grid plot setup
--- a/bokeh/core/compat/bokeh_renderer.py
+++ b/bokeh/core/compat/bokeh_renderer.py
@@ -84,7 +84,7 @@ class BokehRenderer(Renderer):
 
                 _plot.title = ""
                 # and add new tools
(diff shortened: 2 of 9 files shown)


bryevdv:
ping @pzwang Is this enough for now, or do we need to build out more convenience? With this initial commit, you can do:
```
import numpy as np

N = 4000
colors = [
]

TOOLS="resize,crosshair,pan,wheel_zoom,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"

p = figure(tools=TOOLS)

p.scatter(x, y, radius=radii,
          line_color=None)

from bokeh.models.tools import *
p.toolbar.active_drag = p.select(type=LassoSelectTool)[0]
p.toolbar.active_tap = None

show(p)  # open a browser
```
yielding
screen shot 2016-06-16 at 11 13 46 pm

pzwang:
p.toolbar.active_drag = p.select(type=LassoSelectTool)[0]
    p.toolbar.active_tap = None

This all makes sense, but I can't help but notice the asymmetry in the initial tool specification list (strings) and the code it takes to activate a tool.

Maybe if `figure()` also accepted an `active_*` set of kwargs that could take strings?  I dunno...

bryevdv:
@pzwang sure I had kwargs to `Figure` and `Chart` in mind too. I don't want the property to accept strings because then all that logic for deciding what to do with the strings has to move to BokehJS and I'd like to avoid that. I guess my main question is, if this is the correct low-level interface, and will not be expected to change, would that be enough for now, adding the convenience later.

pzwang:
OK that makes sense.  In the docs, though, I would recommend showing creation of a tool instance, and then adding it to the plot, and setting it to active.  The `p.select()` approach should be shown as a separate/second approach for existing tools.

People tend to cut-and-paste, and syntax where we're grabbing `[0]` and stuff just seems kind of magical unless someone is familiar with the architecture of Bokeh.  So, a couple of extra lines of examples would be useful.

bryevdv:
OK, this now works:
```
TOOLS="crosshair,pan,wheel_zoom,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"

p = figure(tools=TOOLS, active_tap="pan")
```

bryevdv:
re: test on `plotting.helpers` I want to clean up/refactor/add tests as part of the pre-1.0 work as one task

bryevdv:
@bokeh/dev if anyone happens to have time for a quick review, it's appreciated. If not, I plan to merge soon.

birdsarah:
I always get confused on this, but something seems off on the visual hierarchy of the docs

screen shot 2016-06-17 at 4 30 37 pm

bryevdv:
Nice catch, I'll add a `Built-in Tools` header

bryevdv:
screen shot 2016-06-17 at 6 37 02 pm

birdsarah:
and remove the colon? :)

birdsarah:
i'm actively testing btw

bryevdv:
was just doing that

birdsarah:
Testing went well. I'm fine with merging as is, its a great step in the right direction. Couple of things:

* You have only implemented on Toolbar and not on ToolbarBox (which is the multi-plot toolbar - actually MultiPlotToolbar is a much better name - hmpf). Any particular reason? It wonder whether you could just share the code via ToolbarBase?

* I tried to updated the active_scroll property programatically using server and failed - did you expect that to work? Probably not but just checking.

* No JS tests, but the Toolbars aren't well tested at all so I'm not going to give you a hard time about it :D.

birdsarah:
I feel pretty sure you could share the code via ToolbarBase and have it available to both kinds of toolbar.

https://github.com/bokeh/bokeh/blob/4a74af2c9dbd996fec6d8857ae419d248f7ec543/bokehjs/src/coffee/models/tools/toolbar_box.coffee#L112-L113

Is exactly the same code you took out of `toolbar.coffee`. 

I can explain more about toolbar and toolbar_box if it's not clear.

birdsarah:
The reason to do all this is that then gridplots (or custom multiplot layouts) will also be able to use this functionality.

bryevdv:
I didn't do it because I didn't think it would work out of the box. If you think it is a simple thing to make it work in the grid case, we can try. 

AFAIK the primary use-case of this is as an initialization. I am not opposed to making it respond to changes but I'd rather make an issue and do it later.

birdsarah:
> AFAIK the primary use-case of this is as an initialization. I am not opposed to making it respond to changes but I'd rather make an issue and do it later. 

Agreed

> I didn't do it because I didn't think it would work out of the box.

You're right. Tried quickly but failed because I forgot we make proxy tools when we merge all the tools together.

birdsarah:
Added follow-on issues #4567 #4568.

LGTM.

tommycarstensen:
When I do:
```
from bokeh.models import WheelZoomTool
p = figure(tools=WheelZoomTool(), active_scroll='wheel_zoom')
```
then I get this `ValueError`:
```
  File "/Users/tc9/lib/python3.6/site-packages/bokeh/plotting/helpers.py", line 404, in _process_active_tools
    raise ValueError("Got unknown %r for 'active_scroll', which was not a string supplied in 'tools' argument" % active_scroll)
ValueError: Got unknown 'wheel_zoom' for 'active_scroll', which was not a string supplied in 'tools' argument
```

I'm using version 0.12.4. Is this a bug or am I using it incorrectly? Thanks.

Would you prefer, if I ask a question like this on the mailing list? Thanks.

tommycarstensen:
Please ignore my comment. The reason I had to do it this way, was because I was also using `HoverTool`. I solved it this way:

```
from bokeh.models import WheelZoomTool, HoverTool
hover = HoverTool(tooltips=[('y', '$y')])
wheel = WheelZoomTool(dimensions='width')
p = figure(tools=[hover, wheel], active_scroll=wheel)
```

PrateekArya:
> OK, this now works:
> 
> ```
> TOOLS="crosshair,pan,wheel_zoom,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"
> 
> p = figure(tools=TOOLS, active_tap="pan")
> ```

Hello!

I am using this approach to set `box_zoom` as active for a grid plot with multiple `figure`. It works, but there are 2 things which make me think that I am not using it the correct way:

1. I get the following warning:

`....py:208: UserWarning: found multiple competing values for 'toolbar.active_drag' property; using the latest value`

2. The box zoom icon in the tool box is not highlighted, but it is active!

Please let me know if I am doing something wrong here :)

PR #4557, opened 2016-06-17
Improvements to layout functions

birdsarah:
Don't have an issue on this yet. Have just been tweaking as I have been using layout and wanting to improve my convenience.

1) supply a plot_width and plot_height to griplot and it sets them all for you (great for a fixed layout with lots of plots and less typing) `gridplot([[p1, p2, p3, p4], [p5, p6, p7, p8]], plot_width=200, plot_height=200)`

2) supply a list of plots (and/or None's) to gridplot and a number of columns (as opposed to a list of lists) and it'll make the grid for you. `gridplot([p1, p2, p3, p4, p5, p6, p7, p8], ncols=4, plot_width=200, plot_height=200)`

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -193,9 +193,9 @@ def layout(children=None, sizing_mode='fixed', responsive=None, *args):
 
     # Make the grid
     rows = []
-    for row in children:
+    for r in children:
         row_children = []
-        for item in row:
+        for item in r:
             if isinstance(item, LayoutDOM):
                 item.sizing_mode = sizing_mode
                 row_children.append(item)
@@ -204,7 +204,7 @@ def layout(children=None, sizing_mode='fixed', responsive=None, *args):
                     """Only LayoutDOM items can be inserted into a layout.
                     Tried to insert: %s of type %s""" % (item, type(item))
                 )
-        rows.append(row(children=row, sizing_mode=sizing_mode))
+        rows.append(row(children=row_children, sizing_mode=sizing_mode))
     grid = column(children=rows, sizing_mode=sizing_mode)
     return grid
 
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -18,12 +18,15 @@ from .util._plot_arg_helpers import _convert_responsive
 #-----------------------------------------------------------------------------
 # Common helper functions
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


birdsarah:
ping @bryevdv this has lots of goodies in it, including the new glyphs notebook.

There are no test for bokeh.layouts (I have an open issue). As these functions have quickly grown, that's starting to feel like more of a priority.

bokeh/layouts.py
```diff
@@ -204,12 +207,21 @@ def layout(children=None, sizing_mode='fixed', responsive=None, *args):
...
+        rows.append(row(children=row_children, sizing_mode=sizing_mode))
     grid = column(children=rows, sizing_mode=sizing_mode)
     return grid
 
 
-def gridplot(children=None, toolbar_location='left', sizing_mode='fixed', responsive=None, toolbar_options=None, *args):
+def chunks(l, n):
+    """Yield successive n-sized chunks from list, l."""
+    for i in range(0, len(l), n):
+        yield l[i: i+n]
```
bryevdv: this is probably built into `itertools`

birdsarah: I didn't see it: https://docs.python.org/2/library/itertools.html

but may be missing something.

birdsarah: This is the SO post where I got that code: http://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks-in-python

bryevdv: color me surprised, I thought it was one of the "recipes" at least.

bokeh/layouts.py
```diff
@@ -252,7 +264,14 @@ def gridplot(children=None, toolbar_location='left', sizing_mode='fixed', respon
     if toolbar_location:
         if not hasattr(Location, toolbar_location):
             raise ValueError("Invalid value of toolbar_location: %s" % toolbar_location)
-    children = _handle_children(children, *args)
+    if ncols:
+        children = _handle_children(children, wrap=False, *args)
+        has_list = [isinstance(child, list) for child in children]
+        if sum(has_list) > 0:
```
bryevdv: `if any(isinstance(child, list) for child in children)` would be a little cleaner / no variable needed

bryevdv:
one small comment otherwise LGTM

PR #4565, opened 2016-06-17
Allow user to choose not to optimize on addConstraint

birdsarah:
Easy performance win when adding all the constraints in add_root.

Ref: #4395 

ping @sccolbert @bryevdv @mattpap 

I need to do some more benchmarking, but I wanted to get early eyes as modding kiwi.js is something I do with trepidation.  This isn't earth shattering but does speed things up.

### DIFF
--- a/bokehjs/src/coffee/core/layout/solver.coffee
+++ b/bokehjs/src/coffee/core/layout/solver.coffee
@@ -36,8 +36,8 @@ class Solver
     if trigger
       @trigger('layout_update')
 
-  add_constraint: (constraint) ->
-    @solver.addConstraint(constraint)
+  add_constraint: (constraint, optimize=true) ->
+    @solver.addConstraint(constraint, optimize)
 
   remove_constraint: (constraint) ->
     @solver.removeConstraint(constraint)
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -172,7 +172,7 @@ class Document
       @_solver.add_edit_variable(edit_variable, strength)
 
     for constraint in constraints
-      @_solver.add_constraint(constraint)
+      @_solver.add_constraint(constraint, false)
 
     if vars.width?
       @_solver.add_constraint(EQ(vars.width, @_doc_width))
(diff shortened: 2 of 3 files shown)


sccolbert:
A couple things:

1) Better to PR the vendor library than modify in-place. Otherwise any update to kiwi will blow away this change.

2) Do you have data which verifies a performance improvement? When writing kiwi, I explicitly *removed* the option to delay optimization specifically because things ran faster optimizing after each add. The reason (my hypothesis) for this, is that the solver has to do less total pivots when optimizing after each addition.

sccolbert:
Ha, I even made a comment to that effect:
https://github.com/bokeh/bokeh/pull/4565/files#diff-ffba01fc7b54d17c4089188d61d7506bL1458

bryevdv:
> Better to PR the vendor library than modify in-place. Otherwise any update to kiwi will blow away this change.

My understanding was that work on kiwi had ceased. :) There was also no npm package/release at the time, which is why it was vendored.

bryevdv:
also @sccolbert I can confirm at least in the case we were having the worst problems with, ~30 plots in a notebook, the qualitative speed-up is 2-3x at least. Before, it

birdsarah:
Thanks for taking the time @sccolbert 

> Better to PR the vendor library than modify in-place. Otherwise any update to kiwi will blow away this change.

Good point, I have already made one mod to kiwi.js - just to improve error reporting; I'll submit a PR to kiwi.

> Do you have data which verifies a performance improvement? When writing kiwi, I explicitly removed the option to delay optimization specifically because things ran faster optimizing after each add. The reason (my hypothesis) for this, is that the solver has to do less total pivots when optimizing after each addition.

Yes, I saw your comment, but still thought it was worth a try. I saw, very annectodally, ~15% speed up on some tests I ran, but definitely want to do a more thorough analysis. But I wanted to throw it out to start gathering thoughts.

In particular, our use case is that on initialization we add a few hundred constraints to the solver all in one batch. There are lots of other places where we add and remove constraints where I have not disabled the sequential _optimize. 

No need to comment further until I have some hard numbers. There have been a lot of moving parts lately which is why I broke this into a seperate PR.

Thanks again!

sccolbert:
A text typical system in Enaml (which uses the c++ solver) has thousands of
constraints. Im very interested in your profiling results when you have
them ready. Thanks!

Brian, development on kiwi hasn't ceased, it's just more or less feature
complete. The two remaining tasks are merging the py3k PRs and making the
JS version a first class citizen.

bryevdv:
There haven't been any perf related complaints, going to close this now, we can always re-visit later.

Issue #4559, opened 2016-06-17
ggplot API Update


HuntJSparra:
The recent update to ggplot API broke all previous code.

Should the examples be update to work with the new API?

birdsarah:
"The recent update to ggplot API broke all previous code."

Can you be a little more specific than "broke all previous code" - can you specify an example or examples that don't work. Most of bokeh is still working I'm assuming!

bryevdv:
ggplot made breaking API changes between `0.6.5` (what we test with from the bokeh conda channel) and some more recent version. I'd like to update to `0.9.4` the most recent version for testing, but we will need to update our compat examples to match (and pin the version to `0.6.5` for testing before uploading the new packages... I have PR but it's not working yet)

bryevdv:
PR to pin ggplot to `0.6.5` has been merged, will upload `0.9.4` packages now

bryevdv:
new packages are uploaded, next step is a PR to update the examples and unpin the ggplot version

PR #4566, opened 2016-06-17
pin ggplot version preceding uploading new versions to anaconda.org

bryevdv:
This is needed as part of #4559, otherwise when new ggplot pkgs are uploaded, the tests will break. The open should be removed when #4559 is completed.

### DIFF
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -9,7 +9,7 @@ MINICONDA="Miniconda-$MINICONDA_VERSION-Linux-x86_64"
 MINICONDA_URL="http://repo.continuum.io/miniconda/$MINICONDA.sh"
 
 PINNED_PKGS=$(cat <<EOF
-ggplot=0.6.5
+ggplot ==0.6.5
 EOF
 )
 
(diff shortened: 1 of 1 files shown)


bryevdv:
OK @bokeh/dev @damianavila I need help here. The build is failing with this:
```
Error:  Package missing in current linux-64 channels: 
  - ggplot=0.6.5
You can search for this package on anaconda.org with
    anaconda search -t conda ggplot=0.6.5
You may need to install the anaconda-client command line client with
    conda install anaconda-client
Note that you have pinned specs in /home/travis/miniconda/conda-meta/pinned:
    ['ggplot=0.6.5']
```
but the package is clearly available on the bokeh channel:
```
$ conda install ggplot=0.6.5
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata ...........
Solving package specifications: ..........

# All requested packages already installed.
# packages in environment at /Users/bryan/anaconda:
#
ggplot                    0.6.5                np19py34_0    bokeh
```

I thought maybe conda demanded that pinned packages be installed now, so I moved the channel add up, but that did not help either.

PR #4571, opened 2016-06-18
minimal server + notebook support

### LINKED ISSUE #3461: Notebook + Server Not Working

bryevdv:
- [x] issues: fixes #3461
- [x] tests added / passed

With this, the notebook can either be run with:

    bokeh serve --allow-websocket-origin=localhost:8888

in one terminal, and

    ipython server_animated.ipynb

and *Cell -> Run All*. This uses `bokeh.client` to connect the python code running in the Jupyter kernel to the server, the same as the examples in `examples/plotting/server`.

Alternatively, the app can be run directly in the server with:

    bokeh server server_animated.ipynb

Running this way, the server monkey patches all the relevant `bokeh.io` and `bokeh.client` to duo nothing, and runs the code from the notebook as a true app, directly in the server.

### DIFF
--- a/bokeh/application/handlers/tests/test_code.py
+++ b/bokeh/application/handlers/tests/test_code.py
@@ -1,6 +1,4 @@
-from __future__ import absolute_import, print_function
-
-import unittest
+import pytest
 
 from bokeh.application.handlers import CodeHandler
 from bokeh.document import Document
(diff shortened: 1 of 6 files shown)


bryevdv:
This is a bit hard to test until another dev build is out, I will push one later today. 

@pzwang @havocp @birdsarah @mattpap I would really appreciate eyes on this. 

@pzwang this is a minimal PR to expose some capability at all. I'd like to get some feedback, use, etc before trying to dive into making it easier or more convenient later.

bryevdv:
Also note to see the send the output inline in the notebook and to there server, there is currently a cell with both:
```
output_server()
output_notebook()
```
I'm loathe to dump a bunch of parameters on `output_notebook` so suggestions welcome. It seems simpler to add a single `notebook=True` to `output_server`, or have an explicit `output_notebook_server` or something.

bryevdv:
Other conveniences, a way to start the server on the right port. Unfortunately this will require BokehJS calling the Jupyter JS kernel API to execute python code as a string to set the port (from the window info) somewhere pre-defined in the kernel namespace. After checking with Jupyter folks, there is no better approach.

bryevdv:
OK, regarding removing the call to `push` that made it work in the notebook. This causes `hover.py` which *only* has `output_server` to fail. The issue seems to be having the `push` and also using the `session` methods at the same time. I can fix this with a check to see if notebook output is also enabled, but I'd be tempted to get rid of `output_server` and just rely on `session.show` etc in scripts.

bryevdv:
OK, I have updated a test. I tested regular and server notebooks, so I am confident that `push` was extraneous (at best) and that changing the test is appropriate

bryevdv:
also ping @jbednar re: dashboard meeting yesterday. I will try to make a short video describing how this could possibly be leveraged in a few days.

bryevdv:
@birdsarah I tried to add some better tests here to check the logging with `capsys` but it was not working.

bokeh/application/handlers/tests/test_code.py
```diff
@@ -36,69 +34,131 @@ class SomeModelInTestScript(Model):
...
+def test_script_uses_client():
+    doc = Document()
+    handler = CodeHandler(source=script_uses_client, filename="/test_filename")
+    handler._logger_text = "%s, %s"
+    handler.modify_document(doc)
+    if handler.failed:
+        raise RuntimeError(handler.error)
+    assert handler.failed is False
+    assert handler.error is None
+    assert len(doc.roots) == 1
```
bryevdv: wow this diff sucks. I just moved all the test methods to be plain funcs and got rid of unneeded `unittest` junk

bryevdv:
I re-orged the docs slightly, on extra heading for *Ways to Use* might need some IA help to split this up soon:
screen shot 2016-06-18 at 3 56 26 pm

havocp:
Seems like a straightforward approach to enable this, love it. Don't see any issues.

bryevdv:
After further testing I think this will have to wait a bit longer. It's more fragile that I realized. If was the case that a kernel restart always worked, perhaps it could go in, but after stopping a kernel in the middle of `loop_until_closed` the plot will now show up on a subsequent run.

bryevdv:
Closing, going to go with a different approach

PR #4576, opened 2016-06-19
General Examples Improvements for 0.12

### LINKED ISSUE #4397: Update all examples to use new layouts
Fairly self-explanatory. All the examples should be updated to remove HBox/VBox etc and to use the latest layout functions. Some should also include the new reponsive features..

ToDo:
- [ ] fix up spectrogram
- [ ] crossfilter
- [ ] surface3d (just needs to be added) 
- [ ] weather (some new pandas date serialization problem?!)
- [ ] embed/animated.py ("model not in session")
- [ ] embed/widget.py (buttons slightly covered, needs widgetbox?)

birdsarah:
A selection of improvements to examples for layouts or other improvements

Partially fixes #4397


bryevdv:
@birdsarah I'd like this to just be a general "examples fixes" I plan on going through all the examples again and just making improvements immediately as I go through.

birdsarah:
> @birdsarah I'd like this to just be a general "examples fixes" 

Everyone should feel free to push at will to this branch. In doing testing before cutting RC I saw lots of little improvements I want to make so looking forward to getting them all in.

bryevdv:
FYI I removed `timeout.py` there is periodic callbacks demonstrated in the much better new gapminder example. And `surface3d` demonstrates custom models better than `random_tiles` which just confused people, we

examples/app/selection_histogram.py
```diff
@@ -59,22 +66,18 @@
...
 vh1 = pv.quad(left=0, bottom=vedges[:-1], top=vedges[1:], right=vzeros, alpha=0.5, **LINE_ARGS)
 vh2 = pv.quad(left=0, bottom=vedges[:-1], top=vedges[1:], right=vzeros, alpha=0.1, **LINE_ARGS)
 
-pv.min_border_top = 80
-pv.min_border_left = 0
-ph.min_border_top = 10
-ph.min_border_right = 10
-p.min_border_right = 10
-layout = vplot(hplot(p, pv), hplot(ph, Paragraph(width=200)), width=800, height=800)
+layout = column(row(p, pv), row(ph, Spacer(width=200, height=200)))
```
birdsarah: Interesting. I'm surprised (and disappointed) the Spacer is necessary.

I was expecting layout([[p, pv],[ph]]) to work.

bryevdv: I don't know that it doesn't work, this was just the first thing I thought to try

bryevdv: It almost works! With the plot axis moved to the right, it aligns the canvas edge instead:
screen shot 2016-06-21 at 9 26 32 pm

bryevdv:
OK lots of cleanup and improvement to all the apps and all of `plotting/file` I would suggest merging this before rc2 and starting a new one after to continue.

bryevdv:
oops'ed a local `git commit --amend` on 40ab7a1, that actually fixes all the `VBoxForm` deprecations in apps as uses `figure` throughout.

bryevdv:
The only deprecations in apps now I think are `ColumnDataSource.from_df` which I actually think at this point we should un-deprecate... setting entire `.data` fields is a very useful thing so that is actually a needed function. Alternatively, making `_data_from_df` public and having that be the way.

Issue #4584, opened 2016-06-19
Set gridplot default location to above


birdsarah:
This is a stop-gap measure for #4579.

PR #4585, opened 2016-06-19
Set default gridplt toolbar position to above

birdsarah:
Fixes: #4584


birdsarah:
I'll fix the ts api in the open pr

birdsarah:
JS API has been updated with the same change in open JSAPI PR

Issue #4578, opened 2016-06-19
Gridplot toolbar broken


birdsarah:
When `above`, on `scale_width`, and on `fixed`

birdsarah:
It is now fixed for `scale_width` on master.

PR #4586, opened 2016-06-19
Fix toolbar display in fixed mode gridplot

birdsarah:
Fixes: #4578

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -315,9 +315,16 @@ def gridplot(
             toolbar_options = {}
         if 'toolbar_location' not in toolbar_options:
             toolbar_options['toolbar_location'] = toolbar_location
+
+        # Fixed sizing mode needs scale_width for the toolbar
+        # for layout to work correctly.
+        if sizing_mode == 'fixed':
+            toolbar_sizing_mode = 'scale_width'
+        else:
+            toolbar_sizing_mode = sizing_mode
         toolbar = ToolbarBox(
             tools=tools,
-            sizing_mode=sizing_mode,
+            sizing_mode=toolbar_sizing_mode,
             **toolbar_options
         )
 


birdsarah:
JS API has been updated with the same change in open JSAPI PR.

bokeh/layouts.py
```diff
@@ -315,9 +315,16 @@ def gridplot(
             toolbar_options = {}
         if 'toolbar_location' not in toolbar_options:
             toolbar_options['toolbar_location'] = toolbar_location
+
+        # Fixed sizing mode needs scale_width for the toolbar
+        # for layout to work correctly.
+        if sizing_mode == 'fixed':
+            toolbar_sizing_mode = 'scale_width'
```
bryevdv: wouldn't this be `scale_width` or `scale_height` depending on the location? top/bottom vs left right

birdsarah: Maybe technically, but it functionally doesn't matter.

`above` and `below` are what's broken for fixed, `left` and `right` work just fine because they get assigned a hard constraint of their width = 30px. In the above/below case, you want the width to be the width of the plot, so setting it to scale width allows it to grow. 

Things are happy to grow down a web page, they're just not happy to grow across it, which is why you have to force the width, but not the height.

Hope that makes sense.

birdsarah: Another thought is that this is a hack. It shouldn't be the final solution. What's there is the "minimum viable hack" which I think is the best way to do it.

zhouhbsea:
来自 魅蓝 note 2

-------- 原始邮件 --------
发件人：Sarah Bird <notifications@github.com>
时间：周日 6月19日 19:25
收件人：bokeh/bokeh <bokeh@noreply.github.com>
主题：[bokeh/bokeh] Fix toolbar display in fixed mode gridplot (#4586)

PR #4588, opened 2016-06-19
fix title props mismatch

### LINKED ISSUE #4587: Title JS models sending bck properties not on python side
`text_align` and `text_baseline` are set internally, resulting in messages about unknown properties on the server.

bryevdv:
issues: fixes #4587

make text_align and text_baseline internal and set up text @visuals by
hand. Also, protect the ctx change in the initializer with a
save/restore.

Side question: seems like JS tests should have complained about this before


bryevdv:
ping @canavandl @birdsarah just to be aware

PR #4590, opened 2016-06-19
Fix up spectrogram

### LINKED ISSUE #4589: Spectrogram example needs fixing up
There are some small CS changes to make but also the layout needs to be fixed as well

bryevdv:
issues: fixes #4589


bryevdv:
As of now the spectrogram runs, but the layout now looks bad:

screen shot 2016-06-19 at 1 26 47 pm

ping @birdsarah

birdsarah:
Cool. Should be straight forward to change. We'll probably switch from using components to just using a fixed layout and a custom template to embed some of our own css (which will probably only be a background). I can take a look next week.

birdsarah:
Feel free to open a new issue and assign me to it.

birdsarah:
Actually, better yet, I made a comment on #4397.

bryevdv:
I have a bokeh server version of the spectrogram in the works that is infinitely better, closing this.

PR #4596, opened 2016-06-20
Gapminder CSS Fix & Server Example

### LINKED ISSUE #4501: Gapminder CSS Issue
The CSS for the gapminder example is forcing the plot the display in the upper left corner:

screen shot 2016-06-14 at 12 15 13 pm

Setting the position of the plot to relative in the HTML fixes the issue, but unsure how to do that with CSS or as to why the CSS does not work in the Bokeh 0.12dev10

HuntJSparra:
Fixed CSS on Gapminder example so it displayed correctly and added a server example of Gapminder using python callbacks instead of CustomJS.

Resolves #4501

Have not touched the Jupyter Notebook example.


HuntJSparra:
ping @birdsarah

HuntJSparra:
EDIT: Moving app example to app directory

Issue #4606, opened 2016-06-21
toolbar alignment on complex layouts


birdsarah:
The use of constructor in LayoutDOM was changed in https://github.com/bokeh/bokeh/commit/56f45b5c8945dced5fe4e64ab5a69c17b553a08e (from constructor to initialize) by @mattpap.

However, constructor is necessary for toolbars to properly lay themselves out in complex layouts.

With initialize:

<img width="572" alt="screen shot 2016-06-20 at 11 59 27 pm" src="https://cloud.githubusercontent.com/assets/1796208/16220667/0f340334-3743-11e6-8654-4d01d7b0a632.png">

With constructor:

<img width="630" alt="screen shot 2016-06-21 at 12 00 43 am" src="https://cloud.githubusercontent.com/assets/1796208/16220689/382726fe-3743-11e6-8d46-486937b2050a.png">

Now.....it may be this is just a symptom of another problem to do with initialization, but for now, am going to change it back until we have a solution that lets us use initialize and allows toolbars to space themselves out correctly.

birdsarah:
However, using constructor instead of initialize causes the JS examples to break. 😭 

Would really appreciate your help @mattpap, as I bet you understand the initialization better than me.

If I'm remembering correctly, the problem is that the call to add_root which goes and picks up all the constraints on all the children happens before the toolbar has been initialialized and as a result it's variables don't get propogated up.

To test, try `examples/howto/layouts/dashboard.py` with either `stretch_both` or `scale_width`.

PR #4607, opened 2016-06-21
Ensure toolbars give themselves space in complex layouts

birdsarah:
Fixes: #4606


mattpap:
@birdsarah, did you verify that JS/TS API and `bokehjs/examples/*` still work?

birdsarah:
was just doing that! And no is the answer.

birdsarah:
So.....we have a problem :D

birdsarah:
Not going to merge this before we cut a release candidate, but we need to fix this before release. Any and all suggestions as to the root cause gratefully appreciated.

mattpap:
Most likely [1] is the cause. This hack is needed to make bokehjs work at all with `initialize` in `LayoutDOM`. Using `initialize` is the right approach (I can explain separately if this is requested). Most likely, the underlying issue is that layout (constraints) are computed way too early in a `Document` (btw. I smell a performance regression here). Alternatively, the mixture of private initialization across different types of toolbar classes may be the reason. Or a combination of both.

[1] https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/tools/toolbar_base.coffee#L98 
[2] https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/tools/toolbar_box.coffee#L142

birdsarah:
It's not [2] because this is not affecting ToolbarBoxes which are used for multiplot toolbars.

> Most likely, the underlying issue is that layout (constraints) are computed way too early in a Document 

Makes sense to me. Any thoughts on a solution?

>  This hack is needed to make bokehjs work at all with initialize in LayoutDOM. 

Right, so this makes me think that my understanding of the problem is correct. Toolbar._sizeable is set in plot initialization. But constraints are getting pulled up to document before all this initialization is complete.

> Using initialize is the right approach (I can explain separately if this is requested).

Would love that. I copied @havocp's original layout work in a couple of places and using constructor on the layout was one of them. But if it's not important to solving this problem then it can wait till later.

> (btw. I smell a performance regression here).

Would be very interested in details on this, but can wait till later I'm guessing.

I have one idea of how to solve this that may meet all our needs - which is to add the sizeable constraint at render time. I'll try that now.

Any other suggestions - like how initializing constraints later greatly appreciated.

birdsarah:
I think I got it. Pushing in a minute or so and would appreciate help.

birdsarah:
@mattpap, what do you think about this. JS examples are running nicely. And toolbars are spacing themselves. Am doing other tests now.

birdsarah:
@mattpap I'm going to merge this and cut the release candidate when the tests pass (~30 mins) if you have any more thoughts, now would be a great time.

mattpap:
@birdsarah, LGTM.

PR #4609, opened 2016-06-21
Remove hack unneeded since #4607 introduced in #4312

mattpap:
Remove leftover after #4607.


birdsarah:
Thanks for the cleanup @mattpap, was thinking that we should do that, but wanted to check with you first!

PR #4615, opened 2016-06-21
Execute range callbacks on reset_range (called by reset_tool)

### LINKED ISSUE #4614: Reset tool does not trigger x/y_range callback events
The reset tool obviously changes the extents of the axes, which should trigger any callback events attached to them but it does not appear to. I could swear there was an issue about this that was closed recently, but I can't find it now so I'm probably misremembering. I've tested it as part of 0.12rc1 and it still appears to be an issue there. This functionality is important for downsampling or any other callbacks that depend on the zoom level such as we make use of in datashader.

canavandl:
- [x] issues: fixes #4614
- [ ] tests added / passed (in progress)

ping @philippjfr for testing

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -357,11 +357,11 @@ class PlotCanvasView extends Renderer.View
     @pause
     if not range_info?
       for name, rng of @frame.get('x_ranges')
-        rng.get('callback')?.execute(rng)
         rng.reset()
+        rng.get('callback')?.execute(rng)
       for name, rng of @frame.get('y_ranges')
-        rng.get('callback')?.execute(rng)
         rng.reset()
+        rng.get('callback')?.execute(rng)
       @update_dataranges()
     else
       range_info_iter = []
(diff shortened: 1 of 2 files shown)


bryevdv:
Should be possible to add JS tests for this

canavandl:
You think it's better to write a selenium test or stub the `execute` callback method and assert it's called?

philippjfr:
Just tested it in datashader and holoviews, works well for me.

philippjfr:
Actually, shouldn't the callback happen after the reset so the ranges are adjusted? Currently my callbacks only update along one of the ranges and I can't figure out if that's due to throttling of the second event or due to the order of resetting.

canavandl:
Yes! Looking at when the callback is executed on a range change event (i.e. via panning), the callback is executed after updating the range data. I'll update the reset callback to match.

philippjfr:
Just switched myself, can confirm that fixes it.

canavandl:
I added a selenium test to check that a CustomJS alert is triggered on clicking the `ResetTool`. Should be a good starting point for additional reset tool testing. Should be merge-able when tests pass.

bryevdv:
Well, I was actually thinking of a gulp unit test, but this probably is ok too

PR #4625, opened 2016-06-22
Revert performance regression "fix" from #4493

### LINKED ISSUE #4611: Plots disappear after zooming in/out in some browsers
Toolbar is still visible though. To make plots appear on screen again, one has to press reset tool button, or reload the page. This happens across all examples with simple and complicated layouts. Reproducible at least in FF and Chrome.

mattpap:
This fixes #4611, but reverts #4493, so #4488 performance regression is still there. I shouldn't be using the word *fixes* at all right now, because I don't fully understand what is going on in the rendering pipeline.

### DIFF
--- a/bokehjs/src/coffee/models/layouts/layout_dom.coffee
+++ b/bokehjs/src/coffee/models/layouts/layout_dom.coffee
@@ -56,11 +56,7 @@ class LayoutDOMView extends BokehView
 
   bind_bokeh_events: () ->
     @listenTo(@model, 'change', @render)
-
-    if @model.sizing_mode == 'fixed'
-      @listenToOnce(@model.document.solver(), 'resize', @render)
-    else
-      @listenTo(@model.document.solver(), 'resize', @render)
+    @listenTo(@model.document.solver(), 'resize', () => @render())
 
     # Note: `sizing_mode` update is not supported because changing the
     # sizing_mode mode necessitates stripping out all the relevant constraints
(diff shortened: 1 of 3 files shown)


bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -56,11 +56,7 @@ class LayoutDOMView extends BokehView
 
   bind_bokeh_events: () ->
     @listenTo(@model, 'change', @render)
-
-    if @model.sizing_mode == 'fixed'
-      @listenToOnce(@model.document.solver(), 'resize', @render)
-    else
-      @listenTo(@model.document.solver(), 'resize', @render)
+    @listenTo(@model.document.solver(), 'resize', () => @render())
```
mattpap: @bokeh/dev, make sure to use `() => @fn()` instead of just `@fn`, if `fn` is not an event handler. `render` is not. This way you make sure that no random arguments will be passed to `fn`. This makes code understanding and refactoring easier. Also, not that important, but `printf`-style debugging is much more convenient this way.

bryevdv: Let's make a small style guide a priority after this release. Not something huge and comprehensive but, at least that collects a few important bits of wisdom like this (e.g. I was not aware of this).

bryevdv:
OK this LGTM for a short term fix

PR #4638, opened 2016-06-22
Keep fixed listen to and solve zoom issue

birdsarah:
Fixes: #4611 #4488


bryevdv:
This seems reasonable at a glance! Requesting a render on a pixel ratio change seems safe and proper to do in general.

birdsarah:
I have only tested this on one browser on OSX. So could do with more testing to make sure it actually resolves the zoom issue over a wider set of browsers.

But yes, as you said, it does seem like a good thing to do anyway.

birdsarah:
ping @mattpap

birdsarah:
OK, fixed my vimrc so hopefully this happens less.

PR #4641, opened 2016-06-22
Should not have a toolbarbox warning when making a grid plot

### LINKED ISSUE #4583: Using gridplot now gives a warning
ToolbarBox must inherit from Box coffeescript side so that everything behaves correctly.

For the defaults tests to pass, that means it must also inherit from Box python side.

But that leads to the following warning when running a correctly functioning gridplot

WARNING:/Users/caged/Dev/bokeh/bokeh/bokeh/core/validation/check.py:W-1002 (EMPTY_LAYOUT): Layout has no children: ToolbarBox, ViewModel:ToolbarBox, ref _id: 1386

ping @bryevdv

birdsarah:
Fixes #4583

### DIFF
--- a/bokeh/models/tests/test_tools.py
+++ b/bokeh/models/tests/test_tools.py
@@ -1,5 +1,8 @@
-from __future__ import  absolute_import
+from __future__ import absolute_import
 
+import mock
+
+from bokeh.core.validation import check_integrity
 from bokeh.models.layouts import Box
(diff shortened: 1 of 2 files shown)


birdsarah:
ping @bryevdv, am happy with this. was much harder to write the test than to write the code!

bryevdv:
LGTM

PR #4644, opened 2016-06-22
Update scale on mainGlGlyph obj and not self

### LINKED ISSUE #4633: TapTool on Line glyph with webgl=True causes renderers to disappears
code to reproduce:

```python
from bokeh.models import TapTool

output_file('line.html')

p = figure(tools='', webgl=True)
p.line(x=[1,2,3], y=[1,2,3], line_width=5, legend='x0')
p.add_tools(TapTool())

show(p)
```

On clicking the line w/ the TapTool, the line and axes disappear.

Raises a `line.js:345 Uncaught RangeError: Invalid typed array length` exception within the LineGLGlyph.prototype._update_scale method.

canavandl:
- [x] issues: fixes #4633

I'm not entirely sure of the consequences of the changes. It seems that the computed attributes were being set on the wrong object (self and not the mainGlGlyph arg), so I just changed it.

The webgl examples still seem to work... 

ping @almarklein 

Discussion:
* Any good strategies for preventing regressions beyond a selenium test?

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/webgl/line.coffee
+++ b/bokehjs/src/coffee/models/glyphs/webgl/line.coffee
@@ -670,7 +670,7 @@ class LineGLGlyph extends BaseGLGlyph
 
       # Do we need to re-calculate segment data and cumsum?
       if Math.abs(@_scale_aspect - (sy / sx)) > Math.abs(1e-3 * @_scale_aspect)
-        @_update_scale(sx, sy)
+        mainGlGlyph.update_scale(sx, sy)
         @_scale_aspect = sy / sx
 
       # Select buffers from main glyph
@@ -692,6 +692,7 @@ class LineGLGlyph extends BaseGLGlyph
       @prog.set_uniform('u_scale_aspect', 'vec2', [sx, sy])
       @prog.set_uniform('u_scale_length', 'float', [scale_length])
 
+      @I_triangles = mainGlGlyph.I_triangles
       if @I_triangles.length < 65535
         # Data is small enough to draw in one pass
         @index_buffer.set_size(@I_triangles.length*2)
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/models/glyphs/webgl/line.coffee
```diff
@@ -670,7 +670,7 @@ class LineGLGlyph extends BaseGLGlyph
 
       # Do we need to re-calculate segment data and cumsum?
       if Math.abs(@_scale_aspect - (sy / sx)) > Math.abs(1e-3 * @_scale_aspect)
-        @_update_scale(sx, sy)
+        mainGlGlyph.update_scale(sx, sy)
         @_scale_aspect = sy / sx
```
almarklein: I think this should also use `mainGlGlyph._scale_aspect`. BTW IMO it's ok to use private props of the mainGlGlyph object since its the same class.

bokehjs/src/coffee/models/glyphs/webgl/line.coffee
```diff
@@ -692,6 +692,7 @@ class LineGLGlyph extends BaseGLGlyph
       @prog.set_uniform('u_scale_aspect', 'vec2', [sx, sy])
       @prog.set_uniform('u_scale_length', 'float', [scale_length])
 
+      @I_triangles = mainGlGlyph.I_triangles
```
almarklein: Right. because lines can only be selected as a whole, i.e. the `indices` arg is always the complete line.

almarklein:
Except small comment this looks good. Testing is hard, because we have yet to find a CI service that can render WebGL ... maybe we need to think of  a series of tests that can be run locally from time to time.

PR #4645, opened 2016-06-22
Fix LOD downsampling

### LINKED ISSUE #4622: LOD downsampling is not working
pan `color_scatter.py` e.g

almarklein:
Fixes #4622


bryevdv:
Ah, of course, `Plot` is basically a container now... but I thought these properties were still on `Plot`?

birdsarah:
> Ah, of course, Plot is basically a container now... but I thought these properties were still on Plot? 

The properties are on Plot, which is what plot_model.plot is referring to. They are no longer on PlotCanvas (because you removed them). I believe that @plot_model is always referring to the PlotCanvas object.

bryevdv:
Oh I see, that's confusing. Well, still intend to do more cleanup around all that design (later)

bryevdv:
LGTM

birdsarah:
yes, I think renaming `plot_model -> plot_canvas` and `plot_view -> plot_canvas_view` would make things clearer and should be straight forward (although tedious)

birdsarah:
@almarklein, some pretty gnarly test errors - does this work locally and we just need to fix JS tests. I can help with that if you need me to.

almarklein:
It works fine locally. Weird. Could there be a difference in initialization in the browser vs in the tests?

birdsarah:
> Could there be a difference in initialization in the browser vs in the tests?

Yes, there's a big difference, these are unit tests, so you need to do a decent amount of self setup.

birdsarah:
The tests are now passing wrt to the LOD. They're actually still failing sporadically but merging  https://github.com/bokeh/bokeh/pull/4655 should help.

Issue #4601, opened 2016-06-20
Toggle button layout broken w/ new widgetbox


bgyarfas:
Spacing and layout of Toggle seem broken.
Running off of 0.12.0dev13. 
```
from bokeh.models import Button, Toggle
from bokeh.plotting import curdoc
from bokeh.layouts import widgetbox

button_1 = Button(label="Regular Button")

button_2 = Toggle(label="Toggle Button")

button_3 = Button(label="Regular Button")

curdoc().add_root(widgetbox([button_1, button_2, button_3]))
```

birdsarah:
Thanks for sharing. Can you post a screenshot? I can definitely imagine this is a bug though. @bryevdv this is probably Toggle not having enough height -  guessing without seeing a screenshot though.

birdsarah:
also, probably won't change anything @bgyarfas but can you reconfirm with the RC1 release that just came out last night.

bgyarfas:
@birdsarah Yes, this was with RC1. See the screenshot below.
![image](https://cloud.githubusercontent.com/assets/6708431/16266403/ab4bd6a6-3852-11e6-99eb-2795ed486f98.png)

birdsarah:
awesome thanks - I didn't realize this one was you too @bgyarfas - we're going to have to appoint you widget czar!

Iceman121:
DataTable still renders weirdly.

![image](https://cloud.githubusercontent.com/assets/16691044/24633494/6e0bc3e8-18fb-11e7-92dc-08f749481514.png)

Any thoughts on how to rectify that?

Issue #4636, opened 2016-06-22
TextInput not obeying width parameter


bgyarfas:
Trying to force a `TextInput` to a certain width does not work as it does with Sliders/MultiSelect/etc. (see last comment in #3346): 
```
from bokeh.models import (
    Slider,
    TextInput)
from bokeh.plotting import  curdoc
from bokeh.layouts import column
curdoc().add_root(column(children=[Slider(width=150)]))
```

I found a workaround to the above by adding `@$el.find('input').width(@mget('width'))` @ [https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/text_input.coffee#L31](url)

Also, a `Slider` with a specified `width` will not be the correct size in Chrome (51.0.2704.103 ) until the browser window is resized.
![image](https://cloud.githubusercontent.com/assets/6708431/16278076/838d8494-3883-11e6-9bea-d946e8e30441.png)
![image](https://cloud.githubusercontent.com/assets/6708431/16278089/8c68c4d4-3883-11e6-81c8-9954e825775f.png)
This behavior is not present on Safari, but _only on the first load within a tab/page_. On subsequent refreshes, with or without cache, the behavior is the same as in Chrome.

bgyarfas:
If using a `widgetbox` then `curdoc().add_root(widgetbox(children=[Slider(width=150)]))` will not obey the `width` parameter. But, if you use `curdoc().add_root(widgetbox(children=[Slider()], width=150))` then it will, at least with the same buggy behavior described above.

bgyarfas:
Buttons will also not obey the width parameter as well when in a `widgetbox`, although the `widgetbox` will obey. Running the following snippet:
```
button_1 = Button(label="Button 1")
button_2 = Button(label="Button 2")
curdoc().add_root(widgetbox(children=[button_1, button_2], width=150))
```
results in the Buttons resizing on browser window size change, but not to the correct size.
Before browser window size change:
![image](https://cloud.githubusercontent.com/assets/6708431/16279347/9e6798c2-3888-11e6-9631-f3d115b20dac.png)

**After:**
![image](https://cloud.githubusercontent.com/assets/6708431/16279335/8f274196-3888-11e6-91fe-fe330c0629d6.png)
Notice the width/height properties of the `widget-box` and `widget` get updated on a size change. But the Button doesn't, the Button is not the right size in these images (they're ~75px).
Putting the `width` param in the `Button` initialization does not work:
```
[...]

Issue #4404, opened 2016-05-28
Tabs only work with fixed layout.


birdsarah:

audiolion:
I ran into this issue, my graphs were stored under different bootstrap tabbed panes, the graph would initially render very small (the bokeh js responsive code declared a min height of 100 so it defaulted to that).

If you hit the `reset` widget button on the graph it would correct it. I used a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) on the  `<div class="tab-pane">` that was being updated and clicked the reset button.

```js
  document.addEventListener("DOMContentLoaded", function() {
    var target = document.getElementById('graphs');

    var callback = function(e) {
      var resetBtn = document.querySelector('.bk-btn-icon.bk-tool-icon-reset');
      resetBtn.click();
    }

    var observer = new MutationObserver(function(mutations) {
      callback();
      this.disconnect();
    });

    var config = { attributes: true };
    observer.observe(target, config);
  });
```

This is vanilla js to accomplish it, the `target` is the `<div class="tab-pane"...>` on the first mutation we click the reset then disconnect the observer so it doesn't fire a bunch of extra times.

omearac:
+1 for this since I desperately need this bug fix

verybadsoldier:
Would be great to have this working. :)

bryevdv:
```
import numpy as np

from bokeh.models import ColumnDataSource, Plot, LinearAxis, Grid, HoverTool
from bokeh.models.widgets import Paragraph
from bokeh.models.layouts import Column, Tabs, Panel
from bokeh.models.glyphs import (
    AnnularWedge, Annulus, Arc, Bezier, Circle, Dash, ImageURL, Line, MultiLine, MultiPolygons, Oval, Hex,
    Patch, Patches, Quad, Quadratic, Ray, Rect, Segment, Square, Text, Wedge, CircleX, Triangle,
    Cross, Diamond, InvertedTriangle, SquareX, Asterisk, SquareCross, DiamondCross, CircleCross, X
)
from bokeh.document import Document
from bokeh.embed import file_html
from bokeh.resources import INLINE
from bokeh.util.browser import view

N = 9

x = np.linspace(-2, 2, N)
y = x**2
sizes = np.linspace(10, 20, N)

xpts = np.array([-.09, -.12, .0, .12, .09])
ypts = np.array([-.1, .02, .1, .02, -.1])

xs = [ xpts + xx for xx in x ]
[...]

PR #4651, opened 2016-06-23
Render widgets and widget box correctly

birdsarah:
This PR takes care of some clean-up that should have been done before, and fixes a bug in widgetbox rendering. Fixes: #4601, #4636, #3346

1) All widgets should not pollute their widget wrapper. The widget wrapper looks something like this `<div id="1234" class="bk-widget bk-layout-fixed">` and may have a width/height/x/y position. All html and classes related to the widget should be inside this box for predictable widget results. I did some of this during the layout PR, but forgot to go back and clean-up the others.  (I'm leaving data_table as an exception to this as it's very complicated and I wouldn't want to break it, and I think it'll be ok).
2) I have fixed some rendering issues with the WidgetBox. Some of this was due to bugs introduced by the initialization PR, I also now set a sensible default for a fixed widgetbox (which has some implications, but is overall a good thing).

There's also some small clean-up of the `buttons_server.py` example, which could have been in a different PR, but snuck in here.

### DIFF
--- a/bokeh/models/widgets/buttons.py
+++ b/bokeh/models/widgets/buttons.py
@@ -60,6 +60,7 @@ class AbstractButton(Widget, ButtonLike):
     A callback to run in the browser whenever the button is activated.
     """)
 
+
 class Button(AbstractButton):
     """ A click button.
 
@@ -68,8 +69,6 @@ class Button(AbstractButton):
     clicks = Int(0, help="""
     A private property used to trigger ``on_click`` event handler.
     """)
-
-    height = Override(default=45)
 
     def on_click(self, handler):
         """ Set up a handler for button clicks.
@@ -82,6 +81,7 @@ class Button(AbstractButton):
 
         """
         self.on_change('clicks', lambda attr, old, new: handler())
+
 
 class Toggle(AbstractButton):
     """ A two-state toggle button.
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -58,6 +58,7 @@ class TextInput(InputWidget):
     widget by hitting Enter or clicking outside of the text box area.
     """)
 
+
 class AutocompleteInput(TextInput):
     """ Single-line input widget with auto-completion. """
 
@@ -66,7 +67,6 @@ class AutocompleteInput(TextInput):
     user upon typing the beginning of a desired value.
     """)
 
-    height = Override(default=65)
 
 class Select(InputWidget):
     """ Single-select widget.
@@ -85,8 +85,6 @@ class Select(InputWidget):
     A callback to run in the browser whenever the current Select dropdown
     value changes.
     """)
-
-    height = Override(default=65)
 
     @classmethod
     def create(self, *args, **kwargs):
--- a/bokehjs/src/coffee/models/layouts/widget_box.coffee
+++ b/bokehjs/src/coffee/models/layouts/widget_box.coffee
@@ -15,51 +15,90 @@ class WidgetBoxView extends LayoutDOM.View
   className: "bk-widget-box"
 
   initialize: (options) ->
-    super()
+    super(options)
+    if not @model.document._unrendered_widgetboxes?
+      @model.document._unrendered_widgetboxes = {}  # poor man's set
+    @model.document._unrendered_widgetboxes[@id] = true
     @render()
 
   bind_bokeh_events: () ->
     super()
-    @listenTo(@model, 'change:children', @build_child_views)
+    @listenTo(@model, 'change:children', () => @build_child_views())
 
   render: () ->
-    super()
 
-    # Go through and make rendering tweaks because of margin
-    # TODO(bird) Make this configurable & less flaky
+    s = @model.document.solver()
 
-    if @model.sizing_mode is 'scale_width'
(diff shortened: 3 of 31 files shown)


birdsarah:
@bryevdv, this PR is not as bad as it looks, and I'm happy to do a pair review to talk you through it.

It does leave a known corner case: https://github.com/bokeh/bokeh/issues/4653

I think it really improves the widget layout experience. In particular, if things don't auto-magically line-up there's much better support for just setting a width/height to be what you want it to be.

@bgyarfas if you're able to take a look at this and see if we're doing better compared to your attempts with widgets yesterday, that'd be great.

birdsarah:
I've restarted the unit tests a couple of times, am being plagued by the flaky solver stub problem in the JS tests.

bryevdv:
@philippjfr can you evaluate this quickly wrt holoviews? I'd like to merge it but not without an assessment from you first

philippjfr:
Just went through all our examples and encountered two small regressions, which weren't there yesterday (in 0.12.0rc2).

Here is a gridplot containing Tabs, which has shifted downwards and now overlaps a little bit:

image

And here are two Tables placed side-by-side, also now overlapping:

image

Otherwise it seems to work okay. Apologies I can't send small self-contained examples but these are generated from within HoloViews. Happy to confirm whether they are working if you come up with a fix though.

bryevdv:
how does holoviews construct those? Maybe there is a better way

bryevdv:
For the tables, I think they have a default size of 600 now, which is a compromise for other fixes. Maybe setting the size on the table explicitly would help with that issue

philippjfr:
>For the tables, I think they have a default size of 600 now, which is a compromise for other fixes. Maybe setting the size on the table explicitly would help with that issue.

Our default Table width is set to 400, but they overlap no matter what I set the width to.

philippjfr:
>how does holoviews construct those? Maybe there is a better way

HoloViews at least for now constructs all it's layouts by supplying list of lists to ``gridplot``. We are looking to improve that in future, now that bokeh supports more powerful layouts, but it seems like ``gridplot`` should just work.

bryevdv:
@bokeh/dev if the Holoviews issues can be identified and are small, I propose this as last PR to be considered and then to evaluate an RC3 with it. The fixes here are substantial enough to warrant consideration in this otherwise quiet period.

bgyarfas:
Probably tangential, but I'm finding myself wanting to put in `Spacer` in between widgets in the context of `row` or `column`. I realize this is probably just equivalent to being able to set the `padding` in a  `bk-grid-row` or `-column` if the children widget-boxes would inherit that parameter. For clarity, I'm just thinking in the context of `sizing_mode='fixed'`. 
What are your thoughts on being able to implement something like this?

birdsarah:
@bgyarfas - I would use spacers for now and open an issue to discuss your use case. generally speaking things don't cascade in the bokeh layout, its a calculated, effectively absolute positioned, set of layouts.

birdsarah:
we're pretty far out of expected use cases here, but wrapping things in widget boxes and specifying widths solves all the overlapping stuff for me. 

The alignment of a plot next to a tabbed set of plots in a grid plot is not expected to work. It had never occurred to me that someone would try to do this.

birdsarah:
> The alignment of a plot next to a tabbed set of plots in a grid plot is not expected to work. It had never occurred to me that someone would try to do this.

I mean I think it would be cool if it did. But there are other layout issues with tabs that we'd already punted to after 0.12.0 so I would put this in that bracket too: https://github.com/bokeh/bokeh/issues/4404

bryevdv:
OK, @philippjfr with the info above do you think you can make simple changes in Holoview to accommodate? If so, I'd say we proceed with this PR

philippjfr:
>we're pretty far out of expected use cases here, but wrapping things in widget boxes and specifying widths solves all the overlapping stuff for me.

Yes, I've solved it too now, had to write a crazy expression that will unpack the lists of lists and get the width for each element in the plot. Doing that I can add ~10% padding to ``Tabs`` and ``Tables``, which solves the overlaps. Thanks a lot for your help!

>The alignment of a plot next to a tabbed set of plots in a grid plot is not expected to work. It had never occurred to me that someone would try to do this.

Someone will always try something crazy and in holoviews that plot can be expressed with just this:

```python
%%opts Overlay [tabs=True width=600 height=600]
x,y = np.mgrid[-50:51, -50:51] * 0.1
img = hv.Image(np.sin(x**2+y**2), bounds=(-1,-1,1,1))
img.relabel('Low') * img.clone(img.data*2).relabel('High') + img
```

which means the barrier to doing crazy things like this is low. If we can get that to work at some point I'd be very happy but definitely shouldn't hold up this release.

bryevdv:
OK, let me check this PR out locally and do some testing. I expect to merge it in a few hours

philippjfr:
>OK, @philippjfr with the info above do you think you can make simple changes in Holoview to accommodate? If so, I'd say we proceed with this PR

Yes, please proceed. Didn't know ``WidgetBox`` would be required for padding but I can make it all work now. Either way it's a huge improvement, couldn't even put a ``DataTable`` in a gridplot before.

bryevdv:
OK tested all apps, several file and notebook examples, and JS api examples. All look good. Will merge this PR and cleanup PR and then cut an RC

PR #4654, opened 2016-06-23
Examples tweaks

birdsarah:
This is branched of #4651 

Tweaks a few examples.


birdsarah:
This is ready for merge whenever, but not marking at such because the other PR isn't merged.

PR #4655, opened 2016-06-23
Skip 3 more flaky JS tests

birdsarah:
Flaky JS tests are really slowing things down.

I have an open issue to go back and really figure out what's going on. All seem to be the same problem of a poorly stubbed Solver.


birdsarah:
@bryevdv - definitely recommend merging and then merging into other outstanding PRs

PR #4659, opened 2016-06-23
Feature/functickformatter args

### LINKED ISSUE #4621: Add `args` parameter to `FuncTickFormatter` similiar to `CustomJS`

canavandl:
Will be merged into 0.12.1 or 1.0

- [x] issues: fixes #4621
- [x] tests added / passed

Adds support for bokeh models as args to FuncTickFormatter.


bokehjs/test/models/formatters/func_tick_formatter.coffee
```diff
@@ -12,6 +13,17 @@ describe "func_tick_formatter module", ->
...
+  it "should handle args appropriately", ->
+    rng = new Range1d()
+
+    # checks that args doesn't cause undeclared var error
+    obj = new formatter.Model
+      code: "function (x) {return x*10};"
+      args: {foo: rng}
+
+    labels = obj.doFormat([-10, -0.1, 0, 0.1, 10])
+    expect(labels).to.deep.equal([-100, -1.0, 0, 1, 100])
```
bryevdv: maybe configure the range with a start/end and include that in the computation?

bryevdv:
@mattpap can you take a look at this? It is a completely isolated change to a new feature. If you think this is good, I'd consider including it, it would be very helpful for some work using the Geo features.

bryevdv:
ping @canavandl do you have an CoffeeScript examples that work? I can't get any CS version I have tried to work .

canavandl:
Missing a line break for the coffeescript lang - the generated code is:
```javascript
["tick", "var func = (x) -> return "javascript_" + xreturn func(tick)"]
```
 Writing fix now.

bryevdv:
@canavandl not sure. ping @mattpap

canavandl:
ping @mattpap 

This is the PR regarded compiling the FuncTickFormatter func from coffeescript that we discussed. Please let me know when you implement the server-side transpiling so that I can finish this PR.

canavandl:
Need to modify this line to load the coffeescript compiler for coffeescript FuncTickFormatters:
https://github.com/bokeh/bokeh/blob/master/bokeh/embed.py#L180

canavandl:
Now the compiler loads for:

```python
from bokeh.models import FuncTickFormatter

output_file("formatter.html")

p = figure(plot_width=500, plot_height=500)
p.circle([0, 2, 4, 6, 8, 10], [6, 2, 4, 10, 8, 0], size=30)

p.yaxis.formatter = FuncTickFormatter(
    lang="coffeescript",
    code="(tick) ->  return Math.floor(tick) + ' + ' + (tick % 1).toFixed(2)"
)

show(p)
```

bryevdv:
@canavandl I'm a bit hazy about the need for the `(tick) ->` in the `code` arg. Can we not make it more like `CustomJS` where the function signature is created for the user? What should be provided if the user does give `args`? Do they have to coordinate the function sig in `code`' manually?

canavandl:
The issue with generating the function signature is that it's unclear (to me) how to have it align with the formatter's doFormat method, where we're passing in a positional argument called `tick`. 

```coffeescript
doFormat: (ticks) ->
  return (@get('func')(tick, @get('values')...) for tick in ticks) # the 'values' are the new args
```

If we generate the function signature from a code block like CustomJS, it's unclear how we'd handle passing the `tick` argument in (require that users have a variable called tick?). 

The CustomJS callbacks don't have to align with any other function signature, so it's a bit simpler:

```coffeescript
execute: (cb_obj, cb_data) ->
  @get('func')(@get('values')..., cb_obj, cb_data, require)
```

bryevdv:
If we generate the function signature, we an just *add* tick to the signature (exactly like `cb_data` and `cb_obj` are done above). In that case `tick` is just implicitly available (exactly like `cb_data` and `cb_obj`). That that's something for us to clearly document, but it seems as though the same framework should be feasible?

bryevdv:
For reference this is where the signature is constructed:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/callbacks/customjs.coffee#L42

So I'd expect correspondingly 
```
new Function(_.keys(@get("args"))..., "tick", "require", code)
```
and then
```
doFormat: (ticks) ->
  return (@get('func')(@get('values')..., tick) for tick in ticks)
```
am I missing something?

canavandl:
That wouldn't work because you have to have a return value from the UDF that the `doFormat` function returns.

canavandl:
Thinking about it, doing the 'tick' namespacing automagic would also break flexx transpilation support because the variable names don't get converted directly.

bryevdv:
> That wouldn't work because you have to have a return value from the UDF that the doFormat function returns.

Why couldn't the code block have a return? This is what I had in mind:

```
p.yaxis.formatter = FuncTickFormatter(lang="coffeescript",
    code="return Math.floor(tick) + ' + ' + (tick % 1).toFixed(2)"
)
```
Asking sincerely, maybe there are technical issues I am not aware?

> Thinking about it, doing the 'tick' namespacing automagic would also break flexx transpilation support because the variable names don't get converted directly.

If that's true I don't see how it works in the `CustomJS` case, too.

bryevdv:
E.g., the example here has a function body with a `return`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function

brendancol:
@bryevdv I would think a `return` statement would only be relevant inside a function.  Why not prescribe a function signature and make the user implement it?  Then inputs/outputs are clear.

birdsarah:
> E.g., the first example here has a function body with a return

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function

I'm not understanding you @bryevdv are you refererring to:

```js
var adder = new Function('a', 'b', 'return a + b');
```

bryevdv:
@brendancol because user errors in supplying the signature are easy to make, but hard to debug. Also it's different from `CustomJS`, I'd like the experience to be identical to `CustomJS`

bryevdv:
@birdsarah yes, it seems perfectly possible have a return in a "programmatically" created function. So it then seems that:
```
p.yaxis.formatter = FuncTickFormatter(lang="coffeescript",
    code="return Math.floor(tick) + ' + ' + (tick % 1).toFixed(2)"
)
```
Should be fine?

birdsarah:
Well the two are different, the example from MDN is using a function constructor. 

But I don't see a problem with what you're proposing.

bryevdv:
I feel like I am communicating really badly. :) What I am proposing:

* *both* `CustomJS` and `FuncTickFormatter` accept a *function body*, on the *python* side

* *both* `CustomJS` and `FuncTickFormatter` pass that function body to the `Function` constructor, on the *JavaScript* side

To reiterate, this is exactly how CustomJS currently works. A function body (not a full function definition) is provided on the python side, and that is passed to `Function` on the JS side. Specifically, `CusotmJS` does this:
```
new Function(_.keys(@get("args"))..., "cb_obj", "cb_data", "require", code)
```
here: 

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/callbacks/customjs.coffee#L42

birdsarah:
@Bryan fwiw I agree with you. In particular "there's magic no matter what, let's have magic as consistent as possible with CustomJS"

mattpap:
@bryevdv, @canavandl, make sure to merge this after https://github.com/bokeh/bokeh/pull/4946.

mattpap:
@canavandl, I updated this PR with `mattpap/precompiler`.

canavandl:
What do you propose doing for coffeescript support? Writing a function body doesn't work because a return statement outside a function causes the compilation to fail.

@bryevdv

bryevdv:
> Writing a function body doesn't work because a return statement outside a function causes the compilation to fail. 

Well that's annoying. What about constructing the enclosing function header (in all cases JS and CS) and compiling that? Then BokehJS could be simplified considerably (we would be sending real complete functions after all) 

If we do that tho I'd like update `CustomJS to operate in the same way`

canavandl:
I managed to wrap the coffeescript function body in a function, compile that to JS, then add a line to call and return the wrapped function.

bryevdv:
I think we would need to compute the function signature (add all the `args`) as well?

canavandl:
fixed ^

canavandl:
```
$ from bokeh.models import FuncTickFormatter, Slider
$ s1 = Slider()
$ s2 = Slider()
$ formatter = FuncTickFormatter.from_coffeescript(code="return tick+10", args={'s1':s1, 's2':s2})
$ formatter.code
> 'var formatter;\n\nformatter = function(s2, s1) {\n  return tick + 10;\n};\n\nreturn formatter(s2, s1)'

**Comment by canavandl** [comment]
The current status is:
* JS - we're accepting a function body and args that assumes `tick` (the unmodified tick label value) and the `args` are in the function namespace.
* Coffeescript - we're accepting a function body and args analogous to  JS. To do this, we're wrapping the code input in a function, transpiling that to JS, then adding a JS snippet to call the wrapped code.
  * essentially the output looks like:
```javascript
var formatter;
formatter = function() { //FUNCTION BODY HERE };
return formatter()"
```

* Python - we're accepting a function object, transpiling it to JS, then adding a JS snippet to call it.

Issue #4663, opened 2016-06-24
Dropdown callbacks broken


bryevdv:
@birdsarah I am pretty sure this commit https://github.com/bokeh/bokeh/commit/89b030bfa1911bb3e6046fc3e911d63013b08695#diff-4b3dcdd7895736bf67d737896bedca07 broke dropdowns. Run `models/transform_interpolator.py` or any other dropdown example. The action is delayed by one click: i.e:

* change from "none" to "linear" --> nothing happens
* change from "linear" to "step before" --> NOW it shows linear interp
* change from "step before" to anything else --> NOW is shows "step before"

PR #4664, opened 2016-06-24
Fix widget calls in wrong order

birdsarah:
Fixes: #4663 

I started using super - and put it in the wrong place - in the recent widgets PR.


birdsarah:
Merging ahead of RC4

PR #4671, opened 2016-06-24
Pretty-up gapminder example

birdsarah:
Just some aesthetic choices and some streamlining

### DIFF
--- a/examples/app/gapminder/main.py
+++ b/examples/app/gapminder/main.py
@@ -42,7 +42,7 @@ cr = plot.circle(x='fertility', y='life', size='population', source=sources[year
 
 plot.add_tools(HoverTool(tooltips="@index", renderers=[cr]))
 
-# This draws a custom legend.
+# this draws a custom legend
 tx, ty = 6.3, 95
 for i, region in enumerate(regions):
     plot.add_glyph(Text(x=tx, y=ty, text=[region], text_font_size='10pt', text_color='#666666'))
--- a/examples/app/gapminder/main.py
+++ b/examples/app/gapminder/main.py
@@ -1,7 +1,7 @@
 import pandas as pd
 
 from bokeh.io import curdoc
-from bokeh.layouts import row, column, widgetbox
+from bokeh.layouts import layout, widgetbox
 from bokeh.models import (ColumnDataSource, HoverTool, Text, Div, Circle,
                           SingleIntervalTicker, Slider, Button, Label)
 from bokeh.palettes import Spectral6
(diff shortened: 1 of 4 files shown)


bryevdv:
@birdsarah I actually don't like most of these changes. I specifically wanted to show how you could embed things like YT videos easily to make very rich SPA's. I also liked the more compact narrow vertical nature, having the right edge line up all the way down. Being *completely* nit-picky, I make it a point to write very short comments *not* as sentences:
```
# this draws a custom legend
```
was intentional with no initial cap or trailing period :) 

Things I would keep: button style switch, plot axes titles

HuntJSparra:
I like the changes. It was sort of strange having the slider and button farther right than anything and not keeping the narrow, vertical form. Also, without the labels I forgot and did not know what the actual data I  was looking at was. I also love how the new button is visually more responsive. Regarding removing the YouTube, the focus of the example is the gapminder recreation and embedding a Bokeh plot. Removing the video brings the focus on the graph (previously small on the page), as it should be imo. Before this, I do not believe we had an embedded app example that did not require launching both a python AND bokeh server AND then running the python. This is a great example of embedding a server without having to jump through so many hoops. If someone were wanting to add a video, it is a HTML and not a Bokeh task. My only nitpick is I miss having the slider cancel the animation when moved.

bryevdv:
I appreciate there are positives, e.g. demonstrating responsiveness, but again my express intent was to show of an `index.html` with real things in it.

bryevdv:
Safari OSX, video is truncated:

screen shot 2016-06-26 at 3 11 31 pm

birdsarah:
The cut off  video was happening for me, but then I updated the css in the index.html. Can you confirm?

> but again my express intent was to show off an index.html with real things in it

I don't understand your comment on this. At this point, my only change to the `index.html` that is currently checked in is to change the width/height properties on the iframe, and to not use `<br/>` tags (which is not good practice). I agree that more could be done, and I would suggest not using the bokeh `Div` widget but just using templated html.

bryevdv:
Also why did `curdoc().title` get removed? I went though all the apps and added that, so they would say something beside "Bokeh Application" on browser tabs

birdsarah:
> The cut off video was happening for me, but then I updated the css in the index.html. Can you confirm?

Never mind, seeing it too. Looking into it.

bryevdv:
> Can you confirm?

Git says I am up to date

> I don't understand your comment on this.

I wanted to show an `index.html` that had *stuff in it*, besides just the Bokeh elements. Without the video the `index.html` is hardly different from the default template built into Bokeh. I want to demonstrate how the template complements and lets the Bokeh parts be augmented with a template.

birdsarah:
> I wanted to show an index.html that had stuff in it, besides just the Bokeh elements. Without the video the index.html is hardly different from the default template built into Bokeh. I want to demonstrate how the template complements and lets the Bokeh parts be augmented with a template. 

I've added the video back, so there's not difference between master and this PR with regards to the html template. So I'm not sure what the objection is. (I'm fixing the video cutoff-ness)

birdsarah:
OK, I added even more into index.html; carved out the stylesheet into a seperate file; and fixed the video.

bryevdv:
> I've added the video back, so there's not difference between master and this PR with regards to the html template. So I'm not sure what the objection is. (I'm fixing the video cutoff-ness)

I don't have one anymore, my comment was from earlier, when the video was gone, and before it was replaced...

bryevdv:
Looks great!

birdsarah:
> Also why did curdoc().title get removed? I went though all the apps and added that, so they would say something beside "Bokeh Application" on browser tabs

Picking up a dead thread here because I have a question. I removed it because the html document had a title so I was expecting that to get used. What's super odd is that it is briefly used and then it changes to Bokeh Application. So my apologies on that, I'm clealy missing a piece of understanding on how bokeh server works. I'm surprised it just overwrites the html template like that.

PR #4674, opened 2016-06-24
remove call to deprecated attribute and replace with new ones

### LINKED ISSUE #4672: Charts Heatmap uses deprecated attribute internally
This causes a deprecation warning everytime Heatmap is used:

```
/Users/fpliger/dev/repos/bokeh/bokeh/charts/builder.py:604: UserWarning: Chart property `sort_legend` was deprecated in 0.12             and will be removed in the future. Use `legend_sort_field` and             `legend_sort_direction` instead.
  `legend_sort_direction` instead.")
```

fpliger:
issues: fixes #4672

### DIFF
--- a/bokeh/charts/builders/heatmap_builder.py
+++ b/bokeh/charts/builders/heatmap_builder.py
@@ -143,8 +143,9 @@ class HeatMapBuilder(XYBuilder):
 
         # sort the legend by the color selection, reversed compared to how the values
         # were assigned to color
-        if len(self.sort_legend) == 0:
-            self.sort_legend = [('color', True)]
+        if self.legend_sort_field is None:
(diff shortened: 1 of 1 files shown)


bryevdv:
FWIW this removes the deprecation warning when making any heat map, tho I also tried using the options and could not find a configuration that did not error. Are we even sure these options make sense with heat map?

fpliger:
@bryevdv does your comment still apply after our conversation? (checking if we should open an issue for this or not... :) )

bryevdv:
I haven't tried to make an example yet. If you could provide a short one that would be helpful

fpliger:
Ok.. Will do it as soon as possible but don't think I'll be able to do it today (having hard day...)

PR #4676, opened 2016-06-24
more examples tweaks for 0.12

bryevdv:
This should try to collect the last set up examples updates.


bryevdv:
OK I'd like to merge this and cut RC5 the last two were needed to deployment on the demo site reasons.

bryevdv:
merging to help with deploy prep. We may not need any more examples work at this point.

Issue #4387, opened 2016-05-27
Layout documentation


birdsarah:
- Document off-canvas annotations
- Note that off-canvas legends require (x, y)
- Document layout & responsiveness
- Changes for use of embed with box responsive (you need to make sure the parent has a size)
- Document modes and modes that are currently less-functional than others.

PR #4678, opened 2016-06-24
general docs improvements for 0.12

### LINKED ISSUE #3329: Docs suggest that add_tools() takes shorthand strings, but it doesn't
In https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/tools.rst#specifying-tools, the text states that: 

    "To use shortcut names but also add tools with properties, one can also call the add_tools method:"

This is misleading, however, because `add_tools()` requires `Tool` instances, and does not accept a shorthand string.

### LINKED ISSUE #4368: add user guide documentation for Title and LabelSet annotations
Missing:
* add user guide documentation for Title and LabelSet annotations
* update Label annotation user guide docs

### LINKED ISSUE #4485: Update docs and docs build to be clear that only HTML output is supported
The innate dependency on JavaScript means that attempting to build other formats will not yield good results. Suggest:

* [x] removing other formats from the makefile
* [x] adding a note to the developer guide stating that only HTML output is supported

### LINKED ISSUE #4669: ArrowHead documentation
Styling the head of an Arrow annotation is not trivial. It seems there is a page missing from the documentation (for this file: https://github.com/bokeh/bokeh/blob/master/bokeh/models/arrow_heads.py for the type of the `end` attribute of the Arrow class here: http://bokeh.pydata.org/en/dev/docs/reference/models/annotations.html). 

Furthermore, the example in the user guide (http://bokeh.pydata.org/en/dev/docs/user_guide/plotting.html#arrows) could show an example of an arrow head with a different color (not just a different shape), for instance:
`p.add_layout(Arrow(end=VeeHead(line_color='blue'), x_start=0.5, y_start=0.7, x_end=0, y_end=0, line_color='red'))`

bryevdv:
#issues: closes #4485 closes #3329 closes #4669 closes #4368 

Other things to close in this PR:
* [x] `FuncTickFormatter` subsection in styling
* [x] legend outside plot area subsection in styling
* [ ] split annotations into separate section
* [x] new JS API section 
* [x]  custom models section
* [x] update apps examples README
* [x] update gallery for new apps examples to be deployed
* [x] docs for `bokeh.layouts`


bryevdv:
@birdsarah I think we should stop forcing gallery thumbnails to be square. I think they should be tiled in square areas, but in side they should scale to width

bryevdv:
Server gallery looks like this now:
screen shot 2016-06-25 at 2 31 46 pm

Still TBD if we can actually deploy the streaming 3d one

bryevdv:
@birdsarah if you do re-style the hover tool we I can make new screenshots easily

bryevdv:
@mattpap please review the new BokehJS docs https://github.com/bokeh/bokeh/pull/4678/commits/fa7db4a878a838a68dc656020015790c38cc8a3f for errors, completeness, etc. 

http://bokeh.pydata.org/en/test/docs/user_guide/bokehjs.html

Obviously I want to automate the generation of the plots and the options, etc, but... later.

bryevdv:
@birdsarah any reason not to embed the real plot at the bottom? It's possible to embed the plot without all the source code, if that was the concern. Or is it just overly big?

birdsarah:
> @birdsarah any reason not to embed the real plot at the bottom? It's possible to embed the plot without all the source code, if that was the concern. Or is it just overly big?

I had forgotten about source code not needing to be displayed. But yes there are other concerns: There's something in the docs css that makes scale_width plots embed wierd that I haven't had time to investigate; I expected stretch_both mode (which is the dashboard mode) to be harder to embed - although I didn't actually try. certainly, try, if it magically works, great!

birdsarah:
> @birdsarah I think we should stop forcing gallery thumbnails to be square. I think they should be tiled in square areas, but in side they should scale to width

Agree. I'm afraid I'm truly out of time now. This probably won't event require a redeploy of docs to update.

bryevdv:
OK, this is a docs only change, I am going to stop the build and merge. When master is green I will kick off RC5 for final testing, which should be the very last RC.

bryevdv:
I skipped splitting out annotations to its own sections, and the narrative on the extensions examples is a bit sparse. Will fill out in a follow on, with a re-deploy if need be.

bryevdv:
also closes #4387

PR #4683, opened 2016-06-26
Broken dialog 

### LINKED ISSUE #4652: Dialog broken
* Dialog - not sure how to use it even (how to get it to appear); adding buttons list doesn't seem to work; adding plot for content gives solver error.

birdsarah:
This is a WIP to get dialog working again.

As far as I can tell this was broken before the layout work - my guess is that the bk-root work broke it - but I'm not sure.

The layout work has complicated this further - you now have two instances of the same object on the model - the children of buttons_box and the buttons - this creates duplicated edit variables which is no longer ok.

Can't get this working quickly, and am out of time.

Close #4652

### DIFF
--- a/bokehjs/src/coffee/models/widgets/dialog.coffee
+++ b/bokehjs/src/coffee/models/widgets/dialog.coffee
@@ -7,6 +7,7 @@ p = require "../../core/properties"
 dialog_template = require "./dialog_template"
 Widget = require "./widget"
 
+
 class DialogView extends Widget.View
 
   initialize: (options) ->
@@ -14,7 +15,6 @@ class DialogView extends Widget.View
     @render()
     @render_content()
     @render_buttons()
-
     @listenTo(@model, 'destroy', @remove)
     @listenTo(@model, 'change:visible', @change_visibility)
     @listenTo(@model, 'change:content', @change_content)
@@ -25,7 +25,7 @@ class DialogView extends Widget.View
 
     content = @mget('content')
     if content?
-      if typeof content is 'object'
+      if typeof(content) != "string"
         @content_view = new content.default_view(model: content)
         @$el.find('.bk-dialog-content').empty()
         @$el.find('.bk-dialog-content').append(@content_view.$el)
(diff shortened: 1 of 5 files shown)


birdsarah:
other problems:
* dialog appears to be rendering twice.
* we've lost the greying out background
* content can disappear - it briefly appears and then a second render seems to eat it
* can't use buttons at all because of above issue
* if want to put on page with anything else then need to use some kind of layout, but that causes it to get laid out - which is not what we want

bokehjs/src/coffee/models/widgets/dialog_template.eco
```diff
@@ -1,4 +1,4 @@
-<div class="bk-bs-modal" tabindex="-1">
+<div class="bk-bs modal" tabindex="-1">
```
mattpap: Typo?

mattpap:
> we've lost the greying out background

Backdrop is unlikely going to work with `bk-root` and without modifications to bootstrap's code. This is because the backdrop's div is added to the top-level of a document.

bokehjs/src/coffee/models/widgets/dialog.coffee
```diff
@@ -65,15 +64,42 @@ class Dialog extends Widget.Model
...
+    for child in @get_layoutable_children()
+      edit_variables = edit_variables.concat(child.get_edit_variables())
+    return edit_variables
+
+  get_constraints: () ->
+    constraints = super()
+    # Go down the children to pick up any more constraints
+    for child in @get_layoutable_children()
+      constraints = constraints.concat(child.get_constraints())
+    return constraints
```
mattpap: I don't follow this. Dialog (a modal) is supposed to be shown on top of everything else. This way there should be no connection to layout. I presume this is necessary to just to lay children of a dialog out. However, I think it's unlikely children will have any connection (in terms of layout) with any elements below a dialog. I think there should be a fresh solver introduced to manage children of a dialog. Another situation where a fresh solver may be useful is laying out plots in tooltips.

bryevdv: I agree that a  separate solver is probably ideal. It will take some investigation to see if this is straightforward or now. I think it may not be so bad, we had already discussed a path for per-root solvers.

bryevdv:
I'm going to close this because it no longer cleanly merges, and for inactivity. But we need to either fix `Dialog`, or remove it, in the very near future.

PR #4686, opened 2016-06-27
Draw annotation level after WebGL blitting

### LINKED ISSUE #4679: Draw Legend After WebGL glyphs
screen shot 2016-06-24 at 6 50 29 pm

almarklein:
Fixes #4679 

cc @bryevdv @brendancol

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -131,6 +131,7 @@ class PlotCanvasView extends Renderer.View
     return @canvas_view.ctx.canvas
 
   init_webgl: () ->
+    ctx = @canvas_view.ctx
 
     # We use a global invisible canvas and gl context. By having a global context,
     # we avoid the limitation of max 16 contexts that most browsers have.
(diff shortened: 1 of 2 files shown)


bryevdv:
@almarklein technically the fix seems trivially correct. I wonder though is there a way to factor out and encapsulate the webgl logic here into a `"glglyph"` renderer? In other words I'd like to be able to write:
```
@_render_levels(ctx, ['image', 'underlay', 'glyph', 'glglyph', 'annotation'], frame_box)
``` 
and have it do the right thing. Ultimately it would be good to factor out most or all the webgl specific code out of `PlotCanvas`

almarklein:
@bryevdv the difficulty is that the webgl canvas must be blitted into the visible canvas at the right moment. This would be in between glyph and annotation. To do the blitting, the hdpi-scaling must be (temporarily) reset. I think the `PlotCanvas` is the most sensible place to do this, since it already deals with stuff like rendering levels and hidpi transforms.

Having a separate `glglyph` level will not solve this. I think it will only complicate things, because the webgl rendering is done by the glyphs themselves; they draw either in 2D or in WebGL. (Technically. the WebGL-enabled glyphs do wrap a special object that is also called glyph, but that's an implementation detail.)

bryevdv:
I'm suggesting exposing a way to do the rescaling to the renderers. Perhaps there is a *method* on `PlotCanvas` to do the rescaling, and then a `glglyph` renderer (that is in between glyph and annotation) can use that to do the rescaling at the right time, I am trying to simplify that specific block of code to "march down a list of things to process uniformly" instead of "process some stuff, stop and do some weird things, keep processing more"

bryevdv:
So looking again, maybe a better name for a new renderer is `glcanvas` ? I just want to get that chunk of code out of the middle of the rendering loop. It seems like it could either go in the `render` method of a `GLCanvas` renderer, or in a `PlotCanvas` method that it calls out to. Tho honestly I'd love to factor all of the GL stuff out into its own class.

almarklein:
Right, I was put on the wrong track by the word "glglyph". Nevertheless, doing the blitting in a renderer would still complicate things. There is another context state to be undone and redone (the clipping).

I just pushed a change that puts the blitting (and also the webgl preparing) in a separate method, so that the render method becomes much cleaner. I think all WebGL code in `PLotCanvas` is now constrained to three methods. We could certainly consider moving these method into a new object. Though that would better be done in a new PR, I think.

bryevdv:
@almarklein OK LGTM for now, restarted spurious test failures.

> We could certainly consider moving these method into a new object. Though that would better be done in a new PR, I think. 

Sure can  you make an issue for this

> Nevertheless, doing the blitting in a renderer would still complicate things. There is another context state to be undone and redone (the clipping).

What I'd suggest is to turn this:
```
@_render_levels(ctx, ['image', 'underlay', 'glyph'], frame_box)
@blit_webgl(ratio)
@_render_levels(ctx, ['annotation'], frame_box)
@_render_levels(ctx, ['overlay', 'tool'])
```
into this:
```
@_render_levels(ctx, frame_box, LEVELS)
```
by associating "do clipping or not" externally, as a mapping between level types and a boolean flag. Then `_render_levels` can just look  up whether to clip or not from the level name.

bryevdv:
please merge when green

Issue #3347, opened 2015-12-15
Larger color ranges (particularly gray scale)


jakirkham:
Often I find myself using `bokeh` to plot some sort of image and overlay it with annotations. To do a reasonable job showing a grayscale image, I would really like to use more than a handful. I have tried using 256 grays, but this still feels limiting. Is there some possibility to have larger color ranges in `bokeh`? If there isn't some way to do this currently, what would be required to have this feature at some point?

bryevdv:
@canavandl we coudl add a `grays` function similar to `viridis(...)` etc.

bryevdv:
@HuntJSparra can you look at making a PR for this?

HuntJSparra:
@jakirkham 
Any color ranges in particular you wanted? I am working to add functions similar to `viridis`, but was curious which palettes would be most useful to a user such as yourself.

Also, regarding the grays, I added the function for them, but only 256 can be displayed due to both hexadecimals just not having the precision for more than 256 and computers not being able to display more than 256 different shades of gray. Larger color ranges are possible with functions such as `magma` and should be with even simple color functions such as (working on it) `blues`.

jakirkham:
> Any color ranges in particular you wanted? I am working to add functions similar to `viridis`, but was curious which palettes would be most useful to a user such as yourself.

I have been using `viridis` with `matplotlib` for some things lately and do like it. Though I must have missed that addition to `bokeh`. Are there some docs or examples I should look at for this functionality? Basically, color ranges that appear perceptually linear are great (the more the merrier). Glad to hear there are some options in that domain.

I often find myself wanting to indicate objects (via polygons or contours) in a larger image. So, having a perceptually linear color scheme for the larger image and having some more discretized color scheme for the objects (so overlaps or neighbors are easily distinguishable) is great.

> Also, regarding the grays, I added the function for them, but only 256 can be displayed due to both hexadecimals just not having the precision for more than 256 and computers not being able to display more than 256 different shades of gray. Larger color ranges are possible with functions such as magma and should be with even simple color functions such as (working on it) blues.

Yeah, I had a function like this that I was using locally (had another dependency). I'm glad there will be one in Bokeh proper. Though, I agree that 256 grays is pretty limiting. Perceptually linear color ranges can take longer paths and thus have more variety. So, that has been what I have tended towards of late.

[...]

PR #4696, opened 2016-06-29
Function for Grays

HuntJSparra:
Partially responds to #3347:

- Adds grey/gray function similar to magma
- Adds greys256 palette
- Adds warning for when more than 256 colors are requested since more than 256 colors is not always possible

### DIFF
--- a/bokeh/palettes.py
+++ b/bokeh/palettes.py
@@ -455,7 +455,7 @@ __palettes__ = [
     "Greens3",    "Greens4",    "Greens5",    "Greens6",    "Greens7",    "Greens8",    "Greens9",
     "Oranges3",   "Oranges4",   "Oranges5",   "Oranges6",   "Oranges7",   "Oranges8",   "Oranges9",
     "Reds3",      "Reds4",      "Reds5",      "Reds6",      "Reds7",      "Reds8",      "Reds9",
-    "Greys3",     "Greys4",     "Greys5",     "Greys6",     "Greys7",     "Greys8",     "Greys9",                                   "Greys256"
+    "Greys3",     "Greys4",     "Greys5",     "Greys6",     "Greys7",     "Greys8",     "Greys9",     "Greys256",
     "PuOr3",      "PuOr4",      "PuOr5",      "PuOr6",      "PuOr7",      "PuOr8",      "PuOr9",      "PuOr10",      "PuOr11",
     "BrBG3",      "BrBG4",      "BrBG5",      "BrBG6",      "BrBG7",      "BrBG8",      "BrBG9",      "BrBG10",      "BrBG11",
     "PRGn3",      "PRGn4",      "PRGn5",      "PRGn6",      "PRGn7",      "PRGn8",      "PRGn9",      "PRGn10",      "PRGn11",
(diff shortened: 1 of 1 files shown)


bokeh/palettes.py
```diff
@@ -432,7 +455,7 @@
     "Greens3",    "Greens4",    "Greens5",    "Greens6",    "Greens7",    "Greens8",    "Greens9",
     "Oranges3",   "Oranges4",   "Oranges5",   "Oranges6",   "Oranges7",   "Oranges8",   "Oranges9",
     "Reds3",      "Reds4",      "Reds5",      "Reds6",      "Reds7",      "Reds8",      "Reds9",
-    "Greys3",     "Greys4",     "Greys5",     "Greys6",     "Greys7",     "Greys8",     "Greys9",
+    "Greys3",     "Greys4",     "Greys5",     "Greys6",     "Greys7",     "Greys8",     "Greys9",                                   "Greys256"
```
canavandl: Missing a comma. I'd also remove the extra whitespace (not try to align with vertically with other 256 palettes) and just put it in the next sequential column after `Greys9`.

canavandl:
Example code:

```python
import numpy as np

from bokeh.models import Range1d, LogColorMapper

output_file('image.html')

N = 100

# scale only shows the spike.

image = Z1 / Z1.min() #normalize to increase log range

mapper = LogColorMapper(palette="Greys256")

p.image(image=[image],
        color_mapper=mapper,
        # palette='Spectral11',
        dh=[1.0],
        dw=[1.0],
        x=[0],
        y=[0])

show(p)
```

yields:

screen shot 2016-06-29 at 3 14 08 pm

compares well to other colormaps in https://github.com/bokeh/bokeh/pull/4549#issuecomment-226641972

canavandl:
LGTM - I think the greyscale levels (rate it goes from black to white) seem appropriate.

@HuntJSparra - If you can try to include some example code to run or screenshots (especially if there aren't accompanying unit tests), it makes it easier for others to test. Next time :)

HuntJSparra:
Will do in the future. Thanks for looking over everything.

bryevdv:
@HuntJSparra Let 's take this opportunity to add tests for `bokeh.palettes`. There is a test file in `bokeh/tests/test_palettes`  it would be good to ensure that: 
* The expected palettes are present in the module
* The expected cmap functions are present in the module
Possibly going further:
* The palettes present all have the expected lengths
* Tests for the proper function of the generic cmap generator

bokeh/palettes.py
```diff
@@ -488,6 +511,8 @@
 
 def _cmap_func_generator(name, cmap):
     def func(n):
+        if n>256:
```
bryevdv: `>=256` or `>255` I think?

bryevdv: Also I would say raise a `ValueError`. But I think this function needs a new name. This 256 restriction only applies to certain specific cases. Something extremely descriptive maybe: `_linear_cmap_from_256_colors`

bryevdv:
@HuntJSparra some small comments

bryevdv:
superceded by #4751

PR #4699, opened 2016-06-30
Change div class to bk-root from bokeh-container

### LINKED ISSUE #4693: Incorrect rendering of embedded Bokeh server app in 0.12
Bokeh server app doesn't render properly when embedded in a Flask app with autoload_server (Bokeh 0.12.0).

This can be reproduced with the following example:

app.py:
```
from flask import Flask, render_template
from bokeh.embed import autoload_server

app = Flask(__name__)

@app.route('/')
def index():
    script = autoload_server(model=None, app_path="/sliders")
    return render_template('index.html', bokeh_script=script)
```

templates/index.html:
```
<!DOCTYPE html>
<html lang="en">

  <body>
    <h1>Bokeh app test</h1>
    <div class="bokeh_app">
      {{ bokeh_script|safe }}
    </div>
  </body>

</html>
```

Run this flask app and the sliders.py example from bokeh/examples/app:
`bokeh serve --host localhost:5000 --host localhost:5006 sliders.py`

This gives the following with Bokeh 0.12.0:
bokeh_0 12 0

Somethink like that (Bokeh 0.11.1) is expected:
bokeh_0 11 1

canavandl:
- [x] issues: fixes #4693 

Testing: this code isn't very well tested. I'm not sure the best way to ensure against regressions.

Example code: https://github.com/beenje/flask_bokeh_app

ping @beenje for testing

### DIFF
--- a/bokehjs/src/coffee/api/plotting.coffee
+++ b/bokehjs/src/coffee/api/plotting.coffee
@@ -3,6 +3,7 @@ $ = require("jquery")
 sprintf = require("sprintf")
 {Document} = require("../document")
 embed = require("../embed")
+{BOKEH_CSS_CLASS_NAME} = require("../embed")
 models = require("./models")
 
 _default_tooltips = [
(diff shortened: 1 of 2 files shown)


beenje:
I tested this branch and it looks good to me.

bryevdv:
The fix LGTM but I think we need a test to prevent regressions in the future

bryevdv:
OK moving to the ;single location protects against not changing thing everywhere (though I don't expect it it to change)  LGTM

Issue #4424, opened 2016-06-02
Strange behaviour and some other issues with ggplot


giancds:
Hey folks, I'm using Bokeh together with ggplot. 

I'm seeing some strange behaviors when converting the ggplot to Bokeh. I'm trying to make some scatterplots using my custom dataset: [dataset.csv.zip](https://github.com/bokeh/bokeh/files/295767/dataset.csv.zip)

If I use only ggplot, the code work as a charm:

```python
import pandas as pd
from ggplot import *
df = pd.read_csv("dataset.csv")
g = ggplot(df, aes("wVerb", "wNoun", "col2")) + geom_point()
print(g)
```
![correct](https://cloud.githubusercontent.com/assets/3305726/15747290/fb3e792e-28d1-11e6-8aeb-c2f559f64a63.png)

If I try to render the same figure using Bokeh, the data gets messed up:

```python
import pandas as pd
from ggplot import *
from bokeh import mpl
from bokeh.plotting import show, output_file
df = pd.read_csv("dataset.csv")
g = ggplot(df, aes("wVerb", "wNoun", color="col2")) + geom_point()
g.make()
output_file("mpl_subplot.html", title="mpl_subplot.py example")
show(mpl.to_bokeh())
```
![wrong](https://cloud.githubusercontent.com/assets/3305726/15747494/e80d0af4-28d2-11e6-9e21-13f9b13ad698.png)

If I try to call Bokeh to render the same plot but without colors, I get a gray figure: 
```python
import pandas as pd
from ggplot import *
from bokeh import mpl
from bokeh.plotting import show, output_file
df = pd.read_csv("dataset.csv")
g = ggplot(df, aes("wVerb", "wNoun")) + geom_point()
g.make()
output_file("mpl_subplot.html", title="mpl_subplot.py example")
show(mpl.to_bokeh())
```
![wrong1](https://cloud.githubusercontent.com/assets/3305726/15747347/3c419be0-28d2-11e6-8161-0b7964395ab0.png)

By the way, if I call `g.draw()` as per the Bokeh examples, I get an error `AttributeError: 'ggplot' object has no attribute 'draw'`. To make it work I had to use `g.make()`

If I try to pass the ggplot object to `mpl.to_bokeh()`, I get an error `AttributeError: 'ggplot' object has no attribute 'canvas'`.

Hope this does turn into a big issue as I would not like to stop using Bokeh.

Thanks for your time,
Gian

bryevdv:
[...]

PR #4718, opened 2016-06-30
Geom_Point Fix

HuntJSparra:
Fixed issue mentioned in #4424 

Previously converting mpl Paths or PathCollections to Bokeh glyphs was not supported. Unfortunately, mpl.gcf() for whatever reason returns a PathCollection instead of a Line2D when handling geom_point from ggplot. This caused the geom_point-turned-PathCollection not to render on the Bokeh plot. Shamelessly copying the draw_markers function from bokeh_renderer into a new draw_path_collection and modifying it a bit to work allows for the paths to be turned into Bokeh glyphs.

Unfortunately there are three limitations:
1. Since the PathCollection stores the shape of the glyph as a path and Bokeh does not yet handle custom glyph shapes, all converted paths default to circle.
2. If someone were to pass a PathCollection for a type of graph other than a scatter plot, it would most likely (untested) result in a scatter plot of the major points instead of, for example, a line plot. Although, this is still better than drawing nothing at all imo.
3. My method of calculating the sizes is no where close to an exact science.

Before Fix:
3c419be0-28d2-11e6-8161-0b7964395ab0

After Fix:
ggplot_works_mostly

### DIFF
--- a/examples/compat/ggplot_point.py
+++ b/examples/compat/ggplot_point.py
@@ -8,7 +8,7 @@ from bokeh.plotting import output_file, show
 
 g = ggplot(mtcars, aes(x='wt', y='mpg', color='qsec')) + geom_point()
 g.make()
-import pdb; pdb.set_trace()
+
 plt.title("Point ggplot-based plot in Bokeh.")
 
 output_file("ggplot_point.html", title="ggplot_point.py example")
--- a/examples/compat/ggplot_point.py
+++ b/examples/compat/ggplot_point.py
@@ -12,5 +12,5 @@ import pdb; pdb.set_trace()
 plt.title("Point ggplot-based plot in Bokeh.")
 
 output_file("ggplot_point.html", title="ggplot_point.py example")
-#plt.show()
+
 show(mpl.to_bokeh())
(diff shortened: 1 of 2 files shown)


bryevdv:
@HuntJSparra this looks good, can you do two more things:

* add the code that generated the above image to a new script in `examples/compat` 
* run all the other `examples/compat` scripts just to make sure this doesn't add any unexpected things to existing examples there

HuntJSparra:
- Added Example
- Added support for multiple colors on the same path
- Added support for path sizes although it is not an exact conversion
- Tested and works with current examples

Some of the new code is very... unpythonic. If someone could help clean it up I would really appreciate it.

examples/compat/ggplot_point.py
```diff
@@ -0,0 +1,16 @@
...
+from bokeh import mpl
+from bokeh.plotting import output_file, show
+
+g = ggplot(mtcars, aes(x='wt', y='mpg', color='qsec')) + geom_point()
+g.make()
+import pdb; pdb.set_trace()
+plt.title("Point ggplot-based plot in Bokeh.")
+
+output_file("ggplot_point.html", title="ggplot_point.py example")
+#plt.show()
```
bryevdv: would remove commented line

bryevdv:
LGTM, will merge when tests are green

PR #3001, opened 2015-10-16
Hotfix/box selection for mobile

birdsarah:
Box selection / zoom for touch was a bit wonky.

This PR doesn't fix it all, but does get it back to the state described in #2997 

I took the opportunity to write an integration test for box selection with mouse, but I can't write a similar test for touch as we don't have selenium setup with browserstack (or equivalent) to run tests against a mobile device.

You can try the before, after here:

Before: https://rawgit.com/birdsarah/bokeh-miscellany/master/tests/touch/box_touch_broken_1de22.html
After: https://rawgit.com/birdsarah/bokeh-miscellany/master/tests/touch/box_touch_hotfix_9b7b.html

PR #4723, opened 2016-07-01
factor JS version into version.coffee, compare to document version

bryevdv:
issues: fixes #3001

versions show up on `debug` when they match:

screen shot 2016-07-01 at 1 21 08 pm

but accompany a warning if they fail to match:

screen shot 2016-07-01 at 1 22 03 pm

Assumption is that `x.y.z-FOO` matches `x.y.z-BAR`

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -2,6 +2,7 @@ _ = require "underscore"
 $ = require "jquery"
 
 {Models} = require "./base"
+js_version = require("./version")
 {EQ, Solver, Variable} = require "./core/layout/solver"
 {logger} = require "./core/logging"
 HasProps = require "./core/has_props"
@@ -567,8 +568,17 @@ class Document
     Document.from_json(json)
 
   @from_json : (json) ->
+    logger.debug("Creating Document from JSON")
     if typeof json != 'object'
       throw new Error("JSON object has wrong type #{typeof json}")
+    py_version = json['version']
+    versions_string = "Library versions: JS (#{js_version})  /  Python (#{py_version})"
+    if js_version.split('-')[0] != py_version.split('-')[0]
+      logger.warn("JS/Python version mismatch")
+      logger.warn(versions_string)
+    else
+      logger.debug(versions_string)
+
     roots_json = json['roots']
     root_ids = roots_json['root_ids']
     references_json = roots_json['references']
--- a/bokehjs/src/coffee/main.coffee
+++ b/bokehjs/src/coffee/main.coffee
@@ -2,7 +2,7 @@ _ = require("underscore")
 
 Bokeh = {}
 Bokeh.require = require
-Bokeh.version = '0.12.0'
+Bokeh.version = require("./version")
 
 # binding the libs that bokeh uses so others can reference them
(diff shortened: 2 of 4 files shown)


bryevdv:
@damianavila can you test or suggest how best to test that the version update machinery is updated correctly for this change?

bryevdv:
FHYI @damianavila I think there is some possibility that the update version script is leaving an extra copy of the "last version" in the sphinx part, but not 100% certain

bryevdv:
OK this needs test updates to stub a python version in the document. As long as doing that will update to add a test that mismatches log an error as well.

bryevdv:
@damianavila I think there was a bug in the sphinx version update code. The old version is already there, so it is the new version that should be added. I can imagine maybe the intent was that the new "current" version should not be in teh dropdown. but I think is should. Might have been my mistake. Anyway, this is the output with the new  version:
```
bryan@0199-bryanv (git:bryanv/3001_version_checks) ~/work/bokeh/scripts $ python version_update.py 0.12.1 0.12.0
Version number changed from 0.12.0 to 0.12.1 in 
['bokehjs/src/coffee/version.coffee', 'bokehjs/package.json']
Version number 0.12.1 added in sphinx/source/conf.py
bryan@0199-bryanv (git:bryanv/3001_version_checks) ~/work/bokeh/scripts $ git diff
diff --git a/bokehjs/package.json b/bokehjs/package.json
index fe8ea16..a91d0b2 100644
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bokehjs",
-  "version": "0.12.0",
+  "version": "0.12.1",
   "description": "Interactive, novel data visualization",
   "keywords": [
     "bokeh",
diff --git a/bokehjs/src/coffee/version.coffee b/bokehjs/src/coffee/version.coffee
index ea8608d..3d8d452 100644
--- a/bokehjs/src/coffee/version.coffee
+++ b/bokehjs/src/coffee/version.coffee
@@ -1,3 +1,3 @@
-version = '0.12.0'
+version = '0.12.1'
 
 module.exports = version
diff --git a/sphinx/source/conf.py b/sphinx/source/conf.py
index 480ff8e..0532fed 100644
--- a/sphinx/source/conf.py
+++ b/sphinx/source/conf.py
@@ -166,7 +166,7 @@ html_context = {
         ('Releases', 'releases/%s' % version),
         ('Developer Guide', 'dev_guide'),
     ),
-    'ALL_VERSIONS': ['0.12.0', '0.11.0', '0.10.0', '0.9.3', '0.8.2'],
+    'ALL_VERSIONS': ['0.12.1', '0.12.0', '0.11.0', '0.10.0', '0.9.3', '0.8.2'],
     'css_server': os.environ.get('BOKEH_DOCS_CSS_SERVER', 'bokehplots.com'),
 }
```
This looks good to me, so I will merge when green unless there is objection.

damianavila:
> @damianavila I think there was a bug in the sphinx version update code. The old version is already there, so it is the new version that should be added. I can imagine maybe the intent was that the new "current" version should not be in the dropdown. but I think it should. Might have been my mistake.

That's OK, seems a sensible change.

PR #4724, opened 2016-07-01
allow scripts failures to fail the build

### LINKED ISSUE #4525: Shout louder on bokehjs build fails
I totally missed a syntax error when I was building bokeh recently because I was working on 3 things at once and the console error was pretty subtle

screen shot 2016-06-15 at 3 16 45 pm

I think it shouldn't say "...Success!" at the top in this case.

bryevdv:
issues: fixes #4525

The problem was that the `scripts` task was configured to "log" errors, instead of letting them fail the build, as described here:

https://github.com/contra/gulp-coffee#error-handling

With this change, e.g. a syntax error now results in:

screen shot 2016-07-01 at 2 02 01 pm

### DIFF
--- a/bokehjs/gulp/tasks/scripts.coffee
+++ b/bokehjs/gulp/tasks/scripts.coffee
@@ -33,13 +33,13 @@ ts = require 'gulp-typescript'
 gulp.task "scripts:coffee", () ->
   gulp.src('./src/coffee/**/*.coffee')
       .pipe(gulpif(argv.incremental, newer({dest: paths.buildDir.jsTree, ext: '.js'})))
-      .pipe(coffee({bare: true}).on('error', gutil.log))
+      .pipe(coffee({bare: true}))
       .pipe(gulp.dest(paths.buildDir.jsTree))
 
 gulp.task "scripts:eco", () ->
   gulp.src('./src/coffee/**/*.eco')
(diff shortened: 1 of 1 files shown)


bryevdv:
ping @mattpap this seems like the right thing to do but I wanted to make sure you hadn't put it the way it was for some particular reason.

bryevdv:
also ping @birdsarah

bryevdv:
Ok this is a tiny change, and the only thing I could think was that on a `gulp build` maybe errors would not get logged without `gutil.log` but they do just fine:
```
$ gulp build
[15:51:07] Using gulpfile ~/work/bokeh/bokehjs/gulpfile.js
[15:51:07] Starting 'typings:install'...
[15:51:18] Finished 'typings:install' after 11 s
[15:51:18] Starting 'build'...
[15:51:18] Starting 'scripts'...
[15:51:18] Starting 'scripts:coffee'...
[15:51:18] Starting 'scripts:eco'...
[15:51:18] Starting 'scripts:ts'...
[15:51:18] Starting 'styles'...
[15:51:18] Starting 'styles:build'...

events.js:141
      throw er; // Unhandled 'error' event
      ^
/Users/bryan/work/bokeh/bokehjs/src/coffee/client.coffee:11:14: error: missing ]
class Message[
```
So, I will merge this when Travis is green

PR #4731, opened 2016-07-02
Fix documented name for ResizeTool

patricktokeeffe:
issues: none

Latest (0.12.0) [docs](http://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#resizetool) specify `'resize_select'` as the name for ResizeTool tool, but correct name is just `'resize'`. Fixes doc.

### DIFF
--- a/sphinx/source/docs/user_guide/tools.rst
+++ b/sphinx/source/docs/user_guide/tools.rst
@@ -204,7 +204,7 @@ respectively.
 ResizeTool
 ''''''''''
 
-* name: ``'resize_select'``
+* name: ``'resize'``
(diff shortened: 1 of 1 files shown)


patricktokeeffe:
Can I apply tags? Can I not? I've been told to. GH help is silent on the issue. I don't get it so sorry here they are:

`type: bug`
`status: ready`
`reso: completed`
`milestone: 0.12.1`
`tag: component: docs`

bryevdv:
> Can I apply tags? 

No, only members of @bokeh/dev can add tags

Thanks for the catch!

PR #4756, opened 2016-07-05
Added `merge_tools` option to gridplot

### LINKED ISSUE #4755: Feature request: Make tool coalescence optional in gridplot

guillochon:
- [x] issues: fixes #4755 
- [ ] tests added / passed


guillochon:
I added a brief description of the option to layout.rst.

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -42,7 +42,10 @@ Valid values are:
 If you would like to hide the toolbar entirely, pass ``None``.
 
 Below is some code that positions the toolbar below the plot. Try
-running the code and changing the ``toolbar_location`` value.
+running the code and changing the ``toolbar_location`` value. Note
+that the toolbar position clashes with the default axes, in this case
+setting the ``toolbar_sticky`` option to ``False`` will move the toolbar
+to outside of the region where the axis is drawn.
```
bryevdv: If you are up for it, what might be ideal is to change the script referenced below to have two plots in a `row`, identical except with one that has `toolbar_sticky=False`

guillochon:
I added a "clashing" example without the `toolbar_sticky` flag, then show a working example with that flag set to `False` below the comment about the flag.

bryevdv:
Thanks @guillochon ! I have a day full of meetings but I will do my best to get this merged tomorrow.

bryevdv:
Actually I might make one more suggestion: how about `gather_tools` ? I know I suggested `coalesce_tools` in gitter chat, but it's quite a mouthful

bryevdv:
@guillochon alternatively an somewhat internal class uses `merge_tools` either of those would be better I think.

bryevdv:
Thanks for the PR @guillochon !

PR #4777, opened 2016-07-08
add cartodb tile provider

### LINKED ISSUE #4775: [feature] add Cartodb positron tile provider
See

https://carto.com/location-data-services/basemaps/

for an example.

PR upcoming

sanga:
- [x] issues: fixes #4775
- [ ] tests added / passed: tile_providers aren't currently tested so far as I can tell and I'm not sure how I would test this at the moment?
- [ ] release document entry (if new feature or API change) -  would you like a release notes entry (seems kinda minor to me)?

Note: I assume retina or non-retina could be detected somehow on the fly and switched. But this is useful as it is at least for me.

### DIFF
--- a/bokeh/tile_providers.py
+++ b/bokeh/tile_providers.py
@@ -1,5 +1,7 @@
 ''' Pre-configured tile sources with urls and attribution for common 3rd-party tile services.
-    Additional Stamen tile service information available at: http://maps.stamen.com/
+    Additional information available at:
+        Stamen tile service - http://maps.stamen.com/
+        CartoDB tile service - https://carto.com/location-data-services/basemaps/
 '''
 from .models.tiles import WMTSTileSource
 
@@ -47,6 +49,7 @@ STAMEN_TERRAIN = WMTSTileSource(
     )
 )
 
+#: Tile Source for CartoDB Tile Service
 CARTODBPOSITRON = WMTSTileSource(
     url='http://tiles.basemaps.cartocdn.com/light_all/{z}/{x}/{y}.png',
     attribution=(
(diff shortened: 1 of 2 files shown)


bryevdv:
@sanga I would suggest that a new example that exercises the new tile providers would be sufficent for testing for now. However, a new jq-ui version breaks things. I have a PR to pin to the old versions, you will need to merge master into this PR once the fix is in master, in order to get tests passing. 

Regarding docs, it would actually be nice to add doc comments to everything in `tile_providers`.  You can see an example of doc comments here:

https://github.com/bokeh/bokeh/blob/master/bokeh/core/enums.py#L80

If you have the time to add some doc comments with a few words about the module contents, it would be very nice, but I would not hold up this PR for it.

bryevdv:
Ok, the jqui version pin PR is merged. If you merge master into your branch, expected tests should start passing.

sanga:
Sounds all fine. To be clear: you'd rather I merge master than rebase.... (I usually always rebase to leave history cleaner)

bryevdv:
@sanga the example actually needs to output an html file with the right name (`airports_map.html` in this case) or else the example CI test will fail. Currently there is no output file being created because you removed the `show` command. If you want to show both in one example, you'll need to put both plots in a layout of some sort (e.g. a `Row` or `Column` and `show` that).

bryevdv:
Alternatively, could restore the original `airports_map.py` and add a completely new example to show off the cartodb tiles.

sanga:
Let's try that. That produces the correct result when I run python airport_map.py (i.e. gives me a correct airports_map.html)

bryevdv:
LGTM

screen shot 2016-07-09 at 4 26 50 pm

ping @jbednar FYI cartodb seems to work great

bryevdv:
Thanks @sanga !

jbednar:
Looks great!

PR #4803, opened 2016-07-13
Make WebGL work on Safari

### LINKED ISSUE #2683: Our Webgl support does not work very well on Safari

### LINKED ISSUE #4795: WebGL line thickness scales inappropriately with browser zoom level
The line width should probably be scaled (or *not* scaled) with the HIDPI `pixel_ratio`.

almarklein:
Fixes #4795 
Fixes #2683 

The main fix has been applied in https://github.com/vispy/vispy.js/blob/master/scripts/gloo2.py, which is the code that generates gloo2.js. While in there I also improved the API a bit, so this PR also updates the Bokeh code accordingly. Further, I snuck in a quick fix for #4795.

Todo:
* [x] Test on other OS / browsers whether all webgl examples still work.
* [x] Test whether line width issue exists on non-retina displays as well, (and that it is now fixed).
* [x] Look into FF warning "Texture has not been initialized prior to a partial upload, forcing the browser to clear it. This may be slow."

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/webgl/base.coffee
+++ b/bokehjs/src/coffee/models/glyphs/webgl/base.coffee
@@ -61,16 +61,16 @@ attach_float = (prog, vbo, att_name, n, visual, name) ->
     # otherwise use VBO to apply array.
     if not visual.doit
       vbo.used = false
-      prog.set_attribute(att_name, 'float', null, 0)
+      prog.set_attribute(att_name, 'float', [0])
     else if visual_prop_is_singular(visual, name)
       vbo.used = false
-      prog.set_attribute(att_name, 'float', null, visual[name].value())
+      prog.set_attribute(att_name, 'float', visual[name].value())
     else
       vbo.used = true
       a = new Float32Array(visual.cache[name + '_array'])
       vbo.set_size(n*4)
       vbo.set_data(0, a)
-      prog.set_attribute(att_name, 'float', [vbo, 0, 0])
+      prog.set_attribute(att_name, 'float', vbo)
 
 attach_color = (prog, vbo, att_name, n, visual, prefix) ->
     # Attach the color attribute to the program. If there's just one color,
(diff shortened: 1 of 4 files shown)


bryevdv:
Awesome.

bryevdv:
Seems to be broken on windows tho. If I use CDN resources, it works in my VM but if I set to inline reources, it stops displaying.

almarklein:
Strange. I just verified it to work fine on my Windows machine (except some warnings on FF that I need to look into), both with inline and relative dev. Any error messages in the console?

bryevdv:
Well it might be an issue on my end. If it's working well for you lets get it in a dev build. That would be easier for me to test on Windows.

bryevdv:
@almarklein more info: the error I get on my windows VM is "can't set initial ranges" if I provide explicit range tuples to `figure`, it works...

bryevdv:
@damianavila if you have 10 minutes can you try the `examples/webgl` examples with this PR on LInux?

caryan:
Seems #4795 is still an issue on HIDPI in Linux. Otherwise the examples seem to work fine in Chrome and Firefox.

screenshot from 2016-07-14 22-49-25

bryevdv:
@caryan we aren't able to reproduce this anymore, maybe a config/install issue? We're going to go ahead and merge this in a bit, and then put out a dev build later today. If you can check out the issue with the dev build, and if it persists in the dev build we can re-open the issue.

caryan:
Sorry about that.  I must have botched the local build/install. With 0.12.1dev3 it works fine.

damianavila:
>@damianavila if you have 10 minutes can you try the examples/webgl examples with this PR on LInux?

Sorry, I just saw this... glad to see it is working.

PR #4809, opened 2016-07-15
Split interaction.rst into three sections

Maggie-M:
Everything's almost ready to go, save for the link descriptions in interaction.rst


bryevdv:
@canavandl this can also go in `dev3`

PR #4812, opened 2016-07-15
fix multiselect to allow values or values and labels

### LINKED ISSUE #4800: MultiSelect not rendering correctly if `options` is `list(dict)`
From the sources of MultiSelect widget I see the possibility to provide `options` as:
`options = List(Either(String, Dict(String, String))`

But the HTML-output looks wrong:
SelectBox with `[object Object]`

Example showing the issue:
```
from bokeh.models.widgets import MultiSelect
from bokeh.io import show
multi_select = MultiSelect(title="Option Works:", options=['a', 'b', 'c'])
multi_select_wrong = MultiSelect(title="Option Wrong:", options=[{'name': 'a', 'value': 'a'}, {'name': 'b', 'value': 'b'}, {'name': 'c', 'value': 'c'}])
show(widgetbox(multi_select, multi_select_wrong))
```
My environment:
```
In [11]: bokeh.__version__
Out[11]: '0.12.1dev2'
```
the `version.js` has `version = '0.12.0'` (perhaps the revision was not updated in the dev-build)

bokeh_multi_select

Thanks for bokeh, it's an incredible tool.

bryevdv:
- [x] issues: fixes #4800
- [x] tests added / passed

### DIFF
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -102,8 +102,11 @@ class MultiSelect(InputWidget):
 
     """
 
-    options = List(Either(String, Dict(String, String)), help="""
-    Available selection options.
+    options = List(Either(String, Tuple(String, String)), help="""
+    Available selection options. Options may be provided either as a list of
+    possible string values, or as a list of tuples, each of the form
+    ``(value, label)``. In the latter case, the visible widget text for each
+    value will be corresponding given label.
     """)
 
     value = List(String, help="""
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 import six
 
 from ...core.properties import abstract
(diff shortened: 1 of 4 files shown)


bryevdv:
ping @bokeh/dev @busino

busino:
Hi Bryan,
The `list(tuple)` approach is perfect. To be consistent the Select widget should be changed too. 

Have a nice week-end.
Raphael

bryevdv:
@busino ah, good catch I will try to get to that tomorrow.

bryevdv:
@bokeh/dev is anyone aware of the list of dicts format working for either `Select` or `MultiSelect` at any time? AFAICT the format of the required dicts was never documented, so I feel that it is fairly safe to change, even if it did work at some point. Would accompany this PR with `MIGRATION` in any case.

bryevdv:
@bokeh/dev OK, I am gong to go ahead and merge this. If there are strong opinions, or known exemplars of the "list of special dicts" usage to consider, we can consider reverting.

Issue #3423, opened 2015-12-28
Add VBar and HBar glyphs


bryevdv:
Very often want to specify *one* of width or height, but not *both* (e.g. almost any financial bar plotting).

Propose:
```
class VBar(Glyph):
    """ Render vertical bars, given a center coordinate, width and (top, bottom) coordinates. """

    __example__ = "tests/glyphs/VBar.py"

    # a canonical order for positional args that can be used for any
    # functions derived from this class
    _args = ('center', 'width', 'top', 'bottom')

    center = NumberSpec(help="""
    The x-coordinates of the centers of the bars.
    """)

    width = NumberSpec(help="""
    The widths of the vertical bars.
    """)

    bottom = NumberSpec(help="""
    The y-coordinates of the bottom edges.
    """)

    top = NumberSpec(help="""
    The y-coordinates of the top edges.
    """)

    line_props = Include(LineProps, use_prefix=False, help="""
    The %s values for the rectangles.
    """)

    fill_props = Include(FillProps, use_prefix=False, help="""
    The %s values for the rectangles.
    """)

class HBar(Glyph):
    """ Render horizontal bars, given a center coordinate, height and (left, right) coordinates. """

    __example__ = "tests/glyphs/HBar.py"

    # a canonical order for positional args that can be used for any
    # functions derived from this class
    _args = ('center', 'height', 'left', 'right')

    center = NumberSpec(help="""
    The x-coordinates of the centers of the bars.
    """)

    height = NumberSpec(help="""
    The heights of the vertical bars.
    """)

    left = NumberSpec(help="""
    The x-coordinates of the left edges.
    """)

    right = NumberSpec(help="""
    The x-coordinates of the right edges.
    """)

    line_props = Include(LineProps, use_prefix=False, help="""
    The %s values for the rectangles.
    """)

    fill_props = Include(FillProps, use_prefix=False, help="""
    The %s values for the rectangles.
    """)
```

ericmjl:
@bryevdv: after talking with @birdsarah, I'd like to take a stab at this between now and SciPy 2016. Can we continue the conversation through here?

bryevdv:
@ericmjl absolutely, this would be a really welcome addition. Are there any questions I can answer to get things started? The old closed PR is probably a decent starting place. The main difficulty as I recall is handing the hit-testing, etc. Perhaps we could make a first PR that just skips that (we do have other glyphs that don't support hit-testing yet for example)

PR #4815, opened 2016-07-16
Vbar hbar

ericmjl:


canavandl:
I think may have accidently checked-in Quad.html and Vbar.html? They'll probably have to be removed.

ericmjl:
Thanks for letting me know. I will take care of that.

ericmjl:
@canavandl just getting back to the thing you mentioned - where did you put Quad.html and Vbar.html?

bryevdv:
@ericmjl I think @canavandl is saying that you accidentally checked those files in to the branch (you can see them in "Files Changed" tab). They need to be removed

bokeh/models/glyphs.py
```diff
@@ -290,6 +290,39 @@ class Gear(Glyph):
...
+    right = NumberSpec(help="""
+    The x-coordinates of the right edges.
+    """)
+
+    line_props = Include(LineProps, use_prefix=False, help="""
+    The %s values for the rectangles.
+    """)
+
+    fill_props = Include(FillProps, use_prefix=False, help="""
+    The %s values for the rectangles.
```
bryevdv: suggest updating both of these to "horizontal bars" instead of "rectangles"

bokeh/models/glyphs.py
```diff
@@ -290,6 +290,39 @@ class Gear(Glyph):
...
+    """ Render horizontal bars, given a center coordinate, height and (left, right) coordinates. """
+
+    __example__ = "tests/glyphs/HBar.py"
+
+    # a canonical order for positional args that can be used for any
+    # functions derived from this class
+    _args = ('y', 'height', 'left', 'right')
+
+    y = NumberSpec(help="""
+    The y-coordinates of the centers of the bars.
```
bryevdv: suggest adding "horizontal" before "bars"

bokeh/models/glyphs.py
```diff
@@ -893,6 +926,39 @@ class Text(Glyph):
...
+
+    __example__ = "tests/glyphs/VBar.py"
+
+    # a canonical order for positional args that can be used for any
+    # functions derived from this class
+    _args = ('x', 'width', 'top', 'bottom')
+
+    x = NumberSpec(help="""
+    The x-coordinates of the centers of the bars.
+    """)
```
bryevdv: suggest adding "vertical" before "bars"

bokeh/models/glyphs.py
```diff
@@ -893,6 +926,39 @@ class Text(Glyph):
...
+
+    width = NumberSpec(help="""
+    The widths of the vertical bars.
+    """)
+
+    bottom = NumberSpec(help="""
+    The y-coordinates of the bottom edges.
+    """)
```
bryevdv: we could be opinionated here, it's possible to add `default=0` and that would make the `bottom=0` unnecessary for bars that do happen to start at 0

tests/glyphs/HBar.py
```diff
@@ -0,0 +1,40 @@
...
+
+N = 9
+y = np.linspace(-2, 2, N)
+x = y**2
+
+source = ColumnDataSource(dict(
+        y=y,
+        right=x,
+    )
+)
```
bryevdv: small style suggestion: this could all easily fit on one line:
```
source = ColumnDataSource(dict(y=y, right=x))
```

tests/glyphs/VBar.py
```diff
@@ -0,0 +1,40 @@
...
+N = 9
+x = np.linspace(-2, 2, N)
+y = x**2
+
+source = ColumnDataSource(dict(
+        x=x,
+        top=y,
+    )
+)
+
```
bryevdv: same comment, suggest one line for `source` definition

bokehjs/src/coffee/models/glyphs/vbar.coffee
```diff
@@ -0,0 +1,67 @@
...
+  # Temporarily commented out. To be put back later.
+  # _index_data: () ->
+  #   index = rbush()
+  #   pts = []
+  #   for i in [0...@x.length]
+  #     if not isNaN(@x[i] + @width[i] + @top[i] + @bottom[i])
+  #       w2 = @width[i]/2
+  #       pts.push([@x[i]-w2, @bottom[i], @x[i]+w2, @top[i], {'i': i}])
+  #   index.load(pts)
+  #   return index
```
bryevdv: did you want to continue this in this PR? If not, let's just delete this and the other block in the other file. I'm almost always -1 on checking in commented-out code

bryevdv:
@ericmjl I left a few other small comments otherwise this is looking great. This will be a nice add!

bryevdv:
@ericmjl one other suggestion: our CI tests kick off on ever push, and we only have 5 workers. Best to collect all the commits locally and then push at the end

bokeh/models/glyphs.py
ericmjl: I'm not quite sure about this. What's the intended use case for VBar and HBar? Is likely isn't a user passing in data, in which case I think it makes more sense for the user to be explicit about what the top and bottom values should be, would that be right?

ericmjl:
@bryevdv yes, i'm sorry about the multiple pushes. thanks for letting me know!

bokeh/models/glyphs.py
bryevdv: I'm not suggesting adding a default for `top`.  I expect `VBar` will be used to easily create bar-like charts using `bokeh.plotting`. The suggestion is merely to make what assume is the common case -- bars starting from `y=0` slightly easier to spell.
```
plot.vbar(x=[1, 2, 3], width=0.5, top=[1,2,3])
```
instead of 
```
plot.vbar(x=[1, 2, 3], width=0.5, bottom=0, top=[1,2,3])
```
It's s minor thing, but common cases are the most important places for minor improvements. Of course if the bars don't start at zero the user can still supply `bottom` explicitly, that's no different.

bokehjs/src/coffee/models/glyphs/vbar.coffee
ericmjl: `_index_data` is needed for hit testing, right? I'm going to need a lesson from you on what hit testing is all about, btw 😛. In any case, I will remove it and the other one first.

bryevdv:
no apologies needed, help and guidance is our responsibility

bokehjs/src/coffee/models/glyphs/vbar.coffee
bryevdv: Yup, all this is to support hit testing, which we can certainly add in a follow-on later. It's potentially possible to just add "brute force" hit testing without an index, too. We can discuss details later

bokeh/models/glyphs.py
bryevdv: And of course being explicit with `bottom=0` is also fine

ericmjl: Ah, I see. My apologies for the previous comment - I had a grammatical breakdown.

I was thinking about a particular case that isn't assumed with the default `bottom=0`:

    plot.vbar(x=[1, 2, 3], width=0.5, bottom=[-3, 0, 0], top=[0, 2, 3])

In this case, there is one negative bar. A user would have to explicitly specify that the bottom is negative, then.

Ok, on thinking about it, I think having a default `bottom=0` makes sense. I'll add this in, and likewise for HBar (left=0).

bryevdv:
FYI while I'm +1 on defaulting `bottom=0` I'm +0 on defaulting `left=0`. I'll leave that to you to decide

bokeh/models/glyphs.py
```diff
@@ -893,6 +926,39 @@ class Text(Glyph):
...
+    The x-coordinates of the centers of the vertical bars.
+    """)
+
+    width = NumberSpec(help="""
+    The widths of the vertical bars.
+    """)
+
+    bottom = NumberSpec(help="""
+    The y-coordinates of the bottom edges.
+    """, default=0)
```
bryevdv: OK, being super picky here, but every other property puts the help last, I think ti would be good to be consistent, so:
```
+    bottom = NumberSpec(default=0, help="""
 +    The y-coordinates of the bottom edges.
 +    """)
 
```

ericmjl:
@bryevdv I think it makes sense to include left=0 as a default, for consistency purposes. Most use cases - data are positively valued, and visually bottom and left are used for zeros.

bokeh/models/glyphs.py
```diff
@@ -290,6 +290,39 @@ class Gear(Glyph):
...
+    The y-coordinates of the centers of the horizontal bars.
+    """)
+
+    height = NumberSpec(help="""
+    The heights of the vertical bars.
+    """)
+
+    left = NumberSpec(help="""
+    The x-coordinates of the left edges.
+    """, default=0)
```
bryevdv: same comment about default being first

bokeh/models/glyphs.py
ericmjl: I'm sorry! I should have looked more carefully at the conventions.

bryevdv: Not at all, it's not the sort of thing that's easy to notice as a patter, again that's why we are here to help!

bryevdv:
Ok, I forgot to mention the defaults on the JS side (in the `@define` blocks) also need to be updated to match, and also the python unit tests need to be updated as well:

https://travis-ci.org/bokeh/bokeh/jobs/145977715#L2468-L2491

bryevdv:
@ericmjl there is one last test still failing, I'm not sure what it is offhand, but I will look into it this afternoon

ericmjl:
@bryevdv I've been trying to debug it too this morning, but I think I'm running into the limits of my knowledge of the `bokeh` library at this point. Thanks for guiding me along this too, btw!

bryevdv:
@ericmjl the coffee default values need to be updated to match. The following patch gets the tests passing:
```
 git diff
diff --git a/bokehjs/src/coffee/models/glyphs/hbar.coffee b/bokehjs/src/coffee/models/glyphs/hbar.coffee
index c176c56..850fdf3 100644
--- a/bokehjs/src/coffee/models/glyphs/hbar.coffee
+++ b/bokehjs/src/coffee/models/glyphs/hbar.coffee
@@ -46,10 +46,10 @@ class HBar extends Glyph.Model
 
   @mixins ['line', 'fill']
   @define {
-      y:      [ p.NumberSpec   ]
-      height: [ p.DistanceSpec ]
-      left:   [ p.NumberSpec   ]
-      right:  [ p.NumberSpec   ]
+      y:      [ p.NumberSpec     ]
+      height: [ p.DistanceSpec   ]
+      left:   [ p.NumberSpec,  0 ]
+      right:  [ p.NumberSpec     ]
     }
 
 module.exports =
diff --git a/bokehjs/src/coffee/models/glyphs/vbar.coffee b/bokehjs/src/coffee/models/glyphs/vbar.coffee
index 5ef9ed3..e8e076b 100644
--- a/bokehjs/src/coffee/models/glyphs/vbar.coffee
+++ b/bokehjs/src/coffee/models/glyphs/vbar.coffee
@@ -45,10 +45,10 @@ class VBar extends Glyph.Model
 
   @mixins ['line', 'fill']
   @define {
-      x:      [ p.NumberSpec   ]
-      width:  [ p.DistanceSpec ]
-      top:    [ p.NumberSpec   ]
-      bottom: [ p.NumberSpec   ]
+      x:      [ p.NumberSpec     ]
+      width:  [ p.DistanceSpec   ]
+      top:    [ p.NumberSpec     ]
+      bottom: [ p.NumberSpec,  0 ]
     }
 
 module.exports =
```
Then we should be good to merge.

bryevdv:
@ericmjl all expected tests passing. Thanks for the great PR! Let me know when you'd like to work on the hit-testing and I can help explain what's going on there.

ericmjl:
@bryevdv thanks for guiding me through! Let me get my life back in order next week, and I'll be back up to try hit testing!

damianavila:
@ericmjl @bryevdv is there any issue opened for this PR? Or is it self-contained?

ericmjl:
@damianavila I think this is linked to issue #3423. Hit testing is still missing, as mentioned above, but I plan to work on it in a separate PR.

damianavila:
Thanks @ericmjl

phobson:
it would seem to me that the hit testing would be identical to the that of the Quad glyph. am I over-simplifying things? I'm messing with this here: https://github.com/phobson/bokeh/pull/1

bryevdv:
@phobson yes that would work but the spatial index also has to be built (i.e. the `_index_data` method also needs to be implemented so that `@index` exists for `_hit_point` to make use of). Alternatively, `_hit_point` could iterate over all the data points and do "point in rect" hit testing by brute force.

bryevdv:
@phobson as a reference the `_index_data` for quads is here: 

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/glyphs/quad.coffee#L10-L42 

The index needs "top, bottom, left, right" coords. vbar has "top, bottom, x, width" and hbar has "y, height, left, right" so the main task in `_index_data` implementations for vbar and hbar would be to convert "x, width" into "left, right" and "y, height" into "top, bottom".

phobson:
@bryevdv thanks for the guidance.

I'll spend sometime on this on my own for a while and come back with more complete thoughts and questions -- perhaps with a proper PR.

bryevdv:
@phobson that would definitely be extremely valuable and appreciated please let us know any questions we can answer

PR #4821, opened 2016-07-16
added black outline and hover color

### LINKED ISSUE #2610: Improve Les Mis example
Two improvements need to be made:

1. Currently, the "non-interacting" squares read out as really really small floats, e.g. "1.27e-123".  We should change these actual values to 0, but also add an ability to suppress the hover tooltip to not display anything for those values.

2. It would be really nice to make the axis labels light up/change color to show when the user hovers over a square, like in the D3 version: http://bost.ocks.org/mike/miserables/

aernlund:
Fixes #2610


birdsarah:
(closed and re-opened to kick travis)

bryevdv:
Thanks @aernlund !

PR #4826, opened 2016-07-17
moved the wheel speed zoom from internal to defined.

### LINKED ISSUE #3125: Expose the wheel zoom speed in Python
The sensitivity of the wheel zoom cannot be configured on the Python side. However, there is a `speed` property which can be set on the JS side. It would be better to expose this property to the Python side too.

sarahheim:
moved the wheel speed zoom from internal to defined. 
Default 1/600.
High values will be clipped.

- [x] issues: fixes #3125

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -197,6 +197,12 @@ class WheelZoomTool(Scroll):
     vertically across the height of the plot.
     """)
 
+    speed = Float(default=1/600, help="""
+    Speed at which the wheel zooms. Default is 1/600. Optimal range is between 
+    0.001 and 0.09. High values will be clipped. Speed may very between browsers.
+    """)
+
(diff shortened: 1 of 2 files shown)


bryevdv:
Thanks for the PR!

bryevdv:
Ack, I should have known the "python <---> JS defaults" checking would need updating for this. reverting for now.

bryevdv:
@sarahheim this turned out to be a python 2 division issue!
```
  WheelZoomTool.speed: coffee defaults to 0.0016666666666666668 but python defaults to 0
```
The python 2 tests failed since `1/600 == 0` in python 2. I reverted the old PR because merging early was my mistake, and I'd like to give you the chance to submit a new one with a fix. If you are busy, etc. I completely understand.

Issue #3250, opened 2015-12-06
Tool icons should have the same dimensions


mattpap:
Currently:

```
BoxSelect.png:     PNG image data, 36 x 32, 8-bit/color RGBA, non-interlaced
BoxZoom.png:       PNG image data, 34 x 32, 8-bit/color RGBA, non-interlaced
Help.png:          PNG image data, 32 x 32, 8-bit/color RGB, non-interlaced
Inspector.png:     PNG image data, 32 x 32, 8-bit/color RGBA, non-interlaced
LassoSelect.png:   PNG image data, 36 x 32, 8-bit/color RGBA, non-interlaced
Pan.png:           PNG image data, 32 x 32, 8-bit/color RGBA, non-interlaced
PolygonSelect.png: PNG image data, 36 x 32, 8-bit/color RGBA, non-interlaced
Reset.png:         PNG image data, 40 x 32, 8-bit/color RGBA, non-interlaced
Resize.png:        PNG image data, 34 x 32, 8-bit/color RGBA, non-interlaced
Save.png:          PNG image data, 32 x 32, 8-bit/color RGBA, non-interlaced
TapSelect.png:     PNG image data, 32 x 32, 8-bit/color RGBA, non-interlaced
WheelZoom.png:     PNG image data, 44 x 32, 8-bit/color RGBA, non-interlaced
```

btw. Who is responsible for creating new icons, etc., after Lila's departure?

bryevdv:
I think amenity created these. Now, maybe Victoria could help.

VelizarVESSELINOV:
:+1: Any plan to move to vector icons, the current icons very pixelized?
Is it possible to customize or update the icons? The save icon is very old style...

bryevdv:
If you know a visual designer who would want to help contribute to an OSS project, we'd be happy to do all of these things. :)

VelizarVESSELINOV:
I will ask few friends for help, but also it is not so difficult to use basic icon and customize a little bit if needed: http://www.flaticon.com/

VelizarVESSELINOV:
@bryevdv Where are stored all the icons? What it is current process to submit you new icons?

canavandl:
Hi @VelizarVESSELINOV 

The icons are located at https://github.com/bokeh/bokeh/tree/master/sphinx/source/_images/icons (or the ./bokeh/sphinx/source/_images/icons directory of the project). In terms of submitting new icons, you'd have to open a PR where the files have been replaced with new image files.

[...]

PR #4841, opened 2016-07-20
Refactor wheel_zoom_tool. Add zoom button tools #916

### LINKED ISSUE #916: Add zoom button that allows zoom by steps
This would add a plus and minus button to the tools that zooms in and out at the current center of the plot.

Examples to consider: Google Maps, Leaflet, Mapbox, basically any slippy map. :-)

timsnyder:
- [x] issues: fixes #916
- [x] tests added / passed

Refactors models/tools/gestures/wheel_zoom_tool into
a scale_range() helper provided by models/tools/zoom_util.

scale_range() is then used to add zoom_in_tool and zoom_out_tool
to models/tools/actions, implementing toolbar buttons that will
zoom

Remaining work on #916 includes:

- [x] add new models to bokehjs/src/coffee/api/typings/models/tools.d.ts. somehow I missed those in my initial commit

- [X] ~~actual design done for icons/Zoom{In,Out}.png.~~ I think what I slapped together is good enough until the tool icons are revisited en masse in #3250 

- [X] ~~ensure zoom-in and zoom-out are "bundled"~~ won't be included in this issue, will address at some point in the future.

- [x] add 'factor' as a parameter on the zoom buttons.  Currently, it
  is hardcoded to 10%

- [x] review user-guide for documentation needs

- [x] ~~possibly rebase this into multiple commits~~. won't do since squash merge will be used per [this comment by mattpap](https://github.com/bokeh/bokeh/pull/4907#issuecomment-238055182)

Possible work for future (or already existing) tickets:
- Does it make sense to expose zoom-only mode on wheel-zoom? I
  personally find it awkward to use wheel-zoom when it pans and zooms
  at the same time.  This could be implemented easily now by
  programmatically eliding the 'center' parameter to scale_range().
  Perhaps it could be implemented as wheel-zoom button having a dropdown
  radio-button list like tap tool that lets the user choose to lock down the center of the plot and only scale in and out
    - I think this should be attempted in #5135. I made this comment [over there as well.](https://github.com/bokeh/bokeh/issues/5135#issuecomment-247895010)


timsnyder:
This should be labeled status:WIP.  I wanted to at least get the initial code I have working checked in for others to look at.  As the PR description states, there are quite a few odds and ends that need to be ironed out still.

bokehjs/src/coffee/models/tools/zoom_util.coffee
```diff
@@ -0,0 +1,86 @@
...
+  zoom_info = {
+    xrs: xrs
+    yrs: yrs
+    factor: factor
+  }
+  return zoom_info
+
+module.exports = {
+    scale_range: scale_range
+}
```
bryevdv: maybe move this to `src/coffee/util/zoom.coffee` ?

bokehjs/src/coffee/models/tools/gestures/wheel_zoom_tool.coffee
```diff
@@ -43,52 +53,14 @@ class WheelZoomToolView extends GestureTool.View
 
     factor  = @mget('speed') * delta
```
bryevdv: As long as we are in this code, every standard `@define` property can just be accessed with `@speed` (when `@` is the model) or `@model.speed` (when `@` is the view) No more `get` or `mget` needed

bokehjs/src/coffee/models/tools/zoom_util.coffee
```diff
@@ -0,0 +1,86 @@
...
+  # args passed as an object:
+  # * args.frame
+  # * args.factor      - scaling factor (clamped: -0.9 < args.factor < 0.9)
+  # * args.center      - center point for scaling
+  # * args.v_axis_only - if true only scale the vertical axis
+  # * args.h_axis_only - if true only scale the horizontal axis
+  frame = args.frame
+  factor = args.factor
+  v_axis_only = args?.v_axis_only ? false
+  h_axis_only = args?.h_axis_only ? false
```
bryevdv: in coffee script these can actually all go in the function parameter list

bokeh/models/tools.py
```diff
@@ -392,6 +392,41 @@ class BoxZoomTool(Drag):
...
+    """ *toolbar icon*: |zoom_in_icon|
+
+    The zoom-in tool allows users to click a button to zoom in
+    by a fixed amount.
+
+    .. |zoom_in_icon| image:: /_images/icons/ZoomIn.png
+        :height: 18pt
+
+    """
+    # TODO ZoomInTool dimensions should probably be constrained to be the same as ZoomOutTool
```
mattpap: Currently this is no tooling in bokeh to make this happen. Neither properties can depend on each other like this, nor there is support for button groups (otherwise there could be a `ZoomInOutTool`).

timsnyder: @mattpap, at the SciPy sprint when I started looking at this, I remember @bryevdv mentioning wanting to look into a way to bundle tools that should come together.  Since it impacts more than these new zoom buttons, maybe that should be broken out into its own issue.  I'd be happy to discuss and/or remove the TODO from the code.

It is also entirely possible that I'm making up this idea or that I misunderstood @bryevdv

bokehjs/src/coffee/models/tools/gestures/wheel_zoom_tool.coffee
timsnyder: @bryevdv, thanks for the tip.  I'll be on the lookout for that in the files I touch.

bokehjs/src/coffee/models/tools/gestures/wheel_zoom_tool.coffee
```diff
@@ -19,20 +19,20 @@ class WheelZoomToolView extends GestureTool.View
     @_scroll(e)
 
   _scroll: (e) ->
-    frame = @plot_model.get('frame')
+    frame = @plot_model.frame
     hr = frame.get('h_range')
     vr = frame.get('v_range')
```
timsnyder: @bokeh/core, I removed quite a few uses of `get()` and `mget()` as @bryevdv [requested in an earlier comment] (https://github.com/bokeh/bokeh/pull/4841#discussion_r71641321).  I was surprised that `CartesianFrame` wouldn't let me access `h_range` and `v_range` without `get()`.

Is this because `@define_computed_property()` and `@define()` properties are different somehow?  Or, am I missing something else?

bryevdv: Yes any computed properties will have to use `get`. Fortunately there are not many of them and the nominal plan is to get rid of computed properties entirely.

bokehjs/src/coffee/core/properties.coffee
```diff
@@ -154,6 +154,10 @@ class Instance extends simple_prop("Instance", (x) -> x.properties?)
 # TODO (bev) separate booleans?
 class Number extends simple_prop("Number", (x) -> _.isNumber(x) or _.isBoolean(x))
 
+# TODO extend Number instead of copying it's predicate
+#class Percent extends Number("Percent", (x) -> 0 <= x <= 1.0)
+class Percent extends simple_prop("Number", (x) -> (_.isNumber(x) or _.isBoolean(x)) and (0 <= x <= 1.0) )
```
timsnyder: Note the addition of a `Percent` property here.  Any suggestions  with the TODO would be appreciated.  Or, if people think that the predicate is fine the way it is, I can delete the TODO comment.

bryevdv: I think it looks fine, @mattpap is this the correct way to "extend" properties on the JS side?

mattpap:
```
diff --git a/bokehjs/src/coffee/core/properties.coffee b/bokehjs/src/coffee/core/properties.coffee
index 0e3854f..f266b1f 100644
--- a/bokehjs/src/coffee/core/properties.coffee
+++ b/bokehjs/src/coffee/core/properties.coffee
@@ -133,11 +133,14 @@ class Property extends Backbone.Model
 
 simple_prop = (name, pred) ->
   class Prop extends Property
-    toString: () -> "#{name}(obj: #{@get(obj).id}, spec: #{JSON.stringify(@spec)})"
+    name: name
+    pred: pred
+
+    toString: () -> "#{@name}(obj: #{@get(obj).id}, spec: #{JSON.stringify(@spec)})"
     validate: (value) ->
-      if not pred(value)
+      if not @pred(value)
         attr = @get('attr')
-        throw new Error("#{name} property '#{attr}' given invalid value: #{value}")
+        throw new Error("#{@name} property '#{attr}' given invalid value: #{value}")
 
 class Any extends simple_prop("Any", (x) -> true)
 
@@ -154,6 +157,10 @@ class Instance extends simple_prop("Instance", (x) -> x.properties?)
 # TODO (bev) separate booleans?
 class Number extends simple_prop("Number", (x) -> _.isNumber(x) or _.isBoolean(x))
 
+class Percent extends Number
+  name: "Percent"
+  pred: (value) -> super(value) and (0 <= value <= 1)
+
 class String extends simple_prop("String", _.isString)
 
 # TODO (bev) don't think this exists python side
@@ -290,6 +297,7 @@ module.exports =
   Location: Location
   Number: Number
   Int: Number                     # TODO
+  Percent   Orientation: Orientation
   RenderLevel: RenderLevel
   RenderMode: RenderMode: Percent
```

mattpap: Alternatively, we could make `simple_prop()` smarter about inheritance.

bokehjs/src/coffee/util/zoom.coffee
```diff
@@ -0,0 +1,80 @@
+scale_range = ({frame, factor, center, h_axis_only = false, v_axis_only = false}) ->
```
timsnyder: @birdsarah, @bryevdv , @mattpap  and anyone else who loves the tests...

I've been learning up on Mocha and sinon and looking at the tests that exist for things like CartesianFrame and Canvas.  It seems like since the current implementation of `scale_range` depends on getting calculated properties like `h_range` and `v_range` and the `mappers`, testing this code is going to be challenging.

1. Are there any JS tests that actually use a document with the solver and everything to test things like this?  Would we need something like CasperJS/PhosphorJS in order to do that?  Would it be worth adding those types of tests?
1. Would it be better to stub CartesianFrame to not require that the  
1. Should I refactor `scale_range` so that it is a pure helper and doesn't depend on having a "solved" CartesianFrame?  In other words, should I pass `h_range`, `v_range`, `x_mappers` and `y_mappers` instead of passing in `frame`?

bryevdv:
I think 3) is definitely a reasonable and good idea, that that function can be tested in isolation more simply. (which would be sufficient for me @birdsarah might have other thoughts)

timsnyder:
mattpap answered my question about history cleanliness in #4907. So I won't worry about rebasing and rely on the final squash merge.

bokeh/models/tools.py
canavandl: I'd remove the TODO for now. Future work around bundling tools together should handle coalescing dimensions or other related properties, but we'll handle it then.

bokehjs/src/coffee/util/zoom.coffee
canavandl: I *think* (always dangerous) that you shouldn't have to worry about stubbing any solver/canvas things because the computed ranges that you're dealing with are all model properties (and not view properties, which involve the solver and canvas stuff).

I added some tests to update the ranges for a wheel pan tool that you should be able to crib: https://github.com/bokeh/bokeh/blob/master/bokehjs/test/models/tools/gestures/wheel_pan_tool.coffee

canavandl:
hi @timsnyder 

this PR is really close. All there's left is:
* merging in master and dealing with whatever conflicts (probably related to set/get stuff)
* writing a couple of tests JS tests
* perhaps dealing with the new `Percent` property (I need to review the code more about that)

I would like to use this feature (in/out zoom tool) in a project. Do you have an idea of whether or not you'll have the availability to make the last couple of changes? If you're not going to be free in the next couple of days, would you mind adding me as a collaborator to your fork (https://github.com/blog/2247-improving-collaboration-with-forks) or letting me branch/supercede your PR?

Thanks again - this is going to be great!

timsnyder:
@canavandl, I aim to definitely wrap it up this weekend. Might get to it before then but I went ahead and enabled the collabo checkbox. If you want to take it across the finish line, feel free.

bokeh/models/tools.py
timsnyder: Sounds good.  I agree with you and mattpap that we shouldn't try to do all of the things here.

bokehjs/src/coffee/util/zoom.coffee
```diff
@@ -0,0 +1,80 @@
...
+  # Util for Tools to Scale ranges of a frame for use by PlotCanvasView.update_range()
+  #
+  # Parameters:
+  # * frame
+  # * factor      - scaling factor (clamped: -0.9 < args.factor < 0.9)
+  # * center      - center point for scaling
+  # * v_axis_only - if true only scale the vertical axis
+  # * h_axis_only - if true only scale the horizontal axis
+
+  hr = frame.get('h_range')
```
birdsarah: Any use of get also need replacing with plain `frame.h_range` etc.

birdsarah:
The following diff gets this PR working (and adds zoom_in, zoom_out tools to the color_scatter example):

```diff
diff --git a/bokehjs/src/coffee/models/tools/actions/zoom_in_tool.coffee b/bokehjs/src/coffee/models/tools/actions/zoom_in_tool.coffee
index c7c3af9..f6f1524 100644
--- a/bokehjs/src/coffee/models/tools/actions/zoom_in_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/actions/zoom_in_tool.coffee
@@ -33,18 +33,9 @@ class ZoomInTool extends ActionTool.Model
   tool_name: "ZoomIn"
   icon: "bk-tool-icon-zoom-in"
 
-
-  initialize: (attrs, options) ->
-    super(attrs, options)
-
-    @override_computed_property('tooltip', () ->
-        @_get_dim_tooltip(
-          @tool_name,
-          @_check_dims(@dimensions, "zoom-in tool")
-        )
-      , false)
-    @add_dependencies('tooltip', this, ['dimensions'])
-
+  @getters {
+    tooltip: () -> @_get_dim_tooltip(@tool_name, @_check_dims(@dimensions, "zoom-in tool"))
+  }
   @define {
     factor: [ p.Percent, 0.1 ]
     dimensions: [ p.Array, ["width", "height"] ]
diff --git a/bokehjs/src/coffee/models/tools/actions/zoom_out_tool.coffee b/bokehjs/src/coffee/models/tools/actions/zoom_out_tool.coffee
index 30bc1a3..c66a41b 100644
--- a/bokehjs/src/coffee/models/tools/actions/zoom_out_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/actions/zoom_out_tool.coffee
@@ -33,18 +33,9 @@ class ZoomOutTool extends ActionTool.Model
   tool_name: "ZoomOut"
   icon: "bk-tool-icon-zoom-out"
 
-
-  initialize: (attrs, options) ->
-    super(attrs, options)
-
-    @override_computed_property('tooltip', () ->
-        @_get_dim_tooltip(
-          @tool_name,
-          @_check_dims(@dimensions, "zoom-out tool")
-        )
-      , false)
-    @add_dependencies('tooltip', this, ['dimensions'])
-
+  @getters {
+    tooltip: () -> @_get_dim_tooltip(@tool_name, @_check_dims(@dimensions, "zoom-out tool"))
+  }
   @define {
     factor: [ p.Percent, 0.1 ]
     dimensions: [ p.Array, ["width", "height"] ]
diff --git a/bokehjs/src/coffee/util/zoom.coffee b/bokehjs/src/coffee/util/zoom.coffee
index bfc868b..b963b8d 100644
--- a/bokehjs/src/coffee/util/zoom.coffee
+++ b/bokehjs/src/coffee/util/zoom.coffee
@@ -8,8 +8,8 @@ scale_range = ({frame, factor, center, h_axis_only = false, v_axis_only = false}
   # * v_axis_only - if true only scale the vertical axis
   # * h_axis_only - if true only scale the horizontal axis
 
-  hr = frame.get('h_range')
-  vr = frame.get('v_range')
+  hr = frame.h_range
+  vr = frame.v_range
 
 
   # clamp the  magnitude of factor, if it is > 1 bad things happen
@@ -18,11 +18,11 @@ scale_range = ({frame, factor, center, h_axis_only = false, v_axis_only = false}
   else if factor < -0.9
     factor = -0.9
 
-  vx_low  = hr.get('start')
-  vx_high = hr.get('end')
+  vx_low  = hr.start
+  vx_high = hr.end
 
-  vy_low  = vr.get('start')
-  vy_high = vr.get('end')
+  vy_low  = vr.start
+  vy_high = vr.end
 
   frame_center = {
     x: (vx_high + vx_low) / 2.0
@@ -56,12 +56,12 @@ scale_range = ({frame, factor, center, h_axis_only = false, v_axis_only = false}
     sy1 = vy_high
 
   xrs = {}
-  for name, mapper of frame.get('x_mappers')
+  for name, mapper of frame.x_mappers
     [start, end] = mapper.v_map_from_target([sx0, sx1], true)
     xrs[name] = {start: start, end: end}
 
   yrs = {}
-  for name, mapper of frame.get('y_mappers')
+  for name, mapper of frame.y_mappers
     [start, end] = mapper.v_map_from_target([sy0, sy1], true)
     yrs[name] = {start: start, end: end}
 
diff --git a/examples/plotting/file/color_scatter.py b/examples/plotting/file/color_scatter.py
index f9b5cd9..08adfe8 100644
--- a/examples/plotting/file/color_scatter.py
+++ b/examples/plotting/file/color_scatter.py
@@ -10,7 +10,7 @@ colors = [
     "#%02x%02x%02x" % (int(r), int(g), 150) for r, g in zip(50+2*x, 30+2*y)
 ]
 
-TOOLS="hover,crosshair,pan,wheel_zoom,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"
+TOOLS="hover,crosshair,pan,wheel_zoom,zoom_in,zoom_out,box_zoom,undo,redo,reset,tap,save,box_select,poly_select,lasso_select"
 
 p = figure(tools=TOOLS)
 

```

birdsarah:
The fact that this PR passed the tests in-spite of not actually working in practice against master indicates that  more tests are needed. A simple selenium test should work. Click on the zoom in button and test the range - you can see other examples of selenium tests that check the range.

birdsarah:
I should add I'm really excited to see this feature finally come into Bokeh and am happy to help however I can especially with selenium tests.

canavandl:
I have unit tests that I need to push.

bokehjs/src/coffee/util/zoom.coffee
```diff
@@ -56,15 +53,17 @@ scale_range = ({frame, factor, center, h_axis_only = false, v_axis_only = false}
...
     [start, end] = mapper.v_map_from_target([sx0, sx1], true)
     xrs[name] = {start: start, end: end}
 
   yrs = {}
-  for name, mapper of frame.get('y_mappers')
+  for name, mapper of frame.y_mappers
     [start, end] = mapper.v_map_from_target([sy0, sy1], true)
     yrs[name] = {start: start, end: end}
 
+  debugger;
```
birdsarah: ????? @canavandl

canavandl: wow. rookie mistake

bryevdv: we could add a check for debugger anywhere in the code quality tests

bryevdv: then mumble mumble pre-commit hook mumble mumble

canavandl: haha

birdsarah: > then mumble mumble pre-commit hook mumble mumble

I think the command is `git mumble mumble pre-commit hook mumble mumble`

bokehjs/src/coffee/models/tools/actions/zoom_in_tool.coffee
```diff
@@ -0,0 +1,56 @@
...
+  default_view: ZoomInToolView
+  type: "ZoomInTool"
+  tool_name: "ZoomIn"
+  icon: "bk-tool-icon-zoom-in"
+
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
+
+    @override_computed_property('tooltip', () ->
```
canavandl: resolved

bokehjs/src/coffee/models/tools/actions/zoom_out_tool.coffee
```diff
@@ -0,0 +1,56 @@
...
+  default_view: ZoomOutToolView
+  type: "ZoomOutTool"
+  tool_name: "ZoomOut"
+  icon: "bk-tool-icon-zoom-out"
+
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
+
+    @override_computed_property('tooltip', () ->
```
canavandl: resolved

bokehjs/src/coffee/util/zoom.coffee
canavandl: resolved

birdsarah:
Tested after luke's changes and looking good :)

bryevdv:
Thanks for the PR @timsnyder !

timsnyder:
You're welcome! Thanks everyone for the help wrapping it up!

PR #4851, opened 2016-07-20
Canavandl/4848 refactor legend as guide

### LINKED ISSUE #4848: create abstract base class for Legend and ColorBar (and future similar classes)
ggplot describes legends and colorbars as guides (http://docs.ggplot2.org/current/guides.html).

Bokeh should refactor the `Legend` model to be a subclass of GuideRenderer (currently it's a subclass of `Annotation`). This will simplify the addition of a `ColorBar` model, as ColorBar and Legend can share some code (off-axis placement, etc) and make the docs easier to navigate.

Eventually (with the addition of a ColorBar), we would add an abstract `Legend` class.

canavandl:
Ground work to support adding ColorBar (also intended to be a Guide subclass)

- [x] issues: fixes #4848
- [x] tests added / passed
- n/a: release document entry (if new feature or API change)

Todo:
* ~~add compat/deprecation warn for importing legend from bokeh.models.annotations~~


bryevdv:
Two comments: 

* how about `legends.py` ? the other guides are in the their own modules, and also I can imagine having specialized legend subclasses. `legends.py` gives some headroom @birdsarah thoughts?

* the transitive import for compat is breaking the refguide build, will take some though

canavandl:
If we move bokeh.models.guides.Legend -> bokeh.models.legends.Legend, where would you propose putting a ColorBar? in legends.py also? in.... colorbars.py?

bryevdv:
ah, right... yah I think just `colorbar.py` maybe, thoughts @birdsarah ?

bryevdv:
The other alternative is to try and consolidate everything ala `annotations.py` but that's a bigger undertaking...

canavandl:
@birdsarah and I agreed to move Legend to bokeh.models.legends.Legend (and then to consider a ColorBar a type of legend and put it in the same module in the future)

canavandl:
Need to add off-plot support:
https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/annotations/annotation.coffee#L34

Issue #4813, opened 2016-07-15
Render bokeh control inside bootstrap carousel


jdimatteo:
Bokeh controls render in the wrong place in a bootstrap carousel.

A couple workarounds: `.bk-toolbar-wrapper.bk-layout-null{ position: absolute; }` or `toolbar_location="above"`

This is described in some more detail here:
http://stackoverflow.com/questions/38388676/render-bokeh-control-inside-bootstrap-carousel

This is reproducible here: https://github.com/jdimatteo/flask_bokeh_app/tree/carousel

bryevdv:
@jdimatteo does explicitly setting `toolbar_location="right"` also solve the problem? If so, my guess is the default is getting set as `None` which translates to `null` on the JS side, and then `None`/`null` is interpreted as "right side" in some places, but also allowing the `null` to make it into the templated CSS class.  If so that should be easy to fix.

jdimatteo:
@bryevdv no, explicitly setting `toolbar_location="right"` does not fix the problem.  Sorry for the late response, and thanks for following up on this

bryevdv:
There's pretty much zero chance that I will be able to fix this, and it seems to not be a common case. Now with PNG/SVG export, those actually seem like better options for a carousel anyway, so closing this.

PR #4856, opened 2016-07-21
Use correct toolbar positioning and not hacky css

### LINKED ISSUE #4853: Hovertool does not show tooltip of last glyph
See the example on the website: 
http://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#hovertool

When hovering over the tooltip of the right-most circle (index 4), the tooltip does not show.

This does not work in Bokeh 0.12.0 but used to work correctly in previous versions as can be verified by going to the user guide webpages of earlier versions such as 0.11.1

birdsarah:
Fixes #4853 

I have tested this manually, I don't think we have any other tests for this - and I'm not sure what that would be off the top of my head.

Note to test this thoroughly you need to test all combinations of:
* sizing_modes 
* toolbar in below and right position (note that the below toolbar was suffering from the same problem as right toolbar, but no-one had caught it yet)
* toolbar_sticky is True and False
* single plots and more complex layouts

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot.coffee
+++ b/bokehjs/src/coffee/models/plots/plot.coffee
@@ -226,10 +226,10 @@ class Plot extends LayoutDOM.Model
       if @toolbar_location is 'below'
         # (3) stack: plot_canvas._dom_top = toolbar._bottom - toolbar._height
         if @toolbar_sticky is false
-          constraints.push(EQ(@toolbar._dom_top, [-1, @plot_canvas._height], @toolbar._bottom))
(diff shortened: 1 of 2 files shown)


bryevdv:
I'm not sure if it is related btw but sometimes the CSS shows up `bk-toolbar-null`

birdsarah:
That's some code we could clean-up but is nothing to worry about. I don't believe you've had any performance problems even if that's the class.

bryevdv:
> That's some code we could clean-up but is nothing to worry about. I don't believe you've had any performance problems even if that's the class.

I think it was causing the problems in #4813

PR #4857, opened 2016-07-22
Correctly perform resize

### LINKED ISSUE #4560: Resize Tool Mostly Broken
Following Issues I have had with resize tool:

- 0.11.1, possibly earlier: Resizing an axis to a near zero values will break the plot (resetting does not help, only reloading the page does), immediately attempting to resize another plot will break it as well (in dev12 all plots are broken); Error: Unkown Constraint

- 0.12.0dev12: Resizing plot blurs it across the original area

- 0.12.0dev12: Tools can still be used in the initial area even if the plot has been scaled to a smaller area

- 0.12.0dev12: Scaling to a larger size clips everything outside the initial plot area

- 0.12.0dev12, possibly earlier dev builds: Resizing on a grid plot does not shift other plots

- 0.12dev12, possibly earlier dev builds: Resetting after resizing a plot snaps the toolbar to the resized area instead of the reset area; resetting again fixes this

screen shot 2016-06-17 at 1 35 00 pm

birdsarah:
Fixes #4560 

ToDo:
- [x] write a test (could also write tests for reset and that true/false on reset_size works)
- [ ] discuss what to do when plot is in a layout

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -269,7 +269,17 @@ class PlotCanvasView extends Renderer.View
       @canvas_view.set_dims([info.dimensions.width, info.dimensions.height])
 
   reset_dimensions: () ->
-    @canvas_view.set_dims([@canvas.initial_width, @canvas.initial_height])
+    @model.plot.width = @canvas.initial_width
+    @model.plot.height = @canvas.initial_height
+    @model.document.resize()
+
+  update_dimensions: (width, height) ->
+    @pause()
+    @model.plot.width = width
+    @model.plot.height = height
+    @model.document.resize()
+    @unpause()
+
 
   get_selection: () ->
     selection = []
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -269,9 +269,7 @@ class PlotCanvasView extends Renderer.View
       @canvas_view.set_dims([info.dimensions.width, info.dimensions.height])
 
   reset_dimensions: () ->
-    @model.plot.width = @canvas.initial_width
-    @model.plot.height = @canvas.initial_height
-    @model.document.resize()
+    @update_dimensions(@canvas.initial_width, @canvas.initial_height)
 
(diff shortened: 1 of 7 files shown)


birdsarah:
@bryevdv - this could go in as is, but could be improved. when testing note that the resize tool is "hidden" from interaction for same reason as hover - fixed in #4856.

In particular, I think we need some kind of `recompute_layout` method on the document.

In the case of fixed layouts: The problem currently is that if the plot being resized is in, for example, a row then we need to (I think) do a `build_child_views` on the row. Of course if that row is part of a more complex layout then we need to recompute the whole layout.

In the case of scale_width layout. Resize is only allowing a resize of the height.

My suggestion would be to add tests to this and merge it, as it's a definite improvement. Then document the layout issues in another issue.

bryevdv:
Your plan seems OK to me but I defer to you in any case. I do wonder about the `build_child_views` that seems a pretty blunt way to do things but maybe its best to start with blunt and simple.

birdsarah:
I've added tests and so marking this as ready. Unfortunately, some of these tests are plot_canvas_view tests that have previously given us intermittent failures due to solver variables, so we may have to skip them for now till we resolve that testing issue.

birdsarah:
I've added a new issue to document the layout problems.

bryevdv:
Failing for flake test. I will fix up and merge. Highly recommend the git commit hook I added to the dev guide :)

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -269,7 +269,17 @@ class PlotCanvasView extends Renderer.View
       @canvas_view.set_dims([info.dimensions.width, info.dimensions.height])
 
   reset_dimensions: () ->
-    @canvas_view.set_dims([@canvas.initial_width, @canvas.initial_height])
+    @model.plot.width = @canvas.initial_width
+    @model.plot.height = @canvas.initial_height
+    @model.document.resize()
```
bryevdv: does this need `pause`/`unpause` too? Or are they not needed below?

birdsarah: I have no idea. I changed this so that reset_dimensions would work. It didn't have a pause() in there before, so I didn't add one. We could.

Issue #4863, opened 2016-07-22
Wheel Zoom tool on Microsoft Edge


birdsarah:
On Bokeh 0.12.0 on Microsoft Surface with Edge, Wheel Zoom tool is not responding to pinch or zooming.

birdsarah:
#4864 partially fixes this and means now that pinch zoom is working at least.

mattpap:
I will assume this refers to the legacy version of Edge browser (EdgeHTML engine). If that's the case, then this issue won't be relevant anymore in bokeh 3.0. If the issue persist on newer versions of Edge, please reopen.

PR #4864, opened 2016-07-22
Support wheelzoom on dual touch monitors better

### LINKED ISSUE #4862: Wheel zoom not working on chrome on touchscreen laptops, when using scroll wheel
This has been reported in a number of places under different titles (#1245, #4762, ...) but I'm opening a new issue to track this specific problem and its fix.

If a user is using:
* Chrome 
* on a device that supports both touch and mouse wheel events (like a Microsoft Surface or a Lenovo touchscreen laptop - these are often windows systems but this is a coincidence I believe)

Then **when wheel zoom tool is enabled it only responds to pinch and not to the mouse scroll.**

We have had other touch and sometimes chrome related fixes before:
* Toolbar buttons not being able to be selected
* Weird plot selection stuff
I'm not sure we've ever specifically fixed this problem though.

birdsarah:
Fixes #4862 

Also partially fixes wheel zoom tool on Edge #4863 - WheelZoom Tool now enables pinch.

### DIFF
--- a/bokehjs/src/coffee/common/ui_events.coffee
+++ b/bokehjs/src/coffee/common/ui_events.coffee
@@ -90,7 +90,7 @@ class UIEvents extends Backbone.Model
     # This is a hack for laptops with touch screen who may be pinching or scrolling
     # in order to use the wheel zoom tool. If it's a touch screen the WheelZoomTool event
     # will be linked to pinch. But we also want to trigger in the case of a scroll.
-    if 'ontouchstart' of window or navigator.maxTouchPoints
+    if 'ontouchstart' of window or navigator.maxTouchPoints > 0
       if et == 'pinch'
         logger.debug("Registering scroll on touch screen")
         tool_view.listenTo(@, "scroll:#{id}", tool_view["_scroll"])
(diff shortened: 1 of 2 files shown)


birdsarah:
@sarahheim are you able to test this branch to see if it gets wheel zoom working on your laptop?

bokehjs/src/coffee/common/ui_events.coffee
```diff
@@ -88,11 +86,30 @@ class UIEvents extends Backbone.Model
       logger.debug("Registering tool: #{type} for event 'doubletap'")
       tool_view.listenTo(@, "doubletap", tool_view._doubletap)
 
+    # Dual touch hack part 1/2
+    # This is a hack for laptops with touch screen who may be pinching or scrolling
+    # in order to use the wheel zoom tool. If it's a touch screen the WheelZoomTool event
+    # will be linked to pinch. But we also want to trigger in the case of a scroll.
+    if 'ontouchstart' of window or navigator.maxTouchPoints
```
mattpap: Shouldn't this line contain `> 0` as in the other cases? Perhaps an utility function could be added to avoid duplication of logic.

birdsarah: Nice catch - thank you!

PR #4887, opened 2016-07-26
Add hit testing support for multiline

### LINKED ISSUE #3110: multi_line and selection callback
Hi,

Is ``multi_line`` supposed to work with a selection callback? For example, when I draw a point plot from a data source, attaching a selection callback to the data source works well, i.e. it returns the index of the selected point, but when I use ''multi_line'' for the same data source and click a line, the callback gets nothing from ``cb_obj.get('selected')``.

canavandl:
Adds hit test support for MultiLine glyph

- [x] issues: fixes #3110

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -22,7 +22,7 @@ class DataSource(Model):
     selected = Dict(String, Dict(String, Any), default={
         '0d': {'glyph': None, 'indices': []},
         '1d': {'indices': []},
-        '2d': {'indices': [], 'point_indices': {}}
+        '2d': {}
     }, help="""
     A dict to indicate selected indices on different dimensions on this DataSource. Keys are:
 
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -22,7 +22,7 @@ class DataSource(Model):
     selected = Dict(String, Dict(String, Any), default={
         '0d': {'glyph': None, 'indices': []},
         '1d': {'indices': []},
-        '2d': {'indices': []}
+        '2d': {'indices': [], 'point_indices': {}}
     }, help="""
     A dict to indicate selected indices on different dimensions on this DataSource. Keys are:
 
--- a/bokehjs/src/coffee/common/hittest.coffee
+++ b/bokehjs/src/coffee/common/hittest.coffee
@@ -35,7 +35,13 @@ class HitTestResult
       indices: []
     }
     # 2d for all for multilines and multi patches
-    @['2d'] = {indices: []}
+    @['2d'] = {
+      # index of the glyph in the multiglyph
+      indices: []
+      # object containing glyph_index: [point_index] of the nearest
+      # point to the crossed segment
+      point_indices: {}
+    }
 
   Object.defineProperty(this.prototype, '_0d', { get: () -> @['0d'] })
   Object.defineProperty(this.prototype, '_1d', { get: () -> @['1d'] })
--- a/bokehjs/src/coffee/common/hittest.coffee
+++ b/bokehjs/src/coffee/common/hittest.coffee
@@ -36,11 +36,8 @@ class HitTestResult
     }
     # 2d for all for multilines and multi patches
     @['2d'] = {
-      # index of the glyph in the multiglyph
-      indices: []
-      # object containing glyph_index: [point_index] of the nearest
-      # point to the crossed segment
-      point_indices: {}
+      # mapping of indices of the multiglyph to array of glyph indices that were hit
+      # e.g. {3: [5, 6], 4, [5]}
     }
 
   Object.defineProperty(this.prototype, '_0d', { get: () -> @['0d'] })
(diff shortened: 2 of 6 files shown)


bokehjs/src/coffee/models/glyphs/multi_line.coffee
```diff
@@ -41,6 +42,79 @@ class MultiLineView extends Glyph.View
...
+      Args:
+        * geometry (object): object with the following keys
+          * vx (float): view x coordinate of the point
+          * vy (float): view y coordinate of the point
+          * type (str): type of geometry (in this case it's a point)
+      Output:
+        Object with the following keys:
+          * 0d (bool): whether the point hits the glyph or not
+          * 1d (array(int)): array with the indices hit by the point
+    ###
```
canavandl: need to update docstring

bokehjs/src/coffee/models/glyphs/multi_line.coffee
```diff
@@ -41,6 +42,79 @@ class MultiLineView extends Glyph.View
...
+      val = @renderer.ymapper.map_from_target(vy)
+      values = @_ys
+    else
+      val = @renderer.xmapper.map_from_target(vx)
+      values = @_xs
+
+    for i in [0...values.length]
+      for j in [0...values[i].length-1]
+        if values[i][j] <= val <= values[i+1]
+          result['2d'].indices.push(i)
```
canavandl: should also set `result['2d'].point_indices` using nearest points

canavandl:
From the example: 

screen shot 2016-07-28 at 2 58 26 pm

(There's also a configured HoverTool)

canavandl:
I added an example to examples/plotting/file. Currently we don't test hit-testing very well - we test the tap and hover tools, but not the hit-test boundaries of glyphs - so I'll open an issue to begin discussion about that.

bokehjs/src/coffee/common/hittest.coffee
```diff
@@ -35,7 +35,13 @@ class HitTestResult
...
     }
     # 2d for all for multilines and multi patches
-    @['2d'] = {indices: []}
+    @['2d'] = {
+      # index of the glyph in the multiglyph
+      indices: []
+      # object containing glyph_index: [point_index] of the nearest
+      # point to the crossed segment
+      point_indices: {}
+    }
```
canavandl: I think this is the only "controversial" part. There has to be an "indices"/array pair to support selections, but it'd be useful to know which points inside a line are selected. My solution was the `point_indices` object which contains the multi_line index and an array of line points.

I'm open to feedback on changing the implementation and/or naming.

bryevdv: we should talk this over. My original intention was that for things like `MultiLine`:

* 1d indices would say "which line" 
* 2d indices would say "which segment" on the lines

Also instead of "nearest" I think for consistency we chose the first point of the segment. But thinking about it now, maybe this would work:
```
@['2d'] = {
    3 : [5],
    6 : [8, 20, 55]
}
```
that would say: lines 3 and 8 were hit, and the values give the segment points. I think we'd still want to fill in the "1d" attr as well (with 3 and 8)

canavandl: >
1d indices would say "which line"
2d indices would say "which segment" on the lines

The hittesting module didn't seem written with that pattern in mind (it was performing lookups for `results['2d'].indices`, but I see thats how hit-testing for patches was implemented. So I changed the multiline hit-testing to match and cleaned up some old code.

I like having the '2d' value be an object better anyway.

canavandl:
>
1d indices would say "which line"
2d indices would say "which segment" on the lines

The hittesting module didn't seem written with that pattern in mind (it was performing lookups for `results['2d'].indices`, but I see thats how hit-testing for patches was implemented. So I changed the multiline hit-testing to match and cleaned up some old code.

I like having the '2d' value be an object better anyway.

bokeh/models/sources.py
```diff
@@ -22,7 +22,7 @@ class DataSource(Model):
     selected = Dict(String, Dict(String, Any), default={
         '0d': {'glyph': None, 'indices': []},
         '1d': {'indices': []},
-        '2d': {'indices': []}
+        '2d': {}
```
canavandl: '2d' 's value is only an object now

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -139,8 +139,6 @@ class GlyphRendererView extends Renderer.View
         selected = indices
       else if selected['1d'].indices.length > 0
         selected = selected['1d'].indices
-      else if selected['2d'].indices.length > 0
```
canavandl: example of old code that was looking for `indices` attr on `'2d'` element

jbednar:
Looks great to me!

bryevdv:
I've added `MIGRATION` tag since this could potentially affect people (I doubt many, but still). This LGTM but would appreciate a few more looks @birdsarah @fpliger @mattpap please comment if you can.

birdsarah:
I will take a look today or over weekend.

PR #4894, opened 2016-07-27
Use of the query string

### LINKED ISSUE #4893: Support access to the request (used to create the session) in bokeh serve.
I would like to have access to the request that was used to create the session in bokeh server. 
I want to select data, based on url query parameters in the request url. For example I want the url `/plot?station=1` to show data for station 1. I found that using the url query parameters is under [discussion/consideration](https://github.com/bokeh/bokeh/wiki/Use-of-the-query-string-in-url-to-control-plot-properties) and was already mentioned in a few issues. To facilitate this discussion I implemented it in the form that I expect it to work. 

The implementation I propose would stores the tornado request in the document. The query parameters are accessible by using the document.request.arguments. 
I can imagine that some parameters can be used as plot properties. I would expect these to be prefixed by "bokeh-", like the session and unlike webgl=1 (which I would replace by bokeh-webgl=1). 

I will create a corresponding pull request which can be used for discussion. 

Bokeh info:
```
Python version      :  3.5.2 (default, Jun 27 2016, 03:10:38)
IPython version     :  4.2.0
Bokeh version       :  0.12.1rc1-8-g90cc3f5
BokehJS static path :  ~/.virtualenvs/main/lib/python3.5/site-packages/bokeh/server/static

```

SiggyF:
-  issues: fixes #4893
- no tests added yet, discuss api first
- no release document entry yet, for discussion

### DIFF
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -287,6 +287,7 @@ class Document(object):
         self._theme = kwargs.pop('theme', default_theme)
         # use _title directly because we don't need to trigger an event
         self._title = kwargs.pop('title', DEFAULT_TITLE)
+        self._request = kwargs.pop('request', None)
         self._template = FILE
 
         # TODO (bev) add vars, stores
@@ -383,6 +384,11 @@ class Document(object):
         if self._title != title:
             self._title = title
             self._trigger_on_change(TitleChangedEvent(self, title))
+
+    @property
+    def request(self):
+        """the request that was used to create this document (server mode)"""
+        return self._request
 
     @property
     def template(self):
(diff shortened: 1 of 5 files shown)


bryevdv:
@SiggyF there's already and issue (#4828) and a PR (#4858) open for this. If you want to take over that would actually be really helpful, but I think we need to proceed along the lined outlined by @havocp  in the other PR comments e.g. putting the request on the session context, and putting that on the document (when it applies).

SiggyF:
Thanks for the comments. The other PR #4858 looks fine to me, please close this one. I'll try and help out putting the request arguments on the session context and the session context on the document. 
One other difference was that I exposed the `request` instead of only the `request.arguments`  PR #4858. I have no preference either way.

SiggyF:
Closing as discussed.

PR #4898, opened 2016-07-28
Version Mismatch Fix

### LINKED ISSUE #4882: Fixed version in version.coffee causes constant "version mismatch" warning
I'm not sure what the intent was behind this, but in development mode this warning is useless. I presume `version.coffee` should be generated/templated and should include the actual version of bokehjs. Otherwise we can just skip it (in development mode).

brittainhard:
Fixes the problem of tagging between Bokeh and BokehJS. Development builds now also run `scripts/version_update.py`, to change the value of the version strings in `version.coffee` and `package.json`. 

Submitting a new release candidate or development release now requires a previous version argument in addition to a new version argument. 

fixes #4882

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -573,7 +573,7 @@ class Document
       throw new Error("JSON object has wrong type #{typeof json}")
     py_version = json['version']
     versions_string = "Library versions: JS (#{js_version})  /  Python (#{py_version})"
-    if js_version != py_version
+    if py_version.indexOf('-') < 0 and js_version != py_version
       logger.warn("JS/Python version mismatch")
       logger.warn(versions_string)
     else
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -573,7 +573,7 @@ class Document
       throw new Error("JSON object has wrong type #{typeof json}")
     py_version = json['version']
     versions_string = "Library versions: JS (#{js_version})  /  Python (#{py_version})"
-    if js_version.split('-')[0] != py_version.split('-')[0]
+    if js_version != py_version
       logger.warn("JS/Python version mismatch")
       logger.warn(versions_string)
     else
--- a/bokehjs/test/document.coffee
+++ b/bokehjs/test/document.coffee
@@ -545,16 +545,32 @@ describe "Document", ->
     logging.set_log_level("warn")
     json = d.to_json_string()
     parsed = JSON.parse(json)
-    parsed['version'] = "0.0.1"
+    parsed['version'] = "#{js_version}"
     out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
-    expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (0.0.1)\n"
-
-    parsed['version'] = "#{js_version}-foo"
-    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
     expect(out).to.be.equal ""
 
+    parsed['version'] = "0.0.1"
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (#{parsed["version"]})\n"
+
     parsed['version'] = "#{js_version}rc123"
-    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (#{parsed["version"]})\n"
+
+    parsed['version'] = "#{js_version}dev123"
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (#{parsed["version"]})\n"
+
+    parsed['version'] = "#{js_version}-foo"
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal ""
+
+    parsed['version'] = "#{js_version}rc123-foo"
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal ""
+
+    parsed['version'] = "#{js_version}dev123-bar"
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
     expect(out).to.be.equal ""
 
   it "can serialize with one model in it", ->
--- a/bokehjs/test/document.coffee
+++ b/bokehjs/test/document.coffee
@@ -553,6 +553,10 @@ describe "Document", ->
     out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
     expect(out).to.be.equal ""
 
+    parsed['version'] = "#{js_version}rc123"
+    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal ""
+
   it "can serialize with one model in it", ->
     d = new Document()
     expect(d.roots().length).to.equal 0
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -84,18 +84,18 @@ elif [[ ! -z "$dtag" && ! -z "$ptag" && -z "$rtag" ]]; then
     # Merge branch into master and push to origin
     git checkout master
     git pull origin
-    git merge --no-ff release_$rtag -m "Merge branch devel_$rtag"
+    git merge --no-ff devel_$dtag -m "Merge branch devel_$dtag"
     git push origin master
-    git branch -d release_$rtag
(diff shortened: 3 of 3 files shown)


bokehjs/src/coffee/document.coffee
```diff
@@ -573,7 +573,7 @@ class Document
       throw new Error("JSON object has wrong type #{typeof json}")
     py_version = json['version']
     versions_string = "Library versions: JS (#{js_version})  /  Python (#{py_version})"
-    if js_version.split('-')[0] != py_version.split('-')[0]
+    if js_version != py_version
```
brittainhard: None of the release tags I've seen use a dash in the name. Regardless, if we are calling `version_update.py` it will automatically change the text values of `version.coffee` and `package.json`, and so all that is needed is a simple equality check.

scripts/deploy.sh
```diff
@@ -71,15 +81,21 @@ elif [[ ! -z "$dtag" && -z "$ptag" && -z "$rtag" ]]; then
...
     fi
 
-    # tag it locally
+    # Merge branch into master and push to origin
+    git checkout master
+    git pull origin
+    git merge --no-ff release_$rtag -m "Merge branch devel_$rtag"
+    git push origin master
+    git branch -d release_$rtag
+
```
brittainhard: @bryevdv we talked about adding this code into creating dev versions, but it seems like it's doing too much. I kind of feel like all we need to do here is call `version_update.py` to update the version strings.

bokehjs/src/coffee/document.coffee
bryevdv: unreleased dev versions will have a dash, though, and we need to not-do this test in that case. This would be the case e.g. running off master with inline resources.

scripts/deploy.sh
bryevdv: you might be right, @damianavila can you comment? One purpose of this PR is to make sure the BokehJS `version` is correct for *all* published artifacts, including dev builds and RCs

scripts/deploy.sh
```diff
@@ -71,15 +81,21 @@ elif [[ ! -z "$dtag" && -z "$ptag" && -z "$rtag" ]]; then
...
     git tag -a $dtag -m "New devel[rc] build $dtag."
+    git push origin $rtag
 
-    # and push the tag
-    git push origin $dtag
     echo "The new devel build was triggered."
 
 else
-    echo "You have to pass a -d tag (dev build) OR -p and -r tags (for releases)."
+    echo "You must pass a -d tag (dev build) OR -r tags (for releases) and a -p tag (previous version)."
```
bryevdv: I think this would read more clearly (paraphrased): "a -d tag and a -p tag OR a -r tag and -p tag" Basically with the giant `OR` it doesn't seem like the "and a -p tag" distributes over both clauses.

scripts/deploy.sh
bryevdv: Actually I do think this is necessary --- because we do need to commit/push the updated files with the updated versions. But I think `release_$rtag` needs to be `devel_$dtag` ?

scripts/deploy.sh
```diff
@@ -71,15 +81,21 @@ elif [[ ! -z "$dtag" && -z "$ptag" && -z "$rtag" ]]; then
...
-    # tag it locally
+    # Merge branch into master and push to origin
+    git checkout master
+    git pull origin
+    git merge --no-ff release_$rtag -m "Merge branch devel_$rtag"
+    git push origin master
+    git branch -d release_$rtag
+
+    # Tag the version locally.
     git tag -a $dtag -m "New devel[rc] build $dtag."
```
bryevdv: maybe `"New devel/rc build $dtag."` not sure what square brackets there would signify.

bokehjs/src/coffee/document.coffee
bryevdv: So maybe (untested): `if py_version.indexOf('-') < 0 and (js_version != py_version)`

scripts/deploy.sh
damianavila: > Actually I do think this is necessary --- because we do need to commit/push the updated files with the updated versions

Yep, I think so...

damianavila:
Btw, at a quick view, now both that `if` branch and the `else if` branch (for dev and rel) are pretty much the same, maybe you could abstract the general thing into a function...

brittainhard:
@damianavila i'll take a swing at that.

bokehjs/test/document.coffee
```diff
@@ -553,6 +553,10 @@ describe "Document", ->
     out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
     expect(out).to.be.equal ""
 
+    parsed['version'] = "#{js_version}rc123"
+    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
+    expect(out).to.be.equal ""
```
bryevdv: I don't understand this test, or why it is passing. This is explicitly setting the python and js versions to be different, supposedly on an RC, so that *should* trigger a warning.

bryevdv: Breaking it down another way, if the python version is of the form:
* `0.1.2`
* `0.1.2dev3`
* `0.1.2rc3` 
Then both python and JS versions should match *exactly* and warn if they don't. The only time to skip is if the python version is somethign like `0.1.2-3-ae56af` (i.e., has a dash in it). In that case the test should just be skipped. Looking at the test above, I'd expect it to fail.

bryevdv: compare: http://jsfiddle.net/8evzh1xh/

brittainhard: Yeah this is... weird. I'll try to figure it out.

bokehjs/test/document.coffee
```diff
@@ -550,11 +550,7 @@ describe "Document", ->
     expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (0.0.1)\n"
 
     parsed['version'] = "#{js_version}-foo"
-    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
-    expect(out).to.be.equal ""
-
-    parsed['version'] = "#{js_version}rc123"
-    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
```
brittainhard: So the test I added was redundant, but the reason it was failing is because it was capturing stdout instead of stderr. I fixed the problem with the previous test, which was capturing stdout. If the test failed you wouldn't see the error.

bokehjs/test/document.coffee
```diff
@@ -550,7 +550,7 @@ describe "Document", ->
     expect(out).to.be.equal "Bokeh: JS/Python version mismatch\nBokeh: Library versions: JS (#{js_version})  /  Python (0.0.1)\n"
 
     parsed['version'] = "#{js_version}-foo"
-    out = stdoutTrap -> Document.from_json_string(JSON.stringify(parsed))
+    out = stderrTrap -> Document.from_json_string(JSON.stringify(parsed))
     expect(out).to.be.equal ""
```
bryevdv: OK but let's also add a test for the other three cases that shouldn't skip, and ensure that they do print warnings when they are supposed to, and don't when they aren't.

bryevdv:
LGTM I think this should be merged then a dev build kicked off after to test everything end to end. @brendancol can you help with the dev build on Monday?

damianavila:
@brendancol Please also update [BEP-2](https://github.com/bokeh/bokeh/wiki/BEP-2:-Release-Management) with the updated info.

brendancol:
@brittainhard I tried running but got runtime errors.  Any idea where I'm going wrong here?

```bash
(bokeh-dev) ➜  scripts git:(bhard/version_mismatch_fix) ✗ bash deploy.sh -d 0.12.1rc2 -p 0.12.1rc
You have triggered the devel build process
M	examples/plotting/file/hover_glyph.py
Switched to a new branch 'devel_0.12.1rc2'
Version number changed from 0.12.0 to 0.12.1rc2 in
['bokehjs/src/coffee/version.coffee', 'bokehjs/package.json']
Version number 0.12.1rc2 added in sphinx/source/conf.py
[devel_0.12.1rc2 9088a6c] Updating version to 0.12.1rc2.
 2 files changed, 2 insertions(+), 2 deletions(-)
deploy.sh: line 78: tagarray: bad array subscript
error: Your local changes to the following files would be overwritten by checkout:
	bokehjs/src/coffee/version.coffee
Please, commit your changes or stash them before you can switch branches.
Aborting
You asked to pull from the remote 'origin', but did not specify
a branch. Because this is not the default configured remote
for your current branch, you must specify a branch on the command line.
Already up-to-date.
Everything up-to-date
error: Cannot delete the branch 'devel_0.12.1rc2' which you are currently on.
Counting objects: 8, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 703 bytes | 0 bytes/s, done.
Total 8 (delta 6), reused 0 (delta 0)
To git@github.com:bokeh/bokeh.git
 * [new tag]         0.12.1rc2 -> 0.12.1rc2
The new devel build was triggered.
(bokeh-dev) ➜  scripts git:(devel_0.12.1rc2) ✗
```

brittainhard:
@brendancol I think the issue was that `version_update.py` was changing `version.coffee`, but the script wasn't adding it to the commit, so it failed when it tried to switch to master. I have no idea why this wasn't causing release builds to fail.

I think this last commit should fix it.

brittainhard:
@damianavila I think we are trying to test out the script before we merge it into master, hence building from within the PR.

PR #4905, opened 2016-07-29
skip flaky tests for now until they can be fixed

bryevdv:


mattpap:
This needs an issue, otherwise those tests will remain skipped forever.

bryevdv:
It' didn't actually work, there are other tests than these? so I am closing the PR anyway.

PR #4907, opened 2016-07-29
Add HoverTool.show_arrow

### LINKED ISSUE #4906: New feature: hide tooltip arrow
As asked on [Stack Overflow](https://stackoverflow.com/questions/38651060/how-to-remove-arrow-on-tooltip-from-bokeh-plot/), it could be great to have the possibility to hide the tooltip arrow.

BoboTiG:
This patch adds a `show_arrow` to `HoverTool`. It is a simple boolean to disable arrow on toolip. `True` by default.

- [x] issues: fixes #4906 and [this question on Stack Overflow](https://stackoverflow.com/questions/38651060/how-to-remove-arrow-on-tooltip-from-bokeh-plot/)
- [x] release document entry

Usage (check the last line):
```python
hover = p.select_one(HoverTool)
hover.point_policy = "follow_mouse"
hover.tooltips = [
    ("Name", "@name"),
    ("Unemployment rate)", "@rate%"),
    ("(Long, Lat)", "($x, $y)"),
]
hover.show_arrow = False
```

### DIFF
--- a/bokehjs/src/less/main.less
+++ b/bokehjs/src/less/main.less
@@ -77,15 +77,27 @@
   display: block;
 }
 
-.bk-tooltip.bk-left::before {
+.bk-tooltip.bk-left.bk-tooltip-arrow::before {
   .bk-tooltip-arrow-horizontal();
   left: -@tooltipArrowWidth;
   border-right-width: @tooltipArrowWidth;
   border-right-color: @defaultTooltipBorder;
 }
 
-.bk-tooltip.bk-right::after {
+.bk-tooltip.bk-left::before {
+  left: -@tooltipArrowWidth;
+  border-right-width: @tooltipArrowWidth;
+  border-right-color: @defaultTooltipBorder;
+}
+
+.bk-tooltip.bk-right.bk-tooltip-arrow::after {
   .bk-tooltip-arrow-horizontal();
+  right: -@tooltipArrowWidth;
+  border-left-width: @tooltipArrowWidth;
+  border-left-color: @defaultTooltipBorder;
+}
+
+.bk-tooltip.bk-right::after {
   right: -@tooltipArrowWidth;
   border-left-width: @tooltipArrowWidth;
   border-left-color: @defaultTooltipBorder;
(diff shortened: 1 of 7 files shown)


bryevdv:
@BoboTiG Thanks for the PR! The tests are currently failing because the python definition of `Tooltip` and the BokehJS version don't match up exactly. You've added a `show_arrow` to the BokehJS side, but not the python side. It would be straighforward to add it on the python side here:

https://github.com/bokeh/bokeh/blob/master/bokeh/models/annotations.py#L646

We can certainly do that, but I think there might be some clunkiness resulting from the fact that don't really expose an actual `Tooltip` object on the python side much, that's worth considering. @mattpap do you have any inputs or suggestions?

BoboTiG:
@bryevdv My pleasure! I added the Python side declaration with the last commit. I refactored to have only one commit. Let me know if something should be changed.

bokehjs/src/less/main.less
```diff
@@ -77,20 +77,32 @@
   display: block;
 }
 
-.bk-tooltip.bk-left::before {
+.bk-tooltip.bk-left.bk-arrow::before {
```
mattpap: `bk-tooltip-arrow` would be a better name to avoid collisions with arrows in other contexts.

mattpap:
>  there might be some clunkiness resulting from the fact that don't really expose an actual Tooltip object on the python side much

This problem have been there for a while. There's nothing we should be fixing in this PR. In general, I'm not sure right now how to improve the API.

bokehjs/src/less/main.less
bryevdv: @mattpap good catch I agree that is more specific and an improvement. @BoboTiG can you make this last change? I think this might be ready after that.

BoboTiG:
@bryevdv No problem, it is done.

bryevdv:
LGTM, @mattpap does this look good from the BokehJS API side? If so I think it's ready to merge.

mattpap:
@bryevdv, correct my if I'm wrong (policywise): @BoboTiG, it's easier easier for reviewers if you add new commits with requested changes. This way a reviewer can focus only one new changes, instead of reviewing everything from scratch. We will squash merge those commits anyway, so you don't have to be worried about repository noise. Anyway, thanks for contributing.

BoboTiG:
@mattpap, you are totally right, I will do things right for the next time :)

PR #4908, opened 2016-07-29
Canavandl/4855 plot fill

### LINKED ISSUE #4855: No fill for background and border doesn't work
Hi,

I'm trying to embed a bokeh plot into a web application that has a gradient background defined in its CSS. I would like my bokeh plot being drawn above that gradient background, without masking it, as if I had "transparent" background and border.

So I set:
fig.background_fill_color = None
fig.border_fill_color = None

But instead of having no fill for the background and the border, I have a ugly gray color that hides my web application's background.

I tried the following trick:
fig.background_fill_color = "#000000"
fig.border_fill_color = "#000000"
fig.background_fill_alpha= 0.0
fig.border_fill_alpha = 0.0

It works, but have a bad side effect: when panning or zooming the plot, previous positions of axis values and ticks are not erase, but drawn above each other.

So, it shall be possible to easily define "no color fill" for background and border (and outline).

Cheers
Mikaël

canavandl:
- [x] issues: fixes #4855
- [x] tests added / passed

Correctly handles `background_fill_color=None` and `border_fill_color=None` on Plot. 
Expands clearRect call to properly clear frame.

example of no-fill plot (on "lightblue" html background):
screen shot 2016-07-29 at 3 47 58 pm


bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -631,12 +631,13 @@ class PlotCanvasView extends Renderer.View
...
-    ctx.fillRect(0, 0,  @canvas_view.mget('width'), @canvas_view.mget('height'))
-    ctx.clearRect(frame_box...)
-    @visuals.background_fill.set_value(ctx)
-    ctx.fillRect(frame_box...)
-
+    ctx.clearRect(0, 0,  @canvas_view.mget('width'), @canvas_view.mget('height'))
+    if @visuals.border_fill.doit
+      @visuals.border_fill.set_value(ctx)
+      ctx.fillRect(0, 0,  @canvas_view.mget('width'), @canvas_view.mget('height'))
+    if @visuals.background_fill.doit
```
mattpap: I don't know if this is desired, but with this code you can't have colored border (why do we call this a border?!) with a transparent frame. For this you would have to restore `ctx.clearRect(frame_box...)` before the second `if` statement.

canavandl: You're right. Thanks for catching that.

canavandl: I don't know why the border area is called a border when in other places we use "border" to refer to outlines (similar to like css).

bryevdv: I probably picked it up from here: http://www.mathworks.com/help/symbolic/mupad_ref/bordercolor.html

If there are concrete proposals for comprehensive, consistent naming schemes that we can adopt that are better, let's make them, and soon (of course `border_fill_color`, etc. would have to get a long deprecation)

bryevdv: Is there anywhere else besides `border_fill_color`? If not, not-repeating that is easy to do if that's the best choice, but let's make an actual decision one way or the other.

PR #4910, opened 2016-07-31
remove line in Makefile to make bad soft link

### LINKED ISSUE #4909: Sphinx Makefile creates a bad soft link
When compiling the documentation using the sphinx Makefile, a soft link is created in `sphinx/_build/html/` pointing to `../../../bokeh/server/static/` which is no longer a valid directory.  This was in the 12.0 version but has since been removed.  

I expect the Makefile should be updated to point at the correct directory or the directory should be returned to this location.

stvn66:
fixes #4909 

- [removes line in Makefile that creates the bad soft link]
- [ ] tests added / passed

### DIFF
--- a/sphinx/Makefile
+++ b/sphinx/Makefile
@@ -30,7 +30,6 @@ clean:
 
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
-	@ln -sf ../../../bokeh/server/static $(BUILDDIR)/html/static
 	@echo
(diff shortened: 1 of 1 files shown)


bryevdv:
@stvn66 can you comment on whether you had  built bokehjs locally first? The directory should exist once `setup.py` is run with `--build-js` (or `--install-js` if you build by hand with `gulp`)

stvn66:
I should be able to try this out this weekend.  I will post back.  I expect it will be as you say.

stvn66:
After building bokehjs and running the `setup.py`, the directory is created.  I had neglected doing this before trying to build the documentation.

PR #4914, opened 2016-08-01
Add support for loading external resources in output template

### LINKED ISSUE #4127: Specifying external URLs for resources

canavandl:
- [x] issues: fixes #4127
- [x] tests added / passed

Needs tests/example

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -381,7 +381,7 @@ class JSResources(BaseResources):
     def js_files(self):
         files, _ = self._resolve('js')
 
-        external_resources = self._collect_external_resources('__external_js_resources__')
+        external_resources = self._collect_external_resources('__javascript__')
         files.extend(external_resources)
 
         return files
@@ -537,7 +537,7 @@ class CSSResources(BaseResources):
     def css_files(self):
         files, _ = self._resolve('css')
 
-        external_resources = self._collect_external_resources("__external_css_resources__")
+        external_resources = self._collect_external_resources("__css__")
         files.extend(external_resources)
 
         return files
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -274,6 +274,25 @@ class BaseResources(object):
         paths = [ join(bokehjs_dir, kind, file) for file in files ]
         return paths
 
+    def _collect_external_resources(self, resource_attr):
+        """ Collect external resources set on resource_attr attribute (expected
+        to be "__external_js_resources__" or "__external_css_resources__") of models.
+        """
+
+        external_resources = []
+
+        for cls in Model.model_class_reverse_map.values():
+            external = getattr(cls, resource_attr, None)
+
+            if isinstance(external, str):
+                external_resources.append(external)
+            elif isinstance(external, list):
+                external_resources.extend(external)
+
+        # Return only unique external resources
+        return list(set(external_resources))
+
+
     def _cdn_urls(self):
         return _get_cdn_urls(self.components, self.version, self.minified)
 
@@ -361,6 +380,10 @@ class JSResources(BaseResources):
     @property
     def js_files(self):
         files, _ = self._resolve('js')
+
+        external_resources = self._collect_external_resources('__external_js_resources__')
+        files.extend(external_resources)
+
         return files
 
     @property
@@ -513,6 +536,10 @@ class CSSResources(BaseResources):
     @property
     def css_files(self):
         files, _ = self._resolve('css')
+
+        external_resources = self._collect_external_resources("__external_css_resources__")
+        files.extend(external_resources)
+
         return files
 
     @property
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -275,9 +275,7 @@ class BaseResources(object):
         return paths
 
     def _collect_external_resources(self, resource_attr):
-        """ Collect external resources set on resource_attr attribute (expected
-        to be "__external_js_resources__" or "__external_css_resources__") of models.
-        """
+        """ Collect external resources set on resource_attr attribute of all models."""
 
         external_resources = []
 
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -19,6 +19,8 @@ import json
 from os.path import basename, join, relpath
 import re
 
+from six import string_types
+
 from . import __version__
 from .core.templates import JS_RESOURCES, CSS_RESOURCES
 from .settings import settings
@@ -282,7 +284,7 @@ class BaseResources(object):
         for cls in Model.model_class_reverse_map.values():
             external = getattr(cls, resource_attr, None)
 
-            if isinstance(external, str):
+            if isinstance(external, string_types):
                 external_resources.append(external)
             elif isinstance(external, list):
                 external_resources.extend(external)
--- a/bokeh/tests/test_resources.py
+++ b/bokeh/tests/test_resources.py
@@ -5,6 +5,7 @@ import unittest
 import os
 
 import bokeh.resources as resources
+from bokeh.models import Model
 from bokeh.resources import _get_cdn_urls, websocket_url_for_server_url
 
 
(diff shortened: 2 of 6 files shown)


canavandl:
Example using external resources (loading KaTex JS and CSS):

screen shot 2016-08-01 at 2 04 57 pm

screen shot 2016-08-01 at 2 06 08 pm

bokeh/resources.py
```diff
@@ -274,6 +274,25 @@ def _file_paths(self, kind):
...
+        """ Collect external resources set on resource_attr attribute (expected
+        to be "__external_js_resources__" or "__external_css_resources__") of models.
+        """
+
+        external_resources = []
+
+        for cls in Model.model_class_reverse_map.values():
+            external = getattr(cls, resource_attr, None)
+
+            if isinstance(external, str):
```
canavandl: I imagined supporting both single and multiple external resources like:

```python
    class CustomModel(Model):
        __external_js_resources__ = "external_js_1"
        __external_css_resources__ = ["external_css_1", "external_css_2"]
```

Alternative we could require only lists, but this seems easy enough to support and one fewer gotcha for users.

mattpap: Using plural in attribute's name brings certain expectations, yet I presume we can leave this as is. Just use `string_types` from `six` instead of `str` directly.

bokeh/resources.py
```diff
@@ -274,6 +274,25 @@ def _file_paths(self, kind):
         paths = [ join(bokehjs_dir, kind, file) for file in files ]
         return paths
 
+    def _collect_external_resources(self, resource_attr):
+        """ Collect external resources set on resource_attr attribute (expected
+        to be "__external_js_resources__" or "__external_css_resources__") of models.
```
mattpap: Although this is an internal method, use appropriate syntax (double backticks) to indicate inline code for the sake of consistency.

examples/models/external_resources.py
```diff
@@ -0,0 +1,78 @@
...
+x = np.arange(0.0, 1.0 + 0.01, 0.01)
+y = np.cos(2*2*np.pi*x) + 2
+
+p = figure(title="LaTex Demonstration", width=500, height=500)
+p.line(x, y)
+
+latex = LatexLabel(text="f = \sum_{n=1}^\infty\\frac{-e^{i\pi}}{2^n}!",
+                   x=35, y=445, x_units='screen', y_units='screen',
+                   render_mode='css', text_font_size='16pt',
+                   background_fill_color='#ffffff'
```
mattpap: Either a trailing comma or parenthesis.

mattpap:
LGTM (up to my comments).

canavandl:
Thanks for the review @mattpap

canavandl:
One item for discussion: 

By loading the external resources as a script in the html doc head, those resources are in the global namespace and not limited to Bokeh. I don't know if that's going to be problematic for users.

examples/models/external_resources.py
```diff
@@ -0,0 +1,77 @@
...
+    else
+      vy = @mget('y')
+    sy = @canvas.vy_to_sy(vy)
+
+    if @model.panel?
+      panel_offset = @_get_panel_offset()
+      sx += panel_offset.x
+      sy += panel_offset.y
+
+    latex = katex.renderToString(@mget('text'), {displayMode: true})
```
canavandl: This works because katex is loaded into the global namespace.

bryevdv:
We have `__implementation__` already, I think I prefer `__javascript__` and `__css__` as the special attributes.

bryevdv:
I guess there's a question of whether we should allow inline JS/CSS to be supplied as well as URLs. Thoughts @canavandl @mattpap ?

mattpap:
> I guess there's a question of whether we should allow inline JS/CSS to be supplied as well as URLs. 

You will be able to include resources with `require()` directives in not so distant future, so I'm not sure if this duplication is necessary. Perhaps it could be easier to use for some users.

bryevdv:
@mattpap OK, so then should we merge this as-is (modulo a decision about the names of the special attrs?)

bryevdv:
OK great, so what do we want to do about docs? Just adding this example to a new "Supplying External Resources" section in "Extending Bokeh", with some minimal narrative instructions would be fine for now to me.

canavandl:
Sounds good ^

bryevdv:
@canavandl I'm getting an error when I try to see the example in the docs:

screen shot 2016-08-05 at 3 11 00 pm

bryevdv:
Wow, it doesn't work locally either unless I delete the `###` comment lines. Not sure what's going on there, maybe just change to single `#` comments

bryevdv:
As an aside this might be worth a bug search / bug report to the coffee compiler.

bryevdv:
code quality issue:
```
E         File contains trailing whitespace: sphinx/source/docs/user_guide/source_examples/extensions_example_latex.py, line 17.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: sphinx/source/docs/user_guide/source_examples/extensions_example_latex.py, line 17.'])
```
FWIW I put a git hook in the dev docs that I use all the time now that runs the code tests before finishing a commit.

canavandl:
> FWIW I put a git hook in the dev docs that I use all the time now that runs the code tests before finishing a commit.

We talked about this. I need to buckle down and do the same.

sphinx/source/docs/user_guide/source_examples/extensions_example_latex.py
```diff
@@ -13,7 +13,8 @@
 class LatexLabelView extends Label.View
   render: () ->
 
-    ### Copied directly from the ``Label.render`` implementation
+    # Start of copied section from ``Label.render`` implementation
+
```
mattpap: CoffeeScript is such a badly designed (if at all) language. Why would the author bother to appreciate historical patterns in code writing. Anyway, soon we will be able to use external files for custom models' implementations, so one will be able to use code highlighting to avoid such mistakes.

examples/models/external_resources.py
mattpap: There isn't much we can do about this. The same is with google maps. We can hide this behind `require()` (as long as katex provides a commonjs version), but then we won't be able to load it externally, but we would have to bundle it with bokeh or (in future) with custom module bundle. This is a limitation of browserify (at least without plugins).

sphinx/source/docs/user_guide/extensions.rst
```diff
@@ -10,7 +10,7 @@ not make it into the core library, either because they are too specialized, or
 for lack of resources. Fortunately, it is possible to extend Bokeh by creating
 custom user extensions.
 
-* Modify the behaviour of existing Bokeh models
+* Modify the behavior of existing Bokeh models
```
mattpap: One might disagree ;)

bokeh/tests/test_resources.py
```diff
@@ -217,3 +218,23 @@ def test_argument_checks(self):
...
+    messes up the Resources state for the other tests.
+    """
+
+    # External resources can be defined as a string or list of strings
+    class CustomModel(Model):
+        __javascript__ = "external_js_1"
+        __css__ = ["external_css_1", "external_css_2"]
+
+    r = resources.Resources()
+    print(r.js_files)
```
bryevdv: `print` should be removed. Also I think it would be ideal to test all the possible cases, single and list for each `_javascript__` and `__css__` separately.

bryevdv:
@canavandl one final small comment on the test then LGTM.

Issue #4526, opened 2016-06-15
Remove "legend" prefix in some of legend's properties


almarklein:
Specifically: `legend_margin`, `legend_padding`, `legend_spacing`. See discusion in #4511.

birdsarah:
I actually think there's a bunch of clean-up for legend properties.

Here they all are:
```python
    location = Either(Enum(LegendLocation), Tuple(Float, Float),
        default="top_right", help="""
    The location where the legend should draw itself. It's either one of
    ``bokeh.core.enums.LegendLocation``'s enumerated values, or a ``(x, y)``
    tuple indicating an absolute location absolute location in screen
    coordinates (pixels from the bottom-left corner).
    """)

    orientation = Enum(Orientation, default="vertical", help="""
    Whether the legend entries should be placed vertically or horizontally
    when they are layed out.
    """)

    border_props = Include(LineProps, help="""
    The %s for the legend border outline.
    """)

    border_line_color = Override(default="#e5e5e5")

    border_line_alpha = Override(default=0.5)

    background_props = Include(FillProps, help="""
    The %s for the legend background style.
    """)

    background_fill_color = Override(default="#ffffff")

    background_fill_alpha = Override(default=0.95)

    label_props = Include(TextProps, help="""
    The %s for the legend labels.
    """)

    label_text_baseline = Override(default='middle')

    label_text_font_size = Override(default={ 'value' : '10pt' })

    label_standoff = Int(5, help="""
    The distance (in pixels) to separate the label from its associated glyph.
    """)

    label_height = Int(20, help="""
    The minimum height (in pixels) of the area that legend labels should occupy.
    """)

    label_width = Int(20, help="""
    The minimum width (in pixels) of the area that legend labels should occupy.
    """)

    glyph_height = Int(20, help="""
    The height (in pixels) that the rendered legend glyph should occupy.
    """)

    glyph_width = Int(20, help="""
    The width (in pixels) that the rendered legend glyph should occupy.
    """)

    legend_margin = Int(10, help="""
    Amount of margin around the legend.
    """)

    legend_padding = Int(10, help="""
    Amount of padding around the contents of the legend.
    """)

    legend_spacing = Int(3, help="""
    Amount of spacing between legend entries.
    """)

    legends = List(Tuple(String, List(Instance(GlyphRenderer))), help="""
    A list of tuples that maps text labels to the legend to corresponding
    renderers that should draw sample representations for those labels.
    .. note::
[...]

PR #4916, opened 2016-08-01
Canavandl/1441 colorbar

### LINKED ISSUE #1441: Colorbar Axis

canavandl:
(ninja edit) Note: this branch is branched from #4851 and should only be merged after the parent.

Proof-of-concept color bar

- [ ] issues: fixes #1441
- [ ] tests added / passed

The code is a mess (lots of copy pasta from axis.coffee and hardcoded values), but serves to generate a POC color bar. (this uses a LogColorMapper, but Linear looks identical except for the tick spacing).

screen shot 2016-08-01 at 6 36 27 pm


canavandl:
Apart from code issues (including the floating tick above the color bar), are there any visual components that I'm missing. Specifically:

* Should a colorbar have a title somewhere? Above or to the right of the ticks?
* I could enclose the entire color bar in a rect with configurable fill props. Is that worthwhile?
* Anything else?

canavandl:
POC bounding rect of color bar (default color would probably be None and padding would be tweaked): 

screen shot 2016-08-01 at 7 05 09 pm

birdsarah:
> Should a colorbar have a title somewhere? Above or to the right of the ticks? 
Yes (optionally) - if the ticks are on the right, the default position should be on the top left - left aligned with the actual color bar (not aligned with the text)

> I could enclose the entire color bar in a rect with configurable fill props. Is that worthwhile? 
I think it's fine to leave that for later

I think this should have the same padding, margin etc properties as legend, specifically as proposed by @almarklein here: https://github.com/bokeh/bokeh/issues/4526 - so the same properties as legend will have!

canavandl:
Now the location/orientation properties are correctly implemented (though still ugly):

screen shot 2016-08-02 at 3 33 47 pm

bokeh/models/legends.py
```diff
@@ -0,0 +1,208 @@
...
+    Int, String, Enum, Instance, List, Dict, Tuple, Include, Either,
+    Float, Override,
+)
+
+from .formatters import TickFormatter, BasicTickFormatter
+from .mappers import ColorMapper
+from .renderers import GuideRenderer, GlyphRenderer
+from .tickers import Ticker, BasicTicker
+
+class Legend(GuideRenderer):
```
mattpap: This change needs an explanation (should have been done in the initial commit). There might be unforeseen consequences of this change across code base. This is also why one should refrain from using "refactoring" near such changes, as this changes both APIs and functionality.

bokeh/models/legends.py
```diff
@@ -0,0 +1,208 @@
...
+    The location where the legend should draw itself. It's either one of
+    ``bokeh.core.enums.LegendLocation``'s enumerated values, or a ``(x, y)``
+    tuple indicating an absolute location absolute location in screen
+    coordinates (pixels from the bottom-left corner).
+    """)
+
+    orientation = Enum(Orientation, default="vertical", help="""
+    Whether the legend entries should be placed vertically or horizontally
+    when they are layed out.
+    """)
```
mattpap: Is it already somehow possible to put a color bar along an axis (outside a plot view)? For me, this should be default behavior. This part of the API will need some thought how to cover both approaches to placing a color bar.

bokeh/models/legends.py
canavandl: The refactoring of `Legend` into a subclass of `GuideRenderer` is isolated in PR #4851 and this branch is merged off of the #4851 parent. I should have mentioned that in the initial comment (I'll add it now).

bokeh/models/legends.py
canavandl: As part of the factoring the `Legend` model out of bokeh.models.annotations, I'm hoping to take advantage of some of the "off-canvas" (off-frame to be more correct) `Legend` code that came out of the layouts rewrite.

I agree that putting the colorbar in the right side panel, against the frame, would be a default. I'll work on attempting that shortly.

canavandl:
superceded by #4931

PR #4873, opened 2016-07-25
Maggie m/4840 quantified code

Maggie-M:
issues: fixes #4840

PR #4921, opened 2016-08-03
Module Crawl and Version Diff

### LINKED ISSUE #4891: Crawl and List all Public Functions, Classes, Methods
Need to discuss data formats. Also need to track function signatures. 

Options to exclude files and folders -- underscore files, tests, mplexplorer.

### LINKED ISSUE #4892: Compare Public API Across Versions

brittainhard:
fixes #4891 fixes #4892 

- [x] Script to crawl folders and grab classes and functions from each file.
- [x] Script to diff between old and new versions


scripts/parse_modules.py
```diff
@@ -0,0 +1,135 @@
...
+        if isinstance(ast_node, ast.FunctionDef) and ast_node.col_offset == 0:
+            return True
+        else:
+            return False
+
+    def is_class(self, ast_node):
+        if isinstance(ast_node, ast.ClassDef):
+            return True
+        else:
+            return False
```
mattpap: Simple `return bool_expression` or `return not (bool_expression)` (depending on the order of cases) is sufficient in all three methods.

scripts/parse_modules.py
```diff
@@ -0,0 +1,135 @@
...
+        files = []
+        tree = os.walk(directory, topdown=True, followlinks=False)
+        for dirpath, dirnames, filenames in tree:
+            for folder in self.exclude:
+                if folder in dirpath:
+                    break
+            else:
+                for name in filenames:
+                    if name.endswith(".py") and not name.startswith("_"):
+                        files.append(dirpath + "/" + name)
```
mattpap: `os.path.join()`

scripts/parse_modules.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def get_classes(self, source):
+        parsed = ast.parse(source)
+        classes = [node for node in ast.walk(parsed) if self.is_class(node) and self.is_public(node.name)]
+        class_defs = {}
+        for x in classes:
+            class_defs[x.name] = {}
+            methods = []
+            for y in x.body:
+                if type(y) == ast.FunctionDef and self.is_public(y.name):
```
mattpap: Is there a reason to use `type()`? If not, `isinstance()` is preferred.

scripts/parse_modules.py
```diff
@@ -0,0 +1,135 @@
...
+
+                if function_diff:
+                    diff[x]["functions"] = list(function_diff)
+                else:
+                    diff[x]["functions"] = []
+
+    # Diff methods
+    for x in union.keys():
+        old_classes = old[x].get("classes", None) if old.get(x, None) else {}
+        new_classes = new[x].get("classes", None) if new.get(x, None) else {}
```
mattpap: `None` is unnecessary in all four cases (see PR #4873).

scripts/parse_modules.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class APICrawler(object):
+    exclude = ("tests", "static", "sampledata", "mplexplorer")
+
+    def __init__(self, directory):
+        self.directory = directory
+
+    def is_public(self, name):
+        if name.startswith("_", 0, 1) and name != "__init__":
```
mattpap: What about other special methods `__get__`, `__set__`, etc.?

brittainhard:
@mattpap can you elaborate on "What about other special methods __get__, __set__, etc.?" comment that you made?

brittainhard:
Here's a gist containing an example diff: https://gist.github.com/brittainhard/c0c682a9f10132486914f8704e316895

brittainhard:
@mattpap , @bryevdv this is ready for review. Please have a look when you have time.

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,235 @@
...
+            for y in x.body:
+                if isinstance(y, ast.FunctionDef) and self.is_public(y.name):
+                    methods.append(y.name)
+            class_defs[x.name]["methods"] = methods
+        return class_defs
+
+    def get_functions(self, source):
+        # For a given source, look for functions and return those functions as a list.
+        parsed = ast.parse(source)
+        functions = [node for node in ast.walk(parsed) if self.is_function(node) and self.is_public(node.name)]
```
canavandl: maybe:
```
[node.name for node in ast.walk(parsed) if self.is_function(node) and self.is_public(node.name)]
```
and remove the second `functions = ` line

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,235 @@
...
+        return functions
+
+    def get_filenames(self, directory):
+        # Go through the file tree and grab the filnames for each file found.
+        # Assign these to an array and return that array.
+        files = []
+        tree = os.walk(directory, topdown=True, followlinks=False)
+        for dirpath, dirnames, filenames in tree:
+            for folder in self.exclude:
+                if folder in dirpath:
```
canavandl: I'm not sure of the how robust (essentially) doing `str.isin` is going to be, in terms of avoiding filename collisions. I could imagine we add a directory named 'static_handler' or similar. We may need to make `self.exclude` contain a real list of directories to ignore instead of trying to match stuff.

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,235 @@
...
+            self._operation = self.diff_operation
+
+
+    def diff_operation(self, a, b):
+        # Use this function to find the sifference between two sets of dictionary
+        # keys. Uses the set difference operation. Called by the diff_modules
+        # function.
+        return list(a - b)
+
+    def combinaton_diff_operation(self, a, b):
```
canavandl: (sp?)

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,235 @@
...
+        # Use a combination of symmetric_difference and difference to return a list
+        # of items additions through the crawl. Allows you to get additions items instead of
+        # removed items. Would produce the same result if you switched the
+        # positions of the former and latter versions. Much more natural and
+        # easy control using the different set operators.
+        return list((a ^ b) - a)
+
+    def diff_files(self):
+        combined = copy.deepcopy(self.former)
+        combined.update(self.former)
```
canavandl: i'm not sure this line does anything

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,235 @@
...
+        diff = {}
+        intersection = {}
+        files_diff = self._operation(set(self.former), set(self.latter))
+
+        # Diff files
+        for x in combined.keys():
+            if x in files_diff:
+                diff[x] = combined[x]
+                diff[x] = {}
```
canavandl: I think this is just overriding the `x` element with an empty dict?

bokeh/util/api_crawler.py
brittainhard: Good point. I'll try to come up with some other way to parse the names.

brittainhard:
@mattpap have you had a chance to look at this?

mattpap:
@brittainhard, I'm doing it now.

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,232 @@
...
+    def __init__(self, directory):
+        self.directory = directory
+
+    def is_public(self, name):
+        # Determines whether a given function or class is public or private. Names
+        # with underscores are not allowed unless the name is "__init__"
+        return not name.startswith("_", 0, 1) or name == "__init__"
+
+    def is_function(self, ast_node):
+        # Checks whether a node is a top level function. Checks whether it is an AST
```
mattpap: > Checks whether a node is a top level function.

Then it should be called `is_toplevel_function`.

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,232 @@
...
+        # function instance. Also uses column offset, i.e. whether the function has
+        # any tabs or spaces before it.
+        return isinstance(ast_node, ast.FunctionDef) and ast_node.col_offset == 0
```
mattpap: I presume there is no better way of doing this, without actually interpreting the code? This can be easily fooled by top-level `try-catch` or `if-else`, though this doesn't happen at all (I presume).

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+    # Point the crawler at this directory.
+    def __init__(self, directory):
+        self.directory = directory
+
+    def is_public(self, name):
+        # Determines whether a given function or class is public or private. Names
+        # with underscores are not allowed unless the name is "__init__"
+        return not name.startswith("_", 0, 1) or name == "__init__"
```
mattpap: Asked about this before, but didn't follow up. `__init__` is not the only public name that starts with an underscore. Arithmetic operators, for example, are public and start with an underscore. I think this code should read `not startswith('_') or (startswith('__') and endswith('__'))` (pseudo code). Or maybe we need to list double-underscore names we want to include, because maybe names like `__getattribute__` shouldn't be considered public (I'm not sure).

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,232 @@
...
+                            diff[x]["classes"][y]["methods"] = methods_diff
+        return diff
+
+    def diff_modules(self):
+        intersection, diff = self.diff_files()
+        diff = self.diff_functions_classes(diff, intersection)
+        diff = self.diff_methods(diff, intersection)
+        return diff
+
+    def parse_diff(self, diff):
```
mattpap: This name is confusing. I think this should be called `pretty_diff`, or something along this.

bokeh/util/api_crawler.py
```diff
@@ -0,0 +1,232 @@
...
+                    elif isinstance(y, list) and y:
+                        for z in y:
+                            class_string = "%s" % z
+                            parsed_diff.append("%s.%s" % (formatted_string, class_string))
+            else:
+                parsed_diff.insert(0, formatted_string)
+        return parsed_diff
+
+    def get_diff(self):
+        self.additions = False
```
mattpap: The way `additions` and in result `_operation`are handled is cumbersome in my opinion, though as a minor thing I won't ask to rewrite this (it would require a major rewrite of the entire module).

mattpap:
This looks good as a basis for further improvements. One thing in particular I dislike is flat diff. I think the final output should be a tree (with an option to flatten it to the form that's currently used), which would say, e.g., that a class was modified and it's method removed. This will be important when the diff will contain more information, e.g. about function arguments, decorators.

bokeh/util/api_crawler.py
brittainhard: Do you have a particular opinion on how it should be done? I'm planning some refactoring when I include function signatures.

brittainhard:
@mattpap I agree that a tree diff is good idea. That exists already before I turn it into the flat diff, so it should be trivial to add some code for picking what kind of diff you want.

bokeh/util/api_crawler.py
brittainhard: This is a good idea, i'll make the change.

brittainhard:
@mattpap @bryevdv this is ready to be merged.

brittainhard:
@bryevdv can you take another look?

bryevdv:
This LGTM. Please merge when green.

PR #4931, opened 2016-08-06
Canavandl/1441 colorbar 2

### LINKED ISSUE #1441: Colorbar Axis

canavandl:
supercedes #4916 

- [x] issues: fixes #1441
- [x] tests added / passed

This version doesn't have the Legends -> GuideRenderer refactor. I need to rethink how to do that part.


bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -0,0 +1,401 @@
...
+  _get_size: () ->
+    geom = @compute_legend_bbox()
+    side = @model.panel.side
+    if side == 'above' or side == 'below'
+      return geom.height
+    if side == 'left' or side == 'right'
+      return geom.width
+
+  _set_image_data: () ->
+    [w, h] = [10, 10]
```
canavandl: Needs to be related to size of the color_mapper's palette (not hardcoded to [10,10]), so that the segmented (smooth_scale) color bar is correct.

canavandl:
Example of adding to side panels (probably need to an ``Auto`` size enum to have them fit to the panel size):

screen shot 2016-08-05 at 8 11 32 pm

Example of ``smoothed_scale=False`` (attr name to be determined) colorbar:

screen shot 2016-08-05 at 8 14 07 pm

The beige bounding box will be default transparent or white. It mostly exists for bounding box calcs.

I'll also put together a list of requirements for the colorbar, so that code review is easier.

mattpap:
> probably need to an Auto size enum to have them fit to the panel size

Definitively. That should be the default in my opinion.

mattpap:
> The beige bounding box will be default transparent or white.

The logic/defaults should be the same for axes and legends.

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -0,0 +1,401 @@
...
+    y = @model.panel._top._value
+    return {x: x, y: -y}
+
+  _get_size: () ->
+    geom = @compute_legend_bbox()
+    side = @model.panel.side
+    if side == 'above' or side == 'below'
+      return geom.height
+    if side == 'left' or side == 'right'
+      return geom.width
```
mattpap: You could use `switch` statement instead:
```
size = switch side
  when 'above', 'below'
    geom.height
  when 'left', 'right'
    geom.width
return size
```
(you can return immediately if you prefer so; given coffescript having "expression statements", I prefer to assign first and then have one return path).

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -0,0 +1,401 @@
...
+
+  type: 'ColorBar'
+
+  @mixins ['text:major_label_',
+           'text:title_',
+           'line:major_tick_',
+           'line:minor_tick_',
+           'line:border_',
+           'fill:background_',
+  ]
```
mattpap: This style is inconsistent with others. Use newline after `[` and then have one extra level of indentation per line.

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -0,0 +1,401 @@
...
+    @visuals.background_fill.set_value(ctx)
+    ctx.save()
+    ctx.fillRect(geom.sx, geom.sy, geom.width, geom.height)
+    ctx.restore()
+
+  _draw_image: (ctx) ->
+    geom = @compute_legend_bbox()
+    @visuals.border_line.set_value(ctx)
+    ctx.save()
+    ctx.setImageSmoothingEnabled(@mget('smooth_scale'))
```
mattpap: Is it safe to make plotting logic depend of image transformation? I'm not sure about this code, I just had different expectations about `smooth_scale` before I started reading.

mattpap: btw. I encourage to use `@model.attr_name` instead of `@mget("attr_name")`. At some point I would like to get rid of by-string references to attributes (e.g. to aid code linting).

jbednar:
The colorbars look good, thanks!  

I have a question about the underlying semantics, when 'smoothed_scale==True'.  Is the smoothing guaranteed to work the same as the underlying colormapping that's used for the data?  I.e., is it just doing smoothing to make it look nice, or is the color interpolation using the same mechanism that the actual mapping from data values into colors would use?  E.g. if one had a logarithmic colormap with logarithmic interpolation between the specific colors in the list (no idea whether this is supported now, but let's say it's supported someday in any case), would the colorbar use the same logarithmic interpolation that is used for data values?  

Ensuring that both data and colormap share the same process of transformation between value and color seems essential for this to be functioning as a legend, accurately reporting the data value represented by a given color even for intermediate color values.

bokehjs/src/coffee/models/annotations/color_bar.coffee
bryevdv: Oh yah, I am only now seeing this, this needs discussion. If the palettes has a lot of colors, things will look smooth, just because there are alot of colors. If not, if there's only 6 colors or whatever, the color bar should look segmented. I'm not sure offhand what that implies about the correct smoothing setting on the ctx but I am pretty sure it should not be configurable.

jbednar: I agree -- when a colorbar is shown with a numerical axis like this, it implies that each of the numerical values along the axis maps to the color shown in the bar, which is what makes the color bar be useful.  So if the actual mapping from data to color is interpolated, the color bar should always be interpolated as well -- the bar should reflect whatever is happening to the actual data.  See my comment in the main conversation for a related issue.

bryevdv:
@jbednar @canavandl the way the color mapper works is like this: The *data* is scaled to number of bins in the palettes, and then the corresponding color from whatever bin the scaled data lands on is the mapped value:
```
value = @_palette[Math.floor(d*scale+offset)]
```
In particular: the *colors* are never interpolated in the actual mapping process. It *is* possible to use interpolation to define a palette from another palette, but once the palette is defined, only those colors are used. 

Given that, I think the colorbar should just display the palette, exactly as-is.

jbednar:
>  It *is* possible to use interpolation to define a palette from another palette, but once the palette is defined, only those colors are used.

Possible in principle, or is there a Bokeh function for that?  Luke and I have been discussing this just now, and it seems like there are some gaps in the way Bokeh handles large numbers of colors (i.e., doesn't), compared to mpl and datashader, but those are probably best discussed in person.

canavandl:
per @mattpap and @bryevdv , I got rid of the canvas-level image smoothing. 

Below are the `Spectral10` and `Viridis256` palettes:

screen shot 2016-08-08 at 6 42 02 pm

The only remaining issue is there is some weird that happens when adding ColorBars to side_panels and having the dimensions set to 'Auto'. I'm not sure this issue is entirely solved for `bokeh.models.annotations.Legend` either, so I'll have to experiment a bit.

canavandl:
supercedes: #4916
issues: fixes #1441

---

### Seeking API feedback (in priority order)
* requires ColorMapper to have `high` and `low` attributes to be set.
  * A ColorMapper instance can have it's `high` and `low` attrs be unset and when
  mapping an image will use the image.data max or min. But this ColorBar implementation
  doesn't have any awareness of what data is being mapped and needs `high` and `low`
  values for the tick coordinates/labels.
  * Right now it fails poorly (javascript error in console). How should we handle this?
  Validation warn? Not plot ticks/labels if high/low are unset?
* `ColorBar.location` and side panel placement
  * The default location is `top_right` to match Legends. However, for ColorBars
  in side panels, this location is computed to be an off-canvas location and no
  visible bar is rendered. Should we add a validation warn? docstring warning?
  Note: AFAICT this issue is also present for Legends (requires explicitly setting
  `location = (0, 0)` for off-frame legends)
* `ColorBar.orientation` and side panel placement
  * I'm not entirely sure what currently happens if you put a vertical ColorBar
  in the top/bottom side bar with `legend_height='auto'`. I'm not sure what
  should happen either.
* `legend_height`, `legend_width`, and determining legend dimensions
  * Should probably be named `image_height`, `image_width` (they currently
    correspond to the color scale image not the outer legend dimensions).
  * The final ColorBar dimensions are additive (image_height + tick_size +
    tick_label_size + ...) versus being subtractive (legend_height - tick_size -
    tick_label_size - ...) because it makes the problems with "bad" attr settings
    more apparent to the user. With additive sizing, the rendered bar is only
    ever "too big" in a visually obvious way. With subtractive sizing, "too
    small" of values causes layout problems that aren't immediately obvious to
    diagnose (because things will solve to have negative values which get
    rendered into the wrong place). Legends are also additive AFAICT (and don't
    have a configurable height/width dimensions).
* `legend_height="auto"` and/or `legend_width="auto"` heuristics
  * With Sarah Bird, we developed a rough set of heuristics to handle to when
  dimensions are set to auto:

    >Heuristics to determine ColorBar image dimensions if set to "auto"
    >
    >Note: Returns the height/width values for the ColorBar's scale image, not
    >the dimensions of the entire ColorBar.
    >
    >If the short dimension (the width of a vertical bar or height of a
    >horizontal bar) is set to "auto", the resulting dimension will be set to
    >25 px.
    >
    >For a ColorBar in a side panel with the long dimension (the height of a
    >vertical bar or width of a horizontal bar) set to "auto", the
    >resulting dimension will be as long as the adjacent frame edge, so that the
    >bar "fits" to the plot.
    >
    >For a ColorBar in the plot frame with the long dimension set to "auto", the
    >resulting dimension will be the greater of:
    >  * The length of the color palette * 25px
    >  * The parallel frame dimension * 0.30
    >    (i.e the frame height for a vertical ColorBar)
    >But not greater than:
    >  * The parallel frame dimension * 0.80

* Major label text rotation
  * Currently only supports horizontal text for both ColorBar orientations. Axes
  accept a "horizontal/vertical" enum or a float (in radians). Adding the enum
  wouldn't be hard. The float would be a PITA in terms of `_tick_extent`
  calculation, but is doable.
* Title placement not configurable (only top/left of bar for both orientations)
  * This seems reasonable to me, but there are probably use cases i'm not aware of.
* `scale_props` property naming.
  * There's a `border_props` line prop for the ColorBar border. In the implementation
  there's a configurable border around the color scale image itself. I need a better
  name.

### ToDo

##### Tasks / bugs (in priority order)
- [ ] Add sphinx documentation
- [ ] The border rect for ColorBars in the right panel get clipped on the
right side by a pixel (I'm missing a Math.floor somewhere)
- [ ] Tick labels aren't factored in the legend dimensions for horizontal cbars,
so "long" labels reach over the computed end and over the border box.
  * This is the behavior for axes, but axes don't have borders to make it apparent.
  I have to think about how to handle this an not break the "snap to frame side"
  behavior.
- [ ] Probably more...

canavandl:
Here's the selenium screenshot:

screen shot 2016-08-12 at 2 46 19 pm

bryevdv:
OK tests are failing because BokehJS grew bigger. @mattpap is going to cleave off `bokeh-api.js` into a separate module, let's wait and merge that in rather than increasing the limit. (But test, review in the mean time.)

mattpap:
> is going to cleave off bokeh-api.js into a separate module, let's wait and merge that

@bryevdv, I gave this a thought and this change interferes badly with precompiler's PR and should be done on top of #4946.

bryevdv:
Ok, let's raise the limit for now.

birdsarah:
Some thoughts:

> Right now it fails poorly (javascript error in console). How should we handle this? Validation warn? Not plot ticks/labels if high/low are unset?

I would say not plot ticks/labels - I can imagine someone would want to do that for whatever reason.

> The default location is top_right to match Legends. However, for ColorBars in side panels, this location is computed to be an off-canvas location and no visible bar is rendered. Should we add a validation warn? docstring warning? Note: AFAICT this issue is also present for Legends (requires explicitly setting location = (0, 0) for off-frame legends)

Yes, this is also an issue for off_canvas legends (requires explicitly setting `location = (0, 0)`). We have not yet decided on the API for off-canvas legends. I think it's fine to not worry about it for this PR and then sort it for ColorBars and Legends.

> I'm not entirely sure what currently happens if you put a vertical ColorBar in the top/bottom side bar with legend_height='auto'. I'm not sure what should happen either.

I think it should create a tall/narrow vertical colorbar in the top/bottom - people might want to do that - basically don't do too much magic just stick to the heuristics which are based on the orientation not on the panel.

----

I would make the default title styling match the axis label default styling - smaller, italicized.

birdsarah:
Is the outline around the image configurable and the axis completely configurable as per a normal axis - I personally, for a low-n palette, would probably typically style more minimally with no ticks, just labels, and no border.

birdsarah:
Definitely going to want an example for the examples folder.

birdsarah:
@canavandl, I took the work I'd done to do data color mapping: https://github.com/bokeh/bokeh/pull/4982 and merged it with this branch. I wanted to create a plot with a mapped data source and a color bar. I've pushed the results here: https://github.com/bokeh/bokeh/tree/birdsarah/color_mapper_with_color_bar

With the following code:
```python
import numpy as np

from bokeh.io import show
from bokeh.models import ColumnDataSource, LinearColorMapper, ColorBar

x = np.random.random(2500) * 10
y = np.random.normal(size=2500) * 2 + 5
source = ColumnDataSource(dict(x=x, y=y))

lin_mapper = LinearColorMapper(palette=Viridis3, low=x.min(), high=x.max())
color_bar = ColorBar(
    color_mapper=lin_mapper,
    legend_width='auto',
    legend_height='auto'
)

p = figure()
opts = dict(x='x', line_color=None, source=source)
p.circle(y='y', fill_color={'field': 'x', 'transform': lin_mapper}, **opts)
p.add_layout(color_bar)

show(p)
```
I got this:
bokeh_plot

I had a quick poke around, but I'm not sure why this is happening.

birdsarah:
I tried again with Spectral10 because your selenium test clearly shows that working correctly.

Here's what I got:

bokeh_plot
https://github.com/bokeh/bokeh/blob/birdsarah/color_mapper_with_color_bar/examples/plotting/file/color_data_map_with_color_bar.py

I guess this means that my data_mapping code breaks something of yours or I'm not using ColorBar correctly - help greatly appreciated.

canavandl:
> I would say not plot ticks/labels - I can imagine someone would want to do that for whatever reason.

I agree. I've made the change.

> I think it should create a tall/narrow vertical colorbar in the top/bottom - people might want to do that - basically don't do too much magic just stick to the heuristics which are based on the orientation not on the panel.

I agree. From the output, it should be visually obvious what has happened: "I've put a vertical ColorBar without a specified height below my plot and it's shrunk my plot way down"

> I would make the default title styling match the axis label default styling - smaller, italicized.

Yup - done.

>Is the outline around the image configurable and the axis completely configurable as per a normal axis - I personally, for a low-n palette, would probably typically style more minimally with no ticks, just labels, and no border.

Yup. However, if you could propose a better name for the border around the image other than "scale_*", that'd be nice.

screen shot 2016-08-18 at 1 17 23 pm

birdsarah:
soooo pretty.

`scale_*` -> `bar_*` ?

Naming things is not my forte.

birdsarah:
I think I feel more strongly about better defaults. What we've got now is a little matplotlibby - love them to bits but not the prettiest defaults.

Looking at http://docs.ggplot2.org/current/guide_colourbar.html and https://stanford.edu/~mwaskom/software/seaborn/examples/network_correlations.html

I would say default of:
* no border
* no minor ticks
* major ticks should be white so the "eat" into the color bar

birdsarah:
On the example you showed - how did you get the label in the middle of an colorblock - instead of at a join?

canavandl:
I used a `FixedTicker` with the `ticks` attr set to locations that'd be in the center of colorblocks.

i.e. with a low/high range of 0-50 and 5 colorblocks, it'd be at `range(5,50,10)`.

canavandl:
> I would say default of:
no border
no minor ticks
major ticks should be white so the "eat" into the color bar

How's:

screen shot 2016-08-18 at 4 33 13 pm

@birdsarah

canavandl:
I also renamed "scale_" to "bar_", which I think is not bad.

birdsarah:
I think the defaults are much better. But still not quite right, I'll play with some variants to see if I can come up with something better - but I think this is a lot cleaner.

birdsarah:
I'm merging this, but two quick thoughts:
* i think Title should get a standoff property so you can move it up/down
* seems like there's code here that's duplicated from other places around bokeh - i'm guessing it wasn't accessible easily - but maybe we can factor it out later

SimonBiggs:
This is amazing work. Thank you so much.

PR #4939, opened 2016-08-09
update dev setup docs

### LINKED ISSUE #4938: "Getting Set Up" section of documentation does not mention the base dependencies of Bokeh
When setting up a python environment for developing Bokeh, in addition to the dependencies listed in the [Getting Set Up page](http://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html#devguide-setup), one must also install the base dependencies for Bokeh.  @birdsarah mentioned a simple method for getting these dependencies by installing then removing Bokeh: `conda install bokeh && conda remove bokeh`.  Either something like this should be added to the documentation, or the explicit dependencies should be listed in addition to the development specific ones.

stvn66:
This change adds directions for installing the base dependencies for Bokeh in addition to the dev specific ones.  Additionally, it changes
```
``--build_js``
``--install_js``
 ```
to 
```
``--build-js``
``--install-js``
```

- [x] issues: fixes #4938 
- [x] tests added / passed

### CODE REVIEW

File: sphinx/source/docs/dev_guide/setup.rst [hunk 1]
```diff
@@ -249,12 +249,32 @@ which include:
...
+In addition to the build and test dependencies, you must also have the base
+dependencies for Bokeh installed. A simple way to install these dependencies
+is to install Bokeh via ``conda install`` or ``pip install`` before running
+``setup.py``.  Alternatively, you can download them indivually. The
+dependencies include:
+
+* backports_abc
+* jinja2
+* markupsafe
+* mkl
```
bryevdv: like the other PR, `mkl` is absolutely 100% not a required dependency. Conda will install it by default for any numpy install, but that has nothing to do with Bokeh.

stvn66:
Updated the dependencies list to match the changes made to `devdeps.py`.  The updated version should only have the essential bokeh dependencies, not the dependencies of dependencies.

bryevdv:
@stvn66 looks great will merge when Travis is green

Issue #4886, opened 2016-07-26
Allow User Defined Models to inherit from User Defined Models


brendancol:
To support external bokeh plugins, we need to be able to extend User Defined Models.  @mattpap any thoughts on this?

mattpap:
@brendancol, in PR #4946, the `examples/models/custom.py` shows how to inherit from custom models.  This assumes all custom models ending up in one bundle. I'm not sure if this will be a problem or not. If so, this can be lifted, but for now I didn't want to solve possibly non-existent problems. Let me know if this matches your expectations for this issue. If not, feel free to elaborate on further expectations for this feature.

mattpap:
I'm closing this, because custom models are now extensible. @brendancol, if there is anything missing in the current implementation, please reopen this with specific issues.

PR #4946, opened 2016-08-11
Pre-compile custom models and code fragments

### LINKED ISSUE #4879: Remove gear glyph from bokehjs to shrink resource size
Since gear and gauge glyphs are rarely used, it'd be better to remove them from bokehjs and make them` __implementation__` glyphs in python.

mattpap:
This removes run-time `bokeh-compiler.js` bundle. Custom models are now precompiled, dependencies extracted (to reasonable extent) and verified. JavaScript is parsed and syntax analyzed as well. There is support for importing Less/CSS stylesheets. node.js is a required dependency for custom models, where CoffeeScript is the default and dependency parsing necessary. In CustomJS/FuncTickFormatter JavaScript is still the default and dependencies are not resolved by default. This way one can create CustomJS without node.js. See `custom.py` and `transform_jitter*.py` under `models/`. This all needs some further polishing and perhaps additional changes to package build files, so that `bokehjs/build/js/{compile.js,modules.json}` are included. I squashed all changes to a single commit (original history was garbage).

fixes #4879, #4886


examples/models/customjs.py
```diff
@@ -27,14 +27,14 @@
 plot.add_layout(LinearAxis(), 'below')
 plot.add_layout(LinearAxis(), 'left')
 
-customjs = CustomJS(args=dict(source=source), lang="coffeescript", code="""
+customjs = CustomJS(args=dict(source=source), code=CoffeeScript("""
```
bryevdv: This is a breaking change. People are definitely making custom extensions at this point, and @birdsarah and I have both been pushing the idea pretty solidly.  I think this needs a deprecation warning. Sholud be pretty simple, add an  `__init__` to `CustomJS` that strips the `lang` kwarg and converts/wraps `code` accordingly?

sphinx/source/docs/user_guide/extensions.rst
```diff
@@ -162,7 +162,7 @@ then the complete Python class might look like:
 
     class Custom(LayoutDOM):
 
-        __implementation__ = open("custom.coffee").read()
+        __implementation__ = "custom.coffee"
```
bryevdv: also a breaking change, is there a way to have a deprecation cycle?

bokeh/core/enums.py
```diff
@@ -158,9 +158,6 @@ def enumeration(*values, **kwargs):
 #: Specify an aggregation type for different charts
 Aggregation = enumeration("sum", "mean", "count", "nunique", "median", "min", "max")
 
-#: Specify the language used in a CustomJS callback
-ScriptingLanguage = enumeration("javascript", "coffeescript")
-
```
bryevdv: prefer to leave in as a deprecated module attr for a short while

bokeh/models/callbacks.py
```diff
@@ -56,17 +64,10 @@ def from_py_func(cls, func):
     """)
 
     code = String(default="", help="""
-    A snippet of JavaScript/CoffeeScript code to execute in the browser. The
+    A snippet of JavaScript code to execute in the browser. The
```
bryevdv: Would be appropriate to mention the JS string can be made with `CoffeeScript` class from etc.

bokeh/util/compiling.py
```diff
@@ -0,0 +1,274 @@
+from __future__ import absolute_import
```
bryevdv: I have a possibly odd but definitely strong aversion to having module names be gerunds (i.e "xyz-*ing*") What about `util.compile` or `util.compiler` or `util.jsbuild` ?

bokeh/util/compiling.py
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    for (var i = 0; i < entry.length; i++) {
+        Bokeh.Models.register_locations(Bokeh.require(entry[i]));
+    }
+  } else {
+    throw new Error("Cannot find Bokeh. You have to load it prior to loading plugins.");
+  }
+})
+"""
```
bryevdv: Is there a way to coordinate this automatically? Maybe `setup.py` could copy to `server/static/js` and this could read from there?

bryevdv:
@mattpap missing new npm dep?
```
    Error: Cannot find module 'acorn' from '/home/travis/build/bokeh/bokeh/bokehjs/src/js'
      at /home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:46:17
      at process (/home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:173:43)
      at ondir (/home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:188:17)
      at load (/home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:69:43)
      at onex (/home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:92:31)
      at /home/travis/build/bokeh/bokeh/bokehjs/node_modules/resolve/lib/async.js:22:47
      at FSReqWrap.oncomplete (fs.js:82:15)
 ```

File: bokeh/util/compiling.py (hunk 5 above)
Comment (mattpap) [inline bokeh/util/compiling.py]: `compile` conflicts with Python's `compile()`, so perhaps `compiler`. `jsbuild` is wrong as well, because you can do more than compiling JavaScript in this module.

File: sphinx/source/docs/user_guide/extensions.rst (hunk 2 above)
Comment (mattpap) [inline sphinx/source/docs/user_guide/extensions.rst]: `__implementation__` accepts a string or an instance implementing `Implementation` interface. Accepting string is a convenience measure and, by default, nothing has changed here. A string is considered CoffeeScript source code. However, if a string is a one-liner which ends in `.coffee`, `.js`, `.css` or `.less`, then it will be interpreted as a path (relative to file where a model is defined). This new convenience feature shouldn't conflict with any earlier code.

File: bokeh/core/enums.py (hunk 3 above)
Comment (mattpap) [inline bokeh/core/enums.py]: My policy is to get rid of old code in the first place, to make sure we use new APIs everywhere and then deprecate. I will do it later.

File: bokeh/util/compiling.py (hunk 6 above)
Comment (mattpap) [inline bokeh/util/compiling.py]: Of course possible in at least a few ways, but for now wasn't significant enough to do. Also, the code was slowly converting from a different implementation to what plugins use. This outcome wasn't that obvious during early work.

**Comment by mattpap** [comment]
> missing new npm dep?

The same with `less`. npm is fun.

**Comment by mattpap** [comment]
@bryevdv, to get things moving, I need conda package for node.js > 5.0 (this includes npm > 3.0).

**Comment by bryevdv** [comment]
@mattpap are you asking me to build such a package? I'm not sure offhand, do we know who built the existing one?

**Comment by mattpap** [comment]
> are you asking me to build such a package? I'm not sure offhand, do we know who built the existing one?

I was asking for feedback, because I don't know my self how this was done before and who did it. Maybe @damianavila knows?

**Comment by bryevdv** [comment]
@mattpap we could try starting with the conda-forge recipe:

https://github.com/conda-forge/nodejs-feedstock/blob/master/recipe/meta.yaml

and then if that works, uploading the result to the bokeh channel on anaconda.org

**Comment by bryevdv** [comment]
@mattpap the following modified `meta.yaml` (with the existing `build.sh` from the link above) seems to be working on OSX. If you build for linux we can upload to the bokeh channel:
```
{% set version = "5.0.0" %}

package:
  name: nodejs
  version: {{ version }}

source:
  fn: node-v{{ version }}.tar.gz
  url: https://nodejs.org/dist/v{{ version }}/node-v{{ version }}.tar.gz
  sha256: 698d9662067ae6a20a2586e5c09659735fc0050769a0d8f76f979189ceaccdf4

build:
  number: 0
  #skip: True  # [not py27]

requirements:
  build:
    # node's build scripts find and use MSVC 2015, regardless of conda's setting.
    - vs2015_runtime  # [win]
    - python          2.7.*
    - gcc             # [linux]
  run:
    - vs2015_runtime   # [win]
    - libgcc           # [linux]

test:
  commands:
    - node -h
    - node -v
    - npm version
    - npm install -h
    - npm config get prefix -g
    - test $(echo "console.log(1 + 3)" | node) == 4  # [not win]

about:
  home: https://nodejs.org/
  license: MIT
  summary: a platform for easily building fast, scalable network applications

extra:
  recipe-maintainers:
    - msarahan
    - pelson
```

**Comment by bryevdv** [comment]
OSX version is up on anaconda.org with these commands:
```
$ anaconda upload /Users/bryan/anaconda/conda-bld/osx-64/nodejs-5.0.0-0.tar.bz2 -u bokeh
Using Anaconda Cloud api site https://api.anaconda.org
detecting package type ...
conda
extracting package attributes for upload ...
done

Uploading file bokeh/nodejs/5.0.0/osx-64/nodejs-5.0.0-0.tar.bz2 ... 
 uploaded 6940 of 6940Kb: 100.00% ETA: 0.0 minutes


Upload(s) Complete

Package located at:
https://anaconda.org/bokeh/nodejs

bryan@0199-bryanv  /tmp $ conda install -c bokeh nodejs
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata ...........
Solving package specifications: ..........

Package plan for installation in environment /Users/bryan/anaconda:

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    nodejs-5.0.0               |                0         6.8 MB  bokeh

The following packages will be UPDATED:

    nodejs: 4.1.0-1 javascript --> 5.0.0-0 bokeh

Proceed ([y]/n)? 
```

**Comment by bryevdv** [comment]
Although I get an error when I `npm install`
```
$ npm install
Error: Cannot find module './lib'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/bryan/anaconda/lib/node_modules/npm/node_modules/request/node_modules/hawk/index.js:1:80)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/bryan/anaconda/bin/node" "/Users/bryan/anaconda/bin/npm" "install"
npm ERR! node v5.0.0
npm ERR! npm  v3.3.6
npm ERR! code MODULE_NOT_FOUND

npm ERR! Cannot find module './lib'
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
```
@msarahan is this somethign you can help with?

**Comment by bryevdv** [comment]
it actually looks like there are binaries available... wouldn't it be easier to just package those, than build them?

**Comment by msarahan** [comment]
Sorry, it looks like your nodejs package built OK.  I don't know what that "Cannot find module" message might come from.  I have never used nodejs - only compiled it.

**Comment by bryevdv** [comment]
let me try 6.x, and then failing that, 4.4.7

**Comment by bryevdv** [comment]
OK I had some "globally" installed things from previously and I think this was the problem. I installed `6.3.1` and it seems to be working

**Comment by bryevdv** [comment]
@mattpap ou should be able to use the recipe above, otherwise I can build `6.3.1` on a linux vm tomorrow

File: bokeh/util/compiler.py [hunk 7]
```diff
@@ -0,0 +1,282 @@
...
+    def __str__(self):
+        return self.text
+
+bokehjs_dir = settings.bokehjsdir()
+
+def nodejs_compile(code, lang="javascript", file=None):
+    compilejs_script = join(bokehjs_dir, "js", "compile.js")
+
+    try:
+        proc = Popen(["nodejs", compilejs_script], stdout=PIPE, stderr=PIPE, stdin=PIPE)
```
canavandl: use "node" instead of "nodejs" for conda-built nodejs compat

bokeh/util/compiler.py
```diff
@@ -0,0 +1,282 @@
...
+    except OSError:
+        raise RuntimeError('node.js is needed to allow compilation of custom models ' +
+                           '("conda install -c bokeh nodejs" or follow https://nodejs.org/en/download/)')
+
+    (stdout, errout) = proc.communicate(input=json.dumps(dict(code=code, lang=lang, file=file)))
```
canavandl: missing `encode` call to convert JSON str to bytes array

bokeh/util/compiler.py
mattpap: @canavandl, at least on Ubuntu, the default is `nodejs` and `node` is a symlink. I think conda package for node.js should create both. Though, I think I will add detection code to bokeh anyway.

bokeh/util/compiler.py
mattpap: Fixed.

bokeh/util/compiler.py
mattpap: Added auto-detection and you can pass a path to a local install with `BOKEH_NODEJS_PATH=foo python some.py`.

bokeh/util/compiler.py
```diff
@@ -85,22 +85,41 @@ def __str__(self):
 
 bokehjs_dir = settings.bokehjsdir()
 
-def nodejs_compile(code, lang="javascript", file=None):
-    compilejs_script = join(bokehjs_dir, "js", "compile.js")
+def _detect_nodejs():
+    if settings.nodejs_path() is not None:
+        nodejs_paths = [settings.nodejs_path()]
+    else:
+        nodejs_paths = ["xnodejs", "node"]
```
canavandl: xnodejs?

mattpap: Yep, it's a NASA version developed for environments with higher levels of background radiation, aka. I'm committing garbage.

canavandl: lol

damianavila:
>Maybe @damianavila knows?

Nop, IDK, but it seems @bryevdv gives you a recipe to build it :+1:

mattpap:
@bryevdv, after all, doing API split off in an unrelated PR and merging changes here, wasn't that big deal.

bokeh/models/callbacks.py
mattpap: This was addressed, just not visible in the short diff.

mattpap:
I removed `Gear` model (and all accompanying code) from bokeh and made a big custom model example out of it. New location is `examples/custom/gears`. This is a nice example, because it spans multiple files in a non-trivial way, integrates coffeescript with javascript and perhaps more. Originally @bryevdv wanted `Gear` to be a custom model, but I think that `Gear` shouldn't have been a part of bokeh in the first place (not this was introduced by me). I doubt this will affect anyone at all and one can still just copy the code from the example.

tests/examples/examples.yaml
```diff
@@ -1,5 +1,6 @@
 - path: "models"
   skip: ["choropleth.py", "maps_cities.py", "data_tables_server.py"]
+- path: "custom/gears"
```
mattpap: There must be added support for the following pattern: `examples/dir/example/__init__.py -> example.html`.

mattpap:
> Sphinx error:
> Unable to generate Bokeh plot at
> /home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/extensions.rst:194:<module '__builtin__' > (built-in)> is a built-in class

Does anyone know what this means? This code works when run manually.

bryevdv:
@mattpap I ran with `SPHINXOPTS=-vv` and got
```
  File "/Users/bryan/work/bokeh/bokeh/sphinxext/bokeh_plot.py", line 275, in html_visit_bokeh_plot
    js, script = autoload_static(plot, resources, filename)
  File "/Users/bryan/work/bokeh/bokeh/embed.py", line 335, in autoload_static
    js_raw = resources.js_raw + [script],
  File "/Users/bryan/work/bokeh/bokeh/resources.py", line 388, in js_raw
    custom_models = gen_custom_models_static()
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 241, in gen_custom_models_static
    impl = model.implementation
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 209, in implementation
    impl = impl.__class__(impl.code, self.file + ":" + self.name)
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 192, in file
    return abspath(inspect.getfile(self.cls))
  File "/Users/bryan/anaconda/lib/python3.4/inspect.py", line 524, in getfile
    raise TypeError('{!r} is a built-in class'.format(object))
TypeError: <module 'builtins' (built-in)> is a built-in class
```

mattpap:
@bryevdv, can we make showing tracebacks the default in travis ci?

bryevdv:
@mattpap sure, we already catch and print exceptions:

https://github.com/bokeh/bokeh/blob/master/bokeh/sphinxext/bokeh_plot.py#L298-L299

I'm not sure why adding `-vv` makes spinx itself somehow output more verbose stack traces. But `-vv` is incredibly voluminous, I'd rather just make that `except` block spit out the stack trace itself, personally.

bryevdv:
@mattpap actually `SPHINXOPTS=-v` also prints the stack trace, and is alot less verbose. If you just want to make that change that's fine with me.

mattpap:
That works for me.

mattpap:
> `TypeError: <module 'builtins' (built-in)> is a built-in class`

The problem here is that sphinx extension uses `compile()` and `eval()` pair to execute Python code, but doesn't configure the environment sufficiently enough for `bokeh.util.compiler` to work properly. bokeh needs to know the path to a custom model to allow for JS module resolution, but that's not possible to obtain, because the default module is `__builtin__` which doesn't configure `__file__` (obviously), which is in turn needed by `inspect.getfile()`.

If a path is known, then the problem can be fixed in the sphinx extensions by adding:
```
import sys, types
sys.modules["fake"] = types.ModuleType("fake")
sys.modules["fake"].__file__ = "path from bokeh-plot directive"
namespace = {"__name__": "fake"}
```
Otherwise we can configure `__file__` to whatever current directory is. Alternatively, we may want to fix the problem in `bokeh.util.compiler`, either by not requiring known custom module location (needs major redesign of the code) or making it work harder to figure out location (use current directory, etc.).

mattpap:
> libgfortran.so.1: cannot open shared object file: No such file or directory

Apparently conda install version 3 and ggplot/seaborn require version 1.

bokeh/sphinxext/bokeh_plot.py
```diff
@@ -235,13 +237,17 @@ def _show(obj=None):
...
 
 
-def _render_plot(source, symbol):
+def _render_plot(source, path=None, symbol=None):
     io._state._document = Document()
     namespace = {}
+    if path is not None:
+        sys.modules["fake"] = types.ModuleType("fake")
+        sys.modules["fake"].__file__ = path
+        namespace["__name__"] = "fake"
```
mattpap: @bryevdv, please review this and the remainder of changes in this file.

bryevdv: Roughly seems OK, I'd suggest building the docs and spot checking plots in various places (and all the extension ones). 

As an aside I think almost all of this complicated code can be ripped out and replaced with a subprocess call to `bokeh json` and some trivial templating.

PR #4953, opened 2016-08-12
Store scss on renderer not on glyph

### LINKED ISSUE #4949: Specifying a selection doesn't work with patches when hit testing tools are present.
Normally you can manually specify an initial selection on a data source by specifying `source.selected = {'0d': ......}` in the correct format.

However if the plot has patches and tools with hit testing are used, this doesn't work due to an initialiazation conflict with the hit testing.

photo 2016-08-10 02_26

From @bryevdv's investigation:

sxss is not defined on the main glyph view, even though render was called because when there is a selection present, the glyph renderer only calls render on the selection and non-selection glyphs, But hit-testing checks the main renderer,

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee#L178-L200

glyph.render is not called when there is a selection present, only selection_glyph.render and nonselection_glyph.render

patches is the only set of shadow glyphs (selection and nonselection glyphs) that have this dependence on something happening in render.  Something that might work is to store sxss on the glyph renderer, not the glyph; or maybe it can be computed in map_data.

birdsarah:
- [x] issues: fixes #4949
- [x] tests added / passed

This allows the sxss to be available for hit testing - which is not done on the main glyph (but on the selection and nonselection glyph)

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/patches.coffee
+++ b/bokehjs/src/coffee/models/glyphs/patches.coffee
@@ -77,8 +77,8 @@ class PatchesView extends Glyph.View
   _render: (ctx, indices, {sxs, sys}) ->
     # @sxss and @syss are used by _hit_point and sxc, syc
     # This is the earliest we can build them, and only build them once
-    @sxss = @_build_discontinuous_object(sxs)
-    @syss = @_build_discontinuous_object(sys)
+    @renderer.sxss = @_build_discontinuous_object(sxs)
+    @renderer.syss = @_build_discontinuous_object(sys)
     for i in indices
       [sx, sy] = [sxs[i], sys[i]]
 
@@ -133,8 +133,8 @@ class PatchesView extends Glyph.View
     hits = []
     for i in [0...candidates.length]
       idx = candidates[i]
-      sxs = @sxss[idx]
-      sys = @syss[idx]
+      sxs = @renderer.sxss[idx]
+      sys = @renderer.syss[idx]
       for j in [0...sxs.length]
         if hittest.point_in_poly(sx, sy, sxs[j], sys[j])
           hits.push(idx)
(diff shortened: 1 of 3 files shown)


bokehjs/src/coffee/models/glyphs/patches.coffee
```diff
@@ -77,8 +77,8 @@ class PatchesView extends Glyph.View
   _render: (ctx, indices, {sxs, sys}) ->
     # @sxss and @syss are used by _hit_point and sxc, syc
     # This is the earliest we can build them, and only build them once
```
bryevdv: Won't this get built every render call, not just once? I'm also a little worried that since render is called with different sxs and sys for selection and non-selection glyphs, that one or the other "wins" by overwriting the other, and that `@renderer.sxss` thus only has the subset of patches for whichever render was called last

bryevdv: I still wonder if this could not be computed in `_map_data`

birdsarah: That comment is old. I agree with you it doesn't make sense. I do remember moving those lines around a bunch when I made this PR. This was also the PR where I made all the patches disappear if I recall correctly, so there may have been a hasty hotfix.

birdsarah: > Won't this get built every render call, not just once? 

https://github.com/bokeh/bokeh/commit/064004bdd64e339f939b055b29f6b35651c2ef70

Here's the commit that moved those lines and made that comment. The commit message is "Only build sxss and syss once - Instead of on every hit!"

So the comment at least makes sense now.
----
>  Maybe we could move to `_map_data`.

Maybe. I definitely haven't got my head deep into this.

> I'm also a little worried that since render is called with different sxs and sys for selection and non-selection glyphs

At the end of the day we only need to be building these sxss & syss once for the main piece of data as the point of it is so that hit testing can work for patches.

I haven't yet understood how selection/nonselection glyphs work and render.

So your concern may be valid but I don't know. I do think this is working at the moment though for what it's worth.

bryevdv:
> Here's the commit that moved those lines and made that comment. The commit message is "Only build sxss and syss once - Instead of on every hit!"

Ah, yes, once per render --- better than "number of hits" per render

> So your concern may be valid but I don't know. I do think this is working at the moment though for what it's worth.

A test would be this: hover over a selected patch first, then load again but hover over an nonselected patch first

Issue #4951, opened 2016-08-12
ModelExtensions - all resources from a whole module are loaded


birdsarah:
I setup a series of ModelExtensions with a base with the implementation and a bunch of subclasses for different skins.

``` python
class IonRangeSliderFlat(IonRangeSliderBase):
    __css__ = [
        "/css/ion.rangeSlider.css",
        "/css/ion.rangeSlider.skinFlat.css",
    ]


class IonRangeSliderHTML5(IonRangeSliderBase):
    __css__ = [
        "/css/ion.rangeSlider.css",
        "/css/ion.rangeSlider.skinHTML5.css",
    ]


class IonRangeSliderModern(IonRangeSliderBase):
    __css__ = [
        "/css/ion.rangeSlider.css",
        "/css/ion.rangeSlider.skinModern.css",
    ]
```

When I used one of them:

``` python
from ion_slider import IonRangeSliderFlat
s = IonRangeSliderFlat(..)
...
```

All the CSS from all the different subclasses were stuck into my html.

birdsarah:
As a work around, putting the sub-classes in different files/modules prevented this.

bryevdv:
@canavandl yah on further thought, the idea would be to only add resources for models that are actually in the document.

bryevdv:
@birdsarah can you expand the original issue to a complete MRE?

bryevdv:
noting this was solved fo the docs by adding a private API that the sphinx code can use to clear extensions in https://github.com/bokeh/bokeh/pull/7942 but need a general way to avoid this.

mattpap:
This was fixed in PR #8749.

bryevdv:
@mattpap do you have any objection to just making `_clear_extensions` public? There has not been much (any) complaint about this so I think it really only comes up in specialized situations like our docs. I think we could just make the method public for anyone who really does need it, and call it a day.

bryevdv:
I intend to proceed as described above.

PR #4954, opened 2016-08-12
Keep ordering of resources

### LINKED ISSUE #4950: Non-deterministic ordering of CSS resources for external resource loading.
I was just using the new external resource loading feature of bokeh. I have a series of three css files. With css the ordering is important. It seems that the order that they're actually put into the document changes on each run.

ping @canavandl - is this likely or do you need more info?

birdsarah:
- [x] issues: fixes #4950
- [x] tests added / passed

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -289,8 +289,7 @@ class BaseResources(object):
             elif isinstance(external, list):
                 external_resources.extend(external)
 
-        # Return only unique external resources
-        return list(external_resources)
+        return external_resources
 
 
     def _cdn_urls(self):
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -285,9 +285,12 @@ class BaseResources(object):
             external = getattr(cls, resource_attr, None)
 
             if isinstance(external, string_types):
-                external_resources.append(external)
+                if external not in external_resources:
+                    external_resources.append(external)
             elif isinstance(external, list):
-                external_resources.extend(external)
+                for e in external:
+                    if e not in external_resources:
+                        external_resources.append(e)
 
         return external_resources
 
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -290,7 +290,7 @@ class BaseResources(object):
                 external_resources.extend(external)
 
         # Return only unique external resources
-        return list(set(external_resources))
+        return list(external_resources)
 
 
     def _cdn_urls(self):
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -281,7 +281,7 @@ class BaseResources(object):
 
         external_resources = []
 
-        for cls in Model.model_class_reverse_map.values():
+        for _, cls in sorted(Model.model_class_reverse_map.items(), key=lambda arg: arg[0]):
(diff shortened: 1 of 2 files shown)


bokeh/resources.py
```diff
@@ -290,7 +290,7 @@ def _collect_external_resources(self, resource_attr):
                 external_resources.extend(external)
 
         # Return only unique external resources
-        return list(set(external_resources))
+        return list(external_resources)
```
mattpap: You actually want to deduplicate resources (alphabetically or by model alphabetically). The result should be sorted in some way. You can gain some insight by looking into custom models code below.

birdsarah: I know we want the de-duplication, but its more important that things actually load for now.

bryevdv:
A couple of comments:
* How hard would it be to only output resources for models that are actually in the doc being serialized, and not simply for every Model that gets imported somewhere?
* What if two models provide the same set of resources, in different and incompatible orderings?

birdsarah:
> What if two models provide the same set of resources, in different and incompatible orderings?

Yes, I thought about that, which is why I was originally going to propose no de-duplication until we've thought it through. But at the end of the day, I don't think that's a use case that we can/should try to support.

All resources are going to end up grouped, probably at the top of a page.

Lets say you have one model with the css files: "normalize" and "style" and you need "normalize" to go before "style". 

Now lets say another model comes along and wants "style" before "normalize". In this case there's no way of supporting both models:
1) If you don't de-duplicate you now have "normalize", "style", "style", "normalize" - which will mean the css for the first model is broken
2) If you do de-duplicate you now have "normalize", "style" - which will mean the css for the second model is broken.

But lets take another example which makes the case for deduplication. You have two custom models, both use `"normalize"`*. The first model is `["normalize", "skin_1"]`, the second model is `["normalize", "skin_2"]`. In this case if we don't de-duplicate we will have `["normalize", "skin_1", "normalize", "skin_2"]` - this may break skin_1 as we will have re-run normalize after applying skin_1. But if we do de-duplicate and have `["normalize", "skin_1", "skin_2"]`.

We will need feedback from real users on how to make this work really well, but in the meantime I think the most important thing is to provide deterministic behavior so that users can easily debug this.

We will:
- load models in a deterministic order (still to do)
- keep resources in the order specified by their entries 
- de-duplicate resources, keeping the first entry of any resource
- we should document this very heavily so users can debug

----
* as an aside custom models with normalize in could be an issue for us as the may override bokeh styling - I think we're good because we're namespaced but maybe something to keep an eye on.

birdsarah:
> How hard would it be to only output resources for models that are actually in the doc being serialized, and not simply for every Model that gets imported somewhere?

I've opened a separate issue for this already - https://github.com/bokeh/bokeh/issues/4951 - because it's pretty annoying. I think we can tackle it there.

canavandl:
The code LGTM. Can you add a mention of resource ordering (both inter and intra-Models) to https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/extensions.rst#supplying-external-resources?

canavandl:
LGTM. I think the deterministic model ordering will be good, even if alphabetical isn't the

PR #4961, opened 2016-08-15
Don't end up with white screen under an unhandled exception

mattpap:
This is a first step to give the user some feedback when bokeh fails (and that happens a lot even if you know how to put pieces together). This only takes care of unhandled exceptions, showing a box with exception's message. Note `<hr>` tag is there intentionally, so that this error stands out from the remainder of a page. The look and feel is modeled after jsmath, so we may want to change that or perhaps attribute the fact.

This is just a tip of the ice berg, because there are plenty of ways bokeh can fail. A nice example is "could not set initial ranges" warning, which can be either a user's mistake resulting in a defunct plot or expected behavior when the plot starts with no data and is updated incrementally.

To test this, use:
```

from bokeh.models import Plot

plot = Plot()

doc = Document()
doc.add_root(plot)
doc.validate()

if __name__ == "__main__":

    filename = "some.html"
        f.write(file_html(doc, INLINE, "some"))
    print("Wrote %s" % filename)
    view(filename)
```


bryevdv:
Only a couple of comments: 

* I think this has to be configurable. I'm fine with default "on" but let's solicit discussion
* I'd tweak the visual saying some too, maybe more like the docs version warning banner:

screen shot 2016-08-15 at 5 25 59 pm


ping @bokeh/dev

birdsarah:
This is brilliant. I tested it on a recent white screen issue and it worked great.

Agree with bryan we can do something a little more modern with the styling - if you'd like me to push something I'd be happy to.

Of course, now I want more. Is is possible to have this pop-up for run-time errors too?

birdsarah:
Also, in my particular error, it rendered as "â€” this.$el.find(...).ionRangeSlider is not a function"

The console errors was: "TypeError: this.$el.find(...).ionRangeSlider is not a function[Learn More]"

This was on firefox in case it matters.

mattpap:
> it rendered as "â€”

I presume this is an issue with unicode character I used for mdash entity. It renders properly for me under FF (42 at least). I will try to use `&mdash;` instead, but it's a pain with DOM APIs.

mattpap:
> I think this has to be configurable. I'm fine with default "on" but let's solicit discussion

I was considering making this work only in non-dev setup of bokeh, but then I realized that (at least in Chrome) rethrown exception's stack trace is preserved, so this doesn't hinder development experience. This feature has to be "on" in "retail" version of bokeh, because otherwise it won't be any help for the average user.

mattpap:
@bryevdv, @birdsarah, as to styling, feel free to make changes. I just used whatever (so jsmath's) styling to make the error stand out on a page. Just keep it plain as the comment in `safely.coffee` indicates. Originally I wanted this to reside outside `bokeh.js` and I may return to this idea, to cover some more cases, like autoload or even guarding bokehjs itself.

bryevdv:
@mattpap I agree it should be on, but there may be folks who would prefer never to have this pop up, even in error cases, for whatever reason, and I would like to make sure that is possible. So is that agreeable, configurable but "on" by default?

mattpap:
@bryevdv, agreed. The question is how to do it, because I don't see the right place where to put the option. Currently we have configurable resources, custom templates, free floating configuration options, environmental variables, all interacting with each other. Additionally there are multiple entry points in the API and multiple targets, each having their requirements and constraints. In longer term I would like to rethink 'bokeh.embed` and `bokeh.resources` and figure out an API that would allow easier configuration and composing different components.

mattpap:
@birdsarah, would you mind testing if the recent commit helps with your issue in Firefox?

bryevdv:
Agree there's lots of cleanup that could happen around configurations and resources to make things more consistent. 

I'd be OK with merging this with just some style work by @birdsarah plus another issue to record the configuration task.

birdsarah:
Styling updated:

before:
screen shot 2016-08-17 at 9 03 46 pm

after:
screen shot 2016-08-17 at 9 03 41 pm

birdsarah:
Tweaked the close button

screen shot 2016-08-17 at 9 15 36 pm

birdsarah:
here's it on a large screen

screen shot 2016-08-17 at 9 16 20 pm

bryevdv:
LGTM, will have to bump the BokehJS size limit

birdsarah:
I've tested on:
OSX - Chrome, Firefox, Safari
Linux (Ubuntu) - Firefox
Windows - IE9 & IE11

birdsarah:
You can also test with just

```python
from bokeh.io import show
from bokeh.models import Plot
show(Plot())
```

mattpap:
Personally I don't like the full width message box, but lets leave this for future iterations. I merged `master`, so size limit test failure should go away.

birdsarah:
screen shot 2016-08-24 at 12 23 37 pm

screen shot 2016-08-24 at 12 23 46 pm

Agree with @mattpap on width. Made a couple of small tweaks. Error message grows to fit message. Message wraps - if it was massive for some reason.

birdsarah:
What do you think about including the stack trace so non webby users could post that in a message to us

screen shot 2016-08-24 at 12 26 53 pm

```diff
diff --git a/bokehjs/src/coffee/safely.coffee b/bokehjs/src/coffee/safely.coffee
index 86ffbdc..8c2c34c 100644
--- a/bokehjs/src/coffee/safely.coffee
+++ b/bokehjs/src/coffee/safely.coffee
@@ -39,11 +39,15 @@ _burst_into_flames = (error) ->
   message = document.createElement("pre")
   message.style["white-space"] = "unset"
   message.appendChild(document.createTextNode(error.message ? error))
+  trace = document.createElement("pre")
+  trace.style["white-space"] = "unset"
+  trace.appendChild(document.createTextNode(error.stack ? error))

   # Add pieces to box
   box.appendChild(button)
   box.appendChild(title)
   box.appendChild(message)
+  box.appendChild(trace)
```

bryevdv:
I like the idea. I would like it slightly better if there is some way to put it behind a collapsible fold with "click for stack trace" or whatever, and also split the lines somehow, but I wouldn't let that hold this up.

bryevdv:
I also think *slightly* more exposition would be helpful, ideally:
```
Bokeh has encountered and error:                                    [x] close

xr is null

<click to expand stack trace>
```

But if nothing else, making the first line a little friendlier, and also add the word "close" to the tiny "x" un the upper right

PR #4962, opened 2016-08-15
Update glyph test examples to fix #4959

### LINKED ISSUE #4959: ImageURL example fails
The ImageURL example at https://github.com/bokeh/bokeh/blob/master/tests/glyphs/ImageURL.py fails with 
`
RuntimeError: Models must be owned by only a single document, <bokeh.models.renderers.GlyphRenderer object at 0x1092b5f90> is already in a doc`

[Bokeh 0.12.1, Python 2.7 and 3.5]

xnx:
- [x] issues: fixes #4959 
As requested by @bryevdv

### DIFF
--- a/tests/glyphs/AnnularWedge.py
+++ b/tests/glyphs/AnnularWedge.py
@@ -1,9 +1,8 @@
 import numpy as np
 
-from bokeh.document import Document
 from bokeh.models import ColumnDataSource, DataRange1d, Plot, LinearAxis, Grid
 from bokeh.models.glyphs import AnnularWedge
-from bokeh.plotting import show
+from bokeh.io import curdoc, show
 
 N = 9
 x = np.linspace(-2, 2, N)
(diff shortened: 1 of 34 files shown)


bryevdv:
Hi @xnx Thanks for the PR! It looks like there is still one line referencing 'doc` in `Gear.py`:
```
./Gear.py:
   61 : doc.add_root(epicyclic)
```
since `doc` is now undefined that's what's causing the CI docs job to fail. I think once that is fixed up this PR will be ready to go.

xnx:
Hi @bryevdv -- I thought I'd fixed that remaining file, but now the unit test fails. I'm not sure why, because it worked previously and the only file I changed, Gear.py works locally for me now.

bryevdv:
Looks like a spurious test failure due to a failed package download, I've restarted the test.

bryevdv:
Tested docs locally, thanks @xnx!

PR #4963, opened 2016-08-16
Add color_mapper model listener to Image glyph for callback driven updates

### LINKED ISSUE #4612: Updating of image ColorMapper
The ability to change the low/high of a ColorMapper for an already existing image (adjusting contrast/brightness) through widget callbacks on a server is desired. I could imagine this being accomplished by the following code, which incidentally doesn't throw any errors as is, it just doesn't do anything ;)

```
import numpy as np
from bokeh.layouts import row, layout
from bokeh.models import Slider
from bokeh.models.mappers import LinearColorMapper
from bokeh.palettes import Greys9

def change_image_contrast(attr, old, new):
    fig_im.glyph.color_mapper.update(low=graph_min_slider.value, high=graph_max_slider.value)
    fig_im.trigger('glyph', fig_im.glyph, fig_im.glyph)

graph_min_slider = Slider(title="Min", start=0, end=99, step=1, value=0)
graph_max_slider = Slider(title="Max", start=1, end=100, step=1, value=100)

graph_min_slider.on_change('value', change_image_contrast)
graph_max_slider.on_change('value', change_image_contrast)

fig = figure(plot_width=500, plot_height=500, x_range=(0, 10), y_range=(0, 10))

fig_im = fig.image(image=[np.random.randint(0, 100, (10, 10), dtype='int16')], x=[0], y=[0], dw=[10], dh=[10],
                   color_mapper=LinearColorMapper(low=0, high=100, palette=Greys9))

curdoc().add_root(layout([fig], [row([graph_min_slider, graph_max_slider])]))
```

canavandl:

- [x] issues: fixes #4612 
- [x] tests added / passed (will do, should be easy pair of mocha tests)

### DIFF
--- a/bokeh/models/mappers.py
+++ b/bokeh/models/mappers.py
@@ -16,6 +16,17 @@ class ColorMapper(Model):
 
     """
 
+    palette = Seq(Color, help="""
+    A sequence of colors to use as the target palette for mapping.
+
+    This property can also be set as a ``String``, to the name of
+    any of the palettes shown in :ref:`bokeh.palettes`.
+    """).accepts(Enum(Palette), lambda pal: getattr(palettes, pal))
+
+    def __init__(self, palette=None, **kwargs):
+        if palette is not None: kwargs['palette'] = palette
+        super(ColorMapper, self).__init__(**kwargs)
+
 class LinearColorMapper(ColorMapper):
     """ Map numbers in a range [*low*, *high*] linearly into a
     sequence of colors (a palette).
@@ -31,13 +42,6 @@ class LinearColorMapper(ColorMapper):
        99 >= x      : 'blue'    # values > high are clamped
 
     """
-
-    palette = Seq(Color, help="""
-    A sequence of colors to use as the target palette for mapping.
-
-    This property can also be set as a ``String``, to the name of
-    any of the palettes shown in :ref:`bokeh.palettes`.
-    """).accepts(Enum(Palette), lambda pal: getattr(palettes, pal))
 
     low = Float(help="""
     The minimum value of the range to map into the palette. Values below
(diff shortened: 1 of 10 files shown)


bokehjs/src/coffee/models/glyphs/image.coffee
```diff
@@ -7,6 +7,13 @@ p = require "../../core/properties"
 
 class ImageView extends Glyph.View
 
+  initialize: (options) ->
+    super(options)
+    @listenTo(@model.color_mapper, 'change', () ->
+      if @_image?
```
canavandl: The nonselection Image glyph doesn't have this._image or this.image_data set, so the `render` raises an exception. This is a non-ideal workaround to that.

It'd probably be better to resolve that issue, so i'll look into it.

bokehjs/src/coffee/models/mappers/linear_color_mapper.coffee
```diff
@@ -22,6 +22,9 @@ class LinearColorMapper extends ColorMapper.Model
       @_reserve_color = parseInt(@get('reserve_color').slice(1), 16)
       @_reserve_val   = @get('reserve_val')
 
+    @listenTo(this, 'change', () ->
```
canavandl: This needs to be factored out onto the parent ColorMapper class, along with some duplicate properties.

bokeh/models/mappers.py
```diff
@@ -16,6 +16,17 @@ class ColorMapper(Model):
...
+    A sequence of colors to use as the target palette for mapping.
+
+    This property can also be set as a ``String``, to the name of
+    any of the palettes shown in :ref:`bokeh.palettes`.
+    """).accepts(Enum(Palette), lambda pal: getattr(palettes, pal))
+
+    def __init__(self, palette=None, **kwargs):
+        if palette is not None: kwargs['palette'] = palette
+        super(ColorMapper, self).__init__(**kwargs)
+
```
bryevdv: Is there a possibility of a `ColorMapper` that does not have a "list of colors" type `palette`? I can sort of imagine that there could possibly be a more general class of color mappers, but I guess we can cross that bridge if it ever happens.

canavandl: I left `low` and `high` on the specific color mappers (log and linear) because I could imagine cases they don't quite work (adding a symmetric log mapper for positive/negative value log mapping).

In the future I could imagine having color functions that are iterators, but that seems a little bit out. Moving palette onto ColorMapper also removes a lot of duplicated code.

bryevdv:
LGTM

PR #4971, opened 2016-08-17
added checks to test that data contains at least one outlier in any c…

### LINKED ISSUE #4970: Box plot example fails if no outliers exist in data
This is an (statistically improbable) edge case, but while experimenting with adapting the code in the boxplot example, I discovered that the code will fail in cases where no outliers exist in the data to be plotted. The problem is that the test on line 32 of `boxplot.py`:
```Python
    if not out.loc[cat].empty:
```
only tests whether there are any outliers for the category, `cat`, in question at the time. If there are no outliers in any of the categories, the script fails here, because actually `out` itself is empty. Error below*:
```
[...]
  File "boxplot_script.py", line 74, in draw_boxplot
    if not out.loc[cat].empty:
  File "/path/to/anaconda/envs/Py35/lib/python3.5/site-packages/pandas/core/indexing.py", line 1296, in __getitem__
    return self._getitem_axis(key, axis=0)
  File "/path/to/anaconda/envs/Py35/lib/python3.5/site-packages/pandas/core/indexing.py", line 1466, in _getitem_axis
    self._has_valid_type(key, axis)
  File "/path/to/anaconda/envs/Py35/lib/python3.5/site-packages/pandas/core/indexing.py", line 1403, in _has_valid_type
    error()
  File "/path/to/anaconda/envs/Py35/lib/python3.5/site-packages/pandas/core/indexing.py", line 1390, in error
    (key, self.obj._get_axis_name(axis)))
KeyError: 'the label [abc] is not in the [index]'
```
I have made a fix for this in the example script, which I will submit as a PR in a couple of minutes.

*note, line number is different because this error is coming from my script, but the code for dealing with outliers is identical between the two scripts.

tobyhodges:
Fixes #4970

### DIFF
--- a/examples/plotting/file/boxplot.py
+++ b/examples/plotting/file/boxplot.py
@@ -27,14 +27,15 @@ def outliers(group):
 out = groups.apply(outliers).dropna()
 
 # prepare outlier data for plotting, we need coordinates for every outlier.
-outx = []
-outy = []
-for cat in cats:
-    # only add outliers if they exist
-    if not out.loc[cat].empty:
-        for value in out[cat]:
-            outx.append(cat)
-            outy.append(value)
+if not out.empty:
+    outx = []
+    outy = []
+    for cat in cats:
+        # only add outliers if they exist
+        if not out.loc[cat].empty:
+            for value in out[cat]:
+                outx.append(cat)
+                outy.append(value)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
LGTM will merge when green, thanks @tobyhodges !

PR #4995, opened 2016-08-22
Split off JS/TS APIs into bokeh-api(.min).js bundle

### LINKED ISSUE #4976: Split off bokehjs JS/TS API

mattpap:
fixes #4976


bokehjs/src/coffee/models/widgets/main.coffee
```diff
@@ -1,34 +1,36 @@
 module.exports = {
-  editors:                  [require('./cell_editors'), 'Editor']
-  formatters:               [require('./cell_formatters'), 'Formatter']
+  models: {
```
bryevdv: OK, good time to make this consistent, but let's add a `MIGRATION` tag and note it

mattpap: Just to note, it's unlikely anyone will actually be affected by this change, because it would require him to `require()` widgets, which only happens once across the repository (in tests).

bryevdv: I agree, I am just trying to develop habits of extreme fastidiousness around any changes.

bryevdv: Also we need to note that `bokeh-api.js` has to be included explicitly now, in any case.

bryevdv:
LGTM

Issue #5001, opened 2016-08-22
availability of CDN resources via HTTPS


jfburkhart:
Documents need updating to reflect availability of CDN resources via HTTPS.

PR #5002, opened 2016-08-22
noted CDN resources are available via HTTPS

jfburkhart:
Fixes documention issue #5001

per discussion here: http://stackoverflow.com/a/39070704/528729

### DIFF
--- a/sphinx/source/docs/user_guide/embed.rst
+++ b/sphinx/source/docs/user_guide/embed.rst
@@ -134,7 +134,11 @@ For example, to use version ``0.12.0``, including widgets support:
 
 .. note::
     You must provide the closing `</script>` tag. This is required by all
-    browsers and the page will typically not render without it.
+    browsers and the page will typically not render without it. You may also
+    encounter a situation in which the browser will not load your page due to
+    an "unsafe" script. In this case, you may need to replace the above URLs
+    with "https" versus "http". That is, CDN resources are also available
+    via HTTPS URLS
 
 The |components| function takes either a single Bokeh Model a list/tuple of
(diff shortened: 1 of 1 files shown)


sphinx/source/docs/user_guide/embed.rst
```diff
@@ -134,7 +134,11 @@ For example, to use version ``0.12.0``, including widgets support:
 
 .. note::
     You must provide the closing `</script>` tag. This is required by all
-    browsers and the page will typically not render without it.
+    browsers and the page will typically not render without it. You may also
+    encounter a situation in which the browser will not load your page due to
+    an "unsafe" script. In this case, you may need to replace the above URLs
+    with "https" versus "http". That is, CDN resources are also available
+    via HTTPS URLS
```
bryevdv: I think it would be best to be as specific about the particular situation as possible:

When embedding in a page served via HTTPS, any scripts and resources must also be loaded via HTTPS or the browser will refuse to load due to an "unsafe" script. For this situation, the Bokeh CDN resources are also available via HTTPS, by replacing "http" with "https" in the above URLS

bryevdv:
@jfburkhart thanks for the PR! I made a suggestion about the wording, then I think this is good to merge.

sphinx/source/docs/user_guide/embed.rst
jfburkhart: @bryevdv thanks. I'll git it another shot. I was trying to squeeze it into the 'note', but really it could be it's own paragraph also.

bryevdv:
Great! stopped tests for docs-only PR, will merge now. Thanks @jfburkhart !

Issue #4987, opened 2016-08-18
Problem with 'Run all' in Jupyter notebooks with bokeh 0.12.1


jbednar:
I can't seem to get 'Run all' to work with Bokeh in my Jupyter 4.1.0 notebooks for Bokeh 0.12.1 (seemed fine in 0.12.0):

![image](https://cloud.githubusercontent.com/assets/1695496/17791331/3c94cafc-6560-11e6-8dde-716bf3c4eb5d.png)

Any ideas?

jbednar:
Code:

```
from bokeh.plotting import figure, output_notebook, show
from bokeh.resources import INLINE
output_notebook(resources=INLINE)
```

```
p = figure()
```

```
show(p)
```

jbednar:
I made it use inline resources only because of the message it showed; it works the same without inline resources.

![image](https://cloud.githubusercontent.com/assets/1695496/17791475/fc46aba4-6560-11e6-98bf-3930c4ba5d4b.png)

jbednar:
Nor is it important that show() is on its own line:

![image](https://cloud.githubusercontent.com/assets/1695496/17791540/4b4cd0de-6561-11e6-9de3-b4f9256d2cab.png)

jbednar:
But it does work fine in bokeh 0.12.0:

![image](https://cloud.githubusercontent.com/assets/1695496/17791626/be239ade-6561-11e6-9157-6b24bd0fb4a4.png)
![image](https://cloud.githubusercontent.com/assets/1695496/17791635/c943f1a2-6561-11e6-8c61-0537a67de7f7.png)

philippjfr:
It comes down to this bit of Javascript not executing in time:

```javascript
Bokeh.$("#%s").text("BokehJS successfully loaded");
var kernel = Jupyter.notebook.kernel
if (kernel.execute !== undefined) {
    kernel.execute("import bokeh.io; bokeh.io._nb_loaded = True");
}
```

When you hit ``Run all`` in the notebook a bunch the cells are queued on the kernel, so asynchronous execution will determine whether the ``Run All`` events or the explicit kernel.execute event is queued first. I don't really see any way around this except for removing the warning.

jbednar:
Thanks!  Seems to work fine if I turn off the checking:

![image](https://cloud.githubusercontent.com/assets/1695496/17794342/25fa2c56-6574-11e6-9265-fdf457e1209a.png)

bryevdv:
[...]

Issue #4980, opened 2016-08-18
Bokeh charts do not render when downloading a Jupyter notebook as HTML


slilienfeld:
I recently updated my version of Bokeh, and all charts look and work exactly as desired in the Jupyter notebook.  However, when downloading the notebook as HTML, none of the charts are showing up.  Version info is as follows:

Python version      :  3.5.2 |Anaconda custom (x86_64)| (default, Jul  2 2016, 17:52:12) 
IPython version     :  5.0.0
Bokeh version       :  0.12.1

bryevdv:
Are there any errors in the JavaScript console of your browser? 

@jbednar did you say you had seen similar?

jbednar:
@bryevdv, sounds like a different issue; when I've had version combinations that made Bokeh charts disappear, they weren't in the notebook version either.

philippjfr:
I just ran into this as well when I used nbconvert on a notebook. This is because this line in the ``output_notebook`` output fails when Jupyter is not available:

```python
var kernel = Jupyter.notebook.kernel
```

slilienfeld:
Yeah just like @philippjfr I'm seeing an 

    Uncaught ReferenceError: Jupyter is not defined

in the JavaScript console of the HTML version of the notebook.

bryevdv:
Well I think it's the same cause? Adding the line `var kernel = Jupyter.notebook.kernel` unconditionally was the mistake. If anything about the `kernel.execute` is to remain for the live notebook case, I guess we just need to test for the existence of `Jupyter` first, and and skip it if not present (on the assumption that it's a static notebook output)

bryevdv:
This is the same of #4987

PR #5014, opened 2016-08-22
Notebook output waits until bokehJS is loaded

philippjfr:
With this PR the code which initializes plots polls whether bokehJS is initialized, delaying rendering of the plot until it is fully loaded. It fixes the issue with Run All (#4987) and nbconvert works again (#4980). To test it you run ``output_notebook``, clear the notebook and reload the page to ensure bokehJS is no longer loaded and then display a plot. The plot will wait to display until you run ``output_notebook`` a second time to load bokehJS. 

- [x] Add a customizable timeout.
- [x] Inject warning into DOM when timeout is reached.
- [x] Update tests

Since keeping around a registry of handles is quite expensive maybe there should be some way to explicitly request a comms handle to be created.

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -34,6 +34,8 @@ calls it with the rendered model.
   if (typeof (window._bokeh_onload_callbacks) === "undefined" || force !== "") {
     window._bokeh_onload_callbacks = [];
     window._bokeh_is_loading = undefined;
+    window._bokeh_init_load = Date.now();
+    window._bokeh_failed_load = false;
   }
 
   function run_callbacks() {
@@ -107,9 +109,12 @@ calls it with the rendered model.
       for (var i = 0; i < inline_js.length; i++) {
         inline_js[i](window.Bokeh);
       }
-    } else {
+    } else if (Date.now() < (window._bokeh_init_load+10000)) {
       setTimeout(run_inline_js, 500);
-    }
+    } else if (!window._bokeh_failed_load) {
+	  console.log("Bokeh: BokehJS failed to load within specified timeout.");
+	  window._bokeh_failed_load = true;
+	}
   }
 
   if (window._bokeh_is_loading === 0) {
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -32,10 +32,10 @@ calls it with the rendered model.
   var force = "{{ force }}";
 
   if (typeof (window._bokeh_onload_callbacks) === "undefined" || force !== "") {
+    {% block init_window %}
     window._bokeh_onload_callbacks = [];
     window._bokeh_is_loading = undefined;
-    window._bokeh_init_load = Date.now();
-    window._bokeh_failed_load = false;
+    {% endblock %}
   }
 
   function run_callbacks() {
@@ -105,16 +105,11 @@ calls it with the rendered model.
   ];
 
   function run_inline_js() {
-    if (window.Bokeh !== undefined || force === "1") {
-      for (var i = 0; i < inline_js.length; i++) {
-        inline_js[i](window.Bokeh);
-      }
-    } else if (Date.now() < (window._bokeh_init_load+10000)) {
-      setTimeout(run_inline_js, 500);
-    } else if (!window._bokeh_failed_load) {
-	  console.log("Bokeh: BokehJS failed to load within specified timeout.");
-	  window._bokeh_failed_load = true;
-	}
+    {% block run_inline_js %}
+    for (var i = 0; i < inline_js.length; i++) {
+      inline_js[i](window.Bokeh);
+    }
+    {% endblock %}
   }
 
   if (window._bokeh_is_loading === 0) {
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -103,8 +103,12 @@ calls it with the rendered model.
   ];
 
   function run_inline_js() {
-    for (var i = 0; i < inline_js.length; i++) {
-      inline_js[i](window.Bokeh);
+    if (window.Bokeh !== undefined || force === "1") {
+      for (var i = 0; i < inline_js.length; i++) {
+        inline_js[i](window.Bokeh);
+      }
+    } else {
+      setTimeout(run_inline_js, 500);
     }
   }
 
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -32,11 +32,13 @@ calls it with the rendered model.
   var force = "{{ force }}";
 
   if (typeof (window._bokeh_onload_callbacks) === "undefined" || force !== "") {
-    {% block init_window %}
     window._bokeh_onload_callbacks = [];
     window._bokeh_is_loading = undefined;
-    {% endblock %}
   }
+
+
+  {% block autoload_init %}
+  {% endblock %}
 
   function run_callbacks() {
     window._bokeh_onload_callbacks.forEach(function(callback) { callback() });
--- a/bokeh/core/_templates/notebook_js.js
+++ b/bokeh/core/_templates/notebook_js.js
@@ -2,13 +2,7 @@ var docs_json = {{ docs_json }};
 var render_items = {{ render_items }};
 var comms_target = "{{ comms_target }}";
 
-(function embed_items() {
-  if (window.Bokeh !== undefined) {
-    Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
-    if (window.Jupyter !== undefined) {
-      Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
-	}
-  } else if (Date.now() < (window._bokeh_init_load)) {
-    setTimeout(run_inline_js, 100);
-  }
-})();
+Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
+if (window.Jupyter !== undefined) {
+  Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
+}
--- a/bokeh/core/_templates/notebook_js.js
+++ b/bokeh/core/_templates/notebook_js.js
@@ -8,7 +8,7 @@ var comms_target = "{{ comms_target }}";
     if (window.Jupyter !== undefined) {
       Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
 	}
-  } else if (Date.now() < (window._bokeh_init_load+10000)) {
-    setTimeout(run_inline_js, 500);
+  } else if (Date.now() < (window._bokeh_init_load)) {
+    setTimeout(run_inline_js, 100);
   }
 })();
--- a/bokeh/core/_templates/notebook_js.js
+++ b/bokeh/core/_templates/notebook_js.js
@@ -1,0 +1,14 @@
+var docs_json = {{ docs_json }};
+var render_items = {{ render_items }};
+var comms_target = "{{ comms_target }}";
+
+(function embed_items() {
+  if (window.Bokeh !== undefined) {
+    Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
+    if (typeof Jupyter !== "undefined") {
+      Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
+	}
+  } else {
+    setTimeout(embed_items, 500)
+  }
+})();
--- a/bokeh/core/_templates/notebook_js.js
+++ b/bokeh/core/_templates/notebook_js.js
@@ -5,10 +5,10 @@ var comms_target = "{{ comms_target }}";
 (function embed_items() {
   if (window.Bokeh !== undefined) {
     Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
-    if (typeof Jupyter !== "undefined") {
+    if (window.Jupyter !== undefined) {
       Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
 	}
-  } else {
-    setTimeout(embed_items, 500)
+  } else if (Date.now() < (window._bokeh_init_load+10000)) {
+    setTimeout(run_inline_js, 500);
   }
 })();
--- a/bokeh/core/_templates/notebook_js.js
+++ b/bokeh/core/_templates/notebook_js.js
@@ -1,8 +1,0 @@
-var docs_json = {{ docs_json }};
-var render_items = {{ render_items }};
-var comms_target = "{{ comms_target }}";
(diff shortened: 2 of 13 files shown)


philippjfr:
Ping @bryevdv and @damianavila.

bokeh/core/_templates/autoload_js.js
```diff
@@ -103,8 +103,12 @@ calls it with the rendered model.
...
 
   function run_inline_js() {
-    for (var i = 0; i < inline_js.length; i++) {
-      inline_js[i](window.Bokeh);
+    if (window.Bokeh !== undefined || force === "1") {
+      for (var i = 0; i < inline_js.length; i++) {
+        inline_js[i](window.Bokeh);
+      }
+    } else {
+      setTimeout(run_inline_js, 500);
```
bryevdv: This might need some care. This template is also used by e.g. `autoload_server` ping @mattpap

philippjfr: I suspected as much, didn't want to create a new template for it just yet though.

bokeh/embed.py
```diff
@@ -462,6 +465,20 @@ def _script_for_render_items(docs_json, render_items, websocket_url=None, wrap_s
...
+
+def _script_for_notebook_render_items(docs_json, render_items, notebook_comms_target=None,
+                                      wrap_script=True):
+    plot_js = _wrap_in_function(NOTEBOOK_JS.render(
+        docs_json=serialize_json(docs_json),
+        render_items=serialize_json(render_items),
+        comms_target=notebook_comms_target
+    ))
+
+    if wrap_script:
```
birdsarah: I don't think you should bother with the wrap_script option. It's very specifically available for `embed.components` which doesn't apply here and as far as I can tell is always going to be `False`.

bokeh/core/_templates/autoload_js.js
birdsarah: So this just runs over and over again until window.Bokeh is defined?

Should we stop after 10 tries.

philippjfr: Yes, definitely should define a configurable timeout of some kind.

bokeh/core/_templates/notebook_js.js
```diff
@@ -0,0 +1,14 @@
...
+
+(function embed_items() {
+  if (window.Bokeh !== undefined) {
+    Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
+    if (typeof Jupyter !== "undefined") {
+      Jupyter.notebook.kernel.execute("import bokeh; bokeh.io._comms_handles['"+comms_target+"'].init()");
+	}
+  } else {
+    setTimeout(embed_items, 500)
+  }
```
birdsarah: I am pretty clueless on how the notebook works so excuse my ignorance. If someone was just looking at a saved notebook on something like nbviewer would this cause a problem?

philippjfr: It did before, the ``if (typeof Jupyter !== "undefined") {`` part should handle that case.

birdsarah:
It'll be interesting figuring out what kind of tests we can add for this?

@bryevdv - should we start adding a list of manual tests under the `tests` directory for stuff like this that's hard to automate.

bryevdv:
> should we start adding a list of manual tests under the tests directory for stuff like this that's hard to automate.

+1

bokeh/core/_templates/autoload_js.js
bryevdv: I'm also not sure we should or would need to re-run in the inline case?

bokeh/core/_templates/notebook_js.js
```diff
@@ -0,0 +1,14 @@
+var docs_json = {{ docs_json }};
+var render_items = {{ render_items }};
+var comms_target = "{{ comms_target }}";
+
+(function embed_items() {
+  if (window.Bokeh !== undefined) {
+    Bokeh.embed.embed_items(docs_json, render_items{%- if websocket_url -%}, "{{ websocket_url }}" {%- endif -%});
+    if (typeof Jupyter !== "undefined") {
```
bryevdv: `window.Jupyter` ? or change the `window.Bokeh` one above? would just like to be consistent.

bokeh/util/notebook.py
```diff
@@ -32,10 +32,6 @@ def load_notebook(resources=None, verbose=False, hide_banner=False):
 
 FINALIZE_JS = """
 Bokeh.$("#%s").text("BokehJS successfully loaded");
```
bryevdv: This should probably change to somethign like `BokehJS loading...` and then if there's a way to add a `done.` at the end after some timeout / check, great.

jbednar:
This sounds promising so far...

philippjfr:
Okay before I continue I have several questions:

1. Are we okay with this approach?
2. What should I do about autoload_js.js, which is not exclusively used by the notebook? Should I create a new template?
3. How should I expose the polling interval and maximum timeout to the user?
4. What message should I display if the the maximum timeout has been exceeded, and how should I display it?
5. Should I just remove the existing tests for this?

bokeh/core/_templates/autoload_js.js
```diff
@@ -103,9 +105,16 @@ calls it with the rendered model.
...
   function run_inline_js() {
-    for (var i = 0; i < inline_js.length; i++) {
-      inline_js[i](window.Bokeh);
-    }
+    if (window.Bokeh !== undefined || force === "1") {
+      for (var i = 0; i < inline_js.length; i++) {
+        inline_js[i](window.Bokeh);
+      }
+    } else if (Date.now() < (window._bokeh_init_load+10000)) {
+      setTimeout(run_inline_js, 500);
```
damianavila: What about using a setInterval and clear the interval when window.Bokeh is defined?
I am OK with this implementation but setInterval feels to me like best suited for this use case...

philippjfr: Not quite sure how that will work. First I'd have to keep handles on all the intervals somewhere, and if I just cleared them all when window.Bokeh is defined that means they would no longer execute. Maybe I'm missing something though.

bryevdv:
ping @mattpap there are questions here that could benefit from your input

bryevdv:
@philippjfr yes I am generally OK with this approach. 

If BokehJS fails to load after a timeout, I think just display the old bug warning, updated to include a line about increasing the timeout (however that is done). 

For now, I's say expose the timeout in the template that `output_notebook` publishes, as a parameter to `output_notebook` @mattpap do you concur?

I'd really like to get @mattpap to chime in on `autoload_js.js`

Remove any existing tests that are now wholly inapplicable. Anything that can be updated, should be updated. 

Lasty, once BokehJS is loaded, it would be good to update the `"BokehJS loading..."` message with a `"done."` at the end.

mattpap:
> I'd really like to get @mattpap to chime in on autoload_js.js

I'm reviewing this.

philippjfr:
@mattpap I just added templating blocks to the autoload_js.js template and extended it with the notebook specific code in a new autoload_nb_js.js template. Let me know what you think.

philippjfr:
@bryevdv The PR is ready for another round of review. The timeout is now customizable, there is a separate ``autoload_nb_js.js`` template and it now inserts a warning into the cells once the timeout is reached. Unfortunately I had to move the string from Python into the template as a really ugly HTML string. Testing it is also a pain as you have to fiddle with the code to either delay or completely stop loading bokehJS. The current failure output looks like this:

image

I've noticed you don't use jQuery in the templates, but I've used it here to quickly grab the parent cell, which I then append the HTML warning to. It should be safe since the notebook uses jQuery anyway, but I can probably also do it without jQuery if you'd prefer.

It's getting late here, so tomorrow I'll insert the Bokeh loaded message and look at the failing tests.

philippjfr:
The message can probably be shortened now that it should only fail if network resources are really slow.

jbednar:
Does that mean that output_notebook no longer needs to be in a cell by itself, at the top, with no other code? If any of those restrictions can now be lifted, it would be great for the message to reflect that.

philippjfr:
>Does that mean that output_notebook no longer needs to be in a cell by itself, at the top, with no other code? If any of those restrictions can now be lifted, it would be great for the message to reflect that.

Right, you should be able to put it in the same cell as your first plot without issues.

jbednar:
That should cut down on a lot of the problems people have.

philippjfr:
Mysteriously TestBokehJS.test_bokehjs is failing but only in Python 2.7.

bryevdv:
We currently suffer some spurious test failures. I've restarted, checking this out now.

philippjfr:
Thought it might be that. My latest commit with a fix when ``hide_banner=True`` isn't quite right. Will fix that in a minute.

bryevdv:
We still have async issues, just different ones. Using "Run all" on the "continuous updating" notebook in `examples/howto/notebook_comms` results in this:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-5-987cc5698708> in <module>()
     16 
     17     # push updates to the plot continuously using the handle (intererrupt the notebook kernel to stop)
---> 18     push_notebook(handle=target)
     19     time.sleep(0.1)

/Users/bryan/work/bokeh/bokeh/io.py in push_notebook(document, state, handle)
    583     else:
    584         msg = Document._compute_patch_between_json(handle.json, to_json)
--> 585     handle.comms.send(json.dumps(msg))
    586     handle.update(document, to_json)
    587 

AttributeError: 'NoneType' object has no attribute 'send'
```

Which I am sure is the same problem as the original, moved to a new place: the JS call to `kernel.execute` does not happen in time when "Run all" is used. 

Given this, I believe there is no use of `kernel.execute` that is safe in any way. 

So here is my suggestion for notebook handles (some of this is we already do):

* Generate a unique ID in python (i.e. in `show`)
* pass this to JS to create JS side of notebook comm
* this should also be optional, see https://github.com/bokeh/bokeh/pull/5033
* store the comm target id somewhere on the python side
* instead of using `kernel.execute` to `init` the comm on the python side, *users must do this explicitly*
```
h = show(plot, notebook_handle=True)

bokeh.io.handles[h].init() # users responsible for this now
```
The part in italics is the new part. 

All that said, I'm *still* not sure this will work, if "Run all" preferentially executes python before the JS then it may simply be literally technically impossible to make notebook handles robust together with "Run all"

philippjfr:
>We still have async issues, just different ones. Using "Run all" on the "continuous updating" notebook in examples/howto/notebook_comms results in this:

D'oh, yes, should have thought about that. The thing is this issue at least won't cause bokeh plots to fall over completely.

>All that said, I'm still not sure this will work, if "Run all" preferentially executes python before the JS then the same may be true even if the users calls handle.init instead of the JS code using kernel.execute to try to do it automatically. It may simply be literally technically impossible to make notebook handles robust together with "Run all"

The only other option I see is for autoload_nb_js.js to register the comm immediately and then pass it to Bokeh.embed.embed_items once BokehJS is loaded, which will then register the ``on_msg`` callback on the Comm. This would involve moving some of the logic in ``Bokeh.embed._init_comms`` into the template but would avoid all async issues I think. The initial push event would then just have to automatically call init on the Comm before sending the data. This would ensure the Comm is registered by the time the user tries to push the first time, although initial events might still get swallowed if the on_msg event has not yet been registered. 

Having the user explicitly request a comms handle seems like a good idea regardless. Let me know what you think and I'll make the appropriate changes to the PR.

bryevdv:
Lets try your suggestion. Can you make the connection return optional as well, in this PR? And we will supersede the other PR

ellisonbg:
How are you loading code on the page? The right way of doing it is to use something like require.js that will not run the code until all of the dependencies are loaded. If you are just throwing a script tag on the page, then you don't have any way of knowing when the code is ready.

philippjfr:
Thanks for your input @ellisonbg! Require.js definitely solves the issue of waiting until the bokehJS dependency is loaded and maybe we should replace my custom solution with require.js (your call @bryevdv). The problem is that we also want to register a ipykernel.comm.Comm to allow pushing updates to the plot and it must be registered on Jupyter.notebook.kernel.comm_manager before you can create the Comm in Python. Currently the only solution I see is to register the target for the Comm immediately, and then attach the on_msg callback when bokehJS is loaded.

ellisonbg:
From conversation with @bryevdv on gitter:

Here is a rough guide:

1. For dependencies that all cells will need, create a bundle of JS code that is installed as an nbextension (doesn't need to be activated)
2. For each cell that needs to use that JS, publish a text/javascript output that uses a require.js call to get the Js bundle from 1

An example is how Altair does this here:

Here is the JS template we use for 2):

https://github.com/vega/ipyvega/blob/master/vega/static/vega.js

Here is the JS bundle that is 1):

https://github.com/vega/ipyvega/blob/master/src/index.js

And the build file for 1)

https://github.com/vega/ipyvega/blob/master/package.json

ellisonbg:
And also:

If you want to have JS that is shared between different cells, then you really need to do something that puts the common JS stuff at a URL location where you can require.js it. That doesn't have to be an extension, but it does have to be some reachable URL:

* npmcdn
* A local file that is written temporarily to a directory the notebook server can serve
* A formal Jupyter nbextension

You just need to load it by URL, not by putting it in a script tag. Otherwise you are going to reinvent require.js.

philippjfr:
@ellisonbg Thanks for the detailed instructions. I'm not sure we're going to commit to shipping an nbextension straight away, so I just discussed this with @bryevdv and we were still a bit unclear on the option of loading a local files. Just to clarify, can the notebook server easily be told to serve an existing path (e.g. the build path of bokehJS) or do we really have to copy it somewhere the notebook can find it?

bryevdv:
Added `MIGRATION` for now having to explicitly request a handle. 

Tested this with all the notebook comms notebooks, all working as expected. @jbednar can you run this through its paces on your Datashader notebooks?

I'd like to ultimately pursue a solution with `require` as @ellisonbg described, but I think it will take more thought and planning, because we could handle "local" resources that way too, but that's different than how "inline" is handled. Either "inline" can behave differently for notebooks or we need a new resource type ("local"?) but that's all a wider discussion. ping @mattpap  

If this is working for:

* single step
* run all
* nbconvert

Across a wide range of notebooks (data shader, some with comms handles) then I would suggest its better to get this in now and make a separate issue to explore the `require` idea for a longer term improvement.

philippjfr:
@jbednar Datashader will need a small fix in ``InteractiveImage`` because the CommsHandle has changed slightly, so I'll open a PR to fix and test it.

philippjfr:
@jbednar Actually the changes I made to the _CommsHandle are no longer necessary so I just undid them, datashader notebook should now just work again.

bryevdv:
@philippjfr can you explain why the explicit comms handle init is not needed? Also there seems to be a real unit test that needs updating, otherwise looking close.

philippjfr:
@bryevdv Because now the template registers a comms target with Jupyter immediately, so it's not dependent on BokehJS being loaded. When bokeh does load it will first look if a comm has already been initialized for the specified target and attach itself with a callback using Comm.on_msg. Hope that makes sense. It feels a little bit dirty but ensures the comm target is always registered in time for the Comm to be created and initialized in Python.

bryevdv:
@philippjfr OK, I guess I was under the impression that *all* JS might get run "late" in the "Run all" case, is it truly just the async script load from CDN that might happen too late?

philippjfr:
@bryevdv I think that is the case yes. If you couldn't rely on that then Comms could never be guaranteed to work with Run All, because they always have to be registered on the frontend first. I can see how it might be a problem but I've never seen it cause any issues.

bryevdv:
OK I tested this with a number of notebooks with inline resources. I don't see any simple path to test it with CDN resources apart from merging it and cutting a dev build, so I will do that now. I suppose this points to an occasional need to cut dev builds from branches.

PR #5017, opened 2016-08-23
Reduce markers size by ~13k with generative techniques

bryevdv:
- [ ] issues: fixes #xxxx
- [ ] tests added / passed

Since this marker code is so completely ossified and established at this point, I think that a little obfuscation at this point can be tolerated, for other gains. in addition to the space savings, since these are now just plain functions, we can easily add tests for them. I expect there is some very slight performance penalty. I will try to characterize it in the future,  but informal testing with ~5k points seem indistinguishable. 

All markers (regular and GL) visually confirmed correct in `webgl/marker_compare.py`

@mattpap can you share your thoughts? I'm also not sure what work on the API might need to be fixed up.


bryevdv:
I think if we cleave off `bokeh-gl.js` then together with you API split we can reduce ~100k off the minified JS.

bokehjs/src/coffee/models/markers/all_markers.coffee
```diff
@@ -0,0 +1,204 @@
...
+    _one_cross(ctx, r)
+    ctx.stroke()
+
+  return
+
+_square_x = (ctx, i, sx, sy, r, vl, vf) ->
+  size = 2*r
+  ctx.rect(-r, -r, size, size)
+
+  if vf.doit
```
mattpap: Proper naming should have been kept. It took me a while to figure out that `vl` stands for and that it's not actually `v1` (stupid fonts).

bokehjs/src/coffee/api/models.coffee
```diff
@@ -81,20 +81,7 @@ module.exports = {
...
-  Cross:                                  require("../models/markers/cross").Model
-  Diamond:                                require("../models/markers/diamond").Model
-  DiamondCross:                           require("../models/markers/diamond_cross").Model
-  InvertedTriangle:                       require("../models/markers/inverted_triangle").Model
-  Square:                                 require("../models/markers/square").Model
-  SquareCross:                            require("../models/markers/square_cross").Model
-  SquareX:                                require("../models/markers/square_x").Model
-  Triangle:                               require("../models/markers/triangle").Model
-  X:                                      require("../models/markers/x").Model
+  markers:                                [require('../models/markers/all_markers')]
```
mattpap: node.js-idiomatic module name is `markers/index`.

bokehjs/src/coffee/api/models.coffee
```diff
@@ -81,20 +81,7 @@ module.exports = {
   LinearColorMapper:                      require("../models/mappers/linear_color_mapper").Model
 
   ## api/typings/models/markers.d.ts
-  Marker:                                 require("../models/markers/marker").Model
-  Asterisk:                               require("../models/markers/asterisk").Model
-  Circle:                                 require("../models/glyphs/circle").Model
```
mattpap: This entry should be restored and moved up into the proper group.

mattpap:
I just want to make sure that with all numbers given we are speaking about savings in minified code?

bokehjs/src/coffee/models/markers/all_markers.coffee
```diff
@@ -0,0 +1,204 @@
...
+  _one_x(ctx, r)
+
+  if vl.doit
+    vl.set_vectorize(ctx, i)
+    ctx.stroke()
+
+  return
+
+module.exports =
+  Asterisk:         M.gen('Asterisk', _asterisk)
```
mattpap: `Marker.gen(...)` for the same reasons as before. It's minifier's job to truncate this name.

mattpap: Actually it would have been even better to import `gen` with `{gen} = require("./marker")`.

mattpap: Or even better, `gen` should be a private function of `all_modules`. Unless you plan to use it elsewhere.

bryevdv: Yes I thought of that this morning in my drive. :) lll make that change.

bryevdv:
Yes reduction is quoted for minified code. We have a *lot* of long strings and identifiers that the minifier won't touch. Combining things into a single index helps too.

bokehjs/src/coffee/api/models.coffee
bryevdv: Ah yes better name thanks

bokehjs/src/coffee/api/models.coffee
bryevdv: Good catch thanks

bokehjs/src/coffee/models/markers/all_markers.coffee
bryevdv: If the minifier will shorten them now that they are passed in as parameters then sure, agree.

bryevdv: Also FYI SourceCodePro (free font) was designed to disambiguate 1/l and 0/O etc

mattpap: Private variables are all going to be shortened.

bryevdv:
@mattpap I m ade your suggestions. Also some informal benchmarking with 15k markers of various types indicated very minimal differences if at all (e.g. 320ms vs 310ms).

bryevdv:
Also, currently:
```
  - bokeh.min.js          :  872.3 KB
```
vs about 885k without the PR

mattpap:
> Also some informal benchmarking with 15k markers of various types indicated very minimal differences if at all

To me it was sufficient to read the code. There's one extra function call per rendering, so there should be no visible speed penalty.

bryevdv:
> To me it was sufficient to read the code. There's one extra function call per rendering, so there should be no visible speed penalty.

Right, I am still unlearning old lessons from different contexts. 

I'd like to pull all the actual functions out into a util module so that they can be tested, but that won't be until later this week, unless you want to make a pass at it.

mattpap:
Anyway, this looks good. I never really understood why every marker has each own model (well I know, we can override features, etc.), so this goes in the right direction. Still the remaining piece of the puzzle is `Circle`, which, across the code base, is understood either as a marker or a glyph. This requires some further work (per markers are all figured out).

bryevdv:
I'd be OK with making `Circle` a `Marker` if there is some reasonable way to have it also accommodate radius (or I guess an alternative is to extend radius support to all markers)

mattpap:
I didn't give the `Circle` issue a thought yet. Just pointing out, because it has moderately bothered me for a long while.

bryevdv:
> I didn't give the Circle issue a thought yet. Just pointing out, because it has moderately bothered me for a long while.

Me too. 

As an aside, at long last I have realized the problem with the original "glyph spec" was not that it existed, but that it was not a standard model. Making things models makes serialization uniform, but that does not mean they all need to be heavyweight. 

Anyway, getting off track, do you want to make a stab at some tests for the pure functions `triangle` etc. with a mock canvas? If not I'll do it in a few days.

mattpap:
If image diff doesn't complain, we can postpone testing.

bryevdv:
`SquareX` is ever so slightly dilated: 

https://s3.amazonaws.com/bokeh-travis/154592447/examples.html

Not sure why, I think it looks fine either way...

bryevdv:
ping @birdsarah realistically I am not going to be able to add tests to this this week. Thoughts? It does not add any (but does also not remove any) but does largely reduce the amount of code to maintain, so I think it's a win for now.

PR #5018, opened 2016-08-23
Canavandl/reverse vertical ticking order

### LINKED ISSUE #4990: Colorbar default direction should be reversed
I just noticed that the color bar default direction is going from low (at top) to high (at bottom).

I think it should be the other way around with high up-top and low at bottom by default.

ping @canavandl  - sorry for noticing this earlier.

canavandl:
- [x] issues: fixes #4990
- [x] tests added / passed

Reverses the direction of the vertical ColorBar (high values are at the top)


canavandl:
from selenium tests:

screen shot 2016-08-22 at 10 45 03 pm

ping @birdsarah

bokehjs/test/models/annotations/color_bar.coffee
```diff
@@ -198,25 +198,51 @@ describe "ColorBar module", ->
 
       beforeEach ->
         @plot.add_layout(@color_bar)
+        @lin_expected = new Float64Array([0, 20, 40, 60, 80, 100])
+        @log_expected = new Float64Array([0, 76.70099985546604, 86.73533304426542, 92.60504167945479, 96.76966623306478, 100])
```
birdsarah: These aren't used anywhere.

birdsarah:
Made one small comment. 

LGTM - Haven't run it locally but with the selenium test am confident it's working as hoped for.

bryevdv:
Crap, I clicked merge by accident. I'l clean those lines up in another open PR as some general cleanup.

bryevdv:
Propose that LGTM and "there are comments" be mutually exclusive :)

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -46,6 +46,9 @@ class ColorBarView extends Annotation.View
   _set_canvas_image: () ->
     palette = @model.color_mapper.palette
 
+    if @model.orientation == 'vertical'
+      palette = palette.slice(0).reverse()
```
mattpap: There must be a better way of expressing an intent to create a copy of an array in a more readable way. Not blaming the author, because I've been using the same idiom myself. I looked at underscore.js, but its documentation says there's no appropriate function, yet the implementation says something different. One can use `_.clone()` for this.

canavandl: good to know. thanks

PR #5022, opened 2016-08-23
Switch resource ordering

### LINKED ISSUE #5021: Render js_raw before js_files
I had a situation where I had custom models with external resources.

I wanted to use bokeh with `resources=local` but because in the template for bokeh_js - https://github.com/bokeh/bokeh/blob/master/bokeh/core/_templates/js_resources.html - files are rendered before raw, my document didn't load because I need bokeh loaded first.

I can't see any harm in switching the order.

birdsarah:
- [x] issues: fixes #5021

### DIFF
--- a/bokeh/core/_templates/js_resources.html
+++ b/bokeh/core/_templates/js_resources.html
@@ -9,12 +9,13 @@ configuration in a Resources object.
 :type js_raw: list[str]
 
 #}
-{%- for file in js_files %}
-<script type="text/javascript" src="{{ file }}"></script>
-{%- endfor %}
 
 {%- for js in js_raw %}
 <script type="text/javascript">
     {{ js|indent(8) }}
 </script>
+{%- endfor %}
+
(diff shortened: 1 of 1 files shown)


mattpap:
I have to give this a thought, but this shouldn't work.

mattpap:
With this change, e.g. setting `log_level` will fail with `Bokeh is not defined`. There is an assumption that `js_raw` may depend on `Bokeh`. This is also nicely visible in `autoload_js` template.

mattpap:
> I had a situation where I had custom models with external resources.

Can I get some specifics (maybe some code)? This may need to be resolved as part of PR #4946.

birdsarah:
> With this change, e.g., setting log_level will fail with Bokeh is not defined.

Yep - discovered that - happy to close this.

> Can I get some specifics (maybe some code)? This may need to be resolved as part of PR #4946.

I'll try and rustle something up on the original issue #5021

PR #5023, opened 2016-08-23
Switch to isNaN

birdsarah:
`_.isNaN` seems to be too strict.


mattpap:
This could have been done across the repository. Additionally, some one could have taken time to review this.

birdsarah:
I don't know about across the repository - I was correcting code I'd just written and realized wasn't working as expected when the values went from python to js, but needed somewhat urgently. Apologies.

PR #5024, opened 2016-08-23
Make core bokeh independent of numpy

### LINKED ISSUE #1280: Hard dependency on numpy
According to the documentation Bokeh shouldn't have a hard dependency on numpy and should work only with python lists. I installed bokeh in a virtualenv with the `--no-deps` option, no numpy is installed inside the environment. When I try to import something from bokeh I get:

```python
>>> from bokeh.plotting import *
  File "testbokeh.py", line 7, in <module>
    from bokeh.plotting import *
  File "/Users/yaser/.virtualenvs/testenv/lib/python2.7/site-packages/bokeh/plotting.py", line 13, in <module>
    from . import _glyph_functions as gf
  File "/Users/yaser/.virtualenvs/testenv/lib/python2.7/site-packages/bokeh/_glyph_functions.py", line 2, in <module>
    from . import glyphs
  File "/Users/yaser/.virtualenvs/testenv/lib/python2.7/site-packages/bokeh/glyphs.py", line 1, in <module>
    from .properties import Align, Bool, DataSpec, Enum, HasProps, Size, Any, Color
  File "/Users/yaser/.virtualenvs/testenv/lib/python2.7/site-packages/bokeh/properties.py", line 16, in <module>
    import numpy as np
ImportError: No module named numpy
```
Steps to reproduce:
```
$ mkvirtualenv testenv
(testenv)$ pip install bokeh --no-deps
(testenv)$ pip install six python-dateutil
(testenv)$ python
>>> from bokeh.plotting import scatter

mattpap:
This is a proof of concept for issue #1280. It's easy to make numpy optional, but the real trouble is testing this.

fixes #1280

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -11,12 +11,11 @@ import decimal
 import json
 import time
 
-import numpy as np
-
 from ..settings import settings
 from ..util.dependencies import import_optional
 from ..util.serialization import transform_series, transform_array
 
+np = import_optional('numpy')
 pd = import_optional('pandas')
 rd = import_optional("dateutil.relativedelta")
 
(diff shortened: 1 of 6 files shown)


bryevdv:
I should have mentioned in the issue, but another reason I am -1 on this is that removing numpy from the deps list means it won't get installed automatically. Which is great for the folks who want to use Bokeh w/o numpy. I would contend the vast number of Bokeh users *do* want to use Bokeh with numpy, which means potentially another install step. In partcular, on a clean env "conda install  bokeh" would not do enough on its own to run many of the examples.

mattpap:
@bryevdv, I agree that numpy should be installed by default with bokeh.

bryevdv:
But then whats the point of making it optional? I guess if people go to the trouble of pip or conda installing with `--no-deps`? I just don't see this as a priority, and moreover, do see it as a risk.

mattpap:
Personally I'm in favor of Python libraries that can be used without installing heavy dependencies, even if the default is to make a package depend on such. There are environments that can take advantage of such fact. Anyway, this PR is a proof of concept worth 10 minutes of work, which may or may not help in removing implementation cost from the overall discussion about this feature.

mattpap:
I doubt we will make bokeh's core numpy-independent, but in any case, the code is here.

Issue #5029, opened 2016-08-24
importing us county data fails on 3.5


crwilcox:
When running `from bokeh.sampledata.us_counties import data as counties` in Python 3.5 an error occurs.

```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-6-a97873fe504b> in <module>()
      2 from bokeh.palettes import Viridis6
      3 from bokeh.plotting import figure, show, output_file, ColumnDataSource
----> 4 from bokeh.sampledata.us_counties import data as counties
      5 from bokeh.sampledata.unemployment import data as unemployment

/home/nbuser/anaconda3_410/lib/python3.5/site-packages/bokeh/sampledata/us_counties.py in <module>()
     24     next(f)
     25     reader = csv.reader(f, delimiter=',', quotechar='"')
---> 26     for row in reader:
     27         name, dummy, state, dummy, geometry, dummy, dummy, dummy, dummy, state_id, county_id, dummy, dummy = row
     28         xml = et.fromstring(geometry)

/home/nbuser/anaconda3_410/lib/python3.5/encodings/ascii.py in decode(self, input, final)
     24 class IncrementalDecoder(codecs.IncrementalDecoder):
     25     def decode(self, input, final=False):
---> 26         return codecs.ascii_decode(input, self.errors)[0]
     27 
     28 class StreamWriter(Codec,codecs.StreamWriter):

UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 2514: ordinal not in range(128)
```

The fix is to provide an encoding (utf-8) to properly decode the file. Otherwise open will use the preferred local and guess. In my case, it attempted ascii decoding which failed.
`with open(join(data_dir, 'US_Counties.csv'), encoding='utf-8') as f:`

birdsarah:
I have a conda env with the following packages installed:

```sh
# packages in environment at /Users/sbird/miniconda3/envs/python_35:
#
bokeh                     0.12.2dev7                 py_1    bokeh/channel/dev
ca-certificates           2016.8.2                      3    conda-forge
certifi                   2016.8.2                 py35_0    conda-forge
jinja2                    2.8                      py35_1    conda-forge
markupsafe                0.23                     py35_0    conda-forge
ncurses                   5.9                           8    conda-forge
openssl                   1.0.2h                        2    conda-forge
pip                       8.1.2                    py35_0    conda-forge
python                    3.5.2                         2    conda-forge
[...]

PR #5030, opened 2016-08-24
Sets encoding to utf-8 for python 3+

crwilcox:
Fixes Issue #5029 where downloading sample US_Counties data on Ubuntu 16 with an ansi preferred local encountered issues decoding the csv encoded as utf-8.


birdsarah:
The integration tests will fail - don't worry about that (I need to fix this).

Thanks for contributing this.

All/most of our sample data files are going to have this issue - if you're feeling motivated you could tack on a few more fixes like this to the other sampledata files.

birdsarah:
This PR breaks python2.7 for me. I get the following error:

```python
Python 2.7.12 | packaged by conda-forge | (default, Aug 12 2016, 19:13:31)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from bokeh.sampledata import us_counties
  File "bokeh/sampledata/us_counties.py", line 28, in <module>
    for row in reader:
UnicodeEncodeError: 'ascii' codec can't encode character u'\ufffd' in position 1: ordinal not in range(128)
```

birdsarah:
This also has  a code quality violation

```sh

E         File contains trailing whitespace: bokeh/sampledata/us_counties.py, line 25.

E         File contains trailing whitespace: bokeh/sampledata/us_counties.py, line 30.

E       assert 2 == 0

E        +  where 2 = len(['File contains trailing whitespace: bokeh/sampledata/us_counties.py, line 25.', 'File contains trailing whitespace: bokeh/sampledata/us_counties.py, line 30.'])
```

crwilcox:
I had tried this on both 2.7.12 and 3.5.1

I will look into this in just a bit. Thanks for verifying that stuff. I will dig in deeper. While I do I will look through the other samples as well.

Thanks!

birdsarah:
Please don't hesitate to ask questions about dev setup if there's anything confusing. We've been trying to make the docs better - but there's a way to go.

crwilcox:
Thanks for your awesome work on bokeh and thanks for being open to contributions. When things are simple it makes sense to submit the request and work to resolve the issue and not just file a bug on the (usually overburdened) project maintainers.

I was able to address the issues. When I was testing from ubuntu on 2 and 3 it all worked with my fix. When I tried on windows I had the same issue you had on osx though. I have committed a revised fix. I still need to look into fixing this for the other samples. 

The root cause seems to be a change to the csv reader between py2 and py3.

I have started setting up my dev environment but kept hitting new things. I now need pandas as well as the stuff in devdeps.py to get going it seems for instance? Might be worth installing anaconda and not doing virtualenvs since a few things are needed. Your dev docs are better than a lot but like most a bit of TLC is usually always needed ;)

bryevdv:
Hi @crwilcox Thanks for your PR! FYI currently the "integration" CI test is expected to fail for anyone not in @bokeh/dev, we are working on improving our test automation to be more resilient, though.

bokeh/sampledata/us_counties.py
```diff
@@ -20,7 +21,15 @@
 nan = float('NaN')
 
 data = {}
-with open(join(data_dir, 'US_Counties.csv')) as f:
+
+# csv differs in Python 2.x and Python 3.x. Open the file differently in each.
+filename = join(data_dir, 'US_Counties.csv')
+if sys.version_info[0] < 3:
```
mattpap: Use `if six.PY2:` instead.

bokeh/sampledata/us_counties.py
```diff
@@ -20,7 +21,15 @@
...
+    f = open(filename, 'rb')
+else:
+    f = open(filename, 'r', newline='', encoding='utf8')
+
```
mattpap: Perhaps using `io.open()` would remove the need for this `if-else` statement? It supports `encoding` on both Py 2 and 3.

crwilcox: @mattpap, I thought about that. The difference in opening r vs rb though and the other settings made me pause on that. Do you have an example of what you are thinking? Truthfully this seemed obvious to me what I was doing and I didn't try to further reduce.

bokeh/sampledata/us_counties.py
crwilcox: done.

bryevdv:
Marking `ready` @mattpap @birdsarah any objections to merging when all the expected tests are green?

mattpap:
I hoped using `io.open()` would shrink that code. Otherwise, it should be put in an utility function.

crwilcox:
@mattpap, I refactored this to use a utility function as you suggested. It removes a bit of duplication and should clean it up nicely.

bryevdv:
Hi @crwilcox we are in the home stretch, but there is a linter violation failing one of the tests:
```
E         File does not end with a newline: bokeh/sampledata/__init__.py, line 130
E       assert 1 == 0
E        +  where 1 = len(['File does not end with a newline: bokeh/sampledata/__init__.py, line 130'])
```
So basically just tap `<enter>` on the end of the last line there and save / push.

bokeh/sampledata/__init__.py
```diff
@@ -120,3 +121,10 @@ def _getfile(base_url, file_name, data_dir, progress=True):
             zip_file.extract(real_name, data_dir)
 
         remove(file_path)
+
+def _open_csv_file(filename):
+    # csv differs in Python 2.x and Python 3.x. Open the file differently in each.
+    if six.PY2:
+        return open(filename, 'rb')
+    else:
+        return open(filename, 'r', newline='', encoding='utf8')
```
mattpap: The error linter complains about is nicely visible at the end of this line.

mattpap: Though it depends on which view one uses, because it looks like it's only visible in "full changes" view.

bryevdv:
@crwilcox FYI if you don't have the bandwidth to make final changes for this PR, GH just announced this:

https://github.com/blog/2247-improving-collaboration-with-forks

Which would allow one of the core team to finish this up.

bryevdv:
@crwilcox just checking in, we are planning a freeze for `0.12.3` next Wednesday. Do you expect you can make that small change before then? Or allow one of the Maintainers to do it by following the instructions in the link above? If you do not have time I understand, in that case I will close this PR and we can try to make a new one by cherry picking your commits keeping author attribution at some future time.

crwilcox:
Sorry to not respond. I haven't been on Github in a while. Let me make the changes and I will reopen.

PR #5032, opened 2016-08-24
Canavandl/4687 hide notebook handle

### LINKED ISSUE #4697: Avoid showing <Bokeh Notebook handle for In[xx]> in Jupyter notebook
As much as I like to fill my notebooks with Bokeh plots, I find the return value of show (p) annoying and distracting, e.g. `<Bokeh Notebook handle for In[6]>` in 
http://nbviewer.jupyter.org/github/bokeh/bokeh-notebooks/blob/master/gallery/texas.ipynb

I personally do not use the handle that is returned, and so I'd be happier if `show()` just didn't return anything unless specifically requested in the call, as in `show(p,notebook_handle=True)`.  But if using the  handle is common for others, that might be tricky to change now.  

Maybe an option could be given to change this behavior at the start, as in `output_notebook(return_handles=False)`?

I'd even be ok with having `repr(handle)` or `str(handle)` (whichever one the notebook is showing) be "", but that's just because I never use the things. :-)

canavandl:

- [x] issues: fixes #4697
- [x] tests added / passed

### DIFF
--- a/bokeh/core/state.py
+++ b/bokeh/core/state.py
@@ -74,6 +74,8 @@ class State(object):
 
         notebook (bool) : whether to generate notebook output
 
+        return_handle (bool) : whether to return the notebook comms handle after `show`
+
         session_id (str) : a default session ID for Bokeh server output
 
         autoadd (bool) : whether certain functions automatically add roots to the document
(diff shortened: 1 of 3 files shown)


canavandl:
superceded by #5033

PR #5033, opened 2016-08-24
Add `request_handle` bool to `show` method to optionally return notebook comms handle

### LINKED ISSUE #4697: Avoid showing <Bokeh Notebook handle for In[xx]> in Jupyter notebook

canavandl:
supercedes #5032 

- [x] issues: fixes #4697 
- [x] tests added / passed

Adds `request_handle` bool to bokeh.io.show method, to optionally return notebook comms handle or not.

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -265,7 +265,7 @@ def curstate():
     '''
     return _state
 
-def show(obj, browser=None, new="tab"):
+def show(obj, browser=None, new="tab", return_handle=False):
     ''' Immediately display a plot object.
 
     In an IPython/Jupyter notebook, the output is displayed in an output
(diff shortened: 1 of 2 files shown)


bryevdv:
The notebooks in `examples/howto/notebook_comms` will need to be updated as well as one of the tutorial notebooks.

bryevdv:
rolling this into #5014 since its working on the same area

PR #5036, opened 2016-08-24
Canavandl/4936 lasso select bugfix

### LINKED ISSUE #4936: Lasso select is broken with non-circle markers
The following code:
```
import numpy as np

p = figure()

x = np.linspace(0,10, 1000)
y = np.sin(x) + np.random.random(len(x)) * 0.1
# Create a figure p with tools="lasso"
p = figure(tools="lasso_select, box_select")

# Add square renderer r at x and y with selection_color="red"
# and nonselection_alpha=0.2
p.triangle(x, y, selection_color="red", nonselection_alpha=0.1)

# This code is already written
output_file("hover_glyph.html")
show(p)
```
results in 

screen shot 2016-08-09 at 7 56 27 am

---

Works fine if a circle renderer is used instead. 

This may be a report of the same issue: https://groups.google.com/a/continuum.io/forum/?pli=1#!searchin/bokeh/super$20jumpy$20and$20barely$20responsive/bokeh/FptxSPlvBUE/NIGh3ZuTCAAJ

canavandl:
- [x] issues: fixes #4936 
- [x] tests added / passed

Todo:
* Make sure that changing canvas coordinate transforms to return new array instead of mutating array (that is also returned?) doesn't introduce new problems or performance issues?
  * Spot testing suggests the PR does neither, but I should be more complete
* Write some mocha tests for the hit testing and/or select tools. I suspect this regression could have been caught with some straightforward unit tests, but we didn't have them.

### DIFF
--- a/bokehjs/src/coffee/models/annotations/arrow.coffee
+++ b/bokehjs/src/coffee/models/annotations/arrow.coffee
@@ -33,8 +33,8 @@ class ArrowView extends Annotation.View
                                        y_name=@mget('y_range_name')
                                        )
     else
-      start = [@canvas.v_vx_to_sx(@_x_start.slice(0)),
-               @canvas.v_vy_to_sy(@_y_start.slice(0))]
+      start = [@canvas.v_vx_to_sx(@_x_start),
+               @canvas.v_vy_to_sy(@_y_start)]
 
     if @mget('end_units') == 'data'
       end = @plot_view.map_to_screen(@_x_end, @_y_end,
@@ -42,8 +42,8 @@ class ArrowView extends Annotation.View
                                      y_name=@mget('y_range_name')
                                      )
     else
-      end = [@canvas.v_vx_to_sx(@_x_end.slice(0)),
-             @canvas.v_vy_to_sy(@_y_end.slice(0))]
+      end = [@canvas.v_vx_to_sx(@_x_end),
+             @canvas.v_vy_to_sy(@_y_end)]
 
     return [start, end]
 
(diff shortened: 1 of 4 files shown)


canavandl:
visual example:

screen shot 2016-08-24 at 5 59 16 pm

bryevdv:
Adding `MIGRATION` I doubt anyone anywhere is using `v_vx_to_sx`, etc. directly but I'd like to develop a habit of fastidiousness around reporting any and all changes to actual behaviour.

canavandl:
I'm marking this as ready. I've done some tools refactoring and testing, but it should go into a new branch/PR.

bryevdv:
Confident about the in-place to copy change? I don't think these functions were used in to many places, but it's probably worth grep'ing and making a cursory inspection of each place.

canavandl:
I grepped around and didn't find any occurrences we're the code is purposely mutating an coordinate array (versus setting a new value). I did remove a couple of now unnecessary `_.clone` usages.

I also don't see any evidence of a decline in performance due to instantiating new arrays, based on the logging messaging of a plot with 3 glyphs of 1000 points:

```python
from bokeh.models import Range1d, LassoSelectTool, PanTool

import numpy as np

p = figure()

x = np.linspace(0,10, 1000)
y = np.sin(x) + np.random.random(len(x)) * 0.1
# Create a figure p with tools="lasso"
p = figure(tools=[LassoSelectTool(select_every_mousemove=True), PanTool()],
           x_range=Range1d(-1,11), y_range=Range1d(-1,11))

# Add square renderer r at x and y with selection_color="red"
# and nonselection_alpha=0.2
p.triangle(x, y, selection_color="red", nonselection_alpha=0.1)
p.scatter(x, y, selection_color="red", nonselection_alpha=0.1)
p.circle(x, y, selection_color="red", nonselection_alpha=0.1)

p.diamond(x, y, selection_color="red", nonselection_alpha=0.1)

# This code is already written
output_file("lasso_master.html")
# output_file("lasso_vectorized.html")
show(p)
```

bryevdv:
OK, LGTM

PR #5037, opened 2016-08-25
Exposing tooltips columns in Charts API

### LINKED ISSUE #4347: Hover in charts not displaying data
It seems there is an issue with plots from bokeh.charts not showing data in the tooltip--at least for the Scatter and TimeSeries plots that I tried.  It will display grouping fields from the color or marker, but not x, y, or any other field in the dataframe.  

While I did see a workaround from http://blog.rtwilson.com/bokeh-plots-with-dataframe-based-tooltips/, but that does not seem well-suited when color and marker groups are used.  

In the code below, if the color and marker arguments are removed, then the tooltip shows no data at all.  As written, both are displayed in the plot, so the displayed content seems to correspond to the PointGlyph label arguments in ScatterBuilder.

```
from bokeh.charts import Scatter, output_file, show

hover = HoverTool(
    tooltips = [
    ("cyl", "@cyl"),
    ("Weight", "@weight"),
    ("Acceleration", "@accel"),
    ('Horsepower', '@hp'),
    ('Origin', '@origin'),
])
tools = [hover]
scatter = Scatter(df, x='mpg', y='hp', color='cyl', marker='origin',
                  title="Auto MPG", xlabel="Miles Per Gallon",
                  ylabel="Horsepower", tools=tools)
output_file('test.html')
show(scatter)
```
screen shot 2016-05-20 at 21 06 35

grromrell:
issues: fixes #4347 

Charts have a kw argument for tooltips, but if the columns do not exist then the hover tool will display '???' This change will find columns that are specified by the tooltips kw argument and add them to the glyph data source if they do not already exist. Currently only the ScatterBuilder is affected. If the change is amenable as supplied I can work on adding the solution to the other Builders.

### DIFF
--- a/bokeh/charts/builders/bar_builder.py
+++ b/bokeh/charts/builders/bar_builder.py
@@ -28,6 +28,7 @@ from ..operations import Stack, Dodge
 from ...core.enums import Aggregation
 from ..stats import stats
 from ...models.sources import ColumnDataSource
+from ..utils import help
 
 # -----------------------------------------------------------------------------
 # Classes and functions
@@ -190,9 +191,7 @@ class BarBuilder(Builder):
         kwargs = self.get_extra_args()
         attrs = self.collect_attr_kwargs()
 
-        groups = []
         for group in self._data.groupby(**self.attributes):
-            groups.append(group)
             glyph_kwargs = self.get_group_kwargs(group, attrs)
             group_kwargs = kwargs.copy()
             group_kwargs.update(glyph_kwargs)
@@ -230,8 +229,8 @@ class BarBuilder(Builder):
         self.max_height = max([renderer.y_max for renderer in self.comp_glyphs])
         self.min_height = min([renderer.y_min for renderer in self.comp_glyphs])
 
-        for i in range(len(self.comp_glyphs)):
-            for sub_renderer in self.comp_glyphs[i].renderers:
+        for renderer in self.comp_glphs:
+            for sub_renderer in renderer.renderers:
                 yield sub_renderer
 
 
--- a/bokeh/charts/builders/bar_builder.py
+++ b/bokeh/charts/builders/bar_builder.py
@@ -234,12 +234,6 @@ class BarBuilder(Builder):
         
         for i in range(len(self.comp_glyphs)):
             for sub_renderer in self.comp_glyphs[i].renderers:
-
-                if self.tooltips:
-                    sub_renderer = add_tooltips_columns(sub_renderer,
-                                                        self.tooltips,
-                                                        groups[i])
-                
                 yield sub_renderer
 
 
--- a/bokeh/charts/builders/bar_builder.py
+++ b/bokeh/charts/builders/bar_builder.py
@@ -28,8 +28,6 @@ from ..operations import Stack, Dodge
 from ...core.enums import Aggregation
 from ..stats import stats
 from ...models.sources import ColumnDataSource
-from ..utils import help, add_tooltips_columns
-
 
 # -----------------------------------------------------------------------------
 # Classes and functions
@@ -191,7 +189,7 @@ class BarBuilder(Builder):
         """
         kwargs = self.get_extra_args()
         attrs = self.collect_attr_kwargs()
-    
+
         groups = []
         for group in self._data.groupby(**self.attributes):
             groups.append(group)
@@ -231,7 +229,7 @@ class BarBuilder(Builder):
         # a higher level function of bar chart is to keep track of max height of all bars
         self.max_height = max([renderer.y_max for renderer in self.comp_glyphs])
         self.min_height = min([renderer.y_min for renderer in self.comp_glyphs])
-        
+
         for i in range(len(self.comp_glyphs)):
             for sub_renderer in self.comp_glyphs[i].renderers:
                 yield sub_renderer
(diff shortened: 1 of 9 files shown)


grromrell:
bokeh_screenshot

birdsarah:
@grromrell - thanks so much for taking this on!

I think the approach you've taken is great. `tooltips` is a convenience for charts, I don't see a reason why that convenience shouldn't do more work. What do you think @bryevdv @fpliger?

There's a few things that are going to be needed to get this to merge though:

1. I don't think this logic can be on builder because some charts change the data significantly - to the extent that its a different shape and you won't be able to add more columns in. Having said that - I think we should add this to all the builders that its easy to.
2. There are a bunch of different ways of expressing the tooltips argument - does this handle all of them?
3. Adding some tests. (This is probably the easiest way to demonstrate qu. 2)
4. Add this feature into at least one charts example
5. Add some documentation

birdsarah:
Also, don't worry about the failing "integration" tests - that's a problem with our selenium tests that I need to fix.

grromrell:
1. I agree, scatterplot was the lowest hanging fruit, but bar, line etc. should be relatively easy as well.
2. The Builder class does a type check that only allows:
   a. List of Tuples of Strings
   b. List of Strings
   c. Boolean
   currently all three are covered, the only functionality that will break it currently is if the '$' operator is    used, which should be a simple fix.
3-5. I am not quite sure where you would like most of this to happen, but I am happy to help out here as well.

bryevdv:
Hi @grromrell thanks ver much for looking into this. Extending this to other charts would be extremely appreciated. 

For 1) I think @birdsarah is probably right about not applying directly to all charts types (I'd have to look close) If so, one approach to not repeat code would be to factor this a helper function that can go into `charts.utils` or wherever, then a method can be added to the builders that makes sense that just calls the help func and nothing else. 

For 3) our tests are usually inline, in a `tests` directory in the same location as the file in question

For 4 and 5: docs could include any or all of:  updating an example or two to uses this in `examples/charts/file`, updating the charts tutorial notebook in the `bokeh-notebooks` repo, adding a small section to the charts chapter of the user's guide in the Sphinx docs.

grromrell:
OK the latest commit should handle all versions of tooltips configurations. I have also added the hover functionality to all relevant charts (Bar, Dot, Line, Scatter, Step, Timeseries). There is one issue that is pretty baffling to me. If I implement the add_tooltips_columns function on the Step builder then half of the chart becomes transparent. I have commented it out for now as I cannot see any reason why this would be the case. Next will be adding tests and documentation.

Two additional notes:

grromrell:
step_screenshot

birdsarah:
>  I have also added the hover functionality to all relevant charts (Bar, Dot, Line, Scatter, Step, Timeseries)

I am wary about some of these additions. Bar, for example, I feel confident that the data shape will have changed between the raw data and the data that gets sent to the client. You cannot have a column data source with some columns of different length so we have to be careful we're not doing that.

> There is one issue that is pretty baffling to me. If I implement the add_tooltips_columns function on the Step builder then half of the chart becomes transparent

I believe this is because of the above issue. I'm guessing that step adds in a bunch of extra data points to make the steps. You can confirm this by adding some print statements to get the length of all the data columns in the ColumnDataSource.

grromrell:
Ok I see what you mean now. I have removed the bar and step builder. Dot, Line and Scatter do not aggregate data up from the row level so those are safe. Timeseries uses the dot and line builders so it doesn't need special handling. I will start working on docs and examples.

Rubyj:
@grromrell Were you able to implement a working solution for bar charts?

bryevdv:
@Rubyj I think the best solution for more sophisticated bar type charts is to use `bokeh.plotting` especially with some of the recent additions from @birdsarah 

@grromrell we are looking at at `0.12.3` release possibly the week after next. Just checking, do you think your docs/examples changes will be in before then?

bryevdv:
Also, as an idea, I wonder if it might not be better (possibly longer term) to have the `ToolTip` have a method that can return a list of all the columns that it will refer to. That could be used for validation at lower levels, and for `bokeh.charts`, simple charts like `Scatter` could take that list and then use it to copy columns into the JSON for the chart (as yo are doing now). But more complicated charts like `Bar` could take that list and potentially do similar aggregations that were done for the bars

bryevdv:
@bokeh/dev opinions on merging this for `0.12.3` without further docs examples?

grromrell:
@bryevdv Sorry for the ghost on this. I will update it today with the PEP fixes and some documentation.

bigreddot:
@grromrell no worries, I understand being busy :) Thanks again for the PR

grromrell:
@bryevdv @birdsarah Ok latest commit should be merge-able. I updated the sphinx 'Charts' documentation with a chart and updated the scatter_multi example to demonstrate the new functionality. The existing example for line_multi should demonstrate the tooltips idea well enough. I think I caught all of the PEP8 problems as well. I neglected to add a test, as I plead ignorance to exactly how that should be done after reviewing some of the tests in the Charts module.

birdsarah:
I'm just checking the docs, then planning to merge. 

Thanks for this @grromrell - it's a great addition.

birdsarah:
Added: #5206 to note need for tests

Docs header could have been out-dented one level but it's fine either way.

Merging.

PR #4933, opened 2016-08-08
Make set() a serializable data type

mattpap:
fixes #4932

Issue #4932, opened 2016-08-07
TypeError: set([...]) is not JSON serializable


hellpanderrr:
If I insert a set instead of a list into some plotting function, like `patches` or `line` I get the following error, and after that bokeh shows it even if I use a list and python interpreter has to be restarted to work again properly:

> TypeError                                 Traceback (most recent call last)
> <ipython-input-2-2f9c7f60120a> in <module>()
>       8 p = figure(title="example",webgl=True)
>       9 p.line([280.0, 290.0, 300.0],[280.0, 290.0, 300.0], alpha=0.2)
> ---> 10 show(p)
> 
> c:\Python27\lib\site-packages\bokeh\io.pyc in show(obj, browser, new)
>     299 
>     300     '''
> --> 301     return _show_with_state(obj, _state, browser, new)
>     302 
>     303 def _show_with_state(obj, state, browser, new):
> 
> c:\Python27\lib\site-packages\bokeh\io.pyc in _show_with_state(obj, state, browser, new)
>     307 
>     308     if state.notebook:
> --> 309         comms_handle = _show_notebook_with_state(obj, state)
>     310 
>     311     elif state.server_enabled:
> 
> c:\Python27\lib\site-packages\bokeh\io.pyc in _show_notebook_with_state(obj, state)
>     328     else:
>     329         comms_target = make_id()
> --> 330         publish_display_data({'text/html': notebook_div(obj, comms_target)})
>     331         handle = _CommsHandle(get_comms(comms_target), state.document, state.document.to_json())
>     332         state.last_comms_handle = handle
> 
> c:\Python27\lib\site-packages\bokeh\embed.pyc in notebook_div(model, notebook_comms_target)
>     247 
>     248     with _ModelInDocument(model):
> --> 249         (docs_json, render_items) = _standalone_docs_json_and_render_items([model])
>     250 
>     251     item = render_items[0]
> 
> c:\Python27\lib\site-packages\bokeh\embed.pyc in _standalone_docs_json_and_render_items(models)
>     564     docs_json = {}
>     565     for k, v in docs_by_id.items():
> --> 566         docs_json[k] = v.to_json()
>     567 
>     568     return (docs_json, render_items)
> 
> c:\Python27\lib\site-packages\bokeh\document.pyc in to_json(self)
>     730         # this is a total hack to go via a string, needed because
>     731         # our BokehJSONEncoder goes straight to a string.
> --> 732         doc_json = self.to_json_string()
[...]

PR #5043, opened 2016-08-29
Make ColumnDataSource.data stricter

### LINKED ISSUE #5042: Disallow set type in `ColumnDataSource.data`

mattpap:
This is a follow up on issue #4932 and PR #4933. In principle this disallows `set` type to be used in `ColumnDataSource.data`. If you wonder why I still left `Any` there, then I already changed that locally, but speed penalty is too great to include it in this PR. I need to optimize property validation first.

fixes #5042

### DIFF
--- a/bokeh/core/tests/test_properties.py
+++ b/bokeh/core/tests/test_properties.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 import datetime
 import unittest
 import numpy as np
+import pandas as pd
 from copy import copy
 
 from bokeh.core.properties import (
@@ -1063,6 +1064,10 @@ class TestProperties(unittest.TestCase):
         self.assertFalse(prop.is_valid({1, 2}))
         self.assertFalse(prop.is_valid({1: 2}))
         self.assertFalse(prop.is_valid(Foo()))
+
+        df = pd.DataFrame([1, 2])
+        self.assertTrue(prop.is_valid(df.index))
+        self.assertTrue(prop.is_valid(df.iloc[0]))
 
     def test_List(self):
         with self.assertRaises(TypeError):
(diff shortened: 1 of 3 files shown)


mattpap:
btw. If you grep for `Seq` in models, then you will see that this PR has more broader positive effect.

bryevdv:
LGTM shouldn't the original issue #4932 be sufficient to close, is a new issue needed?

mattpap:
That's a good question. The original issue asks for a feature or bug fix (presuming its existence), and this PR doesn't do this, but the opposite. That's why I created #5042. This way we can close that one as `resolved` based on this PR and the original as `invalid` or `referred` (I'm not sure).

bryevdv:
OK, that's fine too. I don't mind re-purposing issues when it's clear the originally stated ask is not quite right but making a new issue is fine too :)

bokeh/core/tests/test_properties.py
```diff
@@ -1037,6 +1037,33 @@ def test_Regex(self):
...
+        self.assertTrue(prop.is_valid(np.array([])))
+        self.assertFalse(prop.is_valid(set([])))
+        self.assertFalse(prop.is_valid({}))
+        self.assertTrue(prop.is_valid((1, 2)))
+        self.assertTrue(prop.is_valid([1, 2]))
+        self.assertTrue(prop.is_valid(np.array([1, 2])))
+        self.assertFalse(prop.is_valid({1, 2}))
+        self.assertFalse(prop.is_valid({1: 2}))
+        self.assertFalse(prop.is_valid(Foo()))
+
```
bryevdv: Are all the pandas column types `ndarray` subclasses? (asking, I'm not sure)

mattpap: No they aren't, though it doesn't matter here, because we don't check for any concrete types in `Seq.validate()`. However, I will add tests for pandas' types to make sure they can be handled by validation logic.

PR #5048, opened 2016-08-30
Make different column length error more vocal

### LINKED ISSUE #5000: Warn on ragged length values in ColumnDataSource
Bokeh assumes that the values of a ColumnDataSource are all of the same length.  However users are able to break this contract easily and without being informed.
### Example

``` python
intended_data = {'a': [], 'b': [], 'c': []}

source = ColumnDataSource(intended_data)

actual_data = {'a': [1, 2, 3], 'b': [4, 5, 6]}
source.data.update(actual_data)
```

This may silently result in a client-side plot receiving no data, because the length of the shortest value is zero.  

Instead, it might be nice to raise a warning or exception to inform the user that they have uneven rows.  This might happen on the `update` call, when the data is packaged to be sent up to the client, or on the client side.

mattpap:
This is a proof of concept for replacing validation (where possible) with early and hard failure.

fixes #5000


bokeh/models/tests/test_sources.py
```diff
@@ -148,5 +148,21 @@ def mock(*args, **kw):
...
+        with self.assertRaises(ValueError):
+            ds.data = dict(a=[10, 11], b=[20, 21, 22])
+
+        ds = ColumnDataSource(data=dict(a=[10, 11]))
+        with self.assertRaises(ValueError):
+            ds.data["b"] = [20, 21, 22]
+
+        ds = ColumnDataSource(data=dict(a=[10, 11], b=[20, 21]))
+        with self.assertRaises(ValueError):
+            ds.data.update(dict(a=[10, 11, 12]))
```
mattpap: Note that with the current implementation of property containers, `ds.data` is updated prior to assertions being check, so its value is still mutated, which can be confusing (e.g. in an interactive session). I think we can do something about this, but I'm not sure how expensive this will be.

mattpap:
@bryevdv, I fixed tests and made everything as vocal as possible without actually throwing exceptions.

bryevdv:
LGTM @mattpap @philippjfr do either of you have opinions about whether this or #5544 goes in first?

philippjfr:
I'd be happy for this to go in before, from what I've seen it shouldn't clash too much and tighter and earlier validation will only help.

bryevdv:
OK, I think I can get this merged today.

bryevdv:
@mattpap I think your `warnings.py` is shadowing the python built in (py 2):
```
message = 'bokeh.models.layout.VBoxForm was deprecated in Bokeh 0.12.0 and will be removed, use bokeh.models.layouts.WidgetBox instead.'
stacklevel = 2
    def warn(message, stacklevel=2):
>       warnings.warn(message, BokehDeprecationWarning, stacklevel=stacklevel)
E       AttributeError: 'module' object has no attribute 'warn'

```
I'd recommend renaming the module to `custom_warnings.py` or something just to be completely safe

mattpap:
> I'd recommend renaming the module to custom_warnings.py or something just to be completely safe

I added `from __future__ import absolute_import` to fix this. btw. We should be using this everywhere.

bryevdv:
> I added from __future__ import absolute_import to fix this. btw. We should be using this everywhere.
 
Agreed

bryevdv:
OK this we are till dealing with build two failing with a timeout even after reporting successful exit with code 0. But I checked the reports, and also locally checked the OHCL streaming example. I am going to go ahead and merge.

bryevdv:
@mattpap this is issuing JS console warnings, even when the column lengths are the same. Do you want me to revert this merge? Or do you want to fix in a new PR? For reference here is an example:
```
import bokeh
print(bokeh.__version__)

import numpy as np

from bokeh.palettes import Colorblind8 as palette


values = np.linspace(0.1, 0.9, 21)
print(values)

size = 250
semilogx = figure(title='semilogx test', x_axis_type='log', width=size, height=size)
semilogy = figure(title='semilogy test', y_axis_type='log', width=size, height=size)
loglog = figure(title='loglog test', x_axis_type='log', y_axis_type='log', width=size, height=size)

semilogx.line(values[values != 0.0], values[values != 0.0])
semilogy.line(values, values)
loglog.line(values, values)

fig = gridplot([[semilogx, semilogy, loglog]])
output_file("foo.html")
show(fig)
```
With CDS:
```
{
                        "attributes": {
                          "callback": null,
                          "column_names": [
                            "x",
                            "y"
                          ],
                          "data": {
                            "x": [
                              0.1,
                              0.14,
                              0.18,
                              0.22,
                              0.26,
                              0.30000000000000004,
                              0.33999999999999997,
                              0.38,
                              0.42000000000000004,
                              0.45999999999999996,
                              0.5,
                              0.54,
                              0.58,
                              0.62,
                              0.66,
                              0.7,
                              0.74,
                              0.78,
                              0.82,
                              0.86,
                              0.9
                            ],
                            "y": [
                              0.1,
                              0.14,
                              0.18,
                              0.22,
                              0.26,
                              0.30000000000000004,
                              0.33999999999999997,
                              0.38,
                              0.42000000000000004,
                              0.45999999999999996,
                              0.5,
                              0.54,
                              0.58,
                              0.62,
                              0.66,
                              0.7,
                              0.74,
                              0.78,
                              0.82,
                              0.86,
                              0.9
                            ]
                          }
                        },
```
Yet:

screen shot 2016-12-13 at 11 17 03 am

bryevdv:
This logic is wrong:
```

                switch (lengths.length) {
                    case 0:
                        return null;
                    case 1:
                        msg = "data source has columns of inconsistent lengths";
                        if (soft) {
                            logging_1.logger.warn(msg);
                            return lengths[0];
                        }
                        else {
                            throw new Error(msg);
                        }
                        break;
                    default:
                        return lengths[0];
                }
```
When `lengths==1` that is the correct expected case

bryevdv:
I will make a PR to fix with JS tests

PR #5053, opened 2016-08-30
Add WheelPan tool

### LINKED ISSUE #5052: Feature request: add wheel pan tool
Example Motivation:

When interacting with a time series plot, a user would like to pan through different time eras. This tool would be limited to a single dimension.

Why not WheelZoomTool?

I can imagine wanting to have a specific view window size (say 5 years) and want to explore my data with that specific aspect ratio.

Why not PanTool?

If the data range is much larger than the desired window (say there's 100 years of data but the visualization only "makes sense" when viewed in a 5 year view window, that'd be 20+ click-and-drag pans versus a smoother scrolling interaction.

canavandl:
prototype of WheelPan tool

- [x] issues: fixes #5052
- [ ] tests added / passed (will do)


canavandl:
ping @jbednar

canavandl:
demo code:

```python
from bokeh.models import WheelPanTool
import numpy as np

x = np.arange(0.0, 1.0 + 0.01, 0.01)
y = np.cos(2*2*np.pi*x) + 2

p = figure(title="WheelPanTool Demonstration", width=500, height=500, tools="pan, reset")
p.add_tools(WheelPanTool(dimension='height'))
p.add_tools(WheelPanTool(dimension='width'))

p.line(x, y)

show(p)
```

bokehjs/src/coffee/common/models.coffee
```diff
@@ -151,6 +151,7 @@ module.exports = {
   SelectTool:               require '../models/tools/gestures/select_tool'
   ResizeTool:               require '../models/tools/gestures/resize_tool'
   TapTool:                  require '../models/tools/gestures/tap_tool'
+  WheelPanTool:            require '../models/tools/gestures/wheel_pan_tool'
   WheelZoomTool:            require '../models/tools/gestures/wheel_zoom_tool'
```
jbednar: I'm not sure how whitespace-aware coffeescript is, but it still seems like one more space here would help. :-)

bokeh/models/tools.py
```diff
@@ -174,6 +174,23 @@ class PanTool(Drag):
...
+
+    .. |wheel_pan_icon| image:: /_images/icons/WheelPan.png
+        :height: 18pt
+
+    """
+
+    dimension = Enum(Dimension, default="width", help="""
+    Which dimension the wheel pan tool is constrained to act in. By
+    default the wheel pan tool will pan the plot along the x-axis.
+    """)
```
jbednar: Maybe ought to explicitly state that the y-axis can be selected using "height"?  It's obvious, but still, it's an Enum, and it would be nice to be clear what the enumerated options are...

bryevdv: Tho FYI all `Enum` automatically self-document at least the options:

screen shot 2016-08-31 at 3 36 18 pm

bokehjs/src/coffee/models/tools/gestures/wheel_pan_tool.coffee
```diff
@@ -0,0 +1,101 @@
...
+    }
+    @plot_view.push_state('wheel_pan', {range: zoom_info})
+    @plot_view.update_range(zoom_info, false, true)
+    @plot_view.interactive_timestamp = Date.now()
+    return null
+
+class WheelPanTool extends GestureTool.Model
+  type: 'WheelPanTool'
+  default_view: WheelPanToolView
+  tool_name: "Wheel Pan Zoom"
```
jbednar: Should be just "Wheel Pan", right?

jbednar:
Looks great!  It would be good to make the attribute for controlling the pan speed public, as I'm sure some people will want to change it from Python.

mattpap:
TS API update is missing.

jbednar:
Jean-Luc wonders if there would be a way for the WheelPan tool to have three states.  I.e., right now it has two (disabled and enabled), but if it could have three (disabled, enabled for x, enabled for y), presumably with a different icon each time, then the user could choose which direction to use in a fairly obvious way.  That does sound reasonable to me, but I wouldn't want to hold things up while support for multiple modes and multiple icons for a tool is added!

jbednar:
I'm personally interested in this tool for panning vertically in a special case, but I should mention that I think this will also be very useful for people looking at long time series, e.g. stock prices over the course of a day...

bryevdv:
> Jean-Luc wonders if there would be a way for the WheelPan tool to have three states. I.e., right now it has two (disabled and enabled), but if it could have three (disabled, enabled for x, enabled for y), 

Doubtful. `enabled/disabled` is baked into a parent class, adding this kind of interface would clash with that. 

But, you can set the `tooltip` property programmatically: 

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/tools/gestures/wheel_zoom_tool.coffee#L108-L113

so that the hover over the icon to enable/disable shows something different based on some configuration.

jbednar:
Luke's example already uses tooltips in that way:

image

image

But it's still confusing to have two seemingly indentical tools in the same toolbar...

bryevdv:
Sorry I missed that it already did that. Agreed, but it's probably the best immediate thing that can be done. Letting tools have varying icons would be a chunk of work.

jbednar:
Sure.  So let's definitely not hold up this PR hoping for that. :-)

jbednar:
Those look like good changes, thanks.

PR #5055, opened 2016-08-31
Build hit testing assets in _map_data not _render

### LINKED ISSUE #4967: Consider computing sxss and syss on _map_data
Following on from discussion on #4953 

We compute sxss and syss on every render. Could this just be done at _map_data stage.

birdsarah:
- [x] issues: fixes #4967 
- [ ] tests added / passed

This fixes #4967, but has no effect, as _map_data is called on every render. See #4967 for more info.


bryevdv:
There is no point in doing this in isolation. Until the "over-render" issue is addressed directly, it doesn't matter whether the index is built in `_map_data` or `_render`.

PR #5067, opened 2016-08-31
Canavandl/2136 linked selections

### LINKED ISSUE #2136: Selections only work on the last of two series in one plot

screenshot

screenshot
### Code

Here is the example code:

```
import bokeh.plotting as bk
                     "number": [0.5, 1, 1.5],
                     "number2": [1.5, 2, 2.5]}
bk.show(p)
```

Bokeh 0.8.2
Python 3.4.3

canavandl:
Fixed linked selections

- [x] issues: fixes #2136 
- [x] tests added / passed

todo: 
- [x] make sure `inspected` changes haven't broken anything.
- [x] write a test about the selection clearing
- [ ] ~~there's some weirdness w/ the PolySelectTool regarding the overlay and making the dbl-click selection. May be in a follow up PR (it's present in master)~~ will do in follow up PR


canavandl:
Linked selection was previously broken because a select tool would iterate over it's glyph_renderers and overwrite a ColumnDataSource's `selected` attr (the hittest dict) each time, so that the last iteration would be the result.

This change causes `SelectionManager.select` to mutate the `selected` dict instead of overwriting it. This pushes a tiny bit more code onto the tool (deciding if/when to explicitly clear selections), but it seems pretty natural.

While this PR resolves #2136 , it suggests that a better path forward would be to deprecate SelectionManagers entirely and replace the ColumnDataSource `selected` and `inspected` attrs (now hittest dicts) with `Selector` objects that manage their own updates/clears. Doing so shouldn't be hard, as SelectionManagers are just setting props and triggering events.

canavandl:
superceded by #5158

PR #5078, opened 2016-09-02
Only run integration tests for non external contributors

### LINKED ISSUE #5077: Disable integration tests for external contributors
In fix for #3927 I created a partial solution where screenshot tests wouldn't run for external contributors as SauceLabs is a requirement, but other selenium tests would run locally using firefox on xvfb. However, some of these tests are now failing - presumably because of xvfb restrictions, so just disable integration tests for external contributors.

birdsarah:
Fixes #5077 

Assuming this passes, can merge and then can test with bgyarfas' PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,13 +52,9 @@ before_install:
     - export PATH="$HOME/build/bokeh/bokeh/bokehjs/node_modules/phantomjs-prebuilt/bin:$PATH"
     - echo $TRAVIS_BUILD_NUMBER > __travis_build_number__.txt
     - export TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
+    # We only want to use saucelabs when it's not an external contributor.
     - if [[ -z "${SAUCE_USERNAME}" ]] || [[ -z "${SAUCE_ACCESS_KEY}" ]]; then echo "External committer, not using saucelabs"; else export SAUCELABS=True; fi
-    # We don't use the travis saucelabs add on because it installs on every test run, but we can manually use travis' saucelabs start/stop commands
-    # Furthermore, we only want to use it when it's not an external contributor.
     - if [[ -n "${SAUCELABS}" ]]; then source <(curl -s https://raw.githubusercontent.com/travis-ci/travis-build/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh); fi
-    - if [[ -z "${SAUCELABS}" ]]; then export DISPLAY=:99.0; fi
-    - if [[ -z "${SAUCELABS}" ]]; then sh -e /etc/init.d/xvfb start; fi
-
 
 install:
     - scripts/travis_install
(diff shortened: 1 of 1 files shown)


birdsarah:
Examples test fail isn't related to this PR. Merging.

timsnyder:
@birdsarah and @bryevdv, is there an advantage to continuing to use the CI Sauce account over moving to using [Open Sauce]( https://saucelabs.com/open-source) for bokeh?

Are there specific tests that are known to be more flaky than others with xvfb?  Have you guys looked into higher-level selenium testing with something like [bok-choy](http://bok-choy.readthedocs.io/en/latest/guidelines.html#)?  I find statements like the following from their docs to be especially appealing:

> Selenium calls are notoriously unreliable. They provide no retry logic to protect you from timing issues, which can cause intermittent test failures. In contrast, bok-choy‘s higher-level interface for browser interactions include robust error-checking and retry logic.

birdsarah:
@timsnyder - I'm confused - I thought we did use the open source free account from SauceLabs. It's just that non-members of the bokeh team can't access the API keys that are secrets on Travis: https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions. Can you clarify?

I don't think firefox + xvfb just can't do all the things we want to do - like: cross browser testing; testing on a real browser; testing on multiple os'. 

In addition, we did just use firefox and xvfb to run our selenium tests until we started to do screenshot testing. You want to be able to generate a reference image that can be checked into a repo (although bryan rightly wants to move that out of checking it into main bokeh repo) that developers can work with. But screenshot tests are really finnicky - you bascially have to be pixel perfect - you need to use the same browser, OS, everything. Achieving this seemed easier with saucelabs and lets devs develop new screenshot tests locally.....but I'm definitely open to other suggestions on all of this. We've subsequently added, I believe, one cross browser test - because cross-browser tests are finicky ^ n-browsers and so should be used as sparingly as possible.

I haven't used bok-choy. But a quick glance at the docs and:
1) it's encouraging using the PageObject model for your selenium tests, which is a great design pattern for selenium tests - I've use it before. The way its normally used and described isn't a good match for Bokeh. However, now we're starting to get a modest collection of Selenium tests under our belt we may want to refactor into a "viz object" model of some kind.
2) it uses needle for its diff testing which is nose (not pytest based) and i didn't think was maintained any more, but I could be wrong
3) it's other features feel less bokeh-relevant - but I'm happy to hear otherwise

Overall, given how notoriously flaky selenium tests are, I'm pretty happy with our integration test suite. I restart examples and our flaky js unit tests all the time, but rarely the integration/selenium tests. And they've picked up a few regressions so far. There are many ways they could be improved - there's an issue somewhere that has started to document these issues.

birdsarah:
To clarify I believe it was the new layout resizing tests that were breaking on xvfb.

timsnyder:
@birdsarah, thanks for the extra background info on Sauce and Selenium and taking a look at bok-choy.  I should have started my questions with "Is there a reason that external contribs can't run Sauce tests?".  The link to the travis docs makes complete sense.  I made a poor assumption rather than just asking a question.  Nevertheless, I'm glad to know all of this background now.

I can definitely understand how firefox+xvfb doesn't stack up against Sauce in terms of all the things you did a great job of mentioning.  I also agree that the integration tests are really useful.  I slogged through getting them working locally on my linux setup (without virtual X11 for now) and I found some mistakes in my early work on #4841 with the integration tests.  It seems like if I wanted to enable myself to run Sauce tests, I should be able to setup my own Travis and OpenSauce accounts.  Unless you guys already know that is a dead end, I'll spend some time looking into it after I'm done with #4841.  Alternatively, I do need to setup some selenium+xvfb for a closed source thing I'm also working on and if you feel that having better docs and support for external contribs going that route, I'd be willing to dig into those issues some more.  

On a side note: I recently discovered [Cloud9](http://c9.io) and I'm playing around with getting a bokeh dev environment setup there.  So far, the full env like I have setup on my laptop is too big to fit in a free workspace but it is close enough that I might be able to come up with a workflow that could start out with in free workspaces.  If I kept it tested and up to date with bokeh/bokeh master, it might be a nice way for new contributors to get bootstrapped and into the code quickly.  Then, they could decide if they want to invest the time to get their dev environment running on a local workstation or pay for a Cloud9 subscription for bigger VM through them or via SSH from another cloud.  I may also find that I dislike this idea, I only started looking into it this weekend.

As you can probably tell, I'm new to Selenium testing.  It's something I've wanted to learn for a long time and again, I sincerely appreciate your thoughtful response to my clumsy and confusing questions. :)

Issue #4153, opened 2016-04-10
Strict trigger() and listenTo()


mattpap:
Make `trigger()` and `listenTo()` verify if properties used are actually defined on the target object's model. This is continuation of #4141.

mattpap:
I think that it's infeasible to fix this as-is. We won't get proper level safety anyway and investing time into this weak design is not worth the time. I would prefer to implement a proper signal, slot framework from ground up in TypeScript.

bryevdv:
@mattpap so this can be closed?

mattpap:
I will close this when new signaling API is on board, which I'm working currently on.

PR #5079, opened 2016-09-02
Use custom backbone.js

mattpap:
This PR removes all unneeded code from backbone.js, reducing it's line count by over 60% and shaving off 16k from bokeh.min.js. No changes to logic were done at this point. The main point, however, is to reduce cognitive burden.  This PR is an effect of lack of results in fixing issue #4153 (strict events).


bryevdv:
This is a happy PR

bryevdv:
@mattpap can BB also be removed from [`package.json`](https://github.com/bokeh/bokeh/blob/master/bokehjs/package.json) now as well?

bokehjs/package.json
```diff
@@ -86,7 +86,6 @@
   },
   "dependencies": {
     "underscore": "^1.5.2",
-    "backbone": "^1.1.0",
```
mattpap: @bryevdv, it was removed here.

bryevdv: sorry for the noise I tried searching this page for the wrong thing

Issue #4715, opened 2016-06-30
Focus: GIS Support 


bryevdv:
@brendancol I leave it to you to create a first cut for issues for this epic.

birdsarah:
https://github.com/bokeh/bokeh/wiki/Geo-Ramblings

birdsarah:
@brendancol and I had a meeting and came up with the following plan:

BokehJS is getting bloated and as part of an effort to streamline the core library and have good geo support, we propose splitting out a bokeh-geo library/extension/plugin that will live under the bokeh organization.

The aim of bokeh-geo will initially be to support a core set of basic mapping functionality - we have taken the feature set of LeafletJS as inspiration for what that base set of functionality is.

We propose to add features into the core library that are necessary to support bokeh-geo but also have wider non-geo utility e.g. for datashader or other use cases.

The concrete next things we have identified are:

* Add the following to the core library:
 * some kind of Range2d object that allows
  * the ability to express a bounding box
  * the ability to have a fixed "range-ratio" (like an aspect ratio but for ranges)
 * the ability to express fixed range intervals - at the moment we have a min range interval to prevent zooming in too far, but we need to be able express a series of fixed ranges
 * wrap around axes so you can have multiple worlds (hard)
 * computed tick labels (so you can translate between, for example, meters and lat-lon)

* Take the following items out of the core library, and put in bokeh-geo:
 * GeoJSON datasource
 * Mercator tile source and its children (but not tile sources in general)
 * The built-in tile providers
 * The proj4j dependency (may not go into bokeh-geo - see below)
 * GMapPlot

* Add the following items to bokeh-geo
 * GeoJSON glyphs
  * Glyphs named as per GeoJSON specs - Point, MultiPoint, LineString, MultiLineString, Polygon, MultiPolygon
  * They will accept data in the GeoJSON specification - Lists of co-ordinate tuples e.g. [(1, 2), (2, 3)] not xs and ys separated out
 * A few common transforms to automatically convert data into web mercator - this is far less than proj4 supports and so we may not want to include the library to keep bokeh-geo slimmer for now
 * Add a Map subclass of Plot which has an additional coordinate system property (not necessary immediately)

birdsarah:
Also see: #3148

bryevdv:
Moved to a GH Projects

PR #5080, opened 2016-09-02
Use custom proj4 to reduce bokeh.min.js' size

mattpap:
This shaves off about 55 kB from bokeh.min.js. This PR removes unnecessary imports and projections from `proj4/index.js`, so this isn't the same "custom" approach as e.g. with backbone.js. Just make sure to use `common/proj4` instead of `proj4` in new code.

### DIFF
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -92,7 +92,6 @@
     "jquery-mousewheel": "^3.1.12",
     "jquery-ui": "~1.10.5",
     "jsnlog": "^2.7.5",
-    "mathutils": "0.0.1",
     "rbush": "^2.0.1",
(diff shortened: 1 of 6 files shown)


mattpap:
@brendancol, please review this, especially the selection of projections. I'm not sure if we need anything besides the defaults, i.e. `merc` and `longlat`.

birdsarah:
Thanks for this @mattpap. When we break out bokeh geo (https://github.com/bokeh/bokeh/issues/4715) then proj4j will go entirely. (I guess when is the key word in that sentence :D)

bryevdv:
LGTM ping @brendancol any comments?

PR #5082, opened 2016-09-02
Don't use Any in ColumnDataSource.data

mattpap:
This is nice to have, but it comes with a price: speed degradation. The original type, `Seq(Any)` is pretty much instantaneous when it comes to validation. Here I introduce a special type `Scalar` which encompasses primitive types and date/datetime/delta. This way, `Seq(Scalar) | Seq(Seq(Scalar))` (this is the most precise type) is almost as fast as `Seq(T) | Seq(Seq(T))` where `T` is a primitive type like `Int`. This is about 6 time slower than the original type (when validating all items, i.e. type is correct), when validating a 100k element list. This is in turn about 5~6 times faster than naive `Seq(Either(long_list_here)) | Seq(Seq(Either(long_list_here)))`. Validating 100k element list takes about 200 ms on my machine.

### DIFF
--- a/bokeh/charts/models.py
+++ b/bokeh/charts/models.py
@@ -143,7 +143,8 @@ class CompositeGlyph(HasProps):
             n_rows = len(list(data.values())[0])
 
             # add composite chart index as column
-            data['chart_index'] = [self.label] * n_rows
+            # TODO: data['chart_index'] = [self.label] * n_rows
+            data['chart_index'] = [str(self.label)] * n_rows
 
             # add constant value for each column in chart index
             if isinstance(self.label, dict):
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -947,6 +947,11 @@ class String(PrimitiveProperty):
     """ String type property. """
     _underlying_type = string_types
 
+class Scalar(PrimitiveProperty):
+
+    _underlying_type = bool_types + integer_types + float_types + complex_types + \
+        string_types + date_types + datetime_types + timedelta_types
+
 class Regex(String):
     """ Regex type property validates that text values match the
     given regular expression.
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1051,15 +1051,12 @@ class Seq(ContainerProperty):
         super(Seq, self).validate(value)
 
         if value is not None:
-            if not (self._is_seq(value) and all(self.item_type.is_valid(item) for item in value)):
-                if self._is_seq(value):
-                    invalid = []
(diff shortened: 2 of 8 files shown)


bokeh/charts/models.py
```diff
@@ -143,7 +143,8 @@ def add_chart_index(self, data):
             n_rows = len(list(data.values())[0])
 
             # add composite chart index as column
-            data['chart_index'] = [self.label] * n_rows
+            # TODO: data['chart_index'] = [self.label] * n_rows
+            data['chart_index'] = [str(self.label)] * n_rows
```
mattpap: @fpliger, following up on the E-mail, there are some changes I made, but I doubt they make sense.

bokeh/core/properties.py
```diff
@@ -1053,15 +1051,12 @@ def validate(self, value):
...
-                    for item in value:
-                        if not self.item_type.is_valid(item):
-                            invalid.append(item)
-                    raise ValueError("expected an element of %s, got seq with invalid items %r" % (self, invalid))
-                else:
-                    raise ValueError("expected an element of %s, got %r" % (self, value))
+            if not self._is_seq(value):
+                raise ValueError("expected an element of %s, got %r" % (self, value))
+            elif any(not self.item_type.is_valid(item) for item in value):
+                #invalid = [ item for item in value if self.item_type.is_valid(item) ]
```
mattpap: This is extremely inefficient with the current implementation of validation. This can be restored, but it has to be computed lazily, only when an exception message is actually needed.

bokeh/core/properties.py
```diff
@@ -1543,33 +1541,16 @@ class Datetime(PropertyDescriptor):
...
-            datetime_types += (np.datetime64,)
-        except (ImportError, AttributeError) as e:
-            if e.args == ("'module' object has no attribute 'datetime64'",):
-                import sys
-                if 'PyPy' in sys.version:
-                    pass
-                else:
-                    raise e
-            else:
-                pass
```
mattpap: Someone has to explain this code to me. I'm not sure why we have to make PyPy involved in this.

mattpap:
Realistically speaking, there is little chance this will be efficient enough to be usable. Validation is slow enough already and this doesn't improve the situation.

bryevdv:
@mattpap if you think this is unfeasible perhaps the PR should just be closed as rejected rather than paused?

mattpap:
@bryevdv, yeah, I think I will reject it, though I will submit another PR with `bokeh.core.types`, as I think it's valuable change on its own.

Issue #3856, opened 2016-02-10
Populate legend with rows of data


birdsarah:
This feature idea came up from the discussion on #3843.

Also see notebook: http://nbviewer.jupyter.org/github/birdsarah/bokeh-miscellany/blob/master/3843.ipynb

Normally a legend is populated with a series of GlyphRenderers and their labels.

But if, as in the case of a Donut plot, there is one GlyphRenderer for each of the slices of the pie and the color is controlled by a column; you may well want a legend that is populated by the rows of a data source so that you can document the changing color (which is powered by a column of the ColumnDataSource).

I think this will also relate to making a client side colormappers

For example, in the gapminder example: http://nbviewer.jupyter.org/github/bokeh/bokeh-notebooks/blob/master/tutorial/A1%20-%20Building%20gapminder.ipynb#Embedding I manually build a legend for each of the world's regions. This could be thought of as making a legend based on the data rows, but only displaying unique values.

bryevdv:
I was just thinking this on the way home, thanks for making an issue I think it is a great idea. More generally, we have a notion of a "representative" point but I think that should be expanded to representative pointS, and then also make the label configurable as a column field as well. I think that would cover this case very well. (the reason to make the points configurable, suppose you have some "grouped" data in one source, and want a legend only for each glyph, not for every row -- you could configure the representative points to be one from each group)

fedorova:
Do you have an example of how to use this feature?

bryevdv:
Second example here:

https://bokeh.pydata.org/en/latest/docs/user_guide/annotations.html#legends

PR #5087, opened 2016-09-03
Clarify draw_legend on glyph vs on glyph_renderer

birdsarah:
Previously both glyph_renderer and glyph had the draw_legend function.
The glyph functions would "look up" a reference point - except that this
always returns none. Looking up a reference point is probably best done on
the glyph_renderer where actual data is stored.

This commit renames glyph.draw_legend to glyph.draw_legend_for_index and moves
the responsibility for looking up an index up to the glyph renderer.

In a future PR we will use this capability to allow drawing multiple legend entries for the
same glyph_renderer. For now, this should be a no-op change.

Part of, but does not fix, #3856.


bryevdv:
will merge when green

bryevdv:
@mattpap when I try this branch with `BOKEH_RESOURCES=inline` to test, every example in `examples/howto/layout` errors with the new top error box, e.g.

> Module `Div' does not exists. The problem may be two fold. Either a model was requested that's available in an extra bundle, e.g. a widget, or a custom model was requested, but it wasn't registered before first usage.

Those are just what I tried first, I stopped testing after they failed.

PR #5090, opened 2016-09-03
Add categorical color mapper

### LINKED ISSUE #5013: Discrete/Categorical ColorMapper and ColorBar
Add a CategoricalColorMapper that accepts palette and factors

```python
color_mapper = CategoricalColorMapper(
   palette=Spectral3,
   factors=["cat", "dog", "milk"],
)
```

@canavandl - we'd want color bar to use a CategoricalAxis and I'm not sure we could use image - maybe we could though. It's very similar to low-n palettes.

@bryevdv - `nan_color`, `high_color`, `low_color`. high/low don't make sense but out_of_range does. Which I see as different from No Data. Thoughts on naming.

birdsarah:
Fixes #5013

### DIFF
--- a/bokeh/models/mappers.py
+++ b/bokeh/models/mappers.py
@@ -12,7 +12,7 @@ from .. import palettes
 
 @abstract
 class ColorMapper(Model):
-    """ Base class for color mapper types. `ColorMapper`` is not
+    """ Base class for color mapper types. ``ColorMapper`` is not
     generally useful to instantiate on its own.
 
     """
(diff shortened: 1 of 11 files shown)


bokeh/models/mappers.py
```diff
@@ -80,3 +92,13 @@ class LogColorMapper(ColorMapper):
...
+
+    low = Float(help="""
+    The minimum value of the range to map into the palette. Values below
+    this are clamped to ``low``.
+    """)
+
+    high = Float(help="""
+    The maximum value of the range to map into the palette. Values above
+    this are clamped to ``high``.
+    """)
```
bryevdv: We should probably factor out a `ContinuousColorMapper` for linear and log mappers to inherit from.

canavandl:
The categorical color mapper won't work with a ColorBar (which is fine, a `Legend` is probably a better annotation here anyway), because it lacks high/low attributes.

We should probably add a `ContinuousColorMapper` base class that ColorBar accepts (as per @bryevdv) to limit confusion.

birdsarah:
Split out the ContinuousColorMapper.

bokeh/models/mappers.py
```diff
@@ -24,6 +24,35 @@ class ColorMapper(Model):
...
+        super(ColorMapper, self).__init__(**kwargs)
+
+
+class CategoricalColorMapper(ColorMapper):
+    """ Map categories to colors. Values that are passed to
+    this mapper that aren't in factors will be assigned the nan_color.
+
+    """
+
+    factors = Either(List(String), List(Int), List(Float), List(Datetime), List(Date), help="""
```
bryevdv: `Seq` ? Is there a reason it has to be a list and not a tuple or an array?

birdsarah: Doh! Thanks.

bryevdv:
LGTM I think we should update all examples and notebooks to use this, but I'll make a separate issue for that. Maybe we can also add some slight convenience syntax around these things at that time too.

bokeh/models/mappers.py
bryevdv: Crap, I was in a hurry and forgot about this before merging. I'll revert it.

PR #5099, opened 2016-09-06
Remove computed properties

mattpap:
This attempts at removing computed properties. At this point all computed properties with no dependencies or dependencies on `this` were removed (replaced with getters). The remainder (5 instances) are properties dependent on some other instance (e.g. tools depending on a plot). Those need additional care, because changes to a dependency propagate through change events. I'm not sure if anyone depends on this behavior, but at least no one listens on specific properties. Tests are broken and need considerable amount of work to be fixed (see bokeh-internal).

PR #5106, opened 2016-09-07
Replace mget/mset/get/set with getters and setters

mattpap:
This supersedes PR #5099 and goes all the way to make bokehjs use getters and setters. The issues with the original PR are still present, i.e. some of the tests need major rework and some computed properties are still there. Some of `set()` calls are still there, because they use `silent=true`. Note that we use bare `Backbone.Model` in a few cases, so a few extra instances of `get()` and `set()` are still there. This PR must be reviewed and merged today.


bokehjs/test/common/selection_manager.coffee
```diff
@@ -14,6 +14,7 @@ hittest = utils.require "common/hittest"
 empty_selection = hittest.create_hit_test_result()
 
 describe "SelectionManager", ->
+  return
```
mattpap: Those tests were added by an external contributor, so someone has to volunteer to fix this.

bokehjs/test/models/annotations/color_bar.coffee
```diff
@@ -25,6 +25,7 @@ ColorBar = proxyquire('../../../src/coffee/models/annotations/color_bar',
                       {"../../core/util/text": textStub})
 
 describe "ColorBar module", ->
+  return
```
mattpap: ping @canavandl

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -711,15 +711,19 @@ class PlotCanvas extends LayoutDOM.Model
...
-      layout_renderers = @plot.get(side)
+    collect_panels = (layout_renderers) ->
       for r in layout_renderers
         if r.panel?
           children.push(r.panel)
+
+    collect_panels(@plot.above)
+    collect_panels(@plot.below)
+    collect_panels(@plot.left)
+    collect_panels(@plot.right)
```
mattpap: @birdsarah, please review this change.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -777,30 +781,27 @@ class PlotCanvas extends LayoutDOM.Model
...
-          constraints.push(EQ(last.panel._left, [-1, @left_panel._left]))
-        if side == "right"
-          constraints.push(EQ(last.panel._right, [-1, @right_panel._right]))
+        constraint = switch side
+          when "above" then EQ(last.panel._top, [-1, @above_panel._top])
+          when "below" then EQ(last.panel._bottom, [-1, @below_panel._bottom])
+          when "left"  then EQ(last.panel._left, [-1, @left_panel._left])
+          when "right" then EQ(last.panel._right, [-1, @right_panel._right])
+        constraints.push(constraint)
     return constraints
```
mattpap: @birdsarah, and this one.

bokehjs/src/coffee/models/tools/gestures/pan_tool.coffee
```diff
@@ -39,44 +39,44 @@ class PanToolView extends GestureTool.View
     new_dx = dx - @last_dx
     new_dy = dy - @last_dy
 
-    hr = _.clone(frame.get('h_range'))
```
mattpap: @bryevdv, was this ever necessary to clone this? Currently it's not possible to clone using `_.clone()`. I'm not sure if it was ever working as expected. Currently the safest is to use:
```
frame.h_range.constructor(frame.h_range.attributes)
```

bryevdv: It doesn't look like `hr` is mutated anywhere, so if it was ever needed in the first place, perhaps it is not needed now

mattpap:
@bryevdv, merged master and skipped broken tests.

bokehjs/src/coffee/models/tools/gestures/pan_tool.coffee
mattpap: Apparently there is Backbone.Model's method `clone()` which does exactly what should be done, so `obj.clone()` does the job.

birdsarah:
Both those changes look good to me @mattpap.

This PR needs an issue.

bokehjs/test/models/annotations/color_bar.coffee
canavandl: There's actually a simple fix for this:

```
diff --git a/bokehjs/test/models/annotations/color_bar.coffee b/bokehjs/test/models/annotations/color_bar.coffee
index 60aa69a..73cacad 100644
--- a/bokehjs/test/models/annotations/color_bar.coffee
+++ b/bokehjs/test/models/annotations/color_bar.coffee
@@ -47,9 +47,8 @@ describe "ColorBar module", ->

     beforeEach ->
       # Stub solver computed values with deterministic frame height and width
-      frame_stub = sinon.stub(@plot.plot_canvas.frame, 'get')
-      frame_stub.withArgs('height').returns(500)
-      frame_stub.withArgs('width').returns(500)
+      sinon.stub(@plot.plot_canvas.frame, 'height').returns(500)
+      sinon.stub(@plot.plot_canvas.frame, 'width').returns(500)

     describe "ColorBar.Model._title_extent method", ->
```

Do you want to include or should I follow up in another PR?

bryevdv:
@mattpap All the `examples/howto/layout` examples are still failing for me, it seems like it is trying to load the widgets JS first, and then failing during `require ../../core/properties`. Examples work with `0.12.2`

bryevdv:
same problem with `BOKEH_RESOURCES=server` and 
```
<script type="text/javascript" src="http://localhost:5006/static/js/bokeh.js"></script>
<script type="text/javascript" src="http://localhost:5006/static/js/bokeh-widgets.js"></script>
```
in the page source, so maybe it's a require path that got mucked up instead of the ordering

bryevdv:
This seems to be the `require` that is failing:
```
},{}],"models/widgets/cell_editors":[function(require,module,exports){
var $, $1, $2, CellEditor, CellEditorView, CheckboxEditor, CheckboxEditorView, DateEditor, DateEditorView, IntEditor, IntEditorView, Model, NumberEditor, NumberEditorView, PercentEditor, PercentEditorView, SelectEditor, SelectEditorView, StringEditor, StringEditorView, TextEditor, TextEditorView, TimeEditor, TimeEditorView, Widget, _, p,
  extend = function(child, parent) { for (var key in parent) { if (hasProp.call(parent, key)) child[key] = parent[key]; } function ctor() { this.constructor = child; } ctor.prototype = parent.prototype; child.prototype = new ctor(); child.__super__ = parent.prototype; return child; },
  hasProp = {}.hasOwnProperty;

_ = require("underscore");

$ = require("jquery");

$1 = require("jquery-ui/autocomplete");

$2 = require("jquery-ui/spinner");

p = require("../../core/properties");
```

mattpap:
@bryevdv, apparently it's broken in `master` as well. Widgets bundle is broken, because all core module references are `undefined`. btw. Our tests shouldn't be all green in such situation. Where are all JS errors going? I can't find anything in logs.

mattpap:
Trivial yet critical mistake I made in PR #4946 made all widget code fail (and not only, because plugin bundles were incorrectly resolving dependencies), yet tests were all green all the way. Why JavaScript errors are not taken into account when evaluating examples?

bryevdv:
@mattpap we obviously have some gaps in our JS testing at the moment. I would say this would be a reasonable task for something to look into.

PR #5111, opened 2016-09-07
Add CategoricalColorMapper

birdsarah:
Fixes: #5013

See #5090 for discussion.


birdsarah:
@bryevdv - merge when green?

bryevdv:
There's a real unit test failure (defaults mismatch)

birdsarah:
Why didn't it fail before?

birdsarah:
I've fixed the defaults mismatch. But it seems very odd to me that it didn't fail before but the mismatch still existed (default on bokehjs side, but not on python side).

bryevdv:
Yah it's weird, the JS tests really need a pass over in a few different ways.

bokehjs/src/coffee/models/mappers/categorical_color_mapper.coffee
```diff
@@ -0,0 +1,27 @@
...
+
+  _get_values: (data, palette) ->
+    values = []
+
+    for d in data
+      key = @factors.indexOf(d)
+      if key < 0
+        color = @nan_color
+      else
+        color = palette[key]
```
bryevdv: What if the palette is too small?

birdsarah: It errors miserably :)

bokeh/models/mappers.py
```diff
@@ -24,6 +25,45 @@ class ColorMapper(Model):
...
+
+    def __init__(self, **kwargs):
+        super(ColorMapper, self).__init__(**kwargs)
+        palette = self.palette
+        factors = self.factors
+        if palette and factors:
+            if len(palette) != len(factors):
+                extra_factors = factors[len(palette):]
+                warnings.warn("""Palette length does not match number of
+factors. %s will be assigned to `nan_color` %s""" % (extra_factors, self.nan_color))
```
bryevdv: This seems like an odd message if the palette is too long. `[] will be assigned to 'nan_color'`

birdsarah: Agreed, late night coding sigh.

PR #5114, opened 2016-09-08
Add legend glyph for vbar and hbar

### LINKED ISSUE #5113: VBar / HBar legend missing glyphs

birdsarah:
Fixes #5113 

Adds an example of using hbar and vbar for a stacked and a grouped plot too.

screen shot 2016-09-07 at 10 58 58 pm


bryevdv:
Well, I tried to do it quickly but for whatever reason the plot is boxed in with scrollbars, and the source code is not rendering at all. The source code appears in the HTML output but it looks like "stretch both" causes the plot to obscure it. I guess we can merge this as-is and leave gallery and layout improvements for elsewhere.

PR #5115, opened 2016-09-08
Replace "else if" with switch statement

mattpap:
This way the intent is stated more clearly and the generated code should be marginally smaller.


bryevdv:
Thanks, LGTM. Can you make an issue for this, then merge?

PR #5116, opened 2016-09-08
Make HasProps.id a first class citizen

mattpap:
Previously `id` was a special case in several places, so now it's a property.


bryevdv:
sorry, should have asked to make an issue, just making it an "issue PR" now

PR #5120, opened 2016-09-08
Don't require tornado to run non-server bokeh

### LINKED ISSUE #5119: Non-server bokeh requires tornado
E.g.:
```
$ python -c 'import bokeh.io'
  File "<string>", line 1, in <module>
  File "bokeh/io.py", line 29, in <module>
    from .core.state import State
  File "bokeh/core/state.py", line 46, in <module>
    from ..client import DEFAULT_SESSION_ID
  File "bokeh/client/__init__.py", line 6, in <module>
    from .session import ClientSession, pull_session, push_session, show_session, DEFAULT_SESSION_ID
  File "bokeh/client/session.py", line 10, in <module>
    from ._connection import ClientConnection
  File "bokeh/client/_connection.py", line 11, in <module>
    from tornado import gen, locks
ImportError: No module named tornado
```

mattpap:
I don't like this solution, but don't see anything better right now to make, e.g., `bokeh.io` (and related) not fail to import without tornado.

fixes #5119

### DIFF
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -7,14 +7,11 @@ import logging
 
 log = logging.getLogger(__name__)
 
-from ._connection import ClientConnection
-
 from bokeh.resources import ( DEFAULT_SERVER_WEBSOCKET_URL,
                               server_url_for_websocket_url,
                               _SessionCoordinates )
 from bokeh.document import Document
 from bokeh.util.session_id import generate_session_id
-from bokeh.util.tornado import _DocumentCallbackGroup
 
 DEFAULT_SESSION_ID = "default"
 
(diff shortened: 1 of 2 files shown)


bryevdv:
How about a test to maintain this? Perhaps picking a set of common modules to import, then checking `sys.modules` for tornado?

bryevdv:
Perhaps something like:
```
$ python -c"import sys; import bokeh.io; sys.exit(1 if any('tornado' in x for x in sys.modules.keys()) else 0)"
[bryan:~/work/bokeh/sphinx]$ echo $?
```

bryevdv:
Added the test. LGTM will merge when green.

PR #5127, opened 2016-09-09
Fix deserialization of Instance(HasProps) and add HasProps.__eq__

### LINKED ISSUE #5118: GMapPlot Error AttributeError("'BasicProperty' object has no attribute 'from_json'",)
possibly  a bug in the serialization of MapOptions 

screen shot 2016-09-06 at 9 57 33 am
screen shot 2016-09-06 at 9 59 17 am

Bokeh version 0.12.1 
 OSX 10.11.6 
 
 I am getting an AttributeError("'BasicProperty' object has no attribute 'from_json'",) and I don't know where its coming from. Attached are the log files from terminal and pycharm. 
 
 Here is some example code that gives the same error. 
 
```
 from bokeh.plotting import Figure, ColumnDataSource, show, vplot 
 from bokeh.io import output_file 
 from bokeh.models import (Slider, CustomJS, GMapPlot, 
                           GMapOptions, DataRange1d, Circle, Line) 
 import numpy as np 
 from bokeh.layouts import row, column, widgetbox 
 
 output_file("path.html") 
 
 # Create path around roundabout 
 r = 0.000192 
 
 x1 = np.linspace(-1,1,100)*r 
 x2 = np.linspace(1,-1,100)*r 
 x = np.hstack((x1,x2)) 
 
 f = lambda x : np.sqrt(r**2 - x**2) 
 
 y1 = f(x1) 
 y2 = -f(x2) 
 y = np.hstack((y1,y2)) 
 
 init_x = 40.233688 
 init_y = -111.646784 
 
 lon = init_x + x 
 lat = init_y + y 
 
 # Initialize data sources. 
 location = ColumnDataSource(data=dict(x=[lon[0]], y=[lat[0]])) 
 path = ColumnDataSource(data=dict(x=lon, y=lat)) 
 
 # Initialize figure, path, and point 
 """I haven't been able to verify that the GMapPlot code below works, but 
 this should be the right thing to do. The zoom may be totally wrong, 
 but my latlng points should be a path around a roundabout. 
 """ 
 options = GMapOptions(lat=40.233681, lng=-111.646595, map_type="roadmap", zoom=15) 
 fig = GMapPlot(x_range=DataRange1d(), y_range=DataRange1d(), map_options=options,api_key='XXXXXXXXXXXXXXXX') 
 
 # fig = Figure(plot_height=600, plot_width=600) 
 
 c = Circle(x='x', y='y', size=10) 
 p = Line(x='x', y='y') 
 
 fig.add_glyph(location, c) 
 fig.add_glyph(path, p) 
 
 # Slider callback 
 callback = CustomJS(args=dict(location=location, path=path), code=""" 
     var loc = location.get('data'); 
     var p = path.get('data'); 
 
     t = cb_obj.get('value'); 
 
     /* set the point location to the path location that 
        corresponds to the slider position */ 
     loc['x'][0] = p['x'][t]; 
     loc['y'][0] = p['y'][t]; 
 
     location.trigger('change'); 
 """) 
 
 # The way I have written this, 'start' has to be 0 and 
 #  'end' has to be the length of the array of path points. 
 slider = Slider(start=0, end=200, step=1, callback=callback) 
 
 layout = column(fig, slider) 
 curdoc().clear() 
 curdoc().add_root(layout) 
 
 session.show() 
 session.loop_until_closed() 
```

Sarah Bird - Response 

I'm not sure what's going on, but I've started a little debugging

If you run bokeh-serve with --log-level=debug

You see the source of the error is:
  File "/Users/caged/Dev/bokeh/bokeh/bokeh/core/properties.py", line 1236, in from_json
    attrs[name] = prop.from_json(value, models)
AttributeError: 'BasicProperty' object has no attribute 'from_json'

I stuck a print statement in to print out "prop" and it always prints out a GMapOptions value - although not always the same one - which makes sense because GMapOptions a dictionary.

e.g.
```
zoom:Int
map_type:Enum('satellite', 'roadmap', 'terrain', 'hybrid')
lng:Float
lat:Float
```

I think there's some kind of serialization problem with GMapPlot and the server (although maybe it's wider). Hopefully Bryan has some thoughts.

Bryan Van de Ven - Response

Based on Sarah's observations, I'd agree this seems like a bug that is particular to the serialization of MapOptions. In which case the next step is to file a bug report on GitHub. Fixing it will probably take some time, however, as there are many priorities and few people to work on them. As an immediate workaround I would suggest looking at using a regular Plot with a TileProvider layer:

	http://bokeh.pydata.org/en/latest/docs/user_guide/geo.html#tile-providers

There are a few tile providers built-in already:

	http://bokeh.pydata.org/en/latest/docs/reference/tile_providers.html

CartoDB is probably a good choice.

mattpap:
fixes #5118

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1243,7 +1243,7 @@ class Instance(PropertyDescriptor):
                 attrs = {}
 
                 for name, value in iteritems(json):
-                    prop = self.instance_type.lookup(name)
+                    prop = self.instance_type.lookup(name).descriptor
                     attrs[name] = prop.from_json(value, models)
 
                 # XXX: this doesn't work when Instance(Superclass) := Subclass()
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -703,13 +703,13 @@ class HasProps(with_metaclass(MetaHasProps, object)):
             setattr(self, name, value)
 
     def __eq__(self, other):
-        if not isinstance(other, HasProps):
+        if not isinstance(other, self.__class__):
             return False
         else:
             return self.properties_with_values() == other.properties_with_values()
 
     def __ne__(self, other):
-        if not isinstance(other, HasProps):
+        if not isinstance(other, self.__class__):
             return True
         else:
             return self.properties_with_values() != other.properties_with_values()
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -702,6 +702,18 @@ class HasProps(with_metaclass(MetaHasProps, object)):
         for name, value in properties.items():
             setattr(self, name, value)
 
+    def __eq__(self, other):
+        if not isinstance(other, HasProps):
+            return False
+        else:
+            return self.properties_with_values() == other.properties_with_values()
+
+    def __ne__(self, other):
+        if not isinstance(other, HasProps):
+            return True
+        else:
+            return self.properties_with_values() != other.properties_with_values()
+
     def __setattr__(self, name, value):
         # self.properties() below can be expensive so avoid it
         # if we're just setting a private underscore field
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -702,17 +702,17 @@ class HasProps(with_metaclass(MetaHasProps, object)):
         for name, value in properties.items():
             setattr(self, name, value)
 
-    def __eq__(self, other):
+    def equals(self, other):
+        """ Structural equality of models. """
+        # NOTE: don't try to use this to implement __eq__. Because then
+        # you will be tempted to implement __hash__, which would interfere
+        # with mutability of models. However, not implementing __hash__
+        # will make bokeh unusable in Python 3, where proper implementation
+        # of __hash__ is required when implementing __eq__.
         if not isinstance(other, self.__class__):
             return False
         else:
             return self.properties_with_values() == other.properties_with_values()
-
-    def __ne__(self, other):
-        if not isinstance(other, self.__class__):
-            return True
-        else:
-            return self.properties_with_values() != other.properties_with_values()
 
     def __setattr__(self, name, value):
         # self.properties() below can be expensive so avoid it
--- a/bokeh/core/tests/test_properties.py
+++ b/bokeh/core/tests/test_properties.py
@@ -1469,6 +1469,36 @@ class TestProperties(unittest.TestCase):
         # Invalid values
         self.assertFalse(prop.is_valid((datetime.date(2012, 10, 1), 22)))
 
+def test_HasProps___eq_____ne__():
+    class Foo(HasProps):
+        x = Int(12)
+        y = String("hello")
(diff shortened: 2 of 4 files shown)


bryevdv:
Appears to be a py2/3 issue with these changes. Is #5118 the correct issue BTW? I don't immediately see the connection.

bokeh/core/properties.py
```diff
@@ -1231,7 +1243,7 @@ def from_json(self, json, models=None):
                 attrs = {}
 
                 for name, value in iteritems(json):
-                    prop = self.instance_type.lookup(name)
+                    prop = self.instance_type.lookup(name).descriptor
```
mattpap: @bryevdv, yes. This line appears in the traceback in the issue. The remainder of changes is to allow `assertEquals()` to work.

mattpap:
> TypeError: unhashable type: 'BoxAnnotation'

That's correct. Currently `HasProps` (and `Model`) implement `__eq__` and `__ne__`, but don't implement `__hash__`, because that would conflict with mutability of models. We assume hashability here and there, e.g. by constructing sets of models. One way to deal with this to make `Model` hashable by id and use only id in equality testing. Object will be still mutable, but that won't affect `==` and hash and their relationship. Though `HasProps` would have to stay unhashable.

bryevdv:
@mattpap that seems OK

mattpap:
> that seems OK

No, because it defeats the whole purpose of allowing to structurally compare models, which I need for testing and interactive sessions (well, that's pretty much testing again). To get structural equality working across the board, I removed `__eq__` and implemented `equals()` that just compares properties and their values, and disregards `id`.

I think we can implement `__eq__` in general, but it won't be structural equality, especially that we want also mutability and hashability. What has to be done is to make `HasProps` have `id` field (move from `Model`). This is how bokehjs works and I think that's the correct approach. This will unify both versions. Note that we already implicitly assume this, by allowing Python to use intrinsic object identification (memory location) to be used for hashability.

Also, I would like to rename `HasProps` to `Struct`, because it's hard to refer meaningfully to `HasProps` in writing.

mattpap:
How naive of me:
```
File "/home/travis/build/bokeh/bokeh/bokeh/core/tests/test_properties.py", line 1505
    v = 1.equals(Foo(x=1))
         ^
SyntaxError: invalid syntax
```
assuming that Python is a regular programming language.

bryevdv:
`int(1).equals(Foo(x=1)` might work I'm not sure if it tests something you care about. 

I'm +0 on renaming to `Struct`

I think `HasProps` having an intrinsic `id` is a bit weird, since I associate `id` specifically with things that can be and are serialized, i.e. `Model` and subclasses. I'm not necessarily opposed tho, but I would have to really dig in to things a bit to really evaluate. I'd say make the PR you want then let's review it in detail.

mattpap:
@bryevdv, I will continue discussion on `__eq__` and `__hash__` elsewhere.

PR #5129, opened 2016-09-09
Allow to require("bokehjs") in node.js

### LINKED ISSUE #5125: Bokeh npm install
I'm trying to load Bokeh in an electron context, which means I'm in node with a browser context. However, the object that `require('bokehjs')` returns has no methods (and isn't itself a function). Was there a packaging issue with the npm release or should I be requiring this differently?

mattpap:
It would be nice to share code with `bokehjs/test/index.coffee`, but for whatever reason my attempts to do so failed. The new `require()` would work, but it isn't picked up by node.js.

fixes #5125


bokehjs/package.json
```diff
@@ -99,11 +97,14 @@
     "proj4": "^2.3.10",
     "sprintf": "^0.1.5",
     "timezone": "0.0.38",
-    "coffee-script": "^1.9.3"
```
mattpap: This should have been done in PR #4946.

bokehjs/src/coffee/node_main.coffee
```diff
@@ -0,0 +1,43 @@
...
+    this.constructor._load(modulePath, this)
+
+  overridePath = pkg.browser[modulePath]
+
+  if overridePath?
+    modulePath = path.join(root, overridePath)
+
+  return load(modulePath)
+
+if not (global.window? and global.document?)
```
mattpap: This is to make this work in environments that provide their own web browser-like environment. Note that bokehjs requires `window`, `document` and `navigator` to initialize.

bokehjs/package.json
rgbkrk: 👍 I wondered about this when I was installing too.

bokehjs/src/coffee/node_main.coffee
```diff
@@ -0,0 +1,43 @@
...
+  jsdom = require('jsdom').jsdom
```
rgbkrk: Later, if/when someone tries to webpack bokeh (hint: jupyterlab, other web contexts), they're going to have this annoyingly dynamic require to deal with.

bryevdv: @mattpap any comments?

mattpap: @rgbkrk, to me, being able to repackage bokehjs with webpack and run bokehjs in node.js, are two different targets. There are two reasons bokehjs doesn't work out of the box in node.js. One are module aliases (to `src/vendor`), defined in `package.json`. This is something webpack should easily deal with on its own. The other is lack of appropriate environment in node.js, as bokehjs is a web browser library primarily. This shouldn't concern webpack at all. However, 
if webpack requires a web browser library to be a node.js library as well, then what we can do, is to make bokehjs at least import without requiring a well set up environment. I presume this is how, e.g., jquery works in webpack. 

Also, trying to repackage bokehjs in a custom way may not be as trivial as just including it in a thridparty webpack build, because some of bokehjs' functionality, like custom models, requires specific javascript bundle's encoding and additional code to be present at runtime. Note that at this point we don't use stock browserify to build bokehjs (we use a custom module labeler), and it's likely we will be using a custom approach to unify bokehjs' and custom model build, and allow for building custom, per plot, bokehjs bundles. However, if you just want to include core bokehjs' functionality, that should be easy though.

mattpap: @rgbkrk, do you have a test case on which I could experiment? In the end our build may be too complicated and there might be room for improvement.

mattpap: > is to make bokehjs at least import without requiring a well set up environment.

After initial investigation, this seems to be a lot of work for little gains.

mattpap: According to webpack's documentation, it should be possible to avoid this custom `require()` logic by setting `resolve.packageAlias = "browser"`.

mattpap: So, it looks like the only problem I can see right now, is that there there is one `main` field in `package.json` , but there are two conflicting entry points for either node.js and browserify. If webpack can be told to override the default `main`, then there should be no problem at all.

rgbkrk:
I'm not looking to run bokeh in node proper, but in electron. In electron you have window and document - you can think if it as a browser environment where commonjs loading works.

I'm asking for this ability because I'd like to be able to provide a default transform for bokeh in nteract, much like how we support geojson and Plotly:

GeoJSON support:

https://github.com/nteract/nteract/pull/688

Release with screenshots (note: this is not the latest):

https://github.com/nteract/nteract/releases/tag/v0.0.8

mattpap:
@rgbkrk, OK, I got rid of `jsdom`. There is also new `bokehjs/examples/electron` example showing integration.

rgbkrk:
Awesome! Great to see the electron app example too.

mattpap:
There will be another release in two weeks, so this will be available as bokehjs 0.12.3.

rgbkrk:
Sweet, I look forward to it!

rgbkrk:
:bow: thanks again!

PR #5130, opened 2016-09-09
Correct typo in the notebook docs

imagejan:


bryevdv:
Thanks for the catch @imagejan

PR #5140, opened 2016-09-11
Executing range callbacks on reset_range (additional fix) and initialization (new)

### LINKED ISSUE #4614: Reset tool does not trigger x/y_range callback events

schlichtanders:
I experienced update issues with #4615, concretely the range callbacks are not executed correctly. I had to call reset twice for my callbacks to work correctly. This pull request fixes this.

Additionally I added the analog calls to the callbacks within the initialization such that dynamic ``update_datarange`` is followed by a call to the callbacks here too.

It may be preferable to move the callback executions to ``update_datarange`` directly

- [x] issues: fixes #4614 #4615
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -121,6 +121,11 @@ class PlotCanvasView extends Renderer.View
 
     @update_dataranges()
 
+    for name, rng of @frame.x_ranges
+      rng.callback?.execute(rng)
+    for name, rng of @frame.y_ranges
+      rng.callback?.execute(rng)
+
     @unpause()
 
     logger.debug("PlotView initialized")
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -404,11 +404,15 @@ class PlotCanvasView extends Renderer.View
     if not range_info?
       for name, rng of @frame.x_ranges
         rng.reset()
-        rng.callback?.execute(rng)
       for name, rng of @frame.y_ranges
         rng.reset()
-        rng.callback?.execute(rng)
       @update_dataranges()
+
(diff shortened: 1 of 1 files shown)


bryevdv:
ping @canavandl @mattpap

canavandl:
hi @schlichtanders 

I think the solution to range callbacks not executing properly is to move the `execute` logic onto the range models themselves. That'd simplify the several places in the code where range attributes (often pertaining to ranges of different types) are changed.

I think it would be as simple as adding an `change` eventListener to the range models and making sure that the public methods (`update` and `reset`) cause the callback to be called once and only once.

Also - do you have a [mcve](http://stackoverflow.com/help/mcve) that you can share, where the range callbacks aren't executed? That way we can make sure that your use case is corrected by any changes.

thanks again for contributing!

schlichtanders:
thanks @canavandl
it is good to hear, that this in fact could become an improvement of bokeh

I agree, putting everything into the range model seems appropriate. Next
week I have more time and will look into this probably. However, these are
indeed my first true javascript/coffeescript experiences. So might be that
I get stuck soon.

I thought a kind of minimal working example is
http://bokeh.pydata.org/en/latest/docs/user_guide/interaction/callbacks.html#customjs-for-range-update
(because I copied from it) however, that one seems to work properly with
*reset. *(initialization still does not update the rectangle).
Don't know, what I did differently... maybe it is the use of concrete times
as x-axis. Will check that out

best,
Stephan

schlichtanders:
one weak later, I have to admit, that I won't have enough time to fix by myself.

would be great if someone else could bring the callback executions to the right place within the range model

canavandl:
superceded by #5253

PR #5145, opened 2016-09-12
Fix tooltips after replacing computed properties with getters

### LINKED ISSUE #5123: VBar hover tooltip not working in master
After building and installing master to get a build with colored legends working for VBar plots, I noticed that my hover tooltips (which were working in 0.12.2) are now not working in master. Some sample code that generates my tooltip:
```
fig = plotting.figure(width=1200, height=900, x_range=list(weekList))

step0 = fig.vbar(x='week', bottom=0, top='step0', width=0.5, color='blue', source=source, legend="1.")

step1 = fig.vbar(x='week', bottom='step0', top='step1', width=0.5, color='green', source=source, legend="2")

fig.add_tools(HoverTool(renderers=[step0], tooltips=[("Created", "@step0"), ("Number of samples", "@count")]))

fig.add_tools(HoverTool(renderers=[step1], tooltips=[("Sent", "@data1"), ("Number of samples", "@count")]))
```

No tooltip is appearing upon hovering any part of the plot

@birdsarah

mattpap:
~~@bryevdv's fix in a form of PR.~~ Nope, that didn't work, I had to restore `ttmodels` as a computed property. This needs to be changed appropriately in future.

fixes #5123

### DIFF
--- a/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
@@ -331,9 +331,7 @@ class HoverTool extends InspectTool.Model
     @add_dependencies('computed_renderers', this, ['renderers', 'names', 'plot'])
     @add_dependencies('computed_renderers', @plot, ['renderers'])
 
-  @getters {
-    computed_renderers: () -> @_get_computed('computed_renderers')
-    ttmodels: () ->
+    @define_computed_property 'ttmodels', () ->
       ttmodels = {}
       tooltips = @tooltips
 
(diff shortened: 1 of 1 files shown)


birdsarah:
Thanks so much @mattpap - really appreciate the speedy turnaround.

bryevdv:
@mattpap what case was not working?

mattpap:
Tests were failing and additionally we actually want to recompute `ttmodels` when dependencies change, so for now it's safer to return to the original implementation and reinvestigate this later with other computed properties.

Rubyj:
I can confirm that this does indeed fix #5123

bryevdv:
@mattpap understood, let's get this in an clean up more thoroughly in a follow-on. @birdsarah this LGTM if you want to merge for your dev build.

PR #4660, opened 2016-06-23
Hotfix to make DataTables editable

bgyarfas:
`BooleanFormatters` are still broken, but I'm able to edit a DataTable now

- [x] issues: fixes #4616
- [ ] tests added / passed

* Changed CellEditor to be a subclass of Widgets.Model

* Changed order of initializer for CellEditorView (probably not best practice)

PR #5146, opened 2016-09-12
Make DataTable cells editable

birdsarah:
This is a re-PR of #4660 which had to get reverted after merge.


birdsarah:
There were problems with the tests after merging into master.

bokehjs/src/coffee/models/widgets/cell_editors.coffee
```diff
@@ -5,26 +5,24 @@ $2 = require "jquery-ui/spinner"
...
-
-class CellEditor extends Model
-
-class CellEditorView extends Widget.View
+class CellEditorView extends BokehView
 
   tagName: "div"
-  className: "bk-cell-editor"
+  attributes:
+    class: "bk-cell-editor"
```
mattpap: How this works? I don't see any references to `attributes` on non-Backbone.Model.

mattpap: Never mind, backbone.

bryevdv:
@birdsarah well test are green again on the PR, do we try to merge again?

bryevdv:
Going to merge this and see if it will pass on master.

bryevdv:
Nope, tho now selenium shows it fails because `mget` is gone. I think we need to make PR from a @bokeh/dev contributor (so that the tests are run in the "normal" way) and cherry pick the commits to it keeping attribution so that @bgyarfas gets credit. Alternatively, we could skip the test for now. I guess it needs a merge from master in any case, to get rid of the `mget` in any case.

bgyarfas:
@bryevdv I made you a collaborator on my branch if you wanted to make changes.

bgyarfas:
No worries :)

PR #5157, opened 2016-09-13
session.show(): take into account browser and new keywords

### LINKED ISSUE #5156: session.show() does not take into account browser
This code does not work, it will open new tabs in the current opened browser:

    from bokeh.io import curdoc

    # prepare some data
    x = [1, 2, 3, 4, 5]
    y = [6, 7, 2, 4, 5]

    p = figure(title='simple line example', x_axis_label='x', y_axis_label='y')

    p.line(x, y, legend='Temp.', line_width=2)

    custom_firefox = '/usr/bin/firefox -P ipython --new-tab %s'
    session.show(obj=p, browser=custom_firefox)

I read the code source and see that `browser` and `new` keywords are ignored.

BoboTiG:
This little patch fixes the ignorance of `browser` and `new` keywords into `session.show()` and `show_session()`. 

- [x] issues: fixes #5156
- [x] tests passed

Now, it is possible to call a customized browser:

    # ...
    custom_firefox = '/usr/bin/firefox -P ipython --new-tab %s'
    session.show(browser=custom_firefox)

### DIFF
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -292,7 +292,7 @@ class ClientSession(object):
         """
         if obj and obj not in self.document.roots:
             self.document.add_root(obj)
-        show_session(session=self)
+        show_session(session=self, browser=browser, new=new)
 
     @classmethod
     def _ensure_session_id(cls, session_id):


bokeh/client/session.py
```diff
@@ -164,10 +165,7 @@ def show_session(session_id=None, url='default', app_path='/',
             server_url = coords.server_url
             session_id = coords.session_id
 
-        if controller is None:
-            from bokeh.util.browser import get_browser_controller
```
bigreddot: We don't normally defer imports. The fact that it was done here makes me suspect there is a reason, though can't say offhand what it would've been. Wouldn't the change on line 293 be sufficient in any case? I think if you pass along `browser` as you have done there, then `show_session` will work without any changes? I can't say for certain there are no places that rely on being able to pass in a controller explicitly, so I am reticent to make a breaking change here.

BoboTiG: Sure, I did not see your comment ... Sorry for the rebased PR, promise it will not happen again.

bryevdv:
Thanks for the PR @BoboTiG !

bryevdv:
restarted spurious test failure, will merge when green

BoboTiG:
Does the `gulp-mocha` error is related to this patch?

bigreddot:
No we have some intermittent build/test issues, currently just have to restart tests occasionally. All good now, merging!

Issue #5121, opened 2016-09-08
Fix skipped JS tests as result of getter/setter work


canavandl:
PR #5106 (replacing mget/get with getters) added skips to several test modules that weren't passing.

skipped tests modules include:
- [x] ColorBar 
- [ ] SelectionManager

PR #5066, opened 2016-08-31
Refactor of select tools

canavandl:
supercedes: 5054

- [ ] issues: fixes #5065 
- [ ] tests added / passed
- [ ] release document entry (if new feature or API change)

Issue #2328, opened 2015-05-21
Improve Selection Manager


fpliger:
The intent here is open a discussion regarding the current selection management (SM) implementation to better understand: 
- what would be the desired functionality
- what SM currently provides and what not
- understand what changes are necessary to add what's missing

There are quite a few open issues regarding selectors and inspectors that are related to that. The introduction of the JSActions add an extra area where a different (better?) SM would come handy. To provide an example, let's say there are overlapping glyphs (circle, i.e.) and an user uses the tap tool + OpenURL action to open a link. There is now way to add any minimal logic to prevent opening N new tabs. The same "overlapping" situation is an issue for HoverTool as well.

bryevdv:
reference: https://github.com/bokeh/bokeh/wiki/Selection-and-Hit-test-Managements 

Changing this to a feature issue, since there is work started that could reference this issue in a future PR (ping @mattpap)

bryevdv:
related to #1512

bryevdv:
ping @canavandl this issue and a recent ML post reminds me of the possible need for the selection manager. WE very much need to be able to implement polices like:

* only return the closest
* only return the closest that is also "on top"
* return everything
* return N closest

A selection manager seems like a central place to specify and implement these policies. Thoughts?

asanakoy:
When showing tooltips on the overlapping points it would be nice to be able to have options:
- Stack all tooltips vertically (as it is working right now)
- Display all tooltips in a grid.  
Use-case for this is when you visualize a lot of points (for example TSNE plot), each of them corresponding to an image. Each tooltip shows an image. If there are too many overlapping points tooltips do not fit in the screen. Grid could be a much better option here.

chesucr:
When you select points with some selection tools and you change the tool you are working with, I think the selection should be kept. The current behaviour is to reset and lose the previous selected points

bryevdv:
this is a dupe of #1512

PR #5158, opened 2016-09-13
Fix linked selections, unify select tools and remove selection manager model

### LINKED ISSUE #2136: Selections only work on the last of two series in one plot

canavandl:
This PR 
- Removes the SelectionManager model by moving it's (relatively simple) methods onto the Selector models and replaces the SelectionManager as the selector attribute on DataSources.
  - This has the benefit of fixing linked selections (broken since #2136)
- Factors out a lot of duplicated code onto the parent SelectTool class
  - unifies the behavior of the various select tools

todo: 
- ~~update the API typings~~
- [x] issues: fixes #2136 
- [x] tests added / passed

supercedes #5066 #5067


canavandl:
Linked selection was previously broken because a select tool would iterate over it's glyph_renderers and overwrite a ColumnDataSource's selected attr (the hittest dict) each time, so that the last iteration would be the result.

This change causes SelectionManager.select to mutate the selected dict instead of overwriting it. This pushes a tiny bit more code onto the tool (deciding if/when to explicitly clear selections), but it seems pretty natural.

While this PR resolves #2136 , it suggests that a better path forward would be to deprecate SelectionManagers entirely and replace the ColumnDataSource selected and inspected attrs (now hittest dicts) with Selector objects that manage their own updates/clears. Doing so shouldn't be hard, as SelectionManagers are just setting props and triggering events.

bokeh/models/annotations.py
```diff
@@ -573,7 +573,7 @@ class LabelSet(TextAnnotation):
 
     border_line_color = Override(default=None)
 
-    source = Instance(DataSource, default=lambda: ColumnDataSource(), help="""
```
canavandl: was instance of mismatched JS/python defaults (could have been fixed in separate PR but was done here)

bokehjs/src/coffee/models/annotations/label_set.coffee
```diff
@@ -170,7 +170,7 @@ class LabelSet extends TextAnnotation.Model
       angle:        [ p.AngleSpec,    0                 ]
       x_offset:     [ p.NumberSpec,   { value: 0 }      ]
       y_offset:     [ p.NumberSpec,   { value: 0 }      ]
-      source:       [ p.Instance,     () -> new ColumnDataSource.Model()  ]
```
canavandl: was instance of mismatched JS/python defaults (could have been fixed in separate PR but was done here)

bokehjs/src/coffee/models/annotations/arrow.coffee
```diff
@@ -9,8 +9,6 @@ p = require "../../core/properties"
 class ArrowView extends Annotation.View
   initialize: (options) ->
     super(options)
-    if not @model.source?
```
canavandl: was instance of mismatched JS/python defaults (could have been fixed in separate PR but was done here)

bokehjs/src/coffee/api/typings/models/tools.d.ts
```diff
@@ -78,19 +78,23 @@ declare namespace Bokeh {
     export var BoxSelectTool: { new(attributes?: IBoxSelectTool, options?: ModelOpts): BoxSelectTool };
     export interface BoxSelectTool extends SelectTool, IBoxSelectTool {}
     export interface IBoxSelectTool extends ISelectTool {
+        callback?: Callback | ((source: this) => void);
```
mattpap: Why `this`? Shouldn't this be `DataSource`? If that's the case, perhaps `callback` could be moved to the super class (`SelectTool`). The same should be done on Python side, but there it will conflict with how we do documentation.

bokehjs/src/coffee/common/selector.coffee
```diff
@@ -7,25 +7,77 @@ p = require "../core/properties"
 class Selector extends HasProps
   type: 'Selector'
 
-  update: (indices, final, append, silent=false) ->
+  select: (tool, renderer_view, geometry, final=true) ->
+    if @source != renderer_view.model.data_source
+      logger.warn('select called with mis-matched data sources')
```
mattpap: Is this error really recoverable? Is this something user can inflict? Then this should be `throw new Error('bla')`.

bokehjs/src/coffee/common/selector.coffee
```diff
@@ -7,25 +7,77 @@ p = require "../core/properties"
...
+
+  inspect: (tool, renderer_view, geometry, final=true) ->
+    if @source != renderer_view.model.data_source
+      logger.warn('inspect called with mis-matched data sources')
+
+    indices = renderer_view.hit_test(geometry)
+
+    if indices?
+      inspected = @_update(indices, false, final, true) # append = false
+      # Here for compatibility, should be removed
```
mattpap: This kind of comments should be marked with `TODO`.

bokehjs/src/coffee/api/typings/models/tools.d.ts
bryevdv: Conflict in what way?

mattpap: If `callback` is shared by all subclasses of a class, then it should be defined on the superclass. However, we want custom documentation for `callaback` in each subclass. To do this, we actually have to define it in each subclass. I was hoping `Override` could help here, but there is only support for default values. Perhaps `Override` could be allowed to override `help` as well.

birdsarah: > Perhaps Override could be allowed to override help as well.

+1 I've wanted to do this before for something else.

canavandl: There's two issues here (thanks for bringing this up @mattpap )

1. In terms of unifying the select tools: the TapTool currently returns the DataSource as the cb_obj in the callback (which is required by the working of OpenURL), while all of the other tools return the tool instance. Should all of the tools return the data source or the tool instance (which is perhaps less helpful, but more consistent with the Bokeh pattern of returning the instance of the object that kicked off the callback).

2. In terms of where to place the `callback` attr: the class hierarchy of the tools in python and JS don't match up very well, so there's no python `SelectTool` model.

canavandl:
I think I need to reconsider the Selector `select` / `inspect` function signatures (which are unchanged from master). The current codeflow is:

1.  a select tool will define a geometry (an x,y point for a tap, a rect for a box select)
2.  for each glyph renderer, the tool will check if the geometry yields a positive hit test
3.  for positive hits, it will update the the selections and trigger a `select` event

The issue is that for "hit-ed" renderers that share a data source, multiple `select` events will be triggered and cause any attached callback to be triggered repeatedly.

bokehjs/src/coffee/common/selector.coffee
canavandl: I think this issue will/would be obviated by resolved the goal described [below](https://github.com/bokeh/bokeh/pull/5158#issuecomment-247088450)

If you have any suggestions, that'd be great.

mattpap:
Perhaps this could fix #5121 as well.

bryevdv:
@canavandl please see a comment I made in #2328

bryevdv:
This has alot of conflicts now, and I am no longer sure about removing `SelectionManager` completely.  I'm going to go ahead and close, but leave the branch around for reference.

PR #5161, opened 2016-09-14
use npm phantomjs for tests #5159

### LINKED ISSUE #5159: py.test should use phantomjs from bokehjs/node_modules/.bin by default
phantomjs is required for running the examples tests.  While they aren't typically run locally, it would be nice to have phantomjs listed in the test.requires.  It will enable scripts/dev_environment to include it.

Discussed with @birdsarah [in the gitter](https://gitter.im/bokeh/bokeh?at=57d8bdd6b37816b14fb5c72a)

timsnyder:
- [x] issues: fixes #5159 

As py.test uses `bokehjs/node_modules/.bin/gulp`, it should also use the phantomjs installed by npm as the default rather than expecting it to be found correctly in the user's path.


mattpap:
@timsnyder, can merge most recent `master` into this? There were issues with our travis ci build and there will be test failures otherwise.

timsnyder:
@mattpap, oops!  didn't realize that I was that far behind when I made this.  Sorry for the delayed reaction to the travis build fail.

bryevdv:
@mattpap @birdsarah this seems good to merge too?

bryevdv:
Since it won't fully be tested until a merge, I will merge now, and revert if there is a problem.

bryevdv:
Thanks @timsnyder !

PR #5166, opened 2016-09-15
Bring some structure to *.less files

### LINKED ISSUE #5165: Bring some structure to our *.less sources
Currently it's a chaos, with `continuum.less` being a garbage sink.

mattpap:
fixes #5165


bokehjs/src/less/main.less
```diff
@@ -153,11 +11,3 @@
   border: 1px dashed green;
   z-index: 100;
 }
-
-.bk-toolbar-button.hover:focus {
-  outline: none;
-}
-
-.box .bk-grid-column {
```
mattpap: @birdsarah, was there any use for this? Maybe this should be `bk-box`? Though, `bk-box` being present in another place (also removed in this PR), isn't used across the repository.

bryevdv:
LGTM as far as it goes, it would be good to have @birdsarah comment as well.

bokehjs/src/less/main.less
birdsarah: I can't think of one - fine with me that it's gone. Sorry for the v. late reply.

PR #5172, opened 2016-09-15
Replace jsnlog with a custom, line number preserving logger

### LINKED ISSUE #5171: Replacing jsnlog
I'm not sure what was the rationale for choosing jsnlog, but this seems to be a bad choice. jsnlog is big (1kloc), by default tries to do AJAX calls, which leads to meaningless errors in js console when using `file://` protocol, it is obtrusive to the hosting environment (configures `window.onerror`) and doesn't preserve line numbers. I propose switching to e.g. `loglevel`, which is small, unobtrusive and preserves line numbers. ~~The only deficiency is lack of log message prefixing (with `Bokeh`), but otherwise it's not possible to maintain line numbers~~. I think that having correct line numbers is more important than a prefix (especially since switching to browserify, where we have one big blob of code). @bryevdv, I would like your feedback on this.

mattpap:
This implements a custom logger. It preserves line numbers and allows for logger naming (prefixes). This also slightly reduces bokehjs size and avoids issues that were present in jsnlog (`window.onerror` and AJAX requests). This should work across modern web browsers, but unfortunately it may be broken in some versions of Firefox (you will get '(unknown)' location). This is FF's bug.

fixes #5171

### DIFF
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -93,7 +93,6 @@
     "jquery": "^2.1.1",
     "jquery-mousewheel": "^3.1.12",
     "jquery-ui": "~1.10.5",
-    "jsnlog": "^2.7.5",
     "rbush": "^2.0.1",
     "proj4": "^2.3.10",
     "sprintf": "^0.1.5",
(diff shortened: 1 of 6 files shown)


mattpap:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1280818, this will be fixed in FF 50.

PR #5176, opened 2016-09-16
Add support for low color and high color for linear and log mappers

### LINKED ISSUE #5011: ColorMapping - color values out of high/low
```python
data = [-1, 0, np.NaN, 1, 2]
cmap = LinearColorMapper(palette=["red", "green"], low=0, high=1, nan_color=‘gray’, high_color=‘orange’, low_color=‘pink’)
# expect: [‘pink’, ‘red’, ‘gray’, ‘green’, ‘orange’]
```

LogColorMapper should also receive this treatment.

Maggie-M:
issues: fixes #5011

### DIFF
--- a/bokehjs/src/coffee/models/mappers/linear_color_mapper.coffee
+++ b/bokehjs/src/coffee/models/mappers/linear_color_mapper.coffee
@@ -8,17 +8,19 @@ class LinearColorMapper extends ColorMapper.Model
   type: "LinearColorMapper"
 
   @define {
-      high:          [ p.Number           ]
-      low:           [ p.Number           ]
+      high:       [ p.Number ]
+      low:        [ p.Number ]
+      high_color: [ p.Color  ]
+      low_color:  [ p.Color  ]
     }
 
   _get_values: (data, palette) ->
-    min = @low ? _.min(data)
-    max = @high ? _.max(data)
+    low = @low ? _.min(data)
+    high = @high ? _.max(data)
     max_key = palette.length - 1
     values = []
 
-    norm_factor = 1 / (max - min)
+    norm_factor = 1 / (high - low)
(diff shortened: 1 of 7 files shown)


bokehjs/src/coffee/models/mappers/linear_color_mapper.coffee
```diff
@@ -8,31 +8,47 @@ class LinearColorMapper extends ColorMapper.Model
...
       if isNaN(d)
         values.push(@nan_color)
         continue
 
-      normed_d = (d - min) * norm_factor
+      # This handles the edge case where d == high, since the code below maps
+      # values exactly equal to high to palette.length, which is greater than
+      # max_key
+      if d == high
+        values.push(@palette[max_key])
```
canavandl: should be `values.push(palette[max_key])` not `@palette`

canavandl: should be `values.push(palette[max_key])` not `@palette`

bryevdv:
OK I took a quick look at things too. I wanted to help @Maggie-M write a unit test that would also fail, but there was already a test that seemed like it should have failed! (But wasn't failing) Some things are weird with the `ColorMapper` base class. 

* computing for a single value seems perfectly well defined? It currently returns the empty list, but should just return a single color?
* `v_map_to_screen` and `v_compute` call `_get_values` differently.  One calls with `@palette` and the other with `@_palette`. This seems confusing, I'm not quite sure what's going on, and I am guessing has something to do with why the test we expected to fail was still passing. 

@birdsarah @canavandl any insights welcome, it seems like `v_map_to_screen` and `v_compute` should both essentially do the same thing. I'm not sure why one is not implemented in terms of the other.

PR #5178, opened 2016-09-17
Make Slider handle title=None and title="" properly

### LINKED ISSUE #5175: Slider with no title (feature request)
Currently current bokeh slider always displays a value.  Even when setting the `title` to empty

```
from bokeh.models.widgets import Slider

output_file("slider.html")

slider = Slider(start=0, end=10, value=1, step=.1, title="")

show(widgetbox(slider))
```
image

There are instances when a user does not want to display any value:   One use case is when "grouping" two sliders and displaying a custom title value in some other widget e.g. textbox

mattpap:
1. `title=None` means no title and no current value.
2. `title=""` means just no title (without the annoying colon).

fixes #5175

### DIFF
--- a/bokehjs/src/coffee/models/widgets/slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/slider.coffee
@@ -47,7 +47,8 @@ class SliderView extends InputWidget.View
       slide: @slide
     }
     @$el.find('.slider').slider(opts)
-    @$( "##{ @model.id }" ).val( @$('.slider').slider('value') )
+    if @model.title?
+      @$( "##{ @model.id }" ).val( @$('.slider').slider('value') )
     @$el.find('.bk-slider-parent').height(@model.height)
     return @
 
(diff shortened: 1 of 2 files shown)


bryevdv:
@mattpap can you attach some screenshots

bryevdv:
Actually, am I reading this correctly, `title=None` means the slider itself has no value? Why would that be desirable? AFAIK `title=""` and `title=None` should behave identically?

mattpap:
Slider is rendered as:
```
title colon value
acutal_slider
```
Previously `title=""` would result in ` colon value` (I will be skipping second line), which looks ugly. In this PR `colon` is removed if title is empty. Previously, if `title=None`, then I didn't test this, but this would most likely result in `null colon value`, which is undesirable. As `value` is part of title, it's desirable to not show it altogether when `title=None`. This makes sense, because user (as requested) may want to have slider widget functionality, but display the value (if at all) differently, using a callback.

bigreddot:
Hrm, I guess this is OK for now but I can't say I like it. Seems like the best thing would be for display of title and value to be independently configurable. This with `""` and `None` behaving differently will be surprising to people, I expect.

bryevdv:
LGTM

PR #5179, opened 2016-09-17
Make `from bokeh.palettes import whatever` always return fresh list

### LINKED ISSUE #5170: Viridis6 appears to be reversed
screen shot 2016-09-15 at 10 32 52 am

mattpap:
Previously it was possible to import and modify a palette in-place, making it different from the original for consecutive imports. It's trivial to achieve a copying API, but it would be incompatible with the original, because Python doesn't implement uniform access principle. All work in this PR is directed towards getting the desired functionality and keeping the old API. E.g.
```
In [1]: import bokeh.palettes as p

In [2]: p.Magma7
Out[2]: ['#000003', '#2B115E', '#711F81', '#B53679', '#F0605D', '#FEAE76', '#FBFCBF']

In [3]: p.Magma7[0] = 1

In [4]: p.Magma7
Out[4]: ['#000003', '#2B115E', '#711F81', '#B53679', '#F0605D', '#FEAE76', '#FBFCBF']

In [5]: from bokeh.palettes import Magma7

In [6]: Magma7[0] = 1

In [7]: Magma7
Out[7]: [1, '#2B115E', '#711F81', '#B53679', '#F0605D', '#FEAE76', '#FBFCBF']

In [8]: p.Magma7
Out[8]: ['#000003', '#2B115E', '#711F81', '#B53679', '#F0605D', '#FEAE76', '#FBFCBF']

In [9]: from bokeh.palettes import Magma7

In [10]: Magma7
Out[10]: ['#000003', '#2B115E', '#711F81', '#B53679', '#F0605D', '#FEAE76', '#FBFCBF']
```
Tested in Py 2.7 and 3.5. Also made sure that auto-completion works as expected.

`WIP` status, because tests need updating and I don't have energy right now to deal with nonsense in `test_palettes.py`.

fixes #5170


bryevdv:
@mattpap had to check diff locally to see what your approach was. I figured it was the "module object" trick, which it is, and that seems fine by me. But can you explain this more:
```
+_module_name = 'bokeh.palettes'
+_orig_module = _sys.modules[_module_name]
+if not isinstance(_orig_module, _PalettesModule):
+    _new_module = _PalettesModule(_module_name)
+    _new_dict = { key: val for key, val in _orig_module.__dict__.items() if not key.startswith('_') or key.startswith('__') }
+    _new_module.__dict__.update(_new_dict)
+    _sys.modules[_module_name] = _new_module
```
I usually see it more simply expressed as just replacing the entry in `sys.modules` unconditionally:
```
sys.modules[`bokeh.palettes`] = _PalettesModule(sys.modules[`bokeh.palettes`])
```
Module caching ensures that only happens once, and AFAICT there is nothing in `orig_module` that does not start with `_` so I don't see what the dict update accomplishes.

bryevdv:
@mattpap I tried out the simpler version, it passes tests, it works with tab completion and `dir` so I will merge this when green

Issue #5180, opened 2016-09-18
deprecation warning with matplotlib-2.0.0.b4 and bokeh 0.12.2


stonebig:
Hi,

I get a deprecation warning experiment this code on Bokeh 0.12.2 with Matplotlib 2.0.0b4:
`````
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
import os
from bokeh import mpl
from bokeh.plotting import output_notebook, show
import matplotlib as mplc
# Generate the pandas dataframe
data = np.random.multivariate_normal([0, 0], [[1, 2], [2, 20]], size=100)
data = pd.DataFrame(data, columns=["X", "Y"])
mplc.rc("figure", figsize=(6, 6))

# Just plot seaborn kde
import seaborn as sns
sns.kdeplot(data, cmap="BuGn_d")

plt.title("Seaborn kdeplot in bokeh.")

from bokeh.resources import INLINE
# default solution output_notebook() relies on pydata.org (but spare 2Mo of inline jsscript in your notebook)
# other method to get internal bokeh script can be
#    os.environ['BOKEH_RESOURCES'] = 'inline'
output_notebook(resources=INLINE)

show(mpl.to_bokeh())
````
````
C:\WinPython\basedir35\buildQt5\winpython-64bit-3.5.x.3\python-3.5.2.amd64\lib\site-packages\bokeh\core\compat\bokeh_renderer.py:150: MatplotlibDeprecationWarning: The get_axis_bgcolor function was deprecated in version 2.0. Use get_facecolor instead.
  background_fill_color = convert_color(ax.get_axis_bgcolor())
````
![bokeh-20160918a](https://cloud.githubusercontent.com/assets/4312421/18615135/c8a74884-7d9e-11e6-85d3-0274379de20f.PNG)

stonebig:
https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/axes/_base.py#L2717..L2726

PR #5181, opened 2016-09-18
matplotlib2.0.0b4 compatibility fix

stonebig:
should fix https://github.com/bokeh/bokeh/issues/5180

### DIFF
--- a/bokeh/core/compat/bokeh_renderer.py
+++ b/bokeh/core/compat/bokeh_renderer.py
@@ -147,10 +147,10 @@ class BokehRenderer(Renderer):
 
     def close_axes(self, ax):
         "Complete the axes adding axes-dependent plot props"
-        if LooseVersion(mpl.__version__) <= '1.5.9':
-            background_fill_color = convert_color(ax.get_axis_bgcolor())
-        else:
+        if "get_facecolor" in dir(ax):
             background_fill_color = convert_color(ax.get_facecolor())
+        else:
+            background_fill_color = convert_color(ax.get_axis_bgcolor())
         self.plot.background_fill_color = background_fill_color
         if self.xkcd:
             self.plot.title.text_font = "Comic Sans MS, Textile, cursive"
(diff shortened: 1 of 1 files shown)


bryevdv:
Hi @stonebig this PR is failing tests because `LooseVersion` is undefined:
```
Unable to generate Bokeh plot at /home/travis/build/bokeh/bokeh/sphinx/source/docs/gallery/ggplot_density.rst:12:global name 'LooseVersion' is not defined
make: *** [html] Error 1
```

bokeh/core/compat/bokeh_renderer.py
```diff
@@ -147,7 +147,10 @@ def open_axes(self, ax, props):
 
     def close_axes(self, ax):
         "Complete the axes adding axes-dependent plot props"
-        background_fill_color = convert_color(ax.get_axis_bgcolor())
+        if LooseVersion(mpl.__version__) <= '1.5.9':
```
mattpap: `LooseVersion()` aside, is this really a breaking change in a version differing in patch/build number? Shouldn't this be `< 2`?

bryevdv: I'm guessing that's when the deprecation warning starts, not when the breaking change happened (just a guess). If the intent is specifically to always avoid the warning, it might very well need to check on a point release.

stonebig: I'm never sure what LooseVersion means, so I compared with a never happening release. 

Yes, this is to avoid deprecation warnings before they reach the end-user eye.

bryevdv:
@stonebig I am hesitant to add a new dependency to Bokeh just for this. Let me suggest another possibility: explicitly check if `get_facecolor` exists, and use it if it does (otherwise use the other call). Then there is no need to rely on `LooseVersion` at all.

stonebig:
good idea, is it the following syntax ?
````
if "get_facecolor" in dir(ax):
````

bigreddot:
@stonebig that seems fine, or `hasattr` would work too. If Travis tests pass green and your local MPL testing works as expected I think we can merge this.

mattpap:
> `hasattr` would work too

@stonebig, this is the idiomatic approach.

bryevdv:
perfect will merge when green

bryevdv:
Thanks @stonebig !

PR #5183, opened 2016-09-19
Move palettes in bokehjs to bokeh-api bundle

### LINKED ISSUE #5182: Move JS palettes to bokeh-api.js
@mattpap It was suggested that the palettes are really only useful if using the BokehJS API, and that when using from python, the palette literal (list of colors) should always be sent fully, and never rely on the palette name. Since I don't think this file can be minified, it looks like this might shave off ~44k.

mattpap:
fixes #5182


bokehjs/src/coffee/models/glyphs/image.coffee
```diff
@@ -113,7 +115,7 @@ class Image extends Glyph.Model
       dw:           [ p.DistanceSpec     ]
       dh:           [ p.DistanceSpec     ]
       dilate:       [ p.Bool,      false ]
-      color_mapper: [ p.Instance,  () -> new LinearColorMapper.Model(palette: Greys.Greys9) ]
+      color_mapper: [ p.Instance,  () -> new LinearColorMapper.Model({palette: Greys9()}) ]
```
mattpap: This was a minor hiccup.

bokehjs/test/size.coffee
```diff
@@ -10,8 +10,8 @@ LIMITS = {
   "css/bokeh-widgets.min.css": 527
   "css/bokeh.min.css":         74
   "js/bokeh-widgets.min.js":   290
-  "js/bokeh-api.min.js":        40
-  "js/bokeh.min.js":           918
+  "js/bokeh-api.min.js":        70
+  "js/bokeh.min.js":           780
```
mattpap: Currently `bokeh.min.js` is at 757 kB and I'm still working on webgl split-off, which will result in at least -60 kB.

bryevdv:
LGTM

Issue #5154, opened 2016-09-13
Deprecate output_server


bryevdv:
`output_server` has a number of deficiencies relative to native apps or even `bokeh.client`. I propose to deprecate it and remove it entirely before `1.0`

ping @bokeh/dev

mattpap:
I presume it's possible to detect bokeh being run in a notebook? If so, we could remove `output_*()` functions altogether.

damianavila:
> I presume it's possible to detect bokeh being run in a notebook?

Not possible to know if bokeh is running from a notebook or another client... the kernel does not know anything about the frontend. Several frontends could be potentially "tied" to the same kernel which makes the mapping prone to errors. For instance, I can start a notebook which connect the web frontend with a specific kernel, then I attach a jupyter console to the same kernel, if I ask the kernel from where I am triggering the command, and we previously "linked" it to the web client, it would tell me that I am/running from the web client but I actually used the jupyter console...

bryevdv:
Right, the only way to "know" is to try and publish something and try to see if it fails or not. We in fact did just that a few years ago. Unfortunately it had some bad unintended side-effect (I don't remember the details) and had to be reverted in a hurry in the next release.

Also `output_file` and `output_notebook` are  used ubiquitously. Removing them would break mountains of code. But `output_server` I think we could nip in the bud if we do it soon.

bryevdv:
@bokeh/dev any more thoughts? There seems to be a sudden spike in questions around `output_server` so I am definitely inclined to proceed with a deprecation. `output_server` is basically useless except for static plots and leads people to confusion when callbacks "don't work."

mattpap:
@bryevdv, my understanding of server and its various entry points is insufficient for advice. Though I will vote +1 just for reducing the number of entry points.

bigreddot:
There ar.e no -1 votes, so I will make this a task

bryevdv:
No one chimed in to oppose, so I will make this a task

mattpap:
@bryevdv, solving #5199 I was wondering why callbacks aren't called as expected although everything is set up properly and data flows the way it should between client and server. Yet, to experiment on, I chose the only example that uses `output_server()`. Switching to another example (using `push_session()`) "fixed" the issue immediately.

bigreddot:
Yes, keeping `output_server` was a mistake. It's basically useless. There were so many things breaking/changing in `0.11` that I wanted to try and not add one more incompatibility, but it would have been better to make a clean break then.

PR #5187, opened 2016-09-19
Add deprecation warning if user supplies iterables and user-defined s…

### LINKED ISSUE #2056: deprecate glyph functions accepting datasource and sequence literals simultanously
Right now, when user passes datasource and passed in data on glyph functions [checks for a passed in source](https://github.com/bokeh/bokeh/blob/master/bokeh/_glyph_functions.py#L9) and [tries to figure out](https://github.com/bokeh/bokeh/blob/master/bokeh/_glyph_functions.py#L40) what the pile of parameters correspond to, given whatever glyph we are and then in that function it actually sets columns on the data source.

We should just disallow it. If you pass in a data source, you only get to refer to constant values, or column names and if you pass data in, then you should not specify datasource.

Maggie-M:
issues: fixes #2056

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -119,8 +119,13 @@ def _process_legend_kwargs(kwargs):
             # Otherwise just accept folks were being intentional
             kwargs['label'] = legend
 
+_GLYPH_SOURCE_MSG = """
+Supplying a user-defined data source AND iterable values to glyph methods is deprecated.
 
-def _process_sequence_literals(glyphclass, kwargs, source):
+See https://github.com/bokeh/bokeh/issues/2056 for more information.
+"""
+
+def _process_sequence_literals(glyphclass, kwargs, source, is_user_source):
     dataspecs = glyphclass.dataspecs_with_props()
     for var, val in kwargs.items():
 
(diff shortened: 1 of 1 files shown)


bryevdv:
@bokeh/dev I propose to deprecate this now in `0.12.3`, and raise a `RuntimeError` in the very near future (and add a test for the exception at that time)

bryevdv:
@bokeh/dev any comments on this before merging?

mattpap:
@bryevdv, if this is a deprecation, then it should use appropriate machinery we have in bokeh for this.

bigreddot:
@mattpap we have deprecation machinery for functions and modules, what machinery would cover this case?

bryevdv:
@mattpap do you mean this? 
```
warn(message, BokehDeprecationWarning, stacklevel=2)
```
? If so, sure that seems reasonable

mattpap:
@bryevdv, at minimum this, because `BokehDeprecationWarning` gets special treatment to be always visible. However, I thought we had a function for this `bokeh.util.deprecate`.

bigreddot:
@mattpap That's just a function decorator

@Maggie-M can you change the warning as shown above?

mattpap:
@bryevdv, unfortunately examples weren't updated, so there is some annoying text output (e.g. `plotting/server/hover`).

bigreddot:
Ah, I think that was the only one, thanks for the catch

bigreddot:
Just going to have @Maggie-M delete it as part of #5154 since it is the only `output_server` example as well

PR #5197, opened 2016-09-20
Add very crude HTML representation of models

### LINKED ISSUE #5164: Add support for _repr_html_ to HasProps and Model
`_repr_html_` was originally implemented on `Plot` to render a plot immediately without need for `show()`, at al. This didn't work well, so the code was commented out to be revised in future. Recently, `_repr_pretty_` was implemented to get insight into objects' properties. This works nicely across the board, but could work better in the notebook.

`_repr_html` should be added, so that it produces a shortened version of what `_repr_pretty_` produces, with a `...` button which would expand object's properties, showing more `...` buttons for embedded objects.

This is a followup on #5153.

mattpap:
By default this displays concise representation consisting of model's name, id and expand button. After expansion, it shows all model's properties and their values (formatted with `repr()`) and collapse button, which allows to return to concise representation. This is very crude, doesn't work recursively (thus `repr()`) and doesn't take too much advantage of HTML/CSS (colors, at al.). All this can be improved later, but for now we have something that works and doesn't interfere with notebook workflow. Note this is implemented in vanilla JS, so there are no assumptions about JQuery, etc. We shouldn't use bokehjs' JQuery, because HTML repr of a model may be displayed before `output_notebook()`. Also note that this makes original (commented out) `_repr_html_`, that displayed a plot automatically, obsolete and resolves that discussion as well.

fixes #5164

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -61,11 +61,12 @@ from importlib import import_module
 import inspect
 import numbers
 import re
+import sys
 import types
 from warnings import warn
 from operator import itemgetter
 
-from six import string_types, iteritems
+from six import string_types, iteritems, StringIO
 
 from ..util.dependencies import import_optional
 from ..util.future import with_metaclass
@@ -74,6 +75,7 @@ from .property_containers import PropertyValueList, PropertyValueDict, PropertyV
 from . import enums
 
 pd = import_optional('pandas')
+IPython = import_optional('IPython')
 
 def field(name):
     ''' Convenience function do explicitly mark a field specification for
@@ -918,12 +920,12 @@ class HasProps(with_metaclass(MetaHasProps, object)):
     def unapply_theme(self):
         self.apply_theme(property_values=dict())
 
-    def pprint_props(self, indent=0):
-        """ Prints the properties of this object, nicely formatted """
-        for key, value in self.properties_with_values().items():
-            print("%s%s: %r" % ("  "*indent, key, value))
+    def __str__(self):
+        return "%s(...)" % self.__class__.__name__
 
-    def _repr_pretty_(self, p, cycle):
+    __repr__ = __str__
+
+    def _bokeh_repr_pretty_(self, p, cycle):
         name = "%s.%s" % (self.__class__.__module__, self.__class__.__name__)
 
         if cycle:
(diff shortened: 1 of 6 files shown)


bryevdv:
what does this display in the console IPython shell?

mattpap:
@bryevdv, `_repr_pretty_` as before. This works in frontends that support `_html_repr_`, so pretty much only notebook, maybe qtconsole.

bigreddot:
Right, my point from the other PR is that `_repr_pretty_` outputs too much at the console, that has to be changed before the next release. Getting those printed representations should be easy, but not mandatory.

bryevdv:
Sorry, I have to log into a different account for client work, "bigreddot" is me in case it was not evident.

mattpap:
@bryevdv, yeah, it looked to me as someone is trying to mess with me. However, seeing empty profile and account created recently, shed some light on the matter.

mattpap:
@bryevdv, I thought the problem was with notebook workflow, not console. Note that our representation isn't that much chatty than a non-trivial numpy array, or even a list as IPython's pretty representation as are quite sparse.

bryevdv:
There are two issues but I thought from your previous comment that you'd handle both in this PR. 

In the extremely common case of `p.circle(x, y)` it spits out ~50 rows of info, often completely clearing everything that was previously on display in the console window. This is not reasonable, it makes the ipython console + bokeh unusable:

screen shot 2016-09-20 at 1 52 37 pm

Telling people to put a `;` at the end is also not reasonable. Many people don't even know that is valid python, it's not common in any sense, and is also not what has every been demonstrated for the past three years.

mattpap:
> Many people don't even know that is valid python (...)

Why is that? People coming from matlab, mathematica, etc. backgrounds should be very well familiar with this. Is this us not teaching properly newcomers? This is a feature and should be utilized when necessary. Yet this representation is in fact a bit chatty and it's unfortunate that IPython doesn't have a middle ground. Well, there is `verbose` option, but that's cumbersome and really verbose for any object.

I changed this as follows. The default is trivial repr with only `id` for models and nothing for `HasProps`. To get pretty repr you use `obj.pretty()` or `obj.pprint()`. The API is the same as IPython's and we still use all original code, just under different method names. Thus if we ever decide to show pretty repr directly, then it's a trivial change. If IPython is not available, `obj.pretty()` and `obj.pprint()` won't work and raise a `RuntimeException`.

bryevdv:
> Why is that?

I'm not sure. I just know in nearly 20 years of python it was something I didn't know for a long time, have almost never made use of, an have never seen demonstrated in any online classes or tutorials or materials teaching python. It's just not in the broader python community's collective consciousness. It's the sort of thing you'd only know about because someone specifically told you about it.

bryevdv:
This is terrific. I'm sure we can make it more sophisticated later but this is a great feature:

screen shot 2016-09-21 at 10 56 09 am

PR #5224, opened 2016-09-23
Fix legend merging broken

### LINKED ISSUE #5202: Figure legend not merging glyphs on the same data
So i have a legend created as follows:

```
    fig.line(x="Week of", y="% Under 14", source=source, line_width=2, legend="% Under 14 Days", line_color="blue")
    fig.circle(x="Week of", y="% Under 14", source=source, fill_color="white", size=8, legend="% Under 14 Days", line_color="blue")  
    fig.line(x="Week of", y="% Under 21", source=source, line_width=2, line_color="red", legend="% Under 21 Days")
    fig.circle(x="Week of", y="% Under 21", source=source, fill_color="white", size=8, legend="% Under 21 Days", line_color="red")  
```

Like in the example at http://bokeh.pydata.org/en/latest/docs/gallery/legend.html I would expect my legend to look something like:

image

However, my legend ends up looking like this:

image

Seems like a bug with the legend and the circles not being joined with the lines

birdsarah:
- [x] issues: fixes #5202
- [x] tests added / passed

This fixes #5202. @bryevdv also suggested a legend_mode that lets users explicitly decide the behavior they want.

I'm doing that in a separate PR as this one should be straight forward to merge for an RC. The other one might require futher API discussion.

This restores the behavior where a user sets the same string for multiple legend and auto-magically gets a compound legend glyph.

It also keeps the recently added behavior where if a label is a field on the column data source then the actual data will be used as the legend's label.

This leads to a potential conflict where a user defined label matches with the a column name, and they actually just want a string - and the explicit modes will provide this ability.

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -17,7 +17,7 @@ from ..models import (
     SaveTool, Range, Range1d, UndoTool, RedoTool, ResetTool, ResizeTool, Tool,
     WheelPanTool, WheelZoomTool, ColumnDataSource, GlyphRenderer)
 
-from ..core.properties import ColorSpec, Datetime, value
+from ..core.properties import ColorSpec, Datetime, value, field
 from ..util.deprecate import BokehDeprecationWarning
 from ..util.string import nice_join
 
@@ -107,18 +107,18 @@ def _process_legend_kwargs(kwargs):
     legend = kwargs.pop('legend', None)
     source = kwargs.get('source')
     if legend:
-        if isinstance(legend, string_types):
-            # Try and do something intelligent with a legend string:
-            # * value if it's not in column data source
-            # * field if it is
+        kwargs['label'] = legend
+
+    label = kwargs.get('label')
+    if label:
+        # Try and be smart
+        if isinstance(label, string_types):
+            # Do the simple thing first
+            kwargs['label'] = value(label)
+            # But if there's a source - try and do something smart
             if source and hasattr(source, 'column_names'):
-                if legend in source.column_names:
-                    kwargs['label'] = legend
-                else:
-                    kwargs['label'] = value(legend)
-        else:
-            # Otherwise just accept folks were being intentional
-            kwargs['label'] = legend
+                if label in source.column_names:
+                    kwargs['label'] = field(label)
 
 _GLYPH_SOURCE_MSG = """
 Supplying a user-defined data source AND iterable values to glyph methods is deprecated.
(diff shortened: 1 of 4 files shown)


birdsarah:
this is actually going to just need to be done and done right.

bokeh/plotting/helpers.py
```diff
@@ -170,7 +171,33 @@ def _update_legend(plot, glyph_renderer):
...
+    label = glyph_renderer.glyph.label
+
+    if plot._compound_legend is True:
+        compound_legend = True
+    else:
+        compound_legend = False
+        # But test if we now need to switch
+        if label.get('value'):
+            for r in plot.renderers:
+                if isinstance(r, GlyphRenderer) and hasattr(r.glyph, 'label') and r.glyph.label == label:
```
mattpap: `label` is defined on `Glyph`, so every specific glyph class has this property. `hasattr()` is unnecessary.

bokeh/plotting/helpers.py
```diff
@@ -170,7 +171,33 @@ def _update_legend(plot, glyph_renderer):
         legend = legends[0]
     else:
         raise RuntimeError("Plot %s configured with more than one legend renderer" % plot)
-    legend.legends.append(glyph_renderer)
+
+    label = glyph_renderer.glyph.label
+
+    if plot._compound_legend is True:
```
mattpap: `is True` is unnecessary. `bool` has only two values, so `if plot._compund_glyph:` is sufficient.

bokeh/plotting/helpers.py
```diff
@@ -426,7 +453,9 @@ def _glyph_function(glyphclass, extra_docs=None):
 
     def func(self, *args, **kwargs):
 
+        # Process legend kwargs and remove legend before we get going
```
mattpap: This comment doesn't introduce much information over function name. Perhaps this should be a part of the given function's docstring, if at all needed.

bokeh/plotting/helpers.py
```diff
@@ -162,6 +162,7 @@ def _make_glyph(glyphclass, kws, extra):
         return glyphclass(**kws)
 
 def _update_legend(plot, glyph_renderer):
+    # Get the plot's legend
```
mattpap: The next line says exactly this. Don't repeat yourself.

bokeh/plotting/helpers.py
birdsarah: Good point!

bokeh/plotting/helpers.py
birdsarah: thanks.

PR #5226, opened 2016-09-23
Canavandl/color bar tool

canavandl:
Do Not Merge 

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


banded_color_bar.coffee
```diff
@@ -0,0 +1,65 @@
...
+
+  _set_canvas_image: () ->
+    palette = @model.color_mapper.palette
+
+    if @model.orientation == 'vertical'
+      palette = palette.slice(0).reverse()
+
+    switch @model.orientation
+      when "vertical" then [w, h] = [1, palette.length]
+      when "horizontal" then [w, h] = [palette.length, 1]
```
mattpap: `switch` is an expression in coffeescript, so you can write `[w, h] = switch ...` instead. Not only this tells the intent better, it also would allow to use `const` keyword if such existed in coffeescript (or if we every switch to typescript, but then no `switch` expressions).

image_tool.py
```diff
@@ -0,0 +1,54 @@
...
+from bokeh.models.tools import Drag, DEFAULT_BOX_OVERLAY
+from bokeh.plotting import figure, show
+from bokeh.palettes import Viridis256
+
+### custom tool definition
+
+DEFAULT_COLOR_BAR = lambda: ColorBar(location=(0,0), color_mapper=LinearColorMapper(palette='Spectral9', low=0, high=10))
+
+class ColorBarTool(Drag):
+    __implementation__ = open('color_bar_tool.coffee').read()
```
mattpap: Since 0.12.3 you can use simply `__implementation__ = 'color_bar_tool.coffee'`. Oneliner ending in `.coffee` is interpreted as file path.

color_bar_tool.coffee
```diff
@@ -0,0 +1,164 @@
...
+
+    if dims.x0 <= e.bokeh.sx <= dims.x1
+      x_hit = true
+    else
+      x_hit = false
+
+    if dims.y0 <= e.bokeh.sy <= dims.y1
+      y_hit = true
+    else
+      y_hit = false
```
mattpap: That's pretty verbose, `y_hit = dims.y0 <= ...` would suffice. The same for `x_hit`.

color_bar_tool.coffee
```diff
@@ -0,0 +1,164 @@
...
+
+    did_hit = @_validate_point(e)
+
+    if not did_hit
+      return
+
+    dims = @_get_image_dims()
+
+    @_curpoint = [
+      @plot_view.canvas.sx_to_vx(dims.x1 - 3) # fudge factor, idk why necessary
```
mattpap: This ` - 3` could get some more explanation.

color_bar_tool.coffee
```diff
@@ -0,0 +1,164 @@
...
+
+  @define {
+      color_bar:                [ p.Instance, DEFAULT_COLOR_BAR   ]
+      overlay:                  [ p.Instance, DEFAULT_BOX_OVERLAY ]
+      image_color_mapper:       [ p.Instance                      ]
+    }
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
+    @plot.add_layout(@color_bar, 'right')
```
mattpap: I was hoping for removing `plot` attribute. This line may need some discussion as whether we want proceed this way.

color_bar_tool.coffee
```diff
@@ -0,0 +1,164 @@
...
+
+    @model.color_bar.band_start = start
+    @model.color_bar.band_end = end
+
+    @model.image_color_mapper.low = start
+    @model.image_color_mapper.high = end
+
+    @plot_view.request_render()
+
+    @model.overlay.update({left: null, right: null, top: null, bottom: null})
```
mattpap: I see this being repeated. Isn't there a `clear` function of some sort?

canavandl:
Will supercede as custom extension, not Bokeh core feature.

Issue #5228, opened 2016-09-24
Make legends great again


birdsarah:
Legends are a somewhat unloved corner of Bokeh. See #4526 for some property changes for more consistent API.

One of their odd features is the specification of Legend.legends as `[('label', [glyph_renderer_1, glyph_renderer_2])]`

This allows for a sort-of implicit compound glyph renderers (that is two glyphs representing the same piece of data) which is very handy.

However, it has other limitations and doesn't handle many cases well e.g. 
- Having multiple rows in your legend that represent different visual views on the same glyph_renderer e.g. categorical coloring in gapminder plot
- Being able to select which visual version of a glyph renderer (the hover, selection, non-selection) you wish to be your legend item
- Setting properties of the glyph / label per legend item (e.g. one is bigger or has a bigger font)

PR #5089 started trying to support case (1) from above, but in doing so broke the ability to implictly build compound legend items when using figure - See issue: #5202 

@bryevdv and I had a think about what the ideal solution is, and, recognizing that getting there is a large chunk of work that may not happen any time soon, came up with a compromise solution that lays a path towards ideal while moving us forward without breaking everything.

Worth throwing in a definition at this point. A "compound glyph renderer" is where you have two or more Glyphs with the same data source, and the same data properties (x, y) (they are visually representing the same data concept), but optionally different visual properties (different colors, sizes, etc).

---
### Ideal Case

``` python
class Legend(Annotations):
    items = List(LegendItem)
    # other legend props see #5202 for more info

class LegendItem(HasProps):
    # Defines one (if label is value) or many (if value is field and there are multiple values) rows of a legend
    label = StringSpec
    renderer = GlyphRenderer
    representative = Number  # Which version of your glyph is your representative point (may be an index of your data source)
    kind = Enum(GLyphKind)  # Do you use the visual styling of the "glyph", "hover", "selection", "nonselection", "decimated" - all of which are specified on GlyphRenderer
    # Also in future legend item visual props

class GlyphRenderer(Renderer):
    data_source = Instance(ColumnDataSource)
    glyphs = List(GlyphSpec)  # Always a list so GlyphRenderers can always be compound, but are often lists of length 1
    x_range_name = String
    y_range_name = String
    level = Enum

class GlyphSpec(Model):
    # A collection of Glyphs that together specify the different views on a glyph
    glyph = Instance(Glyph)
    hover = Instance(Glyph)
    selection = Instance(Glyph)
    nonselection = Instance(Glyph)
[...]

PR #5219, opened 2016-09-22
Restore original definition of API for markers using new paths

mattpap:
fixes #5218

PR #5229, opened 2016-09-25
Add LegendItem

### LINKED ISSUE #5202: Figure legend not merging glyphs on the same data

birdsarah:

- [x] issues: fixes #5202 #4526
- [x] tests added / passed

Still TODO:
- [x] Add validation on legend items #5228
- [x] Remaining deprecations as per #4526


birdsarah:
Have checked with JS API (only works after merging #5219). Here's the output

screen shot 2016-09-25 at 1 43 06 pm

bokeh/models/annotations.py
```diff
@@ -35,19 +44,69 @@ class Annotation(Renderer):
...
+
+    @validation.error(NON_MATCHING_DATA_SOURCES_ON_LEGEND_ITEM_RENDERERS)
+    def _check_data_sources_on_renderers(self):
+        if self.label and self.label.get('field'):
+            if len(self.renderers) < 1:
+                return str(self)
+            source = self.renderers[0].data_source
+            for r in self.renderers:
+                if r.data_source != source:
+                    return str(self)
```
bryevdv: I think this is equivalent and a bit simpler:

```
def _check_data_sources_on_renderers(self):
    if self.label and 'field' in self.label: 
        if len({ r.data_source for r in self.renderers }) != 1: 
            return str(self)
```

bokeh/plotting/helpers.py
```diff
@@ -83,65 +88,66 @@ def _pop_colors_and_alpha(glyphclass, kwargs, prefix="", default_alpha=1.0):
...
         # let any non-dataspecs do their own validation (e.g., line_dash properties)
-        if var not in dataspecs: continue
-
+        if var not in dataspecs:
+            continue
         # strings sequences are handled by the dataspec as-is
-        if isinstance(val, string_types): continue
-
+        if isinstance(val, string_types):
+            continue
```
bryevdv: I really prefer the old version, I find the whitespace between each pair of lines much more readable

examples/models/population_server.py
```diff
@@ -48,20 +49,24 @@ def pyramid():
     female_quad = Quad(left=0, right="female", bottom="groups", top="shifted", fill_color="#CFF09E")
     female_quad_glyph = plot.add_glyph(source_pyramid, female_quad)
 
-    plot.add_layout(Legend(legends=[("Male", [male_quad_glyph]), ("Female", [female_quad_glyph])]))
+    plot.add_layout(Legend(items=[
+        LegendItem(label="Male", renderers=[male_quad_glyph]),
+        LegendItem(label="Female", renderers=[female_quad_glyph])
+    ]))
```
bryevdv: What about adding one of those converters to the `items` property so that
```
plot.add_layout(Legend(items=[("Male", [male_quad_glyph]), ("Female", [female_quad_glyph])]))
```
works and converts to a list of `LegendItems` automatically?

sphinx/source/docs/user_guide/annotations.rst
```diff
@@ -78,6 +78,12 @@ if needed by passing a legend that is the column of the column data source.
 .. bokeh-plot:: source/docs/user_guide/source_examples/plotting_legends_by_source.py
     :source-position: above
 
+If you do not want this auto behavior. You can:
+ * use the ``field()`` or ``value()`` functions from :ref:`bokeh.core.properties` to
+   be explicit about your intentions
```
bryevdv: a concrete example always helps users

sphinx/source/docs/user_guide/annotations.rst
```diff
@@ -78,6 +78,12 @@ if needed by passing a legend that is the column of the column data source.
...
+ * not specify a legend argument and manually build a Legend yourself. There are examples of this
+   in the examples directory.
```
bryevdv: worth a link to one with `:bokeh-tree:`

bokehjs/src/coffee/models/annotations/legend_item.coffee
```diff
@@ -0,0 +1,74 @@
...
+  initialize: (attrs, options) ->
+    super(attrs, options)
+    # Validate data_sources match
+    data_source_validation = @_check_data_sources_on_renderers()
+    if not data_source_validation
+      logger.error("Non matching data sources on legend item renderers")
+    # Validate label in data_source
+    field_validation = @_check_field_label_on_data_source()
+    if not field_validation
+      logger.error("Bad column name on label: #{@label}")
```
bryevdv: do these raise exceptions?

bokeh/plotting/helpers.py
```diff
@@ -491,7 +525,8 @@ def func(self, *args, **kwargs):
     kwlines = []
     kws = glyphclass.properties() - set(glyphclass._args)
     for kw in sorted(kws):
-        if kw == "session": continue # TODO (bev) improve or remove
+        if kw == "session":
+            continue  # TODO (bev) improve or remove
```
bryevdv: I think these two lines are complete cruft

bokehjs/src/coffee/models/annotations/legend_item.coffee
birdsarah: No, they do a console.error - so you get the same kind of warning you will see python side - aggressive but doesn't stop execution.

bokeh/models/annotations.py
birdsarah: ok

birdsarah: thanks!

bokeh/plotting/helpers.py
birdsarah: lets give pep8 some love :D

bokeh/plotting/helpers.py
birdsarah: ok, i was just pep8-ing, this has nothing to do with this code. i'm not going to remove some code in this PR. Happy to do it in a follow-on though.

examples/models/population_server.py
birdsarah: I think that'd be great - can I follow on with another PR - so we can maybe get this out soon.

sphinx/source/docs/user_guide/annotations.rst
birdsarah: good point

sphinx/source/docs/user_guide/annotations.rst
birdsarah: ok

examples/models/population_server.py
birdsarah: I think there's a few variations on this which could make it quicker to spell / might be people's guess:

List on glyph_renderer optional:  
```python
items = [("Male", male_quad_glyph), ("Female", female_quad_glyph)]
```

Dictionary (a few ways)

```python
items = {"Male": male_quad_glyph, "Female", female_quad_glyph}
```

```python
items = {"Male": [male_quad_glyph], "Female", [female_quad_glyph]}
```

```python
items = [{"Male": [male_quad_glyph]}, {"Female", [female_quad_glyph]}]
```

bryevdv: Problem with dict is that it is unordered.  And I don't like the list of dicts because it seems much more complicated and easy to get wrong for no value. Also list of tuples is whats already there it's a smaller step for users to migrate.

bokeh/models/annotations.py
```diff
@@ -35,19 +44,64 @@ class Annotation(Renderer):
...
+            if len({r.data_source for r in self.renderers}) != 1:
+                return str(self)
+
+    @validation.error(BAD_COLUMN_NAME)
+    def _check_field_label_on_data_source(self):
+        if self.label and 'field' in self.label:
+            if len(self.renderers) < 1:
+                return str(self)
+            source = self.renderers[0].data_source
+            if self.label.get('field') not in source.column_names:
```
mattpap: It's not a problem with this PR, but whenever I see `prop.get('field')`, knowing that `prop` is `StringSpec` or similar, my head boils. We need to get rid of all those dictionaries that attempt to act like structs.

bokeh/models/annotations.py
```diff
@@ -103,33 +157,72 @@ class Legend(Annotation):
...
-    .. note::
-        The ``legends`` attribute may also be set from a dict or OrderedDict.
-        If a dict is used, the order of the legend entries is unspecified.
+    items = List(Instance(LegendItem), help="""
+    A list of legend items to be rendered in the legend.
+    """)
+
+    @property
+    def legends(self):
+        warnings.warn(DEP_MSG_0_12_3 % ('legends', 'Legend.items'))
```
mattpap: You have to use `BokehDeprecationWarning`. Otherwise most users won't see this. Apply in other instances as well.

bokeh/plotting/helpers.py
mattpap: > lets give pep8 some love

This may bring some hating into this PR ;)

bokehjs/src/coffee/models/annotations/legend_item.coffee
```diff
@@ -0,0 +1,74 @@
...
+    # Always return a list of the labels
+    if @label? and @label.value?
+      return [@label.value]
+    field = @get_field_from_label_prop()
+    if field?
+      if @renderers[0] and @renderers[0].data_source?
+        source = @renderers[0].data_source
+      else
+        return ["No source found"]
+      if source.get_column?
```
mattpap: Do `instanceof ColumnDataSource` instead (I presume this is the intent). If you leave this, next refactoring or "refactoring" (the term `refactoring` is misused all the time) around sources will be fun.

mattpap: This pretty much applies to all uses of `?` operator. If you can avoid it and take advantage of the object model, do it. It states the intent better and is easier on future changes. Also, if you can differentiate between `null` and `undefined`, do it as well (so use e.g. `!= null` instead of `?`). A lot of people in JS community blend those two.

bokehjs/src/coffee/models/annotations/legend_item.coffee
```diff
@@ -0,0 +1,74 @@
...
+        data = source.get_column(field)
+        if data
```
mattpap: On the contrary, here, `?` operator would be in order.

bokeh/plotting/helpers.py
mattpap: I strongly suggest limiting unrelated changes, because you may cause your self a lot of work in merge conflict resolution, and may bias reviewers to not focus on the subject of a PR (I know we shouldn't, but we are just people, usually with strong opinions).

birdsarah:
Can u provide me an example of this. I copied this pattern from Plot. I had assumed, probably wrongly, the __deprecated_attributes__ property did the BokehDeprecationWarning

Sarah Bird
sbird@continuum.io

mattpap:
@birdsarah, `__deprecated_attributes__` aren't about deprecation warnings, but internal bookkeeping in `HasProps`. You have to warn on your own, which you do, just instead of `warnings.warn(msg)` you issue `warnings.warn(msg, BokehDeprecationWarning)` (perhaps with `stacklevel=n` where `n` has to be figured out on case by case basis).

birdsarah:
Thanks so much @mattpap, and thanks for the quick turnaround on a review.

mattpap:
@birdsarah, after PR #5237 there will be only one way do deprecate things.

bryevdv:
@birdsarah I think this should go into `0.12.3` since we have waited, thoughts?

bryevdv:
@birdsarah I know you've got a lot going on with other work, I will try to merge master and address @mattpap couple of comments, I'd like to get this in for `0.12.3` later this week

bokeh/models/annotations.py
bryevdv: @mattpap more precisely, they are acting like unions. If you have a better proposal to support something like a union it is welcome.

mattpap:
Merge conflicts are trivial.

mattpap:
btw. If you want to quickly edit all conflicting files, then this may help (replace `vim -p` with your favorite editor):
```
[alias]
    resolve-merge-conflict = !sh -c 'vim -p $(git status -s | grep '^UU' | cut -c4-)
```
Add it to your `.gitconfig` and then you can issue `git resolve-merge-conflicts`.

bryevdv:
@mattpap that should be added to the dev guide, that's great. Yes the conflicts are trivial. I am fixing those, making the few changes you suggested, then I was going to add support for auto-converting the "list of tuples" in the property for `.items`

bryevdv:
@mattpap @birdsarah there are possibly some minor issues but I think this is ready for now with some followup issues:

* I had to change the `taylor_server.py` example, creating a new `LegendItem` with existing renderers inside was not deserializing correctly.  But I think that is possibly a weird case in general with server comms. In any case it's work-aroundable as seen. 

* I'm not sure about the JS side `LegendItem` logic. It's also work-aroundable. I had to add `value` when I didn't think I'd need to

bokehjs/test/models/annotations/legend_item.coffee
```diff
@@ -0,0 +1,150 @@
...
+
+  describe "get_field_from_label_prop", ->
+
+    it "should return undefined if label property is null", ->
+      legend_item = new LegendItem({'label': null})
+      field = legend_item.get_field_from_label_prop()
+      expect(field).to.be.undefined
+
+    it "should return undefined if label property is value", ->
+      legend_item = new LegendItem({'label': {'value': 'milk'}})
```
mattpap: This code was copied, so no blame here, but I would like to point out that it's preferred it use name literals instead of strings when referring to members of an interface, i.e. `new LegendItem({label: {value: 'milk'}})`. This states the intent better, because, after all, `label` and `value` are members of some interfaces and we aren't building an arbitrary dictionary here. Currently implicit, but when we will use TypeScript, it will be defined somewhere. Actually, it is already under `api/typings`. This is similar to what we do in Python when constructing keyword arguments with `dict(name=value)`, instead of `{"name": value}`.

bryevdv: @mattpap that makes sense, I can push updates to those as well. Do you have any other comments on this PR?

bokeh/models/annotations.py
```diff
@@ -35,19 +44,64 @@ class Annotation(Renderer):
...
+    def _check_field_label_on_data_source(self):
+        if self.label and 'field' in self.label:
+            if len(self.renderers) < 1:
+                return str(self)
+            source = self.renderers[0].data_source
+            if self.label.get('field') not in source.column_names:
+                return str(self)
+
+
+DEP_MSG_0_12_3 = """
```
mattpap: Deprecation warnings have to be changed the new style.

mattpap:
> It's also work-aroundable. I had to add value when I didn't think I'd need to.

Is this an issue with js examples or test?

bryevdv:
> Is this an issue with js examples or test?

Actually I think it just needs to be expicit for `models` examples. The nice convenience checking for field of value is part of the `plotting` API

mattpap:
LGTM. Any inconveniences will emerge sooner than later after merge.

birdsarah:
@bryevdv @mattpap, thanks for taking this PR the last mile - have looked at all the commits that @bryevdv made and they look good - I'd forgotten about "accepts" so that was much easier than I thought it was going to be.

@mattpap, really appreciating the explanations on PR comments I'm learning a lot and hopefully you won't have to repeat yourself in the future.

bryevdv:
OK I will merge this when green, I'll make issues for the other small questions I ran into.

bryevdv:
I can't get the py27 unit tests to pass, the port in use error seems persistent. I wish I knew what changed on travis or whatever else to start this problem.

mattpap:
@bryevdv, is this `5006` port in use error? I complained about this on internal mailing list. This error comes and goes, so you can restart a dozen times with no result and then after an hour or two tests will start to work again.

bryevdv:
Yes, it is the same you mentioned. What is odd is that it is always the same test (`test_log_status`)

bryevdv:
trying specifying a different port in `ManagedServerLoop` (which BTW is used all over the tests, its very strange that only this one test has issue)

bryevdv:
that just kicks the can, now a different test is failing the same way. It's always just one.

bryevdv:
It's also only python 2.7

bryevdv:
OK, `--rerun` "worked", will merge when green

Issue #4739, opened 2016-07-02
Put WebGL related code in CanvasPlot into a separate class.


almarklein:
See discussion in #4686. Some cleaning up was already done in the aforementioned PR, but it would be nice if PlotCanvas can stay agnostic about WebGL (except calling a functions in a few places).

bryevdv:
Ideally, I would like to split all the "webgl" code into a separate `bokeh-webgl.js` JS module. This would be included in the resources automatically when `webgl=True`, but it might also be the nice if things could attempt to degrade gracefully if `webgl=True` but somehow the JS for it is not present. The motivation is that BokehJS needs to go on a serious diet, and it's not really reasonable to demand users who are just drawing small (up to thousands) of things to pay for the extra payload all the time.

almarklein:
I agree.

bryevdv:
this was fixed with #5233 AFAIK just did not get closed automatically

bryevdv:
I'm going to close this one. The current larger goal is just to re-implement all rendering on top of some newer convenient webgl wrapper.

almarklein:
Or wait for [WebGPU](https://en.wikipedia.org/wiki/WebGPU) ...

bryevdv:
As much as I would delay love to restrict support to green-field APIs on only the newest browsers I think we will be stuck supporting relatively older platforms for quite some time. Ideally, someone would write a higher level library that can use fancy new things when they are available, and fall back to webGL when they are not. But I have no idea if anyone is planning such a thing.

almarklein:
I expect that there would be polyfills that target WebGL; The WebGPU is quite high-level (unlike, say Vulkan). The WebGPU API can actually be implemented with desktop GL (they plan to do that [here](https://github.com/gfx-rs/wgpu). WebGL (also 2.0) would be missing a few features, but probably close enough for most stuff...

bryevdv:
we will add it to the list of things to take a look at, this is 2020 work in any case

PR #5233, opened 2016-09-26
Split off webgl code into bokeh-gl bundle

### LINKED ISSUE #4321: Put WebGL functionality in separate bokeh-gl.js?

mattpap:
This reduces size of `bokeh.min.js` to 698 kB. This is based on PR #5189. This isn't as pretty as I expected, but it does the job, well hopefully, because I can't test it. I can only test that it falls back to 2D rendering if webgl is not available. @bokeh/dev, someone has to test this. The API had to be partially changed, because `Plot.webgl` had to be removed, but that didn't affect any examples, because they use `figure(webgl=True)` and this is left as it was. I tried to move as much code as possible to `bokeh-gl` bundle, but there are some leftovers (you can grep for `webgl`), specifically in `PlotCanvas` and `GlyphView`. It may be possible to improve this, but for now it seemed to be unnecessary. One big disadvantage of this PR is that you can't use webgl with gmap plots. This isn't an intrinsic limitation, but just I don't want to add webgl mirror classes for gmap\* classes. I think we should get rid of gmap plots and use tile layers. This would resolve the issue in the proper way.

fixes #4321, #4739


bokeh/plotting/figure.py
```diff
@@ -559,6 +559,16 @@ def scatter(self, *args, **kwargs):
...
+class WebGLFigure(Figure):
+    ''' A subclass of :class:`~bokeh.models.plots.Plot` that simplifies plot
+    creation with default axes, grids, tools, etc., and support rendering
+    with WebGL, if available. This requires addition of `bokeh-gl.min.js`
+    bundle.
+
+    '''
+
+    __subtype__ = "WebGLFigure"
+    __view_model__ = "WebGLPlot"
```
bryevdv: Should this be `bokeh.models.plots.WebGLPlot` instead of `WebGLFigure`? Thinking of anyone wanting to create a WebGL using `bokeh.models`

bryevdv:
I'm a little worried about the removal of `Plot.webgl` As it is I don't see how anyone would create a WebGL enabled plot using `bokeh.models` for instance?

bryevdv:
@mattpap is there any response to the question above about `Plot.webgl` ?

mattpap:
> is there any response to the question above about Plot.webgl ?

There is `WebGLPlot` for this purpose. `webgl` property can be (or may already be) supported in `figure()`.

bryevdv:
Good to know, this should wait for a future release at this point. FYI @birdsarah and I did have a conversation with someone at a conference recently who had done lots of webgl triangulation work, so there might be a lead on very fast patches with holes in the future.

bryevdv:
Moved `MIGRATION` to related issues.

bryevdv:
I think we should hold off on this and close the current PR.

PR #5237, opened 2016-09-27
Unify and simplify deprecation of things

### LINKED ISSUE #5236: Unify and simplify deprecation of things
I think I've seen all possible variations by now.

mattpap:
This replaces the old `bokeh.util.deprecate` (500+ loc) with `bokeh.util.deprecation` (30+ loc). There is now one function `deprecated()` that does the job. There are two signatures:

* `deprecated((x, y, z), old, new, extra=None)`

e.g. `deprecated((0, 12, 4), 'Plot.abc', 'Plot.xyz')`

* `deprecated(message)`

e.g. `deprecated("This needs custom deprecation.")`

Do NOT use `warnings.warn()` directly for this purpose, because it doesn't use `BokehDeprecationWarning`, which is required to always display deprecation warnings, and for consistency sake (previously we had over 10 variations on deprecation styles).

fixes #5236


bigreddot:
Just to be clear, the previous code allowed for deprecating entire modules, it doesn't seem like this does? That's probably OK but I want to make sure all the tradeoffs are know.

mattpap:
@bryevdv, yes, but this boiled down to showing a warning, so the new code does exactly the same thing. Just now you have one way of doing things.

PR #5243, opened 2016-09-28
Catch any exception in `import_optional`

### LINKED ISSUE #5242: import_optional isn't robust to all import failures
In my IIS web app trying to import IPython raises a ValueError.

`import_requires` only protects against an ImportError so trying to import `bokeh` results in an exception which crashes the application

```
Error occurred while reading WSGI handler:

  File "c:\python\envs\py-dev\lib\site-packages\wfastcgi.py", line 779, in main
    env, handler = read_wsgi_handler(response.physical_path)
  File "c:\python\envs\py-dev\lib\site-packages\wfastcgi.py", line 621, in read_wsgi_handler
    handler = get_wsgi_handler(os.getenv('WSGI_HANDLER'))
  File "c:\python\envs\py-dev\lib\site-packages\wfastcgi.py", line 589, in get_wsgi_handler
    handler = __import__(module_name, fromlist=[name_list[0][0]])
  File ".\quantweb.py", line 42, in <module>
    app = create_app()
  File ".\quantweb.py", line 8, in create_app
    from frontend import frontend
  File ".\frontend.py", line 4, in <module>
    from utils import handle_exception
  File ".\utils.py", line 3, in <module>
    import bokeh.plotting as plt
  File "C:\dev\code\bokeh\bokeh\plotting\__init__.py", line 2, in <module>
    from ..document import Document; Document
  File "C:\dev\code\bokeh\bokeh\document.py", line 36, in <module>
    from .model import Model
  File "C:\dev\code\bokeh\bokeh\model.py", line 13, in <module>
    from .core.properties import Any, HasProps, List, MetaHasProps, String
  File "C:\dev\code\bokeh\bokeh\core\properties.py", line 78, in <module>
    IPython = import_optional('IPython')
  File "C:\dev\code\bokeh\bokeh\util\dependencies.py", line 19, in import_optional
    return import_module(mod_name)
  File "c:\python\envs\py-dev\lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "c:\python\envs\py-dev\lib\site-packages\IPython\__init__.py", line 49, in <module>
    from .terminal.embed import embed
  File "c:\python\envs\py-dev\lib\site-packages\IPython\terminal\embed.py", line 16, in <module>
    from IPython.core.interactiveshell import DummyMod, InteractiveShell
  File "c:\python\envs\py-dev\lib\site-packages\IPython\core\interactiveshell.py", line 62, in <module>
    from IPython.utils import io
  File "c:\python\envs\py-dev\lib\site-packages\IPython\utils\io.py", line 92, in <module>
    stdin = IOStream(sys.stdin, fallback=devnull)
  File "c:\python\envs\py-dev\lib\site-packages\IPython\utils\io.py", line 41, in __init__
    for meth in filter(clone, dir(stream)):
  File "c:\python\envs\py-dev\lib\site-packages\IPython\utils\io.py", line 40, in clone
    return not hasattr(self, meth) and not meth.startswith('_')
  File "c:\python\envs\py-dev\lib\site-packages\IPython\utils\io.py", line 79, in closed
    return self.stream.closed
ValueError: underlying buffer has been detached


StdOut: 

StdErr: 
```

dhirschfeld:
Imports can trigger side effects which may result in any exception
potentially being thrown. In this case simply return None rather than
crashing on an optional import.

Fixes #5242

### DIFF
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -1,10 +1,8 @@
 ''' Utilities for checking dependencies
 
 '''
-from __future__ import print_function
 from importlib import import_module
 import logging
-import traceback
 
 
 logger = logging.getLogger(__name__)
@@ -29,7 +27,6 @@ def import_optional(mod_name):
     except Exception as e:
         msg = "Failed to import optional module `{}`".format(mod_name)
         logger.exception(msg)
-        print('\n'.join([msg, traceback.format_exc()]))
         
 
 def import_required(mod_name, error_msg):
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -27,7 +27,6 @@ def import_optional(mod_name):
     except Exception:
         msg = "Failed to import optional module `{}`".format(mod_name)
         logger.exception(msg)
-        
 
 def import_required(mod_name, error_msg):
     ''' Attempt to import a required dependency.
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -3,7 +3,11 @@
 '''
 from importlib import import_module
 import traceback
-import warnings
+import logging
+
+
+logger = logging.getLogger(__name__)
+
 
 def import_optional(mod_name):
     ''' Attempt to import an optional dependency.
@@ -20,9 +24,10 @@ def import_optional(mod_name):
     try:
         return import_module(mod_name)
     except ImportError:
-        return None
+        pass
     except Exception as exc:
-        warnings.warn(traceback.format_exc(), RuntimeWarning)
+        msg = "Failed to import optional module `{}`".format(mod_name)
+        logger.exception(msg)
         
 
 def import_required(mod_name, error_msg):
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -2,6 +2,8 @@
 
 '''
 from importlib import import_module
+import traceback
+import warnings
 
 def import_optional(mod_name):
     ''' Attempt to import an optional dependency.
@@ -17,8 +19,11 @@ def import_optional(mod_name):
     '''
     try:
         return import_module(mod_name)
-    except Exception:
+    except ImportError:
         return None
+    except Exception as exc:
+        warnings.warn(traceback.format_exc(), RuntimeWarning)
+        
 
 def import_required(mod_name, error_msg):
     ''' Attempt to import a required dependency.
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -17,7 +17,7 @@ def import_optional(mod_name):
     '''
     try:
         return import_module(mod_name)
-    except ImportError:
+    except Exception:
         return None
 
 def import_required(mod_name, error_msg):
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -1,8 +1,10 @@
 ''' Utilities for checking dependencies
 
 '''
+from __future__ import print_function
 from importlib import import_module
 import logging
+import traceback
 
 
 logger = logging.getLogger(__name__)
@@ -24,9 +26,10 @@ def import_optional(mod_name):
         return import_module(mod_name)
     except ImportError:
         pass
-    except Exception as exc:
+    except Exception as e:
         msg = "Failed to import optional module `{}`".format(mod_name)
         logger.exception(msg)
+        print('\n'.join([msg, traceback.format_exc()]))
         
 
 def import_required(mod_name, error_msg):
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -2,7 +2,6 @@
 
 '''
 from importlib import import_module
-import traceback
 import logging
 
 
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -24,7 +24,7 @@ def import_optional(mod_name):
         return import_module(mod_name)
     except ImportError:
         pass
-    except Exception as e:
+    except Exception:
         msg = "Failed to import optional module `{}`".format(mod_name)
         logger.exception(msg)
         


dhirschfeld:
NB: I believe the errors are unrelated to this change

mattpap:
Catching only `ImportError` makes sense, because other exceptions can give you valuable feedback about possible misconfiguration of your system.

bryevdv:
@mattpap @dhirschfeld how about this: `ImportError` passes silently, any other exception all continues but logs a message with the (unexpected) exception?

dhirschfeld:
Fine for me - that's what I've tried to implement here. Does bokeh have its own logging system I should have used instead?

mattpap:
@dhirschfeld, yes. It's being used like:
```
import logging
logger = logging.getLogger(__name__)
logger.warning('bla')
```
You initialize once at the top-level and then use it across module. I suggest `git grep logger bokeh/` to see this in action.

bokeh/util/dependencies.py
```diff
@@ -19,6 +21,9 @@ def import_optional(mod_name):
         return import_module(mod_name)
     except ImportError:
         return None
```
mattpap: This should be `pass` and then and the end `return None` once.

bryevdv:
@dhirschfeld agree with @mattpap small suggestions. Will you have time to add those changes in the next few days? If not do you mind one of us pushing to your branch?

dhirschfeld:
Sorry - spread a little thin at the moment. I'll have time tomorrow to push the changes but am happy for you to also push to my branch if that will speed things along...

bryevdv:
@dhirschfeld tomorrow would be great, I am shooting for `0.12.3` at the end of the week.

mattpap:
@dhirschfeld, alternatively you can grant us access to this particular branch (not entire fork) per new GitHub's feature https://github.com/blog/2247-improving-collaboration-with-forks. This way I could finish it and merge.

bokeh/util/dependencies.py
```diff
@@ -18,7 +23,11 @@ def import_optional(mod_name):
     try:
         return import_module(mod_name)
     except ImportError:
-        return None
+        pass
+    except Exception as exc:
+        msg = "Failed to import optional module `{}`".format(mod_name)
+        logger.exception(msg)
+
```
bryevdv: @dhirschfeld I think this needs to print the exception, since the notion is that that is an expected reason for failure (i.e. not an `ImportError`) also to be consistent with the rest of the codebase it would be `e` instead of `exc`

dhirschfeld:
You should be able to push if you want
image

I thought you'd use the logging in preference to a print statement? Happy to do both if that's desirable...

bryevdv:
@dhirschfeld sorry I was using "print" colloquially, I meant the log should include the exception

dhirschfeld:
Was that what you're after?

bryevdv:
@dhirschfeld sorry, now I was thinking of this:
```
msg = "Failed to import optional module '%s', reason: %s", (mod_name, e)
logger.exception(msg)
```

dhirschfeld:
Ah, right. The `log.exception` method will include the traceback automatically whcih is why I used `exception` rather than `warnign` - gives the user the info they need to debut the problem
```python
In [63]: try:
    ...:     1/0
    ...: except:
    ...:     log.exception("Failed to import module `{}`".format('ipython'))
fdsa - ERROR - Failed to import module `ipython`
  File "<ipython-input-63-9d5e31152022>", line 2, in <module>
    1/0
ZeroDivisionError: division by zero

In [64]: 
```

bryevdv:
Oh, I did not know that!

bryevdv:
In that case, just remove the print and we should be good, thanks for your patience @dhirschfeld

dhirschfeld:
np! :)

Should I squash it all into one? I tend to favour that myself, but happy to go with bokeh convention here...

bryevdv:
@dhirschfeld up to you, I will hit the "squash and merge" button in the end, either way :)

dhirschfeld:
I'll leave it as-is then since I've been working in the GitHub UI so can't easily do it from there...

bryevdv:
Looks great, will merge when green!

bryevdv:
@dhirschfeld sorry, at the finish line. There are some code quality checks failing. I tried to clone your fork and push to the PR branch, but GH always said permission denied. Here is a patch that fixes the complaints:
```
iff --git a/bokeh/util/dependencies.py b/bokeh/util/dependencies.py
index c02dcfc..a4c622c 100644
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -24,10 +24,9 @@ def import_optional(mod_name):
         return import_module(mod_name)
     except ImportError:
         pass
-    except Exception as e:
+    except Exception:
         msg = "Failed to import optional module `{}`".format(mod_name)
         logger.exception(msg)
-

 def import_required(mod_name, error_msg):
     ''' Attempt to import a required dependency.
```

bryevdv:
OK I think everything will pass now :)

bryevdv:
I will make a point to add more docs to the dev guide about our linter tests and how to run them locally too.

dhirschfeld:
Sorry, missed that. That's where using a proper IDE rather than the GitHub UI would have helped.

I always configure my editor to show white-space - would be nice if GitHub allowed you to do that too...

bryevdv:
> would be nice if GitHub allowed you to do that too...

+1! :)

dhirschfeld:
Regarding the unused variable - I disagree with the style checker on that one. Yes, unused variables are generally bad as they can clutter the code and obfuscate the intent but even though the exception object `e` was unused by the code it's there for the convenience of the person debugging the code. 

Storing the exception in a local variable allows the person debugging the code to introspect the exception (and message) which actually occurred - something which is even more important for a blanket catch-all `Exception` where it could have been anything at all.

For that reason my personal preference is to always use `except Exception as exc` regardless of whether the `exc` is then used by the code or not.

dhirschfeld:
Not sure why that's failing :| I've re-enabled the allow edits from maintainers checkbox so that might work if you need to push any fixes....

bryevdv:
We have some spurious failures unfortunately, I restarted. Thanks for the PR @dhirschfeld !

dhirschfeld:
Not a problem - we got there in the end! :)

PR #5247, opened 2016-09-29
Replace DatetimeTickFormatter.formats with individual fields

### LINKED ISSUE #5234: Plot not shown if DatetimeTickFormatter partially defined
A plot may not be shown if formats aren't provided for all time scales.

This is reproduced by the following code:

```python
from datetime import datetime as dt
from bokeh.models.formatters import DatetimeTickFormatter
from bokeh.plotting import Figure, output_file, show

output_file('bokeh.html')

p = Figure(title='Demo', x_axis_type='datetime')

x = [dt(2016, 1, 1, 0, 0, 0), dt(2016, 1, 1, 0, 0, 10)]
print(x[1].day, x[1].hour, x[1].minute, x[1].second)
y = [1, 2]

p.scatter(x=x, y=y)

p.xaxis.formatter = DatetimeTickFormatter(formats=dict(
    seconds=["%H:%M:%S"],
    # a plot is displayed if the following lines are *not* commented out
    # minsec=["%H:%M:%S"],
    # minutes=["%H:%M:%S"],
    # hourmin=["%H:%M:%S"],
)
)

show(p)
```

This has been tested with:

Python version      :  3.5.2 (default, Sep 25 2016, 19:32:42) 
Bokeh version       :  0.12.2

mattpap:
`formats` was deprecated and now `DatetimeTickFormatter` has all `format`'s keys as its properties. The second commit isn't strictly necessary, but shows usage of "expression statements" (switch and if) to produce more compact and, I think, more readable code.

fixes #5234

### DIFF
--- a/bokeh/models/formatters.py
+++ b/bokeh/models/formatters.py
@@ -12,7 +12,7 @@ from ..model import Model
 from ..core.properties import abstract
 from ..core.properties import (Bool, Int, String, Enum, Auto, List, Dict,
     Either, Instance)
-from ..core.enums import DatetimeUnits, RoundingFunction, NumeralLanguage
+from ..core.enums import RoundingFunction, NumeralLanguage
 from ..util.dependencies import import_required
 from ..util.deprecation import deprecated
 from ..util.compiler import nodejs_compile, CompilationError
--- a/bokeh/models/formatters.py
+++ b/bokeh/models/formatters.py
@@ -14,6 +14,7 @@ from ..core.properties import (Bool, Int, String, Enum, Auto, List, Dict,
     Either, Instance)
 from ..core.enums import DatetimeUnits, RoundingFunction, NumeralLanguage
 from ..util.dependencies import import_required
+from ..util.deprecation import deprecated
 from ..util.compiler import nodejs_compile, CompilationError
 
 @abstract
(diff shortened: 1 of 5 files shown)


bokeh/models/formatters.py
```diff
@@ -282,26 +283,7 @@ def from_coffeescript(cls, code, args={}):
...
-}
-
-class DatetimeTickFormatter(TickFormatter):
-    """ Display tick values from a continuous range as formatted
-    datetimes.
-
-    """
-
-    formats = Dict(Enum(DatetimeUnits), List(String), default=DEFAULT_DATETIME_FORMATS, help="""
+DATETIME_TICK_FORMATTER_HELP="""
```
mattpap: @bokeh/dev, I didn't update this text, so someone with better English writing capabilities could rewrite this to reflect changes in this PR.

bokeh/models/formatters.py
```diff
@@ -282,26 +283,7 @@ def from_coffeescript(cls, code, args={}):
...
-    'milliseconds': ['%3Nms', '%S.%3Ns'],
-    'seconds':      ['%Ss'],
-    'minsec':       [':%M:%S'],
-    'minutes':      [':%M', '%Mm'],
-    'hourmin':      ['%H:%M'],
-    'hours':        ['%Hh', '%H:%M'],
-    'days':         ['%m/%d', '%a%d'],
-    'months':       ['%m/%Y', '%b%y'],
-    'years':        ['%Y'],
-}
```
bryevdv: I think we have to leave this for now, with a deprecation. Otherwise people that use it currently will break immediately, rendering the other deprecation of `.formats` useless.

bokeh/models/formatters.py
bryevdv: I'll do this now.

bryevdv:
@mattpap I restored `DEFAULT_DATETIME_FORMATS` with a deprecation, and moved the very long datetime help string to the main class. Individual scale properties have much shorter help strings now, that are customized but send users to the main class docs for all the formats. I also added a docstring to `formats` python property getter so that a note about deprecation shows up in the sphinx docs to. I know this is not ideal but we can automate adding warnings to "deprecated property"  docstrings in another PR.

bryevdv:
Also this LGTM I am happy for this to go in immediately if you agree.

PR #5249, opened 2016-09-29
Add a polyfill for Math.log1p()

### LINKED ISSUE #5248: Add a polyfill for Math.log1p() that's not supported by IE
Also examples, like `plotting/file/texas` fail in phantomjs. The error is `undefined is not a function`. This was introduced in PR #4490.

mattpap:
fixes #5248


canavandl:
LGTM

PR #5261, opened 2016-10-01
Fix memory usage regression introduced in PR #5106

### LINKED ISSUE #5260: Plot updates cause heap to grow massively
So I've been working on updating plots from within the notebook and noticed that interaction really slows down after a lot of plot updates. I started profiling a little bit using the ``bokeh serve fourier_animated.py`` example on 0.12.3dev4 and found that the heap just keeps on growing, after about 30 seconds it had grown to 1 GB and the plot starts slowing down.

screen shot 2016-10-01 at 5 05 11 pm

It seems something is not being garbage collected. After a bit of investigation I found that there are a lot of LinearMapper instances that accumulate:

image

and there also seem to be a lot of change events:

image

Here's the heap allocation for the same example in 0.12.2:

screen shot 2016-10-01 at 5 13 34 pm

Bokeh 0.12.3dev4
Chrome Version 53.0.2785.116 (64-bit)
OS X 10.11.6

mattpap:
fixes #5260

### DIFF
--- a/bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
+++ b/bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
@@ -14,46 +14,17 @@ p = require "../../core/properties"
 class CartesianFrame extends LayoutCanvas.Model
   type: 'CartesianFrame'
 
-  @getters {
-    x_ranges: () -> @_get_ranges(@x_range, @extra_x_ranges)
-    y_ranges: () -> @_get_ranges(@y_range, @extra_y_ranges)
-
-    x_mappers: () -> @_get_mappers(@x_mapper_type, @x_ranges, @h_range)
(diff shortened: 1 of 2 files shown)


bryevdv:
Is it possible to add some kind of test to run one of the continuously updating examples for ~1 minute and monitor memory usage?

mattpap:
@bryevdv, sure it is and contrary to running time, this is comparable across builds, as long as hardware architecture is the same.

bryevdv:
@mattpap OK, well I was suggesting such a test should go in this PR. :) Can you make a separate issue to add a test for memory usage of server apps?

mattpap:
@bryevdv, yeah, I will start an issue for this, because it needs some thought how to do it right. Additionally the bug was pretty nasty, so I didn't want this PR to be open for awhile.

PR #5265, opened 2016-10-03
Split off Events and don't use Backbone.Model if not necessary

### LINKED ISSUE #5264: Split off Backbone.Events and don't depend on Backbone.Model if not necessary
Some classes inherit `Backbone.Model` whereas they could simply inherit `Backbone.Events`. Changing this reduces memory footprint of instances of those classes and improves understanding (i.e. I just want event handling, but not everything else that `Model` allows).

mattpap:
One of many PRs to clean up bokehjs. `Events` is the only part of of Backbone that we use without change. That's why I split it off. Backbone `Model` and `View` will likely merge with `HasProps` and `BokehView`.

fixes #5264


bryevdv:
@mattpap this needed to be discussed more. As it turns out there are needs to expose the events so that customJS callbacks and python callbacks can have access to them. That almost certainly means the events bus class will have to be full, serializable Bokeh model, and much of this will have to be un-done.

mattpap:
@bryevdv, I'm fully aware of PR #5276 and I don't mind making `UIEvents` a model. I had to merge this, so we could proceed with `common/` removal and then we can rebase #5276.

PR #5266, opened 2016-10-03
Clean up common/ and util/

### LINKED ISSUE #5263: Move common/* to core/* and merge util/ with core/util/
This duality is confusing and unproductive.

mattpap:
This is based on PR #5265. There is `get_indices()` in `utils/` left, because I wasn't sure what to do with it.

fixes #5263


bryevdv:
There are probably some people using `common/hittest.coffee`

bryevdv:
> There is get_indices() in utils/ left, because I wasn't sure what to do with it.

Can't it just go wherever, in `core/util` ?

bryevdv:
Or maybe it could be a method on column data source until sections get improved. I don't want to go through all this and still leave one thing hanging in `common`, though.

mattpap:
> Can't it just go wherever, in core/util ?

It can, it just I was hoping to remove it entirely (well replace), because it doesn't the current approach doesn't make sense. btw. `common/` is gone completely by now and we are talking about top-level `util/`.

bryevdv:
Sure, let's get rid of top level `util` too. Agreed it does not make much sense, but there is another entire issue to improve selections. It needs some thought and design up front, so for now this could just be moved.

bryevdv:
LGTM

PR #5272, opened 2016-10-03
Changed input to raw_input

### LINKED ISSUE #5271: docserver.py input causes a SyntaxError
On Windows, when using `make.bat serve`, the server can't be stopped - pressing `<ENTER>` gives a syntax error:
```
Press <ENTER> to exit...

  File "docserver.py", line 70, in <module>
    ui()
  File "docserver.py", line 57, in ui
    input("Press <ENTER> to exit...\n")
  File "<string>", line 0
```

Suggestion is changing this to raw_input rather than input,

djpugh:
- [x] issues: fixes #5271
- [x] tests added / passed - no tests for docserver.py
- [x] release document entry (if new feature or API change) - no new feature or API change

Output is now:
```
Press <ENTER> to exit...

Asked Server to shut down.
Server shut down.
```

### DIFF
--- a/sphinx/docserver.py
+++ b/sphinx/docserver.py
@@ -55,8 +55,10 @@ def shutdown_server():
 
 def ui():
     time.sleep(0.5)
-    input("Press <ENTER> to exit...\n")
-
+    try:
+        input("Press <ENTER> to exit...\n")
+    except (SyntaxError, KeyboardInterrupt) as e:
+        pass
 
 if __name__ == "__main__":
 
--- a/sphinx/docserver.py
+++ b/sphinx/docserver.py
@@ -57,7 +57,7 @@ def ui():
     time.sleep(0.5)
     try:
         input("Press <ENTER> to exit...\n")
-    except (SyntaxError, KeyboardInterrupt) as e:
+    except (SyntaxError, KeyboardInterrupt):
         pass
 
 if __name__ == "__main__":
--- a/sphinx/docserver.py
+++ b/sphinx/docserver.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from builtins import input
 
 import flask
 import os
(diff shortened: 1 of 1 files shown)


bryevdv:
IIRC `raw_input` is not in python 3, so this would have to go through `six` or similar. Honestly, I think I'd rather just remove the "keypress" check and let Ctrl-C be what exits, which I think would work on all platforms?

djpugh:
@bryevdv  - I've changed this back to import but changed it to import from builtins which should give python3 like functionality in python 2.7

djpugh:
@bryevdv - I have in the past had problems with fortran libraries intercepting the Ctrl-C and not failing as nicely on windows, the fix I've just pushed should give similar behaviour between python 2 and 3, but I'm happy to remove it and make it a Ctrl-C instead!

bryevdv:
@djpugh well my issue is that Ctrl-C is just what my fingers do naturally, and that actually does not work well anymore since `input` was introduced.  It makes the process hang, and then I have to kill it by hand at the command line. Is there a way to make both "press enter" and "ctrl-c" both work nicely?

djpugh:
@bryevdv I agree, I do tend to hit ctrl-C out of habit - this works for me on windows for both, catching the input and any syntax error or keyboard interrupt and returning from the blocking function. This works for me, but I don't have a python 3 environment to test on to hand.

bryevdv:
Let me try it out on my setup (probably not until tomorrow though)

bryevdv:
we are having a shortage of CI workers so I cancelled this build, just FYI. The change is not anything that is or affects the test. I will test this locally soon.

bryevdv:
OK, tested locally and `enter` and `Ctrl-C` both work on OS X / python 3! Thanks @djpugh !

PR #5274, opened 2016-10-03
Fix deprecation warnings

### LINKED ISSUE #5273: Fix deprecation warnings
It's painful to look through `examples.log` at travis ci for actual errors. There are lots of warnings about column lengths, not using `value()` for `text_font_size`, etc. or related to external libraries.

Also, if we deprecate something, lets make sure that we removed all occurrences before merging. This may not be trivial to do everything manually, but you can take advantage of travis ci. Just remove the thing you're deprecating or replace deprecation with hard error (throw an exception) for a moment, run tests and when everything is fixed, restore deprecation.

mattpap:
@bokeh/dev, please join and help out as time permits.

TODO:

- [x] ~~`text_font_size="10pt"` -> `text_font_size=value("10pt")`~~
- [x] `plot.title = "title"` -> `plot.title.text = "title"`

fixes #5273

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1796,15 +1796,14 @@ class FontSizeSpec(DataSpec):
     or a data source column name referring to column of font size data.
 
     '''
+    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(ex|px|cm|mm|in|pt|pc)$")
+
     def __init__(self, default, help=None):
         super(FontSizeSpec, self).__init__(List(String), default=default, help=help)
 
     def prepare_value(self, cls, name, value):
-        if isinstance(value, string_types):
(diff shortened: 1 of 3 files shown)


bryevdv:
@mattpap well, as I mentioned in the issue, I think maybe the font size thing should be un-deprecated, and things that look like font sizes should be treated as values by default. This was ostensibly deprecated a long time ago but the message/warning got lost and I think it has become too established now.

bryevdv:
`MIGRATION` to mention undeprecation

bryevdv:
LGTM will merge when green

bryevdv:
sigh... `ggplot_density.py` strikes yet again. I checked out the branch and ran it locally just to make sure it works (it does). Will go ahead and merge.

PR #5276, opened 2016-10-03
UIEvents model

### LINKED ISSUE #5278: Feature: emitting tool and ui events and attaching callbacks
The HoloViews team is interested in implementing some complex, custom user interactions for their Bokeh backend, in order to support things like dragging glyphs or hovering over elements to drive output in an adjacent plot.

The current issues they face include:
- Needing combinations of UI events that aren’t supported by Bokeh’s action/gesture/inspector tool UI event grouping (making custom tool models not a complete solution)
- Needing events to be sent during UI events that don’t emit events (e.g. having the inspector tools drive changes, e.g. via mouseover/hover interactions)

The current Bokeh callback system is based on emitting events on model property changes and supports both client side and bokeh server-mediated callbacks. Tools drive callback in this system by either changing the “selected” attribute of a ColumnDataSource (select tools) or by changing the range bounds of a plot (the zoom and pan tools)

Note: There are a couple of counter examples of this property change-driven pattern, including the callbacks emitted on SelectTool events (which don’t support bokeh server callbacks) and some of the input widgets (unclear to me why, but perhaps because they’re wrapped jquery elements). I would argue that the current select tool callbacks are ill-implemented, but perhaps evidence of user desire to customize such tool events.

Proposed is a system of UI events or tool events that can be configured to execute either client-side or bokeh server callbacks. Ideally, the coordinates of the the tool or UI event would be exposed (both in screen and data space coordinates?)

canavandl:
- [ ] issues: fixes #5278 
- [ ] tests added / passed

This PR is a partially an experiment in whether such UI events would accommodate what the custom interactions that the HoloViews team to hopes to implement. Further discussion of UI and tool events is taking place in #5278


mattpap:
This should have been based on PR #5266.

bryevdv:
> This should have been based on PR #5266.

The PRs are incompatible, so there should have been discussion. This one elevates the events bus to a full Bokeh model, whereas #5266 demotes it even from a backbone model.

mattpap:
@bryevdv, I will resolve conflicts when PR #5266 is merged.

bryevdv:
cancelled build, this is not going in for `0.12.3` and there is still some discussion about design to have.

canavandl:
The result of this experiment is:
* Exposing the UI Events in a general matter is useful to downstream libraries that use Bokeh as a visualization backend (e.g. Holoviews)
* Suggests that adding a real UI event message bus is a better path forward (similar to change event handling), then adding only client side callbacks (and not supporting UI events in server implementation.

I'm going to close this PR as rejected

PR #5258, opened 2016-09-30
Double ended sliders extension example

djpugh:
issue: #1996 - this PR adds a documentation extensions example implementing two variants of a double ended slider, which maybe a useful starting point for fixing #1996

PR #5280, opened 2016-10-04
Add support for Eco template in custom models

### LINKED ISSUE #5279: Extensions cannot use own `.eco` templates as compiler won't compile them
As mentioned in #5258, when creating an extension, the template cannot be a separate implementation and compiled, instead it must be updated in the JS/Coffescript implementation of the extension.
e.g.:

```
 ionslidertemplate = require "models/widgets/slidertemplate"
 
 # This model will actually need to render things, so we must provide
 # view. The LayoutDOM model has a view already, so we will start with that
 class IonRangeSliderView extends InputWidget.View
   tagName: "div"
   template: ionslidertemplate
```
And in the `initialize` function:
```
 html = @template(@model.attributes)
 # replace div slider to input slider
 html = html.replace('<div class="slider "','<input type="text" class="slider "')
 html = html.replace('<div class="bk-slider-horizontal">','<div class="bk-slider-horizontal bk-ion-slider">')
```
v0.12.2

It would be really helpful if these could be implemented in a `.eco` file to use as the template directly.

mattpap:
This should be used in PR #5258.

fixes #5279

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -154,6 +154,12 @@ class JavaScript(Inline):
     def lang(self):
         return "javascript"
 
+class Eco(Inline):
+
+    @property
+    def lang(self):
+        return "eco"
+
 class Less(Inline):
 
     @property
@@ -173,6 +179,8 @@ class FromFile(Implementation):
             return "coffeescript"
         if self.file.endswith(".js"):
             return "javascript"
+        if self.file.endswith(".eco"):
+            return "eco"
         if self.file.endswith((".css", ".less")):
             return "less"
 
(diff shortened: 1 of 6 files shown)


bryevdv:
There appears to be a real test failure: https://travis-ci.org/bokeh/bokeh/jobs/164946342#L1447

mattpap:
I don't understand this, I didn't even change anything to make it fail. There's too much magic. Looking at this, I'm not even sure how eco templates were loaded in tests, because there is no hook for that. There is for coffeescript though.

bryevdv:
Is it possible to add an example / tests?

bryevdv:
`ggplot_density.py` failing on this as well. I propose disabling this particular example test for now.  I believe something about the drastic recent TravisCI slowdown is bumping up against our timeouts.

mattpap:
> Is it possible to add an example / tests?

@bryevdv, I added tests for eco compilation and improved less' tests. #5258 will be a nice example.

bryevdv:
@mattpap OK, sounds good.

PR #5282, opened 2016-10-04
Change 0.12.{4->3} deprecation due to delayed release

### LINKED ISSUE #5250: Change 0.12.4 deprecations to 0.12.3 due to delayed release
I did some deprecation on the premise that it will be part of 0.12.4, so this needs to be updated.

mattpap:
fixes #5250


bryevdv:
This only failed due to spurious `ggplot_density.py` failure. I will go ahead and merge.

PR #5287, opened 2016-10-04
misc docs fixes, etc for 0.12.3

### LINKED ISSUE #4897: Subsections of User Guide/Adding Interactions are rendered twice when selected in site guide
For the 0.12.1 site documentation (and in current master), selecting a subsection of http://bokeh.pydata.org/en/latest/docs/user_guide/interaction.html, causes the site guide to render subsection twice:

screen shot 2016-07-28 at 2 10 49 pm

### LINKED ISSUE #4900: Reference Guide page titles are literal HTML string
For the 0.12.1 site docs (and in master), the page titles for pages in the reference guide are literal HTML strings that contain the desired title + some tags.

screen shot 2016-07-28 at 5 47 47 pm

### LINKED ISSUE #5107: bokeh.pydata.org warns that searching 0.12.1 is old but latest isn't pointing to 0.12.2 yet
Here's the behavior:
- go to http://bokeh.pydata.org
 - it will redirect to http://bokeh.pydata.org/en/latest  -- at least it does for me, perhaps we have some kind of locale detection so that 'en' isn't what everyone gets?
 - It may be important to note that when I got to http://bokeh.pydata.org/en/latest v0.12.1 was selected as the version in the dropdown 'navigation-menu'
- enter a search in the 'search-bar' class 'input' at the top right of the page
- you get taken to http://bokeh.pydata.org/en/latest/search.html?q= with your encoded query on the end
- even though we're on a 'latest' URL and v0.12.2 is selected now in the 'navigation-menu' the pink background with firebrick text div warning us about viewing old docs is displayed

image

I think that I might have caught things during a release just at the right time to get really confusing results out of the search box. If the window when this behavior could be observed by users is small, I'm okay with this being marked wontfix and moving on.  However, I'm not positive that this was just unlucky timing and thought I should at least file an issue.  Perhaps there's a bug in whatever injects the pink div into the top of the page results?

### LINKED ISSUE #5262: dev_guide/notes.rst wasn't updated in a year or more
Though it has some historical value, so maybe it could be renamed and preserved, because it shows the state of bokehjs in 2015 (`define()`, defaults, collections, etc.).

bryevdv:
issues: fixes #5107 fixes #5262 fixes #4900 fixes #4897


bryevdv:
I really preferred the "code" formatted titles for the reference guide, but sphinx is dumb about it and "fixing" it would be a giant affair.

bryevdv:
the recent work to make `bokeh.palettes` and "object module" broke the docstring and sphinx docs, this PR also restores that to working.

PR #5290, opened 2016-10-04
fix typo

### LINKED ISSUE #5288: Typo in the Legend Location docs, and why "right_center" instead of "center_right"?
In the current documentation for the legend Location, there sentence says, "For legends in the to the central layout area, such as those created automatically..." but should actually say "For legends in the to the central layout area, such as those created automatically...."  I am fixing this right now.

Additionally, for the location commands, why do 6 of them have the vertical followed by the horizontal alignment but 2 of them have the horizontal followed by the vertical?  This inconsistency seems like it would lead to confusion.

stvn66:
- [ ] issues: fixes #5288 
- [ ] tests added / passed

### DIFF
--- a/sphinx/source/docs/user_guide/styling.rst
+++ b/sphinx/source/docs/user_guide/styling.rst
@@ -707,7 +707,7 @@ property.
 Inside the Plot Area
 ''''''''''''''''''''
 
-For legends in the to the central layout area, such as those created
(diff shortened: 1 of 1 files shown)


bryevdv:
@stvn66 thanks for the correction! Can you open a separate issue about the other thing? perhaps the enum values can be deprecated and replaced.

PR #5291, opened 2016-10-04
docserver.py fix

djpugh:
@bryevdv - I just tested the fix in #5272 on a different machine, and it didn't have the `builtins` module, and I'm not sure where it came from, so I changed it to catch any exception and return, terminating the server.


djpugh:
This should work for python 2 and 3 ok

djpugh:
@bryevdv - I also moved the try before the sleep call as I managed to CTRL-C in/before the sleep call and froze it again...

bryevdv:
tested locally, both enter and ctrl-c work, merging now.

PR #5293, opened 2016-10-05
Bryanv/4926 aload static

### LINKED ISSUE #4926: Autoload_static seems to be broken in version 0.12
As of version 0.12 (tried 0.12.0 , 0.12.1 , and  0.12.2dev1), the autoload_static function seems to be broken. Autoload_static still outputs a script tag and associated javascript, but when placed into an html document, the plot is in the uppermost left corner and has no toolbar. Furthermore, if any text is in the html, the plot will overlay the text and any other plot. I had a rather large working document that works in version 0.11, but not 0.12. The simple example from the [documents](http://bokeh.pydata.org/en/latest/docs/user_guide/embed.html) shows the same error.

```python
from bokeh.resources import CDN
from bokeh.embed import autoload_static

plot = figure()
plot.circle([1,2], [3,4])

js, tag = autoload_static(plot, CDN, "some/path")
```

 Python gives no errors. JavaScript also doesn't seem to give any errors, it may be in the CSS?
[BokehError.txt](https://github.com/bokeh/bokeh/files/403632/BokehError.txt)

I am using python version 3.5 and bokeh versions 0.12.0 and 0.12.1 and 0.12.2dev1 (still works in 0.11.x).

bryevdv:
issues: fixes #4926

The problem was that the function that got called for embedding static plots was passed the `bk-root` div but then called `replaceWith`. Changing to `append` fixes. The custom server example was failing before. After, with `BOKEH_RESOURCES=server` it looks correct:
screen shot 2016-10-04 at 10 21 29 pm

### DIFF
--- a/bokeh/sphinxext/bokeh_plot.py
+++ b/bokeh/sphinxext/bokeh_plot.py
@@ -104,13 +104,6 @@ SOURCE_TEMPLATE = jinja2.Template(u"""
 """)
 
 
-SCRIPT_TEMPLATE = jinja2.Template(u"""
-<div class="bk-root">
-    {{ script|indent(4) }}
-</div>
-""")
-
-
 class bokeh_plot(nodes.General, nodes.Element):
     pass
 
@@ -299,8 +292,7 @@ def html_visit_bokeh_plot(self, node):
             with open(dest_path, "w") as f:
                 f.write(js)
 
-        html = SCRIPT_TEMPLATE.render(script=script)
-        self.body.append(html)
+        self.body.append(script)
     except Exception as e:
         raise SphinxError("Unable to generate Bokeh plot at %s:%d:%s" % (node['rst_source'], node['rst_lineno'], e))
     else:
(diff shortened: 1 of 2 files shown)


bryevdv:
So I'm pretty sure the `bokeh_plot.py` change is also needed, will have to cut a dev build and look at the deployed dev docs for sure.

bryevdv:
@birdsarah this seems to be failing one integration test for real: https://travis-ci.org/bokeh/bokeh/jobs/165131553

I have no clue what in this PR could be causing it tho, any pointers welcome.

PIng @canavandl  too

canavandl:
i assume it's related to #5294 . I'll have a fix shortly

bryevdv:
something else specific to this PR is causing the test to fail

bryevdv:
Ok I am going to go ahead and merge this.  I tested locally in a variety of scenarios.  If there are any problems tomorrow in testing we can just revert it.

PR #5295, opened 2016-10-05
Fix broken responsive layouts on window resize

### LINKED ISSUE #5294: responsive layouts broken in master
It appears that #5205 (Support passing suggested height/width values to document.resize method). Demonstrable with https://github.com/bokeh/bokeh/blob/master/examples/plotting/file/bar_chart.py (resizing the page after initial load causes top two plots to disappear).

canavandl:
In #5205 we added support for width/height argument to the document.resize method, so that we could suggest resize sizes in non-window layouts (i.e. jupyterlab (phosphorjs) panels)). This causes a regression where ``jquery.proxy`` would pass the jquery event as an argument to the resize function as the first arg (this didn't fail previous because JS is fun and ignores extra arguments to functions signatures)

This PR adds a positional argument to placehold the jquery event

- [x] issues: fixes #5294 
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -102,21 +102,24 @@ class Document
   solver: () ->
     @_solver
 
-  resize: (width=null, height=null) ->
+  resize: (jquery_event, width=null, height=null) ->
     # Notes on resizing (xx:yy means event yy on object xx):
     # window:event -> document.resize() -> solver:resize
     #   -> LayoutDOM.render()
     #   -> PlotCanvas.resize() -> solver:update_layout
 
+    # Note: JQuery.proxy passes the resize event as the first argument to the
+    # resize method. We don't currently use that argument for anything, but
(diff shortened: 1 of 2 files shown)


bokehjs/src/coffee/document.coffee
```diff
@@ -102,21 +102,24 @@ class Document
   solver: () ->
     @_solver
 
-  resize: (width=null, height=null) ->
+  resize: (jquery_event, width=null, height=null) ->
```
canavandl: if anybody has feedback on the advisability of the `jquery_event` placeholder arg, i'd appreciate it. It gets passed by JQuery.proxy and I don't know how to get around it.

bryevdv: @mattpap this PR was already merged, with the expectation to fix it up. Can you create and merge a PR for this immediately? @canavandl was also up very late on this and I'd like to get it in sooner than later.

PR #5300, opened 2016-10-05
Use () => @resize() instead of $.proxy(@resize, @)

mattpap:


bryevdv:
Great, will merge when green, thanks!

canavandl:
I pushed an update to the tests, the unit tests should pass now

Issue #5105, opened 2016-09-07
Create conda package for most recent node.js


mattpap:
This is continuation of PR #4946. At this point this isn't strictly necessary, but it would be nice to have a most recent version. Consider either 4.4.x LTS or 6.x. Do not use 5.x.

mattpap:
@bokeh/dev, what's the appropriate channel of communication for this? It shouldn't be us creating packages, as it is very resource inefficient.

damianavila:
I guess you can contact the anaconda team about it and ask for help. Pinging @csoja

csoja:
which version of nodejs do you need? 4.4.7?

mattpap:
@csoja, yes, 4.4.7 (most recent LTS version) is OK.

mingwandroid:
If anyone can assign this to me that'd be great. I guess it may not be possible as I'm part of the team? Moving it to another project (conda? Anaconda Distribution?) should also allow that.

bryevdv:
@mingwandroid Right, it won't let me assign. Can you make the issue linking to here, on one of those projects you are on and self-assign?

canavandl:
conda-forge has nodejs=4.5.0 (https://anaconda.org/conda-forge/nodejs) and I think we've already have conda-forge as an extra channel in our meta.yaml. does that work for you @mattpap ?

mattpap:
@canavandl, yeah, that works. We have `javascript` channel instead of `conda-forge`.

damianavila:
>@canavandl, yeah, that works. We have javascript channel instead of conda-forge.

I think conda-forge channel was added recently by https://github.com/bokeh/bokeh/commit/be6f820b48baf24eb885f57a02435df7657ed0c7 and https://github.com/bokeh/bokeh/commit/60b177518e6f6ecbbd56389b4dfa19fb6430ce3a#diff-bd134fa847b11c75aba174f834eb80dfR15

damianavila:
> I think conda-forge channel was added recently by be6f820 and 60b1775#diff-bd134fa847b11c75aba174f834eb80dfR15

This is something it should be audited later because of channel precedence we are installing a lot of packages from the conda-forge channel... at leat, that was what I saw when I last checked a travisCI run, but I did not have time to look further...

mattpap:
@damianavila, yes, I was on an slightly older branch when checking this.

[...]

PR #5307, opened 2016-10-05
Append conda channels to wrangle channel precendence

### LINKED ISSUE #5299: Use our own bokeh channel and avoid using conda-forge
Using conda-forge channel has raised some issue with timing on TravisCI.
We should avoid to add conda-forge and only get the packages we need from our own bokeh channel.
Previous ref comment: https://github.com/bokeh/bokeh/issues/5105#issuecomment-250470393

@bokeh/dev

canavandl:
- [x] issues: fixes #5299 #5105

This PR:
* Appends (instead of prepends via the --add arg) the bokeh and conda-forge channels so that our channel precedence (high to low) is "defaults, bokeh, conda-forge" instead of "conda-forge,bokeh,default".
  * This appears to reduce the CI unit test matrix times to 5:45-7:00min (as opposed to over 9min). It's unclear if the defaults CDN is faster or we're installing fewer dependencies. [ninja edit: it seems to be both]
* Also removes the (unused) javascript channel from our meta.yaml (we're now installing nodejs from conda-forge)

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -11,7 +11,6 @@ build:
 extra:
   channels:
     - bokeh
-    - javascript
     - conda-forge
 
 requirements:
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -20,9 +20,9 @@ python -V
 
 echo -e "$PINNED_PKGS" > /home/travis/miniconda/conda-meta/pinned
 
-conda config --remove channels defaults
 conda config --add channels conda-forge
 conda config --add channels bokeh
+conda config --remove channels defaults
 conda config --add channels defaults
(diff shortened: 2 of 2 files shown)


canavandl:
from https://travis-ci.org/bokeh/bokeh/jobs/165341830#L275

screen shot 2016-10-05 at 3 07 35 pm

canavandl:
and it's noticeably faster:

screen shot 2016-10-05 at 4 01 50 pm

bryevdv:
I'll take it!

bryevdv:
@bokeh/dev any objections to merging this today?

damianavila:
Sounds reasonable!

Issue #5304, opened 2016-10-05
app/surface3d doesn't work because custom model path is wrongly resolved


mattpap:
This is a peculiar issue. The problem is that custom model's path skips one directory and points to `examples/app/surface3d.pyc` instead of `examples/app/surface3d/surface3d.pyc` and thus coffeescript implementation cannot be found. Easy solution is to restore `open(path).read()`. This doesn't however resolve the bug and custom module compiler still thinks it's somewhere else. I think this is a problem with how `bokeh serve app` runs scripts and paths aren't set up properly, but I have to verify this.

mattpap:
Note that if changing to `open(path).read()`, then any `require()`s in `path` will fail anyway, but `surface3d` example doesn't use any.

bryevdv:
Spectrogram is failing for the same reason. This used to work, I was running spectrogram constantly at strata

bryevdv:
@mattpap OK the reason they worked in `0.12.2` is because they did `open(path)`

bryevdv:
OK I will make a PR with the "quick fix" and open a separate issue for addressing the wider problem in `0.12.4` and we will note the limitation in the release notes.

PR #5310, opened 2016-10-05
Fix Gapminder on Py2 / handle source codings better

### LINKED ISSUE #5302: bokeh serve --show app/gapminder doesn't work
Fails with `SyntaxError: encoding declaration in Unicode string`. Encoding line was added recently. Though the problem is with bokeh server, which should strip those lines before running a script.

bryevdv:
- [x] issues: fixes #5302
- [x] tests added / passed

@birdsarah @mattpap @canavandl @havocp would appreciate quick eyes on this. 

AFAICT the correct thing to do is to read the scripts in as bytes, then pass along any coding comment to `compile` to deal with, instead of using `codecs.open` to always read in `utf-8` (then we have to try and strip out coding comments which seems unreliable)

Additionally, gapminder data has unicode characters, so I added the encoding to `read_csv`. 

With this changes gapminder runs without error on py3 and py3

### DIFF
--- a/bokeh/application/handlers/script.py
+++ b/bokeh/application/handlers/script.py
@@ -1,7 +1,5 @@
 from __future__ import absolute_import, print_function
 
-import codecs
-
 from .code import CodeHandler
 
 class ScriptHandler(CodeHandler):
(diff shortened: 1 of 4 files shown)


bryevdv:
All the app examples with the excetion of the two mentioned in #5304 seem to run fine on py or py3 now

bryevdv:
It would be nice to test this on Windows though

canavandl:
LGTM

PR #5314, opened 2016-10-05
quick fix for custom model implementations in apps

### LINKED ISSUE #5304: app/surface3d doesn't work because custom model path is wrongly resolved

bryevdv:
issues: fixes #5304

### DIFF
--- a/examples/app/spectrogram/waterfall.py
+++ b/examples/app/spectrogram/waterfall.py
@@ -1,9 +1,11 @@
+from os.path import dirname, join
+
 from bokeh.core.properties import List, Float, Seq, Color, Int
 from bokeh.models import DataSource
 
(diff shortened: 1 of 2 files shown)


bryevdv:
Nothing in this is under TravisCI test, so merging immediately.

PR #5321, opened 2016-10-06
Use ES6 import/export instead of require() and module.exports

### LINKED ISSUE #5320: Use ES6 import/export syntax instead of require()
We will still use commonjs-style modules and browserify for bundling. What changes is that we won't be using `require()` and `module.exports` directly, but through `import` and `export` respectively. This is necessary to be able to switch to TypeScript at some point. This also is an opportunity to unify how we import and export things, which is a mess currently.

mattpap:
This takes advantage of import/export syntax in CoffeeScript and uses TypeScript compiler to generate `require()` calls and `module.exports`. CoffeeScript, by design, provides only syntax and leaves to the user how imports and exports will be translated into the chosen module system. Web browsers currently don't support modules. We use commonjs. I originally attempted to use Babel for translating import/export, but this was a disaster. However, TypeScript works pretty well. This approach gives us opportunity to integrate TypeScript modules together with CoffeeScript.

If you want to try this out, it builds as usually now. If you want to see errors reported by TypeScript compiler, you can add `--tsjs` to any build task (starting with `scripts:tsjs` and above). Note that this results in hundreds  of errors being printed on stdout. Some of them are real (yes, I already fixed a few bugs with this) and some are results of TypeScript not getting enough type information (not unexpectedly, as we don't provide almost any; almost because there are typings, so e.g. jquery or underscore are already typed).

fixes #5320


mattpap:
@bokeh/dev, this is ready.

mattpap:
As I didn't get any feedback yesterday, I went ahead and implemented the second stage of imports/exports rewrite in this PR (originally I planned to submit another one). This (mostly) finalizes my work in imports/exports. What changed in the second stage is that modules defining models (and views) now export full names instead of fixed `Model` and `View`. This way if you want to use a model and/or a view, previously you issued:
```
LayoutDOM = require "path/to/layout_dom"
new LayoutDOM.Model()
class SomeWidgetView extends LayoutDOM.View
module.exports = {View: SomeWidgetView}
```
Now you issue (using old syntax):
```
{LayoutDOM, LayoutDOMView} = require "path/to/layout_dom"
new LayoutDOM()
class SomeWidgetView extends LayoutDOMView
module.exports = {SomeWidgetView: SomeWidgetView}
```
And finally, using new syntax:
```
import {LayoutDOM, LayoutDOMView} from "path/to/layout_dom"
new LayoutDOM()
export class SomeWidgetView extends LayoutDOMView
```
Note that new syntax applies to bokehjs' source code and currently can't be used in custom models. I will be integrating TypeScript compiler into `compiler.js` bundle, but that's not ready yet.

mattpap:
By using TypeScript compiler to compile all of our source code, we can start taking some preliminary advantage of static analysis. For example, you can verify if imports are correct. Suppose you misspell an import like this:
```
$ git diff
diff --git a/models/widgets/widget.coffee b/models/widgets/widget.coffee
index c16c699..2f41d37 100644
--- a/bokehjs/src/coffee/models/widgets/widget.coffee
+++ b/bokehjs/src/coffee/models/widgets/widget.coffee
@@ -1,4 +1,4 @@
-import {LayoutDOM, LayoutDOMView} from "../layouts/layout_dom"
+import {ayoutDOM, LayoutDOMView} from "../layouts/layout_dom"
 
 export class WidgetView extends LayoutDOMView
   className: "bk-widget"
```
Then you can use `--tsjs` switch and TypeScripts' error code(s) to filter particular errors, e.g.:
```
$ gulp scripts:compile --tsjs=TS2305
[11:25:34] Using gulpfile ~/repos/bokeh/bokehjs/gulpfile.js
[11:25:34] Starting 'scripts:coffee'...
[11:25:34] Starting 'scripts:js'...
[11:25:34] Starting 'scripts:eco'...
[11:25:34] Starting 'scripts:ts'...
[11:25:34] Finished 'scripts:js' after 746 ms
[11:25:35] Finished 'scripts:eco' after 1.6 s
[11:25:36] Finished 'scripts:ts' after 2.42 s
[11:25:39] Finished 'scripts:coffee' after 5.75 s
[11:25:39] Starting 'scripts:tsjs'...
[11:25:44] build/js/tree_ts/models/widgets/widget.ts(5,3): error TS2305:
Module '"build/js/tree_ts/models/layouts/layout_dom"' has no exported member 'ayoutDOM'.
[11:25:45] Finished 'scripts:tsjs' after 5.91 s
[11:25:45] Starting 'scripts:compile'...
[11:25:45] Finished 'scripts:compile' after 30 μs
```
Currently you have to filter particular classes of errors, because the full output from TypeScript compiler contains hundreds of errors (as explained earlier), so it would be unlikely to get meaningful feedback otherwise.

bryevdv:
> As I didn't get any feedback yesterday 

we were in the middle of a difficult release yesterday. :)

bryevdv:
@mattpap GH won't display a diff this big, is there a way to break this into smaller chunks

bryevdv:
> export full names instead of fixed Model and View

The reason for exporting `Model` and `View` to have a consistent module interface for the JSON deserializer to know how to refer to the Model for a specified type (given as a string). How have you overcome this? Are yo using `instanceof` or something to automatically infer what inside the module is the `Model`, and what is the `View`? That's fine if so, I just want to understand what change allows this to work.

bryevdv:
@mattpap I notice you brought in the TS sources for Kiwi? I believe @birdsarah made some small customizations to the old vendored JS versions for improved logging, and possibly to expose one new flag to toggle a recompute in a certain situation. @birdsarah can you coment?

FWIW so far, the not having to explicitly, manually coordinate `module.exports` is great, checking for import errors is great.

mattpap:
> The reason for exporting Model and View to have a consistent, conventional module interface for the JSON deserializer to know how to refer to the Model for a specified type (given only a string type name). 

This wasn't that consistent, because there things like `module.export = SomeModel` and there were two ways of exporting multiple models from a module (e.g. arrows and cell formatters/editors). Now there is one way. Not that this couldn't have been fixed earlier, but with new syntax it's just natural.

Deserializer doesn't care about modules, it care about models, that's why providing "locations" in `base.coffee` was an understandable, but wrong idea. Now `base.coffee` is provided just with models and you should clearly see how much simpler things are there right now. This also makes `models/index.coffee` much simpler and there is no need for `api/models.coffee`.

mattpap:
> looks like you are handling this by constructing the index.coffee` files carefully to only export the models?

`models/index.coffee` is the public API to the deserializer. Everything else should be imported from exact locations anyway, which was a policy for a long time.

mattpap:
> I notice you brought in the TS sources for Kiwi?

Yeah, I saw it, but there were no API changes, so this doesn't conflict with typings, which btw. aren't picked up currently due to path conflict (`src/vendor` is a big pain). I also wasn't too happy about not knowing where this came from (development branch on  kiwi's repository wasn't my first guess).

bryevdv:
Maybe we could add a kiwi fork like numbro?

mattpap:
> GH won't display a diff this big, is there a way to break this into smaller chunks

I kept bulk, uninteresting changes separate from logic changes, so you can review particular commits. Commit titles are pretty self-explanatory (besides the first few ones, where I wasn't sure if this this going to work), so it should be easy to differentiate. It may be possible to split this into two PRs based on those commits, but it also may result in solving very many merge conflicts.

mattpap:
> Maybe we could add a kiwi fork like numbro?

Maybe, but this should be handled better than currently numbro. We should be publishing packages to npm under our organization and use those (I think npm finally supports this workflow).

bryevdv:
To this extent that I have been able to look over this, LGTM. @mattpap can you confirm that the existing image diff tests look as good as can be expected (we still have server and notebook issues there AFAIK) Can you test out all the custom extension examples in the `examples` folder and also in the user guide section? 

If all of that checks out I'm comfortable merging. This is a huge PR so we should clear it out of the way.

mattpap:
@bryevdv, I actually already did all that, so I reviewed image diff after every successful test run (there was one issue I had to fix), tested custom models after each change to the compiler and tested server examples locally.

bryevdv:
@mattpap great, I was just asking for the confirmation, since it was not stated. :) I'm ok to merge this then, if you'd like to.  Earlier is probably better than later.

Ping @bokeh/dev as a heads up.

bryevdv:
@mattpap FYI this broke the docs for extensions:
```
Unable to generate Bokeh plot at /Users/bryan/work/bokeh/sphinx/source/docs/user_guide/extensions_gallery/latex.rst:9:/Users/bryan/work/bokeh/sphinx/source/docs/user_guide/source_examples/extensions_example_latex.py:LatexLabel:2:1:reserved word 'import'
``` 

I'll make another issue. I think that `:bokeh-plot:` can be drastically simplified and made more robust using the `bokeh json` command line tool.

bryevdv:
Tho, the docs are building on travis. Not sure if its a py 2/3 thing or something else weird about my local install.

mattpap:
@bryevdv, did you run `npm install` after updating `master`?

bryevdv:
nope, that's it, thanks

PR #5327, opened 2016-10-07
fix deprecations

### LINKED ISSUE #5326: Examples with deprecation warnings
As of 0.12.3, the following examples have deprecation warnings

- plotting/file/hover.py
- plotting/file/tap.py

Maggie-M:
issues: fixes #5326

### DIFF
--- a/examples/plotting/file/tap.py
+++ b/examples/plotting/file/tap.py
@@ -1,4 +1,3 @@
-import itertools
 import numpy as np
 
 from bokeh.plotting import figure, show, output_file
--- a/examples/plotting/file/tap.py
+++ b/examples/plotting/file/tap.py
@@ -1,7 +1,7 @@
 import itertools
 import numpy as np
 
-from bokeh.plotting import ColumnDataSource, figure, show, output_file
+from bokeh.plotting import figure, show, output_file
 from bokeh.models import TapTool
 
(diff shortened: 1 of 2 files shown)


examples/plotting/file/tap.py
```diff
@@ -14,20 +14,11 @@
     "#%02x%02x%02x" % (int(r), int(g), 150) for r, g in zip(50+2*x, 30+2*y)
 ]
 
-source = ColumnDataSource(data=dict(
-    x=x,
-    y=y,
-    radius=radii,
-    colors=colors,
-    foo=list(itertools.permutations("abcdef"))[:N],
-    bar=np.random.normal(size=N),
```
mattpap: Those two shouldn't have been removed. They show that you can have extra fields besides what's allowed by the glyph. Though it would be nice to extend this to actually take advantage of those fields, e.g. in a hover tool or a callback.

bryevdv:
@mattpap I suggested removing them, they are duplicative of the hover example and add unnecessary complications in this context.

bryevdv:
@Maggie-M I think there are some unused imports to remove in the tap example

mattpap:
@bryevdv, indeed there is `hover` example that works exactly as I expected. OK, never mind my comment.

PR #5331, opened 2016-10-09
Replace JS' str.trim() with Python's str.strip()

### LINKED ISSUE #5330: syntax error in util/deprecation.py
line 24:
    message += " " + extra.trim()
results in error: AttributeError: 'str' object has no attribute 'trim'

it should be instead:
    message += " " + extra.strip()

that fixes the problem.  I needed that change to get the happiness demo to run

Helmut Strey

mattpap:
fixes #5330


bryevdv:
Let's add some tests to these functions

bryevdv:
Tests added, LGTM when green.

Issue #5318, opened 2016-10-06
Make figure accept Title instance


brendancol:
Right now the example below throws an error:
```
from bokeh.models import Title
from bokeh.io import show, output_file
from bokeh.plotting import figure

t = Title(text='hello')

output_file('title_example.py')
p = figure(title=t)
show(p)
```

birdsarah:
This is a bug. Line 42 and 50 of bokeh/plotting/figure.py

birdsarah:
@brendancol you may also want tot take a look and see if charts have the same problem.

bryevdv:
@brendancol FYI to get the issue to auto close the PR comment has to day "fixes #1234" or "closes #1234"

PR #5338, opened 2016-10-11
made change to fix figure constructor title object bug

brendancol:
https://github.com/bokeh/bokeh/issues/5318

This is still WIP until a test is added...


mattpap:
Maybe this could be handled at the property level, using `.accepts(...)`?

birdsarah:
@bryevdv and I discussed extensively the API for Title, unfortunately I've forgotten all of the nuances.

We agreed at the time that Plot would only accept Title objects, but you could still pass a string to figure and have it auto-populate a Title.

If I recall correctly the problem with having `Plot` accept a string was that it continued to blur the line of the fact that there is now a `Title` object and meant that folks would be more likely, understandable to, in their server callback code write 

```python
p = Plot(title="hello")

def update_title(self, attr, new):
    p.title = new

widget.on_change('value', update_title)
```
which is going to throw up an error.

So our compromise solution was to continue allowing `figure` to accept a string. It's an oversight that we don't allow figure to accept a Title object.

I'm ok with revisiting these assumptions and allowing Plot to accept a string. But I think it needs some careful thought as I do recall @bryevdv and I getting tripped up a little. In the meantime, I am +1 on getting figure accepting `Title` as, in my opinion, it's definitely a bug that it doesn't.

bryevdv:
I wouldn't handle this with `accepts` because then setting `p.title = "foo"` would always create a new `Title` object, and then people would wonder why updating an old one they have a reference to no longer works. This PR (just having `figure` accept both a string and a `Title` seems good to me

brendancol:
@birdsarah I'm a bit confused about the relationship figure / plot in relation to the Title Object.  I believe both accepts Title or String right now.  should the code below change?

```
class TitleProp(Either):

    def __init__(self, default=None, help=None):
        types = (Instance('bokeh.models.annotations.Title'), String)
        super(TitleProp, self).__init__(*types, default=default, help=help)

    def transform(self, value):
        if isinstance(value, str):
            from bokeh.models.annotations import Title
            deprecated("""Setting Plot property 'title' using a string was deprecated in 0.12.0,
            and will be removed. The title is now an object on Plot (which holds all of it's
            styling properties). Please use Plot.title.text instead.

            SERVER USERS: If you were using plot.title to have the server update the plot title
            in a callback, you MUST update to plot.title.text as the title object cannot currently
            be replaced after intialization.
            """)
            value = Title(text=value)
        return value
```

bryevdv:
`Plot` currently accepts a string but that is deprecated and will be removed in the future. I don't think there has been a discussion about when to remove the deprecated use-case.

tests/test_title_property.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import
+
+from bokeh.io import save
+from bokeh.plotting import figure
+from bokeh.models import Title
+
+
+def test_figure_title_should_accept_title_or_string():
```
birdsarah: This should be two separate tests. Ideally tests just test one thing and have a meaningful test name so it's super easy to pin down what went wrong. (I know this isn't true in a lot of our tests, but we can try).

bryevdv:
@brendancol is this `ready` now?

canavandl:
LGTM (will merge when tests pass)

bokeh/plotting/tests/test_figure.py
```diff
@@ -235,6 +237,19 @@ def test_title_kwarg_no_warning(recwarn):
     assert len(recwarn) == 0
 
 
+def test_figure_title_should_accept_title():
+    title = Title(text='Great Title')
+    plot = plt.figure(title=title)
+    l1 = plot.line([1, 2, 3], [1, 2, 3])
```
canavandl: the `l1` is causing flake errors. either add the #noqa or don't assign the renderer to anything

bryevdv: Let's minimize `#noqa` and leave it as a last resort only when necessary

brendancol:
@canavandl ok tests passing, good-to-go-now

bokeh/plotting/tests/test_figure.py
```diff
@@ -235,6 +237,19 @@ def test_title_kwarg_no_warning(recwarn):
     assert len(recwarn) == 0
 
 
+def test_figure_title_should_accept_title():
+    title = Title(text='Great Title')
+    plot = plt.figure(title=title)
+    plot.line([1, 2, 3], [1, 2, 3])
```
birdsarah: @brendancol for future reference. This line is unecessary and adds additional possible future failure modes for this test that don't relate to the thing you're trying to test.

PR #5344, opened 2016-10-12
Support special colors on image glyph

### LINKED ISSUE #5323: ColorMapper special colors are not respected for images
The (Continuous)ColorMapper classes allow defining special colors for the ``nan_color``, ``low_color`` and ``high_color``. However it seems like these are either not respected at all or in some cases even cause errors. Here is a minimal example, I've used named colors here but nothing changes if I use hex colors for example:

```python
import numpy as np
from bokeh.models import LinearColorMapper
from bokeh.palettes import viridis

output_notebook()

p = figure(plot_width=400, plot_height=400, x_range=(0, 10), y_range=(0, 10))

array = np.random.rand(10,10)
array[3:6, 3:6] = np.NaN
cmapper = LinearColorMapper(viridis(256), low=0.25, high=0.75, nan_color='blue', low_color='red', high_color='green')
p.image(image=[array], x=[0], y=[0], dw=[10], dh=[10], color_mapper=cmapper)

show(p)
```

bokeh_plot 20

When using an RGB(A) tuple for a color bokeh breaks completely and I have to refresh the page before any plot will render again. The JS console provides the following error message:

```
Uncaught TypeError: t.toLowerCase is not a function
```

Bokeh 0.12.3rc2
OSX 10.11.6
Chrome Version 53.0.2785.143 (64-bit)

canavandl:
- [x] issues: fixes #5323
- [x] tests added / passed

~~Todo: add tests~~

~~This PR is probably more proof that we need to outline our needs for the color mappers and then refactor.~~ was able to make relatively small changes

~~Note: only hex colors are now supported for Image glyph special colors (nan/high/low)~~ Resolved

### DIFF
--- a/bokeh/models/mappers.py
+++ b/bokeh/models/mappers.py
@@ -25,7 +25,7 @@ class ColorMapper(Model):
     any of the palettes shown in :ref:`bokeh.palettes`.
     """).accepts(Enum(Palette), lambda pal: getattr(palettes, pal))
 
-    nan_color = Color(default="#808080", help="""
+    nan_color = Color(default="gray", help="""
     Color to be used if data is NaN. Default: 'gray'
     """)
 
--- a/bokeh/models/mappers.py
+++ b/bokeh/models/mappers.py
@@ -25,7 +25,7 @@ class ColorMapper(Model):
     any of the palettes shown in :ref:`bokeh.palettes`.
     """).accepts(Enum(Palette), lambda pal: getattr(palettes, pal))
 
-    nan_color = Color(default="gray", help="""
+    nan_color = Color(default="#808080", help="""
     Color to be used if data is NaN. Default: 'gray'
     """)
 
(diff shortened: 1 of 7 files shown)


canavandl:
example from #5323 

code:
```
import numpy as np
from bokeh.models import LinearColorMapper
from bokeh.palettes import viridis

p = figure(plot_width=400, plot_height=400, x_range=(0, 10), y_range=(0, 10))

array = np.random.rand(10,10)
array[3:6, 3:6] = np.NaN
cmapper = LinearColorMapper(viridis(256), low=0.25, high=0.75, nan_color='blue', low_color='red', high_color='green')
p.image(image=[array], x=[0], y=[0], dw=[10], dh=[10], color_mapper=cmapper)

show(p)
```

screen shot 2016-10-12 at 12 35 04 pm

birdsarah:
Ran some examples locally and they look good to me.

bryevdv:
@canavandl this is just intended as a band-aid for `0.12.4` ? That flag is not a long-term fix.

canavandl:
Yes - a bandaid. I'll open a new issue to track resolving this better.

jbednar:
Just to be clear, is there currently any way that works for me to specify "clear" as a color, so that the background shows through where blue shows above?  Tuples are broken, and hex strings are limited to 6 characters, right?, so is there any other way to make this area transparent?

PR #5347, opened 2016-10-12
check for all valid CSS font size units

### LINKED ISSUE #5337: Charts and plots not rendering with user defined title text_font_size in em
As the title states, If I add a defined font size to a title on a plot or a chart, that plot and/or chart does not render in Bokeh 0.12.3. This is a pretty large issue. Reproducible below:

```
#!/usr/bin/python
# -*- coding: utf-8 -*-
from bokeh import plotting
import pandas as pd

data = [{"1": 10, "2": 4, "3": 3,
         "4": 5, "5": 15, "6": 7,
         "7": 5, "8": 10}]

source = pd.DataFrame(data)
source = ColumnDataSource(source)

fig = plotting.figure(width=1200, height=900, title="BL-55-T55555")

fig.vbar(top="1", x=1, width=0.5, fill_color="blue", line_color="black", source=source)
fig.vbar(top="2", x=2, width=0.5, fill_color="green", line_color="black", source=source)
fig.vbar(top="3", x=3, width=0.5, fill_color="yellow", line_color="black", source=source)
fig.vbar(top="4", x=4, width=0.5, fill_color="orange", line_color="black", source=source)
fig.vbar(top="5", x=5, width=0.5, fill_color="red", line_color="black", source=source)
fig.vbar(top="6", x=6, width=0.5, fill_color="purple", line_color="black", source=source)
fig.vbar(top="7", x=7, width=0.5, fill_color="grey", line_color="black", source=source)
fig.vbar(top="8", x=8, width=0.5, color="black", source=source)

fig.title.text_font_size = '1.9em'

curdoc().add_root(fig)

show(fig)
```

bryevdv:
- [x] issues: fixes #5337
- [x] tests added / passed

Took list of units from here:

http://www.w3schools.com/cssref/css_units.asp

Tested every unit by hand to see that:

* it rendered
* it changed size when the value changed

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1802,10 +1802,10 @@ class FontSizeSpec(DataSpec):
 
     A full list of all valid CSS length units can be found here:
 
-    http://www.w3schools.com/cssref/css_units.asp
+    https://drafts.csswg.org/css-values/#lengths
 
     '''
-    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%|rem|vh|vw|ch|em|ex|cm|mm|in|px|pt|pc)$")
+    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%|em|ex|ch|ic|rem|vw|vh|vi|vb|vmin|vmax|cm|mm|q|in|pc|pt|px)$", re.I)
 
     def __init__(self, default, help=None):
         super(FontSizeSpec, self).__init__(List(String), default=default, help=help)
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1805,7 +1805,7 @@ class FontSizeSpec(DataSpec):
     http://www.w3schools.com/cssref/css_units.asp
 
     '''
-    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%%|rem|vh|vw|ch|em|ex|cm|mm|in|px|pt|pc)$")
+    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%|rem|vh|vw|ch|em|ex|cm|mm|in|px|pt|pc)$")
 
     def __init__(self, default, help=None):
         super(FontSizeSpec, self).__init__(List(String), default=default, help=help)
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1796,7 +1796,7 @@ class FontSizeSpec(DataSpec):
     or a data source column name referring to column of font size data.
 
     '''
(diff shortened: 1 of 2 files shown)


bryevdv:
@bokeh/dev this is ready for review

canavandl:
LGTM - a crude for loop shows that all font units are canvas supported.

```

p = figure(x_range=(0,25), y_range=(0,25))

for i, unit in enumerate("%%|rem|vh|vw|ch|em|ex|cm|mm|in|px|pt|pc".split("|")):
    font_size = "5{}".format(unit)
    x = y = i * 1.5
    p.add_layout(Label(x=x, y=y, text=font_size, text_font_size=font_size))

show(p)
```

screen shot 2016-10-12 at 5 02 14 pm

bryevdv:
Hrm I think it should actually just be `%` not `%%` I was thinking it had to be escaped but that's only when used in a format string.

canavandl:
that fixed the "%" unit

screen shot 2016-10-12 at 5 30 12 pm

bokeh/core/properties.py
```diff
@@ -1795,8 +1795,17 @@ class FontSizeSpec(DataSpec):
...
+    interpreted as a CSS length, then the DataSpec is a value spec.
+    Otherwise, setting a string value will result in a field spec.
+
+    A full list of all valid CSS length units can be found here:
+
+    http://www.w3schools.com/cssref/css_units.asp
+
     '''
-    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(ex|px|cm|mm|in|pt|pc)$")
+    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%|rem|vh|vw|ch|em|ex|cm|mm|in|px|pt|pc)$")
```
birdsarah: It would be helpful if these were listed in alphabetical order.

birdsarah:
Found this reference too in case it helps:

https://developer.mozilla.org/en-US/docs/Web/CSS/length

bryevdv:
I think I prefer the `w3schools` link, tho I suppose the definitive source would behttps://drafts.csswg.org/css-values/#lengths

There are some additional units in there but they are awfully obscure. It did make me question, is capitalization important? your link has `q` but the CSSWG link has `Q`

birdsarah:
I think the CSS wg would be better.

Regarding case: https://drafts.csswg.org/css-values/#dimensions

Like keywords, unit identifiers are ASCII case-insensitive.

(i.e., [a-z] and [A-Z] are equivalent).

bryevdv:
Ok I will update the PR later tonight

bryevdv:
@birdsarah OK I updated to not be case sensitive. I put the units in the same order as the csswg tables, which is the most useful order for me.

birdsarah:
Cool!

Issue #3715, opened 2016-01-22
Interactively hide or show lines after plot finished by clicking (without re-ploting like plot browser feature in MATLAB)


bluehope:
It would be great to have a feature to interactively hide or show certain lines after the plot is being finished.

When many lines of data exist (5 or more),
plot once and just by clicking to show or hide, similarly in `plot browser` of MATLAB, feature will be very useful.
## 
### Example of interactively hide or show cetain lines in MATLAB.

<img width="938" alt="interactive_plot" src="https://cloud.githubusercontent.com/assets/1754257/12520835/1de725c0-c18a-11e5-8309-c0560f1784fd.png">
This is one of killing feature in MATLAB.
http://mathworks.com/help/matlab/ref/plotbrowser.html
### Example of interactively hide or show certain lines in Plotly

![plotly_example](https://cloud.githubusercontent.com/assets/1754257/12521526/71e03cee-c18e-11e5-95b6-6b60646edff3.gif)
## 
#### Plot all

``` python
p = figure(title="simple line example", x_axis_label='x', y_axis_label='y')
plot(x, y[1,:])
plot(x, y[2,:])
plot(x, y[3,:])
plot(x, y[4,:])
plot(x, y[5,:])
plot(x, y[6,:])
...
```
#### remove data 2,3,5,6 & re-plot to compare between data 1,4

``` python
p = figure(title="simple line example", x_axis_label='x', y_axis_label='y')
plot(x, y[1,:])
#plot(x, y[2,:])
#plot(x, y[3,:])
plot(x, y[4,:])
#plot(x, y[5,:])
#plot(x, y[6,:])
...
```
#### remove data 3,4,5 & re-plot to compare between data 1,2,6. Keep repeating for bunch of datas

``` python
p = figure(title="simple line example", x_axis_label='x', y_axis_label='y')
plot(x, y[1,:])
plot(x, y[2,:])
#plot(x, y[3,:])
#plot(x, y[4,:])
#plot(x, y[5,:])
plot(x, y[6,:])
...
```

keep repeating...

birdsarah:
Thanks for this request

It's a feature we've had requested a lot before and I believe there are a number of open issues that describe the pieces of the puzzle that would make this work. 

You have described two things that would be implemented separately. 

[...]

PR #5349, opened 2016-10-12
Add support for interactive legends

### LINKED ISSUE #2274: Interactive Legends
Placeholder for tracking, discussion, and eventual roamap inclusion regarding interactive legends.

Three wishlist items for legends that are common in other JS plotting libraries:
1) Legends off the plot area
2) Hover over legend item and respective glyphs highlight in plot
3) Click of legend item hides/shows the respective glyphs

Use cases include data identification for those with color blindness

Original discussion: https://groups.google.com/a/continuum.io/d/msg/bokeh/TymlMcwsZH0/Fs9Wpxle8SwJ

mattpap:
This allows for clickable legends. Try `examples/models/synthetic`, which was designed to outline issues with event handling and propagation, so it's perfect to test this. For example, currently clicking on a legend doesn't prevent from making selection if there's a glyph underneath.

fixes #2274, #3715


birdsarah:
@mattpap, i was happy with the functionality in this PR. Can we get this merged?

bokeh/models/annotations.py
```diff
@@ -120,10 +120,24 @@ class Legend(Annotation):
...
 
+    inactive_props = Include(FillProps, help="""
+    The %s for the legend background style when inactive.
+    """)
+
+    hover_props = Include(FillProps, help="""
+    The %s for the legend background style when hovered.
+    """)
+
+    click_policy = Enum("hide", "mute")
```
bryevdv: This can be documented better if it is a defined enum in the python enums module.

bokeh/models/renderers.py
```diff
@@ -145,6 +145,12 @@ def _check_bad_column_name(self):
     being hovered over by a HoverTool.
     """)
 
+    mute_glyph = Instance(Glyph, help="""
```
bryevdv: woudl prefer `muted_glyph`

mattpap: I chose `mute` over `muted` because the later doesn't follow the convention. Otherwise we would have `selected_glyph`, `hovered_glyph`, etc.

bryevdv:
I understand what you mean about moving some thing further away from tools. I think it's ok to merge for now, a larger rework should be a separate effort.

bryevdv:
@mattpap this is ready to do? Do you want to merge the fix in master in first, or just merge as-is?

bryevdv:
I went ahead and merged master, let's get it green.

bryevdv:
Hrm, after the merge one unit test fails:
```
E       AssertionError: Extra properties: click_policy, hover_fill_alpha, hover_fill_color, inactive_fill_alpha, inactive_fill_color
```

bryevdv:
I updated the test, should go green now.

bryevdv:
@birdsarah @mattpap OK after testing this out on some examples. I did want to bring one thing to attention tho. For the "original" style legends this works great. When the legend is a field (e.g. the gapminder app example) then currently clicking any legend row turns off *all* legend rows. This is not too surprising. So my question is just to confirm that that behavior is OK to start with (with a release note) and second, I think there should be another `click_policy` added: `"none"` and maybe this should actually be the default (i.e. preserve existing default behaviour?)

bryevdv:
@birdsarah the selenium errors are things like:
```
Message: window.get_x_range_end is not a function
```
Could this be a changed version of something causing test problems? (would be third one today!)

mattpap:
> (...) there should be another click_policy added: "none" (...)

I had the same thought today when debugging some user code.

mattpap:
> `Message: window.get_x_range_end is not a function`

This is how selenium tests work. Most likely there is some JavaScript error before that.

bryevdv:
> This is how selenium tests work. Most likely there is some JavaScript error before that.

OK I will leave this in your hands :)

mattpap:
I'm not sure if this is some sort of race condition, but locally those tests don't show any errors.

bryevdv:
Fixed the merge conflict, which was just the added `return null` from another PR. @mattpap please re-pull

birdsarah:
>  click_policy added: "none" and maybe this should actually be the default (i.e. preserve existing default behaviour?)

Yes - I think it's better, especially in the short term, if the default is the current behavior and you have the option of turning on the clickable legends.

mattpap:
I think this should be rescheduled for 0.14.5 and merged just after 0.14.4 release.

bryevdv:
I'm OK with that, @birdsarah would that impact your work?

examples/models/synthetic.py
```diff
@@ -0,0 +1,89 @@
...
+
+doc = Document()
+doc.add_root(plot)
+
+if __name__ == "__main__":
+    filename = "synthetic.html"
+    with open(filename, "w") as f:
+        f.write(file_html(doc, INLINE, "A synthetic example"))
+    print("Wrote %s" % filename)
+    view(filename)
```
bryevdv: This example is confusing with so many legends. I agree there is a reason for the "legends" example to have so many (to show off how/where to have legends) but in this example it will only confuse users. By far the common use case users will have is a single interactive legend, and so that is what we should demonstrate. It would be good to add a `plotting` example` too, maybe just adding an interactive legend to the stocks one or the iris one.

bryevdv: This example is also confusing because the legend "red circles" don't obviously correspond to some particular set of data on the plot. What people want is "click the blue line in the legend and the corresponding blue line in the plot disappears"

bryevdv:
@mattpap I only have a comment about the example(s). Also, IIRC this will not work with recent "grouped" renderer usage, correct? (i.e. when a legend is asked to be pulled from a column) If so that should be documented.

examples/models/synthetic.py
mattpap: @bryevdv, this is a "synthetic" example that's designed to explore bokeh's functionality. I expect this class of examples to turn into integration tests, but the infrastructure is broken, so this will have to wait. Perhaps there could be another directory added under `examples/`. A proper example showing off this new feature should be added or an existing modified.

mattpap:
> Also, IIRC this will not work with recent "grouped" renderer usage, correct?

I don't exactly follow how this feature works, so it may not work together with this PR.

mattpap:
> Actually there should definitely be a section added to the user's guide with an example.

I would be great if you could add this. Perhaps this and dealing with `synthetic.py` example could be done in a followup PR, so that we don't block other work.

bryevdv:
> I would be great if you could add this. Perhaps this and dealing with synthetic.py example could be done in a followup PR, so that we don't block other work.

I have a client call in a few hours. I will add a simple example under `plotting/file` and add user guide docs immediately after, then merge this this afternoon.

bokeh/plotting/helpers.py
```diff
@@ -552,15 +552,23 @@ def func(self, **kwargs):
         else:
             hglyph_ca = None
 
+        # handle the mute glyph, if any properties were given
+        if any(x.startswith('mute_') for x in kwargs):
+            mglyph_ca = _pop_colors_and_alpha(glyphclass, kwargs, prefix='mute_')
+        else:
+            mglyph_ca = None
+
```
bryevdv: @mattpap just a heads up I am going to change this to `muted_` when I push docs changes

bryevdv:
added new doc and example, will merge when green.

bryevdv:
@mattpap hover fill props are not actually used?

mattpap:
> hover fill props on legend are not actually used?

They were supposed to be used, but implementing hover for legends wasn't feasible at this point, because every change in hover would require redrawing a plot (event if not requested).

bryevdv:
So they should be removed for now?

birdsarah:
WOOOOOT :D

guillochon:
Glad this has been added! For me, since I am currently drawing points/errorbars as two separate sets of glyps, the hiding only half works, but I am guessing this functionality will work with the upcoming `FilledArea`/`Whisker`/`Stem` classes?

bryevdv:
@guillochon there's nothing built in. There might be a workaround using he new generic events, I'm not certain. It's a perfectly reasonable feature but it also never entered my mind, so I'd suggest explicitly calling it out in the error bars issue so that its taken into account there.

PR #5359, opened 2016-10-15
Add seaborn and vega color palettes

### LINKED ISSUE #5317: Add a colorblind and D3 palettes
I am interested in adding the seaborn color palettes, particularly the colorblind one.  I find these useful to load in then iterate through when plotting different curves.  Some of the palettes already included (Dark2, Set1, and Set2 from ColorBrewer) work reasonably well, but it would be good to have a better color blind option, and more options in general.  These are the 6 from seaborn I would like to add:

```
SEABORN_PALETTES = dict(
    deep=["#4C72B0", "#55A868", "#C44E52",
          "#8172B2", "#CCB974", "#64B5CD"],
    muted=["#4878CF", "#6ACC65", "#D65F5F",
           "#B47CC7", "#C4AD66", "#77BEDB"],
    pastel=["#92C6FF", "#97F0AA", "#FF9F9A",
            "#D0BBFF", "#FFFEA3", "#B0E0E6"],
    bright=["#003FFF", "#03ED3A", "#E8000B",
            "#8A2BE2", "#FFC400", "#00D7FF"],
    dark=["#001C7F", "#017517", "#8C0900",
          "#7600A1", "#B8860B", "#006374"],
    colorblind=["#0072B2", "#009E73", "#D55E00",
                "#CC79A7", "#F0E442", "#56B4E9"]
)
```

I have little experience with licensing, but looking at [seaborn BSD license](https://github.com/mwaskom/seaborn/blob/master/LICENSE) I believe we can use these as long as we copy the license in as a comment.  I have emailed Michael Waskom to confirm.

Thoughts?

stvn66:
I added the 6 qualitative seaborn and 4 categorical vega palettes.  I also cleaned up `bokeh/palettes.py`, grouping the code for all the new matplotlib palettes together and making the white space in dictionaries match.  A minor point, I noticed some long lines in the matplotlib dictionaries that were not flagged by the flake8 test (because of the comment at the end of the line?) but I did not wrap these as I was not sure if that was intended.

Which file generates [this doc page](http://bokeh.pydata.org/en/latest/docs/reference/palettes.html)?  I may still need to update that, unless it autogenerates from `./bokeh/palettes.py`.  I expect not because I noticed an extra `"` symbol above the BrBG plot.  

- [ ] issues: fixes #5317 
- [ ] tests added / passed

### DIFF
--- a/bokeh/palettes.py
+++ b/bokeh/palettes.py
@@ -1377,134 +1377,134 @@ class _Palettes(object):
     # vega category20b
     @property
     def Category20b_3(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf']
+        return ['#393b79', '#5254a3', '#6b6ecf']
     @property
     def Category20b_4(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede']
     @property
     def Category20b_5(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede', '￼#637939']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede', '#637939']
     @property
     def Category20b_6(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede', '￼#637939', '￼#8ca252']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede', '#637939', '#8ca252']
     @property
     def Category20b_7(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede', '￼#637939', '￼#8ca252', '￼#b5cf6b']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede', '#637939', '#8ca252', '#b5cf6b']
     @property
     def Category20b_8(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede', '￼#637939', '￼#8ca252', '￼#b5cf6b', '￼#cedb9c']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede', '#637939', '#8ca252', '#b5cf6b', '#cedb9c']
     @property
     def Category20b_9(self):
-        return ['￼#393b79', '￼#5254a3', '￼#6b6ecf', '￼#9c9ede', '￼#637939', '￼#8ca252', '￼#b5cf6b', '￼#cedb9c', '￼#8c6d31']
+        return ['#393b79', '#5254a3', '#6b6ecf', '#9c9ede', '#637939', '#8ca252', '#b5cf6b', '#cedb9c', '#8c6d31']
     @property
     def Category20b_10(self):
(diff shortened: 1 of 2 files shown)


stvn66:
I recognized that the [doc page](http://bokeh.pydata.org/en/latest/docs/reference/palettes.html) is the documentation in the header.  I'm not sure why there is a quote above the BrBG...

bokeh/palettes.py
```diff
@@ -450,10 +511,10 @@ def Greys9(self):
         return ["#000000", "#252525", "#525252", "#737373", "#969696", "#bdbdbd", "#d9d9d9", "#f0f0f0", "#ffffff"]
     @property
     def Greys10(self):
-        return ['#000000', '#1c1c1c', '#383838', '#555555', '#717171', '#8d8d8d', '#aaaaaa', '#c6c6c6', '#e2e2e2', '#ffffff']
+        return ["#000000", "#1c1c1c", "#383838", "#555555", "#717171", "#8d8d8d", "#aaaaaa", "#c6c6c6", "#e2e2e2", "#ffffff"]
```
stvn66: Several of the lists used single quotes while others used double quotes.  It was partially for consistency, and partially trying to make sure there were no extra quotes somewhere leading to the extra quote above the BrBg plot on the [docs page](http://bokeh.pydata.org/en/latest/docs/reference/palettes.html) (shown below).  

As there is no functional difference, I am fine leaving it the way it was.  Is there a preference one way or the other in the style guide used for Bokeh?

screen shot 2016-10-16 at 10 05 26 pm

stvn66:
In response to Gael Varoquaux's tweet about the poor quality of the Seaborn palettes, I had a look online for other colorblind palettes (a good colorblind friendly palette was my motivation the edits).  I found [this one](http://jfly.iam.u-tokyo.ac.jp/color/) used by [ggplot2](http://www.cookbook-r.com/Graphs/Colors_(ggplot2)/).  

Noticing how similar all the Seaborn palettes are to each other, and that the colorblind platte is only moderately good, what do people think if I take them out and put in the one ggplot2 uses?  Having the Seaborn palettes may appeal to Seaborn users coming to Bokeh.  Any idea how heavily they are used?

Is this better discussed on the issue page, #5317?

bokeh/palettes.py
stvn66: @mattpap, should I revert the commit changing quotes from single to double quotes?

bryevdv:
@stvn66 I like the code change with slices, but agree with @mattpap let's make that a separate follow-on PR. 

I'd personally say revert the quotes commit. It's not a big deal, but it woudl make this PR cleaner and smaller, and there is no standard across the project so I'm not sure that "cleaning up" this one file is a dramatic improvement to anything.

stvn66:
I will revise this to only bring in the vega catagorical and ggplot2 colorblind palettes (and leave the quotes as they were).

birdsarah:
@stvn66 don't forget to update the palettes in typescript API also

birdsarah:
https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/api/typings/palettes.d.ts

stvn66:
Will do, thanks for pointing this out.  When I added the qualitative brewer palettes there were other additions required but that file no longer exists.

mattpap:
The main thing to update in bokehjs is https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/api/palettes.ts. Though typings should be updated as well, just they are outdated, so more changes will be necessary.

PR #5367, opened 2016-10-17
fix errors in Document.destructively_move

### LINKED ISSUE #5333: Document._destructively_move() (in bokehjs) references undefined variable
I presume it should be `@_all_models` instead of `_all_models`. It looks like three is a big chunk of unused and untested code there.

bryevdv:
- [x] issues: fixes #5333
- [x] tests added / passed

In fixing the typo, I added a test, which uncovered other logic errors. Is this actually used anywhere? If not we can just remove it. 

ping @mattpap @havocp this is ready, comments welcome.

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -159,7 +159,7 @@ export class Document
     finally
       @_pop_all_models_freeze()
 
-  _destructively_move : (dest_doc) ->
+  destructively_move : (dest_doc) ->
     if dest_doc is @
       throw new Error("Attempted to overwrite a document with itself")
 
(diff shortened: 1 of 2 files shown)


bryevdv:
It's used by `replace_with_json` which *does* appear to be in use. But honestly then I'm confused. I couldn't get a simple call to `d2.destructively_move` to not throw an error about `@_all_models` not being empty until I simplified the logic to re-use `clear`

mattpap:
> It's used by replace_with_json which does appear to be in use. But honestly then I'm confused.

This was my observation as well when submitting this issue. It would be good to investigate branches where `replace_with_json()` is used, because there may be some leftover code we may want to remove.

bryevdv:
I do believe `replace_with_json` is used. So my best guess is that by sheer coincidence, it's only ever been used so far to initialize things, i.e. to replace already empty documents, and that this one case happened to pass through those error checks.

bryevdv:
I will try to verify this hypothesis tonight. If that's the case I will merge it later.

bryevdv:
OK that's not quite it either, since any usage of the original function would throw on the `_all_models`. 

Here is a more detailed account. `replace_with_json` is used in these places:

* [_repull_session_doc](https://github.com/bokeh/bokeh/blob/d2ce1971e4b984c7d30728b2378f7c959a3efd06/bokehjs/src/coffee/client.coffee#L244) But I expect the branch to replace an existing doc has never been taken, and that only the first branch that calls `Document.from_json(doc_json)` to make a fresh document has ever been used. 

@havocp any ideas about what could potentially need to trigger a document re-pull on an existing session? I'm not sure there's any actual way into this code path atm but am guessing this branch is "for completeness"

* [_handle_notebook_comms](https://github.com/bokeh/bokeh/blob/d2ce1971e4b984c7d30728b2378f7c959a3efd06/bokehjs/src/coffee/embed.coffee#L14) Again it's in one particular branch (of three) that has almost certainly  never been taken in actual practice. 

I believe the function now works under test. So, we could just merge this. But if there's some way re can reasonably rule-out ever needing to destructively move a document, I'd be OK removing it too. I'm just not sure.

bryevdv:
ping @mattpap any thoughts one way or the other? I'm inclined to just merge this as an improvement, and leave other refactoring or work to an intentional code audit.

mattpap:
@bryevdv, lets merge.

PR #5374, opened 2016-10-18
Drop broken Dialog widget

### LINKED ISSUE #4652: Dialog broken

mattpap:
Joining the "removing things" themed party.

fixes #4652


bryevdv:
It's a popular party (and I a good one).

bryevdv:
LGTM will merge when green

bryevdv:
ping @fpliger just as an FYI for this removal, I know this was used in some work a long time ago.

kongchujun:
So if I need a Dialog, what can I do ?

bryevdv:
@kongchujun That depends on the specifics of your situation, and the appropriate place to discuss it is not here, but on the [project Discourse](https://discourse.bokeh.org/)

PR #5390, opened 2016-10-20
Add tool argument

### LINKED ISSUE #5329: ability to remove tools from plot generated by mpl.to_bokeh
When a bokeh plot is developed from the mpl.to_bokeh function it comes with the default set of tools. There doesn't appear to be a mechanism to then remove tools and the function doesn't provide a means to give a parameter defining tools to include that are different than the defaults.

Here is some example code:

```
from bokeh import mpl
import numpy as np
import matplotlib.pyplot as plt

x = np.arange(0,10,0.1)
y = np.sin(x)

plt.plot(x,y)

fig = plt.gcf()

my_bokeh_fig = mpl.to_bokeh(fig=fig)
my_bokeh_fig.add_tools(HoverTool())
show(my_bokeh_fig)
```

and the generated figure:
screen shot 2016-10-07 at 3 36 36 pm

I just saw @birdsarah give a great talk at BIDS and she suggested that I raise this as an issue.

Maggie-M:
issues: fixes #5329

### DIFF
--- a/bokeh/core/compat/mpl.py
+++ b/bokeh/core/compat/mpl.py
@@ -22,7 +22,7 @@ from .bokeh_renderer import BokehRenderer
 # Classes and functions
 #-----------------------------------------------------------------------------
 
-def to_bokeh(fig=None, use_pandas=True, xkcd=False):
+def to_bokeh(fig=None, tools=None, use_pandas=True, xkcd=False):
     """ Uses bokeh to display a Matplotlib Figure.
 
     You can store a bokeh plot in a standalone HTML file, as a document in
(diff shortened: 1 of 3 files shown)


bokeh/core/compat/mpl.py
```diff
@@ -22,7 +22,7 @@
 # Classes and functions
 #-----------------------------------------------------------------------------
 
-def to_bokeh(fig=None, use_pandas=True, xkcd=False):
+def to_bokeh(fig=None, tools=None, use_pandas=True, xkcd=False):
```
mattpap: For backwards compatibility the default should be `DEFAULT_TOOlS`.

Issue #5315, opened 2016-10-05
extension implementation load path problems in apps


bryevdv:
see ref: #5304

bryevdv:
@mattpap I believe the root cause of this has to do with the cwd is set when the app is run. It works for "directory style" apps, but I am fairly certain it does not work well with single file apps.

mattpap:
@bryevdv, it's "directory style" apps that don't work in particular and the problem is in cwd (and some other factors, including Python's simplicty). 

In `CodeRunner.run()` we have the following code:
```
os.chdir(dirname(self._path))
sys.path.insert(0, '')
```
This makes manual setting of `__imlementation__` work, but not logic in custom module compiler. The reason is the custom models are resolved at later time, when those two lines aren't in effect anymore, and because Python graciously resolves module paths as relative if a module is run from the current directory and it's path is on `sys.path` (both conditions are met here) and as absolute otherwise. Don't know why anyone would invent such a thing, besides making things aesthetically pleasing, but this was fixed in 3.4. The only non-obtrusive solution I found, is to replace those two lines with:
```
sys.path.insert(0, os.path.dirname(self._path))
```  
where `self._path` is already absolute. This makes both:
```
__implementation__ = open(join(dirname(__file__), "surface3d.coffee")).read()
```
and
```
__implementation__ = "surface3d.coffee"
```
work at the same time. Obviously current directory is now set to wherever `bin/bokeh` was run from. I didn't yet fully analyzed this, but this may not have any negative effect, because we uniformly used `__file__` to derive location of auxiliary files.

bryevdv:
Migration note: we will no longer `os.chdir` into app directories, but `__file__` will have correct path that can be used with `open`, etc. (which we already demonstrate, but let's underscore this more)

Issue #5101, opened 2016-09-06
autoload_static JS resource loading deficiencies / compiler caching


canavandl:
Using autoload static for examples in https://github.com/bokeh/bokeh.github.io/blob/source/content/blog/release-0-12/contents.lr , we noted two issues:
1.  In a page with two plots, if the second plot requires resources that aren't loaded by the first plot, the resources aren't ever loaded. I'd guess it's because the autoload template checks to see if Bokeh is already loaded, then returns if it's there.
2.  The autoload_static output appears to load widgets and compiler resources when unneeded.

bryevdv:
@mattpap There are some additional possible deficiencies. Maybe these should have their own issues, but I am trying to record them so as not to lose them:

* the `__implementation__` path bug where it looks for `.coffee` files one directory too high (could not find open issue for this)

* adding custom models anywhere in a process adds to `Model.model_class_reverse_map` Maybe this is fine as it is, but e.g. it's causing a ***huge*** slowdown in our docs build. After all the "extensions" plots are generated, every subsequent plot takes ~5-6 seconds (!!!) The ones previous took ~0.01 seconds. My guess is every call to `autoload_server` is triggering a recompile of all custom models in `model_class_reverse_map` we need to find a way to make it only compile models that are actually  used, or to cache the compilation results.

All of these are fairly high priority after @birdsarah's needs. `0.12.4` for me is all about stability and bugfixes and these are pretty in need of attention.

bryevdv:
Some random ideas that might most or solve all of these: split js and css resources lists into two parts, the core bokeh libraries, and everything else. The core libraries look to see if they are loaded already, and only inject if they are missing. The injection of "everything else" happens unconditionally. Then the last bit is to make the "everything" else specific to what is actually being embedded, not just pull out every last custom model. Also I do think we want to not re-compile every single time.

bryevdv:
Ultimately this means that resources need to be a function of the thing that is being shown. There are two fairly unpleasant ways I could think of to make things work:
```
resources.configure_for(thing)

resources.js_files # tailored for `thing`

resources.configure_for(None)

resources.js_files # current behaviour
```
This makes `Resources` stateful, which sucks, alot. Alternatively we make `js_files` and `css_files` real functions, instead of properties, but that would be a bad breaking change. I guess a compromise could be functions `js_files_for(thing)`, etc.

bryevdv:
[...]

PR #5399, opened 2016-10-24
Improvements to bokeh-plot directive

bryevdv:
[This is WIP]

The `bokeh-plot` directive in `bokeh.sphinxext` has been somewhat rickety for some time. This PR makes some cleanup and improvements:

* keeps track of out-of-date `.py` examples file in a robust way using built-in Sphinx machinery
* re-uses code runner from server instead of customized redundant code runner and fragile custom monkeypatching
* generates a dedicated page for every `.py` example, in addition to wherever the `bokeh-plot` embed is
* initial docstrings in `.py` examples are rendered as ReST above the example in the dedicated page. Can be used to add additional indexing, information, etc. 
* Better sphinx build output
* moves `source_examples` to just `examples` subdir (The vast majority of files changes just reflect this renaming)

Please note: This PR adds a custom `Parser` for `*.py` files. *Any* `.py` files in the Sphinx source dir will be treated as examples to run and build. They should really also always go in an `examples` subdirectory as a convention. 

A few issues to note:
* The source read order had to be controlled, to read  all the `.py` files first. 
* except the "extension" examples, which are currently made last to avoid a giant build slowdown (see https://github.com/bokeh/bokeh/issues/5101#issuecomment-254684325)
* currently assumes `curdoc().add_root` is called, but `show` and `save` are automatically patched to do this. Examples that create and manage their own `Document` don't work (there were only two) 

I've also had to disable the extensions widget example until https://github.com/bokeh/bokeh/issues/5315 is fixed 

Final note: this stanza had to be added to the `server` block in the nginx config for the docs:
```
        location /scripts {
            if ($http_referer ~* (.*)en\/(.*?)\/(.*)) {
                set $v $2;
                rewrite ^(.*)\/(.*).js$ http://bokeh.pydata.org/en/$v/scripts/$2.js redirect;
            }
        }
```
This allows the autoload scripts to have an "absolute" location `/scrips/bokeh-plot-foo.js` etc. but that is correctly prefixed by the version that it was referred from. There might be a better way to configure nginx (separate server block for every version?) that would handle this more automatically, without having to inspect $http_referer but this will do for now.


sphinx/main_gallery.json
```diff
@@ -20,8 +20,6 @@
         { "path": "examples/plotting/file/lorenz.py",          "name": "lorenz"                                      },
         { "path": "examples/plotting/file/color_scatter.py",   "name": "color_scatter"                               },
         { "path": "examples/plotting/file/color_sliders.py",   "name": "color_sliders"                               },
-        { "path": "examples/models/iris_splom.py",             "name": "iris_splom",             "symbol": "grid"    },
-        { "path": "examples/models/anscombe.py",               "name": "anscombe",               "symbol": "grid"    },
```
bryevdv: These are the two examples I had to remove because they do explicit `Document` management. It's possibble `Document` could be patched to a function that returns `curdoc()` tho

sphinx/source/docs/user_guide/examples/layout_vertical.py
```diff
@@ -1,3 +1,10 @@
+"""
+This is an example doc string
+
+* Foo
+* Bar
+
+"""
```
bryevdv: Here's an example of a docstring that can be added to the top of an example. The RST content will *not* currently be renderered where the `bokeh-embed` happens, but *will* be rendered in the dedicated page that gets generated for this example. (Tho maybe that should be configurable in the directive). But this makes it super easy, e.g. to add indexing directives, so that all the examples can be indexed better.

bryevdv:
Also FYI with this PR and the file ordering, a clean docs build takes under two minutes.

bryevdv:
OK, since this is probably the only time in the next few years I'll be able to spend on this stuff, I am going to roll a bunch of `sphinxext improvements into one PR. But I will make a new PR with better commits broken out.

bryevdv:
superceded by #5407

PR #5407, opened 2016-10-25
bokeh.sphinxext improvements

### LINKED ISSUE #3020: Stricter docs build testing
Currently `make html` output is a mess, which hides new issues in the docs. Additionally, we either have to pay more attention to this kind of errors or make ci tests fail if `make html` failed.

### LINKED ISSUE #4774: Docs: Google Maps Down
The Gmaps example under http://bokeh.pydata.org/en/latest/docs/user_guide/geo.html is down due to Google's API update. A warning or key should be added unless it is felt the gmaps example should be removed.

screen shot 2016-07-08 at 3 54 34 pm

### LINKED ISSUE #5411: Document x_axis_location parameter to figure
It is not in reference guide: http://bokeh.pydata.org/en/latest/docs/reference/plotting.html

It also does not appear as a list of options when you mis-guess an option (presumably because it's not an actual property).

bryevdv:
[WIP: Needs tests]

This PR supercedes #5399

- [x] issues: closes #5411 closes #3020 closes #4774
- [ ] tests added / passed

This is a more comprehensive overhaul of `bokeh.sphinxext` machinery. In general, old and unused code (such as non-HTML builder visitors) has been deleted, and things have been streamlined and made to re-use other existing code instead of duplicating fucntionality. Some highlights:

* `bokeh-plot` re-uses server code runner, and adds a custom parser for `.py` files. This also causes a dedicated page for the example to be generated. Any docstring in the example will be rendered as ReST at the top of this page (but *not* at the directive embd location) making indexing and additional information simple to add. 

screen shot 2016-10-25 at 2 15 36 pm

* `bokeh-enum` to collapse huge enums:

    screen shot 2016-10-25 at 2 04 49 pm

* `bokeh-prop` code simplified and shortened, type information and links improved:

    screen shot 2016-10-25 at 2 05 57 pm

* adds `Option` class and `bokeh-options` to validate and document sets of configuration options:

    screen shot 2016-10-26 at 9 35 12 pm


bokeh/core/properties.py
```diff
@@ -1428,14 +1456,31 @@ def validate(self, value):
...
+
+    """
     def __init__(self):
         super(Auto, self).__init__("auto")
 
     def __str__(self):
         return self.__class__.__name__
 
+    def _sphinx_type(self):
+        return _PROP_LINK % self.__class__.__name__
```
mattpap: @bryevdv, I see this duplicated a lot. Can't this be moved to appropriate super-class and overriden only when necessary?

bryevdv: The problem is that it ends up being needed at both the base class level, and at some leave levels, but *not* at intermediate levels. As a concrete example: an actual `Either` property should print out what it is an either/or of. But subclasses of `Either` generally should not. So `Either` overrides to do something more specific, and then further subclasses ovrride back to basic. 

The only solution I can imagine is to make the "basic" version on the base class look for a differently named "hook" like `_sphinx_type_override` and use that when it is present. Happy to do that if you think it's better but it seems like a tossup to me.

bryevdv:
@bokeh/dev a version of the docs built with these changes has been deployed here:

http://bokeh.pydata.org/en/test/

bokeh/sphinxext/bokeh_prop.py
```diff
@@ -43,24 +47,28 @@ class Bar(Model):
 
 PROP_NAMES = [
     name for name, cls in properties.__dict__.items()
-    if isinstance(cls, type) and issubclass(cls, properties.Property)
+    if isinstance(cls, type) and (issubclass(cls, properties.Property) or issubclass(cls, properties.PropertyFactory))
```
bryevdv: This surprised me a bit @mattpap if you can confirm this is the best way.

bokeh/sphinxext/bokeh_plot.py
```diff
@@ -172,7 +175,10 @@ def run(self):
 
     def _get_source(self):
         if self.arguments:
-            source = open(self.arguments[0], "r").read()
+            path = self.arguments[0]
+            if path.startswith("docs"):
+                path = join("source", path)
+            source = open(path, "r").read()
```
bryevdv: This is just some junk to make things work with the path renames below. I wanted to do all renames in one commit by themselves to make reviewing the later commits easier.

bokeh/core/enums.py
```diff
@@ -93,10 +93,10 @@ def enumeration(*values, **kwargs):
 Direction = enumeration("clock", "anticlock")
 
 #: Specify units for mapping values
-Units = enumeration("screen", "data")
+Units = enumeration("screen", "data")  # TODO (bev): can this be deleted?
```
mattpap: @bryevdv, `git grep` shows no usage, besides `SpatialUnits = Units` and being imported in `test_glyphs.py`.

bryevdv: I'll remove it and we'll see how it goes

bokeh/sphinxext/bokeh_prop.py
mattpap: @bryevdv, `Property` and `PropertyFactory` are unrelated, and `PropertyFactory` is used for e.g. `Include`, so at first I would say it makes sense. But, I'm not sure if at this point `Include` shouldn't have been already replaced with included properties. If so, then checking for `PropertyFactory` shouldn't be necessary.

bokeh/core/properties.py
mattpap: @bryevdv, both approaches don't seem OOP compatible, though the current approach is still better than the override thing. I may look into this as time permits.

bryevdv: Well, this is really a delegation problem. So perhaps the solution is to make a little object or function and configure the base with that *value*, and then override the attribute that holds the delegate value as-needed.

bryevdv:
With the examples re-order the CI build went from ~20 minutes to 7 minutes.

bryevdv:
@mattpap @bokeh/dev note with the latest commits any sphinx warnings now result in a build error.

bryevdv:
with strict warnings, problems in prop docstrings, etc bail immediately, so I removed the `if False` debug block, it did not seem to work anymore anyway. I did have to make URL failures in `bokeh_github` roles just `INFO` because they regularly (expectedly) fail on Travis due to non-tag version.

bryevdv:
@birdsarah @mattpap @bokeh/dev OK, also added support for a `GOOGLE_API_KEY` env var. If this is not set, the docs build will fail. I added a secure env var to the Travis build, and the value is also available in the password wallet if you want to set it locally. Alternatively, set `BOKEH_DOCS_MISSING_API_KEY_OK=yes` to allow builds to proceed with broken GMaps.

bryevdv:
OK, I might add some tests if possible but apart from that, this is ready for review @bokeh/dev @mattpap 

The last commit simplifies and improves `bokeh-gallery`. It re-uses the `bokeh-plot` parser, and only re-runs things when absolutely necessary. So subsequent builds are quite faster. 

A few things that can be done later:

* I wanted to add a `bokehjs-plot` directive (and I started another branch for it) but there seems to be some problems with the BokehJS API at the moment @mattpap 

* `bokeh-gallery` is alot closer to being able to create multiple galleries, but it will still need a few tweaks to be made generic

bryevdv:
Well, OK there's one last py27 issue to fix up...

bryevdv:
good to go now

bryevdv:
@bokeh/dev plan to merge this this afternoon CST then cut a dev build, unless someone asked for more time.

bryevdv:
OK, no comments, merging now and kicking off a dev build

birdsarah:
thanks for your hard work bryan

PR #5409, opened 2016-10-26
Fix FuncTickFormatter.from_py_func missing return value and update examples

### LINKED ISSUE #5404: FuncTickFormatter.from_py_func() example ValueError
I am trying the FuncTickFormatter.from_py_func() example and I get the error
`ValueError: Function 'func' may only contain keyword arguments.`

A MWE that produces this error is:
```
from bokeh.models import FuncTickFormatter
import math

def ticker(tick):
    return math.floor(tick) + " + " + (tick % 1).toFixed(2)

p = figure(plot_width=400, plot_height=300)
         
p.yaxis.formatter = FuncTickFormatter.from_py_func(ticker)

p.circle([2, 4, 6], [1, 3, 5])

show(p)
```

I am running Bokeh version 0.12.3 and have flexx installed.

canavandl:
Fix FuncTickFormatter.from_py_func missing return value and update examples

- [x] issues: fixes #5404 
- [x] tests added / passed


bokeh/models/tests/test_formatters.py
```diff
@@ -48,33 +50,33 @@ def has_non_Model_keyword_argument(x=10):
...
-};
-return square(tick);
-"""
+    assert formatter.code == dedent("""\
+                                    "use strict";
+                                    var square;
+                                    square = function (x) {
+                                        return x * x;
+                                    };
+                                    return square(tick);
```
mattpap: Why so much indentation? This doesn't help readability at all.

canavandl: fixed

canavandl:
working formatter in docs:

screen shot 2016-10-26 at 12 54 51 pm

bryevdv:
LGTM

bryevdv:
@canavandl Actually, is it documented somewhere that a python function doesn't take any arguments? If not we should highlight that `tick` is just available (compare it to a module scope var perhaps)

canavandl:
clarified docstrings/user guide

bryevdv:
Perfect, will merge when green

Issue #4680, opened 2016-06-26
DataTable header height not large enough to fit header text - in Jupyter Notebook


sanga:
Code to repro (note: running in jupyter notebook):

```
df = pd.DataFrame({'a': range(5), 'b': range(5)})
columns = [TableColumn(field=k, title=k) for k in df.columns]
source = ColumnDataSource(df)
data_table = DataTable(source=source,columns=columns, fit_columns=True, row_headers=False, height=400)

show(widgetbox(data_table));
```

<img width="487" alt="screen shot 2016-06-26 at 13 21 25" src="https://cloud.githubusercontent.com/assets/2088/16361683/d484dfee-3ba0-11e6-8cdc-3d5b7368a492.png">

```
Python version      :  3.5.1 (default, Dec 26 2015, 18:11:22)
IPython version     :  4.2.0
Bokeh version       :  0.12.0rc4
BokehJS static path :  /usr/local/lib/python3.5/site-packages/bokeh/server/static
```

Observed on most recent Firefox and Chrome.

sanga:
slick-grid appears to be dead btw. Perhaps updating to an actively maintained fork might fix the problem?

i.e.

https://github.com/6pac/SlickGrid

sanga:
Note this is actually the case in your docs too

http://bokeh.pydata.org/en/latest/docs/user_guide/interaction.html#data-table

sanga:
Interestingly this works correctly in the "export_csv" server app - both when I run it locally and for example here:

https://demo.bokehplots.com/apps/export_csv

sophiajiyoung:
I get the same issue! (Previous version works fine.) Is there a workaround for this?

muyiwaolu:
I'm having the same issue. The columns aren't aligning and the font is too big for the headers.

bryevdv:
ping @birdsarah can you take a quick look when you have a chance? Maybe this is a quick fix

birdsarah:
will do

compeanansi:
I'm still having this issue

birdsarah:
Having looked quickly, this does not appear to be a quick fix.

sanga:
How about upgrading the underlying js lib btw? Would you like me to create a separate issue for that? 

(as a reminder/snipped from earlier in this thread):

"slick-grid appears to be dead btw. Perhaps updating to an actively maintained fork might fix the problem?

i.e.

https://github.com/6pac/SlickGrid"

[...]

PR #5410, opened 2016-10-26
Set box-sizing explicitly to make bokehjs work under bootstrap

### LINKED ISSUE #5398: DataTable css conflict with bootstrap css
thanks for bokeh!
Working with the `DataTable` I found an layout issue when combined with bootstrap css. Please look at this kind of minimal working example:
https://gist.github.com/schlichtanders/897b33daa846d4a08ba6e0f2ee5b2d75
In the python file the html one is constructed using jinja2 and bokeh `components`.
The important flag is `USE_BOOTSTRAP`.

If `USE_BOOTSTRAP=True` then you can see that the header of the table is broken (too small height and bad separator placement).
However if `USE_BOOTSTRAP=False` everything works as expected.
Hence some css classes seem to interact unexpectedly.

I hope the example files can help to track down this issue

Python version      :  2.7.6 (default, Jun 22 2015, 17:58:13) 
IPython version     :  4.0.1
Bokeh version       :  0.12.3

mattpap:
fixes #5398, #4680


mattpap:
Apparently we already have this kind of code in bokehjs in https://github.com/bokeh/bokeh/blob/master/bokehjs/src/less/widgets/data_table.less#L25.

mattpap:
This doesn't fix anything. The problem is with how bootstrap's view is created.

PR #5422, opened 2016-10-31
Minor layout cleanup

mattpap:
Improves readability and removes bulky code.

### DIFF
--- a/bokehjs/src/coffee/core/layout/solver.coffee
+++ b/bokehjs/src/coffee/core/layout/solver.coffee
@@ -3,17 +3,10 @@ import {Variable, Expression, Constraint, Operator, Strength, Solver as Constrai
 import {Events} from "../events"
 
 _constrainer = (op) ->
-  () =>
-    expr = Object.create(Expression.prototype)
-    Expression.apply(expr, arguments)
-    return new Constraint(expr, op)
+  (args) -> new Constraint(new Expression(args...), op)
 
 _weak_constrainer = (op) ->
-  () ->
-    args = [null]
-    for arg in arguments
-      args.push(arg)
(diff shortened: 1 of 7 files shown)


bokehjs/src/coffee/core/layout/solver.coffee
```diff
@@ -3,17 +3,10 @@ import {Variable, Expression, Constraint, Operator, Strength, Solver as Constrai
...
-    return new Constraint(expr, op)
+  (args) -> new Constraint(new Expression(args...), op)
 
 _weak_constrainer = (op) ->
-  () ->
-    args = [null]
-    for arg in arguments
-      args.push(arg)
-    new Constraint( new (Function.prototype.bind.apply(Expression, args)), op, Strength.weak )
+  (args) -> new Constraint(new Expression(args...), op, Strength.weak)
```
mattpap: @birdsarah, I presumptuously replaced this, but maybe there was a reason I overlook to write code like this?

bryevdv: I actually got this code from Chris Colbert IIRC, I thought there was some subtlety this was contending with but I can't recall any details. Maybe I was just unaware of `...` at the time, though.

birdsarah: If it works. I have no objection. I do recall some weirdness that required the bind.apply but I don't have anything specific, as @bryevdv said, I took the lead from him and @sccolbert.

birdsarah: Ultimately this just makes convenience functions around solver so it should be pretty easy to know if this didn't work.

mattpap:
This doesn't work. I will revisit this at some point.

PR #5425, opened 2016-10-31
Configure code runner so that full class paths are preserved

### LINKED ISSUE #5315: extension implementation load path problems in apps

mattpap:
fixes #5315


bryevdv:
@mattpap if I try to update the docs widget extension example, I see:
```
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/cmdline.py", line 244, in main
    app.build(opts.force_all, filenames)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/application.py", line 297, in build
    self.builder.build_update()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/builders/__init__.py", line 251, in build_update
    'out of date' % len(to_build))
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/builders/__init__.py", line 265, in build
    self.doctreedir, self.app))
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/environment.py", line 569, in update
    self._read_serial(docnames, app)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/environment.py", line 589, in _read_serial
    self.read_doc(docname, app)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/environment.py", line 742, in read_doc
    pub.publish()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.4.8-py3.4.egg/sphinx/io.py", line 49, in read
    self.parse()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/Users/bryan/work/bokeh/bokeh/sphinxext/bokeh_plot.py", line 162, in parse
    (script, js, js_path, source) = _process_script(source, filename, env.bokeh_plot_auxdir, js_name)
  File "/Users/bryan/work/bokeh/bokeh/sphinxext/bokeh_plot.py", line 125, in _process_script
    js, script = autoload_static(d.roots[0], resources, script_path)
  File "/Users/bryan/work/bokeh/bokeh/embed.py", line 347, in autoload_static
    js_raw = resources.js_raw + [script],
  File "/Users/bryan/work/bokeh/bokeh/resources.py", line 388, in js_raw
    custom_models = gen_custom_models_static()
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 269, in gen_custom_models_static
    impl = model.implementation
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 232, in implementation
    impl = FromFile(join(self.path, impl))
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 178, in __init__
    with io.open(path, encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/bryan/work/bokeh/sphinx/extensions_ion_range_slider.coffee'

Exception occurred:
  File "/Users/bryan/work/bokeh/bokeh/util/compiler.py", line 178, in __init__
    with io.open(path, encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/bryan/work/bokeh/sphinx/extensions_ion_range_slider.coffee'
The full traceback has been saved in /var/folders/t1/48dhz2zd3jgdr0d0zss8kq4h0000gq/T/sphinx-err-ci88vnjb.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 1
```
There were some file renames:
```
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	renamed:    source/docs/user_guide/examples/extensions_example_widget._py -> source/docs/user_guide/examples/extensions_example_widget.py
	deleted:    source/docs/user_guide/examples/extensions_example_widget_plot._py
```
And here's the small diff:
```
diff --git a/sphinx/source/docs/user_guide/examples/extensions_example_widget.py b/sphinx/source/docs/user_guide/examples/extensions_example_widget.py
index 6e8f6bb..8571983 100644
--- a/sphinx/source/docs/user_guide/examples/extensions_example_widget.py
+++ b/sphinx/source/docs/user_guide/examples/extensions_example_widget.py
@@ -16,7 +16,7 @@ class IonRangeSlider(InputWidget):
     # of the custom extension model or a string name of a JavaScript (or
     # CoffeeScript) file with the implementation.

-    __implementation__ = open('extensions_ion_range_slider.coffee').read()
+    __implementation__ = 'extensions_ion_range_slider.coffee'
     __javascript__ = ["https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js",
                       "https://cdnjs.cloudflare.com/ajax/libs/ion-rangeslider/2.1.4/js/ion.rangeSlider.js"]
     __css__ = ["https://cdnjs.cloudflare.com/ajax/libs/normalize/4.2.0/normalize.css",
diff --git a/sphinx/source/docs/user_guide/extensions.rst b/sphinx/source/docs/user_guide/extensions.rst
index 9a13708..6ce15c4 100644
--- a/sphinx/source/docs/user_guide/extensions.rst
+++ b/sphinx/source/docs/user_guide/extensions.rst
@@ -270,7 +270,7 @@ and improvements to this section for future users.
 :ref:`userguide_extensions_examples_latex`
     Include a third-party JavaScript library in order to render LaTex.

- :ref:`userguide_extensions_examples_widget`
+:ref:`userguide_extensions_examples_widget`
     Include a third-party JavaScript library in an extension widget.

 .. _CoffeeScript: http://coffeescript.org
diff --git a/sphinx/source/docs/user_guide/extensions_gallery/widget.rst b/sphinx/source/docs/user_guide/extensions_gallery/widget.rst
index eed726c..a92fbe0 100644
--- a/sphinx/source/docs/user_guide/extensions_gallery/widget.rst
+++ b/sphinx/source/docs/user_guide/extensions_gallery/widget.rst
@@ -8,12 +8,12 @@ This example shows how to add a double-ended slider widget to the plot.
 The single normal Bokeh slider controls the power of the line.
 The double ended sliders control the x range for the line.

-.. .. bokeh-plot:: docs/user_guide/examples/extensions_example_widget_plot.py
-..    :source-position: none
+.. bokeh-plot:: docs/user_guide/examples/extensions_examle_widget.py
+    :source-position: none

 Python script:

-.. literalinclude:: ../examples/extensions_example_widget._py
+.. literalinclude:: ../examples/extensions_example_widget.py
    :language: python

 Coffeescript for ion range slider:
```

bryevdv:
damnit, the close and

mattpap:
@bryevdv, I don't follow. Is this behavior present in `master`? Otherwise I don't see connection with this PR.

bryevdv:
@mattpap no this is on top of your PR. I had disabled an example in the docs, pending this PR to fix it, but it seems there is still an issue

bryevdv:
Can confirm the code runner is getting initialized with the full abs path to the python file, e.g.
```
/Users/bryan/work/bokeh/sphinx/source/docs/user_guide/examples/extensions_example_widget.py
```
But it's looking at the toplevel sphinx dir (i.e. where sphinx is run) to get the base path (and of course failing):
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/bryan/work/bokeh/sphinx/extensions_ion_range_slider.coffee'
```

bryevdv:
Presumably it's running through the `getcwd` path here:
```

    @property
    def path(self):
        if self.file is None:
            return os.getcwd()
        else:
            return dirname(self.file)
```

mattpap:
@bryevdv, can you push your changes into this PR, so I have something to work with? For now I only concerned with apps, but it would be fix both at the same time.

bryevdv:
this is probably why:
```
   @property
    def file(self):
        try:
            file = inspect.getfile(self.cls)
        except TypeError:
            return None
        else:
            return abspath(file)
```
guessing `inspect.getfile` is failing on the programmatically created module

I'll push my changes

mattpap:
@bryevdv, what's wrong is explained in https://github.com/bokeh/bokeh/issues/5315#issuecomment-255844570. Just the fix has to be applied "recursively".

bryevdv:
Build docs with e.g
```
 GOOGLE_API_KEY="foo" BOKEH_DOCS_CDN=local make SPHINXOPTS=-v  clean html serve
```

mattpap:
> `GOOGLE_API_KEY="foo" BOKEH_DOCS_CDN=local make SPHINXOPTS=-v  clean html serve`

That's a lot. I presume `BOKEH_DOCS_MISSING_API_KEY_OK=yes` is also good.

mattpap:
> `RuntimeError: Invalid DISPLAY variable`

So much for building over SSH.

bryevdv:
Yes, `BOKEH_DOCS_MISSING_API_KEY_OK=yes` is also fine the others say to use local resources, and to turn up debugging

bryevdv:
> So much for building over SSH.


sorry, just delete the `serve` from the make invovation

mattpap:
> sorry, just delete the serve from the make invovation

I didn't use it. I just had `BLA make BLA clean html`.

bryevdv:
That's weird and unexpected, `clean` and `html` just do this:
```
clean:
	-rm -rf $(BUILDDIR)/*
	-rm -rf source/docs/gallery/*

html:
	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
	@ln -sf ../../../bokeh/server/static $(BUILDDIR)/html/static
	@echo
	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
```
so must be something to do with `sphinx-build` or one of the extensions

mattpap:
> That's weird and unexpected (...)

@bryevdv, see my edit above. It's matplotlib. The build starts failing on `examples/concepts_other_interfaces.py`.

mattpap:
>  We have a workaround for this in travis ci's scripts.

@bryevdv, here: https://github.com/bokeh/bokeh/blob/master/scripts/travis_install#L64.

mattpap:
Looks like `clean` in `BLA make BLA clean html` is actually required. Otherwise I get `KeyError: 'Annual'`.

bryevdv:
Ah right I forgot to fix that. Happens in some situations not others.

mattpap:
```
Warning, treated as error:
WARNING: intersphinx inventory u'http://docs.scipy.org/doc/numpy/objects.inv'
not fetchable due to <class 'urllib2.HTTPError'>: HTTP Error 404: Not Found
```

bryevdv:
Saw that once today had not seen it previously

mattpap:
`scipy.org` is down, so I can't build our documentation. Well, I can always disable intersphinx.

bryevdv:
Guessing scipy.org is having issues

bryevdv:
You asked for a strict build :)

mattpap:
> You asked for a strict build :)

Not complaining about this. What I'm complaining about this non-robust inter-connectivity mixed up with too much self confidence, that makes very unrelated events and changes rise to major issues. sphinx could just cache those files globally and just symlink them again after a clean.

mattpap:
Another thing I have to complain about is ipdb (or perhaps IPython in general) where those little things are getting broken. E.g. I'm unable to stop debugging neither with C-C nor C-V.

bryevdv:
> Not complaining about this. What I'm complaining about this non-robust inter-connectivity mixed up with too much self confidence, that makes very unrelated events and changes rise to major issues. sphinx could just cache those files globally and just symlink them again after a clean.

Sure, agreed. If we wanted to we could probably write an extension that does this somehow.

mattpap:
> If we wanted to we could probably write an extension that does this somehow.

Yeah, but that's not the point. This is so obvious workflow that I shouldn't be afraid running `clean` command e.g. on a plane (assuming no WIFI), because I may not be able to continue my work.

mattpap:
So, the problem is that `cls.__module__` is a string, so to get `cls`' path, you have to go through `sys.modules`. This isn't an issue with usual modules, but with auto-generated, you have to manually add them there. This is do able, but if we want to clean up after this, then we won't be able to take advantage of this at the time that custom models are resolved.

bryevdv:
I don't follow the last sentence. IIUC you mean the example handler can set `cls.__module__` to the synthetic module it creates? But that there is some drawback?

mattpap:
@bryevdv, can you tell me what kind of magic is this:
```
        # This is to prevent the module from being gc'd before the
        # document is.  A symptom of a gc'd module is that its
        # globals become None.
        if not hasattr(doc, '_CodeHandler__modules'):
            setattr(doc, '_CodeHandler__modules', [])
        doc.__modules.append(module)
```
? When I step debug this and issue `p doc.__modules` I get an exception (as I expect it), but this code actually works. WTF? Just to clarify, I understand why the code is there, just not exactly why it works.

mattpap:
Maybe as temporary (?) measure we could actually modify `sys.modules`, but maintain also a list of per-Document modules and then remove them from `sys.modules` in `doc.__del__`?

bryevdv:
> but this code actually works. WTF? Just to clarify, I understand why the code is there, just not exactly why it works.

That's how double underscore attr names are name-mangled in Python? Or are you asking something else?

> Maybe as temporary (?) measure we could actually modify sys.modules, but maintain also a list of per-Document modules and then remove them from sys.modules in doc.__del__?

That seems worth pursuing

mattpap:
> That's how double underscore attr names are name-mangled in Python? Or are you asking something else?

I'm asking precisely about this. I wasn't aware of this sorcery.

mattpap:
> That seems worth pursuing

I'm implementing this.

mattpap:
@bryevdv, this approach works. I will have to still pursue getting resources bound to documents, but this isn't something that can be done in an hour or two.

bryevdv:
Agreed, there are several related pieces here but anything that can done incrementally, should be done incrementally

bokeh/document.py
```diff
@@ -298,6 +300,12 @@ def __init__(self, **kwargs):
         self._callbacks = {}
         self._session_callbacks = {}
         self._session_context = None
+        self._modules = []
+
+    def __del__(self):
+        for module in self._modules:
+            if module.__name__ in sys.modules:
+                del sys.modules[module.__name__]
```
bryevdv: `__del__` is extremely uncommon I think a note here about why this is where it is would help anyone else looking at this code

bryevdv: And I guess an explanation of purpose could help too. For instance, if there are multiple docs with the same module in `doc._modules`, shouldn't it stay in `sys.modules` until all the docs are gone? (I'm probably missing something here)

mattpap: As far my understanding goes, documents don't share modules, so it should be safe to remove them all.

bryevdv: Tornado is a single python process. Every session served by a given bokeh server has its  own document. If the app uses a custom module, wouldn't the custom models be shared amongst all of the documents for all of the sessions in the one `sys.modules` for the single Tornado process? Maybe I am thinking of something else or missing something obvious.

mattpap: Given this code:
```
 def modify_document(self, doc):
     if self.failed:
         return
     module = self._runner.new_module()
```
 in `bokeh/application/handlers/code.py` I would say it's safe to assume that a module won't be shared across documents. Though, we may have multiple modules referring to one piece of code per document. I'm not sure when or how often `modify_document()` is called, so I'm not sure if it's a problem or not.

bryevdv: `modify_document` is called for every new session, its what generates the document for the session.

bryevdv:
@mattpap need a little context around the removed tests. It looks like we are still inserting a path  (now an explicit one) into the front of `sys.path` if so, this is very important to maintain the specific details of this under test. 

Are we no longer going to `chdir` into app directories? If not I think that will break some exiting code that relied on it? Or is there something I am missing that would make `open("relpath.foo")` just work?

bryevdv:
I fixed the merge caused by the docs example file changes in the other PR, and also verified that the multi-part widgets example works as expected now 👍 

I think with the `sys.path` test restored this will be ready to go.

bokeh/application/handlers/tests/test_code.py
```diff
@@ -58,22 +57,6 @@ def test_script_runtime_error(self):
...
-        assert "path: ''" in handler.error
-
-    def test_script_cwd(self):
-        doc = Document()
-        handler = CodeHandler(source="""import os; raise RuntimeError("cwd: '%s'" % os.getcwd())""", filename="/test_filename")
-        handler.modify_document(doc)
-
-        assert handler.error is not None
-        assert "cwd: '%s'" % abspath(sep) in handler.error
-
```
bryevdv: Can you elaborate? These (or something like them) seem very important to maintain.

bokeh/document.py
bryevdv: I guess the question that would allay my concern here: what is `module.__name__` here? Is it some different random/unique string for every module returned by `self._runner.new_module()` ?

bryevdv: Yes, it is unique
```
module_name = 'bk_script_' + make_id().replace('-', '')
```
OK, this seems fine to me now, sorry for the noise

bokeh/application/handlers/code.py
```diff
@@ -54,12 +55,8 @@ def modify_document(self, doc):
 
         module = self._runner.new_module()
 
-        # This is to prevent the module from being gc'd before the
-        # document is.  A symptom of a gc'd module is that its
-        # globals become None.
```
havocp: without a comment will anyone understand why we store the module anywhere?

bryevdv: agreed I think this was a good comment

mattpap: Originally this was a very important comment, but now we just store modules for further use, so this comment is irrelevant. Perhaps a different one should be used instead.

bryevdv: Sure OK that's true, but it's a really important sub-reason :) I think it's worth noting, maybe:
```
# One reason modules are stored is to prevent the module 
# from being gc'd before the document is.  (A symptom of a 
# gc'd module is that its globals become None.) Additionally
# stored modules are used to provide correct paths. 

```

bryevdv:
@mattpap can  you restore that one test, and add the comment above today? I'd like to get this merged and cut a dev build to do preliminary testing with.

bryevdv:
Awesome will merge when green

PR #5429, opened 2016-11-01
experimental junk

bryevdv:
Experimental PR for discussion: DO NO MERGE

### DIFF
--- a/bokeh/util/serialization.py
+++ b/bokeh/util/serialization.py
@@ -4,6 +4,8 @@ Bokeh objects.
 """
 from __future__ import absolute_import
 
+import base64
+
 from six import iterkeys
 
 from .dependencies import import_optional
@@ -52,8 +54,7 @@ def transform_series(obj):
     return transform_array(vals)
 
 def flattened_and_shape(array):
-    return list(array.flat), array.shape
-
+    return  base64.b64encode(array).decode('utf-8'), array.shape
 
 def transform_array(obj):
     """Transform arrays into lists of json safe types
--- a/bokeh/util/serialization.py
+++ b/bokeh/util/serialization.py
@@ -51,6 +51,10 @@ def transform_series(obj):
     vals = obj.values
     return transform_array(vals)
 
+def flattened_and_shape(array):
+    return list(array.flat), array.shape
+
+
 def transform_array(obj):
     """Transform arrays into lists of json safe types
     also handles pandas series, and replacing
@@ -77,14 +81,14 @@ def transform_array(obj):
     if obj.dtype.kind == 'M':
         if legacy_datetime64:
             if obj.dtype == np.dtype('datetime64[ns]'):
-                return (obj.astype('int64') / 10**6.0).tolist()
+                return flattened_and_shape(obj.astype('int64') / 10**6.0)
         else:
-            return (obj.astype('datetime64[us]').astype('int64') / 1000.).tolist()
+            return flattened_and_shape(obj.astype('datetime64[us]').astype('int64') / 1000.)
     elif obj.dtype.kind == 'm':
-        return (obj.astype('timedelta64[us]').astype('int64') / 1000).tolist()
+        return flattened_and_shape(obj.astype('timedelta64[us]').astype('int64') / 1000)
     elif obj.dtype.kind in ('u', 'i', 'f'):
         return transform_numerical_array(obj)
-    return obj.tolist()
+    return flattened_and_shape(obj)
 
 def transform_numerical_array(obj):
     """handles nans/inf conversion
@@ -92,13 +96,13 @@ def transform_numerical_array(obj):
     if isinstance(obj, np.ma.MaskedArray):
         obj = obj.filled(np.nan)  # Set masked values to nan
     if not np.isnan(obj).any() and not np.isinf(obj).any():
-        return obj.tolist()
+        return flattened_and_shape(obj)
     else:
         transformed = obj.astype('object')
         transformed[np.isnan(obj)] = 'NaN'
         transformed[np.isposinf(obj)] = 'Infinity'
         transformed[np.isneginf(obj)] = '-Infinity'
-        return transformed.tolist()
+        return flattened_and_shape(transformed)
 
 def traverse_data(datum, is_numpy=is_numpy, use_numpy=True):
     """recursively dig until a flat list is found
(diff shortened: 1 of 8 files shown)


bokeh/util/serialization.py
```diff
@@ -112,8 +116,14 @@ def traverse_data(datum, is_numpy=is_numpy, use_numpy=True):
         use_numpy: toggle numpy as a dependency for testing purposes
     """
     is_numpy = is_numpy and use_numpy
-    if is_numpy and not any(isinstance(el, (list, tuple)) for el in datum):
-        return transform_array(np.asarray(datum))
```
bryevdv: I also think this was a bug that just happened to "work" because we usually have a list with just one array, or all the same size. Different sized array, (i.e. different sized images) would behave badly:
```
In [7]: np.asarray(c)
Out[7]:
array([array([1, 2, 3]), array([[1, 2],
       [3, 4]])], dtype=object)
```

bryevdv:
ping @jbednar @philippjfr it would be good to know if these sorts of changes offer any improvement for image performance, especially in the notebook. If so we can figure out how to clean up it to be suitable for inclusion.

philippjfr:
@bryevdv, I'm on it, will look into it later today.

philippjfr:
Sorry @bryevdv, busy week last week. Trying this now.

philippjfr:
@bryevdv As far as I can tell this PR actually makes image updates slightly slower. Using a 100x100 image, the JSON patch takes about 67 ms to generate in python without the PR and 75 ms with the PR. Including the processing on the JS end the each update takes 100 ms without the PR and 120 ms with the PR. A 20% penalty to flatten the array seems like a lot, but it definitely doesn't seem to be helping.

philippjfr:
Edit: Worth noting I compared 0.12.3 to this PR, I'll try current master now.

bryevdv:
@philippjfr well, this is supposed to help with very much larger cases, can you try 3000x3000 ?

jbednar:
We've definitely avoided even trying 3000x3000 in bokeh due to it already
being too slow for 1500x1500, but it's great that it might be worth trying
now!

Jim

philippjfr:
It still seems to be slightly slower at 2000x2000 unfortunately, again mostly on the Python end:

```
       tottime  percall  cumtime  percall      function
w/  PR  0.532    0.532   19.725   19.725 encoder.py:186(encode)
w/o PR   0.517    0.517   18.041   18.041 encoder.py:186(encode)
```

With or without the PR this frequently crashes my browser tab so I'm not even going to try 3000x3000.

bryevdv:
That's incredibly bizarre. It should be doing much less work. I'll keep poking at it.

philippjfr:
I just tried it again something was off about my setup before, using just ``push_notebook`` I did see a small but not significant from 22 seconds to about 21 seconds on a 2000x2000 image.

bryevdv:
@philippjfr OK, this is even more junk, `push_notebook` won't work. But something like this:
```
N = 2000
x = np.linspace(0, 10, N, dtype=np.float64)
y = np.linspace(0, 10, N, dtype=np.float64)
xx, yy = np.meshgrid(x, y)
d = np.sin(xx)*np.cos(yy)

p.image(image=[d], x=0, y=0, dh=10, dw=10, palette="Spectral11")
```
seems to render in about 5-6 seconds in the notebook for me. Can you try to compare again? 

This is a  tentative attempt at sending arrays as base64 encoded, if this junk works, we can start to figure out how to really do it in general, tho that will probably be hard.

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -6,20 +6,113 @@ import {SelectionManager} from "../../core/selection_manager"
...
 # Datasource where the data is defined column-wise, i.e. each key in the
 # the data attribute is a column name, and its value is an array of scalars.
 # Each column should be the same length.
 export class ColumnDataSource extends DataSource
   type: 'ColumnDataSource'
 
+  initialize: (options) ->
+    super(options)
+    for k, v of @data
+      bytes = decode(v[0])
```
mattpap: You may want to try `atob()` built-in instead.

bryevdv: I had read there was some issue with it but I will try now.

bryevdv: simple replacement does not work, for context: http://blog.danguer.com/2011/10/24/base64-binary-decoding-in-javascript/

which claims "Currently all the base64 decoders in javascript are for strings, not suitable for binary data, the common example is if you uses a canvas element and get their base64 representation ( canvas.toDataURL() ) usually you will upload and in the server do the base64 decode, if you want to process the data in javascript you will find the data will got corrupted since it’s processed as string."

philippjfr:
>@philippjfr OK, this is even more junk, push_notebook won't work. But something like this:

@bryevdv I just hacked it by inserting this in ``ImageView._set_data``:

```python
      if @_image? and typeof @_image[0] is 'string'
        bytes = decode_arr(@_image[0])
        @_image = [new Float64Array(bytes)]
```

And was able to update the images. The results are amazing, a 2000x2000 image update now runs in 250ms (vs. the original 21 seconds).

bryevdv:
OK well, I guess we need to figure out a plan to do this for real across all the different use-cases

philippjfr:
Okay, it seems I need to stop doing this. I was testing our custom updating code in HoloViews, which indeed now completes in 250 ms, using ``push_notebook`` it takes 1.6 seconds, which is still a 14x improvement.

bryevdv:
1.6s for 2000x2000 is definitely fine, since that's probably more pixels than most will actually have on the screen in the notebook. If you you are especially ambitions, a plot of the time curve for various sizes from 10x10 up to 2000x2000 would be amazing to have (comparison not necessarily necessary just timings for this PR)

philippjfr:
Here's the graph if I restart the kernel between each update (``push_notebook`` seems to accumulate state and slow down significantly):

image

bokehjs/src/coffee/models/sources/column_data_source.coffee
mattpap: I see. Well, what did I expect from the standard library.

mattpap: Is there an npm package for this code? We shouldn't be maintaining this ourselves. There are multiple implementations available, e.g. `base64-js` which seems do work exactly as this code (uses typed arrays). Performance is perhaps the only thing to consider.

philippjfr:
Is there anything I could do to get this moving? I'm really excited to finally see these improvements and would be happy to help out a bit.

jbednar:
Yes -- for datashader, the speed of transferring images into Bokeh is very often the limiting step, so it would be fabulous to get some improvement there.

bryevdv:
superseded by #5504

PR #5432, opened 2016-11-01
need to ask mapper to return synthetic coords

### LINKED ISSUE #5413: Can't use categorical axis with figure using rects
I have the following code to draw a heatmap, one row for each parameter:

```
params = ['end_brick', 'comms']
date_range = [df['date'].min(), df['date'].max()]
source = ColumnDataSource(df)

p = figure(
    x_axis_type="datetime", x_range=date_range, y_range=params,
    tools='xpan, xwheel_zoom, reset, save, resize, hover', width=800, height=300,
)

for i, par in enumerate(params):
    p.quad(left='start', right='end', bottom=i, top=i + 1,
           color={'field': par, 'transform': cmaps[par]}, source=source)
```

The data looks like this:

```
        date  end_brick    comms  period               start                    end                          
0 2016-10-26       95.6  99.3333       1 2016-10-26 00:00:00  0 2016-10-26 08:00:00                            
1 2016-10-26       75.0  98.8889       3 2016-10-26 08:00:00  1 2016-10-26 16:00:00                            
2 2016-10-26       58.7  96.6667       5 2016-10-26 16:00:00  2 2016-10-27 00:00:00                            
3 2016-10-27       80.1  98.3333       1 2016-10-27 00:00:00  3 2016-10-27 08:00:00                                             
```

It works but as soon I go to zoom in or out, it disappears and the console error is:

```
Uncaught Error: Number property 'start' given invalid value: end_brick(…)
```

The graph works fine when I use y_range=(0, len(params)) but then I lose labels.

The same bug happens with other charts e.g. http://bokeh.pydata.org/en/latest/docs/gallery/boxplot_chart.html

It also happens with zoom_in and zoom_out tools.

bryevdv:
issues: fixes #5413 

The refactor for the zoom in/out tools neglected to maintain the pure synthetic coordinate mappings necessary for zooming to work for categorical axes. 

This does some light refactoring so that some tests can be added for the new functions. Testing `scale_range` itself is too difficult to set up because it accepts a `CartesianFrame` but the new tests should ensure the functionality on categorical mappings.


bryevdv:
ping @birdsarah your review/testing would be appreciated if you have the bandwidth.

bryevdv:
@birdsarah some other folks have had a chance to confirm the fix, and as this adds tests and cleanup I think it's safe to merge. Thanks for the quick look yesterday!

PR #5433, opened 2016-11-02
Extend TextInput with `placeholder` attribute

### LINKED ISSUE #5417: Extend TextInput with `placeholder` and `pattern`
`placeholder` and `pattern` are nice features in HTML to give the user info about the field and to validate the user-input.

Add `placeholder` and `pattern` to `TextInput` in python, extend `text_input.coffee` and the `text_input_template.eco`

busino:
issues: fixes #5417

### DIFF
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -43,6 +43,10 @@ class TextInput(InputWidget):
     A callback to run in the browser whenever the user unfocuses the TextInput
     widget by hitting Enter or clicking outside of the text box area.
     """)
+    
+    placeholder = String(default="", help="""
+    Placeholder for empty input field
+    """)
 
 
 class AutocompleteInput(TextInput):
--- a/bokehjs/src/coffee/models/widgets/text_input.coffee
+++ b/bokehjs/src/coffee/models/widgets/text_input.coffee
@@ -40,4 +40,5 @@ export class TextInput extends InputWidget
 
   @define {
       value: [ p.String, "" ]
+      placeholder: [ p.String, "" ]
     }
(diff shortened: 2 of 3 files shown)


bokeh/models/widgets/inputs.py
```diff
@@ -43,6 +43,10 @@ class TextInput(InputWidget):
     A callback to run in the browser whenever the user unfocuses the TextInput
     widget by hitting Enter or clicking outside of the text box area.
     """)
+    
+    placeholder = String(default="", help="""
+    Placeholder for empty input field
+    """)
```
bryevdv: @busino it occurs to me, does placeholder need to be a single character? If so we can add a new class Character(String) property type that validates len(self.value)==1 easily enough.

bryevdv: Or maybe am I misunderstanding placeholder? I thought was something like `"*"` that would obscure a password field when a user types, but maybe not?

busino: @bryevdv 
The placeholder is used for empty input fields.
See here: http://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_placeholder
I would like to give the user information about the format that the field should have. E.g. for datetime input the placeholder is `2016-12-12 12:00:03`. The user now has the information which format the input should have. An therefore the `pattern` would be a good improvement too.

bryevdv: Ah got it, it's good as is then!

bryevdv:
Oh interesting, the strict sphinx build is failing because secure travis env vars (in this case `GOOGLE_API_KEY` is not available except to folks in @bokeh/dev 

The test failures are spurious, so this can be merged whenever @busino would you be interested in updating one or tow of the existing examples to add a placeholder to them?

busino:
@bryevdv 
I checked the examples and the TextInput is only used in:
 * taylor_server.py (value given)
 * sliders.py (value given)
 * app/movies/main.py (can add Tanrantino and Hanks as `placeholder` hihihi)

bryevdv:
hah I see I thought there were more examples with empty inputs. I'll just go ahead and merge now, thanks  for the PR!

PR #5437, opened 2016-11-02
Don't use tail calls in getters/setters to avoid a bug in Safari

### LINKED ISSUE #5392: tools cause hard crash on Safari after import/export PR
@mattpap OSX / Safari is currently crash easily. AS an example:
- run `unemployment.py`
- pan around
- leave plot area
- pan more

That's pretty reliable way to crash and cause a page reload. Seems to be confined to Safari. A git bisection traced it to the import/export PR:

UPDATE: seems to be related to hover tool being on, see below

```
[bryan:~/work/bokeh]$ git bisect good                                                                                                          (1ef6d22)
d2ce1971e4b984c7d30728b2378f7c959a3efd06 is the first bad commit
commit d2ce1971e4b984c7d30728b2378f7c959a3efd06
Author: Mateusz Paprocki <mattpap@gmail.com>
Date:   Tue Oct 11 19:35:45 2016 +0200

    Use ES6 import/export instead of require() and module.exports (#5321)

    * tsjs

    * underscore

    * jquery

    * import 'name'

    * import * as Mod from "mod"

    * import {name_1, ..., name_n} from "mod"

    * underscore patching

    * mousewheel

    * Remove duplicated Title

    * export {...}

    * Fix imports/exports in models/tiles

    * Adjust imports/exports

    * Split models/index.coffee and use exports

    * import {Model}

    * Use exports in core modules

    * Use exports in models/widgets

    * Use exports in api/

    * More exports from main.coffee

    * Integrate scripts:tsjs in to scripts:compile

    * Use default exports for templates

    * Use exports in main.coffee

    * Use exports in api.coffee

    * Export require in prelude, not in main

    * Don't show tsjs errors unless --tsjs option is used

    * Use exports in core/util/proj4

    * Set up scripts' build dependencies properly

    * Use window.Date() to avoid clash with exported Date

    * Use compiled *.js files for testing

    * Use proper this in pop_pending() in client.coffee

    * Use HasProps instead of HasProps.constructor

    * Fix imports in tests

    * Don't use file extensions in imports in tests

    * Use build/js/tree in proxy require magic

    * Add a polyfill for String.prototype.startsWith()

    * Fix imports of Model in custom model examples

    * Fix broken import in models/canvas/cartesian_frame

    * Add typings for all external libraries

    * Allow to skip selected scripts:tsjs errors

    * Add typings for kiwi

    * Update gulp-typescript to version 3.x

    * Don't compile *.ts files, just copy under tree_ts/

    This allows to compile all *.ts files (handwritten and generated)
    in one compiler run, allowing for proper module resolution, etc.

    * Rewrite api/palettes in TypeScript and fix exports

    * Increase size limit for bokeh-api bundle

    * Switch to direct model/view exports

    * Remove debugging code from gulp/tasks/scripts.coffee

    * Fix broken imports in webgl/{base,line} and tools/toolbar_base

    * Move compile.js build code to compiler:build

    * No need anymore for run-time Eco template compilation

    * Add support for TypeScript to compile.js

    * Use import/export in examples and documentation

    * Allow to use file extensions in custom module paths

    Now you can issue e.g. `import "./custom.less"`.

    * Add support for TypeScript to custom module compiler

    * Fix tests after adding TypeScript to compilation pipeline

    * Access default member by using ['default'] syntax

:040000 040000 8541bccdb203e2062faba1401e5304f106a4e1b2 84a96944fe50f4d3633be352beeccfed4609f3c0 M  bokeh
:040000 040000 ef8f45e4c0988ea4b44a58a2aa39e9f62442cb82 defe848efd3bb6f7fad9677a1ed56ce89188da65 M  bokehjs
:040000 040000 36f13b0ac4bbe15496b3873ee6f02fbe52f6a64f cad0a8cdad9d62c50cc77caf6594747b030cab5c M  examples
:040000 040000 f2d4047399d5c29cd837b9fe56544087f7a4a06c b364c8da71fba56e565fc47bc44c20b19eb0021d M  sphinx
```

mattpap:
@bryevdv, if I understand https://bugs.webkit.org/show_bug.cgi?id=164306#c4, this should fix the problem. We use other getters (setters are uncommon), but they aren't used as intensively as those used for property access.

fixes #5392

### DIFF
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -43,7 +43,7 @@ class TextInput(InputWidget):
     A callback to run in the browser whenever the user unfocuses the TextInput
     widget by hitting Enter or clicking outside of the text box area.
     """)
-    
+
     placeholder = String(default="", help="""
     Placeholder for empty input field
     """)
--- a/bokehjs/src/coffee/core/has_props.coffee
+++ b/bokehjs/src/coffee/core/has_props.coffee
@@ -23,8 +23,9 @@ export class HasProps extends Backbone.Model
           throw new Error("attempted to redefine attribute '#{this.name}.#{name}'")
 
         Object.defineProperty(this.prototype, name, {
-          get: ()      -> this.getv(name)
-          set: (value) -> this.setv(name, value)
+          # XXX: don't use tail calls in getters/setters due to https://bugs.webkit.org/show_bug.cgi?id=164306
+          get: ()      -> value = this.getv(name); return value
+          set: (value) -> this.setv(name, value); return this
         }, {
           configurable: false
           enumerable: true


bryevdv:
@mattpap I added them in the deprecated ones too:
```
diff --git a/bokehjs/src/coffee/core/has_props.coffee b/bokehjs/src/coffee/core/has_props.coffee
index 618aa86..9a12c14 100644
--- a/bokehjs/src/coffee/core/has_props.coffee
+++ b/bokehjs/src/coffee/core/has_props.coffee
@@ -23,8 +23,15 @@ export class HasProps extends Backbone.Model
           throw new Error("attempted to redefine attribute '#{this.name}.#{name}'")
 
         Object.defineProperty(this.prototype, name, {
-          get: ()      -> this.getv(name)
-          set: (value) -> this.setv(name, value)
+          get: () ->
+            # avoid tail call in setter: https://bugs.webkit.org/show_bug.cgi?id=164306
+            value = this.getv(name)
+            return value
+
+          set: (value) ->
+            # avoid tail call in setter: https://bugs.webkit.org/show_bug.cgi?id=164306
+            value = this.setv(name, value)
+            return value
         }, {
           configurable: false
           enumerable: true
@@ -205,12 +212,16 @@ export class HasProps extends Backbone.Model
     return prop_spec
 
   set: (key, value, options) ->
+    value = @setv(key, value, options)
     logger.warn("HasProps.set('prop_name', value) is deprecated, use HasProps.prop_name = value instead")
-    return @setv(key, value, options)
+    # avoid tail call in setter: https://bugs.webkit.org/show_bug.cgi?id=164306
+    return value
 
   get: (prop_name) ->
+    value = @getv(prop_name)
     logger.warn("HasProps.get('prop_name') is deprecated, use HasProps.prop_name instead")
-    return @getv(prop_name)
+    # avoid tail call in setter: https://bugs.webkit.org/show_bug.cgi?id=164306
+    return value
 
   getv: (prop_name) ->
     if not @props[prop_name]?
```
This is was the PR I was going to make, which seems to work when testing

bryevdv:
I will test your PR tho, if it works its more minimal

bryevdv:
Confirm I can't reproduce the crash anymore with this change. @mattpap I will merge when green unless you have anything else.

bryevdv:
@mattpap it looks like the docs failure in an "outside" PR due to missing google API key, made me miss an earlier failure:
```
E         File contains trailing whitespace: bokeh/models/widgets/inputs.py, line 46.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: bokeh/models/widgets/inputs.py, line 46.'])
``` 
Can you fix that up in this PR?

mattpap:
> I added them in the deprecated ones too:

That would be unnecessary. If they say it's just getters and setters, then we shouldn't be adding code "just in case".

bryevdv:
actually I just went ahead and fixed the code quality should be good now

bryevdv:
> That would be unnecessary. If they say it's just getters and setters, then we shouldn't be adding code "just in case". 

Right I mistakenly thought those were also setters then realized they weren't

mattpap:
> actually I just went ahead and fixed the code quality should be good now

@bryevdv, so should I revert last commit?

bryevdv:
> @bryevdv, so should I revert last commit?

? I pushed the whitespace fix to this branch, it should go green on the current build

mattpap:
I pushed as well. Looks like mine is gone.

mattpap:
Nevermind, I pushed nothing. I didn't see push was rejected.

PR #5447, opened 2016-11-07
wrap glyph methods in a better signature

### LINKED ISSUE #5446: improve glyph method function signatures
Came out of some observations and discussion in:

https://groups.google.com/a/continuum.io/forum/?pli=1#!topic/bokeh/szlk3wwuTnk

It's possible to achieve this, PR incoming:

screen shot 2016-11-07 at 11 21 05 am

### LINKED ISSUE #5466: Select widget on_change breaks figures in other tabs
I have two tabs, the first containing one figure, the other containing a different figure with a Select widget. When selecting a value in the Select widget, the figure in tab1 breaks (it seems to become extremely small), despite the widget only changing the figure in tab2.

This is with Bokeh version 0.12.4dev6.

A simplified example is below:
```
import numpy as np
from bokeh.io import curdoc
from bokeh.layouts import WidgetBox, column
from bokeh.models.widgets import Select, Panel, Tabs


def create_fig(a, index):
    n_d2 = a.shape[1]

    x = np.arange(n_d2)
    y = a[index]

    print(a.shape)
    print(index)
    print(a[index])

    fig = figure(width=900, height=360)
    c = fig.circle(x=x, y=y, hover_color="red")

    return fig


def update(attr, old, new):
    group.children[1] = create_fig(array, select.value)

array = np.random.randint(0,1000,(10,5))
options = ['{:.0f}'.format(x) for x in range(array.shape[0])]

# Get figures
fig1 = create_fig(array, 0)
fig2 = create_fig(array, 1)

select = Select(title="i:", value='0', options=options)
wb = WidgetBox(select)

tab1 = Panel(child=fig1, title='1')
group = column([wb, fig2])
tab2 = Panel(child=group, title='2')
tabs = Tabs(tabs=[tab1, tab2])

select.on_change('value', update)

curdoc().add_root(tabs)
```

bryevdv:
issues: fixes #5446 

With this change, e.g. help for `p.square` looks like this in Jupyter notebooks:

screen shot 2016-11-07 at 11 35 01 am

@mattpap @birdsarah if either of you have comments or suggestions for tests that could be added please let me know


bryevdv:
Note: I had to change the order of `left` and `right` in `HBar._args` to satisfy python (can't have non-default arg after a default arg)

bryevdv:
Well, this has a problem to fix (hopefully)

bryevdv:
OK , I think the current code will now work in all cases, but since I took the opportunity to do a little refactoring, I will also add some tests first.

bryevdv:
Actually I want to make some more comprehensive refactoring about docstring generation, I will make a testing strategy then. This seems like a small clear improvement. 

I also took the opportunity to remove some completely unused functions I happened to discover.

bryevdv:
@mattpap any comments on this? If not I will merge later today.

bokeh/plotting/helpers.py
```diff
@@ -466,10 +514,6 @@ def func(self, *args, **kwargs):
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
 
-        # add the positional arguments as kwargs
-        attributes = dict(zip(glyphclass._args, args))
-        kwargs.update(attributes)
-
```
bryevdv: this is now handled by the wrapper with the real function signature

bokeh/plotting/helpers.py
```diff
@@ -505,38 +549,16 @@ def func(self, *args, **kwargs):
...
-            # this awkward syntax is needed to go through Property.__set__ and
-            # therefore trigger a change event. With improvements to Property
-            # we might be able to use a more natural append() or +=
-            tool.renderers = tool.renderers + [glyph_renderer]
+            tool.renderers.append(glyph_renderer)
+
+        self.renderers.append(glyph_renderer)
 
-        # awkward syntax for same reason mentioned above
-        self.renderers = self.renderers + [glyph_renderer]
```
bryevdv: The property containers now detect changes like this correctly, so these can be simplified

bokeh/plotting/helpers.py
```diff
@@ -505,38 +549,16 @@ def func(self, *args, **kwargs):
...
-            desc = ""
-        kwlines.append(_arg_template % (kw, typ, desc, spec.class_default(glyphclass)))
+    _add_sigfunc_info(sigfunc, argspecs, glyphclass, extra_docs)
 
-    func.__doc__ = _doc_template % (glyphclass.__name__, "\n".join(arglines), "\n".join(kwlines))
-
-    if extra_docs:
-        func.__doc__ += extra_docs
-    return func
+    return sigfunc
```
bryevdv: All of this was reduced to:
```
argspecs = _get_argspecs(glyphclass)
sigfunc = _get_sigfunc(glyphclass.__name__.lower(), func, argspecs)
_add_sigfunc_info(sigfunc, argspecs, glyphclass, extra_docs)
return sigfunc
```
and those three functions are defined above

bokeh/util/functions.py
```diff
@@ -1,29 +0,0 @@
...
-            return self
-        value = obj.__dict__[self.func.__name__] = self.func(obj)
-        return value
-
-def arg_filter(func, input_dict):
-    arg_names, vararg_name, keyword_name, defaults = inspect.getargspec(func)
-    result = {}
-    for k in arg_names:
-        result[k] = input_dict.get(k)
-    return result
```
bryevdv: just taking the opportunity to remove this wholly unused module while thinking about it.

bokeh/plotting/helpers.py
```diff
@@ -426,6 +409,49 @@ def _process_active_tools(toolbar, tool_map, active_drag, active_scroll, active_
...
+def _get_argspecs(glyphclass):
+    argspecs = OrderedDict()
+    for arg in glyphclass._args:
+        spec = {}
+        prop = getattr(glyphclass, arg)
+        spec['desc'] = " ".join(x.strip() for x in prop.__doc__.strip().split("\n\n")[0].split('\n'))
+        spec['default'] = prop.class_default(glyphclass)
+        spec['type'] = prop.__class__.__name__
+        argspecs[arg] = spec
+    return argspecs
```
bryevdv: Perhaps this could go directly on `Glyph` classes? I'd feel better about adding a test to it there

bokeh/plotting/helpers.py
```diff
@@ -453,10 +479,32 @@ def _process_active_tools(toolbar, tool_map, active_drag, active_scroll, active_
...
+            desc = ""
+        kwlines.append(_arg_template % (kw, typ, desc, prop.class_default(glyphclass)))
+
+    arglines = []
+    for arg, spec in argspecs.items():
+        arglines.append(_arg_template % (arg, spec['type'], spec['desc'], spec['default']))
+
+    func.__doc__ = _doc_template % (func.__name__, "\n".join(arglines), "\n".join(kwlines))
+    if extra_docs:
+        func.__doc__ += extra_docs
```
bryevdv: This adds `__name__` and `__doc__` to the wrapper. Anything else? Can't use `functools.wraps` because that seems to demolish the signature at least in python 3.x (I think there was a feature added to make the reported signature overridable, and `wraps` seems to use it, resulting in `*args, **kw` signature all over again)

bryevdv:
@mattpap I left some comments and one question to aid review

bokeh/plotting/helpers.py
```diff
@@ -50,23 +50,6 @@ def get_default_alpha(plot=None):
...
-    ----------
-    %s
-    Additionally, the following properties are accepted as keyword arguments: %s
-
-    Returns
-    -------
-    plot : :py:class:`Plot <bokeh.models.Plot>`
-    """ % (desc, params, props)
-
-
```
bryevdv: this is no longer used anywhere

mattpap:
@bryevdv, I presume this was sufficiently tested on Python 3 (or 2 whichever is your default)? If so, LGTM.

bryevdv:
yah tested with python 2 just now.

PR #5451, opened 2016-11-10
Add support for *.tsx source files

### LINKED ISSUE #5450: Add support for *.tsx source files
TSX is TypeScript's take on JSX, which are XML extensions to JavaScript (used mostly, but exclusively, with React). We can take advantage of the nice syntax, hook our own code for generating HTML/DOM, and use this as a more robust replacement for Eco templates. Eco templates are neither well defined (it's a mixture of HTML and CoffeeScript-ish kind of language) and won't work well together with TypeScript (Eco sets new attributes on instances of built-in `String`).

mattpap:
Added support for `*.tsx` (pretty naive see `core/util/dom.ts`) and replaced `toolbar_template.eco` with `toolbar_template.tsx`. This TSX template could be less verbose, but unfortunately TypeScript inherits a lot of garbage from ES, so this is the best we can achieve. For now I converted only one template, as an experiment, but we can follow up with changing the rest. 

fixes #5450


bokehjs/src/coffee/models/tools/toolbar_template.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface ToolbarProps {
+  location: "above" | "below" | "left" | "right";
+  sticky: "sticky" | "non-sticky";
+  logo: "normal" | "grey" | null;
+}
+
+export default (props: ToolbarProps): HTMLElement => {
+  let logo;
+  if (props.logo != null) {
+    const cls = props.logo === "grey" ? "bk-grey" : null;
```
mattpap: No `if` expressions, but we want immutability, so we go back to dark ages.

bokehjs/src/coffee/models/tools/toolbar_template.tsx
```diff
@@ -0,0 +1,30 @@
...
+    logo = <a href="http://bokeh.pydata.org/" target="_blank" class={["bk-logo", "bk-logo-small", cls]}></a>
+  }
+
+  return <div class={[`bk-toolbar-${props.location}`, `bk-toolbar-${props.sticky}`]}>
```
mattpap: I would prefer to have this:
```
  return
   <div ...>
      ...
   </div>
```
instead, but that won't work, because of semicolon inference in ES and thus TypeScript. The above code compiles to:
```
  return;
  <div ...>
    ...
  </div>
```
That's a lot of fun. At least TypeScript warns about unreachable code.

bryevdv:
I'm certainly almost always going to be +1 on removing tools from the toolchain. If we decide to proceed we should have one PR that does all the templates at once and removes all traces of eco at the same time. Regarding verbosity: the templates change only very rarely so I think some slight increase there can be tolerated in exchange for reducing the number of tools people have to interact with.

mattpap:
> If we decide to proceed we should have one PR that does all the templates at once and removes all traces of eco at the same time.

This is my attitude as well, but first I wanted to know your opinion on this.

bryevdv:
I'm +1 for sure. In terms of priorities, I think if it's only a couple of hours work, great. Otherwise there are probably higher immediate priorities.

mattpap:
> Regarding verbosity: (...)

This verbosity is a temporary thing, because when using TSX approach, we don't need templates at all. The reason we used templates is because it's cumbersome to include HTML in CoffeeScript. With TSX, we can maintain HTML together with rendering code. After all, our "templates" aren't actual templates, but snippets (so, small templates) and they can be nicely integrated with code. It's even better, because you can take functional approach and split those snippets into functions and compose HTML nicely. I've been using this approach in web apps I write in Scala (and not only, in bokeh-scala as well), because Scala supports in-line XML.

mattpap:
> Otherwise there are probably higher immediate priorities.

I know what are the priorities, but I also want to do work that will allow sufficient stability of bokehjs for 1.0. Getting rid of Eco template is part of this work. Anyway, this PR costed about one hour of work for now and rewriting templates will take another hour to two. Testing is the pain here.

mattpap:
Apparently some of lesser used widgets (e.g. dropdowns) are broken in `master`.

bokehjs/src/coffee/models/tools/toolbar_template.tsx
mattpap: Actually, we can use the following syntax:
```
  return (
    <div ...>
      ...
    </div>
  )
```
to maintain indentation.

mattpap:
@bryevdv, there is some issue with tabs widget and layout, because tabs widget doesn't render properly when its children aren't plots. I used buttons and it failed on me all the time. The issue is that at first rendering only children are shown and not tabs widget. Resizing browser window restores proper view. I'm not sure what's the underlying cause, whether this is layout itself or how tabs widget is implemented. I presume both.

mattpap:
I restored all the functionality we had in Eco templates and fixed a few widgets. Tabs work as explained above and calendar widget is broken (styling doesn't apply, but I think it was this way for a while).

mattpap:
btw. Do we actually have any use for `DatePicker` widget?

mattpap:
Integration tests fail, because logo's CSS is sensitive to whitespace.

bryevdv:
> btw. Do we actually have any use for DatePicker widget?

A use, certainly, but its styling is badly broken currently, AFAIK

mattpap:
> (...) but its styling is badly broken currently

It think it's not only currently, but it has been for 1.5+ years (since introduction of CSS class prefixes). If no one complained for that amount of time and given that we never had a single example using it, then "a use" seems more like "no use".

mattpap:
@bryevdv, did you have a chance to review this?

bryevdv:
Looking at it now

bokeh/util/compiler.py
```diff
@@ -160,12 +160,6 @@ class JavaScript(Inline):
     def lang(self):
         return "javascript"
 
-class Eco(Inline):
-
-    @property
-    def lang(self):
-        return "eco"
-
```
bryevdv: I very much doubt any users were using this but rates a mention in migration notes

mattpap: @bryevdv, this wasn't released yet. I added this for ion-range-slider example.

mattpap: Actually it may have been in 0.12.3.

sphinx/source/docs/user_guide/examples/extensions_ion_range_slider_template.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div class="bk-slider-parent">
+      {title}
+      {value}
+      <div class="bk-slider-horizontal">
+        <input type="text" class="slider" id={props.id}></input>
+      </div>
+    </div>
+  )
+}
```
bryevdv: needing `.tsx` instead of `.eco` in any extensions also needs a migration note. It would also be good to at least include a mention of `.tsx` and maybe a link to some external docs for it in the "Extending Bokeh" users guide section.

bryevdv:
Generally LGTM. A few small comments. I'd like to run the new example but it will be a few hours.

mattpap:
> I'd like to run the new example (...)

This new example doesn't add anything, just simplifies testing as all widget are conveniently in one place.

bryevdv:
@mattpap these two files were renamed to `._py` to keep them breaking the docs build:
```
extensions_example_widget._py  extensions_example_widget_plot._py
```
When I run either one, with `BOKEH_RESOURCES=inline` I get 
```
RuntimeError: no such module: ../../core/util/dom
```

these should also be rename `.py` when they are working

bryevdv:
This change fixed both:
```
diff --git a/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider_template.tsx b/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider_template.tsx
index 2c41390..63269f4 100644
--- a/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider_template.tsx
+++ b/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider_template.tsx
@@ -1,4 +1,4 @@
-import * as DOM from "../../core/util/dom";
+import * as DOM from "core/util/dom";

 interface RangeSliderProps {
   id: string;
``` 
I'll make the change and fix up the sphinx docs.

mattpap:
> `import * as DOM from "core/util/dom"`;

That's the correct version.

bryevdv:
Unfortunately after making that change and renaming the file, the docs fail to build with either the split up, or inline version. It reports:
```
RuntimeError: no such module: ./extensions_ion_range_slider_template
```

bryevdv:
For reference you I am building the the docs with 
```
BOKEH_DOCS_CDN=local make SPHINXOPTS=-v clean html serve
```

mattpap:
@bryevdv, did this example work before? It smells like #5425.

bryevdv:
yes you are probably right, in that case just make the one change so the files run when run locally but not any of the filename changes or docs un-commenting.

PR #5454, opened 2016-11-11
Fix tab completion for bokeh.palettes

### LINKED ISSUE #5453: Tabe completion on bokeh.palettes doesn't work in python 2
Due to a quirk in python 2, the global `_Palettes` is `None` in the global namespace of the dynamically generated `palettes` module. This causes the implementation of `__dir__` to fail in python 2.

```python
from bokeh import palettes
assert 'viridis' in dir(palettes)   # Fails on python 2
```

jcrist:
Fixes the implementation of `__dir__` on the dynamically generated
`palettes` module. Previously this would fail on python 2, due to a
quirk in how globals were handled (this is fixed in python 3).

Fixes #5453.

### DIFF
--- a/bokeh/palettes.py
+++ b/bokeh/palettes.py
@@ -82,7 +82,10 @@ The complete contents of ``small_palettes`` is show below.
 
 """
 
-class _Palettes(object):
+import sys as _sys
+import types as _types
+
+class _PalettesModule(_types.ModuleType):
     @property
     def YlGn3(self):
         return ["#31a354", "#addd8e", "#f7fcb9"]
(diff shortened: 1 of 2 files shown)


jcrist:
Test failures appear to be unrelated.

bryevdv:
@jcrist yes the one failure is due to a missing credential that is only available to PRs pushed by members of @bokeh/dev I still need to find a way to disable that one specific thing when "outside" PRs arrive. Thanks for the PR, I will merge as soon as the last unit test build turns green

jcrist:
Thanks!

PR #5459, opened 2016-11-11
Fix off-by-one and missing test in LinearInterpolator

### LINKED ISSUE #5370: LinearInterpolator does not work correctly
I am on bokeh 0.12.3 and just tried to replicated the youtube tutorial (https://www.youtube.com/watch?v=9FlUFLmaWvY) using gapmingder data set. However, after following command:

size_mapper=LinearInterpolator(
x=[data.population.min(), data.population.max()],
y=[5,50]
)

China was completely missing on my Bokeh plot. After some fail and trial, I found that I had to change to " data.population.max()+1" to get China back. I don't think it is supposed to work like that. Correct me if I am wrong. 

BTW, the gapminder population data I got was different from the tutorial. The population data ranged from 1952 to 2007. Min and max values were (60011, 1318683096). Thanks.

Add bokeh info:

Python version      :  2.7.12 |Anaconda custom (64-bit)| (default, Jul  2 2016, 17:42:40) 
IPython version     :  5.1.0
Bokeh version       :  0.12.3
BokehJS static path :  /home/da/anaconda2/lib/python2.7/site-packages/bokeh/server/static

bryevdv:
- [x] issues: fixes #5370
- [x] tests added / passed

There was no test that the final control point matched, which is how this bug crept in. The problem was the predicate for `_.findLastIndex`, which was `x >= num`. This could return the last index for `@_x_sorted`, and the subsequent `x2 = @_x_sorted[ind+1]` would then yield `null` -> `NaN`.

The predicate was changed to `num < x` which will always be at most the next to last index. But this also necessitates special casing the first control point, so the following check was added first:
```
if x == @_x_sorted[0]
  return @_y_sorted[0]
```

### DIFF
--- a/bokehjs/src/coffee/models/transforms/interpolator.coffee
+++ b/bokehjs/src/coffee/models/transforms/interpolator.coffee
@@ -78,7 +78,7 @@ export class Interpolator extends Transform
         return ((a.x > b.x) ? -1 : ((a.x == b.x) ? 0 : 1));
       )
 
-    for k in [0..list.length-1]
+    for k in [0...list.length]
       @_x_sorted[k] = list[k].x;
       @_y_sorted[k] = list[k].y;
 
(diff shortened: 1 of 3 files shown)


bryevdv:
ping @birdsarah

bryevdv:
results with the change:

screen shot 2016-11-11 at 5 11 04 pm

Also verified that there were no `NaN` anywhere in the output for the test notebook via the JS console debugger.

bryevdv:
Well, all old and new tests pass, and the output is now correct. As this change does not seem controversial, going to go ahead and merge.

Issue #5306, opened 2016-10-05
embed/embed_multiple fails with unicode error on py2


mattpap:
```
$ python embed_multiple.py 
Traceback (most recent call last):
  File "embed_multiple.py", line 80, in <module>
    f.write(html)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 672004: ordinal not in range(128)
```

mattpap:
We have a tag for py3 but not py2.

joshua-barber:
I am encountering the exact same issue using PY3 and Anaconda.

bryevdv:
I can't reproduce this on python 3, @joshua-barber can you provide more system/platform information.

joshua-barber:
I am using the [Jupyter data-science docker image](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook). Package management on the image is handled by Anaconda, which I used to conda install Bokeh. All other Bokeh examples run with no issue; it is only the 'multiple embed' examples that throw the error @mattpap referenced. I haven't configured a Py2 env with Bokeh so I cannot confirm whether it occurs on **both** Py2 and Py3.

I will update this comment with more specific version info once I am at my dev laptop this afternoon.

joshua-barber:
**[Correction**] The docker image is actually generated from this Dockerfile: [dockerfile.txt](https://github.com/bokeh/bokeh/files/591448/dockerfile.txt)

The referenced environment.yml is used to install:
- python=3.5
- Flask=0.11.1
- sqlalchemy=1.0.13
- psycopg2=2.6.2
- pandas=0.18.1
- pyyaml=3.12

A clean install of the Jupyter data-science image actually runs the examples with no issue. However I'm not clear on what it is in this docker image that is causing _only_ embed.py to fail and none of the other Bokeh examples.

bryevdv:
@joshua-barber if you check out `master` and run the examples from there you should be able to confirm that they now function as expected. Please re-open if there is any continued problem.

PR #5465, opened 2016-11-15
Fixed write file error in examples; encode to UTF-8 (#5306

amyxst:

- [ ] issues: fixes #xxxx
- [ ] tests added / passed

### DIFF
--- a/examples/embed/embed_multiple.py
+++ b/examples/embed/embed_multiple.py
@@ -77,6 +77,6 @@ html = template.render(js_resources=js_resources,
                        div=div)
 
 with open(filename, 'w') as f:
-    f.write(html)
+    f.write(html.encode('utf8'))
 
 view(filename)
--- a/examples/embed/embed_multiple_responsive.py
+++ b/examples/embed/embed_multiple_responsive.py
@@ -63,6 +63,6 @@ html = template.render(js_resources=js_resources,
 filename = 'embed_multiple_responsive.html'
 
 with open(filename, 'w') as f:
-    f.write(html)
+    f.write(html.encode('utf8'))
 
 view(filename)


examples/embed/embed_multiple.py
```diff
@@ -77,6 +77,6 @@
                        div=div)
 
 with open(filename, 'w') as f:
-    f.write(html)
+    f.write(html.encode('utf8'))
```
mattpap: A better solution would be to use `io.open(..., encoding='utf-8')`, especially using it uniformly across code base.

amyxst:
Hi Matt,

I apologize this is my first time attempting to contribute to a project.
Would you suggest that I look through the codebase to find other areas
where the encoding may need to be fixed as well?

Best,

Amy

bryevdv:
Hi @amyxst no apologies, thanks for the contribution! So all of the "higher level" output and embed functions like `file_html` or `autoload_server` in Bokeh automatically use the `encode_utf8` function in this module:

https://github.com/bokeh/bokeh/blob/master/bokeh/util/string.py

So I think the only other areas that might need fixing would be any other example that do a `.write` on raw strings that didn't come from `file_html`, etc. It looks like there are are few other *potential* cases:
```
./embed/animated.py:
   63 :     f.write(html)
./embed/embed_multiple.py:
   80 :     f.write(html)
./embed/embed_multiple_responsive.py:
   66 :     f.write(html)
./embed/widget.py:
  177 :     f.write(html)
./howto/interactive_bubble/gapminder.py:
  128 :     f.write(html)
```
But the `html` variables might have from from `file_html`, `autoload_server`, etc. (in which case everything is fine) but it might be worth a quick look at them. 

Regarding the actual PR, I agree with @mattpap that `io.open(..., encoding='utf-8')` might be a better solution, or alternatively you could use the `encode_utf8` function I mentioned like:
```
 f.write(encode_utf8(html))
```
The reason is that there are some differences between py2 and py3 that these ways account for.

bryevdv:
@amyxst just checking in, do you think you will have an opportunity to update the PR in one of the ways described above (using `io,open` or `encode_utf8`)?

bryevdv:
Hi @amyxst I would like to get this fix in for `0.12.4` but I do think that the solution either needs to use `io.open` as @mattpap described, or the `bokeh.util.string.encode_utf8` as I described. I appreciate that you may not have more time to devote to this, but I'd like to give you an opportunity to close this PR, if you are able. Otherwise this is just a heads up that I will close this PR next week and submit a PR with one of the fixes described above.

bryevdv:
superseded by #5535 

@amyxst thanks for doing the initial work on this, I have closed the issue in a separate PR

PR #5468, opened 2016-11-16
Correcting docstring for function push_notebook

### LINKED ISSUE #5467: Docstring not reflecting correct function signature.
https://github.com/bokeh/bokeh/blob/master/bokeh/io.py#L572

Docstring references invalid function call. 

is: push_notebook(handle)
should be: push_notebook(handle=handle)

tag: docs

sphinxligustri:
#issues: fixes #5467

Correcting docstring for function push_notebook. Adding missing key word, "handle".
https://github.com/bokeh/bokeh/blob/master/bokeh/io.py#L572

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -569,7 +569,7 @@ def push_notebook(document=None, state=None, handle=None):
 
             # Update the plot title in the earlier cell
             plot.title = "New Title"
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the catch and fix @sphinxligustri !

Issue #5389, opened 2016-10-20
Creating a line plot with `x_axis_type='log'` fails when `x_max < 1`


StevenCHowell:
When creating a plot with a logarithmic scale on the x-axis, if the maximum x-value is less than 1 the x-axis decreases from left to right, the scale is linear, not logarithmic, and the default view does not show the data.  I tested various values for the maximum x-value and found that when x_max > 1, the x-axis increases left to right but the default view only shows x > 1.  This code demonstrates the problem

```
import numpy as np
from bokeh.plotting import figure, output_file, show

output_file('test.html')

x = np.linspace(0, 0.95) # x decreases left to right, linear scale
# x = np.linspace(0, 1) # x increases left to right but default view shows, 1 < x < 1.05
# x = np.linspace(0, 10) # x increases left to right but default view shows, 1 < x < 10.5
y = np.sin(x)/x

p = figure(x_axis_type='log')
p.line(x, y)

show(p)
```

I tried declaring the x_range for the figure using `x_range=(x.min(), x.max())` but this had to effect unless I removed the zero point from the data, i.e., `x = x[x > 0]`.  With theses changes, the x-axis increases left to right and the default view shows the range indicated.  Here is a functional workaround

```
import numpy as np
from bokeh.plotting import figure, output_file, show

output_file('test.html')

x = np.linspace(0, 0.95) # x decreases left to right, linear scale
x = x[x > 0]
y = np.sin(x)/x

p = figure(x_axis_type='log', x_range=(x.min(), x.max()))
p.line(x, y)

show(p)
```

I did not test all these same conditions with `y_axis_type` but I expect it to have the same problems.  I did notice that the default view only shows `y > 1`, even if the data begins ay `y < 1`.

Here are the key points that need to be verified for logarithmic x and y axes:
- [ ] make sure the x-axis decreases left to right when the maximum x-value is less than 1
- [ ] make sure the y-axis increases bottom to top when the maximum y-value is less than 1
- [ ] adjust the default view to ignore x<=0 or y<=0 but show all other data (including x<1 and y<1)

StevenCHowell:
Unfortunately, this issue was only partially resolved by #5477.  The axis is still rendered incorrectly if the maximum x-value is ~1 or less.

Here is an example
```
import bokeh
print(bokeh.__version__)

import numpy as np

from bokeh.plotting import figure, show, output_notebook
from bokeh.layouts import gridplot
from bokeh.palettes import Colorblind8 as palette

[...]

Issue #2789, opened 2015-08-31
Range padding possibly discards the log axis properties


ilayn:
This similar to #2431 but in the log-log axis context. If I plot the following 

    x = np.arange(80,2500)
    s = ColumnDataSource(data=dict(x=x,y=y))
    dd = figure(x_axis_type='log',y_axis_type='log')
    dd.line(x,x)
    show(dd)

I get 
![first](https://cloud.githubusercontent.com/assets/1303842/9571643/eab6a73a-4fa0-11e5-8b12-be5ef519ab42.png)

which is a bit strange for the auto padding. But if I plot the following

    x = np.arange(1,20)
    s = ColumnDataSource(data=dict(x=x,y=y))
    dd = figure(x_axis_type='log',y_axis_type='log')
    dd.line(x,x**2)
    show(dd)

![second](https://cloud.githubusercontent.com/assets/1303842/9571664/37ebc33c-4fa1-11e5-86b3-a671cd828eff.png)

In general, I'm trying to understand the reference guide properly. For example, if I read the `DataRange1d` part, I understand that it has a range-padding but then is it the default range object for the plots ?

If I want to explicitly supply an auto range with a different padding value should I go about it as 
`x_r = DataRange1d(range_padding=0.2)` and then supply this to the figure `x_range` attribute? It would be really nice if this can be supplied directly to `x_range`.

Finally, I think in the `DataRange1d` docs, it should be called the fraction but not the percentage as *A percentage of the total range size to add as padding to the range start and end.* sounds like I should enter 20 instead of 0.2 but that makes the range huge.

bryevdv:
@ilayn using `figure`, plots get `DataRange1d` by default, unless you explicitly provide bounds with e.g., `x_range=(0,10)` or similar. So if you want to change the padding you could do:
```
dd = figure(x_axis_type='log',y_axis_type='log')
dd.x_range.range_padding = 0.2
```
I agree the wording is better with fraction. Would you be interested in submitting a PR to change the wording in the docstring?

In general I don't think much thought has been given to the integration of `range_padding` with log axes. I can imagine is it not currently ideal, and might need some development to make it behave better.

ilayn:
Ah, thanks. I tried a few things as such but never occured to me that depending on the existence of fixed range the underlying range options change. 

Because I've received the error message 

    AttributeError: unexpected attribute 'range_padding' to Range1d, possible attributes are callback, end, name, session, start or tags

[...]

PR #5477, opened 2016-11-20
Fix for log scale plotting bug

### LINKED ISSUE #3834: plot is empty when log scale is used
Bokeh incorrectly creates an empty plot with the following:

``` python
ys = [4.471799184102565e-05, 0.0009856299875536934, 0.0011045119899790734]
xs = range(len(ys))
output_file("log.html")
p = figure(y_axis_type="log")
p.line(xs, ys)
show(p)
```

The problem only occurs if `y_axis_type="log"` is used.

It seems when the range of y values is too narrow bokeh fails to compute the y-axis range.  If I multiple the y values by 100, it plots properly.

Tested on bokeh version:  0.11.1 py34_0

clairetang6:
This PR is for fixing a bug in log plotting, where plots are empty or where axes use incorrect initial ranges under some circumstances (related to the minimum and maximum value of the data and the `range_padding`). This bug is caused when a DataRange1d determines that its `start` should be a negative value. This can happen even when all the data are positive values (necessary to be shown on a log plot since log(x) is only defined for x>0) because the `range_padding` calculation is done linearly. 

There are three main ways of approaching a fix. 
1. Have the DataRange1d (or something above it) check whether it's being used linearly or log-ly. 
2. Have the DataRange1d know whether it's linear or log. 
3. Delegate the range bounds calculation to the tickers. 

This PR implements approach 2. by adding a new property `mapper_type` to the DataRange1d. Then the range calculation can be done correctly whether linear or log by the DataRange1d. 

Cons for approach 1.
-It would be a little convoluted to figure out whether the DataRange1d is linear or log. The  DataRange1d doesn't know whether it is an x_range or y_range. The plot_canvas that updates the  DataRange1d knows which dimension the DataRange1d is, but then would need to find the matching Axis from its Plot's `renderers` to inspect that Axis's ticker. 

Cons for approach 3. 
-Ranges are currently in charge of figuring out data-derived bounds which are compared to user-supplied bounds to determine the bounds. If the ranges no longer take into account the padding into their data-derived bounds, then the comparison between data and user bounds will need to use logic moved into the ticker. 
-Range `start` and `end` values are set by pan/zoom tools. If the ticker defaults to adding padding, the ticker will need to know not to add padding when panning/zooming. 

Some other things to consider:
-When a user creates a DataRange1d to use in a log plot instead of using the default, they will need (not absolutely, but they will probably want) to supply `mapper_type="log"` if they don't supply a `start` and `end` and want the correct range padding. If they do supply a `start` and `end` (and they are both positive values), but don't supply a `mapper_type="log"`, then the plot will still be correct. This is because the ticker and mapper are still log and will produce the correct ticks and screen-mapped values. 

- [x] issues: fixes #3834 #5389 #2789
~~- [ ] support different exponential bases~~ (Log bases are all scaled the same, so my original thinking about supporting different bases is not needed. `default_span` could be made to span in terms of powers of `base` instead of only powers of 10, but that would minimally affect ux)
- [x] tests added / passed
~~- [ ] release document entry (if new feature or API change)~~

### DIFF
--- a/bokehjs/src/coffee/models/ranges/data_range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/data_range1d.coffee
@@ -87,9 +87,9 @@ export class DataRange1d extends DataRange
 
       if @_mapper_type == "log"
         log_min = Math.log(min) / Math.log(10)
-        log_max = Math.log(max) / Math.log(10) # TODO bases
+        log_max = Math.log(max) / Math.log(10)
         if max == min
-          span = Math.log(@default_span) / Math.log(10)
+          span = @default_span + 0.001
         else
           span = (log_max-log_min)*(1+range_padding)
         center = (log_min+log_max) / 2.0
--- a/bokehjs/src/coffee/models/ranges/data_range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/data_range1d.coffee
@@ -9,15 +9,14 @@ export class DataRange1d extends DataRange
   type: 'DataRange1d'
 
   @define {
-      start:           [ p.Number         ]
-      end:             [ p.Number         ]
-      range_padding:   [ p.Number, 0.1    ]
-      mapper_type:     [ p.String, "auto" ]
-      flipped:         [ p.Bool,   false  ]
-      follow:          [ p.String         ] # TODO (bev)
-      follow_interval: [ p.Number         ]
-      default_span:    [ p.Number, 2      ]
-      bounds:          [ p.Any            ] # TODO (bev)
+      start:           [ p.Number        ]
+      end:             [ p.Number        ]
+      range_padding:   [ p.Number, 0.1   ]
+      flipped:         [ p.Bool,   false ]
+      follow:          [ p.String        ] # TODO (bev)
+      follow_interval: [ p.Number        ]
+      default_span:    [ p.Number, 2     ]
+      bounds:          [ p.Any           ] # TODO (bev)
       min_interval: [ p.Any ]
       max_interval: [ p.Any ]
     }
@@ -34,6 +33,7 @@ export class DataRange1d extends DataRange
     @_initial_follow = @follow
     @_initial_follow_interval = @follow_interval
     @_initial_default_span = @default_span
+    @_mapper_type = "auto"
 
   @getters {
     min: () -> Math.min(@start, @end)
@@ -85,7 +85,7 @@ export class DataRange1d extends DataRange
     range_padding = @range_padding
     if range_padding? and range_padding > 0
 
-      if @mapper_type == "log"
+      if @_mapper_type == "log"
         log_min = Math.log(min) / Math.log(10)
         log_max = Math.log(max) / Math.log(10) # TODO bases
         if max == min
@@ -136,13 +136,13 @@ export class DataRange1d extends DataRange
     [start, end] = @_compute_range(min, max)
 
     if @_initial_start?
-      if @mapper_type == "log"
+      if @_mapper_type == "log"
         if @_initial_start > 0
-          start = @_initial_start 
+          start = @_initial_start
       else
         start = @_initial_start
     if @_initial_end?
-      if @mapper_type == "log"
+      if @_mapper_type == "log"
         if @_initial_end > 0
           end = @_initial_end
       else
--- a/bokehjs/src/coffee/models/ranges/data_range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/data_range1d.coffee
@@ -9,14 +9,15 @@ export class DataRange1d extends DataRange
   type: 'DataRange1d'
 
   @define {
-      start:           [ p.Number        ]
-      end:             [ p.Number        ]
-      range_padding:   [ p.Number, 0.1   ]
-      flipped:         [ p.Bool,   false ]
-      follow:          [ p.String        ] # TODO (bev)
-      follow_interval: [ p.Number        ]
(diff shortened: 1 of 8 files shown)


bryevdv:
I might propose a slightly different approach. I am definitely not 100% keen on having `Ranges` know about mappers. But if we must, think it would be done more simply with just:
```
mapper = Instance(Mapper)
```
on the range. Since the can be `None` (and defaults to `None`) this property could be considered an optional hint to the range about the mapper type that will be using it. And by setting the actual mapper itself, we avoid an additional enum and all the indirect logic needed to process the enum (we can just inspect the optional mapper directly). 

Possibly taking this further (maybe later) the mapper could expose things like min bounds that that could be taken into account (i.e. a min bound of `"auto"` would mean, see if I have a hinted mapper, and ask it what if any min bound it requires).

clairetang6:
Thinking about this more, the only reason the data_range1ds need to know about the "mapper_type" is to calculate the range-padded bounds, so that the range start/end are nice by default when the user doesn't supply a start/end. So, the "mapper_type" is less about any kind of mapping per se, but is more about how to think about padding the data values. There could a scenario where a user wants the mapping to be linear while the range padding is log or vice versa. Although those scenarios are unlikely, they at least illustrate that the two concepts are independent. So, it may be more correct for the range to have its own independent "mapper_type" (could be called something else). 

That said, the much more common scenario (probably close to 100% of the time) of having the range and mapper match could still be covered by either your proposed `mapper = Instance(Mapper)` or another solution described below. The problem with `mapper = Instance(Mapper)` is that it doesn't look like the LinearMapper and LogMapper are defined on the Python side.

So instead, I tried removing the property `mapper_type` (that I added to data range) and instead adding a private variable `_mapper_type`.  The `_mapper_type` is set to "auto" when a data_range1d is initialized and then is changed by the frame when the frame creates its mappers (and knows whether to create a Linear or Log mapper).

bryevdv:
@clairetang6 Ah right, I'd forgotten. The current handling of mappers is an unfortunate holdover from very early days. We used to handle lots of things by sending string enums and assembling things on the BokehJS side. It became universally clear that assembling things on the python side was better for a variety of reasons, so we stopped doing that pretty much everywhere when our properties system became capable of handling `Instance` well. But mappers and frames are the holdouts that have not been refactored or fixed up yet. That's a larger task, so for this issue I think just forwarding the mapper type as a private var on the JS side is reasonable.

bokehjs/src/coffee/models/ranges/data_range1d.coffee
```diff
@@ -125,9 +136,17 @@ export class DataRange1d extends DataRange
...
+          start = @_initial_start
+      else
+        start = @_initial_start
     if @_initial_end?
-      end = @_initial_end
+      if @_mapper_type == "log"
+        if @_initial_end > 0
+          end = @_initial_end
+      else
+        end = @_initial_end
```
bryevdv: The effect of these two blocks is to silently ignore any values less than zero in the presence of lag mappers, correct?  A log message is probably warranted

StevenCHowell: In my experience using matplotlib and matlab, plotting data on a log scale does silently ignore values less than zero.  This should be understood as part of the logarithm function.

bryevdv: Well MPL at least has `symlog` for which will linearize around zero and allow for negative values, but the regular `log` mode has both `clip` and `mask` options I thought, and in particular, that only applies to *values* AFAICT. You can still set range bounds < 0 if you like, and they are honored as best I can tell. If we are going to prevent *ranges* from extending to negative in the presence of a log mapper, that still seems worth a warning to me. Alternatively we can just let ranges extend below zero, but refuse to map negative values like MPL does.

bryevdv: You're right, it's ony `symlog` that does, I was mistaken. Sorry for the noise.

bryevdv:
So after the above discussion I'm +0 on the warning, but am happy to merge the PR as-is. @clairetang6 the PR is still marked `WIP`, is there any other work or is this `ready`?

clairetang6:
Hi @bryevdv, I'm planning to add some tests and maybe some text into documentation/docstrings that says that the range padding calculation will be linear or log based on whether the axis was linear or log. I'm traveling right now, so it still might be a few days before I can finish this up.

bryevdv:
@clairetang6 great, thanks for the update!

bryevdv:
I've moved all the related issues to the `0.12.4` milestone

clairetang6:
Some screenshots and thoughts for reference for deciding on the data_range1d's `default_span` for log ranges. 

When only one point is plotted, the range determined by the data has a span set by the parameter `default_span`, which is set to 2 by default. When plotting is linear, this means that the `start` is the data value - 1 and the `end` is the data value + 1. 

Added in this PR, when plotting is log, this `default_span` of 2 means that the `start` is one order of magnitude (set to base 10, not determined by the base of the ticker/formatter) smaller than the data value and the `end` is one order of magnitude bigger. However, with larger values that are powers of 10 or close, you end up with rounding errors that make for slightly uglier defaults. 
image

Adding some small fraction, here 0.001, to the `default_span` fixes this. 
image

This is extremely minor as a detail, but is how I've made `default_span` work in the log case.

bryevdv:
@clairetang6 thanks this looks great, will merge when green!

PR #5480, opened 2016-11-21
Making gridplot work properly when complex layouts (Row/Column) added…

### LINKED ISSUE #5479: Merged toolbar is not created properly when row/column layouts added to gridplot
Sometimes it is required to add complex structures to gridplot (i.e. column containing 2 figures with one big figure nearby).

Currently it is not working properly in terms of toolbar creation, because gridplot searches for immediate children Figures only.

cloud-rocket:
- Fixes #5479


bryevdv:
@cloud-rocket in light of #5503 can you pare this PR down to remove the "temporary" css class parts?

cloud-rocket:
All you need is to pull 8f10e60 commit only (not sure how to discard the previous commit from pull)

bokeh/layouts.py
```diff
@@ -368,9 +368,11 @@ def gridplot(*args, **kwargs):
...
         row_children = []
         for item in row:
-            if merge_tools and isinstance(item, Plot):
-                row_tools = row_tools + item.toolbar.tools
-                item.toolbar_location = None
+            if merge_tools:
+                lst = list(item.select({'type': Plot}))
+                for i in range(0, len(lst)):
+                    row_tools = row_tools + lst[i].toolbar.tools
+                    lst[i].toolbar_location = None
```
birdsarah:
```
plots = item.select({'type': Plot})
for plot in plots:
   row_tools = row_tools + plot.toolbar.tools
   plot.toolbar_location = None
```

examples/custom/class/README.md
```diff
@@ -0,0 +1,8 @@
+Sometimes is it convenient to add custom HTML classes to tags generated by Bokeh (i.e. for creating a custom CSS for a single element only).
+
+Until proper support will be added (#5471) it is possible to properly extend Bokeh element View by adding a custom class.
+
+Current example demonstrates how to added custom HTML classes to derived Row/Column layouts.
+
```
birdsarah: "class" seems like a very undescriptive name for the README.md

The README's should include a how to run and probably shouldn't include references to future fixes or personal touches like the signature here.

examples/custom/class/models.py
```diff
@@ -0,0 +1,9 @@
+from bokeh.models import LayoutDOM, Column, Row
+
+class CustomColumn(Column):
+
+    __implementation__ = "custom_col.coffee"
```
birdsarah: This doenst' need to be broken out into so many files

birdsarah:
Please also add some tests. If I recall correclty then I didn't write enough tests for this method anyway. So that's my bad for setting a bad precedent, but you don't need to follow!

bryevdv:
@birdsarah all of that that you commented on should be deleted anyway, the only part of this PR to keep is the first commit

bryevdv:
superseded by #5534

bryevdv:
Thanks for doing the initial footwork @cloud-rocket I have closed the issue in a separate PR.

Issue #5471, opened 2016-11-17
[ENH] Add custom classes to elements


Deeech:
It would be great if you could add custom classes for plots, widgets, layouts and etc.
For example:

```
buttomDiscard = Button(label="Discard", button_type="success", className="customButton")
confirmDialog = widgetbox([buttomDiscard], className="customBox")
```

Volodymyrk:
Can someone familiar with the code base comment on the complexity of this feature? Can it be achieved during a weekend hack? Would core team accept PR for it?

bryevdv:
I don't think it would be that difficult, and it seems like a fairly reasonable escape valve for users to implement more precise styling. I think a PR would be received favorably, tho I'd like to hear thoughts from @birdsarah and @mattpap.

Regarding implementation, it would probably be a single `String` property on `Widget` (or perhaps a different class) and then some small code on the BokehJS side to apply that property value as a class. I guess the question is where exactly the class should be applied.

Volodymyrk:
My first reaction would be try to implement `_repr_html_() `method in a derived class. Thinking about the feature - I'd probably want to set only css class name from python (like @Deeech suggested, and unlike my[ question on SO ](http://stackoverflow.com/questions/40666632/bokeh-widget-custom-css?noredirect=1#comment68585209_40666632)). So the actual style code will be in .css file that I'd include in /template/index.html
Widgets (for now) are not practically usable inside notebooks. But if/when they will be - providing the actual style code would be more practical (like it is possible with[ IPython widgets](http://stackoverflow.com/questions/37013489/how-to-alight-and-place-ipywidgets))

bryevdv:
@Volodymyrk I'm not sure what part `_repr_html` would play in this. That is a feature that is completely specific to Jupyter Notebooks, so I don't think it could be applicable to a general solution (Bokeh must always work the same outside notebooks, too).

cloud-rocket:
Because this feature was urgently needed, I added an example of how to accomplish it by properly extending Bokeh - #5481

bryevdv:
@cloud-rocket @Volodymyrk @Deeech please check out #5503 if you can.

PR #5481, opened 2016-11-21
Adding custom HTML classes to Bokeh layouts

cloud-rocket:
This example addresses the issue until proper support is added to Bokeh.

No internal Bokeh changes are made, only demonstration example 

- [1 ] issues: #5471


bryevdv:
I'm going to work on a PR to add a `css_classes` property to all `LayoutDOM` models for `0.12.4`. Accordingly, I am going to close this, as a temporary example would only have to be removed later. Thanks for the PR and interest, however!

PR #5482, opened 2016-11-21
Replace lists with slices in qualitative brewer palettes

### LINKED ISSUE #5376: Change palette references for the brewer qualitative palettes to be slices
As discussed in pr #5359 and issue #5317, it would be better for redundant lists in `palettes.py` to be slices instead of repeating the list.

StevenCHowell:
This address #5376, but only the python code https://github.com/bokeh/bokeh/blob/master/bokeh/palettes.py.  I am less familiar with JavaScript so I did not edit the coffee code, https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/api/palettes.ts.  I expect that could be changed as well to provide a cleaner implementation.

- [ ] issues: fixes #5376 
- [ ] tests passed


bokeh/palettes.py
```diff
@@ -768,189 +768,189 @@ def RdYlGn11(self):
...
+        return self.Pastel1_9[:8]
     @property
     def Pastel1_9(self):
         return ['#fbb4ae', '#b3cde3', '#ccebc5', '#decbe4', '#fed9a6', '#ffffcc', '#e5d8bd', '#fddaec', '#f2f2f2']
 
     # http://colorbrewer2.org/?type=qualitative&scheme=Pastel2&n=8
     @property
     def Pastel2_3(self):
-        return ['#b3e2cd', '#fdcdac', '#cbd5e8']
+        return self.Pastel2_8[:7]
```
bryevdv: @StevenCHowell seems like copy-paste error on these `Pastel2` colormaps?

bryevdv:
@StevenCHowell one issue about `Pastel2` slices (looks like a copy pasta) otherwise LGTM

StevenCHowell:
@bryevdv just pushed that fix

StevenCHowell:
@bryevdv that push actually had some commits to `palettes.ts` as well (testing out what I was asking about in #5376); wait to merge it until I sort out the JavaScript.  @mattpap did you get to look at those proposed changes?  If that is valid JavaScript I will change the other redundant lists as well.  With the dev installation of bokeh, will using that palette test the validity of `palettes.ts` (or what is the best way to test the JavaScript/CoffeScript code)?

StevenCHowell:
Trying to figure out how to slice a list in the bokehjs file, `palettes.ts`.  I just pushed another attempt.  Is that file CoffeeScript or JavaScript?

bryevdv:
@StevenCHowell sorry, I should have mentioned, it is TypeScript

bryevdv:
Actually, for simplicity, why don't we just proceed with a Python-only PR first

StevenCHowell:
I tried removing the commits but did not have time to sort through what git was doing to the file.  In the end, I manually removed the changes to the TypeScript version.  The Python should be good to merge.

It should be simple to update the TypeScript, just need to identify the correct syntax, but not a significant difference.

bryevdv:
Thanks @StevenCHowell !

bryevdv:
Well, I'm not sure yet why selenium tests are failing, it's not anything to do with this PR

bryevdv:
passing now after five restarts, hopefully just a transient upstream issue

Issue #2326, opened 2015-05-21
Toggle selection of objects with shift+click


jakirkham:
Taking this example ( http://nbviewer.ipython.org/github/jakirkham/bokeh_examples/blob/master/Bokeh_Example_12.ipynb ), a user can shift click to select several objects. Currently, shift click on already selected object does nothing. It would be nice if this resulted in deselection of the currently selected glyph.

bryevdv:
This seems like a pretty reasonable feature, and perhaps not hard to implement. Would you be interested in working up a PR @jakirkham ?

jakirkham:
Could you please give me some pointers on where this would go? I'm a little unsure of how to start.

jakirkham:
My best guess is this [code]( https://github.com/bokeh/bokeh/blob/e701b6df5d29e48b22b3c7dde0123e632548108e/bokehjs/src/coffee/core/selector.coffee#L10-L18 ) would have to change somehow. Does that seem right?

jakirkham:
I think PR ( https://github.com/bokeh/bokeh/pull/5484 ) should fix this.

Unfortunately I had trouble with the development build so was unable to test myself. 😞 Would appreciate advice on that if you are able.

jakirkham:
As noted in the aforementioned PR, I don't think that change is working from a testing point of view and apparently has larger ramifications that I do not necessarily intend.

It would be nice if someone else could try to address this issue as I still would like to see it solved. However, it is becoming clear to me that it is not an issue that I can easily solve myself.

jakirkham:
I've given this some more thought and wanted to run an idea by you and see what you thought. Is there or could there be a way to get just the item(s) that was selected in the last UI event? Simplest case being this shift+click case. It seems like with that information alone one could probably implement something in a callback that would give the desired behavior.

bryevdv:
well, offhand an idea would be that selection tools could save the old state of selection indices before updating them, and then stuff the old value into the not-very-used-or-documented `cb_data` parameter to customJS callbacks. They don't currently do this, but could be made to fairly straightforwardly I imagine.

jakirkham:
Indeed currently I have a way of doing this now. However one of the problems I'm running into is it appears shift+clicking a selected object does not result in the callback being run. Instead it appears to be suppressed. Would it be reasonable to make a change to the selector logic in Bokeh to pick up these events?

jakirkham:
[...]

PR #5484, opened 2016-11-23
WIP, ENH: Deselect previously selected objects on second selection

jakirkham:
Fixes https://github.com/bokeh/bokeh/issues/2326

This should allow things like `Shift+Click`-ing on an object twice to add and then remove it from the selected group. Of course, a third click should reselect them. Currently clicking on any object that is already selected is a no-op.

cc @bryevdv

### DIFF
--- a/bokehjs/src/coffee/core/selector.coffee
+++ b/bokehjs/src/coffee/core/selector.coffee
@@ -11,10 +11,12 @@ export class Selector extends HasProps
     @setv('timestamp', new Date(), {silent: silent})
     @setv('final', final, {silent: silent})
     if append
-      indices['0d'].indices =  _.union(@indices['0d'].indices, indices['0d'].indices)
+      _xor = (arr_1, arr_2) -> _.difference(_.union(arr_1, arr_2), _.intersection(arr_1, arr_2))
+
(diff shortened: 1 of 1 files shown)


jakirkham:
Tried my best to convert this to JS and try it out in Bokeh locally and I don't think this does it. 😞

bryevdv:
@jakirkham how exactly did you try? did you set the env var `BOKEH_RESOURCES=inline` to use local BokehJS instead of release from CDN?

I also think this might need to be a policy somehow... in the case of tap/click this is pretty reasonable to be the only behaviour. But imagine wanting make a selection that is the union of two lasso regions that happened to be adjacent. What if I *don't* want any accidentally "double selected " points to to be unselected? Then it becomes a tedious chore to try an not accidentally points in both lasso regions.

jakirkham:
I patched BokehJS that was installed with Bokeh in site-packages. Would that not work?

Yep, that makes sense. I did not know the ramifications of this sort of change. I'd rather not change those behaviors. So am closing this out.

bryevdv:
@jakirkham just to answer your question: probably not, By default Bokeh uses BokehJS releases from CDN. If you want the local BokehJS (in site packages) to be used you have to set that env var.

PR #5486, opened 2016-11-26
Pass keyword arguments through Serve

mrocklin:
This helps with the manual creation of a Bokeh Server with a predefined Tornado IOLoop.  We pass a few keyword arguments through `invoke` and return the server object.  This enables the following example:

```python
In [1]: from tornado.ioloop import IOLoop

In [2]: loop = IOLoop()

In [3]: import argparse
   ...: argv = ['bokeh', 'examples/app/crossfilter']
   ...: parser = argparse.ArgumentParser(prog=argv[0])
   ...: from bokeh.command.subcommands.serve import Serve
   ...: serve = Serve(parser)
   ...: args = parser.parse_args(argv[1:])
   ...: server = serve.invoke(args, start_loop=False)
   ...: 
2016-11-26 13:25:39,015 Starting Bokeh server version 0.12.3dev4-283-g0db3db9-dirty
2016-11-26 13:25:39,017 Starting Bokeh server on port 5006 with applications at paths ['/crossfilter']
2016-11-26 13:25:39,018 Starting Bokeh server with process id: 15055

In [4]: server.io_loop is loop
Out[4]: True

In [5]: server.io_loop._running
Out[5]: False

In [6]: server.io_loop.start()  # everything works fine
```

### DIFF
--- a/bokeh/command/subcommand.py
+++ b/bokeh/command/subcommand.py
@@ -44,7 +44,7 @@ class Subcommand(with_metaclass(ABCMeta)):
             self.parser.add_argument(*flags, **arg[1])
 
     @abstractmethod
-    def invoke(self, args):
+    def invoke(self, args, **kwargs):
         ''' Takes over main program flow to perform the subcommand.
 
         Args:
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -324,7 +324,7 @@ import argparse
 
 from bokeh.application import Application
 from bokeh.resources import DEFAULT_SERVER_PORT
-from bokeh.server.server import Server
+from bokeh.server.server import Server, _tornado_kwargs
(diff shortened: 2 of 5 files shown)


mrocklin:
With the help of @seanlaw I've found that I can achieve my original goal with the following code:

```python
from bokeh.command.util import build_single_handler_applications
apps = build_single_handler_applications(['examples/app/crossfilter/'], {})
loop = IOLoop()
server = Server(apps, io_loop=loop)
server.start(start_loop=False)
assert server.io_loop is loop
assert server.io_loop._running
server.io_loop.start()
```

mrocklin:
So from my perspective this can be closed.  I'll leave it open in case maintainers see value in it.

bryevdv:
I think it's better to let the handlers continue to be the sole thing responsible for using or supplying or ignoring things like `argv`

PR #5498, opened 2016-11-30
Add Model.js_on_change method

### LINKED ISSUE #5435: [ENH] Add JS callback for streaming data
This feature request follows [preliminary discussion in groups](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/5kar6-glxf4)

It'd be great to have a per-chart JS callback durng data stream update (new data is about to be displayed), so custom frontend elements (Glyphs) will be updated accordingly. 

Thanks
Meir Tseitlin

bryevdv:
- [x] issues: fixes #5435
- [x] tests added / passed

Allows to add a CustomJS callback to any backbone event. As a convenience, if the event name is the same as one of the model properties, the event `"change:prop_name"` is generated instead.


bryevdv:
ping @mattpap @birdsarah just want to get a quick look before I commit to writing the tests and docs for this. I opted for a new method `.js_on_change` instead of overloading the existing `.on_change`. I think it is better for users to be more explicit in this way and it is definitely cleaner to not pollute the python-centric `CallbackManager`. Additionally, to support events on streaming and patching, etc., the interface actually accepts any arbitrary Backbone event, not just property names (tho property names are converted to the appropriate `"change"` event as a convenience)

mattpap:
Two things shouldn't be forgotten is supporting flexx (on Python-side) and (native) functions for TS API.

bokehjs/src/coffee/model.coffee
```diff
@@ -6,10 +6,17 @@ export class Model extends HasProps
...
+    tags:         [ p.Array, [] ]
+    name:         [ p.String    ]
+    js_callbacks: [ p.Any,   {} ]
   }
 
+  initialize: (options) ->
+    super(options)
+    for attr, callbacks of @js_callbacks
+      for cb in callbacks
+        @listenTo(@, "#{attr}", () -> cb.execute(@))
```
mattpap: String interpolation is unnecessary here.

bryevdv: yup thanks, it was changed from something I had earlier

bryevdv:
> Two things shouldn't be forgotten is supporting flexx (on Python-side) and (native) functions for TS API.

The flexx comes along for the ride with `CustomJS`. I'll probably need help for a TS API

bryevdv:
@mattpap @birdsarah this is ready for review if you have the chance

@mattpap regarding adding TS API in this case I think it's better to just let people add event handlers in the standard ways, e.g. `listenTo`, which is already possible, rather than adding a duplicate of the python API in JS just because.

examples/app/ohlc/main.py
```diff
@@ -95,5 +93,7 @@ def update(t):
 
     source.stream(new_data, 300)
 
+curdoc().add_root(column(row(mean, stddev, mavg), gridplot([[p], [p2]], toolbar_location="left", plot_width=1000)))
+
```
bryevdv: I think it's better to demonstrate `curdoc().add_root` as late as possible, as a best practice.  The resolution of model reference happens at `add_root` not at serialization, so confusing situations can happen if `add_root` happens before a model has changes that affect its references.

sphinx/source/docs/user_guide/examples/interaction_callbacks_for_widgets.py
```diff
@@ -9,7 +9,7 @@
 
 source = ColumnDataSource(data=dict(x=x, y=y))
 
-plot = Figure(plot_width=400, plot_height=400)
+plot = figure(plot_width=400, plot_height=400)
```
bryevdv: noticed a couple of lingering uses of `Figure`, fixed these up in this PR

bokehjs/test/model.coffee
```diff
@@ -0,0 +1,67 @@
...
+      spy.reset()
+      expect(spy.called).to.be.false
+      m.bar = "test"
+      expect(spy.callCount).to.be.equal 1
+
+      # check the correct number of calls for m.baz change
+      spy.reset()
+      expect(spy.called).to.be.false
+      m.baz = 10
+      expect(spy.callCount).to.be.equal 0
```
bryevdv: If there's a way to make this test better I am all ears. `spy` seems to wrap the prototype or something, precluding per-instance resolution of methods. So this test just counts the total calls and makes sure that they match the expected counts for each attr.

bryevdv:
I'll merge this tomorrow absent objection

jlstevens:
This looks great to me and I am looking forward to seeing this merged!

bryevdv:
@mattpap I didn't get around to merging this weekend, and I wanted to give you more time to comment too. I will merge this afternoon unless you voice any objection.

bokeh/model.py
```diff
@@ -75,10 +83,60 @@ def get_class(cls, view_model_name):
...
+    Bokeh models:
+
+    .. code:: python
+
+        >>> r = plot.circle([1,2,3], [4,5,6])
+        >>> r.tags = ["foo", 10]
+        >>> plot.select(tags=['foo', 10])
+        [GlyphRenderer(id='1de4c3df-a83d-480a-899b-fb263d3d5dd9', ...)]
+
+    Or simply a convenient way to attach any necessary metadata to a model
```
birdsarah: Thanks for doing this @bryevdv

PR #5503, opened 2016-12-02
Add support for specifying CSS classes on all LayoutDOM

### LINKED ISSUE #5471: [ENH] Add custom classes to elements

bryevdv:
- [x] issues: fixes #5471
- [x] tests added / passed

Adds .css_classes property to LayoutDOM, which accepts a sequence of strings. These are added as-is to the LayoutDOM element.

### DIFF
--- a/bokeh/models/layouts.py
+++ b/bokeh/models/layouts.py
@@ -14,7 +14,7 @@ from ..core.validation.warnings import (
     BOTH_CHILD_AND_ROOT,
 )
 from ..core.enums import SizingMode
-from ..core.properties import abstract, Bool, Enum, Int, Instance, List
+from ..core.properties import abstract, Bool, Enum, Int, Instance, List, Seq, String
 from ..embed import notebook_div
 from ..model import Model
 from ..util.deprecation import deprecated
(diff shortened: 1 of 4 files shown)


bryevdv:
ping @mattpap @birdsarah 

It would be nice to add an example someplace, if anyone has any simple ideas please let me know.

bryevdv:
Also this could probably be made to work:
```
thing.css_classes = "foo"
```
worthwhile?

birdsarah:
Tried this out. 

It added to a Column fine, a Slider fine, but I then added it to Paragraph and it didn't attach.

Probably going to need to audit all widgets.

bryevdv:
@birdsarah it's working fine for me with `Paragraph`, can you double-check? 
```
from bokeh.plotting import output_file, show
from bokeh.models import Paragraph

p = Paragraph(text="foo")
p.css_classes = ["STUFF"]

output_file("foo.html")
show(p)
```

screen shot 2016-12-02 at 8 26 57 pm

bryevdv:
`Paragraph` uses `$el.html(...)` to render inside the `LayoutDOM` div, were you looking at the inside div? I think putting on the outermost `LayoutDOM` div is best because it is the most consistent thing that could be done, is that not correct?

bryevdv:
OK I am going to merge this when green

PR #5504, opened 2016-12-04
Base64 serialization

philippjfr:
This PR implements some of the suggestions in https://github.com/bokeh/bokeh/issues/2204. In further discussion and initial testing (see https://github.com/bokeh/bokeh/pull/5429) we found that simple base64 encoding offers huge performance gains compared to the current implementation. A more complex binary json format is therefore not necessary at this time.

## Motivation

Currently sending large arrays to the browser can take a very long time, particularly when sending relatively large images. In the case of a 2000x2000 image it can take up 30 seconds to serialize and send the image. Through initial investigation we found that simple base64 encoding of arrays can result in huge performance gains for sending array data via web sockets. This is partially due to the reduced network traffic, but mostly due to less work during serialization, i.e. iterating over the array. Overall this has the potential to bring 10-50x improvements in sending large arrays via the comms in the notebook and will also offer some more limited improvements for bokeh server.

## Design

1. The encoding should happen as a last step during serialization on the Python end and as the first step after deserialization on the JS end, ensuring no other code has to deal with base64 encoded data avoiding any backward compatibility issues.
2. The data should be wrapped in a simple JS object along ith the shape and type information. A dedicated model might be considered in future but is not necessary.
3. The column data source on the fronted will unpack and store the shape information on the _shapes attribute.
4. The encoding behavior should be configurable, e.g. do not encode unless the array is larger than a configurable size.
5. All numeric column data source data should potentially be encoded, i.e. both 1-D and 2-D arrays.

## Caveats

Currently I'm encountering issues updating a data source with typed arrays, hence I'm casting to regular array types for now. Need to figure out how to avoid that in this PR.

Currently this PR will correctly encode and decode any numerical types except for 64-bit integers. Since there is no native 64-bit integer type (the maximum safe integer is 2**53), we cannot directly encode and decode 64-bit integer data. Using the npm ``int64-buffer`` package I managed to decode an ``int64`` ArrayBuffer but the advantages and disadvantages of various approaches should be evaluated in a subsequent PR:

```coffeescript
view = new Int32Array(bytes)
array = []
for i in [0...view.length/2]
  int = new Int64LE(view.buffer.slice(i*8,(i+1)*8))
  array.push(int)
```

## To-Do

- [ ] Add configurable options ``BOKEH_USE_BINARY_ARRAYS`` and ``BOKEH_BINARY_ARRAY_CUTOFF``
- [ ] Handle patch events
- [ ] Handle stream events
- [ ] Ensure current tests pass and add extensive tests
- [ ] Add release document entry (if new feature or API change)
- [ ] Document environment variables


bokeh/util/serialization.py
```diff
@@ -51,6 +53,15 @@ def transform_series(obj):
     vals = obj.values
     return transform_array(vals)
 
+def flattened_and_shape(array):
```
jbednar: Docstring?

jbednar:
Looks great to me!  Very excited about the potential performance improvement for images.

bokeh/util/serialization.py
```diff
@@ -54,7 +55,19 @@ def transform_series(obj):
...
+    if (not settings.use_binary_arrays() or
+        array_samples < settings.binary_array_cutoff() or
+        array.dtype.kind in ('U', 'S', 'O') or
+        array.dtype.name == 'int64'):
+        if np.isnan(array).any() or np.isinf(array).any():
+            transformed = array.astype('object')
+            transformed[np.isnan(array)] = 'NaN'
+            transformed[np.isposinf(array)] = 'Infinity'
+            transformed[np.isneginf(array)] = '-Infinity'
+            return transformed.tolist()
```
jbednar: Also, won't this have major negative implications for performance, with huge array sizes for anything that has nans or infs (as in the lines just above this block the array is converted to an object array)?

bryevdv: This is what it is always doing now, so....

philippjfr: >maybe roll this ugly block up into a transform_to_list function or something and reduce the nesting?

Good plan, the function is misnamed right now anyway.

>Also, won't this have major negative implications for performance, with huge array sizes for anything that has nans or infs

This is the fallback condition, i.e. what it was doing before.

bokeh/util/serialization.py
```diff
@@ -54,31 +54,38 @@ def transform_series(obj):
...
+    either a list or a dictionary containing the base64
+    encoded data along with the shape and dtype of the data.
+    """
     if isinstance(array, np.ma.MaskedArray):
         array = array.filled(np.nan)  # Set masked values to nan
     array_samples = np.product(array.shape)
     if (not settings.use_binary_arrays() or
         array_samples < settings.binary_array_cutoff() or
         array.dtype.kind in ('U', 'S', 'O') or
         array.dtype.name == 'int64'):
```
bryevdv: So another idea might be to wrap all these conditions up into a predicate function as well:
```
if need_list_seralization(array):
    return transform_array_to_list(array)
```
Apart from being a bit more readable, it would also be good to put the function `need_list_seralization` under unit test, so that the conditions for when list serialization happens can be rigorously maintained

bokeh/util/serialization.py
```diff
@@ -55,7 +55,7 @@ def transform_series(obj):
     return transform_array(vals)
 
 def transform_array_to_list(array):
-    if np.isnan(array).any() or np.isinf(array).any():
+    if (array.dtype.kind in ('u', 'i', 'f') and (~np.isfinite(array)).any()):
```
bryevdv: `np.isfinite(array)).all()` ?

philippjfr: I've removed the other condition so this seems right to me, if it's numeric and any value is non-finite replace the non-finite values.

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -396,6 +396,8 @@ export class HasProps extends Backbone.Model
       if (prop.optional || false) and prop.spec.value == null and (name not of @_set_after_defaults)
         continue
       data["_#{name}"] = prop.array(source)
+      if name of source._shapes and source._shapes[name]?
```
philippjfr: ToDo: Make note that this could be made into a free function

mattpap: Those two clauses are redundant.

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -388,6 +388,7 @@ export class HasProps extends Backbone.Model
       @document._notify_change(@, attr, old, new_)
 
   materialize_dataspecs: (source) ->
+    # Note: this should be moved to a function separate from HasProps
```
philippjfr: @bryevdv mentioned that it doesn't seem like it should live on HasProps.

mattpap: It should live on `Glyph`, but since data specs were allowed elsewhere (e.g. in annotations), it would have to be duplicated in `Glyph` and `Renderer`. This way, to avoid duplication, it's either `Model` or `HasProps` (and `Model pretty_much_== `HasProps` in bokehjs). I'm not a big fan of having auxiliary functions accessing internals of classes.

bryevdv:
Another consideration just to keep in mind, wrt to PhantomJS, we need to see if the expanded use of typed arrays has any effect on our image-diff tests.

bryevdv:
@philippjfr just FYI the conflict should be small, the other PR replaced `Uint8ClampedArray` with `Uint8Array`

bryevdv:
GH let's you resolve merge conflicts directly here in the web UI now!  (@philippjfr be sure to pull)

philippjfr:
>GH let's you resolve merge conflicts directly here in the web UI now! 

That's pretty cool, thanks for doing that!

philippjfr:
Just a note, probably for my reference only, but I need to go through all the charts and make sure they're working. I've been using this branch for all my work for the past week week, and noticed that while plotting using the glyph API seems to be working, some charts seemingly break.

bryevdv:
@philippjfr that's strange but thanks for noting, I can try to take a look too

bokehjs/src/coffee/core/has_props.coffee
mattpap: At some point our tooling will support mixin composition pattern, so this is a nice candidate to be moved to a trait.

bryevdv:
@philippjfr we should close and re-open this PR directly on `bokeh/bokeh` the image diff reports will not be available otherwise.

bryevdv:
I have sent you an invite to @bokeh/dev

philippjfr:
>@philippjfr we should close and re-open this PR directly on bokeh/bokeh the image diff reports will not be available otherwise.

Ah, yes we definitely want those. I'll do that shortly.

bokehjs/src/coffee/core/util/deserialize.coffee
```diff
@@ -0,0 +1,45 @@
...
+
+export decode_column_data = (data) ->
+    new_data = {}
+    data_shapes = {}
+    for k, v of data
+      if v instanceof Array
+        arrays = []
+        shapes = []
+        for arr in v
+          if arr? and arr.constructor.name == 'Object'
```
philippjfr: Perfect thanks.

bokehjs/src/coffee/document.coffee
```diff
@@ -650,8 +651,13 @@ export class Document
             throw new Error("Cannot apply patch to #{patched_id} which is not in the document")
           patched_obj = @_all_models[patched_id]
           attr = event_json['attr']
-          value = Document._resolve_refs(event_json['new'], old_references, new_references)
-          patched_obj.setv({ "#{attr}" : value })
+          model_type = event_json['model']['type']
+          if attr == 'data' and model_type == 'ColumnDataSource'
```
mattpap: This pretty much makes `ColumnDataSource` final, though I'm not sure if we can do better with this code anyway. Unless we don't want support this encoding in its subclasses, which may not happen, but at least we have to annotate classes appropriately).

bokehjs/src/coffee/models/glyphs/image.coffee
```diff
@@ -30,22 +30,29 @@ export class ImageView extends GlyphView
...
 
     for i in [0...@_image.length]
-      if @_rows?
-        @_height[i] = @_rows[i]
-        @_width[i] = @_cols[i]
+      shape = []
+      if @_image_shape?
+        shape = @_image_shape[i]
+
+      if shape.length
```
mattpap: I would prefer `if shape.length > 0` as `shape.length` is not a boolean.

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -5,21 +5,27 @@ import * as hittest from "../../core/hittest"
...
-      data:              [ p.Any,      {} ]
-      column_names:      [ p.Array,    [] ]
+      data:         [ p.Any,   {} ]
+      column_names: [ p.Array, [] ]
     }
 
   @internal {
     selection_manager: [ p.Instance, (self) -> new SelectionManager({source: self}) ]
     inspected:         [ p.Any ]
+    _shapes:           [ p.Any, {}]
```
mattpap: I would like to have a comment with exact type of `_shape`. It's a shame that `inspected` doesn't show the type and encourage to follow the pattern.

philippjfr:
Closing and reopening based on bokeh/bokeh branch.

bryevdv:
superseded by #5544

PR #5505, opened 2016-12-05
Fix #5494: make Tornado server library-friendly

### LINKED ISSUE #5494: BokehTornado.stop should not stop the IOLoop
`BokehTornado.stop` currently calls `self._loop.stop()`, which kills any other application running on the same IOLoop.  Loop creation and deletion should maybe happen outside at the CLI level.

```python
+++ b/bokeh/server/tornado.py
@@ -368,8 +368,6 @@ class BokehTornado(TornadoApplication):
         if self._ping_job is not None:
             self._ping_job.stop()
 
-        self._loop.stop()
-
     @property
     def executor(self):
         return self._executor
```

pitrou:

- [x] issues: fixes #5494
- [x] tests added / passed

### DIFF
--- a/bokeh/server/tornado.py
+++ b/bokeh/server/tornado.py
@@ -12,7 +12,6 @@ from concurrent.futures import ProcessPoolExecutor
 import os
 from pprint import pformat
 import signal
-import sys
 
 from tornado import gen
 from tornado.ioloop import IOLoop, PeriodicCallback
@@ -323,10 +322,10 @@ class BokehTornado(TornadoApplication):
         return self._resources[root_url]
 
     def start(self, start_loop=True):
-        ''' Start the Bokeh Server application main loop.
+        ''' Start the Bokeh Server application.
 
         Args:
-            start_loop (boolean): False to not actually start event loop, used in tests
+            start_loop (boolean): whether to start the IO loop
 
         Returns:
             None
--- a/bokeh/server/tornado.py
+++ b/bokeh/server/tornado.py
@@ -6,15 +6,13 @@ from __future__ import absolute_import, print_function
 import logging
 log = logging.getLogger(__name__)
 
-import atexit
 # NOTE: needs PyPI backport on Python 2 (https://pypi.python.org/pypi/futures)
 from concurrent.futures import ProcessPoolExecutor
 import os
 from pprint import pformat
-import signal
 
 from tornado import gen
-from tornado.ioloop import IOLoop, PeriodicCallback
+from tornado.ioloop import PeriodicCallback
 from tornado.web import Application as TornadoApplication
 from tornado.web import HTTPError
 from tornado.web import StaticFileHandler
@@ -148,7 +146,7 @@ class BokehTornado(TornadoApplication):
 
     def __init__(self, applications, prefix, hosts,
                  extra_websocket_origins,
-                 io_loop=None,
+                 io_loop,
                  extra_patterns=None,
                  secret_key=settings.secret_key_bytes(),
                  sign_sessions=settings.sign_sessions(),
@@ -249,7 +247,7 @@ class BokehTornado(TornadoApplication):
         super(BokehTornado, self).__init__(all_patterns)
 
     def initialize(self,
-                 io_loop=None,
+                 io_loop,
                  keep_alive_milliseconds=37000,
                  # how often to check for unused sessions
                  check_unused_sessions_milliseconds=17000,
@@ -259,10 +257,7 @@ class BokehTornado(TornadoApplication):
                  stats_log_frequency_milliseconds=15000,
                  **kw):
 
-        if io_loop is None:
-            io_loop = IOLoop.current()
         self._loop = io_loop
-        self._loop_owned = False
 
         for app_context in self._applications.values():
             app_context._loop = self._loop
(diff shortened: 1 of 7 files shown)


bryevdv:
@pitrou whenever you are ready, just add the label `READY` (or if it won't let you, just ping me)

bokeh/server/tornado.py
```diff
@@ -323,10 +322,10 @@ def resources(self, request):
         return self._resources[root_url]
 
     def start(self, start_loop=True):
-        ''' Start the Bokeh Server application main loop.
+        ''' Start the Bokeh Server application.
 
         Args:
-            start_loop (boolean): False to not actually start event loop, used in tests
+            start_loop (boolean): whether to start the IO loop
```
bryevdv: needs `(default: True)` at the end

pitrou:
It won't let me, so ping.

bryevdv:
This LGTM for the most part. I do find `start(start_loop=False)` slight awkward just from a "muddying the meaning of 'start' standpoint". What about splitting `start` into `init` and `start`?
 
The `init` method would basically do everything that `start(start_loop=False)` does, and then `start` would call `init` but also start the `ioloop`. 

Thoughts?

bryevdv:
One other possible idea: instead of a flag to install pre-defined set of "cleanup hooks", have a generic hook that uses cleanup functions that are passed in to `start`, optionally. Then it is entirely the CLI or user code that is responsible for providing the cleanup behaviour that it wants, and the `Server` itself could be slightly simplified.

pitrou:
If you're looking to clean up the current organization, then I would argue the loop starting and shutdown hooks business should be moved to the `Server` class. I would also keep the `start()` and `stop()` method but move the loop-starting code to a separate `run_until_shutdown()` method (name can be discussed :-)).

pitrou:
> One other possible idea: instead of a flag to install pre-defined set of "cleanup hooks", have a generic hook that uses cleanup functions that are passed in to start, optionally. 

I might misunderstand, but that sounds backwards. The Server knows better than the user which of its internal structures need cleaning up at shutdown.

bryevdv:
> If you're looking to clean up the current organization, then I would argue the loop starting and shutdown hooks business should be moved to the Server class. I would also keep the start() and stop() method but move the loop-starting code to a separate run_until_shutdown() method (name can be discussed :-)).

This all seems pretty reasonable. I don't think there are very many people relying on using the either `Server` or `BokehTornado` directly, programmatically *yet*. So if we plan to support these use cases (which is the point of these improvements really) then now is the time to make as many improvements as we can. We just have to make sure to telegraph any potentially breaking changes well in the migration notes. 

> I might misunderstand, but that sounds backwards. The Server knows better than the user which of its internal structures need cleaning up at shutdown.

This is probably true, looking closer at the actual hooks they do seem to be using fairly internal things.

sphinx/source/docs/user_guide/server.rst
```diff
@@ -642,6 +642,32 @@ and the server, there is network traffic between the python client and the
...
+       *server_kwargs* can contain additional Bokeh Server arguments,
+       such as the ``port`` number to listen on.
+       """
+       from bokeh.server.server import Server
+       server = Server(bokeh_applications, io_loop=loop,
+                       **server_kwargs)
+       server.start()
+
+When you ever want to stop the embedded Bokeh Server, call the ``stop()``
+method on the given ``Server`` instance.
```
bryevdv: My only comment: is it worth explicitly mentioning here that `server.start` does *not* start the passed-in `IOLoop`? It might just be my inexperience with Tornado, but I would expect a method `start` to "start" things, whereas at least it seems to me it's really doing more like a setup step. Or maybe this is a recognized pattern, I'm just not familiar with?

bryevdv:
@pitrou this looks great, much cleaner. I only have one comment/question about a method name.

sphinx/source/docs/user_guide/server.rst
pitrou: The method does start the Bokeh server on the given I/O loop. You find this pattern in e.g. Tornado's `TCPServer`: http://www.tornadoweb.org/en/stable/tcpserver.html#tornado.tcpserver.TCPServer

pitrou:
I've tweaked the docs to answer your concerns more clearly.

bokeh/server/tornado.py
```diff
@@ -342,24 +327,18 @@ def start(self, start_loop=True):
...
-                self._loop.start()
-            except KeyboardInterrupt:
-                print("\nInterrupted, shutting down")
-
-    def stop(self):
+    def stop(self, wait=True):
         ''' Stop the Bokeh Server application.
 
+        Args:
+            fast (boolean): whether to wait for orderly cleanup (default: True)
```
mrocklin: fast/wait mismatch between args and docstring

sphinx/source/docs/user_guide/server.rst
```diff
@@ -642,6 +642,36 @@ and the server, there is network traffic between the python client and the
...
+
+       *server_kwargs* can contain additional Bokeh Server arguments,
+       such as the ``port`` number to listen on.
+       """
+       from bokeh.server.server import Server
+       server = Server(bokeh_applications, io_loop=loop,
+                       **server_kwargs)
+       # This returns immediately after having started all timers
+       # and services.
+       server.start()
```
mrocklin: Perhaps consider just providing the body of the function here?  The function definition may not provide much here.

bryevdv: Agreed. I would suggest
```
.. code-block:: python

    server = Server(
        bokeh_applications,  # list of Bokeh applications
        io_loop=loop,        # Tornado IOLoop 
        **server_kwargs      # port, num_procs, etc. 
    )

    # start timers and services and immediately return 
    server.start()

```

bryevdv:
LGTM will merge when green, thanks @pitrou

PR #5508, opened 2016-12-06
Use native methods instead of underscore

### LINKED ISSUE #5507: Replace underscore's functions with native methods were possible
This makes code more readable and will allow us to remove big chunks of underscore later on.

mattpap:
Quick clean up. A lot more could be done, but those changes may be a lot more subtle than these. Note that compatibility isn't an issue, because all browsers and IE 9+ support these methods.

fixes #5507

### DIFF
--- a/bokehjs/src/coffee/api/charts.coffee
+++ b/bokehjs/src/coffee/api/charts.coffee
@@ -151,7 +151,7 @@ export bar = (data, opts={}) ->
     for v, i in row
       columns[i].push(v)
 
-  labels = _.map(columns[0], (v) -> v.toString())
+  labels = columns[0].map((v) -> v.toString())
   columns = columns.slice(1)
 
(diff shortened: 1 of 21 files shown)


mattpap:
JavaScript at its finest:
```
> [undefined].map((x) => NaN)
[ NaN ]
> (new Array(1)).map((x) => NaN)
[  ]
```
This seemingly empty array is `[undefined]`, just `undefined` is printed as empty space. So, `new Array(1)` creates `[undefined]` as I did in the first example. Why doesn't it map the same? Because although seemingly we get `[undefined]`, this time indices of this array are "untouched" (we didn't explicitly set any values of any indices) and such are skipped by `map()`.

bryevdv:
OK the changes look good and I testes ~20 examples, including all the GeoJSON ones. LGTM

PR #5511, opened 2016-12-07
Fix #5493: avoid calling sys.exit from library code

### LINKED ISSUE #5493: Avoid calling sys.exit in server code
Currently the server calls sys.exit if it can't find a good port.  If Bokeh intends to serve non-command-line applications (which I think it should) then it should probably raise here instead and handle the exception in the command line tool.

    bokeh/server/server.py::Server.__init__

pitrou:
- [x] issues: fixes #5493
- [x] tests added / passed


bryevdv:
@pitrou this looks great, just ping me when `ready`

pitrou:
Unrelated failure here: https://travis-ci.org/bokeh/bokeh/jobs/181989004#L2774

Otherwise, this should be ready.

bryevdv:
Hrm, I am at a loss to explain how that error could be in this PR but not in the branch the PR was based off of. I have restarted the build though I don't really see offhand how that would help. If it persists, I will check out the PR locally and investigate.

bryevdv:
It's now failing in your other PR as well, which was passing before, after your last trivial docs change. There must be a dependency change on Travis, or something like that.

bryevdv:
@pitrou this PR has conflicts after merging #5494

pitrou:
Ok, conflicts fixed.

PR #5512, opened 2016-12-07
Fix #5492: expose and print actual port number

### LINKED ISSUE #5492: Support --port 0 for random port
Sometimes we don't know a port that will be open apriori and we want a service to take a random open port.  The convention here is to specify port 0.  Bokeh currently reports the port as 0 when doing this:

```
mrocklin@workstation:~/workspace/bokeh$ bokeh serve examples/app/crossfilter/ --port 0
2016-11-29 10:27:45,282 Starting Bokeh server version 0.11.0dev2-4074-g9c73877-dirty
2016-11-29 10:27:45,285 Starting Bokeh server on port 0 with applications at paths ['/crossfilter']
2016-11-29 10:27:45,285 Starting Bokeh server with process id: 9766
```

It would be nice for it instead to select a random open port and report that port instead.  Tornado servers (and Python sockets generally) will follow this convention if you just call `TCPServer.listen(0)` or whatever.

pitrou:
This makes the Server more usable when port=0 is passed.
Based on PR #5511.

- [x] issues: fixes #xxxx
- [x] tests added / passed


pitrou:
@bryevdv , this PR is ready.

bryevdv:
@pitrou this looks great, one last thing, until we switch to using `Options` somehow, the server args have to be documented "by hand", I think a note about `--port=0` would be appropriate here:

https://github.com/pitrou/bokeh/blob/aa46bc9006ee35a6bc82e9ad0ce0b22818d41e8a/bokeh/command/subcommands/serve.py#L89

bryevdv:
@pitrou great! I am going to cancel the build for this last commit and merge, since there is nothing that changed that is under test. Thanks!

PR #5515, opened 2016-12-07
Reduce bootstrap's size

### LINKED ISSUE #5514: Further slim down boostrap
There is still some bootstrap's code in `src/vendor` that shouldn't be there.

mattpap:
All that's left is what we use.

fixes #5514


bryevdv:
LGTM, report looks good too

PR #5516, opened 2016-12-07
Make location/anchor naming consistent (and add aliases)

### LINKED ISSUE #5292: Inconsistent Legend Location naming
When designating the location of a legend, the options are:
- "top_left"
- "top_center"
- "top_right" (the default)
- "right_center"
- "bottom_right"
- "bottom_center"
- "bottom_left"
- "left_center"
- "center"
  Of the 8 that options other than "center", why is the order for the orientation for 6 of them vertical followed by the horizontal alignment and 2 of them horizontal followed by the vertical? This inconsistency seems like it would lead to confusion.

For what it is worth, I noticed [Matplotlib](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.legend) uses '<vertical> <horizontal>'  (probably just following MATLAB). 

I see 3 options moving forward:
1. do nothing, force people to remember to use "<vertical>_<horizontal>" for top and bottom but "<horizontal> <vertical>" for center.
2. deprecate "left_center" and "right_center", changing them to "center_right" and "center_left". 
3. allow for both combinations for all 8 options, with a little more parsing i.e., add:
- "left_top"
- "center_top"
- "right_top" (the default)
- "center_right"
- "right_bottom"
- "center_bottom"
- "left_bottom"
- "center_left"

Thoughts?

mattpap:
fixes #5292

### DIFF
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import
 from six import string_types
 
 from .. import colors, icons, palettes
+from ..util.deprecation import deprecated
 
 class Enumeration(object):
     ''' Represent an enumerated collection of values.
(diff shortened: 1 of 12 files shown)


bryevdv:
I'm OK with adding a full set of aliases, but I guess at the same time my preference would be a particular convention to endorse (i.e., the only way we document or demonstrate)

mattpap:
@bryevdv, so instead of adding aliases I just deprecated old values. No point in having aliases everywhere.

bryevdv:
@mattpap great, I agree. I've added a `MIGRATION` tag to the issue to note the changes

PR #5517, opened 2016-12-07
small fixups to make sphinx 1.5 work

bryevdv:
Sphinx 1.5 seems to more aggresively try to automatically create ref
targets. Two issues resulted in duplicate targets causing a build error

### DIFF
--- a/bokeh/charts/data_source.py
+++ b/bokeh/charts/data_source.py
@@ -294,7 +294,7 @@ class ChartDataSource(object):
             dims (List(Str), optional): list of valid dimensions for the chart.
             required_dims (List(List(Str)), optional): list of list of valid dimensional
                 selections for the chart.
-            selections (Dict(dimension, List(Column)), optional): mapping between a
+            selections (Dict(String, List(Column)), optional): mapping between a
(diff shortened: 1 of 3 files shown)


bryevdv:
@bokeh/dev when this passes I will merge it. I will then merge some current small PRs that are only failing because of this. Larger/longer running PRs should merge this change into their branches.

bryevdv:
another, second, different dependency change update on Travis is also now causing all BokehJS builds to fail

bryevdv:
ping @mattpap any ideas? it seems to be a JS dependency

bryevdv:
I can reproduce locally by blowing away `node_modules`. Manually running:
```
npm install source-map-support
``` fixes

bryevdv:
Adding it to `packages.json` fixes the immediate problem so I will do that in this hotfix PR

mattpap:
@bryevdv, it doesn't hurt to add it, though I would start from bumping typescript's version to `2.1.4`.

bryevdv:
I'll leave that to you in another PR :) I need to get this merged so ~5 other PRs can go in.

mattpap:
Sure.

PR #5525, opened 2016-12-09
Replace Icon (font-awesome) with an example

### LINKED ISSUE #5513: Consider externalizing font-awesome's icons (or removing altogether)
Currently `bokeh-widgets.min.css` is at 524 kB. Externalizing icons reduces this to 220 kB and removing font-awesome entirely reduces size to 181 kB.

mattpap:
Based on issue #5513. I created an example instead, because we don't have yet any infrastructure for maintaining extensions. However, it will be trivial to create one when we are ready. This also adds support for resolving dependencies with npm. Just note this is not fully though through, so many cases may not work, though resolving external less files (font-awesome) works perfectly. 

fixes #5513


mattpap:
This and recent cleanup of bootstrap reduced `bokeh-widgets.min.css`' size to 146 kB (from 524 kB).

mattpap:
Note also that one can still use icons with button widget. Well, sort of, because you have to implement your self a concrete icon class. This works, because we use abstract base classes, so even we don't have any particular icon implementation, one can still add one and use it with existing button widget implementation.

bryevdv:
have currently been using `MIGRATION` on issues. I think it might be better used on PRs but that should be an intentional decision and release issues tooling should be considered.

mattpap:
@bryevdv, I was pretty sure that we used it on PRs already. I think I always set migration label this way.

examples/custom/font-awesome/named_icon.py
```diff
@@ -67,17 +65,4 @@
...
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
-
-    <ul class="fa-ul" style="list-style: none !important;">
-'''
-
-for icon_name in __icons__:
-    __doc__ += '        <li><i class="fa-li fa fa-%s" style="list-style-type: none !important;"></i>%s</li>\n' % (icon_name, icon_name)
-
-__doc__ += '    </ul>'
+])
```
bryevdv: Can we leave a `bokeh/icons.py` for some time that just has a `warnings.warn` that documents when the module was removed, and points people at the example?

sphinx/source/docs/reference/icons.rst
```diff
@@ -1,7 +0,0 @@
-.. _bokeh.icons:
-
-bokeh.icons
-===========
-
-.. automodule:: bokeh.icons
-  :members:
```
bryevdv: If we leave a `bokeh/icons.py` with a warning, it can also have a module docstring that also discusses the removal, and points people at the example, in which case we should leave this sphinx doc for now. 

Suggest:

```
"""

.. warning::
    FontAwesome icons and the associated ``bokeh.icons`` modules were removed in 
    the Bokeh 0.12.4 release. Please see the :bokeh-tree:`examples/custom/font-awesome`
    directory for a demonstration of how to use these icons with a custom extension.

"""
import warnings
warnings.warn("FontAwesome icons and the associated bokeh.icons modules were removed in the Bokeh 0.12.4 release. Please see the 'examples/custom/font-awesome' directory for a demonstration of how to use these icons with a custom extension.")
del warnings
```

examples/custom/font-awesome/named_icon.py
bryevdv: See comment below for suggested content for a temp `bokeh.icons`

bryevdv:
@mattpap I left two small comments about something I think would be more user-friendly

bryevdv:
Flake8
```
E         bokeh/icons.py:11:161: E501 line too long (239 > 160 characters)
```

bryevdv:
LGTM will merge when green

PR #5527, opened 2016-12-09
fixups for some versions if IE11

### LINKED ISSUE #5526: Some versions of IE 11 do not support Unit8ClampedArray
Additionally, the buffer returned for canvas image data needs to be fixed up to add a `set` method. See:

http://stackoverflow.com/questions/22062313/imagedata-set-in-internetexplorer

See also #5522 for some comments originally made in that issue.

bryevdv:
issues: fixes #5526

This is hard for me to test on different windows VMs from dev. I tested that `image.py` `image_url.py` and `color_data_map.py` work on Win7 + IE 11. I'd like to merge this and make a dev build, which will be much easier to test. 

ping @bokeh/dev

### DIFF
--- a/bokehjs/src/coffee/models/canvas/canvas.coffee
+++ b/bokehjs/src/coffee/models/canvas/canvas.coffee
@@ -32,6 +32,13 @@ export class CanvasView extends BokehView
     fixup_measure_text(@ctx)
     fixup_ellipse(@ctx)
 
+    # fixes up a problem with some versions of IE11
+    # ref: http://stackoverflow.com/questions/22062313/imagedata-set-in-internetexplorer
+    if window.CanvasPixelArray?
+      CanvasPixelArray.prototype.set = (arr) ->
+        for i in [0...@length]
+            @[i] = arr[i]
+
     # map plots reference this attribute
     @map_div = @$('div.bk-canvas-map') ? null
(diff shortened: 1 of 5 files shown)


bryevdv:
As far as I can tell none of our uses of `Uint8ClampedArray` actually would rely on or need the clamping capability. Accordingly, I think `Uint8Array` is fine, and testing so far seems to indicate the same.

bokehjs/src/coffee/models/canvas/canvas.coffee
```diff
@@ -32,6 +32,13 @@ export class CanvasView extends BokehView
     fixup_measure_text(@ctx)
     fixup_ellipse(@ctx)
 
+    # fixes up a problem with some versions of IE11
+    # ref: http://stackoverflow.com/questions/22062313/imagedata-set-in-internetexplorer
+    if window.CanvasPixelArray?
+      CanvasPixelArray.prototype.set = (arr) ->
+        for i in [0...@length]
+            @[i] = arr[i]
```
mattpap: It would be better to modify particular instances of this class than the class itself. It's very bad practice to touch class' prototype.

bryevdv: It also only applies so some sub-versions of one particular version of one particular browser. I'm inclined to just keep it simple this way.

PR #5534, opened 2016-12-10
merge gridplot tools in nested cases

### LINKED ISSUE #5479: Merged toolbar is not created properly when row/column layouts added to gridplot

bryevdv:
- [x] issues: fixes #5479
- [x] tests added / passed

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -368,9 +368,11 @@ def gridplot(*args, **kwargs):
         row_tools = []
         row_children = []
         for item in row:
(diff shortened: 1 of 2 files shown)


bryevdv:
Tested locally. Not sure there aren't other layout issues with nested layouts in grid plots, but the tools coalesce correctly, so I will merge this now.

PR #5535, opened 2016-12-10
use io.open with encoding whenever file writing by hand

### LINKED ISSUE #5306: embed/embed_multiple fails with unicode error on py2

bryevdv:
issues: fixes #5306

Went ahead and updated all the examples that did an explicit `open` and `write`

### DIFF
--- a/examples/embed/animated.py
+++ b/examples/embed/animated.py
@@ -23,6 +23,8 @@ in this directory. Navigate to
 """
 from __future__ import print_function
 
+import io
+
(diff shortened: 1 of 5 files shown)


bryevdv:
Tested all the updated examples by hand on to verify they still work on my system. Merging as this should be uncontroversial.

PR #5543, opened 2016-12-12
disable conda nb extension on travis

### LINKED ISSUE #5375: Notebook image diff tests broken due to "missing kernel"
This seems to be some new jupyter notebook feature we need to account for in how we run during tests. 

screen shot 2016-10-18 at 9 50 00 am

ping @damianavila

bryevdv:
issues: fixes #5375

### DIFF
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -9,6 +9,7 @@ MINICONDA="Miniconda-$MINICONDA_VERSION-Linux-x86_64"
 MINICONDA_URL="http://repo.continuum.io/miniconda/$MINICONDA.sh"
 
 PINNED_PKGS=$(cat <<EOF
+notebook ==4.2.1
 EOF
 )
 
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -57,7 +57,7 @@ conda install --yes $DEPS_RUN
 conda install --yes $DEPS_TEST
 
 # conda nb extension is installed by default and breaks notebook image diff tests
-conda remove --yes nb_conda_kernels
+python -m nb_conda_kernels.install --disable --prefix=/home/travis/miniconda
 
 conda install --yes nodejs
 echo "node version $(node -v)"
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -57,8 +57,7 @@ conda install --yes $DEPS_RUN
 conda install --yes $DEPS_TEST
 
 # conda nb extension is installed by default and breaks notebook image diff tests
-jupyter-nbextension disable nb_conda_kernels --py --sys-prefix
-jupyter-serverextension disable nb_conda_kernels --py --sys-prefix
+conda remove --yes nb_conda_kernels
 
 conda install --yes nodejs
 echo "node version $(node -v)"
--- a/scripts/travis_install
+++ b/scripts/travis_install
@@ -56,6 +56,10 @@ EOF
 conda install --yes $DEPS_RUN
 conda install --yes $DEPS_TEST
 
+# conda nb extension is installed by default and breaks notebook image diff tests
+jupyter-nbextension disable nb_conda_kernels --py --sys-prefix
+jupyter-serverextension disable nb_conda_kernels --py --sys-prefix
+
 conda install --yes nodejs
 echo "node version $(node -v)"
 echo "npm version $(npm -v)"


bryevdv:
ping @damianavila

mattpap:
@bryevdv, maybe this should be also reported upstream that this behavior is just unacceptable.

bryevdv:
OK, that's apparently not the right syntax:
```
+jupyter-nbextension disable nb_conda_kernels --py --sys-prefix
  File "/home/travis/miniconda/bin/jupyter-nbextension", line 6, in <module>
    sys.exit(notebook.nbextensions.main())
  File "/home/travis/miniconda/lib/python2.7/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/home/travis/miniconda/lib/python2.7/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 969, in start
    super(NBExtensionApp, self).start()
  File "/home/travis/miniconda/lib/python2.7/site-packages/jupyter_core/application.py", line 256, in start
    self.subapp.start()
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 877, in start
    self.toggle_nbextension_python(self.extra_args[0])
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 853, in toggle_nbextension_python
    logger=self.log)
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 482, in disable_nbextension_python
    logger=logger)
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 374, in _set_nbextension_state_python
    m, nbexts = _get_nbextension_metadata(module)
  File "/home/travis/miniconda/lib/python2.7/site-packages/notebook/nbextensions.py", line 1143, in _get_nbextension_metadata
    'it is missing the `_jupyter_nbextension_paths()` method.'.format(module))
KeyError: 'The Python module nb_conda_kernels is not a valid nbextension, it is missing the `_jupyter_nbextension_paths()` method.'
```

bryevdv:
OK well this didn't work, now it fails with the notebook showing just a password dialog even though they are not password protected.

bryevdv:
Tho, @mattpap can you comment on why/how *both* the current and the reference images in 

https://s3.amazonaws.com/bokeh-travis/183386517/examples.html

are the password dialog images? I don't see that any previous commit would have ever had this as the screenshot, so where did the reference image come from?

bryevdv:
With this latest change (including a pin to notebook < 4.3) current images seem to be getting generated correctly:

https://s3.amazonaws.com/bokeh-travis/183435067/examples.html

bryevdv:
will merge when green

bryevdv:
Not sure what's going on the second build has finished with 0 exit three times but then stalled and failed due to time out. Seems to be a travis hiccup, so I am going to go ahead and merge

bryevdv:
Well, looks like it is still failing even though it reports "Done. Your build exited with 0." I've sent an email to TravisCI support to advise.

bryevdv:
Confirmed it's happening on re-started previously successful builds, do definitely seems like a Travis issue

damianavila:
I think the `nb_conda_kernles` disabling is probably the quickest thing to do... long term we should look why it was failing...

The "password" thing comes from the fact on the default token auth on notebook 4.3. Short-term you can pin, of course... medium-term, you could use the config option to disable this, see here: http://jupyter-notebook.readthedocs.io/en/latest/security.html

bryevdv:
@damianavila I think you misunderstood, the password thing is not an issue, I pinned to `4.2.3`. But if we *dont* disable the nb conda kernels then the NB tests are worthless (just show an image of the env selection dialog). And if we *do* disable them, then the 2.7 build fails entirely. There is no solution here yet. I am only saying there is no point in running notebook tests *at all* apparently, so we should just disable all notebook tests completely.

damianavila:
> @damianavila I think you misunderstood, the password thing is not an issue, I pinned to 4.2.3

It is not an issue now, but you can not pin forever...

>But if we dont disable the nb conda kernels then the NB tests are worthless (just show an image of the env selection dialog). And if we do disable them, then the 2.7 build fails entirely. There is no solution here yet. I am only saying there is no point in running notebook tests at all apparently, under either scenario, so therefore we should just disable all notebook tests completely.

Yep, I understand, this is why I proposed #5574.

PR #5544, opened 2016-12-13
Base64 serialization

philippjfr:
Reopened PR #5504 as bokeh/bokeh branch to allow running image diffs. This PR implements some of the suggestions in https://github.com/bokeh/bokeh/issues/2204. In further discussion and initial testing (see https://github.com/bokeh/bokeh/pull/5429) we found that simple base64 encoding offers huge performance gains compared to the current implementation. A more complex binary json format is therefore not necessary at this time.

## Motivation

Currently sending large arrays to the browser can take a very long time, particularly when sending relatively large images. This is because all data is sent as raw strings, which is highly inefficient both in terms of size of the data that has to be sent to the frontend and in terms of processing because both backend and frontend iterate over the arrays to validate them. For a simple 2x2 image column data source the JSON representation would previously have looked like this:

```
{'image': [[[0.19593542932534513, 0.8104239473495675], [0.3907588555073921, 0.15267810064366616]]]}
```

the new JSON representation looks as follows:

```
{'image': [{'data': 'yKTPhb3b2j/Qy4BxlXzvP1bfkgIid9Q/kz0MlC4F4T8=', shape: [2,2], dtype: 'float64'}]}
```

In the case of a 2000x2000 image it took up to 30 seconds to serialize and send the image. Through initial investigation we found that simple base64 encoding of arrays can result in huge performance gains for sending array data via web sockets. Overall this has the potential to bring 10-50x improvements in sending large arrays via the comms in the notebook and will also offer some more limited improvements for bokeh server.

## Design

1. The encoding should happen as a last step during serialization on the Python end and as the first step after deserialization on the JS end, ensuring no other code has to deal with base64 encoded data avoiding any backward compatibility issues.
2. The data should be wrapped in a simple JS object along ith the shape and type information. A dedicated model might be considered in future but is not necessary.
3. The column data source on the fronted will unpack and store the shape information on the _shapes attribute.
4. The encoding behavior should be configurable, e.g. do not encode unless the array is larger than a configurable size.
5. All numeric column data source data should potentially be encoded, i.e. both 1-D and 2-D arrays.

## Caveats

Currently this PR will correctly encode and decode any numerical types except for numpy int64 arrays. Since there is no native 64-bit integer type in Javascript (internally ints are handled as double precision floats where the maximum safe integer is 2**53), we cannot directly encode and decode 64-bit integer data. Using the npm ``int64-buffer`` package I managed to decode an ``int64`` ArrayBuffer but the advantages and disadvantages of various approaches should be evaluated in a subsequent PR:

```coffeescript
view = new Int32Array(bytes)
array = []
for i in [0...view.length/2]
  int = new Int64LE(view.buffer.slice(i*8,(i+1)*8))
  array.push(int)
```

## To-Do

- [x] Add configurable options ``BOKEH_USE_BINARY_ARRAYS`` and ``BOKEH_BINARY_ARRAY_CUTOFF`` and document them
- [x] Add server and client side decoding and encoding
- [x] Handle patch events
- [x] Handle stream events
- [ ] Ensure current tests pass and add extensive tests
- [ ] Add release document entry (if new feature or API change)


bryevdv:
FYI currently the second test is failing due to an unrelated TravisCI reason. I have contacted their support. Here is an initial image diff report:

https://s3.amazonaws.com/bokeh-travis/183601852/examples.html

Not bad! As you mention, charts are failing for some reason but I expect this is a pretty minor fix, whatever it is.

philippjfr:
>Not bad! As you mention, charts are failing for some reason but I expect this is a pretty minor fix, whatever it is.

Turned out to be a bug in the deserialization logic, I have to look for a specific key in the object. Currently I'm looking for ``shape`` but I guess there's a chance that could clash. Almar originally suggested using something like ``__array__`` for the data, which would be a safer key to look for. Latest output here, are those remaining differences expected?

https://s3.amazonaws.com/bokeh-travis/183791510/examples.html

philippjfr:
On a separate note, while this PR has been mostly focused on base64 encoding I have investigated some of the alternatives laid out in the original issue. Specifically I looked at both ``ubjson`` and ``msgpack``. After investigating ``ubjson`` a bit I discovered that while the ubjson explicitly supports typed array buffers, none of the Python implementations ([simplejson](https://github.com/kxepal/simpleubjson) and [py-ubjson](https://github.com/Iotic-Labs/py-ubjson))  actually implement the encoding component. ``msgpack`` would work and there is even a numpy extension to let you easily serialize numpy arrays, but I'm questioning whether two additional Python dependencies on top of a msgpack decoder in JS would be worth it.

bryevdv:
> but I'm questioning whether two additional Python dependencies on top of a msgpack decoder in JS would be worth it.

For python it only matters how easily available they are -- care there both pip and conda pkgs ready to go? For JS the main concern is size.

mattpap:
> "Fixed bug in HasProps._value_to_json"

I would prefer, perhaps, `Change {v->value} in HasProps._value_to_json()`. By saying "Fixed bug (...)", without further explanation, I don't know if this is a bug from before this PR or was introduced here. If it's the later, then raising attention so much isn't necessary and just saying what was changed would be more appropriate. Though I still prefer the existing title over, e.g., "fix", as it still happens from time-to-time.

philippjfr:
>I would prefer, perhaps, Change {v->value} in HasProps._value_to_json()

Thanks, I'll be more explicit in future.

bokeh/core/properties.py
```diff
@@ -1181,7 +1185,7 @@ def _sphinx_type(self):
     def from_json(self, json, models=None):
         if json is None:
             return None
-        elif isinstance(json, list):
+        elif isinstance(json, (list, np.ndarray)):
```
philippjfr: Need to guard against numpy not being available.

bryevdv: NumPy is a hard requirement

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -317,6 +317,7 @@ export class HasProps extends Backbone.Model
           immediate = v._immediate_references()
           for obj in immediate
             HasProps._value_record_references(obj, result, true) # true=recurse
+    else if v?.buffer instanceof ArrayBuffer
```
mattpap: `?` is unnecessary here. There's already a check for `null` in the first clause. Does it happen to `v` to be `undefined` as well? If so, the first clause should be rewritten as `v?`.

bokehjs/src/coffee/core/util/serialization.coffee
```diff
@@ -15,12 +15,12 @@ for k, v of ARRAY_TYPES
     DTYPES[v.name] = k
 
 _arrayBufferToBase64 = (buffer) ->
-  binary = ''
+  binary = []
   bytes = new Uint8Array( buffer )
   len = bytes.byteLength
   for i in [0...len]
```
mattpap: Just to note, unless you actually need the index, you can safely write more readable yet efficient code for this in CoffeeScript, i.e.:
```
for b in bytes
 binary.push(String.fromCharCode(b))
```
or even:
```
binary = (String.fromCharCode(b) for b in bytes)
```
One of few good features of CoffeeScript are its `for` loops which get translated into efficient `for(var i, ...)` loops.

bokehjs/src/coffee/document.coffee
```diff
@@ -650,8 +651,13 @@ export class Document
             throw new Error("Cannot apply patch to #{patched_id} which is not in the document")
           patched_obj = @_all_models[patched_id]
           attr = event_json['attr']
-          value = Document._resolve_refs(event_json['new'], old_references, new_references)
-          patched_obj.setv({ "#{attr}" : value })
+          model_type = event_json['model']['type']
+          if attr == 'data' and model_type == 'ColumnDataSource'
+            [data, shapes] = decode_column_data(event_json['new'])
+            patched_obj.setv({ "_shapes": shapes, "data" : data}, {notify: false})
```
mattpap: What's the difference between `notify=false` and `silent=true`?

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -5,21 +5,77 @@ import * as hittest from "../../core/hittest"
...
+export concat_typed_arrays = (a, b) ->
+  c = new (a.constructor)(a.length + b.length)
+  c.set(a, 0)
+  c.set(b, a.length)
+  return c
+
+#exported for testing
+export stream_to_column = (col, new_col, rollover) ->
+  # handle regular (non-typed) arrays
+  if col.concat?
```
mattpap: `col instanceof Array`?

bokehjs/src/coffee/document.coffee
mattpap: Also please use `{_shapes: shapes, ...}` (no quotes).

bokehjs/src/coffee/core/util/serialization.coffee
philippjfr: Thanks, all suggestions for more idiomatic coffeescript welcome.

bokehjs/src/coffee/document.coffee
philippjfr: >What's the difference between notify=false and silent=true?
  
``silent=true`` completely skips telling the document about the change, which means models won't be invalidated and therefore not re-rendered. ``notify=false`` ensures the models are invalidated (and therefore re-rendered) but skips notifying any document callbacks, which means the ``ClientSession`` document listener isn't called and no ModelChangeEvent is generated. This significantly improves performance even independent of the other changes in this PR because when the client receives a ModelChangeEvent it no longer generates another ``ModelChangeEvent``, which has to be compared against the current data before it is invalidated by ``ClientSession._should_suppress_on_change``.

bokehjs/src/coffee/core/has_props.coffee
philippjfr: I don't think ``v`` should ever be undefined here so I think the first clause should be sufficient.

philippjfr:
Just FYI I'll perform a interactive rebase soon to clean up the commits a bit.

bokehjs/src/coffee/document.coffee
philippjfr: Without this change it would have to re-encode the typed arrays so just so that ``ClientSession._should_suppress_on_change`` could compare and suppress them correctly.

philippjfr: I suppose the question is, are there any other kinds of callbacks apart from ``ClientSession`` callbacks that could be installed on a ``Document.on_change`` event?

bokehjs/gulp/tasks/test.coffee
```diff
@@ -4,6 +4,9 @@ mocha = require "gulp-mocha"
 paths = require "../paths"
 utils = require "../utils"
 
+global.atob = require "atob"
+global.btoa = require "btoa"
```
philippjfr: @mattpap ``atob`` and ``btoa`` aren't available in node so I have to add them manually. Is this the proper way to do it?

mattpap: Those should be added in `bokehjs/test/index.coffee` next to other `global.X = ...` lines. Just merge `master` first in to this branch.

mattpap:
> Just FYI I'll perform a interactive rebase soon to clean up the commits a bit.

Given that we squash merge anyway, this seems like a waste of time.

philippjfr:
>Given that we squash merge anyway, this seems like a waste of time.

@bryevdv suggested I do it to make reviewing the PR a bit easier. Up to you guys.

bryevdv:
I'm hoping to have @havocp take a quick pass over this, and I'd like to make it as quick/easy as possible

bokeh/core/properties.py
```diff
@@ -349,7 +352,10 @@ def serializable_value(self, obj):
         value to appear simpler for developer convenience.
 
         """
-        return self.__get__(obj)
+        value = self.__get__(obj)
+        if hasattr(self.descriptor, 'serialize_value'):
```
havocp: why not have a no-op `serialize_value` in the descriptor base class to avoid this hasattr?

bokeh/core/properties.py
```diff
@@ -1181,7 +1187,9 @@ def _sphinx_type(self):
     def from_json(self, json, models=None):
         if json is None:
             return None
-        elif isinstance(json, list):
+        elif isinstance(json, bokeh_seq_types):
+            if isinstance(self.item_type, Any):
+                return json
```
havocp: I'm confused about what's happening here; I haven't read the whole PR yet as I'm reading this, but here are the questions I have looking at these lines:

 * how does json have a numpy array in it now when we didn't used to handle that (shouldn't json be plain json types?)
 * Any.from_json is now only called for single Any values but not for arrays, which could be dangerous
 * what happens to existing code that has a `Seq(Any)` where the Any actually does need from_json to be called; it could be any type at all in there, including objects, as far as I remember?

bokeh/models/sources.py
```diff
@@ -64,7 +64,7 @@ class ColumnDataSource(DataSource):
 
     """
 
-    data = Dict(String, Seq(Any), help="""
+    data = ColumnData(String, Seq(Any), help="""
```
havocp: some sort of `ColumnArray` type instead of `Seq(Any)` might allow us to use a to/from_json override in `ColumnArray` instead of the `isinstance(item_type, Any)` thing. Or `ColumnData` could also have a `from_json` that knows when to bypass the `Seq` from_json, so standalone `Seq(Any)` outside of `ColumnData` would be unaffected.

bokeh/settings.py
```diff
@@ -218,6 +232,14 @@ def nodejs_path(self, default=None):
 
 #: A global settings object that other parts of Bokeh can refer to.
 #:
+#:
+#: ``BOKEH_BINARY_ARRAY_CUTOFF`` --- Number of samples before arrays are encoded.
+#:
+#:   Valid values include all positive integers. Ensures that binary array
+#:   encoding is only applied to large arrays, avoiding the overhead
+#:   in small cases. Only applies when ``BOKEH_USE_BINARY_ARRAYS`` is
+#:   enabled.
```
havocp: Why not hardcode this to something reasonable, having trouble imagining it being important to tune this?

bokeh/settings.py
```diff
@@ -326,6 +348,12 @@ def nodejs_path(self, default=None):
 #:   to an affirmative value will result in more friendly simple numeric IDs
 #:   counting up from 1000.
 #:
+#: ``BOKEH_USE_BINARY_ARRAYS`` --- Whether to encode data arrays before sending them.
+#:
+#:   Accepted values are ``yes``/``no``, ``true``/``false`` or ``0``/``1``.
+#:   Can have a huge effect on performance because encoding and decoding
+#:   arrays as raw strings can be very slow.
```
havocp: There's no mention of any reason to turn this off, here; I think if we can't say why you'd turn it off, it shouldn't be configurable. Being able to toggle this means eventually the non-default option will be broken in a release, probably, because nobody tested it...

havocp: also, if binary allows Nan/Infinity to work, that's a semantic difference right, which isn't ideal to have a config option affecting...

bokeh/core/properties.py
```diff
@@ -1255,6 +1263,23 @@ def from_json(self, json, models=None):
         else:
             raise DeserializationError("%s expected a dict or None, got %s" % (self, json))
 
+class ColumnData(Dict):
+    """Property holding column data in form of a dict. Also applies
+    encoding and decoding to the data.
+    """
```
havocp: You could enforce that this Dict is from `String` to `ColumnArray` and have another class `ColumnArray` that handles the from_json and serialize_value on the arrays maybe, to avoid isinstance checks

havocp: Alternatively, put the virtualization of from_json all here in this `ColumnData` class ... use `Seq(Any)` but in the from_json for this `ColumnData` class, do not call `from_json` on the values immediately; instead, first check whether they are base64-encoded arrays and if so handle them directly, and if not call their `from_json`. So all base64-special-casing would be here in `ColumnData`.

bokeh/core/properties.py
philippjfr: Good point, definitely cleaner that way.

bokeh/settings.py
bryevdv: My thought was that sometimes its nice to see the list output for debugging. However, I seem to be in the minority, so I would not oppose removing these.  I do agree and was thinking yesterday about the nan/inf issue.

bokeh/core/properties.py
philippjfr: I think your suggestion of a ``ColumnArray`` property type is a good one. I'll go with that.

bokeh/core/properties.py
mattpap: > `if isinstance(self.item_type, Any)`

btw. If only inheritance structure was set properly up in `bokeh.core.properties`, this could would be equivalent of `if True`. This is an equivalent of asking if `isinstance(value, object)` for some `value`.

philippjfr: Right, so definitely not a good idea then.

bokeh/core/properties.py
philippjfr: Since it's not straightforward to move both the encoding and decoding on ``ColumnArray`` (because property descriptors aren't called during serialization), I've gone with your second suggestion in the end.

bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -386,9 +387,11 @@ export class HasProps extends Backbone.Model
       if need_invalidate
         @document._invalidate_all_models()
 
-      @document._notify_change(@, attr, old, new_)
+      if not options?.notify? or options.notify == true
+        @document._notify_change(@, attr, old, new_)
```
philippjfr: @havocp I think this is the main thing I need your input on. When a ``ModelChangeEvent`` is received by the frontend it now sets a new ``notify=false`` option which gets passed through to here. This has the effect of preventing all callbacks on the Document from being notified about the change. This results in a fairly large speedup because it doesn't have to compare all the events that are received only to then suppress them anyway.

My question is, is this safe? What other types of callbacks apart from a ClientSession might be attached to a document? Is there a case where you do actually want a received ModelChangeEvent to trigger these callbacks?

philippjfr:
@bryevdv @havocp I've now revised some of the client-side code after some of the great suggestions @havocp made. When receiving a ``ModelChangeEvent`` the ``ClientSession`` now provides a ``setter_id``, which is passed all the way through to newly generated ModelChangeEvent, which means that once that event roundtrips back to the ``ClientSession`` it can just suppress it by comparing the id. This works very well and doesn't have the issues suppressing other event types. Separately I have now changed ``should_suppress_on_change`` to compare the JSON event not the actual event model ensuring that it doesn't compare the encoded and decoded data. This does mean it potentially needlessly encodes data but it should only do so in response to the user setting the data, never because a server based event has been received. Hope that makes more sense, I'm much more comfortable with this approach.

bryevdv:
Just a quick data point for fun and encouragement. Even with a 20% *higher* update rate, the spectrogram CPU usage is now more like:

screen shot 2016-12-21 at 9 38 49 am

which is a huge improvement from where it was. (More qualitatively: my fan does not ramp up anymore). 

 There's probably still room for more improvement, too, if we ever send all arrays (not just CDS columns) or find a way to "sub-stream" slices into 2d arrays in CDS. Right now each "latest" waterfall gram column is sent using lists (it's just a single `Seq` property on the custom renderer)

bokehjs/src/coffee/client.coffee
```diff
@@ -384,32 +384,37 @@ class ClientSession
     @request_server_info().then((ignored) -> undefined)
 
   _should_suppress_on_change : (patch, event) ->
```
havocp: I'm not sure without hand-verifying it. It seems like they should be?

bokehjs/src/coffee/client.coffee
```diff
@@ -384,32 +384,37 @@ class ClientSession
...
     return false
 
   _document_changed : (event) ->
-    if @_current_patch? and @_should_suppress_on_change(@_current_patch, event)
+    # Filter out events that were initiated by the ClientSession itself
+    if event.setter_id == @id
+      return
+
+    json_patch = @document.create_json_patch([event])
+    if @_current_patch? and @_should_suppress_on_change(@_current_patch, json_patch['events'][0])
```
havocp: is it actually possible to need this _should_supress_on_change code now? I guess it can only happen if 1) we have a current patch, which means we have a set() called by us on the stack with our setter_id, and 2) someone called set separately from our own set, and 3) they set the thing to same value that we did ? Pretty far-fetched right? Maybe we can delete all of this and rely on setter_id instead?

bokehjs/src/coffee/document.coffee
```diff
@@ -23,7 +24,10 @@ export class ModelChangedEvent extends DocumentChangedEvent
       throw new Error("'id' field should never change, whatever code just set it is wrong")
 
     value = @new_
-    value_json = HasProps._value_to_json('new_', value, @model)
+    if @model instanceof HasProps
+      value_json = @model.constructor._value_to_json(@attr, value, @model)
+    else
+      value_json = HasProps._value_to_json(@attr, value, @model)
```
havocp: I'm not sure I understand what's going on above... aren't all models instances of HasProps?
It feels like this block of code should be encapsulated in HasProps._value_to_json somehow.

bryevdv: I was also wondering what was going on here for the same reason

bokehjs/src/coffee/document.coffee
```diff
@@ -650,8 +654,13 @@ export class Document
             throw new Error("Cannot apply patch to #{patched_id} which is not in the document")
           patched_obj = @_all_models[patched_id]
           attr = event_json['attr']
-          value = Document._resolve_refs(event_json['new'], old_references, new_references)
-          patched_obj.setv({ "#{attr}" : value })
+          model_type = event_json['model']['type']
+          if attr == 'data' and model_type == 'ColumnDataSource'
+            [data, shapes] = decode_column_data(event_json['new'])
+            patched_obj.setv({_shapes: shapes, data: data}, {setter_id: setter_id})
```
havocp: this makes me think, we aren't handling _resolve_refs in this branch, which means we don't allow column data to contain models; which is probably a good idea; but it isn't captured by `Dict(String, Seq(Any))` in the Python model.
Is this the only way in which that `Any` is a lie or are there tons of other unacceptable column types?

bryevdv: Well, it's not quite a lie, the translation from a NumPy array (which is a `Seq(Any)` to this "little object" that represents typed arrays happens at the very edge of the serialization  boundary (after validation of the NumPy array as a `Seq`). That's basically because the wire-format for typed arrays is something users should never ever see or know about. (tho @philippjfr please correct if I have mis-stated anything) Not sure if there is another way to do or express things.

bokehjs/src/coffee/client.coffee
philippjfr: You're right I think, I don't think there's a good reason to have this check now.

bokehjs/src/coffee/document.coffee
philippjfr: Yes, this is a bit awkward and needless. Is there an issue calling ``@model.constructor._value_to_json`` instead of encapsulating it all on HasProps. The ``ColumnDataSource`` overrides the ``_value_to_json`` static method to perform the serialization. Other suggestions on where this serialization should take place also welcome.

bokehjs/src/coffee/document.coffee
philippjfr: The only options I see are either to override ``setv`` or to add something like ``set_from_json`` that's used in Python.

philippjfr:
@bryevdv @havocp I have now tried to do the same thing on the Python side. On the JS end this involved passing the ``setter_id`` through a few methods, but on the Python end I have to retrofit all ``set_from_json`` and ``__set__`` methods with optional arguments, which gets a bit messy. Let me know if you'd still prefer it for consistency in which case I'll push it.

bryevdv:
I think it's at least worth looking at, we can always revert if need be

philippjfr:
@bryevdv Just pushed the Python component to this. As I said before it's pretty messy but works. Let me know what you think.

bryevdv:
@mattpap @havocp any thoughts on 	c9089d6 ? I agree with @philippjfr that it's a little clunky to have to pass `setter_id` around like that but I also think that getting rid of the expensive and incredible complicated `should_suppress_on_change` is a valuable change.

mattpap:
>  any thoughts on c9089d6?

That's ugly indeed.

bryevdv:
> That's ugly indeed.

I think no one would disagree with that. But `should_suppress_on_change` is also ugly, and expensive, and fragile and complicated. It's also not on the JS side anymore so getting rid of it brings the two implementations into closer alignment. So the question is: is the trade-off worthwhile? (Or, do you have a suggestion for a cleaner way to transmit this information through?)

mattpap:
@bryevdv, no, I have to give this a though. That's why I'm just stating the obvious for now.

philippjfr:
It just got even worse unfortunately, I had to add the handling to RootAdded, RootRemoved and TitleChange events as well.

havocp:
I'd vote for the setter_id change if it's working, the patch really isn't that big, and the ugly is internal (not exposed to app developers). It's 1) more correct and 2) more performant than the should_suppress_on_change goo.

bokeh/server/session.py
```diff
@@ -214,7 +213,7 @@ def _handle_patch(self, message, connection):
         self._current_patch = message
         self._current_patch_connection = connection
         try:
-            message.apply_to_document(self.document)
+            message.apply_to_document(self.document, self.id)
```
havocp: a minor detail but I think you could use simply `self` rather than `self.id` as long as you then test `getattr(event, 'setter_id', None) is self` (`is` rather than `==`).

bokeh/server/session.py
```diff
@@ -165,8 +165,7 @@ def _wrap_session_callbacks(self, callbacks):
         return wrapped
 
     def _document_patched(self, event):
-        may_suppress = self._current_patch is not None and \
-                       self._current_patch.should_suppress_on_change(event)
+        may_suppress = getattr(event, 'setter_id', None) == self.id
```
havocp: making events always have setter_id (but it can be None) is nicer than using getattr perhaps

philippjfr:
@bryevdv Just fixed and updated a number of unit tests. I replaced the tests for ``should_suppress_on_change`` with some tests that make sure that the setter is passed all the way through to the ``document.on_change`` callbacks as an ``event`` attribute, but I guess those tests could be moved elsewhere now.

There is also one test failure I haven't yet figured out yet: ``test_unit_spec_changes_do_not_boomerang``. Would be great if you could figure out what's going on there.

bokeh/server/session.py
philippjfr: Agreed, done.

bokeh/server/session.py
philippjfr: Switched to using ``setter`` rather than ``setter_id`` and checking for identity at least on the python end.

bryevdv:
@philippjfr I will take a look at that last test today. FYI I resolved a conflict with an encoder fix PR that added `deque` support.

bryevdv:
OK I got around to this super late, but I have everything built and up and running, so I will look at it more first thing in the morning. Although it's probably not super helpful, I can say that commenting out that one line:
```
change_to({ 'field' : 'foo' }, { 'field' : 'bar' })
```
 causes the test to pass, so it's just that one part of the test, where the `field` changes, that is problematic

bryevdv:
A few more quite notes:

* changing test to this does NOT cause test to pass:
    ```
    change_to({ 'field' : 'foo', 'units': 'data' }, { 'field' : 'bar', 'units': 'rad' })
    ```

* Here is the actual boomeranged message:

```
(Pdb) got_angry['result'].__dict__
{'_metadata_json': '{}', '_content': {'events': [{'attr': 'distance', 'new': {'units': 'data', 'field': 'foo'}, 'kind': 'ModelChanged', 'model': {'id': '4317785f-2174-4c27-ad84-e721789bfe67', 'type': 'UnitsSpecModel'}}], 'references': []}, '_header_json': '{"msgid": "b343cd48-03ca-4896-b6ca-4c2982dc75c4", "msgtype": "PATCH-DOC"}', '_buffers': [], '_metadata': {}, '_content_json': '{"events": [{"attr": "distance", "new": {"units": "data", "field": "foo"}, "kind": "ModelChanged", "model": {"id": "4317785f-2174-4c27-ad84-e721789bfe67", "type": "UnitsSpecModel"}}], "references": []}', '_header': {'msgid': 'b343cd48-03ca-4896-b6ca-4c2982dc75c4', 'msgtype': 'PATCH-DOC'}}
```
My guess in the previous code would have been that somehow the `units` field was not being properly accounted for in the suppression comparison, now that that is gone, I'm not quite sure offhand. I'll keep looking in the morning.

bokehjs/src/coffee/client.coffee
```diff
@@ -383,33 +382,9 @@ class ClientSession
...
-        for event_json in patch.content['events']
-          if event_json['kind'] == 'TitleChanged' and event_json['title'] == event.title
-            return true
-
-    return false
-
   _document_changed : (event) ->
-    if @_current_patch? and @_should_suppress_on_change(@_current_patch, event)
+    # Filter out events that were initiated by the ClientSession itself
+    if event.setter_id == @id
```
bryevdv: was this meant to change to `setter` instead of `setter_id` ?

bokehjs/src/coffee/client.coffee
```diff
@@ -422,11 +397,7 @@ class ClientSession
     @_connection.send(patch)
 
   _handle_patch : (message) ->
-    @_current_patch = message
-    try
-      @document.apply_json_patch(message.content)
-    finally
-      @_current_patch = null
+    @document.apply_json_patch(message.content, @id)
```
bryevdv: same question re `setter` vs `setter_id`

bryevdv:
@philippjfr if you look at the actual event that is getting through back to the client, it is this:
```
{'hint': None, 'setter': None, 'attr': 'distance', 
 'serializable_new': {'units': 'data', 'field': 'foo'}, 
 'old': {'field': 'foo'}, 
 'new': {'units': 'data', 'field': 'foo'}, 
 'model': UnitsSpecModel(id='fb06a96e-27cc-4c72-8c52-e676ff96f94a', ...), 
 'document': <bokeh.document.Document object at 0x114bf1b38>
}
```
Look at old and new, they differ by only the addition of the `units` field. My current guess is that possibly units are getting set internally by some special case code and this is triggering a spurious additional event.

bryevdv:
OK, I think this is the path that goes awry:

* starts in `_notify_change` with a valid `setter`
* goes down through `UnitsSpec` method `to_serializable`
* a `getattr` seems to lead to `self._notify_owners` in a property container? 
* the notify leads to a `self._trigger(obj, old, value, hint, setter)` with `setter == None`

Step 3 is confusing to me currenty, not sure why `d["units"] = getattr(obj, name+"_units")` leads to a notify change call.

```
> /Users/bryan/work/bokeh/bokeh/document.py(631)_notify_change()
-> serializable_new = model.lookup(attr).serializable_value(model)
(Pdb) l
626                     self._all_models_by_name.remove_value(old, model)
627                 if new is not None:
628                     self._all_models_by_name.add_value(new, model)
629
630             if hint is None:
631  ->             serializable_new = model.lookup(attr).serializable_value(model)
632             else:
633                 serializable_new = None
634             self._trigger_on_change(ModelChangedEvent(self, model, attr, old, new, serializable_new, hint, setter))
635
636         @classmethod
(Pdb) setter
<bokeh.server.session.ServerSession object at 0x10b757c50>
(Pdb) d
> /Users/bryan/work/bokeh/bokeh/core/properties.py(1866)serializable_value()
-> return self.descriptor.to_serializable(obj, self.name, getattr(obj, self.name))
(Pdb) l
1861
1862    class DataSpecProperty(BasicProperty):
1863        """ A Property with a DataSpec descriptor."""
1864
1865        def serializable_value(self, obj):
1866 ->         return self.descriptor.to_serializable(obj, self.name, getattr(obj, self.name))
1867
1868        def set_from_json(self, obj, json, models=None, setter=None):
1869            if isinstance(json, dict):
1870                # we want to try to keep the "format" of the data spec as string, dict, or number,
1871                # assuming the serialized dict is compatible with that.
(Pdb) d
> /Users/bryan/work/bokeh/bokeh/core/properties.py(2133)to_serializable()
-> d["units"] = getattr(obj, name+"_units")
(Pdb) l
2128            return units_props + [ UnitsSpecProperty(descriptor=self, name=base_name, units_prop=units_props[0]) ]
2129
2130        def to_serializable(self, obj, name, val):
2131            d = super(UnitsSpec, self).to_serializable(obj, name, val)
2132            if d is not None and 'units' not in d:
2133 ->             d["units"] = getattr(obj, name+"_units")
2134            return d
2135
2136        def __str__(self):
2137            return "%s(units_default=%r)" % (self.__class__.__name__, self._units_type._default)
2138
(Pdb) d
> /Users/bryan/work/bokeh/bokeh/core/property_containers.py(19)wrapper()
-> self._notify_owners(old)
(Pdb) l
 14         '''
 15         def wrapper(*args, **kwargs):
 16             self = args[0]
 17             old = self._saved_copy()
 18             result = func(*args, **kwargs)
 19  ->         self._notify_owners(old)
 20             return result
 21         return wrapper
 22
 23     class PropertyValueContainer(object):
 24         ''' A base class for property container classes that support change
(Pdb)
```

bryevdv:
Ah, `d` in the `to_serializable` is ` property value dict:
```
(Pdb) type(d)
<class 'bokeh.core.property_containers.PropertyValueDict'>
```
it's setter  is wrapped with `notify_change` so that's the proximate cause I think. Question is what is the right thing to do

bryevdv:
@philippjfr (and @havocp for interest) I pushed a change that fixes the broken test. It seems reasonable but I'd like thoughts. Basically, if we are to modify things in `to_serializable`, need to convert from `PropertyValueDict` to `dict` fist. This seems OK, if it is going through `to_serializable` presumably theres nothing later that would expect / rely on the result being `PropertyValueDict` ?

bryevdv:
@philippjfr just FYI I force pushed a typo correction

havocp:
Having a getter modify a value seems clearly dangerous so change seems logical, though I remember that units code being pretty tricky.
The special property containers are purely to emit change notifications on models so they definitely arent needed in json.

bryevdv:
@philippjfr  @havocp 

OK, CI is green, and reports look good. I just tested this PR with all server examples under `apps`, `models`, and `plotting/server`. I also tested with all the `notebook_comms` notebooks, as well as a scattered selection of file examples (checking that the binary arrays were in the JSON as expected). I did not encounter any errors or problems. 

In my qualitative experience, things are snappier, more efficient, etc. For purposes of release notes/blogs it would be get to get a few hard numbers. 

So what is left before we consider merging this? Are there any other specific tests we could consider adding? I think getting into a dev build would make it easier for a larger group of people to potentially help test it. 

As an aside (not for this PR but should make an issue) we could consider encoding colors as 32bit ints to take advantage of this new capability as well.

bryevdv:
I thing there was one remaining comment by @mattpap about an `instanceof` instead of a duck typing check. @mattpap If you'd like to review and comment on this, please prioritize it accordingly. I'd ideally like to merge this maybe tomorrow to get a dev build into people's hands.

bryevdv:
PIng @jbednar as well, if there are integration/stress tests that you can do they would be valuable at this point.

philippjfr:
>OK, CI is green, and reports look good. I just tested this PR with all server examples under apps, models, and plotting/server on OSX/Safari. I also tested with all the notebook_comms notebooks, as well as a scattered selection of file examples (checking that the binary arrays were in the JSON as expected). I did not encounter any errors or problems.

Awesome, I've also run various examples and have been running this branch for all my holoviews development for the past few weeks, which should provide pretty significant coverage in itself. Getting it into the hands of other users in a dev build seems like a good idea at this point.

>In my qualitative experience, things are snappier, more efficient, etc. For purposes of release notes/blogs it would be get to get a few hard numbers.

I'd be happy to run some profiling to provide some hard numbers. From my own testing the improvements are pretty great, images, as used by datashader, are probably the largest beneficiary here and updating those plots is now dominated by computing the actual aggregates, which wasn't always the case before. I'll try to get some hard numbers on these things and also look at some server examples.

>So what is left before we consider merging this? Are there any other specific tests we could consider adding? I think getting into a dev build would make it easier for a larger group of people to potentially help test it.

Currently the PR is still lacking some tests on JS end to test things aren't boomeranging and the event setter is propagated appropriately. Getting others to test it will definitely be good though.

>As an aside (not for this PR but should make an issue) we could consider encoding colors as 32bit ints to take advantage of this new capability as well.

Yes, that seems like a good idea, although now that there is client-side colormapping I'd hope it's not too frequent an occurrence anyway. If that happens it might also be worth encoding palettes on colormappers this way.

jbednar:
I'd be happy to test out a dev release.

bryevdv:
OK, going to pull the trigger on this, and get a dev build out

jakirkham:
This looks really awesome! Thanks for doing this. 👍

bryevdv:
@jakirkham if you want to test it out, I just build `0.12.4rc1` any feedback / help testing is greatly appreciated.

jakirkham:
Ok, so I took this for a spin. Mostly looks good. Ran into one issue where I might need to change something, which I'm ok with. Am only trying to figure out how I should do this.

One of the things I'm trying to figure out is if I have a list of NumPy arrays in a `ColumnDataSource` that are passed into a `CustomJS` object where some JavaScript code works with them. Before I believe Bokeh was converting them to a list of lists, which worked fine. It seems currently they are packed into this special JavaScript representation of a NumPy array. I can workaround this by converting them to lists of lists beforehand. However, I would rather take this base64 representation of them and convert them in the JavaScript code. Is there some Bokeh function that I can call from the `CustomJS` code snippet to do this conversion?

jakirkham:
Also another observation I'm having is it seems this conversion occurs with a NumPy array. However, it doesn't happen with a list of NumPy arrays (e.g. a bunch of patches). It would be really interesting if we could handle this case. Though that is certainly out of scope for this PR and probably wouldn't happen in this release, but if there is interest I would like to discuss it more in a new issue.

philippjfr:
>Also another observation I'm having is it seems this conversion occurs with a NumPy array. However, it doesn't happen with a list of NumPy arrays (e.g. a bunch of patches).

Is that really true? At least when I was initially testing it this was working and I know that images work, which are supplied as list of one NumPy array. Could you provide an example of a plot where it doesn't encode a list of numpy arrays?

philippjfr:
>One of the things I'm trying to figure out is if I have a NumPy array in a ColumnDataSource that is being passed into a CustomJS object where some JavaScript code works with them. Before I believe Bokeh was converting it to a list of lists, which worked fine. It seems currently it is packed into this special JavaScript representation of a NumPy array.

Just to be clear, when the data source is passed to the CustomJS object it should already have been unpacked as a typed array. Are you saying that inside the CustomJS the data is still appearing as base64?

jakirkham:
>> One of the things I'm trying to figure out is if I have a NumPy array in a ColumnDataSource that is being passed into a CustomJS object where some JavaScript code works with them. Before I believe Bokeh was converting it to a list of lists, which worked fine. It seems currently it is packed into this special JavaScript representation of a NumPy array.

> Just to be clear, when the data source is passed to the CustomJS object it should already have been unpacked as a typed array. Are you saying that inside the CustomJS the data is still appearing as base64?

I wasn't clear what it was, but was clear it was not a 2D array. Now I'd welcome better suggestions on debugging these CustomJS objects, but I just added a couple of `alert` messages. It seems that it is an array of the right type, but the wrong shape. It appears to have been flattened into a 1D array. Is that expected?

philippjfr:
>I wasn't clear what it was, but was clear it was not a 2D array. Now I'd welcome better suggestions on debugging these CustomJS objects, but I just added a couple of alert messages. It seems that it is an array of the right type, but the wrong shape. It appears to have been flattened into a 1D array. Is that expected?

Yes this is expected. The flat version is far more efficient than the nested equivalent. But a utility might indeed be useful to make it easier to work with (thoughts @bryevdv?). The shapes of the arrays should be available on the ``_shapes`` attribute of the ``ColumnDataSource``. If I may ask what exactly are you doing to the nested arrays in the CustomJS callback?

jakirkham:
Thanks for the info. Some utility functions would definitely be nice. Can try hacking something in for the near term so as to play some more. What about striding? Are these always converted to contiguous C-order arrays when they get stored or do they preserve the original order/striding (e.g. F-order)?

Sure, the 2D array contains a set time traces for each identified ROI in the data. These are used as part of an interactive plot where users can select ROIs they want and see the corresponding traces. Link to an example plot below.

ref: http://dudmanlab.org/html/HiContrastForJohn_proj.html

jakirkham:
Cool, thanks for confirming. I'm just providing C-order so should be ok. Just wanted to make sure there weren't other things to consider.

Glad you like it. No worries. It seems that if I just use `tolist` to convert the NumPy array, things keep working with the RC just as they had before. Besides it is a great feature (and I do benefit from it in other ways).

Mainly I'm just trying to figure out what I need to do to leverage this feature in these other cases too.

jakirkham:
> Are these always converted to contiguous C-order arrays when they get stored or do they preserve the original order/striding (e.g. F-order)?

To answer my own question. Yes, it is enforced. Though those lines should probably precede the call to `transform_array_to_list` FWICT.

xref: https://github.com/bokeh/bokeh/blob/8e316faf2a12975c299bdde5fb002c2435f92642/bokeh/util/serialization.py#L190-L191

jakirkham:
>> Also another observation I'm having is it seems this conversion occurs with a NumPy array. However, it doesn't happen with a list of NumPy arrays (e.g. a bunch of patches).
>
> Is that really true? At least when I was initially testing it this was working and I know that images work, which are supplied as list of one NumPy array. Could you provide an example of a plot where it doesn't encode a list of numpy arrays?

To be fair, the case I’m looking at is a list of NumPy arrays of unequal length. So it would be perfectly reasonable to say that is out of scope for this PR. I’m sure a case where they are of equal length works fine as NumPy could detect the array-like nature of them. That all being said, I would be curious to explore optimizing this case more in Bokeh if there is interest.

jakirkham:
Thanks again for all of tips and pointers. Was able to move the traces into a base64 array and work with them that way just fine. As a follow-up question, does the underscore in `_shapes` have any significance? Namely is this a private API subject to change or is this ok for me to use?

Should also note that I had been shipping the colors in the `ColumnDataSource` as hex strings, but was able to leverage this PR to pack them into `uint8` N x 3 array and get them quite small. It was a pretty simple matter to convert them back into appropriate strings on the JavaScript side. It may be helpful for storing palettes compactly for other plots. Would be happy to share what I learned from this experience.

jakirkham:
Thanks again for all of tips and pointers. Was able to move the traces into a base64 array and work with them that way just fine. As a follow-up question, does the underscore in `_shapes` have any significance? Namely is this part of a private API subject to change or is this ok for me to use?

Should also note that I had been shipping the colors in the `ColumnDataSource` as hex strings, but was able to leverage this PR to pack them into a `uint8` N x 3 array and get them quite small. It was a pretty simple matter to convert them back into appropriate strings on the JavaScript side. It may be helpful for storing palettes compactly for other plots. Would be happy to share what I learned from this experience.

jakirkham:
So I think I figured out why the patches were not also getting encoded as `base64`. I wasn't thinking about type as an issue. The points were defaulting to `int64`, which as noted in the caveats section is not supported by this PR. As I can get away with a much smaller type most of the time (in all cases I can think of `uint16`), I went ahead and converted them to that type and encoding proceeded fine. Sorry for the false alarm and confusion.

bryevdv:
Right, unfortunately there is not a 64bit typed array (JavaScript! Everything is a float!  grrr....) so there's not anything we can do. Perhaps there's a warning or more documentation to add.

jakirkham:
Having a warning is a good idea. Maybe can link to docs with more details.

Did some searching and the way people recommend to handle this issue is to use the `Long` implementation in Google's Closure library. I don't know how effective this is or how it compares to things you have already looked at, but maybe it is worth a quick look. It should be doable to break it out if needed.

ref: https://github.com/google/closure-library/blob/master/closure/goog/math/long.js#L53

philippjfr:
>As a follow-up question, does the underscore in _shapes have any significance? Namely is this part of a private API subject to change or is this ok for me to use?

I don't foresee this changing any time soon but if we decide that users should be able to interact with it maybe it shouldn't have the underscore?

>Should also note that I had been shipping the colors in the ColumnDataSource as hex strings, but was able to leverage this PR to pack them into a uint8 N x 3 array and get them quite small. It was a pretty simple matter to convert them back into appropriate strings on the JavaScript side. It may be helpful for storing palettes compactly for other plots.

Yes, that may indeed be worthwhile as those are another fairly common occurrence of large arrays.

>The points were defaulting to int64, which as noted in the caveats section is not supported by this PR.

Sorry should have pointed this out when you mentioned it wasn't working. I think expanding documentation about this is a good idea. @bryevdv Where would be best suited? Should the ``ColumnDataSource`` documentation mention these limitations?

jakirkham:
> > Should also note that I had been shipping the colors in the ColumnDataSource as hex strings, but was able to leverage this PR to pack them into a uint8 N x 3 array and get them quite small. It was a pretty simple matter to convert them back into appropriate strings on the JavaScript side. It may be helpful for storing palettes compactly for other plots.
>
> Yes, that may indeed be worthwhile as those are another fairly common occurrence of large arrays.

Opened issue ( https://github.com/bokeh/bokeh/issues/5676 ) to outline my thoughts on this.

jakirkham:
> > As a follow-up question, does the underscore in _shapes have any significance? Namely is this part of a private API subject to change or is this ok for me to use?
>
> I don't foresee this changing any time soon but if we decide that users should be able to interact with it maybe it shouldn't have the underscore?

Thanks for the info. Wound up serializing the shapes myself as I wanted to remain compatible with 0.12.3 while potentially leveraging this feature from 0.12.4.

bokehjs/src/coffee/core/util/serialization.coffee
```diff
@@ -0,0 +1,90 @@
...
+  int32: Int32Array
+
+DTYPES = {}
+for k, v of ARRAY_TYPES
+    DTYPES[v.name] = k
+
+_arrayBufferToBase64 = (buffer) ->
+  bytes = new Uint8Array( buffer )
+  binary = (String.fromCharCode(b) for b in bytes)
+  return btoa( binary.join("") )
```
mattpap: @philippjfr, @bryevdv, `btoa` isn't imported imported. Is this code used at all? If this is used only in tests (as `global.btoa = require('btoa')` indicates), then this code should be moved to `bokehjs/test/utils.coffee`).

bokehjs/src/coffee/core/util/serialization.coffee
```diff
@@ -0,0 +1,90 @@
...
+
+_base64ToArrayBuffer = (base64) ->
+  binary_string = atob(base64)
```
mattpap: The same here.

bokehjs/src/coffee/core/util/serialization.coffee
philippjfr: This is the core of the serialization/deserialization and should be used all the time. It's used by other functions in this file, which are imported elsewhere.

philippjfr: Probably misunderstood your question, ``global.btoa  = require('btoa')`` is only defined for testing because browsers provide ``btoa`` and ``atob`` natively. Maybe I'm still confused about what you're asking though.

mattpap: @philippjfr, OK, I forgot that browsers provide those and node.js not. No issue then.

PR #5545, opened 2016-12-13
Add support for namespace models

### LINKED ISSUE #3674: Add DataModel base class
See https://github.com/bokeh/bokeh/issues/3674#issuecomment-528179333 for a new, better idea for this feature. 

---

Hi all, I'm having an issue where line plots will sometimes be invisible when rendered. This error seems to be non-deterministic; when I run the same code repeatedly, the plots will sometimes render as normal, and sometimes render invisibly.

An example page that is currently showing these "invisible" plots is available here (2nd plot from the top): https://sne.space/sne/SN1993J-broken.html. One thing you can notice is that hovering over the plot still produces tooltips, and an examination of the data within the page source reveals that the `line_color` attributes are non-white, and that the data is actually available. But for some reason, the lines are not visible!

And here's the exact same page when the render _does_ work: https://sne.space/sne/SN1993J.html. The file sizes are identical but the file contents are _not_ identical. It seems like it might be a different command order in the two files?

Anyone know what's going on?

mattpap:
fixes #3674

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -224,8 +224,7 @@ export class Document
 
     @_all_models = recomputed
 
-  roots : () ->
-    @_roots
+  roots: () -> @_roots
 
   _add_layoutable: (model) ->
     if model.layoutable isnt true
@@ -257,7 +256,6 @@ export class Document
     @_push_all_models_freeze()
     try
       @_roots.push(model)
-      model._is_root = true # TODO get rid of this?
     finally
       @_pop_all_models_freeze()
 
@@ -273,7 +271,6 @@ export class Document
     @_push_all_models_freeze()
     try
       @_roots.splice(i, 1)
-      model._is_root = false
     finally
       @_pop_all_models_freeze()
 
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -11,6 +11,7 @@ import {extend, values} from "./core/util/object"
 import {isEqual} from "./core/util/eq"
 import {isArray, isObject} from "./core/util/types"
 import {ColumnDataSource} from "./models/sources/column_data_source"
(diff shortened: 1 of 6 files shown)


bokehjs/test/models/plots/plot_canvas.coffee
```diff
@@ -272,14 +272,6 @@ describe "PlotCanvasView resize", ->
...
     expect(spy.calledWith([width, height], true)).to.be.true
 
-  it "should call canvas.set_dims and trigger if plot is_root", ->
-    spy = sinon.spy(@plot_canvas_view.canvas_view, 'set_dims')
-    @plot_canvas._is_root = true
-    @plot_canvas.sizing_mode = 'stretch_both'
-    @plot_canvas_view.resize()
-    expect(spy.calledOnce).to.be.true
-    expect(spy.calledWith([width, height], true)).to.be.true
-
```
bryevdv: This particular test needs to be *updated* not *removed*. It's testing an important behavior that should happen to any plot that is a root. Just because we no longer keep track with `_is_root` does not mean that the behavior is no longer important to maintain under test.

mattpap: Yeah, I know. WIP. btw. Can `PlotCanvas` be actually used as a root (without using `Plot`)?

bryevdv: Right, missed the `WIP` at first, sorry about that.  I don't actually know the answer to that question offhand.

bryevdv:
I'm inclined to merge this tomorrow as-is as a small, standalone incremental improvement.

mattpap:
@bryevdv, I still need to restore the test and I'm working on this anyway, so lets wait.

mattpap:
I merged those commits into PR #5976, as they are needed there more. I will open another PR with namespace models' implementation.

PR #5547, opened 2016-12-13
fix JS logic for CDS col lengths check

### LINKED ISSUE #5546: JS column length check logic is backward
See discussion in #5048

bryevdv:
- [x] issues: fixes #5546
- [x] tests added / passed


bryevdv:
ping @mattpap

bryevdv:
Not sure why the JS test is failing. I ran it locally and observed it to pass.

bryevdv:
Uh, wow, it passes when I do `gulp test:models` but fails for a full `gulp test`

Issue #4785, opened 2016-07-11
clean-up use of saucelabs connect on travisCI


birdsarah:
currently done in a very manual way that is unecessary.

can use the matrix.include feature

https://github.com/cotsog/travis_ci_prod_regular/blob/36624277882f3e8b5f8f7bc955696f64ac3e8ca1/.travis.yml

https://docs.travis-ci.com/user/customizing-the-build/#Explicity-Including-Builds

PR #5551, opened 2016-12-13
Update travis.yml

birdsarah:
Fixes: #4785 

- use saucelabs add on (as opposed to doing it ourself)
- switch to matrix include (allows for easy targeted use of saucelabs addon)
- clean-up file in general to hopefully make it easier to read


.travis.yml
```diff
@@ -23,9 +23,7 @@ matrix:
       env: GROUP=unit
     - python: 3.5
       env: GROUP=unit
-    # Examples on python 2 and python 3
-    - python: 2.7
-      env: GROUP=examples
+    # Examples on python 3 (not doing python 2 for now due to nb_conda_kernel issue)
```
bryevdv: I don't think that's actually the issue at all but we will sort it out later

bryevdv:
merging this in order to get a green dev build out. Will make new PR to work on getting py2.7 examples working again.

PR #5561, opened 2016-12-14
hotfix for data<=0 in log axes

### LINKED ISSUE #5549: Correctly handle data values <= 0 on a log scale
This is a continuation from issue #5389, partially adressed by PR #5477.  There persists an issue where negative data is not handled correctly.  All data <= 0 should be discarded before generating the plot.

As is, if `values = np.linspace(-0.1, 0.9), a JS error complains that it "could not set initial ranges", probably because `log(n)` for `n<=0` is not defined.

### LINKED ISSUE #5576: initial range calculation for log plots can cause empty plots
PR [5477](https://github.com/bokeh/bokeh/pull/5477) introduced a bug where plots can show up empty if the initial range calculation returns NaN (when data values are <=0 or None). 

Before PR 5477 the code below produced:
image

After PR 5477:
image

```

output_file('initial_log_range')

x = [1, 2, 3]

# fig1
y = [0.1, 0.2, 0.9]
fig1 = figure(y_axis_type='log', plot_width=300, plot_height=300)
fig1.line(x,y)

# fig2
y = [None, 0.2, 0.9]
fig2 = figure(y_axis_type='log', plot_width=300, plot_height=300)
fig2.line(x,y)

# fig3
y = [-10, 0.2, 0.9]
fig3 = figure(y_axis_type='log', plot_width=300, plot_height=300)
fig3.line(x,y)

# fig4
y = [-10, 0.2, 10]
fig4 = figure(y_axis_type='log', plot_width=300, plot_height=300)
fig4.line(x,y)

grid = gridplot([[fig1, fig2],[fig3, fig4]])

show(grid)
```

clairetang6:
Changing the way log ranges compute their `start` and `end` values in PR [5477](https://github.com/bokeh/bokeh/pull/5477) introduced a bug where plot bounds cannot be computed if there are values <= 0 in the data. 

This first commit is a hotfix and results in (check related issue #5549 for test code):
image
for `values = np.linspace(-0.1, 10, 21)`

and
image
for `values = np.linspace(-0.1, 0.9, 21)`

The reason the second one looks worse than the first is because the maximum value is less than the hotfixed minimum value of 1. Line plots where there is a value of 0 in the log axis also show lines pointing to -Infinity. 

I can add on to this PR once I've figured out a better solution (there's a comment in related issue #5549 about what would be necessary), or start another PR if this new bug needs a fast fix. 

EDIT: I created a new issue #5576 that is fixed by this PR. Addressing #5549 will take some more investigation and I won't have that much time in the next couple days. After this PR the plots generated by the test code in #5576 are restored to their original (semi-broken) selves. 

- [x] issues: fixes #5576
- [x] tests added / passed

### DIFF
--- a/bokehjs/test/models/ranges/data_range1d.coffee
+++ b/bokehjs/test/models/ranges/data_range1d.coffee
@@ -7,6 +7,7 @@ p = utils.require "core/properties"
 
 {CustomJS} = utils.require("models/callbacks/customjs")
 {DataRange1d} = utils.require("models/ranges/data_range1d")
+{GlyphRenderer} = utils.require("models/renderers/glyph_renderer")
 
 class TestObject extends HasProps
   type: 'TestObject'
@@ -107,15 +108,13 @@ describe "datarange1d module", ->
     it "should add renderers from one plot", ->
       r = new DataRange1d()
       p = new TestObject()
-      g = new TestObject()
-      g.type = "GlyphRenderer"
+      g = new GlyphRenderer()
       p.renderers = [g]
       r.plots = [p]
       expect(r.computed_renderers()).to.be.deep.equal [g]
 
       r = new DataRange1d()
-      g2 = new TestObject()
-      g2.type = "GlyphRenderer"
+      g2 = new GlyphRenderer()
       p.renderers = [g, g2]
       r.plots = [p]
       expect(r.computed_renderers()).to.be.deep.equal [g, g2]
@@ -124,13 +123,11 @@ describe "datarange1d module", ->
     it "should add renderers from multiple plot", ->
       r = new DataRange1d()
       p = new TestObject()
-      g = new TestObject()
-      g.type = "GlyphRenderer"
+      g = new GlyphRenderer()
       p.renderers = [g]
 
       p2 = new TestObject()
-      g2 = new TestObject()
-      g2.type = "GlyphRenderer"
+      g2 = new GlyphRenderer()
       p2.renderers = [g2]
 
       r.plots = [p, p2]
@@ -139,13 +136,11 @@ describe "datarange1d module", ->
     it "should respect user-set renderers", ->
       r = new DataRange1d()
       p = new TestObject()
-      g = new TestObject()
-      g.type = "GlyphRenderer"
+      g = new GlyphRenderer()
       p.renderers = [g]
 
       p2 = new TestObject()
-      g2 = new TestObject()
-      g2.type = "GlyphRenderer"
+      g2 = new GlyphRenderer()
       p2.renderers = [g2]
 
       r.plots = [p, p2]
@@ -268,8 +263,7 @@ describe "datarange1d module", ->
     it "should update its start and end values", ->
       r = new DataRange1d()
       p = new TestObject()
-      g = new TestObject()
-      g.type = "GlyphRenderer"
+      g = new GlyphRenderer()
       g.id = 1
       p.renderers = [g]
       r.plots = [p]
@@ -285,8 +279,7 @@ describe "datarange1d module", ->
       r = new DataRange1d()
       r._mapper_type = "log"
       p = new TestObject()
-      g = new TestObject()
-      g.type = "GlyphRenderer"
+      g = new GlyphRenderer()
       g.id = 1
       p.renderers = [g]
       r.plots = [p]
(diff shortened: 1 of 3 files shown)


bokehjs/test/models/ranges/data_range1d.coffee
```diff
@@ -263,6 +263,42 @@ describe "datarange1d module", ->
       expect(r._compute_plot_bounds([g1], bds)).to.be.deep.equal {minX: 0, maxX: 10, minY: 5, maxY: 6}
       expect(r._compute_plot_bounds([g1, g2], bds)).to.be.deep.equal {minX: 0, maxX: 15, minY: 5, maxY: 6}
 
+  describe "update", ->
+
+    it "should update its start and end values", ->
+      r = new DataRange1d()
+      p = new TestObject()
+      g = new TestObject()
+      g.type = "GlyphRenderer"
```
clairetang6: Ok, I can fix that today.

bokehjs/test/models/ranges/data_range1d.coffee
```diff
@@ -263,6 +258,40 @@ describe "datarange1d module", ->
...
+      g.id = 1
+      p.renderers = [g]
+      r.plots = [p]
+
+      bds = {
+        1: {minX: -10, maxX: -6, minY: 5, maxY: 6}
+      }
+
+      r.update(bds, 0, 1)
+      expect(r.start).to.be.equal r.start #test will fail when r.start is NaN because NaN != NaN
```
mattpap: Better would be to explicitly test for `NaN` with `expect(r.start).not.to.be.NaN`. The code above seems functionally equivalent, but tells a very different story. This is why you need the comment. Avoid such comments as much as possible and just write code that tell the same story. Just to be clear, I don't discourage you to write comments at all, just this particular class of comments.

clairetang6: I did try the explicit test for `NaN`, but couldn't get the test to fail. Even `expect(NaN).not.to.be.NaN` didn't fail. But I poked around a little and this bug seems to be fixed in an updated version of chai. I had been using chai version 2.14.7, where `expect(NaN).not.to.be.NaN` doesn't fail. After changing the package.json chai dependency to "^3.0.0", it works with chai version 3.5.0.

clairetang6: Should I update the test and update the dependency in `package.json`?

bryevdv:
+1 from me

bokehjs/test/models/ranges/data_range1d.coffee
mattpap: @clairetang6, yes.

bryevdv:
regarding the hotfix setting to 1, what about if `min<0`, setting to `Number.MIN_VALUE` instead of 1? ping @clairetang6 @StevenCHowell

StevenCHowell:
> what about `if min<=0`, setting to `Number.MIN_VALUE` instead of 1?

That should work, iff `Number.MIN_VALUE > 0` (that's the case right?).  What do you think about zooming out a little using `0.9 * Number.MIN_VALUE`?  This would make it so the view does not come immediately to the edge of the data and would match the other defaults a bit better.

bryevdv:
> The `Number.MIN_VALUE` property represents the smallest positive numeric value representable in JavaScript.

I'll make your suggested change then merge this

bryevdv:
Actually I can't find the test code in the related issue, @StevenCHowell can you point me at it?

bryevdv:
Actually that behaves pretty badly, but investigating it can come later. I will leave it at `min=1` for now

StevenCHowell:
> Actually I can't find the test code in the related issue, @StevenCHowell can you point me at it?

Not sure if you still need it but this code from #5389 should work
```
import bokeh
print(bokeh.__version__)

import numpy as np

from bokeh.palettes import Colorblind8 as palette

output_notebook()

values = np.linspace(-1, 1, 21)
print(values)

size = 250
semilogx = figure(title='semilogx test', x_axis_type='log', width=size, height=size)
semilogy = figure(title='semilogy test', y_axis_type='log', width=size, height=size)
loglog = figure(title='loglog test', x_axis_type='log', y_axis_type='log', width=size, height=size)

semilogx.line(values[values != 0.0], values[values != 0.0])
semilogy.line(values, values) 
loglog.line(values, values)

fig = gridplot([[semilogx, semilogy, loglog]])
show(fig)
```

PR #5564, opened 2016-12-14
Deprecate ImageRGBA.{rows,cols}

### LINKED ISSUE #5562: Correctly deprecating ImageRGBA cols and rows properties
As part of #5544 we are changing how 2D arrays are serialized in bokeh and in reworking that code @bryevdv noticed that unlike the ``Image`` glyph,  the ``ImageRGBA`` glyph supports setting  ``rows`` and ``cols`` properties and setting the image data to a flattened 1D-array. This is both inconsistent and a bit weird so it was suggested that this option should be deprecated.

The question now is how best to do this deprecation. Most of the deprecations currently either rename or move an existing property rather than deprecating it entirely so there is no standard approach here. Currently I see two options:

1) My initial suggestion was to add hidden ``_rows`` and ``_cols`` props which matching rows and cols setter and getter properties write into.

2) As an alternative @bryevdv suggested defining a ``DeprecatedNumberSpec`` property
which would emit a warning on validation.

Neither approach is great, however currently when I try to implement the second approach I noticed that the ``validate`` method is fired twice when supplying the property value to the constructor, causing the deprecation warning to be duplicated. Suggestions welcome on how to correctly perform this deprecation, it would also good to figure out why the value is being set and validated twice in the constructor.

mattpap:
fixes #5562

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -52,7 +52,6 @@ from __future__ import absolute_import, print_function
 import logging
 logger = logging.getLogger(__name__)
 
-import warnings
 import collections
 from copy import copy
 import datetime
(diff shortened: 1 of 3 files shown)


bryevdv:
I think I'd prefer to leave the actual usage (i.e. the second commit) out of this PR (it would be better in the other PR) but for the sake of expediency I will merge as it is.

PR #5565, opened 2016-12-14
Try sudo=true for travis ci problem

bryevdv:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 sudo: 
-  false
+  true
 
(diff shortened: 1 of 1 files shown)


bryevdv:
`sudo: true works` but adds maybe 10 minutes to the build

bryevdv:
Well, I restore `sudo: false` and added `ps` at the end to log processes to try and see what might be hanging around. ...and not it works. I'm inclined to merge as-is and leave the `ps` just in case its needed for debugging in the future.

PR #5567, opened 2016-12-14
explicitly kill stray processes on travis

bryevdv:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,9 @@ matrix:
       env: GROUP=unit
     - python: 3.5
       env: GROUP=unit
-    # Examples on python 3 (not doing python 2 for now as travis never exits)
+    # Examples tests
+    - python: 2.7
+      env: GROUP=examples
     - python: 3.4
       env: GROUP=examples
     # Linting and Docs (no python3 version of fabric)
@@ -96,6 +98,7 @@ script:
     if [[ "$GROUP" == examples ]]; then 
       echo "https://s3.amazonaws.com/bokeh-travis/$TRAVIS_JOB_ID/examples.html";
     fi; 
+    ps -ef;
     echo "********** END POOR MAN LOGGER ************"
 
 after_success: |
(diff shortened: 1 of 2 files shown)


birdsarah:
Regarding my comment earlier about runnign a subset of examples. See:

https://github.com/bokeh/bokeh/blob/master/tests/examples/collect_examples.py#L89

You could have all those methods only return a slice of the list.

bryevdv:
NOT Working:

screen shot 2016-12-14 at 5 04 47 pm

Working:

screen shot 2016-12-14 at 5 04 59 pm

The only difference seems to be in the *number* of jupyter related things

bryevdv:
well, OK it was the conda nb changes somehow after all

bryevdv:
worked with pinned notebook version but not conda nb changes. Putting back conda nb change to confirm thats  the problem.

ping @damianavila it appears that this change to disable the conda nb extension is breaking our build by making something not clean up properly.

bryevdv:
Yup, adding the line back to remove the conda nb extensions breaks the build

bryevdv:
So one option would be to just disable all notebook checks for now. If we revert the conda nb uninstall, then the nb tests are just worthless, so not point in doing that. I'd rather leave the conda nb uninstall, and just disable the notebook tests until there is a solution. Or we could try some kind of `killall` at the end of the script? open to suggestions @bokeh/dev

bryevdv:
@bokeh/dev @damianavila adding `killall python` at the end seems to allow a successful termination:

https://travis-ci.org/bokeh/bokeh/jobs/184349786

This seems like a pretty crappy bandaid, but is it an acceptable one for now?

damianavila:
I proposed this one: https://github.com/bokeh/bokeh/pull/5574, let see how it goes... btw, this is really weird...

bryevdv:
#5574 did not solve the issue, I propose to merge this for now.

PR #5569, opened 2016-12-15
Upgrade jsdom

### LINKED ISSUE #5568: Upgrade to jsdom 9.x
Currently we are on 5.x which doesn't even support `HTMLElement.classList`. Upgrade isn't very straightforward, because of some JavaScript magic. Apparently we were configuring jsdom all wrong (by copying `window`'s properties to `global`), so now we have to figure out how to configure it properly (using envs).

mattpap:
fixes #5568


bokehjs/test/index.coffee
```diff
@@ -30,3 +30,7 @@ blacklist.push('constructor')
 for own key, val of global.window
   if blacklist.indexOf(key) == -1
     global[key] = val
+
+global.HTMLElement = global.window.HTMLElement
+global.Event = global.window.Event
+global.Image = global.window.Image
```
mattpap: Apparently we were doing this wrong all the time. Those new lines are just to make node/jsdom happy, because `HTMLElement` at al. are special and they don't enumerate properly (as I would expect, but no worries, similar issue arises in web browsers). We should follow https://github.com/tmpvar/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global. Unfortunately this is incompatible with how we use mocha. That wiki expects us to organize our code as `mocha -> jsdom`, but we need `jsdom -> mocha`. Perhaps we should get rid of `gulp-mocha` and run mocha manually from within a jsdom env. I don't know. Tests work in this PR, so may proceed with this hack.

mattpap: The last example from that wiki may work for us, but we would need to manually compile scripts (coffeescript) to make it work. Now it's all automated given we use `require()` and on-the-fly compilation.

mattpap: Also, given how things are getting more complicated, I would just run tests in a headless web browser.

mattpap:
Changing status to ready, because I have no idea how to proceed with proper fix. This will have to be resolved at some point.

bryevdv:
> Changing status to ready, because I have no idea how to proceed with proper fix. This will have to be resolved at some point.

If that's the case what is the immediate impetus to change (not saying there isn't one, but I don't understand what current priority depends on this change)

mattpap:
>  but I don't understand what current priority depends on this change)

We use very outdated jsdom, which makes our tests fail whenever I try to use native DOM APIs instead of jquery. And it's annoying switch between different versions of dependencies back and forth (given that npm isn't a speed demon).

bryevdv:
OK well I can't claim enough expertise to properly evaluate all of the considerations here, if you think it is important (and ready for now) please merge.

PR #5533, opened 2016-12-10
Deprecate Bokeh.$ and Bokeh._

mattpap:
fixes #5532

PR #5571, opened 2016-12-15
Don't use FINALIZE_JS if hide_banner is True

### LINKED ISSUE #5570: output_notebook raises javascript error if hide_banner=True
Using latest master the ``output_notebook`` function raises a Javascript error when ``hide_banner=True``, i.e. in a fresh notebook:

```python
output_notebook(hide_banner=True)
```

```
VM9236:92 Uncaught TypeError: Cannot set property 'textContent' of null(…)
```

This originates in the following embedded function:

```javascript
    function(Bokeh) {
      
      document.getElementById("759c6b76-37ad-4246-bc56-d762de3036d9").textContent = "BokehJS is loading...";
    },
```

When additionally setting ``resources=INLINE`` the JS code fails to get executed in the notebook entirely:

```python
output_notebook(hide_banner=True, resources=INLINE)
```

```javascript
TypeError: Cannot set property 'textContent' of null
```

mattpap:
This was revealed after PR #5533. Switching from jquery to less forgiving APIs is definitively step in right direction.

fixes #5570

### DIFF
--- a/bokeh/util/notebook.py
+++ b/bokeh/util/notebook.py
@@ -79,7 +79,7 @@ def _load_notebook_html(resources=None, verbose=False, hide_banner=False,
         elementid = '' if hide_banner else element_id,
         js_urls  = resources.js_files,
         css_urls = resources.css_files,
-        js_raw   = resources.js_raw + [FINALIZE_JS % element_id],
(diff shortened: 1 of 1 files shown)


bryevdv:
LGTM, tested locally

PR #5573, opened 2016-12-15
Make custom models work in Notebook again

### LINKED ISSUE #5227: External resources belonging to custom models fail to load inside of Jupyter/IPython notebooks
The [LatexLabel example](http://bokeh.pydata.org/en/latest/docs/user_guide/extensions_gallery/latex.html) is a great resource but I'm observing problems using it with `output_notebook()`: plots are rendered with essentially no height until the browser window is resized. Removing the `LatexLabel` is all it takes to restore normality. This problem does not appear for `output_file()`.

Here is the output produced by `show()`...

2016_09_23_18_06_15_greenshot

...and it's 'fixed' as soon as I change the window dimensions slightly.

2016_09_23_18_07_07_greenshot

Using bokeh 0.12.2 with Anaconda 4.1.0 (Py35) on Win7 x64 + Chromium 53.0

xref #647

### LINKED ISSUE #5591: Custom models don't work in the notebook due to missing __file__
Another incarnation of our struggle with resolving custom models' module paths.

mattpap:
fixes #5591

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -4,11 +4,12 @@ import logging
 logger = logging.getLogger(__name__)
 
 import io
+import os
+import sys
 import six
 import json
-import inspect
 import hashlib
-from os.path import dirname, join, abspath, exists
+from os.path import dirname, join, abspath, exists, isabs
 from subprocess import Popen, PIPE
 
 from ..model import Model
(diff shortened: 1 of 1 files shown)


mattpap:
Lets merge the first part of it, so that we have at least custom models working in the notebook. Currently I don't know how to properly fix #5227. I have some ideas to hack around this issue, but we may need to completely rethink how we manage external resources.

bryevdv:
@mattpap If this does not fix all of #5227 can you make a new separate issue for this PR to track then?

mattpap:
@bryevdv, done.

Issue #5541, opened 2016-12-12
Label image diff results "expected" vs "actual"


bryevdv:
I have no idea whether this is an old problem being fixed or a new problem showing up:

<img width="658" alt="screen shot 2016-12-12 at 11 12 16 am" src="https://cloud.githubusercontent.com/assets/1078448/21108939/926b733e-c05c-11e6-9ce3-7e05ef65d542.png">

My mind will 100% never be able to remember this arbitrary convention without some kind of labeling, because there is not particular reason it should be one way or the other. 

@mattpap can you say 1) which way it actually is, and 2) where to look to add a label to the output?

mattpap:
It's "current", "diff" and "reference". This should have been labeled for a long time, but there is simple logic behind this. The "current" image always exists, so it's first and is followed by "optional" images. Like with function arguments, optional come last. "diff" vs. "reference" may seem arbitrary though.

mattpap:
> "expected" vs "actual"

I would prefer we didn't use "expected" to refer to "reference" images, because this implies that there exists a set of expected outcomes and this is utterly not true. The reason is that we can't decide computationally (at least within the framework and test model we use) whether the difference is an intentional change or not. Saying "expected" implies that every difference is a bug, though most of the time it's an intentional change.

bryevdv:
That's fine, I just care about being able to know which is which.

bryevdv:
@mattpap In this report:

https://s3.amazonaws.com/bokeh-travis/183601852/examples.html

Why are these reported as "no diff" ?

<img width="623" alt="screen shot 2016-12-13 at 9 47 49 am" src="https://cloud.githubusercontent.com/assets/1078448/21147050/4ffa511a-c119-11e6-94cc-b133ee6335e5.png">

mattpap:
@bryevdv, I'm not sure, but the original examples' testing code didn't allow to proceed to image diff in presence of JavaScript errors. You would get "FAILED" example instead and JavaScript error message in the log. This isn't true anymore after changes early this year, so this may be the effect. Note also that currently there is no way to learn about JavaScript errors in Travis CI's log. One has to go through images for this purpose (as shown in your comment).

damianavila:
Well... I am really note sure the left is current... see this PR where I remove nb_conda_kernels an other extensions... when I see the report: https://s3.amazonaws.com/bokeh-travis/184359356/examples.html

[...]

PR #5574, opened 2016-12-15
Remove nbextensions

damianavila:
Alternative to #5567, let see how it goes...


scripts/travis_install
```diff
@@ -58,7 +58,9 @@ conda install --yes $DEPS_RUN
 conda install --yes $DEPS_TEST
 
 # conda nb extension is installed by default and breaks notebook image diff tests
-python -m nb_conda_kernels.install --disable --prefix=/home/travis/miniconda
+# python -m nb_conda_kernels.install --disable --prefix=/home/travis/miniconda
+# remove all the nbextension/serverextensions
+conda uninstall --yes _nb_ext_conf
```
bryevdv: FWIW when I conda removed the nb extension locally it made my notebook unusable (would not start)

damianavila: mmm... that should not happen... any traceback to know why was not working?

damianavila:
It passed... btw, see my comment here: https://github.com/bokeh/bokeh/issues/5541#issuecomment-267449718

bryevdv:
@damianavila you will have to add back the py 2.7 examples tests to `.travis.yml` they were removed to gets tests green.

bryevdv:
It passed because the failing test was not run :)

bryevdv:
You an see how they were added back to the matrix in my other PR

damianavila:
> You an see how they were added back to the matrix in my other PR

OK, added now... let's see...

damianavila:
mmm... nb_conda_kernels was installed from conda-forge after removal... mmm...

bryevdv:
Nope, same problem. Build just "finished" but is not reporting done/green

bryevdv:
This seems like a problem with the extension or with jupyter or both

damianavila:
>This seems like a problem with the extension or with jupyter or both

It could not be the extension... since it is not there... it seems to be some issue with jupyter itself... but only in python 2... interesting.

bryevdv:
Well it was installed at one point so can't rule out that it could be something about it in particular having had been installed

PR #5577, opened 2016-12-16
Return stable value from ColumnDataSource.get_length()

### LINKED ISSUE #525: ColumnDataSource.prototype.get_length gives arbitrary results
The implementation is as follows:

``` coffeescript
get_length: () ->
      data = @get('data')
      return data[_.keys(data)[0]].length
```

Due to arbitrary zero indexing, this not only can give unexpected results if one data source is shared for multiple purpose, but is also can give different results depending on the platform.

mattpap:
fixes #525


bryevdv:
Sure, LGTM

PR #5580, opened 2016-12-16
Adding size attribute to MultiSelect model

### LINKED ISSUE #5579: Adding size attribute to MultiSelect model
Small feature request: it would be nice to have an attribute to toggle number of visible elements in MultiSelect. Note that this can't exactly be controlled by CSS (the height property can be used, but I'm not sure if there's an easy way to set it to exactly match the height of the options). Since the options can be dynamically set, it seems reasonable to me to also let the size attribute be dynamically set.

This is implemented in these commits: [commit1](https://github.com/azjps/bokeh/commit/504910486751c5e947edd9217f186ede87f82b48), [commit2 which is an unrelated typo fix](https://github.com/azjps/bokeh/commit/0ff1290210af58b59cbecbcd84e98dc6d1248349) I haven't submitted a PR yet, noticed that PR etiquette dictates to submit an issue first.

azjps:
Closes #5579

* Adds size attribute to MultiSelect model
* Small note in docs about BOKEH_MINIFIED env var, as it took me a while to figure out how to directly use the unminified JS.

### DIFF
--- a/bokehjs/src/coffee/models/widgets/multiselect.coffee
+++ b/bokehjs/src/coffee/models/widgets/multiselect.coffee
@@ -39,15 +39,25 @@ export class MultiSelectView extends InputWidgetView
       if values[el.attr('value')]
         el.attr('selected', 'selected')
     )
+    # Note that some browser implementations might not reduce
+    # the number of visible options for size <= 3.
     @$el.find('select').attr('size', this.model.size)
 
   change_input: () ->
+    # Haven't checked if :focus selector works for IE <= 7
+    is_focused = @$el.find('select:focus').size()
     value = @$el.find('select').val()
     if value
       @model.value = value
     else
       @model.value = []
     super()
+    # Restore focus back to the <select> afterwards,
+    # so that even if python on_change callback is invoked,
+    # focus remains on <select> and one can seamlessly scroll
+    # up/down.
+    if is_focused
+      @$el.find('select').focus()
 
 export class MultiSelect extends InputWidget
   type: "MultiSelect"
--- a/bokehjs/src/coffee/models/widgets/multiselect.coffee
+++ b/bokehjs/src/coffee/models/widgets/multiselect.coffee
@@ -20,6 +20,7 @@ export class MultiSelectView extends InputWidgetView
     @listenTo(@model, 'change:options', @render)
     @listenTo(@model, 'change:name', @render)
     @listenTo(@model, 'change:title', @render)
+    @listenTo(@model, 'change:size', @render)
 
   render: () ->
     super()
@@ -38,6 +39,7 @@ export class MultiSelectView extends InputWidgetView
       if values[el.attr('value')]
         el.attr('selected', 'selected')
     )
+    @$el.find('select').attr('size', this.model.size)
 
   change_input: () ->
     value = @$el.find('select').val()
@@ -54,4 +56,5 @@ export class MultiSelect extends InputWidget
   @define {
       value:   [ p.Array, [] ]
       options: [ p.Array, [] ]
+      size:    [ p.Number, 4 ]  # 4 is the HTML default
     }
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -101,7 +101,9 @@ class MultiSelect(InputWidget):
     """)
 
     size = Int(default=4, help="""
-    The number of visible options in the dropdown list.
+    The number of visible options in the dropdown list. (This uses the
+    ``select`` HTML element's ``size`` attribute. Some browsers might not
+    show less than 3 options.)
     """)
 
 class Slider(InputWidget):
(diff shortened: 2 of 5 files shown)


bryevdv:
@azjps this is a nice PR thank you very much! I only have two small comments: 

* there does seem to be some *minimum* size that is always shown (around 3 rows), regardless of what is supplied by the user. Not sure if this warrants a mention, just noting it.

* There is an example that could be updated to demonstrate the use of this property, which I think would be nice. Here is a one-line diff:
    ```
    diff --git a/examples/models/widgets.py b/examples/models/widgets.py
    index 363de99..3bf77e8 100644 
    --- a/examples/models/widgets.py 
    +++ b/examples/models/widgets.py
    @@ -45,7 +45,7 @@ autocomplete_input = AutocompleteInput()

     select = Select(options=["Option 1", "Option 2", "Option 3"])

    -multi_select = MultiSelect(options=["Option 1", "Option 2", "Option 3"])
    +multi_select = MultiSelect(options=["Option %d" % (i+1) for i in range(16)], size=6)

     slider = Slider(value=10, start=0, end=100)
    ```

bryevdv:
@azjps just a quick ping, do you think you will have the opportunity to update that example? If not I think GH would now allow me to do it, if that's OK with you. That's my only comment for this PR

azjps:
Thanks for the kind words,

- I believe this may be browser-specific, there's a note that [in Chrome and Safari, this attribute may not work as expected for size="2" and size="3"](http://www.w3schools.com/TAgs/att_select_size.asp). On Chrome v55 in linux it works correctly for me for both though. I can add a quick note in docstring.
- Sure. I was hoping to add this [example](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/B7i0rdbhI-8) too now that I figured some workaround to get it to work -- registering an on_change callback to Select.value causes the html <select> to lose focus after each change, so one wouldn't be able to scroll up/down in the <select>. Having a js callback reset the focus to the <select> will allow scroll up/down to work. Does that seem like an appropriate way of going about such (and if so mind if I add it as an example)?

bryevdv:
> registering an on_change callback to Select.value causes the html <select> to lose focus after each change

This almost seems like a bug to fix? Should the multi-select JS implementation just do this always, automatically?

azjps:
Ah, I hadn't realized it could potentially be very easy to fix in bokehjs itself. (I don't know anything about how/when the `on_change` callbacks work -- I'd figured the reason the `<select>` was losing focus was that the `on_change` callback re-draws the entire DOM as part of some separate async callback). I'll tack that onto this PR.

bokehjs/src/coffee/models/widgets/multiselect.coffee
```diff
@@ -39,15 +39,25 @@ export class MultiSelectView extends InputWidgetView
...
       @model.value = value
     else
       @model.value = []
     super()
+    # Restore focus back to the <select> afterwards,
+    # so that even if python on_change callback is invoked,
+    # focus remains on <select> and one can seamlessly scroll
+    # up/down.
+    if is_focused
+      @$el.find('select').focus()
```
bryevdv: @mattpap can you comment, does the syntax for `@$el` need to change?

mattpap: No this is fine, for now.

bryevdv: any thoughts on the actual implementation to keep focus? If not I will merge tomorrow.

bokehjs/src/coffee/models/widgets/multiselect.coffee
```diff
@@ -38,14 +39,25 @@ export class MultiSelectView extends InputWidgetView
       if values[el.attr('value')]
         el.attr('selected', 'selected')
     )
+    # Note that some browser implementations might not reduce
+    # the number of visible options for size <= 3.
+    @$el.find('select').attr('size', this.model.size)
 
   change_input: () ->
+    # Haven't checked if :focus selector works for IE <= 7
+    is_focused = @$el.find('select:focus').size()
```
mattpap: More idiomatic approach would be `is_focused = @$el.find('select').is(':focus')`.

bryevdv:
Thanks @azjps !

bokehjs/src/coffee/models/widgets/multiselect.coffee
azjps: Yeah, I assume there is a better way to do this. One (albeit more minor) problem that continues with this implementation is that when scrolling down a `<select>` with a lot of options, when the focus drops off the `<select>` it scrolls back to the top. When this implementation causes the `<select>` to regain focus, it will have been scrolled so that the selected elements are on the bottom of the viewable options. So the viewable options may change and there's also the distraction of seeing the options and scrollbar flicker.

bryevdv: @azjps can you open new issue(s) for any remaining tasks or improvements?

Issue #5585, opened 2016-12-19
Responsive plots don't work with server because of plotdiv


mattpap:
As always if there are two ways of doing the same thing, one won't work. In embed, the target element (aka. `plotdiv`) is either replaced or appended. If replaced, responsive plots do work (e.g. in file output). I discovered this when working on issue #3674.

PR #5586, opened 2016-12-19
Add support for readonly properties

### LINKED ISSUE #5199: Add support bokehjs writable and bokeh readonly properties
There are certain properties in bokehjs like inner plot area that are computed on client-side and it doesn't make sense to allow to configure them in Python, yet it's valuable to get their values in Python.

mattpap:
fixes #5199, #5585

### DIFF
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -510,6 +510,10 @@ class Plot(LayoutDOM):
     the actual plot, without toolbars etc. Note this is computed in a
     web browser, so this property will work only in backends capable of
     bidirectional communication (server, notebook).
+
+    .. note::
+        This is an experimental feature and the API may change in near future.
+
     """)
 
     inner_height = Int(readonly=True, help="""
@@ -517,6 +521,10 @@ class Plot(LayoutDOM):
     the actual plot, without toolbars etc. Note this is computed in a
     web browser, so this property will work only in backends capable of
     bidirectional communication (server, notebook).
+
+    .. note::
+        This is an experimental feature and the API may change in near future.
+
     """)
 
     background_props = Include(FillProps, help="""
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -505,6 +505,20 @@ class Plot(LayoutDOM):
 
     """)
 
+    inner_width = Int(readonly=True, help="""
+    This is the exact width of the plotting canvas, i.e. the width of
+    the actual plot, without toolbars etc. Note this is computed in a
+    web browser, so this property will work only in backends capable of
+    bidirectional communication (server, notebook).
+    """)
+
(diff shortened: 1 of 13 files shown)


bokeh/models/plots.py
```diff
@@ -505,6 +505,20 @@ def _check_snapped_toolbar_and_axis(self):
...
+    bidirectional communication (server, notebook).
+    """)
+
+    inner_height = Int(readonly=True, help="""
+    This is the exact height of the plotting canvas, i.e. the height of
+    the actual plot, without toolbars etc. Note this is computed in a
+    web browser, so this property will work only in backends capable of
+    bidirectional communication (server, notebook).
+    """)
+
```
bryevdv: This seems OK but possibly worth considering: what about just making the entire frame object available as a read-only property with read-only properties on it? This could provide much richer and detailed information, like the exact gutter sizes.

mattpap: The original request was about frame's width and height. Also, frame seems to make like an implementation detail. I will see what can be done.

mattpap: OK.

bryevdv: If it's a quick change to expose frame directly, great, otherwise I'm finally be to proceed with this as a clear improvement

mattpap:
```
GMapPlot: python is missing some properties found in CoffeeScript
    GMapPlot.inner_width: coffee defaults to null but python has no such property
    GMapPlot.inner_height: coffee defaults to null but python has no such property
```
This is likely because "serializable" status of a property is used for what should be "enumerable". Unfortunately we don't have such distinction in our properties system.

bokeh/models/plots.py
mattpap: @bryevdv, so the problem is that `Frame.{width,height}` come from `LayoutCanvas` and are getters of `_width` and `_height` layout variables. They would have to be changed to properties, but I'm not sure of implications of this. Most likely there will be none, besides that those properties should be immutable outside layout engine.

bryevdv: OK, we can hold off, and just do this for now. Might be worth a note that these might be available via frame at some future date

mattpap: @bryevdv, I added a note to both properties that they are experimental.

bryevdv:
@mattpap is this ready ready, or are you still working on it?

mattpap:
@bryevdv, ready.

bryevdv:
FYI ping @jbednar @philippjfr this will be in `0.12.4` and should be useful for datashader / holoviews to get the inner plot dimensions

bryevdv:
Looks good, tested locally. I'd like to add some bokeh sphinx support for automatically noting readonly props but that can happen separately.

PR #5600, opened 2016-12-21
Custom context arguments for the jinja template

### LINKED ISSUE #5583: Custom context arguments for the jinja template
I have a bokeh server app with custom `templates/index.hml`, and I'd like to pass custom arguments to the `render()` method of jinja. I see that I can access the jinja template with `curdoc().template`, but the [doc](http://jinja.pocoo.org/docs/dev/api/#jinja2.Template) says this is an immutable object.

`embed._html_page_for_render_items` already has an argument `template_variables` but it seems to be unused for the server. Can we make this available to the server app?

bkueng:
With this you can use something like:
```
curdoc().template_variables['myvariable'] = '42'
```
And add `{{ myvariable }}` to the `index.html`.

I wasn't quite sure if `Document.template_variables` needs a setter as well. Thoughts?

fixes #5583

### DIFF
--- a/sphinx/source/docs/user_guide/server.rst
+++ b/sphinx/source/docs/user_guide/server.rst
@@ -241,8 +241,7 @@ The optional components are
 
 * A ``templates`` subdirectory with ``index.html`` Jinja template file. The directory may contain additional Jinja templates for ``index.html`` to refer to. The template should have the same parameters as the :class:`~bokeh.core.templates.FILE` template.
 
-Custom variables can be passed to the template via the ``curdoc().template_variables`` 
-dictionary in place:
+Custom variables can be passed to the template via the ``curdoc().template_variables`` dictionary in place:
 
 .. code-block:: python
 
--- a/sphinx/source/docs/user_guide/server.rst
+++ b/sphinx/source/docs/user_guide/server.rst
@@ -240,7 +240,17 @@ The optional components are
 * A ``theme.yaml`` file that declaratively defines default attributes to be applied to Bokeh model types.
 
 * A ``templates`` subdirectory with ``index.html`` Jinja template file. The directory may contain additional Jinja templates for ``index.html`` to refer to. The template should have the same parameters as the :class:`~bokeh.core.templates.FILE` template.
-Custom variables can be passed to the template via the ``curdoc().template_variables`` dictionary.
+
+Custom variables can be passed to the template via the ``curdoc().template_variables`` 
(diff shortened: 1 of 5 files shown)


sphinx/source/docs/user_guide/server.rst
```diff
@@ -240,6 +240,7 @@ The optional components are
 * A ``theme.yaml`` file that declaratively defines default attributes to be applied to Bokeh model types.
 
 * A ``templates`` subdirectory with ``index.html`` Jinja template file. The directory may contain additional Jinja templates for ``index.html`` to refer to. The template should have the same parameters as the :class:`~bokeh.core.templates.FILE` template.
+Custom variables can be passed to the template via the ``curdoc().template_variables`` dictionary.
```
bryevdv: Only suggestion here would be to demonstrate explicitly:

```
Custom variables can be passed to the template via the ``curdoc().template_variables`` 
dictionary in place:

.. code-block:: python

    # set a new single key/value
    curdoc().template_variables["user_id"] = user_id

    # or update multiple at once
    curdoc().template_variables.update(first_name="Mary", last_name="Jones")
```

bryevdv:
@bkueng this is great (and fast!) I only have one small comment about the docs, and then my only other suggestion might be to consider updating one of the existing app examples to demonstrate this (or if you have an interesting app example that you could share, to add one).

bryevdv:
Actually, getting slightly ahead. We should also add a couple of tests around this as well. I'm happy to offer guidance around that if need be.

bryevdv:
There's also one error showing up in our test automation: 
```
/home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/server.rst:243: 
WARNING: Bullet list ends without a blank line; unexpected unindent.
```

Just need to add a blank line before line 243 to fix

bryevdv:
@bkueng BTW if you don't have bandwidth to continue working on this, I understand. Let me know, and I think I can make the small changes on the PR branch directly, as a maintainer.

bkueng:
@bryevdv Thanks for your helpful comments, I completely agree with them. Sorry I had a lot to do today, and would only be able to continue on this tomorrow. So feel free to make the necessary changes, it's much appreciated.

bryevdv:
OK after a few typos (editing on the web is difficult!) everything looks good. Thanks for the PR @bkueng !

bkueng:
Cool, thanks!

PR #5603, opened 2016-12-22
Fix autoload templates to inject data into JS templates properly

### LINKED ISSUE #5558: unicode `__javascript__` external resources breaks output notebook in python 2
Defining `__javascript__` using a list of unicode strings, causes a Javascript error upon calling `output_notebook`:

```
from __future__ import unicode_literals, print_function
import sys
print(sys.version)

from bokeh.models import Model

class TestCustom(Model):
    __javascript__ = ["https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js",
                      "https://cdnjs.cloudflare.com/ajax/libs/ion-rangeslider/2.1.4/js/ion.rangeSlider.js"]

    __implementation__ = """
    p = require "core/properties"
    BokehModel = require "model"

    class TestCustom extends BokehModel
        type: "TestCustom"

    module.exports = 
        Model: TestCustom
    
    """

output_notebook()
```

The cell output is:
```
2.7.12 |Anaconda 4.2.0 (x86_64)| (default, Jul  2 2016, 17:43:17) 
[GCC 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)]
Loading BokehJS ...
SyntaxError: Unexpected string
```

And the console error message is:
```
SyntaxError: Unexpected string
    at OutputArea.append_javascript (outputarea.js:625)
    at OutputArea.append_mime_type (outputarea.js:569)
    at OutputArea.append_display_data (outputarea.js:533)
    at OutputArea.append_output (outputarea.js:319)
    at OutputArea.handle_output (outputarea.js:233)
    at output (codecell.js:356)
    at Kernel._handle_output_message (kernel.js:1060)
    at i (jquery.min.js:4)
    at Kernel._handle_iopub_message (kernel.js:1087)
    at Kernel._finish_ws_message (kernel.js:924)
```

* The Javascript error goes away using a byte string `b"https://cdnjs.cloudflare.com..."`, but then the same error pops up in Python 3.
* The error only happens with `__javascript__` not `__css__`.

Bokeh version: 0.12.3
Jupyter version: 1.0.0
Notebook version: 4.2.3

mattpap:
Previously we relied on Python's `str()` returning valid JS. Sometimes it actually and very deceptively does.

fixes #5558

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -29,7 +29,7 @@ calls it with the rendered model.
     return new Date();
   }
 
-  var force = {{ force|json }};
+  var force = {{ force|default(False)|json }};
 
   if (typeof (window._bokeh_onload_callbacks) === "undefined" || force === true) {
     window._bokeh_onload_callbacks = [];
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -29,9 +29,9 @@ calls it with the rendered model.
     return new Date();
   }
 
-  var force = "{{ force }}";
+  var force = {{ force|json }};
 
-  if (typeof (window._bokeh_onload_callbacks) === "undefined" || force !== "") {
+  if (typeof (window._bokeh_onload_callbacks) === "undefined" || force === true) {
     window._bokeh_onload_callbacks = [];
     window._bokeh_is_loading = undefined;
   }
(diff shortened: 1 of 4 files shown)


bryevdv:
@mattpap is there a test or demonstration that confirms this fixes #5558 ?

mattpap:
@bryevdv, the code from #5558 must work in the notebook without errors. As to testing this automatically, it's possible yet annoying, because we would have to test this big blob of text, which is far from robust.

bryevdv:
Works for me after testing (a screenshot of output here would have saved me the trouble tho)

PR #5605, opened 2016-12-22
Don't inherit __future__ flags in compile()

### LINKED ISSUE #5590: Python 2 incompatibility issue with execfile and bokeh server
I have a bokeh server dashboard written in Python 2.  It includes the use of `print` as a statement.  When run by itself via `bokeh serve --show step3.py`, it runs just fine.

I then write a second script `step4.py` which uses `execfile()` on the first script, and then adds some additional logic of its own.  Now I encounter a very strange problem:

```
2016-12-20 07:52:02,473 Error running application handler <bokeh.application.handlers.script.ScriptHandler object at 0x10bb20f50>: invalid syntax (step3.py, line 41)
File "step4.py", line 3, in <module>:
execfile("step3.py") Traceback (most recent call last):
  File "/Users/pwang/anaconda/envs/bokehdemo/lib/python2.7/site-packages/bokeh/application/handlers/code_runner.py", line 83, in run
  File "/Users/pwang/src/bokeh.demo/webinar/demo1/step4.py", line 3, in <module>
    execfile("step3.py")
  File "step3.py", line 41
    print attribute, "changed from '%s' to '%s'" % (old, new)
                  ^
SyntaxError: invalid syntax
```

I can verify that this is a `print` statement-related SyntaxError because if I put parentheses around this line, everything works fine.  An additional interesting point of information is that in `step4.py` itself, I can put a `print` statement *before* the `execfile("step3.py")` call, and it works just fine.

So somehow, the logic in `code_runner` isn't accounting for `execfile()` behavior...?  Or perhaps the import of `print_function` at the top of `bokeh.application.handlers.code_runner` is screwing things up?

I'll attach the relevant files later (they're being worked on right now).

mattpap:
Apparently `__future__` flags (imports) from the parent environment (so `code_runner.py`) are forwarded to `compile()`. There exists `dont_inherit` boolean argument that doesn't allow that. I tested this manually, but we may want to write tests for this some how. It's sufficient to test for `exec("print 'XXX'")`, but I'm not sure what to do about Python 3. 

fixes #5590

### DIFF
--- a/bokeh/application/handlers/tests/test_code.py
+++ b/bokeh/application/handlers/tests/test_code.py
@@ -5,6 +5,8 @@ import unittest
 from bokeh.application.handlers import CodeHandler
 from bokeh.document import Document
 
+from bokeh.util.testing import skipIfPy3
+
 script_adds_two_roots = """
 from bokeh.io import curdoc
 from bokeh.model import Model
@@ -26,6 +28,16 @@ class TestCodeHandler(unittest.TestCase):
     def test_empty_script(self):
         doc = Document()
         handler = CodeHandler(source="# This script does nothing", filename="path/to/test_filename")
+        handler.modify_document(doc)
+        if handler.failed:
+            raise RuntimeError(handler.error)
+
+        assert not doc.roots
+
+    @skipIfPy3("this test doesn't have a Python 3 equivalent")
+    def test_exec_and___future___flags(self):
+        doc = Document()
+        handler = CodeHandler(source="exec(\"print \\\"XXX\\\"\")", filename="path/to/test_filename")
         handler.modify_document(doc)
         if handler.failed:
             raise RuntimeError(handler.error)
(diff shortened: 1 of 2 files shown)


bryevdv:
I think barring a reason not to, we can assume the standard lib features function as advertised. A single py2 only test for `exec("print 'XXX'")` seems sufficient to guarantee that we are actually using the feature as we should.

bokeh/application/handlers/tests/test_code.py
```diff
@@ -32,6 +34,16 @@ def test_empty_script(self):
 
         assert not doc.roots
 
+    @skipIfPy3("this test doesn't have a Python 3 equivalent")
+    def test_exec_and___future___flags(self):
+        doc = Document()
+        handler = CodeHandler(source="exec(\"print \\\"XXX\\\"\")", filename="path/to/test_filename")
```
bryevdv: Couldn't this be written `source='exec("print \'xxx\'")'` ?

mattpap: It can, I just prefer it that way.

bryevdv: fair enough, will merge as-is then

PR #5612, opened 2016-12-26
Columndatasource autoconvert

### LINKED ISSUE #5578: Auto-conversion to ColumnDataSource
Many times, we want to pass a data structure to a plotting function as a data source. To do so, an explicit conversion is required. Often, the converted object does not serve any purpose, except for the conversion itself.  The modification to handle various supported data structures, such as pandas DataFrame and dicts, is pretty trivial, as you can see [here][1]. Is this a good idea to perform auto-conversion? What are the potential hazards of the way I handled this?

[1]: https://github.com/bgbg/bokeh/commit/aa459a7a89b715c3c8fea9a8cc8a7a5fad5cae9a

bgbg:
Many times, we want to pass a data structure to a plotting function as
a data source. To do so, an explicit conversion is required. Often, the
converted object does not serve any purpose, except for the conversion
itself. This commit adds auto-conversion to reduce boilerplate code.

Reference conversation: https://github.com/bokeh/bokeh/issues/5578

- fixes #5578 
- 2 tests added and  passed


bokeh/plotting/tests/test_figure.py
```diff
@@ -161,6 +161,19 @@ def test_plot_fill_props(self):
...
+        dct = {'x': [1, 2, 3], 'y': [2, 3, 4]}
+        p.circle(x='x', y='y', source=dct)
+
+    def test_columnsource_auto_conversion_from_pandas(self):
+        from bokeh.util.dependencies import import_optional
+        pd = import_optional('pandas')
+        if pd:
+            p = plt.figure()
+            df = pd.DataFrame({'x': [1, 2, 3], 'y': [2, 3, 4]})
+            p.circle(x='x', y='y', source=df)
```
bryevdv: pandas is optional for using Bokeh, but its required for tests, no need to conditionally import

bokeh/plotting/helpers.py
```diff
@@ -513,6 +514,20 @@ def func(self, **kwargs):
...
+            try:
+                # try converting the soruce to ColumnDataSource
+                source = ColumnDataSource(source)
+            except ValueError as err:
+                msg = "Failed to auto-convert {curr_type} to ColumnDataSource.\n" \
+                      "Original error: {err}".format(
+                    curr_type=str(type(source)),
+                    err=err.message
+                )
+                reraise(ValueError, ValueError(msg), sys.exc_info()[2])
```
bryevdv: I was not aware of this from `six`, nice to know abotu

bokeh/plotting/helpers.py
```diff
@@ -513,6 +514,20 @@ def func(self, **kwargs):
         is_user_source = kwargs.get('source', None) is not None
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
+        if not isinstance(source, ColumnDataSource):
+            try:
+                # try converting the soruce to ColumnDataSource
+                source = ColumnDataSource(source)
+            except ValueError as err:
+                msg = "Failed to auto-convert {curr_type} to ColumnDataSource.\n" \
+                      "Original error: {err}".format(
```
bryevdv: slight preference for the string to be a single line with no \ (our linter max line length is 140 I think)

bryevdv:
@bgbg Thanks for the PR!! I left two very small comments inline. Otherwise the only other suggestion I would have is an explicit mention that CDS will try to auto-convert in its docstring.

bryevdv:
Brilliant, merging now. Thanks for the PR @bgbg !

PR #5614, opened 2016-12-26
Resolve overlapping histogram bars (#3315)

### LINKED ISSUE #3315: Overlapping bins in Bokeh charts Histogram example
```
import pandas as pd
import numpy as np

# build some distributions
mu, sigma = 0, 0.5
normal = pd.DataFrame({'value': np.random.normal(mu, sigma, 1000), 'type': 'normal'})
lognormal = pd.DataFrame({'value': np.random.lognormal(mu, sigma, 1000), 'type': 'lognormal'})

# create a pandas data frame
df = pd.concat([normal, lognormal])
df[995:1005]

from bokeh.charts import Histogram
hist = Histogram(df, values='value', color='type', bins=50, legend=True)
show(hist)
```

image

ivan-krukov:
Trivial fix that skips rounding of `center` position strings.

- [x] issues: fixes #3315
- [x] tests passed

### DIFF
--- a/bokeh/charts/stats.py
+++ b/bokeh/charts/stats.py
@@ -434,9 +434,9 @@ class Histogram(BinnedStat):
         for i, b in enumerate(binned):
             width = bin_bounds[i+1] - bin_bounds[i]
             if i == 0:
-                lbl = "[%.1f, %.1f]" % (bin_bounds[i], bin_bounds[i+1])
+                lbl = "[%f, %f]" % (bin_bounds[i], bin_bounds[i+1])
             else:
-                lbl = "(%.1f, %.1f]" % (bin_bounds[i], bin_bounds[i+1])
+                lbl = "(%f, %f]" % (bin_bounds[i], bin_bounds[i+1])
             self.bins.append(Bin(bin_label=lbl, values=[binned[i]], stat=Max(),
                 width=width))
 
--- a/bokeh/charts/tests/test_stats.py
+++ b/bokeh/charts/tests/test_stats.py
@@ -37,7 +37,7 @@ def test_histogram_wo_density():
     h = Histogram(values=values, bins=3)
 
     assert len(h.bins) == 3
-    assert [b.label[0] for b in h.bins] == ['[0.0, 3.0]', '(3.0, 6.0]', '(6.0, 9.0]']
+    assert [b.label[0] for b in h.bins] == ['[0.000000, 3.000000]', '(3.000000, 6.000000]', '(6.000000, 9.000000]']
     assert [b.values[0] for b in h.bins] == [3, 3, 4]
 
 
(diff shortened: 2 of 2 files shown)


bryevdv:
@ivan-krukov wow, thanks! I think ultimately a more comprehensive solution will involve not relying on categorical coordinates for histograms like this. But it will be some time before I can dive into that, so a quick immediate fix is certainly welcome, and I did not discover this approach when I took a brief look a few months ago. 

For completeness of documentation, since the s3 test uploads will not be enabled for your PR, can you attach some before/after screenshots below?

bryevdv:
@ivan-krukov there are also two tests that will need updating:
```
__________________________ test_histogram_wo_density ___________________________
    def test_histogram_wo_density():
        values = list(range(10))
        h = Histogram(values=values, bins=3)
    
        assert len(h.bins) == 3
>       assert [b.label[0] for b in h.bins] == ['[0.0, 3.0]', '(3.0, 6.0]', '(6.0, 9.0]']
E       assert ['[0.000000, ...0, 9.000000]'] == ['[0.0, 3.0]',... '(6.0, 9.0]']
E         At index 0 diff: '[0.000000, 3.000000]' != '[0.0, 3.0]'
E         Use -v to get the full diff
bokeh/charts/tests/test_stats.py:40: AssertionError
___________________________ test_histogram_w_density ___________________________
    def test_histogram_w_density():
        values = list(range(10))
        h = Histogram(values=values, bins=3, density=True)
    
        assert len(h.bins) == 3
>       assert [b.label[0] for b in h.bins] == ['[0.0, 3.0]', '(3.0, 6.0]', '(6.0, 9.0]']
E       assert ['[0.000000, ...0, 9.000000]'] == ['[0.0, 3.0]',... '(6.0, 9.0]']
E         At index 0 diff: '[0.000000, 3.000000]' != '[0.0, 3.0]'
E         Use -v to get the full diff
bokeh/charts/tests/test_stats.py:49: AssertionError
============================= 190 tests deselected =============================
======= 2 failed, 831 passed, 190 deselected, 10 rerun in 41.85 seconds ========
```

ivan-krukov:
I have updated the tests you pointed out.

You are right about this being a temporary fix - if the values passed to `bins` are on the order of magnitude of `1e-6` or below, there still will be bar overlap.

Here is my visual test:

```python
from bokeh.charts import Histogram
from bokeh.io import show
from bokeh.sampledata.autompg import autompg

# Force overlap with decimal data (x < 1)
x = autompg.mpg / 10
show(Histogram(x))
```

Before the fix, I got this:
bokeh_hist_before

After the fix:
bokeh_hist_after

If we pass very small values to bins (_e.g._ `x = autompg.mpg/1e6`), we will get overlap again:

bokeh_hist_small

I would still argue that a temporary fix is better than none.

bryevdv:
> I would still argue that a temporary fix is better than none. 

Definite agree, and happy to merge this PR once tests are green. Thanks @ivan-krukov !

Issue #4918, opened 2016-08-02
Using a Custom JSON Encoder


cperiz:
I am trying to pass a data source which consists of variables of type deque into the Bokeh server. Since Bokeh's JSON serialization currently does not support this format the probable best solution, right now, would be to allow one to write a custom JSON encoder. However, Bokeh currently does not allow for the specification of a custom JSON encoder in its software. This should be made possible.

bryevdv:
I'm not certain this is a reasonable request, in general. Unlike the case where there is one or possibly two processes *both python* using the exact same custom serialization code at both ends, Bokeh is a multi-language project, and the other end of a completely general python custom serializer could need to be a corresponding custom JS serializer. Make the Bokeh JS serializer more extensible would be a much more involved process, and even were it done, it leaves open the door for very hard to debug mis-usages and the corresponding support burden. 

What *might* make sense is to have a way to adapt other types to the core set of Bokeh protocol types (which would be stable, and not change). If you'd care to work up a more detailed proposal for what that might look like from a user perspective, we'd be happy to consider it. 

Alternatively, a much shorter term solution is simply to make `deque` serializable by the built-in serializer. 

Thoughts @bokeh/dev ?

cperiz:
I understand your point. I am not in a position to work up a more detailed proposal right now. If it doesn’t do much damage to the code, I would certainly appreciate it if you could make deque serializable by the built-in serializer.

almarklein:
The relevant code is here: 
* https://github.com/bokeh/bokeh/blob/master/bokeh/core/json_encoder.py
* https://github.com/bokeh/bokeh/blob/master/bokeh/util/serialization.py

It should not be too hard to add support for `deque`. This is something we could do, though a PR from your end would ensure it will be added by the next release ;)

cperiz:
If you do get it done, it would be much appreciated. If we manage to get down to resolving this we will certainly create a PR :)

birdsarah:
Is this relevant: https://github.com/bokeh/bokeh/issues/3493

bryevdv:
I just tried the code in #3493 and it still works fine on master. @cperiz can you provide some code where `deque` is failing?

cperiz:
```python
from bokeh.plotting import figure, output_file, show, curdoc
from collections import deque
from bokeh.client import push_session
from bokeh.models import ColumnDataSource

session = push_session(curdoc())
p = figure(plot_width=400, plot_height=400)
[...]

PR #5615, opened 2016-12-27
Deserialize a deque as a JSON list.

jfinkels:
Fixes issue #4918. I didn't make a changelog entry because I didn't see a heading for version 0.12.4 and I wasn't sure what to do.


bryevdv:
Hi @jfinkels thanks for the PR! Changelogs are auto-generated from GH issues and tags when releases happen, so there is nothing you need to worry about. This LGTM, merging!

PR #5616, opened 2016-12-27
Make use of Fabric3

### LINKED ISSUE #5557: Use python3 version of fabric
There's a port of fabric that supports python3 (conda install fabric3) - https://github.com/mathiasertl/fabric/

consider using it as we move away from python 2.7.

souravsingh:
Since Bokeh is moving towards Python3, it would make sense to use Python 3 fork of Fabric.

Reference Conversation: https://github.com/bokeh/bokeh/issues/5557

- [x] issues: fixes #5557 
- [ ] tests added / passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ matrix:
       env: GROUP=examples
     - python: 3.4
       env: GROUP=examples
-    # Linting and Docs (no python3 version of fabric)
-    - python: 2.7
+    # Linting and Docs
+    - python: 3.4
       env: GROUP=flake_docs
     # Integration test
     - python: 3.4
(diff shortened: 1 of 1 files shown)


bryevdv:
Hi @souravsingh Thanks for the PR! A couple of more steps are needed to complete this: The first is that the linting/docs CI Job should be able to be switched to Python 3, that change can be made here:

https://github.com/bokeh/bokeh/blob/master/.travis.yml#L31-L33

The next change is that our CI scripts need to actually install `fabric3` instead of `fabric`, that change can be made here:

https://github.com/bokeh/bokeh/blob/master/.travis.yml#L108

Note that `fabric3` is not in the default repo, but it is in the conda forge channel. I'd suggest actually splitting that install into two lines like this:

```
conda install anaconda-client --yes
conda install -c conda-forge fabric --yes
```
If you have the opportunity to make these small changes that will be great, and we can test out the machinery on some upcoming dev/rc builds.

bryevdv:
@souravsingh awesome, thanks! I will merge this now and it will get tested "for real" on the next dev build

PR #5620, opened 2016-12-27
Calculate bounds correctly for log axis

### LINKED ISSUE #5549: Correctly handle data values <= 0 on a log scale

clairetang6:
This PR is to fix bugs that are caused when plotting data on a log scale when some of the data values are <= 0. To correctly calculate the minimum and maximum data values that can be plotted in a log axis, you have to take the min and max of only the positive values. Currently, there isn't any way for a range to know what the `min(data[data > 0])` because the ranges are passed `bounds` which contain `min(data)` and `max(data)`. The approach that I took to fix this in this PR was to add a new method `log_bounds` to the `GlyphView` in analogy to the `bounds` method that returns `minX`, `minY`, `maxX`, `maxY` values are determined based on positive values of x and y. Then the `log_bounds` can be passed to the range for a log axis instead of the `bounds`.

- [x] issues: fixes #5549
- [x] tests added / passed
~~- [ ] release document entry (if new feature or API change)~~
- [x] using glyph-specific `_bounds` in the `log_bounds` method
- [x] turn `_mapper_type` into  public `mapper_hint`

(I'm not sure if adding a method on GlyphView is an API change?)


clairetang6:
Before:
image

After:
image

[Test Code](https://github.com/bokeh/bokeh/issues/5389#issuecomment-266791148)
but `values = np.linspace(-0.1, 0.9, 21)`

bokehjs/test/models/glyphs/index.coffee
```diff
@@ -1,3 +1,4 @@
 require "./image"
 require "./image_rgba"
 require "./image_url"
+require "./rect"
```
mattpap: No newline at the end. This will cause code quality tests to fail.

bokehjs/src/coffee/core/util/bbox.coffee
```diff
@@ -5,6 +5,20 @@ export empty = () -> {
   maxY: -Infinity
 }
 
+export positive_x = () -> {
+  minX:  Number.MIN_VALUE,
+  minY: -Infinity,
+  maxX:  Infinity,
+  maxY:  Infinity
```
mattpap: Use trailing comma. This isn't automatically verified, but it's in a good taste (I see where this came from, but still). Or you can use no commas at all (it's allowed by coffeescript).

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -178,24 +178,42 @@ export class PlotCanvasView extends BokehView
     # Update any DataRange1ds here
     frame = @model.frame
     bounds = {}
+    log_bounds = {}
+
+    calculate_log_bounds = false
+    for r in _.values(frame.x_ranges).concat(_.values(frame.y_ranges))
+      if r instanceof DataRange1d
+        if r._mapper_type == "log"
```
mattpap: Here you use a private property in an unrelated class. Either there is a public property that should be preferred or `_mapper_type` should be made public. Using `_` for indicating private members is just a suggestion and we don't enforce it too much. However, having a switch to TypeScript on the horizon, there will be no mercy for this kind of code in the future.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -178,24 +178,42 @@ export class PlotCanvasView extends BokehView
...
     follow_enabled = false
     has_bounds = false
 
     for xr in _.values(frame.x_ranges)
       if xr instanceof DataRange1d
-        xr.update(bounds, 0, @model.id)
+        if xr._mapper_type == "log"
+          xr.update(log_bounds, 0, @model.id)
+        else
+          xr.update(bounds, 0, @model.id)
```
mattpap: Better code would be:
```
use_bounds = if xr._mapper_type == "log" then log_bounds else bounds
xr.update(use_bounds, 0, @model.id)
```
This way you avoid repetition, because the only difference is in bounds. If other arguments differed as well, then the original code would make more sense.

bokehjs/src/coffee/models/glyphs/glyph.coffee
```diff
@@ -55,6 +55,18 @@ export class GlyphView extends BokehView
...
+  log_bounds: () ->
+    if not @index?
+      return bbox.empty()
+    log_x_index = rbush()
+    log_y_index = rbush()
+    positive_x_pts = @index.search(bbox.positive_x())
+    positive_y_pts = @index.search(bbox.positive_y())
+    log_x_index.load(positive_x_pts)
+    log_y_index.load(positive_y_pts)
+    bb = {minX: log_x_index.data.minX, minY: log_y_index.data.minY, maxX:log_x_index.data.maxX, maxY:log_y_index.data.maxY}
```
mattpap: This line should be split into four lines.

mattpap:
> (I'm not sure if adding a method on GlyphView is an API change?)

It is, at least for people implementing custom models. However, such change is far from modifying models' properties.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
bryevdv: I'd be OK making `mapper_type` public, tho I might prefer a name like `mapper_hint`

bokehjs/src/coffee/models/glyphs/glyph.coffee
```diff
@@ -55,6 +55,18 @@ export class GlyphView extends BokehView
...
+    return bb
+
```
bryevdv: @clairetang6 This is a clever use of rbush to find both the positive points in both half planes, then to find the max/min of each set. I definitely think the first use is great, because it uses an existing index. I do wonder about the use to find max/min, just because the overhead to both `load`  and then `search` the index might be more than doing a simple linear search of `positive_x_pts` and `positive_x_pts`. 

My only other comment is that I am not sure every glyph defines `@index` we should verify or catalog any that don't and add a warning or some kind of trivial if not-great fallback in those cases.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
clairetang6: Ok, if I'm going to make this property public, then do I use `@define`? I'm not entirely clear on when `@define`, `@internal`, etc. are supposed to be used.

bokehjs/src/coffee/models/glyphs/glyph.coffee
clairetang6: The seed was planted when I was working with rbush to fix a bug with box select some time ago! I tried out the linear search and it's much faster than using a new index, so I'll change that. 

Another thing that I encountered is that the index `search` returns bboxes (points or rectangles) that are either contained in or intersect with the bbox passed in. So, searching the positive half plane will also return rectangle bboxes (Rect, Quad, etc.) that cover both the positive and negative side of 0.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
mattpap: You use `@internal` when you don't want to propagate changes from bokehjs to bokeh. This way, an internal property in bokehjs doesn't have to have its mirror in bokeh. In this case you want to use `@internal`.

clairetang6: Ok, thanks! Working on that now.

bokehjs/test/models/glyphs/index.coffee
bryevdv: checks ignore test files?

clairetang6: I added the newline back. I think GitHub doesn't consider adding a newline changing the code, so it's still showing up here.

clairetang6:
I haven't added a release document entry yet. I'm not sure where that goes.

bryevdv:
> I haven't added a release document entry yet. I'm not sure where that goes.

Is there anything that needs to to be documented or pointed out? This seems more like a bug fix, in which there might not need to be anything new said.

bryevdv:
I there is anything to add, this is probably the best place for now:

http://bokeh.pydata.org/en/latest/docs/user_guide/plotting.html#log-scale-axes

clairetang6:
Ok, I'll add a sentence or two into the docs.

bryevdv:
Looks great, thanks @clairetang6 !

PR #5621, opened 2016-12-28
Add custom extensions for autocompletion widgets

### LINKED ISSUE #5596: Fix/drop AutocompleteInput
[On linux chrome v55.0.2883.87, bokeh master] The menu revealed by AutocompleteInput does not have any css styling, it is just floating text which overlaps with other elements on the page. I assume at some point some jquery css had been excised causing this widget to break. With the current trend I assume that this widget should just get removed. (I'd find it useful though to have an example of a custom AutocompleteInput widget.)

azjps:
**Not ready to merge.** Just an initial kind of working attempt at replacing autocomplete widget. Please let me know if the scope of this change looks okay, for example since the `textext` and `tokeninput` extensions are currently non-functional, I can drop them out or spend a bit more time seeing if they can be made to work. There are several other issues mentioned in the TODOs.

The existing jquery-ui based autocompletion widget is non-functional (at
the very least, no css is loaded for the dropdown menu etc). These
custom extensions implement several variations of an autocompletion
widget, using a few of the many available libraries (jquery-ui,
awesomplete, jquery-textext, jquery-tokeninput), as well as a demo app
that can be served by bokeh server.

The jquery-ui and awesomplete extensions are roughly functional, the
textext and tokeninput are just proofs of concept and are effectively
non-functional currently.

- [x] issues: fixes #5596
- [ ] tests added / passed: N/A?

### DIFF
--- a/examples/custom/autocomplete/autocomplete_input.coffee
+++ b/examples/custom/autocomplete/autocomplete_input.coffee
@@ -1,0 +1,28 @@
+import * as _ from "underscore"
+# import "jquery-ui/autocomplete"
+
+import {TextInput, TextInputView} from "models/widgets/text_input"
+import * as p from "core/properties"
+
+export class AutocompleteInputView extends TextInputView
+
+  render: () ->
+    super()
+    $input = @$el.find("input")
+    $input.autocomplete({source: @model.completions
+                         minLength: @model.min_chars
+                         autoFocus: @model.auto_first
+                        })
+    # TODO: add css?
+    $input.autocomplete("widget").addClass("bk-ui-autocomplete")
+    return @
+
+export class AutocompleteInput extends TextInput
+  type: "AutocompleteInput"
+  default_view: AutocompleteInputView
+
+  @define {
+    completions: [ p.Array, [] ]
+    min_chars: [ p.Number, 2 ]
+    auto_first: [ p.Bool, false ]
+  }
--- a/examples/custom/autocomplete/autocomplete_input.py
+++ b/examples/custom/autocomplete/autocomplete_input.py
@@ -1,0 +1,22 @@
+from bokeh.core.properties import List, String, Int, Bool
+from bokeh.models.widgets import TextInput
+
+class AutocompleteInput(TextInput):
+    """ Single-line input widget with auto-completion. """
+
+    __implementation__ = "autocomplete_input.coffee"
+    __css__ = ["https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.css"]
+
+    completions = List(String, help="""
+    A list of completion strings. This will be used to guide the
+    user upon typing the beginning of a desired value.
+    """)
+
+    min_chars = Int(default=2, help="""
+    Minimum characteres the user has to type before the autocomplete
+    popup shows up.
+    """)
+
+    auto_first = Bool(default=False, help="""
+    Whether the first element should be automatically selected
+    """)
--- a/examples/custom/autocomplete/textext_input.coffee
+++ b/examples/custom/autocomplete/textext_input.coffee
@@ -1,0 +1,47 @@
+import * as _ from "underscore"
+# Cannot import jquery here, otherwise it overwrites (?) the
+# globally loaded jquery object with function plugins
+# import * as $ from "jquery"
+
+import {TextInput, TextInputView} from "models/widgets/text_input"
+import * as p from "core/properties"
+
+export class TextExtInputView extends TextInputView
+
+  render: () ->
+    super()
+    # TODO: using below hack to use globally loaded jquery, which has
+    # the function plugin extensions, instead of saved @$el object
+    $input = $(@$el.find("input"))
+    completions = @model.completions
+    if $input.textext?
+      # TODO: implement contains instead of match from start:
+      # https://github.com/alexgorbatchev/jquery-textext/issues/168
+      # http://stackoverflow.com/questions/29688035/jquery-textext-set-autocomplete-suggestions-to-filter-based-on-contains-rather
+      $input.textext(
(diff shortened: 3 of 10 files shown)


examples/custom/autocomplete/textext_input.coffee
```diff
@@ -0,0 +1,47 @@
...
+
+export class TextExtInputView extends TextInputView
+
+  render: () ->
+    super()
+    # TODO: using below hack to use globally loaded jquery, which has
+    # the function plugin extensions, instead of saved @$el object
+    $input = $(@$el.find("input"))
+    completions = @model.completions
+    if $input.textext != undefined
```
mattpap: `if $input.textext?`

examples/custom/autocomplete/textext_input.coffee
```diff
@@ -0,0 +1,47 @@
...
+
+export class TextExtInput extends TextInput
+  type: "TextExtInput"
+  default_view: TextExtInputView
+
+  @define {
+      completions: [ p.Array, [] ]
+      min_chars: [ p.Number, 2 ]
+      auto_first: [ p.Bool, false ]
+    }
```
mattpap: Why copying code, you can fix this indentation. This was left this way previously to reduce merge conflicts with other devs' work.

examples/custom/autocomplete/autocomplete_input.coffee
```diff
@@ -0,0 +1,28 @@
...
+
+import {TextInput, TextInputView} from "models/widgets/text_input"
+import * as p from "core/properties"
+
+export class AutocompleteInputView extends TextInputView
+
+  render: () ->
+    super()
+    $input = @$el.find("input")
+    $input.autocomplete(source: @model.completions,
```
mattpap: Use explicit object literals, i.e. `{ ... }`.

examples/custom/autocomplete/autocomplete_input.coffee
```diff
@@ -0,0 +1,28 @@
...
+                        minLength: @model.min_chars,
+                        autoFocus: @model.auto_first
```
mattpap: And trailing commas.

examples/custom/autocomplete/autocomplete_input.py
```diff
@@ -0,0 +1,22 @@
+from bokeh.core.properties import List, String, Int, Bool
+from bokeh.models.widgets import TextInput
+
+class AutocompleteInput(TextInput):
+    """ Single-line input widget with auto-completion. """
+
+    __implementation__ = "autocomplete_input.coffee"
+    __css__ = ["https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.css"]
```
mattpap: This seems unnecessary as you use built-in jquery-ui in the implementation. Note that we use prefixed CSS classes, so there will be unwanted interactions between those two versions or this include will be a no-op.

examples/custom/autocomplete/textext_input.coffee
```diff
@@ -0,0 +1,47 @@
+import * as _ from "underscore"
+# Cannot import jquery here, otherwise it overwrites (?) the
+# globally loaded jquery object with function plugins
```
mattpap: It shouldn't, because it's loaded as a module ("no conflict" mode).

mattpap: Still you shouldn't mix both in one code.

examples/custom/autocomplete/autocomplete_input.py
azjps: The prefixed jquery css does not include ui-autocomplete, so if I remove `__css__` or use the current AutocompleteInput widget, the autocomplete menu is a mess. Maybe it was an oversight, but my takeaway from that and some other open issues was that `jquery-ui` was being dropped from bokeh's packaged css, and shifted to these custom extensions. (The `.addClass("bk-ui-autocomplete")` does not do anything above.)

examples/custom/autocomplete/textext_input.coffee
azjps: Can you [briefly] walk me through this, bearing in mind I have no JS experience? Oddly, if I uncomment the above jquery import, then nothing happens. When I insert a breakpoint in `@render` with the import uncommented, `$.fn.tokenInput === undefined`, and with the import commented, `$.fn.tokenInput !== undefined`.

I'd love to do things the right way here, but this was the culmination of frustration trying to understand the hierarchies/interactions of these modules and imports.

azjps:
The [jquery-tokeninput](http://loopj.com/jquery-tokeninput) widget kind of works now, which would provide a way to select multiple options from `completions` (similar to current `MultiSelect` widget but requiring much less vertical space). Plenty cleanup to do still though.

What should the scope of this PR be? I can move the `jquery-tokeninput` extension to a separate PR and keep this to just the `jquery-autocomplete` extension. I can also remove the existing `AutoComplete` widget in this PR or in a separate PR.

(Apologies for the force push, was in a bit of a hurry.)

bryevdv:
@azjps I'd suggest two PRs:

* one to deprecate `AutoComplete` and add a `jquery-autocomplete` example
* a separate one for the ` jquery-tokeninput` example

We should probably at least have a short deprecation period, unless `AutoComplete` is wholly broken (I can't remember) In that case the first PR could just remove instead of deprecate.

azjps:
Sounds good. `AutoComplete` kind of functions but styling isn't present, it also doesn't call `change_input` when an option is selected which I still have to fix.
bokeh_autocomplete

azjps:
Apologies, I've had functional (but probably not up to standard) code for this but not sure I can provide. Closing for now.

bryevdv:
@azjps Thanks please feel free to open a new PR in the future if you are able.

PR #5622, opened 2016-12-28
Updating versioneer to 0.17+

### LINKED ISSUE #5613: Windows setup.py build from source and versioneer
On fresh install from current bokeh master on Windows 8.1, I can run `python setup.py install --build-js` but `import bokeh` will cause a complaint that version cannot be detected. (Note that I have cygwin installed but I am not using cygwin nor cygwin-git, using `cmd.exe` and windows git installation in `%PATH%`.) This seems to be a known issue on [versioneer issue tracker](https://github.com/warner/python-versioneer/issues/14), but I thought I would file it as an issue here for easier searching. I understand that it would probably take non-trivial work to update the version of versioneer, but it would be nice to make a note in the windows developer guide docs. (Assuming I'm not missing something of course.) I would be happy to submit a PR for such but I don't know how to deal with this besides manually substituting `git_refnames="0.12.x"; git_full="0.12.x"` every time I run `setup.py`. 

git version: 2.10.2.windows.1
```python
(env2016) C:\projects\bokeh\examples\custom>python -m bo
keh serve .
  File "C:\Anaconda\envs\env2016\lib\runpy.py", line 163, in _run_module_as_main
    mod_name, _Error)
  File "C:\Anaconda\envs\env2016\lib\runpy.py", line 111, in _get_module_details
    __import__(mod_name)  # Do not catch exceptions initializing package
  File "C:\Anaconda\envs\env2016\lib\site-packages\bokeh-0.12.4dev9_20_ge4da7e7-
py2.7.egg\bokeh\__init__.py", line 16, in <module>
    from .util.version import __version__; __version__
  File "C:\Anaconda\envs\env2016\lib\site-packages\bokeh-0.12.4dev9_20_ge4da7e7-
py2.7.egg\bokeh\util\version.py", line 36, in <module>
    __base_version__ = base_version()
  File "C:\Anaconda\envs\env2016\lib\site-packages\bokeh-0.12.4dev9_20_ge4da7e7-
py2.7.egg\bokeh\util\version.py", line 34, in base_version
    return VERSION_PAT.search(__version__).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

azjps:
First cut, still need to check local changes. Somehow windows is expanding `"*"` as a glob in `subprocess.Popen([.., "*", ..], shell=False)`, I'm hoping its just an issue with my environment setup 😦 .

Rebuilt versioneer using latest version, which has some windows
compatibility fixes and so forth. The main migration is moving
configuration for versioneer from `setup.py` to `setup.cfg`.

- [x] issues: fixes #5613 
- [ ] tests added / passed


.gitignore
```diff
@@ -113,3 +113,6 @@ win-64
 # generated coffeescript
 /bokehjs/test/common/generated_defaults/models_defaults.coffee
 /bokehjs/test/common/generated_defaults/widgets_defaults.coffee
+
```
azjps: not supposed to be committed, i'll remove

versioneer.py
```diff
@@ -1,469 +1,1143 @@
-#! /usr/bin/python
 
-"""versioneer.py
+# Version: 0.17
```
mattpap: Shouldn't this be installed with conda?

.gitignore
```diff
@@ -113,3 +113,6 @@ win-64
...
+*.styl
+*.png
```
mattpap: Missing trailing newline.

azjps: Yeah, fallout from my attempts to figure out how to use git effectively on windows. This wasn't even supposed to be committed, it was auto-added by the github UI 😐

versioneer.py
azjps: This isn't the `versioneer` library itself, [python-versioneer](https://github.com/warner/python-versioneer) generates a `versioneer.py` to the root of the project repository, so that `setup.py` can use it to compute versions (I guess it was easiest to do this way in order to perform heuristics with the repository directory). In fact `versioneer.py` doesn't exist in the python-versioneer repo. I'll add some additional context to this PR's commit message.

.gitignore
bryevdv: no worries, we are here to help :D

bryevdv:
> The main migration is moving configuration for versioneer from setup.py to setup.cfg.

Ah, that's nice

azjps:
Okay, doesn't seem like there were any local changes made to `_version.py` that are still needed, besides maybe some minor whitespace tweaks. The first commit in PR is direct result of `versioneer install`, the second were some minor changes that I added to add a workaround for myself.

The only other side effect I know of this change is that [the version string uses a "+" now to conform to PEP440](https://github.com/warner/python-versioneer/blob/master/UPGRADING.md).

bryevdv:
> The only other side effect I know of this change is that the version string uses a "+" now to conform to PEP440.

Oh, that's actually significant, unfortunately. There is a significant amount of internal tooling with expectations around the existing format. I would not be oppose to patching our checked in `_version.py` to change things.

bryevdv:
Or alternatively instead of patching `_version.py`, perhaps a conversion step could be applied in 

https://github.com/bokeh/bokeh/blob/master/bokeh/util/version.py

which converts the new format from versioneer to the old format.

azjps:
versioneer has a [`style=git-describe` config](https://github.com/warner/python-versioneer/blob/1d80ff51544e52ff6216751dbede992f65bb6b87/details.md#how-do-i-select-a-version-style) which uses the `git-describe` like string `0.12.4dev10-2-g902c066-dirty` instead of the PEP440 version string versioneer outputs by default `0.12.4dev10+2.g902c066.dirty`. How does that sound? It should probably get tested for a couple of use cases first though, the docs on versioneer aren't up-to-date.

bokeh/_version.py
```diff
@@ -1,527 +1,527 @@
...
-    try:
-        if cfg.parentdir_prefix:
-            return versions_from_parentdir(cfg.parentdir_prefix, root, verbose)
-    except NotThisMethod:
-        pass
-
-    return {"version": "0+unknown", "full-revisionid": None,
-            "dirty": None,
-            "error": "unable to compute version", "date": None}
+
```
azjps: Uh, I'm not sure why git doesn't collapse the diff here. I re-generated this file with versioneer to make sure I didn't leave anything out, maybe versioneer does a git remove + git add.

versioneer.py
```diff
@@ -1043,10 +1050,15 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
 
     # if there is a tag matching tag_prefix, this yields TAG-NUM-gHEX[-dirty]
     # if there isn't one, this yields HEX[-dirty] (no NUM)
-    describe_out, rc = run_command(GITS, ["describe", "--tags", "--dirty",
-                                          "--always", "--long",
-                                          "--match", "%%s*" %% tag_prefix],
```
azjps: note to self -- `%%` accidentally copy+pasted in prior commit

bokeh/_version.py
```diff
@@ -1,191 +1,527 @@
...
+        raise NotThisMethod("'git rev-parse --git-dir' returned error")
+
+    # if there is a tag matching tag_prefix, this yields TAG-NUM-gHEX[-dirty]
+    # if there isn't one, this yields HEX[-dirty] (no NUM)
+    # NOTE: I am encountering a baffling issue, where windows (possibly due
+    # to cygwin, although I am not using it) is expanding * even though
+    # shell=False! As a temporary workaround, I've remove the --match
+    # argument since tag_prefix is empty in the bokeh setup config.
+    describe_args = ["describe", "--tags", "--dirty",
+                     "--always", "--long"]
```
bryevdv: comparing to our old set of flags that were used, this adds `--long`. That might be fine but we should investigate what exactly it does

bryevdv:
@azjps this is looking good, just an FYI re: schedule I'd love to merge this right after the `0.12.4` release or thereabouts (there's enough other build-changes going into `0.12.4` I don't want to add this on top)

bryevdv:
@azjps I would love to merge this soon, I think the only outstanding issue is the file endings in `_version.py`. Our checker is reporting : 
```
File contains carriage returns at end of line: bokeh/_version.py, line 512
```
for every line. Maybe there is an editor setting you can change to unix-style (just line feed) endings, then save and push?

Alternatively, this SO answer suggests a `.gitattributes` file can be set that will have git automatically fix up line endings, which might be nice / worth trying:

http://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings

azjps:
So that's why the git diff wasn't collapsed before (yay [windows](https://github.com/acrogenesis/windows-carriage-return-remover)).

bryevdv:
@azjps No, now it's just:
```
E         File starts with more than 1 empty line: bokeh/_version.py, line 1
E       assert 1 == 0
E        +  where 1 = len(['File starts with more than 1 empty line: bokeh/_version.py, line 1'])
```
This one I can fix in two seconds on the GH UI, so I will do that, and then when it goes green it should!) I will merge

bryevdv:
@azjps I did look into that `--long` option, and the help had this to say:
```
--long
           Always output the long format (the tag, the number of commits and the abbreviated commit name) even when it
           matches a tag. This is useful when you want to see parts of the commit object name in "describe" output, even
           when the commit in question happens to be a tagged version. Instead of just emitting the tag name, it will
           describe such a commit as v1.2-0-gdeadbee (0th commit since tag v1.2 that points at object deadbee....).
```

given that, I think we actual must remove that option to match the previous behaviour

bryevdv:
@azjps I went ahead and made that small change, I'd like to merge this today and get a dev build out to get some early experience with it on more realistic deploys

bryevdv:
Actually this might need more tweaking, unfortunately.  I tested locally and clean, dirty, and `__conda_version__.py` override all worked as expected. But if I made a local tag:
```
[bryan:~/work/bokeh]$ python                                                                                            (vers)
Python 3.4.4 |Anaconda 2.3.0 (x86_64)| (default, Jan  9 2016, 17:30:09)
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import bokeh
  File "/Users/bryan/work/bokeh/bokeh/__init__.py", line 16, in <module>
    from .util.version import __version__; __version__
  File "/Users/bryan/work/bokeh/bokeh/util/version.py", line 36, in <module>
    __base_version__ = base_version()
  File "/Users/bryan/work/bokeh/bokeh/util/version.py", line 34, in base_version
    return VERSION_PAT.search(__version__).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
>>>
```

bryevdv:
proximate problem: there is no more function `get_versions`

bryevdv:
OK it actually looksl like my last change was too hasty and is what broke things, will revert it

bryevdv:
OK! checking locally:
```
[bryan:~/work/bokeh]$ vi setup.py                                                                                       (vers)
[bryan:~/work/bokeh]$ python -c "import bokeh; print(bokeh.__version__)"                                               (vers✱)
0.12.4-11-g6dd1f0f-dirty
[bryan:~/work/bokeh]$ git checkout -- setup.py                                                                         (vers✱)
[bryan:~/work/bokeh]$ python -c "import bokeh; print(bokeh.__version__)"                                                (vers)
0.12.4-11-g6dd1f0f
[bryan:~/work/bokeh]$ git tag 1.2.3                                                                                     (vers)
[bryan:~/work/bokeh]$ python -c "import bokeh; print(bokeh.__version__)"                                                (vers)
1.2.3
[bryan:~/work/bokeh]$ git tag -d 1.2.3                                                                                  (vers)
Deleted tag '1.2.3' (was 6dd1f0f)
[bryan:~/work/bokeh]$ vi bokeh/__conda_version__.py                                                                     (vers)
[bryan:~/work/bokeh]$ python -c "import bokeh; print(bokeh.__version__)"                                                (vers)
2.3.4
```
All looks great. Sorry for the noise earlier I will merge when green.

bryevdv:
Thanks @azjps ! sorry again for the noise edits.

PR #5630, opened 2016-12-29
Drop jquery from core bokehjs

### LINKED ISSUE #5625: Drop jquery from core bokehjs
This is another milestone in making (core) bokehjs more maintainable and smaller. jquery will remain a dependency of bokehjs' widgets (at least data table's, we will see about the other). Setting milestone to 0.12.5, but this will likely take more than that.

mattpap:
fixes #5625, #5670


mattpap:
At this point it's possible to drop jquery by removing its import from `main.coffee`. This gives `{761 -> 677} kB` minified size reduction.

mattpap:
@bryevdv, this PR still needs some polishing, but I want it to be prioritized and merged as soon as possible due to its scope and impact.

bryevdv:
Agreed, happy to merge when it's green.

mattpap:
> Agreed, happy to merge when it's green.

Unfortunately "all green" isn't going to be enough here. There are some parts that need closer scrutiny. I will outline those parts in inline comments.

mattpap:
The last (integration) test failure is this:
```
Failed: Console errors: [{'level': 'SEVERE', 'timestamp': 1484136865755, 'message': 'TypeError: setting a property that has only a getter'}]
```
related to `DataTable` (`interaction/test_datatable.py`).

I'm unable to reproduce this locally and, as you see in the error message, there is no information what property, on what object, where in the source code. Typical attitude in a "simple" programming environment (JS).

mattpap:
Downgrading firefox to what saucelabs uses helped a lot. Apparently there are file names and line numbers is FF's JS console, so its saucelabs that's unable to report the full picture.

bokehjs/src/coffee/models/tools/gestures/wheel_pan_tool.coffee
```diff
@@ -6,18 +6,7 @@ import * as p from "../../../core/properties"
...
-    # we need a browser-specific multiplier to have similar experiences
-    if navigator.userAgent.toLowerCase().indexOf("firefox") > -1
-      multiplier = 20
-    else
-      multiplier = 1
-
-    if e.originalEvent?.deltaY?
-      delta = -e.originalEvent.deltaY * multiplier
-    else
-      delta = e.bokeh.delta
```
mattpap: @bryevdv, this is the code I spoke about in the meeting. `e.bokeh.delta` was computed by jquery-mousewheel plugin and is as browser cross-compatible as possible. However, the plugin was circumvented by the earlier branch containing `deltaY`. The reason `multiplier` is necessary is that firefox returns delta in the number of lines, but other web browsers return it in pixels (it's also possible for it to be returned in pages, but no one uses this, as far as I know).

Tools that depend on wheel event have to be tested under all web browsers we support, especially safari and IE, and FF and Chrome on non-Linux platforms.

bryevdv: OK will test all browsers on OS X an some windows VMs. Can you check several linux combinations?

bryevdv: Tested and working on OS X (safari, chrome, FF) and Win8 (IE, chrome FF)

bryevdv: If is working on linux for you I'd suggest we merge and make an immediate dev build and ask for community to help test it.

mattpap: @bryevdv, it works on Linux, at least in Chrome 51 and FF 41 & 47. My biggest concern was Safari, but if that works we should be good. Still I would like to test a range of browser versions, but I don't have a setup for this right now. On a side note, how hard would it be to have some browser versions as conda packages, so I would easily switch between them as currently I do with Python versions?

mattpap:
Going through examples' report, I noticed that there may be some non-determinism as to ordering of labels in legend items, especially in charts' examples (see e.g. `charts/file/boxplot_single`), but I'm not sure if charts or legends are at fault here. This doesn't seem to be related to this PR (locally `master` and this branch agree on the order).

bokehjs/src/coffee/models/sources/ajax_data_source.coffee
```diff
@@ -26,32 +23,30 @@ export class AjaxDataSource extends RemoteDataSource
...
-        @data = data
-      else
-        logger.error("unsupported mode: " + mode)
-      logger.trace(data)
-      return null
-    ).error(() ->
-      logger.error(arguments)
+    xhr = new XMLHttpRequest()
+    xhr.open(@method, @data_url, true)
+    xhr.withCredentials = false
```
mattpap: @bryevdv, I'm not sure about this. The original code uses `true`, but using `true` here fails with cross-domain errors. This isn't unexpected, because we don't send any credentials.

bryevdv: I'm not sure either, if things work with `false` that seems like the correct value anyway. Maybe ping dev or main for more input tho

bryevdv:
> Going through examples' report, I noticed that there may be some non-determinism as to ordering of labels in legend items, especially in charts' examples (see e.g. charts/file/boxplot_single), but I'm not sure if charts or legends are at fault here. This doesn't seem to be related to this PR (locally master and this branch agree on the order). 

In that case make a separate issue to note it. I do plan a comprehensive overhaul of charts but probably not until `0.12.7` (altair is a higher priority)

bokehjs/test/size.coffee
```diff
@@ -6,11 +6,11 @@ path = require "path"
 build_dir = path.normalize("#{__dirname}/../build")
 
 LIMITS = {
-  "css/bokeh-widgets.min.css": 527
-  "css/bokeh.min.css":         74
-  "js/bokeh-widgets.min.js":   290
+  "css/bokeh-widgets.min.css": 160
+  "css/bokeh.min.css":         60
+  "js/bokeh-widgets.min.js":   380
```
bryevdv: curious, why did this increase?

mattpap: jquery was moved there. After all we aren't dropping it completely.

bryevdv: got it, thanks

bryevdv:
@mattpap is there anything worrying besides `AjaxDataSource`? Should I plan to merge this tomorrow morning?

mattpap:
@bryevdv, `AjaxDataSource` isn't that much worrying. It works, though some changes may be needed and a TODO to resolve.

> Should I plan to merge this tomorrow morning?

Yes, this is ready. The sooner it's merge the better because want to do work on top of this and squash merge prevents me from this.

bryevdv:
OK, I wil merge this now and kick off a dev build tomorrow then. Please make an issue to investigate the ajax source credentials thing, though.

bryevdv:
Actually, need to update migration notes, per new proposed process. I will add a commit later tonight then merge.

bryevdv:
Added migration note, will merge when green.

PR #5637, opened 2016-12-30
Overhaul bokeh.core docs

### LINKED ISSUE #5627: Split up reference docs for bokeh.core
Currently all the modules in `bokeh.core` are generated on a single HTML page. Propose to re-organize the docs structure so that each module generates a separate page.

bryevdv:
issues: fixes #5627 

This PR splits the overly long docs page for `bokeh.core` up into individual pages for each module [*]. It also adds expanded and complete docstrings for all public and private functions, with the exception of:

* `bokeh.compat` I have described this as unmaintained until MEP25, and have also removed the directory from the test reporting

* `bokeh.properties` A follow on PR is planned to move low level things into `property_bases.py` and will update the docs then. 

This PR also:

* removes the `encoder` argument to `serialize_json` as it was unused
* removes `bokeh.core.validation.exceptions` which was unused. 
* fixes a small bug in `export_csv` app example

[*] slight fib, the `bokeh.validation` package is documented as if it was a single module (which it probably should be)

### DIFF
--- a/bokeh/.coveragerc
+++ b/bokeh/.coveragerc
@@ -1,2 +1,2 @@
 [run]
-omit = *tests/*, *sphinxext/*, *compat/mplexporter/*, *sampledata/*,
+omit = *tests/*, *sphinxext/*, *compat/*, *sampledata/*,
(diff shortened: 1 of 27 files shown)


bryevdv:
temp deploy of these docs at http://bokeh.pydata.org/en/test/docs/reference/core.html

bryevdv:
@mattpap any comments on this? I'd like to merge this before splitting/cleaning up/documenting `properties.py`

PR #5641, opened 2016-12-30
Add default_server_port to server_kwargs

### LINKED ISSUE #5629: Bokeh server reports "None" port when there is a port conflict.
When starting a bokeh server on the default port when another server is already running, the port number is incorrectly reported in the error message:
```
$ ../../bin/bokeh serve --show spectrogram
2016-12-28 13:57:32,400 Starting Bokeh server version 0.12.4dev10-77-g450e64c-dirty
2016-12-28 13:57:32,401 Cannot start Bokeh server, port None is already in use
```

souravsingh:
Fixes the issue with "None" reported by Bokeh server. I apologize in advance if the PR is not made as intended.

- [x] issues: fixes #5629 
- [ ] tests added / passed

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -533,6 +533,7 @@ class Serve(Subcommand):
                                                               'unused_session_lifetime_milliseconds',
                                                               'stats_log_frequency_milliseconds',
                                                               'use_xheaders',
+                                                              'default_server_port'
                                                             ]
                           if getattr(args, key, None) is not None }
 
@@ -573,7 +574,8 @@ class Serve(Subcommand):
             address_string = ''
             if server.address is not None and server.address != '':
                 address_string = ' address ' + server.address
-
+         
+          server_kwargs['default_server_port'] = DEFAULT_PORT  
             log.info("Starting Bokeh server on port %d%s with applications at paths %r",
                      server.port,
                      address_string,
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -532,8 +532,7 @@ class Serve(Subcommand):
                                                               'check_unused_sessions_milliseconds',
                                                               'unused_session_lifetime_milliseconds',
                                                               'stats_log_frequency_milliseconds',
-                                                              'use_xheaders',
-                                                              'default_server_port'
+                                                              'use_xheaders'
                                                             ]
                           if getattr(args, key, None) is not None }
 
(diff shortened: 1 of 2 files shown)


bryevdv:
Hi @souravsingh thatnks for the PR. No apologies needed, but this is a little different that I had imagined. I thing the only change that is needed (instead of the two lines you added) is to put:
```
server_kwargs['port'] = DEFAULT_PORT  
```
right before the `with` statement on line 563 of `serve.py` (you might also need to import `DEFAULT_PORT` from wherever it comes from)

bryevdv:
@souravsingh I'm not sure why, but this change is causing the server tests to hang, which I did not expect. I will try to look into it closer next week

bryevdv:
Hi @souravsingh I figured out the problem. I gave you bad advice. The change I told you to make ignores setting the port altogether. Here is a diff that passes tests and also works with my local testing:
```
diff --git a/bokeh/command/subcommands/serve.py b/bokeh/command/subcommands/serve.py
index fa6aca0..81ecb27 100644
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -352,7 +352,7 @@ base_serve_args = (
         metavar = 'PORT',
         type    = int,
         help    = "Port to listen on",
-        default = None
+        default = DEFAULT_SERVER_PORT,
     )),

     ('--address', dict(
diff --git a/bokeh/command/subcommands/tests/test_serve.py b/bokeh/command/subcommands/tests/test_serve.py
index e5c6a80..fdd96b8 100644
--- a/bokeh/command/subcommands/tests/test_serve.py
+++ b/bokeh/command/subcommands/tests/test_serve.py
@@ -11,7 +11,7 @@ import pytest
 import requests

 import bokeh.command.subcommands.serve as scserve
-
+from bokeh.resources import DEFAULT_SERVER_PORT

 def test_create():
     import argparse
@@ -37,7 +37,7 @@ def test_args():
             metavar = 'PORT',
             type    = int,
             help    = "Port to listen on",
-            default = None
+            default = DEFAULT_SERVER_PORT,
         )),

         ('--address', dict(
```
Basically, instead of setting `server_kwargs` we can just set the default on the `--port` argument config. But we need to update the test too, as you can see in the diff. Let me know if I can help or answer any questions.

bryevdv:
@souravsingh Thanks for the PR! There was a small issue with out linter checks, some trailing whitespace on a line. I went ahead and pushed a change to fix that so tests should now pass. I will merge when green!

Issue #5636, opened 2016-12-30
patches incorrectly draws boundaries from GeoJSONDataSource in latest development version on a bokeh server


agoodm:
Today I decided to test bokeh 0.12.4dev13 with an app I am currently developing. It involves plotting US states on a map via the `GeoJSONDataSource`. When I attempt to do so, for the states which have discontinuities in their boundary (eg, their geometry is `MultiPolygon`), the boundaries are drawn incorrectly, as shown here:

![example](https://puu.sh/t5j0B/94615168c2.png)

The code I used is a slightly modified version of the example shown in #4935 
```python
from bokeh.io import output_notebook, show
import json
from bokeh.models import GeoJSONDataSource, HoverTool
from bokeh.plotting import figure, curdoc
from shapely import affinity
from shapely.geometry import mapping, shape

def get_us_coords():
    with open('/Users/goodman/data/boundaries/us_states_simplified_albers.json') as us_states_file:
        geojson = json.loads(us_states_file.read())
        for i, f in enumerate(geojson['features']):
            if f['properties']['NAME'] == 'Alaska':
                geom = affinity.translate(shape(f['geometry']), xoff=1.2e6, yoff=-4.8e6)
                geom = affinity.scale(geom, .45, .45)
                geom = affinity.rotate(geom, 30)
                f['geometry'] = mapping(geom)
            elif f['properties']['NAME'] == 'Hawaii':
                geom = affinity.translate(shape(f['geometry']), xoff=4.75e6, yoff=-1.1e6)
                geom = affinity.scale(geom, 1.15, 1.15)
                geom = affinity.rotate(geom, 30)
                f['geometry'] = mapping(geom)
    return GeoJSONDataSource(geojson=json.dumps(geojson))

geojson = get_us_coords()

x_range = (-3e6, 3e6)
y_range = (-2e6, 1.5e6)

ar = (x_range[1] - x_range[0]) / (y_range[1] - y_range[0])
width = 800

fig = figure(
    plot_width=width, plot_height=int(width / ar),
    x_range=x_range, y_range=y_range,
    x_axis_type=None, y_axis_type=None, min_border=0,
    tools='tap', toolbar_location=None, outline_line_color=None
)
fig.grid.visible = False
fig.add_tools(HoverTool(tooltips='@NAME'))
fig.patches(xs='xs', ys='ys', line_color='#333333', line_width=1, fill_alpha=0.5, source=geojson)
curdoc().add_root(fig)
```
[...]

PR #5642, opened 2016-12-30
GeoJSONDataSource no longer inherits from ColumnDataSource

philippjfr:
As described in https://github.com/bokeh/bokeh/issues/5636, the GeoJSONDataSource has various issues associated with syncing the dynamically computed ``data`` property. This was not an issue previously because before #5544 the ``data`` never ended up roundtripping for reasons that are not quite obvious to me. However I have confirmed that this is not the result of a new issue in the serialization and deserialization introduced by the binary protocol. It seems to be the case that when the JS code sends the unpacked columnar data to Python ‘Nan’ values get interpreted as Nones, and do not get reinterpreted as NaNs on the roundtrip. This might be worth investigating separately.

This PR makes sure that ``GeoJSONDataSource`` no longer directly inherits directly from ``ColumnDataSource``, which means it does not inherit the ``data`` property. On the Python end this is very straightforward because all that has to be sent to the frontend is the geojson representation itself. However on the JS end the ``GeoJSONDataSource`` does have to mirror the ColumnDataSource API without exposing a publicly accessible ``data`` attribute. Therefore I have introduced an intermediate ``ColumnarDataSource`` class, which both the ``ColumnDataSource`` and the ``GeoJSONDataSource`` can inherit from. The PR also avoids the ``Glyph refers to nonexistent column name: xs, ys`` which was generated previously.

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -325,13 +325,12 @@ class ColumnDataSource(DataSource):
 
         self.data._patch(self.document, self, patches, setter)
 
-class GeoJSONDataSource(ColumnDataSource):
+class GeoJSONDataSource(DataSource):
 
     geojson = JSON(help="""
     GeoJSON that contains features for plotting. Currently GeoJSONDataSource can
     only process a FeatureCollection or GeometryCollection.
     """)
-
 
 @abstract
 class RemoteSource(ColumnDataSource):
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -48,7 +48,17 @@ class DataSource(Model):
     A callback to run in the browser whenever the selection is changed.
     """)
 
-class ColumnDataSource(DataSource):
+@abstract
+class ColumnarDataSource(DataSource):
+    """ A baseclass for data source types, which can be mapped onto
+    a columnar format. Not useful to instantiate on its own.
+    """
+
+    column_names = List(String, help="""
+    An list of names for all the columns in this DataSource.
+    """)
+
+class ColumnDataSource(ColumnarDataSource):
     """ Maps names of columns to sequences or arrays.
 
     If the ColumnDataSource initializer is called with a single argument that
@@ -70,9 +80,6 @@ class ColumnDataSource(DataSource):
     """).asserts(lambda _, data: len(set(len(x) for x in data.values())) <= 1,
                  lambda: warnings.warn("ColumnDataSource's columns must be of the same length", BokehUserWarning))
 
-    column_names = List(String, help="""
-    An list of names for all the columns in this DataSource.
-    """)
 
     def __init__(self, *args, **kw):
         """ If called with a single argument that is a dict or
@@ -325,7 +332,7 @@ class ColumnDataSource(DataSource):
 
         self.data._patch(self.document, self, patches, setter)
 
-class GeoJSONDataSource(DataSource):
+class GeoJSONDataSource(ColumnarDataSource):
 
     geojson = JSON(help="""
     GeoJSON that contains features for plotting. Currently GeoJSONDataSource can
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -16,7 +16,7 @@ from ..models import (
     FactorRange, Grid, HelpTool, HoverTool, LassoSelectTool, Legend, LegendItem, LinearAxis,
     LogAxis, PanTool, ZoomInTool, ZoomOutTool, PolySelectTool, ContinuousTicker,
     SaveTool, Range, Range1d, UndoTool, RedoTool, ResetTool, ResizeTool, Tool,
-    WheelPanTool, WheelZoomTool, ColumnDataSource, GlyphRenderer)
+    WheelPanTool, WheelZoomTool, ColumnDataSource, GeoJSONDataSource, GlyphRenderer)
 
 from ..core.properties import ColorSpec, Datetime, value, field
 from ..util.deprecation import deprecated
@@ -518,7 +518,7 @@ def _glyph_function(glyphclass, extra_docs=None):
         is_user_source = kwargs.get('source', None) is not None
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
-        if not isinstance(source, ColumnDataSource):
+        if not isinstance(source, (ColumnDataSource, GeoJSONDataSource)):
             try:
                 # try converting the soruce to ColumnDataSource
                 source = ColumnDataSource(source)
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -16,7 +16,7 @@ from ..models import (
     FactorRange, Grid, HelpTool, HoverTool, LassoSelectTool, Legend, LegendItem, LinearAxis,
     LogAxis, PanTool, ZoomInTool, ZoomOutTool, PolySelectTool, ContinuousTicker,
     SaveTool, Range, Range1d, UndoTool, RedoTool, ResetTool, ResizeTool, Tool,
-    WheelPanTool, WheelZoomTool, ColumnDataSource, GeoJSONDataSource, GlyphRenderer)
+    WheelPanTool, WheelZoomTool, ColumnarDataSource, ColumnDataSource, GlyphRenderer)
 
 from ..core.properties import ColorSpec, Datetime, value, field
 from ..util.deprecation import deprecated
@@ -518,7 +518,7 @@ def _glyph_function(glyphclass, extra_docs=None):
         is_user_source = kwargs.get('source', None) is not None
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
-        if not isinstance(source, (ColumnDataSource, GeoJSONDataSource)):
+        if not isinstance(source, ColumnarDataSource):
             try:
                 # try converting the soruce to ColumnDataSource
                 source = ColumnDataSource(source)
--- a/bokehjs/src/coffee/models/sources/column_data_source.coffee
+++ b/bokehjs/src/coffee/models/sources/column_data_source.coffee
@@ -1,9 +1,7 @@
 import * as _ from "underscore"
 
-import {DataSource} from "./data_source"
-import * as hittest from "../../core/hittest"
+import {ColumnarDataSource} from "./columnar_data_source"
 import {HasProps} from "../../core/has_props"
-import {SelectionManager} from "../../core/selection_manager"
 import {logger} from "../../core/logging"
 import * as p from "../../core/properties"
 import * as serialization from "../../core/util/serialization"
@@ -57,45 +55,6 @@ export patch_to_column = (col, patch) ->
   for i in [0...patch.length]
     [ind, value] = patch[i]
     col[ind] = value
-
-
-# Abstract baseclass for column based data sources, where the column
-# based data may be supplied directly or be computed from an attribute
-export class ColumnarDataSource extends DataSource
-  type: 'ColumnarDataSource'
-
-  @define {
-    column_names: [ p.Array, [] ]
-  }
-
-  @internal {
-    selection_manager: [ p.Instance, (self) -> new SelectionManager({source: self}) ]
-    inspected:         [ p.Any ]
-    _shapes:      [ p.Any, {}]
-  }
-
-  get_column: (colname) ->
-    return @data[colname] ? null
-
-  columns: () ->
-    # return the column names in this data source
-    return _.keys(@data)
-
-  get_length: (soft=true) ->
-    lengths = _.uniq((val.length for _key, val of @data))
-
-    switch lengths.length
-      when 0
-        return null # XXX: don't guess, treat on case-by-case basis
-      when 1
-        return lengths[0]
-      else
-        msg = "data source has columns of inconsistent lengths"
-        if soft
-          logger.warn(msg)
-          return lengths.sort()[0]
-        else
-          throw new Error(msg)
 
 
 # Datasource where the data is defined column-wise, i.e. each key in the
--- a/bokehjs/src/coffee/models/sources/column_data_source.coffee
+++ b/bokehjs/src/coffee/models/sources/column_data_source.coffee
@@ -58,29 +58,28 @@ export patch_to_column = (col, patch) ->
     [ind, value] = patch[i]
     col[ind] = value
 
-# Datasource where the data is defined column-wise, i.e. each key in the
-# the data attribute is a column name, and its value is an array of scalars.
-# Each column should be the same length.
-export class ColumnDataSource extends DataSource
-  type: 'ColumnDataSource'
 
-  initialize: (options) ->
-    super(options)
-    [@data, @_shapes] = serialization.decode_column_data(@data)
+# Abstract baseclass for column based data sources, where the column
+# based data may be supplied directly or be computed from an attribute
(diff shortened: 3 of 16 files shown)


mattpap:
I wouldn't go as far as saying that this PR fixes anything, just avoids running broken code. The underlying bug aside, `GeoJSONDataSource` needs to be redesigned, so that it doesn't set any public `data` property. I'm not sure if `GeoJSONDataSource` has to be a `ColumnDataSource`, but knowing bokeh, the rest of the code works only with `ColumnDataSource`.

philippjfr:
Yes, thinking about it more, this "fix", is more of a hack. I'm not sure what the cleanest way of actually fixing it is. I did try having GeoJSONDataSource inherit directly from DataSource, which worked with some minor changes to plotting and renderer code, but obviously issued a warning about the data property not existing on the Python model. I'll try to investigate how much work it would be to have the coffeescript implementation not to have to subclass from ColumnDataSource, and make data an internal attribute.

philippjfr:
@mattpap @bryevdv @birdsarah I've now repurposed this PR such that the GeoJSONDataSource no longer inherits from ColumnDataSource, which avoids issues with the ``data`` roundtripping. Would be great if you could all have a look to see if you think there's any issues with this approach. We should probably also check whether the other data sources that also inherit from ``ColumnDataSource`` have similar issues causing the ``data`` to be sent back and forth.

Finally it's a bit worrying that NaNs originating from JS seem to turn into Nones when sent to Python, which are then not correctly turned back into NaNs when they are sent back.

bokeh/plotting/helpers.py
```diff
@@ -518,7 +518,7 @@ def func(self, **kwargs):
         is_user_source = kwargs.get('source', None) is not None
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
-        if not isinstance(source, ColumnDataSource):
+        if not isinstance(source, (ColumnDataSource, GeoJSONDataSource)):
```
mattpap: I presume this should be `isinstance(source, DataSource)`.

bokeh/models/sources.py
```diff
@@ -325,14 +325,13 @@ def patch(self, patches, setter=None):
 
         self.data._patch(self.document, self, patches, setter)
 
-class GeoJSONDataSource(ColumnDataSource):
+class GeoJSONDataSource(DataSource):
```
mattpap: Class structure must be the same as in bokehjs, so you have to introduce `ColumnarDataSource` and set bases appropriately.

agoodm:
Just curious, but aside from being a temporary solution to #5636 , what does preventing `GeoJSONDataSource` from inhering directly from `ColumnDataSource` accomplish? I thought the main purpose of the former is to exist as a convenience class which parses information contained in a GeoJSON file on the client-side to something that fundamentally should act just like a `ColumnDataSource`, thus allowing for seamless use of standard glyph methods. In my opinion, this also means that it should have a public `data` property as a regular `ColumnDataSource` does since this offers a more consistent API for users. For the case of `GeoJSONDataSource`, this is especially true for processing callbacks from within a bokeh server, since the data attribute ends up containing other useful information from the GeoJSON file in addition to the polygon boundaries as well. 

Anyways with that in mind I thought it would be interesting to test the example code I used for #5636 again for 0.12.4dev13, except this time I would manually convert the information from the GeoJSON file into a `ColumnDataSource`. Thankfully in python this wasn't too challenging to do with geopandas:

```python
from bokeh.models import ColumnDataSource, HoverTool
import geopandas as gpd

def reducer(x0, x1):
    nan = float('nan')
    return x0 + [nan] + x1


def get_xs(poly):
    return list(poly.boundary.xy[0])
    
    
def get_ys(poly):
    return list(poly.boundary.xy[1])
    

def convert_geojson_to_column(geojson_file):
    df = gpd.read_file(geojson_file)
    xs, ys = [], []
    for geom in df.geometry:
        if geom.type == 'MultiPolygon':
            xs.append(reduce(reducer, [get_xs(poly) for poly in geom]))
            ys.append(reduce(reducer, [get_ys(poly) for poly in geom]))
        else:
            xs.append(get_xs(geom))
            ys.append(get_ys(geom))
            
    df['xs'] = xs
    df['ys'] = ys
    del df['geometry']
    return ColumnDataSource(data=df)


geojson = convert_geojson_to_column('../us_states_simplified_albers.json')
x_range = (-3e6, 3e6)
y_range = (-2e6, 1.5e6)

ar = (x_range[1] - x_range[0]) / (y_range[1] - y_range[0])
width = 800

fig = figure(
    plot_width=width, plot_height=int(width / ar),
    x_range=x_range, y_range=y_range,
    x_axis_type=None, y_axis_type=None, min_border=0,
    tools='tap,save', outline_line_color=None
)
fig.grid.visible = False
fig.add_tools(HoverTool(tooltips='@NAME'))
fig.patches(xs='xs', ys='ys', line_color='#333333', line_width=1, fill_alpha=0.5, source=geojson)
curdoc().add_root(fig)
```
result

I have also observed that python isn't interpreting the received NaN's from JS as None, so it does indeed seem that the problem is likely exclusive to `GeoJSONDataSource`.

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -58,30 +58,29 @@ export patch_to_column = (col, patch) ->
...
-# Each column should be the same length.
-export class ColumnDataSource extends DataSource
-  type: 'ColumnDataSource'
 
-  initialize: (options) ->
-    super(options)
-    [@data, @_shapes] = serialization.decode_column_data(@data)
+# Abstract baseclass for column based data sources, where the column
+# based data may be supplied directly or be computed from an attribute
+export class ColumnarDataSource extends DataSource
```
mattpap: This should be moved to its own file.

mattpap:
>  if you think there's any issues with this approach.

The problem is that bokeh/bokehjs assumes `ColumnDataSource` in a lot of places (see e.g. charts), so by making `GeoJSONDataSource` a `ColumnarDataSource` and not updating everything else to `ColumnarDataSource`, you won't be able to use `GeoJSONDataSource` across bokeh. Changing everything from `ColumnDataSource` to `ColumnarDataSource` is possible, but I'm not sure if we want to do this. Perhaps we should have a broader discussion on the following subjects:

1. What is the usefulness of bokehjs `GeoJSONDataSource`, if the same can be done in Python? After all, we don't have `PandasDataSource`, but do the conversion in Python.
2. What is the point of `DataSource` abstract base class, if it's not used in `isinstance` checks nor as an actual abstract base in the API (`ColumnDataSource` is used as such). bokeh/bokehjs assumes columnar data sources everywhere, but `DataSource` is supposed to be more generic than this. If we actually support non-columnar data sources, I would like to see some code asserting this. Otherwise, lets make `DataSource` a `ColumnarDataSource`.

@philippjfr, I strongly suggest to update TypeScript API. You should do this anyway, but updating it will help you to design the API.

philippjfr:
>Just curious, but aside from being a temporary solution to #5636 , what does preventing GeoJSONDataSource from inhering directly from ColumnDataSource accomplish?

My thinking here was this, since ``GeoJSONDataSource`` does not resolve the geojson data until it is sent client side being able to modify the data directly leads to poorly defined behaviors that can lead to subtle bugs. Specifically ``data`` can't actually be modified until the client-side has converted the data. Once that happens the ``data`` can potentially be modified safely but doing so will mean that the ``geojson`` and ``data`` representation can be out of sync leading to unexpected behaviors, where modifying both the ``geojson`` and ``data`` will effectively override the others. IMO it's therefore cleaner to either provide a utility that converts geojson to a regular column format automatically, i.e. something like a ``from_geojson`` method that exists both in the JS/TS and Python API. The alternative solution which is what implemented here is the ``GeoJSONDataSource`` that does not allow directly modifying the ``data``.

That said I now think that's an issue that needs more discussion and we should leave the hybrid approach that allows modifying both ``data`` and ``geojson`` for now. I'll just try to solve the immediate problem of having JS ``NaNs`` turn into ``None`` while roundtripping.

bryevdv:
@philippjfr @mattpap Agree there is a need for a wider design discussion. But in order not to delay `0.12.4` I think if there is a quick "band aid" fix that can be applied, that it should be. 

----

Brain dump just since thoughts are in mind right now:

Most of the data that Bokeh uses is tabular, which is why CDS are everywhere, and why we convert Pandas automatically (same format, basically). But there are non-columnar data that are worth considering separately, names Geo data, and Graph/Netwotk data. Since a goal has been expressed to have these kinds of data be able to drive existing glyphs, that implies some form of adaptation. That could happen:

* early, on the python side, shoehorn everything into CDS as far as BokehJS is concerned. 
* late, on the BokehJS side during initialization
* later, on the BokehJS side, only during access

If it can happen efficiently, I think the last case is idea. That is, Geo and Graph data sources appear identical to python and JS API, and updating either syncs as expected. Specialized Geo or Graph renderers might take advantage of specific qualities of the specialized data sources. But regular glyphs can also have some way of interacting with these data sources, basically by having their be some adaptive interface that can make them look like CDS at render time (to the extent possible) while pulling data out of the specialized Geo or Graph data source as late as possible.

bryevdv:
OK, FYI re scheduling I'd like to release a `0.12.4` RC tomorrow morning. If this fix is not in bythen I would consider it the one code-related change that could go in afterwards.

philippjfr:
>But in order not to delay 0.12.4 I think if there is a quick "band aid" fix that can be applied, that it should be.

I've just investigated whether there's an easy way to send NaNs back from JS, but it seems like that would bloat the serialization/deserialization code even further. Doing the equivalent to the Python code and converting NaNs and Infinity to strings would work but isn't really correct either and may have some performance penalty.

If you agree we can push forward with the bandaid implemented in this PR. I'd switch to making ColumnarDataSource the baseclass and using that for isinstance checks throughout. That said I quickly want to clarify with what you are saying, the last option makes sense to me:

>later, on the BokehJS side, only during access

But later you say:

>That is, Geo and Graph data sources appear identical to python and JS API, and updating either syncs as expected.

Just wanted to make sure I'm clear on what you mean by that. All these proposals differ in whether you can access and modify the custom format and the common, column-based representation. 
The way I see it the three approaches you outlined have the following pros and cons:

* early: Can only access common format, can modify both formats if ``update_from_geojson`` methods are implemented.

* late: Can access and modify both representations, which can lead to surprising behavior as described above. If ``data`` is private, only the custom representation can be accessed in Python.

* later: Unless getter is implemented for Python and JS can only access common format in JS, limited to modifying the custom representation.

In the **later** approach as you outlined it, you would therefore only be able to access the data in the column based format in JS, and only be able to modify the custom representation, i.e. the geojson. If that's the eventual goal then the **late** approach in this PR, making the ``data`` private should, be functionally identical, since any change to the custom representation will trigger recomputing the data.

bryevdv:
> In the later approach as you outlined it, you would therefore only be able to access the data in the column based format in JS, and only be able to modify the custom representation, i.e. the geojson. 

I'm not quite sure that's an accurate interpretation. In the **later** approach, I Image: data source objects that are not columnar. The specialized non-columnar interface is available for both read or write or update in Python and JS (both sides are identical, like all current models). Only additionally, if one of these sources is used where a column data source is expected (say a circle glyph) then there can optionally be some mechanism to adapt the specialized non-columnar data source into something that the circle glyph can work with. 

> If you agree we can push forward with the bandaid implemented in this PR. I'd switch to making ColumnarDataSource the baseclass and using that for isinstance checks throughout. 

Broadly, yes, I will defer to you and @mattpap on the specific details of the best bandaid for the short term.

philippjfr:
>I'm not quite sure that's an accurate interpretation. In the later approach, I Image: data source objects that are not columnar. The specialized non-columnar interface is available for both read or write or update in Python and JS (both sides are identical, like all current models). Only additionally, if one of these sources is used where a column data source is expected (say a circle glyph) then there can optionally be some mechanism to adapt the specialized non-columnar data source into something that the circle glyph can work with. Or put another way: there can possibly be read-only "columnar views" on these other data sources.

Okay sorry I wasn't quite clear enough, that was my understanding as well as long as these "columnar views" are read-only.

>Broadly, yes, I will defer to you and @mattpap on the specific details of the best bandaid for the short term.

Okay  @mattpap, if you think this is a reasonable bandaid I'll push forward with your suggestions for this PR so it can be merge it in time for 0.12.4rc tomorrow.

mattpap:
This looks good.

philippjfr:
@mattpap Might be worth looking at the typescript changes in detail, I basically just cargo-culted the existing the examples. I suppose I should also add ``GeoJSONDataSource`` typings but wasn't quite clear on whether private attributes should be typed.

mattpap:
> but wasn't quite clear on whether private attributes should be typed.

Currently only (model) properties are typed.

philippjfr:
@mattpap Is this correct? What type should I use for the geojson?

```
  export var GeoJSONDataSource: { new(attributes?: IGeoJSONDataSource, options?: ModelOpts): GeoJSONDataSource };
  export interface GeoJSONDataSource extends ColumnarDataSource, IGeoJSONDataSource {}
  export interface IGeoJSONDataSource extends IColumnarDataSource {
    geojson?: JsObj;
  }
```

mattpap:
> `geojson?: JsObj;`

That makes sense, though `JsObj` is a local type alias to `document.d.ts`. You could move it to `types.d.ts`.

bryevdv:
@philippjfr is this `ready` ?

philippjfr:
@bryevdv Yes, I think it's ready once you and @mattpap are done reviewing.

bryevdv:
do the `charts` etc. need to be generalized to `ColumnardataSource` in this PR? I think this will clash with another PR. I'd prefer to make the minimal change now, and make an issue to make further generalizations later.

philippjfr:
>do the charts etc. need to be generalized to ColumnardataSource in this PR?

Up to you, happy to revert for now.

bryevdv:
This LGTM except for changes under `bokeh/charts` which I would prefer to revert for now. Charts generate data sources internally so unless there is internal chart code to use a more general source, there is no need to make those instance props more general

bryevdv:
cancelled current build

bryevdv:
Also in lieu of tests right now screen shots in this PR to document the fix would be helpful

philippjfr:
I used this minimal example with [this US geojson](https://raw.githubusercontent.com/fabianmoronzirfas/IDJSMap/master/indesign/world.geo.json/countries/USA.geo.json)

```python
import json
from bokeh.models import GeoJSONDataSource, HoverTool
from shapely import affinity
from shapely.geometry import mapping, shape

def get_us_coords():
    with open('US.json') as us_states_file:
        geojson = json.loads(us_states_file.read())
    return GeoJSONDataSource(geojson=json.dumps(geojson))

geojson = get_us_coords()

x_range = (-3e6, 3e6)
y_range = (-2e6, 1.5e6)

ar = (x_range[1] - x_range[0]) / (y_range[1] - y_range[0])
width = 800

fig = figure(
    plot_width=width, plot_height=int(width / ar),
    x_axis_type=None, y_axis_type=None, min_border=0,
    tools='tap', toolbar_location=None, outline_line_color=None
)
fig.grid.visible = False
fig.add_tools(HoverTool(tooltips='@NAME'))
fig.patches(xs='xs', ys='ys', line_color='#333333', line_width=1, fill_alpha=0.5, source=geojson)
curdoc().add_root(fig)
```

Before this PR:

screen shot 2017-01-03 at 4 51 20 pm

After this PR:

screen shot 2017-01-03 at 4 53 31 pm

PR #5654, opened 2017-01-04
Misc docs and tests for 0.12.4 release

### LINKED ISSUE #4877: Need to include ImageSource to docs
They are currently not in reference docs.

My bad.

bryevdv:
issues: fixes #4877


philippjfr:
@bryevdv I've pushed a small extra section to the server docs. It doesn't fit in very well in any of the other sections but I'm still not sure that's the best place to put it. Please modify and/or move as appropriate.

bryevdv:
OK this is ready. I'd like to have added more but it will have to go in later. Will merge in the morning absent comment.

PR #5671, opened 2017-01-08
Use custom code instead of gulp-mocha for running tests

### LINKED ISSUE #5659: gulp test can fail without clear indication
An example from PR #5630:
```
(...)
      ✓ get_width_height should return a constrained width if plot is landscape oriented
[12:20:20] 'test' errored after 14 s
[12:20:20] TypeError in plugin 'gulp-mocha'
Message:
    Cannot set property 'width' of undefined
Details:
    domain: [object Object]
    domainThrown: true
Stack:
TypeError: Cannot set property 'width' of undefined
(...)
  762 passing (13s)
  24 pending
```
To me it's "all green". However, `echo $?` gives `1` indicating a failure and tests actually fail on travis ci. The failure is real, however related to badly written tests (which deserve an issue and a big fat rant on its own) than a bug in the code.  Also note that it's easy to overlook this failure, because the output from the test runner is very sparse and often unreadable.

Expected behavior: big red indication at the end of output that tests failed.

I'm marking this as discussion and not a task, because at this point I think that simply improving things on top of mocha isn't desirable and we should look for a proper test runner (I don't have high hopes for finding one).

mattpap:
gulp-mocha doesn't work and is unmaintained, so just use `child_process.spawn()` and run mocha natively.

fixes #5659

### DIFF
--- a/bokehjs/gulp/tasks/test.coffee
+++ b/bokehjs/gulp/tasks/test.coffee
@@ -1,43 +1,55 @@
 gulp = require "gulp"
-mocha = require "gulp-mocha"
+gutil = require "gulp-util"
+through = require "through"
+cp = require "child_process"
 
 paths = require "../paths"
 utils = require "../utils"
 
-gulp.task "test", ["defaults:generate"], ->
-  gulp.src ["./test", "./test/all.coffee"], read: false
-    .pipe mocha()
+mocha = (options={}) ->
+  return through(
+    (file) ->
+      if @_files? then @_files.push(file.path) else @_files = [file.path]
+    () ->
+      proc = cp.spawn("node_modules/.bin/mocha",
+                      ["--compilers", "coffee:coffee-script/register", "./test/index.coffee"].concat(@_files),
+                      {stdio: 'inherit'})
+
+      proc.on "error", (err) =>
+        @emit("error", new gutil.PluginError("mocha", err))
+
+      proc.on "exit", (code) =>
+        if code != 0
+          @emit("error", new gutil.PluginError("mocha", "tests failed"))
+        else
+          @emit("end")
+  )
+
+gulp.task "test", ["defaults:generate"], () ->
+  gulp.src(["./test/all.coffee"]).pipe(mocha())
 
 gulp.task "test:client", ->
-  gulp.src ["./test", "./test/client.coffee"], read: false
-    .pipe mocha()
+  gulp.src(["./test/client.coffee"]).pipe(mocha())
 
 gulp.task "test:core", ->
-  gulp.src ["./test", "./test/core"], read: false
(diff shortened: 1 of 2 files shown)


mattpap:
Now layout failures are much more prominent, e.g.:
```
$ gulp test
(...)
  Plot
    PlotView
      ✓ render should set the appropriate positions and paddings on the element when it is mode box
      ✓ should call suggest value with the model height and width if sizing_mode is scale_both
      - get_height should return the height from the aspect ratio
      ✓ get_width should return the width from the aspect ratio
      1) "after each" hook
      - get_width should return the width from the aspect ratio

  PlotCanvas
    2) "before each" hook
    3) "after each" hook for "should set the sizing_mode to box by default"


  543 passing (9s)
  18 pending
  3 failing

  1) Plot PlotView "after each" hook:
     Uncaught Error: unknown edit variable
    at Solver.suggestValue (src/vendor/kiwi/kiwi.js:1549:23)
    at Solver.exports.Solver.Solver.suggest_value (build/js/tree/core/layout/solver.js:77:28)
    at PlotView.exports.LayoutDOMView.LayoutDOMView.render (build/js/tree/models/layouts/layout_dom.js:106:15)
    at PlotView.exports.PlotView.PlotView.render (build/js/tree/models/plots/plot.js:35:35)
    at PlotView.<anonymous> (build/js/tree/models/layouts/layout_dom.js:72:34)
    at PlotView.<anonymous> (build/js/tree/core/events.js:199:22)
    at PlotView.<anonymous> (node_modules/underscore/underscore.js:890:21)
    at triggerEvents (build/js/tree/core/events.js:241:43)
    at triggerApi (build/js/tree/core/events.js:227:13)
    at eventsApi (build/js/tree/core/events.js:44:18)
    at Solver.exports.Events.trigger (build/js/tree/core/events.js:216:5)
    at Document.exports.Document.Document._resize (build/js/tree/document.js:200:29)
    at Document.exports.Document.Document.resize (build/js/tree/document.js:158:14)
    at Timeout._onTimeout (node_modules/underscore/underscore.js:768:19)
  

  2) Plot "before each" hook:
     TypeError: Cannot read property 'call' of undefined
  

  3) PlotCanvas "after each" hook for "should set the sizing_mode to box by default":
     TypeError: CanvasView.prototype.get_ctx.restore is not a function
    at Object.unstub_canvas (test/utils.coffee:45:34)
    at Context.<anonymous> (test/models/plots/plot_canvas.coffee:34:11)
  



  4) "before each" hook
[12:17:32] 'test' errored after 15 s
[12:17:32] Error in plugin 'mocha'
Message:
    tests failed
```
Note that layout test failures are non-deterministic and there are several kinds of them, so you may get different output or non at all.

PR #5698, opened 2017-01-10
allow auto for non/selection glyphs

### LINKED ISSUE #2414: use [non]selection_glyph="auto" to generate automatically, not None
reference: http://stackoverflow.com/questions/30915580/how-can-i-configure-the-taptool-so-it-does-not-hide-other-data-points-when-click

Should require `selection_glyph="auto"` and `non selection_glyph="auto"` to create the standard default highlighting (these values should be the new defaults). Setting to `None` should be reserved for "no highlighting".

bryevdv:
issues: fixes #2414 

Makes it possible to specify `selection_glyph=None` to mean no policy (i.e. use regular glyph)

Default behaviors are unchanged.

### DIFF
--- a/bokeh/models/renderers.py
+++ b/bokeh/models/renderers.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 from ..model import Model
 from ..core.enums import RenderLevel
 from ..core.properties import abstract
-from ..core.properties import String, Enum, Instance, Float, Bool, Override
+from ..core.properties import Auto, Bool, Either, Enum, Float, Instance, Override, String
 from ..core import validation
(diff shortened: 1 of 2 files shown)


bryevdv:
E.g. makes this possible:

```
import numpy as np

N = 4000
colors = [
]

p = figure(tools="box_select")

r = p.scatter(x, y, radius=radii,
          line_color=None, selection_color="red")

r.nonselection_glyph = None

show(p)  # open a browser
```

screen shot 2017-01-10 at 5 38 14 pm

bryevdv:
Ping @mattpap seems ok? Not super important but its irritated me for ages, this seem more consistent and correct.

bryevdv:
This seems fairly un-controversial, so I will merge it. If there are any objections later it is easy enough to revert.

PR #5700, opened 2017-01-11
safe-tag-fix

draperjames:
CoffeeScript variable substitution requires double quotes, not single quotes.

### DIFF
--- a/bokehjs/src/coffee/core/util/templating.coffee
+++ b/bokehjs/src/coffee/core/util/templating.coffee
@@ -33,7 +33,7 @@ export replace_placeholders = (string, data_source, i, special_vars = {}) ->
       replacement = "???"
     else
       if format == 'safe'
-        return '#{prefix}#{value}'
+        return "#{prefix}#{value}"
       else if format?
         replacement = Numbro.format(value, format)
       else


bryevdv:
@draperjames can you attach a sample script and a screenshot to this PR as documentation of the fix?

draperjames:
Example code:
```python
%env BOKEH_RESOURCES=inline
from bokeh.plotting import ColumnDataSource, figure
from bokeh.models import HoverTool, Range1d

output_notebook()
fig = figure(tools=[HoverTool(tooltips=[("html", '@html{safe}')])])

fig.quad(left="left", top="top", bottom="bottom", right="right",
         source=ColumnDataSource({"left": [1,3], "bottom": [1,3],
                                  "right": [2,4], "top": [2,4],
                                  "html":["<b>I'm bold</b>", "<span style='color:red;font-size:32px;'>BIG RED TEXT</span>"]}))
show(fig)
```
Screenshot:

screen_cap636197295849170272

Here is an example [notebook](https://nbviewer.jupyter.org/github/draperjames/PythonRef/blob/master/Bokeh%20HoverTool%20Custom%20ToolTips.ipynb) as well.

bryevdv:
@draperjames great, thanks! Merging now

draperjames:
No prob @astraw38 deserves all the credit for the [fix](https://github.com/bokeh/bokeh/pull/5460#issuecomment-271893036).

PR #5717, opened 2017-01-13
Automatically destroying servers in the notebook embed example

jlstevens:
This PR aims to address one of the issues raised in #3461 regarding the new [notebook embed](https://github.com/bokeh/bokeh/blob/master/examples/howto/server_embed/notebook_embed.ipynb) example, namely the problem of cleaning up servers when cells are deleted/re-executed. The approach shown here has been developed with a lot of helpful input from  @bryevdv.

Right now I would consider this PR to illustrate one possible implementation that can work. It does not track cell deletion/replacement in a particularly elegant way and I am still hoping to find a better approach.

## What it does

When a cell with output from ``show_app`` is deleted or replaced, a callback is fired using ``kernel.execute`` to destroy the corresponding server sessions (and then stop it).

## Approach

1. The ``watch_server_cells`` is called at the top of the notebook. This sets up a ``MutationObserver`` on 'notebook-container' div, allowing it to detect when any DOM nodes with ``class='bokeh_class'`` are deleted. To detect removal of DOM nodes, the ``MutationObserver`` needs to be set on a DOM element holding those nodes (i.e I could not get this working by setting the mutation observer on the cells themselves).

2. Invoking ``show_app`` uses the ``server_cell`` function to wrap the HTML output in a div that holds a freshly generated uuid. This uuid is associated with the server instance in the ``_state.uuid_to_server`` dictionary.

3. When these cells containing the wrapping div (with a class of 'bokeh_class') are deleted, ``kernel.execute`` is used to pass the uuid to ``bokeh.io._destroy_server``. This can then lookup the server instance and perform the necessary clean up.

## Outstanding issues

* Running 'Edit -> Undo Delete Cells' in the notebook menu restores a div associated with a server that will have been destroyed. This is why ``_destroy_server`` can receive a uuid corresponding to a  server that has already been destroyed.
* I don't think the clean up code is actually stopping the server instances properly. After running 'Undo Delete Cells', the app is still responsive! I understand it might be waiting for the tornado event loop to terminate?
* The ``MutationObserver`` is on a DOM node nearer the root than is ideal and the callback is invoked a lot, especially using ``subtree:true``.

It might be worth asking the notebook devs how they would recommend detecting the removal/update of cells. At any rate, I think the prospect of being able to embed the bokeh server in notebook cells is very exciting!

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -644,7 +644,7 @@ def _destroy_server(div_id):
     Given a uuid id of a div removed or replaced in the Jupyter
     notebook, destroy the corresponding server sessions and stop it.
     '''
-    server = _state.uuid_to_server.pop(div_id, None)
+    server = _state.uuid_to_server.get(div_id, None)
     if server is None:
         logger.debug("No server instance found for uuid: %r" % div_id)
         return
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -319,7 +319,7 @@ def _show_notebook_app_with_state(app, _state, app_path, notebook_url):
                     allow_websocket_origin=[notebook_url])
     server.start()
     script = autoload_server(model=None, url='http://127.0.0.1:%d' % server.port)
-    display(HTML(server_cell(server, script)))
+    display(HTML(_server_cell(server, script)))
 
 
 def _show_with_state(obj, state, browser, new, notebook_handle=False):
@@ -629,7 +629,7 @@ def _remove_roots(subplots):
         if sub in doc.roots:
             doc.remove_root(sub)
 
-def server_cell(server, script):
+def _server_cell(server, script):
     '''
     Wraps a script returned by autoload_server in a div that allows cell
     destruction/replacement to be detected.
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -652,7 +652,6 @@ def _destroy_server(div_id):
     try:
         for session in server.get_sessions('/'):
             session.destroy()
-        server.stop()
 
     except Exception as e:
         logger.debug("Could not destroy server for id %r: %s" % (div_id, e))
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -22,6 +22,7 @@ import io
 import json
 import os
 import warnings
+import uuid
 
 # Third-party imports
 
@@ -609,3 +610,14 @@ def _remove_roots(subplots):
     for sub in subplots:
         if sub in doc.roots:
             doc.remove_root(sub)
+
+def server_cell(server, script):
+    '''
+    Wraps a script returned by autoload_server in a div that allows cell
+    destruction/replacement to be detected.
+    '''
+    divid = uuid.uuid4().hex
+    _state.uuid_to_server[divid] = server
+    div_html = "<div class='bokeh_class' id='{divid}'>{script}</div>'"
+    return div_html.format(script=script, divid=divid)
+
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -39,6 +39,10 @@ from .util.notebook import load_notebook, publish_display_data, get_comms
 from .util.string import decode_utf8
 from .util.serialization import make_id
 
+from .application import Application
+from .server.server import Server
+from tornado.ioloop import IOLoop
+
 #-----------------------------------------------------------------------------
 # Globals and constants
 #-----------------------------------------------------------------------------
@@ -260,7 +264,8 @@ def curstate():
     '''
     return _state
 
-def show(obj, browser=None, new="tab", notebook_handle=False):
+def show(obj, browser=None, new="tab", notebook_handle=False,
+                            app_path="/", notebook_url="127.0.0.1:8888"):
     ''' Immediately display a plot object.
 
     In an IPython/Jupyter notebook, the output is displayed in an output
@@ -299,9 +304,22 @@ def show(obj, browser=None, new="tab", notebook_handle=False):
         an IPython/Jupyter notebook.
 
     '''
+    if isinstance(obj, Application):
+        return _show_notebook_app_with_state(obj, _state, app_path, notebook_url)
+
     if obj not in _state.document.roots:
         _state.document.add_root(obj)
     return _show_with_state(obj, _state, browser, new, notebook_handle=notebook_handle)
+
+
+def _show_notebook_app_with_state(app, _state, app_path, notebook_url):
+    from IPython.display import HTML, display
+    loop = IOLoop.current()
+    server = Server({'/': app}, io_loop=loop, port=0, host='*',
+                    allow_websocket_origin=[notebook_url])
+    server.start()
+    script = autoload_server(model=None, url='http://127.0.0.1:%d' % server.port)
+    display(HTML(server_cell(server, script)))
 
 
 def _show_with_state(obj, state, browser, new, notebook_handle=False):
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -621,3 +621,20 @@ def server_cell(server, script):
     div_html = "<div class='bokeh_class' id='{divid}'>{script}</div>'"
     return div_html.format(script=script, divid=divid)
 
+def _destroy_server(div_id):
+    '''
+    Given a uuid id of a div removed or replaced in the Jupyter
+    notebook, destroy the corresponding server sessions and stop it.
+    '''
+    server = _state.uuid_to_server.pop(div_id, None)
+    if server is None:
+        logger.debug("No server instance found for uuid: %r" % div_id)
+        return
+
+    try:
+        for session in server.get_sessions('/'):
+            session.destroy()
+        server.stop()
+
+    except Exception as e:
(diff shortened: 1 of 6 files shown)


bokeh/io.py
```diff
@@ -609,3 +610,31 @@ def _remove_roots(subplots):
     for sub in subplots:
         if sub in doc.roots:
             doc.remove_root(sub)
+
+def server_cell(server, script):
```
bryevdv: This can probably be private, or if it is public, `bokeh.embed` is maybe a better place

bokeh/io.py
```diff
@@ -609,3 +610,31 @@ def _remove_roots(subplots):
...
+    '''
+    server = _state.uuid_to_server.pop(div_id, None)
+    if server is None:
+        logger.debug("No server instance found for uuid: %r" % div_id)
+        return
+
+    try:
+        for session in server.get_sessions('/'):
+            session.destroy()
+        server.stop()
```
bryevdv: does cell-undelete work with the `server.stop` ?

bokeh/util/notebook.py
```diff
@@ -118,3 +118,37 @@ def get_comms(target_name):
...
+               var destroyed_id = bokeh_selector[0].id;
+               IPython.notebook.kernel.execute("from bokeh import io;"
+                                               + "io._destroy_server('"
+                                               + destroyed_id + "')");
+               console.log('Destroying server with id:' + destroyed_id);
+             }
+          }
+      }
+  }
+});
```
bryevdv: ping @minrk @damianavila @nickB Comments on this would be very welcome. we need a way to reliably detect when an output cell has been deleted (e.g. re-execution, or actual deletions) in order to destroy bokeh server sessions that are no longer visible. Is there a better way to do this? What about in JupyterLab?

bokeh/util/notebook.py
```diff
@@ -118,3 +118,37 @@ def get_comms(target_name):
...
+observer.observe(target, { childList: true, subtree:true });
+</script>
+"""
+
+def watch_server_cells():
+    """
+    Installs a MutationObserver that detects deletion of cells using
+    io.server_cell to wrap the output.
+    """
+    publish_display_data({'text/html': mutation_observer}, source='bokeh')
```
bryevdv: as long as we are making this a function, let's parameterize the code to pass to `kernel.execute`

bokeh/io.py
jlstevens: Happy to move it! 

This only needs to be used by ``show_app``so it can be made private wherever that goes. Where would you like ``show_app`` to end up?

bokeh/util/notebook.py
jlstevens: Agreed.

bryevdv:
This is a great start! Just FYI this the current implementation of `show` 
```
def show(obj, browser=None, new="tab", notebook_handle=False):
    if obj not in _state.document.roots:
        _state.document.add_root(obj)
    return _show_with_state(obj, _state, browser, new, notebook_handle=notebook_handle)
```
I have this in mind for making it so `show` works with apps too:

```
def show(obj, browser=None, new="tab", notebook_handle=False, app_path="/"):

    # need some sanity checks here, not all arg combinations make sense now

    if isinstance(obj, Application): 
        return _show_notebook_app_with_state(obj, _state, app_path)

    if obj not in _state.document.roots:
        _state.document.add_root(obj)

    return _show_with_state(obj, _state, browser, new, notebook_handle=notebook_handle)
```

The users could just do:
```
show(myapp)
```
I think providing an `app_path` is important, because then if people do:
```
show(app1, app_path="app1")

show(app2, app_path="app2")
```
I believe we can monkey-patch `show` when `bokeh serve notebook.ipynb` is executed. Meaning: full prototype in the notebook, trivial scalable deploy by running `bokeh serve` on the notebook!

bokeh/io.py
bryevdv: I think `show_app` should be the `_show_notebook_app_with_state` outlined below, probably

bryevdv: let's just make this private `_server_cell` for now

jlstevens:
Extending ``show`` as you suggest does seem to be the nicest approach.

> I believe we can monkey-patch show such that when ``bokeh serve notebook.ipynb`` is executed, apps are set up on the app_paths passed in. Meaning: full prototype in the notebook, trivial scalable deploy by running bokeh serve on the notebook!

That would be really cool!

bokeh/util/notebook.py
damianavila: I think there is an event for appending things to the output area: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/outputarea.js#L668, but I don't see an event for the removal... @minrk I am missing some other thing I don't know about? About JupyterLab, lets ping @blink1073...

damianavila: Btw, use Jupyter.notebook instead of IPython.notebook... and even more, are you using kernel.execute... what about using comms as you used above to keep things consistent.

blink1073: If you have a handle on the `NotebookModel` object, its `cells` are observable.

Something like `nbmodel.cells.changed.connect(sender: ICellModel, args: ObservableVector.IChangedArgs<ICellModel>`

https://github.com/jupyterlab/jupyterlab/blob/bcf78198f1220257a92ff95523afc86f7af9e816/src/common/observablevector.ts#L37
https://github.com/jupyterlab/jupyterlab/blob/bcf78198f1220257a92ff95523afc86f7af9e816/src/notebook/notebook/model.ts#L170

bryevdv: @damianavila right I think we knew about the cell added event, but we really want to know about when the cells go away.

@blink1073 how do you get handles on `NotebookModel` ?

blink1073: Actually, if you are a `Renderer`, all you need to know is when the widget you created gets disposed.  In this example for a JSON renderer, you would attach a listener to the created widget as `widget = new MyRendererWidget(); widget.disposed.connect(..)`.

blink1073: cf https://github.com/jupyterlab/jupyterlab_json/blob/5b0534fbc7acc63f629f0a6e584aaa44a263e94f/labextension/src/output.js#L73

bokeh/io.py
jlstevens: Updated as suggested.

bryevdv:
@jlstevens I liked, we have tests to prevent `tornado` and server code from being brought in at the top level. They are dependencies, but also heavyweight:
```
____________________________ test_no_tornado_common ____________________________
    def test_no_tornado_common():
        proc = subprocess.Popen([
            "python", "-c", "import sys; %s; sys.exit(1 if any(('bokeh.client' in x or 'bokeh.server' in x) for x in sys.modules.keys()) else 0)" % ";".join(basic_imports)
        ],stdout=subprocess.PIPE)
        proc.wait()
        if proc.returncode != 0:
>           assert False
E           assert False
tests/test_no_client_server_common.py:19: AssertionError
____________________________ test_no_tornado_common ____________________________
    def test_no_tornado_common():
        proc = subprocess.Popen([
            "python", "-c", "import sys; %s; sys.exit(1 if any('tornado' in x for x in sys.modules.keys()) else 0)" % ";".join(basic_imports)
        ],stdout=subprocess.PIPE)
        proc.wait()
        if proc.returncode != 0:
>           assert False
E           assert False
tests/test_no_tornado_common.py:21: AssertionError
============================= 190 tests deselected =============================
======= 2 failed, 925 passed, 190 deselected, 10 rerun in 59.27 seconds ========
```
so we only want them to be brought in when used. The import just need to go inside the functions where they are used. Hooray for unit tests!

bokeh/io.py
```diff
@@ -298,11 +303,25 @@ def show(obj, browser=None, new="tab", notebook_handle=False):
...
         _state.document.add_root(obj)
     return _show_with_state(obj, _state, browser, new, notebook_handle=notebook_handle)
 
 
+def _show_notebook_app_with_state(app, _state, app_path, notebook_url):
+    from IPython.display import HTML, display
+    from tornado.ioloop import IOLoop
+    loop = IOLoop.current()
+    server = Server({'/': app}, io_loop=loop, port=0, host='*',
+                    allow_websocket_origin=[notebook_url])
```
bryevdv: I just realized, `host='*'` is going to throw out a warning? We should find a way for that to not happen.

jlstevens: There are quite a lot of warnings that probably need to be either addressed or suppressed before this API can be considered polished. You'll see them when running the notebook - I'll take a look to see if I can spot one regarding ``host='*'``.

bokeh/io.py
```diff
@@ -38,6 +39,9 @@
 from .util.string import decode_utf8
 from .util.serialization import make_id
 
+from .application import Application
+from .server.server import Server
```
bryevdv: These need to be protected inside the function too.

jlstevens:
At Bryan's request, I've pushed my branch to the main repo instead of using my fork. Continued discussion of this PR may be found at #5746 and I think this one can now be closed.

bryevdv:
superceded by #5746

Issue #999, opened 2014-08-13
Migration to Bokeh Organization


damianavila:
This is an issue to discuss and list all the tasks involved in the migration of the Bokeh repo from Continuum/bokeh to the Bokeh Organization. 
- [x] transfer the repository: https://help.github.com/articles/how-to-transfer-a-repository
- [x] ~~update secure variables in `.travis.yml` (re-encrypt key pair)~~
- [x] change nbviewer URLs in bokeh-notebooks repository
- [x] transfer bokeh-notebooks to bokeh organization
- [x] decide what to do about ContinuumIO/{bokehjs,bokehjs-build,...}
- [x] ~~update links in StackOverflow answers, etc~~ (https://github.com/bokeh/bokeh/issues/999#issuecomment-59843003)

Feel free to add more points to the list.

PR #5727, opened 2017-01-13
added tooltip font color to avoid being masked by color:white property from outside css

### LINKED ISSUE #5726: Tooltip font color can be turned white by outside css because it's not specified by bokeh
This is to make a case for a fork with a fix.

DylanAlloy:

- [x] issues: fixes #5726
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/less/toolbar.less
+++ b/bokehjs/src/less/toolbar.less
@@ -186,7 +186,7 @@
         .bk-toolbar-button {
           .bk-tip {
              float: right;
-             color: @gray-dark;
+             color: darkgray;
           }
         }
       }
--- a/bokehjs/src/less/toolbar.less
+++ b/bokehjs/src/less/toolbar.less
@@ -186,7 +186,7 @@
         .bk-toolbar-button {
           .bk-tip {
              float: right;
-             color: #999999;
+             color: @gray-dark;
           }
         }
       }
--- a/bokehjs/src/less/toolbar.less
+++ b/bokehjs/src/less/toolbar.less
@@ -186,7 +186,7 @@
         .bk-toolbar-button {
           .bk-tip {
              float: right;
-             color: darkgray;
+             color: #333;
           }
         }
       }


bryevdv:
@DylanAlloy apologies, we had a problem to resolve with our CI testing. Can you merge / rebase master into this PR?

bryevdv:
@mattpap @birdsarah aware of any integration tests that this small change would impact? In absence of the submitter, I'd like to just merge.

birdsarah:
No, but I don't think this is the right color. 

#999 is pretty light. I think it should be @gray-dark (https://github.com/bokeh/bokeh/blob/master/bokehjs/src/vendor/bootstrap-3.1.1/less/variables.less#L11)

You should def test this locally.

bryevdv:
I can change to `@gray-dark` in the GH UI, I don't think I can do the merge easily though.

bryevdv:
@birdsarah is that the correct syntax?

bryevdv:
Actually I replaced the no-arch packages on the bokeh channel, tests may just pass now.

birdsarah:
should be, but do check it's working locally

bryevdv:
CSS building seemed to fail with that change.

bryevdv:
Trying `darkgray`

birdsarah:
see the link above - the variable is called dark-gray - you may need to import it. If not, the quick fix is just `#333`.

mattpap:
btw. bootstrap should not be used anymore in core bokehjs.

mattpap:
I'm curious what's the underlying cause for color propagation in this case? Inheritance? This PR may be a quick fix, but we may want to look for more systematic solution, because we don't want to configure every CSS property for every class we have.

birdsarah:
> btw. bootstrap should not be used anymore in core bokehjs.

good point

bryevdv:
Tested locally, merging now

DylanAlloy:
Just wanted to say sorry I wasn't around to do a merge request again but it looks like you folks have it figured out.

Issue #4135, opened 2016-04-07
Don't expose external typings beyond Bokeh namespace


mattpap:
Currently we use stock typings for jquery and underscore, and those expose global ambient variables which conflict with other typings, e.g. those provided by application in which bokehjs is being embeded. We have to process external typings  and nest them under `Bokeh` namespace.

PR #5730, opened 2017-01-13
Use what we need from underscore

### LINKED ISSUE #5729: Use only what we need from underscore
Pretty much drop underscore as a whole and use only parts we need. Use typescript for extra points.

mattpap:
fixes #5729, #4135


mattpap:
It's interesting to see that a few examples fail on travis ci with:
```
undefined is not a function (evaluating 't.slice()')
```
(note minified resources), but not locally (despite the choice of resources). On a side note, tests are green, despite presence of multiple hard javascript errors.

bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
```diff
@@ -147,9 +147,8 @@ export class HoverToolView extends InspectToolView
 
     for i in indices['1d'].indices
       # multiglyphs will set '1d' and '2d' results, but have different tooltips
-      if not _.isEmpty(indices['2d'])
-        for pair in _.pairs(indices['2d'])
-          [i, j] = [pair[0], pair[1][0]]
+      if not isEmpty(indices['2d'])
+        for i, [j] of indices['2d']
```
bryevdv: is this completely equivalent? What will this do if the second array is longer than length 1?

bokehjs/src/coffee/models/tools/gestures/select_tool.coffee
```diff
@@ -15,7 +14,7 @@ export class SelectToolView extends GestureToolView
         sm.clear()
 
   _save_geometry: (geometry, final, append) ->
-    g = _.clone(geometry)
+    g = clone(geometry)
```
bryevdv: `clone` for objects in `core/util/array` ?

bokehjs/src/coffee/models/tickers/months_ticker.coffee
```diff
@@ -56,7 +54,7 @@ export class MonthsTicker extends SingleIntervalTicker
         month_date.setUTCMonth(month)
         return month_date)
 
-    month_dates = _.flatten(months_of_year(date) for date in year_dates)
+    month_dates = [].concat((months_of_year(date) for date in year_dates)...)
```
bryevdv: fairly obscure syntax, at least to me. Add `flatten` func to `core/util`?

bokehjs/src/coffee/models/tickers/adaptive_ticker.coffee
```diff
@@ -37,9 +35,9 @@ export class AdaptiveTicker extends ContinuousTicker
   initialize: (attrs, options) ->
     super(attrs, options)
 
-    prefix_mantissa =  _.last(@mantissas) / @base
-    suffix_mantissa = _.first(@mantissas) * @base
-    @extended_mantissas = _.flatten([prefix_mantissa, @mantissas, suffix_mantissa])
+    prefix_mantissa = @mantissas[@mantissas.length-1] / @base
+    suffix_mantissa = @mantissas[0]                   * @base
```
bryevdv: suggest remove extra whitespace

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -220,9 +221,9 @@ export class PlotCanvasView extends BokehView
 
     if follow_enabled and has_bounds
       logger.warn('Follow enabled so bounds are unset.')
-      for xr in _.values(frame.x_ranges)
+      for xr in values(frame.x_ranges)
         xr.bounds = null
-      for yr in _.values(frame.y_ranges)
+      for yr in values(frame.y_ranges)
```
bryevdv: if we implement `values` as a convenience we should definitely implement `keys` too I think

bokehjs/src/coffee/core/util/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  return toString.call(obj) === "[object Number]"
+}
+
+export function isString(obj: any): obj is string {
+  return toString.call(obj) === "[object String]"
+}
+
+export function isStrictNaN(obj: any): obj is number {
+  return isNumber(obj) && obj !== +obj
+}
```
bryevdv: I think this should just be `isNan`

bokehjs/src/coffee/core/util/eq.ts
```diff
@@ -0,0 +1,97 @@
...
+  // Remove the first object from the stack of traversed objects.
+  aStack.pop()
+  bStack.pop()
+  return true
+}
+
+// Perform a deep comparison to check if two objects are equal.
+export function isEqual(a: any, b: any) {
+  return eq(a, b)
+}
```
bryevdv: Is this vendored/ported from something? If so a reference in a comment would be advised

bokehjs/src/coffee/core/util/array.ts
```diff
@@ -0,0 +1,294 @@
...
+  return dest
+}
+
+export function copy<T>(array: Array<T>): Array<T> {
+  return array.slice()
+}
+
+export function clone<T>(obj: T): T {
+  return extend({}, obj)
+}
```
bryevdv: This, `values`, and  `extend` seem out of place in a module called `array.ts`

bokehjs/src/coffee/core/properties.coffee
```diff
@@ -79,33 +79,27 @@ export class Property
...
+        else                                 default_value
 
       obj.setv(attr, attr_value, {silent: true, defaults: true})
 
-    # if _.isObject(attr_value) and not _.isArray(attr_value) and not attr_value.properties?
-    #   @spec = attr_value
-    #   if _.size(_.pick.apply(null, [@spec].concat(@specifiers))) != 1
-    #     throw new Error("Invalid property specifier #{JSON.stringify(@spec)}, must have exactly one of #{@specifiers}")
-
-    if _.isArray(attr_value)
```
bryevdv: I can't recall why this was commented. I would prefer to have this check but maybe more work is needed. Please make an issue to investigate validating dataspecs better.

bokehjs/src/coffee/api/typings/bokeh.d.ts
```diff
@@ -35,8 +32,19 @@ declare namespace Bokeh {
...
+    export function sum(array: Array<number>): number;
+    export function cumsum(array: Array<number>): Array<number>;
+    export function min(array: Array<number>): number;
+    export function minBy<T>(array: Array<T>, key: (item: T) => number): T;
+    export function max(array: Array<number>): number;
+    export function maxBy<T>(array: Array<T>, key: (item: T) => number): T;
+    export function argmin(array: Array<number>): number;
+    export function argmax(array: Array<number>): number;
+    export function values<T>(object: {[key: string]: T}): Array<T>;
   }
```
bryevdv: I'm not sure `LinAlg` is the right name here. All of these are basic array operations, not anything to do with actual linear algebra.

bryevdv:
@mattpap I've left a line review. Mostly seems very good. I'd like to look at the image test reports, do some local testing, etc but that may not be until tomorrow. If you can describe the extent of your local testing that would be helpful

bryevdv:
FYI if you would like a non-squash merge, please state that (perhaps we can add a label to request history preserving merges)

bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
mattpap: This works because there is no bounds checking, so `[j] = [1,2,3]` will give you `j == 1`.

bokehjs/src/coffee/models/tickers/months_ticker.coffee
mattpap: `flatten()` has additional meaning assigned to it. It not only concatenates arrays, but also allows scalar values and recursive behavior. We don't need this here. Perhaps I could add a convenience function `concat()` for this purpose. Note that `[].concat(arrays...)` is an idiom.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
mattpap: The situation around `Object.keys()` and `Object.values()` sucks, because `keys()` is widely implemented and `values()` is not, so we have to provide the implementation for `values()`. I did consider adding a trivial wrapper `keys = Object.keys` to `core/util/array`.

bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
bryevdv: OK!

bokehjs/src/coffee/models/tickers/months_ticker.coffee
bryevdv: ok, fine then

bokehjs/src/coffee/models/plots/plot_canvas.coffee
bryevdv: Yes I am suggesting the trivial wrapper just so there is a symmetry of expectation `keys` and `values` and the code looks slightly nicer

bokehjs/src/coffee/core/util/types.ts
mattpap: We use both `global.isNaN()` and this function, which have fairly different corner case behavior. I didn't want to investigate all use cases of `global.isNaN()` right now, so I proceeded this way, so that there is no room for confusion which is which. In a later iteration I will have to resolve this, because ultimately there should be only one `isNaN()`.

bryevdv: Ok, but then please make the issue now to investigate later (and reference this PR)

bokehjs/test/models/formatters/func_tick_formatter.coffee
```diff
@@ -34,7 +34,7 @@ describe "func_tick_formatter module", ->
 
     it "should support imports using require", ->
       formatter = new FuncTickFormatter({
-        code: "var _ = require('underscore'); return _.max([1,2,3])"
+        code: "var max = require('../../core/util/array').max; return max([1, 2, 3])"
```
mattpap: This `../..` seems unnecessary, but otherwise tests fail. This is because this import isn't resolved as in custom models, but relative to `FuctTickFormatter` model (as if we were developing bokehjs here).

bokehjs/src/coffee/models/tools/gestures/select_tool.coffee
mattpap: I moved `clone()`, `extend()` and `values()` to `core/util/object`.

mattpap: And added `keys()`.

bokehjs/src/coffee/core/util/eq.ts
mattpap: This particular code is almost exact copy from underscore. There is some code copied from and some based on underscore in `array.ts`, `object.ts` and `types.ts`. I'm not sure what's the best approach here, but I wanted to have licensing information in one place. What do you think?

bokehjs/src/coffee/api/typings/bokeh.d.ts
mattpap: This will get sorted out in another iteration. I need to look at this from the user facing API standpoint.

bokehjs/src/coffee/core/properties.coffee
mattpap: The check is exactly as below (changed to more efficient version). I think the reason it's commented out is because it's too strict, e.g. doesn't take units into account. I think altogether this logic is reversed, because we already have type information and we are guessing the understanding of an object based on its content. I will start an issue for this, because this should be straightened out.

mattpap:
>  I'd like to look at the image test reports

Those are broken currently, see https://github.com/bokeh/bokeh/pull/5730#issuecomment-272882515.

mattpap:
Issues to resolve:

- [ ] `t.slice()` error
- [ ] update numbro
- [ ] add license information

mattpap:
> `t.slice()` error

This isn't resources, but phantomjs related. I can reproduce this locally in phantomjs. It works fine in Chromium and FF. The problem is in `PatcheView._build_discontinuous_object()` when copying `nanned_qs`.

mattpap:
phantomjs is unable to copy (slice) typed arrays.

mattpap:
If someone wonders how to check if something is a typed array, then the answer is that there is no canonical way. You have to test for each individual type of typed array and there are 9 of those.

bryevdv:
> If someone wonders how to check if something is a typed array, then the answer is that there is no canonical way. You have to test for each individual type of typed array and there are 9 of those.

Amazing.

mattpap:
@bryevdv, I resolved all technical issues. Still I need feedback on license information.

bryevdv:
@mattpap I would just add a comment at the top of the file, I think that is sufficient.

bryevdv:
@mattpap https://github.com/bokeh/numbro/pull/4 is merged, does `package.json` look for a specific commit hash or tag of that repo?

mattpap:
I already updated `package.json`. The dependency depends on SHA1, so as soon I pushed to the repository I could use it here.

mattpap:
> I would just add a comment at the top of the file, I think that is sufficient.

There are several files (5?) affected to varying degrees (from 99% copy to maybe 5% (this number is taken from thin air)). Earlier I was proposing adding the license in one place, maybe a licenses file included in main, but I'm not sure. I also don't want to have the license appear multiple times in the bundle, if possible.

bryevdv:
@mattpap put the license file in one place (main seems ok, but your call) and put one-line comments referring people that place everywhere else.

mattpap:
@bryevdv, I proceeded with adding license information to each file. I may revisit this later, seeing that `bokeh.min.js` contains, e.g., 15 copies of nucleic license and full release notes from sprintf.js.

bryevdv:
LGTM, confirming: ready to merge?

mattpap:
@bryevdv, yes. Just to note, I skipped tests on the last to commits with `[skip ci]` (I tested code quality manually).

bryevdv:
squash or rebase merge?

mattpap:
Squash. I would prefer merge, but at this point most commits wouldn't be workable with under bisection. I need to figure out a system for at least partially testing all commits in CI without exhausting our resources.

bryevdv:
I think the `no ci` also prevented the build on master after the merge :/

Issue #5716, opened 2017-01-13
BokehJS incorrect height calcs of Boxview


Hanterdro:
I found an layouting issue with BokehJS:
If a Boxview is above a plot the height of the bokeh container is too small and bokeh is overlapping the container. Elements outside the bokeh container may be overlapped by bokeh.

Basically the issue is that `BoxView.prototype.get_height` ignores margins.

An easy bugfix for this would be

`height += $(child_view.el).outerHeight(true);`
and remove the silly magic `+ 20` for the height.

```
  WidgetBoxView.prototype.get_height = function() {
    var child_view, height, key, ref1;
    height = 0;
    ref1 = this.child_views;
    for (key in ref1) {
      if (!hasProp.call(ref1, key)) continue;
      child_view = ref1[key];
      height += $(child_view.el).outerHeight(true);
    }
    return height;
  };
```

instead of

```
  WidgetBoxView.prototype.get_height = function() {
    var child_view, height, key, ref1;
    height = 0;
    ref1 = this.child_views;
    for (key in ref1) {
      if (!hasProp.call(ref1, key)) continue;
      child_view = ref1[key];
      height += child_view.el.scrollHeight;
    }
    return height + 20;
  };
```

With this fix `get_height()` returns the expected value. But this value is not used for the container box in `LayoutDOMView.prototype.render`.
![bildschirmfoto 2017-01-13 um 11 15 21](https://cloud.githubusercontent.com/assets/7668392/21926615/cfe11734-d982-11e6-9a8b-e27fe4d04582.png)

bryevdv:
@Hanterdro would you mind submitting a PR with this change? I'd be happy to test it out

Hanterdro:
Yeah... I could do that... but why? :-)
My changes did not fix the issue with the incorrect height calculation of the container box.
I expected that it would.
The bokehJS API function call `get_height` returns with my changes the correct value, but internally this functions seems not to be used to set the height for the container.

Hanterdro:
However, I created the pull request.
I hope this will help.

https://github.com/bokeh/bokeh/pull/5739

Hanterdro:
Good news, my bugfix is working.

I was initially lazy and did my changes directly in the build files.
After I added the bugfix it in the source file and started the build tasks the fix is working as expected.

Hanterdro:
Okay... I don't get it..

[...]

PR #5739, opened 2017-01-16
bokehjs fix for height calc

Hanterdro:
This pull Requests fix incorrect height calcs

See: https://github.com/bokeh/bokeh/issues/5716

### DIFF
--- a/bokehjs/src/coffee/models/layouts/widget_box.coffee
+++ b/bokehjs/src/coffee/models/layouts/widget_box.coffee
@@ -55,12 +55,13 @@ export class WidgetBoxView extends LayoutDOMView
       @el.style.width = css_width
 
   get_height: () ->
+    # We have to add on 10px because widgets container has a padding of 5px on top and button.
     height = 0
     for own key, child_view of @child_views
-      styles = window.getComputedStyle(el)
+      styles = window.getComputedStyle(child_view.el)
       margin = parseFloat(styles['marginTop']) + parseFloat(styles['marginBottom'])
       height += Math.ceil(child_view.el.offsetHeight + margin)
-    return height
+    return height + 10
 
   get_width: () ->
     if @model.width?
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/models/layouts/widget_box.coffee
```diff
@@ -58,8 +58,8 @@ export class WidgetBoxView extends LayoutDOMView
     height = 0
     # We have to add on 10px because widgets have a margin at the top.
     for own key, child_view of @child_views
-      height += child_view.el.scrollHeight
-    return height + 20
+      height += $(child_view.el).outerHeight(true)
```
mattpap: We don't use jquery anymore. This has to be implemented with native DOM functionality.

Hanterdro:
Okay... I replaced the jQuery stuff with vanillaJS.
I readded the magic `+10` for the height, but I don't think it's because of the margin of the elements. I think it's needed for the `widget-box` container, which has a padding-top and padding-bottom of 5px.

bryevdv:
HI @Hanterdro thanks for looking in to this some! The layout system is a bit complicated, ultimately Bokeh uses a constraint solver to lay things out. It's possible the value is not getting fed into that pipeline. I will try to take a look myself using your branch to see if I can spot where the broken link might be, but it will be later this week.

bryevdv:
@mattpap any thoughts on this? Should we just merge, or is it superseded by your other PR?

bryevdv:
ping @mattpap for comment

bryevdv:
@mattpap we should either merge or close this

bryevdv:
Actually, looking at the failing tests, I think this would be non-trivial to fix up. I am going to go ahead and close. @Hanterdro thanks for the PR, I regret that we could not get it merged. Layout is a large effort and one of the more complicated areas, and much code has changed around this PR. I do hope you will consider contributing again.

PR #5746, opened 2017-01-17
Automatically destroying servers in the notebook embed example

jlstevens:
This PR continues on from #5717 but using a local branch. See #5717 for previous comments and discussion.


bryevdv:
rebased on master, merged a few commits. Will look at generalizing server transport layer now.

bryevdv:
@bokeh/dev I plan to merge this later today unless there are comments

PR #5760, opened 2017-01-18
Bryanv/refactor build

### LINKED ISSUE #5209: Stop building noarch conda packages
The status of no-arch packages seems unclear, and continuing to use them seems risky.  A new version of conda-build caused a release candidate build to partially fail (only after several hours and in the middle of the night, of course).

### LINKED ISSUE #5679: Deploy script should automatically create and upload examples tarballs
Currently done manually on full releases.

bryevdv:
Trying again again.

issues: closes #5209 closes #5679

* Refactor the `setup.py` nightmare into simple `setup.py` and `_setup_support.py`
* Update to current tool versions: `conda=4.3.4`, `conda-env=2.6.0`,  `conda-build=2.1.1`
* Adhere to PEP 440 versions:
  - `0.12.5`
  - `0.12.5-dev2`
  - `0.12.5rc1`
  and on travis, local versions like `bokeh-0.12.4+20.gddad8b0-py27`

* remove all `__conda_version__` and `TRAVIS_BUILD_NUMBER` junk
* remove `get_bump_version.py`
* simplify `meta.yaml` with jinja templates
* store classifiers in sidecar file `classifiers.txt`

**HERE is the change that merits discussion:**

* use `setuptools` in ALL cases now

This allows cutting out a special case for windows, removing out custom `develop` and `find_packages` code, and handling entry points correctly. There were some reasons ~5 years ago not to use setuptools but I think that it's now OK. 

ping @mattpap @damianavila @bokeh/dev 

Still to do:
* [x] replace `noarch` packages for: `flexx`, `pytest-base-url`, and `pytest-selenium`
* [x] rewrite `build_and_upload.sh` as a python script with detailed status and error checks
  - [x] use `conda convert` to create non-`noarch` packages
  - [x] update upload for non-`noarch` packages
  - [x] create and upload examples tarball 

We do need to verify that Bokeh works in venv's still but AFAICT it should now be fine still.


bryevdv:
This needs some more work for the final steps of a dev build but it's good enough to keep and fix up in follow on PRs. Also it has a fix all other PRs will need to merge due to `nb_conda_kernels` being removed from default anaconda installers.

bryevdv:
ping @mattpap @bokeh/dev

bryevdv:
just for FYI the remaining problem is just that somehow `conda-build` is not finding `pytest-selenium` to install when it creates an env. Can try adding `-c` or if need be, `--no-test`

conda.recipe/meta.yaml
```diff
@@ -1,10 +1,13 @@
+{% set data = load_setup_py_data() %}
+
 package:
   name: bokeh
-  version: 'nightly'
+  version: {{ data['version'] }}
+
+source:
+  path: ..
```
damianavila: Are you sure you want o use `path` here... some time ago I was told that `path` was going into "deprecation" and that we should use `git_url` instead, not sure if this is the actual recommendation right now, @msarahan thoughts?

scripts/build_upload.py
```diff
@@ -0,0 +1,476 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2017, Continuum Analytics, Inc. All rights reserved.
+#
+# Powered by the Bokeh Development Team.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+''' Bokeh build and upload script for TravisCI release builds
+
+'''
```
damianavila: First of all, thanks for doing this!!

damianavila: I mean porting the script to python (bash did the work enough well, but this is far better for others to understand what the code/process does).

scripts/build_upload.py
```diff
@@ -0,0 +1,476 @@
...
+        c.close()
+
+        data = json.loads(buf.getvalue().decode())
+
+        cdn_token = data["access"]["token"]["id"]
+        cdn_id = [
+            d["endpoints"][0]["tenantId"]
+            for d in data["access"]["serviceCatalog"]
+            if d["name"] == "cloudFiles"
+        ][0]
```
damianavila: Nice to have a caching here... the data structure can change (it happened to me once or twice, IIRC).

scripts/build_upload.py
```diff
@@ -0,0 +1,476 @@
...
+#--------------------------------------
+
+@build_wrapper('conda')
+def build_conda_packages():
+    for v in "27 34 35 36".split():
+        os.environ['CONDA_PY'] = v
+        run("conda build conda.recipe --quiet")
+        # TODO (bev) make platform detected or configurable
+    files = " ".join(glob.glob('/home/travis/miniconda/conda-bld/linux-64/bokeh*'))
+    run("conda convert -p all %s" % files)
```
damianavila: So, we have conda convert again... are you planning to keep this for the future or you plan to use the new arch type package anytime soon?

scripts/travis_install
```diff
@@ -19,29 +19,29 @@ rm -rf $MINICONDA.sh
...
 python -V
 
-echo -e "$PINNED_PKGS" > /home/travis/miniconda/conda-meta/pinned
+echo -e "$PINNED_PKGS" > $HOME/miniconda/conda-meta/pinned
 
-DEPS_TRAVIS="python=$TRAVIS_PYTHON_VERSION conda=4.1.11 conda-env=2.5.2 conda-build=1.21.14"
+DEPS_TRAVIS="python=$TRAVIS_PYTHON_VERSION conda=4.3.4 conda-env=2.6.0 conda-build=2.1.1"
 conda install --yes $DEPS_TRAVIS
 
+conda config --set auto_update_conda off
```
damianavila: :+1:

conda.recipe/meta.yaml
msarahan: @damianavila ``path`` is not going away.  The difference is:

``path``: takes the current state of files/folders as they are at the moment of build.  Copies them to a temporary location.
``git_url``: clones the repo on disk.  This is the state at the most recent commit (when no git_tag is specified, at least).  This will not include changes on disk that have not yet been committed.

What will be going away very soon is support for post-build versioning - creating ``__conda_version__.txt`` files.

setup.py
```diff
@@ -1,614 +1,146 @@
...
-_version = versioneer.get_version()
-_cmdclass = versioneer.get_cmdclass()
+'''
+from os.path import join
+from shutil import copy
+import sys
 
-# Horrible hack: workaround to allow creation of bdist_wheel on pip installation
-# Why, for God's sake, is pip forcing the generation of wheels when installing a package?
+from setuptools import find_packages, setup
```
damianavila: OK, using setuptools... I think right now the things have improved on their enough to use it. So :+1: on this. And entry points for free.
Keep in mind this comment for future improvements: https://github.com/jupyterlab/jupyterlab/pull/613

conda.recipe/meta.yaml
damianavila: Thanks for the clarification @msarahan!!

damianavila:
@bryevdv I left some comments mostly asking things and linking some other that you could potentially use in future improvements. Thanks for doing this work!

scripts/build_upload.py
bryevdv: I want to use new no-arch packages, if for no other reason then 4 plat builds times 5 conda converts takes a really long time. But it would be irresponsible to switch to them before it can be reasonably expected that most users have updated to a conda that supports them.

conda.recipe/meta.yaml
bryevdv: use of `__conda_version__.txt` was removed in this PR. I'm on the fence about the `path`. There are advantages and disadvantages both ways. `..` seems best for now

scripts/build_upload.py
bryevdv: @damianavila I'm not sure I understand, do you mean caching in between test runs? Or that this value might change between the start of the test and the time the creds are used towards the end?

conda.recipe/meta.yaml
damianavila: > There are advantages and disadvantages both ways

Yep, I agree.

scripts/build_upload.py
damianavila: I am saying that I experienced in the past a change in the data structure that broke the script... but it was an API change from them. Pretty sure things will not change in a test run. But could change from one month to another.

scripts/build_upload.py
damianavila: >But it would be irresponsible to switch to them before it can be reasonably expected that most users have updated to a conda that supports them.

OK, I agree.

PR #5766, opened 2017-01-19
Use border-box as the default box model under bk-root

### LINKED ISSUE #5668: Use box-sizing: border-box for everything under bk-root
Make sure everything works afterwards. Also remove any specific `box-sizing`s, but leave one in `data_table.less` (SlickGrid requires content box model). Fix this issue after #2933 and #5102.

mattpap:
fixes #5668

### DIFF
--- a/bokehjs/src/less/bokeh.less
+++ b/bokehjs/src/less/bokeh.less
@@ -1,4 +1,10 @@
 .bk-root {
+  box-sizing: border-box;
+
+  *, *:before, *:after {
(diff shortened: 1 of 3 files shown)


bryevdv:
@mattpap This is almost four months old. Can it be closed?

mattpap:
This needs to be worked on in the context of layout improvements, because it's mostly layout (specifically widgets) that's affected by this change. I will include this in on of upcoming layout PRs.

PR #5767, opened 2017-01-19
Show bokehjs version in "successfully loaded" message

### LINKED ISSUE #5705: Print Bokeh version in notebook "BokehJS loaded" message
The following situation (seen most recently in https://github.com/bokeh/bokeh/issues/5703) has happened to users many times now on gitter, the tracker, stackoverflow, etc, resulting in perceived issues and problems:

* user updates Bokeh in some conda env
* jupyter notebook is in a different env
* user starts notebook thinking they are using updated bokeh
* but really are using whatever bokeh is in the env with the notebook

This is an easy situation to fall into, but there is no good way to detect or warn about it. It's simply a consequence of the fundamental way system paths work. 

Propose to change notebook message from:
```
BokehJS successfully loaded.
```
to 
```
BokehJS 0.12.4 successfully loaded.
```

This is not necessarily expected to lower the incidence of reported problems but will at least make it possible to diagnose whether this is the problem with one simple question. (Any may lower the incidence, if users notice and self-diagnose)

mattpap:
fixes #5705

### DIFF
--- a/bokeh/core/_templates/autoload_nb_js.js
+++ b/bokeh/core/_templates/autoload_nb_js.js
@@ -24,7 +24,8 @@
 
   function display_loaded() {
     if (window.Bokeh !== undefined) {
-      document.getElementById({{ elementid|json }}).textContent = "BokehJS successfully loaded.";
+      var el = document.getElementById({{ elementid|json }});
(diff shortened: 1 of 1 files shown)


bryevdv:
LGTM

PR #5770, opened 2017-01-19
Bryanv/build fixups

bryevdv:
This fixes up some remaining issues with the recent build automation update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -139,6 +139,7 @@ after_success: |
         echo "Building a FULL RELEASE package: $TRAVIS_TAG"
       fi
       python scripts/build_upload.py $TRAVIS_TAG
+      exit $?
   else
       echo "No tag present, not building a release package."
   fi
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -229,10 +229,16 @@ def check_checkout():
         abort()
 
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -139,6 +139,7 @@ after_success: |
         echo "Building a FULL RELEASE package: $TRAVIS_TAG"
       fi
       python scripts/build_upload.py $TRAVIS_TAG
+      exit $?
```
bryevdv: The script failed but the travis job reported success, I believe this should cause the travis job to report failure whenever the script has a non-zero exit.

scripts/deploy.py
```diff
@@ -229,11 +229,17 @@ def check_checkout():
...
     try:
         out = run("git for-each-ref --sort=-taggerdate --format '%(tag)' refs/tags")
         tags = [x.strip("'\"") for x in out.split("\n")]
 
+        if CONFIG.new_version in tags:
+            failed("There is already an existing tag for new version %r" % CONFIG.new_version)
+            abort()
+        else:
+            passed("New version %r does not already have a tag" % CONFIG.new_version)
+
```
bryevdv: This makes sure there is not already a tag for the proposed new version.

PR #5781, opened 2017-01-22
WIP: Allow reselection of objects

jakirkham:
Related https://github.com/bokeh/bokeh/issues/2326

Instead of computing the union of indices, simply concatenate the previously selected indices with the currently provided ones. This should allow for things like deselection on a second click. Though it doesn't require that per se as it could also be interpreted as selecting the same object again thus resulting in later selections being no-ops.

### DIFF
--- a/bokehjs/src/coffee/core/selector.coffee
+++ b/bokehjs/src/coffee/core/selector.coffee
@@ -2,7 +2,7 @@ import {HasProps} from "./has_props"
 import * as hittest from "./hittest"
 import {logger} from "./logging"
 import * as p from "./properties"
-import {union, concat} from "./util/array"
+import {concat} from "./util/array"
 
 merge = (obj1, obj2) ->
   result = {}
@@ -13,7 +13,7 @@ merge = (obj1, obj2) ->
   for key in keys
     arr1 = obj1[key] or []
     arr2 = obj2[key] or []
-    result[key] = union(arr1, arr2)
+    result[key] = concat(arr1, arr2)
 
   return result
 
@@ -24,9 +24,9 @@ export class Selector extends HasProps
     @setv('timestamp', new Date(), {silent: silent})
     @setv('final', final, {silent: silent})
     if append
-      indices['0d'].indices = union(@indices['0d'].indices, indices['0d'].indices)
+      indices['0d'].indices = concat(@indices['0d'].indices, indices['0d'].indices)
       indices['0d'].glyph =  @indices['0d'].glyph or indices['0d'].glyph
-      indices['1d'].indices = union(@indices['1d'].indices, indices['1d'].indices)
+      indices['1d'].indices = concat(@indices['1d'].indices, indices['1d'].indices)
       indices['2d'].indices = merge(@indices['2d'].indices, indices['2d'].indices)
     @setv('indices', indices, {silent: silent})
 


jakirkham:
Hmm...seems like CI is not running. 😕

bryevdv:
@jakirkham it looks like like CI ran and passed? FYI I am interested in looking at and discussing the PR, I am currently buried under a backlog of work, so it will be a few days before I can turn to it.

jakirkham:
> ...it looks like like CI ran and passed?

Weird, maybe webhook deliveries were going really slow. Didn't even indicate that it was running yesterday.

In any event, this only shows we don't break anything in the existing test suite. Probably should add some tests here too. I'll defer to you about where and how to add these.

> FYI I am interested in looking at and discussing the PR, I am currently buried under a backlog of work, so it will be a few days before I can turn to it.

Not a problem. Looks like I'm going to be pretty swamped as well. So no rush. Just curious to hear your thoughts when you have a chance.

mattpap:
>  it looks like like CI ran and passed?

I restarted them. Some jobs were failing for random reasons.

jakirkham:
Have you had a chance to give this some thought, @bryevdv? Would be interested in discussing this when you have some time.

bryevdv:
@jakirkham apologies I this had fallen off my radar.. I've moved the issue to the `0.12.6` milestone so that it will get due consideration after next weeks release. Thanks for you your patience

jakirkham:
I guess there was a `merge` function added recently for the 2D case. Guessing this fixes some issues encountered. Have updated the PR to allow for the `merge` function. However I think there is a fair question as to whether merging is strictly necessary if we are just concatenating user selections and not trying to remove any of them.

bryevdv:
@jakirkham FYI we are in the last stretch of a release I have stopped travis to conserve build workers for now. I'd like to resolve this for `0.12.6` for sure. to set expectations I will be away for a few weeks starting saturday, so it will be after that. Thanks for your patience

jakirkham:
No worries. Was just going through open PRs trying to do some housekeeping.

bryevdv:
That is *always* appreciated, so thanks :)

bryevdv:
@clairetang6 can you comment on your current PR in relation to this one?

jakirkham:
Is there an xref for that PR @bryevdv?

bryevdv:
https://github.com/bokeh/bokeh/pull/6202

jakirkham:
Could I also ask that this be reviewed soon?

clairetang6:
Hi @jakirkham, deselection when a selected point is clicked seems like a nice feature to have. However, I don't think we want to concatenate the indices in the selector. The selector represents the selection itself, so it should just contain the selected indices. 

Right now, adding to a selection with any of the select tools (box select, lasso select, etc) and with the tap tool goes through the same code in the selection manager which updates the selection with unions. This doesn't allow for deselection. For deselection, we need to update the selected indices by removing the ones that were already selected and are being "selected" again. But that kind of deselection probably shouldn't happen with the select tools that select many points at once. 

One possibility is to have the ability to configure whether a tool will update its selection by unioning or by deselection on second select. It might make sense to have the box select, etc. default to unioning and the tap tool default to deselecting.

jakirkham:
For some history, I initially proposed some other changes in PR ( https://github.com/bokeh/bokeh/pull/5484 ) to properly handle deselection, but it was rejected. This originally came out of issue ( https://github.com/bokeh/bokeh/issues/2326 ). This PR was created to serve as a compromise. It doesn't actually handle deselection itself. It merely allows the history of selection to be tracked and available to inspection.

It would allow anyone to handle deselection, specialized duplicate selection (e.g. changing the color of an object), and/or the status quo (with no actual change). It is a simple matter to cull the selected points in say a callback object or do anything else one desires. As it is implemented, it doesn't actually break or change Bokeh's behavior beyond allowing history to accumulate. It also allows users to play and determine what works best for their cases. Over time some of these common strategies can be pulled into Bokeh more formally. In the interim, this opens the door to these considerations without forcing a style.

clairetang6:
Ok, I see how you were approaching the toggling of selection in your last PR. That would be one possibility in getting the toggling in as feature, to have that code be an option to the unioning of indices. The only issue is needing to have that way of updating the indices be optional, so that it wouldn't cause deselection when you didn't want it.

I also see how having the number of times an index was selected could be useful. You could increase the saturation of color as a point gets selected more and more times.

I still don't think that the selector should keep the entire history of indices because it should be a faithful representation of the selection. If it did have the history, then people would need to do a unique operation themselves each time they just wanted the selection. If anyone wanted to have the full history, I think they could write a callback to be attached to a data source. The callback gets called each time there is a new selection, so you could keep track of how many times a specific index was selected there.

bryevdv:
@jakirkham Unfortunately this PR must be closed. My concerns are these:

* This PR is not compatible with #6202 as-is. However #6202 fixes two important bugs and increases testing, and that has to take precedence. I intend to merge #6202 this week. 

* I did not appreciate earlier the implication that this PR changes the interpretation of `indices` from "current selection" to "history of selections". I believe there is a good chance that that would break existing `CustomJS` code for users and possibly internal Bokeh codepaths as well, in hard to detect and fix ways.  I am -1 on doing that. [*]

* I also believe @clairetang6 is correct, and that selection callbacks could afford an existing path to implementing selection history without changes to the core library. 

[*] Because things are going to have to break anyway at some point. Selections need to become real Bokeh models instead of a bare dict. Unfortunately I can't justify breaking things *twice* (once now for this change, and then *again* for a full refactor of selection machinery)

I'd like to start a Wiki page for a real design discussion for a new selection machinery. Your input there would be extremely valuable.

bryevdv:
@jakirkham @clairetang6 Here is the URL for a Wiki design doc:

https://github.com/bokeh/bokeh/wiki/Selection-Architecture-Refactor

It's only a stub page to start. I am traveling the next few days but I will try to add more to it next week. If you would like to start adding requirements and use-cases to it, that would certainly be helpful.

bryevdv:
@jakirkham I'd also just like to re-iterate that your input and contribution is appreciated. I am never happy to close a PR. Most of the problem with adding new features right now (as opposed to patching up bugs) is that the current selection architecture is a muddled mess, and the fault for that lies entirely on me. I hope you will help shape the second iteration of things and make them much better and more robust.

PR #5784, opened 2017-01-23
move abstract out of model.py

### LINKED ISSUE #5704: Move abstract out of properties.py
Want everything in `properties.py` to be of interest to users, `abstract` is almost certainly never. 

Propose `has_props.py` or `model.py` @mattpap do you have a preference?

bryevdv:
issues: fixes #5704

This PR moves `abstract` to `has_props.py`, and also has makes the application of `abstract` add an admonition to the class docstring stating that the class is not useful to instantiate on its own. So this does not have to be remembered to be added by hand every time. There may be other classes that should be marked `abstract` (e.g. `MercatorTileSource`?)


bryevdv:
Since this had to touch so many model files (all of them basically) had hoped to make a sweeping docstring pass, but I don't have time to do that at the moment. Some improvements were made but the rest will have to wait for later.

bryevdv:
A quick note: to make the auto adding to the docstring work, docstrings need a blank line at the end. I went ahead and updated all models (I think) to be this format. This is also in anticipation of adding more content to the docstring anyway. I also would like to propose the following convention: `'''` for docstrings, `"""` for "data"

bokeh/charts/chart.py
```diff
@@ -286,7 +286,7 @@ def make_grid(self, dimension, ticker):
             ticker (obj): the axis.ticker object
 
         Return:
-            grid: Grid instance
+            Grid: A Grid instance
```
bryevdv: new sphinx versions automatically link to types for args and returns, needed to make it the real type other wise there were problems.

bokeh/core/has_props.py
```diff
@@ -37,18 +37,34 @@ def __init__(self, output, verbose=False, max_width=79, newline='\n'):
...
 '''
 
+def abstract(cls):
+    ''' A decorator to mark abstract base classes derived from |HasProps|.
+
+    '''
+    if not issubclass(cls, HasProps):
+        raise TypeError("%s is not a subclass of HasProps" % cls.__name__)
+
+    cls.__doc__ += _ABSTRACT_ADMONITION
```
bryevdv: This could probably be made a little smarter i.e. only have one blank line between the original docstring and this addition, but this is OK for now.

bokeh/models/annotations.py
```diff
@@ -1,59 +1,51 @@
-""" Renderers for various kinds of annotations that can be added to
+''' Renderers for various kinds of annotations that can be added to
 Bokeh plots
 
-"""
+'''
```
mattpap: Why all those unrelated changes?

bryevdv: I mentioned some things above.

bryevdv:
@mattpap I'd like to merge this do you have any other comments?

mattpap:
> (...) do you have any other comments?

No.

Issue #5785, opened 2017-01-23
Extension example breaking docs build


bryevdv:
Building docs on master with 
```
BOKEH_DOCS_CDN=local make SPHINXOPTS=-v clean html serve
```
results in
```
Traceback (most recent call last):
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/cmdline.py", line 296, in main
    app.build(opts.force_all, filenames)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/application.py", line 333, in build
    self.builder.build_update()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/builders/__init__.py", line 251, in build_update
    'out of date' % len(to_build))
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/builders/__init__.py", line 265, in build
    self.doctreedir, self.app))
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/environment/__init__.py", line 549, in update
    self._read_serial(docnames, app)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/environment/__init__.py", line 569, in _read_serial
    self.read_doc(docname, app)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/environment/__init__.py", line 677, in read_doc
    pub.publish()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/Sphinx-1.5-py3.4.egg/sphinx/io.py", line 55, in read
    self.parse()
  File "/Users/bryan/anaconda/lib/python3.4/site-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/Users/bryan/work/bokeh/bokeh/sphinxext/bokeh_plot.py", line 164, in parse
    (script, js, js_path, source) = _process_script(source, filename, env.bokeh_plot_auxdir, js_name)
[...]

PR #5786, opened 2017-01-23
Clean up document.py

bryevdv:
Move things not of general interest into other modules. Add more docstrings.

@mattpap this should be ready but I'd like to check building docs. Am currently blocked by #5785


bryevdv:
@mattpap I'd also like to merge this, but after #5784

PR #5790, opened 2017-01-25
Integration tests for python 2.7

### LINKED ISSUE #5789: Some integration tests fail when run with Python 2.7
Specifically, `test_editable_changes_data` in`test_datatable.py` and `test_scale_width_resizes_plot_while_maintaining_aspect_ratio` in `test_sizing_mode.py`. The first is caused by a unicode/str encoding problem and the second by truncating division.

clairetang6:
Fixes integration tests that fail when run with python 2.7. 

I also added some additional documentation on running screenshot tests. 

- [x] issues: fixes #5789

### DIFF
--- a/tests/integration/interaction/test_datatable.py
+++ b/tests/integration/interaction/test_datatable.py
@@ -43,7 +43,7 @@ def test_editable_changes_data(output_file_url, selenium):
     actions = ActionChains(selenium)
     actions.move_to_element(row_1_cell)
     actions.double_click()
-    actions.send_keys("33\ue007")  # After the backslash is ENTER key
+    actions.send_keys(u"33\ue007")  # After the backslash is ENTER key
     actions.perform()
 
     # Click row_2 (which triggers alert again so we can inspect the data)
--- a/sphinx/source/docs/dev_guide/testing.rst
+++ b/sphinx/source/docs/dev_guide/testing.rst
@@ -58,6 +58,14 @@ To run any of the tests without standard output captured use:
 
   py.test -s
 
+To run a subset of tests by name:
+
+.. code-block:: sh
+
+  py.test -k EXPRESSION
+
+This will run only the tests that include ``EXPRESSION`` in their names (function or class names).
+
 See the py.test documentation at http://pytest.org/latest/ for further information on py.test and it's options.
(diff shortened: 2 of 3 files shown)


tests/integration/interaction/test_datatable.py
```diff
@@ -43,7 +43,7 @@ def test_editable_changes_data(output_file_url, selenium):
     actions = ActionChains(selenium)
     actions.move_to_element(row_1_cell)
     actions.double_click()
-    actions.send_keys("33\ue007")  # After the backslash is ENTER key
+    actions.send_keys(u"33\ue007")  # After the backslash is ENTER key
```
mattpap: This `u` needs some explanation. Perhaps the escape code is wrong or unnecessary.

bryevdv:
Failures in one test looked like spurious PhantomJS failures, restarted.

tests/integration/interaction/test_datatable.py
clairetang6: So, the u is necessary for python 2.7 to indicate something as a unicode literal because otherwise string literals in python 2 only contain ascii characters. The \u escape character is recognized by unicode objects. It works without the u in python 3 because all python 3 strings are stored as unicode.

bryevdv: I am only wondering why ENTER needs to be specified as unicode at all, though this seems to indicate that it is an artifact of selenium somehow: https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.keys.html

bryevdv:
LGTM, @mattpap do you have any additional comments?

tests/integration/interaction/test_datatable.py
mattpap: This isn't enter, it's some tick mark or something. That's why I'm asking if this is necessary at all (+ what are the encoding requirements).

bryevdv: The page I linked, which appear to be selenium official docs, has:
```
ENTER = '\ue007'
```

mattpap: I see. LGTM then.

Issue #5783, opened 2017-01-23
Investigate kdbush 


bryevdv:
Possibly worth investigating, from the author of RBush:

> One recommendation I have is that if your list of points is static (you don't need to add/remove points after indexing), you should use kdbush, my other indexing library that performs point indexing 5-8x faster than RBush.

https://github.com/mourner/rbush/issues/63#issuecomment-268854922

mattpap:
kdbush doesn't work with rectangles, only points, so this could be used for anything that uses `_xy_index` and other glyphs would have to fall back to rbush.

bryevdv:
Investigation completete

PR #5791, opened 2017-01-25
Provide an abstraction layer on top of rbush

mattpap:
Originally this was supposed to fix issue #5783, but as kdbush doesn't support rectangles, it cannot be universally used, just as an optimization for point search, but a cleanup and addition of an abstraction layer is still useful. This can be used to add support for kdbush in fugure.

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/ray.coffee
+++ b/bokehjs/src/coffee/models/glyphs/ray.coffee
@@ -1,10 +1,7 @@
-import {Glyph, GlyphView} from "./glyph"
+import {XYGlyph, XYGlyphView} from "./xy_glyph"
 import * as p from "../../core/properties"
 
-export class RayView extends GlyphView
-
-  _index_data: () ->
-    @_xy_index()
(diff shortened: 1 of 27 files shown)


bokehjs/src/coffee/models/glyphs/ray.coffee
```diff
@@ -1,10 +1,7 @@
-import {Glyph, GlyphView} from "./glyph"
+import {XYGlyph, XYGlyphView} from "./xy_glyph"
 import * as p from "../../core/properties"
 
-export class RayView extends GlyphView
-
-  _index_data: () ->
-    @_xy_index()
```
bryevdv: It seems to me that this indexing for rays would be improved by taking the ray extent into account, but then we'd have to not make it an `XYGlyph` Maybe better just to leave it not an `XYGlyph`?

bryevdv: Actually I guess I am thinking of bounds, nm

PR #5797, opened 2017-01-26
Documented dependency on python versions

### LINKED ISSUE #5796: bokeh 0.12.4 needs Python 3.3 but this isn't documented anywhere
Bokeh `0.12.4` fails to plot from numpy data sources when using python 3.3. 

This is because PR #5544 introduced `base64.b64encode(array.data)` into bokeh/util/serialization.py, and `b64encode` is unable to handle `memoryview` (which is used by numpy). Python addressed this in http://bugs.python.org/issue17839 but that was only released in python 3.4. It's not an issue under python 2.7.

e.g. the following code
```
import numpy as np
x = np.random.rand(5)
y = np.random.rand(5)
p = figure()
p.line(x, y)
show(p)
```
will cause the exception:
```
.../bokeh/bokeh/util/serialization.py in encode_base64_dict(array)
    271     '''
    272     return {
--> 273         '__ndarray__'  : base64.b64encode(array.data).decode('utf-8'),
    274         'shape'        : array.shape,
    275         'dtype'        : array.dtype.name
TypeError: expected bytes, not memoryview
```

I don't consider this to be an error with bokeh, but should be documented so that downstream distributions can use the correct version (e.g. [CentOS 6.8 SCL](http://mirror.centos.org/centos/6/sclo/x86_64/rh/) only supports python 3.3, but they have updated to use bokeh 0.12.4).

danio:
- [X] issues: fixes #5796 
- [ ] tests added / passed

### DIFF
--- a/sphinx/source/docs/installation.rst
+++ b/sphinx/source/docs/installation.rst
@@ -20,7 +20,7 @@ Bokeh with plain Python lists of values.
 
 The Bokeh plot server does take advantage of NumPy, and may have a hard
 dependency on several compiled libraries.  The plot server requires
-a minimum python version of 2.7 or 3.3.
+a minimum python version of 2.7 or 3.4.
 
 For basic usage, have the following libraries installed:
 
--- a/sphinx/source/docs/installation.rst
+++ b/sphinx/source/docs/installation.rst
@@ -19,7 +19,8 @@ on any of the standard NumPy/SciPy stack.  It is entirely possible to use
 Bokeh with plain Python lists of values.
 
 The Bokeh plot server does take advantage of NumPy, and may have a hard
-dependency on several compiled libraries.
+dependency on several compiled libraries.  The plot server requires
+a minimum python version of 2.7 or 3.3.
 
 For basic usage, have the following libraries installed:
 
--- a/sphinx/source/docs/installation.rst
+++ b/sphinx/source/docs/installation.rst
@@ -12,15 +12,12 @@ for those who were unable to complete the process detailed in the
 Dependencies
 ============
 
-Because the Bokeh client library is mostly concerned with providing a nice
-Python interface for generating JSON objects which are then consumed by the
-BokehJS library running in the browser, there shouldn't be a *hard* dependency
-on any of the standard NumPy/SciPy stack.  It is entirely possible to use
-Bokeh with plain Python lists of values.
-
-The Bokeh plot server does take advantage of NumPy, and may have a hard
-dependency on several compiled libraries.  The plot server requires
-a minimum python version of 2.7 or 3.4.
+Bokeh is officially supported (and continuously tested) on CPython versions 2.7 
+and 3.4+ only. Other Python versions may function, possibly in limited capacity. 
+In particular, converting NumPy arrays to lists may be useful with other versions. 
+However, this guidance is only provided as-is, in case it happens to be useful,
+and does not imply any level of official support for other Python versions. All 
+issues opened related to unsupported Python versions will be closed as invalid.
 
 For basic usage, have the following libraries installed:
 
@@ -38,8 +35,17 @@ For basic usage, have the following libraries installed:
 To use the Bokeh server with python 2.7, you also need to install Futures
 package.
 
+Because the Bokeh client library is mostly concerned with providing a nice
+Python interface for generating JSON objects which are then consumed by the
+BokehJS library running in the browser, there shouldn't be a *hard* dependency
+on any of the standard NumPy/SciPy stack.  It is entirely possible to use
+Bokeh with plain Python lists of values. However, the Bokeh plot server does 
+make direct use of NumPy, and it is required to be installed for Bokeh apps
+to function. 
+
 Additionally the ``bokeh.charts`` interface and various examples
-depend on the Pandas library; it is recommended to install Pandas version 0.16.1 or later.
+depend on the Pandas library; it is recommended to install Pandas version 0.16.1 
+or later.
 
 .. _install_packages:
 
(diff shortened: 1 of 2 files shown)


sphinx/source/docs/installation.rst
```diff
@@ -19,7 +19,8 @@ on any of the standard NumPy/SciPy stack.  It is entirely possible to use
 Bokeh with plain Python lists of values.
 
 The Bokeh plot server does take advantage of NumPy, and may have a hard
-dependency on several compiled libraries.
+dependency on several compiled libraries.  The plot server requires
+a minimum python version of 2.7 or 3.3.
```
mattpap: 3.4?

mattpap:
To actually assert what versions of Python we support, I would add appropriate version ranges in `conda.recipe/meta.yaml` under `python` entries, though I'm not sure what the syntax is.

sphinx/source/docs/releases/0.12.4.rst
```diff
@@ -125,6 +125,13 @@ BokehJS has migrated to using ``.tsx`` templates. Provisional support for
 including ``.eco`` templates has been removed from both BokehJS and the
 Bokeh Python library.
 
+Python 3.3 No Longer Supported
```
danio: @mattpap this should probably be moved to 0.10

In the 0.12.4 relnotes I could note instead that the use of b64encode means np arrays no longer work?

mattpap:
> though I'm not sure what the syntax is.

It's something like `python >=2.7,<3|>=3.4` (`,` is `and` and `|` is `or` operator respectively).

sphinx/source/docs/releases/0.12.4.rst
mattpap: Not necessarily. 0.12.3 is the version that definitively broke support for 3.3.

danio: The b64encode change was made in commit 8e316faf2a12975c299bdde5fb002c2435f92642 29/Dec/2016; the 0.12.3 tag predates that (7/Oct/2016).

danio:
> To actually assert what versions of Python we support, I would add appropriate version ranges in conda.recipe/meta.yaml under python entries, though I'm not sure what the syntax is.

My concern with doing would be that it means 3.3 users cannot install bokeh, although it is only particular parts of bokeh that won't work for them. It's a more impactful change than just the documentation change so I will step out of that decision.

sphinx/source/docs/installation.rst
```diff
@@ -19,7 +19,8 @@ on any of the standard NumPy/SciPy stack.  It is entirely possible to use
 Bokeh with plain Python lists of values.
 
 The Bokeh plot server does take advantage of NumPy, and may have a hard
-dependency on several compiled libraries.
+dependency on several compiled libraries.  The plot server requires
+a minimum python version of 2.7 or 3.4.
```
bryevdv: I woulds suggest phrasing to the effect of "2.7 or 3.4+"

sphinx/source/docs/releases/0.12.4.rst
```diff
@@ -125,6 +125,12 @@ BokehJS has migrated to using ``.tsx`` templates. Provisional support for
 including ``.eco`` templates has been removed from both BokehJS and the
 Bokeh Python library.
 
+Used base64 serialisation
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Changes made to use b64encode mean that plotting numpy/pandas data no longer
+works with Python 3.3.
```
bryevdv: A workaround should be to convert the data to pure python lists, which is also probably worth mentioning.

bryevdv:
> To actually assert what versions of Python we support, I would add appropriate version ranges in conda.recipe/meta.yaml under python entries, though I'm not sure what the syntax is.
 
While that seems nice in an ideal sense, I am not so sure about practically. Making an actual restriction adds complications, even more so if we wanted to try an do it for pypi packages too. And if we dont do it for pypi packages then we have a weird inconsistency. Also as stated Bokeh is partially usable on 3.3. I think a documented "this is what we actually" support is sufficient, and then people can make their own choices.

mattpap:
> While that seems nice in an ideal sense,

In the ideal world we wouldn't publish pypi packages at all. However, as we don't live in one and we publish such, but that shouldn't mean that we can't provide full specification of bokeh's dependencies in a conda package, if conda allows to express such.

bryevdv:
Bokeh is partially usable by python 3.3. Basic charts with lists, etc. are not problem. I think it is a mistake to pre-emptively stop people who would want to try to use it to the extent that it is possible, from trying it at all. I'm happy to point to a clear compatibility statement and close issues raised for things outside those boundaries. 

Also, my intention is absolutely to move to "new" conda no-arch packages as soon as the market penetration of recent conda versions that support them increases. We can't have any conditions on python or platform, in order to do that.

bryevdv:
What I would be more amenable to consider is a version check and warning on import for python 3.0 - 3.3

bryevdv:
@danio would you be able to add a "version check" to `bokeh/__init__.py` ? It should check that the version is 2.7 or 3.4+ and I think the warning should say something like:

> Bokeh is officially supported on cpython versions 2.7 and 3.4+ only. Other Python versions may function in limited capacity. In particular, converting NumPy arrays to lists may be useful with other versions. However, this guidance is only provided as-is, in case it happens to be useful, and does not imply any level of official support for other Python versions.  All issues opened related to other versions will be closed as invalid. 

I guess ideally it could check `platform.python_implementation()` to be `CPython` as well.

bryevdv:
@mattpap I have updated the files here to reflect the official support policy. For the time being, I am happy to just point to here, as needed. If it becomes a recurring issue, we can add the version check and warning to `bokeh/__init__.py`.

PR #5798, opened 2017-01-26
Use Element.dataset instead of $(...).data()

### LINKED ISSUE #5778: autoload_static uses Jquery .data()
This function

https://api.jquery.com/jquery.data/

was used in `embed.coffee`:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/embed.coffee#L147

mattpap:
fixes #5778

### DIFF
--- a/bokehjs/src/coffee/embed.coffee
+++ b/bokehjs/src/coffee/embed.coffee
@@ -144,7 +144,7 @@ export inject_raw_css = (css) ->
 
 # pull missing render item fields from data- attributes
 fill_render_item_from_script_tag = (script, item) ->
-  info = script.data()
+  info = script.dataset
   # length checks are because we put all the attributes on the tag
   # but sometimes set them to empty string
   if info.bokehLogLevel? and info.bokehLogLevel.length > 0


mattpap:
@bryevdv, in the issue you said this needs tests and there are such, but what we really need is an example in `examples/embed` showing how to use this and this will test if this works on all levels. Just I'm not sure exactly how this is supposed to work.

bryevdv:
I'm not sure, such an example will need to save a file, and then run a server. I'm no sure how easily such a thing would fit into our examples testing. I agree such an example would be valuable in any case. I was more thinking of some unit tests for `embed.coffee` (of which there are currently zero)

bryevdv:
@mattpap here is a self-contained example you can add, it does not save any files, and running just the one script is all that is needed:
```
from jinja2 import Template

from tornado.web import Application, RequestHandler

from bokeh.embed import autoload_static
from bokeh.resources import CDN

template = Template("""
<!doctype html>

<html lang="en">
<head>
  <meta charset="utf-8">
</head>

<body>
  <div>HERE IS SOME CUSTOM STUFF</div>
  {{ script|safe }}
</body>
</html>
""")

class IndexHandler(RequestHandler):
    def initialize(self, script):
        self.script = script
    def get(self):
        self.write(template.render(script=self.script))

class JSHandler(RequestHandler):
    def initialize(self, js):
        self.js = js
    def get(self): self.write(self.js)

def make_plot():

    p.circle(flowers["petal_length"], flowers["petal_width"],

    return p

if __name__ == '__main__':
    print('Opening Tornado app with embedded Bokeh plot on http://localhost:8888/')

    js, script = autoload_static(make_plot(), CDN, "embed.js")

    app = Application([
        (r"/", IndexHandler, dict(script=script)),
        (r"/embed.js", JSHandler, dict(js=js))
    ])
    app.listen(8888)

    io_loop = IOLoop.current()
    io_loop.add_callback(view, "http://localhost:8888/")
    io_loop.start()
```

bryevdv:
@mattpap please make a separate issue to add tests/examples. #5801 is a critical issue with GMaps has come out of nowhere and is affecting multiple versions of Bokeh. I have a possible, potential fix but the problem is only showing in more complicated embeds like docs, which I can't currently view because of this issue. I am merging this immediately to unblock work on #5801

PR #5669, opened 2017-01-07
fix Rect rendering with log axis

clairetang6:
This PR is to make the Rect glyph behave better when used with log axes. 

Previously, Rect rendering relied on mapping the center point (x, y) to screen space and then rendering a rectangle with the top left corner point (sx - sw/2, sy - sh/2) and a width and height of sw and sh (mapped from w, h). This is a problem for log axis where the center point of the rectangle is not exactly in the center in screen space. 

This PR fixes this by computing the top left corner point in data space, then mapping that point to screen space, and using that mapped point for rendering.

I added some tests, but then realized we need a screenshot test here. Also, hit-testing is not working exactly right yet (because it also uses sw/2 and sh/2). 

- [x] issues: fixes #5336
- [x] tests added / passed
~~- [ ] release document entry (if new feature or API change)~~
- [x] fix hit-testing for log axis
- [x] add screenshot test

PR #5799, opened 2017-01-26
Revert "fix Rect rendering with log axis"

bryevdv:
Reverts bokeh/bokeh#5669


bryevdv:
Examples tests with rects now restored:

https://s3.amazonaws.com/bokeh-travis/195566287/examples.html

PR #5816, opened 2017-01-28
improve the helpstring for CDS selected property

### LINKED ISSUE #5718: Bad formatting in selected docstring
Ends up rendering like this

screen shot 2017-01-13 at 10 38 03 am

Guessing bad use of bullets is causing some text to be bold and the whole thing to be unclear.

While there the wording in this docstring could be a lot clearer.

bryevdv:
fixes #5718

New section renders like this:

screen shot 2017-01-28 at 11 03 55 am

@birdsarah is this reasonable, is there anything else to add?

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -25,21 +25,29 @@ class DataSource(Model):
     }, help="""
     A dict to indicate selected indices on different dimensions on this DataSource. Keys are:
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Well, I think there's more to say but this should do until a more comprehensive improvement to selections can be made

PR #5828, opened 2017-01-31
TableDataSource: the filterable CDS

### LINKED ISSUE #4070: Plotting with subsets (row-wise) of ColumnDataSources (like filtered data)

clairetang6:
This PR is still in its early stages, so feedback on the more basic design and API would be appreciated. 

The goal of this PR is to create a filterable CDS. I wrote about the current pain point and a potential solution in [this wiki working document page](https://github.com/bokeh/bokeh/wiki/Filterable-Data-Source). I just updated it to make it current and to reflect the API I implemented in this PR so far. 

The most important thing for this feature is to separate the data from the filters, so that data would not need to be duplicated. The approach that I took was to create a new DataSource, the TableDataSource, which has an instance of a CDS as a property to hold its data, instead of having a data property. Another approach could be to separate Data and DataSources for all types of DataSources (i.e. make ColumnData into its own model that just holds the data). Then this type of filtering behavior can be given to the CDS.

The TDS is basically a data source with a `cds` and `filter` property. The `cds` is a CDS. The `filter` was initially implemented (first commit) as a `Seq(Int)` and was the row indices to use for the subset. I then made it so that the `filter` could also be a `Seq(Bool)` and also allowed it to be a callable when a TableDataSource is created through its own constructor. I would need some guidance on how to make a new property type that is a callable that returns some other type. 

I implemented a basic `__getitem__` that has two cases. `tds['a']` returns its cds's data['a']. `tds[bool_list]` returns a new tds with a filter of `bool_list` (a list of bools for filtering rows). This allows for the API:

```python
fig.circle(x='x', y='y', source=tds[tds['x'] > 10])
``` 

Example of the result:

tds

```python
from bokeh.models.sources import ColumnDataSource, TableDataSource

output_file("tds.html")

data = {'x': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'y': [10, 5, 2, 6, 1, 4, 8, 2, 7, 3]}

# create a TDS from dict, creating another TDS through the first links them.
tds1_1 = TableDataSource(data)
tds1_2 = tds1_1[tds1_1['y'] == 4]

# create a TDS from a CDS. All TDS's associated with the same CDS are linked.
cds = ColumnDataSource(data)
tds2_1 = TableDataSource(cds)
tds2_2 = TableDataSource(cds, filter=[0, 1, 9]) # filter = Seq(Int)
tds2_3 = TableDataSource(cds, filter=[True, False]*5) # filter = Seq(Bool)
tds2_4 = TableDataSource(cds, filter=(lambda tds: tds['y'] < 4)) # filter = (self) -> Either(Seq(Int), Seq(Bool))
tds2_5 = tds2_1[tds2_1['y'] < 8] # created through __getitem__

plot_size_and_tools = {'plot_height': 220, 'plot_width': 220,
                        'tools':['box_select', 'reset'],
                        'x_range':[0, 11], 'y_range':[0, 11]}

fig1 = figure(title="TDS 1_1, created with data", **plot_size_and_tools)
fig1.circle(x='x', y='y', source=tds1_1, color='black')

fig2 = figure(title="TDS 1_2, created by TDS 1_1", **plot_size_and_tools)
fig2.circle(x='x', y='y', source=tds1_2, color='black')

fig3 = figure(title="Full data set, CDS", **plot_size_and_tools)
fig3.circle(x='x', y='y', source=cds, color='green')

fig4 = figure(title="Full data set, TDS 2_1", **plot_size_and_tools)
fig4.circle(x='x', y='y', source=tds2_1, color='green')

fig5 = figure(title="TDS 2_2, filter=Seq(Int)", **plot_size_and_tools)
fig5.circle(x='x', y='y', source=tds2_2, color='green')

fig6 = figure(title="TDS 2_3, filter=Seq(Bool)", **plot_size_and_tools)
fig6.circle(x='x', y='y', source=tds2_3, color='green')

fig7 = figure(title="TDS 2_4, filter -> Seq(Bool)", **plot_size_and_tools)
fig7.circle(x='x', y='y', source=tds2_4, color='green')

fig8 = figure(title="TDS 2_5, created by TDS 2_1", **plot_size_and_tools)
fig8.circle(x='x', y='y', source=tds2_5, color='green')

show(column(row([fig1, fig2, fig3, fig4]), row(fig5, fig6, fig7, fig8)))
```

- [ ] issues: fixes #4070
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import warnings
+import six
 
 from ..core.has_props import abstract
 from ..core.properties import Any, Bool, ColumnData, Dict, Enum, Instance, Int, JSON, List, Seq, String
@@ -326,6 +327,17 @@ class TableDataSource(ColumnarDataSource):
     filter = Seq(Int, default=[])
 
     cds = Instance(ColumnDataSource)
+
+    def __getitem__(self, key):
+        import numpy as np
+        if isinstance(key, six.string_types):
+            colname = key
+            return np.array(self.cds.data[colname])
+        else:
+            key = list(key) #TODO check whether list is correct length, support int indexing
+            n = list(set(len(x) for x in self.cds.data.values()))[0]
+            filter = [i for i in range(n) if key[i] == True]
+            return TableDataSource(cds=self.cds, filter=filter)
 
 class GeoJSONDataSource(ColumnarDataSource):
     '''
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -4,7 +4,7 @@ import warnings
 import six
 
 from ..core.has_props import abstract
-from ..core.properties import Any, Bool, ColumnData, Dict, Enum, Instance, Int, JSON, List, Seq, String
+from ..core.properties import Any, Bool, ColumnData, Dict, Either, Enum, Instance, Int, JSON, List, Seq, String
 from ..model import Model
 from ..util.dependencies import import_optional
 from ..util.warnings import BokehUserWarning
@@ -324,9 +324,31 @@ class ColumnDataSource(ColumnarDataSource):
 
 class TableDataSource(ColumnarDataSource):
 
-    filter = Seq(Int, default=[])
+    filter = Either(Seq(Int), Seq(Bool), default=[])
 
     cds = Instance(ColumnDataSource)
+
+    def __init__(self, *args, **kw):
+        if len(args) == 1 and "cds" not in kw:
+            cds_or_data = args[0]
+            if isinstance(cds_or_data, ColumnDataSource):
+                kw["cds"] = cds_or_data
+            else:
+                data = cds_or_data
+                kw["cds"] = ColumnDataSource(data)
+
+        n = list(set(len(x) for x in kw["cds"].data.values()))[0]
+
+        filter = kw.pop("filter", list(range(n)))
+        if callable(filter):
+            kw["filter"] = list(range(n))
+        else:
+            kw["filter"] = filter
+
+        super(TableDataSource, self).__init__(**kw)
+
+        if callable(filter):
+            self.filter = list(filter(self))
 
     def __getitem__(self, key):
         import numpy as np
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -321,6 +321,12 @@ class ColumnDataSource(ColumnarDataSource):
 
         self.data._patch(self.document, self, patches, setter)
 
+class TableDataSource(ColumnarDataSource):
+
+    filter = Seq(Int, default=[])
+
+    cds = Instance(ColumnDataSource)
+
 class GeoJSONDataSource(ColumnarDataSource):
     '''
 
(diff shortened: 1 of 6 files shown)


clairetang6:
I am also keeping groupby behavior in mind. I think groupby is more similar to selection/nonselection in that you are still dealing with the full column, just broken into groups, versus plotting only a subset of data. One possibility would be to do something similar to selection/nonselection, where you create glyphs for each group (used for their visuals in rendering). Then in the render function of the glyph_renderer, you get the group indices from somewhere (maybe the tds) and render the different groups with those indices.

bryevdv:
>  Another approach could be to separate Data and DataSources for all types of DataSources (i.e. make ColumnData into its own model that just holds the data). Then this type of filtering behavior can be given to the CDS.

I will say that this idea is one that I have had for a long time, but have never had the chance to act on. I had imagined splitting CDS into `ColumnStore` (or something), which actually has the data/columns, and the CDS that has references to columns in the store (but would otherwise be identical from user's perspective). 

I think it's worth considering this approach, but I recognize it might end up being more work than we decide we want to do at the moment.

bryevdv:
by #5933

YunfengHu:
I like the TableDataSource idea a lost and the example on wiki is great. I was trying to import TableDataSource, but I got the error ImportError: cannot import name 'TableDataSource'. Can @clairetang6 you show me how to install it?

PR #5833, opened 2017-02-01
bump conda and conda-build versions

### LINKED ISSUE #5818: Conda convert windows packages cause an error
Reference: #5814 

According to @kalefranz

> For a workaround, either remove the info/paths.json file in the package tarball itself (if you have access to it on anaconda.org), or remove it locally after it's been unpacked to

bryevdv:
issues: fixes #5818


mattpap:
Maybe this will also help with 3.6?

PR #5837, opened 2017-02-02
fix selection manager tests after getter/setter PR 5106

### LINKED ISSUE #5121: Fix skipped JS tests as result of getter/setter work

clairetang6:
fixes #5121 

These tests originally relied on the `get` method which was removed in PR [5106](https://github.com/bokeh/bokeh/pulls/5106). I updated them and added a comment about why the stub objects are the way there are.


mattpap:
btw. @clairetang6, what code editor do you use? I see those "remove trailing whitespace" are recurring and most code editors allow to remove trailing whitespace automatically. There aren't that many file formats  that require significant whitespace (on top of my head I can only recall signature sparator in E-mails, i.e. `-- `).

PR #5838, opened 2017-02-02
use sinon.test to increase layout test robustness

### LINKED ISSUE #4539: JS tests reporting as failed even when they all passed
I've been seeing this a fair amount over the last few days. The js test is failing even though all the js tests passed. Here's what that output looks like:

``` console
  Jitter transform module

    Jitter with uniform

      ✓ should average the fixed values (57ms)

    Jitter with normal

      ✓ should average the fixed values (48ms)

  Paragraph.View render

    ✓ should set the margin to 0

  Tabs.Model

    ✓ should have children matching tabs.child after initialization

    ✓ should return children from get_layoutable_children

    ✓ get_constraints should return constraints from children

    ✓ get_edit_variables should return edit_variables from children

    ✓ get_constraints should return constraints from children

  Widget.Model

    ✓ should should return 8 constraints

  621 passing (10s)

  8 pending

INFO:tests.plugins.file_server:Shutting down the webserver

F

___________________________ TestBokehJS.test_bokehjs ___________________________

self = <tests.test_bokehjs.TestBokehJS testMethod=test_bokehjs>

    def test_bokehjs(self):

        os.chdir('bokehjs')

        proc = subprocess.Popen([join('node_modules', '.bin', 'gulp'), "test"],

                                stdout=subprocess.PIPE)

        result = proc.wait()

        msg = proc.stdout.read().decode('utf-8', errors='ignore')

        print(msg)

        if result != 0:

>           assert False

E           AssertionError: assert False

../tests/test_bokehjs.py:21: AssertionError

====================== 970 tests deselected by "-m 'js'" =======================

================== 1 failed, 970 deselected in 30.65 seconds ===================

The command "if [[ -z "$TRAVIS_TAG" && "$GROUP" == unit ]]; then py.test -s -m js; fi" exited with 1.
```

Something is causing the python test that runs the bokehjs to think the test failed. Here's the full log for that run: https://travis-ci.org/bokeh/bokeh/jobs/138147284#L2311

clairetang6:
issues: fixes #4539 

In this PR, I wrapped test functions with `sinon.test`. From ["Best Practices for Spies, Stubs and Mocks in Sinon.js"](https://semaphoreci.com/community/tutorials/best-practices-for-spies-stubs-and-mocks-in-sinon-js):

> Use sinon.test Whenever Possible
When you use spies, stubs or mocks, wrap your test function in sinon.test. This allows you to use Sinon's automatic clean-up functionality. Without it, if your test fails before your test-doubles are cleaned up, it can cause a cascading failure - more test failures resulting from the initial failure. Cascading failures can easily mask the real source of the problem, so we want to avoid them where possible.

These changes fix the errors from the layout tests for me. @mattpap

### DIFF
--- a/bokehjs/test/models/annotations/annotation.coffee
+++ b/bokehjs/test/models/annotations/annotation.coffee
@@ -1,6 +1,5 @@
 {expect} = require "chai"
 utils = require "../../utils"
-sinon = require 'sinon'
 
 {SidePanel} = utils.require("core/layout/side_panel")
 
--- a/bokehjs/test/models/axes/axis.coffee
+++ b/bokehjs/test/models/axes/axis.coffee
@@ -102,8 +102,8 @@ describe "AxisView", ->
   it "_axis_label_extent should be 0 if no axis_label", ->
     expect(@axis_view._axis_label_extent()).to.be.equal 0
 
-  it "_get_size should return sum of _tick_extent, _axis_label_extent, and _tick_label_extent", ->
-    sinon.stub(@axis_view, '_tick_extent', () -> 0.11)
-    sinon.stub(@axis_view, '_axis_label_extent', () -> 0.11)
-    sinon.stub(@axis_view, '_tick_label_extent', () -> 0.11)
+  it "_get_size should return sum of _tick_extent, _axis_label_extent, and _tick_label_extent", sinon.test ->
+    this.stub(@axis_view, '_tick_extent', () -> 0.11)
+    this.stub(@axis_view, '_axis_label_extent', () -> 0.11)
+    this.stub(@axis_view, '_tick_label_extent', () -> 0.11)
     expect(@axis_view._get_size()).to.be.equal 0.33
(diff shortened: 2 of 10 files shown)


mattpap:
This seems to work. I didn't observe any failures after running tests 40~50 times. However, I'm still not convinced this a complete fix. It would be nice to know how `sinon.test()` works under failure, e.g. what happens if an exception in thrown in `window.resize` callback, where, if at all, a failure is reported.

bryevdv:
@mattpap I agree there is more to do but this seems to be a clear improvement. I would not vote to wait on it, if that is what you are proposing.

bryevdv:
I will merge this tomorrow unless there is comment/objection.

mattpap:
I did some experimentation with this and it doesn't behave well under failure, but for now it's good enough.

mattpap:
@clairetang6, make sure to assign status label to your future PRs.

Issue #5835, opened 2017-02-01
County Data: most independent cities in Virginia have 'city' left out, which causes confusion when trying to match based on county name


adamwlev:
Hi, love Bokeh, thanks for the great tool.

Just minor comment about some of the names of the counties in the counties data. In Virginia, there are independent cities that are sort of distinct counties. In the data, these cities are present as independent entities, which works when matching by FIPS code. However, if matching by name, some of these cities have identical names as the 'county' counterpart despite being located in very different places. For example: Richmond County and Richmond City are both named 'Richmond', and Fairfax County and Fairfax City are both recorded with the name 'Fairfax'. Other examples are 'Bedford' and 'Franklin'. Also in Maryland, a similar problem exists with Baltimore County and Baltimore City both having names 'Baltimore'.

I would recommend specifying 'city' for these entities that represent independent cities. This makes it possible to match counties on name of entity and state.

Thank you!

adamwlev:
Add 'Roanoke' to that list as both Roanoke City and Roanoke County are under the name 'Roanoke'.

bryevdv:
Hi @adamwlev, I'm missing a bit of context here, but I think you are making a point about the "US Counties" sample data? Perhaps it's a point we need to be clearer on, but the sample data is just that --- sample data. It is made available primarily as a means to illustrate techniques, or how something can be accomplished. There is no implied endorsement regarding its efficacy or correctness. The current data was taken **as-is** from these Google Fusion tables here:

https://support.google.com/fusiontables/answer/1182141?hl=en

In particular for the counties:

https://fusiontables.google.com/data?docid=1xdysxZ94uUFIit9eXmnw1fYc6VcQiXhceFd_CVKa#map:id=2

I am afraid we don't have the resources to do much else. If you know of a better source, or would like to make specific corrections to the existing sample data, we would certainly welcome pull requests.

adamwlev:
Hey @bryevdv, thanks for the response and the info.

The data source is perfect. It's just that the api put forth by bokeh/sampledata/us_counties.py should just allow the user access to the 'Geographic Name' as well as the 'Name', since the 'Name' field is redundant is some rare cases.

Would love to write a pull request to add this field to the api. I'll work on it and submit it to show better what I am talking about.

bryevdv:
@adamwlev oh I see, well that definitely sounds easier, I look forward to see your proposed additions!

PR #5839, opened 2017-02-02
add detailed name in counties sampledata api to enable matching by name

adamwlev:
Deals with #5835. I did not add any tests. I put relevant info in the docstring of the us_counties module.

I know this is just sample data but it is actually an incredibly useful source of data. This PR enables people to match to counties by name and state (in the case that the FIPS code is not in the data). Here is some code showing the data contains duplicates on 'state' and 'name':

dups

and here is an example of an occurrence of a duplicate in the data source:

example

This issue came about because I had some data without FIPS codes for the states and counties but had the names of the states and counties. I happened to be plotting Virginia and I immediately noticed that some of the counties were not being found in the sample data:

before

Then I did my little hack that I propose here so that I could match all the counties:

after

Let me know what you think about having both the 'name' and 'detailed name'. For me I know it was very useful..

### DIFF
--- a/bokeh/sampledata/us_counties.py
+++ b/bokeh/sampledata/us_counties.py
@@ -4,9 +4,13 @@ indexed by the two-tuple containing (state_id, county_id) and has the following 
 associated value:
 
     data[(1,1)]['name']
+    data[(1,1)]['state']
+    data[(1,1)]['detailed name']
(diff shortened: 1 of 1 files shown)


bryevdv:
This looks good to me, thanks @adamwlev !

PR #5844, opened 2017-02-03
Allow to debug tests with `gulp test --debug`

### LINKED ISSUE #5720: It's not possible to debug tests anymore since #5659
You can start a debugger and interact with it, but setting breakpoints in code with `debugger` statement doesn't stop the debugger. It works when using gulp-mocha.

mattpap:
fixes #5720

### DIFF
--- a/bokehjs/gulp/tasks/test.coffee
+++ b/bokehjs/gulp/tasks/test.coffee
@@ -2,6 +2,7 @@ gulp = require "gulp"
 gutil = require "gulp-util"
 through = require "through"
 cp = require "child_process"
+argv = require("yargs").argv
 
 paths = require "../paths"
 utils = require "../utils"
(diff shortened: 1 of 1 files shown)


mattpap:
Note that you shouldn't use `node debug $(which gulp) test` for debugging tests anymore. Anyway, `gulp test --debug` is shorter and makes more semantic sense, because you don't want to debug the build, but bokehjs + tests.

bryevdv:
@mattpap what does this actually do? I'm not familiar with `node debug` or what it does. Should there be information in the dev guide about this?

mattpap:
@bryevdv, `node debug script args` allows you to debug a script with a simple pdb-like debugger. It's useful to have if you need to debug build or test issues. `debug` doesn't propagate well to spawned children, so we have to proceed as in this PR. Alternatively one can use a graphical debugger (basically chrome dev tools) via node-inspector, but it possesses the same issues with spawned processes.

PR #5849, opened 2017-02-05
Ensure all dataspec properties are updated before glyph render

### LINKED ISSUE #5811: Vbar width not updating correctly when too many things updated through `push_notebook`
Running Bokeh 0.12.4, Jupyter 4.3 (but this also happened with 4.2), and it happens both on a 2.7 kernel and a 3.5 kernel.

I have a Jupyter notebook with some interactive charts based on a few filters. I created an example to illustrate the issue in [this gist](https://gist.github.com/aiguofer/927e9d71f6e1e8ea48ade18056ab3322):

In there, there's 2 notebooks. One that behaves as expected, and one that does not. Basically, I calculate the width for some vbars depending on how the data is grouped. The width is correctly updated when there's not too many objects being updated by `push_notebook`. It seems like it works if I have 1 figure with 2 vbars and one line, or if I have 1 figure with 2 vbars and a separate DataTable. However, if I have the 2 vbars, line, and DataTable I start seeing issues.

I tried looking at other bugs to see if they could be related, and it seems like https://github.com/bokeh/bokeh/issues/4911 might have something to do with it?

Here's an example of what I'm seeing even though the width for both bar glyphs is being set to the same value:

selection_015

selection_014

aiguofer:
Currently, if only dataspec properties change for a given glyph, when `push_notebook` is called, the graphs are not updated because the `View`'s properties never get updated. By calling `set_data` in the `render` function, it makes sure that all dataspec properties of the given `Glyph` are updated with the new properties of the `Model`.

- [x] issues: fixes #5811
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -118,8 +118,9 @@ export class GlyphRendererView extends RendererView
     if @model.visible == false
       return
 
-    t0 = Date.now()
+    extend(@glyph, @model.glyph.materialize_dataspecs(@model.data_source))
 
+    t0 = Date.now()
     glsupport = @glyph.glglyph
 
     tmap = Date.now()


bryevdv:
@aiguofer FYI we are about to release `0.12.5` I have marked this issue for `0.12.6` I'll be available to return to discussing/helping in about two weeks.

bryevdv:
FYI haven't forgotten this but theres a big refactor in another PR that needs to land first.

aiguofer:
Awesome, thanks! I haven't had much time to look at this recently, but would be happy to help as I can. I'll probably need to look over recent changes that might affect this.

bryevdv:
@aiguofer good news, this problem seems to have resolved itself as of some recent foundational work:

screen shot 2017-05-11 at 16 51 14
screen shot 2017-05-11 at 16 51 07

accordingly I am closing this PR

PR #5850, opened 2017-02-06
demonstrate argparse usage that works across platofrms

### LINKED ISSUE #5834: Need to use different sintaxis when using command "bokeh serve" in Windows command line.
This issue was originally commented and resolved [here](http://stackoverflow.com/questions/41936031/bokeh-server-not-working-when-deployed-from-windows-server/41980870#41980870). One of the possible arguments to use is `--host='*'` to allow any host. The use of this argument appears somewhere in the documentation I do not find right now, anyway in Linux it works properly. But when used in Windows Command Prompt, not any host is allowed to connect as it is supposed to be. After making some tests with the sintaxis in Windows, the conclussion is that correct way to use arguments is `--host '*'` or `--host=*`. 
Using `--host='*'` is literally interpreted in Windows, hence no wildcard is applied. This issue is probaly obvious for advanced users, but for beginners as me may be not so clear. Thanks to @bryevdv for the help.

bokeh2 1
bokeh4

bryevdv:
issues: fixes #5834 

Usage such as `--host='*'` fails on windows, but not OSX or Linux. Only demonstrate usage such as `--foo bar`, which will work on all platforms, instead of `--foo=bar`

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -85,7 +85,7 @@ argument:
 
 .. code-block:: sh
 
-    bokeh serve app_script.py --port=8080
+    bokeh serve app_script.py --port 8080
 
 To listen on an arbitrary port, pass ``0`` as the port number.  The actual
 port number will be logged at startup.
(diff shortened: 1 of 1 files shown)


bryevdv:
Failures are spurious and will resolve when merged into master.

PR #5851, opened 2017-02-06
Document: fix memory leak by explicitly deleting sys.modules on session destroy

### LINKED ISSUE #5848: Memory leak in bokeh application
In current master, I noticed a memory leak when using a bokeh application. It's only slowly increasing, thus making it hard to detect. I bisected it to commit 18802dd4fe4915d146 (PR #5425).

When I add a `print(len(sys.modules))`, it keeps increasing on every page load, and when I comment the line https://github.com/bokeh/bokeh/pull/5425/files#diff-dc51006b23833be68da1b9ca8a8167a6R63 there's no leak.
Adding a print to [`document.__del__`](https://github.com/bokeh/bokeh/pull/5425/files#diff-665e6cdee790b076b7b90d3ddd57f504R305), I see that it never gets called.

I'm using python 3.5.2 and I have the same issue on a server running python 3.4.3.

@bryevdv @mattpap

bkueng:
Previously, due to circular references, a Document was never GC'd, and thus
sys.modules kept increasing for each page request.

This patch explicitly cleans up sys.modules after a session is destroyed.

Verified that `len(sys.modules)` does not increase over time (which it did before).

Fixes #5848

TODO: add tests

### DIFF
--- a/bokeh/server/tests/test_session.py
+++ b/bokeh/server/tests/test_session.py
@@ -1,5 +1,4 @@
 import mock
-import pytest
 
 from bokeh.document import Document
 
@@ -40,5 +39,3 @@ def test_destroy():
         s.destroy()
         assert s.destroyed
         docroc.assert_called_with(s)
-
-
--- a/bokeh/server/tests/test_session.py
+++ b/bokeh/server/tests/test_session.py
@@ -1,0 +1,44 @@
+import mock
+import pytest
+
+from bokeh.document import Document
+
+import bokeh.server.session as bss
+
+def test_creation():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    assert s.id == 'some-id'
+    assert s.document == d
+    assert s.destroyed is False
+    assert s.expiration_requested is False
+    assert s.expiration_blocked == 0
+
+def test_subscribe():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    assert s.connection_count == 0
+    s.subscribe('connection1')
+    assert s.connection_count == 1
+    s.subscribe('connection2')
+    assert s.connection_count == 2
+    s.unsubscribe('connection1')
+    assert s.connection_count == 1
+    s.unsubscribe('connection2')
+    assert s.connection_count == 0
+
+def test_destroy():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    with mock.patch('bokeh.document.Document.delete_modules') as docdm:
+        s.destroy()
+        assert s.destroyed
+        docdm.assert_called_once()
+
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    with mock.patch('bokeh.document.Document.remove_on_change') as docroc:
+        s.destroy()
+        assert s.destroyed
+        docroc.assert_called_with(s)
+
+
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -97,7 +97,9 @@ class Document(object):
         self._modules = []
         self._template_variables = {}
 
-    def __del__(self):
+    def delete_modules(self):
+        ''' Clean up sys.modules after the session is destroyed.
+        '''
         for module in self._modules:
             if module.__name__ in sys.modules:
                 del sys.modules[module.__name__]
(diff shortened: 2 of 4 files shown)


bryevdv:
I like that this also removes `__del__`, the presence of `__del__` actually makes the GC more conservative and may have in and of itself contributed to the cycle not being reaped. I will try to add some tests to this today and then we can get it merged.

bryevdv:
@bkueng unfortunately it's not letting me push changes to your PR branch, but here is a diff you can apply that adds some minimal tests

```
diff --git a/bokeh/server/session.py b/bokeh/server/session.py
index 6c53dcf..5474009 100644
--- a/bokeh/server/session.py
+++ b/bokeh/server/session.py
@@ -111,8 +111,8 @@ class ServerSession(object):
         return self._expiration_blocked_count

     def destroy(self):
-        self._document.delete_modules()
         self._destroyed = True
+        self._document.delete_modules()
         self._document.remove_on_change(self)
         self._callbacks.remove_all_callbacks()

diff --git a/bokeh/server/tests/test_session.py b/bokeh/server/tests/test_session.py
new file mode 100644
index 0000000..d71f0e4
--- /dev/null
+++ b/bokeh/server/tests/test_session.py
@@ -0,0 +1,44 @@
+import mock
+import pytest
+
+from bokeh.document import Document
+
+import bokeh.server.session as bss
+
+def test_creation():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    assert s.id == 'some-id'
+    assert s.document == d
+    assert s.destroyed is False
+    assert s.expiration_requested is False
+    assert s.expiration_blocked == 0
+
+def test_subscribe():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    assert s.connection_count == 0
+    s.subscribe('connection1')
+    assert s.connection_count == 1
+    s.subscribe('connection2')
+    assert s.connection_count == 2
+    s.unsubscribe('connection1')
+    assert s.connection_count == 1
+    s.unsubscribe('connection2')
+    assert s.connection_count == 0
+
+def test_destroy():
+    d = Document()
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    with mock.patch('bokeh.document.Document.delete_modules') as docdm:
+        s.destroy()
+        assert s.destroyed
+        docdm.assert_called_once()
+
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    with mock.patch('bokeh.document.Document.remove_on_change') as docroc:
+        s.destroy()
+        assert s.destroyed
+        docroc.assert_called_with(s)
+
+
diff --git a/bokeh/tests/test_document.py b/bokeh/tests/test_document.py
index e1a56b6..f0c6f46 100644
--- a/bokeh/tests/test_document.py
+++ b/bokeh/tests/test_document.py
@@ -37,6 +37,20 @@ class TestDocument(unittest.TestCase):
         assert not d.roots
         assert d.template_variables == {}

+    def test_delete_modules(self):
+        d = document.Document()
+        assert not d.roots
+        class FakeMod(object):
+            __name__ = 'junkjunkjunk'
+        mod = FakeMod()
+        import sys
+        assert 'junkjunkjunk' not in sys.modules
+        sys.modules['junkjunkjunk'] = mod
+        d._modules.append(mod)
+        assert 'junkjunkjunk' in sys.modules
+        d.delete_modules()
+        assert 'junkjunkjunk' not in sys.modules
+
     def test_add_roots(self):
         d = document.Document()
         assert not d.roots
```

bkueng:
Strange, I have the `Allow edits from maintainers.` option set

bokeh/server/tests/test_session.py
```diff
@@ -0,0 +1,44 @@
...
+        assert s.destroyed
+        docdm.assert_called_once()
+
+    s = bss.ServerSession('some-id', d, 'ioloop')
+    with mock.patch('bokeh.document.Document.remove_on_change') as docroc:
+        s.destroy()
+        assert s.destroyed
+        docroc.assert_called_with(s)
+
+
```
mattpap: Those two lines have to be removed.

bokeh/server/tests/test_session.py
```diff
@@ -0,0 +1,44 @@
+import mock
+import pytest
```
mattpap: Unnecessary import.

bryevdv:
@bkueng I'm sure it's some operator error on my part. I have no doubt that editing the PR online in the web UI would have worked. But however I cloned your fork did not happen to be the correct way for collaborating. THanks for the PR!

Issue #5871, opened 2017-02-14
ColumnDataSource constructor slow


philippjfr:
Currently the ColumnDataSource constructor is quite inefficient because it iterates over all the columns and calls ``ColumnDataSource.add`` for each of them, which ends up triggering validation for all the columns each time due to the way the data, which is a ``PropertyValueDict``, behaves. In other words validation is triggered on the triangle number of the number of columns.

As far as I can tell there is no reason to use ``ColumnDataSource.add`` in the constructor because it really doesn't do anything except giving columns without names new unique names. Since the constructor only accepts dataframe and dict types there appears to be no case where you'd end up with clashing or non-existent column names in the constructor. Therefore it should be possible to simply use ``self.data.update`` to validate the data once and no more.

Here is some small bit of profiling:

```python
from bokeh.models import ColumnarDataSource
data = {chr(65+c): np.random.rand(1000) for c in range(50)}

%%timeit
cds = ColumnDataSource(data=data)
```

Current speed:

> 1 loop, best of 3: 563 ms per loop

Using ``self.data.update``:

>10 loops, best of 3: 21.6 ms per loop

PR #5872, opened 2017-02-14
ColumnDataSource constructor optimization

philippjfr:
As described in https://github.com/bokeh/bokeh/issues/5871, the ColumnDataSource constructor triggers validation on all the columns for every column that is passed to the constructor, which is highly inefficient. By using ``self.data.update`` the validation is only triggered once resulting in considerably speed improvements for large datasets.

It would be important to know if there is any particular reason why it was using ``.add`` in the first place.

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -105,8 +105,8 @@ class ColumnDataSource(ColumnarDataSource):
             else:
                 raise ValueError("expected a dict or pandas.DataFrame, got %s" % raw_data)
         super(ColumnDataSource, self).__init__(**kw)
-        for name, data in raw_data.items():
-            self.add(data, name)
+        self.column_names[:] = list(raw_data.keys())
+        self.data.update(raw_data)
 
     @staticmethod
     def _data_from_df(df):


mattpap:
I feel like I already fixed this issue in one of my PRs. Anyway, LGTM.

philippjfr:
Looks like there's a bunch of test failures so maybe ``add`` was actually doing something important. I'll keep looking into it.

philippjfr:
Looks like it's mostly because ``column_names`` also needs to be updated (I'll do that now), but I'm wondering how important ``column_names`` actually is, I have often inserted column directly on the ``data`` and everything worked out okay. Is there a reason why the ``column_names`` can't just be a property returning the ``data.keys()``?

bryevdv:
I think `column_names` is mostly historical junk at this point but it's worth validating that hunch.

philippjfr:
Any idea what the ``embed`` test failures are about?

bryevdv:
Some new version of a dependency in Anaconda broke things I think.

philippjfr:
Okay, let me know when that's fixed so I can rebase.

bryevdv:
should be fixed now

philippjfr:
@bryevdv Ready to merge unless you think there are any tests that should be added.

bryevdv:
The existing tests cover checking that `columns` is set as expected so I think they are sufficient.

PR #5873, opened 2017-02-14
Updating docs to show how to disable whitelist in bokeh serve.

leopd:
The only way to figure out how to disable the whitelist in bokeh server has been to read the code.  Here's a simple fix if you want it.

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -107,6 +107,12 @@ acceptable host values with the ``--host`` option:
 
     bokeh serve app_script.py --host foo.com:8081 --host bar.com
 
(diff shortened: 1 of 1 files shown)


bryevdv:
HIt @leopd thanks for the PR. I actually think we may have a way of getting rid of the need for the `--host` parameter entirely, which would be a great simplification for users, but there is still some investigation to do. So, I am going to hold off on this for now, until we can see if getting rid of `--host` is feasible or not.

bryevdv:
Hi @leopd thanks for the contribution. #5979 will be merged soon (probably tomorrow) which makes `--host` no longer needed at all, which is hopefully a great simplification for everyone. Accordingly, I am going to close this PR.

PR #5874, opened 2017-02-14
explicitly require lxml for testing

bryevdv:
Beautiful soup was using different HTML parsers across different  builds, causing tests to fail. Explicitly require `lxml`, which `bs4` regards as the "best" (and which matches our current tests)


mattpap:
> and which matches our current tests

Which are wrong btw.

PR #5875, opened 2017-02-14
Use explicit bs4 HTML parser and fix tests

mattpap:
Fixes failing travis ci's jobs.

### DIFF
--- a/bokeh/tests/test_embed.py
+++ b/bokeh/tests/test_embed.py
@@ -73,14 +73,14 @@ class TestComponents(unittest.TestCase):
 
     def test_result_attrs(self):
         script, div = embed.components(_embed_test_plot)
-        html = bs4.BeautifulSoup(script)
+        html = bs4.BeautifulSoup(script, 'html5lib')
         scripts = html.findAll(name='script')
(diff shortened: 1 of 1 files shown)


bryevdv:
I already just merged a fix of for this.

bryevdv:
I thought about doing it this way but bs4 regards `lxml` as the "best" parser so I opted to just ensure it is always installed.

mattpap:
@bryevdv, on my computer and on travis ci, bs4 considers `html5lib` the default (you don't have to even specify the dependencies), which is correct because HTML5 is the current web standard. You can also see that our tests are incorrect and that's why they were failing when the default parser in bs4 changed. Also those test could be just rewritten to not consider insignificant whitespace.

bryevdv:
It's debatable that these tests should be sensitive to non-significant whitespace. They should test the DOM structure. If you want to improve the tests, I'd support a PR that compares `div.text.strip()` to the empty string. But please note that I made other fixes to the tests that should be preserved (e.g. a vacuous `AssertTrue` that should have been `AssertEquals`

PR #5883, opened 2017-02-17
Test bokehjs examples and run them on travis ci

### LINKED ISSUE #5882: Test bokehjs/examples and run them on travis ci
They are constantly broken due to not being continuously tested.

mattpap:
fixes #5882


bokehjs/examples/anscombe/anscombe.ts
```diff
@@ -1,8 +1,8 @@
 namespace Anscombe {
   console.log(`Bokeh ${Bokeh.version}`);
-  Bokeh.set_log_level("debug");
+  Bokeh.set_log_level("info");
 
-  const anscombe_quartet = Bokeh.LinAlg.transpose([
+  const anscombe_quartet = Bokeh.Utils.transpose([
```
mattpap: This will require a migration note, just I'm still not sure if this is the final API.

bryevdv: This seems "OK" to me. Whatever we end up with for 0.12.5 should be considered final.

bokehjs/examples/anscombe/.gitignore
```diff
@@ -1 +1,2 @@
 anscombe.js
+anscombe-*.png
```
bryevdv: can't we just add a single `.gitignore` at the top with `*.js` an `*.png` ?

bryevdv: Can't we just add a single `.gitignore` one level up with `*.js` an `*.png` ? Having those definitely seems like the exception. If we need to add a `.js` we can make git do that on a case by case basis, instead of having to remember to add a custom `.gitignore` to every new example.

bokehjs/examples/anscombe/anscombe.ts
bryevdv: This seems "OK" to me. Whatever we end up with for 0.12.5 should be considered final, and my preference would be to decide that now and add the migration note now.

bokehjs/examples/anscombe/.gitignore
mattpap: I'm actually going to save files generated by `test_examples` in a dot directory, e.g. `.tests`. Then we can exclude all in one place and don't risk interfering with committed files.

bokehjs/examples/anscombe/anscombe.ts
mattpap: There are a few more things that have to be exposed through this API, that's why I would like to defer this to another PR.

bryevdv: Then there is no point in changing it from `LinAlg` in this PR. A PR should have both a decision and a change, or neither.

PR #5886, opened 2017-02-17
Support visualization for zeppelin

### LINKED ISSUE #5885: Support embed bokeh into apache zeppelin
[Apache Zeppelin](https://zeppelin.apache.org/) is a web-based notebook that enables interactive data analytics. I'd like integrate bokeh into Apache Zeppelin as Jupyter notebook. As the first step, I just want to visualize the chart in zeppelin directly.

zjffdu:
[Apache Zeppelin](https://zeppelin.apache.org/) is a web-based notebook that enables interactive data analytics. I'd like integrate bokeh into Apache Zeppelin as Jupyter notebook. As the first step, I just want to visualize the chart in zeppelin directly. 

- [x] issues: fixes #5885 
- [x] tests: not added yet.

- Screenshot in Zeppelin with this PR
image


zjffdu:
I didn't add test yet, may need some guidance on how to add test for bokeh. Could anyone help review it ? Thanks

bryevdv:
Hi @zjffdu thanks for the PR. I am interested in the idea of supporting Zeppelin, my one concern is that we aren't really set up to maintain this under test directly. But the template looks fairly simple and straightforward so maybe it's OK. 

That said if we are going to generalize from just one notebook type to N notebook types, we will need to be a bit more explicit. I'll leave some comments inline with some proposed changes.

bokeh/core/templates.py
```diff
@@ -35,6 +35,8 @@
 
 FILE = _env.get_template("file.html")
 
+ZEPPELIN_FILE = _env.get_template("zeppelin.html")
+
 NOTEBOOK_LOAD = _env.get_template("notebook_load.html")
```
bryevdv: If we are going do the "one notebook type" to "N notebook types" then we need to make things explicit. I think the new template should be `ZEPPELIN_NOTEBOOK_LOAD` and then the old `NOTEBOOK_LOAD` should be renamed to `JUPYTER_NOTEBOOK_LOAD`

This principle holds in general, and will make this PR necessarily be somewhat bigger. Basically, if we are to support two notebook types, it's not good to have one thing just called "notebook" (which is implicitly jupyter) and then a new thing called "zeppelin notebook". We need to make sure *everywhere* we explicitly distinguish "jupyter notebook" and "zeppelin notebook". So e.g. there is also `bokeh.util.notebook` that has functions like `load_notebook` in it that will need to be renamed to something like `load_jupyter`. Then, in order to be user-friendly a function `load_notebook` should be added that prints a deprecation warning, and defers to `load_jupyter`. There will need to be migration notes about all these changes in the release notes.

bokeh/io.py
```diff
@@ -37,6 +37,8 @@
 from .util.notebook import load_notebook, publish_display_data, get_comms
 from .util.string import decode_utf8
 from .util.serialization import make_id
+from .resources import INLINE
+from .core.templates import ZEPPELIN_FILE
```
bryevdv: This should be wrapped up in a function. e.g. `bokeh.util.notebook.load_zeppelin` In doing so, you can also avoid the import of `INLINE` here. The `load_zeppelin` function should be consistent with the existing `bokeh.util.notebook.load_notebook` (which should be renamed to `bokeh.util.notebook.load_jupyter`) in that passing `None` for a resources arg has a default meaning. 

I'd also carefully consider whether the default should be inline. We did that originally for jupyter notebooks and changed to can default after a heavy user request to do so. Inline resources will make all notebooks substantially larger

Also imports and import lines should be in alphabetical order throughout.

bokeh/core/_templates/zeppelin.html
```diff
@@ -0,0 +1,12 @@
+{#
+Renders for zeppelin notebook.
+#}
```
bryevdv: Need to document all template parameters in this comment block

bokeh/core/_templates/zeppelin.html
```diff
@@ -0,0 +1,12 @@
...
+
+<div>
```
bryevdv: delete blank line

bokeh/core/_templates/zeppelin.html
```diff
@@ -0,0 +1,12 @@
...
+    {{ plot_div|indent(8) }}
+    {{ plot_script|indent(8) }}
+</div>
+<div>
+    {{ bokeh_css }}
+    {{ bokeh_js }}
+</div>
```
bryevdv: This will probably "work". I think it's worth considering if the existing "autoload" style machinery can be used or lightly adapted. I don't know a lot about the Zeppelin notebook , but I am guessing you might run into the same race condition that happened in the jupyter notebook: namely, if you "run all" cells then a later cell that relies on BokehJS being loaded might exectute before the browser actually finishes loading BokehJS.

bokeh/io.py
```diff
@@ -262,7 +267,7 @@ def curstate():
 def show(obj, browser=None, new="tab", notebook_handle=False):
     ''' Immediately display a plot object.
 
-    In an IPython/Jupyter notebook, the output is displayed in an output
+    In an IPython/Jupyter/Zeppelin notebook, the output is displayed in an output
```
bryevdv: As long as we are updating this, remove `IPython`

bokeh/io.py
```diff
@@ -310,7 +315,12 @@ def _show_with_state(obj, state, browser, new, notebook_handle=False):
     shown = False
 
     if state.notebook:
-        comms_handle = _show_notebook_with_state(obj, state, notebook_handle)
```
bryevdv: This needs to be handled more thoroughly. The docstring needs to indicate that `notebook_handle=True` only works with jupyter notebooks and, I think it would be better if trying to use `notebook_handle=True` with zeppelin notebooks raises an explicit error (otherwise we risk bug reports when people expect it to work, but it is silently ignored)

bokeh/core/templates.py
bryevdv: There will also need to be supporting documentation changes. Any place in the sphinx docs where things mention "notebook" generically should be made explicitly "jupyter notebook"

bokeh/io.py
```diff
@@ -310,7 +315,12 @@ def _show_with_state(obj, state, browser, new, notebook_handle=False):
...
+        if state.notebook_name.lower() == "jupyter":
+            comms_handle = _show_notebook_with_state(obj, state, notebook_handle)
+        elif state.notebook_name.lower() == "zeppelin":
+            _show_zeppelin(obj)
+        else:
+            raise Exception("Not supported notebook type:" + state.notebook_name)
```
bryevdv: This should be handled by having a validating property setter, as described above. Then `state.notebook_type` is known to always be good (and no `.lower` is needed either)

bokeh/io.py
```diff
@@ -340,6 +350,9 @@ def _show_notebook_with_state(obj, state, notebook_handle):
             state.last_comms_handle = handle
             return handle
 
+def _show_zeppelin(obj):
```
bryevdv: suggest `show_zeppelin_with_state` to keep a consistent interface. In fact, I'd have it accept all the same args, including `notebook_handle`, then that's the place that can raise a "Zeppelin does not support notebook handles" error if `notebook_handle` is true. Also `_show_notebook_with_state` should be renamed `_show_jupyter_with_state`

bryevdv:
@zjffdu so after reviewing, I just want to be up fron: to officially support this in Bokeh at the API level will mean a fair bit more work: renaming functions that use "notebook" generically to distinguish between jupiter and zeppelin, clean deprecations for the old functions, migration notes, supporting documentation updates, and more tests. 

Another option, that might be simpler to start, is to demonstrate how Bokeh can be loaded and used in Zeppelin with *existing APIs*. A PR that does that would just need to:
* Add an example zeppelin notebook(s) somewhere in the `examples` dir
* Add a single section in the User's guide docs on using Bokeh with zeppelin, referencing the example notebooks.

zjffdu:
@bryevdv Thanks for the careful review. It does need more changes for bokeh to support zeppelin officially. I didn't change the `notebook` api as I am not sure whether these api will be used by users, although it is very unlikely. Zeppelin user can use bokeh via some workaround, but I want to make it easier for zeppelin users like what they can do in juypter.  Let me try to improve this PR further. 
This is my first PR for bokeh, I may didn't follow some rules of bokeh, feel free to let me know and I will improve it. I work in apache zeppelin and spark community, and found Bokeh is very interesting and promising. I think bokeh perfectly match with spark which also has different language bindings. I plan to integrate spark dataframe into bokeh in another PR, bokeh may become a de facto standard visualization tool upon spark.

bokeh/io.py
zjffdu: I plan to create `bokeh.util.jupyter` to replace `bokeh.util.notebook` as all the methods in `bokeh.util.notebook` are for jupyter for now. I will still keep `bokeh.util.notebook` as deprecated.  What do you think ?

Regarding `inline`, I think we can do it in another PR. Because I think one PR should just do one thing.

bryevdv: Right we could either have a `bokeh.util.notebook` that has methods for all notebooks, e.g. `load_jupyter` and `load_zeppelin`, or we could split into two modules `bokeh.util.jupyter` and ``bokeh.util.zeppelin` There are pluses and minuses to both approaches, I don't have a strong opinion on which way we go. 

Regarding the resources, it's also up to you, I was only intending to give some historical context. Given that *some* way of loading resources is always necessary, I guess my natural inclination is to pick the "correct" way from the start. But if it is easier to bootstrap the work (or if you think Zeppelin users would actually prefer inline resources) then certainly that's fine.

bryevdv:
@zjffdu I think Bokeh for the Spark community would be great. It happens to that "doing things right" means taking a bit more thought and effort than a lot of PRs just because of the way it's adding in something completely new. But you PR is appreciated, and we are here to help.

zjffdu:
Hi @bryevdv, I made some improvements on the PR. Could you please help review that.  

Changes:
* bokeh/util/notebook.py --> bokeh/util/jupyter.py 
* bokeh/application/handlers/notebook.py --> bokeh/application/handlers/jupyter.py (
* bokeh/core/_templates/notebook_div.html --> bokeh/core/_templates/jupyter_div.html
* bokeh/core/_templates/notebook_load.html --> bokeh/core/_templates/jupyter_load.html
* Refactor `notebook` to `jupyter` on test code and add `zeppelin` related test code
* Add document for zeppelin.

I only keep `bokeh.util./notebook.py` for compatibility as I thought others are not for public usage.

Besides that, I meet 2 issues which I hope to get help from you.
*  I could not build bokeh doc. Ask the question in bokeh forum as well. https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/6dupW37wafs
*  Sometimes, when I run bokeh in zeppelin for the first time I will hit the following error, but would not hit this issue after the first time. Maybe this is due to some js loading issue. But I am not familiar with bokeh frontend. Maybe you can give me some hits.



Thanks in advance for your review. And I am glad to see the strict policy on external PR of bokeh community,  this is good for an open source project's sustained growing.

bokeh/io.py
```diff
@@ -520,6 +534,57 @@ def push(session_id=None, url=None, app_path=None, document=None, state=None, io
...
+
+                handle = show(plot, notebook_handle=True)
+
+                # Update the plot title in the earlier cell
+                plot.title = "New Title"
+                push_notebook(handle=handle)
+
+    '''
+    deprecated((0, 12, 5), "push_notebook", "push_jupyter")
+    push_jupyter(document, state, handle)
```
bryevdv: Here's the exception that proves the rule. I think `push_notebook` is too well-established to deprecate. But also, since presumably only one notebook can be "activated" at a time, and this is only useful afterwards it seems ok to leave it `push_notebook`

bryevdv:
> And I am glad to see the strict policy on external PR of bokeh community, this is good for an open source project's sustained growing. 

Well, the level of "strictness" is somewhat context sensitive. As this is proposing to add a new capability that will not be straightforward to test and maintain, more work is needed up front...

> I could not build bokeh doc. Ask the question in bokeh forum as well. 

I'm not sure offhand. The best general advice I can give: 
* make sure the GH tag that is checked out matches the version of bokeh that is actually installed (or install Bokeh from source if you want to build docs for `master`)
* make sure all the "docs" dependencies lists in the `conda.recipe/meta.yaml` are installed. They are in the Anaconda defaults channel, or the bokeh channel.

> Sometimes, when I run bokeh in zeppelin for the first time I will hit the following error, but would not hit this issue after the first time. Maybe this is due to some js loading issue. But I am not familiar with bokeh frontend. Maybe you can give me some hits.

Right, this is probably the same race condition that we met working with Jupyter notebooks. Basically, if the "BokehJS load" takes too long in the browser, then subsequent cells that try to show plots (and hence rely on BokehJS being loaded) might execute "too soon" and then you will see the error you describe. Unfortunately fixing this in general is not a trivial prospect. The best I can do at the moment is to refer you to the notebook loading code for Jupyter notebooks. Perhaps that can be reused or adapted? 

I did make a quick pass over the PR, in general I think it is well improved and moving in the right direction. I left a few comments and willbe able to look more closely in the next few days. I should mention that `0.12.5` should be released next week, and I don't think it's possible that that this will be able to go in that release. `0.12.6` is more reasonable target. 

Finally, I should also ask if you think you will have the ability to act as the "maintainer" for this capability? The Bokeh project has not yet got a notion of defined maintainers, but maybe this is the right time to start thinking about that. I ask because basically if there are bugs or things to update in the future, no one else on the current core team has the right experience to make changes.

bokeh/io.py
zjffdu: Fixed

zjffdu:
@bryevdv I have a new push to make some changes
* change 0.12.5 to 0.12.6
* Keep to use push_notebook
* Use js autoload of jupyter for zeppelin. I copy the code from jupyter to zeppelin. No changes except the method signature, so I have concern whether I should merge them together. If the autoload js is stable enough, I may should merge them together.

The CI fails due js build, but I don't know how it relates with my changes. Could you help take a look at it ? Thanks

And I'd love to be the maintainer of zeppelin support and believe I have the time and ability to do that.  :)

bryevdv:
>  If the autoload js is stable enough, I may should merge them together. 

I think this is worth investigating, or if not merged together, perhaps a large common portion can be factored out. That would make things more maintainable. 

Regarding CI I think there have been some issues on Travis so they may be spurious failures. Once you have fixed the small merge conflicts (let me know if you have questions or need help) you will also need to merge/rebase on current master. An upstream dependency changed and broke the build so you will need to merge/rebase master to get the fix. If there are still issues after that, I can take a look. 

> And I'd love to be the maintainer of zeppelin support and believe I have the time and ability to do that.

Great! Once this PR is merged, I will add you to @bokeh/dev

zjffdu:
@bryevdv I find that not only the autoload js can be reused, I can also reuse jupyter div. That means in zeppelin, I can split it into 2 steps the same like it in jupyter. I may be able to merge most of the interfaces.

zjffdu:
@bryevdv Any comment ?

bryevdv:
Sorry I thought I had already but apparently not. If the existing machinery can be re-used as-is, I think that is ideal. Otherwise I think common functionality should be factored out to the fullest extent possible. I'd definitely lean towards adding as few new things as possible, and towards not renaming things if there is no need to differentiate.

zjffdu:
@bryevdv Could you help review it ? Regarding the CI failure, I don't how what does that mean ? I didn't make any change on line 189.
```
sphinx.errors.SphinxWarning: /home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/quickstart.rst:189: WARNING: Inline substitution_reference start-string without end-string.
```

sphinx/source/docs/user_guide/quickstart.rst
```diff
@@ -167,6 +167,20 @@ which show how Bokeh can be used together with Jupyter interactive widgets:
...
 :bokeh-tree:`examples/howto/notebook_comms/Numba Image Example.ipynb`
 
+Zeppelin Notebooks
+==================
+
+At this point we should mention `Apache Zeppelin`_ notebooks.
+
+`Apache Zeppelin`_ is a web-based notebook that enables interactive data analytics.
+You can make beautiful data-driven, interactive and collaborative documents with SQL, Scala and more. It is widely
+used across the `Spark` community. Bokeh integrates seamlessly with Zeppelin
```
bryevdv: This probably needs to be double back ticks (or no back ticks)

sphinx/source/docs/user_guide/quickstart.rst
```diff
@@ -167,6 +167,20 @@ which show how Bokeh can be used together with Jupyter interactive widgets:
...
+notebooks. To view the above examples in a notebook, you would only need
+change |output_file| to |output_notebook|(notebook_type='zeppelin') instead.
+
+The usage of bokeh in Zeppelin notebook is no difference from Jupyter except you need to specify notebook_type
+as `zeppelin`.
```
bryevdv: same comment about backticks

sphinx/source/docs/user_guide/quickstart.rst
```diff
@@ -167,6 +167,20 @@ which show how Bokeh can be used together with Jupyter interactive widgets:
...
+Zeppelin Notebooks
+==================
+
+At this point we should mention `Apache Zeppelin`_ notebooks.
+
+`Apache Zeppelin`_ is a web-based notebook that enables interactive data analytics.
+You can make beautiful data-driven, interactive and collaborative documents with SQL, Scala and more. It is widely
+used across the `Spark` community. Bokeh integrates seamlessly with Zeppelin
+notebooks. To view the above examples in a notebook, you would only need
+change |output_file| to |output_notebook|(notebook_type='zeppelin') instead.
```
bryevdv: This will probably render better with just the double backticks for inline code, instead of the `|` which make a link substution.

sphinx/source/docs/user_guide/setup.rst
```diff
@@ -66,6 +66,13 @@ Execute the lines of python code in the notebook, except with ``output_notebook`
     :scale: 50 %
     :align: center
 
+You can also test your Bokeh installation in an Zeppelin notebook.
+Execute the lines of python code in the notebook, except specifying ``notebook_type`` as ``zeppelin`` in method ``output_notebook``. You should see results like:
```
bryevdv: Try to adhere to 80 column limit in `.rst` files

bokeh/util/notebook.py
```diff
@@ -1,14 +1,16 @@
-''' Functions useful for loading Bokeh code and data in IPython notebooks.
+''' Functions useful for loading Bokeh code and data in Jupyter/Zeppelin notebooks.
 
 '''
 from __future__ import absolute_import
 
 _notebook_loaded = None
 
-def load_notebook(resources=None, verbose=False, hide_banner=False, load_timeout=5000):
+def load_notebook(notebook_type='jupyter', resources=None, verbose=False, hide_banner=False, load_timeout=5000):
```
bryevdv: should make this new parameter at the end instead of the beginning (better for backward compat)

zjffdu: Make sense, I also do the same thing for `output_notebook`

bryevdv:
FYI we are in the middle of a release crunch, so I've stopped the CI builds temporarily to conserve build workers. I'll restart them later.

zjffdu:
ping @bryevdv Could you help continue to review it ? Thanks

bokeh/core/state.py
```diff
@@ -39,6 +39,8 @@ class State(object):
 
     '''
 
+    valid_notebook_types = ['jupyter', 'zeppelin']
```
bryevdv: suggest module level tuple, and all caps `NOTEBOOK_TYPES = ('jupyter', 'zeppelin')`

bokeh/core/state.py
```diff
@@ -81,6 +83,23 @@ def notebook(self):
...
+        return self._notebook_type
+
+    @notebook_type.setter
+    def notebook_type(self, notebook_type):
+        ''' Notebook type, acceptable values are 'jupyter' and 'zeppelin'.
+
+        '''
+        if notebook_type is None or (notebook_type.lower() not in self.valid_notebook_types):
+            raise Exception("Notebook type '" + notebook_type + "' is not supported, the supported notebook types are "
+                            + "({})".format(', '.join(self.valid_notebook_types)))
```
bryevdv: prefer single long line, also this style is more common in our codebase:
```
"Unknown notebook type %r, supported notebook types are: %s" % (notebook_type, ', '.join(NOTEBOOK_TYPES))

File: bokeh/embed.py [hunk 18]
```diff
@@ -323,10 +323,9 @@ def _bundle_for_objs_and_resources(objs, resources):
 
     return bokeh_js, bokeh_css
 
-
 def notebook_div(model, notebook_comms_target=None, theme=FromCurdoc):
-    ''' Return HTML for a div that will display a Bokeh plot in an
-    IPython Notebook
+    ''' Return HTML for a div that will display a Bokeh plot in a
+    Jupyter/Zeppelin Notebook. notebook_comms_target is only supported in Jupyter for now.
```
bryevdv: we do try to keep docstrings to 80 chars tho

bokeh/io.py
bryevdv: I definitely think this should raise an informative `ValueError` if the notebook type is not `'jupyter'`

bryevdv: Ah, I see that is handled below...

bokeh/io.py
```diff
@@ -328,6 +333,12 @@ def _show_notebook_with_state(obj, state, notebook_handle):
         state.last_comms_handle = handle
         return handle
 
+def _show_zeppelin_with_state(obj, state, notebook_handle):
+    if notebook_handle:
+        raise Exception("Zeppelin doesn't support notebook_handle.")
```
bryevdv: suggest `ValueError`

bokeh/util/notebook.py
```diff
@@ -31,8 +34,13 @@ def load_notebook(resources=None, verbose=False, hide_banner=False, load_timeout
...
     '''
     html, js = _load_notebook_html(resources, verbose, hide_banner, load_timeout)
-    publish_display_data({'text/html': html})
-    publish_display_data({'application/javascript': js})
+    if notebook_type=='jupyter':
+        publish_display_data({'text/html': html})
+        publish_display_data({'application/javascript': js})
+    else:
+        print('%html ' + html)
+        print('%html ' + '<script type="text/javascript">' + js + "</script>")
```
bryevdv: probably worth factoring into a small function.

bryevdv:
@zjffdu I left a few more small comments. In general I think it looks pretty good, will just need some manual testing. One new thing that may also need to be addressed: in the last release `show` was modified to accept a Bokeh server application, and can now run and show these inline in a jupyter notebook. I doubt it would work as-is in zeppelin, so there will probably need to be another check in show if an app is passed and notebook type is wrong to raise and exception.

zjffdu:
@bryevdv I have addressed the comments, please help review.

bryevdv:
I think this is pretty close, @mattpap @canavandl can you both also take a quick look at this PR? 

In particular if there are additional unit tests that we could do, it's worth considering, since I doubt this will get much or any manual testing on a regular basis.

bokeh/util/notebook.py
```diff
@@ -31,13 +34,21 @@ def load_notebook(resources=None, verbose=False, hide_banner=False, load_timeout
...
+
 
 FINALIZE_JS = """
 document.getElementById("%s").textContent = "BokehJS is loading...";
 """
 
+def _publish_zeppelin_data(html, js):
+    print('%html ' + html)
+    print('%html ' + '<script type="text/javascript">' + js + "</script>")
+
```
canavandl: I don't know anything about Zeppelin, but is the standard way to output things is to print to stdout?

canavandl:
One question for clarification, but otherwise LGTM

bokeh/util/notebook.py
zjffdu: Thanks @canavandl for review, zeppelin's python interpreter (JVM) would capture python process's stdout and send it to frontend. So yes, this is the standard way for zeppelin to output things to frontend.

canavandl:
Hi @zjffdu 

I'm working to get this PR merged today, but it's current blocked by #6160. The resolution of that bug may require a small change in this PR, i'll let you know though.

nicain:
Hello all; posting here because the example at: https://github.com/ecerami/pydata-essentials/blob/master/bokeh/bokeh_notebook_embed.ipynb

works with bokeh 0.12.5; 0.12.6rc4 fails with the error:

AttributeError                            Traceback (most recent call last)
<ipython-input-1-be6a46af6794> in <module>()
     45 
     46 doc = app.create_document()
---> 47 show(app, notebook_url="localhost:8888")

/local1/anaconda2/lib/python2.7/site-packages/bokeh/io.pyc in show(obj, browser, new, notebook_handle, notebook_url)
    282     # in Tornado) just in order to show a non-server object
    283     if getattr(obj, '_is_a_bokeh_application_class', False):
--> 284         return _show_notebook_app_with_state(obj, _state, "/", notebook_url)
    285 
    286     if obj not in _state.document.roots:

/local1/anaconda2/lib/python2.7/site-packages/bokeh/io.pyc in _show_notebook_app_with_state(app, state, app_path, notebook_url)
    289 
    290 def _show_notebook_app_with_state(app, state, app_path, notebook_url):
--> 291     if state.notebook_type == 'zeppelin':
    292         raise ValueError("Zeppelin doesn't support show bokeh app.")
    293 

/local1/anaconda2/lib/python2.7/site-packages/bokeh/core/state.pyc in notebook_type(self)
     89 
     90         '''
---> 91         return self._notebook_type
     92 
     93     @notebook_type.setter

AttributeError: 'State' object has no attribute '_notebook_type'

bryevdv:
@nicain zeppelin support does not yet extend to bokeh applications, only to "standalone" docs/plots. Adding that capability will probably require a strong commitment from someone to be the maintainer of that functionality

bryevdv:
I just noticed you state that it worked with `0.12.5` I am honestly completely surprised if that is the case. Unfortunately we will not be able to hold up the `0.12.6` release over this.

bryevdv:
Actually, I am confused @nicain you are posting this report in the issue for Zeppelin notebook support. Is this regarding the zeppelin (pyspark) notebook? or the standard jupiter notebook?

bryevdv:
In any case the example works fine for me with `rc4` and `rc5` with standard jupiter notebook. There were no changes to the `State` object in this release so I suspect you might have some kind of install corruption

zjffdu:
I think @nicain means it works in jupyter with 0.12.5 and doesn't work in 0.12.6, I can reproduce this. Let me try to fix this.

bryevdv:
@nicain if this worked as-is with `0.12.5` it was an accident because you are missing a required call to `output_notebook` at the top. With that added it works fine:

screen shot 2017-06-12 at 21 59 05

nicain:
As usual, you nailed it; I was using Jupyter, and the example didn't include output_notebook().  Works just fine with 0.12.5 and the fresh 0.12.6 installed from conda!  Thanks for all your hard work :)

PR #5889, opened 2017-02-17
Use `master@{upstream}` instead of `origin/master` as the default diff-ref

### LINKED ISSUE #5888: tests/examples/examples_report_plugin.py assumes clone has a remote named origin
#### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant pacakges)
* bokeh master sha 3e802d4f6
* conda environment on anaconda.org [tsnyder/bokeh-dev-py34.1-170217-0714](https://anaconda.org/tsnyder/bokeh-dev-py34.1-170217-0714)
* I'm running on a Centos7 VM
* git version 1.8.3.1

The tests are failing to run for me because I don't have a remote named 'origin' in my clone.  When working in a Fork&PR workflow, I generally dislike using the default remote 'origin' and prefer to have two remotes named:
* upstream - points to the upstream repo, in our case bokeh/bokeh
* mygh - points to my github fork.  In my case, tsnyder/bokeh

I do this because I have a hard time remembering whether I cloned from bokeh/bokeh or timsnyder/bokeh.  If I don't have a remote named origin, but instead use a descriptive name, I don't have to try and remember.  I just push to mygh and pull from upstream.

No code changes required.  Steps to reproduce:
```
git config --get-regexp remote.origin && git remote rename origin not-named-origin
py.test -m 'not (js or examples or integration or quality)'
```

##### Observed behavior:
```
.-(~/bokeh)---------------------------------------------------------------------------------------(tsnyder@centos7vm)-
`-- (git)-[master]-> py.test -m 'not (js or examples or integration or quality)'
fatal: Not a valid object name origin/master
Failed to get version for origin/master
```

##### Expected behavior:
I expect that it would work without a remote named 'origin'.  
n/a

n/a

#### Bug Origin
It looks like @mattpap added the `--diff-ref` option to `examples_report_plugin.py` in #5713.  Just mentioning you so that you have a chance to comment on the intended default behavior of `--diff-ref`.

#### Suggested Fix
It may be better to use the special reference named '@{upstream}' as documented by [gitrevisions(7)](https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html)

Or, if the intent is to always check against bokeh/bokeh master, regardless of the current branch, we should do a reverse lookup for the correct remote name to use.

I'm going to try out my fix and submit a PR for this.

mattpap:
Using `origin` is presumptuous and works only for core dev workflow, but not necessarily in forks, which can have the origin (upstream) configured to something different than the literal `origin`.

fixes #5888

### DIFF
--- a/tests/examples/examples_report_plugin.py
+++ b/tests/examples/examples_report_plugin.py
@@ -29,7 +29,7 @@ def pytest_addoption(parser):
         "--report-path", action='store', dest='report_path', metavar='path', default='report.html', help='create examples html report file at given path.'
     )
     parser.addoption(
-        "--diff-ref", type=resolve_ref, default="origin/master", help="compare generated images against this ref"
+        "--diff-ref", type=resolve_ref, default="master@{upstream}", help="compare generated images against this ref"
     )
     parser.addoption(
         "--incremental", action="store_true", default=False, help="write report after each example"


bryevdv:
checked CI report, looks good

PR #5896, opened 2017-02-20
Typo corrections and clarifications for user_guide/server (#5895)

### LINKED ISSUE #5895: Give feedback on bokeh server docs
This is just a ticket so that I can open a PR with some revisions I'd like to suggest to the docs after reading through the User and Dev guide pages for Bokeh server.

timsnyder:
- [x] issues: fixes #5895 

I have some feedback on the server docs.  I've made it through the UserGuide page and wanted to post this to get any dialog started.  I hope to follow up with anything I find in the DevGuide page in the next day or two.


timsnyder:
Just to channel my inner test troll... I'd really like to see something about Best Practices for testing bokeh server apps in user_guide/server.  Has this topic come up in the mailing list or anywhere that we could add to the doc?

sphinx/source/docs/dev_guide/server.rst
```diff
@@ -258,8 +259,8 @@ two sequences of frames, one sequence in each direction ("full duplex").
 On top of websocket frames, we implement our own ``Message`` concept. A Bokeh
 ``Message`` spans multiple websocket frames. It always contains a header frame,
 metadata frame, and content frame. These three frames each contain a JSON
-string. The code permits these three frames to be followed by binary data
-frames, but currently in Bokeh binary data frames are not used.
+string. The code permits these three frames to be followed by optional binary data
+frames. Currently in Bokeh, binary data frames are only used to transport image data.
```
timsnyder: @bryevdv, I'm not sure whether this modification is accurate but since you mentioned that image data was recently optimized to use binary data transfer, I'm thinking this might be closer to true.

philippjfr: @timsnyder Thanks for updating this, I should have grepped for mentions of the binary protocol when I was implementing the protocol. To clarify the binary array protocol is now used to send all numeric array types except for int64 arrays because Javascript does not support 64-bit integers natively.

bryevdv: It's a small list, we should be explicit: all NumPy arrays with `float32`, `float64`, and `int32` dtypes, this is correct?

bryevdv:
> I'd really like to see something about Best Practices for testing bokeh server apps in user_guide/server

That sounds great, I am not sure offhand what would go in that section but I would be happy to have others chime in with ideas.

sphinx/source/docs/dev_guide/server.rst
bryevdv: Also, to be technically pedantic, the original statement is still correct. @philippjfrallowed for a different encoding of numpy arrays, but array data is still sent as part of the message "content" block, as an encoded string. It just happens to be much faster to encode/decode the string than raw JSON encoding of the array was. 

Separate from this, the Bokeh protocol also allows for multi-frame messages. In principle this could allow for example, for sending numpy arrays directly from their memory buffers to the websocket with no additional copies. This might offer even more performance boost. But it is not used yet. There is nothing in Bokeh that makes use of binary frames yet.

timsnyder: Thanks @bryevdv and @philippjfr.   When I was tweaking the docs, I wondered if the recent work was base64 encoding the data into the content block or actually using 'binary' frames.  Now that I have a pointer to @philippjfr's work in #5544, I'll read through his code changes and take another pass at this section of the dev_guide.

timsnyder:
>> I'd really like to see something about Best Practices for testing bokeh server apps in user_guide/server

>That sounds great, I am not sure offhand what would go in that section but I would be happy to have others chime in with ideas.

@birdsarah,  I'm guessing you might be interested in this topic and able to comment regarding best-practice for testing Bokeh Server Apps.  Just FYI.

### To expand on what I was thinking -- the narrative probably goes something like this:

#### Section on Testability for Simple Bokeh Apps
Give a shout-out to the MVC pattern.  Explain that Bokeh Server apps typically have:
* Model in Python code they are writing
* View in BokehJS-provided and tested view code
** Advanced users may want to write custom BokehJS plugin. Point to that doc and make sure that we address testability there.
* Controller is Bokeh widgets and the server machinery to get the stimulus back to the model code.
** Again, point to how someone can create a custom widget

Encourage people to encapsulate their Model logic in a way that is conducive to Python unit testing and pointing them into:
* Bokeh Server demos that have good unit test coverage of their model code.
** I'm not sure whether this is true right now but I'd be willing to submit a PR to make it that way.
* A [cookiecutter](https://cookiecutter.readthedocs.io/en/latest/readme.html) template for bootstrapping a new Bokeh Server App
** Wouldn't have to be cookiecutter, that's only a project I know about and think is neat.  There may be something else/better.

#### Sections for more advanced Bokeh Apps that try to do custom things in the view or controller.
This is where I would put the sub-bullet information regarding user-written View & Controller functionality, pointing into other areas of the Bokeh docs regarding custom JS and widgets

sphinx/source/docs/dev_guide/server.rst
```diff
@@ -313,8 +314,8 @@ Some Current Protocol Caveats
    send the whole giant dictionary whenever any entry in it
    changes.
 
-3. At the moment, we do not optimize binary data by sending it
-   over binary websocket frames.
+3. At the moment, we only optimize image data by sending it
+   over binary websocket frames.  All other binary data is unoptimized.
```
timsnyder: After reading through #5544, I'm going to revert most of my change from earlier in the WebSocket section and update this to explain base64 encoding of NumPy arrays of type `float32`, `float64` and `int32`.  I'll mention the limitation regarding `int64` types in JS.

I aim to push something later today.  Sorry it's taken me so long to get back to this.

sphinx/source/docs/user_guide/server.rst
```diff
@@ -762,38 +760,45 @@ thread, e.g. when using ``bokeh.client`` to create sessions from a web
...
+with it. To share it with other people who are able to install the required
+python stack, we can share the application with them, and let them run it locally
+themselves in the same manner. However, we might also want to deploy the application
+in a way that other people can access it as a service:
+* without having to install all of the prerequisites
+* without needing to have the source code
+* like any other webpage
+
+This section describes some of the considerations that arise when deploying
+Bokeh server applications as a service for others to use.
```
timsnyder: As a general rule, does the Bokeh project mention Continuum's Enterprise offerings where applicable?  I'm asking because this section could possibly use a pointer to [the magic button](https://youtu.be/ZJ8dTg_eKzc) from Anaconda Enterprise v5.0 later this year, right @koverholt ?  However, if we don't point to awesome Enterprise things from the Bokeh docs, I'll remember that as a rule.

bryevdv: We try to keep a pretty light touch. AFAIK the only current content is an email directing anyone interested to `sales@cio` That said, there is a clear interest in hosted apps, so I am happy to have a *brief* pointer to the new things coming out in AEv5. I'd propose doing that in a separate PR, however, once we can get input from @koverholt and others. It's possible they may want to make a specific landing page for folks directed there fro the Bokeh docs.

timsnyder: Definitely a separate PR. Thanks for the feedback.

timsnyder:
This PR is ready from my end.  Let me know if there's anything you guys want changed.

PR #5897, opened 2017-02-20
Global urls for JS and CSS resources

d2rk:
Add urls for JS and CSS resources


bryevdv:
we don't really have a status for "closed by proposer" so marking `rejected` to keep our label compliance checks happy

PR #5898, opened 2017-02-20
Allow to get autoload.js without an element

d2rk:
Sometimes it is necessary to get `autoload.js` without providing element id. Thus you can get necessary scripts and load Bokeh sessions later manually.

### DIFF
--- a/bokeh/server/views/autoload_js_handler.py
+++ b/bokeh/server/views/autoload_js_handler.py
@@ -28,22 +28,22 @@ class AutoloadJsHandler(SessionHandler):
     def get(self, *args, **kwargs):
         session = yield self.get_session()
 
-        element_id = self.get_argument("bokeh-autoload-element", default=None)
-        if not element_id:
-            self.send_error(status_code=400, reason='No bokeh-autoload-element query parameter')
-            return
-
         resources = self.application.resources(self.request)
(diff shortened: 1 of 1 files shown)


bryevdv:
@d2rk can you elaborate more on the use case this is intended to address?

d2rk:
In my case the frontend server is using BokehJS directly in order to inject plots by the given Bokeh session. However, it's necessary to load BokehJS libraries along with custom models. I wasn't able to find any way to do this except to modify `autoload.js`. While I can get BokehJS from CDN or static files directly there is no way to get custom models.

bryevdv:
@d2rk I am interested in exploring your use case. This may or may not be the best approach. Just  a heads up there's not going to be enough people on deck to make anything happen before `0.12.5` (hopefully next week) but let's chat about it in more detail after the release.

d2rk:
@bryevdv any update when this could be merged?

bryevdv:
@d2rk thinking about this more, is there a reason you can't just call `AUTOLOAD_JS` directly, yourself, with `"server"` mode configured resources? That seems like a preferable solution to making this change to the core library.

bryevdv:
Closing for staleness, and also as stated, and because there is already an existing solution,  AUTOLOAD_JS template can be used directly with a null elementid.

PR #5913, opened 2017-02-23
Fix #5793: remove logging.basicConfig() calls from library

### LINKED ISSUE #5793: bokeh calls logging.basicConfig()
When run as a library, bokeh calls logging.basicConfig() in `bokeh.document`. This changes configuration of the root logger (including its default level), even though the embedding application may not expect or want this. Instead, bokeh should just configure its own logger.

pitrou:
issues: closes #5793

### DIFF
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -752,10 +752,6 @@ class Document(object):
             None
 
         '''
-        # logging.basicConfig is a no-op if there's already
-        # some logging configured. We want to make sure warnings
-        # go somewhere so configure here if nobody has.
-        logging.basicConfig(level=logging.INFO)
         root_sets = []
         for r in self.roots:
             refs = r.references()
--- a/bokeh/util/logconfig.py
+++ b/bokeh/util/logconfig.py
@@ -18,9 +18,36 @@ The default logging level is ``none``.
 from __future__ import absolute_import
 
 import logging
+import sys
 
 from ..settings import settings
 
 level = settings.py_log_level()
+bokeh_logger = logging.getLogger('bokeh')
+root_logger = logging.getLogger()
+
 if level is not None:
-    logging.basicConfig(level=level)
+    bokeh_logger.setLevel(level)
+
+if not (root_logger.handlers or bokeh_logger.handlers):
+    # No handlers configured => at least add a printer to sys.stderr for
+    # Bokeh warnings to be displayed
+    default_handler = logging.StreamHandler(sys.stderr)
+    default_handler.setFormatter(logging.Formatter(logging.BASIC_FORMAT))
+    bokeh_logger.addHandler(default_handler)
+    # Avoid printing out twice if the root logger is later configured
+    # by user.
+    bokeh_logger.propagate = False
+else:
+    default_handler = None
+
+
+def basicConfig(*args, **kwargs):
+    """
+    A logging.basicConfig() wrapper that also undoes the default
+    Bokeh-specific configuration.
+    """
+    if default_handler is not None:
+        bokeh_logger.removeHandler(default_handler)
+        bokeh_logger.propagate = True
+    return logging.basicConfig(*args, **kwargs)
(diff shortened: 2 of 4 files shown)


pitrou:
@bryevdv, this will hopefully do the trick.

bokeh/document.py
```diff
@@ -752,10 +752,6 @@ def validate(self):
             None
 
         '''
-        # logging.basicConfig is a no-op if there's already
-        # some logging configured. We want to make sure warnings
-        # go somewhere so configure here if nobody has.
-        logging.basicConfig(level=logging.INFO)
```
bryevdv: My impression of the issue was that you intended for a different custom logger to be configured in `Document`, not that no logger at all be configured. These are validation errors, they explain usage problems that might otherwise result in e.g. a blank graph with no other explanation. We put this here because there are warnings and errors that must be relayed, unconditionally, or else our support burden increases beyond maintainable levels.

pitrou: Calling `basicConfig` in random library functions is completely unpredictable (and broken from the user's POV, as whether bokeh messes with the logging configuration then depends on which exact library functions are called). logging should either be configured at library startup, or shouldn't be configured at all.

bryevdv: `validate` is not some random function. It is called at serialization time. There is basically no usage of Bokeh that will not go through this function. I'm open to suggestions. I don't care about calling or not calling `basicConfig` especially, all that matters is that these usage validation message are not silent by default under any circumstance.

pitrou: > all that matters is that these usage validation message are not silent by default under any circumstance

That is ensured by the fact that logging is configured at startup. Perhaps you can try to PR to check that it does work? I don't know how to exercise this.

bokeh/util/logconfig.py
```diff
@@ -18,9 +18,36 @@
...
 level = settings.py_log_level()
+bokeh_logger = logging.getLogger('bokeh')
+root_logger = logging.getLogger()
+
 if level is not None:
-    logging.basicConfig(level=level)
+    bokeh_logger.setLevel(level)
+
+if not (root_logger.handlers or bokeh_logger.handlers):
+    # No handlers configured => at least add a printer to sys.stderr for
```
pitrou: This is what I'm talking about when I say "logging is configured at startup". The important thing is that the Bokeh logger at least has one handler configured (either directly or through propagation) so that messages don't fall in a black hole.

pitrou: Note that `basicConfig` *also* configures a default handler (as part of other things), but it does so for the entire logging tree starting at the root logger, which is what we want to avoid in a library.

(honestly, I think this part of the logging API is badly designed and easily leads to misuse of the API, since it makes doing the right thing much harder than doing the wrong easy thing)

bokeh/document.py
bryevdv: Right you've added them to `serve.py` so they will be configured (presumably) when people execute `bokeh serve app.py`. These warnings need to show up *always*. E.g. in the jupyter notebook, in random scripts people create. None of that kind if library usage goes through `bokeh serve` (or any `bokeh` subcommand)

pitrou: See my other messages below.

bokeh/util/logconfig.py
bryevdv: Sorry, I'm still abit lost about parts of the situation. I guess a more direct question is: What can be put in `bokeh/__init__.py` unconditionally to make sure these messages never fall in a black hole  (even for "library" usage of Bokeh) without unduly affecting things.

pitrou: Unless I'm not mistaken, `logconfig` is imported unconditionally in `bokeh/__init__.py`.

bryevdv: Ah, I was not aware. Thanks, I will test this PR locally to confirm validation messages are emitted. Can you merge/rebase on master so the fix for the upstream CI issues can be brought in.

pitrou:
Okay, I did the merge. Hopefully the CI tests will pass now.

PR #5915, opened 2017-02-23
Including the color option in the embed simple example

pmreyes2:
The color information was passed to the template but was not used at all. I did very minor changes to:

1. Send the color name instead of the color numerical code in the simple.py app.
2. Use conditional `if` in the `embed.html` template in order to put the `selected` parameter in the right color.

With this changes, the selected color in the form is used for the default value.

### DIFF
--- a/examples/embed/simple/simple.py
+++ b/examples/embed/simple/simple.py
@@ -44,14 +44,14 @@ def polynomial():
     args = flask.request.args
 
     # Get all the form arguments in the url with defaults
-    color = colors[getitem(args, 'color', 'Black')]
+    color = getitem(args, 'color', 'Black')
     _from = int(getitem(args, '_from', 0))
     to = int(getitem(args, 'to', 10))
 
     # Create a polynomial line graph with those arguments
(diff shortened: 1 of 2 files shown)


bryevdv:
Hrm, I feel like I remember the color selection working, but my memory might just be failing! In any case thanks for the the PR @pmreyes2 

Can you do me one favor? If you can attach a couple of screenshots documenting the operation with the changes, that will save me from having to fork/fetch/build/test this PR directly, which would actually lift a considerable burden for me during this busy week.

pmreyes2:
The color selections works, but once you select a color, the page refreshes, the line plots correctly with the selected color, but the form always restarts at the fixed "Black" color. The screen shot here shows ,that with the URL request : http://localhost:5000/?color=Green&_from=0&to=10
the form shows also "Green" as the default selection in the form. Before it always showed Black as default, because it was hard coded in the template.
bokeh_embed

bryevdv:
Ah, I see, great that's definitely an improvement. I will merge when tests are green. Thanks for the contribution!

PR #5920, opened 2017-02-24
Remove ggplot.py examples and references.

bryevdv:
Bokeh has not been able to be used with recent versions of ggplot.py
already for some time. The difficulty of installing older versions for
python 3.6 means the docs are effectively not buildable by python 3.6.

Given the questionable future for all MPL compat for Bokeh, this PR
takes the simple route for making the docs buildable across all Python
versions, and removes the handful of ggplot.py examples, which were not
very compelling to begin with.

ping @bokeh/dev I don't expect this to be especially controversial but any feedback is welcome.


StevenCHowell:
I agree with trimming unhelpful examples to improve the documentation.  The beauty of version control is that if eventually a contributor wants to go back and reincorporate the ggplot examples, the [changes](https://github.com/bokeh/bokeh/pull/5920/files) are documented.  

Also, if more people wanted MPL and ggplot functionality, that would be demonstrated through them raising issues and contributing to fix the deficiencies or find workarounds.

bryevdv:
Also, to be clear, at least with this PR, the removed examples still continue to function (with an old version of ggplot.py), they just won't be in the docs. 

OK, I am going to go ahead and merge this.

StevenCHowell:
In connection with this merge, it may be good to add a comment in those examples.  Do they already  have such a comment?  That would help anyone interested in playing with them.

bryevdv:
Not sure I understand, this merge removed the examples. The version pin was expressed in `meta.yaml` examples debs section, but the fact that no one has ever asked leads me to think no one ever tried them.

StevenCHowell:
nvm, I misread.  I thought you were saying you were removing the documentation but leaving the code for the examples.

PR #5924, opened 2017-02-27
Make sure properties are updated on (deferred) init

### LINKED ISSUE #4875: graph 'running off'
Sometimes, when an on_click/on_change method changes some part of the plot, the entire plot will move towards the bottom right corner of the space it was previously occupying, leaving an empty square in the middle of my document. I've run into this when changing the title, adding labels, and adding lines. Here's some example code that's know to cause the issue (this is run on the server), it tries adding a legend, and this causes the issue. 

```
from random import randint
from bokeh.plotting import Figure,curdoc
from bokeh.models.widgets import Select,Button
from bokeh.models import HBox,VBox,WidgetBox, VBox,HoverTool, WidgetBox, Label, WheelZoomTool, SaveTool, PanTool
from bokeh.io import curdoc
def make_plot():
    plot = Figure(x_axis_type="datetime", plot_width=700,tools=[HoverTool(tooltips=[("vol","$y"),("Date","@formDates")]),PanTool(),WheelZoomTool(),SaveTool()])
    plot.background_fill_color='#EAEBF0'
    plot.title.text = 'a title'
    plot.xaxis.axis_label ='x axis'
    plot.yaxis.axis_label = 'y axis'
    return plot

def make_widgets(plot):
    box = Select(title="Select",options=['a','b','c','d','e'])
    box5 = Select(title="Other Select", options=[])
    butt = Button(label = 'add some line')
    x_values=None
    y_values=None
    def box_change(attrname,old,new):
        box5.options.append(new)
        box5.value = box5.options[0]
    def box5_change(attrname,old,new):
        plot.title.text = "%s - %s" % (box.value,box5.value)
    def butt_click():
        nonlocal x_values
        nonlocal y_values
        y_values=[]
        x_values=[]
        for i in range(30):
            x_values.append(i)
            y_values.append(randint(0,999))
        line = plot.line(x_values,y_values,legend=str(randint(0,50)))
        line.data_source.data['x']=x_values
        line.data_source.data['y']=y_values
    box.on_change('value',box_change)
    box5.on_change('value',box5_change)
    butt.on_click(butt_click)
    return [box,box5,butt]

plot = make_plot()
widgets = make_widgets(plot)
curdoc().add_root(VBox(HBox(WidgetBox(*widgets),VBox(plot,height=1200)),width=1200))
```

bryevdv:
- [x] issues: fixes #4875

Previously, a call to `prop.update` for model properties happened in `attach_document`. But attach_document was not being called for new models added in patches to existing documents, causing problems. This PR adds a (deferred) initializer to HasProps that ensures proper property setup for all models is completed in a consistent way, always.

I don't have a good test to add at the moment. I am going to add a separate issue to improve/refactor BokehJS properties to split up the creation/initialization into more usable and testable chunks.

### DIFF
--- a/bokehjs/test/document.coffee
+++ b/bokehjs/test/document.coffee
@@ -760,6 +760,30 @@ it "can destructively move", ->
     expect(root1.foo).to.equal 57
     expect(root1.child.foo).to.be.equal 44
 
+  it "sets proper document on models added during patching", ->
+    d = new Document()
+    expect(d.roots().length).to.equal 0
+    expect(Object.keys(d._all_models).length).to.equal 0
+
+    root1 = new SomeModel({ foo: 42 })
+    child1 = new SomeModel({ foo: 44 })
+    d.add_root(root1)
+    expect(d.roots().length).to.equal 1
+
+    # can't create copy of doc here like other test. Testing explicitly that
+    # doc attach happens when *not* creating a new document (i.e only patching)
+    # Testing only for/against null .document is not the strongest test but it
+    # should suffice.
+
+    expect(root1.document.roots().length).equal 1
+    expect(root1.child).to.equal null
+
+    event1 = new ModelChangedEvent(d, root1, 'child', root1.child, child1)
+    patch1 = d.create_json_patch_string([event1])
+    d.apply_json_patch_string(patch1)
+
+    expect(root1.document.roots().length).equal 1
+    expect(root1.child.document.roots().length).equal 1
 
   it "sets proper document on models added during construction", ->
     d = new Document()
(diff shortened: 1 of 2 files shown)


bokehjs/test/document.coffee
```diff
@@ -760,6 +760,30 @@ it "can destructively move", ->
...
+
+    expect(root1.document.roots().length).equal 1
+    expect(root1.child).to.equal null
+
+    event1 = new ModelChangedEvent(d, root1, 'child', root1.child, child1)
+    patch1 = d.create_json_patch_string([event1])
+    d.apply_json_patch_string(patch1)
+
+    expect(root1.document.roots().length).equal 1
+    expect(root1.child.document.roots().length).equal 1
```
bryevdv: This test was left over from an earlier solution, but it seems like a reasonable test to have in any case.

PR #5929, opened 2017-02-28
TST: Don't assume current dir is source directory.

### LINKED ISSUE #5928: test_api_crawler assumes it is run in the source directory
* bokeh-0.12.4
* Python 2.7.13
* Fedora 25

#### Description of expected behaviour and the observed behaviour
`test_api_crawler`, when run on the installed package, fails because it assumes it's being run in the source directory. But when testing the installed package, it's quite likely that you aren't in the source directory.

$ pip install --user bokeh && py.test --pyargs bokeh -k test_get_crawl_dict

```
___________________________________________________________ test_get_crawl_dict ____________________________________________________________

test_crawler = <bokeh.util.api_crawler.APICrawler object at 0x7fa9edb97610>

    def test_get_crawl_dict(test_crawler):
        crawl_dict = test_crawler.get_crawl_dict()
>       assert crawl_dict
E       assert {}

.local/lib/python2.7/site-packages/bokeh/util/tests/test_api_crawler.py:37: AssertionError
```

QuLogic:
- [x] issues: fixes #5928
- [ ] tests added / passed

Adding a test for this would probably require a whole extra CI build, unless you feel like changing one to run on the installed copy.

### DIFF
--- a/bokeh/util/tests/test_api_crawler.py
+++ b/bokeh/util/tests/test_api_crawler.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import ast
+import os
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


bryevdv:
LGTM thanks @QuLogic

PR #5931, opened 2017-02-28
Align bokehjs' imports with custom models

### LINKED ISSUE #5930: Align bokehjs' imports with custom models

mattpap:
This allows to import `core/util/array` instead of `../../../core/util/array`. Also typescript is now responsible for reporting bad imports, so pay attention to those `TSxxxx` errors.

fixes #5930


bryevdv:
Do the old paths also still work? Or will this break existing custom extensions?

mattpap:
Nothing changes for custom extensions. As the title suggests, this PR aligns bokehjs' imports to custom extensions. Custom models can't use relative imports for importing bokehjs' modules, because you don't know where you would be in the module tree. You can still use relative imports in bokehjs (I changed only imports from `core`, because those required the most `../`).

PR #5932, opened 2017-02-28
Fix embed themed models

### LINKED ISSUE #4722: curdoc().theme = Theme(json=yaml.load()) is not applied to charts when used in Jupyter notebook
Running the [color slider example](https://github.com/bokeh/bokeh/blob/master/examples/plotting/file/color_sliders.py#L115-L126) in a Jupyter notebook it seems that the defaults defined in the theme are not applied

### LINKED ISSUE #4952: Theme doesn't apply when using components
Can't get themes to apply when using `bokeh.embed.components`.

Have tried manually using `theme.apply_to_model(model)`, but it doesn't seem to work.

jsignell:
Changes to using curdoc rather than new doc when embedding

- [x] issues: fixes #4952 and fixes #4722
- [ ] removed a test that checked for whether `_ModelInEmptyDocument` had been called
- [ ] added example of themed embedding components

### DIFF
--- a/examples/embed/embed_themed.py
+++ b/examples/embed/embed_themed.py
@@ -1,0 +1,60 @@
+import io
+
+from jinja2 import Template
+
+from bokeh.embed import components
+from bokeh.resources import INLINE
+from bokeh.util.browser import view
+from bokeh.io import curdoc, show
+from bokeh.themes import Theme
+from bokeh.plotting import figure
+
+x1 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+y1 = [0, 8, 2, 4, 6, 9, 5, 6, 25, 28, 4]
+
+p1 = figure(title='PINK')
+p1.scatter(x1, y1)
+
+theme = Theme(json={'attrs': {'Title': {'text_color': 'pink'}}})
+curdoc().theme = theme
+
+script, div = components(p1)
+
+template = Template('''<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <title>Bokeh Scatter Plots</title>
+        {{ js_resources }}
+        {{ css_resources }}
+        {{ script }}
+        <style>
+            .embed-wrapper {
+                width: 50%;
+                height: 400px;
+                margin: auto;
+            }
+        </style>
+    </head>
+    <body>
+        <div class="embed-wrapper">
+        {{ div }}
+        </div>
+    </body>
+</html>
+''')
+
+js_resources = INLINE.render_js()
+css_resources = INLINE.render_css()
+
+filename = 'embed_simple.html'
+
+html = template.render(js_resources=js_resources,
+                       css_resources=css_resources,
+                       script=script,
+                       div=div)
+
+with io.open(filename, mode='w', encoding='utf-8') as f:
+    f.write(html)
+
+view(filename)
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -57,7 +57,7 @@ def _find_existing_docs(models):
     existing_docs.discard(None)
 
     if len(existing_docs) == 0:
-        # no existing docs, make a new one
+        # no existing docs, use the current doc
         doc = curdoc()
     elif len(existing_docs) == 1:
         # all existing docs are the same, use that one
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -25,7 +25,7 @@ from .core.templates import (
 )
 from .core.json_encoder import serialize_json
 from .document import Document, DEFAULT_TITLE
-from .model import Model, _ModelInDocument, _ModelInEmptyDocument
+from .model import Model, _ModelInDocument
 from .resources import BaseResources, _SessionCoordinates, EMPTY
 from .util.string import encode_utf8
 from .util.serialization import make_id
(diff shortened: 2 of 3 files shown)


bryevdv:
AFAICT the only remaining place `_ModelInEmptyDocument` is used is in tests for `_ModelInEmptyDocument`

https://github.com/bokeh/bokeh/blob/3ab765c07c9f8f7bf459f0871a9747a8919b1b2d/bokeh/tests/test_model.py#L9-L37

I'd be +1 on removing it altogether in this PR.

jsignell:
@bryevdv I can also probably get rid of `_ModelInDocument` in this PR if you don't think it is out of scope

examples/embed/embed_themed.py
```diff
@@ -0,0 +1,60 @@
...
+from bokeh.themes import Theme
+from bokeh.plotting import figure
+
+x1 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+y1 = [0, 8, 2, 4, 6, 9, 5, 6, 25, 28, 4]
+
+p1 = figure(title='PINK')
+p1.scatter(x1, y1)
+
+theme = Theme(json={'attrs': {'Title': {'text_color': 'pink'}}})
```
bryevdv: Maybe also add a dark (grey?) background, and make the CSS page background match? Then the lighter title will stand out more and also it gives an example of matching a plot to its background in the page.

bryevdv:
> I can also probably get rid of _ModelInDocument in this PR if you don't think it is out of scope

My experience has been that if things like this aren't done immediately as the opportunity arises they often never get done, or languish. So I'm generally +1 on reasonable, incremental code gardening in "regular" PRs as opposed to giant "cleanup" PRs that try to do lots of it at once.

jsignell:
After thinking about it some more it seems like maybe having a context manager for `_ModelInDocument` is actually not such a bad idea. It doesn't just get used in embed.

examples/embed/embed_themed.py
jsignell: do you mean the background of the plot should match background-color for css class `embed-wrapper`?

jsignell:
The example looks like this now: 
screen shot 2017-02-28 at 11 36 38 am

Pretty ugly, but proves the point I think.

bryevdv:
> Pretty ugly, but proves the point I think. 

Maybe more like the crossdfilter demo?



People ask for how to have a "dark" theme with some regularity so have a basic but reasonable one would be nice to demonstrate.

bokeh/model.py
```diff
@@ -712,53 +679,52 @@ def _visit_value_and_its_immediate_references(obj, visitor):
...
-    model._document = None
-    for ref in model.references():
-        ref._document = None
-    empty_doc = Document()
-    empty_doc.add_root(model)
-
-    yield model
+    if len(existing_docs) == 0:
+        # no existing docs, use the current doc
+        doc = curdoc()
```
bryevdv: I'm not sure `curdoc` is advised here. Although I have not seen it done, it's not a huge stretch to imagine a Bokeh app that produces embedded stand alone bokeh plots as some kind of output. Then this code would intertwine the "produced" standalone plots with the bokeh app curdoc and I doubt that is desirable.

bryevdv:
I'm not sure I favor keeping the context manager, but I guess I don't feel too strongly against it. But if we are going to do so we should own it, i.e. rename it `ModelInDocument`, add tests and documentation, and move it some place better like `bokeh.util`

bokeh/model.py
jsignell: that curdoc is what fixes the theme.

bryevdv: It's one way, but it seems that copying any theme to a new `Document` should also be feasible?

examples/embed/embed_themed.py
```diff
@@ -0,0 +1,66 @@
...
+from bokeh.io import curdoc, show
+from bokeh.themes import Theme
+from bokeh.plotting import figure
+
+x1 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+y1 = [0, 8, 2, 4, 6, 9, 5, 6, 25, 28, 4]
+
+p1 = figure(title='PINK TITLE')
+p1.background_fill_color = "#999999"
+p1.border_fill_color = "#999999"
```
bryevdv: these colors can go in the theme, sorry I was not explicit but I was suggesting making the theme have a bit more content.

bokeh/model.py
jsignell: I think it fits the expectation that the `curdoc` would get rendered when embedding. This would happen for show() or save()

jsignell:
screen shot 2017-02-28 at 3 12 23 pm

examples/embed/embed_themed.py
```diff
@@ -0,0 +1,90 @@
...
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <title>Bokeh Scatter Plots</title>
+        {{ js_resources }}
+        {{ css_resources }}
+        {{ script }}
+        <style>
+            body {
+                text-rendering: optimizeLegibility;
```
mattpap: What's the purpose of setting this? `text-rendering` is a SVG property.

jsignell: I think I was copying from another example. I will take it out.

Issue #5884, opened 2017-02-17
Create Marker class that encompasses all markers and allow specific marker type to be specified from data


hafen:
I keep running into an issue with markers where there is a need to vary marker according to some attribute of the data. As far as I understand it, this can only be done by splitting the data up into different data sources and making separate GlyphRenderers for each marker. This is problematic for several reasons.

Take this example, where I'm varying color and marker according to a variable in the data:

<img width="497" alt="screen shot 2017-02-16 at 6 21 44 pm" src="https://cloud.githubusercontent.com/assets/1275592/23050058/dbf32c4e-f474-11e6-938c-71c0ab0ed63b.png">

As illustrated above, one of the problems with breaking the data up and specifying multiple GlyphRenderers is that when you allow tap selection, the selection is restricted to only that group / marker. There are several related issues dealing with the various tools and callbacks when you want to think of the entire set of markers as a whole. The bigger issue is that it makes things a lot more complicated at the interface level for dealing this special case (in particular, it makes things pretty messy inside rbokeh).

To me, the ideal solution would be instead of individual classes "Circle", "Square", etc., having a class "Marker" that has an attribute "marker" that can be any one of: "Asterisk", "Circle", "CircleCross", "CircleX", "Cross", "Diamond", "DiamondCross", "InvertedTriangle", "Square", "SquareCross", "SquareX", "Triangle", "X". As far as I understand, all of the current marker classes have the same attributes, so it seems like this would be possible.

So basically instead of having a specification with things like this:

```json
{
  "type": "GlyphRenderer",
  "attributes": {
    "glyph": {
      "type": "Circle",
      ...
    }
  }
},
{
  "type": "Circle",
  "attributes": {
    "x": {
      "units": "data",
      "field": "x"
    },
    "y": {
      "units": "data",
      "field": "y"
    },
    ...
  }
}

{
  "type": "GlyphRenderer",
  "attributes": {
    "glyph": {
      "type": "Square",
      ...
    }
  }
},
{
  "type": "Square",
  "attributes": {
    "x": {
      "units": "data",
      "field": "x"
    },
    "y": {
      "units": "data",
      "field": "y"
    },
    ...
  }
},
...
```

With each having a separate ColumnDataSource. Instead, something like this would make life so much easier:

```json
{
  "type": "GlyphRenderer",
  "attributes": {
    "glyph": {
      "type": "Marker",
      ...
    }
  }
},
{
  "type": "Marker",
  "attributes": {
    "x": {
[...]

Issue #2989, opened 2015-10-14
Selection bug for plot of multiple columns on the same plot from the same data source


amcpherson:
When I use multiple plotting calls (eg Figure.scatter) to plot multiple columns from the same data source, there is a bug causing problems with selecting data points.

A minimal example reproducing the error is below, and can be run with bokeh-server.  Upon selection of data points they are either unselected or flicker and remain selected.

```
import numpy as np
import pandas as pd

from bokeh.plotting import figure
from bokeh.models import Plot, ColumnDataSource
from bokeh.properties import Instance
from bokeh.server.app import bokeh_app
from bokeh.server.utils.plugins import object_page
from bokeh.models.widgets import HBox


class TestApp(HBox):
    extra_generated_classes = [["TestApp", "TestApp", "HBox"]]

    plot = Instance(Plot)
    source = Instance(ColumnDataSource)

    @classmethod
    def create(cls):
        obj = cls()

        data = pd.DataFrame(np.random.randn(20, 4), columns=['x', 'y', 'z', 'w'])

        obj.source = ColumnDataSource(data=data)

        obj.plot = figure(tools='box_select')
        obj.plot.scatter('x', 'y', asdf=1, source=obj.source)
        obj.plot.scatter('z', 'w', source=obj.source)

        obj.children.append(obj.plot)

        return obj


@bokeh_app.route("/test/")
@object_page("sin")
def make_sliders():
    app = TestApp.create()
    return app

```

azjps:
I don't believe that charting multiple slices of a single ColumnDataSource in a single figure is explicitly supported. It should work if you provided two separate ColumnDataSources, one for ['x', 'y'] and the other for ['z', 'w'].

On the js side, each ColumnDataSource is associated with a single selection manager which stores a single selection slice. [The selection tools](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/tool/gestures/box_select_tool.coffee#L60) update the selection manager for each glyph renderer in a loop, so a selection in your example would probably arbitrarily select rows associated with the second series charted.

I somewhat doubt this will be "fixed" anytime soon, but I'd suppose the way to go about fixing it would be to construct selection managers per renderers instead of per ColumnDataSources. That might not be desirable though since ColumnDataSources selections can be done in other ways in the UI besides through chart renderers.

amcpherson:
[...]

PR #5933, opened 2017-03-01
Filterable and groupable CDS

### LINKED ISSUE #4070: Plotting with subsets (row-wise) of ColumnDataSources (like filtered data)

clairetang6:
This PR is a second attempt to create a filterable CDS. The first attempt in [this PR](https://github.com/bokeh/bokeh/pull/5828) created a new data source called the TableDataSource. This second attempt instead puts the filtering functionality into the ColumnDataSource (CDS). 

In this approach, the data and filters are separated by pulling out the CDS's data into a new model called the ColumnDataStore (data store). The data store would be responsible for holding the data as well as the selection on the full data set. (It would also be responsible for streaming and patching the data and the CDS could have stream and patch methods that called the data store's stream and patch methods.)

I additionally added a group property to the CDS, which specifies a column name and a group name to restrict to. This allows for an API that looks like (described by @bryevdv [here](https://github.com/bokeh/bokeh/issues/5884#issuecomment-280669400))

```python
fig.circle('sepal_width', 'petal_width', source=source, group=('species', 'setosa'))
```

With the first two commits in this PR, you can have something like:

iris_groups

```python
from bokeh.models.data_stores import ColumnDataStore

output_file("filter_group_cds.html")

import numpy as np

data_store = ColumnDataStore(flowers)
source0 = ColumnDataSource(data_store)
source1 = ColumnDataSource(data_store, filter=lambda cds: np.array(cds.data_store.data['species']) == 'setosa')
source2 = ColumnDataSource(data_store, group=['species', 'versicolor'])
source3 = ColumnDataSource(data_store, group=['species', 'virginica'])

plot_size_and_tools = {'plot_height': 300, 'plot_width': 300,
                        'tools':['box_select', 'reset']}

fig1 = figure(title="Full data set", **plot_size_and_tools)
fig1.circle(x='petal_length', y='petal_width', source=source0, color='black')

fig2 = figure(title="Composed of filtered and grouped data", x_range=fig1.x_range, y_range=fig1.y_range, **plot_size_and_tools)
fig2.circle(x='petal_length', y='petal_width', source=source1, color='red')
fig2.square(x='petal_length', y='petal_width', source=source2, color='green')
fig2.diamond(x='petal_length', y='petal_width', source=source3, color='blue')

show(row(fig1, fig2))
```

One issue shown the GIF above is that selecting points on the right figure only works with the last renderer. This issue is discussed in [issue 2136](https://github.com/bokeh/bokeh/issues/2136) and [issue 2989](https://github.com/bokeh/bokeh/issues/2989).

This PR is still in early stages, and I'm looking for feedback at any level. The group property may not be strictly necessary. Users can filter as in source1 in the above example for a specific group on their own and a higher-level plot.scatter method (with a groupby parameter) could use filter internally as well. 

With this PR, each filter is represented through one CDS. Another approach could be to have a CDS contain a mapping of renderers to filters. Then, each renderer could use the same CDS. The filter could be made on the creation of a renderer and be added to the CDS's renderer-filter mapping. 

- [ ] issues: fixes #4070
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/data_stores.py
+++ b/bokeh/models/data_stores.py
@@ -1,0 +1,97 @@
+from __future__ import absolute_import
+
+from ..core.properties import Any, ColumnData, Dict, Instance, List, Seq, String
+from ..model import Model
+from ..util.dependencies import import_optional
+
+pd = import_optional('pandas')
+
+class ColumnDataStore(Model):
+
+    selected = Dict(String, Dict(String, Any), default={
+        '0d': {'glyph': None, 'indices': []},
+        '1d': {'indices': []},
+        '2d': {}
+    }, help="""
+    A dict to indicate selected indices on different dimensions on this DataSource. Keys are:
+
+    .. code-block:: python
+
+        # selection information for line and patch glyphs
+        '0d' : {
+          # the glyph that was selected
+          'glyph': None
+
+          # array with the [smallest] index of the segment of the line that was hit
+          'indices': []
+        }
+
+        # selection for most (point-like) glyphs, except lines and patches
+        '1d': {
+          # indices of the points included in the selection
+          indices: []
+        }
+
+        # selection information for multiline and patches glyphs
+        '2d': {
+          # mapping of indices of the multiglyph to array of glyph indices that were hit
+          # e.g. {3: [5, 6], 4, [5]}
+        }
+
+    """)
+
+    column_names = List(String, help="""
+    A list of names for all the columns in this ColumnDataStore.
+    """)
+
+    data = ColumnData(String, Seq(Any), help="""
+    Mapping of column names to sequences of data. The data can be, e.g,
+    Python lists or tuples, NumPy arrays, etc.
+    """).asserts(lambda _, data: len(set(len(x) for x in data.values())) <= 1,
+                 lambda: warnings.warn("ColumnDataStore's columns must be of the same length", BokehUserWarning))
+
+    def __init__(self, *args, **kw):
+        ''' If called with a single argument that is a dict or
+        pandas.DataFrame, treat that implicitly as the "data" attribute.
+
+        '''
+        if len(args) == 1 and "data" not in kw:
+            kw["data"] = args[0]
+
+        # TODO (bev) invalid to pass args and "data", check and raise exception
+        raw_data = kw.pop("data", {})
+
+        if not isinstance(raw_data, dict):
+            if pd and isinstance(raw_data, pd.DataFrame):
+                raw_data = self._data_from_df(raw_data)
+            else:
+                raise ValueError("expected a dict or pandas.DataFrame, got %s" % raw_data)
+
+        super(ColumnDataStore, self).__init__(**kw)
+
+        self.column_names[:] = list(raw_data.keys())
+        self.data.update(raw_data)
+
+    @staticmethod
+    def _data_from_df(df):
+        ''' Create a ``dict`` of columns from a Pandas DataFrame,
+        suitable for creating a ColumnDataSource.
+
+        Args:
+            df (DataFrame) : data to convert
+
+        Returns:
+            dict(str, list)
+
+        '''
+        index = df.index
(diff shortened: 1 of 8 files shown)


bryevdv:
My initial concern is that we will have to move the data store functionality up into CDS somehow. We have been demonstrating `source.data = new_data` for years now, in both JS and python, and breaking all that code would be too disruptive.

bryevdv:
I have some further thoughts on this. Perhaps we can discuss in a call. A very high level idea description would be this:

Leave CDS basically alone. Make filters/groups be things *external* to a CDS that can provide a *view* on the data that the CDS contains. Then things like glyph renderers can be updated to operate either directly over a CDS, or if a "view" is configured, to operate over the view that wraps the CDS. Ideally we can define something akin to an "iteration protocol" that both CDS and views both have, then things that need to iterate over data don't have to care which one they are iterating over. 

In other words, instead of pushing the current CDS "down" into another object `ColumnDataStore` that is contained by a new and too-different CDS, leave the current CDS as-is and elevate views to be something above it (i.e. that can be configured with a CDS)

clairetang6:
That sounds similar to the first approach I took where I created a new model (TableDataSource) that contained a reference to a CDS. 

What I am trying now as a third approach is to have a mapping of filters-to-renderers stored by a CDS. Then the data for the full data set is still contained in the CDS and .data will be unchanged. If a glyph renderer has an associated filter in that mapping, then they can use that filter.

bryevdv:
> That sounds similar to the first approach I took where I created a new model (TableDataSource) that contained a reference to a CDS.

Yes the thought occurred to me, it might be a matter of naming (names are important!) I think it's important to establish these things as *optional views* on a *data source*, rather than a new kind of at a source. And I think a unified iteration protocol is also important (I don't recall if the other PR did that or not, offhand)

clairetang6:
Superseded by [5982](https://github.com/bokeh/bokeh/pull/5982)

PR #5941, opened 2017-03-01
Generalized event system

jlstevens:
This PR is one I have been working on with @bryevdv on my fork (https://github.com/jlstevens/bokeh/pull/4) that aims to fix issue #5278 and will possibly help address issue #3393. This PR supersedes #5276.

## Aim

The goal is to allow users to specify Python and Javascript callbacks associated with events. Some suggestions for potential events include UIEvents (including hammer events such as 'tap', 'press', keypress events), model related events such as button click events as well as document level events such as 'busy'.  Other proposals for events may be found on [this wiki page](https://github.com/bokeh/bokeh/wiki/Generalizing-Events).

As there are many potential event types, the goal is to introduce a very general mechanism that abstracts over all these potential event types (hammer, DOM, model related, document). This would be a very useful feature to have for building bokeh apps as well as for third party libraries used to build interactive Bokeh visualizations such as HoloViews.

## Architecture outline

The API is very similar to the existing ``on_change`` and ``js_on_change`` API used to register callbacks triggered by property changes. 

Events callbacks are defined on ``Model`` and stored on the JS side as a property ``js_event_callbacks`` and on the Python side in the ``_event_callbacks`` attribute of the``EventCallbackManager`` mixin class (which is analogous to the ``PropertyCallbackManager``  class - formerly known as ``CallbackManager``).

An ``EventManager`` instance exists per document in both on Python and Javascript - its job is to route events from wherever they are generated to the appropriate models that can process them. ``Model`` instances are responsible for actually invoking the registered callbacks in response to an event, triggered on the event manager. The ``EventManager`` also handles the communication of events from Javascript to Python as necessary.

To achieve this, ``Model`` has a ``subscribed_events`` property which specifies which events are being listened to on the Python side. These are the events that are to be communicated to Python so that Python callbacks can respond accordingly. Both Javascript and Python callbacks receive an event instance (e.g ``Tap``) which contains information such as the position of the tap. In terms of the communication protocol, there is a single new ``Event`` message that is converted into corresponding event instances once received.
 
## Current status

You can now use ``on_event`` and ``js_on_event`` with the currently defined set of events to execute either Javascript or Python callbacks. You can see the API in action in the [fourier_animated.py](https://github.com/bokeh/bokeh/files/811526/fourier_animated.py.zip) example I am currently using for testing. I intend to write a new example app from scratch and include it in this PR later.

## Open questions

The main architectural issue right now is that I need a way to supply event instances with a particular model id in order to target a particular model. For instance, events from a given ``UIEvents`` instance only apply to a particular ``Plot`` instance. What is the best way to get the appropriate model id - at the point where the events are generated? For more information about the problem, see  https://github.com/jlstevens/bokeh/pull/4.

## TODO

I have already listed various TODO items at https://github.com/jlstevens/bokeh/pull/4 and here are the most important outstanding items:

- [x] Write an example of a Button widget click event.
- [x] Make sure all classes have decent docstrings
- [x] Illustrate custom events by building a custom model.
- [x] Testing

**Edit**: to avoid having to refer back to the PR on my fork, here are the other TODO items I assigned myself:

- [x] Rename ``js_callbacks`` to ``js_property_callbacks``
- [x] Add more event types and check they have useful contents
- [x] Make a new bokeh app example specifically for illustrating events
 (*edit*: superseded for now with two demos in ``examples/howto`` ).
- [x] Type annotations for typescript

## Possible Future work

Currently events originating in JS can be communicated to Python but events originating in Python cannot currently be communicated to JS. An earlier prototype demonstrated that this feature is possible with the proposed architecture but this feature involves some tricky issues connecting up the communication protocol that are outside the scope of this PR.

In addition, it is trivial to add a Javascript API to the current system. This was successfully prototyped but is omitted here as there was no obvious naming scheme for the JS methods.


jbednar:
Looks very promising!

> events originating in Python cannot currently be communicated to JS. An earlier prototype demonstrated that this feature is possible with the proposed architecture but this feature involves some tricky issues connecting up the communication protocol that are outside the scope of this PR.

Is the idea that such a feature would be submitted as a new PR after this one is merged?  It does seem very important, though I guess not so important as to hold up this PR from being merged.

jlstevens:
> Is the idea that such a feature would be submitted as a new PR after this one is merged? It does seem very important, though I guess not so important as to hold up this PR from being merged.

Exactly.

That said, we don't have that many ideas of what such events could be right now. One of them is the 'busy' event suggested in #3393. @bryevdv also proposed a possible 'demo mode' where Python could drive interaction e.g simulating button clicks etc.

In other words, this PR implements enough to cover most of what we want but a separate PR to communicate events from Python to JS will be needed to complete the event system.

jbednar:
Seems like one might want to have a loop running on the Python side forcing updates on the JS side, whenever new data is available (e.g. from a simulator).  Anyway, sure, can be addressed later.

jlstevens:
@mattpap Any ideas why the unit tests are failing?

mattpap:
> Any ideas why the unit tests are failing?

There are plenty of things you need to update, but the following is unrelated to this PR and peculiar in general:

```
module.exports = class VirtualConsole extends EventEmitter {
                                              ^
TypeError: Class extends value undefined is not a function or null
```
Locally (in a clean setup) I have exactly the same dependencies as on travis ci and can't reproduce this issue. Given the setup, there should be no error here. The error persists after restarts.

mattpap:
OK, so as soon as I clicked "Comment", I realized that this PR is (kind of) at fault. The reason for the failure is that `bokehjs/src/coffee/events.coffee` conflicts with `events` library. Preferably you shouldn't put anything directly in to `bokehjs/src/coffee`, but to its submodules (`core` in this case).

jlstevens:
> Preferably you shouldn't put anything directly in to bokehjs/src/coffee, but to its submodules (core in this case).

Thanks for clarifying the problem! I'll do that now.

jlstevens:
The tests have now gone green except for ``GROUP=integration`` which is complaining about missing PNG files (I assume this is unrelated).

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -4,12 +4,14 @@ import {Events} from "./events"
...
 import {getDeltaY} from "./util/wheel"
+import {HammerEvent, PointEvent, KeyDown} from "./bokeh_events"
+
 
 export class UIEvents
   @prototype extends Events
 
   # new (toolbar: Toolbar, hit_area: Element)
-  constructor: (@toolbar, @hit_area) ->
+  constructor: (@toolbar, @hit_area, @event_manager, @plot) ->
```
mattpap: `@plot` isn't used anywhere.

bokehjs/src/coffee/model.coffee
```diff
@@ -17,6 +19,39 @@ export class Model extends HasProps
       for cb in callbacks
         @listenTo(@, evt, () -> cb.execute(@))
 
+    @listenTo(@, 'change:js_event_callbacks', () -> @_update_event_callbacks)
+
+
+  _process_event : (event) ->
+    # Given an applicable event, execute the associated callback
+    if not event.constructor.event_name
+       console.warn('Event is still passed as a string: ' + event)
```
mattpap: Use bokehjs' logger here and everywhere else.

bokehjs/src/coffee/model.coffee
```diff
@@ -17,6 +19,39 @@ export class Model extends HasProps
...
+
+  _update_event_callbacks : () ->
+    if not @document?
+      # File an issue: SidePanel in particular seems to have this issue
+      console.warn('WARNING: Document not defined for updating event callbacks')
+      return
+    @document.event_manager.subscribed_models.push(@.id)
+
+  _doc_attached : () ->
+    if Object.keys(@js_event_callbacks).length != 0
```
mattpap: `core/util/object` offers `isEmpty`.

bokehjs/src/coffee/core/bokeh_events.coffee
```diff
@@ -0,0 +1,96 @@
+# Granular event classes for use with the event system
```
mattpap: I would like this to be actually moved to `core/`, especially that `core/ui_events` is the only module where this is imported.

bokehjs/src/coffee/model.coffee
jlstevens: Thanks! I'll use this.

bokehjs/src/coffee/model.coffee
jlstevens: Will do.

bokehjs/src/coffee/core/ui_events.coffee
jlstevens: Yes, this is WIP and only partially implemented. It relates to the issue I ask above about finding the appropriate model id to pass to event instances. Any suggestions?

bokehjs/src/coffee/core/bokeh_events.coffee
jlstevens: Is there another ``/core`` as I thought I had just moved this file there?

mattpap: I noticed the rename, but not move.

jlstevens: Ah ok, I was wondering if I was confused about something. :-)

bryevdv:
integration tests temporarily disabled in master. Merge/rebase master and remaining tests sholud go green.

jlstevens:
I'll try to summarize the issue with passing model ids to allow for discussion/suggestions/feedback.

In short, events are processed by models (e.g callbacks are invoked on the associated models) but most of the events I have encountered don't *originate* on the models. Instead, they originate from some code that interacts with the user that is associated with a model in some way.

For instance of ``UIEvents``, the events need to be routed to a particular plot model (so things like ``Tap`` events go to the right place). This is why I am passing ``@plot`` to the ``UIEvents`` constructor - this would allow me to access the plot model id can then be supplied when constructing the various events generated by ``UIEvents``.

Although this approach will work fine for this fairly large class of events, I am wondering how you might get the appropriate model id in other cases. Is there a more general approach? For a concrete example of something I intend to implement shortly, how about routing a button click event to a specific button model by passing the appropriate model id when issuing the event? What about other possible types of event, such as tool related events?

Perhaps it is safe to assume that views (or whatever code will be issuing events) will typically have easy access to the corresponding model(s)? If that is the case, then perhaps there won't be much plumbing necessary without needing a more general mechanism.

bryevdv:
> Perhaps it is safe to assume that views (or whatever code will be issuing events) will typically have easy access to the corresponding model(s)? If that is the case, then perhaps there won't be much plumbing necessary without needing a more general mechanism.
 
All Bokeh views have a corresponding `.model`

jlstevens:
> integration tests temporarily disabled in master. Merge/rebase master and remaining tests should go green.

I've rebased so hopefully all the tests will go green!

jlstevens:
> All Bokeh views have a corresponding .model

So can I assume that getting the appropriate model ids to where events are generated is a non-issue?

bokeh/model.py
```diff
@@ -274,6 +275,24 @@ def __str__(self):
...
+    help="""A mapping of event names to lists of CustomJS callbacks.
+
+    Typically, rather then modifying this property directly, callbacks should be
+    added using the ``Model.js_on_event`` method:)
+
+    .. code:: python
+
+        callback = CustomJS(code="console.log('tap event occured')")
+        plot.js_on_event('tap', callback)
+   """)
```
bryevdv: quotes only indented three spaces

bokeh/model.py
```diff
@@ -274,6 +275,24 @@ def __str__(self):
...
+
+    subscribed_events = List(String, help="""
+      List of events that are subscribed to by Python callbacks. This is
+      the set of events that will be communicated from BokehJS back to
+      Python for this model.
```
bryevdv: text block should be flush with property indentation

bokeh/util/callback_manager.py
```diff
@@ -42,14 +43,49 @@ def _check_callback(callback, fargs, what="Callback functions"):
...
+            callback(event)
+
+    def _update_event_callbacks(self):
+        if self.document is None:
+            return
+
+        events = self._event_callbacks.keys()
+        if events:
+            self.document.event_manager.subscribed_models.add(self)
+
```
bryevdv: will want to check event callbacks as well, to ensure they have the expected signature

bokehjs/src/coffee/model.coffee
jlstevens: Actually, I expect this branch to be removed once the PR is complete. I'll make sure to use the logger anywhere else I might be using console.

bokeh/util/tests/test_callback_manager.py
```diff
@@ -192,7 +192,7 @@ def test_trigger():
     assert good.last_new == 43
 
 def test_trigger_with_two_callbacks():
-    m = cbm.CallbackManager()
+    m = cbm.PropertyCallbackManager()
     good1 = _Good()
     good2 = _Good()
     m.on_change('foo', good1.method)
```
bryevdv: once there is checking for event callbacks, will need similar set of tests for event callbacks as well.

jlstevens: Agreed.

bokeh/util/callback_manager.py
jlstevens: How you you recommend checking the callback signatures? Using the ``inspect`` module?

bokeh/model.py
jlstevens: Fixed in 065a26a

bokeh/model.py
jlstevens: Fixed in 065a26a

bokeh/util/callback_manager.py
bryevdv: Use the existing configurable `_check_callback` function that the property callback manager users

bokehjs/src/coffee/model.coffee
jlstevens: Addressed in 72b9c31

bokeh/util/callback_manager.py
jlstevens: Addressed in f69ec4d

bokehjs/src/coffee/core/ui_events.coffee
jlstevens: @plot is now used to get the model id in 8c6429f

jlstevens: (If I accidentally pinged GitHub use [@plot](https://github.com/plot), sorry that was a mistake!)

jlstevens:
@mattpap @bryevdv 

I've not hooked up the model id matching code for all UI events. That said, I've noticed that the data x,y coordinates processed [here](https://github.com/bokeh/bokeh/pull/5941/files#diff-061192a746e9c5eeb3e560f95d2a2e02R75) results in the correct x-coordinate but the wrong (inverted) y-coordinate. 

Any idea why I seem to be getting -y values even though I am processing x and y in a symmetric way?

bokehjs/src/coffee/model.coffee
jlstevens: Addressed in fdd342b. As I am working in the file, I also replaced three instances of ``console.log`` in document.coffee with ``logger.warn``.

jlstevens:
I've now introduced a button click event for the button widget. For instance you can now use this in ``examples/app/export_csv/main.py`` instead of setting ``button.callback``:

```python
button.js_on_event('button_click', CustomJS(code="console.log('Click event!')"))

def click_callback(event):
    print('Click on button model %s' % event.model_id)

button.on_event('button_click', click_callback)
```

In addition, I've simplified how events are emitted using a model ``trigger_event`` method. Great suggestion @bryevdv!

jlstevens:
Note that I now intend to finish covering all the hammer/UI event types. I'm also happy to add new widget events but you will need to prompt me with suggestions of what you think might be useful.

bryevdv:
@jlstevens here is another event that could be added (and issue closed) https://github.com/bokeh/bokeh/issues/3748

jlstevens:
> Something we also need to think about for this PR: how and where to document all this.

Agreed!

Any ideas about why the y-data coordinate in inverted? Another issue (pointed out to me by @philippjfr ) is that key press events are captured from anywhere in the document and not just when bokeh has focus.

bryevdv:
> Any ideas about why the y-data coordinate in inverted?

You'll have to elaborate on what you are actually getting vs what you are expecting. Actual HTML canvas coords start top left, so `sy` is 0 at the top (`s` for "screen")  We have a conversion internally to a system that has coords start at the bottom left, so `vx` is 0 at the bottom (`v' for "view") You can report either or both. I'm not sure which you are expecting to see.

bryevdv:
> Another issue (pointed out to me by @philippjfr ) is that key press events are captured from anywhere in the document and not just when bokeh has focus.

Not sure if there is anything we can do about that easily or not. Will have to investigate.

jlstevens:
Here is a gif illustrating what I mean by the y axis being inverted:

coords

You can see that the x-coordinate correspond to where I am clicking as measured by the axes. The y coordinate however seems to be inverted with negative values when I am above the central circle and positive values when I am below it (which is the opposite of what is read off the y-axis).

bokehjs/src/coffee/core/bokeh_events.coffee
```diff
@@ -0,0 +1,136 @@
...
+
+  @from_event : (e, model_id=null) ->
+    # Given a hammer event, construct a corresponding Event instance
+    return new @({sx: e.bokeh['sx'], sy : e.bokeh['sy'], model_id: model_id})
+
+  _customize_event : (plot) ->
+    xmapper = plot.plot_canvas.frame.x_mappers['default']
+    ymapper = plot.plot_canvas.frame.y_mappers['default']
+    @x = xmapper.map_from_target(@sx)
+    @y = ymapper.map_from_target(@sy)
```
bryevdv: pretty sure you need to call the `sx_to_vx`, `sy_to_vy` functions first, before passing to `map_from_target`

bryevdv: Though, I thought our hammer plumbing also stored the `vx` `vy` coords already? If so those are what you want to pass to `map_from_target`

jlstevens: Thanks! That fixed it in 35256e6

jlstevens:
> here is another event that could be added (and issue closed) #3748 Could add "lod_start" and "lod_end" to plots, maybe.

That seems very reasonable. I just need to know where I should be issuing these events. My guess that it might be [here](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/plots/plot_canvas.coffee#L506-L513) in ``PlotCanvasView.render``?

```coffeescript
    if Date.now() - @interactive_timestamp < @model.plot.lod_interval
      @interactive = true
      lod_timeout = @model.plot.lod_timeout
      setTimeout(() =>
          if @interactive and (Date.now() - @interactive_timestamp) > lod_timeout
            @interactive = false
          @request_render()
        , lod_timeout)
```

Should 'lod_start' be emitted at the start of the callback in ``setTimeout`` and 'lod_end' be emitted at the end of this callback? Also, I note that I define the event names 'panstart' and 'panend' based on [these lines](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/ui_events.coffee#L28-L30) and not 'pan_start' and 'pan_end'. I don't mind which style we use as long as it is consistent!

bryevdv:
I would say exactly when ` @interactive = true` is set and then exactly when `@interactive = false` is set.

jlstevens:
After adding an ``@lod_started`` attribute to ``PlotCanvasView`` in 2757936, the new ``LODStart`` and ``LODEnd`` events seem to be working quite nicely!

As for the issue with keypress events triggering regardless of focus, do you know why [the key event listener](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/ui_events.coffee#L46-L49) in ``UIEvents`` is currently at the ``document`` level instead of ``@hit_area`` (the ``@canvas_view.el`` of PlotCanvas)?

```coffeescript
    @hit_area.addEventListener("wheel", (e) => @_mouse_wheel(e))
    document.addEventListener("keydown", (e) => @_key_down(e))
    document.addEventListener("keyup", (e) => @_key_up(e))
```

I will try attaching the key listener to ``@hit_area`` to see if it fixes the focus issue. That said, if the listener was attached to ``document`` for a good reason, something else may break. In the worst case, how about I set up some new event callbacks ``_hit_area_key_down`` and ``_hit_area_key_up`` just for emitting events?

jlstevens:
Unfortunately, after playing around with it, I can't seem to attach key event callbacks to anything *other* than document and have the events fire. I'm not sure why but key events don't seem to trigger on ``bk-plot-layout`` or any other node below (e.g ``bk-plot-wrapper``, ``bk-canvas-wrapper`` or ``bk-canvas-events``).

Seems that it is more complicated than I thought - the odd thing is other DOM events such as 'wheel' work fine!

mattpap:
@jlstevens, you need to have a focusable element to catch those events. Using `document` is typical appraoch. Usually done with jquery where you can filter by origin. We don't use jquery anymore, so you will have to do this manually.

jlstevens:
@mattpap Thanks for the suggestion but unfortunately, I'm still having trouble. :-(

I've tried looking at the event target but it always points to the body. In a very simple example (without bokeh) I got this working but I had to click on the element to give it focus. Despite my best efforts, I never managed to get any bokeh element to indicate that it has focus. I'm not sure if this [stackoverflow post](https://stackoverflow.com/questions/56771/how-do-you-set-focus-to-the-html5-canvas-element) helps at all.

What I have managed to do is to limit keypress events to when the mouse is over the plot (b61a4b7). This is marginally better than sending all keypress events but still not ideal: it would be really helpful to have a concept of a plot having focus even if the mouse isn't over it.

I'm afraid I don't know how to do better than this right now! Any suggestions?

bryevdv:
> it would be really helpful to have a concept of a plot having focus even if the mouse isn't over it.

This would be a fairly radical change and I don't think its in scope for his PR. I'm actually wondering if we should postpone all key-related events. I am not interested in providing a tool that makes it trivial for bad actors to create disguised key-logger.

jlstevens:
@bryevdv I agree having focusable plot containers is outside the scope of this PR.

That said, I think it would be a real shame to lose key events as I believe they would be very useful! I think I might have a suggestion that alleviates these concerns.

Why not have an explicit keyboard input tool?

* Solves the focus issue (user explicitly switches the tool on or off)
* Helps alleviates security worries - there would be a visual indicator (tool icon) that keyboard input is being captured. It would be an actual bokeh tool and not a disguised one! :-) 
* It would be a nice demonstration of how tools can integrate the event system.

I think this approach would be very doable and that without keyboard input,  the event system would seem rather incomplete..

jbednar:
Would defining focus for a plot be a fairly radical change, just for keypress events?  Seems like that's the only category of events that having focus would matter for, and it doesn't seem like there is much backwards compatibility to worry about for keypress events, but maybe Bokeh has hidden functionality for keypresses that I don't know about.

bryevdv:
I mean there is currently no notion or machinery to define focus, or for visually indicating it, at all, so that's a distinct change and new functionality, and I'd rather not conflate it with this PR.

jlstevens:
> I mean there is currently no notion or machinery to define focus, or for visually indicating it

Couldn't a bokeh tool address these issues?

jbednar:
Would it make sense to make a separate, presumably small, PR for selecting focus and merge it before this one?  Or is that too much of a can of worms to open, that could derail getting this PR merged before the next release? Keypress events are a very clear requirement for several client projects of mine, so I hope there will be a way to work out something feasible (e.g. with a tool as J-L suggests) to support them.

bryevdv:
> Couldn't a bokeh tool address these issues?

Possibly, but I think there's plenty of details to discuss. I'd rather get a solid initial PR merge then figure out additions later. It's clear this work will *support* whatever solution is decided on, so let's not hold up this (and also let's keep this as small an self-contained as possible)

jlstevens:
> Possibly, but I think there's plenty of details to discuss. I'd rather get a solid initial PR merge then figure out additions later. It's clear this work will support whatever solution is decided on, so let's not hold up this (and also let's keep this as small an self-contained as possible)

Ok, I agree. Let's get this merged ASAP and I hope we will be able to consider a second PR for an explicit keyboard capture tool before the next release. Once this PR is merged, I suspect it would be fairly easy to do and a separate PR would be a good place to continue this discussion.

bryevdv:
As a reminder custom extensions offer a route to keyboard events regardless of what does or does not get merged into the core library or when.

jlstevens:
I've now removed the ``KeyDown`` event and implemented the pinch events. I'm not sure how to test the pinch events though - is there a good way to emulate them?

jlstevens:
One goal of this PR is to demonstrate how users can write their own events for custom models. I quite liked this [custom drawing tool example](http://bokeh.pydata.org/en/latest/docs/user_guide/extensions_gallery/tool.html#userguide-extensions-examples-tool) so I extended that example to define 'draw events' which can be processed with either JS or Python callbacks as shown in this [updated example](https://github.com/bokeh/bokeh/files/826187/draw_events.py.zip). Note that I made some minor changes to run this example as a Bokeh app.

Thankfully, the process of adding events is trivial and this seems like a nice easy way to build a drawing app! For instance, I can imagine running some handwriting recognition process in Python: 

draw_events

I have not yet committed this file as I don't know where it belongs - presumably it will go somewhere in the documentation.

bryevdv:
> I've now removed the KeyDown event and implemented the pinch events. I'm not sure how to test the pinch events though - is there a good way to emulate them?

for one time smoke-testing? android or ios sim I guess? or a real tablet, there are some in the office.

jlstevens:
> for one time smoke-testing? android or ios sim I guess? or a real tablet, there are some in the office.

Thanks. I managed to test it on my android phone. Pinch events seem to be working fine!

I'll add the remaining mouse events shortly - "mouseenter", "mouseleave", "wheel" . Once the rotation events are in, that is everything in ``UIEvents`` covered except key events.

jlstevens:
I've now pushed the pinch events and mouseenter and mouseexit. For some reason, I find mouseenter fires but mouseexit doesn't! Very odd.

This also brings up the issue of naming:

```coffeescript
    @hit_area.addEventListener("mouseleave", (e) => @_mouse_exit(e))
...
  _mouse_exit: (e) ->
    # NOTE: move:exit event triggered unconditionally
    @_bokify_point_event(e)
    @trigger('move:exit', e)
 ```

The DOM event is called 'mouseleave', the callback method name is '_mouse_exit' and the backbone event is 'move:exit'. I've generally worked in order according to the method names but I am happy to use a different convention for naming events if you prefer...

**Comment by jlstevens** [comment]
Actually, I can answer my own question above. As I look up the Bokeh event class using ``e.type`` (by matching the ``event_name`` attribute), it will only work if I follow the DOM event names - which seems the sensible approach anyway.

**Comment by jlstevens** [comment]
I have now fixed the mouse leave event issue and added the mouse wheel event. Just the rotation events in ``UIEvent`` left!

**Comment by mattpap** [comment]
@jlstevens, when you will be done with major work in this PR, please rewrite `core/bokeh_events.coffee` in typescript. This will be our new policy to write all new modules in typescript.

**Comment by bryevdv** [comment]
> when you will be done with major work in this PR, please rewrite core/bokeh_events.coffee in typescript. This will be our new policy to write all new modules in typescript. 

I mean I am generally +1 but is there a record of a discussion where such a decision was made?

@jlstevens FYI the interactive legends PR introduced some merge conflicts

**Comment by jlstevens** [comment]
> FYI the interactive legends PR introduced some merge conflicts

I've just used the online GitHub merge conflict resolution tool for the first time. Fingers crossed I did it correctly!

**Comment by mattpap** [comment]
> I mean I am generally +1 but is there a record of a discussion where such a decision was made?

We had brief discussion about this last week. This isn't written down anywhere nor official at this point, but is an obvious thing to do (even without a policy).

**Comment by bryevdv** [comment]
> but is an obvious thing to do (even without a policy).

Like I said, I'm +1 but if we are to do this, just as a matter of good project management it needs to *be* a policy, and I'd also say that we need to support the decision with docs that help people make that transition.

**Comment by jlstevens** [comment]
Would the typescript file still live somewhere in ``bokeh/bokehjs/src/coffee``? I have noticed the ``coffee`` directory already seems to contain a mix of ``.coffee`` and ``.ts`` files  (e.g in ``core/util``).

**Comment by mattpap** [comment]
@jlstevens, just change the extension. `src/coffee` isn't too accurate about the contents of that directory.

**Comment by jlstevens** [comment]
@mattpap I'm thinking about how to convert from coffeescript to typescript. 

Am I able to write typescript without the type annotations (so it is more like ES6) and then add the type annotations once the untyped code is working? That would offer a much easier migration path for anyone like me who lacks familiarity with typescript...

**Comment by jlstevens** [comment]
A bunch of unit tests in ``bokeh.models.tests.test_glyphs`` seem to be failing with:

```
AssertionError: Properties missing: visible
```

I can fix it with this diff:

```diff
modified   bokeh/models/tests/utils/property_utils.py
@@ -10,7 +10,7 @@ TEXT = ["text_font", "text_font_size", "text_font_style", "text_color", "text_al
 ANGLE = ["angle", "angle_units"]
 
 PROPS = ["name", "tags", "js_callbacks", "js_event_callbacks", "subscribed_events"]
-GLYPH = ["visible"]
+GLYPH = []
 
 MARKER = ["x", "y", "size", "angle", "angle_units"]
```

Is this something I should do or did I mess up something with the merge commit?

mattpap:
> Is this something I should do or did I mess up something with the merge commit?

`Glyph.visible` was removed in the PR merged earlier, so you have to fix this.

bryevdv:
@jlstevens fairly sure that is the right diff (ping @mattpap) it occurs to me that that should have gotten a migration note in the other PR and I think it was overlooked.

mattpap:
@jlstevens, currently TS errors aren't fatal and code is generated anyway, so you can commit whatever works and then we can fix it gradually (in this PR).

bokehjs/src/coffee/document.coffee
```diff
@@ -12,6 +12,25 @@ import {isEqual} from "./core/util/eq"
...
+    # Send message to Python via session
+    @session.send_event(event)
+
+  trigger: (event) ->
+    for model_id in @subscribed_models.values
+        if event.model_id != null and event.model_id != model_id
+            continue
+        model = @document._all_models[model_id]
+        if model?
+            model._process_event(event)
```
mattpap: This code is improperly indented (4 vs 2).

bokehjs/src/coffee/core/bokeh_events.coffee
```diff
@@ -0,0 +1,170 @@
...
+    return @
+
+
+export class ButtonClick extends Event
+
+  @event_name = 'button_click'
+  @applicable_models = ['Button']
+
+  constructor: (options) ->
+    super(options)
```
mattpap: Is there a reason for redefining constructor in each of these classes?

bokehjs/src/coffee/core/bokeh_events.coffee
```diff
@@ -0,0 +1,170 @@
+# Granular event classes for use with the event system
+import {logger} from "./logging"
+
+
+export class Event
+
+  @_event_classes = {}
+
+  constructor : (options) ->
```
mattpap: You could use `options={}` so that you don't have to `cls({})` every time you construct an event.

jlstevens: Good plan.

bokehjs/src/coffee/document.coffee
jlstevens: Thanks, though I am now converting to typescript which means this is all being rewritten.

bokehjs/src/coffee/core/bokeh_events.coffee
jlstevens: You're right - the constructor here isn't necessary.

jlstevens:
As a first step towards typescript, I've converted ``bokeh_events.coffee`` to ES6 in ``bokeh_events.ts``. Everything still seems to be working, although I had to add a ``register_event_class`` method. A corresponding method already existed in Python but in not coffeescript but now I seem to need it to register the event classes at the bottom of the file (instead of as part of the class definitions).

I have to say converting to ES6 has made me appreciate how much more pleasant it is to read and write coffeescript!

jlstevens:
@mattpap I might need some help/pointers regarding the type annotations.

Does the type information have to be in the file itself or can it be in a separate ``.d.ts`` file?

mattpap:
> Does the type information have to be in the file itself or can it be in a separate .d.ts file?

If you write a `*.ts` file, then it contains type information. If you have a `*.js` file (or `*.coffee -> *.js`), then you write a `*.d.ts`.

jlstevens:
>  If you have a *.js file (or *.coffee -> *.js), then you write a *.d.ts.

Does this mean I could have kept the file as coffeescript and added a ``*.d.ts`` file? If so, that does seem preferable to me...

mattpap:
>  I had to add a register_event_class method.

ES/TS classes don't have static initializers. For now your solution is fine, later we could use decorators.

mattpap:
No, I mean that you have to fill in types in `core/bokeh_events.ts`. However, there are some dependencies on other module, like `core/logger`, which currently don't have typings (aren't `*.ts` files, nor have associated `*.d.ts` files), so you will have to explicitly cast to `any` type in a few places, to bypass strict type checking.

jlstevens:
Thanks for clarifying! 

If you could give me an example of a type signature or two in ``bokeh_events.ts``, I think that would help me get started.

bokehjs/src/coffee/core/bokeh_events.ts
```diff
@@ -0,0 +1,200 @@
...
+    static _event_classes = {}
+
+    constructor(options = {}) {
+        this._options = options;
+        if (options.model_id) {
+            this.model_id = null
+        }
+    }
+
+    set_model_id(id) {
```
mattpap: `set_model_id(id: string): this`

bokehjs/src/coffee/core/bokeh_events.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export class Event {
+    static _event_classes = {}
+
+    constructor(options = {}) {
+        this._options = options;
+        if (options.model_id) {
+            this.model_id = null
+        }
+    }
```
mattpap: I don't follow this. You don't want to set `model_id` when it's available?

bokehjs/src/coffee/core/bokeh_events.ts
mattpap: And if unsure or there is no good type available, you can just use `any`. There are a lot of errors regarding implicit `any`, so if you make this explicit, they will go away. We don't want to have any `any` types floating around, but at this point we can allow for them.

mattpap:
I need to give a though about how to type `options`. I will let you know tomorrow.

mattpap:
btw. Indentation is wrong in `core/bokeh_events.ts`.

bokehjs/src/coffee/core/bokeh_events.ts
jlstevens: Well spotted - I messed up the conversion here. The code was working because when the ``model_id`` is null, it is broadcast to all the models.

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+    # example setting up python callback
+
+'''
+from __future__ import absolute_import
+
+class Event(object):
+    ''' Base class for all Bokeh events.
+
+    '''
+    event_classes = []
```
bryevdv: This should be private

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+    event_name = None
```
bryevdv: not sure this is necessary

bokehjs/src/coffee/core/bokeh_events.ts
jlstevens: Fixed in d49e10a

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+        print('Warning: Could not find appropriate Event class')
+
+
+class ButtonClick(Event):
+    ''' Announce a button click event.
+
+    '''
+    event_name = 'button_click'
+
+    def __init__(self, model_id=None):
```
bryevdv: I was thinking it might be better for event initializers to accept models after all, but to pick of the model id from the model and use that. The reason for that is that some events only make sense with certain kinds of models. E.g. it makes no sense to register a `ButtonClick` on a plot. So it would be more user-friendly to detect bad combinations and raise meaningful errors when they happen. Offhand, that seems easiest to do here in the initializer. So maybe:
```
def __init__(self, model):
    if not isintance(model, Button): 
        raise ValueError(...)
    self.model_id = model.id
```

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+
+class Pan(PointEvent):
+    '''
+
+    '''
+    event_name = 'pan'
+
+    def __init__(self, deltaX=None, deltaY=None, direction=None, **kwargs):
+        self.deltaX = deltaX
+        self.deltaY = deltaY
```
bryevdv: Snake case *alway* for attrs/vars on the python side. If something needs to match JS, and it conventions conflict, then the JS convention is the one that yields.

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+        '''
+        cls.event_classes.append(event_cls)
+
+    @classmethod
+    def from_JSON(cls, json):
+        eventclasses = [eventcls for eventcls in cls.event_classes]
+        for eventscls in eventclasses:
+            if eventscls.event_name == json['event_name']:
+                return eventscls(**json['event_values'])
+        print('Warning: Could not find appropriate Event class')
```
bryevdv: Instead of this, we should define whatever hook we need to define in order to plug into the standard `json` module so that the usual `json.loads(json_str)` works

jlstevens:
> btw. Indentation is wrong in core/bokeh_events.ts.

Should be fixed now in 3065905

bokeh/events.py
```diff
@@ -0,0 +1,178 @@
...
+
+    '''
+    event_classes = []
+    event_name = None
+
+    def __init__(self, model_id=None):
+        self.model_id = model_id
+
+    @classmethod
+    def register_event_class(cls, event_cls):
```
bryevdv: Still not terribly fond of this registry.  Will have to think on it. 

In any case, if this is a class method, it should just be `register` and only take one arg, then e.g.
```
MyCustomEvent.register()
```
will work (`cls` will be `MyCustomEvent`)

Of course we could automate that with a metaclass or possibly even just a class decorator on the`Event` base class.... It's just a little odd having models take care of all this automagically, but having events be much more explicit.

bokeh/events.py
jlstevens: Fixed in 25cc341

bokeh/events.py
jlstevens: Fixed in 29d4e5a

bokeh/events.py
```diff
@@ -88,6 +87,13 @@ def __init__(self, sx=None,sy=None, x=None, y=None, model_id=None):
 
 # --- Point Events ------------------------------------------------------------
 
+class Tap(PointEvent):
+    ''' Announce a tap or click event on a Bokeh plot.
+
+    '''
+    event_name = 'tap'
+
+
```
bryevdv: oops! (sorry)

bokeh/events.py
jlstevens: Seems reasonable although it seems I will need to inline the import to avoid cyclic import issues. It would be...

```python
def __init__(self, model):
    from .models.widgets import Button
    if not isinstance(model, Button): 
        raise ValueError(...)
    self.model_id = model.id
```

I would also consider introducing a ``PlotEvent`` base class so this validation would only need to be done once for everything that is associated with ``Plot``.

bokeh/events.py
jlstevens: I noticed ``Tap`` had gone missing when I tried running my example app. :-)

bokeh/events.py
jlstevens: I'm not too fond of it either.  A decorator sounds reasonable but I would need to think a bit before implementing it - I don't yet see how a single decorator would capture all event class definitions. Wouldn't you need to apply the decorator on every event class definition?

A metaclass would do it though....

bokeh/events.py
jlstevens: Here would you prefer ``deltax`` or ``delta_x``? Personally, I would lean towards the former...

bokeh/events.py
bryevdv: Right I am not sure about the class decorator either, it just seemed like it might be *possible* definitely could go the metaclass route, though. 

Another option, a bit more drastic, would be to make events themselves be models. That would require some consideration, though.

bokeh/events.py
bryevdv: `delta_x`

bokeh/events.py
bryevdv: internal import for this is fine. And interposing a class in the hierarchy to de-dupe code is good too.

bryevdv: The other option is to make it declarative and generic, i.e. classes get configured with  list of acceptable model types and a generic chunk of code in `Event` checks them

bokeh/events.py
jlstevens: Or we could add the current class to ``_event_classes`` (declared as a set) in ``Event.__init__`` (assuming it always gets called). It would be trivial to implement and I think that would be better than switching to models. It does mean ``super`` has to be called appropriately...

jlstevens: I've implemented a solution in 4ddcdc0 by introducing a simple metaclass. Note that I re-introduced ``Event.event_name`` as ``None`` as it simplifies the logic in the ``from_JSON`` method - it lets me guarantee that all events have an ``event_name`` I can check against.

bokeh/events.py
jlstevens: I thought of that, but inline imports need to be deferred - if you try to list acceptable model types at the class level, you need to import them first which runs into trouble.

bokeh/events.py
jlstevens: Fixed in 45990c4

bokeh/events.py
jlstevens: One other thing, should we still allow the model to be ``None`` to allow 'broadcast' events that aren't associated with a particular model?

bokeh/events.py
jlstevens: I think that would involve setting the [``object_hook`` argument](https://docs.python.org/2/library/json.html#json.load) when using ``json.load`` or ``json.loads``. Either that or making a custom decoder which would need to be supplied as an argument to ``load``/``loads`` anyway. 

You would then need to use something like this in ``Document.apply_json_event``:

```python
json.loads(json_str, object_hook = Event.decode_events)
```

Seems reasonable.

jlstevens: I would also add a check to make sure the return value is an ``Event`` because the above line can deserialize any JSON and in ``apply_json_event`` it should definitely be an event object.

jlstevens: Implemented in 4607de4.

volkerjaenisch:
I have a question to this very important improvement.
_The following is nether intended as a rant nor as a demand._
We have implemented for our Bokeh application a separate event protocol utilizing AJAX calls for the communication between JS-Events and Python-Workloads. We have done this because we were unhappy with the current event handling capabilities of bokeh.
The main problem we found in the current implementation is that events initiated by the JS cannot be acknowledged by the python handlers. So if for instance the user changes an axis scaling (which in our case fetches new data from a web based data source) the sending instance of the event (the axis) cannot determine if the datasource has completed the sending of new data (or refused to send at all, or ran in a timeout) and react appropriately.
Also the receiving instance (Python Handler) receive lots of intermediate events since there in no active debouncing on the JS side and debouncing on the receiver side is due to network latency a fruitless idea.
From my (surely limited) understanding the current event handling code is like "fire and forget". The event triggering is communicated solely over changes in model properties and not via some kind of "RPC" like for instance CORBA does.
**My question is if and how acknowledgement of events is planned to be implemented with this PR.**
I would love to help implementing such a thing. Surely there are lots of problems like blocking vs. not blocking, or timeouts, etc to be addressed to realize such a feedback. Maybe generally such a feedback mechanism may interfere negatively with other components or core concepts of Bokeh, I am not aware of. Please enlight me.

Cheers,
Volker

bryevdv:
> From my (surely limited) understanding the current event handling code is like "fire and forget". 

This is correct

> The event triggering is communicated solely over changes in model properties

But this is not. This PR adds a new websocket protocol message for events that are not tied to any model properties. Previous to this PR, model properties were (ab)used to hackily simulate events like button clicks, but this PR provides a clean mechanism to report those events to that fake properties are not needed. 

> My question is if and how acknowledgement of events is planned to be implemented with this PR.

I'd say it's out of scope for this PR. This PR is motivated by concrete and pressing uses cases going from other projects like Holoviews, Dask, and Datashader. It's important that this go in the next release which means it needs to be merged next week basically. There is not time for significant change at this point. 

A second part of this work, which did not make it in this PR and may also not make it to the next release, it to allow events to flow the other direction, i.e. for events to generated on the server so that `CustomJS` callbacks could respond to them. My particular interest is in `BUSY` and `DONE` events that can be triggered at the start and end of long running work (e.g. a button kicks of a scikit learn computation, and you want the UI to put up some kind os spinner or other indication that work is happening). Perhaps once events are bi-diretional, combined with the fact that the events system can be extended with user defined events, that will be enough for you to "roll your own" notifications in the short term? 

Regarding further works in the core library, you've touched on several possible avenues of development debouncing and handling backpressure better. What would help move these forward are very detailed use-cases to study, and detailed proposals for behavior and implementation. As I said the use case for the event system in this PR came from concrete needs of other projects. There's basically nothing more helpful, especially for foundational changes.I will just say that there is not enough context for me to appreciate your use case or why you made the decision to use a separate AJAX protocol, from just the brief description above. I'd encourage you to start a new separate issue to discuss and consider further improvements in much greater detail.

jlstevens:
@volkerjaenisch 

I think this is a question for @bryevdv who has been closely involved with this PR and who has a better idea of possible future directions than I do. I can say a few things though...

> From my (surely limited) understanding the current event handling code is like "fire and forget". 

At this time, this is correct. I don't see any technical reasons why debouncing events couldn't be implemented on the client side if necessary.

> The event triggering is communicated solely over changes in model properties

The actual communication of events from JS to Python uses an extension to the message protocol (using a new EVENT message type) which means it isn't relying exclusively on model property changes to trigger events.

> My question is if and how acknowledgement of events is planned to be implemented with this PR.

Currently only JS->Python communication of events is implemented, so I don't think this feature will arrive in this particular PR. I do have a rough prototype of event communication working in the opposite direction and this will hopefully be implemented in a follow up PR soon. Supporting a bidirectional flow of events was always a goal of this event architecture.

One of the concrete aims we have for Python->JS events is to send 'busy' events to JS from Python so that the visualization can respond appropriately when the server is busy. Once bidirectional communication of events is possible, I don't why you couldn't try to send events one way and acknowledgement events in the other. @bryevdv Do you think such a feature would be a good idea and a plausible extension of the current event architecture?

bryevdv:
> Do you think such a feature would be a good idea and a plausible extension of the current event architecture?

Possibly, it's worth discussing for later. Again, with lots of details, example proposed code and use cases, etc. AFAICT @volkerjaenisch is asking for something like a real RPC mechanism. That's not trivial, but it's entirely possible that a simple one could be built as an abstraction on top of a bi-diretional event system (that this PR is half of). So I'd say the best plan is to: 

* merge this PR
* get a PR for the "other direction"
* discuss and propose ideas for abstractions built on top, i.e. RPC

Or in other words, I think this work supports, rather than precludes, a more "RPC" like mechanism that could be possibly added in the future.

volkerjaenisch:
>     The event triggering is communicated solely over changes in model properties
> 
> But this is not. 

Sorry! My remarks on the event protocol were based on the vanilla Bokeh code without this PR. I should have stated this more clearly. 
I have noticed the new "Event" token for the websocket protocol of this PR already, and it is surely a step in the right direction - but there could be so much more :-)

My team has yet only invested a month or so in bringing up a prototype of a typical scientific task that is comparable to Bokeh + Datashader + Interaction, where the last part is not trivial. We want to build scientific (generic and concrete) visualizations for our customers that all share a common goal: Working on very large datasets fluently in the browser. Bokeh is our first choice (and there is no second) since it is pythonic and it has a astonishing low barrier for newcomers. 

I have build lots of interacting systems in the recent four decades, and a well defined communication layer was the most crucial thing in each of them. Second Problem was the performance of data transfer. Looking at Bokeh closely I dicovered it is using websockets, the most advanced HTTP protocol yet, but Bokeh only implements in its higher level protocol-layer a hand full of protocol tokens, only. I can assume this results from the evolution of Bokeh. Probably started with AJAX and then migrated to websockets. No shame in this - software evolves, period.

But it would be so cool if the websockets layer with the already existing synchronized model layer would be open for custom extension protocols. But as @bryevdv states my suggestions are out of scope of this PR. So I will open a new topic on Bokeh-dev to discuss them further.

Volker

bryevdv:
@volkerjaenisch thank you very much for your kind words. Also all the things you describe are absolutely in line with what we want for people to be able to do with Bokeh (basically: do more with less code) Extensibility is also a very important priority to continue to develop. I think your input and use cases would be extremely valuable for helping guide development in these areas, I definitely look forward to discussing this further.

volkerjaenisch:
@jlstevens

>     My question is if and how acknowledgement of events is planned to be implemented with this PR.
> 
> Currently only JS->Python communication of events is implemented, so I don't think this feature will arrive in this particular PR. I do have a rough prototype of event communication working in the opposite direction and this will hopefully be implemented in a follow up PR soon. Supporting a bidirectional flow of events was always a goal of this event architecture.

On "bidirectional event communication":
The usecases discussed in this thread on this topic address something like Python has a simulation running and likes to inform the JS by invoking a JS event. That will for sure be helpful.

But it cannot be a basis for a feedback mechanism like I supposed. This is because Events in Bokeh will still be "fire and forget" after merging this PR.

One thing Bokeh does not address yet (IMHO) is synchronization in "time". Where "time" stands for  session, or interaction, or transaction or what ever you like to call them. I have seen some approaches on this topic using Session-IDs in the code but not for JS-to-Python and back. It would be important to have such a Session-ID/Event-ID available on both sides.

Volker

jlstevens:
I've now added ``examples/howto/js_events.py`` to demo client-side events that currently looks like this.

image

This howto example works as a static HTML file meaning it could also be integrated into a useful reference page in the docs. In addition, it should help us prompt people  for ideas regarding a new bokeh app example to demonstrate this system with a more involved example.

Once I've added the corresponding version for Python events (i.e an app in ``examples/howto/python_events.py``), I think that will suffice as a concrete events demo for this PR. Building a decent, meaningful app to really showcase events will take a fair bit of thought and time.

jlstevens:
I have now added ``examples/howto/events_app.py`` which is identical to the above but 1) also prints the events to the Python console 2) needs to be run as a bokeh app.

I've also added a ``pprint`` method to ``Event`` so you can easily see the event information you are interested in (e.g the model id may not always be helpful). I wanted to use this to implement ``__str__`` and ``__repr__`` methods but then I would need to list the event attributes when calling ``pprint``.

I'm starting to think that it is worth adding an ``event_attributes`` class attribute to the event classes. This would be a bit redundant with the names in the event constructors but it would allow me to implement ``__str__`` and ``__repr__`` once. In addition, such a declaration would let me simplify the callbacks in the howto demos.

bokeh/events.py
jlstevens: @bryevdv I'm happy to implement this but I do have three questions:

1. Should we allow ``model=None`` to allow 'broadcast' events?
2. Would you want this validation mirrored on the JS side or can I leave the use of ``model_id`` alone there?
3. On the JS side, there is the ``  static applicable_models = ['Button']`` attribute which accepts a list of strings. If I mirrored this on the Python side then a) I could avoid inline imports by checking against the class name b) define events that apply to more than one type of model. It would then be easier to add the same validation approach to the JS side.

jlstevens:
@mattpap I'm afraid I'm still going to need help to type ``bokeh_events.ts``.

Even after sprinkling ``: any`` all over the file, I still had a ton of warnings/type errors and doing that would defeat the point of trying to specifying types in the first place.

bokehjs/src/coffee/core/bokeh_events.ts
```diff
@@ -0,0 +1,217 @@
+import { logger } from "./logging"
+
+export class Event {
```
mattpap: This needs to be renamed to `BokehEvent` (or something), so that it doesn't conflict with DOM `Event`.

jlstevens: sure, I'll do that now.

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -177,13 +180,24 @@ export class UIEvents
       sx: x - left
       sy: y - top
     }
+    e.bokeh = extend(e.bokeh, extras)
+    event_cls = PointEvent.event_class(e)
```
mattpap: Why `PointEvent`? I presume this should be `Event` (or `BokehEvent` after changes). The same applies below.

jlstevens: Sure, that also works. I'll put those two changes together.

bokehjs/src/coffee/core/bokeh_events.ts
jlstevens: Fixed in 0e7813b

bokehjs/src/coffee/core/ui_events.coffee
jlstevens: Fixed in ed942a2

bokehjs/src/coffee/core/bokeh_events.ts
```diff
@@ -0,0 +1,217 @@
...
+
+export class PinchStart extends PointEvent {
+  static event_name = 'pinchstart'
+}
+
+export class PinchEnd extends PointEvent {
+  static event_name = 'pinchend'
+}
+
+Event.register_event_class('lodstart', ButtonClick);
```
mattpap: Every event type has `event_name`, yet `register_event_class()` still need `name` argument.

bokehjs/src/coffee/model.coffee
```diff
@@ -17,6 +21,42 @@ export class Model extends HasProps
...
 
+    @listenTo(@, 'change:js_event_callbacks', () -> @_update_event_callbacks)
+
+
+  _process_event : (event) ->
+    # Given an applicable event, execute the associated callback
+    if not event.constructor.event_name
+       logger.warn('Event is still passed as a string: ' + event)
+       return
+    if not event.constructor.applicable_models.some((m) => m == @.type)
```
mattpap: This invalidates the point of inheritance of models. I know we have still this have kind of code in bokehjs, but new code shouldn't go this path.

bokehjs/src/coffee/core/bokeh_events.ts
jlstevens: Good point. Fixed in 08a1e3c

bokehjs/src/coffee/model.coffee
jlstevens: Events aren't Models. This is something I discussed with @bryevdv and we didn't find any strong reason to make them into models. In addition, ``applicable_models`` can be a list so one event may have multiple applicable models and I'm not entirely sure the appropriate specification would always follow the structure of the inheritance hierarchy.

mattpap: I'm speaking about `m == @.type`. `m` and `@` are models, so this should use `instanceof` check.

mattpap: Well, `m` is a string type representing a model. I know it's easier this way, but for the purpose of being able to reason about the code, this and anything else that uses `@type`, must respect model hierarchy. Otherwise someone will be badly surprised with this behavior.

jlstevens:
@mattpap Thanks! It now compiles without lots of warnings/errors.

I tried many of the things I see in your diff and I suspect the one thing I didn't try that was the cause of many of the warnings was declaring ``_options: any``.

bokehjs/src/coffee/document.coffee
jlstevens: Now addressed - file is now converted to typescript with the correct indentation.

bokehjs/src/coffee/model.coffee
jlstevens: If I import the various models into ``bokeh_events.ts`` then I could set the actual model classes in ``applicable_models``. Then I could use an isinstance check here. I'm happy to do that if I am understand it right.

jlstevens: @mattpap I had a go at this but didn't have any luck. I don't get why the ``instanceof`` check never passes in the following diff:

```diff
modified   bokehjs/src/coffee/core/bokeh_events.ts
@@ -1,4 +1,6 @@
 import { logger } from "./logging"
+import {Button} from "../models/widgets"
+import {Plot} from "../models/plots/plot"
 
 export class BokehEvent {
   static _event_classes: any = {}
@@ -54,14 +56,14 @@ export class BokehEvent {
 
 export class ButtonClick extends BokehEvent {
   static event_name = 'button_click'
-  static applicable_models = ['Button']
+  static applicable_models = [Button]
 }
 
 
 export class UIEvent extends BokehEvent {
   // A UIEvent is an event originating on a PlotCanvas this includes
   // DOM events such as keystrokes as well as hammer events and LOD events.
-  static applicable_models = ['Plot']
+  static applicable_models = [Plot]
 }
 
 
modified   bokehjs/src/coffee/model.coffee
@@ -29,7 +29,8 @@ export class Model extends HasProps
     if not event.constructor.event_name
        logger.warn('Event is still passed as a string: ' + event)
        return
-    if not event.constructor.applicable_models.some((m) => m == @.type)
+
+    if not event.constructor.applicable_models.some((m) => m instanceof @.constructor)
       return
 
     event = event._customize_event(@)
```

mattpap: You should reverse this check. You want `this` to be an instance of any of applicable classes.

jlstevens: Thanks, I got myself confused but it is now fixed in 2c7781e using the reversed check as you suggested.

jlstevens:
@mattpap Any idea why the ``flake_docs`` tests are failing? There seem to be connection timeout errors such as:

```
requests.packages.urllib3.exceptions.NewConnectionError: 
<requests.packages.urllib3.connection.HTTPConnection object at 0x2ad4f49c0438>: Failed to
 establish a new connection: [Errno 110] Connection timed out
```

bryevdv:
> Any idea why the flake_docs tests are failing? There seem to be connection timeout errors such as:

`docs.scipy.org` is down and our docs use intersphinx which relies on it being up

bokeh/events.py
jlstevens: Addressed in 88f4ec5 and 5deb786

jlstevens:
@bryevdv @mattpap 

I think I have now implemented all the code related changes and ticked off everything from my TODO list other than improving the docstrings and adding suitable tests.

Now would be a good time to give me any more comments you may have regarding the implementation before I start setting up the tests and adding the docstrings.

jlstevens:
@bryevdv 

I was just wondering if the ``_trigger_event`` method on ``Model`` (inherited from the ``EventCallbackManager`` mix-in) should be made public (i.e called ``trigger_event`` instead)?

This is something I noticed while writing the tests - I don't see why you shouldn't want to trigger event callbacks on ``Models`` yourself if you want to, especially now there is additional validation when constructing events...

jlstevens:
I've been running my new unit tests locally to make sure they pass which is why I have only just noticed that the unit tests are now failing on travis.

It seems they started failing from [this build](https://travis-ci.org/bokeh/bokeh/builds/210631987) but I don't see how the failures could relate to the corresponding commit (2c7781e). Any ideas?

mattpap:
@jlstevens, this is because `Button` is being imported in `core/bokeh_events`, which makes our build pull widget stuff into `bokeh.js`, which in turn make tests fail, because now `bokeh.min.js` is at about 1 MB instead of ~700 kB.

Models shouldn't be referenced from the core. Registration of applicable models should be moved to models that need this. This is a change I wanted to ask for anyway, independently of test failures.

jlstevens:
I have now fixed the tests and done the refactor suggested by @mattpap . The reason the flake tests failed seems to be due to intersphinx issues only.

jlstevens:
I have now added unit tests of the event classes themselves, test to make sure event callbacks are triggered correctly and tests for the ``EventCallbackManager``.

bokeh/util/tests/test_callback_manager.py
jlstevens: Addressed in 8d6ac34

jlstevens:
I have now filled in all the event class docstrings and added code examples to the module docstring. That completes all the tasks I assigned myself for this PR!

jlstevens:
@mattpap 

You may have already noticed this in the howto example:

image

The good news is that the data seems to be there but ``cb_obj.event.constructor.event_name`` which used to supply the event name is now ``undefined``.

jlstevens:
Nevermind, ``cb_obj.event.event_name`` works which means I can just update the howto example.

mattpap:
@jakirkham, yeah, forgot to update the other example.

jakirkham:
Guessing you didn't mean me. 😉

mattpap:
@jakirkham, yes, unfortunately when I type `@j<TAB>`, github auto-completes your account's name instead of the author's of this PR.

jakirkham:
No worries. Just wanted to make sure I wasn't missing something.

bryevdv:
Thanks @jlstevens

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -238,24 +252,22 @@ export class UIEvents
     @_trigger('rotate:end', e)
 
   _mouse_enter: (e) ->
-    @_bokify_jq(e)
-    @_trigger('move:enter', e)
+    @_bokify_point_event(e)
+    @trigger('move:enter', e)
```
mattpap: @jlstevens, @bryevdv, why `@_trigger()` was changed back to `@trigger()`?

jlstevens: This looks like an unintentional change. As far as I know it should still be ``@_trigger()``.

mattpap: @jlstevens, OK. It used to be actually `@trigger()`. I changed it to `@_trigger()` in PR #5349, to gain full control over event flow. Now seeing this change to original, I was wondering what will break when I change it to what I need.

jlstevens: I'm fairly sure that everything this PR needs is in ``@_bokify_point_event(e)`` and I don't think switching between ``@trigger`` and ``@_trigger`` should have any effect on how events are processed.

PR #5942, opened 2017-03-01
Creates tempfile when output_filename is not writable (notebook)

### LINKED ISSUE #5677: Permission Denied error when invoking show() from inside Jupyter Notebook
Hi,

I'm using bokeh 0.12.3 and jupyter 1.0.0. I'm writing some code that produces a simple bar chart and calls show() on this bar chart inside the notebook.

Here's the backtrace that I get:
```
PermissionError                           Traceback (most recent call last)
<ipython-input-3-e94210215dfe> in <module>()
      1 results = r._get_results()
----> 2 v.visualize(results, None, None)

/home/lukas/src/tcpsp/TCPSPsuite/scripts/visualizer.py in visualize(self, results, x_axis, y_axis)
     11         chart = Bar(df, 'job_count', values='elapsed', agg='mean')
     12 
---> 13         show(chart)

/home/lukas/.virtualenvs/tcpsp/lib/python3.5/site-packages/bokeh/io.py in show(obj, browser, new, notebook_handle)
    312     if obj not in _state.document.roots:
    313         _state.document.add_root(obj)
--> 314     return _show_with_state(obj, _state, browser, new, notebook_handle=notebook_handle)
    315 
    316 

/home/lukas/.virtualenvs/tcpsp/lib/python3.5/site-packages/bokeh/io.py in _show_with_state(obj, state, browser, new, notebook_handle)
    330 
    331     if state.file or not shown:
--> 332         _show_file_with_state(obj, state, new, controller)
    333 
    334     return comms_handle

/home/lukas/.virtualenvs/tcpsp/lib/python3.5/site-packages/bokeh/io.py in _show_file_with_state(obj, state, new, controller)
    335 
    336 def _show_file_with_state(obj, state, new, controller):
--> 337     filename = save(obj, state=state)
    338     controller.open("file://" + filename, new=_new_param[new])
    339 

/home/lukas/.virtualenvs/tcpsp/lib/python3.5/site-packages/bokeh/io.py in save(obj, filename, resources, title, state, validate)
    394 
    395     filename, resources, title = _get_save_args(state, filename, resources, title)
--> 396     _save_helper(obj, filename, resources, title, validate)
    397     return os.path.abspath(filename)
    398 

/home/lukas/.virtualenvs/tcpsp/lib/python3.5/site-packages/bokeh/io.py in _save_helper(obj, filename, resources, title, validate)
    462         html = standalone_html_page_for_models(obj, resources, title)
    463 
--> 464         with io.open(filename, "w", encoding="utf-8") as f:
    465             f.write(decode_utf8(html))
    466 

PermissionError: [Errno 13] Permission denied: '/usr/lib/python3.5/runpy.html'
```

It seems like I'm not the only one experiencing the problem: see for example https://github.com/yhat/rodeo/issues/211

I'm pretty sure the problem is in _detect_filename() in io.py: If no filename to save a chart to is supplied, it tries to save the file (with an autodetected filename) into the directory of the currently executing file (more exactly: the file executing in the topmost stackframe, which is assumed to be the main file being run.) This fails spectacularly if that file is located in a read-only location as is the case when you run an jupyter notebook, which on my system seems to use "/usr/lib/python3.5/runpy.py" to run the notebook.

I think the cleanest solution would be to use a temporary directory if no save-to directory is specified. If no directory is specified, the user is probably not really interested in the file, but only needs it temporarily: This happens in the show() method, for example.

Is there a reason for saving temporary files in the directory of the script? In that case, should the save() method be amended by a "temporary" argument (with default False), which in turn tells _detect_filename() to use a temporary directory? In that case, show() could set that flag and avoid the permission problem…?

I'd be willing to cook up a pull request if I get any feedback on how you think this problem should be solved.

jsignell:
- [x] issues: fixes #5677
- [ ] tests added / passed

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -22,6 +22,7 @@ import io
 import json
 import os
 import warnings
+import tempfile
 
 # Third-party imports
 
@@ -413,6 +414,8 @@ def _get_save_args(state, filename, resources, title):
     if filename is None:
         warn = False
         filename = _detect_filename("html")
+        if not os.access(filename, os.W_OK):
+            filename = tempfile.NamedTemporaryFile().name
 
     if filename is None:
         raise RuntimeError("save() called but no filename was supplied or detected, and output_file(...) was never called, nothing saved")
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -394,14 +394,21 @@ def _detect_filename(ext):
     None if the script could not be found (e.g. interactive mode).
     """
     import inspect
-    from os.path import isfile, dirname, basename, splitext, join
+    from os.path import isfile, dirname, basename, splitext, join, curdir
 
     frame = inspect.currentframe()
     while frame.f_back and frame.f_globals.get('name') != '__main__':
         frame = frame.f_back
 
     filename = frame.f_globals.get('__file__')
-    if filename and isfile(filename):
+
+    if filename is None:
+        return tempfile.NamedTemporaryFile().name
+
+    elif not os.access(dirname(filename) or curdir, os.W_OK | os.X_OK):
+        return tempfile.NamedTemporaryFile().name
+
+    elif isfile(filename):
         name, _ = splitext(basename(filename))
         return join(dirname(filename), name + "." + ext)
 
@@ -414,8 +421,6 @@ def _get_save_args(state, filename, resources, title):
     if filename is None:
         warn = False
         filename = _detect_filename("html")
-        if not os.access(filename, os.W_OK):
-            filename = tempfile.NamedTemporaryFile().name
 
     if filename is None:
         raise RuntimeError("save() called but no filename was supplied or detected, and output_file(...) was never called, nothing saved")
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -403,14 +403,8 @@ def _detect_filename(ext):
 
     filename = frame.f_globals.get('__file__')
 
-    if filename is None:
-        return tempfile.NamedTemporaryFile().name
-
-    # dirname(filename) returns empty str if file called from within directory
-    directory = dirname(filename) or curdir
-
-    if not os.access(directory, os.W_OK | os.X_OK):
-        return tempfile.NamedTemporaryFile().name
+    if filename is None or not os.access(dirname(filename) or curdir, os.W_OK | os.X_OK):
+        return tempfile.NamedTemporaryFile(suffix="." + ext).name
 
     name, _ = splitext(basename(filename))
     return join(dirname(filename), name + "." + ext)
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -391,10 +391,11 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
 
 def _detect_filename(ext):
     """ Detect filename from the name of the script being run. Returns
-    None if the script could not be found (e.g. interactive mode).
+    temporary file if the script could not be found or the location of the
+    script does not have write permission (e.g. interactive mode).
     """
     import inspect
-    from os.path import isfile, dirname, basename, splitext, join, curdir
+    from os.path import dirname, basename, splitext, join, curdir
 
     frame = inspect.currentframe()
     while frame.f_back and frame.f_globals.get('name') != '__main__':
@@ -405,12 +406,14 @@ def _detect_filename(ext):
     if filename is None:
         return tempfile.NamedTemporaryFile().name
 
-    elif not os.access(dirname(filename) or curdir, os.W_OK | os.X_OK):
+    # dirname(filename) returns empty str if file called from within directory
+    directory = dirname(filename) or curdir
+
+    if not os.access(directory, os.W_OK | os.X_OK):
         return tempfile.NamedTemporaryFile().name
 
-    elif isfile(filename):
-        name, _ = splitext(basename(filename))
-        return join(dirname(filename), name + "." + ext)
+    name, _ = splitext(basename(filename))
+    return join(dirname(filename), name + "." + ext)
 
 def _get_save_args(state, filename, resources, title):
     warn = True
@@ -421,9 +424,6 @@ def _get_save_args(state, filename, resources, title):
     if filename is None:
         warn = False
         filename = _detect_filename("html")
-
-    if filename is None:
-        raise RuntimeError("save() called but no filename was supplied or detected, and output_file(...) was never called, nothing saved")
 
     if resources is None and state.file:
         resources = state.file['resources']
--- a/sphinx/source/docs/releases/0.12.5.rst
+++ b/sphinx/source/docs/releases/0.12.5.rst
@@ -118,3 +118,12 @@ Attempting to access any other attribute on ``request`` will result in an
 error.
 
 .. _project roadmap: http://bokehplots.com/pages/roadmap.html
+
+Default save file
+~~~~~~~~~~~~~~~~~
+
+If user-specified or default destination cannot be written to, a temporary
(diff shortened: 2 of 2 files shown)


bryevdv:
Several examples in python 2 seem to be reliably failing on Travis examples tests with e.g.
```
tests/examples/test_examples.py::test_file_examples[/home/travis/build/bokeh/bokeh/examples/plotting/file/bar_chart.py-example58] 
[INFO] Example run in 1.245s
[INFO] Example rendered in 0.352s
[JS] Error opening /home/travis/build/bokeh/bokeh/examples/plotting/file/bar_chart.html: No such file or directory
FAILED
```
I note that, at least for that bar chart example, the script does not call `output_file`:

https://github.com/bokeh/bokeh/blob/master/examples/plotting/file/bar_chart.py

I think that is a mistake (I think all examples should set a filename explicitly), but it also seems to show some issue with this change and python 2?

mattpap:
> it also seems to show some issue with this change and python 2

phantomjs part of examples testing is done only on 2.7.

bokeh/io.py
```diff
@@ -413,6 +414,8 @@ def _get_save_args(state, filename, resources, title):
     if filename is None:
         warn = False
         filename = _detect_filename("html")
+        if not os.access(filename, os.W_OK):
```
mattpap: `_detect_filename()` can return `None`, indicating that detection failed, so `os.access()` check should be moved there.

mattpap: `os.access("non_existent_filename", os.W_OK)` returns `False`, so you have to proceed with the solution proposed in the issue, i.e. check `W_OK | X_OK` for the base directory.

jsignell: Shouldn't we just be checking `W_OK` be on the base directory? Why should `X_OK` allow a pass?

bryevdv: agree with the suggestion to move to `_detect_filename`

mattpap: You need `X_OK` (executable bit) on a directory to make it "useful". Try `chmod -x on_some_directory` and then `touch on_some_directory/whatever` and see what happens.

bryevdv: @jsignell the syntax is a bit confusing, `W_OK | X_OK` is bitwise-or'ing two flags together. It's actually stating the directory has to be both writeable *and* "executable" (i.e. listable, for a dir) .

bokeh/io.py
```diff
@@ -393,14 +394,21 @@ def _detect_filename(ext):
...
         frame = frame.f_back
 
     filename = frame.f_globals.get('__file__')
-    if filename and isfile(filename):
+
+    if filename is None:
+        return tempfile.NamedTemporaryFile().name
+
+    elif not os.access(dirname(filename) or curdir, os.W_OK | os.X_OK):
+        return tempfile.NamedTemporaryFile().name
```
bryevdv: I think this usage of `or` might be surprising to many python programmers. But in any case, why the fallback check to `curdir`? The branch below only ever uses `dirname(filename)` so the accessibility of `curdir` does not seem relevant?

As a personal preference, I don't use `elif` if all the branches `return`

bokeh/io.py
```diff
@@ -393,14 +394,21 @@ def _detect_filename(ext):
...
+
+    elif isfile(filename):
         name, _ = splitext(basename(filename))
         return join(dirname(filename), name + "." + ext)
```
bryevdv: I wonder if it is also worth checking whether the final filename exists and is not writable. Maybe the standard exception that python will raise if trying to write to an unwritable file is actually the clearest thing, though.

bokeh/io.py
```diff
@@ -393,14 +394,21 @@ def _detect_filename(ext):
...
 
     frame = inspect.currentframe()
     while frame.f_back and frame.f_globals.get('name') != '__main__':
         frame = frame.f_back
 
     filename = frame.f_globals.get('__file__')
-    if filename and isfile(filename):
+
+    if filename is None:
+        return tempfile.NamedTemporaryFile().name
```
bryevdv: Can `NamedTemporaryFile` ever fail? If so maybe wrap in `try`/`except` and return `None` if it fails.

bryevdv: Otherwise, if we just want to let any exceptions propagate, the the lines that check for `None` from this function are no longer needed.

bokeh/io.py
```diff
@@ -393,14 +394,21 @@ def _detect_filename(ext):
     None if the script could not be found (e.g. interactive mode).
     """
```
bryevdv: This docstring can be updated/clarified.

bokeh/io.py
jsignell: If you are in the folder with the file and run the file from there then `dirname(filename)` evaluates to `""` which then fails the writeability check. But really what you want is to check the writeability of the current directory.

bokeh/io.py
```diff
@@ -390,19 +391,29 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
...
-        return join(dirname(filename), name + "." + ext)
+
+    if filename is None:
+        return tempfile.NamedTemporaryFile().name
+
+    # dirname(filename) returns empty str if file called from within directory
+    directory = dirname(filename) or curdir
+
+    if not os.access(directory, os.W_OK | os.X_OK):
+        return tempfile.NamedTemporaryFile().name
```
mattpap: This redundancy has to be removed. Also make sure that all code paths given a path with the same extension.

bokeh/io.py
```diff
@@ -390,19 +391,29 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
...
 
     filename = frame.f_globals.get('__file__')
-    if filename and isfile(filename):
-        name, _ = splitext(basename(filename))
-        return join(dirname(filename), name + "." + ext)
+
+    if filename is None:
+        return tempfile.NamedTemporaryFile().name
+
+    # dirname(filename) returns empty str if file called from within directory
```
mattpap: This shouldn't be unexpected outcome, thus the later `or curdir` doesn't seem necessary, unless you're worried that between `path = _detect_filename()` and using `path` current directory changes.

jsignell: I am just trying to check the wherever the file will get written to that dir is writeable. So if dirname(filename) returns empty string then we need to check that the current directory is writeable.

bokeh/io.py
jsignell: Which redundancy? The `tempfile.NamedTemporaryFile().name`? Having `filename is None or s.access(directory, os.W_OK | os.X_OK)` seemed less legible, but is that preferred?

bokeh/io.py
bryevdv: I can confirm that that this function returns `"foo.py"` if run e.g as `python foo.py` and `"somedir/foo.py"` if run as `python somedir/foo.py` in which in the former case the dirname ends up returning the empty string. I can also confirm the `os.access` does not interpret the empty string as the current dir:
```
In [9]: os.access(curdir, os.W_OK | os.X_OK)
Out[9]: True

In [10]: os.access("", os.W_OK | os.X_OK)
Out[10]: False
```
So I think testing against `curdir` makes sense.

bokeh/io.py
bryevdv: I'm OK with either but I do agree that the tmp file should have the requested `ext` in all cases

bryevdv:
This LGTM. It's probably worth adding a migration note about this (just trying to develop good fastidious habits around reporting things that change) This is the file:

https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/releases/0.12.5.rst

A new section and short sentence or two under "Migration Guide" Once that's there I think we can merge!

PR #5948, opened 2017-03-02
Custom 404 for docs

### LINKED ISSUE #4241: Create a custom 404 page for Bokeh Docs
As suggested [here](https://github.com/bokeh/bokeh/pull/4239#issuecomment-215233233) the JS warning banner for old/dev docs could try to redirect to users the same specific page, in the latest doc. But in cases where the specific page no longer exists, it would be better to have users land on something better than a default 404 page

Proposed to create a better designed and more informative 404 page for the docs site at http://bokeh.pydata.org 

Possible features:
- incorporate Bokeh logo
- provide link to current docs main page
- include the GCSE search box

jsignell:
- [x] issues: fixes #4241

This builds a reasonable looking 404.html, but the nav links are all relative.


jsignell:
screen shot 2017-03-02 at 2 37 35 pm

bryevdv:
Looks good, I will merge when green.

jsignell:
@bryevdv I don't know how this will work in the deploy environment, but running the docs locally, the relative links don't work very well because they are relative to the initial location of the 404.rst.

bryevdv:
We might just want to craft a custom HTML page after all. As a "quick fix" we could just take the HTML generated here and fix up the relative links and then throw away the rst source

jsignell:
http://bokeh.pydata.org/en/latest/slfjsg

bryevdv:
(location of static asset TBD in later issue/PR)

PR #5960, opened 2017-03-06
added visible checks to legend and box_annotation

### LINKED ISSUE #5488: hide annotations via callback
# Original question
I have a Bokeh plot where I add some data in the form of a `LabelSet` and `BoxAnnotation` as overlay, but I want to be able to dynamically enable/disable this overlay.

I can enable/hide some of the lines in the plot [already](http://stackoverflow.com/q/40592165/1562285), but the system for the Annotations seems to be different. I've got so far already

## Initialising

    from ipywidgets import interact
    from bokeh.plotting import figure as bf
    from bokeh.layouts import layout as bl
    from bokeh.models import Toggle, BoxAnnotation, CustomJS
    from bokeh.io import push_notebook, show, output_notebook
    output_notebook()

## Widget generation

    p = bf(title='test', x_range=(0,1), y_range=(0,1))
    x = [1/3, 2/3]
    y=[1/3, 2/3]
    p.circle(x=x, y=y, size=15)
    
    box = BoxAnnotation(left=None, right=0.5, fill_color='red', fill_alpha=0.1)
    
    p.add_layout(box)

## Interactivity

    code = '''\
    if toggle.active
        box.visible = true
        console.log 'enabling box'
    else
        box.visible = false
        console.log 'disabling box'
    '''
    callback = CustomJS.from_coffeescript(code=code, args={})
    toggle = Toggle(label="Red Box", button_type="success", callback=callback)
    callback.args = {'toggle': toggle, 'box': box}
    
    layout = bl([p], [toggle])
    show(layout)

When I check the JS console, the if/else clauses get triggered as expected, so the `Toggle` works but the red box stays in place, both in Firefox as in IE.;
When I try with a `Labelset` instead of a `BoxAnnotation` it doesn't hide either.

 ### Bokeh info
Python version      :  3.5.2 |Anaconda custom (64-bit)| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]
IPython version     :  5.1.0
Bokeh version       :  0.12.3
BokehJS static path :  C:\Anaconda3\lib\site-packages\bokeh\server\static

the notebook can be found at [My github repo](https://github.com/vadella/StackOverflow-questions/blob/master/bokeh_glyph.ipynb)

# Preliminary answer
This is a copy of [a StackOverflow question](http://stackoverflow.com/q/40849500/1562285) where @bryevdv [answered](http://stackoverflow.com/a/40852280/1562285)  there might be some plumbing that is not hooked up on the BokehJS side to respond to `visible` and asked me to repeat the question here.

As a workaround he suggested to use alpha settings for the time being, which works, both on the `LabelSet` as on the `BoxAnnotation`

jsignell:
- [x] issues: fixes #5488


mattpap:
@jsignell, you may want avoid duplication and do this once in the super class of both changed classes (`Annotation`) or perhaps even higher up in the hierarchy. Preferably you should add code handling `visible` property in the class (or related, because we have model-view separation) that defines it (so `Renderer`). `Renderer` doesn't implement this property and you can find `if @model.visible` all over bokehjs. What you have to do is to implement `RendererView.render()` and make sure that all overrides of `render()` in descendants of `RendererView` call `super()`. Also check if all views that should inherit from `RendererView` actually do this, because due to lack of static typing, inheritance is often broken in bokehjs.

bokehjs/src/coffee/models/annotations/legend.coffee
```diff
@@ -91,7 +91,7 @@ export class LegendView extends AnnotationView
   render: () ->
     if not @model.visible
       return
-  
+
```
mattpap: I highly recommend enabling trailing whitespace trimming in your code editor. You can also run `py.test -m quality` locally before pushing changes. This will spare you some work and leave travis ci's resources for more interesting changes.

bryevdv:
@mattpap I advised to handle separately, because some of the annotations render using DOM elements which may need to be actively hidden, as opposed to passively just "not drawing" on the canvas.

jsignell:
@bryevdv which annotations render using DOM elements?

bryevdv:
label, labelset, title, text_annotation *may*, tooltip always does

mattpap:
@jsignell, grep for `render_mode`. Doing this myself I realize that I directed you towards a mine field

jsignell:
Ok so leave it how it is for now?

mattpap:
@jsignell, after commit 6bfa2be annotations implementing `render_mode` are broken, so either you have to revert the last commit or proceed with the proper implementation (which isn't that hard, but requires custom handling of `visible` property when `render_mode == "css"` in `Span`, `BoxAnnotation` and `TextAnnotation`).

mattpap:
Actually, the first commit is also wrong because `BoxAnnotation` needs special handling, i.e.:
```
if not @model.visible and @model.render_mode == "css"
  hide(@el)
```
(which applies to other annotations listed in my earlier comment).

jsignell:
We can just always `hide(@el)` right?

    if not @model.visible
      hide(@el)
      return

mattpap:
> We can just always `hide(@el)` right?

No, this doesn't make sense for anything that uses only canvas for rendering.

jsignell:
sorry I meant we can just use:
```
if not @model.visible
  hide(@el)
  return
```
if render_mode is a kwarg

jsignell:
Ok I added the checks for render_mode. Things seem to be working locally now.

sphinx/source/docs/user_guide/examples/styling_visible_annotation_with_interaction.py
```diff
@@ -0,0 +1,31 @@
...
+box = BoxAnnotation(left=1.5, right=2.5, fill_color='green', fill_alpha=0.1)
+p.add_layout(box)
+
+# We write coffeescript to link toggle with visible property of box and line
+code = '''\
+if toggle.active
+    object.visible = true
+else
+    object.visible = false
+'''
```
mattpap: This could be just `object.visible = toggle.active`.

mattpap:
OK, it looks like with current design it can't get any better than this. I hate this duplication, but we will have to make bigger restructuring of renderers  to improve this.

bryevdv:
LGTM will merge when green

PR #5975, opened 2017-03-09
ColumnDataSource.from_dataframe using arrays

### LINKED ISSUE #5973: CDS creation from DataFrames should not use tolist()
Discovered fixing #5152: Currently Pandas DataFrames still convert all their columns to python lists using `.tolist` Not only is this slow, it means that no CDS creates this way will ever make use of the binary array protocol. One option is to use:
```
ds.to_dict('series')
```
which returns a "dict of series" however the series objects are references the actual same series objects in the dataframe. Not sure if sharing is advised. Copying to equivalent numpy array seems advised.

jsignell:
- [x] issues: fixes #5973

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -69,8 +69,8 @@ class ColumnDataSource(ColumnarDataSource):
     ''' Maps names of columns to sequences or arrays.
 
     If the ColumnDataSource initializer is called with a single argument that
-    is a dict, pandas.DataFrame, or pandas.Series, that argument is used as
-    the value for the "data" attribute. For example::
+    is a dict or pandas.DataFrame, that argument is used as the value for the
+    "data" attribute. For example::
 
         ColumnDataSource(mydict) # same as ColumnDataSource(data=mydict)
         ColumnDataSource(df) # same as ColumnDataSource(data=df)
@@ -100,10 +100,10 @@ class ColumnDataSource(ColumnarDataSource):
         raw_data = kw.pop("data", {})
 
         if not isinstance(raw_data, dict):
-            if pd and (isinstance(raw_data, pd.DataFrame) or isinstance(raw_data, pd.Series)):
+            if pd and isinstance(raw_data, pd.DataFrame):
                 raw_data = self._data_from_df(raw_data)
             else:
-                raise ValueError("expected a dict, pandas.DataFrame, or pd.Series got %s" % raw_data)
+                raise ValueError("expected a dict or pandas.DataFrame, got %s" % raw_data)
         super(ColumnDataSource, self).__init__(**kw)
         self.column_names[:] = list(raw_data.keys())
         self.data.update(raw_data)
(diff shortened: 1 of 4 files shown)


bokeh/util/serialization.py
```diff
@@ -150,6 +150,10 @@ def transform_array_to_list(array):
         transformed[np.isposinf(array)] = 'Infinity'
         transformed[np.isneginf(array)] = '-Infinity'
         return transformed.tolist()
+    elif (array.dtype.kind == 'O' and pd.isnull(array).any()):
```
bryevdv: we don't assume `pandas` as a hard dependency, `pd` can be none if `optional_import` above fails, so will need to check that.

jsignell: It it better to look for matches to np.nan? That seems dicey. And np.isnan doesn't work

bryevdv: if the problem is that there are actual pandas nulls then the solution is just to and `and pd` to the if above I think? Can't have pandas nulls without pandas present.

jsignell: you can have an object array with nan in it even if pandas isn't there though right?

bryevdv: certainly but they won't have pandas nulls in them without pandas installed :)

jsignell: This is what I am worried about:

```python
>>> a = np.array(['a', 'b' ,4, np.nan])
>>> np.isnan(a)
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
>>> pd.isnull(a)
array([False, False, False, False], dtype=bool)
```

jsignell: oh wait I read that wrong

bokeh/models/sources.py
```diff
@@ -117,19 +117,20 @@ def _data_from_df(df):
...
+        _df = df.copy()
+        index = _df.index
+        new_data = _df.to_dict('series')
+
         if index.name:
-            new_data[index.name] = index.tolist()
+            new_data[index.name] = index.values
         elif index.names and not all([x is None for x in index.names]):
-            new_data["_".join(index.names)] = index.tolist()
+            print("_".join(index.names))
```
bryevdv: `print` should be removed

bokeh/models/sources.py
```diff
@@ -141,7 +142,7 @@ def from_df(cls, data):
             data (DataFrame) : data to convert
 
         Returns:
-            dict[str, list]
+            dict[str, np.array]
```
bryevdv: I think this will always return a `pd.Series`? We can be specific if so

bokeh/util/serialization.py
```diff
@@ -150,6 +150,10 @@ def transform_array_to_list(array):
         transformed[np.isposinf(array)] = 'Infinity'
         transformed[np.isneginf(array)] = '-Infinity'
         return transformed.tolist()
+    elif (array.dtype.kind == 'O' and pd and pd.isnull(array).any()):
+        transformed = array.astype('object')
+        transformed[pd.isnull(array)] = 'NaN'
+        return transformed.tolist()
```
bryevdv: We need a test for this.

bokeh/models/sources.py
```diff
@@ -141,7 +142,7 @@ def from_df(cls, data):
...
 
         '''
         return cls._data_from_df(data)
```
bryevdv: I'd suggest a couple of tests, at a minimum we should maintain that passing a DataFrame results in a `.data` with `pd.Series` as column values

bokeh/models/sources.py
jsignell: the index is an `np.array`

bryevdv:
Test failures are purely due to intersphinx faillures caused by `scipy.org` being down. Going to go ahead and merge now.

Issue #5518, opened 2016-12-07
Add new child to existing column


mrocklin:
When adding a child plot to an existing column and then updating the plot appears but then I get the following error:

    bokeh.server.protocol.server_handler - ERROR - error handling message Message 'PATCH-DOC' (revision 1): RuntimeError('Cannot apply patch to 8cf5ddd0-9b96-49fc-8099-c792733d0ff5 which is not in the document',)

And my figures come out wonky:

![image](https://cloud.githubusercontent.com/assets/306380/20989991/d7a81e04-bca4-11e6-814a-edf8757e71f0.png)

```python
 
    def __init__(...):
        ...
        self.sizing_mode = 'scale_width'
        self.figures = {}
        self.sources = {}
        self.root = column(sizing_mode=self.sizing_mode)

    def add_figure(self, name):
        source = ColumnDataSource({'x': [], 'y': []})
        fig = figure(title=name, tools='', height=150, sizing_mode=self.sizing_mode)
        fig.line(source=source, x='x', y='y')
        self.sources[name] = source
        self.figures[name] = fig
        self.root.children = self.root.children + [fig]

    def update(self):
        for name, counter in self.server.counters.items():
            if name not in self.figures:
                self.add_figure(name)
            else:
                xs = [counter.quantile(i / 100) for i in range(101)]
                ys = [xs[i + 1] - xs[i] for i in range(len(xs) - 1)]
                self.sources[name].data.update({'x': xs, 'y': ys})
```

mattpap:
That error is unexpected. Failing on layout is something that would normally happen. Layout, as currently implemented, doesn't like changes.

mrocklin:
I'm also fine with a "this is not supported" answer.  I can find ways around this if necessary.

bryevdv:
It should be supported, we've always intended to have a second pass to shore up some rough edges around layout there just have not been resources to do it yet.

mattpap:
> I'm also fine with a "this is not supported" answer.

No, `'PATCH-DOC' (revision 1): RuntimeError('Cannot apply patch (...)'` always indicates a bug in core bokeh's (or bokehjs') infrastructure.

bryevdv:
Yes, the protocol is supposed to afford adding new models, not just patching existing ones.

azjps:
Encountered similar issue, although was not adding a new figure but instead adding a new completely unrelated DOM element. This apparently causes the constraint solver to fail, but only when certain other models like figures are in the document.

[...]

Issue #4810, opened 2016-07-15
Trouble Swapping out layout contents when using server


brendancol:
I'm looking for a simple way to swap out figures inside a layout...basically hiding one figure and displaying another.  Below I have an example of my intent.  What Im seeing is that when a figure is dynamically added to a Column, its plot_height/plot_width is set to 0.  

Example: 

``` python
from bokeh.io import curdoc
from bokeh.plotting import Figure
from bokeh.models import Row, Column, Button
from random import randint

def _create_fig():
    fig = Figure(plot_width=100, plot_height=100)
    fig.circle(x=0, y=0)
    return fig

def callback():
    num = randint(0,9)
    if num > 5:
        fig1_col.children = [_create_fig()]
        fig2_col.children = []
    else:
        fig2_col.children = [_create_fig()]
        fig1_col.children = []

fig1 = _create_fig()
fig2 = _create_fig()
button = Button()
button.on_click(callback)

button_row = Row(children=[button])

fig1_col = Column(children=[fig1])
fig2_col = Column(children=[fig2])

fig_row = Row(children=[fig1_col, fig2_col])

layout = Column(children=[button_row, fig_row])
curdoc().add_root(layout)
```

brendancol:
funny enough, the example above works if empty column line is execute before re-populating the columns:

```
fig2_col.children = []
fig1_col.children = [_create_fig()]
```

bryevdv:
ping @mattpap I think this might be a dupe, or if not, might be covered by your ongoing layout work. Moving to `0.12.6` to keep track

mattpap:
Definitively a duplicate, though I'm not sure of which issue (there are a few). This will be fixed (or already is) in PR #5976.

bryevdv:
Just confirming that this is not fixed by any work up to 0.12.6dev7

Also here is updated code:

```
from bokeh.io import curdoc
from bokeh.plotting import figure
from bokeh.layouts import row, column
from bokeh.models import Button
from random import randint

def _create_fig():
    fig = figure(plot_width=200, plot_height=200)
    fig.circle(x=0, y=0)
    return fig

def callback():
    num = randint(0,9)
    if num > 5:
        fig1_col.children = [_create_fig()]
        fig2_col.children = []
    else:
        fig2_col.children = [_create_fig()]
        fig1_col.children = []

fig1 = _create_fig()
fig2 = _create_fig()
button = Button()
button.on_click(callback)

button_row = row(button)

[...]

Issue #1650, opened 2015-01-08
Issue Validation error when plots are duplicated in one document


bryevdv:
Code to reproduce, top four plots do not render glyphs (tools work and versions below respond accordingly)

```
import numpy as np

from bokeh.plotting import figure, output_file, show
from bokeh.layouts import column, gridplot

N = 50

x = np.linspace(0, 4*np.pi, N)
y = np.sin(x)

output_file("grid.html", title="grid.py example")

TOOLS = "pan,wheel_zoom,box_zoom,reset,save,crosshair"

l = figure(title="line", tools=TOOLS, plot_width=300, plot_height=300)
l.line(x,y, line_width=3, color="gold")

aw = figure(title="annular wedge", tools=TOOLS, plot_width=300, plot_height=300)
aw.annular_wedge(x, y, 10, 20, 0.6, 4.1, color="navy", alpha=0.5,
    inner_radius_units="screen", outer_radius_units="screen")

bez = figure(title="bezier", tools=TOOLS, plot_width=300, plot_height=300)
bez.bezier(x, y, x+0.4, y, x+0.1, y+0.2, x-0.1, y-0.2,
    line_width=2, color="olive")

q = figure(title="quad", tools=TOOLS, plot_width=300, plot_height=300)
q.quad(x, x-0.2, y, y-0.2, color="tomato", alpha=0.4)


p1 = gridplot([[l,aw],[bez,q]])

# specify "empty" grid cells with None
p2 = gridplot([[l,None,aw],[bez,q,None]])

show(column(p1, p2))
```

bryevdv:
Nevermind, problem was because `q` was re-used...

bryevdv:
This is still a problem in `0.12` due to "duplicate edit variables"

birdsarah:
Just to clarify: what is expected to happen in this use case. Are the two plot objects the same i.e. if you pan/select/.... one the other one shows the same?

bryevdv:
I still can't think of any real reasonable use-case for duplicating plots (or more generally, any "layoutable") in a document. I think "solving" this would be a waste of effort to fix a non-problem, so instead I propose to add a validation check for situations like this, which should be simple to check.

bryevdv:
[...]

Issue #5959, opened 2017-03-06
notebook comms "Basic Usage.ipynb" broken


bryevdv:
This used to work as of `0.12.4`, does not work with `0.12.5dev14`

<img width="680" alt="screen shot 2017-03-06 at 10 29 17" src="https://cloud.githubusercontent.com/assets/1078448/23619000/d442a592-0257-11e7-9392-deb8f48b2f29.png">

@mattpap could your recent PR have affected this? I'm not aware of any other layout related PRs. 

In general, plots must continue to be multiply-"showable" in notebooks

mattpap:
> could your recent PR have affected this?

Definitively, but that's good. I will handle this.

mattpap:
This will be fun. You can reproduce this without notebook with:
```
from bokeh.plotting import figure, show
from bokeh.layouts import row

p1 = figure()
p1.line([1, 2, 3], [1, 2, 3])

p2 = figure()
p2.line([1, 2, 3], [1, 4, 9])

show(row(p1, p2))
show(p1)
```
What happens is that you can have plots repeated in a document, which means that layout's solver will collect variables and constraints multiple times and this results in "duplicate edit variable". Previously this error must have been conveniently hidden behind `try-catch` block. This issue is even more prominent in the notebook, because all cells share one solver.

bryevdv:
We should probably have a new document per output cell, to be honest.

mattpap:
That's one solution, but I'm not sure if this wouldn't affect comms in any way and perhaps could result in explosion in number of documents floating around. It would also undermine the idea of roots. 

Another idea is to have one solver per root. After all, what's a document with multiple roots, a row, a column or else? Having a solver per root layoutable would open room for having embedded layouts, e.g. in tabs, which currently are badly broken by reusing the global solver.

bryevdv:
> Another idea is to have one solver per root. 

Actually, this has been discussed, and I thought, implemented already? But maybe I am forgetting, and it used to be a solver-per-page and now its solver-per-doc. I don't personally think more docs is big problem, but solver per-root seems worth investigating.

bryevdv:
Thinking more, I think solver per root is definitely advised. I think having bokeh apps that can be templated ala `components` will require that. I still think there was at least a branch in the past that implemented solver per-root. @birdsarah do you remember?

Also adding `critical`, as I would not release until this is fixed.

birdsarah:
[...]

PR #5976, opened 2017-03-10
Another attempt to make layout more reliable

### LINKED ISSUE #4449: Ensure canvas state reset at start of drawing
Continuing discussion from #4405. The [`render()` method](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/plots/plot_canvas.coffee#L374) in `plot_canvas.coffee` is the main entry point for the rendering; we should assume a clear state of the context because at the start of that function, we [make a call](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/plots/plot_canvas.coffee#L407-L411) to `@canvas_view.prepare_canvas()`, which might resize the canvas (which resets context state).

All our `ctx.save()` and `ctx.restore()` usage pairs up as it should, however, if an error occurs (for whatever reason), any matching `ctx.restore()` may not be called, leaving the context in an unpredictable state. Since one of the first things we do in `plot_canvas.render()` is set a scaling transformation for the HIDPI, the plot will generally "blow up" in such a situation.

Although this is not a terrible thing per-see, as it only occurs when something is going wrong anyway, I think it can be a nasty / distracting side-effect. People might be wondering (and reporting) the plot from blowing up while the error is due to something completely unrelated.

Therefore, I propose to _ensure_ that the state of the context is clear before we start rendering, by making it reset in the `@canvas_view.prepare_canvas()` method (which already does a reset when it resizes the canvas).

Since there is no `ctx.reset()`, I propose to call `ctx.restore()` a couple of times. I agree that the _number_ of times to call `ctx.restore()` is rather arbitrary, and it generally means that we'll be doing a few unnecessary calls to `ctx.restore()` but I don't think the effect would be significant. 

A more sophisticated approach might be to keep tack of the depth of the stack by replacing `ctx.save()` and ctx.restore()` and updating a counter, and resetting based on that, but that feels a bit hacky.

### LINKED ISSUE #5967: Unexpected behavior when deleting from layout.children
I have encountered a bug when updating row contained in a column.

When `r1.children[0]` is deleted, r2 disappears and only the tools are left.  I would expect a column with `s1` above `s2`.  Is deleting from the children array not supported?

Example IPython notebook [https://github.com/paddymul/bokeh_bugs/blob/master/Bokeh%20row%20update%20artifact.ipynb](url)

```
#this messes up the rendering row2
r1 = row(s1,s2)
r2 = row(s3)
displayed_obj = column(r1, r2)

handle = show(displayed_obj, notebook_handle=True)
time.sleep(5)
del(r1.children[0])
push_notebook(handle=handle)
```
screen shot 2017-03-08 at 4 40 31 pm

Versions
python 2.7.11
IPython notebook 4.1.2
Bokeh 0.12.4

Thanks

mattpap:
- [x] implement per-root layout solver
- [ ] make layout reinitializable

fixes #5967 
? #5518
? #4810
fixes #4449

? #1650 -- not fixed because constraints are still bound to models, and there are other issues to resolve (e.g. we shouldn't touch `plot.renderers`).


mattpap:
I just noticed that sometimes (e.g. in grid plots), layout enter an infinite loop of changing constraints, recomputing solution and triggering events. I'm not sure right now how much effort is needed to fix this.

philippjfr:
I've just been testing this PR and comparing it to master. As far as I can tell this PR severely degrades performance in the layout solver. Notebooks containing a lot of plots now take considerably longer to render, pegging CPU load at 100% throughout and even once all the plots are done rendering, to the point that the notebook becomes unresponsive. 

When plotting column or gridplots I can now watch each subplot appear one by one when before they would all (at least appear to) render in one go. Specifically I tried the two main holoviews bokeh tutorials ([1](http://holoviews.org/Tutorials/Bokeh_Elements.html) and [2](http://holoviews.org/Tutorials/Bokeh_Backend.html)), which would take a little while to render before but basically freeze my browser tab when I try to regenerate them with this PR.

Sorry for the negative feedback and unfortunately I don't have any suggestions on how to fix/improve this either.

bryevdv:
As a concrete example, the `models/file/iris_splom.py` example comes up with some of the subplots missing, taking a few seconds to show them.

bryevdv:
I think there may be regressions in master too, though. If I run the "howto" LInked Paning notebook with master and  this PR, and repeatedly run the last three cells over and over, then the `show` gets progressively slower and worse. But it seems to get worse much faster on master

mattpap:
I figured out the reason for performance regression. I'm working on a solution.

bryevdv:
@mattpap can you confrim you are talking about a solver related and not the notebook show issue in #6037

mattpap:
@bryevdv, no, I'm talking about regression introduced in this PR. Also, I'm working on a better fix, because what committed earlier is not a complete fix.

philippjfr:
>Also, I'm working on a better fix, because what committed earlier is not a complete fix.

Just to say I can confirm it's no longer cripplingly slow now to the point of freezing my browser tab, but it's still considerably slower than before. I'll test again when there's a more complete fix.

mattpap:
> `models/file/iris_splom.py`

This example is slow for another, unrelated reason. In this case, x and y ranges are shared across all plots and initialization of each next plot forces rerendering of all previously rendered plots (due to ranges being "updated").

bryevdv:
I just merged the other PR for notebook slowdown. @philippjfr can you merge master into this PR locally, and see how the combination is?

bryevdv:
@mattpap actually can you merge/rebase master?

mattpap:
@bryevdv,  I think I resolved all performance regressions of this PR.

bryevdv:
Making any selection in `crossfilter` blanks plot (no console messages)

screen shot 2017-03-22 at 15 23 01

bryevdv:
Also get this error with lass select tool:
```
TypeError: this.throttled_render is not a function. (In 'this.throttled_render()', 'this.throttled_render' is undefined)
```

mattpap:
> `TypeError: this.throttled_render is not a function. (In 'this.throttled_render()', 'this.throttled_render' is undefined)`

That's messed up. This shouldn't have ever worked. Fixed 477aafa.

bryevdv:
Confirmed lasso select working now. Also all other apps seem to be functioning as expected. Checking notebooks next.

mattpap:
> Making any selection in crossfilter blanks plot (no console messages)

Plot appears again when browser is resized or `resize()` method called. This is something trivial yet annoying.

bryevdv:
notebook comms, notebook server embed and general notebook all appear to be working well on OSX.

philippjfr:
Just tested the holoviews tutorials, performance even on large grids is now back to normal.

mattpap:
btw. This is a very useful example I've been working with during work on this PR:
```
from bokeh.io import curdoc
from bokeh.models.widgets import Select

def get_figure(color, n):
    p = figure(title="markers.py example")
    x = range(100)
    y = [ xi**n for xi in x ]
    p.line(x, y, color=color, name="color=%s, n=%s" % (color, n))
    return p

w = Select(options=['red', 'green', 'blue'], value='red')

f1 = get_figure('red', 1)
f2 = get_figure('red', 2)
f3 = get_figure('red', 3)
c = column(w, f1, f2, f3)

def update(name, old, new):
    c.children[1] = get_figure(w.value, 1)
    c.children[2] = get_figure(w.value, 2)
    c.children[3] = get_figure(w.value, 3)

"""
def update(name, old, new):
    c.children[1:] = [
        get_figure(w.value, 1),
        get_figure(w.value, 2),
        get_figure(w.value, 3),
    ]
"""

w.on_change('value', update)

curdoc().add_root(c)
```
The commented out code should work as well, but it's a much simpler and less useful case. This should be turned into a selenium tests.

bryevdv:
OK, if the crossfilter example issue can be resolved, I'd support this plan:

* whatever needed (e.g. skipping) to get tests green
* more manual testing (by me) 
* merge and RC for tomorrow

bryevdv:
> The commented out code should work as well, but it's a much simpler and less useful case. This should be turned into a selenium tests. 

Absolutely agree we need some selenium tests around apps at this point. I think we could run bokeh server on saucelabs? ping @canavandl

mattpap:
> if the crossfilter example issue can be resolved

It must be. The above code is now broken as well (used to work earlier today).

bryevdv:
> It must be. 
 
Obviously :) I was outlining the minimum requirements for a release this week.

mattpap:
This can be changed to a non-server example as well. I've been working also with TypeScript examples as well, just then you won't see `PATCH-DOC` errors.

mattpap:
I'm chasing my tail here. I fix one thing, another gets broken.

bryevdv:
How can I help? Are there avenues I can help investigate?

bokehjs/src/coffee/models/layouts/layout_dom.coffee
```diff
@@ -1,42 +1,128 @@
...
+  _reset_solver: () ->
+    if not @is_root
+      @parent._reset_solver()
+    else
+      @_solver.clear()
+      @_init_solver()
+
+  _init_solver: () ->
+    @_doc_width = new Variable("document_width")
+    @_doc_height = new Variable("document_height")
```
bryevdv: for later: `_root_width` seems more descriptive

bryevdv:
@mattpap my observation is that `prepare_canvas` is not getting called enough. In the interest of calling it more (but not all the time) I tried this:
```
index 603db3240..01a5406d3 100644
@@ -514,6 +514,9 @@ export class PlotCanvasView extends BokehView

   render: () ->
     logger.trace("PlotCanvas.render() for #{@model.id}")
+    if not @_canvas_prepared?
+      @canvas_view.prepare_canvas()
+      @_canvas_prepared = true

     if Date.now() - @interactive_timestamp < @model.plot.lod_interval
       if not @lod_started
```
This fixes the crossfilter example. I am checking others now, but so far so good. Would this be "good enough" for now?

mattpap:
@bryevdv, this change is likely to break at last some forms of resizing. I have a different solution ready, but that in turn breaks toolbar positioning. There are so many things badly implemented that when I do one thing right, I need to do at least one other as well, and so on.

mattpap:
Actually, I read this wrong.

bryevdv:
@mattpap I just tried with all the various embed resizing and other resizing examples, and they seem to work as expected

mattpap:
Lets proceed then.

bryevdv:
@mattpap OK, can you skip or update whatever tests need to be? If not now, then in the morning? I will merge and get an RC out when it is green

bryevdv:
@mattpap gmap not rendering until a resize, then reporting 
```
[Error] Error: detached element
```

bryevdv:
Actually I think I had an old kernel running, after a restart all I can get for gmap is 
```
TypeError: undefined is not an object (evaluating 'window.google.maps')
```

bryevdv:
so, FWIW the `CRITICAL` push notebook issue seemed to be fixed by the other hotfix PR for notebook perf regression. Given that, the problems remaining are problems that already existed. I think there is just too much risk to merge this for `0.12.5`. Let's get the other events PR in and this can be your primary focus for a (much smaller) `0.12.6` milestone.

bryevdv:
FYI with fix above gmaps behave very badly on reloads, reshows. Basically `request_render` can get called before `initialize` somehow, then it falls over when `is_paused` is undefined.

bryevdv:
I verified #5959 is working on master so I am closing it separately.

mattpap:
> I think you are calling render earlier

Yes, I changed the (DOM) rendering model, so now there is no delayed or deferred rendering. This requires further changes in other places to make it reliable.

mattpap:
> `[Error] Error: detached element`

This shouldn't happen, but can because we start rendering before root `el` is added to DOM. This is one of changes I referenced in the earlier comment.

mattpap:
> and this can be your primary focus for a (much smaller) 0.12.6 milestone.

OK. This will allow to actually fix things.

bryevdv:
Just FYI the examples under `howto/layout` do not look good/correct with this PR currently

bryevdv:
I can confirm that the canvas is the correct size, the drawing is just 1/4 too small

bryevdv:
@mattpap this it the fix for the HiDPI problem
```
index e2329f5eb..f2da733f1 100644
@@ -544,7 +544,7 @@ export class PlotCanvasView extends DOMView
     @model.frame._update_mappers()

     ctx = @canvas_view.ctx
-    ctx.pixel_ratio = ratio = @canvas_view.pixel_ratio  # Also store on cts for WebGL
+    ctx.pixel_ratio = ratio = @canvas.pixel_ratio  # Also store on cts for WebGL

     # Set hidpi-transform
     ctx.save()  # Save default state, do *after* getting ratio, cause setting canvas.width resets transforms
```

bryevdv:
`examples/plotting/server/responsive.py` no longer trigger updates on resize

mattpap:
> this is the fix for the HiDPI problem

Sure, this is what happens when the same value is stored all over the code.

bryevdv:
FYI `plotting/server/responsive.py` still not firing events, probably just some missing plumbing

bryevdv:
Unfortunately all the examples under `howto/layouts` are now seriously broken for me after the latest commits.

mattpap:
> examples under howto/layouts are now seriously broken

None are tested on travis ci, so that's not unexpected. I'm investigating.

canavandl:
The bokeh/tests/integration/interaction/test_datatable failure is real, when clicking on a cell to edit a ``Error: model of a view wasn't configured`` exception is raised.

canavandl:
It appears that the ``sizing_mode='scale_width'`` isn't working, but the others are fine.

mattpap:
> The bokeh/tests/integration/interaction/test_datatable

I already fixed that one locally.

bryevdv:
Let's plan to discuss this in the morning. This PR is blocking several other avenues of work (due to the merges that would be extremely difficult to resolve). If this is still having problems by tomorrow AM CST then we will need to take a step back, let other work proceed, and figure out how to accomplish this in a series of smaller PRs. There is a fair amount of cleanup and incidental refactoring in this PR that could be done incrementally and verified, to reduce the size of the actual relevant changes that need review and evaluation.

mattpap:
> plotting/server/responsive.py still not firing events, probably just some missing plumbing

Those events are intentionally disabled at this point to make debugging easier. Otherwise there is many "unable to apply patch" errors in JS console. Those errors are indicative of design issues in the server, but are otherwise harmless, at least to layout.

mattpap:
> Unfortunately all the examples under howto/layouts are now seriously broken

I broke it when making adjustments to figure out what's wrong with unit tests. It's fixed now. I also enabled `howto/layouts` on travis ci.

mattpap:
Given that tests should be all green at this point, I should be able to finish this PR tomorrow.

mattpap:
One regression to note is that examples' tests run now 22 minutes instead of 16 (I'm excluding newly enabled examples). A lot more examples timeout in JS. After looking at the log, I suspect the recently added "finished rendering" heuristic may be at fault, because otherwise everything works as expected and there are no image diff failures (besides two spurious).

canavandl:
What's the "finished rendering" heuristic that you mention?

mattpap:
> What's the "finished rendering" heuristic that you mention?

https://github.com/bokeh/bokeh/blob/master/tests/plugins/phantomjs_screenshot.js#L75. Note this is pretty limited in scope. I'm working on something that will cover more bases (e.g. image loading, maps).

bryevdv:
`plotting/file/bar_chart.py` displays incorrectly to begin (fixes itself after a window resize)

screen shot 2017-04-26 at 08 41 05


A bit more troubling, the `movies` example fails to load, closing the WS connection immediately

bryevdv:
Annoyingly, I if I try to break on all exceptions and run, `movies` works.

bryevdv:
It seems to be working on Chrome and FF but not safari

mattpap:
@bryevdv, everything works fine as much as my testing capabilities allow. I have a suspicion what may not work in safari, but I would need access to it to tests things efficiently.

bryevdv:
OK, well I can look into getting a mac for you to test on. But that will take time.; So my proposal to is spit this PR up. There are lot of changes here that could go into individual PRs:

* changes to settings, embed, client.coffee, etc. 
* style improvements to `get_constraints`
* canvas fixup changes
* renaming constraints to use underscores

Those are probably not the cause of these problems, but they add alot of bulk to this PR. Let's verify that those changes are OK by performing them in isolation, and make the remainder smaller and easier to evaluate.  We can discuss more at the meeting.

bryevdv:
@mattpap  Here is the console screen shot

screen shot 2017-04-26 at 10 40 15

mattpap:
@bryevdv, btw. is this the only failing server example?

bryevdv:
I will check now

mattpap:
`bar_chart` example was broken yesterday. It should be easy to fix. What's worrying is that tests on travis ci didn't pick this up.

bryevdv:
> btw. is this the only failing server example?

Yes, all other examples are working. My only observation is that a couple (surface3d and spectrogram) *seem* to take a bit longer to show up but I'm not 100% certain

mattpap:
> seem to take a bit longer to show up but I'm not 100% certain

That's perfectly possible. I expect minor performance regressions as a cost for correct behavior.

mattpap:
I reviewed all changes to code that may affect the server and there isn't much of that:

* ebf679233bf74bae7fbd5e0e1c8a8bca58230cd1 changes client.coffee, but you said you already tried to revert this, so it's not it.
* f6225b8664328126cc22dbbdc71022d09e9f1135 changes embed.coffee and seems to be a nice candidate for unexpected behavior.

@bryevdv, can you revert the later commit and see if the example works?

bryevdv:
trying now

bryevdv:
@mattpap no that does not solve the problem. However, if I I reduce the data size, even by half:
```
movies = movies[::2]
```
Then it starts working. I believe I may have heard something about safari changing the default max websocket frame size. I thought that was for the future. Perhaps it's not for the future, and somehow is interacting with other changes here. If there is a way to raise that limit we could investigate.

mattpap:
Just to make it clear, this behavior doesn't show up on `master`?

bryevdv:
In fact it is broken on master [1] and is broken (now) on `0.12.5` as well. Does anyone have an old version of safari they can test with? For reference 

safari: 10.1 (12603.1.30.0.34)
OSX: 10.12.4

[1] with the "renders but WS closes and sliders dont work" I described in the call

bryevdv:
OK this def does not have to be fixed here, just the bar chart one. Please make another issue. Also reference this ML discussion:

https://groups.google.com/a/continuum.io/forum/#!searchin/bokeh/websocket$20safari%7Csort:relevance/bokeh/KvrSS3CMBC4/mU0ekn1WAQAJ

I have to step out for an hour or so

ping @canavandl FYI

bryevdv:
6013.1.30 came out a month ago, guessing its the culprit https://en.wikipedia.org/wiki/Safari_version_history

mattpap:
Googling for "safari 10.1 websockets" reveals that we aren't the only unhappy customer.

canavandl:
here's an issue for the safari/websocket problem: https://github.com/bokeh/bokeh/issues/6196

bryevdv:
OK all the relevant layout examples in the repo are working. I'm not sure #4810 or all of #5518 are fixed tho the problems now may be server sychronization. The original example for from #5518 is working, it seems to be adding figures that does not work. I thing there may be an issue adding objects that have references to other instances.

bryevdv:
@mattpap I would support merging this PR as it currently is. There is work to add tests that can probably be rolled into some client work. Are you done / ok to merge?

mattpap:
@bryevdv, lets merge. I still have to finalize making layout reinitializable, so that's why some of the listed issues aren't fully fixed. I will submit a new PR to finish this, today or tomorrow.

bryevdv:
OK, I've updated the top so those issues won't auto close. Merging now.

lukauskas:
With this PR merged, the `document.resize()` function is removed.
I was relying on this function to resize the image from `CustomJS`, an approach similar to the one described [here](http://stackoverflow.com/a/40095304) by birdsarah.

Can someone suggest how to trigger such resize from CustomJS in the new Layout API ?

This is sort-of related to #5381

canavandl:
Thanks for pointing that out @lukauskas , i'm was also relying on ``document.resize`` in another project. 

@mattpap what's the best way to trigger a ``resize`` on a layout with the changes?

bryevdv:
@mattpap this should get a migration note if there is not one already.

mattpap:
@lukauskas, what's described on SO looks like a hack rather than a good practice. Just setting `width` of a figure should do all the necessary resizing, though I doubt this works as expected currently. For now you can use this instead:
```
for (var id in Bokeh.index) {
  Bokeh.index[id].resize()
}
```

lukauskas:
@mattpap oh yes, absolutely, it was definitely a hack.

What I found is that setting `width` (or `height`) is sufficient in practice with the new code.
The caveat is that I do not know any way of telling the layout that the width has changed.
The figure only updates when window is resized.

For my use case, I can afford to automatically trigger the window resize event every time I want to change the size. I am aware though this is as hacky as it gets.

If anyone is interested here's what works:

```python
resize_callback = CustomJS(args=dict(plot=plot),
                                   code="""
                                   var new_height = 12345; // your height here
                                   plot.height = new_height;

                                   // This is the only way to trigger layout re-rendering
                                   // that I can figure out right now
                                   // IE9+ only
                                   var event = document.createEvent('HTMLEvents');
                                   event.initEvent('resize');
                                   window.dispatchEvent(event);
                                   """.format(row_height)
                                   )
```

PR #5978, opened 2017-03-10
Correctly infer document from bokeh.io._state for bokeh.io.save (instead of creating new one)

### LINKED ISSUE #5977: Themes aren't applied to document when using `save`
Themes aren't being applied to a document when using `bokeh.io.save`. Themes seem to work when using `bokeh.io.show`.

Toggle the save/show lines in the below code example to reproduce.

```python
import colorsys
import yaml

from bokeh.layouts import column, row, widgetbox
from bokeh.models import ColumnDataSource, HoverTool, CustomJS, Slider
from bokeh.plotting import figure, output_file, show, curdoc, save
from bokeh.themes import Theme

# for plot 2: create colour spectrum of resolution N and brightness I, return as list of decimal RGB value tuples
def generate_color_range(N, I):
    HSV_tuples = [ (x*1.0/N, 0.5, I) for x in range(N) ]
    RGB_tuples = map(lambda x: colorsys.hsv_to_rgb(*x), HSV_tuples)
    for_conversion = []
    for RGB_tuple in RGB_tuples:
        for_conversion.append((int(RGB_tuple[0]*255), int(RGB_tuple[1]*255), int(RGB_tuple[2]*255)))
    hex_colors = [ rgb_to_hex(RGB_tuple) for RGB_tuple in for_conversion ]
    return hex_colors, for_conversion

# convert RGB tuple to hexadecimal code
def rgb_to_hex(rgb):
    return '#%02x%02x%02x' % rgb

# convert hexadecimal to RGB tuple
def hex_to_dec(hex):
    red = ''.join(hex.strip('#')[0:2])
    green = ''.join(hex.strip('#')[2:4])
    blue = ''.join(hex.strip('#')[4:6])
    return (int(red, 16), int(green, 16), int(blue,16))

# plot 1: create a color block with RGB values adjusted with sliders

# initialise a white block for the first plot
hex_color = rgb_to_hex((255, 255, 255))

# initialise the text color as black. This will be switched to white if the block color gets dark enough
text_color = '#000000'

# create a data source to enable refreshing of fill & text color
source = ColumnDataSource(data=dict(color=[hex_color], text_color=[text_color]))

# create first plot, as a rect() glyph and centered text label, with fill and text color taken from source
p1 = figure(x_range=(-8, 8), y_range=(-4, 4),
            plot_width=600, plot_height=300,
            title='move sliders to change', tools='')

p1.rect(0, 0, width=18, height=10, fill_color='color',
        line_color = 'black', source=source)

p1.text(0, 0, text='color', text_color='text_color',
        alpha=0.6667, text_font_size='36pt', text_baseline='middle',
        text_align='center', source=source)

# the callback function to update the color of the block and associated label text
# NOTE: the JS functions for converting RGB to hex are taken from the excellent answer
# by Tim Down at http://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
callback = CustomJS(args=dict(source=source), code="""
    function componentToHex(c) {
        var hex = c.toString(16);
        return hex.length == 1 ? "0" + hex : hex;
    }
    function rgbToHex(r, g, b) {
        return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
    }
    var data = source.data;
    color = data['color'];
    text_color = data['text_color'];
    var R = red_slider.value;
    var G = green_slider.value;
    var B = blue_slider.value;
    color[0] = rgbToHex(R, G, B);
    text_color[0] = '#ffffff';
    if ((R > 127) || (G > 127) || (B > 127)) {
        text_color[0] = '#000000';
    }
    source.trigger('change');
""")

# create slider tool objects with a callback to control the RGB levels for first plot
SLIDER_ARGS = dict(start=0, end=255, value=255, step=1, callback=callback)

red_slider = Slider(title="R", **SLIDER_ARGS)
callback.args['red_slider'] = red_slider

green_slider = Slider(title="G", **SLIDER_ARGS)
callback.args['green_slider'] = green_slider

blue_slider = Slider(title="B", **SLIDER_ARGS)
callback.args['blue_slider'] = blue_slider

# plot 2: create a color spectrum with a hover-over tool to inspect hex codes

brightness = 0.8 # change to have brighter/darker colors
crx = list(range(1,1001)) # the resolution is 1000 colors
cry = [ 5 for i in range(len(crx)) ]
crcolor, crRGBs = generate_color_range(1000,brightness) # produce spectrum

# make data source object to allow information to be displayed by hover tool
crsource = ColumnDataSource(data=dict(x=crx, y=cry, crcolor=crcolor, RGBs=crRGBs))

# create second plot
p2 = figure(x_range=(0,1000), y_range=(0,10),
            plot_width=600, plot_height=150,
            tools='hover', title='hover over color')

color_range1 = p2.rect(x='x', y='y', width=1, height=10,
                       color='crcolor', source=crsource)

# set up hover tool to show color hex code and sample swatch
p2.select_one(HoverTool).tooltips = [
    ('color', '$color[hex, rgb, swatch]:crcolor'),
    ('RGB levels', '@RGBs')
]

# theme everything for a cleaner look
curdoc().theme = Theme(json=yaml.load("""
attrs:
    Plot:
        toolbar_location: null
    Grid:
        grid_line_color: null
    Axis:
        axis_line_color: null
        major_label_text_color: null
        major_tick_line_color: null
        minor_tick_line_color: null
"""))

layout = row(
    widgetbox(red_slider, green_slider, blue_slider),
    column(p1, p2)
)

output_file("color_sliders.html", title="color_sliders.py example")

## toggle this to see document being applied
# show(layout)
save(layout)
```

canavandl:
- [x] issues: fixes #5977
- [x] tests added / passed

Refactor _save_helper to use internally use `_ModelInDocument` (via `bokeh.embed.file_html`) in order to correctly handle document state.

For testing purposed, you can use the bug example from the issue (#5977)

Migration Note: 

* `bokeh.io.save` now will only accept a LayoutDOM item and no longer a document. I don't think this should be much of an interruption, as I don't think any users are saving documents. This also brings the `save` method in line with the `bokeh.io.show` method, making things more intuitive.
* `validate` kwarg to `bokeh.io.save` has been deprecated
* `bokeh.embed.standalone_html_page_for_models` has been marked as deprecated in place of `bokeh.embed.file_html`

### DIFF
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -84,7 +84,7 @@ def _find_existing_docs(models):
         doc = existing_docs.pop()
     else:
         # conflicting/multiple docs, raise an error
-        msg = ('Multiple items in models conatain documents or are '
+        msg = ('Multiple items in models contain documents or are '
                'themselves documents. (Models must be owned by only a '
                'single document). This may indicate a usage error.')
         raise RuntimeError(msg)
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -28,6 +28,7 @@ from .core.json_encoder import serialize_json
 from .document import Document, DEFAULT_TITLE
 from .model import Model
 from .resources import BaseResources, _SessionCoordinates, EMPTY
+from .util.deprecation import deprecated
 from .util.string import encode_utf8
 from .util.serialization import make_id
 
@@ -340,10 +341,8 @@ def file_html(models,
               template_variables={}):
     '''Return an HTML document that embeds Bokeh Model or Document objects.
 
-    The data for the plot is stored directly in the returned HTML.
-
-    This is an alias for standalone_html_page_for_models() which
-    supports customizing the JS/CSS resources independently and
+    The data for the plot is stored directly in the returned HTML, with
+    support for customizing the JS/CSS resources independently and
     customizing the jinja2 template.
 
     Args:
@@ -659,6 +658,7 @@ def standalone_html_page_for_models(models, resources, title):
         UTF-8 encoded HTML
 
     '''
+    deprecated((0, 12, 5), 'bokeh.io.standalone_html_page_for_models', 'bokeh.io.file_html')
     return file_html(models, resources, title)
 
 def server_html_page_for_models(session_id, model_ids, resources, title, websocket_url, template=FILE):
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -52,22 +52,22 @@ def _wrap_in_onload(code):
 """ % dict(code=_indent(code, 4))
 
 @contextmanager
-def _ModelInDocument(models, theme=None):
+def _ModelInDocument(models, apply_theme=None):
     doc = _find_existing_docs(models)
     old_theme = doc.theme
 
-    if theme is FromCurdoc:
+    if apply_theme is FromCurdoc:
         from .io import curdoc; curdoc
         doc.theme = curdoc().theme
-    elif theme is not None:
-        doc.theme = theme
+    elif apply_theme is not None:
+        doc.theme = apply_theme
 
     models_to_dedoc = _add_doc_to_models(doc, models)
 
     yield models
 
     for model in models_to_dedoc:
-        doc.remove_root(model, theme)
+        doc.remove_root(model, apply_theme)
     doc.theme = old_theme
 
 
@@ -197,7 +197,7 @@ def components(models, wrap_script=True, wrap_plot_info=True, theme=FromCurdoc):
         models = values
 
     # 2) Append models to one document. Either pre-existing or new and render
-    with _ModelInDocument(models, theme=theme):
+    with _ModelInDocument(models, apply_theme=theme):
         (docs_json, render_items) = _standalone_docs_json_and_render_items(models)
 
     script = _script_for_render_items(docs_json, render_items, websocket_url=None, wrap_script=wrap_script)
@@ -302,7 +302,7 @@ def notebook_div(model, notebook_comms_target=None, theme=FromCurdoc):
     model = _check_one_model(model)
 
     # Append models to one document. Either pre-existing or new and render
-    with _ModelInDocument([model], theme=theme):
+    with _ModelInDocument([model], apply_theme=theme):
         (docs_json, render_items) = _standalone_docs_json_and_render_items([model])
 
     item = render_items[0]
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -438,7 +438,7 @@ def _save_helper(obj, state, filename, resources, title, validate):
     remove_after = False
     doc = state.document
 
-    if obj not in doc.roots:
+    if obj.document is None:
         doc.add_root(obj)
         remove_after = True
 
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -347,7 +347,7 @@ def _show_server_with_state(obj, state, new, controller):
     show_session(session_id=state.session_id_allowing_none, url=state.url, app_path=state.app_path,
                  new=new, controller=controller)
 
-def save(obj, filename=None, resources=None, title=None, state=None, validate=True):
+def save(obj, filename=None, resources=None, title=None, state=None, **kwargs):
     ''' Save an HTML file with the data for the current document.
 
     Will fall back to the default output state (or an explicitly provided
@@ -370,12 +370,15 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
             If None, use the default state title value, if there is one.
             Otherwise, use "Bokeh Plot"
 
-        validate (bool, optional) : True to check integrity of the models
-
     Returns:
         filename (str) : the filename where the HTML file is saved.
 
     '''
+
+    if 'validate' in kwargs:
+        deprecated((0, 12, 5), 'The `validate` keyword argument', 'None', """
+        The keyword argument has been removed and the document will always be validated.""")
+
     if state is None:
         state = _state
 
@@ -435,7 +438,7 @@ def _get_save_args(state, filename, resources, title):
     return filename, resources, title
 
 def _save_helper(obj, filename, resources, title):
-    html = file_html([obj], resources, title=title)
+    html = file_html(obj, resources, title=title)
 
     with io.open(filename, mode="w", encoding="utf-8") as f:
         f.write(decode_utf8(html))
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -30,6 +30,7 @@ import tempfile
 from .core.state import State
 from .document import Document
 from .embed import notebook_div, standalone_html_page_for_models, autoload_server
+from .models.layouts import LayoutDOM
 from .layouts import gridplot, GridSpec ; gridplot, GridSpec
 import bokeh.util.browser as browserlib  # full import needed for test mocking to work
 from .util.deprecation import deprecated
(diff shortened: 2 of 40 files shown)


bokeh/io.py
```diff
@@ -358,11 +357,10 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
     ``/foo/myplot.html``)
 
     Args:
-        obj (Document or model object) : a plot object to save
+        obj (LayoutDOM object) : a Layout (Row/Column), Plot or Widget object to display
 
         filename (str, optional) : filename to save document under (default: None)
-            If None, use the default state configuration, otherwise raise a
-            ``RuntimeError``.
```
canavandl: This docstring is out-of-date. The implementation now will create a tempfile matching the `__file__` name instead of raising a RuntimeError.

bryevdv: Please update any docstrings that need updating :)

bokeh/io.py
```diff
@@ -439,17 +434,13 @@ def _get_save_args(state, filename, resources, title):
...
-            Document().add_root(obj)
-            remove_after = True
-        doc = obj.document
-    elif isinstance(obj, Document):
-        doc = obj
-    else:
-        raise RuntimeError("Unable to save object of type '%s'" % type(obj))
+    doc = state.document
+
+    if obj not in doc.roots:
```
bryevdv: I think checking `obj.document is None` is better? That handles the case where the `obj` is in a doc whether the doc is the state.document "curdoc" or not. And if `obj.document is None` then the `obj` is not in `doc.roots` so good to add as a fallback

bokeh/tests/test_io.py
```diff
@@ -105,11 +105,6 @@ def test_default_filename(self):
         filename, resources, title = io._get_save_args(io._state, None, "resources", "title")
         self.assertEqual(filename, "filename")
 
-    def test_missing_filename(self):
```
canavandl: This test didn't actually test for a missing filename (which now is handled without raising an exception and tested elsewhere). It just happened to raise a RuntimeError related to trying to save an object of `type(str)`.

bokeh/io.py
```diff
@@ -439,17 +434,13 @@ def _get_save_args(state, filename, resources, title):
 
     return filename, resources, title
 
-def _save_helper(obj, filename, resources, title, validate):
+def _save_helper(obj, state, filename, resources, title, validate):
     remove_after = False
-    if isinstance(obj, LayoutDOM):
-        if obj.document is None:
-            Document().add_root(obj)
```
canavandl: This is cause of the noted bug (themes not being applied). The solution is the get the document from io._state and not create a new one.

bokeh/io.py
canavandl: I did... where did it go?...

bokeh/io.py
bryevdv: See what was done for themes in https://github.com/bokeh/bokeh/pull/5932/files I think we have to do the same thing. Is there a reason `save` can't just be implemented in terms of `components` ?

bokeh/io.py
canavandl: done

bokeh/io.py
canavandl: It's currently implemented via `file_html` which uses the new `_ModelInDocument` code to handle theming. I'm not sure what you'd hope to gain from rewriting.

bryevdv: If that's the case why do we need to do *anything* with docs or roots right here? I'm trying not to duplicate what `_ModelInDocument` is doing, and it seems like we are at least partially doing that below.

bokeh/io.py
```diff
@@ -439,17 +434,13 @@ def _get_save_args(state, filename, resources, title):
...
-    else:
-        raise RuntimeError("Unable to save object of type '%s'" % type(obj))
+    doc = state.document
+
+    if obj.document is None:
+        doc.add_root(obj)
+        remove_after = True
 
     if validate:
         doc.validate()
```
bryevdv: @canavandl just checking, decisions was to remove all the above code and just call `standalone_html_page_for_models` with `obj`, directly at the start of `_save_helpers`, since it uses `_ModelInDocument` internally?

canavandl: done

bokeh/io.py
```diff
@@ -377,15 +375,12 @@ def save(obj, filename=None, resources=None, title=None, state=None, validate=Tr
...
-    Raises:
-        RuntimeError
-
     '''
     if state is None:
         state = _state
 
     filename, resources, title = _get_save_args(state, filename, resources, title)
-    _save_helper(obj, filename, resources, title, validate)
+    _save_helper(obj, filename, resources, title)
```
canavandl: need to warn that the `validate` arg has been no-op'ed. Now it always happens inside of `_ModelInDocument`

bryevdv: This is a private function. Just get rid of it.

canavandl: I commented that poorly, the `validate` value comes from the `bokeh.io.save` function signature, which is user facing. How do we deprecate kwargs in Bokeh?

bryevdv: If they are on public APIs by checking for them and calling `deprecated` explicitly, and then doing whatever fix up is necessary to not propagate the parameter further inside the function.

bryevdv: Alternatively we can pass `validate` through to `_ModelInDoc`

canavandl: We were previously validating an empty document, so essentially not doing validation. I've just removed/deprecated the `validate` kwarg.

canavandl:
I believe this is ready for final review. ping @bryevdv

bokeh/io.py
bryevdv: Fair enough. FYI CI will fail until scripy docs site is back up.

bokeh/io.py
bryevdv: migration note for `obj` param type change and `validate` deprecation / ignore

bokeh/embed.py
```diff
@@ -659,6 +658,7 @@ def standalone_html_page_for_models(models, resources, title):
         UTF-8 encoded HTML
 
     '''
+    deprecated((0, 12, 5), 'bokeh.io.standalone_html_page_for_models', 'bokeh.io.file_html')
```
bryevdv: migration note for function deprecation

bryevdv:
LGTM I think it just needs a migration note

sphinx/source/docs/releases/0.12.5.rst
```diff
@@ -128,6 +128,19 @@ interactive session which formerly could result in a ``PermissionError``.
 
 For details see :bokeh-pull:`5942`.
 
+The ``validate`` keyword argument to ``bokeh.io.save`` has been put into a
+deprecation cycle. Future usage of ``bokeh.io.save`` will always validate
+the document before outputting a file.
+
+The ``bokeh.io.save`` method will now only accept a ``LayoutDOM``object and no
+longer a ``Document`` for its ``obj`` argument. This aligns the
```
bryevdv: missing space before "object"

bryevdv: also pure preference but I usually just say "has been deprecated"

bryevdv:
@canavandl there is a conflict to resolve?

bokeh/embed.py
canavandl: added

bokeh/io.py
canavandl: added

bryevdv:
Failing only due to known issue with scipy.org being down. Merging now.

PR #5979, opened 2017-03-11
Support relative URLs / deprecate --host arg 

### LINKED ISSUE #5692: Make it possible to use relative URLs

bryevdv:
- [ ] issues: fixes #5692
- [ ] tests added / passed

### DIFF
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -7,10 +7,35 @@ import logging
 log = logging.getLogger(__name__)
 
 from ..document import Document
-from ..resources import DEFAULT_SERVER_WEBSOCKET_URL, server_url_for_websocket_url, _SessionCoordinates
+from ..resources import _SessionCoordinates, DEFAULT_SERVER_HTTP_URL
 from ..util.session_id import generate_session_id
 
 DEFAULT_SESSION_ID = "default"
+
+def websocket_url_for_server_url(url):
+    if url.startswith("http:"):
+        reprotocoled = "ws" + url[4:]
+    elif url.startswith("https:"):
+        reprotocoled = "wss" + url[5:]
+    else:
+        raise ValueError("URL has unknown protocol " + url)
+    if reprotocoled.endswith("/"):
+        return reprotocoled + "ws"
+    else:
+        return reprotocoled + "/ws"
+
+def server_url_for_websocket_url(url):
+    if url.startswith("ws:"):
+        reprotocoled = "http" + url[2:]
+    elif url.startswith("wss:"):
+        reprotocoled = "https" + url[3:]
+    else:
+        raise ValueError("URL has non-websocket protocol " + url)
+    if not reprotocoled.endswith("/ws"):
+        raise ValueError("websocket URL does not end in /ws")
+    return reprotocoled[:-2]
+
+DEFAULT_SERVER_WEBSOCKET_URL = websocket_url_for_server_url(DEFAULT_SERVER_HTTP_URL)
 
 def push_session(document, session_id=None, url='default', app_path='/', io_loop=None):
     """ Create a session by pushing the given document to the server,
@@ -53,7 +78,7 @@ def push_session(document, session_id=None, url='default', app_path='/', io_loop
 
     """
     coords = _SessionCoordinates(dict(session_id=session_id, url=url, app_path=app_path))
-    session = ClientSession(session_id=coords.session_id, websocket_url=coords.websocket_url, io_loop=io_loop)
+    session = ClientSession(session_id=coords.session_id, websocket_url=websocket_url_for_server_url(coords.server_url), io_loop=io_loop)
     session.push(document)
     return session
 
@@ -102,7 +127,7 @@ def pull_session(session_id=None, url='default', app_path='/', io_loop=None):
 
     """
     coords = _SessionCoordinates(dict(session_id=session_id, url=url, app_path=app_path))
-    session = ClientSession(session_id=session_id, websocket_url=coords.websocket_url, io_loop=io_loop)
+    session = ClientSession(session_id=session_id, websocket_url=websocket_url_for_server_url(coords.server_url), io_loop=io_loop)
     session.pull()
     return session
 
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -8,6 +8,7 @@ log = logging.getLogger(__name__)
 
 from ..document import Document
 from ..resources import _SessionCoordinates, DEFAULT_SERVER_HTTP_URL
+from ..util.deprecation import deprecated
 from ..util.session_id import generate_session_id
 
 DEFAULT_SESSION_ID = "default"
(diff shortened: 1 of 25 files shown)


bryevdv:
@marami52 @dhuebner  do you guys have the ability to build from source and test? This is one that I can use all the help I can get on. I don't want to merge this until it's in good shape, and it's also verified to work in your use case. 

With the current state of this PR, existing app example seem to work as before, the `server_embed` examples have all been updated as needed but also run.  There are a couple of trivially failing tests but I want to make sure there isn't a better way of doing some things before committing time to updating tests. 

This PR works by allowing an optional absolute url and app path to be passed through to `embed_items` via `autoload_server`.  

@havocp as always any thoughts welcome

marami52:
@bryevdv I just build from the source and checking out your PR. To make sure, I have version `0.12.5dev14+21.g6ce656b`.
It works perfectly fine with this version. Thank you!

bryevdv:
@marami52 yes that is the expected version thank for confirming. And that is good to hear! As I said I want to do some clean up and additional work on this, so now that you have revealed that you can test from source fairly easily, hopefully it won't be too much to ask for testing a few more times. :D

marami52:
@bryevdv by all means! :)

bryevdv:
@havocp offhand do you remember what the need for `app_path` is at all? Is it just to be able to support this:
```
def output_server(self, session_id=None, url="default", app_path='/'):
```
Because if so I am fine stipulating that the default also implies `app_path="/"` Basically, I'd like to deprecate `app_path` and just have the python io/embed functions take the entire URL to an app, always.

bryevdv:
Motivation: if `app_path` can be dispensed with then I believe the only case that would need any extra configuration would be the "embed a server programmatically in Flask" type case, and it would just need a flag to say "pass the full app URL to BokehJS to use for loading resources and creating websockets instead of using relative locations". 

That's slightly annoying, but reasonable (and could possibly be made even better by defining some new helper functions for embedding in different situations based on the situation)

bryevdv:
I guess it's also used to allow `_SessionCoordinates` to make this distinction?
```
    @property
    def server_url(self):
        """ Server URL including app path derived from the kwargs provided."""
        return self._server_url

    @property
    def url(self):
        """ Server base URL derived from the kwargs provided (no app path)."""
        return self._base_url
```
But I think, given the full URL, the app path and the base url could be reliably computed?

bryevdv:
@marami52 can you check the version I just pushed as well? It needs some new tests and minor cleanups, but barring any problems from your end I think this will be the final implementation.

marami52:
@bryevdv I did the test but this time it is not working. I don't get any error message. However, I get an empty page when I run the app. Here is the version: `0.12.5dev14+23.g4669b51`

bryevdv:
Well, that's surprising and unfortunate. Are there no JS console error messages either? Can you set `BOKEH_LOG_LEVEL=debug` as well?

bryevdv:
Actually I can reproduce this, guess this is what you will see:

screen shot 2017-03-14 at 16 52 45


I'll fix that up later tonight

bryevdv:
I think I can repro this but not sure my setup is exactly like yours so the JS console output will be very helpful. Should be able to fix this up later tonight.

marami52:
My JS console doesn't show any error. It says websocket 0 is open but in the network tab it shows pending:

screen shot 2017-03-14 at 3 06 12 pm
screen shot 2017-03-14 at 3 07 21 pm

bryevdv:
Can you try with the logging env var above set?

marami52:
I just ran the app with `--log-level debug` option. Got the same result and nothing changed. I did't get any error message in my logs either.

marami52:
@bryevdv I just deleted everything from last and this version and reinstalled everything again. Now everything is working well! I'm sorry for previous wrong message.

bryevdv:
@marami52 now I'm confused because I can definitely reproduce a problem with a simple nginx setup. I am going to push one  more set of changes soon hopefully it won't betook hard for you to test those too.

marami52:
@bryevdv I know. It was strange for me as well. For testing, for sure. Just let me know when you pushed to start my testing.

bryevdv:
@marami52 I just pushed new changes 
```
$ bokeh info                                                                                      (bryanv/5692_relative_urls)
Python version      :  3.6.0 |Anaconda custom (x86_64)| (default, Dec 23 2016, 13:19:00)
IPython version     :  5.1.0
Bokeh version       :  0.12.5dev14-24-g186614562
```

bryevdv:
I believe it is necessary for the python side to communicate the `app_path` to BokehJS `embed_items`.  What I don't like is that notebook embed vs server embed usage of `DOC_JS` passes two vs four arguments. If this current build works for you I would then add more tests and docs and leave a cleanup of that for later work.

marami52:
@bryevdv yes it works! I keep my container on for new few hours that you can check on your side:
https://ride.r-brain.io:8443/719ecc896ad643438efc0fa2dc5a802c/p5006/myApp 
Also you can check my development env here: 
https://app.r-brain.io/#/ride/719ecc896ad643438efc0fa2dc5a802c/
It's actually running in JLab!

bryevdv:
@marami52 that's great to hear. I am going to push a few docs and tests updates then we should be good to go

bryevdv:
@marami52 one last test? I pushed docs/tests but also some small changes to accomodate some corner cases. Also, can you share what your `autoload_server` call looks like for my info?

marami52:
@bryevdv this one also works well (version `0.12.5dev14+26.g3a65c20`)
how can I get `autoload_server` call to share with you?

bryevdv:
@marami52 sorry I was not completely clear, I was asking for an example of how you are calling `autoload_server`, what arguments you pass it.

marami52:
I see. I just run `bokeh serve myApp.py`

bryevdv:
@marami52 I see, when I heard about you and @dheubner running behind nginx I assumed you'd also be embedding the app in some other HTML page (which would require using `autoload_server`). You are just going to / offering navigation directly to the proxied URL to the app that nginx exposes?

marami52:
Yes. That's correct.

bryevdv:
OK, there was an actual docs build failure coincident with the `scipy.org` problem. PR #5990 erroneously left a monkey patch for `push` in which broke the docs. I am fixing that here. The `scipy.org` problem appears resolved as well. Plan to merge this when green, then put out a dev build for easier testing.

dhuebner:
@bryevdv Great! Will the PR be in the next release? If yes, when?

bryevdv:
The release is planned for next week. This PR is merged, so barring any unknown and unforeseen showstoppers, it will be that release. A new dev build is available `0.12.5dev15` that includes this PR, so any help testing it would be appreciated.

dhuebner:
Thanks

bryevdv:
@dhuebner @marami52 I need some feedback, I am thinking that the default should be switched. I think relative urls are only *needed* in a small number of cases, meaning that it should default to false. Do you have any strong comments or objections to making the default value here:

https://github.com/bokeh/bokeh/blob/master/bokeh/embed.py#L473

be `False` for `relative_urls` ?

marami52:
@bryevdv thank you for asking. I think for all cloud services like ours, it should be relative url by default. So, our answer is biased toward current setup. However, if the majority is in the opposite way, we can give the hint to our users for `relative_urls` and it doesn't make any technical problem for us.

bryevdv:
made separate issue to discuss (#6066)

PR #5995, opened 2017-03-15
Add support for CustomJSTransform

### LINKED ISSUE #5015: CustomJSTransform
Adding new transforms is easy, but people should be able to self serve something like this:

``` python
from bokeh.models import CustomJSTransform

transform = CustomJSTransform.from_coffeescript(
    compute="""return 2*x""",
    v_compute="""
        vals = []
        for x in xs:
            vals.push(2*x)
        return vals
    """
)

.....

p.circle(x='x', y='y', size={field: 'x', transform: transform})
```

Just like https://github.com/bokeh/bokeh/pull/4659, user would _know_ that single `x` is passed to compute and array `xs` is passed to v_compute. User would specify a return value and would also be able to pass in args - like for CustomJS.

canavandl:
issues: fixes #5015

Todo:
- [x] tests added / passed
- [x] Add `from_py_func` method for transform via flexx
- [x] Add `from coffeescript`
- [x] Add example to docs


bokehjs/src/coffee/models/transforms/customjs_transform.coffee
```diff
@@ -0,0 +1,31 @@
...
+
+  @getters {
+    func: () -> @_make_func()
+    values: () -> @_make_values()
+  }
+
+  compute: (x) -> @func(@values..., require, x)
+
+  v_compute: (xs) ->
+    result = new Float64Array(xs.length)
```
bryevdv: is this a safe assumption? What if users want to transform to colors or categorical values? In general I think we might want to require users to supply both a "scalar" and "vector" version of the transform, so that we don't have to make assumptions.

bokehjs/src/coffee/models/transforms/customjs_transform.coffee
```diff
@@ -0,0 +1,31 @@
...
+    for x, idx in xs
+      result[idx] = @func(@values..., require, x)
+    return result
+
+  _make_func: () ->
+    # this relies on Object.keys(args) and values(args) returning keys and values
+    # in the same order
+    new Function(Object.keys(@args)..., "require", "x", @code)
```
mattpap: `x` should be placed before `require` and `require` should be followed by `exports`.

bokehjs/src/coffee/models/transforms/customjs_transform.coffee
mattpap: Apparently this is how all other transforms are implemented. At this point I doubt we even care about typed arrays floating around, maybe besides reducing memory usage.

bryevdv: The other transforms are not generic. It can be safely assumed the input and output of a `Jitter` transform is numeric. That's not the case here, though.

canavandl: You are correct, it shouldn't be a typed array. Will fix

bryevdv: Well my point is as really that if users are responsible for supplying the vector version themselves , they could supply one that uses a typed array or not, appropriately

canavandl:
I'm breaking this out so that the comment thread doesn't get collapsed.

> In general I think we might want to require users to supply both a "scalar" and "vector" version of the transform, so that we don't have to make assumptions that might not always hold. It's a little more work for users but it they are reaching for CustomJSTransform they can probably handle providing both func and v_func?

We had a similar discussion when we implemented the FuncTickFormatter, regarding having users write scalar or vectorized transforms for tick values. We ended up going with a scalar transform and looping over all of the tick values in the implementation.

Pros:
* No concerns about data column "raggedness" (wrong length)
* Simpler to implement/debug

Cons:
* No access to column values for aggregate-level stats or neighbor values (can't support normalization, differentiation, etc).

Because of the need to do the aggregate-level stats (something that's not really required for tick transformations AFAICT), I think it's necessary to make users implement vectorized versions of their
transforms (in addition to the scalar version).

canavandl:
I'm also not stoked with the name "CustomJSTransform". Idk if there's something better

bryevdv:
> We had a similar discussion when we implemented the FuncTickFormatter, regarding having users write scalar or vectorized transforms for tick values. We ended up going with a scalar transform and looping over all of the tick values in the implementation.

Right but the situations are not entirely analogous. The output of a tick formatter is *always* an array of strings. That's not the case here. Transforms can have arbitrary output types depending on where he output of the transform gets used. For this reason and also those you mention I agree that in the specific case of `CustomJSTransform` it's best to place the burden on users to supply the vectorized version. 

A possible concession in the future: some function to auto-generate a "dumb" vectorized version that just collects the scalar values in a (non-Typed) array. That might be useful for some people, some of the time. I'd be +0 on it. 

> I'm also not stoked with the name "CustomJSTransform". Idk if there's something better

Agreed, esp since `CustomJS` should then really be `CustomJSCallback`, but isn't. I don't have any better suggestions tho.

bokeh/models/tests/test_transforms.py
```diff
@@ -0,0 +1,133 @@
...
+
+def test_customjstransform_from_py_func_no_args():
+
+    def cosine():
+        from flexx.pyscript import window
+        return window.Math.cos(x) # noqa
+
+    def v_cosine():
+        from flexx.pyscript import window
+        return [window.Math.cos(x) for x in xs] # noqa
```
mattpap: Isn't it possible to just `from flexx.pyscript import Math`?

canavandl: Something like that is mentioned in the docs (https://flexx.readthedocs.io/en/stable/pyscript/api.html#Math), but they may be out of date. I can't find the `Math` stub in 0.4.1

canavandl:
Example looks like: 

screen shot 2017-03-20 at 3 25 45 pm

canavandl:
So the name is `CustomJSTransform`, which isn't amazing but I didn't come up with anything else.

bokeh/models/tests/test_transforms.py
bryevdv: I am fairly down on "python->js transpilation" in general at this point. So, I agree this can be lived with as is for now.

PR #5997, opened 2017-03-16
Move webgl glyphs and related code to bokeh-gl bundle

### LINKED ISSUE #4321: Put WebGL functionality in separate bokeh-gl.js?

mattpap:
Reduces `bokeh.min.js` size from 699 to 636 kB. `gloo2` and `models/glyphs/webgl` were moved to `bokeh-gl` bundle, which is automatically included if `Plot.webgl == True`. The logic is the same, i.e. there are no new models and `webgl` is left. This also means that the code is still a mess, with webgl popping up everywhere, but at least the bulk of the webgl code is in the new bundle.

fixes #4321

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/glyph.coffee
+++ b/bokehjs/src/coffee/models/glyphs/glyph.coffee
@@ -4,7 +4,6 @@ import * as proj from "core/util/projections"
 import {BokehView} from "core/bokeh_view"
 import {Model} from "../../model"
 import {Visuals} from "core/visuals"
-import * as bokehgl from "./webgl/main"
 import {logger} from "core/logging"
 import {extend} from "core/util/object"
 import {isString, isArray} from "core/util/types"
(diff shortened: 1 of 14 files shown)


bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -88,9 +88,8 @@ export class PlotCanvasView extends BokehView
     @canvas_view.render(true)
 
     # If requested, try enabling webgl
-    if @model.plot.webgl or window.location.search.indexOf('webgl=1') > 0
-      if window.location.search.indexOf('webgl=0') == -1
-        @init_webgl()
```
bryevdv: migration note

bokehjs/src/coffee/models/glyphs/glyph.coffee
```diff
@@ -21,11 +20,21 @@ export class GlyphView extends BokehView
...
-        Cls = bokehgl[@model.type + 'GLGlyph']
-        if Cls
+    ctx = @renderer.plot_view.canvas_view.ctx
+
+    if ctx.glcanvas?
+      try
+        glglyphs = require("models/glyphs/webgl/index")
+      catch e
+        if e.code == 'MODULE_NOT_FOUND'
+          logger.warn('WebGL was requested and is supported, but bokeh-gl(.min).js is not available, falling back to 2D rendering.')
```
bryevdv: prefer "back to HTML canvas rendering"

bryevdv:
@mattpap you will need to also update

https://github.com/bokeh/bokeh/blob/master/scripts/build_upload.py#L349

bokehjs/src/coffee/models/glyphs/glyph.coffee
mattpap: @bryevdv, webgl is HTML canvas rendering, just with a different context.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
mattpap: @bryevdv, I though this was a developers' tool, but apparently it's in the documentation.

PR #5999, opened 2017-03-16
Make custom models' bundle part of embed instead of resources

### LINKED ISSUE #5353: Extensions do not render when using bokeh.embed.components / bokeh.embed.file_html APIs
I'm attempting to generate HTML that includes extensions (LaTeX) using the `bokeh.embed.components` and `bokeh.embed.file_html` APIs.
http://bokeh.pydata.org/en/latest/docs/reference/embed.html#bokeh.embed.components

I used the LaTeX example in the documentation as a starting point:
http://bokeh.pydata.org/en/latest/docs/user_guide/extensions_gallery/latex.html#userguide-extensions-examples-latex

Please see the attached code ([latexTest.py.txt](https://github.com/bokeh/bokeh/files/529342/latexTest.py.txt)) that illustrates the issues below.

When using the `file_html` API, the extension code (LatexLabel) appears to be included in the resulting HTML/script section -- however, the figure does not render.  In the case of the `components` API, the extension code does not appear to be included, and as expected, the figure does not render and the following error message is displayed:

_Bokeh Error
Module `LatexLabel' does not exists. The problem may be two fold. Either a model was requested that's available in an extra bundle, e.g. a widget, or a custom model was requested, but it wasn't registered before first usage._

References:
https://github.com/bokeh/bokeh/issues/647 [Support latex labels]
https://github.com/bokeh/bokeh/pull/4914 [Add support for loading external resources in output template]
https://github.com/bokeh/bokeh/issues/3193 [desired_num_ticks not working when using custom x_range]

mattpap:
This allows to use custom models with `components()` by moving custom models' bundles from resources to embed.

fixes #5353

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -27,7 +27,6 @@ from .settings import settings
 
 from .util.paths import bokehjsdir
 from .util.session_id import generate_session_id
-from .util.compiler import gen_custom_models_static
 from .model import Model
 
 DEFAULT_SERVER_HOST = "localhost"
@@ -331,10 +330,6 @@ class JSResources(BaseResources):
         if self.log_level is not None:
             raw.append('Bokeh.set_log_level("%s");' % self.log_level)
 
-        custom_models = gen_custom_models_static()
-        if custom_models is not None:
-            raw.append(custom_models)
-
         return raw
 
     def render_js(self):
@@ -459,5 +454,3 @@ class Resources(JSResources, CSSResources):
 CDN = Resources(mode="cdn")
 
 INLINE = Resources(mode="inline")
-
-EMPTY = Resources(mode="inline", components=[], log_level=None)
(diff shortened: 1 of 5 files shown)


bryevdv:
Just checking, this does not solve the issue of accumulating all extension models, even if they are not used in the document currently being rendered?

mattpap:
@bryevdv, no.

bryevdv:
This PR needs some tests

mattpap:
However, this will improve performance, because now custom models are computed only when generating embedding, not every time when `Resources.js_raw` is used.

bokeh/resources.py
```diff
@@ -459,5 +454,3 @@ def render(self):
 CDN = Resources(mode="cdn")
 
 INLINE = Resources(mode="inline")
-
-EMPTY = Resources(mode="inline", components=[], log_level=None)
```
bryevdv: Pretty sure that at least a few folks have relied on this. Is there a pressing need to remove it?

PR #6003, opened 2017-03-17
check for transform listener on glyph

### LINKED ISSUE #5174: Task: support updating computed transforms fields if dependencies change
Currently, when a transform source field (or transform itself) is changed, the computed field isn't updated.

For example: when using a color mapper as a transform, if the color mapper instance or the field being transformed is updated, the `fill_color` attribute isn't computed.

``` python
p.patches('x', 'y', source=source,
          fill_color={'field': 'rate', 'transform': color_mapper},
          fill_alpha=0.7, line_color="white", line_width=0.5)
```

jsignell:
- [x] issues: fixes #5174
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/core/has_props.coffee
+++ b/bokehjs/src/coffee/core/has_props.coffee
@@ -131,6 +131,8 @@ export class HasProps extends Backbone.Model
     # that subsequent updates do not duplicate that setup work.
     for name, prop of @properties
       prop.update()
+      if prop.spec.transform
+        @listenTo(prop.spec.transform, "change", () -> @trigger('transformchange', this))
 
   setv: (key, value, options) ->
     # backbones set function supports 2 call signatures, either a dictionary of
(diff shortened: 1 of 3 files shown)


bokehjs/src/coffee/core/has_props.coffee
```diff
@@ -405,6 +405,10 @@ export class HasProps extends Backbone.Model
       # this skips optional properties like radius for circles
       if (prop.optional || false) and prop.spec.value == null and (name not of @_set_after_defaults)
         continue
+      if prop.spec.transform
+        if not prop.spec.transform._listeners[this._listenId]
```
bryevdv: It would be better not to dig into private Backbone API. What is this trying to accomplish? Is this to prevent the same transform from being listened to multiple times?

jsignell: > Is this to prevent the same transform from being listened to multiple times?

exactly

bryevdv: It might also be worth investigating if this could be done here in this loop instead:

https://github.com/bokeh/bokeh/blob/15037b3e4e9fc7aa4eb061a94acaec48af897af1/bokehjs/src/coffee/core/has_props.coffee#L132-L133

That's already looping over all the properties so it seems like a nice place to set things up. Also it only happens once so I dont think the check for listeners would be needed

jsignell: yep ok that works great.

bryevdv:
One thing we are not doing, and we don't have to do here, is updating things in case the transform model itself changes. i.e. not just changing a property on a current transform, but swapping out an entirely new transform object. I think it would probably actually be a bit hard to even accomplish  this at the moment, which is why it's not a priority. But we might want to make a follow on issue just to record the possibility.

PR #6008, opened 2017-03-17
deprecate mpl compat

### LINKED ISSUE #5788: migration notes and exceptions about MPL compat.
MPL compat does not work with MPL 2.0, and is unlikely to ever until or unless MEP 25 is implemented. The `0.12.5` release notes should make mention of this, and importing the MPL compat fiunctions should raise an error if MPL >= 2.0 is installed.

bryevdv:
- [x] issues: fixes #5788

This PR deprecates all MPL compat capability, to be removed when Bokeh 1.0 is released. It also cleans up the release notes and adds a rough cut of "what's in the release".

@bokeh/dev and especially @pzwang, any comments on the text of the deprecation announcement?


bokeh/core/compat/__init__.py
```diff
@@ -24,3 +24,9 @@
 .. _mplexporter: https://github.com/mpld3/mplexporter/tree/master/mplexporter
 
 '''
+from ...util.deprecation import deprecated
+
+deprecated(
+    "MPL compatibility can no longer be successfully maintained, and is now deprecated. " +
+    "All MPL compat functions will be removed completely on the release of Bokeh 1.0. " +
```
jbednar: I might suggest saying "prior" instead of "on", because it might be good to remove them slightly ahead of 1.0 so that we can be sure nothing relies on them.

bryevdv: nothing where? I can guarantee nothing inside the project does. My definite preference is to give folks as long as possible and also for the breakpoint for such a big deletion to be "logical"

jbednar: In the examples, training materials, etc.

bryevdv: I'd be amendable to start removing the examples and docs before 1.0 but leaving the actual hard cutoff at for when then modules are removed at 1.0 There are not many places, I am confident that they can be removed fully in one go and if not, we can mop up any stragglers...

jbednar: Sounds good!

jbednar: Well, except that I'd remove examples *immediately*, as soon as anything becomes deprecated, to reduce the number of people who start relying on anything scheduled to disappear...

bryevdv: That's not unreasonable. I could do that in this PR.

bryevdv:
OK I've added a second commit that removes all compat examples and compat usage in the docs. As a result we can drop `seaborn` as a dev dependency. I have left the few basic MPL tests just to make sure we don't break anything while the deprecation is still active. After the deprecation is finished we can drop MPL as a dev dependency as well.

bryevdv:
@jbednar as an aside I'd suggest adding a data shader to the "other libraries" section

jbednar:
Ok, I'll add datashader once this PR is merged, to avoid conflicts.

mattpap:
```
/home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/compat.rst:: WARNING: document isn't included in any toctree
```

bryevdv:
@mattpap removing compat was a mistake (it has holoviews and JS and data shader info) I will have to revert now. I also still wanted to get comment from @pzwang

PR #6011, opened 2017-03-18
Bryanv/5788 deprecate mpl

### LINKED ISSUE #5788: migration notes and exceptions about MPL compat.

bryevdv:
One more attempt. Would like to have comment from @bokeh/dev  and give @pzwang time to offer any input as well  

- [x] issues: fixes #5788
- [x] tests added / passed


bryevdv:
Perhaps we need to have another status to discriminate between 'ready for comments" and "ready to merge"

mattpap:
Possibly, but at this point I think that the earlier PR should have had WIP status.

bryevdv:
> Possibly, but at this point I think that the earlier PR should have had WIP status.

It was my best judgement that it was done, I did not catch the mistake until later. If there more people involved regularly we could probably afford a process with more explicit steps for acceptance. Some day...

mattpap:
@bryevdv, with WIP status I'm not talking about the mistake, which could have been fixed in a followup PR, but about giving others and especially @pzwang time to comment. Alternatively to WIP, one could use "ready" status and ask for review. This actually would be even better, because it indicates that your job is done and you're just waiting for thirdparties.

bryevdv:
Well, I stated in the beginning of the PR that I was intending to get feedback. I also generally *do* interpret "ready" to mean "ready for comment/review"

PR #6013, opened 2017-03-19
Make sure `node --version` terminates and is evaluated lazily

### LINKED ISSUE #6012: Importing bokeh.models creates zombie process
The following was brought to my attention: 

Importing `bokeh.models` seems to create a zombie process.  Here is a minimal failing example:

```python
import bokeh.models, psutil, time
time.sleep(3)
print(psutil.Process().children())
print(psutil.Process().children()[0].status())
```

prints

```
[<psutil.Process(pid=4851, name='node') at 140524007092176>]
zombie
```

This may be caused by https://github.com/bokeh/bokeh/blob/master/bokeh/util/compiler.py#L116 

I mentioned this to @bryevdv and he suggested that I ping and assign @mattpap

mattpap:
fixes #6012

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -113,22 +113,32 @@ def _detect_nodejs():
 
     for nodejs_path in nodejs_paths:
         try:
-            Popen([nodejs_path, "--version"], stdout=PIPE, stderr=PIPE)
+            proc = Popen([nodejs_path, "--version"], stdout=PIPE, stderr=PIPE)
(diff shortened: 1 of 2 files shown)


bryevdv:
I think mocks should be used sparingly, but in this instance I think it is appropriate to maintain that `.wait` is called, otherwise it would be easy for a regression to slip in in the future.

bryevdv:
@mattpap I know you are working on several things at once, I've added the test and will merge when green

PR #6015, opened 2017-03-19
Fixed bug in Model._process_event method

jlstevens:
Small fix to code merged from #5941: in ``Model._process_event`` the variable``event_name`` is undefined and it should be ``event.event_name``. 

After this fix, there are no more exceptions in the web console when running ``events_app.py``. Note this code path isn't executed when running ``js_events.py`` which is why I didn't notice this problem earlier.

### DIFF
--- a/bokehjs/src/coffee/model.coffee
+++ b/bokehjs/src/coffee/model.coffee
@@ -30,7 +30,7 @@ export class Model extends HasProps
       for callback in @js_event_callbacks[event.event_name] ? []
         callback.execute({event: event}, {})
 
-      if @subscribed_events.some((m) -> m == event_name)
+      if @subscribed_events.some((m) -> m == event.event_name)
         @document.event_manager.send_event(event)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Are there any tests we can add to this?

jlstevens:
I'm happy to set up Javascript tests (which is where this error originated) given a little bit of guidance. Figuring how to break down the tests properly will need some thought...

mattpap:
> Are there any tests we can add to this?

This should have been written in TypeScript, then you would get nice compile-time error message about undeclared variable.

bryevdv:
> This should have been written in TypeScript, then you would get nice compile-time error message about undeclared variable.

Given that not everyone knows TS well (or at all) yet, it's going to be a long transition process.

bryevdv:
In the interest of getting a dev build or RC out that can be used for testing the next two days, I am going to merge this as-is

PR #6033, opened 2017-03-21
Make multiple selection work again

### LINKED ISSUE #6032: Multiple selections via shift+select don't work
v bokeh 0.12.5dev16
python3
Chrome 57

Confirmed for the LassoSelectTool and BoxSelectTool, but thought to exist for all selection tools.

Best to reproduce with examples/app/selection_histogram.py

After doing a first box selection (which works normally), holding shift to perform another box selection causes the second selection to not "end" (i.e. the overlay will continue following the cursor). The glyphs in the second selection area will be rendering as their selection_glyphs after ending the second selection, but it seems that some "selected" event isn't occuring, because the selection histogram app demo doesn't update the histograms.

mattpap:
fixes #6032

### DIFF
--- a/bokehjs/test/core/util/array.coffee
+++ b/bokehjs/test/core/util/array.coffee
@@ -5,7 +5,88 @@ array = utils.require "core/util/array"
 
 describe "array module", ->
 
+  it "copy should return a new copy", ->
+    a = [0, 1, 2]
+    b = array.copy(a)
+    expect(b).to.deep.equal [0, 1, 2]
+    a[0] = 10
+    expect(a).to.deep.equal [10, 1, 2]
+    expect(b).to.deep.equal [0, 1, 2]
+
+  it "concat should append all arrays", ->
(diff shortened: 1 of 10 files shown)


bryevdv:
@canavandl will add a selenium test for multiple selection to this PR

bryevdv:
I'm also going to add some unit tests to `array.ts` while I have it open since there are none

bokehjs/test/core/util/array.coffee
```diff
@@ -5,7 +5,84 @@ array = utils.require "core/util/array"
 
 describe "array module", ->
 
+  it "copy should return a new copy", ->
+    a = [0, 1, 2]
+    expect(array.copy(a)).to.deep.equal [0, 1, 2]
```
mattpap: This test doesn't assert that a copy is new. To assert the test statement you have to test whether inplace modification of copy doesn't affect the original. This can be improved later.

PR #6034, opened 2017-03-21
Improve propagation of events in UIEvent._trigger()

### LINKED ISSUE #6030: BoxZoomTool and BoxSelectTool doesn't clear the overlay when end event is off the plot frame
v bokeh 0.12.5dev16
python3
Chrome 57

BoxZoomTool and BoxSelectTool doesn't clear the overlay when end event is off the plot frame

screen shot 2017-03-21 at 12 26 57 pm

screen shot 2017-03-21 at 12 26 50 pm

mattpap:
This fixes the problem, but this logic is very cumbersome. This needs to be revisited in the next dev cycle.

fixes #6030

### DIFF
--- a/bokehjs/src/coffee/core/ui_events.coffee
+++ b/bokehjs/src/coffee/core/ui_events.coffee
@@ -131,11 +131,20 @@ export class UIEvents
       switch event_type
         when "tap"
           view.on_hit?(sx, sy)
-    else if not @_hit_test_frame(sx, sy)
-      @trigger("move:exit", e)
 
+      if base_type == "pan"
+        @_trigger_event_for_gesture(event_type, e)
+    else if not @_hit_test_frame(sx, sy)
       if base_type == "move"
         @plot_view.set_cursor()
+
+      if event_type != "move:exit"
+        @trigger("move:exit", e)
+
+      if base_type == "move"
+        @trigger(event_type, e)
+      else
+        @_trigger_event_for_gesture(event_type, e)
     else
       if base_type == "move"
         active = any(@toolbar.inspectors, (t) -> t.active)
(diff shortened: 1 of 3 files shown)


bryevdv:
> This fixes the problem, but this logic is very cumbersome. This needs to be revisited in the next dev cycle.

unfortunately it breaks multi-select, second shift selections highlight points but don't update selections and box selections fail to hide after pan_end

bryevdv:
needed a re-base on master to pick up the other fix for multi-select. Will merge when green (happily we now know the selenium tests for multi-select works!)

bryevdv:
@mattpap @canavandl @bokeh/dev OK after merging master and fixing the unit tests, there was one saucelabs test failing. After investigating, I think the new behaviour is actually much preferred, so I changed the test itself. The new behaviour is this: if making a selection and the mouse leaves the plot area, instead of clearing, it makes the selection. This is very useful for making selections near the plot border easier. I'm going to go ahead and merge this and get an RC out, but please comment if you have thoughts or disagreement with this change.

canavandl:
This also changes the behavior of the PanTool and BoxZoomTool. Previously, a pan or zoom end outside of the frame would stop the interaction and now the pan continues as you drag beyond the frame (I'm actually surprised a selenium test isn't failing for this) and now the zoom occurs.

I don't have a strong opinion on whether this is good or not, I just wanted to point out the change.

jlstevens:
I don't think this change affects the events system as it *looks* like ``_bokify_hammer`` and ``_bokify_point_event`` are being used in the same way as before. If you don't think this is true, please tell me and I'll test that the events system is still working as expected.

bryevdv:
@canavandl good catch. I think for the pan and zoom, the previous better was better. But for select, the new behavior is better. I think we mostly just need to define and document the expected behaviors in tests (AFAIK theres not one for the pan out of plot area) Can you make an issue for this?

AS far as what to do, I'm personally OK to release with pan and zoom having the new behavior as a minor known issue. Does anyone have strong thoughts otherwise?

jbednar:
I'd have to try it to be sure, but the new behavior sounds like an improvement to me, even for panning and zooming.  Seems like it will make it simpler to make a large pan across a small viewport, and similarly for zooming. But maybe I'm not imagining it very clearly.

bryevdv:
It's a mixed bag to be honest, there are definitely different cases where different behaviors are preferable. I'd just like to nail it down with tests, whatever it is decided.

StevenCHowell:
I agree that being able to pan/zoom outside the plot frame would be a nice change.  This would be especially helpful for smaller plot frames.

pzwang:
I'm +1 on continuing an initiated pan/zoom even if the mouse moves outside of the plot frame.

My logic is this: when the user is initiating a pan or zoom by clicking somewhere within the plot, the location of the plot border is meaningful and significant, since it has a meaning in data space as well.

But once they begin the interaction, then the linearity and continuity of the interaction is the important thing.  Unless some aspect of the interaction is relative to the plot frame or the data coordinates of where the plot boundary maps to, the frame is a then merely a "bit of UI" that doesn't map to anything meaningful in their mental model.  In a deeper sense, the "mouse position" has ceased to be the primary concept, and instead "the mapping of mouse motion to data space motion"   is now the primary concept in the user experience.  So, any artifact of the mouse position should not abruptly and jarringly interrupt the interaction.

Using this conceptual framing (pardon the pun), then we can also see that if we ever build a selection mode that can pan the plot while the user is selecting, then it is improper to terminate the selection interaction as long as the plot is panning.  (I am aware that this is not what our current selection operation does.)

Additionally, this reasoning for doing mouse capture then also implies that if the mouse is over things like interaction legends, other widgets, etc., it should not trigger them.

bryevdv:
Great, it sounds like we have accidentally made an improvement. I just think it needs to now be maintained under test.

jbednar:
I agree with Peter's explanation; that's why panning outside the plot frame makes sense to me.

PR #6043, opened 2017-03-23
Avoid isinstance checks in _visit_value_and_its_immediate_references

mrocklin:
Isinstance checks can be surprisingly costly when applied to many small
objects.  This avoids many calls to isinstance in
_visit_value_and_its_immediate_references by doing two things:

1.  Short circuiting on common types.  These appear to be 90% of all
    inputs to this funciton
2.  Rearranging the if-tree to prefer more commonly succeeding checks
    before less commonly succeeding ones

In my benchmark isinstance calls took up around 5% of computation time
and now they take up around 2%

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -700,8 +700,12 @@ _common_types = {int, float, str}
 
 
 def _visit_value_and_its_immediate_references(obj, visitor):
-    '''
+    ''' Recurse down Models, HasProps, and Python containers
 
+    The ordering in this function is to optimize performance.  We check the
+    most comomn types (int, float, str) first so that we can quickly return in
+    the common case.  We avoid isinstance and issubclass checks in a couple
+    places with `type` checks because isinstance checks can be slow.
     '''
     typ = type(obj)
     if typ in _common_types:  # short circuit on common base types
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -695,19 +695,27 @@ def _visit_immediate_value_references(value, visitor):
     else:
         _visit_value_and_its_immediate_references(value, visitor)
 
+
+_common_types = {int, float, str}
+
+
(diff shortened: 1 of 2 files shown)


bokeh/util/serialization.py
```diff
@@ -216,16 +216,16 @@ def traverse_data(obj, is_numpy=is_numpy, use_numpy=True):
...
+        if type(item) is float:
             if np.isnan(item):
                 item = 'NaN'
             elif np.isposinf(item):
                 item = 'Infinity'
             elif np.isneginf(item):
                 item = '-Infinity'
             obj_copy.append(item)
+        elif isinstance(item, (list, tuple)):
+            obj_copy.append(traverse_data(item))
```
bryevdv: I 'd like to mere this for `0.12.5` but it needs a comment at the top explaining the rationale for the ordering, otherwise it risks being changed later accidentally.

bryevdv: I"d like to merge this for `0.12.5` today but it needs a comment at the top explaining the reason for the order, or else it risks being accidentally or unintentionally changed later

bokeh/model.py
```diff
@@ -695,19 +695,27 @@ def _visit_immediate_value_references(value, visitor):
...
+    elif issubclass(typ, dict):
         for key, value in iteritems(obj):
             _visit_value_and_its_immediate_references(key, visitor)
             _visit_value_and_its_immediate_references(value, visitor)
+    elif issubclass(typ, HasProps):
+        if issubclass(typ, Model):
+            visitor(obj)
+        else:
+            # this isn't a Model, so recurse into it
+            _visit_immediate_value_references(obj, visitor)
```
bryevdv: same comment as below, this needs a comment for future devs to explain why it is the way it is

PR #6045, opened 2017-03-23
fixed hover on gapminder to display country

jsignell:
Fixes a hover tools bug introduced in #5972.

Country names were not being shown on hover. 

Before:
screen shot 2017-03-23 at 8 54 34 am

After:
screen shot 2017-03-23 at 8 53 38 am


jsignell:
@birdsarah, @bryevdv and I decided that the duplicated versions of the gapminder example were confusing and hard to keep updated. There is just one version now and it runs in an app.

PR #6053, opened 2017-03-23
Do not replace NaN or Infinite with 0

### LINKED ISSUE #6035: NA values correctly skipped in linear scale but not in log scale
Hi,

I have an issue with plotting "NA" values (as imported with a CSV reader on R) as lines when using a log scale. Here is the code example in a Jupyter notebook:
```
import numpy as np

# prepare some data
x = [1, 2, 3, 4, 5, 6, 7, 8]
y = [0.01, 0.03, 0.1, 'NA', 1, 3, 10, 30]
#chartData = {"x": x, "y": y}

# output to Jupyter notebook
output_notebook()

# create a plot
p = figure(plot_width=600, plot_height=400,
          )

p.line(x, y, line_width=2)
# show the results
show(p)
```
bokeh_plot_na_linear

The 'NA' value is correctly skipped and shows as a break in the line.

Now if I plot the same data but using a log scale for the Y axis:

```
# create a plot
p2 = figure(plot_width=300, plot_height=250,
y_axis_type="log",
          )

p2.line(x, y, line_width=2)
# show the results
show(p2)
```
bokeh_plot_na_log

As you can see, the 'NA' value is arbitrarily plotted at the level of the X axis (whatever the Y axis range I can specify). It also happens when using numpy.nan instead of 'NA'.

Am I doing anything wrong? Is this a bug, and is there a workaround, other than removing the 'NA' data (which is a pain for large matrices and multiple line plots, where the 'NA' are not always consistently sampled in series)? Thanks in advance.

zebulon2:
NaN/Infinite values should not be replaced with a finite value, since this breaks line plotting code with 'NA' in data series using a log scale. Fixes #6035.


bryevdv:
@canavandl This would be a nice little fix, I do think the behaviour of the `LogColorMapper` might need very minor tweaks. Do you concur?

zebulon2:
Just a comment regarding the failed tests: there seems some inconsistency regarding the handling of +/-Infinity value. With ColorBar/LogColorMapper test, +Infinity should become 100 and Infinity should become 0. On the other hand, with log_mapper/Range1D, -Infinity should be equal to 10. I suppose that before my patch, those transforms were made when the functions were receiving the value 0 (that replaced Infinity in the log mapper). We should now catch Infinity in those functions and do the tick transform directly.

zebulon2:
I understand the test failure: when Infinity was replaced with 0, the map_to_target(0) test function in https://github.com/bokeh/bokeh/blob/master/bokehjs/test/models/mappers/log_mapper.coffee#L21-L24 would return 0*100+10 = 10.
So now all the question is: should the test be corrected, and in bokehjs code, should the tick calculation that returns 10 be made in another function that tests the value for Infinity?

bryevdv:
I think it would be good to add a `nan_color` property to handle special casing colormaping nans (if desired) But that's for a different PR. For +/-inf I think they should map to the upper and lower range respectively. Is that the behavior now? If so I agree the test should simply be updated.

zebulon2:
Actually we have 5 fails:

- with ColorBar module ColorBar ColorBar._tick_coordinates we should have +Infinity equals 100 and -Infinity equals 0. Instead we get null. I suppose, but correct me if I am wrong, that 0/100 result is fine and the code for _tick_coordinates should be updated?

- log_mapper module creation with Range1d ranges where -Infinity = 10. This is where probably the test is inaccurate.

-  PlotCanvasView get_canvas_element "after each" hook: I am unsure about this error.

Any comments?

bryevdv:
Hi @zebulon2 sorry for the delay, I was on vacation. The "after each" error is sporadic error we currently suffer, it can be ignored (restart the test and it usually goes away). I agree with your assessment of the other two cases, but also ping @canavandl for input.

canavandl:
Sorry for the delay in responding. @zebulon2 I have some code that came out of some related work with the LogTicker that would move your PR forward, do you mind if I push it onto your PR?

zebulon2:
Hi, no problem with the delay, hope you had great vacation. @canavandl yes please do, I am more than happy to test.

bokehjs/src/coffee/models/mappers/log_mapper.coffee
```diff
@@ -20,11 +21,14 @@ export class LogMapper extends Model
     if inter_scale == 0
       value = 0
     else
-      value = (Math.log(x) - inter_offset) / inter_scale
-      if isNaN(value) or not isFinite(value)
-        value = 0
+      clamped_x = clamp(x, @source_range.start, @source_range.end)
```
canavandl: We were previously clamping the start value, but it's not clear (to me) why. Should I remove the clamping in it's entirety? We don't clamp for the LinearMapper.

bokehjs/src/coffee/models/mappers/log_mapper.coffee
```diff
@@ -20,11 +21,14 @@ export class LogMapper extends Model
...
+      _x = (Math.log(clamped_x) - inter_offset) / inter_scale
+      if isFinite(_x)
+        value = _x * scale + offset
+      else
```
canavandl: We're treating both NaN and Infinity values as NaN. Is that reasonable?

bokehjs/test/models/glyphs/rect.coffee
```diff
@@ -297,14 +297,14 @@ describe "Rect", ->
           expect(result3['1d'].indices).to.be.deep.equal([1])
 
         it "should work when axis is log", ->
-          data = {x: [60, 100, 140], y: [60, 100, 140]}
+          data = {x: [1, 10, 100, 1000], y: [1, 10, 100, 1000]}
```
canavandl: I'm unsure about this test. I understand the intent is to that `_hit_point` works by passing in vx/vy coordinates and checking that the method is able to calculate the glyph location in log coords. But in practice it's hard to know what the vx/vy coords should be or whether the test is passing for some unrelated reason (i.e. the glyph is massive and takes up the whole plot area).

It don't know if it's better to have a better structure for how to write these sorts of tests or to just replace it with a selenium test. Due to my changes, I don't think the tests covers anything meaningful.

canavandl:
Working example:

```python
from bokeh.plotting import show, figure
import numpy as np

p = figure(x_axis_type='log')
p.line(x=[1, 10, 100, np.nan, 500, 1000], y=[1, 3, 5, 4, 2, 6])

show(p)
```

screen shot 2017-04-10 at 10 38 02 pm

bokehjs/test/models/glyphs/rect.coffee
bryevdv: selenium test seems preferable but let's make a separate issue to add that, and remove this at the same time.

bokehjs/src/coffee/models/mappers/log_mapper.coffee
bryevdv: I think so, there's not much else we can do with `inf` at the low level besides "not draw"

bokehjs/src/coffee/models/mappers/log_mapper.coffee
bryevdv: I'm not sure why, and agree it does not seem necessary and even possibly erroneous. This maps all points outside a plot region to the edge of the plot region? Seems that would produce bad results for large size markers or line points that are off screen that get moved to the edge.

canavandl: Looking at the tests, it looks like we were only left-side clamping. I'd assume it was a hack to handle negative values, which has been obviated with this PR. I think I should remove the clamping.

bryevdv: Agree with removing (but please test existing examples for good measure)

bokehjs/src/coffee/models/mappers/log_mapper.coffee
canavandl: agreed.

canavandl:
@zebulon2 if you have an opportunity to test this PR, that'd be fantastic.

bryevdv:
Thanks for the contribution @zebulon2 !

canavandl:
We're created a dev release with this PR that can be installed via ``conda install -c bokeh/label/dev bokeh=0.12.6dev1``. Let us know if you're able to try it and it doesn't work.

zebulon2:
Hi, I have been able to test 0.12.6dev1 and indeed it works fine with discontinuous series. Thanks a lot for merging it to the main branch and publish it in the next release.

Issue #5984, opened 2017-03-12
Improve data transfer, using a binary transfer protocol


volkerjaenisch:
Currently data is transfered between Bokeh-Server and client as a JSON structure.
While JSON is a established format for AJAX, JSON is not capable as high performance data transfer mechanism for scientific applications.

Currently NP-Arrays that could be transfered in binary mode, were base 64 encoded and transfered as text embedded in a JSON structure.
On the JS side the decoding of base64 and a byte wise shuffling into JS typed Arrays takes lot of time and RAM.

As an example we implemented our own **binary** protocol for data transfer and it is much faster than Bokeh.
Our home-brew-data-structure to transfer data between Bokeh Server and client is crude but at least a order of magnitude faster than Bokeh.

From the user perspective one defines targets and associates data in a meta data dict:

    data_map = {
            'source.data.x': x_data,
            'source.data.index': x_data,
            'source.data.y': y_data,
            'source.data.y_above': y_above,
            'source.data.y_below': y_below,
        }

We align it in a binary data structure:

Length of Metadata
Metadata
dataset1
dataset2

0) The pos/length of all data structures to be submitted is calculated
1) The length of the meta data is coded into a string with 8 digits. So we beat endianess problems.
2) The metadata is JSON coded and added binary.
3) The metadata maps a Target (Plot1.source.data.X) to a Dataset defined by its position and length in the binary data structure. 
4) A Dataset is a binary representation of a numpy array fit for being 1:1 mapped to as JS typed array. 

On the JS side we disentangle the metadata first. Then we can map the binary data of the NP-Arrays to their targets JS typed arrays very fast, while respecting endianess.

This works quite well and we like to implement is to Bokeh.

Volker

bryevdv:
A few quick comments:

At the lowest level, the Bokeh protocol already supports multi-fragment messages, which was done explicitly with the notion in mind of sending zero-copy binary buffers (e.g. NumPy memory views) as separate fragments, directly over websocket to JS typed arrays. However, this capability has not been put to any real use yet. But, you can see the details here e.g.

https://github.com/bokeh/bokeh/blob/master/bokeh/server/protocol/receiver.py#L21-L34
https://github.com/bokeh/bokeh/blob/master/bokeh/server/protocol/message.py#L80-L121

[...]

PR #6055, opened 2017-03-23
Use math library instead of numpy for nan/inf checks

mrocklin:
Math is much faster than NumPy for scalar checks

### DIFF
--- a/bokeh/util/serialization.py
+++ b/bokeh/util/serialization.py
@@ -11,6 +11,7 @@ performance and efficiency. The list of supported dtypes is:
 from __future__ import absolute_import
 
 import base64
+import math
 
 from six import iterkeys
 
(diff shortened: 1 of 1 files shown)


bryevdv:
@mrocklin Are you sending lists instead of arrays in your CDS? I would not actually expect you to hit this code path if you CDS columns are real arrays.

mrocklin:
I frequently stream small lists

bryevdv:
OK, fair enough. This change seems reasonable in any case I just wanted to check that nothing unexpected was happening.

mrocklin:
You're right though that I should consider switching to numpy arrays when lists get large.  I don't suppose you have any intuition on where the tipping point is here?

bryevdv:
Not offhand, it would be nice to measure. Roughly when the cost of JSON encoding exceeds the cost of base64 encoding. Though I would add a corollary that if you have non-finite values in your data it's probably always better to use real arrays, just so the nans are shipped as nans

mrocklin:
Oh, that's interesting.  NumPy arrays avoid the JSON serialization costs?

bryevdv:
> NumPy arrays avoid the JSON serialization costs?

Well, let me be precise. NumPy arrays are encoded using base64 and the resulting string is serialized in a JSON message. This turns out to scale *much* better, especially for > 1d arrays (JSON serialization of nested arrays is horrendously slow). This also allows non-finite values to be sent faithfully ("fake" handling them in lists can also be quite slow)

However , since the very lo level WS message protocol does allow for multi-part messages, there is still a possibility for true zero-copy sending of binary buffers (e.g. a numpy memory buffer) directly into a WS and then directly into a JS typed array. This is future work, though: https://github.com/bokeh/bokeh/issues/5984

PR #6056, opened 2017-03-23
Use cross-browser compatible offset()

### LINKED ISSUE #5986: bug with firefox hover tool coordinates
Here the sample code demonstrating a bug in Firefox hover tool coordinates.
(various late versions in osx) with various late versions of bokeh)
No faulty behaviour in Safari.

This happens when the browser window is smaller than the window itself (i.e. the window can be scrolled). In this case hover tool coordinates are wrong: 
-coordinates are shifted by the amount of scrolling

The attached image shows an example: the mouse location is represented by the star. Hover tooltip shows up at wrong coordinates, as can also be seen in the js console.

firefox_coo_bug

```
from bokeh.io import curdoc
from bokeh.models import ColumnDataSource,HoverTool,CustomJS

hover_tool = HoverTool()
hover_tool_callback = CustomJS(code="""console.log('x=',cb_data.geometry['x'],'y=',cb_data.geometry['y']);""")
hover_tool.callback = hover_tool_callback

p = figure(tools = [hover_tool])
cds = ColumnDataSource({'x':[1,2,3,4],'y':[1,2,3,4]})
p.line(x='x',y='y', source=cds)
curdoc().add_root℗
```

### LINKED ISSUE #6051: middle and bottom plot won't pan in y-dimension in custom/gears/gears.py demo on FireFox
OSX
Firefox 52
Bokeh 0.12.5rc1

Middle and bottom plot won't pan in y-dimension in custom/gears/gears.py demo on FireFox

mattpap:
fixes #6051
fixes #5986

### DIFF
--- a/bokehjs/src/coffee/core/dom.ts
+++ b/bokehjs/src/coffee/core/dom.ts
@@ -101,8 +101,8 @@ export function position(element: HTMLElement) {
 export function offset(element: HTMLElement) {
   const rect = element.getBoundingClientRect()
   return {
-    top:  rect.top  + document.body.scrollTop,
-    left: rect.left + document.body.scrollLeft,
+    top:  rect.top  + window.pageYOffset - document.documentElement.clientTop,
+    left: rect.left + window.pageXOffset - document.documentElement.clientLeft,
   }
 }
 


bryevdv:
I've asked @jsignell to do some quick local testing with this PR then we can get it merged.

bryevdv:
@mattpap is this the same issue https://github.com/bokeh/bokeh/issues/5986 ?

Issue #6059, opened 2017-03-23
Python event callback not called if no JS event callback is registered


kevinsa5:
I am using the latest github release, 0.12.5rc1, which has the new generalized event callback system.  If I register a javascript callback, then the corresponding python callback is called correctly, but if the javascript callback is omitted, the python callback isn't called.  Example code:

    from bokeh.io import curdoc
    from bokeh.plotting import figure
    from bokeh import events
    from bokeh.models import CustomJS  

    x = range(1,10)
    y = range(1,10)
    p = figure(tools="pan,wheel_zoom,zoom_in,zoom_out,reset")
    p.scatter(x, y)

    def callback(attrs):
        print "python callback"

    # uncomment this line:
    #p.js_on_event(events.Tap, CustomJS(code="",args={}))
    p.on_event(events.Tap,  callback)

    curdoc().add_root(p)

If you uncomment `p.js_on_event(events.Tap, CustomJS(code="",args={}))`, then the python callback runs as expected.  If the `js_on_event` line is left commented, then the python callback is not called.  

System info, can provide more if desired:

    webdashboard@dash:~$ bokeh --version
    0.12.5rc1
    webdashboard@dash:~$ python --version
    Python 2.7.12
    webdashboard@dash:~$ cat /etc/lsb-release 
    DISTRIB_ID=Ubuntu
    DISTRIB_RELEASE=16.04
    DISTRIB_CODENAME=xenial
    DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"

Browser:  Chrome `Version 56.0.2924.87 (64-bit)` on Windows 7.

bryevdv:
can reproduce, @jlstevens the `events_app.py` example only works because it also happens to have `js_on_event` calls.

jlstevens:
Thanks for reporting this issue! I'll be investigating it shortly.

jlstevens:
This issue should be fixed by the PR referenced above.

canavandl:
resolved by #6062

PR #6062, opened 2017-03-24
Event system fixes and improvements

jlstevens:
This PR fixes the issue described in #6059 and tweaks how JS callbacks are used.

The fix was to make sure the event managers ``subscribed_models`` attribute was being updated when Python callbacks are registered without needing a corresponding JS callback. The model now also listens to changes to the ``subscribed_events`` property instead of just listening to changes to the ``js_event_callbacks`` property.

In addition, the event instance is now passed as the ``cb_obj`` which is more consistent with how the Python callbacks are defined as well as with other usage of ``CustomJS`` callbacks.

### DIFF
--- a/bokeh/events.py
+++ b/bokeh/events.py
@@ -49,17 +49,20 @@ logger = logging.getLogger(__file__)
 
 from .util.future import with_metaclass
 
-EVENT_CLASSES = set()
+_CONCRETE_EVENT_CLASSES = dict()
 
 class _MetaEvent(type):
     ''' Metaclass used to keep track of all classes subclassed from Event.
 
-    All Event classes will be added to the EVENT_CLASSES set which is
-    used to decode event instances from JSON.
+    All Concrete Event classes (i.e. not "abstract" event base classes with
+    no ``event_name``) will be added to the _CONCRETE_EVENT_CLASSES set which
+    is used to decode event instances from JSON.
+
     '''
     def __new__(cls, clsname, bases, attrs):
         newclass = super(_MetaEvent, cls).__new__(cls, clsname, bases, attrs)
-        EVENT_CLASSES.add(newclass)
+        if newclass.event_name is not None:
+            _CONCRETE_EVENT_CLASSES[newclass.event_name] = newclass
         return newclass
 
 class Event(with_metaclass(_MetaEvent, object)):
@@ -76,20 +79,42 @@ class Event(with_metaclass(_MetaEvent, object)):
 
     @classmethod
     def decode_json(cls, dct):
-        ''' Custom json decoder for Events for use with the object_hook
-        argument of json.load or json.loads.
+        ''' Custom JSON decoder for Events.
+
+        Can be used as the ``object_hook`` argument of ``json.load`` or
+        ``json.loads``.
+
+        Args:
+            dct (dict) : a JSON dictionary to decode
+                The dictionary should have keys ``event_name`` and ``event_values``
+
+        Raises:
+            ValueError, if the event_name is unknown
+
+        Examples:
+
+            .. code-block:: python
+
+                >>> import json
+                >>> from bokeh.events import Event
+                >>> data = '{"event_name": "pan", "event_values" : {"model_id": 1, "x": 10, "y": 20, "sx": 200, "sy": 37}}'
+                >>> json.loads(data, object_hook=Event.decode_json)
+                <bokeh.events.Pan object at 0x1040f84a8>
+
         '''
         if not (('event_name' in dct) and ('event_values' in dct)):
             return dct
 
+        event_name = dct['event_name']
+
+        if event_name not in _CONCRETE_EVENT_CLASSES:
+            raise ValueError("Could not find appropriate Event class for event_name: %r" % event_name)
+
         event_values = dct['event_values']
-        for eventscls in EVENT_CLASSES:
-            if eventscls.event_name == dct['event_name']:
-                model_id = event_values.pop('model_id')
-                event = eventscls(model=None, **event_values)
-                event._model_id = model_id
-                return event
-        logger.warn("Could not find appropriate Event class for %r" % dct['event_name'])
+        model_id = event_values.pop('model_id')
+        event = _CONCRETE_EVENT_CLASSES[event_name](model=None, **event_values)
+        event._model_id = model_id
+        return event
 
 class ButtonClick(Event):
     ''' Announce a button click event on a Bokeh Button widget
--- a/bokeh/events.py
+++ b/bokeh/events.py
@@ -41,6 +41,10 @@ event object that triggered the callback.
 
     button.on_event(ButtonClick, callback)
 
+.. note ::
+    There is no throttling of events. Some events such as ``MouseMove``
+    may trigger at a very high rate.
+
 '''
 from __future__ import absolute_import
 
@@ -68,11 +72,20 @@ class _MetaEvent(type):
 class Event(with_metaclass(_MetaEvent, object)):
     ''' Base class for all Bokeh events.
 
+    This base class is not typically useful to instantiate on its own.
+
     '''
     _event_classes = []
     event_name = None
 
     def __init__(self, model):
+        ''' Create a new base event.
+
+        Args:
+
+            model (Model) : a Bokeh model to register event callbacks on
+
+        '''
         self._model_id = None
         if model is not None:
             self._model_id = model._id
@@ -117,7 +130,7 @@ class Event(with_metaclass(_MetaEvent, object)):
         return event
 
 class ButtonClick(Event):
-    ''' Announce a button click event on a Bokeh Button widget
+    ''' Announce a button click event on a Bokeh Button widget.
 
     '''
     event_name = 'button_click'
@@ -130,7 +143,7 @@ class ButtonClick(Event):
         super(ButtonClick, self).__init__(model=model)
 
 class PlotEvent(Event):
-    ''' PlotEvent is the baseclass for all events applicable to Plot models
+    ''' PlotEvent is the base class for all events applicable to Plot models.
 
     '''
 
@@ -144,11 +157,23 @@ class PlotEvent(Event):
 class LODStart(PlotEvent):
     ''' Announce the start of "interactive level-of-detail" mode on a plot.
 
+    During interactive actions such as panning or zooming, Bokeh can
+    optionally, temporarily draw a reduced set of the data, in order to
+    maintain high interactive rates. This is referred to as interactive
+    Level-of-Detail (LOD) mode. This event fires whenever a LOD mode
+    has just begun.
+
     '''
     event_name = 'lodstart'
 
 class LODEnd(PlotEvent):
     ''' Announce the end of "interactive level-of-detail" mode on a plot.
+
+    During interactive actions such as panning or zooming, Bokeh can
+    optionally, temporarily draw a reduced set of the data, in order to
+    maintain high interactive rates. This is referred to as interactive
+    Level-of-Detail (LOD) mode. This event fires whenever a LOD mode
+    has just ended.
 
     '''
     event_name = 'lodend'
@@ -156,10 +181,20 @@ class LODEnd(PlotEvent):
 class PointEvent(PlotEvent):
     ''' Base class for UI events associated with a specific (x,y) point.
 
+    Attributes:
+        sx (float) : x-coordinate of the event in *screen* space
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
+
+    Note that data space coordinates are relative to the default range, not
+    any extra ranges, and the the screen space origin is at the top left of
+    the HTML canvas.
+
     '''
     event_name = None
 
-    def __init__(self, model, sx=None,sy=None, x=None, y=None):
+    def __init__(self, model, sx=None, sy=None, x=None, y=None):
         self.sx = sx
         self.sy = sy
         self.x = x
@@ -171,11 +206,23 @@ class PointEvent(PlotEvent):
 class Tap(PointEvent):
     ''' Announce a tap or click event on a Bokeh plot.
 
+    Attributes:
+        sx (float) : x-coordinate of the event in *screen* space
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
+
     '''
     event_name = 'tap'
 
 class DoubleTap(PointEvent):
     ''' Announce a double-tap or double-click event on a Bokeh plot.
+
+    Attributes:
+        sx (float) : x-coordinate of the event in *screen* space
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
 
     '''
     event_name = 'doubletap'
@@ -183,11 +230,27 @@ class DoubleTap(PointEvent):
 class Press(PointEvent):
     ''' Announce a press event on a Bokeh plot.
 
+    Attributes:
+        sx (float) : x-coordinate of the event in *screen* space
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
+
     '''
     event_name = 'press'
 
 class MouseEnter(PointEvent):
     ''' Announce a mouse enter event onto a Bokeh plot.
+
+    Attributes:
+        sx (float) : x-coordinate of the event in *screen* space
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
+
+    .. note::
+        The enter event is generated when the mouse leaves the entire Plot
+        canvas, including any border padding and space for axes or legends.
 
     '''
     event_name = 'mouseenter'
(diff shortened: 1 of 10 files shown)


jlstevens:
Tests are passing except for the 'integration' test with what looks like a selenium error.

bryevdv:
This looks good but let's take the opportunity to add some tests and more docs. I can push some docs updates, @canavandl can you help with suggestions, etc for testing?

jlstevens:
I'm very happy to write tests - it is just that I don't know my way around the testing infrastructure on the JS side.

jlstevens:
@bryevdv It has been mentioned before but just to reiterate - I think embedding the output of ``js_events.py`` into a documentation page seems like a great way to kick off the docs!

If that demo could be a the top of the page, then I would be happy to write some other information about the events system underneath.

bryevdv:
@jlstevens right, I was asking @canavandl to help with information, advice etc. about JS unit tests or possibly integration tests since he he very familiar with both. 

So let's divid up docs this way: I will flesh out the docstrings/refdocs info. If you want to add the "js event watcher" to the docs, there's two things to do:

* add the script to here: https://github.com/bokeh/bokeh/tree/master/sphinx/source/docs/user_guide/examples

* use `bokeh-plot` sphinx directive to embed the plot. Maybe here? https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/interaction/callbacks.rst 

I think the organization of the "interactions" section could stand to be revisited, but let's keep things simple in this PR.

jlstevens:
@bryevdv Sounds good. And how to I trigger a docs rebuild and view the updated docs?

bryevdv:
@jlstevens you'll need to build them locally. Basically: install all the deps in `meta.yaml` then in `sphinx` dir, execute `make clean all serve`

bryevdv:
@jlstevens I have expanded the docstrings and also added the ref guide section, which was missing. 

I also took the chance to do some minor cleanup, most of which should be uncontroversial, except perhaps for one thing, which I will leave an inline comment about. In any case the commits are granular so it's easy enough to drop any specific change if necessary.

bokeh/events.py
```diff
@@ -7,90 +7,130 @@
...
+                <bokeh.events.Pan object at 0x1040f84a8>
+
         '''
         if not (('event_name' in dct) and ('event_values' in dct)):
             return dct
 
+        event_name = dct['event_name']
+
+        if event_name not in _CONCRETE_EVENT_CLASSES:
+            raise ValueError("Could not find appropriate Event class for event_name: %r" % event_name)
```
bryevdv: This is the change. If the class can't be found, an exception is now raised. Since this should only be used in the context of our low level code decoding protocol messages, this should really never happen. I'd prefer if it *does* happen for it to fail bigly.

bokeh/events.py
```diff
@@ -7,90 +7,130 @@
...
-from .util.future import with_metaclass
+
 import logging
 logger = logging.getLogger(__file__)
 
-EVENT_CLASSES = set()
-
+from .util.future import with_metaclass
 
+_CONCRETE_EVENT_CLASSES = dict()
```
bryevdv: I also changed this, in two ways: a dict instead of a set (makes looking up event names cleaner) and also it only stores concrete event types ("abstract" event classes with no event_name) do not get put in the index.

examples/howto/events_app.py
```diff
@@ -18,18 +18,19 @@ def display_event(div, attributes=[]):
...
+        args = [];
+        for ( i=0; i<attrs.length; i++ ) {
+            args.push(attrs[i] + '=' + Number(cb_obj[attrs[i]]).toFixed(2));
+        }
+        line = "'<span style=%r><b>" + cb_obj.event_name + "</b>(" + args.join(", ") + ")</span>\\n";
+        text = div.text.concat(line);
+        lines = text.split("\\n")
+        if ( lines.length > 35 ) { lines.shift(); }
+        div.text = lines.join("\\n");
+    """ % (attributes, style))
```
bryevdv: I realize this is a few lines longer but I find the mixed python/js string substitution substitution impenetrable, and a single JS block much preferable.

examples/howto/events_app.py
```diff
@@ -27,13 +27,19 @@ def display_event(div, attributes=[]):
...
 def print_event(attributes=[]):
     """
     Function that returns a Python callback to pretty print the events.
     """
     def python_callback(event):
-        print(event.pprint(attributes))
+        cls_name = event.__class__.__name__
+        attrs = ', '.join(['{attr}={val}'.format(attr=attr, val=event.__dict__[attr])
+                       for attr in attributes])
+        print('{cls_name}({attrs})'.format(cls_name=cls_name, attrs=attrs))
```
bryevdv: I"d actually like to have a generic object printing mechanism for Bokeh, but let's do that as a dedicated effort, and not add one-off and ad-hoc pprints all over. This was only used in the example so I moved it there.

bokeh/events.py
jlstevens: Agreed. This exception shouldn't occur but we definitely want to see it if it does!

bokeh/events.py
jlstevens: Both seem like sensible changes to me.

examples/howto/events_app.py
jlstevens: Much more readable, thanks!

examples/howto/events_app.py
jlstevens: One alternative might be to have a ``repr`` for events that includes the ``model_id`` and a ``str`` that doesn't. The idea them is that the ``repr`` has everything needed to re-create the event from the representation whereas the ``str`` is more readable (which means it could be used here). Does that seem reasonable?

canavandl:
I'm adding tests to cover the JS side of this in #6076 . I think it's possible to add some selenium tests to check the roundtripping of callbacks via bokeh server, but we need to add some machinery to support selenium + bokeh server. I'll try to get started later tonight.

canavandl:
There's an issue where the `on_event` callbacks aren't triggered using client_sessions. Apart from being broken, this is the route through which we'd add bokeh server testing (running a global Bokeh Server event loop fixture and updating the document within the tests).

```python

plot = figure(height=800, width=1000, tools="pan")
plot.rect(x=[1, 2], y=[1, 1], width=1, height=1)

def event_callback(event):
    print("EventCallback:Pan")

plot.on_event("pan", event_callback)

def property_callback(attr, old, new):
    print("PropertyCallback:RangeChange")

plot.x_range.on_change("start", property_callback)

doc = curdoc()
doc.add_root(plot)

session = push_session(doc)
session.show()
session.loop_until_closed()
```

ping @jlstevens for a fix
ping @mattpap because you have opinions about Events :)

examples/howto/events_app.py
mattpap: We already have a printing capability implemented as `HasProps.pretty()` or `HasProps.pprint()`.

examples/howto/events_app.py
```diff
@@ -18,22 +18,29 @@ def display_event(div, attributes=[]):
...
-    l1 = "text = {name} + '(' + {args} + ')' + \
-             div.text.replace('<b>JS Events (see console for Python events)</b>','');".format(name=name, args=args )
-    l2 = "div.text = '<b>JS Events (see console for Python events)</b><br><font size=\"0.5pt\">' + text.split('<br>',20).join('<br>')+ '</font>';"
-    return CustomJS(code=l1+l2, args={'div': div})
-
+    style = 'float:left;clear:left;font_size=0.5pt'
+    return CustomJS(args=dict(div=div), code="""
+        attrs = %s;
+        args = [];
+        for ( i=0; i<attrs.length; i++ ) {
```
mattpap: This isn't CoffeeScript. Variable declarations must be prefixed with `var`.

examples/howto/events_app.py
jlstevens: @bryevdv @mattpap  I see three options:

1. Make events inherit from Model, linking to a previous discussion we had.
2. I could easily support the same API on Event, implementing those two methods appropriately.
3. Leave it as it is for now.

examples/howto/events_app.py
jlstevens: Fixed in 194f019ed

jlstevens:
@canavandl 

How are you trying to trigger the pan event for testing? The only place the pan event is triggered is [in ui_events.coffee](https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/ui_events.coffee#L231). 

It is possible that maybe there is some place where the pan event should be triggered but isn't? I would only expect the pan event to fire in response to a corresponding hammer event as processed by ``UIEvents``. In other words, if``@_bokify_hammer`` is receiving the pan hammer event and the bokeh pan event isn't firing, then that would probably be a bug.

canavandl:
@jlstevens I'm trying to trigger the pan via selenium, but the above example is runnable manually. Does the pan event callback execute for you?

jlstevens:
@canavandl 

I get what looks like an IOError error when trying to use ``bokeh serve``:

```
raise IOError("Cannot push session document because we failed to connect to the server (to start the server, try the 'bokeh serve' command)") Traceback (most recent call last):
  File "/Users/jstevens/Desktop/development/bokeh/bokeh/application/handlers/code_runner.py", line 81, in run
  File "/Users/jstevens/Desktop/development/bokeh/foo.py", line 20, in <module>
    session = push_session(doc)
  File "/Users/jstevens/Desktop/development/bokeh/bokeh/client/session.py", line 86, in push_session
    session.push(document)
  File "/Users/jstevens/Desktop/development/bokeh/bokeh/client/session.py", line 298, in push
    raise IOError("Cannot push session document because we failed to connect to the server (to start the server, try the 'bokeh serve' command)")
IOError: Cannot push session document because we failed to connect to the server (to start the server, try the 'bokeh serve' command)
```

canavandl:
Are you running a `bokeh serve` process in one terminal and the example as a script in another?

jlstevens:
Thanks, now I see how this is meant to be used. I'll investigate!

jlstevens:
@canavandl I see the event message for pan being generated on the client side so it seems that event something is different about how ``bokeh.client.push_session`` is processing the websocket messages on the Python side. I'll keep investigating...

jlstevens:
I now think the problem is that the other message types ('PULL-DOC-REQ', 'PUSH-DOC' 'PATCH-DOC' 'SERVER-INFO-REQ') are handed in [_connection.py](https://github.com/bokeh/bokeh/blob/master/bokeh/client/_connection.py) but 'EVENT' currently isn't.

The client needs to be updated to handle 'EVENT' appropriately so that when you use ``bokeh.client.push_session``, the [event classmethod](https://github.com/bokeh/bokeh/blob/master/bokeh/server/session.py#L230) gets called appropriately on the server. I will now try to understand the client code a bit better - if anyone has any pointers on how to do this, that would help speed things up!

jlstevens:
After discussion with @canavandl, we came to the conclusion that we can test ``js_on_event`` with the proposed testing infrastructure but not ``on_event`` as that requires bokeh server to send events as well as receive them. In other words, it will be supported once Python->JS events are possible, such as the proposed 'busy' event.

examples/howto/js_events.py
```diff
@@ -16,15 +16,19 @@ def display_event(div, attributes=[]):
...
-               + " + ', '").format(attr=attr) for attr in attributes]
-    args = '+'.join(attrs) if attributes else repr('')
-    l1 = "text = {name} + '(' + {args} + ')' + \
-             div.text.replace('<b>Events</b>','');".format(name=name, args=args )
-    l2 = "div.text = '<b>Events</b><br><font size=\"0.5pt\">' + text.split('<br>',20).join('<br>')+ '</font>';"
-    return CustomJS(code=l1+l2, args={'div': div})
-
+    style = 'float:left;clear:left;font_size=0.5pt'
+    return CustomJS(args=dict(div=div), code="""
+        attrs = %s;
```
mattpap: This example has to be updated as well (`var`).

jlstevens: Yes, I fixed this in a local commit already. I'll push it now (3e23e7ca7)

jlstevens:
@bryevdv 

I've had a go adding a section to the user guide. As you pointed out earlier, the whole "interactions" section needs an overall revision but that is outside the scope of this PR. In particular, I had trouble coming up with a title for this section when there is already a 'CustomJS for Generic Events', making it harder to explain the difference between ``js_on_change`` and ``js_on_event``.

The main thing I've done with ``user_guide/examples/js_events.py`` is to try and compress it a bit vertically, without sacrificing the number of events types shown and hopefully without sacrificing too much readability.

It may come down to personal preference but I do prefer the look of the 60 line version for the purposes of the user guide than the 92 line version currently in ``examples/howto``. For one thing, this example would appear a *lot* longer than the other examples without this compression, making it look out of place. The best line saving trick I found was to simply loop over all events that have ``point_attributes`` instead of repeating those ``js_on_event`` lines over and over.

bokehjs/src/coffee/document.coffee
```diff
@@ -21,15 +21,16 @@ class EventManager
 
   send_event: (event) ->
     # Send message to Python via session
-    @session.send_event(event)
+    if @session
+      @session.send_event(event)
```
canavandl: the coffeescript convention is `@session?.send_event(event)`

bokehjs/src/coffee/document.coffee
```diff
@@ -21,15 +21,16 @@ class EventManager
...
-        if event.model_id != null and event.model_id != model_id
-            continue
-        model = @document._all_models[model_id]
-        if model?
-            model._process_event(event)
+      if event.model_id != null and event.model_id != model_id
+        continue
+      model = @document._all_models[model_id]
+      if model?
+        model._process_event(event)
```
canavandl: again, `model?.process_event(event)`

bokeh/tests/test_events.py
```diff
@@ -22,11 +22,11 @@ def __call__(self, event):
 
 def test_event_metaclass():
     # All events currently in the namespace should be in the EVENT_CLASSES set
-    assert len(all_events - events.EVENT_CLASSES) == 0
+    assert len(concrete_events - set(events._CONCRETE_EVENT_CLASSES.values())) == 0
 
 def test_common_decode_json():
-    for event_cls in events.EVENT_CLASSES:
-        if event_cls.event_name is None: continue # Skip abstract base class
+    for event_name, event_cls in events._CONCRETE_EVENT_CLASSES.items():
```
canavandl: It's generally not recommended to loop over iterables inside of pytest tests. The suggested thing to do is use the `@pytest.mark.parametrize` decorator. Idk if it's worth messing with now though

bokehjs/src/coffee/document.coffee
jlstevens: Fixed in 0870e05

bokehjs/src/coffee/document.coffee
jlstevens: Fixed in 0870e05

sphinx/source/docs/user_guide/interaction/callbacks.rst
```diff
@@ -77,10 +77,57 @@ is executed to update some data:
     :source-position: above
 
 
+
+CustomJS for Event classes
```
canavandl: I'd rename this to "CustomJS for User Interaction Events" and rename the above to "CustomJS for Model Property Change Events"

canavandl: actually - you'll have to divide the above section into "CustomJS for Generic Events" and "CustomJS for Model Property Change Events"

bokeh/events.py
```diff
@@ -146,43 +206,103 @@ def __init__(self, model, sx=None,sy=None, x=None, y=None):
...
+        sx (float) : y-coordinate of the event in *screen* space
+        x (float) : x-coordinate of the event in *data* space
+        y (float) : y-coordinate of the event in *data* space
+
+
+    By default, Bokeh plots do not capture scroll events, because prevents
+    the page from scrolling which can be annoying unless it was explicitly
+    requested. In order for ``MouseWheel`` events to fire, either the
+    ``WheelZoomTool`` must be active, or some custom extension must have
+    requested wheel UI events to propagate.
```
canavandl: Is this accurate? I don't see where in the code this is handled.

sphinx/source/docs/user_guide/interaction/callbacks.rst
```diff
@@ -77,10 +77,57 @@ is executed to update some data:
...
+    console.log('Tap event occured at x-position: ' + event.x)
+    """)
+
+    p = figure()
+    # execute a callback whenever the plot canvas is tapped
+    p.js_on_event('tap', callback)
+
+
+Although the event can be specified as a string such as ``'tap'`` above,
+it is often more appropriate to specify the first argument of
```
canavandl: Why is it often more appropriate? I'd remove or make clearer

sphinx/source/docs/user_guide/interaction/callbacks.rst
```diff
@@ -77,10 +77,57 @@ is executed to update some data:
     :source-position: above
 
 
+
+CustomJS for Event classes
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In addition to responding to Backbone events using ``js_on_change``,
```
canavandl: Maybe:

In addition to responding to property change events using ``js_on_change``, Bokeh allows 
CustomJS callbacks to be triggered by specific interaction events with the 
plot canvas, on button click events, and LOD events.

These event callbacks are defined on models using the ``js_on_event`` method,
with the callback receiving the event object as a locally defined ``cb_obj``
variable.

sphinx/source/docs/user_guide/interaction/callbacks.rst
```diff
@@ -77,10 +77,57 @@ is executed to update some data:
...
+    # execute a callback whenever the plot canvas is tapped
+    p.js_on_event('tap', callback)
+
+
+Although the event can be specified as a string such as ``'tap'`` above,
+it is often more appropriate to specify the first argument of
+``js_on_event`` using the event class. The full set of event classes can
+be imported from the ``bokeh.events`` module.
+
+The following code imports ``bokeh.events`` and registered all the
```
canavandl: imports/registered - different tense

sphinx/source/docs/user_guide/interaction/callbacks.rst
jlstevens: I'm looking for a way to convey that the first arguments can be strings, but that passing the event class is a more robust way of specifying things (which is why I think I would recommend this style).

We should probably decide whether we are happy seeing strings in ``js_on_event`` and ``on_event`` or whether we might prefer to always specify via event class. The only real advantage of using strings that I see is it helps avoid one import...

bokeh/events.py
canavandl: replaced with:

By default, Bokeh plots do not prevent default scroll events unless a WheelZoomTool or WheelPanTool is active. This may change in future releases.

bryevdv:
> After discussion with @canavandl, we came to the conclusion that we can test js_on_event with the proposed testing infrastructure but not on_event as that requires bokeh server to send events as well as receive them. In other words, it will be supported once Python->JS events are possible, such as the proposed 'busy' event.

just FYI there is a difference between this and `"busy"`. The difficultly with `"busy"` was that it was "on the document' instead of "on a model", which complicated matters because where do the callbacks live, if not on a model? I would expect making the current events work with `bokeh.client` to be somewhat simpler, because they (and the callbacks) are all associated with real models. But at this point agree, for the next release.

jlstevens:
@bryevdv 

I agree that the issue of where to put the callbacks (document vs model) complicates things when considering 'busy'. From what you said we can tackle this in two stages 1. communicating events defined on models from Python to JavaScript, hopefully solving the ``bokeh.client`` issue 2. figuring our where to keep the callbacks for events such as 'busy'. Looking forward to getting this in after the release!

@canavandl Thanks for implementing the doc changes and merging!

PR #6067, opened 2017-03-24
change default for relative_url to False

### LINKED ISSUE #6066: Change default for
Supplying relative URLs for Bokeh server resources only seems to be *strictly necessary* in a few specialized cases (e.g. behind nginx with path to port mapping). But making it the default requires explicitly turning this off in cases like:

https://github.com/bokeh/bokeh/blob/master/examples/howto/server_embed/flask_embed.py#L64

I think it is likely this case will be quite a bit more common, in which case the default behavior should be swapped. I propose to update the default value of `relative_urls` here:

https://github.com/bokeh/bokeh/blob/master/bokeh/embed.py#L473

to be `False`. A few examples will also need to be updated (simplified) 

ping @dhuebner @marami52 @bokeh/dev thoughts? This is new API so it really should be figured out now, before it goes in a release.

bryevdv:
- [x] issues: fixes #6066
- [x] tests added / passed

### DIFF
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -28,7 +28,7 @@ from .core.templates import (
 from .core.json_encoder import serialize_json
 from .document import Document, DEFAULT_TITLE
 from .model import Model
-from .resources import BaseResources, _SessionCoordinates
+from .resources import BaseResources, DEFAULT_SERVER_HTTP_URL, _SessionCoordinates
 from .util.deprecation import deprecated
 from .util.string import encode_utf8
 from .util.serialization import make_id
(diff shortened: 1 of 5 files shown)


bryevdv:
@bokeh/dev this changes the default to *not* be relative URLs for bo

 I also made the default value for `model=None` as well, which eliminates having to explicitly and annoyingly say `model=None` in the common case of calling `autoload_server` to embed a URL. 

I tested this with all the `server_embed` examples as well as with `example/app` demos (including with prefix and port set) as well as `models/server`. I think these changes are safe. If we agree to go with this I would add some additional user guide docs to this PR describing different situations and what to do in them explicitly.

bryevdv:
@marami52 I just had a happy realization. This change only affects embedding plots using `autoload_server` which I forgot that you are not using. For directly accessing a bokeh server  (including through a proxy) the resources still default to relative. So we may have stumbled upon a solution that works best for everyone all the time. If you can test with this PR to confirm that, that would be extremely helpful.

marami52:
@bryevdv good news! For sure, I will test it out over by today or tomorrow.

bryevdv:
OK I wanted to add more docs (and may later) but this will do for now.

@canavandl assuming @marami52 comes back with a positive report, then I would say this is ready to go, though some extra testing of all the various bokeh server example by someone other than me woudl probably be good too.

marami52:
@bryevdv just tested and works very well (version `0.12.5rc3+2.gc17ecc3`). Thanks!

canavandl:
LGTM

PR #6068, opened 2017-03-24
Style improvements to plotting/file/ examples

jsignell:
Before | After
screen shot 2017-03-24 at 6 10 38 pm screen shot 2017-03-27 at 10 40 02 am


bryevdv:
@jsignell there are probably some other examples that could use style improvements, if you are interested I'd suggest collecting several of them in one PR

bryevdv:
Could also add `vbar` and `hbar` to this example (seem to be missing)

jsignell:
@bryevdv ok, did you have particular ones in mind?

jsignell:
> Could also add vbar and hbar to this example (seem to be missing)

Also Image?

bryevdv:
Not offhand I'd just say feel free to clean up/improve as you see fit. I'd suggest concentrating on `plotting` dir tho

bryevdv:
> Also Image? 

I left it off this example just to keep the code for the example shorter/smaller. That said if there's way to add it in a couple of lines of code, +1

mattpap:
This example appears cropped on travis ci. I would prefer this to be changed to 4 x 5 grid to accommodate for current limitations of our examples' tests.

jsignell:
## Gmap changes:
Before | After
screen shot 2017-03-27 at 10 40 57 am screen shot 2017-03-27 at 10 40 34 am

## Jittering changes:

Before:
screen shot 2017-03-27 at 10 41 32 am


After:
screen shot 2017-03-27 at 10 33 33 am

jsignell:
## Line Select

Before | After
screen shot 2017-03-27 at 12 00 04 pmscreen shot 2017-03-27 at 12 00 59 pm

## Markers
Before:
screen shot 2017-03-27 at 12 56 33 pm screen shot 2017-03-27 at 12 56 41 pm

mattpap:
> Line Select

Difference in line widths should be restored, as it shows a feature.

jsignell:
screen shot 2017-03-27 at 2 12 56 pm

jsignell:
## Unemployment
Updated data to include through 2016
screen shot 2017-03-28 at 12 09 07 pm

## Unemployment polar
New example showing the unemployment data on polar coordinates
screen shot 2017-03-28 at 12 08 27 pm

mattpap:
> New example showing the unemployment data on polar coordinates

I'm not sure how to read this plot. I have only basic clues about what it what, because I know the original plot. I would argue that this is the wrong tool for the job, though having this kind of example would be nice. If unsure how to proceed, you may want to reach out to other employees. @jbednar, maybe you give some feedback?

jsignell:
> I'm not sure how to read this plot.

@mattpap, maybe I should add labels, I was thinking that hovertools would be enough, but obviously in a screenshot that doesn't come across. The innermost ring is 1948 and the outermost is 2016, the months are like numbers on a clock with January at 12 o'clock.

jbednar:
>  @jbednar, maybe you give some feedback?

Definitely difficult to interpret without a key or labels.  Seems like it at least needs a color key, for the z axis, but the radial and angular axes also need keys.  Yes, the hover information helps, but it would be much better to be direct.  Adding angular axis labels seems straightforward.  Adding radial axis labels is difficult due to overplotting.  

Using a polar plot in this case does make sense because the data is cyclic to some extent (seasonal differences), but it has a clear disadvantage of giving each year a different area on the page, which will affect how we interpret the plot.  Plus polar plots are just generally more difficult for most people, so in this case I'd stick with the Cartesian heatmap version; it's not bad (apart from the missing color key, at least when shown above).

jsignell:
Maybe I wasn't clear. I didn't mean for the polar plot to replace the cartesian one. I just thought it might be nice to have as an additional example. In any case, I added labels and colormaps. 
screen shot 2017-03-29 at 1 25 11 pm
screen shot 2017-03-29 at 1 02 20 pm

bryevdv:
Just a couple of super quick comments: 

* The jitter example would be improved if the columns were made different colors. Not necessarily unique, just each neighbor different. The use of jitter fixes the spatial overlap in the columns, but now the columns themselves are not really distinguishable from one another.

* Hate to be contrary, but I'm -1 on the polar example. As Jim originally said, polar plots are hard to interpret, and I don't think the polar transformation helps here. In particular it makes the months which are all otherwise the "same size" have wildly different visual sizes, which I think is not efficacious. Also people ask for polar examples so Think a polar example should be purpose-built and have it's own `polar.py` filename. I also think a more conventional polar example would be appropriate (e.g. radar chart or trajectory)

canavandl:
LGTM

PR #6069, opened 2017-03-25
fixed broken link to gapminder example

tommycarstensen:

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
@tommycarstensen thanks for the catch! The link was broken due to removing that one version of the gap minder example, but the other version doesn't actually demonstrate `file_html`, so I just edited to removed the link entirely.

Issue #6075, opened 2017-03-26
ResetTool does not trigger change events on Range1d


philippjfr:
When attaching a callback on the x- and y-range ``start`` and ``end`` atttributes you expect it to be triggered whenever the range changes, including when resetting the range with the reset tool. While this seems to be handled correctly for for ``DataRange1d``, when using explicitly defined ``Range1d`` models no event is triggered as the example below shows:

```python
from bokeh.plotting import figure, output_notebook
from bokeh.models import CustomJS
from bokeh.models import Range1d

output_notebook()

x_range = Range1d(start=0, end=10)
y_range = Range1d(start=0, end=10)

p = figure(
      # Commenting out this line will stop ResetTool triggering the callback 
      # x_range=x_range, y_range=y_range) 
)

cb = CustomJS(code='console.log("Range changed")')
p.x_range.js_on_change('start', cb)
p.x_range.js_on_change('end', cb)
p.y_range.js_on_change('start', cb)
p.y_range.js_on_change('end', cb)

p.circle([1, 2, 3, 4, 5], [2, 5, 8, 2, 7], size=10)

show(p)
```

This also applies to ``on_change`` events on the server.

philippjfr:
@bryevdv Sorry to bug you but it would be a shame if we couldn't fix this for 0.12.5 as it affects all datashader plots. If you could have a quick look at this to tell me what should be happening I'd be happy to work on a PR.

mattpap:
@philippjfr, look into `reset()` in `models/ranges/range1d.coffee`. There's `setv()` call with `silent: true`, which doesn't allow propagation of events. I'm not sure why this is the way it is. There may or may not be a reason.

philippjfr:
>There's setv() call with silent: true, which doesn't allow propagation of events.

I saw that, but DataRange1d appears to do the same thing but still triggers the change events so I think it's something to do with ``update_dataranges``, which only applies to DataRange1d models.

mattpap:
In `DataRange1d`'s case, `start` and `end` are managed elsewhere (in `update()`) and are not guarded with `silent: true`.

philippjfr:
Okay, that makes sense, maybe removing the silent true in Range1d reset is sufficient then, I tried that initially and it seemed to work, but I had assumed there was a reason for it.

canavandl:
The reason for it was to only trigger one change event per property change. But I see we're only triggering a "change" event and not a "change:start" or "change:end" event.

[...]

PR #6082, opened 2017-03-28
Ensure Range1d triggers start/end change events on reset

philippjfr:
After discussing https://github.com/bokeh/bokeh/issues/6075 with @canavandl we determined that for the time being it should be safe to trigger Range1d change events on reset. Unlike DataRange1d the ranges on Range1d are not recomputed by ``PlotCanvasView.update_dataranges``, which ends up triggering a "loud" change when it eventually call ``DataRange1d.update``. So this change simply ensures that both trigger the appropriate changes. In future it might be nice to move some of the logic on PlotCanvasView onto the ranges and handle triggering of events more consistently but for the time being this addresses the immediate issue.

WIP as I probably need to update some tests.

### DIFF
--- a/bokehjs/src/coffee/models/ranges/range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/range1d.coffee
@@ -40,6 +40,7 @@ export class Range1d extends Range
   }
 
   reset: () ->
-    @setv({start: @_initial_start, end: @_initial_end}, {silent: true})
+    # start/end change events not silenced to trigger callbacks
+    @setv({start: @_initial_start, end: @_initial_end})
     @_set_auto_bounds()
     @trigger('change')
--- a/bokehjs/src/coffee/models/ranges/range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/range1d.coffee
@@ -43,4 +43,3 @@ export class Range1d extends Range
     # start/end change events not silenced to trigger callbacks
     @setv({start: @_initial_start, end: @_initial_end})
     @_set_auto_bounds()
-    @trigger('change')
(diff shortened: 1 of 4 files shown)


mattpap:
@philippjfr, there is one trivial error to fix:
```
 1) range1d module reset should execute update callback once:
     AssertionError: expected false to be true
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:296:33)
      at .<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:286:25)
      at get (node_modules/chai/lib/chai/utils/overwriteProperty.js:50:37)
      at Context.<anonymous> (test/models/ranges/range1d.coffee:104:35)
```

philippjfr:
Perfect thanks, I was waiting on the test infrastructure to be working again, I'll make that fix shortly.

philippjfr:
After looking at the test (and making an incorrect fix) I now see the issue, if you attach a regular callback or presumably a generic ``change`` event (if there is such a thing) to the ``Range1d`` object, it will now be triggered twice. @canavandl Any suggestions?

bokehjs/src/coffee/models/ranges/range1d.coffee
```diff
@@ -40,6 +40,7 @@ export class Range1d extends Range
   }
 
   reset: () ->
-    @setv({start: @_initial_start, end: @_initial_end}, {silent: true})
+    # start/end change events not silenced to trigger callbacks
+    @setv({start: @_initial_start, end: @_initial_end})
     @_set_auto_bounds()
     @trigger('change')
```
mattpap: This line should have been removed. `@setv()` triggers `change` if `silent: false`.

philippjfr: Ah, yes, makes sense. Thanks!

philippjfr:
@mattpap @canavandl After removing the explicit ``@trigger('change')`` one of the Selenium tests is now indicating that a regular callback is no longer triggered. Shouldn't a change to the ``start`` and ``end`` also trigger the main callback?

PR #6086, opened 2017-03-30
Polar example

jsignell:
See discussion in #6068

screen shot 2017-03-30 at 11 23 35 am

### DIFF
--- a/examples/plotting/file/polar.py
+++ b/examples/plotting/file/polar.py
@@ -1,0 +1,101 @@
+from math import pi, sin, cos
+import pandas as pd
+
+from bokeh.io import show, output_file
+from bokeh.plotting import figure
+from bokeh.models import (
+    ColumnDataSource,
+    HoverTool,
+    LinearColorMapper,
+    BasicTicker,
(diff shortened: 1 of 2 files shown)


jsignell:
Static image
screen shot 2017-03-31 at 11 04 15 am

bryevdv:
I'm not really on board with these examples. The clock example is not bad, but I'm also not sure why anyone would bring in all of Bokeh just to draw a clock face. If it was heavily styled then it also becomes a demo of styling which is more useful (esp. rarer options like line-cap) 

I think there are some fundamental objections against the unemployment one, though. Basically all the criticisms against pie-charts hold, plus a few more. I'd much prefer to see a polar chart of the kind scientists might really use. A few random examples:

p1
p2
p3

bryevdv:
@jsignell do you expect to be able to put any additional time into this PR? If not that's understandable and fine, but I will close in that case.

bryevdv:
I'm going to go ahead and close this, we an work up new polar examples in a later PR

PR #6088, opened 2017-03-31
Fixing up examples/models/file/

jsignell:
## Before and After 

screen shot 2017-03-31 at 4 18 35 pmscreen shot 2017-03-31 at 4 17 33 pm
screen shot 2017-03-31 at 4 18 22 pmscreen shot 2017-03-31 at 4 17 54 pm
screen shot 2017-03-31 at 4 19 24 pmscreen shot 2017-03-31 at 4 17 03 pm

### DIFF
--- a/examples/models/file/image_url.py
+++ b/examples/models/file/image_url.py
@@ -33,7 +33,7 @@ plot.add_glyph(source, image1)
 image2 = ImageURL(url="url", x="x2", y="y2", w=20, h=20, anchor="top_left")
 plot.add_glyph(source, image2)
 
-image3 = ImageURL(url=dict(value=url), x=200, y=-100, anchor="bottom_right")
+image3 = ImageURL(url=dict(value=url), x=200, y=-100, w=50, h=50, angle=np.pi, anchor="bottom_right")
 plot.add_glyph(source, image3)
 
 xaxis = LinearAxis()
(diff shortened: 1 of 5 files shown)


examples/models/file/image_url.py
```diff
@@ -33,7 +33,7 @@
 image2 = ImageURL(url="url", x="x2", y="y2", w=20, h=20, anchor="top_left")
 plot.add_glyph(source, image2)
 
-image3 = ImageURL(url=dict(value=url), x=200, y=-100, anchor="bottom_right")
+image3 = ImageURL(url=dict(value=url), x=200, y=-100, w=50, h=50, angle=np.pi, anchor="bottom_right")
```
mattpap: This may be a regression in `ImageURL` glyph. Those arguments shouldn't be necessary. As I recall, this example did it on purpose, so that we test that width and height can be derived from an image.

jsignell: >This may be a regression in ImageURL glyph. 

I don't think so. `ImageURL`s that don't have 'w' and 'h' set don't show up in 0.12.4 either.
screen shot 2017-03-31 at 5 20 01 pm

Python version      :  3.5.3 
Bokeh version       :  0.12.4

mattpap: By regression I meant a long term one. This was broken on and off since 0.11.

mattpap: This is a bug in the definition of `ImageURL.{w, h}`, see #6095.

bryevdv: Agree we should fix the real problem and not demonstrate a temp workaround in an example.

bryevdv:
@Maggie-M is going to change back the image url examples and then we can merge

bryevdv:
The failed test is due to the tornado problem. This will be green when merged, so merging now.

PR #6100, opened 2017-04-05
Correct cursor handling

canavandl:
Only use crosshair cursor if inspector-type tool is active.

### DIFF
--- a/bokehjs/src/coffee/core/ui_events.coffee
+++ b/bokehjs/src/coffee/core/ui_events.coffee
@@ -130,13 +130,13 @@ export class UIEvents
         if view?
           if view.model.cursor?
             cursor = view.model.cursor()
-          if active_inspectors?
+          if active_inspectors
             # override event_type to cause inspectors to clear overlays
             event_type = "move:exit"
 
         # the event happened on the plot frame but off a renderer
         else if @_hit_test_frame(e.bokeh.sx, e.bokeh.sy)
-          if active_inspectors?
+          if active_inspectors
             cursor = "crosshair"
 
         @plot_view.set_cursor(cursor)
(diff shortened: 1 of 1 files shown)


canavandl:
It's unclear to me why this bug didn't cause these tests to fail: https://github.com/bokeh/bokeh/blob/master/bokehjs/test/common/ui_events.coffee#L53-L111

bryevdv:
I'd say make a separate issue to visit the test gap an get this merged for release

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -130,13 +130,13 @@ export class UIEvents
         if view?
           if view.model.cursor?
             cursor = view.model.cursor()
-          if active_inspectors?
```
mattpap: This name is deceptive, as it indicates we're dealing with a list of objects and not a boolean.

canavandl: That was the source of the bug. I'll change the name.

canavandl: @mattpap do you know why the tests didn't catch the bug though? It seems the test interpreter eval-ed that line differently than a browser

mattpap: Yeah, I reviewed this code and though that this was thoughtful change.

mattpap: `gulp test` tests something (slightly) different than bokeh.js bundle in a very different environment than a web browser, so that's not unexpected, though I'm not sure if this is the case here (I would need a closer look).

mattpap: Pretty much everything that touches DOM may work differently in our testing environment, because jsdom, which we use to emulate DOM, is only and API and doesn't provide any functionality. Though code may still work by chance or with aid of stubs.

mattpap: I'm hoping to run all of our tests in headless chrome in near future. It shouldn't bring any performance penalty and we will have a proper testing environment. Just we need to wait for support for OSX (along with existing for Linux).

PR #6110, opened 2017-04-07
Fix regression in 'get_best_ticker' handling empty data ranges

### LINKED ISSUE #6108: Error using DataRange1d with auto start/end and DatetimeAxis
Minimal example:

```python
from bokeh.plotting import show
from bokeh.models import Plot, DataRange1d, DatetimeAxis

x_range = DataRange1d()
y_range = DataRange1d()

p1 = Plot(x_range=x_range, y_range=y_range)

p1.add_layout(DatetimeAxis(), "below")

show(p1)
```

Error:

```javascript
composite_ticker.js:72 Uncaught TypeError: Cannot read property 'get_ticks_no_defaults' of undefined
    at DatetimeTicker.exports.CompositeTicker.CompositeTicker.get_ticks_no_defaults (composite_ticker.js:72)
    at DatetimeTicker.exports.Ticker.Ticker.get_ticks (ticker.js:19)
    at DatetimeAxis.exports.Axis.Axis._tick_coords (axis.js:397)
    at DatetimeAxis.tick_coords (axis.js:298)
    at DatetimeAxisView.exports.AxisView.AxisView._tick_label_extent (axis.js:181)
    at DatetimeAxisView.exports.AxisView.AxisView._get_size (axis.js:41)
    at Object.exports.update_constraints (side_panel.js:127)
    at PlotCanvasView.exports.PlotCanvasView.PlotCanvasView.update_constraints (plot_canvas.js:767)
    at PlotCanvasView.exports.PlotCanvasView.PlotCanvasView.render (plot_canvas.js:708)
    at later (throttle.js:17)
```

identified by @mrocklin

canavandl:
- [x] issues: fixes #6108
- [x] (hella) tests added / passed


bryevdv:
We have basically zero test coverage for axes on the BokehJS side. I'll add some (a start, at least) to this PR, unless you want to.

canavandl:
I'm looking into what add tests would require right now. I'll let you know.

bryevdv:
Is it safe to say this only affects plots that are *initially empty*? if so, there might be a workaround for @mrocklin (and I'd also be content with a quick dev build instead of an accelerated `0.12.6`)

canavandl:
AFAICT The bug only affects plots with a DataRange1d -type range, a CompositeTicker -type ticker (including a DatetimeTicker), and no initial data.

bryevdv:
OK that's my best understanding as well. Let's go with quick dev build then

canavandl:
> Will this fix allow me to stop making dummy data for empty plots in the future?

Yes

bryevdv:
side thought: it might be nice to add an optional hint to data ranges to allow for a suggested initial range in case there is no data.

bokehjs/src/coffee/models/tickers/composite_ticker.coffee
```diff
@@ -39,7 +39,7 @@ export class CompositeTicker extends ContinuousTicker
     # this can happen if the data isn't loaded yet, we just default to
     # the first scale
     best_index = argmin(errors)
-    if best_index == Infinity
+    if best_index is undefined
       return @tickers[0]
```
mattpap: I would prefer not to depend on this corner case in `argmin()` and check for empty `errors` and compute argmin only if `errors` is non-empty.

bryevdv: Agreed

canavandl: fixed in 8f6582fb74bfde3d00b040e1a402a35a174bf659

bokehjs/src/coffee/models/tickers/log_ticker.coffee
```diff
@@ -31,32 +31,31 @@ export class LogTicker extends AdaptiveTicker
...
+        minor_offsets = (i*minor_interval for i in [0...num_minor_ticks])
+        for x in minor_offsets[1..minor_offsets.length]
           minor_ticks.push(ticks[0]-x)
         for tick in ticks
           for x in minor_offsets
             minor_ticks.push(tick+x)
     else
-      startlog = Math.ceil(log_low)
-      endlog = Math.floor(log_high)
+      startlog = Math.ceil(log_low * 0.9999)
```
canavandl: I'm unsure about the wisdom of dealing with floating point errors in this way. The issue is for a range between 1 and 1000:
```coffeescript
data_low, data_high = 1, 1000
log_log, log_high = 1, 2.999999997
start_log, end_log = 1, 2
```
so we end up missing a right-hand tick. This adds a small fudge factor that's supposed to be larger than the floating point error, so the tick is present.

bryevdv: Does it *always* miss the right and tick? If so just compute the extra tick, unconditionally. If it's only sometimes, then something like this may be ok. Can we guarantee this doesn't sometimes generate extra ticks that are not correct?

bokehjs/test/models/tickers/log_ticker.coffee
```diff
@@ -0,0 +1,50 @@
...
+      ticker = new LogTicker({desired_num_ticks: 3, num_minor_ticks: 1})
+      ticks = ticker.get_ticks_no_defaults(1, 999, null, 3)
+      expect(ticks.major).to.be.deep.equal([ 1, 10, 100 ])
+      expect(ticks.minor).to.be.deep.equal([ 0.1, 1, 10, 100, 1000 ])
+
+    it "should do stuff on short scale", ->
+      ticker = new LogTicker({desired_num_ticks: 3, num_minor_ticks: 2})
+      ticks = ticker.get_ticks_no_defaults(1, 999, null, 3)
+      expect(ticks.major).to.be.deep.equal([ 1, 10, 100 ])
+      # 0.2 is weird, should be 0.5 i think
```
bryevdv: We should figure this out but doesn't have to happen here. Just make a separate issue that references this one.

bokehjs/src/coffee/models/tickers/log_ticker.coffee
canavandl: It doesn't always miss the tick, only when there's enough floating point error introduced in ``log_low = Math.log(data_low) / Math.log(base)``. And it's not limited to the right hand side.

We would incorrectly add a tick if the range is supposed to be 2.9999997 (for example), but I'm unsure how common of a case or how undesirable a 3.0 tick is in that case.

canavandl:
We're still missing tests for the MercantorTicker, the DatetimeTicker and all of it's components (DaysTicker, MonthsTicker, YearsTicker).

I'll open an issue to follow up on those (and the <1 LogTicker minor ticks issue) and mark this PR as ready. It resolves the initial issue (a CompositeTicker regression not handling empty ranges) and represents a notable increase in overall test coverage.

bokehjs/test/models/tickers/log_ticker.coffee
```diff
@@ -0,0 +1,50 @@
...
+      ticks = ticker.get_ticks_no_defaults(1, 1000, null, 4)
+      expect(ticks.major).to.be.deep.equal([ 1e0, 1e1, 1e2, 1e3 ])
+      expect(ticks.minor).to.be.deep.equal([ 1e-1, 1e0, 1e1, 1e2, 1e3, 1e4 ])
+
+    it "should have four major ticks and ten minor ticks for long range", ->
+      ticker = new LogTicker({num_minor_ticks: 2})
+      ticks = ticker.get_ticks_no_defaults(1, 1000, null, 4)
+      expect(ticks.major).to.be.deep.equal([ 1e0, 1e1, 1e2, 1e3 ])
+      # 0.2 is weird, should be 0.5 i think
+      expect(ticks.minor).to.be.deep.equal([ 0.2, 1e-1, 1e0, 5e0, 1e1, 5e1, 1e2, 5e2, 1e3, 5e3, 1e4 ])
```
bryevdv: Also for later: we need to figure out a way to not have minor ticks outside bounds when explicit axis bounds for ticks are present.

bryevdv:
LGTM @mattpap unless you have further comments I will merge in the next hour or so.

PR #6111, opened 2017-04-08
Fix typo in selection documentation.

shea-parkes:
Do this because it misled me for quite a while until I figured out what was going on.

I respect wanting matching issues, but I really couldn't bring myself to do so for such a small thing.  Feel free to reject of course.


bryevdv:
Thanks for the catch and fix @shea-parkes !

PR #6119, opened 2017-04-10
Update build_upload.py

### LINKED ISSUE #6104: twine upload in build/upload release script incorrect
Currently it does:
```
run("twine register upload dist/*")
```

I believe it needs to do instead simply:
```
run("twine upload dist/*")
```

Maggie-M:
issues: fixes #6104

### DIFF
--- a/scripts/build_upload.py
+++ b/scripts/build_upload.py
@@ -381,7 +381,7 @@ def upload_anaconda(token, dev):
 
 @upload_wrapper('pypi')
 def upload_pypi():
-    run("twine register upload dist/*")
+    run("twine upload dist/*")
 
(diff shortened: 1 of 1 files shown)


scripts/build_upload.py
```diff
@@ -381,7 +381,7 @@ def upload_anaconda(token, dev):
 
 @upload_wrapper('pypi')
 def upload_pypi():
-    run("twine register upload dist/*")
+    run("twine upload dist/*")
```
bryevdv: I just realized I think this can be made slightly better and more bulletproof if we use the exact actual version:
```
run("twine upload dist/bokeh-%s.tar.gz" % CONFIG.version)```

PR #6122, opened 2017-04-11
Canavandl/log zero handling

### LINKED ISSUE #6035: NA values correctly skipped in linear scale but not in log scale

canavandl:
- [x] issues: fixes #6035 
- [x] tests added / passed


bokehjs/test/models/glyphs/rect.coffee
```diff
@@ -297,14 +297,14 @@ describe "Rect", ->
...
-          data = {x: [60, 100, 140], y: [60, 100, 140]}
+          data = {x: [1, 10, 100, 1000], y: [1, 10, 100, 1000]}
           glyph_view = create_glyph_view(@glyph, data)
 
           @set_mappers(glyph_view, type="log")
           glyph_view.map_data()
 
-          result4 = glyph_view._hit_point(@geometry4)
-          result5 = glyph_view._hit_point(@geometry5)
+          result4 = glyph_view._hit_point({ vx: 200, vy: 200 })
```
canavandl: This test should probably be replaced with a selenium test. I understand that the intent is to assert that the _hit_point hits the correct glyph at a vx/vy (based on the mapper), but in practice it's difficult to figure out the vx/vy of the glyph. Currently the test passes, but doesn't check much of value.

bokehjs/src/coffee/models/mappers/log_mapper.coffee
```diff
@@ -20,11 +21,14 @@ export class LogMapper extends Model
     if inter_scale == 0
       value = 0
     else
-      value = (Math.log(x) - inter_offset) / inter_scale
-      if isNaN(value) or not isFinite(value)
-        value = 0
+      clamped_x = clamp(x, @source_range.start, @source_range.end)
```
canavandl: I'm not clamping the opposite path (`map_from_target` and `v_map_from_target`). I'm not sure if it's necessary to do that.

canavandl:
superceded by #6053

PR #6126, opened 2017-04-11
Use noUnusedLocals to discover unused imports

### LINKED ISSUE #6125: Use noUnusedLocals to discover unused imports, etc.

mattpap:
fixes #6125


bryevdv:
Great, LGTM

bokehjs/gulp/tasks/scripts.coffee
```diff
@@ -62,6 +62,7 @@ gulp.task "scripts:tsjs", ["scripts:coffee", "scripts:js", "scripts:ts"], () ->
         gutil.log("#{gutil.colors.red(real)}#{rest}")
         return
 
+      # XXX: can't enable "6133", because CS generates faulty code for closures
```
mattpap: This is annoying and significantly weakens this PR, but there's nothing we can do about this. Given code like this:
```
arr = [0...10]
```
CS generates code like this:
```
var arr, i, ref1;
arr = function() {
  var i, ref1;
}
```
(I skipped irrelevant code) which has obvious unused variables outside the closure.

bryevdv: We use coffee `1.10` it looks like the latest release is `1.12.5` is it also broken in newer releases?

mattpap: @bryevdv, I have 0.12.4 installed. `^1.10.0` version specification in `package.json` means versions `>= 1.10.0 && <2.0.0` are allowed, so most of the time we use most recent versions of our dependencies (as long as packages are updated on regular basis).

bryevdv: Bummer. Can we just (not now) replace all of these with a `range(0, 10)` that we write?

mattpap: Sure, that's always an option.

PR #6128, opened 2017-04-11
Add ability to use github api token

### LINKED ISSUE #6102: Have deploy script use Github API token to avoid rate-limiting issues
Currently if there are release issues and the deploy script has to be run more than once, the script fails due to GitHub API rate-limiting. This is because the deploy script queries a bunch of issues to make sure they're resolved.

It was suggested that API called with a GitHub token aren't rate-limited, so a solution would be to have the deploy check if a ``GITHUB_API_KEY`` env var is set and, if so, use it.

Maggie-M:
issues: fixes #6102 

I ran this six times without a token and got a 403 Forbidden message.

With the token, I was able to consistently get the following output

``` 
$ BOKEH_GITHUB_API_TOKEN=<token> python issues.py -c -p 0.12.4
INFO:root:finding issues...
WARNING:root:issue with no type label: #6109 Notebook embed example is not working
WARNING:root:closed issue with no reso label: #6109 Notebook embed example is not working
```

### DIFF
--- a/scripts/issues.py
+++ b/scripts/issues.py
@@ -9,13 +9,14 @@ import dateutil.tz
 import gzip
 import json
 import logging
+import os
 import pickle
 import sys
 
 from collections import OrderedDict
 from functools import partial
 from itertools import count, groupby
-from six.moves.urllib.request import urlopen
+from six.moves.urllib.request import urlopen, Request
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
We should add an encrypted API key on TravisCI and update the build and upload script to use it, but that can be separate work.

bryevdv:
@bokeh/dev anyone running `deploy.py` to make a full release can (should) set `BOKEH_GITHUB_API_TOKEN` now

PR #6129, opened 2017-04-12
Fix typo in ISSUE_TEMPLATE.md

shea-parkes:
Do this so it is a bit more polished.


canavandl:
mergng. thanks @shea-parkes

PR #6131, opened 2017-04-12
Add support for Toolbar.active_inspect property

### LINKED ISSUE #5599: Feature: Hover Tool Inspector Default
Need ability to set default behavior for hover tool inspector, currently it is always on. 
capture

Wanted to capture this, since I've seen it in group messages a handful of times now. I can see this being an issue where lots of meta is pushed to overlapping inspectors when exploring data. There is potential for the browser script to lag. @bryevdv Maybe combine with #5497 efforts?

canavandl:
Add support for Toolbar.active_inspect property, so that users can specify which inspect tools are active on plot initialization.

- [x] issues: fixes #5599 
- [x] tests added / passed

### DIFF
--- a/bokeh/plotting/figure.py
+++ b/bokeh/plotting/figure.py
@@ -9,6 +9,7 @@ from ..core.properties import Auto, Either, Enum, Float, Int, Seq, Instance, Str
 from ..core.enums import HorizontalLocation, VerticalLocation
 from ..models import Plot, Range, Title, Tool
 from ..models import glyphs, markers
+from ..models.tools import Drag, Inspection, Scroll, Tap
 from ..util.options import Options
 from ..util._plot_arg_helpers import _convert_responsive
 from .helpers import _get_range, _process_axis_and_grid, _process_tools_arg, _glyph_function, _process_active_tools
@@ -54,15 +55,20 @@ class FigureOptions(Options):
     A label for the y-axis.
     """)
 
-    active_drag = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_drag = Either(Auto, String, Instance(Drag), default="auto", help="""
     Which drag tool should initially be active.
     """)
 
-    active_scroll = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_inspect = Either(Auto, String, Instance(Inspection), Seq(Instance(Inspection)),
+    default="auto", help="""
+    Which drag tool should initially be active.
+    """)
+
+    active_scroll = Either(Auto, String, Instance(Scroll), default="auto", help="""
     Which scroll tool should initially be active.
     """)
 
-    active_tap = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_tap = Either(Auto, String, Instance(Tap), default="auto", help="""
     Which tap tool should initially be active.
     """)
 
@@ -123,7 +129,7 @@ class Figure(Plot):
 
         tool_objs, tool_map = _process_tools_arg(self, opts.tools)
         self.add_tools(*tool_objs)
-        _process_active_tools(self.toolbar, tool_map, opts.active_drag, opts.active_scroll, opts.active_tap)
+        _process_active_tools(self.toolbar, tool_map, opts.active_drag, opts.active_inspect, opts.active_scroll, opts.active_tap)
 
     annular_wedge = _glyph_function(glyphs.AnnularWedge)
 
--- a/bokeh/plotting/figure.py
+++ b/bokeh/plotting/figure.py
@@ -59,8 +59,7 @@ class FigureOptions(Options):
     Which drag tool should initially be active.
     """)
 
-    active_inspect = Either(Auto, String, Instance(Inspection), Seq(Instance(Inspection)),
-    default="auto", help="""
+    active_inspect = Either(Auto, String, Instance(Inspection), Seq(Instance(Inspection)), default="auto", help="""
     Which drag tool should initially be active.
     """)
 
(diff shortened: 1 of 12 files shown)


bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -73,10 +72,10 @@ export class UIEvents
     else if et == "move"
       logger.debug("Registering tool: #{type} for event '#{et}'")
       if tool_view._move_enter?
-        tool_view.listenTo(@, "move:enter", tool_view._move_enter)
-      tool_view.listenTo(@, "move", tool_view["_move"])
+        tool_view.listenTo(@, "move:enter:#{id}", tool_view._move_enter)
```
canavandl: Have to make the events be tool specific (like gesture tools), instead of generic. I can't think of any issues this introduces.

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -123,24 +122,25 @@ export class UIEvents
...
         # the event happened on the plot frame but off a renderer
         else if @_hit_test_frame(e.bokeh.sx, e.bokeh.sy)
-          if has_active_inspectors
+          if not isEmpty(active_inspectors)
             cursor = "crosshair"
 
         @plot_view.set_cursor(cursor)
-        @trigger(event_type, e)
+        for inspector in active_inspectors
+          @trigger("#{event_type}:#{inspector.id}", e)
```
canavandl: Is there going to be a timing issue related to looping over triggers like this?

bokehjs/src/coffee/models/tools/toolbar.coffee
```diff
@@ -40,6 +40,15 @@ export class Toolbar extends ToolbarBase
           @gestures[et].tools = @gestures[et].tools.concat([tool])
         @listenTo(tool, 'change:active', @_active_change.bind(tool))
 
+    if @active_inspect is 'auto'
+      # do nothing as all tools are active be default
```
canavandl: Is there a more idiomatic coffeescript way to implement a `pass` statement?

mattpap: You can put a stray `;` there. Also please don't use `is` operator.

canavandl: fixed

canavandl:
This PR continues inspector tool handling where multiple inspectors can be active simultaneously, as opposed to gesture tools.

Overall questions:
* Is this bad (having multiple inspectors active at the same time)?
* Is the documentation clear enough about this?

bryevdv:
> Is this bad (having multiple inspectors active at the same time)?

It's definitely reasonable to have/want multiple active inspectors at one time.

bokeh/plotting/figure.py
```diff
@@ -54,15 +55,20 @@ class FigureOptions(Options):
...
     """)
 
-    active_drag = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_drag = Either(Auto, String, Instance(Drag), default="auto", help="""
     Which drag tool should initially be active.
     """)
 
-    active_scroll = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_inspect = Either(Auto, String, Instance(Inspection), Seq(Instance(Inspection)),
+    default="auto", help="""
```
bryevdv: definitely prefer one line with `help="""` at the send that is the style of almost every (every?) other property

bokehjs/src/coffee/core/ui_events.coffee
bryevdv: AFAIK I know anything that responding to `"move:enter"` before will also respond to this. @mattpap can you confirm?

bokehjs/src/coffee/core/ui_events.coffee
bryevdv: There was a loop before, it was just implicit, so my guess is no

bryevdv:
A couple of tiny comments otherwise this looks great

bokehjs/src/coffee/core/ui_events.coffee
canavandl: I don't quite understand what you're asking. Are you saying that something that a listener on "move:enter" will also respond agnostically to "move:enter:id-hash"?

bryevdv: Yes, that is my understanding. I'm suggesting that anything that still responds to `"move:enter"` will still work with this change. But perhaps that's incorrect (in which case) everything that listened to `"move:enter"` has to be updated. (and maybe has been, I have not checked)

bokeh/plotting/gmap.py
```diff
@@ -44,15 +45,19 @@ class GMapFigureOptions(Options):
...
+    active_inspect = Either(Auto, String, Instance(Inspection), Seq(Instance(Inspection)), default="auto", help="""
+    Which drag tool should initially be active.
+    """)
+
+    active_scroll = Either(Auto, String, Instance(Scroll), default="auto", help="""
     Which scroll tool should initially be active.
     """)
 
-    active_tap = Either(Auto, String, Instance(Tool), default="auto", help="""
+    active_tap = Either(Auto, String, Instance(Tap), default="auto", help="""
```
bryevdv: Maybe we could factor out some mixing for `figure` and `gmap`

canavandl:
This PR has a problem with isolating the `move` events between the tools. Having only one hover tool activated causes both tooltips to appear. I'm apparently not groking something about how to events are listened for.

Here's an example to demostrate:

```python

p = figure(plot_width=400, plot_height=400, title="Click the Dots")
p.circle(x=[1,2,3,4,5], y=[2,5,8,2,7],
         fill_color=["navy", "orange", "olive", "firebrick", "gold"], size=20)

hover1 = HoverTool(tooltips = [
    ("index", "$index"),
    ("(x,y)", "($x, $y)"),
    ("radius", "@radius"),
])
hover2 = HoverTool(tooltips = [
    ("fill color", "$fill_color"),
])

p.add_tools(hover1, hover2)
p.toolbar.active_inspect = None

show(p)
```

canavandl:
ping @bryevdv @mattpap ^

canavandl:
Works now. I didn't understand how showing the tooltip is triggered.

PR #6136, opened 2017-04-13
Enable TS' strict options and fix errors

### LINKED ISSUE #5854: Enable typescript's strictNullChecks
This should be done sooner than later in the process of switching to typescript. Otherwise we will have a lot of changes to do.

mattpap:
fixes #5854

### DIFF
--- a/bokehjs/src/coffee/core/util/wheel.ts
+++ b/bokehjs/src/coffee/core/util/wheel.ts
@@ -7,7 +7,11 @@
  */
 
 function fontSize(element: Element): number | null {
-  return parseInt(getComputedStyle(element).fontSize, 10) || null
+  const value = getComputedStyle(element).fontSize
+  if (value != null)
+    return parseInt(value, 10)
+  else
+    return null
 }
 
 function lineHeight(element: HTMLElement): number {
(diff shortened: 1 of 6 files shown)


bokehjs/src/coffee/core/util/wheel.ts
```diff
@@ -7,7 +7,11 @@
  */
 
 function fontSize(element: Element): number | null {
-  return parseInt(getComputedStyle(element).fontSize, 10) || null
+  const value = getComputedStyle(element).fontSize
+  if (value != null)
+    return parseInt(value, 10)
+  else
+    return null
```
bryevdv: Personally I prefer
```
if (value == null)
  return null

return parseInt(value, 10)
```
slightly shorter, doesn't have unneeded branching (when there is a return), and I definitely find comparisons `==` easier to causally reason about then `!=`

bryevdv: I'm assuming there's not some weirdo JS reason a direct comparison against `null` will fail

bryevdv:
Left one small comment. Generally looks good, I'd like it if we take opportunities like this to add more JS tests.

bokehjs/src/coffee/core/util/wheel.ts
mattpap: My preference is to have the more likely branch first in such code.

bryevdv: In that case I'd still suggest and prefer
```
if (value != null)
  return parseInt(value, 10)

return null
```

bryevdv: @mattpap do you object to changing it to the above?

bryevdv:
Going to assume the small compromise change is OK, will merge when green

PR #6137, opened 2017-04-13
Make invisible glyph renderers not participate in hit testing

### LINKED ISSUE #6120: Hover responds to hidden data when legend.click_policy = 'hide'
Python 2.7.13, Bokeh 0.12.5, Windows 10x64, Chrome 57.0.2987.133 (64-bit) and Firefox 52.0.2 (32-bit)

A hover tooltip will respond to data that the user has hidden by clicking on legend items when legend.click_policy = 'hide'. The hover tooltip should not respond to data that is not visible.

image

Code:
```python
from bokeh.plotting import figure, output_file, show, ColumnDataSource

x1 = [1, 2, 3, 4]
y1 = [10, 12, 13, 14]
x2 = [1, 2, 3, 4]
y2 = [4, 5, 6, 6]

#Standard set of tools. We'll add the custom hover instance afterwards.
plotTools = 'box_zoom, wheel_zoom, pan, resize, tap, crosshair, reset, save'

p = figure(plot_width = 800, plot_height = 800,
           x_axis_label = 'X',
           y_axis_label = 'Y',
           tools = plotTools,            #Add the standard tools above.
           active_scroll = 'wheel_zoom') #Activate mouse wheel zooming.

hover = HoverTool(
    tooltips = [
        ('index', '$index'),
        ('X coord', '@x'),
        ('Y coord', '@y')
    ]
)

p.add_tools(hover) #Add the instance of HoverTool that we modified

data1 = ColumnDataSource(
        name = 'data 1',
        data = dict(
            x = x1,
            y = y1
        )
    )

data2 = ColumnDataSource(
        name = 'data 2',
        data = dict(
            x = x2,
            y = y2
        )
    )

output_file('hoverBug.html')

p.line('x', 'y', legend=data1.name, source=data1,
       line_width=2, line_color='red')
p.line('x', 'y', legend=data2.name, source=data2,
       line_width=2, line_color='blue')

p.legend.click_policy = 'hide'

show(p)
```

mattpap:
fixes #6120


bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -230,7 +230,10 @@ export class GlyphRendererView extends RendererView
     @glyph.draw_legend_for_index(ctx, x0, x1, y0, y1, index)
 
   hit_test: (geometry) ->
-    @glyph.hit_test(geometry)
+    if @model.visible
+      return @glyph.hit_test(geometry)
+    else
+      return null
```
bryevdv: I would suggest putting this in `Glyph.hit_test` instead, where it would be much easier to add a test for (and then suggest adding a unit test to make sure `null` is returned if `glyph.visible = false`)

mattpap: Glyphs don't have `visible` property anymore.

bryevdv: Right I missed `@model` / `@glyph` dichotomy. My point is this needs a test, to prevent regressions. This would be both fairly easy to break and fairly hard to notice given our existing tests.

bryevdv: An option (to make testing simpler) would be to have this just call a new `GlyphRenderer.hit_test`) that does what the code above does. That could be tested in the manner described.

bryevdv: @mattpap do you intend to add the test or would to prefer that I do it?

bryevdv:
It's probably a matter of preference whether muted glyphs participate in hit testing or not, but we can wait until someone raises an issue to do anything about making it configurable.

nbosc:
As op of  #6120 , I think the hover tooltip should not be visible if the data is hidden.

bryevdv:
@nbosc that's exactly what the PR did?

nbosc:
@bryevdv sorry for asking but in which version is it supposed to work? 12.7 installed

bryevdv:
The OP code (after removing deprecated `resize` tool) works as expected for me with the latest release `0.12.14`. I don't have the bandwidth to check any other releases.

nbosc:
Indeed it works. Is it possible to configure the same behaviour when legend.click_policy="mute" ?

bryevdv:
No, that was my comment above. Some people will certainly want it to still display when muted, and others maybe not. If you would like it to not display, then that would require new development so a new GH issue should be made.

PR #6140, opened 2017-04-13
Use tslib instead of per module generated helpers

### LINKED ISSUE #6139: Use tslib to reduce size of bokehjs
TypeScript (and CoffeeScript for that matter) has to generate some helper functions when compiling into ES5 (classes, decorators, iterators, async stuff). Currently those helpers are generated per file and as we have over 300 source files at this point, this will have a significant contribution. We can use `tslib` and `importHelpers` compiler option to import those helper functions instead of generating them. This reduces the size of helpers to import per file + 10 kB (unminified) for tslib.

mattpap:
fixes #6139

### DIFF
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -59,6 +59,7 @@
     "jsdom": "^9.8.3",
     "websocket":"^1.0.22",
     "typescript": "^2.3.0",
+    "tslib": "^1.6.0",
     "coffee-script": "^1.10.0",
     "detective": "^4.3.1",
     "jslint": "^0.10.3",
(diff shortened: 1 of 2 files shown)


bryevdv:
So the tests are pretty uniformly that
```
E         + "use strict";
E         + Object.defineProperty(exports, "__esModule", { value: true });
```
is now no longer in the output. If this is the new "correct" then it's easy enough to update the tests and I am happy to do that. But I am not sure if the test is correct and something else needs to be fixed. @mattpap can you advise?

mattpap:
As far as I can tell, those should still be there. I need to investigate this.

bryevdv:
OK, I will leave this to you to resolve then!

PR #6143, opened 2017-04-14
Properly apply --prefix to links on the app index page

### LINKED ISSUE #6080: prefix bug on "Active Bokeh Applications" page
I use a command like "bokeh serve --prefix=prefix app1.py app2.py" to serve my applications.
If I then access ip_address/prefix I can see a list of my applications running but the link points to ip_address/app1 instead of ip_address/prefix/app1.
I think this can be solved by changing  [bokeh/bokeh/server/views/root_handler.py](https://github.com/bokeh/bokeh/blob/master/bokeh/server/views/root_handler.py), line 32
from
`self.render("app_index.html", items=self.applications.keys())`
to
`self.render("app_index.html", items=[self.prefix+k for k in self.applications.keys()])`

regards.

bryevdv:
issues: fixes #6080

### DIFF
--- a/bokeh/server/views/app_index.html
+++ b/bokeh/server/views/app_index.html
@@ -41,7 +41,7 @@
        Current Running applications:
        <ul>
          {% for item in items %}
-           <li><a href={{ item }}>{{ item[1:] }}</a></li>
+           <li><a href={{ prefix }}{{ item }}>{{ item[1:] }}</a></li>
          {% end %}
        </ul>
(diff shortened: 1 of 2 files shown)


bryevdv:
Tested locally with various prefixes and without a prefix.

canavandl:
LGTM (tested locally with and without prefix)

canavandl:
Needs tests to prevent regression.

PR #6146, opened 2017-04-14
Allow callback add functions to accept args and kwargs

### LINKED ISSUE #6041: Support *args and **kwargs in add_next_tick_callback
This is a minor inconvenience, but it would be nice for `curdoc().add_next_tick_callback` to support `*args` and `**kwargs`

    doc.add_next_tick_callback(func, *args, **kwargs)

Currently the solution is to wrap these in a lambda:

    doc.add_next_tick_callback(lambda: func(*args, **kwargs))

This is low priority, I just thought I'd report it.  This is also what Tornado does:

```python
In [1]: from tornado.ioloop import IOLoop

In [2]: IOLoop.add_callback?
Signature: IOLoop.add_callback(self, callback, *args, **kwargs)
Docstring:
Calls the given callback on the next I/O loop iteration.

It is safe to call this method from any thread at any time,
except from a signal handler.  Note that this is the **only**
method in `IOLoop` that makes this thread-safety guarantee; all
other interaction with the `IOLoop` must be done from that
`IOLoop`'s thread.  `add_callback()` may be used to transfer
control from other threads to the `IOLoop`'s thread.

To add a callback from a signal handler, see
`add_callback_from_signal`.
File:      ~/Software/anaconda/lib/python3.6/site-packages/tornado/ioloop.py
Type:      function
```

bryevdv:
- [x] issues: fixes #6041
- [x] tests added / passed

OK I am not super happy with these tests, but I think they are the best that can be added for now absent a refactor of some of the plumbing. (Unless I am missing a different approach)

Ultimately I think some of these private classes can/should be made public so that testing them directly is less unreasonable.

### DIFF
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -196,12 +196,15 @@ class Document(object):
     def title(self, title):
         self._set_title(title)
 
-    def add_next_tick_callback(self, callback):
+    def add_next_tick_callback(self, callback, *args, **kwargs):
         ''' Add callback to be invoked once on the next tick of the event loop.
 
         Args:
             callback (callable) :
                 A callback function to execute on the next tick.
+
+        Any additional arguments or keyword arguments will be applied when
+        the callback is invoked.
 
         Returns:
             NextTickCallback :
@@ -214,9 +217,9 @@ class Document(object):
         '''
         from .server.callbacks import NextTickCallback
         cb = NextTickCallback(self, None)
-        return self._add_session_callback(cb, callback, one_shot=True)
+        return self._add_session_callback(cb, callback, one_shot=True, args=args, kwargs=kwargs)
 
-    def add_periodic_callback(self, callback, period_milliseconds):
+    def add_periodic_callback(self, callback, period_milliseconds, *args, **kwargs):
         ''' Add a callback to be invoked on a session periodically.
 
         Args:
@@ -225,6 +228,9 @@ class Document(object):
 
             period_milliseconds (int) :
                 Number of milliseconds between each callback execution.
+
+        Any additional arguments or keyword arguments will be applied when
+        the callback is invoked.
 
         Returns:
             PeriodicCallback :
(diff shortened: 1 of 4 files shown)


bokeh/server/tests/test_callbacks.py
```diff
@@ -258,3 +260,36 @@ def cleanup(callback):
...
+        assert list(group._group._periodic_callbacks.keys())[0]() == (10, 20, 30, 40)
+        group.remove_session_callback(cb)
+
+        d = Document()
+        cb = d.add_timeout_callback(func, 1, 10, 20, c=30, d=40)
+
+        group.add_session_callback(cb)
+        assert len(group._group._timeout_callbacks) == 1
+        assert list(group._group._timeout_callbacks.keys())[0]() == (10, 20, 30, 40)
+        group.remove_session_callback(cb)
```
bryevdv: This is the "worst" test. The other tests in `test_document` maintain that a proper `callback_obj` is created. This tries to test that, given a callback obj with args/kwargs, it ultimately stores a function with those parameters baked in.

bokeh/tests/test_document.py
```diff
@@ -543,6 +598,33 @@ def cb(): pass
...
+        assert callback._args == (10, 20)
+        assert callback._kwargs == dict(c=30, d=40)
+
+        callback = d.remove_next_tick_callback(cb)
+        assert len(d.session_callbacks) == 0
+        assert len(events) == 2
+        assert isinstance(events[0], SessionCallbackAdded)
+        assert isinstance(events[1], SessionCallbackRemoved)
+
+
```
bryevdv: These tests just ensure that a proper "callback obj" is created that reflects any args/kwargs passed in. But it's only about half the journey.

bryevdv:
@mrocklin I added tests, but haven't actually tried things out. Can you give this PR a quick spin?

bokeh/document.py
```diff
@@ -790,6 +799,12 @@ def _add_session_callback(self, callback_obj, callback, one_shot):
                 Whether the callback should immediately auto-remove itself
                 after one execution.
 
+            args (tuple) :
+                Arguments to apply when the callback is invoked
+
+            kwargs (tuple) :
+                Keywoard Arguments to apply when the callback is invoked
```
canavandl: spelling "Keyword"

PR #6154, opened 2017-04-18
Update to provide Tornado 4.5 compat

### LINKED ISSUE #6152: Bokeh Server is incompatible with tornado=4.5
There's a currently undiagnosed issue where Bokeh Server applications are incompatible with the newest tornado v4.5 release.

The server appears to start normally, but doesn't receive any requests or open any sessions when queried.

To reproduce:
* Install Bokeh v0.12.5 via conda or pip
* Run a Bokeh Server app example via ``bokeh serve app.py``
(below is an example app)
* Try (and fail) to open an application session at http://localhost:5006/line_on_off in a browser or via cURL. You'll see that any request just hangs.

contents of app.py
```python
""" Example demonstrating turning lines on and off - with bokeh server

"""

import numpy as np

from bokeh.io import curdoc
from bokeh.layouts import row
from bokeh.models import CheckboxGroup

p = figure()
props = dict(line_width=4, line_alpha=0.7)
x = np.linspace(0, 4 * np.pi, 100)
l0 = p.line(x, np.sin(x), color=Viridis3[0], legend="Line 0", **props)

checkbox = CheckboxGroup(labels=["Line 0", "Line 1", "Line 2"], active=[0, 1, 2], width=100)

def update(attr, old, new):
    l0.visible = 0 in checkbox.active
    l1.visible = 1 in checkbox.active
    l2.visible = 2 in checkbox.active

checkbox.on_change('active', update)

layout = row(checkbox, p)
curdoc().add_root(layout)
```

Platform: MacOS and Ubuntu 14.04
Bokeh = 0.12.5
Tornado = 4.5

bryevdv:
issues: fixes #6152

This PR updates functions in `bokeh.util.tornado` to no longer rely on old
behaviour where `gen.convert_yielded` would raise `gen.BadYieldedError` when
passed `None`. Additionally code was made safer against infinite loops by
explicitly checking that potential arguments to `convert_yielded` are
Futures to begin with.

Comments also edited for length and style

### DIFF
--- a/bokeh/util/tornado.py
+++ b/bokeh/util/tornado.py
@@ -1,6 +1,6 @@
-""" Internal utils related to Tornado
+''' Internal utilities related to Tornado
 
-"""
+'''
 from __future__ import absolute_import, print_function
 
 import logging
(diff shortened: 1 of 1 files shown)


bryevdv:
will submit a slightly different PR after comments from bdarnell

PR #6157, opened 2017-04-18
Optimized python execution fails due to missing __doc__ instantiation.

### LINKED ISSUE #6156: Optimized python execution fails due to missing __doc__ instantiation.
When you run the newest bokeh version in python 2.7 with the parameters --OO
the ```__doc__``` variables will not be set. They become None. 
"man python"  -> ```--OO Discard docstrings in addition to the -O optimizations```
This is a problem when you want to run bokeh on a deployment server, since exceptions will be raised. 

```
2017-04-18 16:51:23 ERROR    Error on request:
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/werkzeug/serving.py", line 193, in run_wsgi
    execute(self.server.app)
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/werkzeug/serving.py", line 181, in execute
    application_iter = app(environ, start_response)
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/werkzeug/wsgi.py", line 599, in __call__
    return self.app(environ, start_response)
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/werkzeug/contrib/lint.py", line 337, in __call__
    app_iter = self.app(environ, checking_start_response)
  File "/data/projects/python/server-monitoring/serverMonitoring/Main.py", line 40, in __call__
    response, endpoint = self.dispatch_request(request)
  File "/data/projects/python/server-monitoring/serverMonitoring/Main.py", line 31, in dispatch_request
    func = getattr(importlib.import_module(module[0]), module[1])
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/data/projects/python/server-monitoring/serverMonitoring/Handler/Delivery.py", line 20, in <module>
    from serverMonitoring.Handler.Dashboard.Widgets import ChartParameters, Chart
  File "/data/projects/python/server-monitoring/serverMonitoring/Handler/Dashboard/Widgets.py", line 20, in <module>
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/bokeh/embed.py", line 28, in <module>
    from .core.json_encoder import serialize_json
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/bokeh/core/json_encoder.py", line 47, in <module>
    from ..util.serialization import transform_series, transform_array
  File "/data/projects/python/server-monitoring/env/lib/python2.7/site-packages/bokeh/util/serialization.py", line 39, in <module>
    __doc__ = __doc__ % ("\n".join("* ``np." + str(x) + "``" for x in BINARY_ARRAY_TYPES))
TypeError: unsupported operand type(s) for %: 'NoneType' and 'str'
```

I suggest a simple if-clause whenever ```__doc__``` is accessed. 
Unfortunately this has to be fixed in various places.

mondainai:
I just added some simple if-clauses before using accessing ```__doc__```s.

[x] issues: fixes #6156

### DIFF
--- a/bokeh/command/__init__.py
+++ b/bokeh/command/__init__.py
@@ -10,6 +10,9 @@ def _build_docstring():
     from . import subcommands
 
     for cls in subcommands.all:
+        # running python with -OO will discard docstrings -> __doc__ is None
+        if __doc__ is None:
+            __doc__ = ''
         __doc__ += "%8s : %s\n" % (cls.name, cls.help)
 
 _build_docstring()
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -296,7 +296,7 @@ from bokeh.application import Application
 from bokeh.resources import DEFAULT_SERVER_PORT
 from bokeh.server.server import Server
 from bokeh.util.logconfig import basicConfig
-from bokeh.util.string import nice_join
+from bokeh.util.string import nice_join, format_docstring
 from bokeh.settings import settings
 
 from os import getpid
@@ -308,14 +308,12 @@ LOGLEVELS = ('debug', 'info', 'warning', 'error', 'critical')
 SESSION_ID_MODES = ('unsigned', 'signed', 'external-signed')
 DEFAULT_LOG_FORMAT = "%(asctime)s %(message)s"
 
-# running python with -OO will discard docstrings -> __doc__ is None
-if __doc__ is not None:
-    __doc__ = __doc__.format(
-        DEFAULT_PORT=DEFAULT_SERVER_PORT,
-        LOGLEVELS=nice_join(LOGLEVELS),
-        SESSION_ID_MODES=nice_join(SESSION_ID_MODES),
-        DEFAULT_LOG_FORMAT=DEFAULT_LOG_FORMAT
-    )
+__doc__ = format_docstring(__doc__,
+    DEFAULT_PORT=DEFAULT_SERVER_PORT,
+    LOGLEVELS=nice_join(LOGLEVELS),
+    SESSION_ID_MODES=nice_join(SESSION_ID_MODES),
+    DEFAULT_LOG_FORMAT=DEFAULT_LOG_FORMAT
+)
 
 def _fixup_deprecated_host_args(args):
     if args.host is not None and len(args.host) > 0:
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -308,12 +308,14 @@ LOGLEVELS = ('debug', 'info', 'warning', 'error', 'critical')
 SESSION_ID_MODES = ('unsigned', 'signed', 'external-signed')
 DEFAULT_LOG_FORMAT = "%(asctime)s %(message)s"
 
-__doc__ = __doc__.format(
-    DEFAULT_PORT=DEFAULT_SERVER_PORT,
-    LOGLEVELS=nice_join(LOGLEVELS),
-    SESSION_ID_MODES=nice_join(SESSION_ID_MODES),
-    DEFAULT_LOG_FORMAT=DEFAULT_LOG_FORMAT
-)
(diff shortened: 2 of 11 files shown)


bryevdv:
@mondainai thanks for the catch and the fix. To avoid regressions it would be nice to add some sort of test to this if possible. Is it sufficient to simply compile `.pyc` files with `-OO` to trigger the problem? Or do the compiled modules have to be executed to reveal the problem. If the former, we can probably find a way to adda step to our CI builds. 

It might be nice to have a small comment for each of these changes too, as to why they are necessary. 

Build failure was spurious, I have restarted it.

bryevdv:
@mondainai I believe the following code will maintain this against regressions. Can you add it to your PR as `tests/test_OO_import.py` ?
```

import os
import subprocess

def test_no_tornado_common():
    imports = []

    for path, dirs, files in os.walk("bokeh"):
        if "tests" in path: continue

        for file in files:
            if not file.endswith(".py"):continue
            if file.endswith("__main__.py"): continue

            if file.endswith("__init__.py"):
                mod = path.replace("/", ".")
            else:
                mod = path.replace("/", ".") + "." + file[:-3]

            imports.append("import " + mod)

    proc = subprocess.Popen(["python", "-OO", "-c", ";".join(imports)], stdout=subprocess.PIPE)

    proc.wait()
    if proc.returncode != 0:
        assert False

``` 

This test fails on my master branch (but passes if I remove the "-OO") so I think it's a good improvement, even if it may not be bulletproof. If you add this, and a short comment above each `if` you added, explaining why it is needed, then I think this will be good to go.

bryevdv:
I think the test failure was due to a spurious network issue, I've restarted it.

bryevdv:
@mondainai I am happy to merge this PR as-is once Travis is green, but another idea might be god if you are interested in making another small change. Basically just to wrap this up in function to make the code a little cleaner everywhere:
```
from bokeh.util.string import format_docstring

autoload_server.__doc__ = format_docstring(autoload_server.__doc__,  
                                           DEFAULT_SERVER_HTTP_URL)
```

Then`format_docstring` would be something like this:
```
def format_docstring(docstring, *args):
    ''' Safely format docstrings. 

    When Python is executed with the ``-OO`` option, doc strings are removed and 
    replaced the value ``None``. This function guards against applying the string 
    formatting options in that case. 

    Args:
        docstring (str or None) : The docstring to format, or ``None``
        args (tuple) : string formatting arguments for the docsring

    Returns:
        str or None

    '''
    if docstring is None: return None

    return docstring % args 
```

Then basic tests coudl be added to `ls bokeh/util/tests/test_string.py` as well. 

LMK what you think!

mondainai:
Hi @bryevdv,
A function is of course a nicer and better way to express something ;)
I'll try to commit a newer version later. There will be however two new functions (format_docstring, concatinate_docstring). I will also have to make some changes in the some string; I will try to get rid of those ```%```-formattings in those cases.
Cheers, 
Thomas

bryevdv:
@mondainai Are two functions really needed? String formatting is built into python (in a variety of ways, keeping `%`-formatting would keep things the most consistent with the rest of the codebase) so the only operation that needs to be guarded against is working against `None` by accident. I think one function will suffice?

mondainai:
@bryevdv 
You are right, only one function is really needed.
I made some small changes to some docs by using ```{variable_name}``` instead of ```%s```. I think this should be encouraged for the docstrings because it is more descriptive and docstrings in particular can get very long. In some instances I had to replace .format() notation anyway. 

I can change the .format() back to using % in format_docstring if you want to.

bryevdv:
@mondainai No I think the `.format` is good, I was just being set in my ways. :) 

I think the test failures were spurious, so I've restarted them. Otherwise LGTM will merge when green, thanks!

bryevdv:
Model PR @mondainai, thanks again!

PR #6159, opened 2017-04-18
Steven c howell/6158 muted color

### LINKED ISSUE #6158: For Interactive Legends, a note about `muted_color` would be helpful
For the [Interactive Legend guide](https://github.com/bokeh/bokeh/blob/d8fcffa1c472bf641517ef81698bb6d057cbd30f/sphinx/source/docs/user_guide/interaction/legends.rst#id7), in addition to the note about `muted_alpha`, it would be helpful to also demonstrate `muted_color`.

StevenCHowell:
This updates the example by including the visual property `muted_color`. 

- [ ] issues: fixes #6158

### DIFF
--- a/sphinx/source/docs/user_guide/examples/interaction_legend_mute.py
+++ b/sphinx/source/docs/user_guide/examples/interaction_legend_mute.py
@@ -11,7 +11,8 @@ for name, color in zip(['AAPL', 'IBM', 'MSFT', 'GOOG'], Spectral4):
         "http://ichart.yahoo.com/table.csv?s=%s&a=0&b=1&c=2005&d=0&e=1&f=2014" % name,
         parse_dates=['Date']
     )
-    p.line(df['Date'], df['Close'], line_width=2, color=color, alpha=0.8, muted_alpha=0.2, legend=name)
+    p.line(df['Date'], df['Close'], line_width=2, color=color, alpha=0.8,
+           muted_color=color, muted_alpha=0.2, legend=name)
 
 p.legend.location = "top_left"
 p.legend.click_policy="mute"
--- a/sphinx/source/docs/user_guide/interaction/legends.rst
+++ b/sphinx/source/docs/user_guide/interaction/legends.rst
@@ -30,11 +30,11 @@ Muting Glyphs
 Other times it is preferable for legend interaction to mute a glyph, instead
 of hiding it entirely. In this case, set ``click_policy`` property to
 ``"mute"``. Additionally, the visual properties of a "muted glyph" also
-need to be specified. In general this is does exactly in the same way as for
+need to be specified. In general this is does in exactly the same way as for
 :ref:`userguide_styling_selected_unselected_glyphs` or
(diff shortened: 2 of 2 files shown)


StevenCHowell:
@mattpap was there consideration to use a default value for `muted_color` and `muted_alpha`?  I would like to see the automatic defaults depend on the `color` and `alpha` kwargs, e.g., `muted_color=color` and `muted_alpha=0.2*alpha`.

StevenCHowell:
This is related to #5349.

bryevdv:
@StevenCHowell LGTM can you attach a screen shot for posterity?  

Re: defaults, I'm not sure color for muted glyphs can be vectorized. i.e. it is currently limited to drawing all instances of single glyph with a single color. This does not really show up with line, where there's only a single color to begin with. But it's an issue for, e.g. `circle` where individual circles can have different colors. In that case `color` might be a column name, and passing it to `muted` will not really work. So we chose grey as a reasonable default that will work.

StevenCHowell:
Here is the plot before muting any lines

screen shot 2017-04-18 at 2 21 14 pm

Without suppling the `muted_color` kwarg, muted lines will all be blueish
screen shot 2017-04-18 at 2 23 37 pm

Supplying the kwarg `muted_color=color`, similar to `color=color`, muted lines look like this
screen shot 2017-04-18 at 2 20 56 pm

StevenCHowell:
This is a minor change but I think as an example, demonstrating both the `muted_alpha` and `muted_color` kwargs shows users how to set all the applicable options.  

> But it's an issue for, e.g. circle where individual circles can have different colors. In that case color might be a column name of different colors, and passing it to muted as-is will not work. So we chose grey as a reasonable default that will work.

Perhaps it would be interesting to provide another example using something like circle where the glyphs have a range of colors, e.g., dataset 1 uses shade of red but when muted all show the same red color.  (if done appropriately, e.g., setting `muted_alpha=1`, this could actually be used to highlight a selection rather that hide it)

StevenCHowell:
> Re: defaults, I'm fairly sure color for muted glyphs cannot be vectorized. i.e. it is currently limited to drawing all instances of single glyph with a single color. Only the "main" glyph can be vectorized. This does not really show up or matter with line, where there's only a single color to begin with.

@bryevdv what about setting the default `muted_alpha` either to a fixed value, maybe 0.2, or a percentage of the unmuted `alpha`?  Perhaps the percentage could similarly not be vectorized, but a fixed value should work, right?  I think the current default is 1, essentially requiring the user to **always** set the `muted_alpha` value.

bryevdv:
@StevenCHowell so, maybe. A reason not to do that, is that if no `muted_` parameters are suppled, then no "muted glyph" is created. Setting a default would then mean every glyph gets an associated muted glyph created and serialized and used by BokehJS, regardless of whether a glyph is in a legend or even if there is no legend at all. Maybe that's OK, but it's something to consider, as it's not entirely without cost.

bryevdv:
Also probably worth making separate issue to discuss, as this is closed and will drop off my radar if the tab gets closed.

PR #6166, opened 2017-04-20
Add hover tool support for Segment

### LINKED ISSUE #6161: Extend hit testing and hover support to Segment
Currently possible to hit test and have hover tools for lines and multi-lines. Should implement hit-testing methods on the `Segment` glyph as well. Doing so will allow hover tools on graphs.

bryevdv:
issues: closes #6161

adds _hit_point and _hit_span to the Segment glyph so that it can work with hover tools.

Also updates HoverTool docstring to indicate segments is not functional with hover tools (as well as multi line which was erroneosly listed as unsupported)

Also makes a minor change to line hit-testing. From inspection the existing code would appear to fail on lines that "go backwards"

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/line.coffee
+++ b/bokehjs/src/coffee/models/glyphs/line.coffee
@@ -69,7 +69,7 @@ export class LineView extends XYGlyphView
       values = @_x
 
     for i in [0...values.length-1]
-      if values[i]<=val<=values[i+1]
+      if values[i]<=val<=values[i+1] or values[i+1]<=val<=values[i]
         result['0d'].glyph = this.model
         result['0d'].get_view = (() -> this).bind(this);
         result['0d'].flag = true  # backward compat
--- a/bokehjs/src/coffee/models/glyphs/segment.coffee
+++ b/bokehjs/src/coffee/models/glyphs/segment.coffee
@@ -36,8 +36,6 @@ export class SegmentView extends GlyphView
     x = @renderer.xmapper.map_from_target(vx, true)
     y = @renderer.ymapper.map_from_target(vy, true)
 
-    threshold = Math.max(2, @visuals.line.line_width.value() / 2)
-
     point =
       x: this.renderer.plot_view.canvas.vx_to_sx(vx)
       y: this.renderer.plot_view.canvas.vy_to_sy(vy)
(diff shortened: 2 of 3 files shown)


bokehjs/src/coffee/models/glyphs/line.coffee
```diff
@@ -69,7 +69,7 @@ export class LineView extends XYGlyphView
       values = @_x
 
     for i in [0...values.length-1]
-      if values[i]<=val<=values[i+1]
+      if values[i]<=val<=values[i+1] or values[i+1]<=val<=values[i]
```
bryevdv: So I'm not 100% sure on this, but it looks like the original code would miss hits on any line segment that "went backward" ? Worth validating before merging, though.

bryevdv:
After adding a hover tool to vector example:

screen shot 2017-04-19 at 19 44 32

screen shot 2017-04-19 at 19 44 49

bokehjs/src/coffee/models/glyphs/line.coffee
canavandl: You are correct that this change is necessary. Without it, it won't properly inspect lines going "backwards".

```python

p = figure(plot_width=400, plot_height=400, tools='')
p.line(x=[5,4,3,2,1], y=[4,2,4,5,3])
p.add_tools(HoverTool(mode='vline'))
show(p)
```

bokehjs/src/coffee/models/glyphs/segment.coffee
```diff
@@ -30,6 +31,65 @@ export class SegmentView extends GlyphView
...
+
+    hits = []
+
+    candidates = @index.indices({
+      minX: @renderer.xmapper.map_from_target(hr.min),
+      minY: @renderer.ymapper.map_from_target(vr.min),
+      maxX: @renderer.xmapper.map_from_target(hr.max),
+      maxY: @renderer.ymapper.map_from_target(vr.max)
+    })
+    console.log (candidates)
```
canavandl: `console.log` artifact

bryevdv: good catch, thanks.

bryevdv:
OK assuming I didn't leave some whitespace problem by editing in the GH web ui, this should be good merge when green

canavandl:
LGTM. Tested Line and Segment glyphs using the different HoverTool modes.

talbertc-usgs:
Thank you, I was just looking for this enhancement!

PR #6169, opened 2017-04-20
implement _hit_span for all markers

### LINKED ISSUE #5907: Hover Inspection does not work for all glyphs
Using [Hover Inspection](http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#hover-inspections) with glyphs other than a circle does not work, or at least square a diamond (I did not test further).  A simple test is to replace `circle` in the linked example with `square`.  

In this screenshot below, I copied the example, added circles vertically offset by 100, then replaced the original circles by squares.  You can see that when I hover, only the circles are highlighted.

image

As there is no error thrown in python, I wonder if the bug is on the js end.

bryevdv:
issues: fixes #5907

screen shot 2017-04-19 at 22 43 42

### DIFF
--- a/bokehjs/src/coffee/models/markers/marker.coffee
+++ b/bokehjs/src/coffee/models/markers/marker.coffee
@@ -80,6 +80,32 @@ export class MarkerView extends XYGlyphView
         hits.push([i, dist])
     return hittest.create_1d_hit_test_result(hits)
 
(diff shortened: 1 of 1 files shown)


canavandl:
LGTM. Tested with CircleX example and hline/vline modes.

PR #6176, opened 2017-04-20
be more strict about valid dataspec keys

### LINKED ISSUE #2680: Non-unit specs accept still allow `units` field
At some point `x` and `y` properties' types were changed to `NumberSpec` which is `data` units only. However, you can still set `screen` units. Obviously nothing will happen in bokehjs and no errors or warnings will be shown anywhere. This results in `iris_splom` example being broken, because text is displayed in `data` units, very far from actual data.

bryevdv:
- [x] issues: fixes #2680
- [x] tests added / passed

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1432,21 +1432,21 @@ class DataSpec(Either):
             color = ColorSpec(help="docs for color") # defaults to None
 
     '''
-    def __init__(self, typ, default, help=None):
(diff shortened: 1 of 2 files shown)


bryevdv:
OK I think this is small and I think should be uncontroversial, so I am going to merge. But @mattpap please take a quick look when you have a chance, if there are any concerns we can revert or make a follow on.

Issue #2352, opened 2015-05-29
Feature Request: Error Bar Plots


MaxNoe:
For scientists one hugely important feature would be plots with error bars und lower/upper limits like these:

http://matplotlib.org/examples/statistics/errorbar_limits.html

Also see this stackexchange question:
http://stackoverflow.com/questions/29166353/how-do-you-add-error-bars-to-bokeh-plots-in-python

MaxNoe:
Any news on this?

michaelaye:
:+1:

bsipocz:
:+1:  

And thanks to @MaxNoe for the SO answer, that will do until it gets built in.

bsipocz:
Now looking into it more closely, the SO solution is not working with interactive plots. Back to square one.

MaxNoe:
What's the problem?

bsipocz:
```
errorbar(pp, x='hjd', y='flux', xerr='fluxerr', source=s2)

​

source.callback = CustomJS(args=dict(s2=s2), code="""

        var inds = cb_obj.get('selected')['1d'].indices;

        var d1 = cb_obj.get('data');

​

        s2.get('data')['hjd'] = d1['hjd'][inds[0]]

        s2.get('data')['flux'] = d1['flux'][inds[0]]

        s2.get('data')['fluxerr'] = d1['fluxerr'][inds[0]]

        

        cb_obj.trigger('change');

        s2.trigger('change');

    """)



---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-38-524c4b97c72f> in <module>()
     30 #    pp.line(hjd[i], flux[i])
     31 
---> 32 errorbar(pp, x='hjd', y='flux', xerr='fluxerr', source=s2)
     33 
     34 source.callback = CustomJS(args=dict(s2=s2), code="""

<ipython-input-29-ce33a5d4f725> in errorbar(fig, x, y, xerr, yerr, color, source, point_kwargs, error_kwargs)
      7       x_err_y = []
      8       for px, py, err in zip(x, y, xerr):
----> 9           x_err_x.append((px - err, px + err))
     10           x_err_y.append((py, py))
     11       fig.multi_line(x_err_x, x_err_y, color=color, source=source, **error_kwargs)

TypeError: unsupported operand type(s) for -: 'str' and 'str'

```

MaxNoe:
You need to give it the data as something iterable.

[...]

PR #6177, opened 2017-04-20
FilledArea Annotation

canavandl:
Part of #2352

- [x] tests added / passed


canavandl:
initial output:

screen shot 2017-04-20 at 4 28 18 pm

```python
from bokeh.models import FilledArea, ColumnDataSource

p = figure(plot_width=400, plot_height=400, tools='pan')
    x=[1,2,3,4,5],
    y=[4,2,4,5,3],
    y_upper=[5,3,5,6,4],
    y_lower=[3,1,3,4,2]
))

p.line(x='x', y='y', source=source)

area = FilledArea(
    coords='x', dimension='height', lower='y_lower', upper='y_upper', source=source,
    level='underlay'
)
p.add_layout(area)

show(p)
```

bokeh/models/annotations.py
```diff
@@ -530,6 +530,52 @@ class BoxAnnotation(Annotation):
...
+    '''
+    lower = NumberSpec(help="""
+    The lower coordinates to locate the text anchors.
+    """)
+
+    upper = NumberSpec(help="""
+    The upper coordinates to locate the text anchors.
+    """)
+
+    coords = NumberSpec(help="""
```
canavandl: What's a better name for this property?

canavandl:
Questions:
* Do we need to support screen coordinates in addition to data coords?
* Are we happy with the upper/lower/dimension coordinate system instead of (lower_x, lower_y, upper_x, upper_y?)

bryevdv:
I think `upper` and `lower` is great. Do we want to add an orientation for filled areas across the horizontal dimension?

bokeh/models/annotations.py
bryevdv: `base`?

canavandl:
> I think upper and lower is great. Do we want to add an orientation for filled areas across the horizontal dimension?
If I'm understanding your question correctly, that's what the `dimension` property was for. Though I realize I should be using 'orientation' instead.

bokeh/models/annotations.py
canavandl: base! That'll work

bryevdv: is this necessary though? I think we just have users provide upper and lower coords as absolute coords, and if they want to tie them to some other base set of coords a transform will work for that.

bryevdv:
`Band` might also be a good but shorter name, just a thought.

bryevdv:
> If I'm understanding your question correctly, that's what the dimension property was for. Though I realize I should be using 'orientation' instead. 

Sorry I just missed that, `dimension` is fine.

bokeh/models/annotations.py
bryevdv: Don't have to worry about not but worth considering the case where upper and lower are different lengths... I think that implies multiple sources but may theres just a new different annotation for that.

canavandl:
I started writing writing JS unit tests, but there's so little testable surface in the view (most of the code is canvas drawing) and nothing in the model. I'll probably just add some selenium tests instead.

canavandl:
Current status:

screen shot 2017-04-21 at 10 33 05 am

I think it'd look better without the lines on the outer sides.

bryevdv:
> I think it'd look better without the lines on the outer sides.
  
Agree

bokehjs/src/coffee/models/annotations/band.coffee
```diff
@@ -0,0 +1,125 @@
...
+  set_data: (source) ->
+    super(source)
+    @visuals.warm_cache(source)
+
+  _map_data: () ->
+    x_mapper = @plot_view.frame.x_mappers[@model.x_range_name]
+    y_mapper = @plot_view.frame.y_mappers[@model.y_range_name]
+
+    non_base_mapper = if @model.dimension == "height" then y_mapper else x_mapper
+    base_mapper = if @model.dimension == "height" then x_mapper else y_mapper
```
bryevdv: maybe `base_mapper` and `limit_mapper`?

bokeh/models/annotations.py
```diff
@@ -530,6 +530,71 @@ class BoxAnnotation(Annotation):
...
+    ''' Render a filled area band along a dimension.
+
+    '''
+    lower = NumberSpec(help="""
+    The coordinates of the lower portion of the filled area band.
+    """)
+
+    lower_units = Enum(SpatialUnits, default='data', help="""
+    The unit type for the lower attribute. Interpreted as "data space" units
+    by default.
```
canavandl: Fixed

bokehjs/src/coffee/models/annotations/band.coffee
canavandl: Fixed

canavandl:
Here's the User Guide example:

screen shot 2017-04-24 at 10 58 07 am

canavandl:
Related to the above example, it seems useful to also implement a ``center`` or ``middle`` line (perhaps with unique styling), so that it's possible to indicate the rolling mean.

Does that seem reasonable?

bryevdv:
> Related to the above example, it seems useful to also implement a center or middle line (perhaps with unique styling), so that it's possible to indicate the rolling mean.

> Does that seem reasonable?

I don't think we should be in the business of computing means ourselves, and there are already other ways to draw lines so I don't see the benefit of extra complexity

canavandl:
> I don't think we should be in the business of computing means ourselves, and there are already other ways to draw lines so I don't see the benefit of extra complexity

That's fine. I'll mark this PR as ready then.

canavandl:
I forgot to add the TS Typings.

mattpap:
120 changed files? Something is pretty wrong with this PR.

canavandl:
Nah - I just messed up a rebase.

StevenCHowell:
@canavandl this looks nice!  I was just wanting to generate a plot like this and wondered if quad was still the best option.  Looking forward to using this!

PR #6178, opened 2017-04-20
use proper alignment for vertical ticks below a plot

### LINKED ISSUE #5914: Unable to right align tick labels
The following is a MWE based on @MatejKovacic's example from the gitter channel.

``` python
# coding: utf-8

from bokeh.models import PrintfTickFormatter

output_notebook()

x = range(0, 100)
# average temerature on both sensors
y = range(0, 100)
#temperature on single sensor only
y_red = y
y_yellow = y

p = figure(plot_width=900, 
           plot_height=400,
           tools="save",
           x_range=['really_long_label_{}'.format(i) for i in x], 
           y_range=(round(min(y),0)-1,round(max(y),0)+2), 
           x_axis_location='below')

p.vbar(x=x, width=0.7, bottom=0, top=y, color="firebrick")

p.xaxis[0].formatter = PrintfTickFormatter(format='%s')
p.xaxis.major_label_orientation = "vertical"

p.line(x, y_red, line_width=2, line_color="red")
p.circle(x, y_red, line_width=2, line_color="red", fill_color="red", size=4)

p.line(x, y_yellow, line_width=2, line_color="yellow")
p.circle(x, y_yellow, line_width=2, line_color="yellow", fill_color="yellow", size=4)

show(p)
```

bokeh_plot

It appears that one cause of the issue might be that [ctx.fillText](https://github.com/bokeh/bokeh/blob/3995f0f/bokehjs/src/coffee/models/axes/axis.coffee#L106) is not sufficient for right aligning text. It seems that the [textalign](https://www.w3schools.com/tags/canvas_textalign.asp) property of a `ctx` must be used.

Currently at this point, no matter that "padding" I use in the PrintfTickFormatter, I am unable move the labels below the xaxis.

@bryevdv Thoughts?

bryevdv:
issues: fixes #5914

with this fix:

screen shot 2017-04-20 at 14 35 18

### DIFF
--- a/bokehjs/src/coffee/core/layout/side_panel.coffee
+++ b/bokehjs/src/coffee/core/layout/side_panel.coffee
@@ -106,7 +106,7 @@ _align_lookup = {
     parallel   : CENTER
     normal     : LEFT
     horizontal : CENTER
-    vertical   : RIGHT
+    vertical   : LEFT
   left:
     justified  : CENTER
     parallel   : CENTER


canavandl:
We need to check that this doesn't impact the ColorBar, which uses the same alignment lookup table.

bryevdv:
> We need to check that this doesn't impact the ColorBar, which uses the same alignment lookup table.

Sure tho I'm pretty confident it was just broken in the same way for colorbar (but never noticed)

canavandl:
It appears (AFAICT) that ColorBar doesn't support rotating the tick labels, so my point is mute. So, the PR LGTM!

PR #6179, opened 2017-04-21
better validation for font size values

### LINKED ISSUE #4948: Plot title ignores sizing and appears italicized if number is provided but no units are specified
When setting `plot.title.text_font_size` to a number without providing any units (IE  `plot.title.text='20'`), the title defaults to 10pt and the text appears italicized regardless of the value set in plot.title.text_font_style (though it does respect the `plot.title.text_font_align` attribute). This behaviour doesn't match the behaviour if a non-number value is provided (which would cause the plot not to render).

If there is no default unit for sizing, I would expect that the behaviour would be the same as if I had provided a completely erroneous value (IE `plot.title.text_font_size='apples'`)

bryevdv:
- [x] issues: fixes #4948
- [x] tests added / passed

This PR doesn't add any validation on the JS side but should alert to bad values set from python. 

makes the assumption that column names do not begin with digits (or if they do, that `field` is used explicitly)

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1547,7 +1547,7 @@ class FontSizeSpec(DataSpec):
     https://drafts.csswg.org/css-values/#lengths
 
     '''
-    _font_size_re = re.compile("^[0-9]+(\.[0-9]+)?(%|em|ex|ch|ic|rem|vw|vh|vi|vb|vmin|vmax|cm|mm|q|in|pc|pt|px)$", re.I)
+    _font_size_re = re.compile(r"^[0-9]+(.[0-9]+)?(%|em|ex|ch|ic|rem|vw|vh|vi|vb|vmin|vmax|cm|mm|q|in|pc|pt|px)$", re.I)
 
     def __init__(self, default, help=None):
         super(FontSizeSpec, self).__init__(List(String), default=default, help=help)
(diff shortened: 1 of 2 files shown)


bryevdv:
This also is small and seems uncontroversial. I will merge in a few hours when it is green unless there is comment.

bokeh/core/properties.py
```diff
@@ -1559,6 +1559,15 @@ def prepare_value(self, cls, name, value):
             value = dict(value=value)
         return super(FontSizeSpec, self).prepare_value(cls, name, value)
 
+    def validate(self, value):
+        super(FontSizeSpec, self).validate(value)
+
+        if isinstance(value, dict) and 'value' in value:
+            value = value['value']
+
+        if isinstance(value, string_types) and value[0].isdigit() and self._font_size_re.match(value) is None:
```
mattpap: What about empty strings under the second clause?

mattpap: Also I'm not sure what's the need for that clause (an optimization?). The regex tests for the first character to be a digit.

bryevdv: fair enough, should empty string just raise `ValueError` as well?

mattpap: Yes.

PR #6180, opened 2017-04-21
stab at fixing selections + sorting in data table

### LINKED ISSUE #3564: Getting values of selected rows from a DataTable after reordering with sort
I'm trying to find a way to query the values on selected rows of a DataTable after it has been sorted by clicking on one of the column headers.

Before the table row order gets modified by the sort operation I could use the selected indexes available in a CustomJS callback to get the values contained in the selected rows from the ColumnDataSource used as input to the table.  However, when the table rows are reordered with a sort those selection indexes relate to the new order of the rows that are displayed. (i.e. the order oder of the source stays the same but the display order is different.) The reported selection indexes correctly match the display order (e.g. if the third row is selected a selection index of 2 will be reported).  However, I can't figure out how to get the values that are displayed on that row since it now isn't the third row of the input data source.

---

Bryan Van de ven commented:

Looking at the current code: 

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/data_table.coffee#L52-L86 

It _looks_ like it actually is trying to update the data source order. I'm not sure how I feel about that. I think maybe I hope that doesn't work, I think a better approach would be to maintain an explicit sorter to use to map to the original data. 

---

My initial reaction is that I agree with Bryan that actually reordering the data source might not be a great idea.  However a bit more investigation shows that the data source is not actually being reordered. Either that or the reordered one is not getting exposed to either the javascript in the callback or python executing in a jupyter notebook.  

I should add that I've been testing this on v.11RC2 but will try with the released one also.

### LINKED ISSUE #6115: descending order with TableColumn still ascending
Default_sort is ascending, I am using descending but either this is not working or I am using the wrong kind of marker for default_sort

from bokeh.layouts import row
        from bokeh.models import ColumnDataSource, WidgetBox, HoverTool
        from bokeh.models.widgets import  DataTable, TableColumn

        xu = [1,2,3]
        ttnwwrd = [90, 78, 150]
        dte = [100,200,300]
        #data = {'id':[1,2,3],'ttnwwrd':[3,4,6],}

        sourcet5 = ColumnDataSource(data=dict(x=xu,top=ttnwwrd,date=dte))
        p5 = figure(plot_width=500, plot_height=650, title="GJT Unique",
            tools = "pan,wheel_zoom,box_zoom,reset,hover,previewsave") #, x_axis_type = "datetime")
        # x = "x",
        glyph = VBar(x="x", top="top", bottom=-1, width=.5, fill_color="#b3de69")
        p5.add_glyph(sourcet5, glyph)
        people = ['un', 'deux','trois']
        sourcedt1 = ColumnDataSource(data=dict(people=people))

        columns1 = [
            TableColumn(field="people", title="People",  default_sort="descending"),

        ]
        dttbl1 = DataTable(source=sourcedt1, columns=columns1, width=220, height=130)
        wdgtdt1 = WidgetBox(dttbl1)
        hover = p5.select(dict(type=HoverTool))
        hover.tooltips = [("New words", "@top"),('date',' @date')]
        l5 = row(p5,wdgtdt1)
        show(l5)
I see:
screen shot 2017-04-09 at 9 53 17 am
 but I should see:
screen shot 2017-04-09 at 9 54 47 am

Thank you
Pierre

bryevdv:
- [x] issues: fixes #3564 
- [x] issues: fixes #6115
- [x] tests added / passed

This PR allows for sorting and selection on data tables to work properly together. Except for explicit cell editing, this PR changes the table implementation so that no modifications are made to the input data source (except by explicit user edits). Instead, an index is created (and stored outside the source data). The index is permuted by sorts, and is used to inform get, set, and selections operations.

### DIFF
--- a/bokehjs/src/coffee/models/widgets/data_table.coffee
+++ b/bokehjs/src/coffee/models/widgets/data_table.coffee
@@ -12,6 +12,8 @@ import {any} from "core/util/array"
 import {TableWidget} from "./table_widget"
 import {WidgetView} from "./widget"
 
+export DTINDEX_NAME = "__bkdt_internal_index__"
+
 wait_for_element = (el, fn) ->
   handler = () =>
     if $.contains(document.documentElement, el)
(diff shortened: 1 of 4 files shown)


bokehjs/src/coffee/models/widgets/data_table.coffee
```diff
@@ -77,10 +74,13 @@ export class DataProvider
...
-    records.sort (record1, record2) ->
+    old_index = @index.slice()
+
+    # TODO (bev) this sort is unstable, which is not great
+    @index.sort (i1, i2) ->
       for [field, sign] in cols
-        value1 = record1[field]
-        value2 = record2[field]
+        value1 = records[old_index.indexOf(i1)][field]
+        value2 = records[old_index.indexOf(i2)][field]
```
bryevdv: Slightly worried about the scaling performance of this inside the sort comparator. If there's a better way I'd love to know.

bryevdv:
@jms90h5 @dennisobrien @nicain @chanansh if any of you have the oportunity to test this it would be appreciated. 

@mattpap @canavandl this is done modulo some tests I'd like to add.

bryevdv:
@mattpap @canavandl this is ready for review

screenflow-4

bryevdv:
I'll merge this tomorrow unless there is further comment.

nicain:
Extremely exciting!  Do you have a sense of when this fix will be available via conda update?  I am ready to use this immediately.

bryevdv:
> Extremely exciting! Do you have a sense of when this fix will be available via conda update? I am ready to use this immediately. 

I'd like to publish a Dev Build tomorrow. Of course, all the usual caveats about dev builds apply. There are definitely some currently known issues, you may or may not run into them. We don't have a firm schedule yet but I am hoping for a `0.12.6` freeze by the end of the month.

nicain:
Fantastic.  Ill be ready and waiting.

PR #6183, opened 2017-04-22
Provide richer formatting options for tooltip fields

### LINKED ISSUE #1239: hover.tooltip convenience function for 'datetime' info
It would be nice to make it easy to create a "hover" tooltip template that could do some simple conversions/formatting on "standard" data point information: x, y, size, color, etc.  Maybe also some convenience conversions for `source` attributes of data points.

My particular scenario has to do with providing a hover tooltip which displays a human-readable time stamp for the data point.  The x-axis is configured as `datetime`, but when I display `$x` I get back a float (seconds since the epoch) which isn't very readable.  To get around this, I need to add a ColumnDataSource with a single field, like so:

``` python
    time=df.map(lambda x: x.strftime('%d-%m-%Y'))
))
```

and then I tell my tooltip to display `("time", "@time")`.  This is workable, but it would be nicer if I could just do:

``` python
("time", "`stftime('%m-%d-%Y', $x)`"
```

or something similar to say "do a client-side conversion of `$x` into something more human-readable"

### LINKED ISSUE #2595: Update HoverTool's documentation with field formatting, etc.
This should include:
- [ ] formatting options (just copy or link documentation from `NumeralTickFormatter`)
- [ ] ability to reference complex field names using curly brackets, e.g. `@{My Field}`
- [ ] support for HTML templating

Perhaps formatting options should be moved to common location and just referenced from all other locations (there are at least three: tick formatter, hover tool and data table).

### LINKED ISSUE #6079: Clarify docs for HoverTool
I recently read http://bokeh.pydata.org/en/0.12.4/docs/user_guide/tools.html#hovertool looking for the `hline` or `vline` behavior.  I was thrown off by the first sentence:

> The hover tool pops up a tooltip div whenever the cursor is over a glyph.

Reading down the page further to http://bokeh.pydata.org/en/0.12.4/docs/user_guide/tools.html#hover-tool my confusion was reinforced by:

> The hover tool displays informational tooltips whenever the cursor is directly over a glyph.

The words "_cursor is **directly** over_" led me to believe that I wasn't remembering things correctly.  So, I bugged Brian in the gitter:
image

A little tweaking of the HoverTool user-guide text would have kept me reading and eventually gotten me into the refguide where I would have seen `hline` and `vline`.

The limitations of `hline` and `vline` should be documented in the refguide as well.

Items:
- [ ] make link into refguide more prominently featured.  This may be as simple as having 'Reference' included in the link text.  When I saw the 'HoverTool' link, I didn't bother to click it because I thought it might just be a self reference to the user-guide.
- [ ] allude to the possibility of `hline` and `vline` in user-guide without being a full repeat of all the detial in the refguide
- [ ] clarify limitations of `hline` and `vline` in the refguide
- [ ] document the use of '~' in TOOLTIPS as in hover_span.py example.  It seems needed for spanning hover modes.
- [ ] update the 'Warning' block in [HoverTool Reference](http://bokeh.pydata.org/en/0.12.4/docs/reference/models/tools.html#bokeh.models.tools.HoverTool) to accurately reflect the glifs that are supported.

bryevdv:
Adds a `.formatters` property to HoverTool that maps CDS column names to a fomatter type (one of: "numeral", "datetime", "printf") When column values are looked up, any format is interpreted according to the `.formatters` dict. If no type is specified, "numeral" is assumed, which preserves existing behavior.

- [x] fixes #1239 
- [x] fixes #2595
- [x] fixes #6079 
- [x] tests added / passed

WIP PR, adding docs next


bryevdv:
@canavandl @timsnyder These changes should be ready. I'd appreciate a look-over, especially the new expanded hover tool docs.

bokeh/models/tools.py
```diff
@@ -724,16 +724,33 @@ class HoverTool(Inspection):
...
+
+    .. code-block:: python
+
+        "@foo{0,0.000}"    # formats 10000.1234 as: 10,000.123
+
+        "@foo{(.00)}"      # formats -10000.1234 as: (10000.123)
+
+        "@foo{($ 0.00 a)}" # formats 1230974 as: $ 1.23 m
+
+    Specifcying a format ``{safe}`` after a field name will override automatic
```
canavandl: *Specifying

bokeh/models/tools.py
```diff
@@ -724,16 +724,33 @@ class HoverTool(Inspection):
...
+    more precisely. Fields can be modified by appending a format specified to
+    the end in curly braces. Some examples are below.
+
+    .. code-block:: python
+
+        "@foo{0,0.000}"    # formats 10000.1234 as: 10,000.123
+
+        "@foo{(.00)}"      # formats -10000.1234 as: (10000.123)
+
+        "@foo{($ 0.00 a)}" # formats 1230974 as: $ 1.23 m
```
canavandl: If I have a field name with a space AND want to format a field, what do I do? (granted, I probably just shouldn't have a space in the name)

bryevdv: `"@{foo bar}{($ 0.00 a)}"` I can add a note that those things are compatible

bokehjs/test/core/util/templating.coffee
```diff
@@ -0,0 +1,93 @@
+{expect} = require "chai"
```
canavandl: thanks for finally getting this added. we've handle multiple regressions because we never tested it.

bryevdv: there's more to test but it's a start

bokeh/models/tools.py
bryevdv:
```
+Note that format specifications are also compatible with column names that
+have spaces. For example ```@{adjusted close}{($ 0.00 a)}`` applies a format
+to a column named "adjusted close".
```

PR #6189, opened 2017-04-24
Canavandl/export png

### LINKED ISSUE #538: Headless static (svg, png) image generation

canavandl:
Canavandl/export png

- [x] issues: fixes #538 
- [x] tests added / passed

Todo:
- [x] check that phantomjs is installed and raise dependency exception if not
- [x] better figure out layout width/height
- [x] check browser console for errors and warn
- [x] test that themes work
- [ ] better figure out render wait time
- [x] test that custom templates work

Questions:
* Do we allow for specified output size or just use document size?
* How will we support or warn about sizing modes?
* Is it ok to force inline resources?


bokeh/io.py
```diff
@@ -559,3 +563,92 @@ def _destroy_server(div_id):
...
+    driver = webdriver.PhantomJS()
+
+    # generate the html as tempfile
+    temp_html = tempfile.NamedTemporaryFile(suffix=".html").name
+    save(obj, filename=temp_html, resources=INLINE, title="")
+
+    if filename is None:
+        filename = _detect_filename("png")
+
+    driver.get("file:///" + temp_html)
```
canavandl: Is `get`-ing the html page like this going to be a problem. The screenshot tests serve the pages via a simple HTTPServer, but I don't know if that's necessary.

bokeh/io.py
```diff
@@ -559,3 +563,92 @@ def _destroy_server(div_id):
...
+
+    class element_to_start_resizing(object):
```
canavandl: This is only temporary until we add a Bokeh "render complete" message.

We need some sort of wait until the plot is correctly rendered before we take the screenshot.

bokeh/util/tests/test_dependencies.py
```diff
@@ -14,3 +14,11 @@ def test_required_fail():
     with pytest.raises(RuntimeError) as excinfo:
         dep.import_required('bleepbloop', 'nope')
     assert 'nope' in str(excinfo.value)
+
+def test_detect_phantomjs_success():
+    assert dep.detect_phantomjs() is not None
+
+def test_detect_phantomjs_fail(monkeypatch):
```
canavandl: The ``RuntimeError`` exception isn't being raised. For some reason the monkeypatching isn't working.

canavandl:
Current status:

There's currently some weird border lines around the plots which appear to be a CSS issue (they're not present using the Firefox driver), and we're clipping the right size of the output.

```python
from bokeh.io import export
from bokeh.layouts import column, row

p1 = figure(plot_width=500, plot_height=500)
p1.scatter(x=[1,2,3], y=[1,2,3])

p2 = figure(plot_width=500, plot_height=500)
p2.scatter(x=[1,2,3], y=[1,2,3])

p3 = figure(plot_width=1000, plot_height=500)
p3.scatter(x=[1,2,3], y=[1,2,3])

layout = column(row(p1, p2), row(p3))

export(layout, 'output.png')
```

yields:

output

bokeh/io.py
bryevdv: We may need to force inline resources for file access to work.

bokeh/util/dependencies.py
```diff
@@ -48,3 +50,18 @@ def import_required(mod_name, error_msg):
...
+    if settings.phantomjs_path() is not None:
+        phantomjs_path = settings.phantomjs_path()
+    else:
+        phantomjs_path = "phantomjs"
+
+    try:
+        proc = Popen([phantomjs_path, "--version"], stdout=PIPE, stderr=PIPE)
+        proc.wait()
+    except OSError:
+        raise RuntimeError('PhantomJS is needed to to export')
```
mattpap: `to to`

bokeh/io.py
canavandl: I am already forcing inline resources for the html tempfile that's being generated. I'll add this to my list of general questions though (is forcing inline resources bad?)

bokeh/tests/test_io.py
```diff
@@ -293,3 +295,13 @@ def _test_children_removed_from_root(layout_generator, children=None):
...
+
+def test__get_screenshot_as_png(monkeypatch):
+    def mockreturn(self):
+        return "PNG"
+    monkeypatch.setattr(selenium.webdriver.remote.webdriver.WebDriver, "get_screenshot_as_png", mockreturn)
+
+    layout = Plot(x_range=Range1d(), y_range=Range1d())
+
+    png = io._get_screenshot_as_png(layout)
+    assert png == "PNG"
```
canavandl: I'm unsold on this test. I think it might be better to actually create a tiny, empty Plot and compare ``png`` against an expected binary string. If the plot is empty (a white square), there shouldn't be any test failures due to changing defaults.

bokeh/io.py
canavandl: fixed

bokeh/util/tests/test_dependencies.py
canavandl: fixed

bokeh/util/dependencies.py
canavandl: fixed

bokeh/io.py
```diff
@@ -559,3 +563,69 @@ def _destroy_server(div_id):
...
+        """
+    driver.execute_script(script)
+
+    def is_bokeh_render_complete(driver):
+        return driver.execute_script('return window._bokeh_render_complete;')
+
+    try:
+        WebDriverWait(driver, 5, poll_frequency=0.1).until(is_bokeh_render_complete)
+    except TimeoutException:
+        pass
```
canavandl: should do something smarter in case of webdriver timeout (likely Bokeh image has failed to render)

damianavila:
Just a quick comment from a thousand of foots... so disregard if I am missing context... but do you really want spend resources on this solution when the chrome headless support is coming? Just curious :wink:

bryevdv:
@damianavila Yes, selenium can (and will) drive both but the folks driving this immediate work are interested meeting very minimal and specific requirements.

damianavila:
> @damianavila Yes, selenium can (and will) drive both but the folks driving this immediate work are interested meeting very minimal and specific requirements.

Got it. Thanks!

canavandl:
I'm going to mark this as "ready". It works on OS-X, Ubuntu and Windows 10.

There's some "unanswered"

* Do we allow for specified output size or just use document size?
  * Currently we generate a png that's the size of the layout. We could potentially resize the png based on height/width kwargs, but i'd be worried about changing the aspect ratio of things. There's also plenty of other tools to resize/crop a png, so maybe not dealing with size is fine.
* How will we support or warn about sizing modes?
  * The sizing modes "work", but the default PhantomJS window size is 350x300 so the plots end up pretty small. I think it might be best to add a warning to the ``export`` docstring to notify users that sizing modes probably won't behave well.
* Is it ok to force inline resources?

bryevdv:
@canavandl 

I think a warning in the docstring is fine for now. We should make separate issue to add some narrative documentation to the user's guide about this feature. I think it's OK to force inline resources in this context but that might also be worth mentioning in the docstring. 

What do users need to install to enable this feature? We should provide instructions for that as well (can go in the separate docs PR)

canavandl:
Users need: selenium, phantomjs (the binary) and pillow

bryevdv:
> Users need: selenium, phantomjs (the binary) and pillow

are those all available via conda (maybe the bokeh channel)? would could make a metapackage to install them, of so

bryevdv:
I'm def OK to merge this now as-is just make any necessary follow on issues to coincide.

PR #6192, opened 2017-04-24
Fix deprecated call to datetime64 for NP_EPOCH.

### LINKED ISSUE #6191: Fix deprecated datetime64 use for NP_EPOCH
From https://github.com/numpy/numpy/pull/6453 this causes a deprecation warning.
/cc @shoyer

chinasaur:
Fixes #6191
/cc @shoyer

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -49,7 +49,7 @@ from ..util.serialization import transform_series, transform_array
 pd = import_optional('pandas')
 rd = import_optional("dateutil.relativedelta")
 
-NP_EPOCH = np.datetime64('1970-01-01T00:00:00Z')
+NP_EPOCH = np.datetime64(0, 's')
 NP_MS_DELTA = np.timedelta64(1, 'ms')
 
 class BokehJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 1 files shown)


bokeh/core/json_encoder.py
```diff
@@ -49,7 +49,7 @@
 pd = import_optional('pandas')
 rd = import_optional("dateutil.relativedelta")
 
-NP_EPOCH = np.datetime64('1970-01-01T00:00:00Z')
+NP_EPOCH = np.datetime64(0, 's')
```
shoyer: Note that you probably want to set the appropriate time units here for the constant, though I think this gets cast properly in arithmetic. NumPy's default time resolution is `us`.

bryevdv: Bokeh serializes to all date time values to FP ms since epoch. Would that just be `np.datetime64(0, 'ms')`? Agreed arithmetic probably makes things work but no reason not to be precise.

chinasaur: The 's' was for consistency with the current string parsed result, but yeah 'ms' makes more sense.  Might be a good idea to make this _NP_EPOCH while we're at it?

bryevdv: Sure that seems reasonable too

bryevdv:
Thanks @chinasaur !

PR #6198, opened 2017-04-26
Remove confusing aliases from LayoutCanvas

### LINKED ISSUE #6197: Remove confusing aliases from LayoutCanvas
Follow up on PR #5976.

mattpap:
fixes #6197

### DIFF
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -104,7 +104,7 @@
     "jquery-ui": "~1.10.5",
 
     "numbro": "https://github.com/bokeh/numbro.git#e1b6c52",
-    "kiwi": "https://github.com/bokeh/kiwi.git#b26f6cb",
+    "kiwi": "https://github.com/bokeh/kiwi.git#3091d4d",
     "root-require": "^0.3.1"
   },
   "scripts": {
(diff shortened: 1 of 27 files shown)


bryevdv:
Should `_left` etc just be made public? Hate to see `foo._private.value` all over the place.

mattpap:
> Should _left etc just be made public?

Currently it's not possible due to confusion between `width` vs. `_width` in multiple places. Member starting with `_` are layout variables. When layout variables are moved to views, there will be clear separation between those.

bryevdv:
OK LGTM then

mattpap:
Also, `_value` will be soon replaced with `value`, but I have to clean up kiwi and use getters and setters across the code base.

mattpap:
@bryevdv, I went ahead and renamed `_value` to `value`. That's all changes for this PR.

mattpap:
I've been seeing this error lately:
```
Exception occurred:
  File "/home/travis/miniconda/lib/python3.4/ssl.py", line 620, in read
    v = self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out
```
It's from `flake_docs` job. Restarting the job helps.

bryevdv:
> It's from flake_docs job. Restarting the job helps. 

The sphinx GH extensions like `:bokeh-pull:` currently try to check that the link is valid and reachable. We can probably relax this.

PR #6202, opened 2017-04-27
Fix selections on plots with multiple glyph renderers

### LINKED ISSUE #2136: Selections only work on the last of two series in one plot

### LINKED ISSUE #5044: Shift selection in linked brushing plots
I expected to have shift selection in between two linked plots to select the union in the shared common column data source. For instance in the linked brushing example in the docs  http://bokeh.pydata.org/en/latest/docs/user_guide/interaction/linking.html#userguide-interaction-linking
I assume that selecting points in the left plot, holding shift, and the select some points in the right plot should lead to the union selection in the source. It looks like the individual plots does union within itself ?

clairetang6:
This PR fixes the issue where a selection made on a plot containing multiple renderers that share the same data source only works with the last renderer added to the plot. The problem is that the selection manager writes over its data sources `.selected` property on every select call, but the select tool can make multiple select calls per selection event because the tools loop through the renderers and call select for each renderer. 

This PR fixes the issue by changing the `renderer_view` parameter for the select method on the selection manager to `renderer_views` so that multiple renderers (which are associated with that data source/selection manager) can be passed in together for one selection event. So then, each select tool groups its renderers by the data sources that they share and calls select for each group of renderers. 

Also, I removed the selection manager's `selectors` dict (which mapped renderers to selectors) in favor of a single `selector` because:

1. the `selected` rows of a column data source should be shared across all glyph renderers that use it.
2. each cds is associated with a selection manager in a one-to-one way.

This PR additionally fixes #5044, so that you can now shift-select to append to the selection across different plots when they have renderers that share the same data source.

- [x] issues: fixes #2136, #5044
- [x] tests added / passed

selection

```python

output_file("selection_multiple_renderers.html")

x = list(range(-20, 21))
y0 = [abs(xx) for xx in x]
y1 = [xx**2 for xx in x]
y2 = range(len(x))

source = ColumnDataSource(data=dict(x=x, y0=y0, y1=y1, y2=y2))

TOOLS = "box_select,lasso_select,help"

left = figure(tools=TOOLS, plot_width=300, plot_height=300, title=None)
left.circle('x', 'y0', source=source)
left.circle('x', 'y2', source=source, color='red')

right = figure(tools=TOOLS, plot_width=300, plot_height=300, title=None)
right.circle('x', 'y1', source=source)

p = gridplot([[left, right]])

show(p)
```


bokehjs/src/coffee/core/hittest.coffee
```diff
@@ -19,6 +19,19 @@ export point_in_poly = (x, y, px, py) ->
 
 nullreturner = () -> null  # stub function shared by all hittests by default
 
+merge = (obj1, obj2) ->
```
mattpap: @clairetang6, can you move this to `core/util/object.ts` and add tests?

mattpap: The signature should be `function merge(obj1: object, obj2: object): object`.

clairetang6: Ok, will do.

bryevdv:
@clairetang6 FYI a merge/rebase on master will fix up the current test failures

bokehjs/src/coffee/core/util/object.ts
```diff
@@ -27,6 +29,21 @@ export function clone<T>(obj: T): T {
   return extend({}, obj)
 }
 
+export function merge<T>(obj1: {[key: string] : Array<T>}, obj2: {[key: string]: Array<T>}): object {
```
clairetang6: @mattpap I tried the type signature with obj1 and obj2 being objects, but there was some sort of warning. This seems to have fixed it, but I'm not certain it's the right thing to do. Also, I don't know if there is common notion of what "merge" should mean for objects. I guess this function has to take objects with arrays as values and unions those array values for two objects.

bryevdv: > Also, I don't know if there is common notion of what "merge" should mean for objects.

I was actually going to suggest it *not* go in `util/objects` for this reason (tho I think tests for it are good)

bokehjs/src/coffee/core/util/object.ts
```diff
@@ -27,6 +29,21 @@ export function clone<T>(obj: T): T {
...
+  let result: {[key: string]: Array<T>} = Object.create(null);
+
+  let keys = concat([Object.keys(obj1), Object.keys(obj2)])
+
+  for (let i = 0; i < keys.length; i++) {
+    let key = keys[i]
```
mattpap: `for (const key of keys)`

bokehjs/src/coffee/core/util/object.ts
```diff
@@ -27,6 +29,21 @@ export function clone<T>(obj: T): T {
   return extend({}, obj)
 }
 
+export function merge<T>(obj1: {[key: string] : Array<T>}, obj2: {[key: string]: Array<T>}): object {
+  let result: {[key: string]: Array<T>} = Object.create(null);
```
mattpap: `const`. The same elsewhere.

bokehjs/src/coffee/core/util/object.ts
mattpap: @clairetang6, I was wrong with the original signature. We define `merge()` for specific class of objects, so the definition here is correct. Though the return type could be more specific (the same as `obj1` and `obj2`).

bokehjs/src/coffee/core/util/object.ts
```diff
@@ -27,6 +29,21 @@ export function clone<T>(obj: T): T {
...
+
+  let keys = concat([Object.keys(obj1), Object.keys(obj2)])
+
+  for (let i = 0; i < keys.length; i++) {
+    let key = keys[i]
+    let arr1: Array<T> = obj1.hasOwnProperty(key) ? obj1[key] : []
+    let arr2: Array<T> = obj2.hasOwnProperty(key) ? obj2[key] : []
```
mattpap: Those type annotations shouldn't be necessary.

bryevdv:
LGTM

bokehjs/src/coffee/models/tools/gestures/select_tool.coffee
```diff
@@ -80,3 +80,12 @@ export class SelectTool extends GestureTool
...
     computed_renderers: () -> @_get_computed('computed_renderers')
   }
+
+  _computed_renderers_by_data_source: () ->
+    renderers_by_source = {}
+    for r in @computed_renderers
+      if !(r.data_source of renderers_by_source)
+        renderers_by_source[r.data_source] = [r]
+      else
+        renderers_by_source[r.data_source] = renderers_by_source[r.data_source].concat([r])
```
mattpap: Plain JS object can have only string keys. This should use `r.data_source.id`.

mattpap: Later on we should switch from using plain objects as dictionaries to `Map` or `WeakMap`.

clairetang6:
@mattpap, yeah I had noticed that I couldn't get the selection manager straight from the data source key since it had been converted to a string. I'm assuming with the Map that then the key would be the actual object? Then, you could get the selection manager from the ds without having to go through a renderer (just ds.sm instead of r's[0].ds.sm).

bryevdv:
Failing the file size limit test 
```
  1) js/bokeh.min.js file size should remain below limit:
     AssertionError: expected 666329 to be below 665600
      at Context.<anonymous> (test/size.coffee:22:37)
``` 

Will have to increase limit or find some savings somewhere.

canavandl:
For some reason, the build size on my machine is (fractionally) lower on my machine than on Travis. So the tests will pass locally and not on CI.

bryevdv:
@canavandl @clairetang6 I will fix the merge conflict and update the size limits this afternoon, then merge for a dev build.

mattpap:
> the size limits this afternoon

The size limit was bumped to 700 kB in PR #6233, so it should be sufficient.

mattpap:
```
<<<<<<< clairetang6/2136_selection_multiple_renderers
      source.trigger('select')

      for renderer_view in renderer_views
        source.trigger('select-' + renderer_view.model.id)
=======
      source.select.emit(renderer_view.model.id)
>>>>>>> master
```
I'm not sure of resolution of this. If we need those ids, then this would work:
```
source.select.emit([ view.model.id for view in renderer_views ])
```

bryevdv:
@mattpap why would it not be
```
for renderer_view in renderer_views
 source.select.emit(renderer_view.model.id)
```
?

mattpap:
> why would it not be

Because you are emitting the same signal multiple times, just with different argument. Then each slot connected to this signal will trigger multiple times. Currently there is only one slot connected in glyph_renderer (`request_render`, so you wouldn't even notice that something is wrong due to throttling). Given that no one currently uses those ids, I would just not emit them.

clairetang6:
In the selection manager before this PR and before the change to signaling, there were two calls to source.trigger:

```python
source.trigger('select')
source.trigger('select-' + renderer_view.model.id
```

I don't think the second call with the id was listened to by anything in bokehjs, but I thought maybe it was there in case a user wanted to listen to a select event that was triggered by a specific renderer. Were users able to access the backbone events before? If they weren't then, could we just change that line to `source.select.emit()`?

bryevdv:
@clairetang6 not easily and it was never demonstrated or advertised as a supported thing to do. I don't think it needs to be replaced.

bryevdv:
@clairetang6 I only just saw that you removed `ready` do you want this reverted? What was left to do?

clairetang6:
Hi @bryevdv, no I only removed ready to fix the merge conflict. I think any other work related to this could be in a separate PR.

Issue #6184, opened 2017-04-22
bokeh-api documentation improvement suggestions


aerusso:
The [Developing with JavaScript](http://bokeh.pydata.org/en/latest/docs/user_guide/bokehjs.html) page is very helpful, but a few (important) things were not mentioned on this page:

1. All *three* of bokeh-x.y.z.min.css bokeh-x.y.z.min.js and bokeh-api-x.y.z.min.js must be used and **in that order**.
2. Rendering into an html element is supported by  `Bokeh.Plotting.show(object,html_object)`
3. Changes to data sources can be propagated to graphs by calling `data_source.trigger("update")`

Is there a way for me to submit a full minimal html example to be included on that page?

bryevdv:
> Is there a way for me to submit a full minimal html example to be included on that page?

Not trivially. I've been meaning to work on a new `bokehjs-plot` Sphinx directive that would make it easy to embed BokehJS examples in our docs, but I have not had the opportunity to do so yet. It's possible to include raw HTML in Sphinx docs though, so that's probably the best approach to get "real" plots on that page for now. If you are interested I can provide more information.

bryevdv:
For reference, the ReST source for that page is here:

https://raw.githubusercontent.com/bokeh/bokeh/master/sphinx/source/docs/user_guide/bokehjs.rst

And some docs about raw HTML pass through are here:

http://docutils.sourceforge.net/docs/ref/rst/directives.html#raw-data-pass-through

aerusso:
Maybe in the short term, just the HTML can be included, and later on we can figure out the best way to properly include an illustration of the example? My pull request #6203 does exactly this, and also identifies some other outdated comments about BokehJS API.

PR #6203, opened 2017-04-27
Update documentation regarding BokehJS API split

aerusso:
Addresses issues in #6184.

### DIFF
--- a/sphinx/source/docs/user_guide/bokehjs.rst
+++ b/sphinx/source/docs/user_guide/bokehjs.rst
@@ -130,9 +130,9 @@ These higher level interfaces currently comprise  ``Bokeh.Plotting`` and
 ``Bokeh.Charts``.
 
 .. note::
-    The APIs described below are currently part of the ``bokeh.js`` file,
-    however it is likely that in the future these interfaces will be split
-    into separate files for distribution.
+    As of ``0.12.2`` the APIs described below have been split into BokehJS
+    API, in the ``bokeh-api.js`` file, which must be imported in addition
+    to ``bokeh.js``.
 
 .. _userguide_bokehjs_interfaces_plotting:
 
--- a/sphinx/source/docs/user_guide/bokehjs.rst
+++ b/sphinx/source/docs/user_guide/bokehjs.rst
@@ -337,3 +337,63 @@ function, with the plot it generates shown below:
 
 .. image:: /_images/bokehjs_bar_charts.png
     :width: 100%
+
+
+``Minimal Complete Example``
+''''''''''''''''''''''''''''
+
+A minimal example follows, demonstrating a proper import of the libraries,
+and dynamic creation and modification of plots.
+
+.. code-block:: html
+
+    <!doctype html>
+    <html lang="en">
+    <head>
+    <meta charset="utf-8">
+    <title>Complete Example</title>
+    <link rel="stylesheet" href="https://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.css" type="text/css" />
+    <script type="text/javascript" src="https://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.js"></script>
+    <script type="text/javascript" src="https://cdn.pydata.org/bokeh/release/bokeh-api-0.12.5.min.js"></script>
+    <!-- The order of CSS and JS imports above is important. -->
+    </head>
+
+    <body>
+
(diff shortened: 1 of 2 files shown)


sphinx/source/docs/user_guide/bokehjs.rst
```diff
@@ -337,3 +337,63 @@ function, with the plot it generates shown below:
...
+
+    <body>
+
+    <button onclick="addPoint()">Add some data!</button><br/>
+
+    <div>
+    <script type="text/javascript">
+
+    // arrays to hold data
+    source = new Bokeh.ColumnDataSource({
```
mattpap: `var source`. The same for `plot` and `scatterData`.

aerusso: Whoops! I tested with `use strict` now.

sphinx/source/docs/user_guide/bokehjs.rst
```diff
@@ -337,3 +337,63 @@ function, with the plot it generates shown below:
...
+.. code-block:: html
+
+    <!doctype html>
+    <html lang="en">
+    <head>
+    <meta charset="utf-8">
+    <title>Complete Example</title>
+    <link rel="stylesheet" href="http://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.css" type="text/css" />
+    <script type="text/javascript" src="http://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.js"></script>
+    <script type="text/javascript" src="http://cdn.pydata.org/bokeh/release/bokeh-api-0.12.5.min.js"></script>
```
bryevdv: CDN links should be `https` (this will work but it's always better to demonstrate with `https`)

bryevdv:
@aerusso This looks great, maybe for the short term a JSFiddle link or embed would be helpful to go along? I got your code running in this fiddle:

https://jsfiddle.net/bokeh/ck5udmxb/

I did have to change slightly, to split out the JS and HTML as jsfiddle prefers. Let me know what you think.

bryevdv:
also I should add, for embedding (I think either the script or iframe would work), there is a raw HTML directive for sphinx:
```
.. raw:: html 

    [ html/css/js here ]

```

sphinx/source/docs/user_guide/bokehjs.rst
aerusso: Thanks! I changed this.

aerusso:
@bryevdv I actually ran into that "covering up other text" issue, but wasn't sure if I was just doing something wrong.

I don't know the proper way to add an html file into the Sphinx build process, but I implemented your suggestion with iframes in aerusso/bokeh@5eb9c8086842ed3d56292b842eb822e7633b2d29 . It builds with rst2html and does what you'd expect.

bryevdv:
@aerusso Sorry I was pretty vague in my suggestion. I was not proposing to add an HTML file like that. If you click through the jsfiddle link above, there is a button to "embed" the fiddle. It gives you either s JS snippet, or an iframe, that you can drop into a page. I mean that you could use `raw:: html` with one of those things. Something like:

```
.. raw:: html 

    <script async src="//jsfiddle.net/bokeh/ck5udmxb/embed/"></script>
```

or 

```
.. raw:: html
 
    <iframe width="100%" height="300" src="//jsfiddle.net/bokeh/ck5udmxb/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
```

maybe with a note like "you can see this code live n the embedded JSFiddle below" or similar.

bryevdv:
@mattpap can you comment on the "div covering up" issue? Does the user supplied div need to have `bokeh-root` class or something?

aerusso:
I've added the JSFiddle, with a slightly larger height to make scrolling simpler.

bryevdv:
FYI tests failing spuriously due to a downstream dependency changing

bryevdv:
@aerusso It think this looks good to go, for my benefit can you attach a screenshot of the rendered docs page here?

aerusso:
Attaching as requested. I built this by simply calling `rst2html` on `bokehjs.rst`, running a Python SimpleHTTPServer, and viewing the page through Google Chrome 58.

bokehjsapi-example-rendered

bryevdv:
Thanks @aernlund ! Merging since the unrelated test issue is fixed in master.

PR #6204, opened 2017-04-27
Bryanv/mappers

### LINKED ISSUE #5268: Improve mappers' inheritance structure
Currently we have:

```
```

bryevdv:
- [x] issues: fixes #5268
- [x] tests added / passed

@mattpap @canavandl @bokeh/dev This PR cleans up the transform/mapper/scale structure. It is the basis of some other work around fixing categoricals,  adding custom mappers, and getting rid of the gorpy `mapper_type` junk. But I'd like to get the big disruptive re-org part merged first, and then follow on with other work in other PRs. 

Question: do we update `@xmapper`, `@x_mappers[]` etc to use the "scale" terminology everwhere? It's definitely the "right" thing to do, but is also somewhat disruptive. I think we'd have to leave back-compat aliases in for awhile at least. We should decide to do it (or not) and if yes, then I will do it as part of this PR. 

I also plan to add some docs (including some migration notes) and tests to this PR before merging but I wanted to get these questions out of the way. 

@mattpap I also need your help specifically to fix up the TS typings appropriately. 

---

Description:

* `Transform` is represents a computation in JS, and always has `compute` and `v_compute`
* `ColorMapper` is now subclasses of `Transform` which allows `DataSpec` to be cleaner. Continue to call color "mapper" because that is common terminology
* `Scale` is a subclass of `Transform` and represents and *invertible* computation. It adds `invert` and `v_invert`. All the old non-color mappers are now scales, e.g. `LinearScale`

Future work this supports:

* get rid of `mapper_type` gorp, in favor of sending scale models explicity 
* add a `CustomJSScale`
* clean up categorical handling


bokeh/core/properties.py
```diff
@@ -1440,7 +1440,6 @@ def __init__(self, key_type, value_type, default, help=None):
                 Either(
                     String,
                     Instance('bokeh.models.transforms.Transform'),
-                    Instance('bokeh.models.mappers.ColorMapper'),
```
bryevdv: This has bothered me for ages.

bokehjs/src/coffee/api/typings/models/scales.d.ts
```diff
@@ -0,0 +1,3 @@
+declare namespace Bokeh {
+
+}
```
bryevdv: @mattpap need your help here. I think the `mappers.d.ts` file also needs updating not that color mapper inherits from `Transform`

mattpap: There are no typings for transforms yet, so this has to be added first. Otherwise changes should be straightforward (move `mappers.d.ts -> scales.d.ts`, change the base and do renaming).

bryevdv:
@mattpap all tests (including master) are failing with this
```
_______________ test_functickformatter_from_coffeescript_no_arg ________________
    def test_functickformatter_from_coffeescript_no_arg():
        coffee_code = dedent("""
            square = (x) -> x * x
            return square(tick)
            """)
    
        formatter = FuncTickFormatter.from_coffeescript(code=coffee_code)
    
>       assert formatter.code == dedent("""\
            "use strict";
            Object.defineProperty(exports, "__esModule", { value: true });
            var square;
            square = function (x) {
                return x * x;
            };
            return square(tick);
            """)
E       assert 'var square;\n...quare(tick);\n' == '"use strict";...uare(tick);\n'
E         + "use strict";
E         + Object.defineProperty(exports, "__esModule", { value: true });
E           var square;
E           square = function (x) {
E               return x * x;
E           };
E           return square(tick);
bokeh/models/tests/test_formatters.py:60: AssertionError
```
etc

bokehjs/src/coffee/api/typings/models/scales.d.ts
bryevdv: @mattpap  does this look right? 
```
diff --git a/bokehjs/src/coffee/api/typings.d.ts b/bokehjs/src/coffee/api/typings.d.ts
index 6c67b29af..16e7c7a5d 100644
--- a/bokehjs/src/coffee/api/typings.d.ts
+++ b/bokehjs/src/coffee/api/typings.d.ts
@@ -42,3 +42,4 @@
 /// <reference path="./typings/models/tiles.d.ts" />
 /// <reference path="./typings/models/toolbars.d.ts" />
 /// <reference path="./typings/models/tools.d.ts" />
+/// <reference path="./typings/models/transforms.d.ts" />
diff --git a/bokehjs/src/coffee/api/typings/models/mappers.d.ts b/bokehjs/src/coffee/api/typings/models/mappers.d.ts
index fb0dbfbb6..78eab8d5b 100644
--- a/bokehjs/src/coffee/api/typings/models/mappers.d.ts
+++ b/bokehjs/src/coffee/api/typings/models/mappers.d.ts
@@ -1,9 +1,6 @@
 declare namespace Bokeh {
-  export interface Mapper extends Model, IMapper {}
-  export interface IMapper extends IModel {}
-
-  export interface ColorMapper extends Mapper, IColorMapper {}
-  export interface IColorMapper extends IMapper {
+  export interface ColorMapper extends Transform, IColorMapper {}
+  export interface IColorMapper extends ITransform {
     palette?: Array<Color>;
     nan_color?: Color;
   }
diff --git a/bokehjs/src/coffee/api/typings/models/scales.d.ts b/bokehjs/src/coffee/api/typings/models/scales.d.ts
index 70de6b53f..dac98eed6 100644
--- a/bokehjs/src/coffee/api/typings/models/scales.d.ts
+++ b/bokehjs/src/coffee/api/typings/models/scales.d.ts
@@ -1,3 +1,16 @@
 declare namespace Bokeh {
+  export interface Scale extends Transform, IScale {}
+  export interface IScale extends ITransorm {}

+  export var LinearScale: { new(attributes?: ILinearScale, options?: ModelOpts): LinearScale };
+  export interface LinearScale extends Scale, ILinearScale {}
+  export interface ILinearScale extends IScale {}
+
+  export var LogScale: { new(attributes?: ILogScale, options?: ModelOpts): LogScale };
+  export interface LogScale extends Scale, ILogScale {}
+  export interface ILogScale extends IScale {}
+
+  export var CategoricalScale: { new(attributes?: ICategoricalScale, options?: ModelOpts): CategoricalScale };
+  export interface CategoricalScale extends Scale, ICategoricalScale {}
+  export interface ICategoricalScale extends IScale {}
 }
```
with a new file for transforms:
```
declare namespace Bokeh {
  export type Int = number;
  export type Percent = number;
  export type Color = NamedColor       |
            string           |
            [Int, Int, Int]         |  // RGB (r, g, b)
            [Int, Int, Int, Percent];  // RGBA(r, g, b, a)
  export type FontSize = string;
  export type Map<T> = {[key: string]: T};
  export type JsObj = Map<any>;
}
```

mattpap:
@bryevdv, typescript 2.3.0 was officially released. What's seen here is what I've been investigating in PR #6139 without success. I think we will just have to remove those lines, as I have no idea why this doesn't work anymore. This shouldn't affect anything. I will submit a PR shortly.

mattpap:
> This shouldn't affect anything

1. `"use strict"` is added by browserify anyway.
2. `__esModule` doesn't really have any function right now.

bokehjs/src/coffee/api/typings/models/transforms.d.ts
```diff
@@ -0,0 +1,4 @@
+declare namespace Bokeh {
+  export interface Transform extends Model, ITransform {}
+  export interface ITransform extends IModel {}
+}
```
bryevdv: @mattpap these typings changes compile without error, but please check them. I only added the transform base, I don't have enough TS experience to add the rest.

mattpap: Typings look good. Also, you should run `gulp examples` just to make sure everything compiles. Though examples don't use mappers directly, so there should be no failures.

bryevdv:
@mattpap @canavandl @bokeh/dev OK I went ahead and pushed a commit to implement the var renaming. We can revert it if that's decided.

bokehjs/src/coffee/models/scales/linear_scale.coffee
```diff
@@ -0,0 +1,60 @@
...
+
+export class LinearScale extends Scale
+
+  initialize: (attrs, options) ->
+    super(attrs, options)
+
+    @define_computed_property('state', @_state, true)
+    @add_dependencies('state', this, ['source_range', 'target_range'])
+    @add_dependencies('state', @source_range, ['start', 'end'])
+    @add_dependencies('state', @target_range, ['start', 'end'])
```
mattpap: @bryevdv, as long as you're cleaning things up, you could replace all those computed properties with straight implementation.

bryevdv: I'm happy to do that, but not in this PR which is more than large enough already. There will be more mapper PRs shortly that are smaller and would be better for stashing this kind of incidental cleanup.

bryevdv:
If there's no other comments, I will add:

* migration notes
* some docs
* "compat" properties for `@xmapper`, etc.

then merge this later today.

bokehjs/src/coffee/models/ranges/data_range1d.coffee
```diff
@@ -22,7 +22,7 @@ export class DataRange1d extends DataRange
   }
 
   @internal {
-      mapper_hint:     [ p.String, 'auto' ]
+      scale_hint:     [ p.String, 'auto' ]
```
canavandl: Is the intention to remove the scale hinting (and pass first class models) in a follow-up PR?

bokehjs/src/coffee/models/scales/scale.coffee
```diff
@@ -0,0 +1,20 @@
+import {Transform} from "../transforms"
+import {isNumber} from "core/util/types"
+
+export class Scale extends Transform
+
+  # map to/from methods provided for back compat, should be
+  # considered deprecated.
+
+  # TODO (bev) "rest" args can be removed when categoricals fixed up
+  map_to_target: (x, rest...) ->
```
canavandl: is this just for backwards compat?

bryevdv: Yup

bokehjs/src/coffee/models/ranges/data_range1d.coffee
bryevdv: yes this PR is already big and disruptive, and that would be an isolated chunk of work once this PR is in

bryevdv: (I only renamed it at all because it was easier to do and verify during the mass rename)

canavandl:
LGTM. I both like the direction this PR takes the project (having first class Scale models) and that the PR is pretty tightly scoped. I assume you've grep-ed around the find all usages of x_mappers in the examples?

bryevdv:
>  I assume you've grep-ed around the find all usages of x_mappers in the examples?

I've searched in the sphinx docs an the examples

canavandl:
> I've searched in the sphinx docs an the examples

You found the instances that I was able to think of. So I think it's fine.

bokehjs/src/coffee/models/annotations/arrow.coffee
```diff
@@ -10,8 +10,6 @@ export class ArrowView extends AnnotationView
     if not @model.source?
       this.model.source = new ColumnDataSource()
     @canvas = @plot_model.canvas
-    @xscale = @plot_view.frame.xscales[@model.x_range_name]
-    @yscale = @plot_view.frame.yscales[@model.y_range_name]
```
bryevdv: seems unused anywhere

bokehjs/src/coffee/models/annotations/box_annotation.coffee
```diff
@@ -32,15 +32,15 @@ export class BoxAnnotationView extends AnnotationView
...
-    stop    = @canvas.vy_to_sy(@_calc_dim(@model.top,    @model.top_units,    @yscale, @frame.v_range.end))
+    frame = @plot_model.frame
+    canvas = @plot_model.canvas
+    xscale = @plot_view.frame.xscales[@model.x_range_name]
+    yscale = @plot_view.frame.yscales[@model.y_range_name]
+
+    sleft   = canvas.vx_to_sx(@_calc_dim(@model.left,   @model.left_units,   xscale, frame.h_range.start))
+    sright  = canvas.vx_to_sx(@_calc_dim(@model.right,  @model.right_units,  xscale, frame.h_range.end))
+    sbottom = canvas.vy_to_sy(@_calc_dim(@model.bottom, @model.bottom_units, yscale, frame.v_range.start))
+    stop    = canvas.vy_to_sy(@_calc_dim(@model.top,    @model.top_units,    yscale, frame.v_range.end))
```
bryevdv: no reason to store these on the object that I can see

bokehjs/src/coffee/models/annotations/label.coffee
```diff
@@ -37,13 +36,13 @@ export class LabelView extends TextAnnotationView
...
     if @model.x_units == "data"
-      vx = @xscale.compute(@model.x)
+      vx = xscale.compute(@model.x)
     else
       vx = @model.x
     sx = @canvas.vx_to_sx(vx)
 
     if @model.y_units == "data"
-      vy = @yscale.compute(@model.y)
+      vy = yscale.compute(@model.y)
```
bryevdv: same, no reason to store on the object

bokehjs/src/coffee/models/annotations/label_set.coffee
```diff
@@ -40,14 +37,17 @@ export class LabelSetView extends TextAnnotationView
...
     if @model.x_units == "data"
-      vx = @xscale.v_compute(@_x)
+      vx = xscale.v_compute(@_x)
     else
       vx = @_x.slice(0) # make deep copy to not mutate
     sx = @canvas.v_vx_to_sx(vx)
 
     if @model.y_units == "data"
-      vy = @yscale.v_compute(@_y)
+      vy = yscale.v_compute(@_y)
```
bryevdv: ditto

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -91,6 +91,10 @@ export class CartesianFrame extends LayoutCanvas
     y_ranges: () -> @_y_ranges
     xscales:  () -> @_xscales
     yscales:  () -> @_yscales
+
+    # These are deprecated and should not be used in new code
+    x_mappers: () -> @_xscales
+    y_mappers: () -> @_yscales
   }
```
bryevdv: @mattpap please confirm this is the correct way to do things. The passing test seems to indicate that it is

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -87,10 +87,6 @@ export class PlotCanvasView extends DOMView
 
     # compat, to be removed
     @frame = @model.frame
-    @x_range = @frame.x_ranges['default']
-    @y_range = @frame.y_ranges['default']
-    @xscale = @frame.xscales['default']
-    @yscale = @frame.yscales['default']
```
bryevdv: I couldn't find these in use anywhere, in BokehJS or in any examples. They were marked "to be removed" years ago so now seems like a good time

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -52,6 +52,10 @@ export class GlyphRendererView extends RendererView
     @xscale = @plot_view.frame.xscales[@model.x_range_name]
     @yscale = @plot_view.frame.yscales[@model.y_range_name]
 
+    # compat, deprecated, to be removed
+    @xmapper = @xscale
+    @ymapper = @yscale
+
```
bryevdv: This should allow existing JS code to continue to function. If theres an easy way to yell loudly in the console once about a deprecation (i.e not on every single access) I guess that would be good to do

bryevdv:
@mattpap @canavandl can you both review just the latest commit

bokehjs/src/coffee/models/plots/plot_canvas.coffee
bryevdv: A few things do still use `@x_range`, etc after lunch I will try to fix them up to not do that, or put this back otherwise.

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
mattpap: @bryevdv, you mean creating getters? If so, yes.

bryevdv: just so that `frame.x_mappers['default']` etc still works

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
mattpap: You could use a getter logging a deprecation warning and returning a value.

canavandl: you could have a getter and an underscore property "_has_warned" that gets updated?

mattpap: My assumption is that the library (bokehjs in this case) uses the new naming convention across the board. Then a chatty approach makes sense to force the user to deal with issues sooner than later. After all, this layer is not provided there to last, but to help user run his code in the mean time of fixing it.

bryevdv:
OK I think I have addressed all comments, will merge when green

mattpap:
Heads up for anyone using pre 0.12.5 coffeescript. Due to a bug compilation fails with:
```
bokeh/bokehjs/src/coffee/models/glyphs/patches.coffee:67:45: error: unexpected default
    xr = @renderer.plot_view.frame.x_ranges.default
                                            ^^^^^^^
```

PR #6211, opened 2017-04-28
Initialize properly plot canvas in plot's subclasses

### LINKED ISSUE #6210: bokehjs doesn't work in notebook
A side effect of bad programming practice in gmap's code and PR #5976.

mattpap:
fixes #6210

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot.coffee
+++ b/bokehjs/src/coffee/models/plots/plot.coffee
@@ -103,7 +103,7 @@ export class Plot extends LayoutDOM
       title = if isString(@title) then new Title({text: @title}) else @title
       @add_layout(title, @title_location)
 
-    @_plot_canvas = new PlotCanvas({plot: @})
+    @_plot_canvas = @_init_plot_canvas()
 
     @toolbar.toolbar_location = @toolbar_location
     @toolbar.toolbar_sticky = @toolbar_sticky
@@ -128,6 +128,9 @@ export class Plot extends LayoutDOM
 
     _set_sizeable(@)
     _set_sizeable(@plot_canvas)
+
+  _init_plot_canvas: () ->
+    return new PlotCanvas({plot: @})
 
   @getters {
     plot_canvas: () -> @_plot_canvas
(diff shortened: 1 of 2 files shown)


bryevdv:
What do we need to get some form of minimal notebook testing back in our Travis builds?

bokehjs/src/coffee/models/plots/plot.coffee
```diff
@@ -103,7 +103,7 @@ export class Plot extends LayoutDOM
       title = if isString(@title) then new Title({text: @title}) else @title
       @add_layout(title, @title_location)
 
-    @_plot_canvas = new PlotCanvas({plot: @})
+    @_init_plot_canvas()
```
bryevdv: I think `@_plot_canvas = @_init_plot_canvas()` (i.e. having the method return the canvas)  might be nicer than relying on the side effect of setting the property.

mattpap:
> What do we need to get some form of minimal notebook testing back in our Travis builds?

I don't know. notebooks didn't work for a long time, so there may be several issues to fix.

bryevdv:
> I don't know. notebooks didn't work for a long time, so there may be several issues to fix.

I'd settle for a couple of canary image diffs that hard fail, at this point. 

Do you have any thoughts about the other comment? I'm happy to make the change, if you are busy (but agree)

bokehjs/src/coffee/models/plots/plot.coffee
bryevdv: For that matter, I'd prefer not to have a methods at all, can't we add a class variable for the type? Then
```
@_plot_canvas = new @canvas_type({plot: @})
```
ping @mattpap

mattpap:
> I'd settle for a couple of canary image diffs that hard fail, at this point.

This is not a question of scope, but just making our testing infrastructure work with notebook again. I don't remember what was the problem, I need to refresh my memory on this. I just suspect that there was an issue with detecting when notebook evaluated cells and bokehjs finished rendering, so possibly existing heuristic or an improved one will help with this.

bryevdv:
> This is not a question of scope, but just making our testing infrastructure work with notebook again. I don't remember what was the problem, I need to refresh my memory on this. I just suspect that there was an issue with detecting when notebook evaluated cells and bokehjs finished rendering, so possibly existing heuristic or an improved one will help with this.

right getting notebooks tested regularly again is definitely for a separate issue. I'd like to get this merged an in a dev build. Can you address the comment about avoiding side effects?

mattpap:
> Can you address the comment about avoiding side effects?

@bryevdv, I already did.

bryevdv:
@bryevdv, I already did.

I missed it, sorry! merging now

PR #6229, opened 2017-05-02
Proof of concept: client-side SVG generation

### LINKED ISSUE #538: Headless static (svg, png) image generation

canavandl:
Experimental PR to investigate generating SVGs on the client side instead of 

borrows from: #3867

- [x] issues: fixes #538 
- [ ] tests added / passed

What doesn't currently work:
* ~~SaveTool (should be easy to fix, just needs to use `getSerializedSvg` method instead of `getDataURI` on mock canvas element)~~ fixed
* Glyph clipping needs some tweaking, they are able to be dragged outside of the frame
* There's some artifact-ing from glyphs that are dragged off the canvas, leaving some lines on the canvas edge
* Performance (when panning, which causes lots of rerenders) isn't as fast as canvas. Unclear if is limitation of svg output.
* ~~Exporting layouts. This makes each plot a unique SVG, so it'd require some work to grab all of the SVGs then stitch them together.~~ out-of-scope for PR
* ~~It's not headless. I'd waiting until the ``export`` PR is merged then build on top of that to support headless generation using PhantomJS/headless Chrome. I also haven't experimented with node-canvas to see what the issues are with using that approach recently, though I suspect there are several.~~ out-of-scope for PR

Things that won't work with this approach:
* Exporting plot with a toolbar (the tool icons are overlaid HTML elements, so the exported SVG won't contain the toolbar)
* Exporting plot with webgl glyphs (the webgl glyphs are contained in a overlaid webgl layer, this also won't be exported along with the svg)

Usage/Test suggestions:
* set `Plot.background_fill_color = None` and `Plot.border_fill_color` so that the plot whitespace is transparent and not filled white.
* set `Plot.toolbar_location = None`, so that the toolbar (that isn't exported) doesn't figure into the layout solver (making the plot unnecessarily small to accomodate). But then you can't use the save tool... I need to think about this. Alternatively, you could right-click and save element (at least in FireFox)


canavandl:
screen shot 2017-05-02 at 12 41 02 pm

canavandl:
Output of SaveTool (note missing toolbar):

screen shot 2017-05-02 at 1 09 15 pm

mattpap:
> the webgl glyphs are contained in a overlaid webgl layer, this also won't be exported along with the svg)

Just to note, soon I will be splitting the rendering pipeline (levels) over multiple canvases, so lets make sure that this approach won't fall apart completely then.

bryevdv:
> Just to note, soon I will be splitting the rendering pipeline (levels) over multiple canvases

Where has there been discussion of this?

canavandl:
Question:
* Do we want to expose replacing the typical Canvas backend with SVG elements to users (i.e. create a SVG mode instead of only being a detail of an ``export_svg`` method)
* Is it acceptable to for an ``export_svg`` method to return several svg elements inside of a layout instead of a single svg of patches together plots?

canavandl:
In summary - this PR adds an ``Plot.output_backend`` enum property. When set to "svg" (the default is "canvas"), the canvas element is replaced with a svg element with a mocked ctx object that translates the canvas calls into svg output. There's some minor artifacting and mediocre performance on plot interactions, but it's unclear how resolvable those are.

I haven't added any tests. I'm unsure of the best way to do so, short of a simple integration test.

This could probably use a validation warning about using webgl=True and output_backend="svg" on the same plot, as the saved plot won't have the rendered webgl glyphs.

bryevdv:
Maybe webl and svg should both be different categories of output backends?

canavandl:
currently blocked until this PR is merged/release is made: https://github.com/gliffy/canvas2svg/pull/48

(The "Attempted to apply path command to node g" issue affects several of our glyph render implementations)

bryevdv:
ping @canavandl There was no response from the gliffy contact about the status of `canvas2svg` so I think we have to assume it is abandoned/unmaintained. I have forked the repo into the Bokeh org and also opened the PR that I believe you are waiting on on the fork. (Will leave it to you to confirm and merge)

canavandl:
Though there's several small improvements to be made re: testing, documentations, etc, i'm marking this as READY and will resolve in followup PRs.

PR #6233, opened 2017-05-03
Replace backbone events with signaling API

### LINKED ISSUE #4153: Strict trigger() and listenTo()

### LINKED ISSUE #6242: Task: Remove computed properties implementation
Remove computed properties

Current used by:
* LogScale
* LinearScale
* HoverTool
* SelectTool
* Axis

### LINKED ISSUE #6244: Fails to build in windows 7
Hi,

I have been trying to build Bokeh on Windows 7 using python 3.5.3. 

With a fix to _version.py, which is implemented later on, I can build 0.12.4. This displays the simple line example as expected.

When I try to build with 0.12.5rc1 the line example does not display as expected but raises an exception as outlined below. All the later builds I have tested do not resolve this.

Chrome

chrome
 
IE 11

ie11

### LINKED ISSUE #6247: Allow to write bokehjs unit tests in TypeScript

mattpap:
fixes #4153 
fixes #6244
fixes #6242
fixes #6247


bryevdv:
There's a lot to like here, just a few preliminary comments:

* `signaling.ts` needs complete unit test coverage
* we have to keep deprecated versions of `listenTo` and `trigger` for now, they have been publicly demonstrated too much to just drop without warning. I'd suggest a new function `@connect` for signals to replace `listenTo` rather than trying to support multiple call conventions in one function
* will ned migration notes

mattpap:
> we have to keep deprecated versions of listenTo and trigger for now

Yes, but per my policy this happens last in a PR, so that all code is forced to use new APIs.

> I'd suggest a new function `@connect` for signals

Actually I already did this locally, but used `@connectTo`, so that there's a clear distinction between `signal.connect(slot)` and `@connectTo(signal, slot)`.

bryevdv:
> Yes, but per my policy this happens last in a PR, so that all code is forced to use new APIs.

That's fine but was not stated, just making sure we are on the same page

> Actually I already did this locally, but used @connectTo, so that there's a clear distinction between signal.connect(slot) and @connectTo(signal, slot)

`connect_to` would be more in line with the rest of the project. I know camel case is more standard in the JS world but we are already well away from that, and I'd rather favor consistency at this point.

bryevdv:
@mattpap let me know if you have any questions about resolving the merge conflict (or want me to do it)

.travis.yml
```diff
@@ -64,7 +64,7 @@ script:
 
   # unit tests
   - if [[ -z "$TRAVIS_TAG" && "$GROUP" == unit ]]; then
-      py.test --rerun=3 -s -m js;
+      py.test -s -m js;
```
mattpap: I'm quite confident with PR #5976 and this we don't need to retry anymore.

bokehjs/gulp/tasks/test.coffee
```diff
@@ -12,71 +12,79 @@ mocha = (options={}) ->
...
 
-gulp.task "test:coverage", ["defaults:generate"], () ->
-  gulp.src(["./test/all.coffee"]).pipe(mocha({coverage: true}))
 
 gulp.task "test", ["defaults:generate"], () ->
-  gulp.src(["./test/all.coffee"]).pipe(mocha())
+  gulp.src(["./test/unit.coffee", "./test/defaults.coffee", "./test/size"]).pipe(mocha())
 
-gulp.task "test:client", ->
+gulp.task "test:unit", () ->
```
mattpap: `gulp test:unit` is now a way to run only unit tests, so nothing that depends on external factors (defaults, size).

bokehjs/src/coffee/core/properties.coffee
```diff
@@ -107,9 +110,6 @@ export class Property
 
     @init()
 
-    if trigger
-      @trigger("change")
```
mattpap: @bryevdv, I tried hard to figure out what's the purpose of this. Looks like a leftover, maybe from the time when `Property` was a backbone model. Anyway, no tests failed after removing this.

bokehjs/src/coffee/core/selection_manager.coffee
```diff
@@ -30,8 +30,7 @@ export class SelectionManager extends HasProps
 
       @source.selected = selector.indices
 
-      source.trigger('select')
-      source.trigger('select-' + renderer_view.model.id)
```
mattpap: It's not possible to trigger events like this and this this wasn't used anywhere so I removed this. It's always possible to filter `select` on its argument.

bokehjs/src/coffee/core/selection_manager.coffee
```diff
@@ -62,13 +61,7 @@ export class SelectionManager extends HasProps
 
       @source.setv({inspected: inspector.indices}, {"silent": true })
 
-      source.trigger(
-        'inspect', indices, tool, renderer_view, source, data
-      )
-      source.trigger(
-        "inspect#{renderer_view.model.id}", indices, tool, renderer_view,
-        source, data
-      )
```
mattpap: Same here.

bokehjs/src/coffee/core/ui_events.coffee
```diff
@@ -61,36 +79,42 @@ export class UIEvents
...
+        if v._move_enter?   then v.connectTo(@move_enter,   (x) -> if x.id == id then v._move_enter(x.e))
+        if v._move?         then v.connectTo(@move,         (x) -> if x.id == id then v._move(x.e))
+        if v._move_exit?    then v.connectTo(@move_exit,    (x) -> if x.id == id then v._move_exit(x.e))
+      when "tap"
+        if v._tap?          then v.connectTo(@tap,          (x) -> if x.id == id then v._tap(x.e))
+      when "press"
+        if v._press?        then v.connectTo(@press,        (x) -> if x.id == id then v._press(x.e))
+      when "scroll"
+        if v._scroll?       then v.connectTo(@scroll,       (x) -> if x.id == id then v._scroll(x.e))
+      else
```
mattpap: @canavandl, you may want to review changes to this file. What we have here isn't pretty, but events like `move:exit:#{id}` are a big no-go at this point.

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -37,6 +38,9 @@ export class ColorBarView extends AnnotationView
     return {x: x, y: -y}
 
   _get_size: () ->
+    if not @model.color_mapper?
```
mattpap: This is a recurring change, because `color_mapper` doesn't have a default value. This needs to be improved in a future PR. For now this is just to make tests run.

bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
```diff
@@ -317,44 +318,48 @@ export class HoverTool extends InspectTool
       callback:     [ p.Any                    ] # TODO: p.Either(p.Instance(Callback), p.Function) ]
     }
 
-  initialize: (attrs, options) ->
-    super(attrs, options)
+  connect_signals: () ->
+    super()
+    # TODO: @connectTo(@plot.properties.renderers.change, () -> @_computed_renderers = @_ttmodels = null)
```
mattpap: I will address this in another PR, because this kind of listeners need more attention across bokehjs.

bokehjs/src/coffee/polyfill.ts
```diff
@@ -0,0 +1,9 @@
+declare function require(moduleName: string): any
+
+if (typeof WeakMap !== "function") {
+  require("es6-weak-map/implement")
+}
+
+if (typeof Set !== "function") {
+  require("es6-set/implement")
+}
```
mattpap: This is annoying, but phantomjs and IE 10 don't support this. phantomjs isn't going to be maintained anymore, so we need to switch to chrome/chromium headless sooner than later. I'm not sure if we support IE 10, but at least as long as we use phantomjs, it's convenient though expensive to have this on-board. It's expensive, because it adds some 40~50 kB minified. When we drop phantomjs, then for IE we can provide a separate file with polyfills, so that users of the default, modern browsers don't have to suffer bigger downloads.

mattpap:
@bryevdv, @canavandl, as this is a large PR, I left comments in places where there is something interesting going on.

bokehjs/src/coffee/polyfill.ts
bryevdv: We are currently being funded to support PhantomJS as a target for static output generation by folks who can't really use Chrome. So either we find a way to include this only when generating output with PhantomJS, or we are stuck with it for a long time.

bryevdv: ping @canavandl

bryevdv:
I will check this out and test with it today. I'e asked @philippjfr and @jlstevens to do some light testing with DS/HV as well. I'd like to get this merged by tomorrow, given its size

philippjfr:
>I will check this out and test with it today. I'e asked @philippjfr and @jlstevens to do some light testing with DS/HV as well.

Just ran through all our HoloViews/Datashader example notebooks, particularly the ones that make use of events, without any issues.

bokehjs/src/coffee/polyfill.ts
mattpap: @bryevdv, phantomjs support is actually the easy case, because we control the environment, at least as far testing infrastructure and static image generation code (which is based on it) go, because we can always inject whatever code we need prior to running bokehjs and client code. IE is harder, because we don't know where bokehjs will actually run. However, I just recalled there is this MS' proprietary conditional syntax in HTML comments, that can be used for conditionally including polyfills. I will go ahead and implement this in another PR (when this is merged).

mattpap:
This PR adds support for tests written in TypeScript. The same can be done for our build, which I already did locally, so when this is merged, I will submit a PR.

bokehjs/src/coffee/polyfill.ts
bryevdv: OK, I expected that would be the case, and that's good to hear,  I just wanted to make sure the requirement was voiced explicitly.

bryevdv:
@mattpap the `pivot` example seems to have some issues, though I'm not sure if they are related to this PR yet or not.

sphinx/source/docs/releases/0.12.6.rst
```diff
@@ -58,3 +58,19 @@ Old                            New
...
+to bokehjs and writers of extensions.
+
+===================================== ==============================================
+Old                                   New
+===================================== ==============================================
+
+``@listenTo(obj, "change", fn)``      ``@connectTo(obj.change, fn)``
+``@listenTo(obj, "change:some", fn)`` ``@connectTo(obj.properties.some.change, fn)``
+``obj.trigger("change", arg)``        ``obj.change.emit(arg)``
+``obj.trigger("change:some", arg)``   ``obj.properties.some.change.emit(arg)``
```
bryevdv: @mattpap I would still prefer `connect_to` for stylistic consistency. I think `attr` would be a less confusingthign name than `some`. Also, is it impossible to make `obj.some.change.emit(arg)` work? Have versions of the old functions been added back with deprecation warnings?

bryevdv:
@mattpap I've tested  few dozen examples an things all seem to be going well. I have a few comments/questions to address, I left in a comment on the migration notes.

mattpap:
>  pivot example seems to have some issues, 

I tested on `master` and there are issues with this example.

sphinx/source/docs/releases/0.12.6.rst
mattpap: > I would still prefer connect_to for stylistic consistency

I forgot about this. I did rename to `connetTo` prior your earlier comment.

> I think attr would be a less confusingthign name than some.

Fine with me. I use `some` as a placeholder, like `foo`.

> Have versions of the old functions been added back with deprecation warnings?

Yes, you can still use `@trigger("change:some")`, though there will be deprecation warning.

bryevdv: `foo` would also be fine :) Do you mind making those two small changes? (attr name in notes and `connect_to`) ? Then we can merge and get a dev build out

bokehjs/src/coffee/core/ui_events.coffee
canavandl: I agree it isn't pretty, but it does the necessary work of isolating events between tools.

bokehjs/src/coffee/model.coffee
```diff
@@ -18,11 +18,15 @@ export class Model extends HasProps
...
       for cb in callbacks
-        @listenTo(@, evt, () -> cb.execute(@))
+        if attr != null
+          @connectTo(@properties[attr][evt], () -> cb.execute(@))
+        else
+          @connectTo(@[evt], () -> cb.execute(@))
 
-    @listenTo(@, 'change:js_event_callbacks', () -> @_update_event_callbacks)
-    @listenTo(@, 'change:subscribed_events', () -> @_update_event_callbacks)
+    @connectTo(@.properties.js_event_callbacks.change, () -> @_update_event_callbacks)
```
canavandl: remove the period from `@.properties`?

bokehjs/src/coffee/core/ui_events.coffee
bryevdv: Maybe a local `_filter_id` function could reduce the boilerplate? Not a huge priority for now tho.

sphinx/source/docs/releases/0.12.6.rst
mattpap: I renamed `some` to `attr`.

On a second thought I think that just `connect()` is sufficient. I went with, hopefully it's fine.

bokehjs/src/coffee/model.coffee
mattpap: Fixed.

bokehjs/src/coffee/core/selection_manager.coffee
canavandl: I agree, this is fine.

bokehjs/src/coffee/core/selection_manager.coffee
canavandl: I agree, this is fine.

bokehjs/src/coffee/polyfill.ts
canavandl: Great ^

canavandl:
This LGTM related to UIEvents and the annotations code.

PR #6235, opened 2017-05-03
Deprecate x/y_mapper_type Plot kwargs in place of first-class Scale models

### LINKED ISSUE #5268: Improve mappers' inheritance structure

canavandl:
Deprecate x/y_mapper_type Plot kwargs in place of first-class Scale models

- [ ] issues: fixes #5268
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -13,6 +13,7 @@ from ..core.validation import error, warning
 from ..core.validation.errors import REQUIRED_RANGE
 from ..core.validation.warnings import (MISSING_RENDERERS, NO_DATA_RENDERERS,
                                         MALFORMED_CATEGORY_LABEL, SNAPPED_TOOLBAR_ANNOTATIONS)
+from ..util.deprecation import deprecated
 from ..util.plot_utils import _list_attr_splat, _select_helper
 from ..util.string import nice_join
 
@@ -23,6 +24,7 @@ from .grids import Grid
 from .layouts import LayoutDOM
 from .ranges import Range, FactorRange
 from .renderers import DataRenderer, DynamicImageRenderer, GlyphRenderer, Renderer, TileRenderer
+from .scales import Scale, CategoricalScale, LinearScale, LogScale
 from .sources import DataSource, ColumnDataSource
 from .tools import Tool, Toolbar, ToolEvents
 
@@ -36,16 +38,28 @@ class Plot(LayoutDOM):
             kwargs["tool_events"] = ToolEvents()
 
         if "toolbar" in kwargs and "logo" in kwargs:
-            raise ValueError("Conflicing properties set on plot: toolbar, logo.")
+            raise ValueError("Conflicting properties set on plot: toolbar, logo.")
 
         if "toolbar" in kwargs and "tools" in kwargs:
-            raise ValueError("Conflicing properties set on plot: toolbar, tools.")
+            raise ValueError("Conflicting properties set on plot: toolbar, tools.")
 
         if "toolbar" not in kwargs:
             tools = kwargs.pop('tools', [])
             logo = kwargs.pop('logo', 'normal')
 
             kwargs["toolbar"] = Toolbar(tools=tools, logo=logo)
+
+        if "x_mapper_type" in kwargs and "x_scale" in kwargs:
+            raise ValueError("Conflicting properties set on plot: x_mapper_type, x_scale.")
+
+        elif "x_mapper_type" in kwargs:
+            kwargs["x_scale"] = self._scale(kwargs.pop("x_mapper_type"))
+
+        if "y_mapper_type" in kwargs and "y_scale" in kwargs:
+            raise ValueError("Conflicting properties set on plot: y_mapper_type, y_scale")
+
+        elif "y_mapper_type" in kwargs:
+            kwargs["y_scale"] = self._scale(kwargs.pop("y_mapper_type"))
 
         super(LayoutDOM, self).__init__(**kwargs)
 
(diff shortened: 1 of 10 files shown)


canavandl:
So I've run into an issue that we need to clarify. It's possible to specify a mapper_type that conflicts with the range type on a Plot (see example below).

```python

x_range = FactorRange(factors=["a", "b", "c"])
# here we're specifying a factor range and a log-type mapper
p = figure(x_range=x_range, x_mapper_type="log")
p.scatter(x=["a", "b", "c"], y=[1,2,3])
show(p)
```

The current behavior is to ignore the mapper_type kwarg and generate an appropriate mapper in this method: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/canvas/cartesian_frame.coffee#L45

Instead of creating a ranges lookup dict and a scales lookup dict, does it make more sense for ranges to have a ``scale`` property?

ping @bryevdv

bryevdv:
I don't think ranges should *have* scales. I would say we can specify and detect nonsensical combinations and report errors (on both python and JS sides). Basically factor range and categorical scale go with each other and nothing else.

canavandl:
We could certainly detect nonsensical scale/range combinations with setters.

Are there use cases where users are going want have ranges with different mappers on the same dimension (i.e. plotting a glyph in log space and another in linear space on the same x-dimension)?

bryevdv:
> Are there use cases where users are going want have ranges with different ~~~mappers~~~ scales on the same dimension (i.e. plotting a glyph in log space and another in linear space on the same x-dimension)?

I can't imagine one offhand, but you know that's been true about lots of things in the past. It might be worth trying to do a survey and see if other tools support this or if there are examples of this. If not or if it's extremely rare/contrived then supporting a single scale (but possibly multiple ranges) would be *much* simpler.

canavandl:
It appears that both matplotlib and ggplot2 only support a single scale on a dimension (along with multiple ranges). I'd propose continuing to match those libraries and revisit if/when there's a reasonable use case to expand to supporting multiple scales.

bokeh/models/plots.py
```diff
@@ -371,17 +385,50 @@ def _check_snapped_toolbar_and_axis(self):
...
+    def y_mapper_type(self):
+        deprecated((0, 12, 6), "y_mapper_type", "y_scale")
+        return self.y_scale
+
+    @y_mapper_type.setter
+    def y_mapper_type(self, mapper_type):
+        deprecated((0, 12, 6), "y_mapper_type", "y_scale")
+        self.y_scale = self._scale(mapper_type)
+
+    x_scale = Either(Auto, Instance(Scale), help="""
```
canavandl: My only other question was whether to have the default x/y_scale be ``Auto``, which would a LinearScale unless there's a FactorRange, then it'd be a Categorical Scale (I can extend the docstring to explain this).

bryevdv: Yah it kind of sucks but we definitely need a "non scale thing" to indicate "choose a scale for me" OTOH what about just `None` for that purpose? It's already accepted... I'm not opposed to `Auto`, either way the "auto" behavior should be be the default since it is the current default.

bryevdv:
> It appears that both matplotlib and ggplot2 only support a single scale on a dimension (along with multiple ranges). I'd propose continuing to match those libraries and revisit if/when there's a reasonable use case to expand to supporting multiple scales.

definitely in agreement, single scale per dimension

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -42,24 +42,21 @@ export class CartesianFrame extends LayoutCanvas
         ranges[name] = extra_range
     return ranges
 
-  _get_scales: (scale_type_name, ranges, frame_range) ->
+  _get_scales: (scale, ranges, frame_range) ->
```
canavandl: There's some weirdness here:

By building this 'scales' dict, we're implicitly supporting having a LinearScale and CategoricalScale (or LogScale and CategoricalScale), but not a LinearScale and LogScale. It's not clear (to me) why we'd make that distinction.

canavandl:
So I chatted with Caswell and said that matplotlib does support multiple scales (http://matplotlib.org/examples/api/two_scales.html) along the same dimension. Their implementation is that plots have axes (which are similar to Bokeh ranges + axes + tickers + formatters) that also have a `set_scale` method that allows users to define the scale for the axis.

I feel like my opinion is shifting back towards having ranges have a scale property (which would also handle the issue of validation and prevent getting into situations where a FactorRange is paired with a numeric scale). I might make an experimental PR to see how it looks/works

bryevdv:
What's a concrete use case?

canavandl:
You're right, it's not worth worrying about without any user interest. I'm going to mark this as ready then.

bryevdv:
@canavandl so the only other thing I think might be worth considering is possibly not ever sending, null, auto, etc to BokehJS. I.e., figure those out pyhton side, and always and only send a real `Scale` to BokehJS. I think to do this we'd have to have some pre-serialization hook to "fix up" auto, etc. It would have to happen right at the "last moment" because users could change ranges at any time before serialization.

canavandl:
>  I think to do this we'd have to have some pre-serialization hook to "fix up" auto, etc. It would have to happen right at the "last moment" because users could change ranges at any time before serialization.

Do we do something similar elsewhere? It seems a little gorpy instead of being able to rely on our property system.

bryevdv:
> Do we do something similar elsewhere? It seems a little gorpy instead of being able to rely on our property system.

It's vaguely similar to the validation pas, but otherwise no. I agree it's definitely a slight departure and would be a little work. OTOH there is some benefit to making BokehJS simpler, and being able to always assume that there is a real mapper model. So, wondering about the tradeoff.

PR #6237, opened 2017-05-04
WIP: Add NumberInput.

### LINKED ISSUE #6173: Add numeric input widget
Currently the only way I could find to input numbers directly is by using sliders. However if the range is quite large or you require high precision this is impractical. Currently I'm using a TextInput and casting, but a new widget which accepts numbers without casting would be useful. This could translate to `<input type="number">` in HTML.

herm:
This adds a new input for numbers with spin buttons. It works but the rendering is broken. In #6173 I was asked to open an pull request anyways. PLEASE DO NOT COMMIT THIS AT THIS TIME!

Current rendering
image

Correct rendering (from https://jqueryui.com/spinner/):
image

The code is quite similar to that of the slider implementation. I suspect it might have to do something with that I don't run `@_prefix_ui()` but if I include this line the control doesn't work at all. 

- [x] issues: fixes #6173
- [ ] tests added / passed


bryevdv:
@herm this PR is nearly three weeks old, do you intend to continue work on it?

herm:
I am not really sure how to fix the rendering problem. Could you please have a look at the code? You probably understand the interaction between the bokeh code and jqueryui better than I do.

mattpap:
To make things render properly `@_prefix_ui()` is required. Why it breaks spinner's functionality? This needs some debugging to know, but at this point I think that we need to end this prefixing business. It's a hack and it doesn't seem to work very well.

mattpap:
After PR #6335 is merged, there will be no more `ui-` prefixing. However, there will be also no jquery and jquery-ui, so this PR needs to be rewritten from scratch using DOM APIs. This would be actually useful, if we want to restore spinner cell editor, that will be broken by that PR.

bryevdv:
@herm I am going to go ahead and close this PR, since it is based on jq-ui, and recent work removed the jq-ui dependency from Bokeh. I think this is still a good feature and we'd be happy to help you work up a new PR that uses our `dom` module if you have the time and inclination

PR #6239, opened 2017-05-05
fixed reference to delta in scroll event - fixes #6238

### LINKED ISSUE #6238: MouseWheel event: delta=None at all times
Bug can be demonstrated with this app, modified from [here](https://github.com/bokeh/bokeh/blob/master/examples/howto/js_events.py).
```python
""" Demonstration Bokeh app of how to register event callbacks in both
Javascript and Python using an adaptation of the color_scatter example
from the bokeh gallery. This example extends the js_events.py example
with corresponding Python event callbacks.

Removed all events except MouseWheel to illustrate bug in event data
"""

import numpy as np

from bokeh.io import curdoc
from bokeh import events
from bokeh.models import CustomJS, Div, Button
from bokeh.layouts import column, row


def display_event(div, attributes=[]):
    """
    Function to build a suitable CustomJS to display the current event
    in the div model.
    """
    style = 'float:left;clear:left;font_size=0.5pt'
    return CustomJS(args=dict(div=div), code="""
        var attrs = %s;
        var args = [];
        for (var i=0; i<attrs.length; i++ ) {
            args.push(attrs[i] + '=' + Number(cb_obj[attrs[i]]).toFixed(2));
        }
        var line = "<span style=%r><b>" + cb_obj.event_name + "</b>(" + args.join(", ") + ")</span>\\n";
        var text = div.text.concat(line);
        var lines = text.split("\\n")
        if ( lines.length > 35 ) { lines.shift(); }
        div.text = lines.join("\\n");
    """ % (attributes, style))

def print_event(attributes=[]):
    """
    Function that returns a Python callback to pretty print the events.
    """
    def python_callback(event):
        cls_name = event.__class__.__name__
        attrs = ', '.join(['{attr}={val}'.format(attr=attr, val=event.__dict__[attr])
                       for attr in attributes])
        print('{cls_name}({attrs})'.format(cls_name=cls_name, attrs=attrs))
    return python_callback

# Follows the color_scatter gallery example

N = 4000
colors = [
]

p = figure(tools="pan,wheel_zoom,zoom_in,zoom_out,reset")

p.scatter(x, y, radius=radii,
          line_color=None)

# Add a div to display events and a button to trigger button click events

div = Div(width=1000)
button = Button(label="Button", button_type="success")
layout = column(button, row(p, div))


point_attributes = ['x','y','sx','sy']
wheel_attributes = point_attributes+['delta']

## Register Javascript event callbacks
# Mouse wheel event
p.js_on_event(events.MouseWheel, display_event(div,attributes=wheel_attributes))

## Register Python event callbacks
# Mouse wheel event
p.on_event(events.MouseWheel, print_event(attributes=wheel_attributes))

curdoc().add_root(layout)
```

Run the app using `bokeh serve --show events_app.py`

Move the mouse to the plot window, then scroll. 

Example browser output (js_on_event callback):
> wheel(x=61.83, y=65.67, sx=358.40, sy=208.07, delta=NaN)
> wheel(x=26.34, y=61.05, sx=184.40, sy=232.07, delta=NaN)

Example python console output (on_event callback):
> MouseWheel(x=61.827411078543115, y=65.6738621521518, sx=358.3999938964844, sy=208.06666564941406, delta=None)
> MouseWheel(x=26.336113449936775, y=61.05292612403528, sx=184.39999389648438, sy=232.06666564941406, delta=None)

The [documentation ](http://bokeh.pydata.org/en/latest/docs/reference/events.html) states
>delta (float) : the (signed) scroll speed

Which implies that delta should represent scroll speed and direction, but it appears to be getting lost somewhere.

schoolie:
I came across a bug in the MouseWheel event. The bug can be shown by running this app from the howto section:
bokeh/examples/howto/events_app.py

The value for "delta" of the scroll event was showing up as NaN / None in both js_on_event and on_event callbacks.

Without fully understanding how everything's connected, I was able to trace the problem to line 162 of bokeh/bokehjs/src/coffee/core/bokeh_events.ts

Changing 
~~~
  static from_event(e: any, model_id: string = null) {
    return new this({
      sx: e.bokeh['sx'],
      sy: e.bokeh['sy'],
      delta: e.delta,
      model_id: model_id,
    })
  }
~~~
to 
~~~
  static from_event(e: any, model_id: string = null) {
    return new this({
      sx: e.bokeh['sx'],
      sy: e.bokeh['sy'],
      delta: e.bokeh['delta'],
      model_id: model_id,
    })
  }
~~~
fixes the problem, and allows the delta parameter to pass through.

There may be other issues with this, but it appears to simply be a typo. I took a quick look to see if I could add a test for this, but didn't come up with much.

- [x] issues: fixes #6238
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/core/bokeh_events.ts
+++ b/bokehjs/src/coffee/core/bokeh_events.ts
@@ -159,7 +159,7 @@ export class MouseWheel extends PointEvent {
     return new this({
       sx: e.bokeh['sx'],
       sy: e.bokeh['sy'],
-      delta: e.delta,
+      delta: e.bokeh['delta'],
       model_id: model_id,
     })
   }


mattpap:
>  but it appears to simply be a typo.

That's it. Although we use TypeScript here, lack of type information (`e: any`) allows for such simple mistakes.

bryevdv:
Thanks @schoolie !

PR #6260, opened 2017-05-08
Canavandl/5268 mapper type

### LINKED ISSUE #5268: Improve mappers' inheritance structure

canavandl:
- [x] issues: fixes #5268 
- [x] tests added / passed

supercedes reverted PR #6235


bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -45,8 +45,10 @@ export class CartesianFrame extends LayoutCanvas
...
     for name, range of ranges
-      if range.type == 'Range1d' or range.type == "DataRange1d"
-        s = if scale? then scale else new LinearScale()
+      if scale?
+        s = scale
+      else if range.type == 'Range1d' or range.type == "DataRange1d"
+        s = new LinearScale()
       else if range.type == "FactorRange"
         s = new CategoricalScale()
       else
```
bryevdv: As you pointed out this weirdness in the earlier PR. I'd say this is another reason why BokehJS should just always expect a real scale. There's no way for it to know to use a log scale without either sending the scale itself, or a hinting of some sort (what we were trying to get rid of).

canavandl: I agree

mattpap: On unrelated note, all those `range.type == "something"` should be rewritten as `instanceof` checks.

canavandl:
In terms of passing Scale models to BokehJS. Is it better to have a Plot.x_scale/y_scale default (of a LinearScale) or have a validation error (similar to the REQUIRED_RANGE if the scale props aren't set?

We can have some magic inside of ``figure`` to take care of this, but need to handle it for ``Plot``.

bryevdv:
> In terms of passing Scale models to BokehJS. Is it better to have a Plot.x_scale/y_scale default (of a LinearScale) or have a validation error (similar to the REQUIRED_RANGE if the scale props aren't set?

I'm not sure offhand. I guess it's worth trying with a default and seeing how that looks/seems, the main issue is just not having some usage that easily leads to a linear scale with a factor range

bokeh/core/validation/errors.py
```diff
@@ -38,10 +46,12 @@
     1001: ("BAD_COLUMN_NAME",        "Glyph refers to nonexistent column name"),
     1002: ("MISSING_GLYPH",          "Glyph renderer has no glyph set"),
     1003: ("NO_SOURCE_FOR_GLYPH",    "Glyph renderer has no data source"),
-    1004: ("REQUIRED_RANGE",         "A required Range objects missing"),
+    1004: ("REQUIRED_RANGE",         "A required Range object is missing"),
     1005: ("MISSING_GOOGLE_API_KEY", "Google now requires API keys for all Google Maps usage"),
     1006: ("NON_MATCHING_DATA_SOURCES_ON_LEGEND_ITEM_RENDERERS", "LegendItem.label is a field, but renderer data sources don't match"),
     1007: ("MISSING_MERCATOR_DIMENSION", "Mercator Tickers and Formatters must have their dimension property set to 'lat' or 'lon'"),
+    1008: ("REQUIRED_SCALE",         "A required Scale object is missing"),
```
canavandl: I'm unsure whether this should be it's own error or whether it could/should be folded into "INCOMPATIBLE_SCALE_AND_RANGE". The only time it'd ever come up (IIRC), is if both ranges and scales are set to None (overriding the default for the latter)

bokeh/models/plots.py
```diff
@@ -371,22 +416,45 @@ def _check_snapped_toolbar_and_axis(self):
     The (default) data range of the vertical dimension of the plot.
     """)
 
-    x_mapper_type = Either(Auto, String, help="""
-    What kind of mapper to use to convert x-coordinates in data space
-    into x-coordinates in screen space.
+    @classmethod
+    def _scale(cls, scale):
```
canavandl: note: doesn't handle case where user sets "x_mapper_type=None", though I'm not sure that anybody does that anyways.

bokeh/models/plots.py
```diff
@@ -371,22 +416,45 @@ def _check_snapped_toolbar_and_axis(self):
...
+    def y_mapper_type(self):
+        deprecated((0, 12, 6), "y_mapper_type", "y_scale")
+        return self.y_scale
+
+    @y_mapper_type.setter
+    def y_mapper_type(self, mapper_type):
+        deprecated((0, 12, 6), "y_mapper_type", "y_scale")
+        self.y_scale = self._scale(mapper_type)
+
+    x_scale = Instance(Scale, default=lambda: LinearScale(), help="""
```
canavandl: note: this adds a `LinearScale` default

bokeh/models/tests/test_plots.py
```diff
@@ -159,3 +161,73 @@ def test_plot_raises_error_if_toolbar_and_tools_are_set():
...
+    assert check == []
+
+
+def test__check_compatible_scale_and_ranges_incompat_numeric_scale_and_factor_range():
+    plot = Plot(x_scale=LinearScale(), x_range=FactorRange())
+    check = plot._check_compatible_scale_and_ranges()
+    assert check != []
+
+
+# def test__check_compatible_scale_and_ranges_incompat_factor_scale_and_numeric_range():
```
canavandl: There's some weirdness here because `CategoricalScale` is a subclass of `LinearScale`. Examples will work because a categorical_scale will use the linear_scale compute/invert methods if the passed values are numbers.

We need to decide we should prevent this situation on the python side though.

bokeh/plotting/figure.py
```diff
@@ -123,6 +125,9 @@ def __init__(self, *arg, **kw):
         self.x_range = _get_range(opts.x_range)
         self.y_range = _get_range(opts.y_range)
 
+        self.x_scale = _get_scale(self.x_range, opts.x_axis_type)
```
canavandl: This should clean up any range/scale incompatibilities.

bokeh/models/tests/test_plots.py
bryevdv: I am going to work on substantially re-working categoricals after this PR.

bokeh/plotting/helpers.py
```diff
@@ -200,6 +201,17 @@ def _get_range(range_input):
     raise ValueError("Unrecognized range input: '%s'" % str(range_input))
 
 
+def _get_scale(range_input, axis_type):
+    if isinstance(range_input, (DataRange1d, Range1d)) and axis_type in ["linear", "datetime", "auto"]:
+        return LinearScale()
+    elif isinstance(range_input, (DataRange1d, Range1d)) and axis_type == "log":
+        return LogScale()
+    elif isinstance(range_input, FactorRange):
```
canavandl: The current behavior to allow:

```python
figure(x_range=FactorRange(), x_axis_type="log") #or any axis type, i.e. linear/datetime
```

Though it may barf on the BokehJS side. Should we prevent users from getting into that state?

bokehjs/src/coffee/api/typings/models/plots.d.ts
```diff
@@ -55,8 +55,8 @@ declare namespace Bokeh {
   extra_x_ranges?: Map<Range>;
   extra_y_ranges?: Map<Range>;
 
-  x_mapper_type?: Auto | "log";
-  y_mapper_type?: Auto | "log";
+  x_scale?: Scale;
```
canavandl: I'm not sure what needs to be here.

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -42,24 +42,13 @@ export class CartesianFrame extends LayoutCanvas
...
-      else
-        logger.warn("unknown range type for range '#{name}': #{range}")
-        return null
-      scales[name] = new scale_type({
-        source_range: range
-        target_range: frame_range
-      })
+      s = scale.clone()
+      s.setv({source_range: range, target_range: frame_range})
+      s.connect_range_signals()
```
canavandl: This is ugly and could potentially be removed, see comment below

bokehjs/src/coffee/models/scales/linear_scale.coffee
```diff
@@ -4,21 +4,20 @@ import * as p from "core/properties"
...
 
-  connect_signals: () ->
-    super()
+  @getters {
+    state: () ->
+      if not @_state? then @_state = @_compute_state()
+      return @_state
+  }
+
+  connect_range_signals: () ->
```
canavandl: All of this signal connection stuff serves to merely cache the Scale state. I'm not sure there's much (if any) performance benefit to be gains from the indirection. We could do away with the getter and signals and instead call `_compute_state` every time `compute` or `invert` is called.

canavandl:
The only "migration note" worthy change is when users to

```python
p = Plot(x_range=FactorRange())
```

Because the default x/y_scales are LinearScale, this is going to emit a validation error saying the LinearScale/FactorRange are incompatible. Using the ``bokeh.plotting.figure`` should make sure they're compatible and avoid this error. 

This change seems acceptable and minimally invasive to me.

bryevdv:
I guess that's a reason a "fixup" step prior to serialization could be useful. There are some things that just can't be deduced for certain until the last moment and I think this is one of those things. How many examples do we have that would need to be modified to not emit this warning?

canavandl:
It's also unclear to me how the JS GMapPlot x_scale/y_scale defaults are Models and not LinearScales

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
canavandl: this could all be wrapped in a utility fxn, i guess (the clone, settings the ranges and the connects)

canavandl:
>I guess that's a reason a "fixup" step prior to serialization could be useful. There are some things that just can't be deduced for certain until the last moment and I think this is one of those things. How many examples do we have that would need to be modified to not emit this warning?

How about a third validation check level "fatal" (in addition to warn and error) that would raise a RuntimeError or activating ``settings.strict`` to raise exceptions on errors? I'm hesitant to implement a pre-serialization fixup because a) it seems difficult to intuit a user's intention if there's several ranges (and some are incompatible) and b) if feels too "magical" and like it's just be confusing.

Additionally, we need to figure out a more graceful for BokehJS to fail with an incompat scale/range. Right now it's an unclear exception inside of Axis.get_ticks.

bryevdv:
>  I'm hesitant to implement a pre-serialization fixup because a) it seems difficult to intuit a user's intention if there's several ranges (and some are incompatible) 

To be clear I am proposing only to do exactly what is already done in BokehJS, only to do it in python instead of JS.

canavandl:
>To be clear I am proposing only to do exactly what is already done in BokehJS, only to do it in python instead of JS.

I guess it's not clear to me what is already done in BokehJS. Can you elaborate?

bryevdv:
I guess I should have specified, what was *previously* done in BokehJS, i.e.
```
-      if range.type == "Range1d" or range.type == "DataRange1d"
 -        if scale_type_name == "log"
 -          scale_type = LogScale
 -        else
 -          scale_type = LinearScale
 -        range.scale_hint = scale_type_name
 -      else if range.type == "FactorRange"
 -        scale_type = CategoricalScale
 -      else
 -        logger.warn("unknown range type for range '#{name}': #{range}")
 -        return null
 -      scales[name] = new scale_type({
 -        source_range: range
 -        target_range: frame_range
 -      })
```

canavandl:
I'm -1 on that solution because it gets away from our "one scale per dimension goal" and feels non-intuitive.

It allows having both FactorRanges and (an implicitly linear) Range1d on the same dimension, unless the scale is specified, then you can only have one or the other? Also, it can't be a combination of a LogScale and CategoricalScale or LogScale and LinearScale?

```
-      if range.type == "Range1d" or range.type == "DataRange1d"
 -        if scale_type_name == "log"
 -          scale_type = LogScale
 -        else
 -          scale_type = LinearScale
 -        range.scale_hint = scale_type_name
 -      else if range.type == "FactorRange"
 -        scale_type = CategoricalScale
 -      else
 -        logger.warn("unknown range type for range '#{name}': #{range}")
 -        return null
 -      scales[name] = new scale_type({
 -        source_range: range
 -        target_range: frame_range
 -      })
```

How about mirroring the same scale/range validation and throwing an Error if there's an incompatibility. In that case the message would be clear to the user.

bryevdv:
> I'm -1 on that solution because it gets away from our "one scale per dimension goal" and feels non-intuitive.

I'm definitely not being precise enough if I gave the impression of supporting more than one scale per dimension. Maybe another way to say things:

> I am proposing that no examples should need to be updated or changed as a result of this effort. 

Or in code, in particular, that:

```
p = Plot(x_range=FactorRange())
```
work without warning or error as it does now. 

That would mean letting Python code decide to install a `CategoricalScale` and wire it up. And since `p.x_range = whatever` is possible at any time, it also means that decision has to happen as late as possible. Orthogonal and independent to this, there are combinations of ranges and extra ranges that don't make sense (that we could issue validation errors or warning about).

----

That said, if 
```
p = Plot(x_range=FactorRange())
``` 
generating a validation message (unless the user sets up the right scale too) is the only divergence with this PR then maybe it can be lived with, or addressed later in the context of other categorical work.

canavandl:
Thanks for the clarification, it helped distill our point of disagreement.

As a self-described bokeh.models API power-user, I wouldn't want Bokeh to try and infer the scale from the range in the case of ``p = Plot(x_range=FactorRange())``. I'd rather have a validation error or other hard error. When using that API, I think it's best to have zero "magic" and make everything be very explicit (I feel like I'm developing the JSON graph in my head).

I'm totally ok w/ the ``bokeh.plotting`` API inferring the scale, but I'd be reluctant to add similar magic to bokeh.models.

bryevdv:
> I'm totally ok w/ the bokeh.plotting API inferring the scale, but I'd be reluctant to add similar magic to bokeh.models.

Fair enough

bryevdv:
> I'm totally ok w/ the bokeh.plotting API inferring the scale, but I'd be reluctant to add similar magic to bokeh.models.

Actually I have talked myself out of thinking this is sufficient. It does not address the fact that users can update ranges at any time, not just in the call to `figure`:
```
p  = figure()
p.x_range = FactorRange()  # broken combination now
```
The only way to make that work (and I think it has to work) is to fix things up as late as possible. nd if we do that we might as well do it for `Plot` too.

canavandl:
Just to be clear, breaking things by changing the range at the end of a script is the current Bokeh behavior. The below raises an obtuse BokehJS Ticker error.

```

p = figure(x_mapper_type='linear')
p.scatter(x=[1,10,100,1000], y=[1,2,3,4])

p.x_range = FactorRange()

show(p)
```

I'd argue that this isn't a huge problem because I haven't seen a ton of user reports about this issue, probably because it's a pretty obvious instance of user error. I think it's friendly to raise a transparent validation error about incompatible ranges/scales than trying to auto-magically clean things up.

bryevdv:
That's an apples-and-oranges comparison. My point is that *this* does work now:
```

p = figure()
p.scatter(x=['a', 'b', 'c', 'd'], y=[1,2,3,4])

p.x_range = FactorRange('a', 'b', 'c', 'd')

show(p)
```
but will *not* work subsequently. I think that's pretty unfriendly to users but if you think it's not a big issue then I will defer to that.

canavandl:
I would argue that your examples *doesn't* work now. In master it creates a plot with a LinearAxis with numeric values (and not a Categorical one) that looks like:

screen shot 2017-05-12 at 3 19 04 pm

Then when you pan the plot, the axis changes (I think to numeric factors?) but is still wrong:

screen shot 2017-05-12 at 3 20 12 pm

I think the validation error is more user friendly that allowing users to get into the above state.

bryevdv:
@canavandl OK you've convinced me, what is left for this PR now that that is decided?

canavandl:
I'm currently fixing up the broken examples (there are 2 or 3). Can you look at the GMapPlot default issue that's failing the tests? I can't figure it out.

bryevdv:
Def, checking it out now

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -42,24 +44,23 @@ export class CartesianFrame extends LayoutCanvas
...
-        source_range: range
-        target_range: frame_range
-      })
+      if range instanceof DataRange1d or range instanceof Range1d
+        if not scale instanceof LogScale or not scale instanceof LinearScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
+      if range instanceof FactorRange
+        if not scale instanceof CategoricalScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
+      if scale instanceof LogScale and range instanceof DataRange1d
```
canavandl: This PR punts on the DataRange1d.scale_hint stuff. It's going to have to be addressed in it's own PR, especially if we intend to support custom scales.

bokeh/models/tests/test_plots.py
canavandl: fixed (for now)

bryevdv:
I am not sure what's going on here either! @mattpap do you have any ideas? There are other inherited computed instance defaults (e.g `tool_events`) that *do* pass the defaults comparison.

bryevdv:
ping @mattpap this PR is otherwise ready to go, can you see if you an offer any help with the defaults mismatch issue? It's possible that computed, inherited defaults are simply being ignored in some instance e.g. tickers for datetime ticker, but I don't think that is always the case.

mattpap:
@canavandl, `LinearScale` class in bokehjs doesn't have `type` attribute. Make sure all other classes in `models/scales` have it.

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -42,24 +44,28 @@ export class CartesianFrame extends LayoutCanvas
...
+      if range instanceof FactorRange
+        if scale not instanceof CategoricalScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
+
+      if scale instanceof LogScale and range instanceof DataRange1d
+        range.scale_hint = "log"
+
+      s = scale.clone()
+      s.setv({source_range: range, target_range: frame_range})
+      s.connect_range_signals()
```
mattpap: You are introducing a memory leak here, because you're connecting new slots every time `_get_scales()` is called. Either clear old connections or make sure that you always connect the same slots. Note that you don't have to provide a cleanup procedure if you connect only once, because all slots are disconnected when model is destroyed or view removed.

bokehjs/src/coffee/models/canvas/cartesian_frame.coffee
```diff
@@ -42,24 +44,28 @@ export class CartesianFrame extends LayoutCanvas
...
+      if range instanceof DataRange1d or range instanceof Range1d
+        if scale not instanceof LogScale and scale not instanceof LinearScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
+        # special case because CategoricalScale is a subclass of LinearScale, should be removed in future
+        if scale instanceof CategoricalScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
+
+      if range instanceof FactorRange
+        if scale not instanceof CategoricalScale
+          throw new Error("Range #{range.constructor.name} is incompatible is Scale #{scale.constructor.name}")
```
mattpap: `Function.name` is not widely supported and it's unreliable under minification. Use `model.type` instead.

canavandl: fixed

bokehjs/src/coffee/models/scales/log_scale.coffee
```diff
@@ -2,24 +2,10 @@ import {Scale} from "./scale"
...
-  @getters {
-    state: () ->
-      if not @_state? then @_state = @_compute_state()
-      return @_state
-  }
+  type: "LogScale"
 
   compute: (x) ->
-    [factor, offset, inter_factor, inter_offset] = @state
+    [factor, offset, inter_factor, inter_offset] = @_compute_state()
```
mattpap: So this doesn't have significant negative impact on performance?

canavandl: As far as I can tell, this doesn't meaningfully impact performance. The "slow" part of mapping is converting a large array from data space into screen space, not computing the mapper state once at the beginning.

bokehjs/src/coffee/api/typings/models/plots.d.ts
mattpap: This is correct.

bokehjs/src/coffee/models/scales/log_scale.coffee
bryevdv: I'm willing to believe the overhead is negligible when using `v_compute`. It's also worth noting that the cache is *currently* invalidated very often, every time there is any pan or zoom at all, e.g., and performance is fine. The code around this has change considerably so I'm also fine taking a step bak and doing the simple thing, and re-introducing caching with a fresh start later if it is determined to be necessary.

sphinx/source/docs/releases/0.12.6.rst
```diff
@@ -38,10 +38,14 @@ Old                            New
...
-are not expected to impact any user code.
-
-Additionally, these new scales models are now accessible from the Python.
+Previously, these Mapper classes were internal implementation details. The
+new Scale classes are first-class Bokeh models and are accessible from
+Python.
+
+There is a new Plot validation check to ensure that Scales and Ranges on a
+dimension are compatible. Users employing a ``FactorRange`` in the
+``bokeh.plotting`` API will have to specify ``CategoricalScale`` on the same
```
bryevdv: meant ``bokeh.models`` here? Things are handled automagically in ``bokeh.plotting`` for most cases? I think its worth showing an example line or two of code here to demonstrate exactly what users will need to change, and possibly a link to a complete example that does this.

PR #6262, opened 2017-05-09
properly check for ImageURL w, h for None

### LINKED ISSUE #6095: `ImageURL.{w,h}` use data units by default
This is broken yet again. `w` and `h` uses `DistanceSpec` which uses `data` units by default. Even help entries for both of those properties say that values are in pixels.

bryevdv:
- [x] issues: fixes #6095
- [x] tests added / passed

The issue was that the view properties like `@_w` are now converted to lists `[null]` when `@model.w` is `null` (or `None` on the python side). I have no opinion on whether this is reasonable (it could probably be argued either way). So this PR fixes things by checking `@model.w` directly, which is an unequivocal and always-correct source of truth in this question. 

I have also removed the note about possibly changing the property names to `dw` and `dh`. That ship has sailed. 

Finally, I changed the example to pass an empty CDS to the last "values-only" call to `ImageURL`. This is because we normally "broadcast" values to the size of the CDS. this results in the same image being rendered 5 times. The broadcasting is a reasonable thing in general. It could be simply considered a usage error to pass a CDS with data when there are not data specs with fields. (Maybe a validation warning would be advised?) Otherwise, to avoid this we would need to check for and special-case whenever a CDS has data but a glyph has all "value" data specs. 

I added a unit test to confirm that when `model.w` and `model.h` are `null` then expected `NaN` is generated for `view.sw` and `view.sh` but this is not ideal. Perhaps a strict image diff test?

screen shot 2017-05-09 at 08 34 26

### DIFF
--- a/examples/models/file/image_url.py
+++ b/examples/models/file/image_url.py
@@ -34,7 +34,7 @@ image2 = ImageURL(url="url", x="x2", y="y2", w=20, h=20, anchor="top_left")
 plot.add_glyph(source, image2)
 
 image3 = ImageURL(url=dict(value=url), x=200, y=-100, anchor="bottom_right")
-plot.add_glyph(source, image3)
+plot.add_glyph(image3)
 
 xaxis = LinearAxis()
 plot.add_layout(xaxis, 'below')
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -467,28 +467,18 @@ class ImageURL(Glyph):
     The y-coordinates to locate the image anchors.
     """)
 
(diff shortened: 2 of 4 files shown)


examples/models/file/image_url.py
```diff
@@ -34,7 +34,7 @@
 plot.add_glyph(source, image2)
 
 image3 = ImageURL(url=dict(value=url), x=200, y=-100, anchor="bottom_right")
-plot.add_glyph(source, image3)
+plot.add_glyph(ColumnDataSource(), image3)
```
mattpap: No need for dummy data source, `plot.add_glyph(image3)` will work.

PR #6270, opened 2017-05-10
Make bokehjs work in IE again and improve tests

### LINKED ISSUE #6212: Issue with test build 0.12.6dev3 on Windows 7 with IE11
I encountered an issue with test build 0.12.6dev3 uisng IE11 on Windows 7 with Python 3.5.3

Using the code below from the User Guide Getting Started page.

```python

# prepare some data
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]

# output to static HTML file
output_file("lines.html")

# show the results
show(p)
```

Expected result
good_simple_line

Actual result:

bokeh_obj_doesnt_support

In Chrome, with the above configuration, the result is as expected.

Build 0.12.6dev2 does not have this issue.

mattpap:
fixes #6212

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -589,8 +589,11 @@ export class PlotCanvasView extends DOMView
 
     ctx.restore()  # Restore to default state
 
-    event = new Event("bokeh:rendered", {detail: @})
-    window.dispatchEvent(event)
+    try
+      event = new Event("bokeh:rendered", {detail: @})
+      window.dispatchEvent(event)
+    catch
+      # new Event() is not supported on IE.
 
   _on_resize: () ->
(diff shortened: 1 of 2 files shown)


bryevdv:
What versions of IE do not support `Event`?  having a "renderered" event that e.g. `js_on_event` can respond to would be useful on all platforms.

mattpap:
This is event is for testing purpose, so it doesn't matter. This will be replaced anyway with proper rendering heuristic. Event constructors are supported only in Edge. It's possible to construct those events in IE with other methods, but once again, no need for this.

al6ayb:
Hi,
the issue of IE not supported still persists. Even on 0.12.6. Latest I could manage to produce working output on IE is on 0.12.6dev2

bryevdv:
@al6ayb please make a new separate issue, with more information (exact library, OS, browser versions) as well as JavaScript console output

PR #6278, opened 2017-05-11
Allow datetime values for number specs by default

### LINKED ISSUE #5499: Datetimes on plot are always treated as local time and shifted to UTC
##### Description
When displaying `datetime.datetime` values along the x-axis using `x_axis_type = "datetime"`, the values are shifted if the system's timezone is not UTC. 

For example, if you live in UTC-05:00, then all the dates are shifted ahead 5 hours (7am appears at noon).

This happens regardless of the `tzinfo` on the datetimes, including for naive datetimes where `tzinfo` is `None`.

##### Sample Code
```
from datetime import datetime
from bokeh.models.tools import CrosshairTool, HoverTool

hours = range(24)
datetimes = [datetime.now().replace(hour=h, minute=0, second=0, tzinfo=None) for h in hours]

example = figure(
	x_axis_label = 'Date/Time',
	x_axis_type = 'datetime',
	y_axis_label = 'Expected Hour',
)
example.circle(
	x = 'x',
	y = 'y',
	fill_color = 'blue',
	size = 6,
	source = ColumnDataSource({
		'x': datetimes,
		'y': hours,
		'time_fmt': [dt.strftime("%H:%M") for dt in datetimes],
	}),
)
example.add_tools(CrosshairTool())
example.add_tools(HoverTool(
	tooltips = [
		("index", "$index"),
		("expected time", "@time_fmt"),
	],
))

show(example)
```
##### Expected Behaviour
I would have expected that naive datetimes are displayed as is.

##### Actual Behaviour
screenshot at 2016-12-01 12 53 01

##### System, Library, Browser, Version information
I am using Python 3.5.2 (64-bit) with bokeh 0.12.3 and Firefox 50.0 as well as Chrome 54.0.2840.100 (64-bit) on Ubuntu MATE 16.04.1 LTS and on CentOS release 6.8.

bokeh info (Ubuntu):
```
Python version      :  3.5.2 (default, Sep 10 2016, 08:21:44) 
IPython version     :  5.1.0
Bokeh version       :  0.12.3
BokehJS static path :  /usr/local/lib/python3.5/dist-packages/bokeh/server/static
```

bokeh info (CentOS):
```
Python version      :  3.5.2 (default, Nov 24 2016, 13:03:33) 
Bokeh version       :  0.12.3
BokehJS static path :  /usr/local/lib/python3.5/site-packages/bokeh/server/static
```
##### Other notes
You can work around this issue by setting your system timezone to UTC. This is not ideal though.

### LINKED ISSUE #5831: Numberspec coordinates should accept datetimes
The following example should ideally make a simple plot and place a label using data space coordinates:

```
from bokeh.plotting import show, output_file, figure
from datetime import datetime

f = figure(x_axis_type="datetime")
f.line(x=[datetime(2015, 1, 1), datetime(2016, 1, 1)], y=[1,2])

label = Label(x=datetime(2015, 1, 1), y=2, text='Some label')
f.add_layout(label)

output_file('test.html')
show(f)

>>> ValueError: expected a value of type Real, got 2015-01-01 00:00:00 of type datetime
```

This seems to fail because Label coordinates do not accept datetime even tough it seems natural. The workaround is to convert the datetime coordinates to milliseconds since epoch, and try again:

```
from bokeh.plotting import show, output_file, figure
from datetime import datetime

f = figure(x_axis_type="datetime")
f.line(x=[datetime(2015, 1, 1), datetime(2016, 1, 1)], y=[1,2])

x_label = datetime(2015, 1, 1).timestamp() * 1000

label = Label(x=x_label, y=2, text='Some label')
f.add_layout(label)

output_file('test.html')
show(f)
```

## Desired feature:
The suggestion is to have bokeh perform this conversion automatically when asked to use a datetime object as a coordinate, as this would be the most intuitive interaction when coordinates are specified in data space units.

```
Python version  :   3.4.5 |Continuum Analytics, Inc.|(default, Jul 5 2016, 14:53:06)  [MSC v.1600 64bit (AMD64)]
IPython version :   (not installed)
Bokeh version   :   0.12.4
node.js version :   v0.12.2
```

### LINKED ISSUE #6274: FutureWarning from pandas in BokehJSONEncoder

pandas >= 0.20.0
bokeh master

On master:

```python
In [1]: from bokeh.core.json_encoder import BokehJSONEncoder
imp
In [2]: import pandas as pd

In [3]: BokehJSONEncoder().default(pd.Timestamp('2017'))
 /Users/taugspurger/sandbox/bokeh/bokeh/core/json_encoder.py:73: FutureWarning: pandas.tslib is deprecated and will be removed in a future version.
You can access Timestamp as pandas.Timestamp
  if pd and isinstance(obj, pd.tslib.Timestamp):
Out[3]: 1483228800000.0
```

PR inbound.

bryevdv:
This PR adds a parameter `accept_datetime` to NumberSpec that defaults to True. When enabled, NumberSpecs will accept and immediately convert datetime values to serialized ms-since-epoch values. This allows glyphs and annotations to be positioned using datetime values.

NumberSpec was also modified to always accept and convert timedelta objects. 

This also fixes a future warning from pandas regarding `pd.tslib`

Datetime/timedelta detection and conversion was moved to `bokeh.util.serializtion` so that the code might be shared. As part of this, the handling of Python `datetime` values was corrected to match the handling of NumPy `datetime64`

- [x] issues: fixes #6274, fixes #5831 fixes #5499
- [x] tests added / passed

@mattpap @canavandl I would appreciate a close review of this PR

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -106,7 +106,7 @@ from six import string_types, iteritems
 
 from ..colors import RGB
 from ..util.dependencies import import_optional
-from ..util.serialization import transform_column_source_data, decode_base64_dict
+from ..util.serialization import convert_datetime_type, decode_base64_dict, transform_column_source_data
 from ..util.string import nice_join, format_docstring
 
 from .property.bases import ContainerProperty, DeserializationError, ParameterizedProperty, Property, PrimitiveProperty
@@ -636,6 +636,9 @@ class Either(ParameterizedProperty):
             return self._type_params[0]._raw_default()
         default = kwargs.get("default", choose_default)
         super(Either, self).__init__(default=default, help=help)
+        self.alternatives = []
+        for tp in self._type_params:
+            self.alternatives.extend(tp.alternatives)
 
     # TODO (bev) get rid of this?
     def __or__(self, other):
@@ -1029,7 +1032,7 @@ class Angle(Float):
     pass
 
 class Date(Property):
-    ''' Accept Date (not datetime) values.
+    ''' Accept Date (but not DateTime) values.
 
     '''
     def __init__(self, default=datetime.date.today(), help=None):
@@ -1455,7 +1458,7 @@ class DataSpec(Either):
         property.
 
         Args:
-            name (str) : the name of the property these descriptors are for
+            base_name (str) : the name of the property these descriptors are for
 
         Returns:
             list[DataSpecPropertyDescriptor]
@@ -1491,7 +1494,13 @@ class DataSpec(Either):
 _FieldValueTransform = Enum("field", "value", "transform")
 
 class NumberSpec(DataSpec):
-    ''' A |DataSpec| property that accepts numeric fixed values.
+    ''' A |DataSpec| property that accepts numeric and datetime fixed values.
+
+    By default, date and datetime values are immediately converted to
+    milliseconds since epoch. It it possible to disable processing of datetime
+    values by passing ``accept_datetime=False``.
+
+    Timedelta values are interpreted as absolute milliseconds.
 
     .. code-block:: python
 
@@ -1500,8 +1509,12 @@ class NumberSpec(DataSpec):
         m.location = "foo" # field
 
     '''
-    def __init__(self, default=None, help=None, key_type=_FieldValueTransform):
+    def __init__(self, default=None, help=None, key_type=_FieldValueTransform, accept_datetime=True):
         super(NumberSpec, self).__init__(key_type, Float, default=default, help=help)
+        self.accepts(TimeDelta, convert_datetime_type)
+        if accept_datetime:
+            self.accepts(Datetime, convert_datetime_type)
+
 
 class StringSpec(DataSpec):
     ''' A |DataSpec| property that accepts string fixed values.
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -10,13 +10,14 @@ from ..core.enums import (accept_left_right_center, AngleUnits, DeprecatedLegend
                           FontStyle, LegendClickPolicy, LegendLocation, Orientation, RenderMode,
                           SpatialUnits, TextAlign)
 from ..core.has_props import abstract
-from ..core.properties import (Angle, AngleSpec, Auto, Bool, ColorSpec, DistanceSpec, Either, Enum, Float,
-                               FontSizeSpec, Include, Instance, Int, List, NumberSpec, Override,
+from ..core.properties import (Angle, AngleSpec, Auto, Bool, ColorSpec, Datetime, DistanceSpec, Either,
+                               Enum, Float, FontSizeSpec, Include, Instance, Int, List, NumberSpec, Override,
                                Seq, String, StringSpec, Tuple, value)
 from ..core.property_mixins import FillProps, LineProps, TextProps
 from ..core.validation import error
 from ..core.validation.errors import BAD_COLUMN_NAME, NON_MATCHING_DATA_SOURCES_ON_LEGEND_ITEM_RENDERERS
 from ..model import Model
+from ..util.serialization import convert_datetime_type
 
 from .formatters import BasicTickFormatter, TickFormatter
 from .mappers import ContinuousColorMapper
@@ -402,6 +403,9 @@ class BoxAnnotation(Annotation):
 
     left = Either(Auto, NumberSpec(), default=None, help="""
     The x-coordinates of the left edge of the box annotation.
+
+    Datetime values are also accepted, but note that they are immediately
+    converted to milliseconds-since-epoch.
     """)
 
     left_units = Enum(SpatialUnits, default='data', help="""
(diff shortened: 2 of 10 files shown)


bryevdv:
Example with Box annotation
```
from datetime import *

output_file("box_annotation.html", title="box_annotation.py example")

#reduce data size

left_box = BoxAnnotation(plot=p, right=date(2010,10,7), fill_alpha=0.1, fill_color='blue')
mid_box = BoxAnnotation(plot=p, left=date(2010,10,8), right=date(2010,10,9), fill_alpha=0.1, fill_color='yellow')
right_box = BoxAnnotation(plot=p, left=date(2010,10,10), fill_alpha=0.1, fill_color='blue')

p.renderers.extend([left_box, mid_box, right_box])

p.xaxis.axis_label = 'Time'
p.yaxis.axis_label = 'Value'

show(p)
```
screen shot 2017-05-11 at 10 46 36

bryevdv:
Example with label:

```
from bokeh.plotting import show, output_file, figure
from datetime import datetime

f = figure(x_axis_type="datetime")
f.line(x=[datetime(2015, 1, 1), datetime(2016, 1, 1)], y=[1,2])

label = Label(x=datetime(2015, 1, 1), y=2, text='Some label')
f.add_layout(label)

output_file('test.html')
show(f)
```
screen shot 2017-05-11 at 10 46 16

bokeh/core/properties.py
```diff
@@ -636,6 +636,9 @@ def choose_default():
             return self._type_params[0]._raw_default()
         default = kwargs.get("default", choose_default)
         super(Either, self).__init__(default=default, help=help)
+        self.alternatives = []
+        for tp in self._type_params:
+            self.alternatives.extend(tp.alternatives)
```
bryevdv: This change expands `Either` to consider any `.accept` calls that its type params may have made

bokeh/models/annotations.py
```diff
@@ -543,7 +556,10 @@ class Label(TextAnnotation):
 
     x = Float(help="""
     The x-coordinate in screen coordinates to locate the text anchors.
-    """)
+
+    Datetime values are also accepted, but note that they are immediately
+    converted to milliseconds-since-epoch.
+    """).accepts(Datetime, convert_datetime_type)
```
bryevdv: Note that the coords for label had to have an `accepts` added explicitly since they are not number specs to begin with. Perhaps should have/should change label to be vectorized like other annotations.

bokeh/util/serialization.py
```diff
@@ -7,55 +7,132 @@
...
-        np.dtype(np.uint8),
-        np.dtype(np.int8),
-        np.dtype(np.uint16),
-        np.dtype(np.int16),
-        np.dtype(np.uint32),
-        np.dtype(np.int32),
-    ])
-except ImportError:
-    is_numpy = False
-    BINARY_ARRAY_TYPES = set()
```
bryevdv: NumPy was imported directly in `json_encoder.py` for some time which AFAIK meant Bokeh could not function in any way without numpy. I think we should give up the pretense that Bokeh can be used without numpy.

bokeh/util/tests/test_serialization.py
```diff
@@ -19,6 +20,46 @@ def test_id_with_simple_ids():
...
+
+def test_convert_datetime_type():
+    assert bus.convert_datetime_type(datetime.datetime(2016, 5, 11)) == 1462942800000.0
+    assert bus.convert_datetime_type(datetime.timedelta(3000)) == 259200000000.0
+    assert bus.convert_datetime_type(datetime.date(2016, 5, 11)) == 1462942800000.0
+    assert bus.convert_datetime_type(datetime.time(3, 54)) == 14040000.0
+    assert bus.convert_datetime_type(np.datetime64("2016-05-11")) == 1462924800000.0
+    assert bus.convert_datetime_type(np.timedelta64(3000, 'ms')) == 3000.0
+    assert bus.convert_datetime_type(pd.Timedelta("3000ms")) == 3000.0
+    assert bus.convert_datetime_type(bus._pd_timestamp(3000000)) == 3.0
```
bryevdv: BTW some of these datetime tests for 2016-05-11 seem to give different results depending on the local TZ setting? AFAIK `np.datetime64` is always UTC, but I gather `datetime.datetime` may not be

bryevdv:
@mattpap @canavandl these tests fail on travis, but pass on my machine:
```
    
        f.dt = datetime.datetime(2016, 5, 11)
>       self.assertEqual(f.dt, 1462942800000.0)
E       AssertionError: 1462924800000.0 != 1462942800000.0
```
On my machine the 1462942800000.0 is returned and the test passes. I can only conclude this has something to do with local timezone settings. I will try updating the calls to `datetime` with `tzinfo=None` but is that the correct solution? (assuming it works)

bryevdv:
Not actually surprising, but adding `tzinfo=None` did not fix the test failure.  I guess this is actually #5499 ? I'd like to understand and fix that here as well. Suggestions welcome.

bryevdv:
I just pushed a change that should also fix #5499 It will require a migration note, because it seems we have been handing built in datetime values badly forever. I would appreciate close scrutiny of this entire PR @bokeh/dev

canavandl:
This change seems correct to me. The box annotation example now aligns the boxes on the day borders now:

screen shot 2017-05-12 at 5 54 53 pm

bertop89:
I tried this in Bokeh 0.12.9 and the problem is still there. Has this been fixed in any of the latest releases? Thanks!

PR #6281, opened 2017-05-11
Add plumbing for streaming and patching annoations

### LINKED ISSUE #4872: Arrows not updating from streaming data source
Hi, maybe I am doing it all wrong (I assume the documentation is WIP for app plots, events and stuff), but I am trying to plot arrows that are not annotations (my exact use case is drawing a directed graph)  and I can't get it to work.
The graph changes over time so I would like to write something like what I would do with Segments:

``` python
renderer = plot.arrow(end=Arrow(...), x0='x0', y0='y0', x1='x1', y1='y1', source=links)
```

Unfortunately, writing `plot.add_layout(Arrow(...), ..., source=links))` doesn't work. It don't know what exactly goes wrong, but nothing gets displayed. It might be that the render isn't updated as the source changes (I am updating it with `source.stream(new_data)` and the source is initially empty).

I am going to investigate to do it with something like [custom glyphs](http://bokeh.pydata.org/en/latest/docs/user_guide/extensions.html#putting-it-together) but as I am not really fluent in js/coffee I may not achieve it that way either (that's why I am bringing this up here). Did I miss something? Is there a reason why arrows are annotations and not glyphs?

bryevdv:
fixes #4872

Arrow, Band, and LabelSet were missing signal connections to `source.streaming` and `source.patching` necessary for these annotations to update when their data sources are streamed or patched.

This PR adds the missing signal connections. When reasonable, a `request_render` was moved to `set_data` (this is ok when there is not any CSS/dom rendering path)

Test case:

```
from random import randrange
from bokeh.io import curdoc
from bokeh.models import ColumnDataSource, Arrow, NormalHead

plot = figure(x_range=(0, 10), y_range=(0, 10))
source = ColumnDataSource(data={'x0': [], 'y0': [], 'x1': [], 'y1': []})
# if the plot only contains annotations it screws up range and scrolling
plot.circle(x=0, y=0)
plot.add_layout(
        Arrow(end=NormalHead(size=4), x_start='x0', y_start='y0', x_end='x1',
        y_end='y1', source=source))

def callback():
    source.stream({k: [randrange(10) for _ in range(5)] for k in ('x0', 'y0', 'x1', 'y1')})

curdoc().add_root(layout([[plot]]))
curdoc().add_periodic_callback(callback, 1000)
```

Results in:

screen shot 2017-05-11 at 18 21 05

### DIFF
--- a/bokehjs/src/coffee/models/annotations/arrow.coffee
+++ b/bokehjs/src/coffee/models/annotations/arrow.coffee
@@ -15,13 +15,14 @@ export class ArrowView extends AnnotationView
   connect_signals: () ->
     super()
     @connect(@model.change, () => @plot_view.request_render())
-    @connect(@model.source.change, () ->
-      @set_data(@model.source)
-      @plot_view.request_render())
+    @connect(@model.source.streaming, () -> @set_data(@model.source))
+    @connect(@model.source.patching, () -> @set_data(@model.source))
+    @connect(@model.source.change, () -> @set_data(@model.source))
 
   set_data: (source) ->
     super(source)
     @visuals.warm_cache(source)
+    @plot_view.request_render()
 
   _map_data: () ->
     if @model.start_units == 'data'
--- a/bokehjs/src/coffee/models/annotations/label_set.coffee
+++ b/bokehjs/src/coffee/models/annotations/label_set.coffee
@@ -22,11 +22,23 @@ export class LabelSetView extends TextAnnotationView
       @connect(@model.change, () ->
         @set_data(@model.source)
         @render())
+      @connect(@model.source.streaming, () ->
+        @set_data(@model.source)
+        @render())
+      @connect(@model.source.patching, () ->
+        @set_data(@model.source)
+        @render())
       @connect(@model.source.change, () ->
         @set_data(@model.source)
         @render())
     else
       @connect(@model.change, () ->
+        @set_data(@model.source)
+        @plot_view.request_render())
+      @connect(@model.source.streaming, () ->
+        @set_data(@model.source)
+        @plot_view.request_render())
+      @connect(@model.source.patching, () ->
         @set_data(@model.source)
         @plot_view.request_render())
       @connect(@model.source.change, () ->
--- a/bokehjs/src/coffee/models/annotations/band.coffee
+++ b/bokehjs/src/coffee/models/annotations/band.coffee
@@ -10,13 +10,14 @@ export class BandView extends AnnotationView
 
   connect_signals: () ->
     super()
-    @connect(@model.source.change, () ->
-      @set_data(@model.source)
-      @plot_view.request_render())
+    @connect(@model.source.streaming, () -> @set_data(@model.source))
+    @connect(@model.source.patching, () -> @set_data(@model.source))
+    @connect(@model.source.change, () -> @set_data(@model.source))
 
   set_data: (source) ->
     super(source)
     @visuals.warm_cache(source)
+    @plot_view.request_render()
 
   _map_data: () ->
     x_scale = @plot_view.frame.xscales[@model.x_range_name]


bokehjs/src/coffee/models/annotations/label_set.coffee
```diff
@@ -22,13 +22,25 @@ export class LabelSetView extends TextAnnotationView
...
     else
       @connect(@model.change, () ->
         @set_data(@model.source)
         @plot_view.request_render())
+      @connect(@model.source.streaming, () ->
+        @set_data(@model.source)
+        @plot_view.request_render())
+      @connect(@model.source.patching, () ->
+        @set_data(@model.source)
+        @plot_view.request_render())
```
bryevdv: @mattpap is it possible to connect multiple signals to the same callback with one call to `@connect`

bokehjs/src/coffee/models/annotations/arrow.coffee
```diff
@@ -15,13 +15,14 @@ export class ArrowView extends AnnotationView
...
-      @set_data(@model.source)
-      @plot_view.request_render())
+    @connect(@model.source.streaming, () -> @set_data(@model.source))
+    @connect(@model.source.patching, () -> @set_data(@model.source))
+    @connect(@model.source.change, () -> @set_data(@model.source))
 
   set_data: (source) ->
     super(source)
     @visuals.warm_cache(source)
+    @plot_view.request_render()
```
canavandl: since `set_data` is called in the arrow `initialize` statement, is the arrow rendered twice on the page load? (and same with the other annotations?)

bokehjs/src/coffee/models/annotations/label_set.coffee
mattpap: Currently not, but you wouldn't depend on this anyway. There is an intrinsic order of running callbacks (slots), which is in order of connection, but it's preferable that if you have a chain of actions happening in order for a given signal, you should state this explicitly like in this code. There are no other guarantees, so there may be interruptions between slots or even batches of slots running in different stack frames.

bokehjs/src/coffee/models/annotations/arrow.coffee
bryevdv: `request_render` actually precipitates a call to `throttled_render` so I'm not overly worried about it. I guess I'd rather have the mental model that  setting new data always implies a corresponding request to render in any case. `set_data` doesn't actually trigger any changes in the data source, so that can't be what was causing a render previously anyway. My guess is that lots of things are requesting renders around startup. But repeated calls to `request_render` within the throttle window should have no effect. (That was really the intent of `request_render`, to reduce the bookkeeping burden on developers).

I think if we wanted to prevent multiple renders on startup the cleanest way would just be to have a synchronization condition. I.e. *no* render ever happens until some condition is met, such as "all models have passed initialization" then after condition is met, the  flag blocking `render` is flipped, and a `request_render` is issued.

bryevdv: Meant to add: and maybe that's worth doing, but it should be a separate issue.

bokehjs/src/coffee/models/annotations/label_set.coffee
bryevdv: Fair enough. What about this to clean things up some then?
```
index 0981da9e8..d57912e91 100644
--- a/bokehjs/src/coffee/models/annotations/label_set.coffee
+++ b/bokehjs/src/coffee/models/annotations/label_set.coffee
@@ -8,15 +8,21 @@ export class LabelSetView extends TextAnnotationView
   initialize: (options) ->
     super(options)

-    @set_data(@model.source)
-
     if @model.render_mode == 'css'
       for i in [0...@_text.length]
         @title_div = div({class: 'bk-annotation-child', style: {display: "none"}})
         @el.appendChild(@title_div)

+    @set_data(@model.source)
+
   connect_signals: () ->
     super()
+    @connect(@model.change, () => @plot_view.request_render())
+    @connect(@model.source.streaming, () -> @set_data(@model.source))
+    @connect(@model.source.patching, () -> @set_data(@model.source))
+    @connect(@model.source.change, () -> @set_data(@model.source))
+
+
     if @model.render_mode == 'css'
       # dispatch CSS update immediately
       @connect(@model.change, () ->
@@ -48,6 +54,10 @@ export class LabelSetView extends TextAnnotationView
   set_data: (source) ->
     super(source)
     @visuals.warm_cache(source)
+    if @model.render_mode == 'css'
+      render()
+    else
+      @plot_view.request_render())

   _map_data: () ->
     xscale = @plot_view.frame.xscales[@model.x_range_name]
```

bryevdv: Summary: move the canvas/css test to `set_data` and call `render`/`request_render` there appropriately.

bryevdv: Alternatively can add a separate `set_data_and_render` function and keep `set_data` as it is. Could do that in response to @canavandl comment above too.

bokehjs/src/coffee/models/annotations/arrow.coffee
canavandl: Definitely another issue. Thanks for clarifying.

bokehjs/src/coffee/models/annotations/label_set.coffee
canavandl: That might be a little cleaner, but i'm not opposed to the PR as-is.

PR #6282, opened 2017-05-12
allow plot ranges to accept (start, end) tuples

### LINKED ISSUE #1112: line tutorial Range ValueError
I just worked through the lines tutorial (I like it - thanks), but encountered problems with the very end exercise of setting x_range and y_range parameters.  I first tried passing a list [-10,10](which is what is done in the solutions).  I got a ValueError:

ValueError: expected an instance of type Range, got [-10, 10] of type list

OK, so perhaps I'm stubborn, so I tried a tuple, wondering if better conversion existed for it:

ValueError: expected an instance of type Range, got (-10, 10) of type tuple

However, it looks like I really should be using bokeh.objects.Range1D objects, as pointed out in the next (scatter.py) example.

I've read in other issues (searching for duplicate issues) that this tutorial will be replaced soon.  Is this a general issue (bug) that lists and tuples _should_ work for this argument - or should the tutorial rework include a step to create Range1D objects?

This was done with Bokeh 0.5.2 - please disregard if the master branch fixes this behavior.

bryevdv:
- [x] issues: fixes #1112
- [x] tests added / passed

Just trying to pass over old tickets to close or complete if they are trivial. This should be uncontroversial.

### DIFF
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 from six import string_types
 
 from ..core.enums import Location
-from ..core.properties import Auto, Bool, Dict, Either, Enum, Include, Instance, Int, List, Override, String
+from ..core.properties import Auto, Bool, Dict, Either, Enum, Float, Include, Instance, Int, List, Override, String, Tuple
 from ..core.property_mixins import LineProps, FillProps
 from ..core.query import find
 from ..core.validation import error, warning
@@ -21,7 +21,7 @@ from .axes import Axis
 from .glyphs import Glyph
 from .grids import Grid
 from .layouts import LayoutDOM
-from .ranges import Range, FactorRange
+from .ranges import FactorRange, Range, Range1d
 from .renderers import DataRenderer, DynamicImageRenderer, GlyphRenderer, Renderer, TileRenderer
 from .sources import DataSource, ColumnDataSource
 from .tools import Tool, Toolbar, ToolEvents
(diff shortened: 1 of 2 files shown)


canavandl:
So I'm a soft -1 on this PR. It's counter to my "no magic" bokeh.models API feelings.

Additionally, ``bokeh.plotting.figure`` accepts a list of strings and generates a FactorRange, so this change doesn't fully reflect that API either. I'd rather just close the parent issue as won't fix.

bryevdv:
> Additionally, bokeh.plotting.figure accepts a list of strings and generates a FactorRange, so this change doesn't fully reflect that API either. I'd rather just close the parent issue as won't fix.

It would be easy enough to accept a list of strings and create a factor range. The main reason I am +1 on this is that it alot of examples are getting fairly import-heavy, and this would allow users to do something simple and common thing without having  to reach for yet another import. Apart from that I think being able to set a range as a tuple is just a very reasonable ask that is probably supported by almost any other library.

canavandl:
Ok - if we add support for creating FactorRanges in the same way, i'm ok with this change.

For consistency, should Title ``accept`` a string? Was there a reason why we didn't do that before? (The docstring is out of date there at least).

Am I correct in assuming there's minimal likelihood of user confusion by setting a tuple and creating a new range instead of updating a current one.

```python
p.x_range.callback = some_callback
p.x_range = (0,1) # creates a new range without the above callback
```

mattpap:
> ```
> p.x_range.callback = some_callback
> p.x_range = (0,1) # creates a new range without the above callback
> ```

This and linked plots. I'm not against this feature, but I would prefer to update existing range instead of creating a new one. I think we don't have support for this pattern in properties.

bryevdv:
OK I can agree with the notion that it would be better to update existing ranges in place, in cases where that is feasible. But I think that would require some additional machinery to pass more information to the `.accepts` converter in order to implement. I will close this PR but leave the original issue open.

PR #6296, opened 2017-05-16
Pin sphinx==1.5.1 and clean up dev_guide/documentation deps section

### LINKED ISSUE #6228: Task: add missing `packaging` dependency to documentation requirements
Task: add missing `packaging` dependency to documentation requirements in documentation at: http://bokeh.pydata.org/en/latest/docs/dev_guide/documentation.html#devguide-documentation

Also, ``packaging`` is listed twice in the meta.yaml file. We should clean that up.

canavandl:
* Pin to sphinx=1.5.1
* Remove duplicate `packaging` entry in meta.yaml
* Clean up dev_guide/documentation deps sections

- [x] issues: fixes #6228
- [ ] tests added / passed

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -25,7 +25,6 @@ requirements:
     - pycurl
     - packaging
     - six
-    - packaging
     - python-dateutil
     - numpy
     - yaml
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -49,11 +49,11 @@ test:
   requires:
     # docs
     - pygments
-    - sphinx >=1.3.1
+    - sphinx==1.5.1
+    # pyyaml installed by bokeh
+    # packaging installed by bokeh
 
     # tests
-    # pyyaml installed by bokeh
-    # jinja2 installed by bokeh
     - colorama
     - coverage
     - flake8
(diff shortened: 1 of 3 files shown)


bryevdv:
Is the CI build actually picking up bad versions of sphinx? If not I'd rather not pin. There fix for this has already been merged in Sphinx, I'd rather set a min version when when the next version of sphinx comes out.

conda.recipe/meta.yaml
```diff
@@ -50,11 +49,11 @@ test:
   requires:
     # docs
     - pygments
-    - sphinx >=1.3.1
+    - sphinx >=1.4
```
canavandl: This matches the current conf.py (and 1.3.1 doesn't work)

conda.recipe/meta.yaml
```diff
@@ -25,7 +25,6 @@ requirements:
     - pycurl
     - packaging
     - six
-    - packaging
```
bryevdv: Isn't this needed by the TravisCI docs build?

canavandl: it's a duplicate (look up two rows)

bryevdv: doh :)

PR #6310, opened 2017-05-20
Enabling/ Disabling re-ordering of columns in DataTable (freezing columns in place)

### LINKED ISSUE #6286: Enabling/ Disabling re-ordering of columns in DataTable (freezing columns in place)
Just putting in a simple request to allow the user to specify whether the columns in a DataTable can  be re-ordered or not. Currently the user can always move the columns around, but sometimes you may want to freeze say the first column or all columns in their initial positions within the table.

As Bryan pointed out, the option is currently hard coded and could easily be exposed. (https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/data_table.coffee#L158)

anthonydouc:
Addresses allowing fixed columns in datatable. https://github.com/bokeh/bokeh/issues/6286

Assuming no extra documentation is needed. 

If a test is needed I can do that too tomorrow.

* [x] issues: fixes #6286 
* [ ] tests added / passed
* [X] release document entry (if new feature or API change)

### DIFF
--- a/bokeh/models/widgets/tables.py
+++ b/bokeh/models/widgets/tables.py
@@ -353,7 +353,13 @@ class DataTable(TableWidget):
     sort direction. Use Ctrl + click to return to natural order. Use
     Shift + click to sort multiple columns simultaneously.
     """)
-
+    
+    reorderable = Bool(True, help="""
+    Allows the reordering of a tables's columns. To reorder a column, 
+    click and drag a table's header to the desired location in the table.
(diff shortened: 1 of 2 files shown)


bryevdv:
@anthonydouc Looks great! I've tested locally and everything works exactly as expected. 

For something like this, the only simple test is to to make sure the Python/JS sides have the same default values, and that happens automatically, so there's not any other test I'd suggest. 

Thanks for the contribution!

PR #6313, opened 2017-05-21
Extend CDS patching to sub-elements in columns

### LINKED ISSUE #6285: Extend patching to sub-items
Given the following CDS:
```
ColumnDataSource(data={
    "foo" : [ 1,    2,    3,    4,    5    ],
    "bar" : [ 1.7,  3.2,  5.8,  9.7,  10.4 ],
    'baz" : [ img1, img2, img3, img4, img5 ],
})
```
where `img1`, etc. are 2d NumPy arrays, it is currently possible to patch *individual* elements using the existing patch format:
```
patches = {
    'foo' : [ (0, 1) ],
    'bar' : [ (0, -2), (1, 201) ],
    'baz': [ (0, img) ],
}

source.patch(patches)
```

This suffers from two drawbacks: 

* The indexing scheme does not extend *inside* 2d images, it is only possible to patch entire images at a time. 

* Even for the 1d case, the indexing scheme is somewhat cumbersome in cases where it is desired to patch contiguous regions of a column. 

Propose to extend the patching syntax to accept slices, and to provide a simple mechanism for generating slices:
```
# patch the first three elements of a 1d column
p1 = patcher[0:3]
patches = { 'foo': (p1, [1, 2, 3] } 

# patch the entire second column of a the 10th image
p2 = patcher[10, :, 2]
patches = { 'baz' : (p2, [...] }
```
Note: use of an object like `patcher` with a custom `__getitem__` seems necessary, as slice syntax is only available inside indexing brackets. 

Would serialize slices `slice(1, 10, None)` as `(1, 10, null)`. Bounds and consistency checking could occur on the python side, but handling of the slice would occur in BokehJS in the CDS `patch` method:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/sources/column_data_source.coffee#L96-L101

bryevdv:
This PR extends the CDS patch capability to more kinds of patch values. In particular, it can now patch sub-portions of "array" element. This is most useful for, e.g. updating data for image, patches, and multi-line glyphs.

Patches are still of the form `(index, new_data)` but now the index can be a slice, or a list of integers and slices (with correspondingly sized new data) The kinds of patches now accepted are summarized below:
```
# patch entire elements of "typical" columns
(10, 100)
(slice(2), [100, 101])

# column is "list of 1d array"
([1, 4], [6])
([0, slice(2, 4)], [100, 101])

# column is "list of 2d array"
([1, 4, slice(None)], [1, 2, 3, 4, 5])
([0, slice(2, 4), 2], [100, 101])
```

- [x] issues: fixes #6285
- [x] tests added / passed

### DIFF
--- a/bokeh/core/property/containers.py
+++ b/bokeh/core/property/containers.py
@@ -363,15 +363,20 @@ class PropertyValueDict(PropertyValueContainer, dict):
         synchronize.
 
         .. warning::
-            This function assumes the integrity of ``new_data`` has already
+            This function assumes the integrity of ``patches`` has already
             been verified.
 
         '''
+        import numpy as np
         old = self._saved_copy()
 
         for name, patch in patches.items():
             for ind, value in patch:
-                self[name][ind] = value
+                if isinstance(ind, (int, slice)):
+                    self[name][ind] = value
+                else:
+                    shape = self[name][ind[0]][ind[1:]].shape
+                    self[name][ind[0]][ind[1:]] = np.array(value, copy=False).reshape(shape)
 
         from ...server.events import ColumnsPatchedEvent
 
--- a/bokehjs/src/coffee/models/sources/column_data_source.coffee
+++ b/bokehjs/src/coffee/models/sources/column_data_source.coffee
@@ -3,7 +3,7 @@ import {HasProps} from "core/has_props"
 import * as p from "core/properties"
 import {Set} from "core/util/data_structures"
 import * as serialization from "core/util/serialization"
-import {isArray, isObject} from "core/util/types"
+import {isArray, isNumber, isObject} from "core/util/types"
 
 # exported for testing
 export concat_typed_arrays = (a, b) ->
@@ -65,7 +65,7 @@ export patch_to_column = (col, patch, shapes) ->
     # make the single index case look like the length-3 multi-index case
     if not isArray(ind)
 
-      if Number.isInteger(ind)
+      if isNumber(ind)
         value = [value]
         patched.push(ind)
       else
--- a/bokehjs/src/coffee/models/sources/column_data_source.coffee
+++ b/bokehjs/src/coffee/models/sources/column_data_source.coffee
@@ -1,6 +1,7 @@
 import {ColumnarDataSource} from "./columnar_data_source"
 import {HasProps} from "core/has_props"
 import * as p from "core/properties"
+import {Set} from "core/util/data_structures"
 import * as serialization from "core/util/serialization"
 import {isArray, isObject} from "core/util/types"
 
@@ -56,17 +57,26 @@ export slice = (ind, length) ->
 
 # exported for testing
 export patch_to_column = (col, patch, shapes) ->
+  patched = new Set()
+  patched_range = false
+
   for [ind, value] in patch
 
     # make the single index case look like the length-3 multi-index case
     if not isArray(ind)
+
       if Number.isInteger(ind)
         value = [value]
+        patched.push(ind)
+      else
+        patched_range = true
+
       ind = [0, 0, ind]
       shape = [1, col.length]
       item = col
 
     else
+      patched.push(ind[0])
       shape = shapes[ind[0]]
       item = col[ind[0]]
 
@@ -84,8 +94,12 @@ export patch_to_column = (col, patch, shapes) ->
 
     for i in [istart...istop] by istep
       for j in [jstart...jstop] by jstep
+        if patched_range
+          patched.push(j)
         item[i*shape[1] + j] = value[flat_index]
         flat_index++
+
+  return patched
 
 # Data source where the data is defined column-wise, i.e. each key in the
 # the data attribute is a column name, and its value is an array of scalars.
@@ -127,7 +141,8 @@ export class ColumnDataSource extends ColumnarDataSource
 
   patch: (patches) ->
     data = @data
+    patched = new Set()
     for k, patch of patches
-      patch_to_column(data[k], patch, @_shapes[k])
+      patched = patched.union(patch_to_column(data[k], patch, @_shapes[k]))
     @setv('data', data, {silent: true})
-    @patching.emit()
+    @patching.emit(patched.values)
--- a/bokehjs/src/coffee/models/sources/column_data_source.coffee
+++ b/bokehjs/src/coffee/models/sources/column_data_source.coffee
@@ -2,7 +2,7 @@ import {ColumnarDataSource} from "./columnar_data_source"
 import {HasProps} from "core/has_props"
 import * as p from "core/properties"
 import * as serialization from "core/util/serialization"
-import {isObject} from "core/util/types"
+import {isArray, isObject} from "core/util/types"
 
 # exported for testing
 export concat_typed_arrays = (a, b) ->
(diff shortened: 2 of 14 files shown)


bryevdv:
@mattpap when the `patching` signal is emitted, I'd like to communicate the specific indices that got patched. What can I pass to the `emit` call, and how do I access that data when the signal is handled?

bryevdv:
Here is an example that can be used for testing
```
from bokeh.io import curdoc
from random import randint

import numpy as np

N = 200
for i in range(N):
    for j in range(N):
        view[i, j, 0] = int(i/N*255)
        view[i, j, 1] = int(j/N*255)
        view[i, j, 2] = 158
        view[i, j, 3] = 255

source = ColumnDataSource(data=dict(img=[img]))

p.image_rgba(image='img', x=0, y=0, dw=10, dh=10, source=source)

def update():

    ind = [0, slice(70, 91, 5), slice(None)]
    new = np.roll(source.data['img'][0][70:91:5, :], 2, axis=1).flatten()
    patches = { 'img' : [(ind, new)] }

    source.patch(patches)

curdoc().add_root(p)

curdoc().add_periodic_callback(update, 50)
```

There are a few more python tests that could be added, and the docs could be expanded a bit, but otherwise this should be close. The only additional feature I'd like to possibly add is the optimization described above.

bryevdv:
screenflow-4

mattpap:
@bryevdv, regarding `patching` signal. You can emit one argument (this can be a tuple, so effectively you can emit as many arguments you want). Data is received as the first argument to a slot (there is one more which is a reference to the sender). So, for a list of indices you would do:
```
patching = Signal(this, "patching") # <number[], ColumnarDataSource>
```
```
obj.patching.connect((indices) => console.log(indices))
```
```
obj.patching.emit([1, 2, 3])
```

Make sure that type information is left in comments, so that I don't have to reverse engineer this in a few weeks or months.

bryevdv:
@mattpap I may add a little to docs and possibly an example, but the core implementation is done, please review as soon as you can, would like to include in a dev build for DS devs to start integrating with.

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -49,13 +50,58 @@ export stream_to_column = (col, new_col, rollover) ->
...
+  patched = new Set()
+  patched_range = false
+
+  for [ind, value] in patch
+
+    # make the single index case look like the length-3 multi-index case
+    if not isArray(ind)
 
-# Datasource where the data is defined column-wise, i.e. each key in the
+      if Number.isInteger(ind)
```
mattpap: Not supported on IE. Also, does this have to be an integer check? If you are just differentiating between a number and something else you should use `isNumber` from `core/util/array`.

bryevdv: `core/util/types`?

bryevdv: fixed 

FYI it really should be an integer, as it is an array index. But I was mostly just looking for way to determine "not array and not object" and the other things I tried failed.

mattpap: > `core/util/types`?

Maybe I'm just a bit tired right now, because I clearly though `types` and wrote `arrays`.

bokeh/core/property/containers.py
```diff
@@ -363,15 +363,20 @@ def _patch(self, doc, source, patches, setter=None):
...
         old = self._saved_copy()
 
         for name, patch in patches.items():
             for ind, value in patch:
-                self[name][ind] = value
+                if isinstance(ind, (int, slice)):
+                    self[name][ind] = value
+                else:
+                    shape = self[name][ind[0]][ind[1:]].shape
+                    self[name][ind[0]][ind[1:]] = np.array(value, copy=False).reshape(shape)
```
bryevdv: An upshot of this is that sub-patching into "columns of 1d" or "columns of 2d" can only wrk when the individual column elements are all arrays. Will add a note to the docs about this.

bokehjs/src/coffee/models/sources/column_data_source.coffee
mattpap: > it really should be an integer, as it is an array index

Is there a distinction between and integers and floats in this code? If not (hopefully), then you shouldn't be validating here, just using minimal logic to distinguish javascript types. As to distinguishing non objects, this is tricky and you should always check for as specific types as possible, because otherwise:
```
> let t = require("./types")
undefined
> t.isObject(1)
false
> t.isObject(new Number(1))
true
```
Though I'm not sure if any code in bokehjs or any libraries we use produce this kind of objects.

bryevdv: there's not a distinction between ints and floats, I have changed to `isNumber` and all the unit tests pass so it seems fine.

bokehjs/src/coffee/models/sources/column_data_source.coffee
```diff
@@ -49,13 +50,58 @@ export stream_to_column = (col, new_col, rollover) ->
...
-  for i in [0...patch.length]
-    [ind, value] = patch[i]
-    col[ind] = value
+export slice = (ind, length) ->
+  if isObject(ind)
+    return [ind.start ? 0, ind.stop ? length, ind.step ? 1]
+  return [start, stop, step] = [ind, ind+1, 1]
 
+# exported for testing
+export patch_to_column = (col, patch, shapes) ->
```
mattpap: @bryevdv, I will need a signature for this function. My rough approximation is this:
```
(col: any[],
 patch: [number | [number, number] | [number, number, number], any][],
 shape: [number, number][]) => Set<number>
```
btw. `[type_0, ..., type_n]` syntax stands for a tuple (on the type level, because this isn't (easily) distinguishable at runtime).

mattpap: Obviously we can make this a bit more readable with type aliases, e.g.:
```
type Slice = [number, number] | [number, number, number]
type Shape = [number, number]
(col: any[],
 patch: [number | Slice, any][],
 shape: Shape[]) => Set<number>
```

bryevdv: The slices get turned into `[number, number, number]` inside this function. In the patch that is passed in they are objects: `{start: 0, stop: 10, step: 1}`

But I don't think what you have is quite correct. Here is it put another way: 

* A `patch` is a list of "tuples", where each tuple is `(index, new_data)`
* `new_data` is probably `any` (or `any[]`?)
* However `index` can either be:

- a single number
- a single slice object (see above)
- a length 2 or length three list, where the first element is always a number, and the subsequent elements are either numbers or slice objects


Shapes can also be different lengths. The length is dependent on the corresponding item in the CDS column. I.e. if its "lists of 1d" like for multiline, then individual shapes are just `[number]` but if its "lists of 2d" like images then individual shapes are `[number, number]`

bryevdv: E.g. in this example:
```
 +            # replace the entire 10th column of the 2nd array:
 +
 +              +------------- index of item in column data source
 +              |
 +              |     +------- row subindex into array item
 +              |     |
 +              |     |     +- column subindex into array item
 +              V     V     V
 +            ([2, slice(None), 10], new_values)
```
the `index` on the JS side is `[2, {start:null, stop:null, step:null}, 10]`

bryevdv: Other valid index values:
```
# replace item in "normal" CDS col at index 2
2 

# replace a range of values in a "normal" CDS column
{start:0: stop:10, step: 2} 

# replace an item in a "list of 1d" CDS column, at index 5, subindex 10
[5, 10] 

# replace a range of values in a "list of 1d" CDS column, at index 5
[5, {start:0: stop:10, step: 2}] 

# replace a range of values in a "list of 2d" CDS column, at index 5 in the 10th column of the subitem
[5, {start:0: stop:10, step: 2}, 10] 

```

bryevdv:
Added a howto example that shows all three kinds of patching:

screen shot 2017-05-22 at 15 39 05

@mattpap any objections to merging?

mattpap:
@bryevdv, looks good, though I didn't have time to go through all details.

Issue #6316, opened 2017-05-22
Bug in ScreenDistanceSpec serialization


philippjfr:
Unfortunately I haven't been able to create a small reproducible example for this issue but I think I've tracked down why it occurs. The specific error I'm referring to is this:

>ValueError: expected an element of either String, Dict(Enum('field', 'value', 'transform'), Either(String, Instance(Transform), Float)) or Float, got {'units': 'screen', 'field': 'size'}
 
The traceback mentions this line:

```python
def to_serializable(self, obj, name, val):
    d = super(ScreenDistanceSpec, self).to_serializable(obj, name, val)
    d["units"] = "screen" <----
    return d
```

The issue appears to be that in descriptors.py the container object is passed through to be serialized:

```python
self.property.to_serializable(obj, self.name, getattr(obj, self.name))
```

And the ``DataSpec.to_serializable`` just passes the container object through without making a copy, which means ScreenDistanceSpec tries to directly modify the container triggering the ValueError. The error occurs in 0.12.6dev{4, 5} but not 0.12.5. I'll continue trying to make a minimal example as currently I can only trigger it with a holoviews gridplot of scatter points.

mattpap:
>  but not 0.12.5

btw. Last commit (57068d6) that touches this code was merged in 0.12.4.

philippjfr:
Thanks, I'll do a bisection to figure out the actual source of the issue then.

bryevdv:
I think having the superclass always  return a new dict is reasonable. There is one branch where it is not doing that:

https://github.com/bokeh/bokeh/blob/master/bokeh/core/properties.py#L1489

There also seems to be a branch that could return `None` (at the top) not sure if that's actually not a real concern in practice or if it should be guarded against in subclasses.

Can you make a PR for this and test it out?

PR #6318, opened 2017-05-23
Ensure DataSpec.to_serializable returns new dict

philippjfr:
Fixes issue #6316 associated with ``ScreenDistanceSpec`` and ``DataDistanceSpec`` attempting to set ``units`` directly on a container, triggering a validation error. I'm still looking for a small reproducible example that could be tested but this seems to be the safer approach regardless.

### DIFF
--- a/bokeh/core/tests/test_properties.py
+++ b/bokeh/core/tests/test_properties.py
@@ -1766,6 +1766,19 @@ def test_strict_dataspec_key_values():
         with pytest.raises(ValueError):
             f.x = dict(field="foo", units="junk")
 
+def test_dataspec_dict_to_serializable():
+    for typ in (NumberSpec, StringSpec, FontSizeSpec, ColorSpec, DataDistanceSpec, ScreenDistanceSpec):
+        class Foo(HasProps):
+            x = typ("x")
+        foo = Foo(x=dict(field='foo'))
+        props = foo.properties_with_values(include_defaults=False)
+        if typ is DataDistanceSpec:
+            assert props['x']['units'] == 'data'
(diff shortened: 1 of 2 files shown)


bryevdv:
This definitely seems like a correct thing to do. Do you want to wait and try to isolate a test case, or should I merge this now? Actually I guess it would be possible to add a test case that maintains the dict is a *new* dict in any case, would you mind adding that?

philippjfr:
@bryevdv Added a unit tests which helped me catch another bug ``DataDistanceSpec`` was referencing ``ScreenDistanceSpec`` in the call to super. Not sure why these cases didn't get caught before or cause more errors.

bokeh/core/tests/test_properties.py
```diff
@@ -1766,6 +1766,18 @@ class Foo(HasProps):
...
+        class Foo(HasProps):
+            x = typ("x")
+        foo = Foo(x=dict(field='foo'))
+        props = foo.properties_with_values(include_defaults=False)
+        if typ is DataDistanceSpec:
+            assert props['x']['units'] == 'data'
+        elif typ is ScreenDistanceSpec:
+            assert props['x']['units'] == 'screen'
+        assert props['x']['field'] == 'foo'
+
```
bryevdv: could add `assert props['x'] is not foo.x` to maintain that serialized dict is a copy?

philippjfr: Yes, good point, that was the whole point of the change after all. I was only implicitly testing that because it would otherwise raise the ValueError.

PR #6321, opened 2017-05-23
Exposed initial_start and initial_end in models/Range1d. This allows …

### LINKED ISSUE #6320: Expose initial_start and initial_end in Range1d
In some cases you might not know the range of data when a bokeh application initially loads. In this case the `Range1d` is set to a default of `[0,1]` for the figure you intend to plot in. After plotting the data you can of course change `figure.x_range.start` and `figure.x_range.end` to have the graph contain your newly plotted data. But, if you have the `ResetTool` available and use it, the range will reset to `[0,1]`. If the properties `_initial_start` and `_initial_end` are exposed you can change the 'default' range so reset works properly.

bgyarfas:
…the user to change the 'default' range after a graph has been created and before the user knows the range of the data that is going to be plotted

- [x] issues: fixes #6320 
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/ranges.py
+++ b/bokeh/models/ranges.py
@@ -75,6 +75,14 @@ class Range1d(Range):
     maximum visible interval. Can be a timedelta. Note that ``bounds`` can
     impose an implicit constraint on the maximum interval as well. """)
 
+    initial_start = Either(Float, Datetime, Int, default=0, help="""
+    The default start of the range. (When user hits reset)
+    """)
+
+    initial_end = Either(Float, Datetime, Int, default=1, help="""
+    The default end of the range. (When user hits reset)
+    """)
+
     def __init__(self, *args, **kwargs):
         if args and ('start' in kwargs or 'end' in kwargs):
             raise ValueError("'start' and 'end' keywords cannot be used with positional arguments")
--- a/bokehjs/src/coffee/models/ranges/range1d.coffee
+++ b/bokehjs/src/coffee/models/ranges/range1d.coffee
@@ -10,12 +10,14 @@ export class Range1d extends Range
       bounds: [ p.Any       ] # TODO (bev)
       min_interval: [ p.Any ]
       max_interval: [ p.Any ]
+      initial_start: [p.Number, 0]
+      initial_end: [p.Number, 1]
     }
 
   _set_auto_bounds: () ->
     if @bounds == 'auto'
-      min = Math.min(@_initial_start, @_initial_end)
-      max = Math.max(@_initial_start, @_initial_end)
+      min = Math.min(@initial_start, @initial_end)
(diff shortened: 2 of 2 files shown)


bryevdv:
My concern with exposing these properties on ranges is that they can then be set before the doc is rendered — and then will be overwritten and ignored when they are computed in the client. I think that would be unexpected an possibly a source of  extra support issues. 

Think there are two possible avenues, not far off:

* if the intent is for the reset tool to always automatically just use computed bounds but not otherwise be configurable, then there is no need to make these properties public. The tool could look for `_initial_start`, or better a read-only getter method `get_computed_initial_start` that could be added.

* if the goal is for the bounds that the reset tool uses to be *configurable* then I think new properties (that are *not* computed in the client) are needed. 

It both things are needed then you'd probably want new properties, and to have some precedence order for the reset tool to check things, probably: use new prop values if they've explicitly been set by the user, other wise use the computed initial start/end

bgyarfas:
Doesn't `DataRange1d` use auto-computed bounds? If so, then I think the first bullet is covered. This was intended to be user configurable so I think I'll need to work on the second bullet. To give you a use case of why... I have 3 different plots where the data doesn't get plotted until after the doc is rendered. I have them sharing an x_range but for the y_range I compute on the server side some bounds that takes into consideration the data going to all 3 plots This range is then applied to all of the plots, so in this case any sort of auto computed bounds wouldn't work.

bryevdv:
Yes `DataRange1d` auto-computes bounds, but I think that is orthogonal to my point. I was talking about this situation:
```
plot = figure()

#  data range will reset this in the client, it will have no effect. That's confusing. 
plot.x_range.initial_start = -100

curdoc.add_root(plot)
```
Although looking at the PR it doesn't seem it updates the usage of `_initial_start` in `DataRange1d` so while the above may not be true I think changing `_initial_start` *only* on `Range1d` but not on its subclasses is likely to have other issues.

I started writing up a proposal but looking further I was reminded of what a gorpy corner of BokehJS this is. I will have to allot some dedicated time to think about this / look into it more. But if you have another implementation to consider, please feel free to update this PR in the mean time. I think adding separate `reset_start` and `reset_end` is probably the way to go in any case.

bryevdv:
Complications that will arise in any of these cases: if these "reset overrides" are settable, then its possible for users to set values that are inconsistent with and contradict the `bounds` setting. That has to be handled  in a predictable, reasonable  way somehow. 

I'm starting to think this might not be a trivial change to do right, and wonder if either a custom extension range or reset tool tailored to your specific use case might not be easier.

bryevdv:
@bgyarfas I am going to go ahead and close this PR. I think it will take some more thought. I have bumped the issue to `0.12.8` though, to try and have a more concrete deadline

PR #6323, opened 2017-05-24
add bkcharts shim and dependency

bryevdv:
Add compat shim for `bokeh.charts` and dependency on `bkcharts` (both to be removed at `1.0`)


bryevdv:
Merging into master that has broken example fixed.

PR #6324, opened 2017-05-24
Add curdoc().add_root() to bokeh command example

### LINKED ISSUE #5634: Fix documentation of 'bokeh html foo.py'
The [user guide](http://bokeh.pydata.org/en/latest/docs/user_guide/concepts.html#output-methods) shows an example of `bokeh html foo.py`, however the `foo.py` seems to be missing two lines.

The full code should be:
```python
from bokeh.io import curdoc    #This is missing from userguide

p = figure() 
p.line(x=[1, 2, 3], y=[4,6,2]) 

curdoc().add_root(p)    #This is missing from userguide
```

tdsmith:
Otherwise, the resulting document is empty.

Closes #5634.

### DIFF
--- a/sphinx/source/docs/user_guide/concepts.rst
+++ b/sphinx/source/docs/user_guide/concepts.rst
@@ -120,10 +120,11 @@ above example would be simplified to:
 
 .. code-block:: python
 
-    from bokeh.plotting import figure
+    from bokeh.plotting import figure, curdoc
 
     p = figure()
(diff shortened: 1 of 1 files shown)


bryevdv:
FYI a downstream URL used by one of our tested examples is having SSL issues, it's going to cause the tests to fail for now.

bryevdv:
Merging into `master` which now has the broken example disabled. Thanks @tdsmith !

PR #6327, opened 2017-05-25
Adds the renderer object to cb_data

### LINKED ISSUE #6326: Passing renderer object in cb_data on HoverTool CustomJS callback
Previously there was no way to determine which renderer invoked a callback within JS. By passing the renderer in cb_data it will then be available in the callback.

bgyarfas:
To be able to identify which renderer invoked the callback from within the JS

- [x] issues: fixes #6326 
- [ ] tests added / passed


bryevdv:
Thanks @bgyarfas! will merge now, changes will be included in a dev build out later today

PR #6334, opened 2017-05-29
add script to dump complete JSON specification for all Bokeh models

### LINKED ISSUE #5950: Expose all JSON representations in one place
In order to make life much simpler for other language bindings, publish a single plain-text JSON file that has default representation of every model in one place. 

Additionally, it would be good to generate another file in some format that specifies the type for each model property as well. 

cc. @hafen

bryevdv:
issues: closes #5950

### DIFF
--- a/scripts/spec.py
+++ b/scripts/spec.py
@@ -1,0 +1,52 @@
+from datetime import date
+import json
+
+import bokeh.models as bm; bm
+import bokeh.models.widgets as bmw; bmw
+
+from bokeh.model import Model
+
+def _proto(obj, defaults=False):
+    return json.dumps(obj.to_json(defaults), sort_keys=True, indent=None)
+
+data = {}
+for name, m in sorted(Model.__class__.model_class_reverse_map.items()):
+    item = {
+        'name'  : name,
+        'bases' : [str(base)[8:-2] for base in m.__bases__],
+        'desc'  : m.__doc__.strip() if m.__doc__ is not None else '',
+        'proto' : _proto(m(), True),
+    }
+    props = []
+    for prop_name in m.properties():
+        descriptor = m.lookup(prop_name)
+        prop = descriptor.property
+
+        detail = {
+            'name'    : prop_name,
+            'type'    : str(prop),
+            'desc'    : prop.__doc__.strip() if prop.__doc__ is not None else '',
+        }
+
+        default = descriptor.instance_default(m())
+        if isinstance(default, date):
+            default = str(default)
+
+        if isinstance(default, Model):
+            default = _proto(default)
+
+        if isinstance(default, list) and any(isinstance(x, Model) for x in default):
+            default = [_proto(x) for x in default]
+
+        if isinstance(default, dict) and any(isinstance(x, Model) for x in default.values()):
+            default = { k: _proto(v) for k, v in default.items() }
+
+        detail['default'] = default
+
+        props.append(detail)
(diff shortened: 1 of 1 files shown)


bryevdv:
ping @hafen

New gist of output here:

https://gist.githubusercontent.com/bryevdv/118ad623883e607ad08563472414eb18/raw/756c7dee83541208c3935845c61a31de06093f5e/gistfile1.txt

bryevdv:
Instance defaults are handled by providing a "without-defaults" prototype, e.g:

```
{
        "name": "overlay",
        "type": "Instance(BoxAnnotation)",
        "desc": "A shaded annotation drawn to indicate the selection region.",
        "default": "{\"bottom_units\": \"screen\", \"fill_alpha\": {\"value\": 0.5}, \"fill_color\": {\"value\": \"lightgrey\"}, \"id\": \"5fbd1c1a-a219-4cf1-af45-fe9480f4f73e\", \"left_units\": \"screen\", \"level\": \"overlay\", \"line_alpha\": {\"value\": 1.0}, \"line_color\": {\"value\": \"black\"}, \"line_dash\": [4, 4], \"line_width\": {\"value\": 2}, \"render_mode\": \"css\", \"right_units\": \"screen\", \"top_units\": \"screen\"}"
      },
```

scripts/spec.py
```diff
@@ -0,0 +1,52 @@
...
+from bokeh.model import Model
+
+def _proto(obj, defaults=False):
+    return json.dumps(obj.to_json(defaults), sort_keys=True, indent=None)
+
+data = {}
+for name, m in sorted(Model.__class__.model_class_reverse_map.items()):
+    item = {
+        'name'  : name,
+        'bases' : [str(base)[8:-2] for base in m.__bases__],
```
mattpap: `base.__module__ + '.' + base.__name__` or is there a reason to use code that I needed to run to know what it does?

bryevdv: Nope, just an oversight, thanks, been fixed

scripts/spec.py
```diff
@@ -0,0 +1,52 @@
+from datetime import date
+import json
+
+import bokeh.models as bm; bm
+import bokeh.models.widgets as bmw; bmw
```
mattpap: BMW wouldn't be too happy about such comparison, quality-wise.

mattpap:
Not much to review here, just I'm curious what's the relationship of this code with `bokehjs/gulp/tasks/generate_defaults.py`?

scripts/spec.py
bryevdv: @mattpap there's plenty to aspire to in terms of quality. Until we have hundreds of QA people like BMW certainly has, I'm sure that will always be the case. In the mean time we've managed to create a tool that is evidently useful to a fairly large number of people, with only a handful of full-time contributors.

bryevdv:
> Not much to review here, just I'm curious what's the relationship of this code with bokehjs/gulp/tasks/generate_defaults.py?

This script is more general, in that it needs to report type and docs information. It's probably the case that `generate_defaults.py` could be replaced by this script at some point.

bryevdv:
@hafen I will go ahead and merge this today. If there is additional work or features needed just let mw know!

scripts/spec.py
mattpap: @bryevdv, I was just joking about widgets, not this PR.

PR #5774, opened 2017-01-20
Unify all types of sliders with nouislider

mattpap:
- [ ] restore title attribute

fixes #5502
fixes #2527

PR #6335, opened 2017-05-29
Proof of concept removal of jquery and jquery-ui

### LINKED ISSUE #2527: Rendered slider* plots have missing bk-* classes

### LINKED ISSUE #5502: Consider nouislider to replace all our slider widgets
Available at https://refreshless.com/nouislider/. It's small, feature full and under liberal license. Doesn't use jquery-ui, which I would like to get rid of as much as possible.

### LINKED ISSUE #5506: Fix DatePicker widget
Based on discussion in PR #5451:

>> (...) but its styling is badly broken currently

> It think it's not only currently, but it has been for 1.5+ years (since introduction of CSS class prefixes). If > no one complained for that amount of time and given that we never had a single example using it, then "a use" seems more like "no use".

I think we should consider removing it. Just note that I introduced this widget, I think just for the sake of having a widget for whatever jquery-ui supports.

### LINKED ISSUE #6308: Sliders :: CustomJS Can Not Dynamically Change Title
@mattpap 

Sorry to keep bothering about sliders! Not sure if this is a bug or just simply not implemented.
I can dynamically update the callback object's start, end, & value attributes as I please in JS and I see the updates visually. However, the **title** attribute doesn't seem to want to update visually. The actual object value does change; I have confirmed by using console.log()

    reset_output()
    output_file("slider_callback.html")
    callback_slider = CustomJS(code = '''
        console.log(cb_obj.title);
        if (cb_obj.value == 0) {
           cb_obj.title = "foo";
        }
        else {
            cb_obj.title = "bar";
         }  
       cb_obj.trigger('change');
    ''')
    slider = Slider(start = 0, end = 1, value = 0, step = 1, title = "YOU SHALL NOT CHANGE ME!")
    layout = column(slider)
    show(layout)

### LINKED ISSUE #6332: Remove CSS 'bk-' prefixing
While working on issue #4503 I agreed that attempting to remove the "bk-" CSS prefix would help because of how jquery-ui widgets generate elements and how _prefix_ui in bokeh was being implemented and used.

There's worry this may cause breakages for people using the bk- prefixed stylesheets/code and there will be the need for a deprecation release.

### LINKED ISSUE #6474: New panels to a Tabs widget does appear
https://gist.github.com/nicain/5a0236b688c166cb195d61e7735bcdcd

Given the behavior of modifying the "children" attribute of a layout, I expected adding a new tab to work in a similar fashion.  In the gist above, I think that pressing the button should add a new tab (and new figure), but it doesn't seem to work.

bokeh 0.12.6
firefox 54 and Chrome 58

### LINKED ISSUE #6490: WebGL was requested and is supported, but bokeh-gl(.min).js is not available, falling back to 2D rendering.
I can't use `webgl` mode in Jupyter notebook, and the here is the warning:

```
WebGL was requested and is supported, but bokeh-gl(.min).js is not available, falling back to 2D rendering.
```

the source code:

```
output_notebook()

fig = figure(output_backend="webgl")

circle_source = ColumnDataSource(data={"x":[0], "y":[0], "r":[1], "color":["#000000"]})
fig.circle(x="x", y="y", radius="r", fill_color="color", line_width=0, line_color=None, source=circle_source)
show(fig)
```

### LINKED ISSUE #6530: Split bokeh-widgets into bokeh-widgets and bokeh-tables
This way jquery will be included only were needed and size of main widget bundle will go below 100 kB minified (as in PR #6335).

mattpap:
- [x] remove `bk-ui-` prefixing
- [x] rewrite widgets using DOM APIs
- [x] drop backbone's javascript
- [x] remove dependency on jquery{-ui}
- [x] make sure all widgets sill work
- [x] reimplement autocomplete widget
- [x] import slick grid from fork (don't use vendor)
- [x] upgrade slick grid to newest version
- [x] fix slick grid's cell editors
- [x] use pikaday in date picker widget
- [x] merge PR #5774 into this
- [x] restore title attribute in sliders
- [x] make sliders' style less bulky
- [x] make slickgrid respect strict mode
- [x] include jquery for data table widget

fixes #6332 
fixes #6474 
fixes #6490
fixes #5502 
fixes #2527 
fixes #5506
fixes #6308
fixes #6530


mattpap:
@bryevdv, do you have any preference what CDN should be used for jquery-ui? E.g. https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js?

mattpap:
I'm marking this as ready from functional perspective. Unfortunately, there is an unrelated deficiency in how external resources are handled which needs to be resolved to make this PR complete. Currently external resource are included based on models, not by objects (models actually in use).

mattpap:
Note that there will be image differences, because internal and external (jquery-ui) CSS changed.

bokeh/models/widgets/widget.py
```diff
@@ -17,3 +17,15 @@ class Widget(LayoutDOM):
...
+    """ A mixin for widgets requiring JQuery and/or JQuery UI. """
+
+    __javascript__ = [
+        "https://code.jquery.com/jquery-3.2.1.min.js",
+        "https://code.jquery.com/ui/1.12.1/jquery-ui.min.js",
+    ]
+
+    __css__ = [
+        "https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css",
+    ]
```
bryevdv: A comment and a question:

* Not fond of the name `NeedsJQuery` but that's a small thing

* Much more important: how does this work with inline resources? A very important use case that must be considered is people working on isolated networks who will not have access to the wider internet.

mattpap: `NeedsJQuery` is defined in `bokeh/models/widgets/widget.py`.

mattpap: > : A very important use case that must be considered is people working on isolated networks (...)

I didn't consider this yet, as this touch a much broader topic that wasn't addressed when external resources feature was added. Another situation to consider is people not wanting this being included automatically at all. I'm going to address those in future PRs in this release cycle.

mattpap: Unfortunately, just downloading and including jquery in `js_raw` won't work, because this wound't include any images or fonts, etc.

bryevdv: Yah I am not certain about the viability of this approach. I think perhaps we have to either:
* re-implement widgets without jq/jq-ui *at all*
* cleave `bokeh-widgets` into two pieces `bokeh-widgets` and `bokeh-jqwidgets` where the latter has all the jq stuff bundled inside as now 
* building/bundling our own `bokeh-jqui` that we publish can be included when necessary.

bryevdv: To elaborate for extensions: I don't think its the case that extensions relying on external resources necessarily work in the situations I described. we've never made that claim, and presumably anyone in that situation would understand that it would not work on an isolated network. But the *built-in* widgets, do absolutely always need to work, inline or on isolated networks.

mattpap: > re-implement widgets without jq/jq-ui at all

I would prefer that. In fact, most important widgets are already jquery independent. Most notably sliders and data table depend on jquery(ui) in this PR. Sliders are already reimplemented in PR #5774. This leaves us with data table. Additionally there is date picker and auto complete input, where the former needs some work to make it jquery independent (there's an issue open for that) and the later is pretty trivial to reimplement.

mattpap: > bokeh-jqwidgets where the latter has all the jq stuff bundled inside as now

That would mean going back to prefixes. You can't have two conflicting versions of jquery ui. This isn't even going to work in the notebook. I mean it would work to some extent, but there would be CSS clashes, and this is what we want to avoid.

mattpap: > building/bundling our own bokeh-jqui that we publish can be included when necessary.

This looks like the only solution if we want to keep jquery widgets. Then we can package everything in one bundle including images and fonts, and include the bundle only when requested.

bryevdv: AFAIK date picker is hopelessly broken and has been for a long time. DataTable is very important but probably also the largest widget implementation. How about cleaving it off and bundling JQ-ui with just it? then we can Fix autocomplete and date picker to not use JQ at all.

bryevdv: At some point e.g we may want to switch to the phosphor table, so cleaving off data table now actually seems like a good prelude for that later.

mattpap: > AFAIK date picker is hopelessly broken

Date picker works in this PR. It's just a matter of not using prefixes with jquery ui.

bryevdv: OK, I am fine with `bokeh-jqui` as well.

bryevdv: Can you give an estimate to go the `bokeh-jqui` route? To make sure I understand, there would be some notebook issues with this route? Or we'd keep prefixes for the few things that need them? If this is going to take more than a day or so we'd need to discuss prioritization.

mattpap: > To make sure I understand, there would be some notebook issues with this route?

Notebook provides its own version of jquery(ui), so we can expect clashes between both versions. This is similar problem we have with their global `define()`, which interacts with our external resources written in UMD style. We should expect either bokehjs or notebook not working (which is something we went already before).

> Or we'd keep prefixes for the few things that need them?

I don't want to keep prefixes. The whole original point of this PR to get rid of this maintenance nightmare.

mattpap: At this point I think that the only viable option is to drop jquery-ui completely. I already reimplemented autocomplete (naively, but can be improved later). I don't care about date picker, so it can be either removed or rewritten (if someone want to devote time to this). This leaves us with SlickGrid. I see that it uses jquery-ui only for sortable column and some trivia (icons), so I should be able to fix this in a day or two.

bryevdv: OK, certainly getting rid of JQ altogether has many good benefits. But we need to have it happen early in the cycle so the changes can have some exposure in dev builds. Please do the following:
* make separate issues for each of `autocomplete`, `datepicker`, and `datatable` 
* pride (at least) a daily status update

mattpap: So I actually don't have to fix slick grid right now, because it works perfectly fine without jquery-ui, as long as `DataTable.reorderable == false` (which I can set appropriately when jquery-ui isn't present). To wrap up this PR I need to fix cell editors which are pretty broken after upgrading slick grid.

bryevdv:
Will probably need a migration note about removal of `bk-` prefix

bokehjs/src/coffee/models/widgets/data_table.coffee
```diff
@@ -149,9 +149,15 @@ export class DataTableView extends WidgetView
     if @model.row_headers
       columns.unshift(@newIndexColumn())
 
+    reorderable = @model.reorderable
+
+    if reorderable and not $?.fn?.sortable?
+      logger.warn("jquery-ui is required to enable DataTable.reorderable")
+      reorderable = false
+
```
bryevdv: What's the longer term plan then? sorting is one of the more important features of tables

bokehjs/src/coffee/models/widgets/date_picker.coffee
```diff
@@ -12,16 +12,16 @@ export class DatePickerView extends InputWidgetView
...
       maxDate: if @model.max_date? then new Date(@model.max_date) else null
       onSelect: @onSelect
     })
-    @$el.append([@label, @input])
-    @_prefix_ui()
+    @el.appendChild(@label)
+    @el.appendChild(@input)
     return @
 
   onSelect: (dateText, ui) =>
```
bryevdv: We should wrap this non-jq datepicker: https://github.com/dbushell/Pikaday

bokehjs/src/coffee/models/widgets/data_table.coffee
mattpap: This is about sorting rows, but reordering columns (aka. drag & drop). As said before, I will change slickgrid to not use ui for this, but as there is a quick workaround, this can be done in a future PR, so that we don't block this behemoth.

bryevdv: OK yes, that's fine. It's sorting that is important.

bokehjs/src/coffee/models/widgets/date_picker.coffee
mattpap: I started working on this.

mattpap: That's done.

bryevdv:
@mattpap I think we can close https://github.com/bokeh/bokeh/issues/5506 with this too

mattpap:
For some reason phantomjs is unable to load bokeh-widgets anymore. No explanation is given. I can reproduce this locally. Also, there is no color output on travis ci (at least in this PR).

mattpap:
I narrowed this down to slickgrid upgrade, but I have no idea what's wrong with this. Works everywhere expect in phantomjs, and phantom doesn't even throw me a bone, to give me at least minor clue why this fails. It just doesn't load, though bokeh-widgets is properly resolved.

mattpap:
I'm sick and tired of this BS. Last day I wasted on learning that lesscss use different base directories for file resolution in different features, and fails silently, or actually succeeds in a way, just with useless result. And now this.

canavandl:
@mattpap I'll shift gears and a look at the PhantomJS issues now, if you don't mind.

mattpap:
@canavandl, any help will be useful. Currently 62f287 is first failing commit and 15a40ecf is the last working. There are a few in between but they are removals only, so shouldn't be affecting anything. Make sure to use `npm install; npm prun; gulp dev-build` (or `gulp build`) when switching commits. I'm pretty sure this is a syntax error or some other thing, maybe related to strict mode, in slick grid.

mattpap:
Yeah it's strict mode.

mattpap:
Now I recall that we previously had to modify slick grid to work in strict mode. Looks like this wasn't fixed upstream and all our changes were removed when upgrading and removing slick grid from `vendor`.

mattpap:
@bryevdv, @canavandl, another pile of nonsense below:
```
INTERNALERROR>   File "/home/travis/miniconda/lib/python3.5/site-packages/_pytest/vendored_packages/pluggy.py", line 745, in __call__
  File "/home/travis/miniconda/lib/python3.5/site-packages/_pytest/main.py", line 105, in wrap_session
During handling of the above exception, another exception occurred:
During handling of the above exception, another exception occurred:
Exception ignored in: 
```
and
```
___________________________ TestBokehJS.test_bokehjs ___________________________
self = <tests.test_bokehjs.TestBokehJS testMethod=test_bokehjs>
    def test_bokehjs(self):
        os.chdir('bokehjs')
        proc = subprocess.Popen([join('node_modules', '.bin', GULP), "test"],
                                stdout=subprocess.PIPE)
        msg = out.decode('utf-8', errors='ignore')
        os.chdir('..')
>       print(msg)
tests/test_bokehjs.py:25: 
../../../miniconda/lib/python2.7/site-packages/colorama/ansitowin32.py:40: in write
    self.__convertor.write(text)
../../../miniconda/lib/python2.7/site-packages/colorama/ansitowin32.py:141: in write
    self.write_and_convert(text)
../../../miniconda/lib/python2.7/site-packages/colorama/ansitowin32.py:169: in write_and_convert
    self.write_plain_text(text, cursor, len(text))
self = <colorama.ansitowin32.AnsiToWin32 object at 0x7f1ba5b3b250>
text = "[14:51:50] Using gulpfile ~/build/bokeh/bokeh/bokehjs/gulpfile.js\n[14:51:50] Starting 'defaults:generate'...\n[14:51:5...size\n\r    \u2713 should remain below 750 kB limit\n\n\n  950 passing (16s)\n  8 pending\n\n[14:52:18] Finished 'test' after 26 s\n"
start = 0, end = 61004
    def write_plain_text(self, text, start, end):
        if start < end:
>           self.wrapped.write(text[start:end])
E           UnicodeEncodeError: 'ascii' codec can't encode character u'\u2713' in position 467: ordinal not in range(128)
../../../miniconda/lib/python2.7/site-packages/colorama/ansitowin32.py:174: UnicodeEncodeError
```
(that's 3.5/3.6 and 2.7 unit tests respectively; 3.4 apparently works fine).

bryevdv:
I just commented about the encoding problem in the flow, I'm only seeing that on 2.7 FWIW 

The others seem intermittent which sucks for sure.

mattpap:
@bryevdv, I ended up splitting of `bokeh-widgets` into `bokeh-widgets` and `bokeh-tables` (which includes jquery). Nice side effect of this is that the basic widgets weigh now only 90 kB minified.

mattpap:
This is ready, though there are some rough edges to resolve later:

- data table works, but there is something wrong with positioning of scroll bars
- cell editors need further changes (I commented out some code that still uses jquery-ui)

On the other hand, I improved styling of sliders, so they aren't as bulky as they were in PR #5774.

mattpap:
Also note that currently slickgrid's and pikaday's CSS classes aren't prefixed (besides being under `.bk-root`). nouislider is prefixed, but that's because I added support for this to the library, so we don't have to modify it just set one variable in less. Prefixing can be added later in the release cycle if needed, but I think we are good with what we have. Current setup doesn't allow bokehjs to interfere with the environment, though environment may interfere with bokehjs (stupid CSS specificity rules), but that's not very likely.

bryevdv:
I will make an effort to look at this and test with it locally to get it merged over the weekend. There is currently a real (but probably trivial to fix error) in the js tests:
```
Error: Cannot find module '/home/travis/build/bokeh/bokeh/bokehjs/test/../build/js/tree/models/widgets/data_table'
```

bryevdv:
@mattpap I pushed changes to report the now js/css file sizes and also a small fix for the defaults test but please review my changes

bryevdv:
@mattpap `models/server/widgets.py` does not display anything really and reports JS console error:
```
TypeError: undefined is not an object (evaluating 'this.grid.invalidate')
```

bryevdv:
OK apart from that one issue above pretty much every widget example that is expected to work, seems to (`linked_tap.py` has an unrelated issue)

mattpap:
> models/server/widgets.py does not display anything really and reports JS console error:

I overlooked one jquery call in cell_formatters.

bryevdv:
OK,I had a cursory glance at the diffs. Nothing jumped out, so I will merge this today to avoid having such a large PR in linger. There are definitely some follow up things, will make some issues for those as well.

mattpap:
@bryevdv, do sliders look OK now or should there be further improvements?

bryevdv:
> do sliders look OK now or should there be further improvements?

I think we will want to tweak them more, but we should probably solicit some input from various design-y people to help guide the changes

PR #6337, opened 2017-05-29
Silence all change events during initialization of models

### LINKED ISSUE #6336: Silence all change signals during initialization of models
This will increase performance and allow for easier debugging (less signal noise).

mattpap:
This is a side effect of work on layout. Changes in this PR:

- [x] silence change events on init
- [x] fix double initialization bug in `Signal.constructor`
- [x] better printing of properties (to better visualize change signals)
- [x] don't include `@type` in `@id`, because `toString()` already handles this
- [x] make `uniqueId` produce simple ids in dev mode (requires setting `dev` in `core/util/string`)

fixes #6336


bryevdv:
OK, LGTM and tested with ~30 various examples

PR #6340, opened 2017-05-30
Add support for namespace models

### LINKED ISSUE #3674: Add DataModel base class

mattpap:
fixes #3674


mattpap:
@bryevdv, if we are going to proceed with this, can you add docstrings?

bokeh/document.py
```diff
@@ -241,6 +241,27 @@ def add_periodic_callback(self, callback, period_milliseconds):
...
+    def set_var(self, name, value):
+        '''
+        '''
+        self._get_namespace().vars[name] = value
+
+    def get_var(self, name):
+        '''
+        '''
+        return self._get_namespace().vars[name]
+
```
bryevdv: A similar issue is coming up in some continued events work @jlstevens is working on for "document events" (like startup/busy/done/etc). Its the same situation where an object needs to be added as a root in order to interact with properties or events on it. I'd like to handle these cases consistently. What we discussed:
* always just add the object to `roots` at doc creation
* keep an internal reference (so no need to search for it in `_roots`
* disallow removing the object from roots (check in `remove`, throw an exception)

I'd like to do that here as well. Then also have a readonly `vars` property that defines `getattr` and `setattr` so that this works:

```
curdoc().vars.foo = 10
```

bryevdv:
>  if we are going to proceed with this, can you add docstrings?

Sure, no problem. Also should mention a prime motivation is also something like this:
```
plot.title.text = { 'var': 'foo' }

# then this will update the title
curdoc().vars.foo = "new title"
```
With the idea that if there is a small set of things people care about, this would allow them to make meaningful and easily accessible aliases to them, rather than have to dig through a potentially complicated object graph to find the property they want to update. 

This could potentially be later work, though

bryevdv:
@mattpap also to be clear about priorities, this can wait another release. Bug fixes are the most important thing right now.

mattpap:
@bryevdv, I know but I needed to distract my self with some easy work.

bryevdv:
@mattpap if you will fix up the merge conflict I will finish up the part about making it a mandatory  `root`. It will help me figure out what to to with the events bus case as well.

bryevdv:
OK @mattpap I got this PR up to date with master, I will try to push an "API" commit by tomorrow morning

bryevdv:
@mattpap This fails on server apps because it tries to render a view for the namespace. I thought we added a `Viewable` base class at some point? Was that in an unmerged PR? Once that is fixed I think I'd be satisfied with something like 
```
diff --git a/bokeh/document.py b/bokeh/document.py
index 0aabdc37d..2588b6cde 100644
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -80,6 +80,16 @@ class EventManager(object):
         self.subscribed_models = set() # Models subscribed to events


+class _NamespaceProxy(object):
+    def __init__(self, namespace):
+        object.__setattr__(self, '_namespace', namespace)
+
+    def __getattr__(self, attr):
+        return self._namespace.vars[attr]
+
+    def __setattr__(self, attr, value):
+        self._namespace.vars[attr] = value
+
 class Document(object):
     ''' The basic unit of serialization for Bokeh.

@@ -263,6 +273,13 @@ class Document(object):
         '''
         return self._get_namespace().vars[name]

+    @property
+    def vars(self):
+        '''
+
+        '''
+        return _NamespaceProxy(self._get_namespace())
+
     def add_root(self, model, setter=None):
```

bryevdv:
Actually I think this will require some re-thinking. I need to elaborate the use-cases I'd like to be able to support and we can work from that to what really needs to be implemented.

p-himik:
I'm trying to nicely integrate React-ish input forms with Bokeh. So, basically, when I change some input, Bokeh plots some new data based on that input.
It's fairly easy to implement using Bokeh widgets, but I don't like the approach as it makes it much harder to integrate plots into an existing app.

In the UI that I'm implementing, most of the views subscribe to values in a global data storage. My intent is to issue changes this global data storage (hence, update the view) in a reaction to some changes on the Bokeh server side.

I think that for this case it would be great to have some kind of a DataStore with `values: Dict(String, Any)` and that can be added to a document without any warnings/errors regarding it not being renderable.
It would make it possible to subclass such DataStores so they update the global data storage in the way I need.


With that being said, I think current implementation in this PR could be improved by allowing multiple named Namespace objects that can be accessed individually with `doc.get_model_by_name()`.
Also, I can probably update this PR with my suggestion if everyone agrees.

p-himik:
Also related: #3349

bryevdv:
Lots of conflicts now. Closing this, will return to it in a new PR

PR #6342, opened 2017-05-30
Implement proper rendered finished (idle) algorithm

### LINKED ISSUE #5879: Make "bokeh finished rendering heuristic" work with non-plot examples
Currently they time out, because "finished" event isn't ever triggered.

mattpap:
- [x] support `ImageURL`
- [x] support `GMapPlotCanvas`
- [ ] support `TileRenderer`

fixes #5879

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -572,7 +572,7 @@ def _wait_until_render_complete(driver):
     script = """
     // add private window prop to check that render is complete
     window._bokeh_render_complete = false;
-    window.addEventListener("bokeh:rendered", function() {
+    window.addEventListener("bokeh:idle", function() {
         window._bokeh_render_complete = true;
     });
     """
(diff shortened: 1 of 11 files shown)


bryevdv:
ping @jlstevens FYI once this is merged we can/should also expose as a `bokeh.events` event as well.

bryevdv:
@mattpap AFAIK image url and gmap are by far the more common cases. So I would definitely not hold up merging this just to get the tile renderer done. We can note a known issue and make a separate GH issue to add support for it later.

bryevdv:
@mattpap I have resolved the trivial conflict in the GH UI

mattpap:
@bryevdv, I was experimenting with this PR over weekend and it doesn't yet work reliably (with `TileRenderer` put aside). This needs some more work.

mattpap:
I improved this PR and merged into #6384, because it's needed there to proceed with testing.

Issue #6345, opened 2017-05-30
Move canvas2svg to separate optional bundle


canavandl:
Move canvas2svg to optionally loaded bundle (like webgl)

birdsarah:
I ran examples/howto/layouts/dashboard.py with the bollinger plot set to svg output_backend.

The patch bleeds outside the plot area. Maybe this is captured by an above point, but I thought I'd mention it.
![screen shot 2017-05-30 at 7 20 41 pm](https://cloud.githubusercontent.com/assets/1796208/26610445/1ff83666-456d-11e7-8bd2-e4f6d88ac274.png)
![screen shot 2017-05-30 at 7 20 28 pm](https://cloud.githubusercontent.com/assets/1796208/26610450/2246f092-456d-11e7-9cb8-21d6828ac801.png)

mattpap:
canvas2svg was integrated into bokehjs' core in PR #10402.

PR #6347, opened 2017-05-30
Canavandl/6345 deprecate webgl attr

canavandl:
Deprecate ``webgl`` attr in place of ``output_backend="webgl"``.

This is one to prevent conflicts between the new SVG backend and webgl.

- [x] issues: related to #6345
- [ ] tests added / passed

### DIFF
--- a/examples/models/file/maps_cities.py
+++ b/examples/models/file/maps_cities.py
@@ -26,7 +26,7 @@ plot = GMapPlot(
     plot_height=500,
     map_options=map_options,
     api_key=API_KEY,
-    webgl=True,
+    output_backend=True,
 )
 
 if plot.api_key == "GOOGLE_API_KEY":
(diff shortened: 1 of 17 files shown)


bryevdv:
What about `canvas-gl` as the mode name? indicating canvas plus webgl when feasible

examples/models/file/maps_cities.py
```diff
@@ -26,7 +26,7 @@
     plot_height=500,
     map_options=map_options,
     api_key=API_KEY,
-    webgl=True,
+    output_backend=True,
```
mattpap: `"webgl"`?

canavandl:
> What about canvas-gl as the mode name? indicating canvas plus webgl when feasible

I'm -0.5 on "canvas-gl" because WebGL is a known thing, but I'm willing the make the change. I definitely to migrate the old ``webgl`` attr docstring over, where it mentions that glyphs without webgl implementations will fallback to canvas.

examples/models/file/maps_cities.py
canavandl: yup - that test just failed :)

PR #6349, opened 2017-05-31
gallery toctree path should not use os separator, always backslash

### LINKED ISSUE #6331: Re-building docs - KeyError gallery exception
Bokeh - all versions.
Env: anaconda on win 64 python 3
When building first time  - all is good.
Next round gives following error:
Output:
`Running Sphinx v1.4.8
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 30 added, 6 changed, 29 removed
bk_script_f0df36cdd9f143e087b632c47702ab71:41: RuntimeWarning: divide by zero encountered in divide
bk_script_f0df36cdd9f143e087b632c47702ab71:41: RuntimeWarning: divide by zero encountered in log
bk_script_f0df36cdd9f143e087b632c47702ab71:41: RuntimeWarning: invalid value encountered in multiply
bk_script_f0df36cdd9f143e087b632c47702ab71:42: RuntimeWarning: divide by zero encountered in log
reading sources... [100%] index
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
adding links to sitemap... [100%] http://bokeh.pydata.org/en/0.12.6dev6/docs/reference/models/markers.html
copying bokeh-plot files... [100%] bokeh-plot-ff7d4e4a58f23d8a18428273c04d26ea.js.jsternal-0.jssss0.js-0.js

Exception occurred:
  File "C:\Users\Maxim\Anaconda3\envs\docs-unified\lib\site-packages\Sphinx-1.4.8-py2.7.egg\sphinx\environment.py", line 1210, in get_toc_for
    tocdepth = self.metadata[docname].get('tocdepth', 0)
KeyError: u'docs\\gallery\\anscombe'
The full traceback has been saved in c:\users\maxim\appdata\local\temp\sphinx-err-7ognmc.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!`

After that "pickled environment" file is removed and next make html passes ok but it builds everything again.

maxgamurar:

- [ ] issues: fixes #6331
- [ ] tests added / passed

### DIFF
--- a/bokeh/sphinxext/bokeh_gallery.py
+++ b/bokeh/sphinxext/bokeh_gallery.py
@@ -42,7 +42,7 @@ class BokehGalleryDirective(BokehDirective):
             src_path = abspath(join("..", detail['path']))
             dest_path = join(dest_dir, detail['name'] + ".py")
 
-            # sphinx pickeled env works only with backslash
+            # sphinx pickled env works only with forward slash
             docname = join(env.app.config.bokeh_gallery_dir, detail['name']).replace("\\","/")
 
             try:
--- a/bokeh/sphinxext/bokeh_gallery.py
+++ b/bokeh/sphinxext/bokeh_gallery.py
@@ -41,7 +41,7 @@ class BokehGalleryDirective(BokehDirective):
         for detail in details_iter:
             src_path = abspath(join("..", detail['path']))
             dest_path = join(dest_dir, detail['name'] + ".py")
-            docname = join("docs", "gallery", detail['name'])
+            docname = join("docs", "gallery", detail['name']).replace("\\","/")
 
             try:
                 copyfile(src_path, dest_path)
--- a/bokeh/sphinxext/bokeh_gallery.py
+++ b/bokeh/sphinxext/bokeh_gallery.py
@@ -41,7 +41,9 @@ class BokehGalleryDirective(BokehDirective):
         for detail in details_iter:
             src_path = abspath(join("..", detail['path']))
             dest_path = join(dest_dir, detail['name'] + ".py")
-            docname = join("docs", "gallery", detail['name']).replace("\\","/")
+
+            # sphinx pickeled env works only with backslash
+            docname = join(env.app.config.bokeh_gallery_dir, detail['name']).replace("\\","/")
 
             try:
                 copyfile(src_path, dest_path)
@@ -66,5 +68,6 @@ def env_updated_handler(app, env):
     return getattr(env, 'gallery_updated', [])
 
 def setup(app):
+    app.add_config_value('bokeh_gallery_dir', join("docs", "gallery"), 'html')
     app.connect('env-updated', env_updated_handler)
     app.add_directive('bokeh-gallery', BokehGalleryDirective)


bryevdv:
@maxgamurar At a glance this seems like the wrong solution to me. I guess the question I'd rather ask is "why do the keys in the env dict not have os-separators to begin with?" then fix *that* problem.

bryevdv:
Tested locally with spectrogram, CPU usage in line with previous levels (and remain there over time).  

@mattpap if there is an underlying other issue to resolve or look into, I'd suggest and issue to keep track of it.

maxgamurar:
@bryevdv I think sphinx has a strict rule how to format urls in toctree. So, if you think its a bug with sphinx, then most likely we will never get that fixed. So anyway, this solution works for us and we are going to use it in our local version of docsgallery sphinxext...

bryevdv:
> I think sphinx has a strict rule how to format urls in toctree

Do you know if that is documented somewhere? If that's the case, that's the case and that's fine I just had never heard of that. 

>  So, if you think its a bug with sphinx, then most likely we will never get that fixed. 

I've had them accept PRs from me in a few days. :) 

> we are going to use it in our local version of docsgallery sphinxext.

It would be a mistake to maintain your own separate version of sphinxext IMO. If this is the best solution there is, great we will merge it. I'd just like to expend a minimal effort to make sure there isn't a better option. Also if this is the only solution that line should have a comment explaining the purpose of the substitution.

maxgamurar:
About blackslash, its not documented but I come to that conclusion myself after trying to solve that issue, it seems that picked env accepts only black slash and forward slash makes a toctee key unusable.

About sphinx and PR - I know they will accept but we can not use latest version of sphinx, for now we use 1.4. So for now we will use our local sphinxext with gallery, not only because of slash issue, also we have a custom path for gallery files and maybe I will add one more PR to allow custom gallery path.

bryevdv:
In that case you should just add an explanatory comment so this PR can be merged.

maxgamurar:
ok, I will do.
Honestly for me it looks like a temporary solution too but the fastest one.

maxgamurar:
@bryevdv added comment and custom dir config option

bryevdv:
Where was the config option discussed? Is there an issue? What is the use-case?

maxgamurar:
See my comment https://github.com/bokeh/bokeh/pull/6349#issuecomment-305360494
Do you want me to create separate PR for that?

bryevdv:
Yes please make a separate issue and PR or distinct features and fixes. In the issue please also describe what the purpose of this new config option Not it's mechanical action, which is evident, but the intended purpose or an example use case of why it is necessary. Perhaps there is a better way to accomplish whatever it is you are trying to accomplish without adding new features.

bokeh/sphinxext/bokeh_gallery.py
```diff
@@ -41,7 +41,9 @@ def run(self):
         for detail in details_iter:
             src_path = abspath(join("..", detail['path']))
             dest_path = join(dest_dir, detail['name'] + ".py")
-            docname = join("docs", "gallery", detail['name'])
+
+            # sphinx pickeled env works only with backslash
```
bryevdv: typo: pickled

bryevdv: also the comment is wrong? It should say "only works with forward slash" ?

bryevdv: typo: should be "pickled", also the content of the comment is wrong? It should say "only works with forward slash" ?

bryevdv:
Actually, since both things affect the same line of code I am OK with one PR, but please provide a justification for why this config option is necessary in a comment here.

maxgamurar:
I need custom gallery location, because bokeh will be  ported into main docs.continuum.io portal and location of gallery files would be different.

bryevdv:
I understand what the option *does* I'm asking why you think you need to put the gallery files in a different location at all.

maxgamurar:
We have different file structure and it does not match hardcoded gallery path in sphinx extension.
Bokeh docs are located in another folder, so plugin fails to generate and upload gallery on correct folder.

bokeh/sphinxext/bokeh_gallery.py
bryevdv: @maxgamurar can you update the comment? then I will merge

bryevdv:
> We have different file structure and it does not match hardcoded gallery path in sphinx extension.
Bokeh docs are located in another folder, so plugin fails to generate and upload gallery on correct folder.

OK, but please be aware this option may have to be removed or changed later when more work is done on the gallery directive.

maxgamurar:
Why you may have to remove it ? It does not affect original functionality and just adds new sphinx option to point different location for gallery files.

bryevdv:
I accidentally added a comment in an edit to the wrong place, so you probably did not see it:

> My reticence here is because I do intend to make the gallery directive more general, to support e.g. multiple galleries. In which case the notion of a single "gallery dir" is a bad one to introduce.

maxgamurar:
Ok, I see now, thanks for clarification. But anyway, having one with hardcoded dir path is not good and ability to set custom path is better.

PR #6350, opened 2017-05-31
server.py: add max_buffer_size argument

### LINKED ISSUE #6338: Feature request: configurable maximum upload size for Tornado server
I'd like to lift the default maximum Tornado upload size of 100MB. This requires adding a limit to [server.Server](https://github.com/bokeh/bokeh/blob/master/bokeh/server/server.py#L85), which then passes it to [tornado.httpserver.HTTPServer](https://github.com/bokeh/bokeh/blob/master/bokeh/server/server.py#L131) as `max_buffer_size=`.

bkueng:
Configurable maximum upload size, see  #6338

Fixes  #6338

### DIFF
--- a/bokeh/server/server.py
+++ b/bokeh/server/server.py
@@ -80,9 +80,14 @@ class Server(object):
     Kwargs:
         num_procs (str):
             Number of worker processes for an app. Default to one. Using 0 will autodetect number of cores
+        tornado_server_kwargs (dict):
+            Additional arguments passed to tornado.httpserver.HTTPServer. E.g. max_buffer_size to
+            specify the maximum upload size. More details can be found at:
+
+            http://www.tornadoweb.org/en/stable/httpserver.html#http-server
     '''
 
-    def __init__(self, applications, io_loop=None, **kwargs):
+    def __init__(self, applications, io_loop=None, tornado_server_kwargs=None, **kwargs):
         log.info("Starting Bokeh server version %s (running on Tornado %s)" % (__version__, tornado.version))
 
         if isinstance(applications, Application):
@@ -115,6 +120,10 @@ class Server(object):
         port = kwargs.get('port', DEFAULT_SERVER_PORT)
         self._address = kwargs.get('address') or None
 
+        if tornado_server_kwargs is None:
+            tornado_server_kwargs = {}
+        tornado_server_kwargs.setdefault('xheaders', kwargs.get('use_xheaders', False))
+
         self._num_procs = kwargs.get('num_procs', 1)
         if self._num_procs != 1:
             assert all(app.safe_to_fork for app in self._applications.values()), (
@@ -128,7 +137,7 @@ class Server(object):
             tornado_kwargs['redirect_root'] = kwargs.get('redirect_root', True)
 
             self._tornado = BokehTornado(self._applications, self.prefix, **tornado_kwargs)
-            self._http = HTTPServer(self._tornado, xheaders=kwargs.get('use_xheaders', False))
+            self._http = HTTPServer(self._tornado, **tornado_server_kwargs)
             self._http.start(self._num_procs)
             self._http.add_sockets(sockets)
 


bryevdv:
This looks good, would it be worth also making it a command line option `--max-buffer-size`?

bkueng:
Correct me if I'm wrong, but in order to use this you need a custom Tornado upload request handler, and for that you need a custom serve/startup script. Meaning I don't see any way to make direct use of a `--max-buffer-size` argument.

bryevdv:
@bkueng you might be right, I thought I remembered something about a new (lower) default websocket payload limit for tornado, and assumed this was in response to that but I think I was assuming incorrectly based on your statement. 

Are there many other tornado args that users might want to supply when starting a server programmatically? If there are more than just one or two, maybe it would make more sense to make a generic `tornado_kwargs` parameter that accepts a dict, then this entire dict is applied as kwargs to  Tornado. That way there is a path to customizing any tornado args, not just this one.

bkueng:
I thought about this, and the only thing I saw that might be useful are SSL options, but from reading the docs, one should rather use an SSL proxy for that.
I leave it up to you if you want to be more generic.

bryevdv:
> I thought about this, and the only thing I saw that might be useful are SSL options, but from reading the docs, one should rather use an SSL proxy for that.

Actually I think there is interest in bing able to run Bokeh in an existing Tornado app, and to add SSL (the question came up explicitly on the ML recently). It's demo'd with nginx in the docs because that happens to be what I know how to do. I'd love to show the "pure tornado" option if possible.

I think I have a slight preference for the general tornado kwargs dict, with a link to the relevant section of the tornado docs that list all the possible kwarrgs, as well as calling out a few of the most useful ones (like this one) explicitly in example usage. Thoughts?

bkueng:
I updated for the more general dict. It makes it also more flexible if tornado adds further arguments in the future.

I haven't found a list of all supported arguments, just this: http://www.tornadoweb.org/en/stable/httpserver.html#http-server

bokeh/server/server.py
```diff
@@ -128,7 +134,7 @@ def __init__(self, applications, io_loop=None, **kwargs):
             tornado_kwargs['redirect_root'] = kwargs.get('redirect_root', True)
 
             self._tornado = BokehTornado(self._applications, self.prefix, **tornado_kwargs)
-            self._http = HTTPServer(self._tornado, xheaders=kwargs.get('use_xheaders', False))
+            self._http = HTTPServer(self._tornado, xheaders=kwargs.get('use_xheaders', False), **tornado_server_kwargs)
```
bryevdv: There is a slight chance for user error here. If someone puts `xheaders` in `tornado_server_kwargs` there will be a duplicate kwarg passed to `HTTPServer`. I think we can be slightly more careful here. How about:
```
tornado_server_kwargs.setdefault('xheaders', kwargs.get('use_xheaders', False))
self._http = HTTPServer(self._tornado, **tornado_server_kwargs)
```

This should have any `xheaders` value already in `tornado_server_kwargs` (which should only really happen if we are *not* running in the from the command line) take precedence, otherwise use the `use_xheaders` or default to `False`

bokeh/server/server.py
```diff
@@ -80,9 +80,12 @@ class Server(object):
     Kwargs:
         num_procs (str):
             Number of worker processes for an app. Default to one. Using 0 will autodetect number of cores
+        tornado_server_kwargs (dict):
+            Additional arguments passed to tornado.httpserver.HTTPServer. E.g. max_buffer_size to
+            specify the maximum upload size.
```
bryevdv: suggest adding 
```
specify the maximum upload size. More details can be found at:

http://www.tornadoweb.org/en/stable/httpserver.html#http-server
```

bryevdv:
@bkueng left two small comments, otherwise looking good!

bokeh/server/server.py
bryevdv: I might also suggest the name `http_server_kwargs` tho I would not object to keeping it as is either

bkueng:
Thanks for the feedback, I updated accordingly

bryevdv:
Thanks @bkueng !

PR #6352, opened 2017-05-31
Hack around unset document in Box's layout computations

### LINKED ISSUE #6261: Adding new sub-layouts fails to set documemt
ref: #5856 

BOTH versions of @cutright code in the referenced issue "good" and "bad" do not work (for the same reason), namely, `.document` on the newly added layout is `null`

Hopefully this is just a small oversight, and making sure `document` is set on new models in `PATCH-DOC` updates is all that is needed

ping @mattpap 

screen shot 2017-05-08 at 17 49 16

Here is the "good" version of the code without tabs (that is now also failing):
```
from bokeh.layouts import layout, row
from bokeh.models.widgets import Select, Button, Tabs, Panel

from bokeh.io import curdoc

source = ColumnDataSource(data={'x': [0, 1, 2, 3, 4],
                                'y': [0, 1, 4, 9, 16]})


def update():
    pass


def button_add_selector_row():
    layout_query.children.append(AddSelectorRow().row)


class AddSelectorRow:
    def __init__(self):
        self.select_category = Select(value="option1", options=["option1", "option2"])
        self.row = row([self.select_category])



# set up layout
plot = figure(plot_width=1000, plot_height=400)
plot.line('x', 'y', source=source)

# Set up initial buttons
main_add_selector_button = Button(label="Add Selector", button_type="default", width=400)
main_add_selector_button.on_click(button_add_selector_row)
layout_query = layout([[main_add_selector_button], [plot]])

curdoc().add_root(layout_query)

update()
```

The referenced issue has the other version (with tabs)

mattpap:
I feel dirty writing code like this, but I don't see any other option, perhaps besides restoring old logic with invalidating all models every time `children` change (which isn't any better). Now I know why this code was there before. There are two proper solutions to this issue. One is fix document to attach itself to all new objects, which, to my surprise, doesn't work or at least doesn't work early enough. The other is move layout logic to views, but due to `LayoutCanvas` model and lack of `LayoutCanvasView`, it's pretty much impossible right now to do so.

fixes #6261

### DIFF
--- a/bokehjs/src/coffee/models/layouts/box.coffee
+++ b/bokehjs/src/coffee/models/layouts/box.coffee
@@ -312,7 +312,12 @@ export class Box extends LayoutDOM
   # once with horizontal=true and once with horizontal=false)
   _align_inner_cell_edges_constraints: () ->
     constraints = []
-    if @ in @document.roots()
+
+    # XXX: checking for `@document?` is a temporary hack, because document isn't always
+    # attached properly. However, if document is not attached then we know it can't be
+    # a root, because otherwise add_root() would attach it. All this layout logic should
+    # be part of views instead of models and use is_root, etc.
+    if @document? and @ in @document.roots()
       flattened = @_flatten_cell_edge_variables(@_horizontal)
       for key, variables of flattened
         if variables.length > 1


bryevdv:
OK I tested both the code in the issue as well as several other example apps. 

@mattpap please make an issue to revisit the commented fix, or make a reference to this PR in an appropriate existing issue.

PR #6361, opened 2017-05-31
fix promblems with render pause/unpause

### LINKED ISSUE #6328: GMap in Notebook unreliable again
I believe (but have not yet confirmed) that this happened in the `0.12.6dev2` to `dev3` time frame, which would possibly implicate the big layout PR. 

Behavior is for plot areas to render solid grey on the first `show` but often render correctly after a save and reload of the notebook. No out of the ordingary JS console messages or errors. 

ping @mattpap

bryevdv:
issues: fixes #6328

This fixes issues with `pause` and `unpause`. These functions were being called in the `PlotCanvasView` initializer, but *not* in the subclass `GMapPlotView` initializer. This could result in calls to `request_render` too early.  Specifically wrt to gmaps, this could cause `_build_map` to be called before the plot was otherwise initialized and ready to render, which in turn results in a blank gmap. 

To fix pause/unpause for now, needed to change from simple boolean to a condition counter to keep track of "pause level", so that on pause/unpause pair nested in side another (via a function call, say) would be handled correctly. Additionally, put an explicit guard in `render` itself to prevent any renders during a pause regardless of how they were triggered. I don't think this is strictly necessary, but seems like a reasonable preventative measure. 

With this change I am unable to reproduce any issues with gmaps inside or outside of notebooks, with either inline or server resources. 

I also fixed a problem with zoom range updates that could lead to unbalanced pause/unpause.

### DIFF
--- a/bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
@@ -17,7 +17,10 @@ load_google_api = (api_key) ->
 export class GMapPlotCanvasView extends PlotCanvasView
 
   initialize: (options) ->
+    @pause()
+
     super(options)
+
     @zoom_count = 0
 
     mo = @model.plot.map_options
@@ -31,6 +34,8 @@ export class GMapPlotCanvasView extends PlotCanvasView
       if not window._bokeh_gmaps_callback?
         load_google_api(@model.plot.api_key)
       gmaps_ready.connect(() => @request_render())
+
+    @unpause()
 
   update_range: (range_info) ->
     # RESET -------------------------
@@ -47,13 +52,14 @@ export class GMapPlotCanvasView extends PlotCanvasView
 
     # ZOOM ---------------------------
     else if range_info.factor?
-      @pause()
 
       # The zoom count decreases the sensitivity of the zoom. (We could make this user configurable)
       if @zoom_count != 10
         @zoom_count += 1
         return
       @zoom_count = 0
+
+      @pause()
 
       super(range_info)
 
@@ -166,7 +172,7 @@ export class GMapPlotCanvasView extends PlotCanvasView
     @map.setOptions({zoom: @model.plot.map_options.zoom})
     @_set_bokeh_ranges()
 
-  # this method is expected and called by PlotView.render
+  # this method is expected and called by PlotCanvasView.render
   _map_hook: (ctx, frame_box) ->
     [left, top, width, height] = frame_box
     @canvas_view.map_el.style.top    = "#{top}px"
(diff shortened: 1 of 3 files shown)


bryevdv:
@mattpap I expect some day you will have a different but more robust solution but this seems like the conservative correct path for now.

bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
```diff
@@ -47,14 +52,15 @@ export class GMapPlotCanvasView extends PlotCanvasView
...
-      @pause()
 
       # The zoom count decreases the sensitivity of the zoom. (We could make this user configurable)
       if @zoom_count != 10
         @zoom_count += 1
         return
       @zoom_count = 0
 
+      @pause()
+
```
bryevdv: early return above could leave pause/unpause unbalance, moved after zoom count check

bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
```diff
@@ -17,7 +17,10 @@ load_google_api = (api_key) ->
 export class GMapPlotCanvasView extends PlotCanvasView
 
   initialize: (options) ->
+    @pause()
+
```
bryevdv: pause immediately, before anything else, in all canvas view initializers

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -42,17 +42,21 @@ export class PlotCanvasView extends DOMView
...
+      @is_paused += 1
 
   unpause: (immediate=false) ->
-    @is_paused = false
-    if immediate
-      @render()
-    else
-      @request_render()
+    @is_paused -= 1
+    if @is_paused == 0
```
mattpap: Is there a realistic possibility for calling `unpause()` more often than `pause()`? If so, this code won't catch that. Also, later on you use a non-strict condition, which could indicate such possibility. I would keep all conditions in line, so either use `<= 0` for unpaused and `> 0` for paused, or `== 0` for unpaused and `!= 0` for paused. Also, I would rename `@is_paused` to something else and private, and use `@is_paused` for a getter, so that you don't have to repeat the condition multiple times later.

mattpap:
I had a similar code floating around in earlier layout work, but didn't see a real need for it at the time.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
bryevdv: sounds reasonable I will do that and add tests as well

bryevdv:
@mattpap I made the requested changes

Issue #6341, opened 2017-05-30
Add parameter to autoload_server to control loading of JS/CSS resources


jphalip:
When embedding a plot from a separate server, the JS/CSS resources are then systematically loaded from that same server as well. For optimization purposes, I'd like to load those resources from a separate caching server instead living at a separate domain (e.g. 'cdn.pydata.org'). To achieve this I'd need a way of telling the plot server to *not* return those resources.

So it'd be great if a new `load_resources` parameter (defaults to `True`) could be added to `bokeh.embed.autoload_server()`. This parameter would then be passed on as a querystring parameter to the `autoload.js` script, which would then be used by the `AutoloadJsHandler` on the plot server to decide whether or not to include those resources when rendering the `AUTOLOAD_JS` template.

Does that sound reasonable? If so then I'm happy to provide a patch for this.

Thanks!

bryevdv:
Well, half the point of `autoload_server` is really to load the resources (It's right there on the label! :) ). If you want to load resources yourself,  I think that's reasonable, but rather than adding a parameter to `autoload_server` that makes it not-load-things in some situations (which is complicated/confusing) I'd suggest a new function with a different name, e.g `embed_server` or something. I think it could (should) re-use the existing machinery (i.e. that kind of parameter would be fine, internally) so I don't think it would be much more work than what you've proposed. Thoughts?

jphalip:
That makes sense to me! Since the actual purpose of this is to only load the websockets connection, maybe this new function could be named `websockets_server`?

jphalip:
Yes `connect_server` is probably better. I'll work on a patch to see how that works. Thanks!

bryevdv:
Great, look forward to seeing it, thanks!

PR #6363, opened 2017-06-01
Embed exclude resource files

jphalip:
This is a WIP pull request for issue #6341.

After looking more closely at the codebase, I think it might be best to introduce a parameter to `autoload_server` instead of introducing an entirely new method (e.g. `connect_server` as previously suggested) after all. I understand that the term "auto" sort of implies that it would automatically load everything. However, the point of this PR is just to exclude the resource *files* so that they can be served from elsewhere (most typically a caching server). After this change, `autoload_server` would still provide other resources (e.g. inline JS/CSS).

Also, a more practical reason is that the term "auto" is already used in a lot of places (`AutoloadJsHandler`, `AUTOLOAD_JS`, etc.) and therefore extending the terminology would probably require adding a lot of boilerplate code for little benefit in the end.

Finally, I've added the parameter in a way that would be backwards-compatible. Let me know if I've missed anything though!

Let me know what you think. I'm happy to make adjustments and also to write tests and documentation once we settle on the final API.

 Thanks!


bokeh/server/views/autoload_js_handler.py
```diff
@@ -8,6 +8,7 @@
 
 from six.moves.urllib.parse import urlparse
 from tornado import gen
+from tornado.web import MissingArgumentError
```
jphalip: I forgot to remove this line. We don't need it so I'll remove it in a subsequent PR.

bryevdv: It will have to be removed in this PR, unused imports trigger hard CI test failures:
```
    @pytest.mark.quality
    def test_flake8():
        chdir(TOP_PATH)
    
        out, err = proc.communicate()
    
E         ./bokeh/server/views/autoload_js_handler.py:11:1: F401 'tornado.web.MissingArgumentError' imported but unused
E         ./build/lib/bokeh/server/views/autoload_js_handler.py:11:1: F401 'tornado.web.MissingArgumentError' imported but unused
E         
E       assert 1 == 0
E        +  where 1 = <subprocess.Popen object at 0x7f0c7d504710>.returncode
```
You should just be able to make the change in a new commit and push your branch, and it will update this PR automatically.

bryevdv:
Regarding the content of the PR, I'd support renaming the new function (with the new parameter) to be a private function, then having `autoload_server` and `connect_server` be thin wrappers around that new private function. It's definitely the case that the "auto" `autoload_server` specifically and has always referred to *automatically loading the BokehJS resources*, i.e. the things the new option would turn off loading. (This is in contrast to other methods like `components` where users are responsible for loading BokehJS by hand). So I continue to think that adding this parameter to `autoload_server` is confusing and hard to document.

bryevdv:
@jphalip let's actually iterate on this a bit. I'd really actually like to deprecate `autoload_server` before 1.0, because it does too much already (it combines embedding docs or sessions in one complicated function). I'd much rather see something like this:
```
def _connect_session_or_doc(....):
    # this is the "new" autoload_server you wrote, with the new parameter

def autoload_server(...):
    # calls _connect_session_or_doc

def server_doc(url, relative_urls=False, resources="default")
    
def server_session(model, session_id, relative_urls=False, resources="default")

```

These would be much easier to document, have better names, good for users all around. Since they are simpler and have better names, I would not mind having them have a `resources` parameter! The parameter should accept `None` (which does what you want and not embed any resources) or "default" (does what it currently does with default resources object) or a custom configured `Resources` object.

jphalip:
Thanks @bryevdv, I think this all makes sense and I agree the API looks much cleaner that way. I've updated the PR with your instructions. I had a few more comments and questions:

- The "autoload" terminology is still used in a lot of places (`AutoloadJsHandler`, `AUTOLOAD_JS`, `autoload.js`, etc.). Should that be cleaned up as well now or can that wait till a later cleanup phase in the future?
- I've added some documentation for the new `resources` parameter, but currently it's in the `autoload_server()` docstring, which doesn't make much sense. We'll probably have to rewrite all those docstrings anyways so I can clean that up in the end with the rest of the documentation.
- I wonder if at present the `None` value is correct for `resources` is correct. This prevents resource *files* from being loaded, but other resources are still loaded (see `js_raw`, `css_raw` in `AutoloadJsHandler`). Do you think maybe that could be confusing? Is `None` ok or should we use something like `no-files` instead?

bokeh/embed.py
```diff
@@ -528,6 +582,17 @@ def autoload_server(model=None, app_path=None, session_id=None, url="default", r
...
+            along with the document.
+
+            If ``default`` then the default JS/CSS bokeh files will be loaded.
+
+            If None then none of the resource files will be loaded. This is
+            useful if you prefer to serve those resource files via other means
+            (e.g. from a caching server). Be careful, however, that the resource
+            files you'll load separately are from the same version as the
+            server, otherwise the rendering may not work correctly.
+
```
bryevdv: I would not add this to the `autoload_server`, just pass `resources="default"` unconditionally in the call to `_connect_session_or_document` below. Then `autoload_server` does not change at all which is desirable since the plan would be to deprecate it soon.

bryevdv: It looks like it isn't actually added to the params, just the docstring entry should be removed.

bryevdv:
@jphalip this is looking good. It looks like some tests are failing due to the new URL param:
```
E       -  'src': 'http://localhost:8081/foo/bar/sliders/autoload.js?bokeh-autoload-element=750f5649-27e0-4277-a4cc-3b80e8e8c978&bokeh-app-path=/foo/bar/sliders'}
E       +  'src': 'http://localhost:8081/foo/bar/sliders/autoload.js?bokeh-autoload-element=750f5649-27e0-4277-a4cc-3b80e8e8c978&bokeh-app-path=/foo/bar/sliders&resources=default'}
E       ?                                                                                                                                                       ++++++++++++++++++
```
There are two possible approaches:
* update the tests
* allow the `resources` param to be omitted when `"default"` (with default value on the JS side being `"default"` if it is omitted. 

 I think I'd prefer the latter, then none of the current tests need to change. It would be good to add some tests around the other cases, though.

bokeh/embed.py
jphalip: Yep, that was the second point in my comment above. I'll clean this up eventually when we settle on the API and have to write docstrings for other methods as well.

jphalip:
@bryevdv Sure, that sounds good. I already have it so that the JS already assumes `"default"` by default. Are you saying that you'd like the `autoload_server()` output to omit the parameter altogether when the value is `default`?

jphalip:
I'll take a look at fixing the tests and maybe add a few more to consolidate the new API. I'll update the PR.

bryevdv:
@jphalip Not just `autoload_server`, basically am saying that 
```
&resources=default
```
should never show up in a rendered URL (should just, er, default to "default")

jphalip:
@bryevdv Right, that makes sense. I'll adjust accordingly.

jphalip:
@bryevdv So I've made some adjustments and added some tests and documentation. When you have a chance let me know what you think.

jphalip:
Oh, and just a note that I haven't included a test for the default=none case, I'll add that soon in the coming days.

jphalip:
Ok I've added the test. It seems that one of the runs on Travis has failed but apparently due to a glitch. It'd need to be re-run, I think.

sphinx/source/docs/user_guide/embed.rst
```diff
@@ -369,6 +370,35 @@ will look something like this:
...
+id and model must be provided. Another difference from ``autoload_server`` is
+that with those two methods one may choose to not load the JS/CSS resource
+files by passing a ``resources="none"`` parameter.
+
+Here is an example using ``server_document``:
+
+.. code-block:: python
+
+    from bokeh.embed import server_document
+    script = server_document(url="https://demo.bokehplots.com/apps/slider")
```
bryevdv: I think this can be passed as a positional argument now
```
server_document("https://demo.bokehplots.com/apps/slider")
```
which I think is preferable to demonstrate.

bryevdv:
I've restarted the test on Travis. This is looking really good! Thank you for paying attention to tests and docs, those make PRs exceptional. 

What do you think about expanding the `resources` parameter to accept `Resources` objects? I had in mind that all of these could work:
```
# supply js/css URLs from an explicit Resources object
server_document(..., resources=Resources(mode="server", ...))  

# don't supply any js/css URLs
server_document(..., resources=None)  

#  supply the default js/css URLs
server_document(...)
server_document(..., resources="default")  
```

Apart from that, I would only suggest expanding the docstrings for `server_document. Basically just copy any relevant info and args from `_connect_session_or_document` so that all the user-facing docstrings are complete on their own.

bryevdv:
@jphalip Just an FYI we are freezing today for `0.12.6` so I have marked this issue for `0.12.7` I think thats better, we can merge soon after this week's release and have a whole release cycle after to make sure there are no issues. Given that, I'd like to also to these things in this PR:

* deprecate `autoload_server`
* adjust any example/docs usage to use the new functions

Let me know what you think

jphalip:
@bryevdv Sure, that sounds reasonable. I'll also make the adjustments you've suggested in the next PR.

bryevdv:
@jphalip unless you object, I'd say go ahead and make those changes (adding a call to `deprecate` to `autoload_server` and updating any examples) in this PR so that all the work is in one unit. There are only a handful of examples to update:
```
❯ grip autoload_se
./embed/animated.py:
   31 : from bokeh.embed import autoload_server
   62 : """ % autoload_server(p, session_id=session.id, relative_urls=False)
./embed/autoload_static.py:
   23 :     <fixed>autoload_server</fixed>. For more information see the
./embed/widget.py:
   32 : from bokeh.embed import autoload_server
  169 : """ % autoload_server(layout, session_id=session.id, relative_urls=False)
./howto/server_embed/flask_embed.py:
    9 : from bokeh.embed import autoload_server
   64 :     script = autoload_server(url='http://localhost:5006/bkapp')
./howto/server_embed/tornado_embed.py:
   10 : from bokeh.embed import autoload_server
   22 :         script = autoload_server(url='http://localhost:5006/bkapp')
```

bryevdv:
FYI we are now targeting next Tuesday for the release (there are always delays) and could merge this any time after that.

jphalip:
@bryevdv Sure, I'll work on that. By the way, is it preferred to squash all intermediary commits into a single one?

bryevdv:
We default to "Squash and Merge" when merging PRs so the commit structure of the PR is really up to what you prefer

bryevdv:
@jphalip FYI, only because we are in the middle of a release candidate build, I have temporarily stopped the CI build for this PR

bryevdv:
@jphalip I have restarted the CI build now that the release is over :) If you can update the examples listed about I think we are good to merge!

jphalip:
@bryevdv I think I took care of all those examples in my previous commit. Could you take a look, see if anything's missing?

bryevdv:
@jphalip You correct, my apologies for missing those changes. Thanks for your contribution!

PR #6368, opened 2017-06-01
Canavandl/svg headless export

canavandl:
- [x] issues: related #6345 
- [x] tests added / passed

The main point of discussion is that I have this functionality as a separate function than ``bokeh.io.export``. Is that the correct idea? I think it is because A) users would potentially want a PNG screenshot of an SVG plot-containing layout and B) the product is different enough (a single PNG vs multiple SVGs) that I think it'd cause confusion if they were the same method. Perhaps we'd better rename ``export`` to ``export_as_png``?

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -669,7 +669,6 @@ def _get_svg(obj):
     webdriver = import_required('selenium.webdriver',
                                 'To use bokeh.io.export you need selenium ' +
                                 '("conda install -c bokeh selenium" or "pip install selenium")')
-
     # assert that phantomjs is in path for webdriver
     detect_phantomjs()
 
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -665,7 +665,7 @@ def export(obj, filename=None):
 
     return os.path.abspath(filename)
 
-def _get_svg(obj):
+def _get_svgs(obj):
     webdriver = import_required('selenium.webdriver',
                                 'To use bokeh.io.export you need selenium ' +
                                 '("conda install -c bokeh selenium" or "pip install selenium")')
@@ -700,7 +700,7 @@ def export_svgs(obj, filename=None):
     ''' Export the LayoutDOM object as a PNG.
 
     If the filename is not given, it is derived from the script name
-    (e.g. ``/foo/myplot.py`` will create ``/foo/myplot.png``)
+    (e.g. ``/foo/myplot.py`` will create ``/foo/myplot.svg``)
 
     Args:
         obj (LayoutDOM object) : a Layout (Row/Column), Plot or Widget object to display
@@ -719,7 +719,7 @@ def export_svgs(obj, filename=None):
         Glyphs that are rendered via webgl won't be included in the generated PNG.
 
     '''
-    svgs = _get_svg(obj)
+    svgs = _get_svgs(obj)
 
     if len(svgs) == 0:
         logger.warn("No SVG Plots were found.")
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -565,39 +565,17 @@ def _destroy_server(div_id):
     except Exception as e:
         logger.debug("Could not destroy server for id %r: %s" % (div_id, e))
 
-def _crop_image(image, left=0, top=0, right=0, bottom=0, **kwargs):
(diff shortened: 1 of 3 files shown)


bokeh/io.py
```diff
@@ -658,3 +664,75 @@ def export(obj, filename=None):
...
+    return serialized_svgs
+    """
+
+    svgs = driver.execute_script(svg_script)
+
+    driver.quit()
+
+    return svgs
+
+def export_svgs(obj, filename=None):
```
canavandl: maybe ``export_as_svgs``?

bokeh/io.py
```diff
@@ -658,3 +664,75 @@ def export(obj, filename=None):
...
+    save(obj, filename=html_path, resources=INLINE, title="")
+
+    driver = webdriver.PhantomJS()
+    driver.get("file:///" + html_path)
+
+    _wait_until_render_complete(driver)
+
+    svg_script = """
+    var serialized_svgs = [];
+    var svgs = document.getElementsByClassName('bk-root')[0].getElementsByTagName("svg");
```
canavandl: there's never going to be a case where a SVG is under the ``bk-root`` node, is there?

bokeh/io.py
```diff
@@ -658,3 +664,75 @@ def export(obj, filename=None):
...
+    svgs = _get_svgs(obj)
+
+    if len(svgs) == 0:
+        logger.warn("No SVG Plots were found.")
+        return
+
+    if filename is None:
+        filename = _detect_filename("svg")
+
+    for i, svg in enumerate(svgs):
```
canavandl: this seems brittle

bokeh/io.py
bryevdv: I don't have a preference

bokeh/io.py
bryevdv: Will `ImageURL` load SVGs? That's the only possibility I can think of without the presence of custom extensions. I think this is fine for now in any case, if there are corner cases later we can deal with them as they arise.

bokeh/io.py
bryevdv: More tests are always welcome but not sure what specifically you are referring to as brittle here

bokeh/io.py
canavandl: I think it's going to be ``export_svgs`` and renaming ``export`` -> ``export_png`` for clarity.

bokeh/io.py
canavandl: ImageURL doesn't work w/ SVGs, so I think we're fine.

bryevdv:
@canavandl last CI error is just a docs ref issue:
```
  File "/home/travis/miniconda/lib/python3.4/site-packages/Sphinx-1.5.1-py3.4.egg/sphinx/cmdline.py", line 296, in main
    app.build(opts.force_all, filenames)
  File "/home/travis/miniconda/lib/python3.4/site-packages/Sphinx-1.5.1-py3.4.egg/sphinx/application.py", line 333, in build
    self.builder.build_update()
  File "/home/travis/miniconda/lib/python3.4/site-packages/Sphinx-1.5.1-py3.4.egg/sphinx/builders/__init__.py", line 251, in build_update
    'out of date' % len(to_build))
  File "/home/travis/miniconda/lib/python3.4/site-packages/Sphinx-1.5.1-py3.4.egg/sphinx/builders/__init__.py", line 268, in build
    self.warn(*warning, **kwargs)
  File "/home/travis/miniconda/lib/python3.4/site-packages/Sphinx-1.5.1-py3.4.egg/sphinx/application.py", line 399, in warn
    raise SphinxWarning(warntext)
sphinx.errors.SphinxWarning: /home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/export.rst:25: ERROR: Undefined substitution referenced: "export_png".
Warning, treated as error:
/home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/export.rst:25: ERROR: Undefined substitution referenced: "export_png".
```

bryevdv:
@canavandl I just edited the file here in the GH UI, seemed simple enough that I couldn't botch it :) Will  merge when green.

bryevdv:
> that I couldn't botch it

ha-ha

PR #6371, opened 2017-06-01
Add height/width kwarg support to bokeh.io.export

### LINKED ISSUE #6369: Add height/width kwargs to bokeh.io.export to support resizing the exported PNG
Add height/width kwargs to bokeh.io.export to support resizing the exported PNG

canavandl:
- [x] issues: fixes #6369 
- [x] tests added / passed

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -668,7 +668,7 @@ def export(obj, filename=None, height=None, width=None):
     image = _get_screenshot_as_png(obj)
 
     if height is not None or width is not None:
-        image = _resize_image(height=height, width=width)
+        image = _resize_image(image, height=height, width=width)
 
     if filename is None:
         filename = _detect_filename("png")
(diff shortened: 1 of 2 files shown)


bokeh/io.py
```diff
@@ -652,6 +667,9 @@ def export(obj, filename=None):
     '''
     image = _get_screenshot_as_png(obj)
 
+    if height is not None or width is not None:
+        image = _resize_image(height=height, width=width)
```
mattpap: `_resize_image()` has different signature (`image`).

canavandl: fixed

bryevdv:
@canavandl I think this may need some consideration. I think we need to do more than call an image resize after the fact. I think we should try to generate images of the right size from the start. e.g. we could tinker with plot sizes before running through the export machinery.

bryevdv:
Will revisit in another PR

PR #6373, opened 2017-06-02
Don't use implicit slots, e.g. `@connect(@change, @set_data)`

### LINKED ISSUE #6365: ImageRGBA error during change events
Since https://github.com/bokeh/bokeh/pull/6313 was merged there is an error in the ``ImageRGBA._set_data`` method where it is being passed two copies of the source. This means that the ``indices``variable incorrectly refers to the source resulting in an error on this line

```coffeescript
if indices? and indices.indexOf(i) < 0
```

    Exception in Comm callback TypeError: e.indexOf is not a function

because the source does not have an ``indexOf`` method. I haven't tried reproducing this using bokeh server yet but here is a minimal example to reproduce this in the notebook:

```python
    
    import numpy as np
    
    output_notebook()
    
    N = 20
    for i in range(N):
        for j in range(N):
            view[i, j, 0] = int(i/N*255)
            view[i, j, 1] = 158
            view[i, j, 2] = int(j/N*255)
            view[i, j, 3] = 255
    
    cds = ColumnDataSource(data=dict(image=[img]))
    
    # must give a vector of images
    p.image_rgba(image='image', source=cds, x=0, y=0, dw=10, dh=10)
    
    handle = show(p, notebook_handle=True)  # open a browser
    
    img[:20, :20] += 10
    cds.data['image'][0] = img
    push_notebook(handle=handle)
```

Tested on: bokeh 0.12.6dev7

mattpap:
I went ahead and fixed everything, besides:
```
bokehjs/src/coffee/models/tools/toolbar.coffee:        @connect(tool.properties.active.change, @_active_change.bind(null, tool))
bokehjs/src/coffee/models/tools/toolbar_box.coffee:          @connect(proxy.properties.active.change, @_active_change.bind(null, proxy))
```
which looks suspicious on its own.

fixes #6365

### DIFF
--- a/bokehjs/src/coffee/models/annotations/box_annotation.coffee
+++ b/bokehjs/src/coffee/models/annotations/box_annotation.coffee
@@ -17,8 +17,8 @@ export class BoxAnnotationView extends AnnotationView
     # "data only updates" that tools might want to use
     if @model.render_mode == 'css'
       # dispatch CSS update immediately
-      @connect(@model.change, @render)
-      @connect(@model.data_update, @render)
+      @connect(@model.change, () -> @render())
+      @connect(@model.data_update, () -> @render())
     else
       @connect(@model.change, () => @plot_view.request_render())
       @connect(@model.data_update, () => @plot_view.request_render())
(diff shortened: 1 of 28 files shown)


mattpap:
> `Failed: Console errors: [{'level': 'SEVERE', 'message': 'TypeError: t is undefined',`

One real failure to fix. Unfortunately, there is no line information and minified resources result in useless variable name information.

PR #6374, opened 2017-06-02
Bryanv/6317 fix warnings

### LINKED ISSUE #6317: DeprecationWarnings on Python 3.6
We get the following deprecation warnings in our CI jobs:
```
distributed/bokeh/tests/test_scheduler_bokeh.py::test_simple
  /home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/bokeh/util/callback_manager.py:19: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
    return getargspec(callback.func)

distributed/bokeh/tests/test_scheduler_bokeh.py::test_TaskProgress_empty
  /home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/bokeh/core/property/bases.py:173: DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
    return new == old
```

Bokeh version is 0.12.5.

bryevdv:
issues: fixes #6317

@mattpap @canavandl I'd appreciate close review of this PR.

### DIFF
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -16,6 +16,7 @@ from copy import copy
 import types
 
 from six import string_types
+import numpy as np
 
 from ...util.string import nice_join
 from .containers import PropertyValueList, PropertyValueDict
@@ -163,20 +164,21 @@ class Property(PropertyDescriptorFactory):
         return self._readonly
 
     def matches(self, new, old):
-        # XXX: originally this code warned about not being able to compare values, but that
-        # doesn't make sense, because most comparisons involving numpy arrays will fail with
-        # ValueError exception, thus warning about inevitable.
-        try:
-            if new is None or old is None:
-                return new is old           # XXX: silence FutureWarning from NumPy
-            else:
-                return new == old
-        except (KeyboardInterrupt, SystemExit):
-            raise
-        except Exception:
-            # if we cannot compare (e.g. arrays) just punt return False for match
-            pass
-        return False
+        ''' Whether two parameters match values.
+
+        If either ``new`` or ``old`` is a NumPy array, then the result of
+        ``np.array_equal`` will determine if the values match..
+
+        Otherwise, the result of standard Python equality will be returned.
+
+        Returns:
+            True, if new and old match, False otherwise
+
+        '''
+        if isinstance(new, np.ndarray) or isinstance(old, np.ndarray):
+            return np.array_equal(new, old)
+
+        return new == old
 
     def from_json(self, json, models=None):
         ''' Convert from JSON-compatible values into a value for this property.
--- a/bokeh/util/callback_manager.py
+++ b/bokeh/util/callback_manager.py
@@ -3,46 +3,20 @@ mixin classes for adding ``on_change`` and ``on_event`` callback
 interfaces to classes.
 '''
 from __future__ import absolute_import
-from functools import partial
-from inspect import formatargspec, getargspec, isfunction, ismethod
-from types import FunctionType
+
 from ..events import Event
-
-def _callback_argspec(callback):
-    '''Bokeh-internal function to get argspec for a callable'''
-    if not callable(callback):
-        raise ValueError("Callbacks must be callables")
-
-    if isfunction(callback) or ismethod(callback):
-        return getargspec(callback)
-    elif isinstance(callback, partial):
-        return getargspec(callback.func)
-    else:
-        return getargspec(callback.__call__)
+from ..util.future import get_param_info, format_signature, signature
 
 def _check_callback(callback, fargs, what="Callback functions"):
     '''Bokeh-internal function to check callback signature'''
-    argspec = _callback_argspec(callback)
-    formatted_args = formatargspec(*argspec)
-    margs = ('self',) + fargs
+    sig = signature(callback)
+    formatted_args = format_signature(sig)
     error_msg = what + " must have signature func(%s), got func%s"
-    defaults_length = len(argspec.defaults) if argspec.defaults else 0
 
-    if isinstance(callback, FunctionType):
-        if len(argspec.args) - defaults_length != len(fargs):
-            raise ValueError(error_msg % (", ".join(fargs), formatted_args))
+    all_names, default_values = get_param_info(sig)
 
-    elif isinstance(callback, partial):
-        expected_args = margs if ismethod(callback.func) else fargs
-        keyword_length = len(callback.keywords.keys()) if callback.keywords else 0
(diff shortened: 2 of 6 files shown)


bokeh/core/property/bases.py
```diff
@@ -163,20 +164,21 @@ def readonly(self):
...
+        Otherwise, the result of standard Python equality will be returned.
+
+        Returns:
+            True, if new and old match, False otherwise
+
+        '''
+        if isinstance(new, np.ndarray) or isinstance(old, np.ndarray):
+            return np.array_equal(new, old)
+
+        return new == old
```
bryevdv: `array_equal` avoids the `FutureWarning` (except erroneously in one probably uncommon case: https://github.com/numpy/numpy/issues/9210) so this new code seems both more robust as well as simpler. Also by using `array_equal` I don't think there are any "expected" exceptions, so we can remove the `try...except` including the bad `except Exception` block

It's worth noting that the old notes about never comparing arrays were wrong, the `old==new` *did* compare arrays but returned bool arrays. We have existing tests that depend on arrays getting compared. All of them pass unchanged with this new code.

bokeh/util/callback_manager.py
```diff
@@ -3,45 +3,49 @@
...
-            raise ValueError(error_msg % (", ".join(fargs), formatted_args))
+    all_names, default_values = get_param_info(sig)
 
-    elif isinstance(callback, partial):
-        expected_args = margs if ismethod(callback.func) else fargs
+    # This is a bit hacky. Py2 reports self in param info for bound methods
+    # but Py3 doesn't. Handling by just dropping any the first argument any
+    # time it is 'self', which is probably more useful anyway.
+    if ismethod(callback) and len(all_names) and all_names[0] == 'self':
+        all_names = all_names[1:]
```
bryevdv: This is worth noting. Not 100% happy with it, but it seems ok. The upshot is that the reported signature for bound methods is more in line with the way Py3 does it (`self` is ignored)

bokeh/util/callback_manager.py
```diff
@@ -3,45 +3,49 @@
...
-    formatted_args = formatargspec(*argspec)
-    margs = ('self',) + fargs
+    sig = _callback_signature(callback)
+    formatted_args = format_signature(sig)
     error_msg = what + " must have signature func(%s), got func%s"
-    defaults_length = len(argspec.defaults) if argspec.defaults else 0
 
-    if isinstance(callback, FunctionType):
-        if len(argspec.args) - defaults_length != len(fargs):
-            raise ValueError(error_msg % (", ".join(fargs), formatted_args))
```
bryevdv: not worrying about `self` led to a simplification, only have to handle two cases:`partial` functions, and everything else.

bryevdv:
Py2 callback tests failing, I thought I rant them locally but I think I was using ` py.test` from a py3 env unintentionally

bokeh/util/callback_manager.py
```diff
@@ -3,46 +3,20 @@
...
-            raise ValueError(error_msg % (", ".join(expected_args), formatted_args))
-    # testing against MethodType misses callable objects, assume everything
-    # else is a normal method, or __call__ here
-    elif len(argspec.args) - defaults_length != len(margs):
-        raise ValueError(error_msg % (", ".join(margs), formatted_args))
 
+    all_names, default_values = get_param_info(sig)
+
+    if len(all_names) - len(default_values) != len(fargs):
+        raise ValueError(error_msg % (", ".join(fargs), formatted_args)+ " %r, %r" % (all_names, default_values))
```
bryevdv: this was greatly simplified. There is only one test, all the gorpy py2/py2 handling is in the `future` module

bokeh/util/future.py
```diff
@@ -56,3 +58,57 @@ def wraps(obj, attr_names=WRAPPER_ASSIGNMENTS, wraps=wraps):
...
+            sig = getargspec(func.__call__)
+            sig.args.remove('self')
+            return sig
+
+    def format_signature(sig):
+        from inspect import formatargspec
+        return formatargspec(*sig)
+
+    def get_param_info(sig):
+        return (sig.args, sig.defaults or [])
```
bryevdv: Not super happy with the modification to `args` here but they seem to work fine (all tests pass), this code is only used when there are errors anyway, and some day maybe we can drop python 2 support and delete all of this.

bokeh/util/callback_manager.py
```diff
@@ -3,46 +3,20 @@
...
-            raise ValueError(error_msg % (", ".join(expected_args), formatted_args))
-    # testing against MethodType misses callable objects, assume everything
-    # else is a normal method, or __call__ here
-    elif len(argspec.args) - defaults_length != len(margs):
-        raise ValueError(error_msg % (", ".join(margs), formatted_args))
 
+    all_names, default_values = get_param_info(sig)
+
+    if len(all_names) - len(default_values) != len(fargs):
+        raise ValueError(error_msg % (", ".join(fargs), formatted_args))
```
bryevdv: This was greatly simplified, this is basically the entire function, all the gorpy p2/py3 differences are handled in `bokeh.util.future`

bryevdv:
@mattpap  @canavandl I will merge this in an hour or so. If there's issues after the RC we can back it out.

bokeh/core/property/bases.py
philippjfr: I'm getting errors here now because it seems to be comparing the ``.data`` of a CDS (i.e. a dictionary of arrays), which raises errors like this:

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

philippjfr: I'll try to come up with a simple example.

bryevdv: @philippjfr This doesn't make any sense? The *only* path before was `new == old` and the new code just add a special case for a subset where the args are arrays. Everything else (e.g. including dicts?) goes through the same `new == old`. I don't understand, definitely need an example.

bryevdv: Where is the execution coming from, please *always* give that information, not just a single exception line without context. I am wondering if it's inside `array_equal` e.g.

philippjfr: Sorry, should have been clearer. I filed https://github.com/bokeh/bokeh/issues/6393. The error comes from the ``old==new`` comparison, because old and new are dictionaries of arrays.

mattpap: @philippjfr, that issue also doesn't provide full traceback, nor code for fast reproduction (e.g. to know the types of `new` and `old`).

PR #6375, opened 2017-06-03
Developer docker tools

jphalip:
I've created this docker container so I could more easily run the tests. I thought maybe this might be of interest to the project in order to facilitate external contributions. Let me know what you think!


bryevdv:
@jphalip I will have to go and dig it up the info, but it's actually possible to get ahold of the exact docker images that TravisCI uses, and run the travis install machinery in it. I did this by hand extensively when developing some of our automation and it was a huge help. Unfortunately I don't have the necessary docker experience to make that more general or easier for other people to do. But I would be a but more interested in something like that that mirrors the travis environment, because having paths to testing our build and test infra outside of actually running on Travis woudl be extremely valuable. Would you be interested in adapting this PR more towards that? I can dig up the links to the Travis containers, if so.

jphalip:
@bryevdv That sounds like a good idea and I'm happy to look into it. Are you referring to this specifically? https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally

bryevdv:
@jphalip I beliece that's the same info, I recall I was using one of the quay.io images fairly sure we use `travis-python` tho I suppose we could use `travis-default` since we manage python versions ourself with conda

jphalip:
Ok that sounds good, I'll take a look into it.

bryevdv:
@jphalip Awesome, thanks! BTW if you are interested in continuing to contribute to Bokeh after these PRs you have open now, I'd be happy to add you to @bokeh/dev

jphalip:
Ok thanks @bryevdv, I'm happy to be included in the dev team.

bryevdv:
FYI I just updated the Travis CI build to use the new "trusty" ubuntu dist. I' not sure offhand where the docker images for that are, but if we can provide instructions and any tools that help folks get setup to use those locally that would be valuable.

bryevdv:
FYI here is some info about obtaining the "trusty" images:

https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally

bryevdv:
Probably also worth mentiong that https://github.com/bokeh/bokeh/pull/6542 should be merged soon. That refactors the CI automation with several scripts under `scripts/ci/` The most relevant of which are probably `build` and `install`. I'd definitely love to see see docker tools that:

* download the official Travis generic "Trusty" images
* utilize the new ci scripts to perform dependency installs and builds

jphalip:
@bryevdv Thanks for the headsup about that PR. I'll keep an eye on it and will try to integrate those scripts with the Travis image.

bryevdv:
FYI that PR was merged today

jphalip:
So I came up with this simple Dockerfile as an initial attempt:

```
FROM travisci/ci-garnet:packer-1496954857

ENV PATH="/home/travis/miniconda/bin:${PATH}"
ENV MINICONDA_FILENAME="Miniconda3-latest-Linux-x86_64.sh"
ENV CONDA_REQS="conda=4.3.21 conda-env=2.6.0 conda-build=2.1.15"

ADD . /tmp/bokeh
RUN chown -R travis /tmp/bokeh

USER travis

RUN cd /tmp/bokeh && scripts/ci/install:test

RUN rm -rf /tmp/bokeh
```

This can be built like so:

```
docker build -t bokeh-dev .
```

And then you may run the tests like so, for example:

```
docker run -it -v $PWD:/bokeh bokeh-dev /bin/sh -c "cd /bokeh && pytest -m 'unit'"
```

A few notes:
- Could we remove the line "`git fetch origin master`" from `scripts/ci/install`? We wouldn't want that line when working locally.
- Running the tests still yield a few failures. I haven't had a chance to look into those yet. The majority of tests pass though.
- I'm not too familiar with the current Travis setup or with all the things that get built/tested on the Travis platform. Are you looking to replicate everything, or just a subset (e.g. just be able to run the tests)?
- This current Dockerfile works with Python 3 only. I assume we'd want a separate container for each version of Python (where we'd set a different value for the `PYTHON` environment variable). Any thoughts about that?
- Could you detail what the encrypted environment variables are used for? I assume we won't be able to use those locally so would need a workaround for those.

jphalip:
Also, this `.dockerignore` is useful to speed up the docker build:

```
**/.git
**/*.pyc
**/*.pyo
bokehjs/node_modules/*
bokehjs/build/*
```

bryevdv:
* I don't think we can remove the `git fetch origin master` but perhaps it can be made contingent on  some environment variable that the docker file sets.

* Definitely not trying to replicate everything. If people can have a one-step setup that gets them the ability to build BokehJS and the docs, and run unit, js, and quality tests, I think that would be great. The "examples" image diff tests would be a nice bonus, and the selenium/saucelabs tests are definitely a reach. 

* I don't know much about how configurarable docker containers are. e.g whether one can be "paramterized" by the value of `PYTHON`, or whether it makes more sense to make multiple containers that each hard code a different value. I don't really have a strong opinion.

* the encrypted variables are mostly used for the automated release deployment, so they are not a concern. There's two env vars used for running selenium tests on saucelabs, but I think it's fine to punt on that. There's also a `GOOGLE_API_KEY` needed to build the docs. This can be set to a junk value (or people will have to supply their own key to see map plots)

bryevdv:
Also FYI I may move `CONDA_REQS` back into the script instead of "outside config" It seemed like a good idea at the time but has some drawbacks.

jphalip:
@bryevdv I've just pushed the updates to use Travis' containers. Would you mind giving it a try? The instructions are in `docker-tools/README.md`.

I believe there are still a few test failures but it might be easier for you to track down the issue. Let me know what you think!

bryevdv:
@jphalip sure, FYI I am traveling this week so it may be next week before I can take a look.

jphalip:
@bryevdv No worries, enjoy your vacation :)

bryevdv:
If only! Headed to PyData Seattle :)

bryevdv:
just an FYI I'm back in town my goal is to try this out/get it merged this week .

Dockerfile-py2
```diff
@@ -0,0 +1,16 @@
...
+ENV CONDA_REQS="conda=4.3.21 conda-env=2.6.0 conda-build=2.1.15"
+
+ADD . /tmp/bokeh
+RUN chown -R travis /tmp/bokeh
+
+USER travis
+
+RUN cd /tmp/bokeh && scripts/ci/install:test
+
+RUN rm -rf /tmp/bokeh
```
bryevdv: These are at the very top level? Our top-level dir is already cluttered with too much, I'm looking for ways to clean it up. I'd prefer a `docker` subdir that all of this could go under, is that reasonable?

jphalip: I had initially started out inside a subdirectory, however the issue is that Docker doesn't allow referencing objects in parent directories so I couldn't access `scripts/ci` or `conda.recipe` which are both necessary to build the container. So I resorted to moving the Dockerfiles to the root.

Could all of those directories be moved to a common parent directory, for examples named `tools` or such?

jphalip: See the [documentation](https://docs.docker.com/engine/reference/builder/#add):

"The <src> path must be inside the context of the build; you cannot COPY ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon."

bryevdv: OK .Maybe things could be moved but that's a nontrivial undertaking to coordinate, so probably not worth doing at this point. If it would not be too unusual for docker users  I'd prefer shorter and lowercase filenames like `docker-py2` and `docker-py3`

I'm also interested in centralizing conda req and miniconda filename so that that is not repeated in (at least) three places. Can docker files read this from other files? Or only env vars?

bryevdv:
@jphalip when I run one of the docker build commandsI get this error:

```
Removing intermediate container 1acb7222d834
Step 10/11 : RUN cd /tmp/bokeh && scripts/ci/install:test
 ---> Running in 9d8f0ece4639
+ [[ ! -z '' ]]
+ bash scripts/ci/install
+ git fetch origin master
fatal: Not a git repository (or any of the parent directories): .git
The command '/bin/sh -c cd /tmp/bokeh && scripts/ci/install:test' returned a non-zero code: 128
```

bryevdv:
I suppose that's the issue you mentioned about about `git fetch` I think the solution is to pass a command line arg (or set an env var) and have `install.sh` do or not do the git fetch accordingly

jphalip:
@bryevdv Oh yes that's the issue. That's why I had introduced the `NO_GIT_FETCH` env var in the Dockerfile, but it looks like I forgot to add the conditional test inside `scripts/ci/install` to my latest commit. I'll try to update the PR later today. In the meantime you may just temporarily comment out the `git fetch origin master` line if you'd still like to do some testing.

jphalip:
FYI, I've added the conditional statement so the docker build command should proceed without error now.

bryevdv:
@jphalip awesome, I think this is either done or close :) Did you have any input on this question:

> I'm also interested in centralizing conda req and miniconda filename so that that is not repeated in (at least) three places. Can docker files read this from other files? Or only env vars?

Also would renaming to shorter lowercase filenames e.g. `docker-py2` be weird ?

jphalip:
I think it's probably best to keep the `Dockerfile` naming convention as it's pretty standard.

Also, I was incorrect earlier: the build command must be run from the root, but the Dockerfiles don't need to be. So I've just moved them to the `docker-tools` directory.

Regarding the environment variables, I had tried to load them from a separate file but that didn't seem to work. I think the only way is to use the `ENV` clause as currently, or to pass them via the `-e` option.

bryevdv:
Thanks @jphalip !

PR #6379, opened 2017-06-04
Do not over-execute range slider render

### LINKED ISSUE #6123: RangeSlider Bug
This bug was introduced with 0.12.5 drop. 

capture
```
from bokeh.models.widgets import RangeSlider
output_file("range_slider.html")
range_slider = RangeSlider(start=0, end=10, range=(1,9), step=.1, title="Stuff")
show(widgetbox(range_slider))

```
Env:
Python 3.5
Bokeh 0.12.5
Windows

bryevdv:
fixes #6123

Not 100% about this solution but it seems like a "good-enough" solution for the immediate term. It makes things usable at all (and for probably most common use cases) until broader jq refactor can take place. 

This code:
```
from bokeh.io import curdoc
from bokeh.models import RangeSlider

slider = RangeSlider(start=0, end=10, range=(1,9), step=.1, title="Stuff")

def cb(attr, old, new):
    print(slider.range)

slider.on_change('range', cb)

curdoc().add_root(slider)
```

Behaves as expected:

screen shot 2017-06-04 at 11 03 25

### DIFF
--- a/bokehjs/src/coffee/models/widgets/range_slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/range_slider.coffee
@@ -14,7 +14,10 @@ export class RangeSliderView extends InputWidgetView
 
   initialize: (options) ->
     super(options)
-    @connect(@model.change, () -> @render())
+    @connect(@model.properties.start.change, () -> @_render())
+    @connect(@model.properties.end.change, () -> @_render())
+    @connect(@model.properties.step.change, () -> @_render())
+    @connect(@model.properties.orientation.change, () -> @_render())
     @$el.empty()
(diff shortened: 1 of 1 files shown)


bryevdv:
@mattpap I also plan to merge this in a few hours

mattpap:
@bryevdv, I don't follow this change, but feel free to merge if this helps. I will finalize PR #5774 for next release, so this will be no issue at some point.

bryevdv:
@mattpap my guess is that the widget defines its own `render` and our events are causing it to be called too often. In particular a change handler for `range` was problematic, maybe setting up a ping pong between our events and the sliders.

PR #6380, opened 2017-06-04
add a toggleable property to inspectors

### LINKED ISSUE #6064: remove the hover menu item, and keep the hover function working
Feature request.

I would like to remove the hover menu item, and keep the hover function working

bryevdv:
issues: fixes #6064

Until there is a better solution to consolidating multiple hover tools in toolbars, this feature at least allows users who have multiple hover tools the option of hiding some or all of the buttons in the toolbar.

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -90,7 +90,11 @@ class Inspection(Tool):
     ''' A base class for tools that perform "inspections", e.g. ``HoverTool``.
 
     '''
-    pass
+    toggleable = Bool(True, help="""
+    Whether an on/off toggle button should appear in the toolbar for this
+    inpection tool. If ``False``, the viewers of a plot will not be able to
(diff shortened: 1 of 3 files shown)


bryevdv:
@mattpap I will merge this in a few hours unless there is comment. I tested this with a modified `color_scatter.py` and it worked perfectly to remove any desired inspectors from the toolbar.

rajesh-iiith:
@bryevdv I am not able to hide the hover tool in the toolbar. I tried this setting toggleable = False. This attribute seems to be absent in 0.12.5.

bryevdv:
@rajesh-iiith this property was introduced in `0.12.6`, to use it you will have to upgrade.

rajesh-iiith:
@bryevdv Thanks. Its working as expected with 0.12.7.

PR #6381, opened 2017-06-04
Whisker annotation

### LINKED ISSUE #2352: Feature Request: Error Bar Plots

canavandl:
- [x] issues: fixes #2352 
- [x] tests added / passed


canavandl:
looks like (could perhaps use some better defaults):

screen shot 2017-06-04 at 8 17 23 pm

bryevdv:
Just an idea: What about adding a `TeeHead` for arrows, then implementing this in terms on an Arrow internally? Then users could swap out different end things if they wanted and all the drawing code ould be reused

canavandl:
> Just an idea: What about adding a TeeHead for arrows, then implementing this in terms on an Arrow internally? Then users could swap out different end things if they wanted and all the drawing code ould be reused

That's a good idea, i'll make the change.

bryevdv:
I guess `plus_end` and `minus_end` would allow users to control the sides independently, with the interpration of `base + thing` and `base - thing` to define which side is which in a way independent of dimensions? Alternatively just provide one head that is always used for both ends.

bryevdv:
I also wonder if this ought to default to a different level? I know we have an "annotations" level under the assumption that most annotations will want to be "on top" but I feel like maybe whiskers are usually underneath? not sure

bokehjs/src/coffee/models/annotations/whisker.coffee
```diff
@@ -0,0 +1,119 @@
...
+      _upper_vx = limit_scale.v_compute(@_upper)
+    else
+      _upper_vx = @_upper
+
+    if @model.base.units == "data"
+      _base_vx = base_scale.v_compute(@_base)
+    else
+      _base_vx = @_base
+
+    @_canvas = @plot_model.canvas
```
mattpap: This should be only a local variable.

bokehjs/src/coffee/models/annotations/whisker.coffee
```diff
@@ -0,0 +1,119 @@
...
+    else
+      _lower_vx = @_lower
+
+    if @model.upper.units == "data"
+      _upper_vx = limit_scale.v_compute(@_upper)
+    else
+      _upper_vx = @_upper
+
+    if @model.base.units == "data"
+      _base_vx = base_scale.v_compute(@_base)
```
mattpap: Where did `@_base` came from? The same thing happens in `Band` annotation.

mattpap: Never mind, `base` is a spec, so `_base` is magically created in `materialize_dataspecs()`.

canavandl:
Final:
screen shot 2017-06-05 at 2 46 50 pm

bryevdv:
I think the example would look a little better with an alpha on the circles, but I'll be making a separate PR tomorrow for a pass over the docs, I can play around with it then. Looks good will merge when green

bokeh/models/annotations.py
```diff
@@ -983,3 +983,57 @@ class Tooltip(Annotation):
...
+    upper_head = Instance('.models.arrow_heads.ArrowHead', default=_DEFAULT_TEE, help="""
+    Instance of ArrowHead.
+    """)
+
+    base = DistanceSpec(help="""
+    The orthogonal coordinates of the upper and lower values.
+    """)
+
+    dimension = Enum(Dimension, default='height', help="""
+    The direction of the band.
```
bryevdv: Will update this in follow on docs PR

PR #6382, opened 2017-06-05
attempt to properly redraw when gmaps change size

### LINKED ISSUE #6091: sizing_mode breaks GMap alignment
Hi!

I'm using bokeh 0.12.5rc3, Python 3.6.1, Chrome 57.0.2987.110 on Ubuntu 16.10. Using this script:

```

from bokeh.models.glyphs import Patch
from bokeh.models import (
     GMapPlot, Range1d, ColumnDataSource, GMapOptions, WheelZoomTool, PanTool
)

from bokeh.sampledata.us_states import data as states

API_KEY = "YOUR API KEY"

map_options = GMapOptions(lat=39.5501, lng=-105.7821, map_type="roadmap", zoom=6)

plot = GMapPlot(x_range=Range1d(), y_range=Range1d(), map_options=map_options,
                api_key=API_KEY, sizing_mode='scale_width', width=300, height=100)

source = ColumnDataSource(data=dict(x=states['CO']["lons"], y=states['CO']["lats"]))

patch = Patch(x="x", y="y", line_color="red", fill_color=None)
plot.add_glyph(source, patch)

doc = Document()
doc.add_root(plot)

plot.add_tools(WheelZoomTool(), PanTool())

filename = "maps.html"
    f.write(file_html(doc, INLINE, "Google Maps Example"))
view(filename)
```

At first sight there's no problem with the map:

image

but if i do anything that resizes the GMap (like resizing the window, or inspecting an element) the polygon get's misaligned:

image

if i return the gmap to its initial width the misalignment seems to dissapear, but just panning the map shows that this 're alignment' it's a kind of glitch:

image

Maybe this a corner case of, or it's related to #2964, but that was closed since there wasn't any code to reproduce the bug, so i'm opening this one instead. Sorry if this was already reported (i searched but didn't find any similar)

bryevdv:
issues: fixes #6091

As motioned in the issue this does not seem to be a 100% fix, but an improvement. Tested also that notebook usage was not affected.

### DIFF
--- a/bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
@@ -110,8 +110,13 @@ export class GMapPlotCanvasView extends PlotCanvasView
     # create the map with above options in div
     @map = new maps.Map(@canvas_view.map_el, map_options)
 
-    # update bokeh ranges whenever the map idles, which should be after any UI action
-    maps.event.addListenerOnce(@map, 'idle', @_set_bokeh_ranges)
+    # update bokeh ranges whenever the map idles, which should be after most UI action
+    maps.event.addListener(@map, 'idle',
+      @_set_bokeh_ranges
+      @render())
+
+    # also need an event when bounds change so that map resizes trigger renders too
+    maps.event.addListener(@map, 'bounds_changed', () => @_set_bokeh_ranges())
 
     # wire up listeners so that changes to properties are reflected
     @connect(@model.plot.properties.map_options.change, () => @_update_options())


bokehjs/src/coffee/models/plots/gmap_plot_canvas.coffee
```diff
@@ -110,8 +110,13 @@ export class GMapPlotCanvasView extends PlotCanvasView
     # create the map with above options in div
     @map = new maps.Map(@canvas_view.map_el, map_options)
 
-    # update bokeh ranges whenever the map idles, which should be after any UI action
-    maps.event.addListenerOnce(@map, 'idle', @_set_bokeh_ranges)
+    # update bokeh ranges whenever the map idles, which should be after most UI action
+    maps.event.addListener(@map, 'idle',
+      @_set_bokeh_ranges
+      @render())
```
mattpap: @bryevdv, ^^^ this doesn't make sense. I presume this should be:
```
maps.event.addListener @map, 'idle', () =>
  @_set_bokeh_ranges()
  @render()
```
Obviously the original compiles as it's perfectly valid CoffeeScript. I hate this language.

mattpap: Also, I think it would be better to `request_render()` instead, unless you already verified calling `render()` directly is necessary.

bryevdv: It does seem to perform better with a synchronous render but I can try again with a request to if it's acceptable when I push a change to make the callback syntax better. I'll submit a follow on PR in a short while.

Issue #6259, opened 2017-05-08
Adding new button widgets results in too-wide buttons


bryevdv:
ref: https://github.com/bokeh/bokeh/issues/5856

Hopefully this is something simple, like a re-compute not being triggered when it should 

ping @mattpap 

code:
```
import numpy as np

from bokeh.io import curdoc
from bokeh.layouts import row
from bokeh.models import ColumnDataSource
from bokeh.plotting import figure

# Set up data
x = np.linspace(0, 4*np.pi, 200)
y = np.sin(x)
source = ColumnDataSource(data=dict(x=x, y=y))

# Set up plot
plot = figure()

plot.line('x', 'y', source=source)

# Problem Code:
from bokeh.models.widgets import Button
from bokeh.models.layouts import VBox

button = Button(label="Add")
vbox = VBox(children=[button])

def create_button():
    button = Button(label=str(len(vbox.children)))
    vbox.children.append(button)

button.on_click(create_button)

curdoc().add_root(row(vbox, plot))
curdoc().title = "Sliders"
```

Results in:

<img width="980" alt="screen shot 2017-05-08 at 17 25 39" src="https://cloud.githubusercontent.com/assets/1078448/25828406/9907594c-3415-11e7-9f56-86a697fbbec3.png">

bryevdv:
Just confirming that this is not fixed by any work up to `0.12.6dev7`

mattpap:
This isn't a problem with layout per se, but with its API. Widget have to be wrapped in `WidgetBox` to make them appear properly (I'm not sure why this is needed). The following code works:
```
def create_button():
    from bokeh.models.layouts import WidgetBox
    button = Button(label=str(len(vbox.children)))
    vbox.children.append(WidgetBox(children=[button]))
```
The problem is that some API functions that work with layout do this automatically, but if you append to children like this, you have to do wrapping manually.

bryevdv:
Oh, another probably slightly better is to put all the buttons in one widget box

bryevdv:
fixed in #8085:

<img width="947" alt="screen shot 2019-01-30 at 21 20 34" src="https://user-images.githubusercontent.com/1078448/52032207-e844cb00-24d4-11e9-8cd0-624398545040.png">

PR #6384, opened 2017-06-05
Use a two pass bottom-up algorithm for computing layout

### LINKED ISSUE #4764: Issue with interactions between widgets and plots using bokeh server
When creating two plots, each one with select box acting on the data and try to replace the plot in the "on_change" callback, the change in one select box updates the corresponding plot and messes up the other (the plot disappears or some text appears behind, ...).

To reproduce the issue, simply run the code below as server and change the selection of the first select box. The first graph will update and the second will get affected. Then, changing the second select box fixes the display of the second plot and affects the first one. 

``` python
from bokeh import layouts
from bokeh.io import curdoc
from bokeh.sampledata import iris
from bokeh.models import ColumnDataSource, widgets


df = iris.flowers

species = df.species.unique().tolist()
COLORS = [
    "green", "red",  "blue", "cyan",
    "purple", "magenta", "orange", 'darkgray', 'navy', 'brown'
]

df['color'] = df['species'].apply(lambda x: COLORS[species.index(x)])
df.head()


def get_figure(min_length):
    min_length = int(min_length)
    p = figure(title="markers.py example")
    source = ColumnDataSource(df[df.sepal_length >= min_length])
    p.scatter(x='sepal_width', y='sepal_length', color='color', source=source)
    return p


def make_combo():
    w = widgets.Select(options=['2', '4', '6'], value='2')
    fig = get_figure(2)
    c = layouts.Column(w, fig)

    def update(name, old, new):
        c.children[-1] = get_figure(w.value)

    w.on_change('value', update)
    return c


layout_ = layouts.layout([make_combo()], [make_combo()])
curdoc().add_root(layout_)
```

Output of `bokeh info`:

``` shell
Python version      :  3.4.3 |Continuum Analytics, Inc.| (default, Dec  1 2015, 11:59:33) [MSC v.1600 32 bit (Intel)]
IPython version     :  4.0.0
Bokeh version       :  0.12.0
BokehJS static path :  c:\PyEnv\py34_32\20160511\lib\site-packages\bokeh\server\static
```

### LINKED ISSUE #4810: Trouble Swapping out layout contents when using server

### LINKED ISSUE #5131: Unexpected initial layout with DataTable and layout()
Dear all,

thanks for this marvelous library! I have written a small dashboard to follow some processes. The dashboard consists of three tables that I render with the layout() function. The problem is that the initial layout has the wrong width making the tables overlap each other.

initial_layout
After moving the browser's window (with the mouse I mean) the elements re-accommodate and they layout as I expected. 

final_layout
In order to reproduce the behavior I saved the code below and I run `bokeh serve test.py`

test.py file:

```
from bokeh.layouts import widgetbox, layout
from bokeh.models.widgets import DataTable, TableColumn, NumberFormatter
import pandas as pd

df = pd.DataFrame({'Host': ['foo', 'foo', 'bar', 'bar', 'foo', 'foo', 'bar', 'bar'],
                   'JobRuntime': [215, 351, 756, 654, 684, 654, 654, 658],
                   'Exitval': [0, 0, 0, 255, 0, 255, 0, 0]})
source = ColumnDataSource(df)
columns = [
        TableColumn(field="Host", title="Host"),
        TableColumn(field="JobRuntime", title="JobRuntime",
                    formatter=NumberFormatter(format='00:00:00')),
        TableColumn(field="Exitval", title="Exitval"),
    ]
data_table = DataTable(source=source, columns=columns, fit_columns=True,
                       width=400, height=400)

counts = df.pivot_table(columns='Host', index='Exitval', values='JobRuntime',
                        aggfunc=len, margins=True)
counts = counts.reset_index()
counts.columns = counts.columns.map(str)
columns2 = [TableColumn(field=val, title=val) for val in counts.columns]
counts = ColumnDataSource(counts)
data_table2 = DataTable(source=counts, columns=columns2, fit_columns=True,
                        width=800, height=200)

means = df.pivot_table(columns='Host', index='Exitval', values='JobRuntime',
                       margins=True)
means = means.reset_index()
means.columns = means.columns.map(str)
columns3 = [TableColumn(field='Exitval', title='ExitVal')]
columns3 += [TableColumn(field=val, title=val, formatter=NumberFormatter(format='00:00:00')) for val in means.columns[1:]]
means = ColumnDataSource(means)
data_table3 = DataTable(source=means, columns=columns3, fit_columns=True,
                        width=800, height=200)

curdoc().add_root(layout([[widgetbox(data_table), widgetbox(data_table2, data_table3)]], width=3000, sizing_mode='stretch_both'))

```

The output of `bokeh info` is
Python version      :  2.7.11 |Anaconda 4.0.0 (64-bit)| (default, Dec  6 2015, 18:08:32) 
IPython version     :  4.1.2
Bokeh version       :  0.12.2
BokehJS static path :  /home/andres/anaconda2/lib/python2.7/site-packages/bokeh/server/static

Thanks again

### LINKED ISSUE #5518: Add new child to existing column

### LINKED ISSUE #5879: Make "bokeh finished rendering heuristic" work with non-plot examples

### LINKED ISSUE #6213: Appending layout regression
@mattpap not sure if this is covered by other issues but I am making a new one specifically because it is a regression. This code works in `0.12.5` and `0.12.4` (plots are added below) but does not work in master (plots stack on top of each other)

```
import numpy as np

from bokeh.models.layouts import VBox

io_loop = IOLoop.current()
layout = VBox()


def modify_doc(doc):
    doc.add_root(layout)
    doc.add_periodic_callback(add_plot, 3000)


def add_plot():
    print("add plot")
    x = np.linspace(0, 4 * np.pi, 1000)
    y = np.sin(x)
    plot2 = figure(y_range=(-1.2, 1.2), plot_width=300, plot_height=300)
    plot2.line(x, y, line_width=2, legend="x")
    layout.children.append(plot2)

bokeh_app = Application(FunctionHandler(modify_doc))
server = Server({'/': bokeh_app}, io_loop=io_loop)
server.start()


if __name__ == '__main__':
    print('Opening Bokeh application on http://localhost:5006/')
    io_loop.add_callback(server.show, "/")
    io_loop.start()
```

### LINKED ISSUE #6287: Add migration note for removal of Document.resize()
Apparently this was advertised and needs a replacement. This is a follow up on PR #5976.

mattpap:
This PR switches from a chaotic signal-based layout algorithm to a bottom-up one. This is still a two pass approach, but at least it was narrowed down and doesn't introduce performance penalty due to excessive painting (painting is done only after second pass). This PR is a bit chaotic on its own, but for now I just wanted to make things work and naming convention, etc. I will leave for later. I tested this on server examples from 0.12.6 milestone and a few `plotting/file`. After running examples on travis ci I will be able to figure out how badly I broke things and how much further work is needed. Note that unit tests are broken right now due to code shuffling and changes in method names.  

fixes #5518 
fixes #6213 
fixes #4810
fixes #4764
fixes #5131
fixes #5879
fixes #6287
possibly more

### DIFF
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -100,6 +100,7 @@ class Document(object):
         self._all_models_freeze_count = 0
         self._all_models = dict()
         self._all_models_by_name = MultiValuedDict()
+        self._all_former_model_ids = set()
         self._callbacks = {}
         self._session_callbacks = {}
         self._session_context = None
@@ -355,7 +356,11 @@ class Document(object):
             if event_json['kind'] == 'ModelChanged':
                 patched_id = event_json['model']['id']
                 if patched_id not in self._all_models:
-                    raise RuntimeError("Cannot apply patch to %s which is not in the document" % (str(patched_id)))
+                    if patched_id not in self._all_former_model_ids:
+                        raise RuntimeError("Cannot apply patch to %s which is not in the document" % (str(patched_id)))
+                    else:
+                        logger.warn("Cannot apply patch to %s which is not in the document anymore" % (str(patched_id)))
+                        break
                 patched_obj = self._all_models[patched_id]
                 attr = event_json['attr']
                 value = event_json['new']
@@ -1095,6 +1100,7 @@ class Document(object):
             if m.name is not None:
                 recomputed_by_name.add_value(m.name, m)
         for d in to_detach:
+            self._all_former_model_ids.add(d._id)
             d._detach_document()
         for a in to_attach:
             a._attach_document(self)
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -572,9 +572,16 @@ def _wait_until_render_complete(driver):
     script = """
     // add private window prop to check that render is complete
     window._bokeh_render_complete = false;
-    window.addEventListener("bokeh:rendered", function() {
-        window._bokeh_render_complete = true;
-    });
+    function done() {
+      window._bokeh_render_complete = true;
+    }
+
+    var doc = window.Bokeh.documents[0];
+
+    if (doc.is_idle)
+      done();
+    else
+      doc.idle.connect(done);
     """
     driver.execute_script(script)
 
@@ -585,7 +592,7 @@ def _wait_until_render_complete(driver):
         WebDriverWait(driver, 5, poll_frequency=0.1).until(is_bokeh_render_complete)
     except TimeoutException:
         logger.warn("The webdriver raised a TimeoutException while waiting for \
-                     a 'bokeh:rendered' event to signify that the layout has rendered. \
+                     a 'bokeh:idle' event to signify that the layout has rendered. \
                      Something may have gone wrong.")
     finally:
         browser_logs = driver.get_log('browser')
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -330,6 +330,9 @@ class JSResources(BaseResources):
         if self.log_level is not None:
             raw.append('Bokeh.set_log_level("%s");' % self.log_level)
 
+        if self.dev:
+            raw.append('Bokeh.settings.dev = true')
+
         return raw
 
     def render_js(self):
--- a/bokeh/tests/test_io.py
+++ b/bokeh/tests/test_io.py
@@ -304,31 +304,29 @@ def test__crop_image():
 
 def test__get_screenshot_as_png():
     layout = Plot(x_range=Range1d(), y_range=Range1d(),
-                  plot_height=2, plot_width=2, toolbar_location=None,
+                  plot_height=20, plot_width=20, toolbar_location=None,
                   outline_line_color=None, background_fill_color=None,
                   border_fill_color=None)
 
     png = io._get_screenshot_as_png(layout)
-    assert png.size == (2, 2)
-    # a 2x2px image of transparent pixels
-    assert png.tobytes() == b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+    assert png.size == (20, 20)
+    # a 20x20px image of transparent pixels
+    assert png.tobytes() == ("\x00"*1600).encode()
 
 def test__get_svgs_no_svg_present():
     layout = Plot(x_range=Range1d(), y_range=Range1d(),
-              plot_height=2, plot_width=2, toolbar_location=None)
+              plot_height=20, plot_width=20, toolbar_location=None)
 
     svgs = io._get_svgs(layout)
     assert svgs == []
 
 def test__get_svgs_with_svg_present():
     layout = Plot(x_range=Range1d(), y_range=Range1d(),
-                  plot_height=2, plot_width=2, toolbar_location=None,
+                  plot_height=20, plot_width=20, toolbar_location=None,
                   outline_line_color=None, border_fill_color=None,
                   background_fill_color=None, output_backend="svg")
 
     svgs = io._get_svgs(layout)
     assert svgs[0] == ('<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" '
-                       'width="10" height="10" style="width: 10px; height: 10px;"><defs/><g><g/><g transform="scale(1,1) '
-                       'translate(0.5,0.5)"><rect fill="#FFFFFF" stroke="none" x="0" y="0" width="10" '
-                       'height="10"/><g/><g/><g/><g/></g><g transform="scale(1,1) translate(0.5,0.5)"><rect fill="#FFFFFF" '
-                       'stroke="none" x="0" y="0" width="10" height="10"/><g/><g/><g/><g/></g></g></svg>')
+                       'width="20" height="20" style="width: 20px; height: 20px;"><defs/><g><g/><g transform="scale(1,1) '
+                       'translate(0.5,0.5)"><rect fill="#FFFFFF" stroke="none" x="0" y="0" width="20" height="20"/><g/><g/><g/><g/></g></g></svg>')
--- a/bokeh/tests/test_resources.py
+++ b/bokeh/tests/test_resources.py
@@ -156,13 +156,13 @@ class TestResources(unittest.TestCase):
         self.assertEqual(r.mode, "server")
         self.assertEqual(r.dev, True)
 
-        self.assertEqual(len(r.js_raw), 1)
+        self.assertEqual(len(r.js_raw), 2)
         self.assertEqual(r.css_raw, [])
         self.assertEqual(r.messages, [])
 
         r = resources.Resources(mode="server-dev", root_url="http://foo/")
 
-        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW])
+        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW, "Bokeh.settings.dev = true"])
         self.assertEqual(r.css_raw, [])
         self.assertEqual(r.messages, [])
 
@@ -180,7 +180,7 @@ class TestResources(unittest.TestCase):
         self.assertEqual(r.mode, "relative")
         self.assertEqual(r.dev, True)
 
-        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW])
+        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW, "Bokeh.settings.dev = true"])
         self.assertEqual(r.css_raw, [])
         self.assertEqual(r.messages, [])
 
@@ -198,7 +198,7 @@ class TestResources(unittest.TestCase):
         self.assertEqual(r.mode, "absolute")
         self.assertEqual(r.dev, True)
 
-        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW])
+        self.assertEqual(r.js_raw, [DEFAULT_LOG_JS_RAW, "Bokeh.settings.dev = true"])
         self.assertEqual(r.css_raw, [])
         self.assertEqual(r.messages, [])
 
--- a/bokehjs/examples/burtin/burtin.ts
+++ b/bokehjs/examples/burtin/burtin.ts
@@ -7,6 +7,7 @@ namespace Burtin {
 
   console.log(`Bokeh ${Bokeh.version}`);
   Bokeh.set_log_level("info");
+  Bokeh.settings.dev = true
 
   type Gram = "negative" | "positive";
 
--- a/bokehjs/src/coffee/api/typings/bokeh.d.ts
+++ b/bokehjs/src/coffee/api/typings/bokeh.d.ts
@@ -25,6 +25,12 @@ declare namespace Bokeh {
   }
   var logger: Logger;
 
+  class Settings {
+    dev: boolean;
+  }
+
+  var settings: Settings;
+
   function sprintf(fmt: string, ...args: any[]): string;
 
   namespace embed {
--- a/bokehjs/src/coffee/core/dom_view.coffee
+++ b/bokehjs/src/coffee/core/dom_view.coffee
@@ -13,7 +13,17 @@ export class DOMView extends View
     DOM.removeElement(@el)
     super()
 
+  layout: () ->
+
   render: () ->
+
+  renderTo: (element, replace=false) -> # HTMLElement, boolean
+    if not replace
+      element.appendChild(@el)
+    else
+      DOM.replaceWith(element, @el)
+
+    @layout()
 
   @getters {
     solver:  () -> if @is_root then @_solver else @parent.solver
--- a/bokehjs/src/coffee/core/dom_view.coffee
+++ b/bokehjs/src/coffee/core/dom_view.coffee
@@ -7,6 +7,7 @@ export class DOMView extends View
 
   initialize: (options) ->
     super(options)
+    @_has_finished = false
     @el = @_createElement()
 
   remove: () ->
@@ -25,8 +26,14 @@ export class DOMView extends View
 
     @layout()
 
+  has_finished: () -> @_has_finished
+
+  notify_finished: () ->
+    @root.notify_finished()
+
   @getters {
     solver:  () -> if @is_root then @_solver else @parent.solver
+    is_idle: () -> @has_finished()
   }
 
   _createElement: () ->
--- a/bokehjs/src/coffee/core/layout/layout_canvas.coffee
+++ b/bokehjs/src/coffee/core/layout/layout_canvas.coffee
@@ -23,3 +23,18 @@ export class LayoutCanvas extends Model
     return editables
 
   get_constraints: () -> []
+
+  @getters {
+    layout_bbox: () ->
+      return {
+        top: @_top.value,
+        left: @_left.value,
+        width: @_width.value,
+        height: @_height.value,
+        right: @_right.value,
+        bottom: @_bottom.value,
+      }
+  }
+
+  dump_layout: () ->
+    console.log(this.toString(), @layout_bbox)
--- a/bokehjs/src/coffee/core/layout/side_panel.coffee
+++ b/bokehjs/src/coffee/core/layout/side_panel.coffee
@@ -143,7 +143,7 @@ _align_lookup_positive = {
 #         height or width. Extending to full height or width means it's easy to
 #         calculate mid-way for alignment.
 
-export update_constraints = (view) ->
+export update_panel_constraints = (view) ->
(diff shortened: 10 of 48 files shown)


mattpap:
Note that there are still PATCH-DOC errors, though fewer than before. Those are harmless. They appear in situations like when updating children in a loop. Suppose first plot is updated on the server side. This gets sent to bokehjs immediately and rendered. During rendering `inner_{width, height}` of unchanged plots are updated and sent to the server. However, at this point second and possibly more plots were change on the server side. Those plots aren't anymore in a document, so change events from `inner_{width,height}` have nowhere to go. This should be further investigated as a server/protocol bug.

bryevdv:
@mattpap from my testing this resolves alot of issues (except #6259 which still seems to make too-wide buttons) so I would like to try to make getting this in a priority. I would be willing to extend the freeze deadline until tomorrow. Priorities are:

* fixing up tests
* ensuring export functionality still works 
* testing as much as possible before merging 

@canavandl can help with some testing and especially with characterizing any export issues. What other kind of assistance would help see this finished?

canavandl:
From spot testing, it appears that the ``bokeh.io.export_svgs`` functionality still works. Previously for a transparent canvas without renderers the svg had two white, unstroked rects. This change actually makes the raw SVG make a little more sense by not having those rects (as I would expect from our ``doit`` machinery).

There's still an issue that ``bokeh.io.__wait_until_render_complete`` isn't capturing any "bokeh:rendered" event. Though that may be resolved by #6342 ?

Here's the patch for the test failure:

```
diff --git a/bokeh/tests/test_io.py b/bokeh/tests/test_io.py
index df5a608dd..570b48d0d 100644
--- a/bokeh/tests/test_io.py
+++ b/bokeh/tests/test_io.py
@@ -328,7 +328,4 @@ def test__get_svgs_with_svg_present():
 
     svgs = io._get_svgs(layout)
     assert svgs[0] == ('<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" '
-                       'width="10" height="10" style="width: 10px; height: 10px;"><defs/><g><g/><g transform="scale(1,1) '
-                       'translate(0.5,0.5)"><rect fill="#FFFFFF" stroke="none" x="0" y="0" width="10" '
-                       'height="10"/><g/><g/><g/><g/></g><g transform="scale(1,1) translate(0.5,0.5)"><rect fill="#FFFFFF" '
-                       'stroke="none" x="0" y="0" width="10" height="10"/><g/><g/><g/><g/></g></g></svg>')
+                       'width="10" height="10" style="width: 10px; height: 10px;"><defs/><g><g/></g></svg>')
```

mattpap:
> There's still an issue that bokeh.io.__wait_until_render_complete isn't capturing any "bokeh:rendered" event. Though that may be resolved by #6342 ?

Unfortunately not and currently all examples' tests timeout. I'm working on this.

bryevdv:
My feeling is that this is too disruptive to try to get in so late. I'll take a look at whatever is done in the morning, but reasonably I expect to cut an RC without this PR (i.e. so don't go overboard trying to finish it)

mattpap:
@bryevdv, yeah, there still some issue to resolve and I doubt I will be able to finish this today. Most notably I need to provide a way for change events to trigger full relayout when necessary. Previously this was handled by `layout_update` signal, but that doesn't exist anymore. Also tabs are broken and unit tests are a big headache to fix (mostly because of stubs).

bryevdv:
OK I will cut an RC this morning (need to try to fix select tool regression first)

mattpap:
Examples' tests look really good. There are only two real failures. Run time was significantly reduced and takes about 12 minutes, which is only 2 minutes slower than running examples' tests with `--no-js`.

bryevdv:
@mattpap please see the email I just sent, it seems some JS dependency just updated and broke our build, fixing this is priority 1

mattpap:
@bryevdv, I added a trivial solution to PATCH-DOC errors in f49219d.

bkueng:
I can also confirm that this fixes #6407.
However `bokeh.io.export_png` fails with:
```
ERROR:bokeh.application.application:Error running application handler <bokeh.application.handlers.directory.DirectoryHandler object at 0x7fda33f473c8>: Message: {"errorMessage":"Can't find variable: Bokeh","request":{"headers":{"Accept":"application/json","Accept-Encoding":"identity","Connection":"close","Content-Length":"386","Content-Type":"application/json;charset=UTF-8","Host":"127.0.0.1:55477","User-Agent":"Python http auth"},"httpVersion":"1.1","method":"POST","post":"{\"args\": [], \"script\": \"\\n    // add private window prop to check that render is complete\\n    window._bokeh_render_complete = false;\\n    function done() {\\n      window._bokeh_render_complete = true;\\n    }\\n\\n    var doc = Bokeh.documents[0];\\n\\n    if (doc.is_idle)\\n      done();\\n    else\\n      doc.idle.connect(done);\\n    \", \"sessionId\": \"68d44790-4bac-11e7-ab6c-8d0af3a72563\"}","url":"/execute","urlParsed":{"anchor":"","query":"","file":"execute","directory":"/","path":"/execute","relative":"/execute","port":"","host":"","password":"","user":"","userInfo":"","authority":"","protocol":"","source":"/execute","queryKey":{},"chunks":["execute"]},"urlOriginal":"/session/68d44790-4bac-11e7-ab6c-8d0af3a72563/execute"}}
Screenshot: available via screen
```
This used to work for me on a previous state on master, when it was still called `bokeh.io.export`. Not sure if this PR is the cause.

mattpap:
@bkueng, thanks for pointing this out. This is definitively introduced by this PR. @canavandl, I'm not sure why `Bokeh` is not available. Maybe I should use `window.Bokeh`?

mattpap:
@bryevdv, current status is as follows:

- unit tests: `get_svgs_with_svg_present` is broken and I don't know why. I suspect idle algorithm may be at fault. @canavandl must look into this.
- examples: `plotting/file/categorical` has slightly offset horizontal axes. I think this example have been broken for a while, but now is a bit more broken.
- integration: two tests fail. The first is the one with too big canvas. However, canvas being too big is a problem I tracked to at least 0.12.5 (I expect it to be a lot earlier). I'm not sure why now the test fail, especially that plots generated for 0.12.5 and this PR are exactly the same. The other failure I didn't have chance to investigate yet, but it looks like a flaky test.

bryevdv:
@mattpap this is looking good I ran basically all the app examples under `app` and `server` dirs, as well as many static layout examples. The only issues I've found:

* `sizing_mode.py` the "both" modes cover up the dropdown (not a big deal for now)
* `responsive.py` starts of size 0x0 but fixes after a resize

mattpap:
>  the "both" modes cover up the dropdown

Yes and this is a preexisting issue. I think we never really tested responsive modes like this.

mattpap:
> responsive.py starts of size 0x0 but fixes after a resize

Fixed.

mattpap:
> `plotting/file/categorical` has slightly offset horizontal axes.

I experimented with all sizing modes and this example, and currently `scale_width` exhibits this behavior. However, resizing window restores previous bad behavior. Also, `scale_both` is completely broken at last in this case. `scale_height` and `strech_both` work correct. I don't mention `fixed`, because that's the most tested mode and that must work.

mattpap:
> `scale_both` is completely broken

Actually this is a feature, because `scale_both` is only implemented for `Plot` and categorical example uses `Row(Plot, Plot)`, so this mode can't work. It actually should raise an exception that it's not supported or we should implement it in `LayoutDOM` (I'm not sure why this case is omitted).

mattpap:
> The other failure I didn't have chance to investigate yet, but it looks like a flaky test.

This was resolved by a fix to `scale_both` mode.

bryevdv:
So it looks like there are only two test failures at this point? 

* unit test `test__get_svgs_with_svg_present`
* selenium test `test_scale_width_plot_starts_at_correct_size`

Is that correct? 

@canavandl are you available to help with the former today?

mattpap:
> selenium test test_scale_width_plot_starts_at_correct_size

This is is the too big canvas issue. As commented before, generated plots are the same for 0.12.5 and this PR, and I don't understand why this test started to fail now. This issue is reproducible in many examples in both versions. What's funny that there are very similar examples where one has canvas of correct size and the other of wrong size. I'm inclined to XFail this test for now and investigate this issue properly in next cycle.

mattpap:
There is also `plotting/file/categorical` that I'm still investigating.

bryevdv:
That might be an option, it's also possible @canavandl might have some useful insight on that one as well, worth a quick look before xfailing it

mattpap:
> (...) before xfailing it

That's why I'm leaving failing. I hoped that maybe some other fixed I did may resolve this.

canavandl:
current the ``bokeh.io.export_svg`` functionality is failing due to an "unsatisfiable constraint" error which causes nothing to render. This is causing the svg tests failure (you can notice that the test svg output is 500x500 when it should be 2x2). ( @mattpap )

mattpap:
@canavandl, I don't see any error like this on travis ci. Can you provide code for me to reproduce this (or just how to debug selenium)?

canavandl:
It also appears that ``bokeh.io.export_png`` isn't working for the same reason.

The error is ``type: WARNING`` in PhantomJS while it's ``SEVERE`` (and thus captured) in Firefox.

For debugging selenium, I generally change the driver to Firefox (it requires esr 45.8, unless you want to mess w/ the geckodriver stuff) so that I get a browser window.

Here's a patch to swap out the driver:
```
diff --git a/bokeh/io.py b/bokeh/io.py
index 8b162bffd..28ebbd1d8 100644
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -620,7 +620,7 @@ def _get_screenshot_as_png(obj):
     html_path = tempfile.NamedTemporaryFile(suffix=".html").name
     save(obj, filename=html_path, resources=INLINE, title="")
 
-    driver = webdriver.PhantomJS()
+    driver = webdriver.Firefox()
     driver.get("file:///" + html_path)
 
     ## resize for PhantomJS compat
@@ -682,7 +682,7 @@ def _get_svgs(obj):
     html_path = tempfile.NamedTemporaryFile(suffix=".html").name
     save(obj, filename=html_path, resources=INLINE, title="")
 
-    driver = webdriver.PhantomJS()
+    driver = webdriver.Firefox()
     driver.get("file:///" + html_path)
 
     _wait_until_render_complete(driver)
```

canavandl:
@mattpap it appears the issue is related to attempting to solve a 2x2 plot size. It's sort of a contrived test case, so i'm willing to change the tests.

mattpap:
> it appears the issue is related to attempting to solve a 2x2 plot size.

Yes, there used to be a safety check for 50x50 if I recall correctly. I will debug this outside selenium. I'm not sure if this will be easily solvable. I see you set `toolbar_location=None`, so you did enough to help this work.

Would it be possible to propagate JS errors to py.test? Currently the output doesn't give me much information about what's really going on. Also, this may be related to #6422.

canavandl:
> Would it be possible to propagate JS errors to py.test? Currently the output doesn't give me much information about what's really going on. Also, this may be related to #6422.

Yes, I'll do that now (I should have done it previously).

mattpap:
37 px is the magic number (at last for `plotting/file/color_scatter` without toolbar). The reason it used to work is because there was a safety check in layout which prevented layout from proceeding, so really it didn't work, but at least there were no errors.

canavandl:
>> Would it be possible to propagate JS errors to py.test? Currently the output doesn't give me much information about what's really going on. Also, this may be related to #6422.

>Yes, I'll do that now (I should have done it previously).

I'm actually going to open an issue to resolve this in another PRs. It's slightly more involved than I initially though (but very doable).

mattpap:
@canavandl, just FYI, in case you need to push more commits, I will let the tests run and then rebase this branch, because you've pushed a commit I removed earlier (8d0c7ec).

canavandl:
> @canavandl, just FYI, in case you need to push more commits, I will let the tests run and then rebase this branch, because you've pushed a commit I removed earlier (8d0c7ec).

I'm sorry about that. I don't need to push any more commits (the tests *should* pass now).

canavandl:
My only question about this PR: Is the migration note about deprecating the ``document.resize`` method and suggested work-around still accurate?

mattpap:
> Is the migration note about deprecating the `document.resize` method and suggested work-around still accurate?

Yes. It wasn't actually, but I restored `LayoutDOM.resize()`.

mattpap:
OK. Unit tests work. I xfailed one integration tests, so new it should be all green.

canavandl:
All green. We good to merge @mattpap ?

mattpap:
Currently we have the following image differences:

- `burtin/burtin`: toolbar is slightly off vertically (barely noticable, but still)
- `models/file/glyphs`: dimensions differ, which is a common thing with this example, most likely due to tabs (which are don't work very well)
- `plotting/file/categorical`: I've been writing about this earlier. I narrowed this down, but still not sure how to fix this.
- `howto/layouts/words_and_plots`: this is the same problem as above. Resize restores proper size in both cases. I think the problem is the order of layout or applying constraints. I tried earlier a top-down approach and it fixed `categorical` (I'm not sure about this) one, but broke a lot of other things.

mattpap:
Regarding idle algorithm:

- I finally add (compared to original PR) support for `TileRenderer`, however that doesn't seem to work reliably yet (there are some timeouts from time to time)
- external resources are not considered at all right now, so there may be spurious image differences.

mattpap:
@bryevdv, I figured this out. We need three passes. Two to compute layout and third to update CSS, but as rendering (DOM) and layouting are done in the same method (`render()`), it appears like we need three layout passes. Later I will need to clearly separate rendering and layouting (I already did this for painting).

bryevdv:
@mattpap what is your feeling about possible performance issues?

I restarted the one failed test in case it was spurious

mattpap:
> what is your feeling about possible performance issues?

@bryevdv, current situation is suboptimal, but performance penalty, if any, should be small. Although I don't have hard numbers at this point, layouting and DOM rendering is cheap, compared to painting, and painting is done the exact number of times it must. Currently, even after adding the third pass, examples tests run several minutes faster than before this PR.

bryevdv:
🎊

mattpap:
> What's funny that there are very similar examples where one has canvas of correct size and the other of wrong size.

Code for making toolbar sticky is at fault here. If there is no toolbar or toolbar is non-sticky, then canvas is sized properly (e.g. burtin examples). This isn't a big problem, because frame is still correctly sized, but nevertheless this needs to be resolved at some point.

PR #6390, opened 2017-06-06
handle renderers that return null hit test results

### LINKED ISSUE #6388: Selection tool regression
Using bokeh 0.12.6dev7 the selection tools appear to be broken in certain cases. A minimal example that reproduces the issue uses a set of ``points`` and ``multi_line`` glyphs:

```

# output to static HTML file
output_notebook()

p = figure(plot_width=400, plot_height=400, tools=['box_select'])

p.multi_line([[1, 3, 2], [3, 4, 6, 6]], [[2, 1, 4], [4, 7, 8, 5]],
             color=["firebrick", "navy"], alpha=[0.8, 0.3], line_width=4)

# show the results
show(p)
```

When trying to apply a box_select in this example the console issues the following warning:

```
Uncaught TypeError: Cannot read property 'update_through_union' of null
    at e.r.SelectionManager.e.select (bokeh-0.12.6dev7.min.js:2)
```

And the box_select is never applied. I suspect this is probably related to https://github.com/bokeh/bokeh/issues/6202 but haven't had the time to investigate further. I can confirm it happens in both 0.12.6dev6 and 0.12.6.dev7.

bryevdv:
- [x] issues: fixes #6388
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/core/selection_manager.coffee
+++ b/bokehjs/src/coffee/core/selection_manager.coffee
@@ -24,6 +24,12 @@ export class SelectionManager extends HasProps
 
     indices_renderers = (r.hit_test(geometry) for r in renderer_views)
 
+    # renderers that don't support hit testing a particular kind of geometry
+    # will return null, need to filter those out and see if anything is left
+    indices_renderers = (i for i in indices_renderers when i != null)
+    if indices_renderers.length == 0
+      return false
+
     if indices_renderers?
       indices = indices_renderers[0]
       for indices_other in indices_renderers
--- a/bokehjs/test/core/selection_manager.coffee
+++ b/bokehjs/test/core/selection_manager.coffee
@@ -4,6 +4,7 @@ utils = require "../utils"
 {SelectionManager} = utils.require "core/selection_manager"
 
 SomeMarker = utils.require("models/markers/index").CircleX
+NullMarker = utils.require("models/glyphs/index").MultiLine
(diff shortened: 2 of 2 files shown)


bryevdv:
@philippjfr can you test with this PR branch immediately? I want to try to add a quick test but if you can test locally in the cases that showed the problem I can cut an RC as soon as the test is merged.

bryevdv:
ping @clairetang6 FYI this was complicated by the fact that `hit_test` returns `null` when a given geometry is not supported (my fault). It's probably better to change it sometime to have it always return a hit test restult (just an empty one)

philippjfr:
Can confirm the issue I reported is fixed with this PR applied.

bryevdv:
Great, will merge when green then cut `RC1`

bokehjs/src/coffee/core/selection_manager.coffee
```diff
@@ -24,6 +24,12 @@ export class SelectionManager extends HasProps
 
     indices_renderers = (r.hit_test(geometry) for r in renderer_views)
 
+    # renderers that don't support hit testing a particular kind of geometry
+    # will return null, need to filter those out and see if anything is left
+    indices_renderers = (i for i in indices_renderers when i != null)
+    if indices_renderers.length == 0
+      return false
+
     if indices_renderers?
```
mattpap: @bryevdv, btw. this looks unnecessary.

PR #6394, opened 2017-06-06
fix property match checking when there are dicts of arrays

### LINKED ISSUE #6393: Modifying datasources broken
In bokeh master and presumably any rc/dev releases cut in the last few days, modifying a CDS is broken completely. This is because https://github.com/bokeh/bokeh/pull/6374 introduced a change in the way bokeh properties check whether an old and new value has changed. The approach there guards against checks on arrays, but since the ``.data`` property on a CDS is not itself an array it ends up doing a simple equality between the dictionary of arrays, which raises this error:

>ValueError�: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

This occurs for any change to a ``ColumnDataSource.data``, which means that any server or notebook examples that modify a CDS in python are currently broken.

bryevdv:
- [x] issues: fixes #6393
- [x] tests added / passed

@philippjfr @mattpap please test/have a look at this

### DIFF
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -178,7 +178,18 @@ class Property(PropertyDescriptorFactory):
         if isinstance(new, np.ndarray) or isinstance(old, np.ndarray):
             return np.array_equal(new, old)
 
-        return new == old
+        # this handles the special but common case where there is a dict with numpy
+        # arrays as values (e.g. the .data property of a ColumnDataSource)
+        if isinstance(new, dict) and isinstance(old, dict):
+            return list(new.keys()) == list(old.keys()) and \
+                all(self.matches(*args) for args in zip(new.values(), old.values()))
+
+        try:
+            return new == old
+
+        # if the comparison fails for some reason, just punt and return no-match
+        except ValueError:
+            return False
 
     def from_json(self, json, models=None):
         ''' Convert from JSON-compatible values into a value for this property.
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -181,8 +181,9 @@ class Property(PropertyDescriptorFactory):
         # this handles the special but common case where there is a dict with numpy
         # arrays as values (e.g. the .data property of a ColumnDataSource)
         if isinstance(new, dict) and isinstance(old, dict):
-            return list(new.keys()) == list(old.keys()) and \
-                all(self.matches(*args) for args in zip(new.values(), old.values()))
+            if set(new.keys()) != set(old.keys()):
+                return False
+            return all(self.matches(new[k], old[k]) for k in new)
 
         try:
             return new == old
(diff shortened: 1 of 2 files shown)


bokeh/core/property/tests/test_bases.py
```diff
@@ -65,3 +73,35 @@ def test_property_matches_incompatible_arrays(capsys):
...
+def test_property_matches_dicts_with_array_values(capsys):
+    p = pb.Property()
+    d1 = dict(foo=np.arange(10))
+    d2 = dict(foo=np.arange(10))
+
+    assert p.matches(d1, d1) is True
+    assert p.matches(d1, d2) is True
+
+    # XXX not sure if this is preferable to have match, or not
+    assert p.matches(d1, dict(foo=list(range(10)))) is True
```
mattpap: numpy agrees with this:
```
In [1]: import numpy as np

In [2]: a = np.arange(10)

In [3]: b = list(range(10))

In [4]: np.array_equal(a, b)
Out[4]: True
```

bokeh/core/property/bases.py
```diff
@@ -178,7 +178,18 @@ def matches(self, new, old):
         if isinstance(new, np.ndarray) or isinstance(old, np.ndarray):
             return np.array_equal(new, old)
 
-        return new == old
+        # this handles the special but common case where there is a dict with numpy
+        # arrays as values (e.g. the .data property of a ColumnDataSource)
+        if isinstance(new, dict) and isinstance(old, dict):
+            return list(new.keys()) == list(old.keys()) and \
+                all(self.matches(*args) for args in zip(new.values(), old.values()))
```
philippjfr: Is this safe? The ordering of dicts isn't guaranteed right, so the order of keys and values ``old`` and ``new`` isn't necessarily the same (although they probably are in some python versions and implementations).

bokeh/core/property/tests/test_bases.py
bryevdv: Right, that's the reason it returns `True` but I'm not sure that's what we'd actually want given that we do treat arrays and lists differently in some places

bokeh/core/property/bases.py
bryevdv: You're right, I had in mind that that `keys` and `values` return things in the same relative order. Let me think on it

bokeh/core/property/tests/test_bases.py
bryevdv: (I think it's probably fine, but just wanted to document the question in case something does arise)

bryevdv:
@philippjfr I have updated the test I think it should be better now

philippjfr:
Looks good and works for me.

bokeh/core/property/bases.py
mattpap: The order is deterministic and sufficient here.

PR #6397, opened 2017-06-07
More explicit match cases

bryevdv:
This PR adds handling of Pandas types explicitly, as well as adds back overall exception fallback in any case. 

Without this, `export_csv` example would work but show errors like

```
2017-06-06 20:36:45,200 error handling message Message 'PATCH-DOC' (revision 1): ValueError('The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().',)
```

With this, dicts of arrays, series, and indices are compared correctly (tests added)

ping @mattpap

### DIFF
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -18,10 +18,13 @@ import types
 from six import string_types
 import numpy as np
 
+from ...util.dependencies import import_optional
 from ...util.string import nice_join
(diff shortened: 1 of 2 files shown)


bryevdv:
I tested all the example apps with `0.12.6rc2` patched with this. I am going to go ahead and merge and cut a new rc. They all worked except `pivot` which seems to have its own issues.

PR #6403, opened 2017-06-07
Bryanv/6398 webgl canvas regression

### LINKED ISSUE #6398: All webgl examples fail
All webgl examples fail with the Bokeh error in the browser:
```
undefined is not an object (evaluating 't.setLineDash')
```

bryevdv:
- [x] issues: fixes #6398
- [ ] tests added / passed

This PR fixes the missing canvas error.  On safari. some scatter examples still fail with 
```
RuntimeError: RuntimeError:OpenGL got errors (after draw): 1282
```

I believe this was an existing problem on safari, but if there are ideas to fix it I would like to fix it in this PR. 

Would also like to add tests for the canvas creation.


bryevdv:
All right, I am not sure offhand how to add a test for this so I will punt for now.

PR #6404, opened 2017-06-07
extract data source correctly on line selections

### LINKED ISSUE #6400: line_select.py example broken
JS console reports:
```
TypeError: undefined is not an object (evaluating 'cb_obj.selected["0d"]')
```

bryevdv:
issues: fixes #6400

The loop was iterating over an object indexed by data source id's and passing the id as the data source to the callback. This PR correctly grabs the actual data source.

screen shot 2017-06-07 at 00 31 48

If there is any good test that is reasonable to add I'd appreciate suggestions.

### DIFF
--- a/bokehjs/src/coffee/models/tools/gestures/tap_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/gestures/tap_tool.coffee
@@ -28,9 +28,9 @@ export class TapToolView extends SelectToolView
 
       renderers_by_source = @model._computed_renderers_by_data_source()
 
-      for ds_id, renderers of renderers_by_source
+      for _, renderers of renderers_by_source
         ds = renderers[0].data_source
-        sm = renderers[0].data_source.selection_manager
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/models/tools/gestures/tap_tool.coffee
```diff
@@ -28,7 +28,8 @@ export class TapToolView extends SelectToolView
 
       renderers_by_source = @model._computed_renderers_by_data_source()
 
-      for ds, renderers of renderers_by_source
+      for ds_id, renderers of renderers_by_source
+        ds = renderers[0].data_source
         sm = renderers[0].data_source.selection_manager
```
mattpap: `ds_id` isn't used so I would use `_` instead. Also, `sm = ds.selection_manager`.

bryevdv:
@mattpap I have made the requested changes

PR #6410, opened 2017-06-07
make buttons respect disabled property

### LINKED ISSUE #6402: disabled button raises button click protocol event
In `models/server/buttons.py` clicking the disabled button prints "clicked" at the console.

bryevdv:
- [x] issues: fixes #6402
- [x] tests added / passed

ping @bokeh-dev @jlstevens This PR makes it so that buttons with `enabled=False` no longer generate click events. 

Perhaps this merits a migration note?

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -279,7 +279,7 @@ class Model(with_metaclass(MetaModel, HasProps, PropertyCallbackManager, EventCa
     help="""A mapping of event names to lists of CustomJS callbacks.
 
     Typically, rather then modifying this property directly, callbacks should be
-    added using the ``Model.js_on_event`` method:)
+    added using the ``Model.js_on_event`` method:
 
     .. code:: python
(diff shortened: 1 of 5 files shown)


mattpap:
This seems unnecessary, because we set `disabled` HTML element attribute, and disabled buttons should not receive click events.

mattpap:
The underlying problem is that `click` event is attached to `@el`, not the actual `<button>` element (which is a child of `@el`. This way `click` doesn't obey `disabled` attribute. I also observed that, for whatever reason, `click` event is received multiple times 2~3 per click when attached to `@el`.

bryevdv:
OK, is there a simple way to attach the event to the actual button?

mattpap:
@bryevdv, I submitted my take on this issue in PR #6412.

bryevdv:
superseded by #6412

PR #6415, opened 2017-06-08
Pass HTTP arguments to autoload_server (#5992)

### LINKED ISSUE #5992: Pass HTTP request arguments to autoload_server and pull_session
#4858 makes it possible to pass HTTP requests arguments to app documents. However, the functions `autoload_server` and `pull_session` do not allow for sending request parameters, because they already add a query string to the URL.

For instance, `autoload_server` adds `/autoload.js?bokeh-autoload-element=...&bokeh-session-id=...` to the end of the URL. If I add the query string `?N=10` to the end of my `app_path`, then the URL will result in `?N=10/autoload.js?bokeh-autoload-element=...&bokeh-session-id=...` which is not a valid URL.

It would be great if there was a way to pass request arguments to these functions so that they automatically add it to the URL in a correct way.

ThmsLa:
Needed that feature for an app so I figured I'd do a PR. 
It's my first one though, so all suggestions are welcome :)

- [x] issues: fixes #5992
- [x] tests added / passed

### DIFF
--- a/examples/embed/arguments/flask_server.py
+++ b/examples/embed/arguments/flask_server.py
@@ -1,0 +1,65 @@
+'''This example demonstrates embedding an autoloaded Bokeh server
+into a simple Flask application, and passing arguments to Bokeh.
+
+To view the example, run:
+
+    python flask_server.py
+
+in this directory, and navigate to:
+
+    http://localhost:5000
+
+'''
+import atexit
+import subprocess
+
+from flask import render_template_string, Flask
+from bokeh.embed import server_document
+
+home_html = """
+<!DOCTYPE html>
+<html lang="en">
+  <body>
+    <div class="bk-root">
+      <h1><a href="/batch/1"> Batch 1 (cos)</a></h1>
+      <h1><a href="/batch/2"> Batch 2 (sin)</a></h1>
+      <h1><a href="/batch/3"> Batch 3 (tan)</a></h1>
+    </div>
+  </body>
+</html>
+"""
+
+app_html = """
+<!DOCTYPE html>
+<html lang="en">
+  <body>
+    <div>
+        <h2><a href="/batch/1">Batch 1 (cos)</a> - <a href="/batch/2">Batch 2 (sin)</a> - <a href="/batch/3">Batch 3 (tan)</a></h2>
+    </div>
+    <div class="bk-root">
+      {{ bokeh_script|safe }}
+    </div>
+  </body>
+</html>
+"""
+
+app = Flask(__name__)
+
+bokeh_process = subprocess.Popen(
+    ['bokeh', 'serve', '--allow-websocket-origin=localhost:5000', 'bokeh_server.py'], stdout=subprocess.PIPE)
+
+@atexit.register
+def kill_server():
+    bokeh_process.kill()
+
+@app.route('/')
+def home():
+    return render_template_string(home_html)
+
+@app.route('/batch/<int:batchid>')
+def visualization(batchid):
+    bokeh_script = server_document(url='http://localhost:5006/bokeh_server', arguments=dict(batchid=batchid))
+    return render_template_string(app_html, bokeh_script=bokeh_script)
+
+if __name__ == '__main__':
+    app.run(debug=True)
--- a/sphinx/source/docs/user_guide/embed.rst
+++ b/sphinx/source/docs/user_guide/embed.rst
@@ -368,6 +368,30 @@ will look something like this:
     data-bokeh-doc-id=""
     ></script>
 
+You can also pass arguments to your Bokeh server by passing them in a dictionary to ``arguments``.
+The following illustrates how to pass and retrieve arguments.
+
+.. code-block:: python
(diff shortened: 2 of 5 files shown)


sphinx/source/docs/user_guide/embed.rst
```diff
@@ -367,6 +367,30 @@ will look something like this:
...
+
+.. code-block:: python
+
+    # Bokeh server
+    # request.arguments is a dict that maps argument names to lists of strings,
+    args = curdoc().session_context.request.arguments
+
+    try:
+        foo = int(args.get('foo_id')[0])
+    except:
```
mattpap: This is a bad practice, because this catchs also system events, like `SystemExit` or `KeyboardInterrupt`. At minimum `exception Exception` should be used and preferably a most specific exception possible, so `except ValueError` (I presume). Even better would be to limit the code in `try` branch to minimum, so that you don't hide possible bugs, so:
```
foo = args.get('foo_id')[0]
try:
    foo = int(foo)
except ValueError:
    foo = 1
```

mattpap: `examples/embed` also has to be updated.

examples/embed/arguments/flask_server.py
```diff
@@ -0,0 +1,65 @@
...
+def home():
+    return render_template_string(home_html)
+
+@app.route('/batch/<int:batchid>')
+def visualization(batchid):
+    bokeh_script = autoload_server(None, url='http://localhost:5006/bokeh_server', arguments=dict(batchid=batchid))
+    return render_template_string(app_html, bokeh_script=bokeh_script)
+
+if __name__ == '__main__':
+    app.run(debug=True)
```
mattpap: Code quality tests will fail due to lack of trailing newline.

sphinx/source/docs/user_guide/embed.rst
ThmsLa: Thanks for the input! I'll update that.

I don't know if it should be marked as an issue, but the [official doc](http://bokeh.pydata.org/en/latest/docs/user_guide/server.html#accessing-the-http-request) uses this piece of code too.

ThmsLa:
Seems like Travis is failing on test_no_tornado_common, which hasn't even been modified... How to fix that?

bryevdv:
@ThmsLa thanks for the PR, looks like one test is failing:
```
____________________________ test_no_tornado_common ____________________________
    def test_no_tornado_common():
        code = "import sys; %s; sys.exit(1 if any(('bokeh.client' in x or 'bokeh.server' in x) for x in sys.modules.keys()) else 0)"
        proc = subprocess.Popen(["python", "-c", code % ";".join(basic_imports)],stdout=subprocess.PIPE)
        proc.wait()
        if proc.returncode != 0:
>           assert False
E           assert False
```
Although Tornado is technically listed as a dependency of Bokeh, it's only needed if you are using the server. We strive to keep all "basic" modules from importing anthing client/server related at the top level. My guess is that you can just move your `bokeh.client` import into `autoload_server`.

HOWEVER, I think this should wait a bit, and be based of https://github.com/bokeh/bokeh/pull/6363 That PR is splitting `autoload_server` into two simpler functions (and will deprecate `autoload_server`) We plan to merge that soon after `0.12.6` release so then we could merge your PR as a follow and cut a dev build so its available for you to use. What do you think?

ThmsLa:
@bryevdv Just updated the branch, let's hope the tests pass this time.

Regarding #6363, I think it's a great idea! I'll update this PR as soon as it's merged :)

bryevdv:
@ThmsLa the other PR is merged, so as soon as you update this one to resolve the conflicts we can get it merged as well, thanks!

bryevdv:
@ThmsLa FYI the one CI error is a linter error, we have a fairly permissive line length policy, but not infinite :) 

```
    @pytest.mark.quality
    def test_flake8():
        chdir(TOP_PATH)
    
        out, err = proc.communicate()
    
E         ./bokeh/embed.py:668:166: E501 line too long (174 > 165 characters)
E         ./build/lib/bokeh/embed.py:668:166: E501 line too long (174 > 165 characters)
E         
E       assert 1 == 0
E        +  where 1 = <subprocess.Popen object at 0x7f6ab1e3a9e8>.returncode
```

bryevdv:
Also I would ask that the docstrings for the public functions also have a description for `arguments` added to them (there does not seem to be one currently unless I am missing it somehow)

Otherwise looks good to merge when green with those changes!

bryevdv:
Also I would ask that the docstrings for the public functions also have a description for `arguments` added to them (there does not seem to be one currently unless I am missing it somehow)

Otherwise looks good to merge when green with those changes!

ThmsLa:
@bryevdv Sorry, did, the last push in a rush. Everything should be fine now!

bryevdv:
Thanks @ThmsLa will merge when green

bryevdv:
have restarted one test, probably a spurious failure

PR #6418, opened 2017-06-08
fix nearest mode for hover on multi-lines

### LINKED ISSUE #6416: HoverTool broken with multi_line plots in 0.12.5
Hi,

I noticed a regression in 0.12.5 from 0.12.4 using the hover tool in a multi-line plot. 
This was verified in Windows 7, with Firefox and Chrome, using both Python 2.7 and 3.5.
The hover tool works with circle plots (as in the documentation), but not with a multi_line plot. Using a multi_line plot, the console indicates a JS error: "TypeError: renderer.glyph.sx is undefined" when the cursor is hovered above a point of the multi-line. The multi-lin graph renders fine (and other tools like zooming do work). This works with 0.12.4.

Here is a code example:
```
from bokeh.plotting import figure, show, ColumnDataSource
        
source = ColumnDataSource(
                data=dict(
                    x=[[1, 3, 2], [3, 4, 6, 6]],
                    y=[[2, 1, 4], [4, 7, 8, 5]],
                    desc=['a', 'b'],
                )
            )
        
hover = HoverTool(
                tooltips=[
                    ("index", "$index"),
                    ("(x,y)", "($x, $y)"),
                    ("desc", "@desc"),
                ]
            )
        
p = figure(plot_width=400, plot_height=400, tools=[hover],
                   title="Mouse over the dots")
        
p.multi_line('x', 'y', line_width=2, source=source)
       
show(p)
```

bryevdv:
issues: fixes #6416

Tested with the code in the issue. All this complicated code in HoverTool needs much better unit tests at some point.

### DIFF
--- a/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
@@ -230,12 +230,12 @@ export class HoverToolView extends InspectToolView
               jj = j+1
 
             when "nearest"
-              d1x = renderer.glyph.sx[i][j]
-              d1y = renderer.glyph.sy[i][j]
+              d1x = renderer.glyph.sxs[i][j]
+              d1y = renderer.glyph.sys[i][j]
               dist1 = hittest.dist_2_pts(d1x, d1y, sx, sy)
 
-              d2x = renderer.glyph.sx[i][j+1]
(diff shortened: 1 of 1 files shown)


bryevdv:
Im going to ahead and merge this as it seems straightforward (as well as narrow in scope) and was tested to function but please let me know of any concerns.

PR #6420, opened 2017-06-08
Move js_on_change() connections to connect_signals()

### LINKED ISSUE #6419: Restore events for "patch", "stream" and "do"
Previously, `js_on_event` etc. could accept string names of arbitrary Backbone events. This afforded a way for users to add callbacks to CDS `patch` and `stream` events as well as the `do` event when a toolbar button is clicked. When Backbone was removed, this capability was lost, breaking some users' code. Should add three explicit events:
```
Patching
Streaming
Do
```
to `bokeh.events` and have them trigger at the appropriate places on the JS side.

mattpap:
- [x] fix initialization order
- [ ] add tests for initialization order
- [x] add aliases for `streaming` and `patching`
- [x] add deprecation warning

fixes #6419

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -20,6 +20,7 @@ from .themes import default as default_theme
 from .util.callback_manager import PropertyCallbackManager, EventCallbackManager
 from .util.future import with_metaclass
 from .util.serialization import make_id
+from .util.deprecation import deprecated
 from .events import Event
 
 def collect_models(*input_values):
@@ -377,7 +378,7 @@ class Model(with_metaclass(MetaModel, HasProps, PropertyCallbackManager, EventCa
 
         .. code:: python
 
-            source.js_on_change('stream', callback)
+            source.js_on_change('streaming', callback)
 
         '''
         if len(callbacks) == 0:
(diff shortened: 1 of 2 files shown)


mattpap:
@bryevdv, btw. this needs to be merged for 0.12.6.

bryevdv:
I don't think it is mandatory, but I would like to see it go in if possible

mattpap:
But this is a regression. At least what's already done should be merged, because this fixes `js_on_change()` to work with `streaming`, `patching` and `do`.

bryevdv:
@mattpap Oh I see, I didn't realize that was already accomplished. I leave it to you: if you want to merge is as-is please go ahead, just make sure to create follow up issues. But it seems isolated and small so I am happy to merge it later today or early tomorrow

bryevdv:
@mattpap do you plan to add to this PR?

mattpap:
@bryevdv, yes, I will update it soon.

bryevdv:
OK well FYI we would like to cut a last rc in the next 3 hours.

mattpap:
@bryevdv, this is ready.

PR #6438, opened 2017-06-11
Fix merging of 2d selections

### LINKED ISSUE #6433: selecting a line from a multi-line when one is already selected causes an error
The error (copied and pasted at the bottom from console) is raised when you use the tap tool to select one of the lines of the multi-line when one is already selected. There is no error if there is nothing selected or if there are multiple lines selected (selected using a box select on other data in a shared CDS). 

You can see the error using the example below and or examples/plotting/file/multi_line.py
```python
import numpy as np
from bokeh.layouts import row
from bokeh.io import show

N = 9
x = np.linspace(-2, 2, N)
y = x**2

xpts = np.array([-.09, -.12, .0, .12,  .09])
ypts = np.array([-.1,   .02, .1, .02, -.1])

source = ColumnDataSource(dict(
        xs=[xpts*(1+i/10.0)+xx for i, xx in enumerate(x)],
        ys=[ypts*(1+i/10.0)+yy for i, yy in enumerate(y)],
        x=np.arange(9),
        y=np.arange(9)
    )
)

fig1 = figure(tools=['tap'])
fig1.multi_line(xs="xs", ys="ys", line_color="#8073ac", line_width=4, source=source)
fig2 = figure(tools=['box_select', 'tap'])
fig2.circle(x='x', y='y', source=source)

show(row(fig1, fig2))
```

The error looks to be caused by the equality algorithm, used by HasProps when it sets attributes, for the hit-test-result/selection object ({'0d':.., '1d': .., '2d':.. }) when the ['2d']['indices'] object doesn't inherit from Object.prototype. After more digging, it seems like that happens after the ['2d']['indices'] objects get "merged" by core/util/object merge because that function returns an object created by Object.create(null). An easy fix seems to be changing that line to Object.create(Object.prototype).

```
Uncaught TypeError: b.hasOwnProperty is not a function
    at eq (bokeh.js:4821)
    at eq (bokeh.js:4821)
    at eq (bokeh.js:4821)
    at Object.isEqual (bokeh.js:4832)
    at Selector.exports.HasProps.HasProps._setv (bokeh.js:1390)
    at Selector.exports.HasProps.HasProps.setv (bokeh.js:1447)
    at Selector.exports.Selector.Selector.update (bokeh.js:3124)
    at SelectionManager.exports.SelectionManager.SelectionManager.select (bokeh.js:3032)
    at TapToolView.exports.TapToolView.TapToolView._select (bokeh.js:25476)
    at TapToolView.exports.TapToolView.TapToolView._tap (bokeh.js:25456)
```

clairetang6:
This PR fixes the merging of 2d selections. The merge function was moved into object.ts from hittest in [PR 6202](https://github.com/bokeh/bokeh/pull/6202). Before the move, the result of merge was created by `result = {}`. When it was moved to typescript, it became `result = Object.create(null)`. But `result = {}` creates an object whose prototype is Object.prototype, so in object.ts, it should be `result = Object.create(Object.prototype)`. We need the returned object to have Object methods, like `hasOwnProperty`. 

The test "Selector should merge 2d selections with append=true" fails before the change.  

```
  1) Selector should merge 2d selections with append=true:
     TypeError: obj1.hasOwnProperty is not a function
      at Object.merge (build/js/tree/core/util/object.js:43:25)
      at HitTestResult.exports.HitTestResult.HitTestResult.update_through_union (build/js/tree/core/hittest.js:63:46)
      at Selector.exports.Selector.Selector.update (build/js/tree/core/selector.js:27:21)
      at Context.<anonymous> (test/core/selector.coffee:73:9)
```

- [x] issues: fixes #6433 
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/core/util/object.ts
+++ b/bokehjs/src/coffee/core/util/object.ts
@@ -33,7 +33,7 @@ export function merge<T>(obj1: {[key: string] : Array<T>}, obj2: {[key: string]:
   /*
    * Returns an object with the array values for obj1 and obj2 unioned by key.
    */
-  const result: {[key: string]: Array<T>} = Object.create(null);
+  const result: {[key: string]: Array<T>} = Object.create(Object.prototype);
 
   const keys = concat([Object.keys(obj1), Object.keys(obj2)])
 
--- a/bokehjs/test/core/selector.coffee
+++ b/bokehjs/test/core/selector.coffee
@@ -10,6 +10,11 @@ empty_selection = hittest.create_hit_test_result()
 some_1d_selection = hittest.create_1d_hit_test_result([[0, 1], [1, 2]])
 other_1d_selection = hittest.create_1d_hit_test_result([[4, 1], [5, 2]])
 
+some_2d_selection = hittest.create_hit_test_result()
+some_2d_selection['2d']['indices'] = {2: [0, 1]}
+other_2d_selection = hittest.create_hit_test_result()
+other_2d_selection['2d']['indices'] = {2: [2, 3]}
+
 describe "Selector", ->
 
   it "should be constructable", ->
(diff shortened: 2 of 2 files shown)


bryevdv:
documenting OK with late merge due to isolated code path and addition of tests. Thanks @clairetang6

PR #6449, opened 2017-06-12
protect against exceptions when running inline code

bryevdv:
Without this `try/finally` if there is an exception in the the load callback then it is not deleted, and the next time output is attempted, the old callback is tried to run again, again resulting in an error. (i.e. permanently corrupting a notebook until a clear restart)


jbednar:
Thanks!

canavandl:
tested locally and works for me.

philippjfr:
Just tested it as well, works well.

PR #6452, opened 2017-06-13
Bokeh application note doesn't work in jupyter

### LINKED ISSUE #6451: Bokeh application note doesn't work in jupyter

Reported in https://github.com/bokeh/bokeh/pull/5886#issuecomment-280671299

bokeh: 0.12.6dev2
python: 3.5
notebook: jupyter

https://github.com/ecerami/pydata-essentials/blob/master/bokeh/bokeh_notebook_embed.ipynb works in bokeh 0.12.5, but doesn't owrk in 0.12.6dev2

https://github.com/ecerami/pydata-essentials/blob/master/bokeh/bokeh_notebook_embed.ipynb

```
works with bokeh 0.12.5; 0.12.6rc4 fails with the error:

AttributeError Traceback (most recent call last)
in ()
45
46 doc = app.create_document()
---> 47 show(app, notebook_url="localhost:8888")

/local1/anaconda2/lib/python2.7/site-packages/bokeh/io.pyc in show(obj, browser, new, notebook_handle, notebook_url)
282 # in Tornado) just in order to show a non-server object
283 if getattr(obj, '_is_a_bokeh_application_class', False):
--> 284 return _show_notebook_app_with_state(obj, _state, "/", notebook_url)
285
286 if obj not in _state.document.roots:

/local1/anaconda2/lib/python2.7/site-packages/bokeh/io.pyc in _show_notebook_app_with_state(app, state, app_path, notebook_url)
289
290 def _show_notebook_app_with_state(app, state, app_path, notebook_url):
--> 291 if state.notebook_type == 'zeppelin':
292 raise ValueError("Zeppelin doesn't support show bokeh app.")
293

/local1/anaconda2/lib/python2.7/site-packages/bokeh/core/state.pyc in notebook_type(self)
89
90 '''
---> 91 return self._notebook_type
92
93 @notebook_type.setter

AttributeError: 'State' object has no attribute '_notebook_type'
```

zjffdu:
Trivial fix for the issue of #6451, also unit test is added

- [ ] issues: fixes #6451
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/state.py
+++ b/bokeh/core/state.py
@@ -45,6 +45,7 @@ class State(object):
         self.last_comms_handle = None
         self.uuid_to_server = {} # Mapping from uuid to server instance
         self.watching_cells = False
+        self._notebook_type = State.NOTEBOOK_TYPES[0]
         self.reset()
 
     @property
--- a/bokeh/core/tests/test_state.py
+++ b/bokeh/core/tests/test_state.py
@@ -21,6 +21,7 @@ def test_creation():
     assert s.file == None
     assert s.notebook == False
     assert s.watching_cells == False
+    assert s.notebook_type == 'jupyter'
 
 def test_default_file_resources():
     s = state.State()


bryevdv:
I'm not sure about this. As I described in both issues, the problem is that the example notebook never called `output_notebook`. If things worked without that before, it was entirely unintentional and accidental. I maintain that users should always call `output_notebook`, and we always demonstrate that in our examples. I would be more inclined to accept a PR that generates a better warning that directs users to call `output_notebook`.

bryevdv:
Just to be clear about motivations: One of the worst things to have to try and document is "Here is a thing you have to do, *but only sometimes*" It's much simpler and better for everyone concerned to just say "`output_notebook` is *always* required in the notebook".

bryevdv:
Thanks for the PR @zjffdu however I am going to go ahead and close this. I would prefer to see a solution that recognizes that `_notebook_type` is unset that then warns the user that calling `output_notebook` is required in any notebook.

Issue #6481, opened 2017-06-16
Rewrite bokehjs in TypeScript


mattpap:
This is a meta issue for moving bokehjs from coffeescript to typescript.

Module status:
- [x] api
- [x] core
  - [x] layout
  - [x] util
- [x] models
  - [x] annotations
  - [x] axes
  - [x] callbacks
  - [x] canvas
  - [x] expressions
  - [x] filters
  - [x] formatters
  - [x] glyphs
    - [x] webgl
  - [x] graphs
  - [x] grids
  - [x] layouts
  - [x] mappers
  - [x] markers
  - [x] plots
  - [x] ranges
  - [x] renderers
  - [x] scales
  - [x] sources
  - [x] tickers
  - [x] tiles
  - [x] tools
    - [x] toolbars
  - [x] transforms
  - [x] widgets
    - [x] tables
- [x] client
- [x] protocol
- [x] document
- [x] embed

mattpap:
btw. If anyone else would like to join in, then make sure to commit all changes in two stages. First commit rename from `path/file.coffee` to `path/file.ts` and then commit changes necessary to make that file compile. This helps git to maintain history of renamed files.

bryevdv:
@mattpap just FYI there are some outside folks potentially interested in helping parallelize this effort after the low level core classes are done, so just request to keep this issue updated with progress as things go.

jakirkham:
Asking this because I'm curious, what is the motivation for switching from CoffeeScript to TypeScript?

mattpap:
> what is the motivation for switching from CoffeeScript to TypeScript?

The main reason is static typing of bokehjs to reduce maintenance burden and straighten the APIs. I'm not even half through the rewrite, and I already fixed a few dozen of bugs and API inconsistencies. Another reason is that CoffeeScript is a bad language. It's nice on the surface and often eye pleasing, but its certain features are straight from a nightmare, e.g. scoping rules, grammar, implicit returns (to name a few). A nice example this is issue #6588. TypeScript isn't perfect either, but we don't have much choice, and still this transition is a big improvement of the current state of the art. Additionally, at the time of making this decision, CoffeeScript was essentially a dead language. That changed recently with CoffeeScript 2, which isn't a big improvement, but at least allowed us to speed up the transition. There are other smaller yet important reasons (e.g. tooling).

PR #6482, opened 2017-06-16
Start rewriting bokehjs in TypeScript

mattpap:
Rewrite randomly picked modules in TypeScript (mostly core, utils and some tests).

addresses #6481


bryevdv:
Report and tests look good, cursory read of the code also looks good.

PR #6483, opened 2017-06-16
Restore workaround to properly trigger changes in source.data from data_tables

### LINKED ISSUE #6478: Editable data tables not modifying python source.data - bokeh server 0.12.6
Upgrading to 0.12.6, I found that my editable data tables changes where no longer being shown when I called the data.source on the python side of my application. It showed the original value of the data source. However on the JS side the data.source was being properly updated to the new values.

Browser: Chrome 58.0.3029.110
OS: Windows 7 

Environment on which the bug is found
```
- bkcharts=0.2=py35_0
- bokeh=0.12.6=py35_0
- jinja2=2.9.6=py35_0
- markupsafe=0.23=py35_2
- mkl=2017.0.1=0
- numpy=1.13.0=py35_0
- pip=9.0.1=py35_1
- python=3.5.3=3
- python-dateutil=2.6.0=py35_0
- pyyaml=3.12=py35_0
- requests=2.14.2=py35_0
- setuptools=27.2.0=py35_1
- six=1.10.0=py35_0
- tornado=4.5.1=py35_0
- vs2015_runtime=14.0.25123=0
- wheel=0.29.0=py35_0
```

Environment on which there is no bug
```
- bokeh=0.12.5=py35_1
- jinja2=2.9.6=py35_0
- markupsafe=0.23=py35_2
- mkl=2017.0.1=0
- numpy=1.13.0=py35_0
- pip=9.0.1=py35_1
- python=3.5.3=3
- python-dateutil=2.6.0=py35_0
- pyyaml=3.12=py35_0
- requests=2.14.2=py35_0
- setuptools=27.2.0=py35_1
- six=1.10.0=py35_0
- tornado=4.4.2=py35_0
- vs2015_runtime=14.0.25123=0
- wheel=0.29.0=py35_0
```

```
from bokeh.io import curdoc
from bokeh.models import *

x = [1, 2, 3, 4]

source = ColumnDataSource(data=dict(x=x))
columns = [
    TableColumn(field="x", title="x", sortable=True, formatter=NumberFormatter(format='0.')),
]
data_table = DataTable(source=source, editable=True,
                       columns=columns, sizing_mode='fixed')

button_print_py_func = Button(label="Print Python")


def print_python(source=source):
    data = source.data
    print(data)


button_print_py_func.on_click(print_python)

button_print_js = Button(label="Print JS")
button_print_js.callback = CustomJS(args=dict(source=source), code="""
    var data = source.data;
    console.log(data)
""")

buttons = row(button_print_py_func, button_print_js)
layout = column(data_table, buttons)

curdoc().add_root(layout)
```
I expect when I'm printing the data on the python side to see in the terminal the updated values, but on 0.12.6, I see the original source.data not the ones in the table. This was not case with 0.12.5. 
For both version the current values are properly shown on the JS console.

I don't know if this is related, but I also found that on 0.12.5 source.data included an index column which can be seen on both the JS console and the terminal this index column was removed in 0.12.6.

Signapur:
Restore workaround to properly trigger changes in source.data 
Fix typo in tests for data_table.setItem

The workaround used was restored from 260f5765b63060b60680f7f3c7e91df0a5ed637e using the new emit API.

- [x] issues: fixes #6478 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/data_table.coffee
+++ b/bokehjs/src/coffee/models/widgets/data_table.coffee
@@ -99,8 +99,7 @@ export class DataTableView extends WidgetView
     @grid.invalidate()
     @grid.render()
 
-    # XXX: Workaround for `@model.source.trigger('change')` not triggering an event within python.
-    # But we still need it to trigger render updates
+    # This is only needed to call @_tell_document_about_change()
     @model.source.data = @model.source.data
     @model.source.change.emit()
 
--- a/bokehjs/src/coffee/models/widgets/data_table.coffee
+++ b/bokehjs/src/coffee/models/widgets/data_table.coffee
@@ -99,6 +99,11 @@ export class DataTableView extends WidgetView
     @grid.invalidate()
     @grid.render()
 
+    # XXX: Workaround for `@model.source.trigger('change')` not triggering an event within python.
+    # But we still need it to trigger render updates
+    @model.source.data = @model.source.data
+    @model.source.change.emit()
+
   updateSelection: () ->
     if @in_selection_update
       return
--- a/bokehjs/test/models/widgets/data_table.coffee
+++ b/bokehjs/test/models/widgets/data_table.coffee
@@ -144,7 +144,7 @@ describe "data_table module", ->
       expect(r).to.equal null
       expect(dp.source.data).to.deep.equal {index: [0,1,100,10], bar: [3.4, 1.2, 200, -10]}
 
-    it "should set fields when sorted", ->
+    it "should set items when sorted", ->
       source = new ColumnDataSource({data: {index: [0,1,2,10], bar: [3.4, 1.2, 0, -10]}})
       dp = new DataProvider(source)
       fake_col = {sortAsc: true, sortCol: {field: "bar"}}


bokehjs/src/coffee/models/widgets/data_table.coffee
```diff
@@ -99,6 +99,11 @@ export class DataTableView extends WidgetView
     @grid.invalidate()
     @grid.render()
 
+    # XXX: Workaround for `@model.source.trigger('change')` not triggering an event within python.
+    # But we still need it to trigger render updates
+    @model.source.data = @model.source.data
```
bryevdv: @Signapur just to clarify, do things not work without this self-assignment? I would expect just the `@model.source.change.emit()` below to be sufficient

bryevdv: ping @mattpap for comment

mattpap: This should be a no-op.

bryevdv: @Signapur can you remove this line and test to see if it works as expected?

bryevdv: @mattpap assuming things work with the line removed, do you agree this is  reasonable fix at least for now?

mattpap: I see why this works. This line won't emit change signals, because of `isEquals()` check (setting the same data), but server notification is done regardless of that. Lets leave this as-is, but change the comment to indicate that this is only needed to call `@_tell_document_about_change()`.

bryevdv: OK, @Signapur can you update the comment to mention `@_tell_document_about_change()`? And then we will get this merged!

bryevdv:
Thanks @Signapur !

bokehjs/src/coffee/models/widgets/data_table.coffee
```diff
@@ -99,6 +99,10 @@ export class DataTableView extends WidgetView
     @grid.invalidate()
     @grid.render()
 
+    # This is only needed to call @_tell_document_about_change()
+    @model.source.data = @model.source.data
+    @model.source.change.emit()
```
rs2: @DuCorey , @bryevdv This doesn't propagate the update back to Python's `source.data`, does it?

mattpap: It does propagate changes back to Python. This happens unconditionally (almost, there's `silent` option) when a property is assigned to. Additional `emit` is necessary to trigger internal bokehjs' signals, because this requires `!isEqual(old_data, new_data)`, which isn't the case here.

rs2: @mattpap I'm almost certain that I saw `change.emit()` propagate changes on the js side, but not back to Python side... WOuld you have a counterexample?

mattpap: > I'm almost certain that I saw change.emit() propagate changes on the js side, but not back to Python side.

That's true, but `emit` and propagation of changes to Python are two independent features. To propagate changes to Python it's sufficient to assign to a property. If you look at https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/core/has_props.ts#L254-L283, knowing that `obj.prop = val` translates to `obj.setv({prop: val})`, then it should be clear that changes are propagated as long as `silent` is not set or is `false`.

PR #6485, opened 2017-06-17
add mark for tests needing selenium and support pytest --unmarked

### LINKED ISSUE #6441: Enable exclusion of selenium tests for static image export
Running Bokeh unit tests on 0.12.6.dev7...

py.test -m "not (examples or integration or js or quality)" bokeh

reports 3 failed tests. In discussion with @bryevdv looks like they are selenium tests for the new static image export feature. Would be useful to add a new mark for them so that they can be excluded easily.

bryevdv:
issues: fixes #6441

Non-integration tests that require selenium can now be selected with `py.test -m selenium` This includes, e.g.some png and svg export unit tests. 

Additionally `py.test -m unit` will run all tests otherwise unmarked (i.e the basic python unit tests)

### DIFF
--- a/bokeh/command/subcommands/tests/test_png.py
+++ b/bokeh/command/subcommands/tests/test_png.py
@@ -67,6 +67,7 @@ bokeh png: error: %s
 """ % (too_few)
         assert out == ""
 
+@pytest.mark.unit
 @pytest.mark.selenium
 def test_basic_script(capsys):
     def run(dirname):
@@ -81,6 +82,7 @@ def test_basic_script(capsys):
     with_directory_contents({ 'scatter.py' : basic_scatter_script },
                             run)
 
+@pytest.mark.unit
 @pytest.mark.selenium
 def test_basic_script_with_output_after(capsys):
     def run(dirname):
@@ -94,6 +96,8 @@ def test_basic_script_with_output_after(capsys):
 
     with_directory_contents({ 'scatter.py' : basic_scatter_script },
                             run)
+
+@pytest.mark.unit
 @pytest.mark.selenium
 def test_basic_script_with_output_before(capsys):
     def run(dirname):
--- a/bokeh/command/subcommands/tests/test_png.py
+++ b/bokeh/command/subcommands/tests/test_png.py
@@ -67,6 +67,7 @@ bokeh png: error: %s
 """ % (too_few)
         assert out == ""
 
+@pytest.mark.selenium
 def test_basic_script(capsys):
     def run(dirname):
         with WorkingDir(dirname):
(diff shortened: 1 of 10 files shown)


bryevdv:
@marcuskd Can you check out this PR and see if it helps make tests easier to run?

bryevdv:
Somewhat unrelated, but also changed test deps to depend on `notebook` instead `of `jupyter` as the latter brings in Qt unecessrily

marcuskd:
Hi Bryan,

I just ran *py.test -m unit* on branch bryanv/6441_test_labels. Results:

*990 passed, 1 skipped, 186 deselected in 102.19 seconds*

No selenium errors, looks good to me!

Hope that helps,
Marcus.


On Sat, Jun 17, 2017 at 10:12 PM, Bryan Van de Ven <notifications@github.com

setup.cfg
```diff
@@ -19,11 +19,14 @@ max-line-length = 165
 norecursedirs = build _build node_modules
 python_files = *_tests.py *_test.py test_*.py
 selenium_exclude_debug = html logs
+implicit_marker = unit
```
canavandl: I didn't know this was an option. Good to know.

bryevdv: It's not, this PR adds it

canavandl: ah yes, I didn't scroll down enough :)

bryevdv:
I was thinking of changing this slightly. I think the implicit mark could also be applied explicitly. So the selenium tests could also be marked unit. Then unit by itself runs everything and "unit and not selenium" would run unit test minus selenium. Thoughts? Could also use generic "slow" which seems common instead of selenium

mattpap:
> Could also use generic "slow" which seems common instead of selenium

`slow` has a different meaning, so lets better not make this more confusing. `slow` should be used with different classes of tests, indicating that a particular test is much slower than the average.

canavandl:
> Then unit by itself runs everything and "unit and not selenium" would run unit test minus selenium.

I like that. It makes the `pytest -m unit` directive simpler. I'd stick w/ `selenium` over `slow` though.

bokeh/command/subcommands/tests/test_png.py
```diff
@@ -92,7 +94,7 @@ def run(dirname):
 
     with_directory_contents({ 'scatter.py' : basic_scatter_script },
                             run)
-
+@pytest.mark.selenium
```
canavandl: Could use an empty line above.

bryevdv:
OK I will make the change later today (keeping "selenium" but adding a "unit" mark explicitly)

bryevdv:
@bokeh/dev OK I have made the small change described above

@marcuskd note the change `py.test -m unit` will run all unit tests, *including* the selenium unit tests. To exclude them you can run `py.test -m"unit and not selenium"`

bryevdv:
It's worth noting that any unit test that has any explicit mark at all will also need to be explicitly marked unit, to be considered a unit test. That is a little crappy but still an overall improvement. Will merge when green.

PR #6487, opened 2017-06-17
remove all 0.12.4 deprecations

bryevdv:
This removes all the deprecations up to those made in `0.12.4`

Our recent informal policy has been to remove deprecations after two release cycles which means we should do our best to get any remaining deprecations into `0.12.7` before removing *all* deprecations for 1.0


bryevdv:
@bokeh/dev any objection merging this sooner rather than later?

PR #6489, opened 2017-06-18
add option to upload binary files to CDN

### LINKED ISSUE #6464: examples upload failed on full release build
Reported error:

```
 'utf-8' codec can't decode byte 0xcd in position 12: invalid continuation byte
```

https://travis-ci.org/bokeh/bokeh/jobs/242442407#L1826

bryevdv:
issues: fixes #6464

This adds an option to upload binary data such as the examples zip file correctly to CDN. 

Tested locally by hand, however for good measure examples upload has also been moved last so that if it fails, nothing else is impacted. 

Would like to merge this before `0.12.7dev2`

### DIFF
--- a/scripts/build_upload.py
+++ b/scripts/build_upload.py
@@ -173,9 +173,10 @@ def cdn_upload(local_path, cdn_path, content_type, cdn_token, cdn_id, binary=Fal
                             "Origin: https://mycloud.rackspace.com",
                             "Content-Type: %s" % content_type])
     if binary:
-        c.setopt(pycurl.POSTFIELDS, open(local_path, "rb").read())
+        data = open(local_path, "rb").read()
     else:
-        c.setopt(pycurl.POSTFIELDS, open(local_path).read().encode('utf-8'))
+        data = open(local_path).read().encode('utf-8')
+    c.setopt(pycurl.POSTFIELDS, data)
     c.perform()
     c.close()
 
--- a/scripts/build_upload.py
+++ b/scripts/build_upload.py
@@ -161,7 +161,7 @@ def upload_wrapper(name):
         return wrapper
     return decorator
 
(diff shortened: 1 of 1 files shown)


scripts/build_upload.py
```diff
@@ -172,7 +172,10 @@ def cdn_upload(local_path, cdn_path, content_type, cdn_token, cdn_id):
     c.setopt(c.HTTPHEADER, ["X-Auth-Token: %s" % cdn_token,
                             "Origin: https://mycloud.rackspace.com",
                             "Content-Type: %s" % content_type])
-    c.setopt(pycurl.POSTFIELDS, open(local_path).read().encode('utf-8'))
+    if binary:
+        c.setopt(pycurl.POSTFIELDS, open(local_path, "rb").read())
+    else:
+        c.setopt(pycurl.POSTFIELDS, open(local_path).read().encode('utf-8'))
```
mattpap: It's a minor thing, but you should try to avoid this kind of duplication. Much better would be:
```
if binary:
    data = open(local_path, "rb").read()
else:
    data = open(local_path).read().encode('utf-8')
c.setopt(pycurl.POSTFIELDS, data)
```

mattpap:
I'm just curious if we can't upload everything as binary?

scripts/build_upload.py
bryevdv: Sure, that's better, I'll make the change before merging.

bryevdv:
> I'm just curious if we can't upload everything as binary?

Not sure, but I know the other works, and also that it's extremely hard to test "for real" so I'm not inclined to change things for the text files.

mattpap:
> (...) hard to test "for real" so I'm not inclined to change things for the text files.

That's reasonable.

bryevdv:
@mattpap I've made the change you requested

PR #6491, opened 2017-06-18
fix line glyph rendering with selected points

### LINKED ISSUE #515: line plots render selections wierd
if you do a line plot, and then try to select on it, it draws lines strangely skipping the unselected points

clairetang6:
This PR fixes the weird behavior of the line glyph with selected and non-selected points. 

In the example, I added some NaNs to show how they are handled.
Before:
line_selection_before

After:
line_selection

One potential way to improve the rendering of the non-selected points in the line would be to extend from the non-selected points one back and one forward, so that there wouldn't be a gap between the selected and non-selected points. But then the line glyph view would need to know whether it is rendering selected or non-selected points, which is info at the glyph renderer level. The glyph renderer could tell the line glyph when it calls glyph.render(...) (with an additional parameter as a selected flag?), but I wasn't sure if that was worth it or necessary. 

line_selection_example

- [x] issues: fixes #515 
- [x] tests added / passed

```python
import numpy as np
from bokeh.io import show

N = 30
x = np.linspace(-2, 2, N)
y = np.sin(2*x)
y[15] = np.NaN
y[17] = np.NaN
y[20] = np.NaN

source = ColumnDataSource(dict(x=x, y=y))

fig1 = figure(tools=['tap', 'box_select', 'hover', 'save'], plot_width=500, plot_height=300)
fig1.line(x="x", y="y", source=source, line_color="#F46D43", line_width=6, line_alpha=0.6)
fig1.circle(x="x", y="y", source=source)

show(fig1)
```


mattpap:
@clairetang6, you should try adding an integration test for this. This should work as is in the gif, create plot, make selection gesture (that's the tricky part, if there is no template for this in integration tests already) and verify a screenshot.

clairetang6:
@mattpap Ok. There are some tests that have selenium making box selections, so I should be able to base it off of that.

bryevdv:
@clairetang6 I think I had in mind your more difficult improvement, which is perhaps why I have put off trying to implement it :) This simpler approach did not occur to me. I think this is a clear and immediate improvement, I would say we should get this merged then make a follow-on issue for refinements.

bryevdv:
@clairetang6 if we go that route, is this PR `ready`?

bryevdv:
Thanks @clairetang6 ! Can you add new issues for any follow on work (incremental improvements, or maybe extension to multi lines, etc)

PR #6498, opened 2017-06-19
update TravisCI to latest (supported) linux dist

### LINKED ISSUE #6497: Update linux dist on TravisCI
I contacted TravisCI support about the recent slowdown. It seems the current dist we use is EOL. Updating may bring back performance but should probably be done in any case.

> Thanks for reaching out and I'm sorry for this slowdown of your builds.
> 
> We are unsure what is causing this behavior and, unfortunately, we probably won't be able to investigate further. The main reason is that this specific infrastructure (i.e. container-based Precise) will be decommissioned in the near future given that Ubuntu Precise has now reached its end of life. You can have more details about your plan in this blog post. Sorry for the inconvenience.
> 
> Hence, we are now advising all our customers to move away from this infrastructure and instead use our container-based Trusty infrastructure. You can route your builds over there by adding the following to your .travis.yml file:
> 
> sudo: false
> dist: trusty
> 
> Can you try this out and let us know the results you get?

bryevdv:
fixes #6497

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
-sudo:
-  false
+sudo: false
+dist: trusty
 
 env:
   global:


bryevdv:
Sample size of 1 seems to suggest test time improvement. In any case, tests pass, so merging.

mattpap:
Test times are at the lower end of what was previously. Hopefully, there will be lest fluctuation.

bryevdv:
yes seems to be holding up in further runs. Fingers crossed. Will also be working on streamlining the build using stages soon too.

PR #6534, opened 2017-06-25
Add password input (Simplified)

### LINKED ISSUE #6492: Expose TextInput type as a property of widget. (For Password Typing)
Inspired by issue  [#5417](https://github.com/bokeh/bokeh/issues/5417) (which discusses the addition of placeholder on the TextInput widget) I think it is appropriate to expose also the **type** to the widget. 

One of the advantages of doing that should be to allow password input boxes. 

Currently, the Textinput does not allow to type a password without displaying what the user is typing. The only parameters allowed are "callback", "value", "placeholder" (see [bokeh doc](http://bokeh.pydata.org/en/latest/docs/reference/models/widgets.inputs.html#bokeh.models.widgets.inputs.TextInput)).

A quick inspection to a Bokeh generated page gives the following for an input text object:
```
<input class="bk-widget-form-input" **type="text"** id="f8c02d2b-0630-49b6-b01f-c8071c97370c" value="***" placeholder="">
```
However, if we could have the same with only the type edited to 'password', one could enter passwords with the default dots being displayed instead of the letters typed.
```
<input class="bk-widget-form-input" **type="password"** id="f8c02d2b-0630-49b6-b01f-c8071c97370c" value="***" placeholder="">
```
I believe this should be a simple addiction to bokeh, and I imagine I could do it myself if someone could point me on the right direction. I still did not fully understand how bokeh transform python into elements in HTML, so I got lost tracking the necessary changes.

mauhcs:
This branch adds a Input text for passwords.

- [ ] issues: fixes #6492
- [ ] tests added / passed


bokehjs/src/coffee/models/index.ts
```diff
@@ -17,3 +17,4 @@ export * from "./sources"
 export * from "./tickers"
 export * from "./tiles"
 export * from "./tools"
+export * from "./widgets"
```
mattpap: This needs to be removed. Widgets are part of a separate bundle and can't be included here.

bokehjs/src/coffee/models/widgets/password_input.coffee
```diff
@@ -0,0 +1,11 @@
+import {TextInputWidget, TextInputWidgetView} from "./text_input_widget"
+import template from "./password_input_template"
```
mattpap: This isn't needed anymore. Also the tsx file can be removed as well.

bokeh/models/widgets/inputs.py
```diff
@@ -51,6 +51,21 @@ class TextInput(InputWidget):
     """)
 
 
+class PasswordInput(InputWidget):
```
mattpap: This needs to extend `TextInput`. Then `callback` and `placeholder` can be removed, as they are already defined in `TextInput`.

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,36 @@
...
+                                     secret.text = 'Right Password. The Secret is 42.';
+                                      return;
+                                     }
+                               }
+                               secret.text = "Wrong Password."; """)
+#user.callback = verify_pwd # Check password pressing enter. -BUG Works also with tab :(
+pwd.callback = verify_pwd # Check password pressing enter.  -BUG Works also with tab :(
+btn.callback = verify_pwd # Check password clicking on the Button.
+
+output_file("password.html", title="Password Field")
```
mattpap: This file name must much py file name (up to extension), otherwise tests will fail.

bokehjs/src/coffee/models/widgets/password_input.coffee
```diff
@@ -0,0 +1,10 @@
+import {TextInputWidget, TextInputWidgetView} from "./text_input_widget"
```
mattpap: `TextInput`, `TextInputView`, `./text_input`.

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,36 @@
+from bokeh.models.widgets import PasswordInput, TextInput, PreText, Button
+from bokeh.layouts import column, row
+from bokeh.plotting import show, output_file
+from bokeh.models.callbacks import CustomJS
+import bokeh
+
+print bokeh.models.widgets.__file__
```
mattpap: This line has to be removed.

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,36 @@
...
+
+## Verify if the password typed is bokeh using a JS script
+verify_pwd = CustomJS(args=dict(user=user, pwd=pwd, secret=secret),
+                      code=""" if (pwd.value == '"""+PASSWD+"""') {
+                                    if (user.value == '"""+USER+"""') {
+                                     secret.text = 'Right Password. The Secret is 42.';
+                                      return;
+                                     }
+                               }
+                               secret.text = "Wrong Password."; """)
```
mattpap: This could be rewritten to something more readable, e.g.:
```
verify_pwd = CustomJS(args=dict(user=user, pwd=pwd, secret=secret), code="""
  if (pwd.value === "%s" && user.value === "%s") {
    secret.text = 'Right Password. The Secret is 42.';
  } else {
    secret.text = "Wrong Password.";
  }
""" % (PASSWD, USER))
```

bryevdv:
@bokeh/dev I think this is definitely a worthwhile feature that had been over looked. Now that we have a concrete PR to focus, we just need to decide: should this be a new separate model? Or should a `type` property be added to the existing input? Thoughts?

Thanks @mauhcs !

mattpap:
> Or should a type property be added to the existing input? Thoughts?

It can't be a property on `TextInput`, because there is `AutocompleteInput` inheriting from it, and you don't want to end up with autocomplete password input. Unless, the class hierarchy changes, which may be a good thing, but requires some broader thought, this is the proper course of action.

bryevdv:
OK in that case I think this just needs some light documentation support and will be good to go.

bokeh/models/widgets/inputs.py
```diff
@@ -51,6 +51,12 @@ class TextInput(InputWidget):
     """)
 
 
+class PasswordInput(TextInput):
+    ''' Single-line password input widget.
+
+    '''
```
bryevdv: Probably worth a mention of what exactly the means, i.e. the value can't be inspected in browser inspectors

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Set User and Password:
+USER = "Mau"
+PASSWD = "Bok3h"
+
+## Create Texts, inputs and button
+text = PreText(text="LOGIN TO KNOW\nTHE SECRET:")
+user = TextInput(placeholder="username", title="(UserName: "+USER+")")
+pwd = PasswordInput(placeholder="password", title="(Password: "+PASSWD+")")
+btn = Button(label="GO!",width=150)
```
bryevdv: This comment and the one above can be omitted

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,32 @@
...
+verify_pwd = CustomJS(args=dict(user=user, pwd=pwd, secret=secret),
+                      code=""" if (user.value = '{}') & (pwd.value == '{}') {{
+                                   secret.text = 'Right Password. The Secret is 42.';
+                                   return;
+                               }}
+                               secret.text = 'Wrong Password.';
+                               """.format(USER, PASSWD))
+#user.callback = verify_pwd # Check password pressing enter. -BUG Works also with tab :(
+pwd.callback = verify_pwd # Check password pressing enter.  -BUG Works also with tab :(
+btn.callback = verify_pwd # Check password clicking on the Button.
```
bryevdv: I'd suggest making an issue for the bug on GH and omitting the comments here

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,32 @@
...
+
+output_file("using_password_input.html", title="Password Field")
+page = row(column(text,user, pwd,btn),secret)
+show(page)
```
bryevdv: It's also probably worth some kind of comment that stresses that the usage here is purely for demonstration, and that putting the password plaintext in a `CustomJS` is not advised since it would expose the password.

bryevdv:
@mauhcs I left a few small comments otherwise I think this is very close, thanks for the PR!

examples/howto/using_password_input.py
mauhcs: Which comments? 
The title for the Inputs?

examples/howto/using_password_input.py
mauhcs: GH?

examples/howto/using_password_input.py
```diff
@@ -0,0 +1,32 @@
+from bokeh.models.widgets import PasswordInput, TextInput, PreText, Button
+from bokeh.layouts import column, row
+from bokeh.plotting import show, output_file
+from bokeh.models.callbacks import CustomJS
+
+# Set User and Password:
+USER = "Mau"
+PASSWD = "Bok3h"
+
+## Create Texts, inputs and button
```
mauhcs: You mean this comment and the one above?

examples/howto/using_password_input.py
bryevdv: `# Set User and Password:` and `## Create Texts, inputs and button` both seem redundant to me

mauhcs:
Thank you @bryevdv and @mattpap for the feedback. I use bokeh a lot, so it is a pleasure to help. Too bad my coffee and JS skills aren't that amazing, though. That just made me appreciate Bokeh even more.

examples/howto/using_password_input.py
pzwang: Agreed - For anything which suggests it could/should be used for security purposes, it's important that we include documentation in both the example, as well as the docstring for the class.  It's not merely to disclaim liability, but also to clearly lay out what aspects of security are *not* addressed by this widget: various ways the password text may be stored in plaintext, various potential attack vectors, SSL/non-HTTPS connections, etc. etc.

I realize this is not a trivial amount of work, but it also doesn't have to be super extensive.. It just needs to give the user the impression that this is not "automagically secure".

mauhcs: @pzwang, Thanks for the feedback.
I wrote a few lines on the example and on the docstring.  Have a look if that fits the bill.

bryevdv:
Thanks @mauhcs, I think I will want to edit the messages a bit but that can happen later and these are good for the time being. I will merge this now. Thanks for your contribution!

PR #6544, opened 2017-06-26
Canavandl/187 graph vis

### LINKED ISSUE #187: Support graphs/trees/networks
Need to add support for visualizing graph data. Need to integrate with various formats (dot, networkx, etc.), support things like force directed layouts, etc.

canavandl:
- [ ] issues: fixes #187 
- [ ] tests added / passed

Todo:
- [x] Add a `from_networkx`
- [x] GraphDataSource (python) validation (assert nodes.index col, assert edges.start/end col, that layout nodes exist in nodes?)
- [x] add ``Plot.add_graph`` method.
- [x] connect signals
- [x] rename GraphDataSource -> GraphSource
- [ ] documentation


bokehjs/src/coffee/models/graphs/graph_data_source.coffee
```diff
@@ -0,0 +1,12 @@
+import {Model} from "../../model"
+import {ColumnDataSource} from "../sources/column_data_source"
+import {logger} from "../../core/logging"
+import * as p from "../../core/properties"
+
+export class GraphDataSource extends Model
+  type: 'GraphDataSource'
+
+  @define {
+    nodes: [ p.Instance, new ColumnDataSource({data: {index: []}})          ]
```
mattpap: `() -> new ...`

bokehjs/src/coffee/models/renderers/graph_renderer.coffee
```diff
@@ -0,0 +1,68 @@
...
+
+export class GraphRendererView extends RendererView
+
+  initialize: (options) ->
+    super(options)
+    @edges = @build_glyph_view(@model.edges)
+    @nodes = @build_glyph_view(@model.nodes)
+
+    # hackery
+    @edges._index_data = () -> null
```
canavandl: this will be removed

bokehjs/src/coffee/models/graphs/static_layout_provider.coffee
```diff
@@ -0,0 +1,24 @@
+import {LayoutProvider} from "./layout_provider"
+import * as p from "../../core/properties"
+
+export class StaticLayoutProvider extends LayoutProvider
+  type: "StaticLayoutProvider"
+
+  get_node_locations: (graph_source) ->
+    [x, y] = [[], []]
+    for i in graph_source.nodes.data.index
+      x.push(@layout[i][0] ? null)
```
mattpap: This needs to be renamed to something less confusing with existing layout. `graph_layout` would suffice.

canavandl: fixed

bokehjs/src/coffee/models/graphs/graph_data_source.coffee
canavandl: fixed

canavandl:
current status:

```python
from bokeh.models import Range1d, GraphDataSource, GraphRenderer, StaticLayoutProvider

p = figure(x_range=Range1d(0,3), y_range=Range1d(0,3))

graph_source = GraphDataSource()
graph_source.nodes.data['index'] = [1,2,3,4]
graph_source.edges.data = {'start': [1,1,1], 'end': [2,3,4]}

layout_provider = StaticLayoutProvider(
    graph_layout={1: [1,1], 2: [1,2], 3: [2,1], 4: [2,2]}
)

r = GraphRenderer(graph_source=graph_source, layout_provider=layout_provider)

p.renderers.append(r)

show(p)
```

yields:
screen shot 2017-06-27 at 1 49 12 pm

bokeh/models/graphs.py
```diff
@@ -0,0 +1,102 @@
...
+        edges_start = [edge[0] for edge in edges]
+        edges_end = [edge[1] for edge in edges]
+
+        node_source = ColumnDataSource(data=dict(index=nodes))
+        edge_source = ColumnDataSource(data=dict(
+            start=edges_start,
+            end=edges_end
+        ))
+
+        graph_layout = nx.circular_layout(G)
```
canavandl: there are several networkx layout methods. What's the best way to expose them all without adding a networkx dep outside of this classmethod?

maybe have users pass in callables and if func=None, then use some default:
```python
def from_networkx(cls, G, func=None):
    if func is None:
        func = nx.circular_layout
    ....
```

bokeh/models/graphs.py
```diff
@@ -0,0 +1,102 @@
...
+    """)
+
+
+_DEFAULT_NODE_SOURCE = lambda: ColumnDataSource(data=dict(index=[]))
+
+
+_DEFAULT_EDGE_SOURCE = lambda: ColumnDataSource(data=dict(start=[], end=[]))
+
+
+class GraphDataSource(Model):
```
canavandl: Also need to validate that all start/end nodes are present in node.index

canavandl:
```python
import networkx as nx
from bokeh.models import Range1d, GraphSource

G=nx.Graph()
G.add_nodes_from([0,1,2,3])
G.add_edges_from([[0,1], [0,2], [2,3]])

graph_source = GraphSource.from_networkx(G)

p = figure(x_range=Range1d(-1.1,1.1), y_range=Range1d(-1.1,1.1))
p.add_graph(graph_source)

show(p)
```

hsparra:
What about adding a helper method to generate a graph based upon a passed data frame. 

1. The source node and destination node columns would need to be specified.
2. If the destination node was not populated then it would be an unconnected node.
3. If a directed graph, then direction would be from source to destination node. If connection was bi-directional then would have another row entry with source and destination nodes swapped.
3. Additional columns could be specified as providing meta-data, or just any additional columns would be meta-data.
4. Possibly be able to specify a column to use for weighting, or could just be meta-data.

bokeh/models/graphs.py
bryevdv: Sure that's probably the best approach. I'd even consider just making the argument required, then we don't have import `networkx` ourselves at all anywhere.

bokeh/models/renderers.py
```diff
@@ -168,6 +169,46 @@ def _check_bad_column_name(self):
...
+    graph_source = Instance(GraphSource, help="""
+    Instance of a GraphSource which contains the information about the
+    nodes and edges of a network graph.
+    """)
+
+    nodes = Instance(Marker, default=_DEFAULT_NODE, help="""
+    Instance of a Marker glyph that will be rendered as the graph nodes. The
+    marker attribute fields will be matched against columns in the
+    ``graph_source.nodes`` ColumnDataSource.
+    """)
```
bryevdv: Is `Marker` general enough? I can imagine `Rect`, `Oval`, `Ellipse` as reasonable as well. Basically any glyph that has an x/y "center"

bokeh/models/renderers.py
```diff
@@ -168,6 +169,46 @@ def _check_bad_column_name(self):
...
+
+    edges = Instance(MultiLine, default=_DEFAULT_EDGE, help="""
+    Instance of a MultiLine glyph that will be rendered as the graph edges. The
+    multiline attribute fields will be matched against columns in the
+    ``graph_source.edges`` ColumnDataSource.
```
bryevdv: I wonder if we should not expose this publicly just yet. If the intent is to allow room for future "curved paths" between nodes, I think it might be better to have the graph renderer be able to deduce what is needed based on some property of the layout provider. The reason I worry about making a publicly property to try and coordinate that here, is that most arbitrary combinations of values here with arbitrary layout providers would probably not work together

bokehjs/src/coffee/models/renderers/graph_renderer.coffee
bryevdv: Right, using the glyphs as implementation details is a perfectly reasonable thing to want to do, but is also a thing that has not really *been done* before. So I would expect some changes might be necessary to accommodate this case, likely loosening some assumptions in various places.

canavandl:
defaults test failure should be resolved by #6573

bokeh/models/renderers.py
canavandl: I agree. I think we just need to mirror the XYGlyph abstract class into Bokeh and require an instance of that.

bryevdv: +1

bokehjs/src/coffee/models/renderers/graph_renderer.coffee
canavandl: this has been resolved

bokeh/models/graphs.py
canavandl: donezo

bokeh/models/renderers.py
canavandl: commenting for posterity: The MultiLine glyph is exposed so that users can style the visual attributes of the graph edges. The LayoutProvider will supply the coordinate data.

PR #6547, opened 2017-06-27
clean up phantomjs procs in unit tests

### LINKED ISSUE #6525: py.test bokeh/tests/test_io.py doesn't kill phantomjs precesses
Run:
```
py.test bokeh/tests/test_io.py
ps aux | grep phantom
```

canavandl:
clean up phantomjs procs in unit tests

- [ ] issues: fixes #6525 
- [ ] tests added / passed


canavandl:
I need to investigate using a selenium test fixture with a session scope. I tried this previously and had issues, but I don't remember why. This would resolve having to manually start/stop a phantomjs process in each unit test and probably speed the tests up a bit.

PR #6563, opened 2017-06-29
Fix up stages install order issues

bryevdv:
This PR attempts to 

* prevent installs when they are not needed
* make sure installs happen when they are needed in the deploy stage

### DIFF
--- a/scripts/ci/install:build
+++ b/scripts/ci/install:build
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e # exit on error
+set -x # echo commands
+
+if  [[ ! -z "${TRAVIS_TAG}" ]]; then
+    echo
+    echo "SKIPPING BUILD STAGE INSTALL --- This is a build for release tag ${TRAVIS_TAG}"
+    echo
+    exit 0
+fi
+
+# do the base install
+bash scripts/ci/install
+
+# install build time dependencies
+BUILD_DEPS=`python scripts/deps.py build`
+conda install --yes ${BUILD_DEPS}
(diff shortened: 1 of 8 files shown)


scripts/ci/install:build
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
pratapvardhan: colon in filenames are invalid on windows. It is understandable that files under `scripts/ci/*` are meant for ci only. 

However, on windows, with git, these files get removed and creates clutter. Would you be okay to rename files with colon to underscore maybe?

bryevdv: I would be -1 on underscores, I want it to be visually evident that the there is a component:subcomponent relationship in these files and I don't think an underscore would convey that as it would just look like a snake case name. Possible alternatives:

* Use other unicode characters that happen to be similar to the ascii colon: 
https://stackoverflow.com/questions/10386344/how-to-get-a-file-in-windows-with-a-colon-in-the-filename

* Use other ascii punctuation, `install.build`, `install~build`...

pratapvardhan: +1 for using any of `install.build`, `install~build`  ...

mattpap: I would be wary about using `~`. In fact, `:` is problematic in some shells on linux as well (e.g. I don't have auto-completion with `:` in file names). `.` perhaps the safest.

bryevdv: @pratapvardhan please feel free to submit a PR with a change to `.`

PR #6569, opened 2017-06-30
hotfix for Travis log truncation

### LINKED ISSUE #6568: hotfix for TravisCI log truncation
Got the following response from TravisCI re: log truncation

> Thanks for following up. We apologize for the inconveniences caused. We believe that the cut-off of the logs is due to the rollout of the filter for secret environment variables. See this incident as 
> 
> We are working on a better implementation.
> 
> In the mean while, you could run your builds with the following snippet in the top level of .travis.yml:
> 
> filter_secrets: false
> 
> This will disable the filtering of the contents of environment variables marked as secret in the build log, should they accidentally be leaked there, and will bring you back to the situation as it was before May.
> 
> Please let us know how that works out for you.
> 
> We will let you know when our implementation has been improved, so you can remove this workaround.
>

bryevdv:
issues: fixes #6568

ping @canavandl @mattpap just FYI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: generic
 sudo: false
 dist: trusty
 
+# This does not show secret variables, but only disables a filter for accidental spillage. The filter
+# is causing problems with log trunction and TravisCI support advised to turn it off for now. See the
+# issue https://github.com/bokeh/bokeh/issues/6568 for more information.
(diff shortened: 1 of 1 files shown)


canavandl:
huh.

bryevdv:
This also seems to fix the missing log coloration.

PR #6570, opened 2017-06-30
Remove unused daylight data download func to remove requests build dep

### LINKED ISSUE #6543: remove requests as runtime dependency?
This only seems to be used for one sampledata download, in which case we should find a different way to grab the data with stdlib functions and remove the dependency.

ping @mattpap

canavandl:
Remove unused daylight data download func to remove requests build dep

- [ ] issues: fixes #6543 
- [ ] tests added / passed

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -67,6 +67,7 @@ test:
     - selenium >=3
     - pytest-xdist
     - pytest-rerunfailures
+    - requests >=1.2.3
     - beautifulsoup4
     - scipy
     - pillow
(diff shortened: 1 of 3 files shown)


conda.recipe/meta.yaml
```diff
@@ -67,6 +67,7 @@ test:
     - selenium >=3
     - pytest-xdist
     - pytest-rerunfailures
+    - requests >=1.2.3
```
bryevdv: This one is still needed I think, for tests. It's the runtime dependency in `setup.py` I wanted to get rid of. Tho if the tests pass I guess I am wrong and this can go!

bryevdv: Turns out requests is part of miniconda / required by conda. So that's why it works for our tests on travis with this removed. I worry that someone working in a non-root conda env, or without conda at all could get tripped up by this being missing. I'd vote to leave this in as a test-dep only.

bryevdv: disregard, I just need more sleep

canavandl: Yes. This line adds requests as a test dep to our meta.yaml

PR #6585, opened 2017-07-04
Mattpap/6584 webpack

### LINKED ISSUE #6584: Add a sample configuration for bundling with webpack
I don't think we will actually switch to webpack, but at least lets have a sample configuration available, as this was asked many times before.

mattpap:
fixes #6584


mattpap:
We really need to replace phantomjs with headless chrome. I upgraded sprintf with sprintf-js (for several reasons, one that the former isn't maintained anymore) and the new one doesn't work in phantomjs.

bryevdv:
@mattpap in the linked issue you mention not thinking we'd switch to webpack, has this opinion changed? I'd like to understand the plan for this PR.

mattpap:
@bryevdv, no, there was no intent to switch to webpack, just to show how to configure it if user would like to repack bokehjs with his own code using webapck. Even that much is currently barely possible, due to how webpack works and some bugs. So, I hijacked this PR to implement a custom linker that uses webpack-inspired encoding scheme (much more efficient than browserify). To finish this PR, I need to restore support for imports/require in CustomJS and extensions, which I'm almost done with, and restore (or fix) support for source maps.

bryevdv:
@mattpap I would regard source maps as a low priority. They can be done in a follow-on PR tho I wonder if we are porting wholesale to TS if they are worth worrying about at all. What is the timeframe to complete just the CustomJS work needed for this PR?

mattpap:
> What is the timeframe to complete just the CustomJS work needed for this PR?

Tomorrow.

mattpap:
So, I added support for source maps. If you compile bokehjs with `gulp build --sourcemaps`, you will be able to debug in typescript. This is behind a switch, because it makes `scripts:bundle` a lot slower (8s vs 2s without `--sourcemaps` on my computer). Note however, that even without `--sourcemaps`, source maps are actually generated, but in much weaker version (and thus faster). Unfortunately, source maps generated for minified bokehjs (with or without `--sourcemaps`) are broken due to bugs in `gulp-uglify` (they appear to be working, but are offset).

mattpap:
Thanks to TypeScript, work on `bokehjs/gulp/linker.ts` was a breeze. I hardly had to fix any issues after fixing compiler errors (assuming strong typings, which isn't often the case).

mattpap:
> (...) we are porting wholesale to TS if they are worth worrying about at all.

Having any source maps is an improvement, but having original source code is really pleasant to work with. Thought there are caveats, because e.g. what you see in source code, isn't what you can access in JS console.

PR #6586, opened 2017-07-05
Canavandl/6521 deprecate tool events

### LINKED ISSUE #6521: Deprecate ToolEvents
The `ToolEvent` class does not work well at all. It should be replaced with new style events from `bokeh.events` that tools can generate when appropriate. 

Adding this to `0.12.7` because we should try to mark all planned deprecations by `0.12.7` if possible. 

ping @jlstevens

canavandl:
- [x] issues: fixes #6521 
- [x] tests added / passed


bokehjs/src/coffee/models/tools/gestures/select_tool.coffee
```diff
@@ -59,16 +60,7 @@ export class SelectToolView extends GestureToolView
...
-      tool_events = @plot_model.plot.tool_events
-      if append
-        geoms = tool_events.geometries
-        geoms.push(g)
-      else
-        geoms = [g]
-
-      tool_events.geometries = geoms
-    return null
+    @plot_model.plot.trigger_event(new Selection({geometry: g}))
```
bryevdv: We might want to include `final` in the event? That would provide a way for event handlers to filter and ignore intermediate selection geometries if desired.

canavandl: that's a good idea. I'll add it now.

bryevdv:
@canavandl is `Selection` the best name? I think it might be better to be more specific `SelectionGeometry` or something that conveys that it conveys the geometry information (and not e.g. the selection indicies)

PR #6589, opened 2017-07-06
Enabling server example tests experiment

### LINKED ISSUE #3935: How to get the server examples testable again
The server examples tests don't currently work because the test doesn't know the session id that the server plot ends up running.

---
### Proposal 1: specify sessionid

To make them run, we could change code like:

``` python
```

to be:

``` python
session = push_session(curdoc(), sessionid="name of file")
```

This would be similar to a constraint we impose on the file examples, and means we would know where to look.

However, it has the massive downsides of:
- documenting a bad habit (specifying the sessionid isn't good practice)
- not working for all examples

---
### Proposal 2: add a --test option to bokeh serve

This would monkey patches the sessionid to be the same as the filepath of the example.

This has the disadvantages of:
- being more work than Proposal 1
- adding testing infrastructure into our main code

But it seems like a clean route

---

canavandl:
Enabling server example tests experiment

- [ ] issues: fixes #3935 
- [ ] tests added / passed

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -565,7 +565,7 @@ def _destroy_server(div_id):
     except Exception as e:
         logger.debug("Could not destroy server for id %r: %s" % (div_id, e))
 
-def _wait_until_render_complete(driver):
+def _wait_until_render_complete(driver, timeout=5, poll_frequency=0.1):
     from selenium.webdriver.support.ui import WebDriverWait
     from selenium.common.exceptions import TimeoutException
 
@@ -576,12 +576,15 @@ def _wait_until_render_complete(driver):
       window._bokeh_render_complete = true;
     }
 
-    var doc = window.Bokeh.documents[0];
+    var documents = window.Bokeh.documents;
+    if (documents.length > 0) {
+      var doc = documents[0];
 
-    if (doc.is_idle)
-      done();
-    else
-      doc.idle.connect(done);
+      if (doc.is_idle)
+        done();
+      else
+        doc.idle.connect(done);
+    }
     """
     driver.execute_script(script)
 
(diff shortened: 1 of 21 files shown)


tests/examples/test_examples.py
```diff
@@ -214,6 +214,11 @@ def _run_example(example):
 import warnings
 warnings.filterwarnings("ignore", ".*", UserWarning, "matplotlib.font_manager")
 
+def static_seed(*args, **kwargs):
+    return "abc"
+from bokeh.util import session_id
+session_id._get_random_string = static_seed
```
canavandl: This patches the ``bokeh.util.session_id._get_random_string`` at runtime to return a deterministic string that's internally used as the session_id.

bokeh/io.py
```diff
@@ -576,12 +576,14 @@ def _wait_until_render_complete(driver):
       window._bokeh_render_complete = true;
     }
 
-    var doc = window.Bokeh.documents[0];
+    if (window.Bokeh) {
```
canavandl: Some of the server apps load slowly, so sometimes the page is loaded before Bokeh is appended to the window. There's probably a better way to handle this.

tests/examples/test_examples.py
```diff
@@ -62,31 +68,38 @@ def test_file_examples(file_example, example, report):
...
+    app = build_single_handler_application(example.path)
+    doc = app.create_document()
 
-    if example.no_js:
-        if not pytest.config.option.no_js:
-            warn("skipping bokehjs for %s" % example.relpath)
-    else:
-        _assert_snapshot(example, url, 'server')
+    session_id = "session_id"
+    push_session(doc, session_id=session_id)
```
canavandl: I'm not sure the benefit of pushing the document into a running bokeh server process vs just saving/showing the document directly. Since we just taking a screenshot, there's not much benefit to having a server backed, "live" document versus a static one other than some implicit integration-y testing of ``push_session``. Thoughts?

tests/examples/test_examples.py
```diff
@@ -62,31 +68,38 @@ def test_file_examples(file_example, example, report):
...
-        if example.no_diff:
-            warn("skipping image diff for %s" % example.relpath)
-        else:
-            _get_pdiff(example)
+    webdriver.get("http://localhost:5006/?bokeh-session-id=%s" % session_id)
+    webdriver.set_window_size(1200, 800)
+
+    _wait_until_render_complete(webdriver, timeout=20)
+
+    assert (has_no_console_errors(webdriver))
```
canavandl: This is cruft and should be removed. ``_wait_until_render_complete`` checks the browser logs already, though it might be better if that's broken out.

canavandl:
This gets the server examples to work by using the ``build_single_handler_application`` functionality to create an Application instance and return a document instead of exec-ing the file via a subprocess.

One win is that this eliminates the logging handing from stdout because we're executing the doc in the main process any exceptions are raised there as normal exceptions.

Another win is that we're screenshotting the layout via selenium instead of a subprocess call to the phantomjs executable. This allows us to do examples testing with other backends by making the webdriver instance configurable, even potentially many backends via SauceLabs.

bryevdv:
@canavandl what was the rationale for also changing the style of the examples in `models/server`? until now all the "app" examples e.g `bokeh serve app.py` were in `examples/app` and the examples in `models/server` and `plotting/server` were specifically for `bokeh.client` style usage.

canavandl:
> @canavandl what was the rationale for also changing the style of the examples in models/server? until now all the "app" examples e.g bokeh serve app.py were in examples/app and the examples in models/server and plotting/server were specifically for bokeh.client style usage.

I didn't realize that was the rationale for writing the examples that way. I thought it was boilerplate so that the example tests could run. I think the ScriptHandler strategy should still work with the examples written in the previous way. I'll revert and check.

canavandl:
superceded by #6705

PR #6595, opened 2017-07-10
Add support for image diff in headless chrome

### LINKED ISSUE #6594: Add support for headless chrome/chromium
At some point this should replace phantomjs in our image diff tests.

mattpap:
This isn't going to work right now, because it's not yet hooked into testing infrastructure, and we need newer version of nodejs available as a conda package first. To play with this locally, install most recent google chrome (chromium won't work) and nodejs 8.1.x, and then issue:
```
/opt/google/chrome/chrome --headless --remote-debugging-port=9222 & # or in another console
node bokehjs/headless.js examples/models/file/anscombe.html
```
This will generate `output.png` with rendered plots. You can visit `localhost:9222` for remote debugger.

EDIT: `--disable-gpu` is needed only on Windows at this point.

fixes #6594


canavandl:
While trying to get the server examples working I generated a competing strategy for image diffing by using Bokeh's ScriptHandlers and selenium ( #6589 ) that could be extended to all of the example types.

I'll comment later, but we should consider the pros and cons of using selenium to drive a browser vs calling chrome headless via a subprocess.

bryevdv:
@mattpap what is the rationale for not dog-fooding our own built-in export capability. I would much prefer to do that, barring some exceptional reason not to.

mattpap:
@bryevdv, it may be possible to have the same code for static image generation and testing, or at least a common core, but it may also be not desirable. Testing is not only about generating pretty images, but more importantly about dealing with failure. I need to know what failed and why (see integration tests and how fun it is to work with when things fail). If this can be done with selenium, sure we can proceed this way, but I doubt it. After all, selenium has to implement the least common value of all backends.

Here I experiment with Chrome Debug Protocol, and this PR is more about it, than about switching to headless chrome (I would do this anyway, even if I had to use xvfb). With CDP we have access to chrome devtools, aka. complete debugger, so we have new possibilities open. We could do network traffic or memory usage monitoring, run our code with different settings (e.g. throttling, to see how it behaves under different kinds of load), or maybe even use debugger to dump stack frames in case of exceptions being thrown. I don't know how much this is doable or feasible, but it's worth exploring. At this point I don't want to be limited by unrelated stuff.

bryevdv:
@mattpap your point about additional capabilities over and above images for other kinds of testing is well-taken, but an approach using selenium *for things that can be covered that way* is also appealing, both for the possibility of using many actual browsers at some point, but also for some to simply make sure selenium usage is well-exercises constantly. https://github.com/bokeh/bokeh/pull/6589/ adds minimal support testing server examples, which I would regard as highly valuable, so I think my inclination wrt to the two PRs would be "both" but on different time scales: finish the other PR in the short term, and find areas and ways in which chrome headless and CDP can off additional specific benefits and use those (later) where appropriate.

Thoughts (both of you)?

canavandl:
I think there's a middle road here where we could use selenium (and perhaps SauceLabs) to do image diffing to help solve our problem of being blindsided by platform-specific issues.

I like the concept of using CDP for writing real integration tests - I agree that our current integration test infra is difficult to extend or debug and CDP could be a real improvement there.

mattpap:
Another benefit of using chrome directly would be the ability to use emulation mode and test painting under different pixel ratios (per issue #6840).

mattpap:
Image differences are due to chrome having different defaults (fonts in particular).

bryevdv:
@mattpap do we still need to install phantomjs at all then? What is it still used for?

PR #6602, opened 2017-07-11
GH6601 Added major_label_overrides to ColorBar

### LINKED ISSUE #6601: Add major_label_overrides to ColorBar
axes already support the use of major_label_overrides that use a dictionary to change major tick labels.  It would be nice if ColorBar objects supported the same.  This could close out #5618

js3711:
First PR.

==1168 tests deselected ==
== 4 passed, 2 skipped, 1168 deselected, 1 warnings in 27.02 seconds ==

- [X] issues: fixes  #6601
- [X] tests added / passed
- [X] release document entry (if new feature or API change)

# Useful Test Code:
```
## Imports:
import bokeh.plotting as bkp
from bokeh.models import ColorBar, FixedTicker, Range1d, LinearColorMapper
from bokeh.palettes import viridis

import numpy as np

output_notebook(INLINE)

## Create data
ar = np.zeros((20,20))
ar[10:20, 10:20] = 1
ar[0:10, 10:20] = 2
ar[10:20, 0:10] = 3

## Plotting
plt = bkp.figure()

ticks = np.array([0.5, 1.5, 2.5, 3.5])
color_mapper = LinearColorMapper(palette=viridis(4), low=0, high=4)

color_bar = ColorBar(color_mapper=color_mapper, ticker = FixedTicker(ticks=ticks),
                     label_standoff=5, border_line_color=None, location=(0,0),
                     major_label_overrides={0.5:'a', 1.5:'b', 2.5:'c', 3.5:'d'})

plt.image([ar], x=0, y=0, dh=1, dw=1, color_mapper=color_mapper)
plt.add_layout(color_bar, 'right')

plt.x_range = Range1d(0,1)
plt.y_range = Range1d(0,1)

show(plt)
```


bryevdv:
@canavandl what are your thoughts? I guess in an ideal world I'd prefer that color bars *have` an `Axis`, and all the axis formatting, etc. flows from that. But maybe at this point it'd more realistic to simple try to share some of the implementation in a few helper functions rather than trying to share an actual Axis? If so I think we could proceed with this approach.

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -515,6 +522,7 @@ export class ColorBar extends Annotation
       minor_coords[i] = new Float64Array((scale_length - coord for coord in minor_coords[i]))
 
     return {
+      "major_ticks": major_ticks
       "major": major_coords
       "minor": minor_coords
       "major_labels": major_labels
```
bryevdv: This is getting a little cluttered. I think if we are touching this code it would be a good time to clean it up a bit with some light refactoring, e.g. perhaps return instead
```
return {
  ticks: ticks
  coords: coords
  labels: labels
```

Where each of the above has `.major` and (optionally) `.minor` as attributes. 

I think it's also worth considering applying the formatting (including label overrides) here, in this function, rather than in the view. In any case the view is calling `_tick_coordinates()` twice! We should definitely fix that since it's unnecessarily wasteful. I also think these methods could be made public.

canavandl:
> @canavandl what are your thoughts? I guess in an ideal world I'd prefer that color bars *haveanAxis`, and all the axis formatting, etc. flows from that. But maybe at this point it'd more realistic to simple try to share some of the implementation in a few helper functions rather than trying to share an actual Axis? If so I think we could proceed with this approach.

When creating the ColorBar model, I attempted to use a real Axis but found that it's implementation was too tied to being attached to along a plot side/have a plot range. So I went with the current implementation with the hopes of being "axis=like", so this approach seems to fit. It's disappointing to have to remember to play catch-up with any Axis model changes, but I don't know if there's a better alternative unless Legends become real layout-able models.

bryevdv:
Ok that's what I thought. I think we can proceed with this and plan to see what functions can be factored out later. @js3711 will you be able to also do some of the light refactoring I outlined above?

bokehjs/src/coffee/models/annotations/color_bar.coffee
canavandl: I agree about doing the label overriding here in the model and then adding a JS test to ensure it works/doesn't break in the future.

bryevdv: Alternatively, slice the other dimension:
```
return {
  major: major
  minor: minor
}
```
Where `major` and `minor` each have `.ticks` `.labels` and `.coords`

js3711:
@bryevdv sure, I'll take a stab at it

bryevdv:
@js3711 that's great, let me know if I can provide any additional guidance or help

js3711:
@bryevdv I updated the pull request with the refactoring and updated the tests.  Let me know what you think.

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -203,7 +203,8 @@ export class ColorBarView extends AnnotationView
     image = @model._computed_image_dimensions()
     [x_offset, y_offset] = [image.width * nx, image.height * ny]
 
-    [sx, sy] = @model._tick_coordinates().major
+    tick_coords = @model._tick_coordinates()
```
bryevdv: `@model._tick_coordinates()` looks to be called at least three times. I think this could be improved:
* rename `_tick_coordinates` to `tick_info` (no underscore, update name to reflect changes)
* call `tick_info()` once,  before `_draw_major_ticks`, etc, and pass the info to these methods instead of having them each call `tick_info()` separetely, e.g. `@_draw_major_ticks(ctx, tick_info)`

bokehjs/src/coffee/models/annotations/color_bar.coffee
```diff
@@ -337,6 +336,7 @@ export class ColorBar extends Annotation
       scale_alpha:    [ p.Number,         1.0         ]
       ticker:         [ p.Instance,    () -> new BasicTicker()         ]
       formatter:      [ p.Instance,    () -> new BasicTickFormatter()  ]
+      major_label_overrides:   [ p.Any,      {}           ]
```
bryevdv: picky, but these blocks should all be lined up

bryevdv:
@js3711 looks good! I have one more improvement suggestion, if you have bandwidth to work on it. 

Otherwise, there are two test issues:

The first is a failing JS test:
```
  1) ColorBar module ColorBarView ColorBarView._get_label_extent method (orientation='vertical') and no major_labels:
     TypeError: Cannot read property 'major' of undefined
      at ColorBarView.exports.ColorBarView.ColorBarView._get_label_extent (build/js/tree/models/annotations/color_bar.js:327:42)
      at Context.<anonymous> (test/models/annotations/color_bar.coffee:400:36)
```

The second is a python test with a trivial fix:

```
bokeh/models/tests/test_annotations.py:113: 
model = ColorBar(id='330d7f4a-020a-43bb-b6e6-69e9dcedb010', ...)
props = (['plot', 'level', 'visible', 'location', 'orientation', 'height', ...], ['title_text_font', 'title_text_font_size', '....], ['bar_line_color', 'bar_line_width', 'bar_line_alpha', 'bar_line_join', 'bar_line_cap', 'bar_line_dash', ...], ...)
expected = set(['background_fill_alpha', 'background_fill_color', 'bar_line_alpha', 'bar_line_cap', 'bar_line_color', 'bar_line_dash', ...])
found = set(['background_fill_alpha', 'background_fill_color', 'bar_line_alpha', 'bar_line_cap', 'bar_line_color', 'bar_line_dash', ...])
missing = set([])
    def check_properties_existence(model, *props):
        expected = set(chain(PROPS, *props))
        found = set(model.properties())
        missing = expected.difference(found)
        extra = found.difference(expected)
        assert len(missing) == 0, "Properties missing: {0}".format(", ".join(sorted(missing)))
>       assert len(extra) == 0, "Extra properties: {0}".format(", ".join(sorted(extra)))
E       AssertionError: Extra properties: major_label_overrides
```
The new `major_label_overrides` property needs to be added to the test that checks the expected properties in the test at`bokeh/models/tests/test_annotations.py:113`

js3711:
@bryevdv Thank you for the feedback, it is much appreciated.  I just made the requested changes and pushed; let me know what you think.

I have two process type questions for you (if you have time to answer):
1) How did you catch the python test?  Is there something I should be doing to run those tests before I push?
2)  Do you have any suggestions for a free coffeescript or javascript IDE? How do you work across the Python -> Coffeescript -> Javascript boundaries?

bryevdv:
@js3711 you can see the status of the automated test on Travis by clicking the "Details" 
link in the PR:

screen shot 2017-07-16 at 19 40 24

The various tests can also be run locally, with more or less difficulty (the python unit tests are pretty easy to get running). The Developer's guide has info about that.

I use SublimeText for my editor which is nagware unless you pay for it. But lots of people seem to like Atom which is very similar, and I have also been meaning to try VS.CODE which is based on Atom but apparently much faster.

bryevdv:
looks geat, thanks for the PR @js3711 !

screen shot 2017-07-16 at 19 59 58

PR #6448, opened 2017-06-12
Review and improve layout constraints

mattpap:
This is an attempt to fix issues like #4972, where constraints are badly created, resulting in collapsed plots or "unsatisfiable constraint" error.

fixes #4972
fixes #4599 
fixes #6578

PR #6604, opened 2017-07-11
6592 webgl save

### LINKED ISSUE #6592: Save tool not functioning for webgl backend figures 0.12.6 regression
Currently when using the save tool with a plot that has been generated with WebGL nothing will happen.  The image will not be saved and no errors are reported in the console. 

This is a depreciation from 0.12.5 where the image saving works correctly.

There's also an error in the current webgl docs. The first example is used to illustrate the plot when not using webgl, but the code includes webgl in it.

```
output_file("scatter10k.html", title="scatter 10k points (no WebGL)")

p = figure(output_backend="webgl")
```

Go to http://bokeh.pydata.org/en/latest/docs/user_guide/webgl.html and the save button does not work for any of the webgl plots.
```
import numpy as np

N = 10000

x = np.random.normal(0, np.pi, N)
y = np.sin(x) + np.random.normal(0, 0.2, N)

output_file("scatter10k.html", title="scatter 10k points (with WebGL)")

p = figure(output_backend="webgl")
p.scatter(x, y, alpha=0.1)
show(p)
```

If we got back to the 0.12.5 docs http://bokeh.pydata.org/en/0.12.5/docs/user_guide/webgl.html the save button works without an issue.

```
import numpy as np

N = 10000

x = np.random.normal(0, np.pi, N)
y = np.sin(x) + np.random.normal(0, 0.2, N)

output_file("scatter10k.html", title="scatter 10k points (with WebGL)")

p = figure(webgl=True)
p.scatter(x,y, alpha=0.1)
show(p)
```

Browser: Chrome 59.0.3071.115
OS: Windows 7

DuCorey:
Fix save tool when `output_backend= "webgl"`.
Fix webgl docs for the example without webgl.

- [x] issues: fixes #6592 
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -663,7 +663,7 @@ export class PlotCanvasView extends DOMView
       ctx.fillRect(frame_box...)
 
   save: (name) ->
-    if @model.plot.output_backend == "canvas"
+    if @model.plot.output_backend in ["canvas", "webgl"]
       canvas = @canvas_view.get_canvas_element()
       if canvas.msToBlob?
         blob = canvas.msToBlob()
--- a/bokehjs/test/size.ts
+++ b/bokehjs/test/size.ts
@@ -7,7 +7,7 @@ const build_dir = path.normalize(`${__dirname}/../build`)
 
 const LIMITS: {[key: string]: number} = {
   // js
-  "js/bokeh.min.js":           730,
+  "js/bokeh.min.js":           732,
   "js/bokeh-widgets.min.js":    90,
   "js/bokeh-tables.min.js":    200,
   "js/bokeh-api.min.js":        75,
--- a/sphinx/source/docs/user_guide/webgl.rst
+++ b/sphinx/source/docs/user_guide/webgl.rst
@@ -67,7 +67,7 @@ Examples
 
     output_file("scatter10k.html", title="scatter 10k points (no WebGL)")
 
-    p = figure(output_backend="webgl")
+    p = figure(output_backend="canvas")
     p.scatter(x, y, alpha=0.1)
     show(p)
 


bryevdv:
Looks like this tiny addition was enough to bump past our current file size limits:
```
 1) bokehjs/build/*/*.min.{js,css} file sizes js/bokeh.min.js file size should be 731 kB <= 730 kB:
     AssertionError: expected 748532 to be below 747520
      at Context.<anonymous> (test/size.ts:28:34)
```
So that will need to be adjusted before tests will pass. Can you increase the limit to 732k here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/test/size.ts#L10

bryevdv:
@mattpap Is there an easy way to identify what modules are contributing the most to the final built and minified file? I'm surprised after excising jq and backbone and widgets and webgl that things are still this big. Is this an artifact of our granular modules? At this stage I'd consider some consolidation if it offered enough saving.

bryevdv:
thanks for the PR @DuCorey will merge when green

mattpap:
@bryevdv, funny you ask, because today I reduced bokeh.min.js from 729 kB to 659 kB, by replacing browserify with my own bundler/linker (this work will be part of PR #6585).

> Is there an easy way to identify what modules are contributing the most to the final built and minified file?

Currently we don't have any tooling for inspecting contributions of particular modules. At most you can use `--verbose` flag to see what modules are included in each bundle, but without sizes. It's interesting to see that such insignificant change made a significant contribution. I think that some dependencies changed in the mean time and they did contribute. We have to consider using exact versions of our runtime dependencies.

> I'm surprised after excising jq and backbone and widgets and webgl that things are still this big. Is this an artifact of our granular modules?

Don't forget where we were starting from. Also, bokehjs is growing all the time (e.g. addition of canvas2svg). Maybe TypeScript is contributing more that CoffeeScript (I didn't research that)? bokehjs still has heavy dependencies, like kiwi, and there is still a lot of code duplication (e.g. in PR #6448 I removed a dozen of duplicates of the same layout code; annotations are a mess with code being copied left and right).

PR #6609, opened 2017-07-13
fix localhost for autoload_server in io.py

### LINKED ISSUE #6598: Server url hard coded as localhost. With possible fix.
bokeh: 0.12.6
python: 3.6
notebook: 5.0.0
jupyter: 1.0.0
jupyter_client 5.0.1
jupyter_console: 5.1.0
jupyter_core: 4.3.0
ipython: 6.0.0
ipython_genutils: 0.2.0
OS: ubuntu 16.04
brower: Chrome 58.0.3029.110 (64-bit)
What I tried to do was running a jupyter server on one computer and then opening up notebook_embed.ipynb on another computer through internet. The `show(app)` didn't show any figure.

This is caused by this line `script = autoload_server(url='http://127.0.0.1:%d%s' % (server.port, app_path))`  in the function `_show_notebook_app_with_state`. 

This is fine when jupyter notebook is run on localhost. But suppose now my server ip is 10.0.1.1 and client ip is 10.0.1.2. When this line is called, it creates js code on server (10.0.1.1). The code is sent to the client (10.0.1.2). When the browser on client sees this js code, it tries to fetch the figure from 127.0.0.1 instead of the server (10.0.1.1).

So I fixed the bug by changing the line to
```
script = autoload_server(url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path))
```
In this case, you can specify the server ip by calling `show(app, notebook_url='10.0.1.1:8888')`. Now the browser on the client computer correctly sees to the bokeh server is on the server ip.

enfeizhan:
- 1 issue: fixes #6598
- 0 tests added / passed
- 0 release document entry (if new feature or API change)

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -303,7 +303,7 @@ def _show_notebook_app_with_state(app, state, app_path, notebook_url):
     loop = IOLoop.current()
     server = Server({app_path: app}, io_loop=loop, port=0,  allow_websocket_origin=[notebook_url])
     server.start()
-    script = autoload_server(url='http://127.0.0.1:%d%s' % (server.port, app_path))
+    script = autoload_server(url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path))
     display(HTML(_server_cell(server, script)))
 
 def _show_with_state(obj, state, browser, new, notebook_handle=False):
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -303,7 +303,8 @@ def _show_notebook_app_with_state(app, state, app_path, notebook_url):
     loop = IOLoop.current()
     server = Server({app_path: app}, io_loop=loop, port=0,  allow_websocket_origin=[notebook_url])
     server.start()
-    script = server_document(url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path))
+    url = 'http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path)
+    script = server_document(url)
     display(HTML(_server_cell(server, script)))
 
 def _show_with_state(obj, state, browser, new, notebook_handle=False):
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -28,7 +28,7 @@ import uuid
 # Bokeh imports
 from .core.state import State
 from .document import Document
-from .embed import autoload_server, notebook_div, file_html
+from .embed import server_document, notebook_div, file_html
 from .layouts import gridplot, GridSpec ; gridplot, GridSpec
 from .models import Plot
 from .resources import INLINE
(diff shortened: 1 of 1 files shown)


bokeh/io.py
```diff
@@ -303,7 +303,7 @@ def _show_notebook_app_with_state(app, state, app_path, notebook_url):
     loop = IOLoop.current()
     server = Server({app_path: app}, io_loop=loop, port=0,  allow_websocket_origin=[notebook_url])
     server.start()
-    script = autoload_server(url='http://127.0.0.1:%d%s' % (server.port, app_path))
+    script = autoload_server(url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path))
```
bryevdv: `autoload_server` is actually deprecated, split into `server_document` and `server_session`. This should have been updated earlier, but was overlooked. This PR would be a good place to change it to use the simpler `server_document(app_url)`

bryevdv:
@enfeizhan I only had one suggestion, will you have the opportunity to make that change?

bokeh/io.py
enfeizhan: Have replaced autoload_server with server_document imported from embed.py as I understand that 's what you meant?

enfeizhan:
Have replaced autoload_server with server_document imported from embed.py as I understand that 's what you meant?

bokeh/io.py
```diff
@@ -303,7 +303,7 @@ def _show_notebook_app_with_state(app, state, app_path, notebook_url):
     loop = IOLoop.current()
     server = Server({app_path: app}, io_loop=loop, port=0,  allow_websocket_origin=[notebook_url])
     server.start()
-    script = autoload_server(url='http://127.0.0.1:%d%s' % (server.port, app_path))
+    script = server_document(url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path))
```
bryevdv: very minor thing but one of the reasons to split `autoload_server` was so that it could be typically called with a single positional arg, also for debugging it's often nice to save computed args like this first, in case anyone ever needs to single step through the code. I'd suggest this:
```
url='http://%s:%d%s' % (notebook_url.split(':')[0], server.port, app_path)
script = server_document(url)
```

bryevdv:
@enfeizhan awesome thanks, I had one more small comment that should be a quick edit. 

Otherwise, have you had a chance to test this change locally and  remotely? This will be hard to have unit tests for so it would be good to attach screenshots of working local and remote usage to a comment here as documentation that the change is working as expected.

enfeizhan:
@bryevdv have had the small change as you pointed out.

Sorry I know this might sound unprofessional. I believe the change works as I have been using the fixed version in my actual work. That change was made on the codes installed through conda. However I am quite struggling to create a dev env with bokeh installed from local source code as my skill set mainly encompass data analysis rather than software development.

bryevdv:
>  I believe the change works as I have been using the fixed version in my actual work.

That's really what I was looking for, confirmation that the code has been exercised. Thanks for the PR @enfeizhan 

Re getting setup, until the next release, the best current info is available here:

http://bokeh.pydata.org/en/dev/docs/dev_guide/setup.html

I've tried to streamline things considerably but let me know if anything is confusing or could be improved.

PR #6622, opened 2017-07-15
Prevent user select of toolbar

### LINKED ISSUE #4295: Investigate disabling user-select
Recently, sometimes Bokeh plots get annoyingly "selected" i.e. are blued out like selected text. I believe this is because user-select needs to be disabled, something like:

```
this.$el.css({
  "user-select": "none",
  "ms-user-select": "none",
  "moz-user-select": "none",
  "khtml-user-select": "none",
  "webkit-user-select": "none"});
```

dkatten:
Fixes #4295. 

Tested against the lasso tool in color_scatter example (per issue), and tests to select within the toolbar and its buttons passed manually.

Please let me know if we're able to see *any* places to select text/elements that I might have missed.

- [x] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Thanks @dkatten !

PR #6624, opened 2017-07-15
Issue6577_make stream setter accessible only in the private API

### LINKED ISSUE #6577: ColumnDataSource.stream() setter argument is not documented
I'm trying to figure out how to store state in my [callback()](https://stackoverflow.com/questions/44841436/how-to-store-and-retrieve-state-in-a-callback) method so that I can maintain a pointer to the current location in my source database.  I've been looking through the bokeh source code and I noticed an argument that looks interesting, but it isn't documented:

```
    def stream(self, new_data, rollover=None, setter=None):
        ''' Efficiently update data source columns with new append-only data.
        In cases where it is necessary to update data columns in, this method
        can efficiently send only the new data, instead of requiring the
        entire data set to be re-sent.
        Args:
            new_data (dict[str, seq]) : a mapping of column names to sequences of
                new data to append to each column.
                All columns of the data source must be present in ``new_data``,
                with identical-length append data.
            rollover (int, optional) : A maximum column size, above which data
                from the start of the column begins to be discarded. If None,
                then columns will continue to grow unbounded (default: None)
        Returns:
            None
        Raises:
            ValueError
```

https://github.com/bokeh/bokeh/blob/master/bokeh/models/sources.py#L211

gregfreeman:
- [x] issues: fixes #6577 
- [x] tests added / passed

Addresses issue 6577 by making the stream setter argument only accessible in a private method (_stream).

test__stream_good_data  tests the private method and 
test_stream_good_data tests the public method.

Tested on linux using 
py.test -m "unit and not selenium"
999 passed

### DIFF
--- a/bokeh/document.py
+++ b/bokeh/document.py
@@ -372,7 +372,7 @@ class Document(object):
                 source = self._all_models[source_id]
                 data = event_json['data']
                 rollover = event_json['rollover']
-                source.stream(data, rollover, setter)
+                source._stream(data, rollover, setter)
             elif event_json['kind'] == 'ColumnsPatched':
                 source_id = event_json['column_source']['id']
                 if source_id not in self._all_models:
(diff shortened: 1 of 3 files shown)


bryevdv:
This PR is `ready` the broken streaming issue is due to #6628 (should be fixed first)

bryevdv:
Thanks @gregfreeman !

PR #6630, opened 2017-07-15
Add metadata support to bokeh server applications

### LINKED ISSUE #6621: Add metadata support to bokeh server
The goal here is to add a new <app>/metadata endpoint to each application running on the bokeh server.  This will return a json dictionary which contains at a minimum a 'name' but can also contain:

* description (markup?) 
* description_html

Other possibilities:

API Change:

bokeh.application.Application will take an additional optional parameter: metadata.

metadata can either be a dict which will be returned as part of metadata.  It can also be a function which will return a dict.

A default /metadata will be created for every application even if no user metadata is provided.

yerfdogyrag:
- [ ] issues: fixes #6621 
- [ ] tests added / passed


bokeh/application/application.py
```diff
@@ -100,9 +101,23 @@ class Application(object):
     # will function without bringing in tornado.
     _is_a_bokeh_application_class = True
 
-    def __init__(self, *handlers):
+    def __init__(self, *handlers, metadata=None):
```
bryevdv: Kwargs-only args do not wok with Python2, we will need to find another way to express this, perhaps a Python getter/setter for a `metadata`property

bryevdv:
One test is failing on (some) python 3 versions, not sure offhand:
```
________________________________ test_metadata _________________________________
    def test_metadata():
        application = Application(metadata=dict(hi="hi", there="there"))
        with ManagedServerLoop(application) as server:
            t = tornado.BokehTornado({}, "", [])
            meta_url = url(server) + 'metadata'
            meta_resp = http_get(server.io_loop, meta_url)
>           meta_json = json.loads(meta_resp.buffer.read())
bokeh/server/tests/test_tornado.py:124: 
s = b'{"url": "/", "data": {"there": "there", "hi": "hi"}}', encoding = None
cls = None, object_hook = None, parse_float = None, parse_int = None
parse_constant = None, object_pairs_hook = None, kw = {}
    def loads(s, encoding=None, cls=None, object_hook=None, parse_float=None,
            parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
        """Deserialize ``s`` (a ``str`` instance containing a JSON
        document) to a Python object.
    
        ``object_hook`` is an optional function that will be called with the
        result of any object literal decode (a ``dict``). The return value of
        ``object_hook`` will be used instead of the ``dict``. This feature
        can be used to implement custom decoders (e.g. JSON-RPC class hinting).
    
        ``object_pairs_hook`` is an optional function that will be called with the
        result of any object literal decoded with an ordered list of pairs.  The
        return value of ``object_pairs_hook`` will be used instead of the ``dict``.
        This feature can be used to implement custom decoders that rely on the
        order that the key and value pairs are decoded (for example,
        collections.OrderedDict will remember the order of insertion). If
        ``object_hook`` is also defined, the ``object_pairs_hook`` takes priority.
    
        ``parse_float``, if specified, will be called with the string
        of every JSON float to be decoded. By default this is equivalent to
        float(num_str). This can be used to use another datatype or parser
        for JSON floats (e.g. decimal.Decimal).
    
        ``parse_int``, if specified, will be called with the string
        of every JSON int to be decoded. By default this is equivalent to
        int(num_str). This can be used to use another datatype or parser
        for JSON integers (e.g. float).
    
        ``parse_constant``, if specified, will be called with one of the
        following strings: -Infinity, Infinity, NaN.
        This can be used to raise an exception if invalid JSON numbers
        are encountered.
    
        To use a custom ``JSONDecoder`` subclass, specify it with the ``cls``
        kwarg; otherwise ``JSONDecoder`` is used.
    
        The ``encoding`` argument is ignored and deprecated.
    
        """
        if not isinstance(s, str):
            raise TypeError('the JSON object must be str, not {!r}'.format(
>                               s.__class__.__name__))
E           TypeError: the JSON object must be str, not 'bytes'
```

bryevdv:
There are also a couple of linter errors to fix up:
```
    @pytest.mark.quality
    def test_flake8():
        chdir(TOP_PATH)
    
        out, err = proc.communicate()
    
E         ./bokeh/server/tests/test_tornado.py:133:9: F841 local variable 't' is assigned to but never used
E         ./bokeh/server/views/metadata_handler.py:27:9: F841 local variable 'session' is assigned to but never used
E         
E       assert 1 == 0
E        +  where 1 = <subprocess.Popen object at 0x7f94e0c91ef0>.returncode
```

bryevdv:
ping @havocp just in case you have any general comments you'd like to add. Intention is to allow apps to be configured with an arbitrary chunk of metadata that could be accessed via an endpoint. another PR will add an optional endpoint to list existing/running app endpoints. This will allow a minimal amount of introspection so that system that want to integrate different apps from a single server can query a server for purposes of embedding or providing navigation to whatever is available.

yerfdogyrag:
Eep - OK, let me also try and figure out why I didn't get notices on the Travis fail.

bokeh/application/application.py
yerfdogyrag: Are you OK if I go:
 def __init__(self, *handlers, **kwds):
       metadata = kwds.get('metadata', None)

?

bryevdv: Sure that's also fine, we do something similar in many places. The nicest version would check for any additional unexpected kwargs and raise an exception.

bryevdv:
@mattpap @canavandl any comments? If not I will plan to merge later today.

bokeh/server/views/metadata_handler.py
```diff
@@ -0,0 +1,37 @@
...
+from tornado import gen
+
+from .session_handler import SessionHandler
+
+class MetadataHandler(SessionHandler):
+    ''' Implements a custom Tornado handler for document display page
+
+    '''
+    def __init__(self, tornado_app, *args, **kw):
+        super(MetadataHandler, self).__init__(tornado_app, *args, **kw)
```
bryevdv: I think this should be deleted if it doesn't do anything other than call `super`

bokeh/server/views/metadata_handler.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def initialize(self, *args, **kw):
+        pass
```
bryevdv: similar, I think think this can just be deleted (but maybe some implementation is needed... worth checking to find out)

bokeh/server/views/metadata_handler.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @gen.coroutine
+    def get(self, *args, **kwargs):
+        url = self.application_context.url
+        userdata = self.application_context.application.metadata
+        if callable(userdata):
+            userdata = userdata()
+        if userdata is None:
+            userdata = {}
```
bryevdv: Is there a specific use-case for deferring the execution to this point when `userdata` is callable? I.e., is there a reason not to do it once, in the `Applictication` initializer?

sphinx/source/docs/dev_guide/server.rst
```diff
@@ -31,9 +31,10 @@ widgets, or other content it desires. The application can also set up
 callbacks, to run periodically or to run when the document changes.
 
 Applications are represented by the ``Application`` class. This class is
-little more than a list of ``Handler`` instances. Handlers can be created
-in lots of ways; from JSON files, from Python functions, from Python files,
-and perhaps many more ways in the future.
+contains list of ``Handler`` instances and optional metainformation. Handlers
+can be created in lots of ways; from JSON files, from Python functions, from
+Python files, and perhaps many more ways in the future.  The optional metadata
+is available as a json blob via the ``/metadata`` endpoint.
```
bryevdv: probably worth giving an example JSON output, with the fixed `url` and `data` fields, to demonstrate where exactly the user supplied metadata goes (in the `data` value)

bokeh/application/application.py
```diff
@@ -100,9 +101,26 @@ class Application(object):
...
+        Application factory.
+
+        Args:
+            handlers (`bokeh.server.views.doc_handler.DocHandler`):
+                List of handlers called.  The url is taken from the first one
+                only.
+
+            metadata (dict):
+                May be requested by http://applicationurl/metadata as a json
+                blob.
```
bryevdv: probably worth giving an example JSON output, with the fixed `url` and `data` fields, to demonstrate where exactly the user supplied metadata goes (in the `data` value)

bryevdv:
@yerfdogyrag I went though once more and left a few small comments, I do think there a few more small docs/cleanup thing that would be good to do if you have the bandwidth.

bokeh/server/views/metadata_handler.py
yerfdogyrag: *blush* - yup.

bokeh/server/views/metadata_handler.py
yerfdogyrag: Yes - I can see situations where the metadata can change at runtime (for instance, reporting last update time for the underlying data).

bokeh/server/views/metadata_handler.py
yerfdogyrag: Ah - I'll use the excuse that I got the code from DocHandler - I'll update in both places.

yerfdogyrag: Actually three places.  It looks like tornado.web.RequestHandler.initialize is just a pass, so it should be safe to delete.

yerfdogyrag: It doesn't look like it can be deleted.  The application_context is passed to it in some cases so it barfs if we don't have the **kw.

I went ahead and moved it to SessionHandler and deleted it from the other places.

bryevdv: seems good!

bokeh/application/application.py
yerfdogyrag: Just as a note: it looks like Sphinx doesn't currently show the __init__ documentation, so I'm going to move the docstring to the class docs.

yerfdogyrag:
Thanks for the comments.  As a note: I added an application.rst which will conflict with Maggie's in #6636 just to make sure it was clean.  I had to include all the handlers explicitly because of the import * stuff in __init__.py

It looks like Sphinx isn't configured to show the __init__() documentation.  In theory, it can be changed with a setting to concatenate the class documentation with the __init__ documentation.  Have y'all already gone down that path?  Should I try it?

bryevdv:
@yerfdogyrag I am going to collaborate on the other PR to expand the documentation, do you mind removing the `application.rst` from this PR to avoid the conflict?

yerfdogyrag:
np, done.

bryevdv:
Awesome, will merge when green!

bryevdv:
Thanks @yerfdogyrag !

Issue #6634, opened 2017-07-17
Add example of custom tooltip to example library


jordanfreedmanPM:
There's no current example with a custom tooltip in example library.

As requested at Scipy2017 sprint.

PR #6635, opened 2017-07-17
Custom Tooltip Example #6634.

jordanfreedmanPM:
Adds example to example plotting library utilising the custom tooltip.

image


bryevdv:
@jordanfreedmanPM This is a nice example! I realized I after we spoke I was a unclear in my description, it's actually custom formatters that are undemonstrated:

http://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#formatting-tooltip-fields

I think perhaps one could be added to this example, for example a format on the atomic mass that fixes the number of digits of precision, etc? What do you think?

jordanfreedmanPM:
@bryevdv Sure - sounds good. See updated version:

image

bryevdv:
Thanks @jordanfreedmanPM !

jordanfreedmanPM:
@bryevdv no problem. Let me know if there's any other functionality which is missing in the examples!

PR #6641, opened 2017-07-18
normalize tap tool callback calling convention

### LINKED ISSUE #6581: deprecate OpenURL?
`OpenURL` has slightly different semantics that standard callbacks, causing confusion and problems (e.g. https://github.com/bokeh/bokeh/issues/6580). I'd like to propose deprecating it in `0.12.7` and removing in `1.0` The change would be accompanied by documentation support in the user's guide showing how to do the same sort of thing with a standard `CustomJS` callback, and a deprecation message pointing there. 

ping @bokeh/dev

bryevdv:
- [x] issues: fixes #6581

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -321,8 +321,25 @@ class TapTool(Tap):
     """)
 
     callback = Instance(Callback, help="""
-    A client-side action specification, like opening a URL, showing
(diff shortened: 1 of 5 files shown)


mattpap:
@bryevdv, shouldn't migration tag be associated with issues?

bryevdv:
>  shouldn't migration tag be associated with issues?

Yes I will move them

PR #6645, opened 2017-07-19
Fix streaming with filterable CDS

### LINKED ISSUE #6628: Filterable CDS broke CDS streaming
ping @clairetang6 

Filterable CDS PR has broken CDS streaming, see e.g. OHLC example:

screenflow-2

Since auto-ranging is still working, it seems as thought the data is still making it to BokehJS in some form, but that rendering is somehow suppressed. 

Git bisection reveals Filterable CDS merge as first bad commit:

```
❯ git bisect good
90b941d7a5d8ed6fc3aa649a94bf568d001ff06b is the first bad commit
commit 90b941d7a5d8ed6fc3aa649a94bf568d001ff06b
Author: Claire Tang <clairetang6@gmail.com>
Date:   Mon Jul 3 10:48:35 2017 -0700

    Filterable and groupable CDS with new model CDSView (#5982)

    * filterable/groupable CDS using new model CDSView

    * allow filter to be Seq(Bool)

    * update test to add cds_view to plot object references

    * only call get_indices if a view's source is a ColumnarDataSource

    * make python and coffee defaults closer for new models

    * add some comments for mapping selection from cds subset to full set

    * merge master after selection with multi-renderers and change to signaling

    * added new filters, IndexFilter and BooleanFilter

    * add cdsview to datatable

    * have the indices emitted by the inspect signal be in the full cds space

    * make using views with data table more robust, and fix tests

    * rename selection converting from_subset or to_subset

    * add CustomJSFilter

    * fix 6433: indices['2d']['indices'] object needs to have Object prototype

    * only convert selections that aren't null

    * make cds_view work with Line glyphs

    * change filters to return null for no filtering, add warn logs

    * add tests for filters

    * add tests for cds_view

    * call glyphrenderer set_data when cds view changes

    * cds_view's source must be ColumnarDataSource, not ColumnDataSource

    * fix defaults matching tests

:040000 040000 07e3ac2a8ed9e68119f7428cd5be10a8322f8de3 46807a79cbe11e1a5dacc44b32f20d2a71f5901f M	bokeh
:040000 040000 6f647b0136744db874ddbcdb565d67afa37c3227 3e7c7fd515dc81432b0a37fe696556700bb9c168 M	bokehjs
~/work/bokeh bisect/good-7580c53bb0ed8ae0d8d0ba5ff11e62223c102dee|bisect
```

clairetang6:
This PR fixes streaming in the presence of CDSViews. The if statement is to guard against cases where the data source of a renderer is not a columnar data source and therefore doesn't have a `streaming` signal and when there is no data source.

One thing I considered was that for each streaming event, of the CDSView and GlyphRenderer, you would want the CDSView to respond first to update its indices before the GlyphRenderer responds with its set_data. Otherwise the first streaming event would not cause a re-render of the new data. This seems to be what happens (I checked with the following script that streams new_data on a button click), but I don't know if there is a way to guarantee that one slot is called before another one? 

- [x] issues: fixes #6628 
- [x] tests added / passed

```python

from numpy import random

source = ColumnDataSource(dict(x=[1], y=[2]))
x_val = 2

p = figure(plot_height=500, tools="xpan,xwheel_zoom,xbox_zoom,reset", y_range=(0, 10))

def add_data():
    global x_val
    new_data = dict(x=[x_val], y=[random.randint(10)])
    source.stream(new_data, 300)
    x_val = x_val + 1
    
button = Button(label="Add")
button.on_click(add_data)

curdoc().add_root(column(p, button))
```

### DIFF
--- a/bokehjs/src/coffee/models/sources/cds_view.coffee
+++ b/bokehjs/src/coffee/models/sources/cds_view.coffee
@@ -29,6 +29,8 @@ export class CDSView extends Model
     )
     if @source?.streaming?
       @connect(@source.streaming, () -> @compute_indices())
+    if @source?.patching?
+      @connect(@source.patching, () -> @compute_indices())
 
   compute_indices: () ->
     indices = (filter.compute_indices(@source) for filter in @filters)
--- a/bokehjs/src/coffee/models/sources/cds_view.coffee
+++ b/bokehjs/src/coffee/models/sources/cds_view.coffee
@@ -27,6 +27,8 @@ export class CDSView extends Model
       @compute_indices()
       @change.emit()
     )
+    if @source?.streaming?
+      @connect(@source.streaming, () -> @compute_indices())
 
   compute_indices: () ->
     indices = (filter.compute_indices(@source) for filter in @filters)
--- a/bokehjs/test/models/sources/cds_view.coffee
+++ b/bokehjs/test/models/sources/cds_view.coffee
@@ -61,3 +61,12 @@ describe "CDSView", ->
     it "convert_indices_from_subset", ->
       view = new CDSView({source: cds, filters: [filter1, filter2]})
       expect(view.convert_indices_from_subset([0, 1])).to.be.deep.equal [1, 2]
+
+  it "should update its indices when its source streams new data", ->
+    cds = new ColumnDataSource({data: {x: [], y: []}})
+    new_data = {x: [1], y: [1]}
+
+    view = new CDSView({source: cds})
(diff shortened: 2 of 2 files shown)


bryevdv:
A guess, but this probably needs to check for patching as well as streaming

clairetang6:
On first thought, I didn't think you would need to update the indices when data is patched because it doesn't change the length of data. But the cds_view also takes care of filtering, so it definitely should compute its indices again when new data is patched. Thanks for the catch.

PR #6648, opened 2017-07-19
re-enable "secrets filter"

bryevdv:
From Travis Support

> This is a follow-up to say that we believe that we have a solution in place for the truncated logs (though not yet for the colorized output). Could you try and remove `filter_secrets: false` from your configuration? Please let us know if you would encounter a truncated log again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: generic
 sudo: false
 dist: trusty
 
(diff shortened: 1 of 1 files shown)


bryevdv:
still seems to truncate, definitely removes colorized output in any case, so leaving as-os

PR #6665, opened 2017-07-22
Fix DataTable resize regression by computing indices when source data changes

### LINKED ISSUE #6642: export_csv example under 0.12.7dev11 doesn't resize table
Running example export_csv from https://github.com/bokeh/bokeh/tree/master/examples/app/export_csv

under 0.12.7dev11 the table doesn't resize and extra rows are filled with 'undefined', '$NaN'.
The number of rows is 248 and doesn't change when moving the slider.
The rows after 248 are not shown.

Under 0.12.6 everything works perfect.

clairetang6:
This PR fixes the DataTable resize regression. It also updates the export_csv app, so that the range of the slider better matches the example data. 

- [x] issues: fixes #6642
- [x] tests added / passed

### DIFF
--- a/examples/app/export_csv/main.py
+++ b/examples/app/export_csv/main.py
@@ -4,7 +4,7 @@ import pandas as pd
 
 from bokeh.layouts import row, widgetbox
 from bokeh.models import ColumnDataSource, CustomJS
-from bokeh.models.widgets import Slider, Button, DataTable, TableColumn, NumberFormatter
+from bokeh.models.widgets import RangeSlider, Button, DataTable, TableColumn, NumberFormatter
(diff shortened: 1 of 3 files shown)


examples/app/export_csv/main.py
```diff
@@ -19,7 +19,7 @@ def update():
         'years_experience' : current.years_experience,
     }
 
-slider = Slider(title="Max Salary", start=10000, end=250000, value=150000, step=1000)
+slider = Slider(title="Max Salary", start=10000, end=110000, value=50000, step=1000, format="0,0")
```
bryevdv: Don't we have a working double slider now? It might be nice to update to that for a salary range

PR #6672, opened 2017-07-25
Add selection/inspection interactions to GraphRenderers

### LINKED ISSUE #6612: Task: Graph visualization improvements
based on #6544 

Needs:

- [x] user guide documentation
- [x] support for tool interactions
- [x] add "Examples" example
- [x] add GraphRenderer.hit_test tests
- [x] have GraphRenderer figure into Plot.bounds (won't fix)
- [x] figure out GraphHitTestPolicy NodesAndLinkedEdges.do_inspection test issue (they're commented out)
- [x] Add EdgesAndLinkedNodes policy
- [x] Add Figure.graph method and handle kwargs
- [ ] update TS API

canavandl:
- [ ] issues: fixes #6612 
- [ ] tests added / passed

Adds selection/inspection interactions to GraphRenderers


bokehjs/src/coffee/core/selection_manager.coffee
```diff
@@ -24,7 +24,7 @@ export class SelectionManager extends HasProps
 
   inspect: (renderer_view, geometry) ->
     did_hit = false
-    @_set_inspector(renderer_view)
+    @_set_inspector(renderer_view.model)
```
canavandl: This is an annoying side-effect to do. The inspector should probably be created in the renderer.initialize method.

bokehjs/src/coffee/models/renderers/graph_renderer.coffee
```diff
@@ -66,11 +70,48 @@ export class GraphRenderer extends Renderer
...
+        node_indices = (@node_renderer.data_source.data.index[i] for i in indices["1d"].indices)
+        edge_source = @edge_renderer.data_source
+        edge_indices = []
+        for i in [0...edge_source.data.start.length]
+          if contains(node_indices, edge_source.data.start[i]) or contains(node_indices, edge_source.data.end[i])
+            edge_indices.push(i)
+
+        linked_index = create_hit_test_result()
+        for i in edge_indices
+          linked_index["2d"].indices[i] = [0]
```
canavandl: should "select" entire array instead of just [0] element

canavandl: Selecting the entire array is tough because it's coordinates are supplied by the layout_provider. Currently we only support 2-element multilines (i.e. one segment), so this selects the whole segment. This will probably need to be readdressed in the future,

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -157,8 +157,7 @@ export class GlyphRendererView extends RendererView
       else if selected['1d'].indices.length > 0
         selected = selected['1d'].indices
       else
-        selected = Object.keys(selected["2d"].indices)
-        selected = (parseInt(i) for i in selected)
+        selected = (parseInt(i) for i in Object.keys(selected["2d"].indices))
```
mattpap: @bryevdv, this would also take advantage of a proper map data structure.

canavandl: yup. I spent 20 minutes trying to figure out why I couldn't get values I'd just written to this dict.

bokeh/models/renderers.py
```diff
@@ -242,6 +242,18 @@ def _check_malformed_graph_source(self):
...
+    With the "default" selection mode, node glyphs will be selected in the same
+    way as normal glyphs. With "linked" mode, selecting a node will also select
+    the attached edges.
+    """)
+
+    inspection_mode = Enum("default", "linked", help="""
+    With the "default" inspection mode, node glyphs will be inspected in the same
+    way as normal glyphs. With "linked" mode, inspecting a node will also inspect
+    the attached edges.
+    """)
```
bryevdv: I we want to make these as flexible as possible, we might want to make these policies configurable with named models, e.g. `DefaultGraphSelection` and `LinkedGraphSelection` that would leave the door open for a `CustomJSGraphSelection` which I think would be a valuable relief valve for the inevitable fancy things people want to do that we can't maintain in the library proper

canavandl:
here a demo example (requires networkx):

```python
import networkx as nx

from bokeh.io import show
from bokeh.models.graphs import from_networkx
from bokeh.models import Plot, Range1d, TapTool, HoverTool, Circle, MultiLine

G=nx.Graph()
G.add_nodes_from([0,1,2,3])
G.add_edges_from([[0,1], [0,2], [2,3]])

p = Plot(x_range=Range1d(-1.1, 1.1), y_range=Range1d(-1.1, 1.1))

r = from_networkx(G, nx.circular_layout)

r.node_renderer.glyph.radius = 0.10
r.node_renderer.hover_glyph = Circle(fill_color='red')
r.selection_mode = "linked"
r.inspection_mode = "linked"
r.edge_renderer.glyph.line_width = 10
r.edge_renderer.glyph.line_color = 'gray'
r.edge_renderer.nonselection_glyph = MultiLine(line_color="green", line_width=5, line_alpha=0.2)
r.edge_renderer.hover_glyph = MultiLine(line_color="red", line_width=15)

p.renderers.append(r)

p.add_tools(HoverTool(), TapTool())

show(p)
```

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -157,7 +157,7 @@ export class GlyphRendererView extends RendererView
       else if selected['1d'].indices.length > 0
         selected = selected['1d'].indices
       else
-        selected = []
+        selected = (parseInt(i) for i in Object.keys(selected["2d"].indices))
```
bryevdv: why are the indices strings?

bokehjs/src/coffee/models/renderers/graph_renderer.coffee
```diff
@@ -66,11 +70,48 @@ export class GraphRenderer extends Renderer
...
+        for i in edge_indices
+          linked_index["2d"].indices[i] = [0] #currently only supports 2-element multilines, so this is all of it
+
+        # manually get or create inspector
+        @edge_renderer.data_source.selection_manager._set_inspector(@edge_renderer)
+        edge_inspector = @edge_renderer.data_source.selection_manager.inspectors[@edge_renderer.id]
+        edge_inspector.update(linked_index, final, append)
+        @edge_renderer.data_source.inspected = edge_inspector.indices
+        @edge_renderer.data_source.inspect.emit()
+
```
bryevdv: making selection polices be models would also let the functionality be factored into chunks so that this does not have to be be a "huge switch" kind of function.

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
canavandl: because ``Object.keys`` casts objects keys to strings. Or, more precisely, it returns the internally stored values (which are strings) instead of casting back the dtypes (I think).

bokehjs/test/models/renderers/glyph_renderer.coffee
```diff
@@ -55,7 +55,7 @@ describe "GlyphRenderer", ->
       hit_test: (geometry) -> return create_1d_hit_test_result([])
 
     class HitTestHit
-      hit_test: (geometry) -> return create_1d_hit_test_result([0, 1])
+      hit_test: (geometry) -> return create_1d_hit_test_result([[0], [1]])
```
bryevdv: can you explain this change?

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
mattpap: Because we're dealing with JS objects acting as maps. E.g. `{1: []}` has one key of value `"1"`.

bokehjs/test/models/renderers/glyph_renderer.coffee
canavandl: ``create_1d_hit_test_result`` accepts a list of lists. There's optionally a second element that the hit index is sorted by. The old behavior was returning an array of ``undefined`` values.

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
bryevdv: right, of course, thanks

bokeh/models/renderers.py
bryevdv: Also we should check with @hsparra but I think in a default "linked" mode, the desire might also be for adjacent nodes to get selected, not just the edges leading to them.

bryevdv: Also also wonder if we need modes per node and per edge. i.e. maybe you want selecting an edge to select the adjacent nodes, but you don't want selecting a node to do any linking. I don't think that is expressible here?

hsparra: I think a way to encapsulate modes would be useful. An example might be to select other nodes with a specified attribute that the selected node has. Say I have a graph representing relationships among individuals. Perhaps when I select a given node I want to see all the other nodes (i.e. individuals) that have the same type of favorite animal or all the individuals that live in the same state.

bokehjs/test/models/graphs/graph_hit_test_policy.coffee
```diff
@@ -0,0 +1,193 @@
...
+        @edge_source.selection_manager.inspectors[@edge_renderer.id].indices = initial_inspection
+
+        @stub.returns(create_hit_test_result())
+        policy = new NodesAndLinkedEdges()
+
+        policy.do_inspection("geometry", @gv, true, false)
+
+        indices = @edge_source.selection_manager.inspectors[@edge_renderer.id].indices
+
+        # expect(indices.is_empty()).to.be.true
```
canavandl: This is commented out because it fails. I think it's a test setup issue and not an implementation issue.

bokehjs/test/models/graphs/graph_hit_test_policy.coffee
```diff
@@ -0,0 +1,193 @@
...
+        expect(@edge_source.inspected.is_empty()).to.be.true
+
+      it "should select linked edges if hit_Test result is not empty", ->
+        @stub.returns(create_1d_hit_test_result([[0], ]))
+        policy = new NodesAndLinkedEdges()
+
+        policy.do_inspection("geometry", @gv, true, false)
+
+        indices = @edge_source.selection_manager.inspectors[@edge_renderer.id].indices
+        # expect(indices['2d'].indices).to.be.deep.equal({ '0': [ 0 ], '1': [ 0 ] })
```
canavandl: This is commented out because it fails. I think it's a test setup issue and not an implementation issue.

canavandl:
todo:
* fix up test weirdness (see comments)
* add/fixup policy-related docstrings
* decide on default policies?

bokehjs/test/models/graphs/graph_hit_test_policy.coffee
canavandl: I've wasted enough time trying to figure this out. It's probably a shallow vs deep copy issue, but I'm going to punt and open an issue.

PR #6685, opened 2017-07-27
Add a note on jinja2 and pyyaml dependencies

### LINKED ISSUE #6684: Following dev documentation fails for a clean conda env
The documentation implicitly assumes that all is done in the root env which already has `jinja2` and `pyyaml`. In the case of a clean conda env, these packages have to be installed manually.

p-himik:
Fixes #6684

### DIFF
--- a/sphinx/source/docs/dev_guide/setup.rst
+++ b/sphinx/source/docs/dev_guide/setup.rst
@@ -93,12 +93,12 @@ sources:
     basis with the ``-c`` command line option to ``conda``, e.g.
     ``conda install -c bokeh -c conda-forge <pkgs>``
 
-It's also necessary to install `conda-build` first, to bootstrap the rest
-of these instructions. To do that, execute:
+It's also necessary to install `jinja2` and `pyyaml` first, to bootstrap
+the rest of these instructions. To do that, execute:
 
 .. code-block:: sh
 
-    conda install conda-build
+    conda install jinja2 pyyaml
 
 From the top level of the *source checkout* directory, execute the following
 command at your command prompt to install all the required packages:


sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -103,6 +103,10 @@ of these instructions. To do that, execute:
 From the top level of the *source checkout* directory, execute the following
 command at your command prompt to install all the required packages:
 
+.. note::
+    If you want to use a clean conda environment for Bokeh development, you
+    may have to install ``jinja2`` and ``pyyaml`` in that environment manually
```
mattpap: Missing trailing period.

bryevdv:
@p-himik what about simply adding `jinja2 and pyyaml` to the `conda install` line right above?
```
conda install conda-build jinja2 pyyaml
```
I think that should work in all cases, and then users are not buried with trying to figure out whether a note applies to them, they can just always run the command and things will work.

p-himik:
@bryevdv Two things that I don't like about it:

1. It would override packages in the root env in case a user has added bokeh and/or conda-forge channels to conda config
2. It simply won't work for a non-root env - `conda-build` can only be installed in the root env.

bryevdv:
Oh right, I forgot that needs to be changed any. The `deps.py` script has been fixed to no longer depend on conda build, so we should not instruct users to install it. The `conda-build` install should be removed entirely. So then just this:
```
conda install jinja2 pyyaml
```

?

From experience, having instructions with branches "if your situation is this, do this, otherwise do that" is *really undesirable*, it's easily one of the biggest contributing factors to hard-to-follow instructions.

bryevdv:
Actually tho, I am confused, `pyyaml` and `jinja2` are explicit dependencies lists in the recipe:

https://github.com/bokeh/bokeh/blob/master/conda.recipe/meta.yaml#L32-L33

so this line:

```
conda install `python scripts/deps.py build run test`
``` 
should install them, regardless of environment. To be clear you are saying this did not happen?

bryevdv:
Yeah I just realized that (had to try things out out to figure it out) :)

p-himik:
Hmm, after looking at `meta.yaml` I think it can be made even simpler unless I'm missing something:

1. Move `REQUIRES` from `setup.py` to `meta.yaml` - removes the need to manually install `jinja2` and `yaml`
2. Parse `meta.yaml` in `setup.py` to fill in requirements
3. Remove `deps.py`
4. Create files `environments/windows.yaml` and `environments/osx_linux.yaml` that would import relevant sections from `meta.yaml` (`meta.yaml` will have to have anchors)

Users will have to just run `conda install -f environment/%suitable_os%.yaml`.

I just wanted to share the idea - I don't think that this PR is the right place for its implementation, and this plan may fail miserably or just be not applicable for some reason that I don't see yet.

bryevdv:
> Parse meta.yaml in setup.py to fill in requirements

This is the step I very strongly object to. :) After many years of churn and problems, I want the `setup.py` to be as simple and possible and I am currently very happy with `setup.py` being the single source of truth for runtime requirements. 

> Remove deps.py

> Create files environments/windows.yaml and environments/osx_linux.yaml that would import relevant sections from meta.yaml (meta.yaml will have to have anchors)

unfortunately env files only let you create *new* envs, not specify a set of packages to install into an *existing* env. But this is something we have to have for our CI builds because we use `conda-build', and as you've noted that can only be installed into the root env. So we need tools that can be used with the root env. Creating env files means having `deps.py` *and* env files and maintaining things in two separate places and that is definitely a non-starter for me.

p-himik:
I see, thanks for explaining.

bryevdv:
> I see, thanks for explaining. 

My pleasure, I should probably put aloof this info in the dev guide as well...

bryevdv:
Thanks @p-himik !

PR #6688, opened 2017-07-27
Make bokehjs work in notebook

### LINKED ISSUE #6679: BokehJS fails to load at all in notebooks after 0.12.7dev12
(bryan) Just clarifying: the issue is that BokehJS no longer works at all in notebooks

----

When moving from bokeh 0.12.7dev11 to 0.12.7dev12, the following code 

   import holoviews as hv
   hv.notebook_extension('bokeh','matplotlib')

goes from working in a notebook to giving the following output:

```
TypeError: undefined is not an object (evaluating 'Bokeh.set_log_level')
```

The JS console shows the following errors:
```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (preact.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (preact-compat.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (index.js.map, line 0)
```

Holoviews is version 1.8.0.

@mattpap

mattpap:
I ended up commenting UMD code. This is suboptimal, but a proper solution is much harder than I expected, and requires deeper changes to how autoload works.

fixes #6679

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -118,4 +118,4 @@ calls it with the rendered model.
       run_inline_js();
     });
   }
-}(this));
+}(window));
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -15,16 +15,16 @@ calls it with the rendered model.
 :type css_urls: list
 
 #}
-(function(global) {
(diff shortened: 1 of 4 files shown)


bryevdv:
confirmed working from @hsparra

PR #6686, opened 2017-07-27
Add EdgesAndLinkedNodes policy

canavandl:
- [x] addresses some of #6612 
- [x] tests added / passed
- [ ] release document entry (if new feature or API change)

PR #6702, opened 2017-07-31
Add examples/models/files/graphs.py example

canavandl:
needs to be merged after #6686 

looks like (each with different interactions):
screen shot 2017-07-31 at 5 25 01 pm

### DIFF
--- a/examples/examples.yaml
+++ b/examples/examples.yaml
@@ -1,6 +1,7 @@
 - path: "models/file"
   slow: ["choropleth.py"]
   no_js: ["maps_cities.py", "maps.py", "trail.py"]
+  no_diff: ["graphs.py"]
 
 - path: "models/server"
   type: server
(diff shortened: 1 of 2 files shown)


examples/examples.yaml
```diff
@@ -1,6 +1,7 @@
 - path: "models/file"
   slow: ["choropleth.py"]
   no_js: ["maps_cities.py", "maps.py", "trail.py"]
+  no_diff: ["graphs.py"]
```
canavandl: it'll be a no_diff example because the networkx layout methods aren't all deterministic

mattpap: Why is that? Is there anything we can do other than simply silencing this example? If there is a way, it should be preferred over no_diff or no_js.

canavandl: You're right that doing no_diff reduces its test value. I'll modify the example to use the two deterministic layout algos and remove the no_diff flag.

canavandl: Looking at it, the only deterministic layout function is nx.circular_layout. Having only that layout reduces the example value too much in my opinion. Would you be opposed to an additional screenshot test of a circular layout example?

PR #6705, opened 2017-08-01
Canavandl/3935 examples tests 3

### LINKED ISSUE #3935: How to get the server examples testable again

canavandl:
supercedes: #6589 

- [x] issues: fixes #3935 
- [ ] tests added / passed

Overall - this PR does the minimum to get the server examples working by using our application handler code to parse scripts into Documents and push them into a running bokeh server process. It doesn't use any of our screenshot code but continues to our our phantomjs CLI script code.

Feedback points of discussion:
* Bokeh.documents race condition issue, see inline comment below
* Is minor refactor of server examples structure ok?
  * Using curdoc, not Document, and explicitly adding roots
  * Moving session stuff into ``if __name__ == "__main__": conditional
  (benefits include examples now work via bokeh server and python script.py into running bokeh server process. Also they feel "cleaner" to me)
* The server logs get included in the Travis logs re: each example. They're a little verbose but may be helpful in bug situations. Are they ok as-is or should be try to silence them?

Here's the report: 
https://s3.amazonaws.com/bokeh-travis/259919782/examples.html#models/server/buttons

The fourier_animated.py example is slow enough to load that it intermittently isn't screengrabbed before a timeout. We'll need to resolve this issue too.

As an aside, I think we could/should refactor the script examples tests to use our handlers in the same way. It's reduce some code duplication and test complexity (right now we're eval-ing code strings :( ).


tests/examples/test_examples.py
```diff
@@ -198,7 +201,9 @@ def _assert_snapshot(example, url, example_type):
 
     assert success, "%s failed to load" % example.relpath
     assert no_resources, "%s failed with %d missing resources" % (example.relpath, len(resources))
-    assert no_errors, "%s failed with %d errors" % (example.relpath, len(errors))
+
+    if example_type != "server":
```
canavandl: The issue is that all server examples end up with an error because of a race condition related to: https://github.com/bokeh/bokeh/blob/master/tests/plugins/phantomjs_screenshot.js#L85

I think it's a race condition where the document isn't evaluated and appended to the Bokeh variable before this line is evaluated, so we write an error. It's unclear to me what the best work-around is.

ping @mattpap

bryevdv:
I'm OK with the example refactors and verbose logs.

tests/examples/test_examples.py
canavandl: per @mattpap : 
>I'm aware of this issue, but right now I don't have anything useful to contribute. This code was meant to be used only with standalone examples. In server examples we also need to wait for documents to be created and this asks for `Bokeh.is_idle` to be added.

mattpap: @canavandl, I will look into this. It's likely that a simple generalization of the existing algorithm should be sufficient, though I have to look into when server creates documents.

canavandl: ok. as a work-around, we could add a blocking time.sleep to the ``test_server_examples`` code to ensure the doc is complete. While ugly it would let us get the server examples working again.

canavandl: I was able to use this code snippet from the phantomjs docs to wait for slow server responses: https://github.com/ariya/phantomjs/blob/master/examples/waitfor.js

So we won't have to add a blocking sleep in python.

canavandl:
ok. I'm finally happy w/ the generated screenshots and robustness of things. You can view the report here:

https://s3.amazonaws.com/bokeh-travis/262064379/examples.html

PR #6733, opened 2017-08-07
Removed hard-coded zoom levels on Mercator Tile calculations

### LINKED ISSUE #6709: Increase the number of zoom levels available to Bokeh Tile sources
The current version of Bokeh pre-calculates the number of available zoom resolutions for Mercator (and by inheritance BBox) Tile sources. In the *bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee* file this is hard-coded on line 19 as:

 `@_resolutions = (@get_resolution(z) for z in [0..30])`

This assumes we can only ever zoom to level 30, but there are cases where this is insufficient. In my example, I use an ArcGIS server as a BBox tile source, visualising aerial photography at a much higher resolution than commercial services provide.

Attached is an example of the resolution we need to be able to zoom to. (We do have imagery at a few more levels of zoom, but they look less meaningful than this one). If I change the range of resolutions to 50 I can reach all these levels, but maybe this should be a configurable option.

I have only tested this by modifying the generated *bokeh.js* files directly, but I understand that the coffee file is the original source.

railway

euandekock:
- [ ] issues: fixes #6709 
- [ ] tests added / passed

Removed the hard coded 0 and 30 for Zoom levels during resolution calculation, and used the preconfigured min_zoom and max_zoom options, which are defaulted to 0 and 30, but are configurable when defining the tile source.

### DIFF
--- a/bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee
+++ b/bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee
@@ -16,7 +16,7 @@ export class MercatorTileSource extends TileSource
 
   initialize: (options) ->
     super(options)
-    @_resolutions = (@get_resolution(z) for z in [0..30])
+    @_resolutions = (@get_resolution(z) for z in [@min_zoom..@max_zoom])
 
   _computed_initial_resolution: () ->
     if @initial_resolution?


bryevdv:
@euandekock perfect, what a pleasant surprise, thanks!

PR #6734, opened 2017-08-07
Fix False renderer args not being set

### LINKED ISSUE #6731: Setting visible=False doesn't work with glyph functions
Bokeh 0.12.6

```python

p = figure()
p.triangle(x=[0], y=[0], visible=False)
c = p.circle(x=[1], y=[1])
c.visible = False

curdoc().add_root(p)
```

The triangle is visible, although it should not be.
The circle is invisible.

The issue is that `bokeh.plotting.helpers._pop_renderer_args` uses `if val` instead of `if val is not None`.

p-himik:
Fixes #6731

Apart from fixing `False`, also fixes potential use of `None` as an argument value

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -94,12 +94,15 @@ def get_default_alpha(plot=None):
     return 1.0
 
 
+_RENDERER_ARGS = ['name', 'x_range_name', 'y_range_name',
+                  'level', 'view', 'visible', 'muted']
+
+
 def _pop_renderer_args(kwargs):
-    result = dict(data_source=kwargs.pop('source', ColumnDataSource()))
-    for attr in ['name', 'x_range_name', 'y_range_name', 'level', 'view', 'visible', 'muted']:
-        val = kwargs.pop(attr, None)
-        if val:
-            result[attr] = val
+    result = {attr: kwargs.pop(attr)
+              for attr in _RENDERER_ARGS
+              if attr in kwargs}
+    result['data_source'] = kwargs.pop('source', ColumnDataSource())
     return result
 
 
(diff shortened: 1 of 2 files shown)


bryevdv:
@p-himik I think for this we could add some tests to:

https://github.com/bokeh/bokeh/blob/master/bokeh/plotting/tests/test_helpers.py

which is a bit thin as it is. Is this something you could do? If not do you mind if I push some to your branch later this week?

p-himik:
Sure, I'll add some tests.

bryevdv:
@p-himik I'm not sure what is wrong offhand, but Py 2.7 tests are failing due to this:
```
==================================== ERRORS ====================================
____________ ERROR collecting bokeh/plotting/tests/test_helpers.py _____________
ImportError while importing test module '/home/travis/build/bokeh/bokeh/bokeh/plotting/tests/test_helpers.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
bokeh/plotting/tests/test_helpers.py:1: in <module>
    from unittest import mock
E   ImportError: cannot import name mock
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
============================ 1090 tests deselected =============================
================== 1090 deselected, 1 error in 10.01 seconds ===================

```

bryevdv:
I *think* that changing to
```
from mock import mock
```

will work in both py2 and py3, we list `mock` as a test dep which will be installed for both

bryevdv:
Alternatively, we use `pytest.mock` in other tests, if that is workable here then it's probably preferable for keeping things consistent

bryevdv:
Actually in the interest if trying to clear this today, I made the change to try importing from `mock` instead.

PR #6735, opened 2017-08-07
Fix zero range_padding causing empty plot

### LINKED ISSUE #6730: Setting range_padding to 0 can lead to an empty plot
Bokeh 0.12.6

Running
```python

ds = ColumnDataSource(dict(x=[0], y=[0]), name='initial_ranges_fix_ds')
p = figure(width=200, height=200)
p.circle(x='x', y='x', source=ds)
p.x_range.range_padding = 0

curdoc().add_root(p)
```

leads to
image

while JS console shows only
```
[bokeh] setting log level to: 'info'
bokeh.min.js:1 [bokeh] Websocket connection 0 is now open
bokeh.min.js:22 Bokeh items were rendered successfully
```

The issue is gone if `x` column has some different values - adding more dots on the same vertical line will still result in an empty plot.
Setting `range_padding` to a value other than 0, however small, leads to a correctly rendered plot.

The reason, it appears, is that the lines
```coffee
        if max == min
          span = @default_span
```
are present only in the `if range_padding? and range_padding > 0` section in `DataRange1d._compute_range`.

p-himik:
Fixes #6730

### DIFF
--- a/bokehjs/test/models/ranges/data_range1d.coffee
+++ b/bokehjs/test/models/ranges/data_range1d.coffee
@@ -156,6 +156,8 @@ describe "datarange1d module", ->
       expect(r._compute_range(3, 3)).to.be.deep.equal [2, 4]
       r.default_span = 4
       expect(r._compute_range(3, 3)).to.be.deep.equal [1, 5]
+      r.range_padding = 0
+      expect(r._compute_range(3, 3)).to.be.deep.equal [1, 5]
 
     it "should use default_span as powers of 10 when scale_hint='log'", ->
       r = new DataRange1d()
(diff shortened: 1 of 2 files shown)


bryevdv:
@p-himik thanks for the PR! Fortunately `DataRange1d` does have OK test coverage currently:

https://github.com/p-himik/bokeh/blob/master/bokehjs/test/models/ranges/data_range1d.coffee

But if there was a specific case we missed that this fixed it would be good to add a test or two to exercise that code path. Do you want to take a crack at that? If not, do you mind if I push a commit to this branch at some point this week?

p-himik:
Yeah, sure, I'll add a couple of tests.

bryevdv:
@p-himik just a heads up, for this to go into `0.12.7` it will need t be wrapped up by Monday or so, do you still anticipate being able to add a few tests?

bokehjs/test/models/ranges/data_range1d.coffee
```diff
@@ -225,11 +231,20 @@ describe "datarange1d module", ->
       [a, b] = r._compute_range(0.01, 10)
       expect(a).to.be.closeTo(0.0017782794100389264, 1e-12)
       expect(b).to.be.closeTo(56.23413251903488, 1e-12)
+      r.range_padding = 0
+      [a, b] = r._compute_range(0.01, 10)
+      expect(a).to.be.closeTo(0.01, 1e-12)
+      expect(b).to.be.closeTo(10, 1e-12)
```
bryevdv: Nice to see this, we probably need to use `closeTo` in a variety of places elsewhere, I somehow missed that it existed until now.

bryevdv:
Thanks @p-himik !

PR #6737, opened 2017-08-07
Make sure AjaxDataSource is initialized only once

### LINKED ISSUE #6736: Attaching AjaxDataSource to multiple glyphs leads to multiple AJAX requests
Bokeh 0.12.6

Started because of https://stackoverflow.com/questions/45301848/bokehflask-multiple-ajaxdatasource-calls

Because of 
https://github.com/bokeh/bokeh/blob/891c5ab40af9b2ea8a7cd5c5122acdd4ff878471/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee#L59-L60
 the code at 
https://github.com/bokeh/bokeh/blob/90e14b8e377dafe29b044c0f86245d891cb4f8ad/bokehjs/src/coffee/models/sources/ajax_data_source.coffee#L25-L26
 will be run multiple times if an `AjaxDataSource` is attached to multiple glyphs.

p-himik:
Fixes #6736

### DIFF
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -57,7 +57,7 @@ export class GlyphRendererView extends RendererView
     @set_data(false)
 
     if @model.data_source instanceof RemoteDataSource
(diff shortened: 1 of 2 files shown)


bryevdv:
Thanks @p-himik !

PR #6738, opened 2017-08-07
Fix steaming for DataFrame-based ColumnDataSource

### LINKED ISSUE #6667: Creating a ColumnDataSource with a DataFrame makes it impossible to use streaming
Bokeh 0.12.6

Executing
```python
import pandas as pd

d = pd.DataFrame({'a': [1]})
cds = ColumnDataSource(d)
cds.stream(ColumnDataSource.from_df(d))
```
leads to
```
AttributeError: 'Series' object has no attribute 'extend'
```

The reason is that `ColumnDataSource._data_from_df` uses `DataFrame.to_dict('series')` but `PropertyValueDict._stream` assumes the columns to be lists, or at least something that has `extend` method.

On a related note - probably making `steam` also accept a raw `DataFrame` would make the interface more cohesive.

p-himik:
Fixes #6667

### DIFF
--- a/bokeh/models/tests/test_sources.py
+++ b/bokeh/models/tests/test_sources.py
@@ -40,7 +40,7 @@ class TestColumnDataSource(unittest.TestCase):
         ds = ColumnDataSource(df)
         self.assertTrue(set(df.columns).issubset(set(ds.column_names)))
         for key in data.keys():
-            self.assertIsInstance(ds.data[key], pd.Series)
+            self.assertIsInstance(ds.data[key], np.ndarray)
             self.assertEquals(list(df[key]), list(ds.data[key]))
         self.assertIsInstance(ds.data['index'], np.ndarray)
         self.assertEquals([0, 1], list(ds.data['index']))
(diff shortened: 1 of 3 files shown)


bryevdv:
@p-himik just a heads up, for this to go into `0.12.7` it will need t be wrapped up by Monday or so

bryevdv:
FYI the changed code in the merge conflict is to deal with the multi-index case, it makes a CDS key by joining the sub-index col names with `_`. The new code may be OK as-is for the rest of your PR, but you should take  look.

p-himik:
@bryevdv The tests contain a fair amount of repetitions - not only mine but many tests in `test_sources.py`. What do you think about rewriting this file in pytest? I've been using it quite intensively lately, and I think that file can be rewritten in a way that offers a better coverage with a smaller code size.

bryevdv:
@p-himik we try to use `pytest` for all new tests, and I would not be opposed to converting old tests as feasible. I will say even though we use pytest for new tests we  don't have alot of experience e.g. with parameterized tests so some good examples of doing that or other best practices would be great

bokeh/models/tests/test_sources.py
```diff
@@ -190,6 +190,170 @@ def mock(*args, **kw):
...
+        data_stream = ds.data._stream
+
+        def stream_wrapper(*args, **kwargs):
+            stream_stuff['args'] = args
+            stream_stuff['kwargs'] = kwargs
+            data_stream(*args, **kwargs)
+
+        ds.data._stream = stream_wrapper
+
+        ds._stream(pd.Series([11, 21, 31], index=list('abc')), 7)
```
bryevdv: OH this is interesting, I didn't realize `Series` could be used this way.

bryevdv:
@p-himik Thanks the tests look good. I agree that there is a fair amount of boilerplate, but I'm willing to defer cleanup up on that until the entire module gets converted to use `pytest` at some point. 

Thanks!!

PR #6742, opened 2017-08-08
Add additional instructions to /docs/dev_guide/setup.html

### LINKED ISSUE #6627: Add additional instructions to /docs/dev_guide/setup.html
In the section 

http://bokeh.pydata.org/en/dev/docs/dev_guide/setup.html#next-steps

it might be helpful to have some instructions on how to run some of the example scripts and apps.

e.g. 

```
cd examples/plotting/
env BOKEH_RESOURCES=inline python iris.py
```

with some details on what the `BOKEH_RESOURCES` environment variable does. And

```
cd examples/app
python -m bokeh serve --show sliders.py
```

malramsay64:
There are instructions added for running the examples from the examples
directory in the development setup guide. This is modelled on the user_guide
setup.

- [x] issues: fixes #6627 
- [ ] tests added / passed


sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -309,6 +309,41 @@ You should see output similar to:
...
+from http://cdn.pydata.org Other options can be found in the
+`resources reference`_.
+
+Another method of running bokeh is as a server. An example of this mode of operation can be run
+using the command
+
+.. code-block:: sh
+
+    python -m bokeh serve &
+    python examples/plotting/server/line_animate_widget.py
```
bryevdv: @malramsay64 I'd much prefer this to be a `bokeh serve` style app, using something in `examples/app`. It's not a stretch to say that I actively discourage using the bokeh server in this way (with `bokeh.client` in a separate script) unless there is an absolute need for it, so I don't want it to be the method that is prominently demonstrated.

sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -309,6 +309,41 @@ You should see output similar to:
...
     npm version         :  3.10.10
 
+The next check that can be made is to generate one of the example figures.
+There are a examples for the different ways in which bokeh can be used.
+
+To create an html file,
+
+.. code-block:: sh
+
+    env BOKEH_RESOURCES=inline python examples/plotting/file/iris.py
```
bryevdv: This is `csh`/`tcsh` specific , need to demonstrate also with Bash syntax (which I also think should be first) 
```
BOKEH_RESOURCES=inline python examples/plotting/file/iris.py
```
as well as for Windows (I don'r recall what that looks like offhand)

sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -309,6 +309,41 @@ You should see output similar to:
...
+
+.. image:: /_images/bokeh_iris_html.png
+    :scale: 50 %
+    :align: center
+
+The variable ``BOKEH_RESOURCES`` determines where the css and JavaScript
+resources required by bokeh are found. In this case they are included inline in
+the html file. Another possible value is ``BOKEH_RESOURCES=cdn`` which will fetch the resources
+from http://cdn.pydata.org Other options can be found in the
+`resources reference`_.
```
bryevdv: CDN is the default so I don't think this last sentence is necessary. Also if you are trying to test locally build BokehJS using CDN resources is always what you *don't* want to do, so we could state why `BOKEH_RESOURCES` is needed instead.

sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -309,6 +309,41 @@ You should see output similar to:
     node.js version     :  v6.10.3
     npm version         :  3.10.10
 
+The next check that can be made is to generate one of the example figures.
+There are a examples for the different ways in which bokeh can be used.
```
bryevdv: I'd suggest "generate one of the example figures" -> "run some of the examples" and "There are a examples for the different ways" -> "There are different ways"

bryevdv:
Thanks for the PR @malramsay64 I have just a few comments/suggestions.

sphinx/source/docs/dev_guide/setup.rst
```diff
@@ -309,6 +309,44 @@ You should see output similar to:
...
+
+.. image:: /_images/bokeh_iris_html.png
+    :scale: 50 %
+    :align: center
+
+The variable ``BOKEH_RESOURCES`` determines where the css and JavaScript
+resources required by bokeh are found. In this case they are included inline in
+the html file. The ``BOKEH_RESOURCES`` variable is required as the default
+behaviour is to use CDN resources.
+`resources reference`_.
```
bryevdv: I think this is an extra line? Also I think it is specifically worth mentioning that the reason `BOKEH_RESOURCES` is needed is because we want to use the local version of BokehJS that we just built.

bryevdv:
@malramsay64 awesome I had one last comment then I think it's good to go

malramsay64:
I think that should be everything. Thanks @bryevdv for your work making bokeh awesome.

bryevdv:
Thanks for your contribution @malramsay64 !

PR #6743, opened 2017-08-08
Allow arbitrary nested lists in layout function

### LINKED ISSUE #4538: Allow layout function to accept arbitrarily nested lists
Currently the layout function only accepts a list of lists.

It would be nice if we could hand it arbitrarily deeply nested lists of things to make it quicker to construct a complex layout.

malramsay64:
I have moved the creation of the grid in the layout into a separate function which recursively calls itself which should allow for arbitrary lists.

At this point I haven't added any tests. I couldn't find any current tests of the layout function, and am not sure how to test correctness.

- [ ] issues: fixes #4538
- [ ] tests added / passed

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -248,21 +248,28 @@ def layout(*args, **kwargs):
     children = _handle_children(*args, children=children)
 
     # Make the grid
-    rows = []
-    for r in children:
-        row_children = []
-        for item in r:
-            if isinstance(item, LayoutDOM):
-                item.sizing_mode = sizing_mode
-                row_children.append(item)
-            else:
-                raise ValueError(
-                    """Only LayoutDOM items can be inserted into a layout.
-                    Tried to insert: %s of type %s""" % (item, type(item))
-                )
-        rows.append(row(children=row_children, sizing_mode=sizing_mode))
-    grid = column(children=rows, sizing_mode=sizing_mode)
-    return grid
+    return _create_grid(children, sizing_mode)
+
+
+def _create_grid(iterable, sizing_mode, layer=0):
+    """Recursively create grid from input lists."""
+    return_list = []
+    for item in iterable:
+        if isinstance(item, list):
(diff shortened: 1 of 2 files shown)


bokeh/layouts.py
```diff
@@ -248,21 +248,28 @@ def layout(*args, **kwargs):
...
-    return grid
+    return _create_grid(children, sizing_mode)
+
+
+def _create_grid(iterable, sizing_mode, layer=0):
+    """Recursively create grid from input lists."""
+    return_list = []
+    for item in iterable:
+        if isinstance(item, list):
+            return_list.append(_create_grid(item, sizing_mode, layer+1))
```
mattpap: A test has to be added for this new case.

bryevdv:
I agree with @mattpap I realize the tests for `layouts.py` were already light, but this is the opportunity to help improve that. This function is fairly straightforward in terms if inputs and outputs and no side effects, so I think it would be sufficient to simply exercise the function with a variety of inputs, comparing to whatever the current outputs are (it's worth a little initial examination to make sure the current outputs make sense). The file to add tests to is here:

https://github.com/malramsay64/bokeh/blob/master/bokeh/tests/test_layouts.py

bokeh/tests/test_layouts.py
```diff
@@ -29,3 +32,35 @@ def test_gridplot_merge_tools_nested():
...
+        assert isinstance(row, lyt.Row)
+        for col in row.children:
+            assert isinstance(col, lyt.Column)
+
+
+@pytest.mark.parametrize('sizing_mode', SizingMode)
+def test_layout_sizing_mode(sizing_mode):
+    p1, p2, p3, p4 = figure(), figure(), figure(), figure()
+
+    grid = lyt.layout([[p1, p2], [p3, p4]], sizing_mode=sizing_mode)
```
mattpap: `grid = ` needs to be removed.

bryevdv:
Thanks @malramsay64 !

PR #6750, opened 2017-08-08
0.12.7 docs/examples

bryevdv:
@bokeh/dev if there are smaller docs/examples tasks let's collect them in this one PR.


README.md
```diff
@@ -208,7 +208,7 @@ very large or streaming datasets.
 
 Installation
 ------------
-We recommend using the [Anaconda Python distribution](https://www.continuum.io/why-anaconda) and conda to install Bokeh. Enter this command at a Bash or Windows command prompt:
+We recommend using the [Anaconda Python distribution](https://anconda.com/why-anaconda) and conda to install Bokeh. Enter this command at a Bash or Windows command prompt:
```
mattpap: an**a**conda

bryevdv: fixed, thanks

bryevdv:
I don;t know why the integration test is failing it started failing  the branding update commit, which changed no code

mattpap:
> I don;t know why the integration test is failing it started failing the branding update commit, which changed no code.

No clue either. The report is useless. I would suspect a browser version change to be responsible, though looking at metadata, I think it's the same (at last major number).

bryevdv:
I guess we will see in a few minutes when the latest merges to master finish, whether this is a general problem or confined to this branch.

mattpap:
btw. What's up with the last horizontal tick in `plotting/file/custom_datetime_axis`?

image

mattpap:
I compared all packages used for testing in this PR and the last good one, and all are the same.

bryevdv:
Oh, I changed it to use a much simpler tick override mechanism... but that's going to make any extra ticks look stupid... bounds would fix that

bryevdv:
Master is failing now too: https://travis-ci.org/bokeh/bokeh/jobs/269230310 This is a general issue

canavandl:
The failing test (``test_scale_both_resizes_width_and_height_with_fixed_aspect_ratio``) passes locally for me. The issue appears to originate w/ SauceLabs.

mattpap:
It gets even better in https://travis-ci.org/bokeh/bokeh/jobs/269230310 (`master` after PR #6823).

mattpap:
> Yes, it appears SauceLabs has upgraded from Firefox 45 -> Firefox 50.

Metadata doesn't confirm this: https://saucelabs.com/beta/tests/01a511ac16314c7aabde1c48752bb96d/metadata#25.

canavandl:
The Travis logs say so: https://travis-ci.org/bokeh/bokeh/jobs/269213598#L337

mattpap:
OK, that's bloody mess. I think we just have to force version 45, if possible, and later update our tests to work with more modern versions.

bryevdv:
wheres a passing test link, maybe compare metadata against that

mattpap:
We depend on the default version (https://github.com/bokeh/bokeh/blob/master/tests/integration/integration_tests_plugin.py#L65), so this (and a few other references) has to be fixed.

bryevdv:
Also can configure here: https://github.com/bokeh/bokeh/blob/master/tests/integration/integration_tests_plugin.py#L39

But doesn't metadata indicate the version has *not* changed?

mattpap:
> wheres a passing test link, maybe compare metadata against that

> But doesn't metadata indicate the version has not changed?

Yeah, e.g. PR #6823 (sliders styling) also has FF 50 installed, but that's unrelated to sauce labs, because we don't run FF locally.

mattpap:
In the case of lack of ideas to fix this, we can temporarily XFail this test (we have one similar already XFailed). At this point I have no idea what's wrong.

bryevdv:
em either, I will xfail the four tests...

bryevdv:
Not sure if it's coincidence but all four failing tests were the last tests in their respective files.

bryevdv:
passing now will merge when green

PR #6751, opened 2017-08-08
fix BOKEH_DOCS_CDN handling

### LINKED ISSUE #6739: plots don't render in dev docs due to missing resources
The docs incorrectly looking for `/en/0.12.7dev15/static/js/bokeh-gl.min.js` instead of `dev`

ping @bryevdv

bryevdv:
issues: fixes #6739


bryevdv:
@canavandl FYI this should be merged before another dev build

PR #6752, opened 2017-08-08
remove resize tool

### LINKED ISSUE #4196: Remove the Resize Tool?!
Speak now or forever hold you peace :D

The new layout work has some conflicts with the resize tool.

In the new layout work sizes are determined from the outside-in, rather than the inside-out. 

In the layout work, the resize tool will be removed from the set of default tools and it may be made no-op (at least for a while) to move things forward.

After some discussion with @bryevdv we thought it would be good to get some feedback on whether the resize tool is useful at all, and if so to describe the use cases so we can be thoughtful about adding it back in.

It may be, for example, that resize tool is useful on the outside layout box (which may contain one or many plots) so that everything inside resizes to fit that new user specified size.

bryevdv:
- [x] issues: fixes #4196

This makes the resize a no-op (with warning) on the python side and removes it entirely from BokehJS.

documentation:

screen shot 2017-08-08 at 16 36 23


mattpap:
@bryevdv, I think we will need a follow up on this, because there is more related code to remove. This includes `PlotCanvas.reset_dimensions`, `ResetTool.reset_size`, and possibly more.

Issue #6755, opened 2017-08-10
ColorMapper special colors do not support alpha


philippjfr:
Back in https://github.com/bokeh/bokeh/issues/5324 I described a bug handling RGB(A) tuples as special colors (nan_color, low_color and high_color) for color mappers. While that issue was partially resolved actually specifying an alpha value using this tuple was not supported. This is a problem that's particularly annoying in geographic contexts where you want to see the background to see the context of the Image you are viewing. Here is a small example demonstrating the issue:

```python
arr = np.random.rand(100, 100)
arr[:50, :50] = np.NaN

p = figure(x_range=(-.5, .5), y_range=(-.5, .5), width=300, height=300)
p.scatter(x=np.random.rand(100)-.5, y=np.random.rand(100)-.5)
r = p.image(image=[arr[::-1]], x=-.5, y=-.5, dw=1, dh=1, palette=hv.Cycle('Reds').values)

# Set nan color
r.glyph.color_mapper.nan_color = (0, 0, 0, 0)
show(p)
```

<img width="277" alt="screen shot 2017-08-10 at 11 24 31 am" src="https://user-images.githubusercontent.com/1550771/29166514-89789b88-7dbe-11e7-902c-3aeb4fd24ac6.png">

Additionally either due to issues parsing the rgba string or due to Javascript's 31-bit limitation of bitwise math you can get weird results like this:

```python
r.glyph.color_mapper.nan_color = (255, 255, 255, 0.5)
```

<img width="279" alt="screen shot 2017-08-10 at 11 24 39 am" src="https://user-images.githubusercontent.com/1550771/29166533-a2ea4742-7dbe-11e7-8d1b-2049f8356e5e.png">

The output I would expect in this case looks like this:

<img width="258" alt="screen shot 2017-08-10 at 11 26 03 am" src="https://user-images.githubusercontent.com/1550771/29166549-ba5eb520-7dbe-11e7-8f38-ae201d5afa3d.png">

I have already worked up a fix since we need this functionality urgently and will submit a PR shortly.

jbednar:
Thanks for documenting this longstanding problem.

PR #6756, opened 2017-08-10
Add support for special color transparency in ColorMapper

philippjfr:
In https://github.com/bokeh/bokeh/issues/6755 I described an issue handling transparency for special colors (e.g. nan_color, low_color and high_color) when specifying the color as an RGBA tuple. It took me a very long time to work out that bitwise math in Javascript tops out at 31-bits, which means the previous approach to bit shifting the colors no longer worked and I had to use a bit of actual math to handle the alpha value correctly. Additionally the ``color2hex`` utility was not correctly converting the alpha value to a hex value.

I have also put together a notebook demonstrating the original issue and the new behavior with the fix: https://anaconda.org/philippjfr/bokeh_alpha/notebook

Let me know what tests you think I should add and whether you have any concerns with the approach I had to take.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/6755
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/mappers/color_mapper.coffee
+++ b/bokehjs/src/coffee/models/mappers/color_mapper.coffee
@@ -31,7 +31,7 @@ export class ColorMapper extends Transform
         ind = i*4
         # Bitwise math in JS is limited to 31-bits, to handle 32-bit value
         # this uses regular math to compute alpha instead (see issue #6755)
-        color[ind] = parseInt((value/4278190080.0) * 255)
+        color[ind] = Math.floor((value/4278190080.0) * 255)
         color[ind+1] = (value & 0xff0000) >> 16
         color[ind+2] = (value & 0xff00) >> 8
         color[ind+3] = value & 0xff
--- a/bokehjs/src/coffee/models/mappers/color_mapper.coffee
+++ b/bokehjs/src/coffee/models/mappers/color_mapper.coffee
@@ -29,6 +29,8 @@ export class ColorMapper extends Transform
       for i in [0...data.length]
         value = values[i]
         ind = i*4
+        # Bitwise math in JS is limited to 31-bits, to handle 32-bit value
+        # this uses regular math to compute alpha instead (see issue #6755)
         color[ind] = parseInt((value/4278190080.0) * 255)
         color[ind+1] = (value & 0xff0000) >> 16
         color[ind+2] = (value & 0xff00) >> 8
--- a/bokehjs/src/coffee/models/mappers/color_mapper.coffee
+++ b/bokehjs/src/coffee/models/mappers/color_mapper.coffee
@@ -24,17 +24,17 @@ export class ColorMapper extends Transform
   v_map_screen: (data, image_glyph=false) ->
     values = @_get_values(data, @_palette, image_glyph)
     buf = new ArrayBuffer(data.length * 4)
-    color = new Uint32Array(buf)
-
     if @_little_endian
+      color = new Uint8Array(buf)
       for i in [0...data.length]
         value = values[i]
-        color[i] =
-          (0xff << 24)                   | # alpha
-          ((value & 0xff0000) >> 16)     | # blue
-          (value & 0xff00)               | # green
-          ((value & 0xff) << 16);          # red
+        ind = i*4
+        color[ind] = parseInt((value/4278190080.0) * 255)
+        color[ind+1] = (value & 0xff0000) >> 16
+        color[ind+2] = (value & 0xff00) >> 8
+        color[ind+3] = value & 0xff
     else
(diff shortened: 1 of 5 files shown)


jbednar:
Looks great to me, thanks!

bokehjs/src/coffee/models/mappers/color_mapper.coffee
```diff
@@ -24,17 +24,17 @@ export class ColorMapper extends Transform
...
+      color = new Uint8Array(buf)
       for i in [0...data.length]
         value = values[i]
-        color[i] =
-          (0xff << 24)                   | # alpha
-          ((value & 0xff0000) >> 16)     | # blue
-          (value & 0xff00)               | # green
-          ((value & 0xff) << 16);          # red
+        ind = i*4
+        color[ind] = parseInt((value/4278190080.0) * 255)
```
bryevdv: This merits a comment explaining or linking back to the issue

philippjfr: Definitely a good idea given that this line alone cost me hours of confusion.

jbednar:
Not sure if you intended "use a bit of actual math" to be such a perfect pun in this context. :-)

bokehjs/src/coffee/models/mappers/color_mapper.coffee
```diff
@@ -24,17 +24,19 @@ export class ColorMapper extends Transform
...
         value = values[i]
-        color[i] =
-          (0xff << 24)                   | # alpha
-          ((value & 0xff0000) >> 16)     | # blue
-          (value & 0xff00)               | # green
-          ((value & 0xff) << 16);          # red
+        ind = i*4
+        # Bitwise math in JS is limited to 31-bits, to handle 32-bit value
+        # this uses regular math to compute alpha instead (see issue #6755)
+        color[ind] = parseInt((value/4278190080.0) * 255)
```
mattpap: Why `parseInt()`? Wouldn't `Math.trunc()` be more meaningful?

philippjfr: Yes, definitely, will push that in a minute.

bokehjs/src/coffee/models/mappers/color_mapper.coffee
```diff
@@ -24,17 +24,19 @@ export class ColorMapper extends Transform
...
         value = values[i]
-        color[i] =
-          (0xff << 24)                   | # alpha
-          ((value & 0xff0000) >> 16)     | # blue
-          (value & 0xff00)               | # green
-          ((value & 0xff) << 16);          # red
+        ind = i*4
+        # Bitwise math in JS is limited to 31-bits, to handle 32-bit value
+        # this uses regular math to compute alpha instead (see issue #6755)
+        color[ind] = Math.trunc((value/4278190080.0) * 255)
```
mattpap: Of course IE and phantomjs don't support this. @philippjfr, can you implement polyfill from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc in `core/util/math` and use it here?

philippjfr: Yay, JS being its usual fun self. Will try to get on that asap.

mattpap: This may be actually a solvable problem in typescript, because we can provide our own typings for the standard library, and explicitly say what we expect to be implemented by all supported runtimes. This way using `Math.trunc` would results in an error (e.g. "`Math` namespace doesn't implement `trunc` function").

philippjfr: Actually since values are guaranteed to be positive here I could just use ``Math.floor``. You okay with that or should I still do the polyfill?

bryevdv: I would vote for `Math.floor` just to avoid adding more code to BokehJS

bokehjs/src/coffee/models/mappers/color_mapper.coffee
```diff
@@ -24,17 +24,19 @@ export class ColorMapper extends Transform
...
+        # this uses regular math to compute alpha instead (see issue #6755)
+        color[ind] = Math.floor((value/4278190080.0) * 255)
+        color[ind+1] = (value & 0xff0000) >> 16
+        color[ind+2] = (value & 0xff00) >> 8
+        color[ind+3] = value & 0xff
     else
+      color = new Uint32Array(buf)
       for i in [0...data.length]
         value = values[i]
         color[i] = (value << 8) | 0xff     # alpha
```
mattpap: This should have been updated to reflect changes below.

PR #6759, opened 2017-08-10
Use screen coordinates in canvas layout

### LINKED ISSUE #6447: Switch layout's coordinate system from view to screen
Currently we use a mixture of view (origin is left-bottom corner) and screen (origin is left-top corner) coordinates. View coordinates originate from Cartesian frame, however "the rest of the world", so canvas, DOM, events, etc. are all in screen coordinates. To simplify implementation, screen coordinates should be used uniformly across bokehjs.

mattpap:


mattpap:
At this point this doesn't fix much. I will resubmit those changes when I have concrete progress on issue #6447.

PR #6764, opened 2017-08-12
Example test improvements

### LINKED ISSUE #6762: Random data in server examples causes image diff to fail
The following examples:

- models/server/widgets
- plotting/server/responsive
- plotting/server/selection_histogram

fail. File examples' tests configure a seed (stdlib, numpy) to generate stable images. If possible, this should be extended to server examples, or those shouldn't use random data. If nothing works, those examples shouldn't participate in image diff, but that's the final measure (no image diff makes examples pretty useless).

canavandl:
* [x] Seed random algos inside of examples
* [x] Remove session callbacks from documents inside of server examples
* [x] Enable app examples

- [ ] issues: fixes #6762
- [ ] tests added / passed

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -85,7 +85,8 @@ test:
     - icalendar
     - flask
     - flexx
-
+    - h5py
+    - scikit-learn
 
   commands:
     - bokeh -h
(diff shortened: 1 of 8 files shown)


canavandl:
I'm marking this is as ready. I've marked a couple of app examples as skipped due to lingering non-determinism or a (seemingly real) property issue. I'll address those in another PR

PR #6772, opened 2017-08-15
Add Figure.graph method

### LINKED ISSUE #6612: Task: Graph visualization improvements

canavandl:
Add Figure.graph method

- [x] issues: fixes #6612
- [x] tests added / passed


bokeh/plotting/helpers.py
```diff
@@ -68,6 +67,110 @@ def _stack(stackers, spec0, spec1, **kw):
...
+
+    node_renderer = GlyphRenderer(glyph=node_glyph,
+                                  nonselection_glyph=nsnode_glyph,
+                                  selection_glyph=snode_glyph,
+                                  hover_glyph=hnode_glyph,
+                                  muted_glyph=mnode_glyph,
+                                  data_source=node_source)
+
+    ## edge stuff
+    edge_kwargs = {k: v for k, v in kwargs.copy().items() if k in MultiLine.properties()}
```
canavandl: I'm unsure what the best route is here. Should all of the Glyph props be prefixed with `edge_` and `node_`?

The issue i'm running into is if a user does:
```python
plot.graph(..., line_width=10)
```
The line_width kwarg is applied to be the node (Circle) glyph and the edge (MultiLine) glyph.

bryevdv: > Should all of the Glyph props be prefixed with edge_ and node_?

I think that makes sense. Similar to e.g. `nonselection_line_color`

canavandl:
Have to merge after #6903 in order to example to work.

bokeh/plotting/helpers.py
```diff
@@ -28,8 +29,6 @@
 
 pd = import_optional('pandas')
 
-DEFAULT_PALETTE = ["#f22c40", "#5ab738", "#407ee7", "#df5320", "#00ad9c", "#c33ff3"]
```
canavandl: this appears to be dead code

bokeh/plotting/helpers.py
```diff
@@ -90,10 +193,6 @@ def get_default_color(plot=None):
         return colors[0]
 
 
-def get_default_alpha(plot=None):
```
canavandl: this appears to be dead code

bokeh/plotting/helpers.py
```diff
@@ -68,6 +67,110 @@ def _stack(stackers, spec0, spec1, **kw):
 
     return _kw
 
+def _graph(node_source, edge_source, **kwargs):
```
canavandl: This could be DRY-ed up a bit, but it's only repeated once so the benefit isn't huge.

examples/plotting/file/graphs.py
```diff
@@ -1,56 +1,51 @@
...
-gs = GridSpec((n + 1)//ncols, 1 + ncols)
-gs[0, 0] = graph_draw(g)
-
-for i, coloring in enumerate(colorings):
-    f = graph_draw(g, node_color=[ str(color) for color in coloring ], text_color="white")
-    gs[i//ncols, 1 + i%ncols] = f
-plot = gridplot(gs, toolbar_location=None)
-
-output_file("graphs.html", title="Graph k-coloring with computer algebra")
-show(plot)
```
mattpap: @canavandl, what was the rationale for removing this example?

bryevdv: To avoid confusing users with a file simply named `graphs.py` that demonstrated the way that most users in most cases should *not* do things any longer. If you want to re-rewrite this using the new graph renderer, or add it back as is but name it something like `explicit_manual_graph.py` with some comments explaining that this is now a low level approach with pointers to the docs for graph renderers that would be OK.

PR #6774, opened 2017-08-16
Have output_notebook register custom bokeh mimetype to renderer ``show`` msg

### LINKED ISSUE #6791: Task: Refactor Jupyter Notebook integration to use custom mimetypes
Task: Refactor Jupyter Notebook integration to use custom mimetypes

canavandl:
Have output_notebook register custom bokeh mimetype to renderer ``show`` msg

todo:
- [x] unbreak Zeppelin support (fix notebook_div return value)
- [x] implement cell clean-up logic
- [x] don't register mimetype in certain cases (not in classic notebook or already registered)
- [x] show server cells using custom mimetype
- [x] add server cell cleanup logic (done, but need to remove element from Bokeh.index)

old:
- [ ] issues: fixes #6791 
- [ ] tests added / passed


bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,102 @@
...
+      this.keyboard_manager.register_events(toinsert);
+      // Render to node
+      var props = {data, metadata: metadata[MIME_TYPE]}
+      render(props, toinsert[0])
+      element.append(toinsert)
+      return toinsert;
+    }
+
+    /* Handle when an output is cleared or removed */
+    output_area.events.on('clear_output.CodeCell', handleClearOutput);
```
canavandl: /* Handle when an output is cleared or removed */
  events.on('clear_output.CodeCell', handleClearOutput);
  events.on('delete.Cell', handleClearOutput);

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,102 @@
...
+
+    // function to render output of bk mime renderer
+    function append_mime(data, metadata, element) {
+      // create a DOM node to render to
+      var toinsert = this.create_output_subarea(
+        metadata,
+        CLASS_NAME,
+        MIME_TYPE
+      );
+      this.keyboard_manager.register_events(toinsert);
```
canavandl: i'm not 100% sure this does.

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,102 @@
...
+    script.textContent = props["data"]["script"]
+
+    node.appendChild(div);
+    node.appendChild(script);
+  }
+
+  /**
+   * Handle when an output is cleared or removed
+   */
+  function handleClearOutput(event, { cell: { output_area } }) {
```
mattpap: This syntax (destructuring) is not supported in IE. There are more instances to fix.

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,102 @@
...
+    function append_mime(data, metadata, element) {
+      // create a DOM node to render to
+      var toinsert = this.create_output_subarea(
+        metadata,
+        CLASS_NAME,
+        MIME_TYPE
+      );
+      this.keyboard_manager.register_events(toinsert);
+      // Render to node
+      var props = {data, metadata: metadata[MIME_TYPE]}
```
mattpap: This syntax as well (short object initializer).

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,102 @@
...
+  var CLASS_NAME = 'output_bokeh rendered_html';
+
+  /**
+   * Render data to the DOM node
+   */
+  function render(props, node) {
+    var div = document.createElement("div")
+    div.innerHTML = props["data"]["div"]
+    var script = document.createElement("script")
+    script.textContent = props["data"]["script"]
```
mattpap: Semicolon insertion is a tricky subject, so I would recommend being consistent with semicolons in JS. Either not use them at all, unless necessary (knowing the rules), or play it safe and using them everywhere. My approach is to not use semicolons in typescript (due to compiler's feedback) and use then in plain javascript.

bokeh/embed.py
```diff
@@ -326,10 +326,10 @@ def _bundle_for_objs_and_resources(objs, resources):
 
     return bokeh_js, bokeh_css
 
-def notebook_div(model, notebook_comms_target=None, theme=FromCurdoc):
-    ''' Return HTML for a div that will display a Bokeh plot in a
-    Jupyter/Zeppelin Notebook. notebook_comms_target is only supported
-    in Jupyter for now.
+def notebook_content(model, notebook_comms_target=None, theme=FromCurdoc):
```
canavandl: this is poorly tested.

bokeh/embed.py
```diff
@@ -379,6 +379,37 @@ def notebook_div(model, notebook_comms_target=None, theme=FromCurdoc):
     )
     div = _div_for_render_item(item)
 
+    return (js, div)
+
+
+def notebook_div(model, notebook_comms_target=None, theme=FromCurdoc):
```
canavandl: this is poorly tested.

bokeh/io.py
```diff
@@ -291,20 +295,20 @@ def _show_notebook_app_with_state(app, state, app_path, notebook_url):
...
-        watch_server_cells(_destroy_server_js)
-        state.watching_cells = True
-
     logging.basicConfig()
-    from IPython.display import HTML, display
     from tornado.ioloop import IOLoop
     from .server.server import Server
     loop = IOLoop.current()
     server = Server({app_path: app}, io_loop=loop, port=0,  allow_websocket_origin=[notebook_url])
+
```
canavandl: this is untested

bokeh/io.py
```diff
@@ -529,38 +534,24 @@ def _remove_roots(subplots):
...
-    return div_html.format(script=script, divid=divid)
-
-_destroy_server_js = """
-var cmd = "from bokeh import io; io._destroy_server('<%= destroyed_id %>')";
-var command = _.template(cmd)({destroyed_id:destroyed_id});
-Jupyter.notebook.kernel.execute(command);
-"""
-
-def _destroy_server(div_id):
+def _destroy_server(server_id):
```
canavandl: this is untested

bokeh/core/_templates/autoload_nb_js.js
canavandl: resolved

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,116 @@
...
+    var server_id = cell.output_area._server_id;
+    // Clean up Bokeh references
+    if (id !== undefined) {
+      Bokeh.index[id].remove();
+      delete Bokeh.index[id];
+    }
+
+    if (server_id !== undefined) {
+      // Clean up Bokeh references
+      var cmd = "from bokeh.io import _state; print(_state.uuid_to_server['" + server_id + "'].get_sessions()[0].document.roots[0]._id)";
```
canavandl: @bryevdv this is my solution for removing a server app from ``Bokeh.index``. I would appreciate help re: considering all of the implications of it's implementation.
* Is there going to be more than one session for a notebook app?
* Is there going to be more than one root for a server doc?

canavandl:
I'm marking this as ready. I'll try to add the missing pair of tests before merging. 

ping @mattpap , @bryevdv

bokeh/core/_templates/autoload_nb_js.js
bryevdv: I don't think there will ever normally be more than one session. In order for there to be multiple sessions a user would have to jump through many hoops, enough that I'd say they are on their on at that point. I am less certain that there would be only one root ever. Certainly that's the current common case but once we make it possible to template roots into different parts of the dom separately I can imagine folks wanting more sophisticated embedded apps

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,116 @@
...
+   */
+  function handleClearOutput(event, handle) {
+    var cell = handle.cell;
+
+    var id = cell.output_area._bokeh_element_id;
+    var server_id = cell.output_area._server_id;
+    // Clean up Bokeh references
+    if (id !== undefined) {
+      Bokeh.index[id].remove();
+      delete Bokeh.index[id];
```
canavandl: Is this the best way to clean up a Bokeh document? (cc @mattpap )

bokeh/core/_templates/autoload_nb_js.js
blink1073: I think it is meant for the `keyboard_manager` itself to enable/disable itself appropriately: https://github.com/jupyter/notebook/blob/0ccb548238fb884dbbf0b96492f48e1d980103db/notebook/static/notebook/js/keyboardmanager.js#L238.

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,116 @@
...
+  }
+
+  /**
+   * Handle when a new output is added
+   */
+  function handleAddOutput(event,  handle) {
+    var output_area = handle.output_area;
+    var output = handle.output;
+    // store reference to embed id or server id on output_area
+    output_area._bokeh_element_id = output.metadata[MIME_TYPE]["id"];
```
blink1073: Perhaps store this locally in the module rather than dangling attributes?  If not, the second should at least be called `_bokeh_server_id`.

bryevdv: The rationale for attaching it to the output area is that that is exactly what triggers the notification when it is added/removed, so all of the necessary information is collected locally where it is needed. Keeping it in the module would necessitate some more complicated mapping between all output areas and their respective corresponding server id's which seems more complicated for no gain.

bokeh/core/_templates/autoload_nb_js.js
canavandl: Testing both, it still doesn't seem like the document is being appropriated removed. I'm unsure what the solution is here.

bokeh/core/_templates/autoload_nb_js.js
bryevdv: Agree about a leading underscore, though

canavandl:
There's some weirdness where ``bokeh.io.show`` in the notebook reloads the entire autoload_nb_js bundle, instead of just whatever it needs to embed. I'm moving this back to WIP until I figure it out.

canavandl:
tested locally w/ static notebook output (via show), push_notebook and embedded server apps.

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,116 @@
...
+      /* Is output safe? */
+      safe: true,
+      /* Index of renderer in `output_area.display_order` */
+      index: 0
+    });
+  }
+
+  // register the mime type if in Jupyter Notebook environment and previously unregistered
+  var OutputArea = root.Jupyter.OutputArea;
+  if ((root.Jupyter !== undefined) && (!OutputArea.prototype.mime_types().includes(MIME_TYPE))) {
```
mattpap: `Array.includes()` is not supported in IE.

canavandl: [internal screaming]

thanks. i changed it to use ``indexOf``

zjffdu:
@canavandl I am working on show bokeh app in Apache Zeppelin. The remaining issue is to stop the server when cell is removed. After looking at the code of jupyter, I am not sure how to verify the server is stoped, could you guide me how to do that ?

blink1073:
@zjffdu, how are you stopping the notebook server?

bryevdv:
@blink1073 this is not about stopping the notebook server, it is about stopping a bokeh server that has been started on the IOLoop to show an embedded app. @zjffdu I'm not sure whar you mean by "verify" we have unit tests to make sure that `Server.stop` etc. do what they are supposed to. You just need to keep track of the servers you start and stop them when you are done with them (you can look at the `_destroy_servers` function that was removed e.g)

PR #6783, opened 2017-08-18
Added support for nbconvert preprocessors in NotebookHandler

jlstevens:
This PR extends the ``NotebookHandler`` which processes``.ipynb`` files in ``bokeh serve`` to support nbconvert preprocessors. There are two main aims:

* Strip out magics from notebooks so that the bokeh server does not choke on them. This will be beneficial to all bokeh users that serve notebooks, whether or not they are using HoloViews.
* Add the HoloViews preprocessors to support holoviews magics. This will make it much easier for notebooks using holoviews to use the bokeh server with minimal modification.

Here is an example of what this enables. Given the following notebook:

image

You can now run:

```
bokeh serve --show notebook.ipynb
```

To get:

image

Note that the holoviews magic setting the curve to 'red' is now respected due to the preprocessors used to turn the magics into standard Python syntax. I have tried to write this code fairly defensively so that this is only applied when holoviews is available, making sure that any error backs off the the old behavior of not using any preprocessors at all.

There is a correspond PR (ioam/holoviews#1811) where I am working on further improving the HoloViews preprocessors.

## Outstanding actions

- [ ] Discuss how preprocessors might be enabled/disabled in general.
- [ ] Add the ``StripMagicsProcessor`` to bokeh so that all bokeh users can take advantage of this step without needing holoviews installed.
- [ ] Explore ways the boilerplate code in the notebook can be further reduced.
- [ ] Add new unit tests both to bokeh and holoviews.

I have marked this PR as WIP for now as I am still experimenting with the concept.

### DIFF
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -82,5 +82,3 @@ class NotebookHandler(CodeHandler):
             return [OptsMagicProcessor(), OutputMagicProcessor(), StripMagicsProcessor()]
         except:
             return []
-
-
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -11,7 +11,7 @@ notebook code is executed, the Document being modified will be available as
 '''
 from __future__ import absolute_import, print_function
 
-from bokeh.util.dependencies import import_required
+from bokeh.util.dependencies import import_required, import_optional
 
 from .code import CodeHandler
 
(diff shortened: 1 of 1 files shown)


bokeh/application/handlers/notebook.py
```diff
@@ -32,17 +32,53 @@ def __init__(self, *args, **kwargs):
...
+        if holoviews is None:
+            return []
+
+        elif holoviews.__version__.release < (1,8,0):
+            return []
+        try:
+            from holoviews.ipython.preprocessors import (OptsMagicProcessor,
+                                                         OutputMagicProcessor,
+                                                         StripMagicsProcessor)
+            return [OptsMagicProcessor(), OutputMagicProcessor(), StripMagicsProcessor()]
```
bryevdv: I thought the consensus was that stripping magics should always be done unconditionally? Perhaps `StripMagicsProcessor` should be added to Bokeh directly and always added here?

bokeh/application/handlers/notebook.py
```diff
@@ -32,17 +32,53 @@ def __init__(self, *args, **kwargs):
...
         if 'filename' not in kwargs:
             raise ValueError('Must pass a filename to NotebookHandler')
         filename = kwargs['filename']
 
+        preprocessors = self._preprocessors()
+        try:
+            source, meta = self.export_python(filename, preprocessors)
+        except:
+            # If any preprocessors resulted in an exception, convert without them
+            source, meta = self.export_python(filename, preprocessors=[])
```
bryevdv: I the try/except necessary? This seems like it might silently cover up errors, or obscure the real source of problems.

bokeh/application/handlers/notebook.py
bryevdv: Also, I'd rather use duck typing here. This should check for `holoviews.get_notebook_preprocessors` or something, and just use the result of that instead of making a bunch of gorpy version checks of external libraries.

bokeh/application/handlers/notebook.py
```diff
@@ -32,17 +32,53 @@ def __init__(self, *args, **kwargs):
...
-            kwargs['source'] = source
+            for preprocessor in preprocessors:
+                exporter.register_preprocessor(preprocessor)
+            return exporter.from_notebook_node(nb)
 
-        super(NotebookHandler, self).__init__(*args, **kwargs)
+
+    def _preprocessors(self):
+        '''
+        Returns a list of suitable nbconvert preprocessors. Attempts to
```
bryevdv: docstrings start on the same line as the initial quotes, with a space between

bryevdv: Also width of up to 80 cols (this looks narrow) and a blank line at the end

bryevdv: maybe this should be a property as well?

bokeh/application/handlers/notebook.py
```diff
@@ -32,17 +32,53 @@ def __init__(self, *args, **kwargs):
...
+        nbconvert = import_required('nbconvert', 'The Bokeh notebook application handler requires Jupyter Notebook to be installed.')
+
         with open(filename) as f:
             nb = nbformat.read(f, nbformat.NO_CONVERT)
             exporter = nbconvert.PythonExporter()
-            source, meta = exporter.from_notebook_node(nb)
-            kwargs['source'] = source
+            for preprocessor in preprocessors:
+                exporter.register_preprocessor(preprocessor)
+            return exporter.from_notebook_node(nb)
```
bryevdv: Is this function really necessary? Esp if the try/except above can be removed.  It doesn't refer to anything though `self` why is it a method not a free function? Alternatively it could use `self.preprocessors` directly and not take it as a parameter.

bokeh/application/handlers/notebook.py
```diff
@@ -32,17 +32,53 @@ def __init__(self, *args, **kwargs):
...
+
+        kwargs['source'] = source
+        super(NotebookHandler, self).__init__(*args, **kwargs)
+
+
+    def export_python(self, filename, preprocessors):
+        '''
+        Given an .ipynb notebook, convert to plain Python syntax using
+        the supplied list of nbconvert preprocessors.
+        '''
```
bryevdv:
```
''' Given an .ipynb notebook, convert to plain Python using any supplied nbconvert 
preprocessors.

'''
```

bryevdv:
@jlstevens I want to clean up our PR list. Please submit a new PR if you intend to pursue this in the future.

jlstevens:
@bryevdv No problem! 

I still plan to do this but I agree this PR was left stagnating...

Issue #6603, opened 2017-07-11
Circle radius with radius_units='data' not round?


rsgoodwin:
#### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)
bokeh=12.4
python=3.6.1
notebook=5.0
os=Windows 10
browser=edge, chrome
monitor resolution = 4k and 3000x2000 tested.

#### Description of expected behavior and the observed behavior
*expected behavior*:  A circle with center 0,0  and a specific radius=r should extend in x and y direction such that xmin, xmax = ymin, ymax = +/- r

*actual behavior*:  The xmin, xmax = +/-r as expected,  however,  the ymin, ymax does not extend to +/-r.  It appears the conversion between screen pixels and plot units is different in x vs y.  

Note:  The square below is plotted just as a reference.  On the monitor,  the square is not actually square - it is physically longer in Y direction (measured with a ruler).  This my be part of the problem.  However,  one might expect that a circle and square drawn with equal centers such that length of square = diameter of circle would result in the circle being nested perfectly in the square, with tangents at the center of the squares sides.

Perhaps I need some other "data" in the plot to force the scale differently? 

#### Complete, minimal, self-contained example code that reproduces the issue

```
from bokeh.plotting import figure, output_notebook, show
output_notebook()
```

trial 1 = no plot width, height specified (or set to equal values)

```
p = figure()
p.rect(0, 0, 300, 300, line_color='black')
p.circle(x=0, y=0, radius=150, line_color='black', 
         fill_color='grey',  radius_units='data')
p.axis.minor_tick_out = 0
show(p)
```

Changing the plot dimensions,  can get circle to fit in square with ~these settings

```
p2 = figure(plot_width=500, plot_height=465)
p2.rect(0, 0, 300, 300, line_color='black')
p2.circle(x=0, y=0, radius=150, line_color='black', 
         fill_color='grey',  radius_units='data')
p2.axis.minor_tick_out = 0

show(p2)
```

#### Stack traceback and/or browser JavaScript console output
NA

#### Screenshots or screencasts of the bug in action
![image](https://user-images.githubusercontent.com/2280241/28071432-bfdc9f0c-6604-11e7-928c-99ce906f4f0f.png)

![image](https://user-images.githubusercontent.com/2280241/28072068-a549096c-6606-11e7-98f9-fa8e67dd4ea6.png)

bryevdv:
[...]

PR #6784, opened 2017-08-18
Basic aspect ratio control

### LINKED ISSUE #474: Easily control aspect ratio
Includes being able to set aspect (e.g. if you want `1:1` aspect) and also have resize and zoom tools, etc. respect the existing aspect ratio.

### LINKED ISSUE #517: Better control over data/screen aspect ratios needed for large circles
**UPDATE SUMMARY**

Te underlying browser API accepts one radius parameter (in pixels). If data/screen aspect ratios are not matched, then the the *data-space* radius in different dimensions will not match, because "10" in data units will be a different number of pixels in the x- vs y-dimensions. This can also cause issues for hit testing. Things that can be done:

* using something that scales in both dimensions (e.g. `p.ellipse` instead of circles)

* make the aspect ratios match up by controlling padding, labels size and orientation, and canvas size and data ranges explicitly. 

Making the aspect is a slight chore currently, but still doable. There is an open issue to make controlling aspects ratios easier to do.

In the event different aspect ratios are *necessary*, then this situation is mathematically unavoidable with canvas circles, and so they may simply *not be appropriate to use in that case* (i.e. use ellipses instead) 

---

As far as I can see, the issue is caused by a large (read: orders of magnitude) scale difference between the x and y coordinates. Here's an example, modified from `hover.py`, that triggers the issue in 0.4.2:

```

import numpy as np
from six.moves import zip
from collections import OrderedDict
from bokeh.plotting import *
from bokeh.objects import HoverTool, Range1d

TOOLS="pan,wheel_zoom,box_zoom,reset,hover,previewsave"

x = range(0, 1000, 10)
y = range(10000,20000,100) # <--

radii = [5]*100
colors = ['black']*100

source = ColumnDataSource(
    data=dict(
        x=x,
        y=y,
        radius=radii,
        colors=colors,
    )
)

output_file("hover.html")

hold()

circle('x', 'y', radius='radius', source=source, tools=TOOLS,
       fill_color='colors', fill_alpha=0.6,
       line_color=None, Title="Hoverful Scatter")


# We want to add some fields for the hover tool to interrogate, but first we
# have to get ahold of the tool. This will be made easier in future releases.
hover = [t for t in curplot().tools if isinstance(t, HoverTool)][0]

hover.tooltips = OrderedDict([
    ("index", "$index"),
    ("(x,y)", "($x, $y)"),
    ("radius", "@radius"),
])

show()  # open a browser
```

DuCorey:
## Summary
This PR adds a simple aspect ratio control tool.

- [x] issues: fixes #474, #517
- [ ] tests 

## Latest commit
This PR adds a two properties to the plot object.

`match_aspect` which indicates whether Bokeh should attempt the match the (width/height) of *data space* to the (width/height) in pixels of *screen space*.

`aspect_scale` which allows for further control of aspect control behavior by multiplying on top of the ratio calculated at `match_aspect`. This property allows the user to distort the axis at will. For example, with `aspect_scale= 2` a square in *data* units will be drawn as a rectangle with twice the *pixel* units in width than in height.

#### `match_aspect=True`
This fixes the bug shown in #6603.
bokeh_plot 32

#### `match_aspect=True, aspect_scale=2`
bokeh_plot 47

## Explanation

Hello Bokeh team,

From my understanding the current behavior when plotting the data ranges is to use the bounds of the glyphs. The method proposed in this PR for aspect control is to instead set the data ranges values with new ones calculated based on the glyph bounds and the inherent aspect ratio of the plot. It also allows an arbitrary user defined aspect ratio. 

#### Current behavior
bokeh_plot 42

#### New behavior with aspect ratio of 1:
bokeh_plot 40

#### New behavior with aspect ratio of 16/9:
bokeh_plot 41

### What it doesn't fix
This isn't a perfect fix, since it simply uses the built in ranges for trying to maintain the aspect ratio. As described in #517, there seems to be a deeper problem with pixel per distance being different on each axis. I'm currently studying the information in those threads to see if my current solution can be improved.

### Todo
- [ ] tests 

I'm assuming a new feature like this requires tests and unfortunately I haven't used mocha before, but I will do my best to get the tests done.

### Final remarks
I hope this PR is appreciated. It's always of pleasure contributing to a project like Bokeh. Feel free to dismiss this PR if you are working on something more comprehensive that also fixes issues like #6603 and #517. I've also left some more comments on specific parts about the code. I'm welcome to all suggestions or improvements. I'll be available to work some more on the PR after the weekend.

To the whole team, enjoy your weekend and thank you for such a great project.
Corey

### DIFF
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -695,7 +695,8 @@ class Plot(LayoutDOM):
 
     aspect_ratio = Either(Bool, Float, default=False, help="""
     Specify the aspect ratio behavior of the plot. Default is ``False`` which
-    means that the glyph will fill the whole plot area. ``True`` indicates that
-    the glyphs should maintain the an aspect ratio of 1. A ``Float``value can
-    also be given for arbitrary aspect_ratio control.
+    means that axis range will be set to the glyph's bounds. ``True`` indicates
+    that the axis range should maintain an aspect ratio of 1. A ``Float`` value
+    can also be given for arbitrary aspect_ratio control. Aspect ratio is
+    defined as the ratio of width over height.
     """)
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import
 
 from ..core.enums import Location, OutputBackend
-from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String
+from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String, Either, Float
 from ..core.property_mixins import LineProps, FillProps
 from ..core.query import find
 from ..core.validation import error, warning
@@ -691,4 +691,11 @@ class Plot(LayoutDOM):
     .. note::
         When set to ``webgl``, glyphs without a WebGL rendering implementation
         will fall back to rendering onto 2D canvas.
+    """)
+
+    aspect_ratio = Either(Bool, Float, default=False, help="""
+    Specify the aspect ratio behavior of the plot. Default is ``False`` which
+    means that the glyph will fill the whole plot area. ``True`` indicates that
+    the glyphs should maintain the an aspect ratio of 1. A ``Float``value can
+    also be given for arbitrary aspect_ratio control.
     """)
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -693,12 +693,34 @@ class Plot(LayoutDOM):
         will fall back to rendering onto 2D canvas.
     """)
 
-    aspect = Either(Auto, Float, default=None, help="""
+    match_aspect = Bool(default=False, help="""
     Specify the aspect ratio behavior of the plot. Aspect ratio is defined as
-    the physical distance relationship between a figure's width and height. It
-    is defined by the ratio of the width over the height.
+    the ratio of width over height. This property controls whether Bokeh should
+    attempt the match the (width/height) of *data space* to the (width/height)
+    in pixels of *screen space*.
 
-    Default is ``None`` which indicates no aspect ratio. ``"auto"`` indicates
-    that the plot should maintain an aspect ratio of 1. A ``Float`` value can
-    be given for arbitrary aspect ratio control.
+    Default is ``False`` which indicates that the *data* aspect ratio and the
+    *screen* aspect ratio vary independently. ``True`` indicates that the plot
+    aspect ratio of the axes will match the aspect ratio of the pixel extent
+    the axes. The end result is that a 1x1 area in data space is a square in
+    pixels, and conversely that a 1x1 pixel is a square in data units.
+
+    .. note::
+        This setting only takes effect when there are two dataranges. This
+        setting only sets the initial plot draw and subsequent resets. It is
+        possible for tools (single axis zoom, unconstrained box zoom) to
+        change the aspect ratio.
+    """)
+
+    aspect_scale = Float(default=1, help="""
+    A value to be given for increased aspect ratio control. This value is added
+    multiplicatively to the calculated value required for ``match_aspect``.
+    ``aspect_scale`` is defined as the ratio of width over height of the figure.
+
+    For example, a plot with ``aspect_scale`` value of 2 will result in a
+    square in *data units* to be drawn on the screen as a rectangle with a
+    pixel width twice as long as its pixel height.
+
+    .. note::
+        This setting only takes effect if ``match_aspect`` is set to ``True``.
     """)
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import
 
 from ..core.enums import Location, OutputBackend
-from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String, Either, Float, Auto
+from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String, Float
 from ..core.property_mixins import LineProps, FillProps
 from ..core.query import find
 from ..core.validation import error, warning
--- a/bokeh/models/plots.py
+++ b/bokeh/models/plots.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import
 
 from ..core.enums import Location, OutputBackend
-from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String, Either, Float
+from ..core.properties import Bool, Dict, Enum, Include, Instance, Int, List, Override, String, Either, Float, Auto
 from ..core.property_mixins import LineProps, FillProps
 from ..core.query import find
 from ..core.validation import error, warning
@@ -693,10 +693,12 @@ class Plot(LayoutDOM):
         will fall back to rendering onto 2D canvas.
     """)
 
-    aspect_ratio = Either(Bool, Float, default=False, help="""
-    Specify the aspect ratio behavior of the plot. Default is ``False`` which
-    means that axis range will be set to the glyph's bounds. ``True`` indicates
-    that the axis range should maintain an aspect ratio of 1. A ``Float`` value
-    can also be given for arbitrary aspect_ratio control. Aspect ratio is
-    defined as the ratio of width over height.
+    aspect = Either(Auto, Float, default=None, help="""
+    Specify the aspect ratio behavior of the plot. Aspect ratio is defined as
+    the physical distance relationship between a figure's width and height. It
+    is defined by the ratio of the width over the height.
+
+    Default is ``None`` which indicates no aspect ratio. ``"auto"`` indicates
+    that the plot should maintain an aspect ratio of 1. A ``Float`` value can
+    be given for arbitrary aspect ratio control.
     """)
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -210,9 +210,8 @@ export class PlotCanvasView extends DOMView
     follow_enabled = false
     has_bounds = false
 
-    if @model.plot.aspect? and @frame._width.value != 0 and @frame._height.value != 0
-      aspect = if @model.plot.aspect == "auto" then 1 else @model.plot.aspect
-      r = 1/aspect*(@frame._width.value/@frame._height.value)
+    if @model.plot.match_aspect != false and @frame._width.value != 0 and @frame._height.value != 0
+      r = 1/@model.plot.aspect_scale*(@frame._width.value/@frame._height.value)
 
       for k, v of bounds
         width = v.maxX - v.minX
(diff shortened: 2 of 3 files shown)


bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -210,6 +210,32 @@ export class PlotCanvasView extends DOMView
     follow_enabled = false
     has_bounds = false
 
+    if @model.plot.aspect_ratio != false
+      aspect = if @model.plot.aspect_ratio is true then 1 else @model.plot.aspect_ratio
```
DuCorey: Should there be more specific checks for the float number?

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -210,6 +210,32 @@ export class PlotCanvasView extends DOMView
...
+      r = aspect*(@model.plot.width/@model.plot.height)
+
+      for k, v of bounds
```
DuCorey: I'm still a beginner in Coffescript, but is comprehension the correct way of looping through the object? Would closing the scope be preferred? i.e. 
```
for k, v of bounds 
  do(k, v) -> 
```

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -210,6 +210,32 @@ export class PlotCanvasView extends DOMView
...
+
+        if width < r*height
+          width = r*height
+        else
+          height = width/r
+
+        bounds[k].maxX = xcenter+0.5*width
+        bounds[k].minX = xcenter-0.5*width
+        bounds[k].maxY = ycenter+0.5*height
+        bounds[k].minY = ycenter-0.5*height
```
DuCorey: Would it be better to create a new dictionary with new values instead of overwriting bounds?

bryevdv:
Whoa, unexpected and awesome

> I hope this PR is appreciated.

An enthusiastic YES. If I may wax slightly, this is the power of OSS. One reason that this issue has lingered for so long is that I had framed the problem in entirely different terms, namely: given fixed/determined data bounds, how can the layout (i.e. the actual pixel dimensions of the central plot area) be adjusted dynamically to maintain some constraints. I imagined lots of bad cases and it made getting started very difficult. But you've come at it from the opposite direction, which never occurred to me at all (taking the pixel bounds as determined and adjusting the data ranges appropriately). But I think this approach here (and limiting things to when there are two data ranges) is actually *much* simpler as well as more promising!

We are in a bit of a crush right before the next release freeze, so my inclination is to let this simmer for the next week or two. That will allow for time to add tests, tweak docs, etc without being rushed. I also think the pixel-distance issue might be fixable with a small delta off this code, so we could potentially discuss and iterate on that as well. Then the goal would be to merge and get it into a dev build soon after `0.12.7` comes out. Does that sound like a reasonable plan of attack?

Again, thank you for the extremely pleasant surprise.

bryevdv:
Basically for the pixel distance problem, off hand I might suggest the following: add another property option `iso` (or maybe `auto`) which simply looks at the pixel aspect ratio of the "central" plot frame, and then automatically uses that as the numerical aspect ratio when updating data ranges. 

The reason that per-pixel distances may not match in different directions is that the aspect ration of the pixel dimensions is different from the aspect ratio of the "data" dimensions. But te former is easily obtainable, and you just opened up an easy path to setting the latter.

bokehjs/src/coffee/models/plots/plot_canvas.coffee
```diff
@@ -210,6 +210,32 @@ export class PlotCanvasView extends DOMView
     follow_enabled = false
     has_bounds = false
 
+    if @model.plot.aspect_ratio != false
+      aspect = if @model.plot.aspect_ratio is true then 1 else @model.plot.aspect_ratio
+      r = aspect*(@model.plot.width/@model.plot.height)
```
bryevdv: Pretty sure this should be using the plot frame's "inner width" and "inner height", those are the pixel dimensions of the area between the axes, i.e.  not counting any space taken up by title, axes, tick labels, padding, etc.

bryevdv:
Another reason to take time is we will need to figure out how things like interactions with tools should behave. If an aspect is set to something specific, and a user tries to zoom just one axis... what happens? Maybe that's fine, and aspect is only applied when autoranging, so if a user explicily changes the bounds (e.g. by zooming) then it's fine to ignore the aspect constraint until a reset. Mostly we just need to make sure we understand the behaviours, and document it well.

bokeh/models/plots.py
```diff
@@ -692,3 +692,10 @@ def webgl(self, webgl):
         When set to ``webgl``, glyphs without a WebGL rendering implementation
         will fall back to rendering onto 2D canvas.
     """)
+
+    aspect_ratio = Either(Bool, Float, default=False, help="""
+    Specify the aspect ratio behavior of the plot. Default is ``False`` which
+    means that the glyph will fill the whole plot area. ``True`` indicates that
+    the glyphs should maintain the an aspect ratio of 1. A ``Float``value can
```
bryevdv: lack of space after last backtick is what is causing the one test failure (sphinx barfs)

cwhy:
I have not tried it yet but why in the "New behavior with aspect ratio of 1" example plot have a aspect ratio that is not exactly 1:1?

bryevdv:
> I have not tried it yet but why in the "New behavior with aspect ratio of 1" example plot have a aspect ratio that is not exactly 1:1?

As I commented above, it's using the overall canvas pixel dimensions, instead of the pixel dimensions of just the axes that border the plot area, and I believe that is an issue.

DuCorey:
Good news, following @bryevdv suggestions of using the `inner_width` and `inner_height` works perfectly now for setting the aspect ratio. Measuring the pixels on both axis, I'm finding they match up perfectly. I've updated the examples in the PR. 

Second, getting access to the values of `inner_width` and `inner_height` was a little more tricky. From my understanding of the code, the `inner_width` and `inner_height` are set from the `frame._width` and `frame._height` properties which are in turn calculated by `update_constraints()` in the `render()` function of the model. Unfortunately `update_dataranges()` is called in the init before `frame._width` and `frame._height` have been calculated. To go around this, I've added `update_dataranges()` again in the `render()` function. This now properly calculates the range values with aspect ratio. However, from running the code on my side seems to be called multiple times and portions of the `update_dataranges()` are being run unnecessarily. I'm unsure how best to proceed to avoid unnecessary calls to `update_dataranges()`. Let me know what you think.

bryevdv:
>  However, from running the code on my side seems to be called multiple times and portions of the update_dataranges() are being run unnecessarily. 

It's possible, given that the data bounds depend on the pixel dims of the axes which in turn depend on the extents of the tick labels which depends on the data bounds that iterating/calling twice may be unavoidable

bryevdv:
```

p = figure(aspect_ratio=1, plot_height=400)

p.circle([2], [2], radius=0.4)

show(p)
```
screen shot 2017-08-21 at 13 13 59


Hooray! 

@DuCorey What do you think about changing `Bool` to `Auto`. Then the default is `None` (`null` on the JS side`) which means don't do anything to set aspect (corresponds to `False` in the current code) and `"auto"` means match the plot area's aspect (corresponds to `True`)?

Also I think I'd have a slight preference for `aspect` instead of `aspect_ratio`

bryevdv:
Also this behaves like I expected when zooming just one axis or the other:

screen shot 2017-08-21 at 13 18 42
screen shot 2017-08-21 at 13 18 54

This seems fine, we just need to make sure to document that aspect matching only happens on initial auto-ranging and resets. Operations like single axis zoom explicitly will *not* preserve aspect.

bryevdv:
lastly! it would be ideal to factor out the code that computes new bounds based on pixel aspect into its own little function and add some tests for it. I can help with that if necessary.

bryevdv:
Also regarding re-calling `update_dataranges` that could be skipped if `aspect` is `None` (false)

bryevdv:
@DuCorey do you mind if I push changes to your branch? I would like to try and get this into 0.12.7 after all

bryevdv:
Alternatively if you have time to make these two changes today:

* rename `aspect_ratio` to `aspect` 

* use `Auto` instead of `Bool`

Then this PR could be merged  and I could add tests and docs in a separate PR. 

I understand if you don't have time on short notice, which is why I am also happy to push changes to your branch directly.

DuCorey:
@bryevdv Changes are done. However, even with the added check we are still calling update_dataranges() twice. Going through the code execution, the behavior is consistent because when we call `update_dataranges() `and update the x and y ranges, this calls a request_paint() with `final = false` which then recalculates both the constraints and the updates the ranges however since the ranges aren't being changed this time around the `request_paint()` is sent with` final = true` and the code goes elsewhere. I haven't tried changing this logic.

Second, I've inversed the behavior of aspect to now increase in width when we increase the aspect ratio. To me this makes more sense, since looking at definitions of aspect ratio online we see that for example an aspect ratio of 16:9 indicates a ratio of width over height of 1.77. The method now does the same thing. It sets the ranges such as the physical width over height of plot ranges are equal to 1.77. I think most users using this feature will expect such a behavior.

bryevdv:
@DuCorey thanks! It's almost certain we are "over-calling" things in a few different places. There is an open issue about `_map_data` being called more than necessary, e.g. It'll be best to try and investigate and improve these things in their own PR. (Performance generally seems adequate despite over-calling)

> Second, I've inversed the behavior of aspect to now increase in width when we increase the aspect ratio

Yes, good catch, aspect is normally defined by width/height

bryevdv:
OK tests are green I will plan on kicking the tires just a bit then merging later this afteroon

bryevdv:
Well tests *were* green until I accidentally hit the "close" button I've stopped the new test run, no sense in taking up test workers.

bokeh/models/plots.py
```diff
@@ -692,3 +692,13 @@ def webgl(self, webgl):
...
     """)
+
+    aspect = Either(Auto, Float, default=None, help="""
+    Specify the aspect ratio behavior of the plot. Aspect ratio is defined as
+    the physical distance relationship between a figure's width and height. It
+    is defined by the ratio of the width over the height.
+
+    Default is ``None`` which indicates no aspect ratio. ``"auto"`` indicates
+    that the plot should maintain an aspect ratio of 1. A ``Float`` value can
+    be given for arbitrary aspect ratio control.
```
bryevdv: We can edit this in a follow-on PR before the release, but this is a little off. `"auto"` indicates that the aspect ratio of the axes dimensions should be automatically made to match the aspect ratio of the corresponding pixel extents of the axes. The end result is that a 1x1 area in data space is a square in pixels, and conversely that a 1x1 pixel is a square in data units, or (perhaps more importantly) that circles drawn inside squares will always touch all four sides. 

We should also add a note about how this setting only affect when there are two dataranges, and then only affects initial plot draw and resets -- it is possible for the use of tools (single axis zoom or unconstrained bo zoom) to change the aspect ratio

bryevdv:
> This shouldn't be merged without image diff or integration tests.

Plan is to add tests and more docs in follow-on PRs during the freeze so that this can go in an RC tomorrow morning. I'm merging when travis goes green.

bryevdv:
@canavandl I'm not sure I've ever added a new image diff integration test from scratch. Can we pair on that in the office in the next day or two? Then I will update those docs with more detail in the docs PR. I have asked @DuCorey for an `examples/plotting/file/aspect.py` in a follow on PR that will get picked up by the image diff tests. (I think we need a dedicated integration test too, though)

cc @mattpap

bryevdv:
@mattpap Additionally, in any case integration tests will really need to be added by someone on @bokeh/dev in order to make sure they actually function

PR #6799, opened 2017-08-22
add mechanism to install external notebook display hooks

### LINKED ISSUE #6788: Make notebook display extendible by external libraries
I'd like to embed bokeh app into Apache Zeppelin just as embed bokeh app in Jupyter

bryevdv:
- [x] issues: fixes #6788
- [x] tests added / passed

cc @canavandl @zjffdu @mattpap @clairetang6

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -77,7 +77,7 @@ def install_notebook_hook(notebook_type, load, show_doc, show_app, overwrite=Fal
             A function for loading BokehJS in a notebook type. The function
             will be called with the following arguments:
 
-            .. code-block:: pythonm
+            .. code-block:: python
 
                 load(
                     resources,   # A Resources object for how to load BokehJS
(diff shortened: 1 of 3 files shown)


bryevdv:
The main idea is that the `io._notebook` dict maps notebook type names to sets of functions for displaying in the notebook. then when `show` is called it looks up the right functions based on notebook type and whether it is a standalone doc or app that is being shown.

bryevdv:
I think this PR needs a tiny bit more work to add a hook action for "load" so that the zeppelin specific code in`util.notebook` can be removed eventually but I will do that after merging your PR @canavandl

bryevdv:
@canavandl @mattpap this is ready

bryevdv:
This also includes a small fix for the nb/lab mimetype PR. I'd like to merge this tonight then RC tomorrow morning.

PR #6801, opened 2017-08-22
Derive Slider.format from other arguments

### LINKED ISSUE #6800: Automatic configuration of Slider.format
Integer sliders should use integer formatting.

mattpap:
fixes #6800

### DIFF
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import
 
 from ...core.has_props import abstract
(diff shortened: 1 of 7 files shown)


mattpap:
@bryevdv, given the default formatting, this still doesn't format years in sliders the way you want it. We can get rid of `,` as thousandths separator, though it would be better to not treat years sliders as arbitrary integer sliders.

bryevdv:
Personal preference I would rather not have any thousands separator on by default (in general)

mattpap:
Superseded by PR #6804.

PR #6805, opened 2017-08-23
Position labels relative to frame if in center region

### LINKED ISSUE #6786: Label using screen units bound to frame is positioned relative to canvas
E.g. `plot.add_layout(Label(x=0, y=0, x_units="screen", y_units="screen"))` will be hidden behind below panels (if any).

mattpap:
fixes #6786


mattpap:
@bryevdv, this fixes my example from the tutorial. I'm not sure though if this will break other examples. If so, just go ahead with RC, because it may need some deeper changes.

bryevdv:
I ran other examples that used label and they all seemed to function as expected, so I am inclined to merge for RC, we can cleanly revert if need be

mattpap:
@bryevdv, yeah, there are no image differences (finally!).

PR #6807, opened 2017-08-23
Add plotting/file/eclipse example

### LINKED ISSUE #6796: Add examples from PyCon.PL tutorial

mattpap:
fixes #6796


examples/plotting/file/eclipse.py
```diff
@@ -0,0 +1,75 @@
...
+    color_mapper=mapper,
+    location="bottom_left", orientation="horizontal",
+    title="Popularity of \"solar eclipse\" search term",
+    title_text_font_size="12pt", title_text_font_style="bold",
+    background_fill_alpha=0.0)
+p.add_layout(color_bar)
+
+notes = Label(
+    x=0, y=0, x_units="screen", y_units="screen",
+    x_offset=40, y_offset=20,
```
mattpap: Originally, prior to PR #6805, I had to use the following hack:
```
p.extra_x_ranges["x_screen"] = Range1d(0, 1, bounds="auto")
p.extra_y_ranges["y_screen"] = Range1d(0, 1, bounds="auto")
notes = Label(
    x=0, y=0, x_offset=40, y_offset=20,
    x_range_name="x_screen",
    y_range_name="y_screen",
    ...)
```

examples/plotting/file/eclipse.py
```diff
@@ -0,0 +1,75 @@
...
+
+mapper = LinearColorMapper(palette=list(reversed(YlOrRd5)), low=0, high=100)
+
+source = ColumnDataSource(data=dict(
+    state_xs=list(states.lons),
+    state_ys=list(states.lats),
+    trend=states.trend,
+))
+us = p.patches("state_xs", "state_ys",
+    fill_color=dict(field="trend", transform=mapper),
```
mattpap: Is it possible to configure a transform without need for using explicit data source and field references?

examples/plotting/file/eclipse.py
```diff
@@ -0,0 +1,75 @@
...
+p.grid.grid_line_color = None
+
+p.title.text = "Google Search Trends and the Path of Solar Eclipse, 21 August 2017"
+p.title.align = "center"
+p.title.text_font_size = "16pt"
+
+mapper = LinearColorMapper(palette=list(reversed(YlOrRd5)), low=0, high=100)
+
+source = ColumnDataSource(data=dict(
+    state_xs=list(states.lons),
```
mattpap: I'm not sure what's going on here, but if I remove `list()`, then I get:
```
ValueError: Out of range float values are not JSON compliant: nan
```

examples/plotting/file/eclipse.py
bryevdv: there is some initial work in the new `bokeh.transform` that we can expand upon

examples/plotting/file/eclipse.py
mattpap: @bryevdv, do you have any immediate idea what's going on here?

bryevdv: no but it will have to wait, suggest another issue

mattpap: Sure. This "out of range float" has been popping out here and there for a while now.

PR #6811, opened 2017-08-25
Call load_notebook via notebook hook

### LINKED ISSUE #6810: Call load_notebook via notebook hook
In `outpute_notebook`, we should call

 `_run_notebook_hook(_state.notebook_type, 'load', resources, verbose, hide_banner, load_timeout)`
instead of calling load_notebook directly,

zjffdu:
- [ ] issues: fixes #6810 
- [ ] tests added / passed

It is just a trivial and straightforward change to allow non-builtin notebook works


bryevdv:
@zjffdu this has two test failing tests with:
```
        if notebook_type not in _notebook_hooks:
>           raise RuntimeError("no display hook installed for notebook type %r" % notebook_type)
E           RuntimeError: no display hook installed for notebook type <Mock name='mock.notebook_type' id='140566165867600'>
```
You can see full output here: https://travis-ci.org/bokeh/bokeh/jobs/268207362

The easy solution (and fine with me) is to change the string string comparison that is messed mum by mocks to at `startswith("no display hook installed for notebook type")`

zjffdu:
@bryevdv I think the failed test is not due to string comparision. It is due to `State` is mocked. So I changed 
```
_run_notebook_hook(_state.notebook_type, ...
```
to 
```
_run_notebook_hook(notebook_type, ...
```

And besides that, I found another issue that the patched `bokeh.io.load_notebook` will not be called because the hook registration happens before the test. So I explicitly call the registration again in test to ensure the patched `bokeh.io.load_notebook` is registered properly.

bryevdv:
Thanks @zjffdu

PR #6812, opened 2017-08-25
use separate display_data to display status div

### LINKED ISSUE #6809: Inline, Minified Resources do not work in classic notebooks
This is due to an interaction with the classic notebooks use of JQuery, when output is published as `text/html`. New notebook code published a div and a script together as `text/html`. Propose to solve by publishing a single script as `application/javascript` (which should work) that creates the necessary div itself

bryevdv:
- [x] issues: fixes #6809
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/_templates/autoload_nb_js.js
+++ b/bokeh/core/_templates/autoload_nb_js.js
@@ -140,7 +140,7 @@
   function display_loaded() {
     if (root.Bokeh !== undefined) {
       var el = document.getElementById({{ elementid|json }});
-      if (el !== undefined) {
+      if (el != null) {
         el.textContent = "BokehJS " + Bokeh.version + " successfully loaded.";
       }
     } else if (Date.now() < root._bokeh_timeout) {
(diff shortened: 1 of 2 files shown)


bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -138,7 +140,9 @@
   function display_loaded() {
     if (root.Bokeh !== undefined) {
       var el = document.getElementById({{ elementid|json }});
-      el.textContent = "BokehJS " + Bokeh.version + " successfully loaded.";
+      if (el !== undefined) {
```
mattpap: `getElementById()` returns `null` if element is not found. You should use `el != null` idiom here.

bryevdv:
@mattpap Thanks, FYI see flowdock for plans re: this miss and getting a release out

canavandl:
This PR causes ``undefined`` to be printed in the output_notebook cell in Jupyter Lab. This could presumably be handled in the JLab extension though.

screen shot 2017-08-25 at 10 45 10 am

canavandl:
And apparently a file save error in JLab (presumably related to the ``undefined`` div)

screen shot 2017-08-25 at 10 46 50 am

bryevdv:
@canavandl any ideas about the errors above?

PR #6814, opened 2017-08-25
Add aspect example

DuCorey:
Add example for the new aspect ratio control behavior introduced in #6784 

- [ ] tests added / passed

### DIFF
--- a/examples/plotting/file/aspect.py
+++ b/examples/plotting/file/aspect.py
@@ -1,0 +1,15 @@
+from bokeh.plotting import figure, output_file, show
+from bokeh.layouts import row
+
+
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks @DuCorey !

mattpap:
I think it would be useful to add other plots showing various aspect ratios from #6784.

DuCorey:
@mattpap Would you like examples that distort the plot dimensions as well as show various aspect ratios?

bryevdv:
@DuCorey I think having a good selection do different scenarios is good, e.g tall/wide plots etc

mattpap:
@DuCorey, yes, this is what I had in mind.

bryevdv:
Thanks @DuCorey !

PR #6818, opened 2017-08-27
Have notebook mime renderer use application/javascript mime message for nbconvert compat

canavandl:
Have notebook mime renderer use application/javascript mime message for nbconvert compat

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bokeh/io.py
```diff
@@ -436,7 +433,9 @@ def _show_file_with_state(obj, state, new, controller):
 
 def _show_jupyter_doc_with_state(obj, state, notebook_handle):
     comms_target = make_id() if notebook_handle else None
-    publish_display_data({'text/html': notebook_div(obj, comms_target)})
+    (script, div) = notebook_content(obj, comms_target)
+    publish_display_data({HTML_MIME_TYPE: div})
+    publish_display_data({JS_MIME_TYPE: script, EXEC_MIME_TYPE: ""}, metadata={EXEC_MIME_TYPE: {"id": obj._id}})
```
canavandl: publishing two messages here isn't ideal. I should refactor the ``script`` to create and embed a div.

canavandl:
In the current classic notebooks seem to work. However only nbconvert=5.1.1 seems to work for properly creating an html doc of the notebook. Using nbconvert=5.2.1 causes the ``application/javascript`` mime message that's published from ``bokeh.io.show`` to be stripped out and the plot doesn't render.

Here's a gist of an executed notebook: https://gist.github.com/canavandl/0c6e6ed79cec3b459c4f808f4a1bc6b5
Here's the properly converted html w/ nbconvert=5.1.1: https://gist.github.com/canavandl/44165de06479d68fa57084ff86bd6897
Here's the unrendered plot html w/ nbconvert=5.2.1: https://gist.github.com/canavandl/1f1e58ba99eebc98da4d9cd91b5ec776

You can search the latter two for "output_subarea" to see that the JS content is stripped out in the 5.2.1 converted html. It should be possible to download the notebook and use nbconvert directly to reproduce.

damianavila:
Pinging @mpacer, the lead nbconvert dev. Can you help us with this M? Thanks!

mpacer:
I'm not sure why this is happening… 

When I run nbconvert 5.2.1 on this it still has that application/javascript information. I checked with @choldgraf this also is the case for him.

I'm guessing that there might be a config file that you have defined somewhere that is overwriting the (`display_data_priority`)[https://github.com/jupyter/nbconvert/blob/2bb7994d8f93c114913b95c0be18f5d8f3008743/nbconvert/utils/base.py#L16] traitlet value. Otherwise I can't figure out where this issue is coming from.

bryevdv:
I will try to reproduce this with 5.11 locally later today. @mpacer where might an `nbconvert` config file as you describe be located?

bryevdv:
OK, I can confirm that I can also successfully convert the test notebook above with nbconvert 5.2.1, @canavandl maybe try on a clean system or VM? We should try to figure out what's different about your system tho too.

mpacer:
If it were me, I'd use https://github.com/mpacer/jupyter_conf_search (`pip install jupyter_conf_search`)… it's not perfect but there are a lot of places you'd need to look.

canavandl:
thanks @mpacer for the ``jupyter_conf_search`` protip. I was pulling in some config from my root env that was causing things not to work. I think we know what to do to move forward now.

bokeh/core/_templates/autoload_nb_js.js
```diff
@@ -1,5 +1,129 @@
...
+      /* Is output safe? */
+      safe: true,
+      /* Index of renderer in `output_area.display_order` */
+      index: 0
+    });
+  }
+
+  // register the mime type if in Jupyter Notebook environment and previously unregistered
+
+  if (root.Jupyter !== undefined) {
```
canavandl: required so that our classic notebook mime renderer isn't registered in JLab or nbconvert environments.

bokeh/core/_templates/onload_js.js
```diff
@@ -0,0 +1,12 @@
+var id = setInterval(embed_document, 100, window);
```
canavandl: This is a potential solution for embedding bokeh docs in notebooks/nbconvert situations where it's necessary to wait until Bokeh is attached to the window. I don't know if it's better quit trying after some number of intervals, in that case that loading BokehJS has totally failed.

canavandl: If we keep this, I'll dedupe this and DOCS_JS.js

bokeh/io.py
bryevdv: I think to two messages is fine. Now that I realize it's an option, I'd personally rather keep the JS scripts simpler in any way possible.

bokeh/tests/test_embed.py
```diff
@@ -111,6 +111,13 @@ def test_output_is_without_script_tag_when_wrap_script_is_false(self, mock_make_
         self.maxDiff = None
         self.assertEqual(rawscript.strip(), script_content.strip())
 
+class TestNotebookContent(unittest.TestCase):
+
+    def test_return_type(self):
+        r = embed.notebook_content(_embed_test_plot)
+        self.assertTrue(isinstance(r, tuple))
+        self.assertTrue(len(r), 2)
+
```
bryevdv: More can be tested here? The script and idv are both probably amend to aaic t least some b`startwith` testing, or the div could be parsed entirely with beatiful soup and have its structure and expected attributes verified

bokeh/core/_templates/onload_js.js
bryevdv: Hrm maybe, sure. But 100ms startup for every plot is alot. If a notebook has 100 plots that will not be negligible. would suggest 10ms max. It would maybe be better to split into `embed_document` and a helper that does the render. Then `embed_document` can try to render immediately (by calling the helper), and only if it fails the first attempt, add the helper on an interval.

bokeh/io.py
canavandl: I'm leaning towards agreeing at this point. Especially since notebook apps don't require a div, but a script that the mime renderer handles.

bokeh/core/_templates/onload_js.js
bryevdv: Also def think there should be a max timeout

bokeh/tests/test_embed.py
canavandl: I'll likely refactor most of these tests.

bokeh/core/_templates/doc_js.js
```diff
@@ -1,4 +1,24 @@
...
+    var timer = setInterval(function(root) {
+      if (root.Bokeh !== undefined) {
+        embed_document(root);
+        clearInterval(timer);
+      }
+      attempts++;
+      if (attempts > 10) {
+        clearInterval(timer);
+      }
+    }, 100, root)
```
canavandl: unsure of best interval/number of attempts

bryevdv: I'd suggest skewing the other direction, 10ms and 100 attempts. Tho if Bokeh exists it might do to clear the interval first before the embed. We can always try something out and tune it more later

bokeh/core/_templates/doc_js.js
```diff
@@ -1,4 +1,25 @@
...
+      if (root.Bokeh !== undefined) {
+        embed_document(root);
+        clearInterval(timer);
+      }
+      attempts++;
+      if (attempts > 10) {
+        console.log("Bokeh: ERROR: Unable to embed document because BokehJS library is missing")
+        clearInterval(timer);
+      }
+    }, 100, root)
```
bryevdv: I still think 10ms interval with 100 tries is better

mattpap: In my subjective view, 100ms is already a visible delay. Given that the test is cheap, then I would go for more fine-grained intervals.

bryevdv: cc @canavandl

canavandl: Fixed

PR #6828, opened 2017-08-28
Don't fail silently when inline resources aren't available

### LINKED ISSUE #6827: Don't fail silently when inline resources aren't available
Currently we get empty `<script>` and `<style>` tags in generated HTML. This should pretty much revert #644.

mattpap:
fixes #6827


bryevdv:
good change

PR #6830, opened 2017-08-28
Use setv to silent set data_source.inspected (not not trigger render)

### LINKED ISSUE #6829: Degraded HoverTool performance in 0.12.7rc4
Using the examples/plotting/file/texas.py example, there's notable lagging of the tooltip from the cursor. Setting logging to debug it appears that the canvas is redrawing on every mousemove.

Bokeh=0.12.7rc4

canavandl:
Use setv to silent set data_source.inspected (not not trigger render)

- [ ] issues: fixes #6829 
- [ ] tests added / passed


bryevdv:
LGTM we need to just make sure to always test `texas.py` specifically, often, until we can devise a better way to test this. Perhaps the `setv` call can be put inside another function just for the purpose of mocking it in some tests.

mattpap:
LGTM, though in future, we may need to reconsider this liberal approach of acting on any changes to objects' properties.

canavandl:
We were just discussing having things act on changes to data_source.selected instead of the real data_source.select event. Perhaps something we can address in the future.

mattpap:
> Perhaps something we can address in the future.

Definitively. I've been dealing with a similar issue in layout for a while.

PR #6832, opened 2017-08-29
Simplify data frame length calculation.

bje-:

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Thanks for the PR @bje- !

PR #6834, opened 2017-08-29
Use default action when detecting file name in jupyter

### LINKED ISSUE #6560: using ``export_png`` or ``save`` without filename from within jupyter notebook saves png file to lib/python
Using ``bokeh.io.export_png``, ``bokeh.io.export_svg`` or ``bokeh.io.save`` from within a jupyter notebook without specifying an output filename causes the file to be written to lib/python (on my machine it's '/Users/lcanavan/miniconda3/envs/bokeh/lib/python3.5/runpy.png'.

It's likely due to how if there's no filename arg, we infer the name of the script the document is in. A notebook is apparently executed by this ``runpy.py`` script in lib/python, so that's the generated name.

Is there a consistent way to get the path of a jupyter notebook from inside the session?

Bokeh 0.12.7dev3

mattpap:
This covers common cases where previously detected file names were unhelpful (e.g. `runpy.html`).

fixes #6560

### DIFF
--- a/bokeh/io.py
+++ b/bokeh/io.py
@@ -489,15 +489,34 @@ def save(obj, filename=None, resources=None, title=None, state=None, **kwargs):
     _save_helper(obj, filename, resources, title)
     return os.path.abspath(filename)
 
-def _detect_filename(ext):
-    """ Detect filename from the name of the script being run. Returns
-    temporary file if the script could not be found or the location of the
-    script does not have write permission (e.g. interactive mode).
-    """
-    import sys
-    import inspect
-    from os.path import dirname, basename, splitext, join
+def _no_access(basedir):
+    ''' Return True if the given base dir is not accessible or writeable
 
+    '''
+    return not os.access(basedir, os.W_OK | os.X_OK)
+
+def _shares_exec_prefix(basedir):
+    ''' Whether a give base directory is on the system exex prefix
+
+    '''
+    import sys
+    prefix = sys.exec_prefix
+    return (prefix is not None and basedir.startswith(prefix))
+
+def _temp_filename(ext):
+    ''' Generate a temporary, writable filename with the given extension
+
+    '''
+    return tempfile.NamedTemporaryFile(suffix="." + ext).name
+
+def _detect_current_filename():
+    ''' Attempt to return the filename of the currently running Python process
+
+    Returns None if the filename cannot be detected.
+    '''
+    import inspect
+
+    filename = None
     frame = inspect.currentframe()
     try:
         while frame.f_back and frame.f_globals.get('name') != '__main__':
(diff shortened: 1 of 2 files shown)


bryevdv:
@mattpap  @hsparra is interested in this change, can you provide some concrete examples of operation? Will this always prevent writing output into site packages? Or only when the user does not have write permissions there?

mattpap:
This will prevent usage of auto-detected name if it's on the executable path. This works in all scenarios that unfolded so far. I tested this with notebook, ipython, runpy and in a conda environment.

bryevdv:
OK, I took a closer look and LGTM tho if there is any way to split out some smaller pieces such that they could have unit tests it would be even better.

bryevdv:
@mattpap please feel free to delete my commit if you don't approve. It splits the function up into small pieces that can be tested (sometimes with mocks) and adds docsrings. In general as we head to 1.0 I'd like for functionality like this to be rigidly locked down under test.

bryevdv:
@mattpap are you OK with my changes?

bokeh/io.py
```diff
@@ -489,25 +489,79 @@ def save(obj, filename=None, resources=None, title=None, state=None, **kwargs):
...
+    basedir = dirname(filename) or os.getcwd()
 
-    if filename is None or not os.access(dirname(filename) or curdir, os.W_OK | os.X_OK):
-        return tempfile.NamedTemporaryFile(suffix="." + ext).name
+    if _no_access(basedir) or _shares_exec_prefix(basedir):
+        return _temp_filename(ext)
 
-    name, _ = splitext(basename(filename))
-    return join(dirname(filename), name + "." + ext)
+    name, current_ext = splitext(basename(filename))
```
mattpap: `current_ext` is unused. I'm surprised that flake doesn't complain about this.

bokeh/tests/test_io.py
```diff
@@ -411,3 +411,78 @@ def test__get_svgs_with_svg_present_with_driver():
...
+def test__shares_exec_prefix():
+    import sys
+    old_ex = sys.exec_prefix
+    sys.exec_prefix = "/foo/bar"
+    assert io._shares_exec_prefix("/foo/bar") == True
+    sys.exec_prefix = "/baz/bar"
+    assert io._shares_exec_prefix("/foo/bar") == False
+    sys.exec_prefix = None
+    assert io._shares_exec_prefix("/foo/bar") == False
+    sys.exec_prefix = old_ex
```
mattpap: This needs to be in `try-finally` block (or use some py.test magic). Otherwise it can influence other tests under failure.

bokeh/io.py
```diff
@@ -489,25 +489,79 @@ def save(obj, filename=None, resources=None, title=None, state=None, **kwargs):
...
-    while frame.f_back and frame.f_globals.get('name') != '__main__':
-        frame = frame.f_back
+    try:
+        while frame.f_back and frame.f_globals.get('name') != '__main__':
+            frame = frame.f_back
+
+        filename = frame.f_globals.get('__file__')
+    finally:
+        del frame
+        return filename
```
mattpap: This isn't a big issue, but it would recommend not using `return` statement in `finally` block. This is because the body `finally` block doesn't participate normal execution path, so it may be confusing when someone adds (perhaps accidentally) another `return` statement.

bokeh/tests/test_io.py
```diff
@@ -411,3 +411,78 @@ def test__get_svgs_with_svg_present_with_driver():
...
+    io._detect_current_filename = lambda : None
+    io._no_access = lambda x: False
+    io._shares_exec_prefix = lambda x: False
+    r = io.default_filename("test")
+    assert r != "/a/b/foo.test"
+    assert r.endswith(".test")
+
+    io._detect_current_filename = old__detect_current_filename
+    io._no_access = old__no_access
+    io._shares_exec_prefix = old__shares_exec_prefix
```
mattpap: Similar story here. One failure above and there will be a number of unrelated failures in other tests.

mattpap:
@bryevdv, yes, just tests need to be made more robust.

bryevdv:
@mattpap thanks, I believe I have made all the suggested changes

mattpap:
LGTM. Apparently PR author can't approve their own PRs, even when modified by other.

bryevdv:
That's a bit annoying but I guess I understand. Anyone should merge this when green. (whoever sees it first)

PR #6837, opened 2017-08-29
Remove MPL and bokeh.charts

bryevdv:
This PR removes all support for MPL and `bkcharts` combat.


bryevdv:
@mattpap thanks for the catches

PR #6838, opened 2017-08-29
Don't test on Python 3.4 anymore

### LINKED ISSUE #6819: Switch examples tests to use Python 3.5 or 3.6
I recall that there was a reason for sticking with 3.4, but I don't remember what exactly. At this point 3.4 is not supported by Anaconda, and there are no updates available to certain packages. For example, in PR #6807 I need pyshp in most recent version, but that's available only for 2.7 and 3.5+. Perhaps it's time to use more recent version of Python 3 as the basis for testing.

mattpap:
fixes #6819

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ jobs:
       script: scripts/ci/test
       env:
       - GROUP=examples
-      - PYTHON=3.4
+      - PYTHON=3.5
       - EXTRA_CMD=--no-js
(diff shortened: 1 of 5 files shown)


bryevdv:
@mattpap let's merge the mpl/charts PR first to get the release notes file, then you can merge/rebase on top and add an explicit mention that we are no longer testing on python 3.4

PR #6839, opened 2017-08-29
Make Grid.dimension an enum

### LINKED ISSUE #1000: Specify grid dimensions with an enum
If we don't support more than two dimensions, then it should take an enum.

mattpap:
fixes #1000

### DIFF
--- a/bokeh/core/tests/test_query.py
+++ b/bokeh/core/tests/test_query.py
@@ -39,8 +39,8 @@ def large_plot():
     plot.add_layout(LogAxis(), 'left')
     plot.add_layout(LinearAxis(y_range_name="liny"), 'left')
 
-    plot.add_layout(Grid(dimension=0), 'left')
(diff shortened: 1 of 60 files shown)


bryevdv:
This needs a migration note about the deprecation. 

I also wonder if there aren't better enums:
```
plot.add_layout(Grid(dimension="height", ticker=yticker))
```
That is for a grid with *horizontal* lines (that happen at the tick location given by the *y* ticker) I wonder if horizontal/vertical to describe the dimension the grid line spans might be clearer (with horizontal going with y tickers, and vertical going with x tickers). i.e. if someone says to me "I want a horizontal grid" the first thing I will imagine is a grid with horizontal lines.

bryevdv:
ping @mattpap we should decide this and get it merged

mattpap:
After all, I'm not sure if I really want to proceed with this.

PR #6854, opened 2017-08-31
Enable notebook examples in examples tests

### LINKED ISSUE #6853: Enable notebook examples in test_examples
Enable notebook examples in test_examples. Currently they're all skipped.

Bokeh 0.12.7

canavandl:
Enable notebook examples in examples tests

- [ ] issues: fixes #6853 
- [ ] tests added / passed


scripts/ci/test:examples
```diff
@@ -10,6 +10,9 @@ if [[ $TRAVIS_COMMIT_MSG == *"[ci disable examples]"* ]]; then
     exit 0
 fi
 
+jupyter notebook --generate-config
+echo "c.NotebookApp.token = u''" >> ~/.jupyter/jupyter_notebook_config.py
```
mattpap: I presume this can't be configured on the command line?

canavandl: This was the Jupyter team's suggested approach. I wish i could do something like ``jupyter notebook --use-token=false`` but that doesn't appear possible.

tests/examples/test_examples.py
```diff
@@ -93,10 +93,8 @@ def test_server_examples(server_example, example, report, bokeh_server):
         else:
             _get_pdiff(example)
 
-
-### {{{ THIS IS BROKEN and all examples are skipped in examples.yaml
 @pytest.mark.examples
-def test_notebook_examples(notebook_example, example, jupyter_notebook, report):
+def test_notebook_examples(notebook_example, example, jupyter_notebook, bokeh_server, report):
```
canavandl: Have to add the ``bokeh_server`` test fixture so that jupyter notebooks can load the resources.

scripts/ci/test:examples
canavandl: I lied. Looking at their docs, ``jupyter notebook --NotebookApp.token=""`` should work.

mattpap: Nice to know.

canavandl:
Here's an example report: https://s3.amazonaws.com/bokeh-travis/270565989/examples.html#plotting/notebook/color_scatterplot

One unresolved issue is knowing when a notebook with plots in multiple cells is done being executed. Our current screenshot code looks for the first document in ``window.Bokeh.documents`` to be idle, so we're potentially missing later plots being rendered.

Here's an example of that: https://s3.amazonaws.com/bokeh-travis/0.12.7-fdfeddf/howto/notebook_comms/Basic%20Usage.png

Perhaps skipping such situations would be a short-term solution.

canavandl:
The "Basic Usage.ipynb" example would have surfaced our regression in ``bokeh.io.push_notebook``.

mattpap:
Examples using random numbers need to be seeded somehow.

canavandl:
> Examples using random numbers need to be seeded somehow.

good catch. i'll do that now.

canavandl:
I'm currently skipping all howto/notebooks until we decide how to handle notebooks w/ multiple plot cells. I'll open a separate issue to track.

canavandl:
Merging master in hopes that the syntax error goes away.

bryevdv:
Still not-failing failing. I'd suggest we should get the test to actually fail as-is first, then figure out why it's failing.

bryevdv:
I'm going to go ahead and close this as stale

PR #6856, opened 2017-08-31
Checks for finite values of bounds before doing the calculation for match_aspect bounds

### LINKED ISSUE #6851: Plots which contain glyph's with infinite bounds fail to set initial ranges when match_aspect is set to True
From my understanding, some glyphs can have infinite bounds when they are added to a plot without any initial data, but the plot still has a well defined glyph. 

In the current implementation of match_aspect, calculations will be done on these +- Infinity values and return NaNs which then cause the failure to set the initial ranges.

I'll be pushing a PR, by the end of this week to rectify this issue. My current idea is to simply skip calculations on those infinity values.

DuCorey:
# Summary
Checks for finite values of bounds before doing the calculation for `match_aspect `bounds.
- [x] issues: fixes #6851
- [ ] tests added / passed

## Explanation
When trying to plot a figure which contains a glyph without specified bounds. The default value given to these glyphs is Infinity. If you also specified the `match_aspect` for the figure, it would fail to draw the initial ranges since the computation done to to insure `match_aspect` would turn the `Infinity` values into `NaN`. This fixes the issue by checking that all values are finite before doing any calculation for the bounds. 

## Example
```
p = figure(match_aspect=True)
source = ColumnDataSource(data={'x': [], 'y': []})
p.circle(x=1, y=2)
```
Without the fix the ranges fail to set. However, with `match_aspect=False` the figure will set the ranges properly.

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.coffee
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.coffee
@@ -212,28 +212,28 @@ export class PlotCanvasView extends DOMView
 
     if @model.plot.match_aspect != false and @frame._width.value != 0 and @frame._height.value != 0
       r = 1/@model.plot.aspect_scale*(@frame._width.value/@frame._height.value)
-
       for k, v of bounds
-        width = v.maxX - v.minX
-        if width <= 0
-          width = 1.0
+        if isFinite(v.maxX) and isFinite(v.minX) and isFinite(v.maxY) and isFinite(v.minY)
+          width = v.maxX - v.minX
+          if width <= 0
+            width = 1.0
 
-        height = v.maxY - v.minY
-        if height <= 0
-          height = 1.0
+          height = v.maxY - v.minY
+          if height <= 0
+            height = 1.0
 
-        xcenter = 0.5*(v.maxX + v.minX)
-        ycenter = 0.5*(v.maxY + v.minY)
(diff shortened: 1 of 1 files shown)


bryevdv:
@DuCorey did you want to add the `is_empty` method to bbox, or merge this as-is? I am ok with either.

DuCorey:
@bryevdv Will push up an updated version adding the `is_empty `method in the next few days.

DuCorey:
Apologies, I've tried implementing the functionality as we discussed in the comments. Unfortunately, the way we discussed doing this doesn't seem to be working on my end. I'm unsure whether this issue stems from my lack of understanding in JavaScript or the code as it is currently. 

From looking at the inspector, it seems that the data being passed to bounds is simply an object without inheritance and is not a `BBox` class. Any methods I add to `BBox` are simply not defined for the bounds object. 

More specifically, from my understanding, bounds are gotten by calling the `bounds()` function in glyphs which returns in turns calls the `get bbox` from the `RBush` class. `get bbox` does not return a `BBox` class as defined in `bbox.ts` but simply a `Rect` type. 

I'd rather leave it to someone more familiar with the inner-workings of bokehjs to fix such an issue. I'd recommend just pulling my first fix as it stands.

bokeh-integrations:
Sounds good to me, thanks @DuCorey !

bryevdv:
(that was me)

PR #6857, opened 2017-08-31
Fix positioning of axis label

### LINKED ISSUE #6472: Make layout canvas' panels not overlap in corners
Currently panels overlap in corners, which makes e.g. labels hard to use. My proposal is to make horizontal (above, below) panels extend to full canvas length and vertical (left, right) to extend only to frame length (sandwich between above and below panels). This would work nicely with titles and would preserve all examples in their existing form (as far as I can tell). We could consider making this configurable, but I'm not sure if it's useful. If one wonders why axis don't overlap, it's because their rendering is limited to frame size, but their panels are overlapping nevertheless.

### LINKED ISSUE #6850: Log-axis label layout issue
Bokeh 0.12.7 seems to introduce a layout error when adding an title to a log-axis. The axis-title is suddenly shifted upwards and cut off.

Using the log-axis example as basis and adding an axis-title seems to replicate the problem:

**MWE**
```
import numpy as np

x = np.linspace(0.1, 5, 100)

p = figure(title="log axis example", y_axis_type="log",
           y_range=(0.001, 10**22))

p.line(x, np.sqrt(x), legend="y=sqrt(x)",
       line_color="tomato", line_dash="dotdash")

p.line(x, x, legend="y=x")
p.circle(x, x, legend="y=x")

p.line(x, x**2, legend="y=x**2")
p.circle(x, x**2, legend="y=x**2",
         fill_color=None, line_color="olivedrab")

p.line(x, 10**x, legend="y=10^x",
       line_color="gold", line_width=2)

p.line(x, x**x, legend="y=x^x",
       line_dash="dotted", line_color="indigo", line_width=2)

p.line(x, 10**(x**2), legend="y=10^(x^2)",
       line_color="coral", line_dash="dashed", line_width=2)

p.yaxis.axis_label = 'New ylabel'  # Here is the problem

output_file("test.html", title="log plot example")

show(p)  # open a browser
```

**Screenshot**
screenshot

Let me know if additional information is needed

mattpap:
This takes advantage of axis' panel to position axis label. ~~Note that currently this results in incorrect positioning, because panels are aligned to canvas and not frame (see #6472). I intentionally went with the correct solution and not using frame for alignment, so that we finally start using proper tooling (assuming #6472 is fixed).~~

fixes #6850
fixes #6472

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -156,7 +156,8 @@ class Legend(Annotation):
     """)
 
     padding = Int(10, help="""
-    Amount of padding around the contents of the legend.
+    Amount of padding around the contents of the legend. Only applicable when
+    when border is visible, otherwise collapses to 0.
     """)
 
     spacing = Int(3, help="""
--- a/bokehjs/src/coffee/core/layout/layout_canvas.coffee
+++ b/bokehjs/src/coffee/core/layout/layout_canvas.coffee
@@ -13,6 +13,8 @@ export class LayoutCanvas extends Model
     @_height = new Variable("#{@toString()}.height")
     @_right = new Variable("#{@toString()}.right")
     @_bottom = new Variable("#{@toString()}.bottom")
+    @_hcenter = new Variable("#{@toString()}.hcenter")
+    @_vcenter = new Variable("#{@toString()}.vcenter")
 
   get_editables: () ->
     return [@_width, @_height]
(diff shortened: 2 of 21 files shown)


rocheseb:
Does it also fix labels showing up in the top left corner here ?

```
from bokeh.io import show

fig = figure()
fig.xaxis.axis_label = "xaxis label"
fig.yaxis.axis_label = "yaxis label"

show(fig)
```

mattpap:
> Does it also fix labels showing up in the top left corner here ?

Yes. I'm actually surprised any labels are shown in an empty figure with the current implementation.

mattpap:
I finally figured out how to fully define panel constraints and not collapse the frame. This in turn allowed to fix overlapping panels issue (#6472). All this was holding me back for weeks. I will need to do some tweaking around legends, but otherwise image diff looks good after such an intrusive change.

bryevdv:
@mattpap can you attach a screen shot

mattpap:
and new example `models/file/panels`:

Note that some of the titles are misaligned, but that's title's fault, and something to be improved later.

mattpap:
On a side note, travis ci has been so slow and unresponsive recently.

Issue #6831, opened 2017-08-29
color slider example can show hex FP values


bryevdv:
Not sure if it's in the example or in Bokeh

<img width="830" alt="screen shot 2017-08-28 at 19 04 59" src="https://user-images.githubusercontent.com/1078448/29798646-dfc31c5c-8c23-11e7-80bf-a44a321aaccb.png">

crashMOGWAI:
Looks like the sliders can return a float on occasion, haven't nailed it down yet. In the mean time, I've modified the callback to prohibit this behavior. 

```
callback = CustomJS(args=dict(source=source), code="""
    function componentToHex(c) {
        var hex = c.toString(16);
        return hex.length == 1 ? "0" + hex : hex;
    }
    function rgbToHex(r, g, b) {
        return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
    }
    function toInt(value) {
       return value | 0; 
    }
    var data = source.data;
    color = data['color'];
    text_color = data['text_color'];
    var R = toInt(red_slider.value);
    var G = toInt(green_slider.value);
    var B = toInt(blue_slider.value);
    color[0] = rgbToHex(R, G, B);
    text_color[0] = '#ffffff';
    if ((R > 127) || (G > 127) || (B > 127)) {
        text_color[0] = '#000000';
    }
    source.change.emit();
""")
```

bryevdv:
It may be that slider values sholud always just be considered float, and any more specialized interpretation is left up to users (e.g. as your PR) but perhaps we can special case when start/end/stop are all ints and cast the value just in that case or something. I will let the PR close this issue, though. (can open a new issue if you have a follow on PR)

crashMOGWAI:
Most likely just an artifact of float imprecision, which is expected behavior. 
```
In[1]: 0.99999999999999998 == 1.0
Out[1]: True
```

bryevdv:
@mattpap this diff fixes, is OK?
```
diff --git a/bokehjs/src/coffee/models/widgets/slider.coffee b/bokehjs/src/coffee/models/widgets/slider.coffee
index 6eaa633af..a2021b5b1 100644
--- a/bokehjs/src/coffee/models/widgets/slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/slider.coffee
@@ -12,7 +12,11 @@ export class SliderView extends AbstractSliderView
       step: @model.step
     }

-  _calc_from: ([value]) -> value
[...]

PR #6860, opened 2017-09-01
Patch color_sliders.py

crashMOGWAI:
Partially addresses #6831 by preventing non int values from being passed to hex conversion

### DIFF
--- a/examples/plotting/file/color_sliders.py
+++ b/examples/plotting/file/color_sliders.py
@@ -62,7 +62,7 @@ callback = CustomJS(args=dict(source=source), code="""
         return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
     }
     function toInt(v) {
-       return v | 0; 
+       return v | 0;
     }
     var data = source.data;
     color = data['color'];
--- a/examples/plotting/file/color_sliders.py
+++ b/examples/plotting/file/color_sliders.py
@@ -61,12 +61,15 @@ callback = CustomJS(args=dict(source=source), code="""
     function rgbToHex(r, g, b) {
         return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
     }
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the fix! will merge when green

bryevdv:
@crashMOGWAI the one tests failure is from our linter:
```
E         File contains trailing whitespace: examples/plotting/file/color_sliders.py, line 65.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: examples/plotting/file/color_sliders.py, line 65.'])
```
Once you remove the trailing whitespace on that line and push a new commit the tests will go green

mattpap:
Although this helps with the example, this doesn't fix the underlying cause. I would rather investigate why numerical slider with integer start and step, end can return inexact values. Otherwise this issue will pop up elsewhere sooner than later.

bryevdv:
I'm going to go ahead and merge this so the example works better for the release, but leave the issue open to solve the general problem

bryevdv:
@mattpap this diff fixes, is OK?
```
diff --git a/bokehjs/src/coffee/models/widgets/slider.coffee b/bokehjs/src/coffee/models/widgets/slider.coffee
index 6eaa633af..a2021b5b1 100644
--- a/bokehjs/src/coffee/models/widgets/slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/slider.coffee
@@ -12,7 +12,11 @@ export class SliderView extends AbstractSliderView
       step: @model.step
     }

-  _calc_from: ([value]) -> value
+  _calc_from: ([value]) ->
+    if Number.isInteger(@model.start) and Number.isInteger(@model.end) and Number.isInteger(@model.step)
+      return value|0
+    else
+      value

 export class Slider extends AbstractSlider
   type: "Slider"
```

PR #6866, opened 2017-09-02
Allow supplying explicit edge paths for graphs

### LINKED ISSUE #6865: Allow supplying explicit edge paths to GraphRenderer
Currently when plotting a graph using the GraphRenderer and StaticLayoutProvider there is no way to supply an explicit set of paths because StaticLayoutProvider draws straight line paths from the start and end points. It is trivial to allow supplying explicit ``xs`` and ``ys`` for the MultiLine glyph, so I will open a PR to allow that shortly.

philippjfr:
When plotting a graph using the GraphRenderer and StaticLayoutProvider there is no way to supply an explicit set of paths because StaticLayoutProvider draws straight line paths from the start and end points. This PR allows supplying explicit 'xs' and 'ys' to the ``MultiLine`` CDS, which the ``StaticLayoutProvider`` will use rather than computing the straight line paths. Since the ``StaticLayoutProvider`` already expects ``start`` and ``end`` columns I think it's fine and consistent if it also looks for xs and ys, although I'd be open to respecting names set for 'xs' and 'ys' on the MultiLine glyph. Still needs tests and documentation. 

- [x] issues: fixes #6865 
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/graphs/static_layout_provider.coffee
+++ b/bokehjs/src/coffee/models/graphs/static_layout_provider.coffee
@@ -14,6 +14,9 @@ export class StaticLayoutProvider extends LayoutProvider
     return [xs, ys]
 
   get_edge_coordinates: (edge_source) ->
+    if edge_source.data.xs? and edge_source.data.ys?
+      return [edge_source.data.xs, edge_source.data.ys]
+
     [xs, ys] = [[], []]
     starts = edge_source.data.start
     ends = edge_source.data.end
--- a/bokehjs/src/coffee/models/graphs/static_layout_provider.coffee
+++ b/bokehjs/src/coffee/models/graphs/static_layout_provider.coffee
@@ -14,19 +14,22 @@ export class StaticLayoutProvider extends LayoutProvider
     return [xs, ys]
 
   get_edge_coordinates: (edge_source) ->
-    if edge_source.data.xs? and edge_source.data.ys?
-      return [edge_source.data.xs, edge_source.data.ys]
-
     [xs, ys] = [[], []]
     starts = edge_source.data.start
     ends = edge_source.data.end
+    has_paths = edge_source.data.xs? and edge_source.data.ys?
     for i in [0...starts.length]
-      if @graph_layout[starts[i]]? and @graph_layout[ends[i]]?
-        [start, end] = [@graph_layout[starts[i]], @graph_layout[ends[i]]]
+      in_layout = @graph_layout[starts[i]]? and @graph_layout[ends[i]]?
+      if has_paths and in_layout
+        xs.push(edge_source.data.xs[i])
+        ys.push(edge_source.data.ys[i])
       else
-        [start, end] = [[NaN, NaN], [NaN, NaN]]
-      xs.push([start[0], end[0]])
-      ys.push([start[1], end[1]])
+        if in_layout
+          [start, end] = [@graph_layout[starts[i]], @graph_layout[ends[i]]]
+        else
+          [start, end] = [[NaN, NaN], [NaN, NaN]]
+        xs.push([start[0], end[0]])
+        ys.push([start[1], end[1]])
     return [xs, ys]
 
   @define {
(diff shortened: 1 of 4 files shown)


jbednar:
Sounds great.  Can you show an example of it in use?

philippjfr:
Ping @canavandl to provide feedback. I'll work on some examples shortly. Once I have one should I add it to the User Guide section?

bryevdv:
> Once I have one should I add it to the User Guide section?

Yes please do

philippjfr:
Does this seem like a reasonable example? I just adapted one of the existing examples and computed bezier paths between the nodes:

```python
import math

N = 8
node_indices = list(range(N))

graph = GraphRenderer()

    index=node_indices,
    fill_color=Spectral8)

    start=[0]*N,
    end=node_indices)

### start of layout code

### Draw quadratic bezier paths
def bezier(sx, sy, ex, ey, bx, by, steps=100):
    xs, ys = [], []
    for t in np.linspace(0, 1, steps):
        xs.append((1-t)*(1-t)*sx + 2*(1-t)*t*bx+t*t*ex)
        ys.append((1-t)*(1-t)*sy + 2*(1-t)*t*by+t*t*ey)
    return xs, ys

sx, sy = graph_layout[0]
xs, ys = [], []
for node_index in node_indices:
    ex, ey = graph_layout[node_index]
    x, y = bezier(sx, sy, ex, ey, 0, 0)
    xs.append(x)
    ys.append(y)
graph.edge_renderer.data_source.data['xs'] = xs
graph.edge_renderer.data_source.data['ys'] = ys

plot.renderers.append(graph)

show(plot)
```

screen shot 2017-09-05 at 2 19 38 pm

jbednar:
That example looks perfect, thanks!

canavandl:
I like the concept of optionally supplying some edge coordinates that supply the curves then falling back to straight lines. There's just a couple of conceptual questions that I want to address:

* By storing the node coordinates on the StaticLayoutProvider we're able to support masking parts of the network (by not including some nodes in the graph_layout) and/or "animating" node locations by only changing the layout provider. How could we get this approach to support those features? Right now, it appears that an edge would still be drawn to a masked node.
* Perhaps being part of the solution to above, could/should edge coordinates belong to the layout provider model?
* I also wanted to point out that it's possible to provide discontinuous, unconnected or otherwise non-sensical edges via this strategy. I'm not sure it's possible to allow users to provide line coordinates without also allowing them to draw "stupid" edges, I just wanted to make sure that this is apparent.

philippjfr:
Thanks @canavandl, all good points.

>How could we get this approach to support those features? Right now, it appears that an edge would still be drawn to a masked node.

Good point, this implementation technically still requires ``start`` and ``end`` points for some of the interactive features, so I could extend it to drop edge paths whose start or end points are not in the layout provider. Animating would require updating both the data source and the layout provider however. So we don't necessarily have to move them onto the layout provider but I'll defer to your judgement on this. 

>I'm not sure it's possible to allow users to provide line coordinates without also allowing them to draw "stupid" edges, I just wanted to make sure that this is apparent.

That's true, we could check that the start and end points match but that might suffer from numerical precision issues and I've already run into cases where an algorithm shifts the path slightly (e.g. the edge bundling we implemented in datashader). My suggestion would be to just to make sure to document this appropriately.

philippjfr:
@bryevdv @canavandl Ready for review. Should go green this time.

bryevdv:
@philippjfr I think we need to consider @canavandl's question:

> Perhaps being part of the solution to above, could/should edge coordinates belong to the layout provider model?

In particular

> The benefit of keeping them on the data source is that it will benefit from base64 encoding and in future the binary protocol.

If you make a `data = ColumnData` on the layout provider then I think as of a few days ago the serialization will come along for free. I'm inclined to think putting the data on the LP is preferable unless we can prove a specific drawback (i.e. if I am wrong about just simply adding a `ColumnData`

philippjfr:
I'm happy to give it a try.

philippjfr:
The nice thing about keeping it on the data source was that it automatically validates whether the number of paths matches the number of supplied edges. Not sure how to do that validation if I move it onto the LayoutProvider.

canavandl:
> Perhaps being part of the solution to above, could/should edge coordinates belong to the layout provider model?

Since edges don't have explicit indices in the way that nodes do, I think @philippjfr is right that the benefit of having validation of static edge coords via the normal edge_renderer.data_source validation is the best route forward. I struggling to think of how we would reliably keep the edge coords on the layout provider model. Since the layout provider is managing the passing of these coords, I think it fulfills our stated goals for the StaticLayoutProvider model. 

There's going to be some weirdness around updating both node and edge coordinates in a callback, which will send two different change messages and cause a stutter, but I believe this will be resolved when we add batched, document-level update eventing.

PR #6868, opened 2017-09-02
Align attribution to frame's lower right corner

### LINKED ISSUE #4876: Tile attribution misaligned (see toolbar on right-side example)
ping @birdsarah

mattpap:
fixes #4876

### DIFF
--- a/bokehjs/src/coffee/models/tiles/tile_renderer.coffee
+++ b/bokehjs/src/coffee/models/tiles/tile_renderer.coffee
@@ -35,17 +35,17 @@ export class TileRendererView extends RendererView
 
     if isString(attribution) and attribution.length > 0
       if not @attributionEl?
-        border_width = @map_plot.outline_line_width
-        bottom_offset = @map_plot.min_border_bottom + border_width
-        right_offset = @map_frame._right.value - @map_frame._width.value
-        max_width = @map_frame._width.value - border_width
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/models/tiles/tile_renderer.coffee
```diff
@@ -35,17 +35,17 @@ export class TileRendererView extends RendererView
...
           class: 'bk-tile-attribution'
           style: {
-            position: 'absolute'
-            bottom: "#{bottom_offset}px"
-            right: "#{right_offset}px"
+            position: "absolute"
+            bottom: "#{bottom}px"
+            right: "#{right}px"
             'max-width': "#{max_width}px"
+            padding: "2px"
```
bryevdv: wouldn't `padding: "#{@map_plot.outline_line_width}px` accommodate plot outline styles more flexibly?

mattpap: That's a different thing. Previously it was used to position the attribution, but as nothing else respect outline in bokejs, then for consistency sake I removed it. At some point I will need to figure out a generic scheme for handling frame outline (or in fact border).

bryevdv: OK

PR #6872, opened 2017-09-02
Switch from `pt` font size units to `px`

mattpap:
This changes all `pt` font size units to `px`. I rounded inexact `px` numbers to whole numbers. Integration tests' images need updating, thus WIP.


mattpap:
I'm not exactly happy how this PR turned out. I will get back to this at some point.

Issue #6876, opened 2017-09-02
New feature: add a "ResetTool clicked" event 


nleclercq:
Add a `Reset`event so that one can attach a js or python callback to the `ResetTool`.
Adopt an implementation a la `SelectionGeometry`.

mattpap:
@bryevdv, I presume this can be closed?

PR #6877, opened 2017-09-02
added new Reset event

nleclercq:
See bokeh issue #6876

### DIFF
--- a/bokeh/events.py
+++ b/bokeh/events.py
@@ -195,6 +195,15 @@ class SelectionGeometry(PlotEvent):
         self.final = final
         super(SelectionGeometry, self).__init__(model=model)
 
+class Reset(PlotEvent):
+    ''' Announce a button click event on a plot ResetTool.
+
+    '''
+    event_name = "reset"
+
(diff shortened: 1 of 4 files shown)


bryevdv:
FYI @nleclercq I pushed a small change to remove some trailing whitespace (linter check was failing)

bryevdv:
@nleclercq The problem is that the trigger line needs to be this:
```
@plot_model.plot.trigger_event(new Reset())
```
the `plot_model` is actually a `PlotCanvas` not a `Plot` due to uninteresting historical reasons. If you pull (to get my whitespace change) then make that change and push, we can get this merged.

nleclercq:
Thanks Bryan. The `Reset` event works now. Anything else to do for this new feature? You wrote : 

> It would probably be useful to report the updated plot ranges in the Reset event as well

Hum... I'm not sure to understand what it means exactly cause I don't see which one of the existing events I could fire to announce the ranges update.

bryevdv:
Thanks @nleclercq !

PR #6878, opened 2017-09-02
Bryanv/document cleanup

bryevdv:
This work started as an effort to add binary array transport. I got a good part of the way and then realized there was a little BokehJS work (probably) to accept buffers, and also that I had been generally cleaning up code related to document along the way. I decided to finish the latter task, leaving a clean foundation for the remaining array transport work. 

This PR:

* adds a custom descriptor and property container for `ColumnData` so that all accesses to `source.data` can be easily specialized
* adds a `ColumnDataChanged` document event (hint) and uses it to implement the existing base64 array encoding 
* removes some duplicate/cruft/unused functions
* splits the enormous `document.py` in a `document` package with:
  - `document.py` much slimmed down (almost under 900 lines!)
  - `events.py` for document events like `ModelChanged` (**finally** they are in the right place)
  - `locking.py` for funcs and classes around doc locking
  - `util.py` alot of things that don't need to clutter `document.py`
* make properties responsible for doing their own container wrapping if necessary instead of unpleasant global centralizing in property base
* make doc events themselves encapsulate the code to generate their JSON message (as well as updating refs or buffers)

I will add docs and alot of tests tomorrow before marking `ready` but do not expect any additional code changes. 

@mattpap, @canavandl  the commits are more or less compartmentalized and atomic (I tried to run tests in between every commit) if you'd like to examine them individually. 

@havocp I don't expect you to have any time to look at this in depth but if you have time for a quick scan, and anything happens to jump out at you, comments certainly appreciated. 

---
After this, the path or binary array transport is pretty straightforward:
* use the isolated code path for `ColumnData` updates to generate buffers in the protocol message in one place
* use those buffers on the BokehJS side (need to double check BokehJS message receiving buffers correctly)

As a bonus we could also use the new descriptor and `ColumnDataChanged` event to send only the parts of CDS that change. e.g 
```
source.data['foo'] = array(...)  # only send foo column, not all of CDS
```


bryevdv:
@bokeh/dev FYI plan to merge this in a few hours.

PR #6879, opened 2017-09-03
Add support for multi line text

### LINKED ISSUE #994: Text doesn't respect new lines
The code below does not render the `\n` as a new line. Replacing it with `<br/>` doesn't help either.

```
import bokeh.plotting as plt

plt.output_file("test.html")
plt.text(x=[1,2,3], y = [0,0,0], text=['hello\nworld!', 'hello\nworld!', 'hello\nworld!'], angle = 0)
plt.show()
```

mattpap:
fixes #994

### DIFF
--- a/bokeh/core/property_mixins.py
+++ b/bokeh/core/property_mixins.py
@@ -43,7 +43,7 @@ from __future__ import absolute_import
 
 from .enums import LineJoin, LineCap, FontStyle, TextAlign, TextBaseline
 from .has_props import HasProps
-from .properties import ColorSpec, DashPattern, Enum, FontSizeSpec, Int, NumberSpec, String, value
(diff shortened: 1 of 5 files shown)


mattpap:
image

bryevdv:
I would suggest ditching the grid lines, and adding labels under each group that state what the settings or it are. Also tbh I don't understand what's going on in some of them e.g. this one where the rows seem to overlap vertically

screen shot 2017-09-03 at 20 13 22

mattpap:
@bryevdv, I did some improvements to the example:
image

bryevdv:
@mattpap I still found it confusing. I also wanted to take advantage of new categoricals, so I updated to do that, and also have a slider for angle since it can apply to all combos.

screenflow

mattpap:
This is very nice and something I had in mind this example to become.

PR #6883, opened 2017-09-04
Bryanv/6852 bar hittests

### LINKED ISSUE #6852: HoverTool not rendering on HBar glyphs in 0.12.7

python 3.5.2 (also tested in 3.6.2 with same result)
bokeh 0.12.7
OSX 10.11.4
Chrome 60.0.3112.113

In a figure with `HBar` glyphs and a `HoverTool` specified, the Hover tooltips are not rendering in bokeh 0.12.7.  Confirmed the same code running in 0.12.6 renders the `HoverTool` as expected.  Prior code displaying `HoverTool`s on `VBar` glyphs still works under 0.12.7.

```python
from bokeh.io import curdoc
from bokeh.models import ColumnDataSource, Range1d, HoverTool

source = ColumnDataSource(data={'x1': ['a','b'], 'x2': [1,2]})
p = figure(tools="hover",y_range=["a","b"])
p.x_range = Range1d(0,3)
p.hbar(y="x1", right="x2", source=source, height=0.9)
p.select(HoverTool).tooltips = [("x1","@x1"), ("x2","@x2")]

curdoc().add_root(p)
```

no errors generated.

bryevdv:
issues: fixes #6852

This PR fixes the point hover problem with `HBar`. This issue is that the coordinate specification was given in reverse. This is fixed in the first commit. The PR goes on to refactor bars and quads to share most implementation in a non-public `Box` base. Finally, in the third commit it adds a simple `_hit_span` so that all bars and quads now respond to `hline` and `vline` hovers.

An example is shown below (cursor was to the right of the bar, with mode `hline`)

screen shot 2017-09-04 at 11 12 00

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/box.coffee
+++ b/bokehjs/src/coffee/models/glyphs/box.coffee
@@ -44,6 +44,27 @@ export class BoxView extends GlyphView
     result['1d'].indices = hits
     return result
 
+  _hit_span: (geometry) ->
+
+    [vx, vy] = [geometry.vx, geometry.vy]
+
+    if geometry.direction == 'v'
+      y = @renderer.yscale.invert(vy)
+      hr = @renderer.plot_view.frame.h_range
+      minX = @renderer.xscale.invert(hr.min)
+      maxX = @renderer.xscale.invert(hr.max)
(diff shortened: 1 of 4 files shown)


bryevdv:
I'm gonna go ahead and merge this.

PR #6886, opened 2017-09-04
fix table DateFormatter using tz

### LINKED ISSUE #6859: DateFormatter not working in Tables
Hi,

I tried using the DateFormatter but it does not work anymore in the last version (0.12.7).

I think it's a regression as this works in 0.12.6.

Minimal example to test with: http://bokeh.pydata.org/en/latest/docs/user_guide/examples/interaction_data_table.html

You can see in the example that the dates are not correctly formatted.

Switching back to 0.12.6 and running the exact same example, the dates are correctly displayed.

bryevdv:
- [x] issues: fixes #6859
- [x] tests added / passed

This PR fixes `DateFormatter` but in doing so updates it to use `strftime` formats instead of the JQ style formats. The previous special format names are still accepted, except `TICKS` which has been removed. 

Original example now shows:

screen shot 2017-09-04 at 17 02 17

### DIFF
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@ -202,7 +202,7 @@ MapType = enumeration("satellite", "roadmap", "terrain", "hybrid")
 
 #: Specify a format for printing dates
 DateFormat = enumeration("ATOM", "W3C", "RFC-3339", "ISO-8601", "COOKIE", "RFC-822",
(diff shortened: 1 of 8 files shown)


bryevdv:
This should be fairly uncontroversial I will merge when green.

PR #6889, opened 2017-09-05
remove cruft from top level bokeh module

bryevdv:
Just cleaning out some junk.


bryevdv:
Should be uncontroversial

PR #6890, opened 2017-09-05
add trace log level

### LINKED ISSUE #6888: Add trace log level on Python side
Add a custom `trace` log level to match BokehJS, and also to avoid having commented `log.debug` statements.

bryevdv:
issues: closes #6888 

This adds a `trace` log level and uncomments any commented `debug` log calls, replacing them with uncommented `trace` instead.

@bokeh/dev any comments on whether there are any downsides to monkey patching `logging` ?

### DIFF
--- a/bokeh/core/has_props.py
+++ b/bokeh/core/has_props.py
@@ -11,7 +11,7 @@ serializable properties.
 from __future__ import absolute_import
 
 import logging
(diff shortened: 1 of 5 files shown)


bryevdv:
Well, AFAIK it should be OK, we can always revert or change if need be

PR #6898, opened 2017-09-07
[doc] Adjust docstring indentation

chbrandt:
@bryevdv If this is not a reasonable request, never mind, just discard...

When submitting this pull request it says:

> All pull requests must have an associated issue in the issue tracker. If there
> isn't one, please go open an issue describing the defect, deficiency or desired
> feature. You can read more about our issue and PR processes in the
> [wiki](https://github.com/bokeh/bokeh/wiki/BEP-1:-Issues-and-PRs-management).
> 
> - [ ] issues: fixes #xxxx
> - [ ] tests added / passed
> - [ ] release document entry (if new feature or API change)

Great.

I wonder whether for such small corrections like I'm doing now there could be a simpler approach (create an issue). On my side I am almost ashamed of submiting such small correction to you, but since I think documentation is really important (and painful to maintain) I think it is still valid. And since it is that small, I just wanted it to be as fast as possible.

Best,
Carlos

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -155,7 +155,7 @@ def widgetbox(*args, **kwargs):
 
     Args:
         children (list of :class:`~bokeh.models.widgets.widget.Widget` ): A list
-        of widgets for the WidgetBox.
+            of widgets for the WidgetBox.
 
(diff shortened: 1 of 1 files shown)


bryevdv:
@chbrandt all contributions are welcome and appreciated, thanks!

PR #6904, opened 2017-09-07
Cleanup BokehJS client code and add tests

bryevdv:
This is another supporting PR to lead up to binary array support. The buffer handling on the BokehJS had never been exercised, appeared to be wrong, and had no tests. This PR:

* makes a clean port of `Message` and `Receiver` from python and adds full tests
* splits up client session and client connection and has the use the new `Message` and `Receiver` classes
* adds some missing tests on the python side

None of the existing BokehJS tests for client connections were changed (they did move location, though)

I am still not completely happy with the split of code between BokehJS client connections and sessions. I'd like for the python and JS classes to be essentially as near to straight ports of each other as they can be (or at least each responsible for the same things). But this is good enough for now.


bokeh/protocol/message.py
```diff
@@ -309,10 +309,8 @@ def metadata_json(self):
...
 
+    # buffer properties
+
     @property
     def buffers(self):
         return self._buffers
-
-    @buffers.setter
-    def buffers(self, value):
-        self._buffers = list(value)
```
bryevdv: unused, dangerous, removed

bokeh/protocol/message.py
```diff
@@ -161,7 +161,7 @@ def assemble_buffer(self, buf_header, buf_payload):
         Raises:
             ProtocolError
         '''
-        if self.header['num_buffers'] <= len(self._buffers):
+        if self.header.get('num_buffers', 0) <= len(self._buffers):
```
bryevdv: `num_buffers` may be omitted if there are no buffers

PR #6906, opened 2017-09-08
Binary array transport

### LINKED ISSUE #5984: Improve data transfer, using a binary transfer protocol

bryevdv:
- [x] issues: fixes #5984
- [x] tests added / passed

Comments:

This PR adds support for transporting arrays as pure binary data from the server to clients, when the array type is compatible with JavaScript typed array type. The following sorts of updates to a column data source will trigger binary transport:

```
# replacing an entire data dict
source.data = dict(x=x, y=y)

# setting a single column
source.data[x] = x

# updating multiple columns
source.update(x=x, y=y)
```
Additionally, in the latter two cases, all updates now only send the columns there were actually set, instead of always sending the entire data dict (whether binary transport is available or not).

Since the many various functions involved in serialization were geared towards *returning* JSON or JSON strings, the implementation approach take was to add an optional **out-paramter** to collect binary buffers in to the functions, and to propagate it through the call chain when present. When not supplied, the default behaviour remained identical to the previous behaviour, making this approach reasonably non-intrusive. 

On the JS side, few changes were needed apart from providing an option to circumvent old/new value equality testing on ` ColumnDataChangedEvent` since it is a given that the update should take place. This results in a considerable performance improvement. 

Some limitations:

* base64 encoding still used for initial page load, hopefully can fix in a follow-on effort
* binary transport TO the server is not supported (including from python clients)
* ~~~no detection or correction for byte order is undertaken. Binary data is sent out as-is~~~

Finally, code under `bokeh.client` has been split up and refactored with additional tests added.

### DIFF
--- a/bokeh/client/__init__.py
+++ b/bokeh/client/__init__.py
@@ -1,4 +1,15 @@
-''' Provides client API for connecting to a Bokeh server
+''' Provides client API for connecting to a Bokeh server from a Python
+process.
+
+The primary uses for the ``bokeh.client`` are:
+
+* Implementing testing infrastructure around Bokeh applications
+* Creating and customizing specific sessions of a Bokeh application
+  running *in a Bokeh Server*, before passing them to a viewer.
+
+While it is also possible to run Bokeh application code "outside" a
+Bokeh server using ``bokeh.client``, this practice is **HIGHLY DISCOURCED**.
+
 
 '''
 from __future__ import absolute_import
--- a/bokeh/client/_connection.py
+++ b/bokeh/client/_connection.py
@@ -192,7 +192,15 @@ class ClientConnection(object):
         raise gen.Return(None)
 
     def _send_patch_document(self, session_id, event):
-        msg = self._protocol.create('PATCH-DOC', [event])
+        # XXX This will cause the client to always send all columns when a CDS
+        # is mutated in place. Additionally we set use_buffers=False below as
+        # well, to suppress using the binary array transport. Real Bokeh server
+        # apps running inside a server can handle these updates much more
+        # efficiently
+        from bokeh.document.events import ColumnDataChangedEvent
+        if hasattr(event, 'hint') and isinstance(event.hint, ColumnDataChangedEvent):
+            event.hint.cols = None
+        msg = self._protocol.create('PATCH-DOC', [event], use_buffers=False)
         self.send_message(msg)
 
     def _send_message_wait_for_reply(self, message):
--- a/bokeh/client/connection.py
+++ b/bokeh/client/connection.py
@@ -5,116 +5,32 @@ Server. Users will always want to use ``bokeh.client.session`` instead.
 from __future__ import absolute_import, print_function
 
 import logging
-
 log = logging.getLogger(__name__)
 
-from tornado import gen, locks
+from tornado import gen
 from tornado.httpclient import HTTPRequest
 from tornado.ioloop import IOLoop
 from tornado.websocket import websocket_connect, WebSocketError
-from tornado.concurrent import Future
 
 from bokeh.protocol import Protocol
 from bokeh.protocol.exceptions import MessageError, ProtocolError, ValidationError
 from bokeh.protocol.receiver import Receiver
 
-class _WebSocketClientConnectionWrapper(object):
-    ''' Used for compat across Tornado versions and to add write_lock'''
-
-    def __init__(self, socket):
-        if socket is None:
-            raise ValueError("socket must not be None")
-        self._socket = socket
-        # write_lock allows us to lock the connection to send multiple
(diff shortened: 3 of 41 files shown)


bryevdv:
Here is a specific example that tries to minimize compute for the image
```

import numpy as np

from bokeh.io import curdoc
from bokeh.models import ColumnDataSource, Slider

N = 512
for i in range(N):
    for j in range(N):
        view[i, j, 0] = int(i/N*255)
        view[i, j, 1] = 0
        view[i, j, 2] = int(j/N*255)
        view[i, j, 3] = 255

def compute(G):
    view[:, :, 1] = G
    return img

source = ColumnDataSource(data=dict(img=[compute(0)]))

def update(attr, old, new):
    source.data=dict(img=[compute(new)])

p.image_rgba(image='img', x=0, y=0, dw=10, dh=10, source=source)

slider = Slider(start=0, end=255, value=0, step=10)
slider.on_change('value', update)

curdoc().add_root(column(p, slider))
```

philippjfr:
Here are my profiling results using a periodic callback:

bokeh_plot

bryevdv:
@philippjfr are you using the code above? Have you subtracted out any of the time to actually compute the new array?

philippjfr:
>@philippjfr are you using the code above?

A slight modification that uses a periodic callback instead.

>Have you subtracted out any of the time to actually compute the new array?

As far as I can tell that time is negligible (<1ms).

bryevdv:
Can you run a range of sizes from N=100 to 1000?

bryevdv:
@philippjfr also check with the change I just pushed, it seems to make a pretty huge difference, I can scrub a 1000x1000 image with very little lag.

philippjfr:
Here's what I'm seeing, maybe I need to find a better way to profile though:

bokeh_plot

bokeh_plot

philippjfr:
Separately, if I push too much data, too quickly across the socket (1000x1000 image with a low periodic timeout) I frequently see this:

    2017-09-08 15:09:04,094 Write error on <socket.socket fd=13, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 5006, 0, 0), raddr=('::1', 58854, 0, 0)>: [Errno 22] Invalid argument

bryevdv:
> Here's what I'm seeing, maybe I need to find a better way to profile though:

Yah that does not seem to agree with my subjective experience at least. 1000x1000 is almost smooth when scrubbing a slider with the new code, and basically unusable without.

As for date rate limit, yes there seems to be one, I doubt we can do much about but it would be nice to know what it is.

bryevdv:
@philippjfr Here is slider interactive with 1000x1000

screenflow


Without the change it basically does not work or function at all. What are you actually measuring? Can you share your benchmark code?

bryevdv:
@bokeh/dev I am going to try to add more tests but this is otherwise ready.

bryevdv:
OK, limited capability is one thing, I wouldn't be able to live with myself leaving that endianness oversight. New code works on my laptop, should work on big endian and across one to the other, but will need to track down a big endian resource to test for sure (will add any unit tests I am able to tomorrow)

bryevdv:
OK, I am going to merge this when green and make dev build

PR #6911, opened 2017-09-08
Add convenience function for datetime values

### LINKED ISSUE #6895: Bokeh DateRangeSlider returns tuple of integers instead of dates
In 0.12.7 the DateRangeSlider `.value` returns a tuple of integers. Additionally, in order to convert them back to dates you have to divide by 1000 (i.e. `datetime.fromtimestamp(the_tuple[0]/1000)`).

Expected behavior: Return a tuple of Date objects.

```
from datetime import datetime
from bokeh.models.widgets import DateRangeSlider
from bokeh.io import curdoc

def date_range_update(attrname, old, new):
    print('-- range values:', date_slider.value)
    # Works
    d1 = datetime.fromtimestamp(date_slider.value[0] / 1000) 
    # Does not Work, gives error
    d2 = datetime.fromtimestamp(date_slider.value[0])

date_slider = DateRangeSlider(value=(date_start,date_end), start=date_start, end=date_end)
date_slider.on_change('value', date_range_update)

curdoc().add_root(date_slider)

```

Maggie-M:
issues: fixes #6895

### DIFF
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -126,15 +126,19 @@ class DateRangeSlider(AbstractSlider):
 
     @property
     def value_as_datetime(self):
+        ''' Convenience property to retrieve the value tuple as a tuple of
+        datetime objects.
+
+        '''
         if self.value is None:
             return None
         v1, v2 = self.value
         if isinstance(v1, numbers.Number):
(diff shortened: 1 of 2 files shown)


bryevdv:
@hsparra can you test this out?

Issue #6908, opened 2017-09-08
Running output_notebook with hide_banner=True broken


philippjfr:
Using latest master ``output_notebook`` is broken when using the ``hide_banner`` option, e.g. in the example below the plot does not display unless I set ``hide_banner=False``.

```python
from bokeh.io import output_notebook
from bokeh.plotting import figure, output_file, show

output_notebook(hide_banner=True)

p = figure(plot_width=400, plot_height=400)

p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], size=20, color="navy", alpha=0.5)

show(p)
```

The JS console reveals the following errors:

```
Bokeh: ERROR: autoload.js configured with elementid 'bdb33685-6dc0-4eb1-8acb-60291c9a5833' but no matching script tag was found. 
(unknown) Bokeh: ERROR: Unable to embed document because BokehJS library is missing
```

bryevdv:
This should hopefully be a fairly simple fix. We are abusing the autoload template to show the banner div (it's usually for the display div). The real solution is to not re-use the autoload template or at least not re-use `elementid`, but there's probably something simpler in the immediate term.

PR #6913, opened 2017-09-08
Canavandl/6909 notebook fixes

### LINKED ISSUE #6909: Notebook output generates console errors
Since introducing JupyterLab compatibility in bokeh master displaying bokeh plots in the classic notebook generates various JS errors in certain conditions (which I haven't yet narrowed down). Here's an example that reliable reproduces the issue:

```python
import math

from bokeh.io import show
from bokeh.layouts import row

output_notebook(hide_banner=False)


def create_graph():
    N = 8
    node_indices = list(range(N))

    graph = GraphRenderer()

        index=node_indices,
        fill_color=Spectral8)

        start=[0]*N,
        end=node_indices)

    ### start of layout code

    plot.renderers.append(graph)
    return plot

p1 = create_graph()
p2 = create_graph()

show(row(p1, p2))
```

This produces the following error on the console, which appears twice:

```
Exception in event handler for output_added.OutputArea TypeError: Cannot read property 'hasOwnProperty' of undefined
    at window._Events.handleAddOutput (eval at append_javascript (outputarea.js:727), <anonymous>:67:21)
    at window._Events.dispatch (jquery.min.js:5)
    at window._Events.y.handle (jquery.min.js:5)
    at Object.trigger (jquery.min.js:5)
    at window._Events.<anonymous> (jquery.min.js:5)
    at Function.each (jquery.min.js:4)
    at init.each (jquery.min.js:4)
    at init.trigger (jquery.min.js:5)
    at init.events.trigger (events.js:31)
    at OutputArea.append_output (outputarea.js:339) (2) ["output_added.OutputArea", {…}, callee: (...), Symbol(Symbol.iterator): ƒ]
```

This is presumable occurs due to an issue here in the autoload_nb_js template: https://github.com/bokeh/bokeh/blob/master/bokeh/core/_templates/autoload_nb_js.js#L59

canavandl:
Canavandl/6909 notebook fixes

- [ ] issues: fixes #6909 #6908 
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -74,10 +74,12 @@ calls it with the rendered model.
       document.getElementsByTagName("head")[0].appendChild(s);
     }
   };
+
   {%- if elementid -%}
   var element = document.getElementById({{ elementid|json }});
   if (element == null) {
     console.log("Bokeh: ERROR: autoload.js configured with elementid '{{ elementid }}' but no matching script tag was found. ")
+    return
   }
   {%- endif %}
 
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -74,12 +74,10 @@ calls it with the rendered model.
       document.getElementsByTagName("head")[0].appendChild(s);
     }
   };
-
   {%- if elementid -%}
   var element = document.getElementById({{ elementid|json }});
   if (element == null) {
     console.log("Bokeh: ERROR: autoload.js configured with elementid '{{ elementid }}' but no matching script tag was found. ")
-    return false;
   }
   {%- endif %}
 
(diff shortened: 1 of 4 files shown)


bokeh/core/_templates/autoload_js.js
```diff
@@ -74,12 +74,10 @@ calls it with the rendered model.
       document.getElementsByTagName("head")[0].appendChild(s);
     }
   };
-
   {%- if elementid -%}
   var element = document.getElementById({{ elementid|json }});
   if (element == null) {
     console.log("Bokeh: ERROR: autoload.js configured with elementid '{{ elementid }}' but no matching script tag was found. ")
-    return false;
```
canavandl: The issue is that we now optionally publish an html div for the banner, then publish the autoload script. When the banner isn't loaded, this elementid doesn't exist and the load script is returned before embedding the resources.

I need to think a little more about the impact of this change on other autoload situations.

PR #6918, opened 2017-09-11
js code error in documentation

busino:
Attributes are accessible directly from `cb_obj`.
Use `cb_obj.x` instead.

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Thanks for the edit busino

PR #6920, opened 2017-09-12
added render line to allow programatic change of tabs, fixes Issue #6502

### LINKED ISSUE #6502: Programmatically setting the active Tab does not change the tab in 0.12.6 & 0.12.7
When moving from Bokeh 0.12.5 to 0.12.6 on python 3.6, setting the active tab programmatically no longer changes the tab in the browser.  Clicking on the "Switch Tabs" button generated by the below code should cause the active tab to switch to Tab 2, but no longer does this under Bokeh 0.12.6. Maybe this is related to Issue #6501 ?

```python
from bokeh.layouts import column, widgetbox
from bokeh.models.widgets import Div, Tabs, Panel

label1 = Div(text='Tab 1')
label2 = Div(text='Tab 2')
panel1 = Panel(child=widgetbox(label1), title='Tab 1')
panel2 = Panel(child=widgetbox(label2), title='Tab 2')
tabs = Tabs(tabs=[panel1, panel2])


def callback():
    print('before', tabs.active)
    tabs.active = 1
    print('after', tabs.active)


button = Button(label="Switch Tabs")
button.on_click(callback)

curdoc().add_root(column(widgetbox(button), tabs))
```

Zyell:

- [x] issues: fixes #6502
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/tabs.coffee
+++ b/bokehjs/src/coffee/models/widgets/tabs.coffee
@@ -9,6 +9,7 @@ export class TabsView extends WidgetView
   connect_signals: () ->
     super()
     @connect(@model.properties.tabs.change, () => @rebuild_child_views())
+    @connect(@model.properties.active.change, () => @render())
 
   render: () ->
     super()


Zyell:
Fixes issue #6502

bryevdv:
@Zyell Just to set expectiations, I'd like @mattpap to look at this, given his comment about wanting to understand why infinite event ping-pong was not happening. He will be gone for another ~10 days but we will get this evaluated when he returns.

bryevdv:
@mattpap woudl you oppose merging this for `0.12.10` ?

bryevdv:
ping @mattpap

mattpap:
I have reservations about this change, but if it works then lets proceed. I will revisit tabs later on anyway, because it needs a proper implementation as a first-class layout.

bryevdv:
Thanks @Zyell !

PR #6928, opened 2017-09-12
Bryanv/cleanups

bryevdv:
Remove several deprecations and dead code, and add one for `responsive`. Work is outlined in individual commits. 

The only thing possibly of note:

> Glyphs methods now only accept either all sequence literals, OR all column names

i.e. mixing will trigger an error

### DIFF
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -28,7 +28,6 @@ from six.moves.urllib.parse import quote_plus
 from ..document import Document
 from ..resources import _SessionCoordinates, DEFAULT_SERVER_HTTP_URL
 from ..util.browser import NEW_PARAM
-from ..util.deprecation import deprecated
 from ..util.session_id import generate_session_id
 from ..util.string import format_docstring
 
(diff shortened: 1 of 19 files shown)


bryevdv:
cc @zjffdu This PR removes all *built-in* remains of zeppelin support, so you will want to make sure your notebook hook is providing all the necessary tasks to load and show

bryevdv:
@canavandl please review if you have time but no worries if you do not.

bryevdv:
I don't think there should be anything controversial here, I am going to go ahead and merge.

zjffdu:
@bryevdv I just found that `bokeh.util.notebook._load_notebook_html` is deleted in some PR, could we expose it as public function so that zeppelin hook (bkzep) can reuse the this function to load bokeh js ?

bryevdv:
@zjffdu Bokeh itself has no use for that function anymore so I'd suggest a better option is to copy it the old version into your project. Bokeh is a very large library, as we approach 1.0 I am trying to aggressively pare down any cruft, duplicate, or unused code to make ongoing maintenance more approachable.

zjffdu:
Make sense, will do that.

zjffdu:
I updated `bkzep` and bokeh still works in zeppelin. Thanks @bryevdv

bryevdv:
Brilliant, I am glad to hear it! I think this separation of concerns will keep things more easily maintainable for everyone

PR #6930, opened 2017-09-13
Issue #6704: Fix hover hittest on horizontal/vertical segments

### LINKED ISSUE #6704: HoverTool hit detection fails on vertical and near-vertical segment glyphs

*ISSUES THAT DO NOT CONTAIN NECESSARY INFORMATION MAY BE CLOSED, IMMEDIATELY*

The issue tracker is NOT the place for general support. For questions and
technical assistance, come ask the [Bokeh mailing list](https://groups.google.com/a/continuum.io/forum/#!forum/bokeh) or join the chat on [Gitter](https://gitter.im/bokeh/bokeh). For feature requests, please provide a detailed description or proposal of the new capability or behavior.

bokeh 0.12.6

Hovertool on segments should work for all segments, but hit detection appears to be failing on vertical and near-vertical segments. Not sure what the max slope is, but as slope nears infinity, hit detection fails.

```

hover = HoverTool(tooltips=[
    ("index", "$index"),
    ("(x,y)", "($x, $y)")
])

p = figure(title="Background Analysis", 
           toolbar_location="above",
           tools=[hover],
           plot_width=400, 
           plot_height=400)

p.segment(
    x0=[1, 2, 3], 
    x1=[1, 2.4, 3.1],
    y0=[1, 2, 3], 
    y1=[2, 2.5, 3.7], 
)

show(p)
```
None

image

pankajp:
Horizontal/vertical segments have zero bbox area and hence
are almost impossible to coincide with a point for hittest search.
Hence RBush.indices() was never returning such segments.
We now instead search rbush with a square of size 4x4 instead of a
point to find such segments as well.
Ideally, the square size should be maximum of the the line_width,
but the current state is in itself a great improvement.

- [x] issues: fixes #6704 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/segment.coffee
+++ b/bokehjs/src/coffee/models/glyphs/segment.coffee
@@ -42,9 +42,13 @@ export class SegmentView extends GlyphView
 
     hits = []
     lw_voffset = 2 # FIXME: Use maximum of segments line_width/2 instead of magic constant 2
-    lw_sxoffset = @renderer.xscale.invert(vx-lw_voffset, true)
-    lw_syoffset = @renderer.yscale.invert(vy-lw_voffset, true)
-    candidates = @index.indices({minX: lw_sxoffset, minY: lw_syoffset, maxX: 2*x - lw_sxoffset, maxY: 2*y - lw_syoffset})
+    candidates = @index.indices({
+      minX: @renderer.xscale.invert(vx-lw_voffset, true), 
+      minY: @renderer.yscale.invert(vy-lw_voffset, true),
+      maxX: @renderer.xscale.invert(vx+lw_voffset, true),
+      maxY: @renderer.yscale.invert(vy+lw_voffset, true)
+    })
+
     for i in candidates
       threshold2 = Math.pow(Math.max(2, @visuals.line.cache_select('line_width', i) / 2), 2)
       [p0, p1] = [{x: @sx0[i], y: @sy0[i]}, {x: @sx1[i], y: @sy1[i]}]
--- a/bokehjs/src/coffee/models/glyphs/segment.coffee
+++ b/bokehjs/src/coffee/models/glyphs/segment.coffee
@@ -41,13 +41,15 @@ export class SegmentView extends GlyphView
       y: this.renderer.plot_view.canvas.vy_to_sy(vy)
 
     hits = []
-
-    candidates = @index.indices({minX: x, minY: y, maxX: x, maxY: y})
+    lw_voffset = 2 # FIXME: Use maximum of segments line_width/2 instead of magic constant 2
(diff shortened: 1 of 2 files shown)


bokehjs/src/coffee/models/glyphs/segment.coffee
```diff
@@ -41,13 +41,15 @@ export class SegmentView extends GlyphView
       y: this.renderer.plot_view.canvas.vy_to_sy(vy)
 
     hits = []
-
-    candidates = @index.indices({minX: x, minY: y, maxX: x, maxY: y})
+    lw_voffset = 2 # FIXME: Use maximum of segments line_width/2 instead of magic constant 2
+    lw_sxoffset = @renderer.xscale.invert(vx-lw_voffset, true)
+    lw_syoffset = @renderer.yscale.invert(vy-lw_voffset, true)
+    candidates = @index.indices({minX: lw_sxoffset, minY: lw_syoffset, maxX: 2*x - lw_sxoffset, maxY: 2*y - lw_syoffset})
```
bryevdv: Can you explain this a bit more? I am confused why you'd multiply the *coordinate* by 2? If the intent is to check a box around the point `x,y` then I would expect maybe:
```
candidates = @index.indices({
    minX: x - lw_sxoffset, 
    minY: y - lw_syoffset, 
    maxX: x + lw_sxoffset, 
    maxY: x + lw_syoffset})
```

bryevdv: Oh I see `lw_sxoffset` already has the displacement baked in. I still think the right way to do things is then to transform all the values independently:
```
candidates = @index.indices({
    minX: @renderer.xscale.invert(vx-lw_voffset, true), 
    minY: @renderer.xscale.invert(vx-lw_voffset, true),
    maxX: @renderer.xscale.invert(vx+lw_voffset, true),
    maxY: @renderer.xscale.invert(vx+lw_voffset, true)
```

bryevdv:
@pankajp I just have one comment to discuss above, thanks for the PR!

pankajp:
@bryevdv Thanks for the comment, indeed it was a misguided optimization which was incorrect in non-linear scales. I have updated the code.

bryevdv:
Looks good, I'd like to merge when green. I am traveling and don't have the ability to check out the PR and test myself, can you attach a screenshot of a hover tool over a vertical or horizontal segment to this PR as documentation?

bryevdv:
@pankajp there was a code linter violation (trailing whitespace) for the sake of expediency I went ahead and edited the PR branch from the GitHub UI, hope that's ok.

pankajp:
@bryevdv Thanks a lot for the fix and merge. I actually just finished building and testing this change, so here's the screenshot. And this is the first time I realized neither windows print-screen nor the sniping tool captures the mouse cursor, and windows has another nice tool called "Steps Recorder" which does that :)

image

PR #6936, opened 2017-09-15
Refactor bokeh.io as "model" package with API version declarations

bryevdv:
- [ ] issues: fixes #xxxx
- [x] tests added / passed

The `bokeh.io` module had gotten very large and cumbersome (~1 kLOC). This work started as merely breaking up the module in a package and submodules, but while doing do, I decided to pursue some ideas about how an "ideal" Bokeh package should be structured. 

First, some clarifying definitions about the types of API objects in the project:

* **public** commonly used by all users, e.g, `bokeh.plotting.figure` 
* **internal** used to implement public API, or used by developers or advanced users, e.g. `bokeh.io.notebook.destroy_server`
* **private** The objects have names with leading underscores, e.g. `bokeh.io.saving._get_save_args`

  *A distinguishing feature of private objects is that they should never be imported or used outside the module in which they are defined*

Any **public** or **internal** object is "visible" (i.e. no leading underscore) and will be rigorously maintained between non-major releases. Any **private** object is subject to change at any (they should not be imported or relied on by anything outside the immediate module they reside in)

To facilitate these notions, this PR adds the `bokeh.util.api` module with `@public` and `@internal` decorators. There are used to decorate any visible functions, classes, methods, or properties. (i.e. non-private, no leading underscore) with the version they were first introduced in, e.g.
```
@public(1,0,0)
def a_public_function(): pass

@internal(1,0,0)
def an_internal_function(): pass
```

This PR adds a new test mark `api` that can be used to verify a module's API against a specifcation:
```
api = {
    PUBLIC: (
        ( 'curdoc', (1, 0, 0) ),
    ), INTERNAL: (
        ( 'set_curdoc', (1, 0, 0) ),
    )
}

test_public_api, test_internal_api, test_all_declared, test_all_tested = verify_api(bid, api)
```
Tests will fail if:

* any visible functions, classes, methods, or properties of the module are not declared `public`  or `internal`
* and visible that are declared, are not accounted for int the specifiction
* the declaration in the module does not match the specification in the test

Additionally all tests were converted to `pytest` style. 

Later work will add Sphinx extension to inspect these annotations to automatically add "first introduced in version..." information to the reference guide, as well as to off the option to see only smaller public API when viewing the reference. 

---

With regards to the actual refactoring, `bokeh.io` was split into the following pieces:
```
doc.py
export.py
notebook.py
output.py
saving.py
showing.py
state.py
util.py
```
Every submodule has the exact same top header: 
```
# Copyright (c) 2012 - 2017, Anaconda, Inc. All rights reserved.
#
# Powered by the Bokeh Development Team.
#
# The full license is in the file LICENSE.txt, distributed with this software.
'''

'''

# Boilerplate
#----------------------------------------------------------------------------
from __future__ import absolute_import, division, print_function, unicode_literals

import logging
logger = logging.getLogger(__name__)

from bokeh.util.api import public, internal ; public, internal

# Imports

# Standard library imports

# External imports

# Bokeh imports
```

There is a similar header for test modules. I'd like to eventually have these the start of every Python module under `bokeh`.

Modules are split into sections for public, internal, private. Objects are in alphabetical order in each section, as are all imports in their sections.

cc @bokeh/dev for any thoughts. Once there is agreement about any details that anyone might comment on, this structure can be phased in in smaller PRs, one package at a time.


bryevdv:
Also to note and to do: The tests required a bit of massaging, but all existing tests were preserved. I intend to add more test to this PR before merging. I am leaving Dev Guide docs for this structure to a later PR.

bryevdv:
@bokeh/dev just and FYI planning on merging this this afternoon

bokeh/io.py
```diff
@@ -1,946 +0,0 @@
...
-def reset_output(state=None):
-    ''' Clear the default state of all output modes.
-
-    Returns:
-        None
-
-    '''
-    _state.reset()
-
-def _remove_roots(subplots):
```
canavandl: i'm pretty sure this is dead code. has it been removed (I didn't see it grepping around)?

bryevdv: Yup, removed it

p-himik:
@bryevdv I'm working on #7679 and the main test ordering issue comes from `bokeh.util.testing.verify_api`. I agree that the tests for public and internal members being declared are needed, but I don't see the point of having some specification and testing for it. In my opinion, such tests are not different from

```python
def my_fn():
    pass

def test_my_fn_name():
	assert my_fn.__name__ == 'my_fn'
```

The only two possibilities when such tests break are:
- A developer forgot to add/change the specification in the test files
- A developer has accidentally changed how some member is decorated with `general`/`dev`

The first one doesn't have any value since it's related solely to tests.
The second one doesn't have any value either since all accidents should be taken care of during code review sessions - otherwise, it would make sense to just have a copy of all code and make sure that it's always in sync.

Please share your thoughts.

bryevdv:
> The second one doesn't have any value either since all accidents should be taken care of during code review sessions 

This is the crux of things. Certainly things would work this way in an ideal world, but in practice not everything is always caught in reviews, so that's the intent behind placing the version specification under test. I guess I would consider changing/dropping it if it is causing a serous problem in some other place tho.

PR #6938, opened 2017-09-15
stop computing unused and expensive bokeh.__base_version__

bryevdv:

This PR removes ``bokeh.__version__`` which was computed on every import of Bokeh and could take hundred of milliseconds. 

Ref: https://github.com/dask/distributed/issues/1399#issuecomment-329834414


mrocklin:
Thanks @bryevdv !

PR #6945, opened 2017-09-18
Use Bokeh protocol to implement push_notebook

bryevdv:
- [x] tests added / passed

@bokeh/dev I am marking this as `ready` I still plan to add some docs and tests later today but no implementation changes.

### DIFF
--- a/bokeh/embed.py
+++ b/bokeh/embed.py
@@ -72,7 +72,7 @@ def _ModelInDocument(models, apply_theme=None):
     if settings.perform_document_validation():
         doc.validate()
 
-    yield models
+    yield doc
 
     for model in models_to_dedoc:
         doc.remove_root(model, apply_theme)
@@ -80,6 +80,10 @@ def _ModelInDocument(models, apply_theme=None):
 
 @contextmanager
 def _ModelInEmptyDocument(model, apply_theme=None):
+
+    # Note: Comms handling relies on the fact that the new_doc returned
+    # has models with the same IDs as they were started with
+
     from .document import Document
     doc = _find_existing_docs([model])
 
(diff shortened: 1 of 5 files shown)


philippjfr:
@bryevdv This looks great and shouldn't be too painful for us. Very excited to try it out tomorrow.

rs2:
👍

nleclercq:
I'm sorry for asking, but what is the underlying gain? Will this PR promote `push_notebook` to the level of features provided by an embedded booked server? I used to base my code on `push_notebook` but I finally had to switch to the 'embed server approach' to obtain the required level of interaction between JS and python. Thanks.

rs2:
@nleclercq Have you seen `notebook_div` as an alternative?

bryevdv:
@nleclercq `push_notebook` only offers uni-directional updates, i.e from Python *to* BokehJS. But with this work, that use capability is on par with (as fast, and as reliable) as using a Bokeh server would be. As to why:

* some people prefer to use `ipywidgets` (or already have large existing use of them) and `push_notebook` is useful for them
* some people only need the unidirectional update (i.e. just updating a plot continuously) and this capability is probably less work / more familiar in that specific use-case

nleclercq:
@bryevdv ok, so using an embedded server is still mandatory when you want to connect a python callback to bokeh event. right?

bryevdv:
If you want to connect to events originating from BokehJS (selections, range changes, property value updates like slider values, etc) then yes that requires a Bokeh server.

bryevdv:
@canavandl I had to make a few changes to `notebook_content` and tests, if you have a few moments a quick look would be appreciated. The gist is this: it now also returns the new doc that all the models get "re-homed" into (by `ModelInEmptyDocument`) so that that "copy" doc can be kept by the comms handle.

bryevdv:
OK  I am going to go ahead and merge this, there is some more cleanup I would like to do but it's more appropriate to do when I "canonicalize" the APIs of other subpackages.

PR #6952, opened 2017-09-19
canonicalize bokeh.client

bryevdv:
This PR canonicalizes `bokeh.client` with module and test module templates and API declarations. 

Additionally it adds some additional tests, and fixes a few style updates changed in this work back to `bokeh.io`


bryevdv:
@bokeh/dev there's not really much interesting here, I plan to merge later today

PR #6963, opened 2017-09-22
Correct gridplot doctext so that ncols parameter name is rendered

### LINKED ISSUE #6962: Gridplot doctext formatting error
Current text for `ncols` parameter [formats](https://bokeh.pydata.org/en/latest/docs/reference/layouts.html#bokeh.layouts.gridplot) as   

     Int (ncols) – Specify the number of columns you would like ...

while something similar to  

     plot_width (int, optional) – The width you would like all your plots to be

is probably correct.

ahjulstad:
Doctext formatting change.

- [ ] issues: fixes #6962 

(I have not yet been able to run sphinx myself due to an error running sphinx/build)

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -301,7 +301,7 @@ def gridplot(*args, **kwargs):
             toolbar will be located, with respect to the grid. Default is
             ``above``. If set to None, no toolbar will be attached to the grid.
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the catch @ahjulstad !

PR #6965, opened 2017-09-22
#6960: Derive ResizeTool from Tool not object

### LINKED ISSUE #6960: The deprecated ResizeTool is meant to be NO-OP, but fails validation Python-side
`bokeh 0.12.7`

Repro:
```python
from bokeh.models import ResizeTool, Tool

print(issubclass(ResizeTool, Tool))  # False
Figure(tools=[ResizeTool()])  # ValueError as expected Seq(Either(String, Instance(Tool)))
```

Should be derived from `Tool` not `object`: https://github.com/bokeh/bokeh/blob/master/bokeh/models/tools.py#L273

rs2:

- [x] issues: fixes #6960 
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -270,7 +270,7 @@ class ResetTool(Action):
     """)
 
 
-class ResizeTool(object):
+class ResizeTool(Tool):
     ''' DEPRECATED in 0.12.7, attempting to use will be a NO-OP
 
     '''


bryevdv:
Deriving from `Tool` means that a corresponding empty `ReizesTool` would have to be added back to BokehJS, since we have tests to maintain 100% parity between the python and JS models. I'm definteily -1 on that, especially since `ResizeTool` will probably be removed after the next release. The intent of the loud warning is to get people to not add a `ReszieTool` any more at all, so I 'm not sure what benefit there is to making the `issubclass` check "work" for a tool that doesn't really exist.

rs2:
Oh well, nice try. It is unfortunate that the error message is not clear when `ResizeTool` is added among other tools.

bryevdv:
@rs2 I want to make sure I understand, are you getting an actual error/exception instead of just a warning/no-op? It's not clear to me if you want the `isinstance` check to work because *you* are want to be able to to that, or because the Bokeh codebase is doing so somewhere?

rs2:
@bryevdv Sure, do you think the following traceback that is produced by the snippet below is sufficient to understand which of the tools is deprecated (or what the problem is, in the first place)?

```python
from bokeh.models import HoverTool, ResetTool, ResizeTool

Figure(tools=[HoverTool(), ResizeTool(), ResetTool()])
```

This results in
```
ValueError: expected an element of either String or Seq(Either(String, Instance(Tool))), got [HoverTool(id='7e420722-4735-4b93-8f9e-e024ae02a297', ...), <bokeh.models.tools.ResizeTool object at 0x000001B825BC2048>, ResetTool(id='16a98af5-b6b6-4982-bbb0-92c4b8098031', ...)]
```
You can kind of find the odd one in the list (`bokeh.models.tools.ResizeTool object`), but a bit more clarity would certainly help...

Please feel free to re-close this ticket...

bryevdv:
@rs2 No, it's not supposed t do that, it's supposed to just print a loud warning and simply ignore the tool. I think this could be fixed in this function:

https://github.com/bokeh/bokeh/blob/master/bokeh/plotting/helpers.py#L494

with an explicit filter to remove any resize tools. I would support such a change if you are interested in making a small PR to do that. Alternatively we could simply accelerate removing the tool entirely. I would be OK with either.

bryevdv:
I think at this point I'd favor just removing the tool entirely. It's technically a short deprecation cycle but we have not demonstrated its use in a vey long time. At this point I think the priority for getting to the 1.0 stabilized API (i.e. jettisoning anything we ever want to be able to jettison) is to get there *quickly*.

bryevdv:
I will just go ahead an close this, and make the removal PR along with the necessary release notes.

PR #6966, opened 2017-09-24
Clairetang6/6947 cdsview colors

### LINKED ISSUE #6947: color mapping in circle fill colors does not take current view (CDSView) into account
bokeh 0.12.8rc2, Tornado 4.5.1, Windows 10, Firefox 55.0.3, python 3.6.1

The attached code shall demonstrate a bug:
The source contains 5 data points as a simple example.
The y-values equal to 1 shall be filtered (using BooleanFilter) and additionally colored green, using a linear color mapper.

Filtering works, only the points number 1, 2 and 5 are plotted.
But the coloring is "green, green, blue", instead of "green, green green" as expected.
It seems to take into account the indices of the points number 1,2 and 3 and not of 1,2 and 5.
So obviously the coloring is based on the original, non-filtered source and does not take the current provided view into account.

I would appreciate if this bug could be solved.
And also, if anyone has an idea of a work around in the meantime.

Thanks a lot and best regards,
Juerg

```
# code goes here between backticks
from bokeh.models import ColumnDataSource, CDSView, BooleanFilter, LinearColorMapper
from bokeh.io import show

data = {'x_values': [1, 2, 3, 4, 5],
        'y_values': [1, 1, 0, 0, 1]}

colors = ["blue", "green"]
boolean_colors = LinearColorMapper(palette=colors, low=0, high=1)



source = ColumnDataSource(data=data)

booleans = [True if val == 1 else False for val in source.data['y_values']]

view1 = CDSView(source=source, filters=[BooleanFilter(booleans)])

p=figure()
p.circle("x_values","y_values", size=20, source=source, view=view1, fill_color={'field': "y_values", 'transform': boolean_colors})

show(p)
```

clairetang6:
This PR addresses the issue that vectorized visual properties were not working with the CDSView and filtered data sources. 

The problem was the `@visuals` of a glyph would contain the full lists of visual property values when a property was vectorized, but would use subset indices to index into those lists. That is, if you had a data source with three items and a filtered view with only the last item (index 2 in full data space), the last item has an index of 0 in subset space. Then if the `fill_color`'s were `['blue', 'blue', 'red']` and you indexed with 0, you would get `'blue'` instead of the correct `'red'`.

I considered two ways of fixing this. One would be to change the index at the glyph level. This would mean saving the full data space indices on the glyph and changing all lines like `@visuals.fill.set_vectorize(ctx, i)` to `@visuals.fill.set_vectorize(ctx, @all_indices[i])`. But there are >50 lines across all of the glyphs that would need to be changed. 

The second way, which is implemented here, is to have the `Visuals` object know the full data space indices. The indices are passed to the `visuals` by the glyph in `set_data`. Then the `visuals` instance of each glyph can correctly pass back the correct values.

- [x] issues: fixes #6947 
- [x] tests added / passed


bryevdv:
@clairetang6 This looks good, but it would be better to add a test that ensures that glyph actually sets `all_indices` as well (a mock would do, if that's easiest)

Issue #6955, opened 2017-09-19
Possible bug: Hover tool does not work with filtered source


GenericUser123:
bokeh 0.12.8rc2, Tornado 4.5.1, Windows 10, Firefox 55.0.3, python 3.6.1

A simple ColumnDataSource with 5 example points, splittet up into 3 "groups" ("A", "B" and "C") is filtered using a select widget.
The hover tool of the generated plot works for the points of group "A", which is the default value of the select widget. Selecting group "B" or "C", the points are plotted correctly, but the hover tool will not work anymore.

I tried to change the default value to "B", but the hover tool will still only works with group "A" and not with "B" or "C", so it does not seem to depend on the default value of the select widget.

I guess that might be a bug?
Is there any workaround?

Thanks a lot and best regards,
Juerg

```
# code goes here between backticks
from bokeh.plotting import figure
from bokeh.models import ColumnDataSource, CDSView, GroupFilter
from bokeh.models.widgets import Select
from bokeh.layouts import row, widgetbox
from bokeh.io import curdoc


data = {'x_values': [1, 2, 3, 4, 5],
        'y_values': [1, 1, 0, 0, 1],
        'cat': ["A", "A", "B", "B", "C"]}


cat_filter = Select(title='category', options=["A", "B", "C"], value="A")
source = ColumnDataSource(data=data)


def create_plot():
    view1 = CDSView(source=source, filters=[GroupFilter(column_name='cat', group=cat_filter.value)])
    p=figure(tools="hover")
    p.circle("x_values","y_values", size=20, source=source, view=view1)
    return p


def update(attr, old, new):
    l.children[1]=create_plot()
    
cat_filter.on_change('value', update)


inputs=widgetbox(cat_filter)
l=row(inputs,create_plot())
update

curdoc().add_root(l)

```

aavanian:
I get the same (although using CustomJS callbacks). Color mapping also seems buggy:

Snippet:
```python
import random

from bokeh.layouts import widgetbox
from bokeh.models.mappers import LinearColorMapper
from bokeh.palettes import Viridis256
from bokeh.layouts import row, column
from bokeh.plotting import figure, show, output_notebook
from bokeh.models import ColumnDataSource, CDSView, GroupFilter, CustomJS, Select, HoverTool

data = {
    'age': [i for i in range(100)],
    'label': [random.choice(['a', 'b', 'c']) for _ in range(100)],
    'y': [random.uniform(0, 1) for _ in range(100)]
}
[...]

Issue #6910, opened 2017-09-08
The new feature Filter (CDSView) not behaving has expected.


afonit:
Moving this over from the mailing list to a github issue at Bryan's request.

Here is the mailing list post:
https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/OFrEpYWsTvk

Background:
I am using some of the new features, and trying to combine http://bokeh.pydata.org/en/0.12.7/docs/user_guide/categorical.html#visual-dodge with boolean filters http://bokeh.pydata.org/en/0.12.7/docs/user_guide/data.html#booleanfilter
I am changing the visual dodge to have the data all in a column (like one would have to do with altair)
So I am creating a view for each of the three vbar data sets
but it is not working.

Here is the example:

```python
from bokeh.core.properties import value
from bokeh.io import show, output_notebook
from bokeh.models import ColumnDataSource
from bokeh.plotting import figure
from bokeh.transform import dodge
from bokeh.models import ColumnDataSource, CDSView, BooleanFilter

output_notebook()

fruits = ['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries']
years = ['2015', '2016', '2017']

data = {'fruits' :['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries',
                  'Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries',
                  'Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries'],
        'date': ['2015','2015','2015','2015','2015','2015',
                '2016','2016','2016','2016','2016','2016',
                '2017', '2017','2017','2017','2017','2017'],
        'nums': [2, 1, 4, 3, 2, 4,
                 5, 3, 3, 2, 4, 6,
                 3, 2, 4, 4, 5, 3]
       }

source = ColumnDataSource(data=data)

view2015 = CDSView(source=source, filters=[BooleanFilter([True if date == '2015' else False for date in source.data['date']])])
view2016 = CDSView(source=source, filters=[BooleanFilter([True if date == '2016' else False for date in source.data['date']])])
view2017 = CDSView(source=source, filters=[BooleanFilter([True if date == '2017' else False for date in source.data['date']])])

p = figure(x_range=fruits, y_range=(0, 10), plot_height=250, title="Fruit Counts by Year",
           toolbar_location=None, tools="")

[...]

PR #6967, opened 2017-09-24
fix rendering of filtered data sources for previously broken glyphs

### LINKED ISSUE #6820: CDSView not working with Text glyph
This code does not display the filtered text. Adding the filter to a circle does filter the circles, however. 

cc @clairetang6 

```

import yaml

from bokeh.models import (Circle, ColumnDataSource, DataRange1d, Grid, HoverTool, Label, LabelSet,
                          LinearAxis, Plot, Range1d, SingleIntervalTicker, CDSView, IndexFilter, Text)
from bokeh.sampledata.sprint import sprint
from bokeh.themes import Theme

# Based on http://www.nytimes.com/interactive/2012/08/05/sports/olympics/the-100-meter-dash-one-race-every-medalist-ever.html

abbrev_to_country = {
    "USA": "United States",
    "GBR": "Britain",
    "JAM": "Jamaica",
    "CAN": "Canada",
    "TRI": "Trinidad and Tobago",
    "AUS": "Australia",
    "GER": "Germany",
    "CUB": "Cuba",
    "NAM": "Namibia",
    "URS": "Soviet Union",
    "BAR": "Barbados",
    "BUL": "Bulgaria",
    "HUN": "Hungary",
    "NED": "Netherlands",
    "NZL": "New Zealand",
    "PAN": "Panama",
    "POR": "Portugal",
    "RSA": "South Africa",
    "EUA": "United Team of Germany",
}

fill_color = { "gold": "#efcf6d", "silver": "#cccccc", "bronze": "#c59e8a" }
line_color = { "gold": "#c8a850", "silver": "#b0b0b1", "bronze": "#98715d" }

def selected_name(name, medal, year):
    return name if medal == "gold" and year in [1988, 1968, 1936, 1896] else ""

t0 = sprint.Time[0]

sprint["Abbrev"]       = sprint.Country
sprint["Country"]      = sprint.Abbrev.map(lambda abbr: abbrev_to_country[abbr])
sprint["Medal"]        = sprint.Medal.map(lambda medal: medal.lower())
sprint["Speed"]        = 100.0/sprint.Time
sprint["MetersBack"]   = 100.0*(1.0 - t0/sprint.Time)
sprint["MedalFill"]    = sprint.Medal.map(lambda medal: fill_color[medal])
sprint["MedalLine"]    = sprint.Medal.map(lambda medal: line_color[medal])
sprint["SelectedName"] = sprint[["Name", "Medal", "Year"]].apply(tuple, axis=1).map(lambda args: selected_name(*args))

source = ColumnDataSource(sprint)

xdr =Range1d(start=25, end=-0.5)
ydr = DataRange1d(range_padding=2, range_padding_units="absolute")

plot = Plot(plot_width=1000, x_range=xdr, y_range=ydr, toolbar_location=None)
plot.title.text = "Usain Bolt vs. 116 years of Olympic sprinters"

xticker = SingleIntervalTicker(interval=5, num_minor_ticks=0)
xaxis = LinearAxis(ticker=xticker, axis_label="Meters behind 2012 Bolt")
plot.add_layout(xaxis, "below")

xgrid = Grid(dimension=0, ticker=xaxis.ticker)
plot.add_layout(xgrid)

yticker = SingleIntervalTicker(interval=12, num_minor_ticks=0)
yaxis = LinearAxis(ticker=yticker, major_tick_in=-5, major_tick_out=10)
plot.add_layout(yaxis, "right")

filters = [IndexFilter(list(sprint.query('Medal == "gold" and Year in [1988, 1968, 1936, 1896]').index))]

medal = Circle(x="MetersBack", y="Year", size=10, fill_color="MedalFill", line_color="MedalLine", fill_alpha=0.5)
medal_renderer = plot.add_glyph(source, medal)

#sprint[sprint.Medal=="gold" * sprint.Year in [1988, 1968, 1936, 1896]]
plot.add_glyph(source, Text(x="MetersBack", y="Year", x_offset=10, text="Name"), view=CDSView(source=source, filters=filters))

plot.add_glyph(source, Text(x=7.5, y=1942, text=["No Olympics in 1940 or 1944"],
                            text_font_style="italic", text_color="silver"))

tooltips = """
<div>
    <span style="font-size: 15px;">@Name</span>&nbsp;
    <span style="font-size: 10px; color: #666;">(@Abbrev)</span>
</div>
<div>
    <span style="font-size: 17px; font-weight: bold;">@Time{0.00}</span>&nbsp;
    <span style="font-size: 10px; color: #666;">@Year</span>
</div>
<div style="font-size: 11px; color: #666;">@{MetersBack}{0.00} meters behind</div>
"""

hover = HoverTool(tooltips=tooltips, renderers=[medal_renderer])
plot.add_tools(hover)

theme = Theme(json=yaml.load("""
attrs:
    Plot:
        outline_line_color: None
    Axis:
        axis_line_color: None
        major_tick_line_color: None
        axis_label_text_font_style: "bold"
        axis_label_text_font_size: "10pt"
    Grid:
        grid_line_dash: "dashed"
    Text:
        text_baseline: "middle"
        text_align: "center"
        text_font_size: "9pt"
"""))

doc = Document(theme=theme)
doc.add_root(plot)

if __name__ == "__main__":

    doc.validate()
    filename = "sprint.html"
        f.write(file_html(doc, INLINE, plot.title.text))
    print("Wrote %s" % filename)
    view(filename)
```

clairetang6:
This PR fixes the rendering of glyphs that use a subset of data, through filtering with CDSView. It also fixes some bugs with the hover tool and filtered plots that were introduced with GraphRenderer work. 

The problem was at line 141 in bokehjs/src/coffee/models/renderers/glyph_renderer.coffee. I had mistakenly thought that `all_indices` was in subset space when it really is in full data space. `all_indices` contains the indices of the filtered data source (e.g. [1, 20, 45]) in the full data space. It needs to be converted to `indices` in subset space (i.e. [0, 1, 2]). This previously was working for the circle glyph, because `mask_data` using the spatial index would return the correct subset space `indices`. However, for glyphs that don't have `_mask_data`, `indices` was incorrectly in the full data space. The fix is just to convert `indices` to subset space if it wasn't converted by `mask_data`. 

I also made sure that the Line glyph was being handled properly. It needs special selected/non-selected handling ([PR where line selection rendering is improved](https://github.com/bokeh/bokeh/pull/6491)). I made it so that hovering when a line renderer has a hover_glyph displays the whole line (can be a subset using CDSView) using the hover style.

- [x] issues: fixes #6820 #6910 #6955
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -140,7 +140,7 @@ export class GlyphRendererView extends RendererView
     # all_indices is in full data space, indices is converted to subset space
     # either by mask_data (that uses the spatial index) or manually
     indices = @glyph.mask_data(@all_indices)
-    if '' + indices == '' + @all_indices
+    if indices.length == @all_indices.length
       indices = [0...@all_indices.length]
     dtmask = Date.now() - tmask
 
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -114,7 +114,7 @@ export class GlyphRendererView extends RendererView
     lod_factor = @plot_model.plot.lod_factor
     @decimated = []
     for i in [0...Math.floor(@all_indices.length/lod_factor)]
-      @decimated.push(@all_indices[i*lod_factor])
+      @decimated.push(i*lod_factor)
 
     dt = Date.now() - t0
     logger.debug("#{@glyph.model.type} GlyphRenderer (#{@model.id}): set_data finished in #{dt}ms")
@@ -137,8 +137,11 @@ export class GlyphRendererView extends RendererView
     dtmap = Date.now() - t0
 
     tmask = Date.now()
-    # all_indices and indices are in cdsview subset space
+    # all_indices is in full data space, indices is converted to subset space
+    # either by mask_data (that uses the spatial index) or manually
     indices = @glyph.mask_data(@all_indices)
+    if '' + indices == '' + @all_indices
+      indices = [0...@all_indices.length]
     dtmask = Date.now() - tmask
 
     ctx = @plot_view.canvas_view.ctx
@@ -150,10 +153,7 @@ export class GlyphRendererView extends RendererView
       selected = []
     else
       if selected['0d'].glyph
-        if @glyph instanceof LineView
-          selected = indices
-        else
-          selected = @model.view.convert_indices_from_subset(indices)
+        selected = @model.view.convert_indices_from_subset(indices)
       else if selected['1d'].indices.length > 0
         selected = selected['1d'].indices
       else
@@ -165,10 +165,7 @@ export class GlyphRendererView extends RendererView
       inspected = []
     else
       if inspected['0d'].glyph
-        if @glyph instanceof LineView
-          inspected = indices
-        else
-          inspected = @model.view.convert_indices_from_subset(indices)
+        inspected = @model.view.convert_indices_from_subset(indices)
       else if inspected['1d'].indices.length > 0
         inspected = inspected['1d'].indices
       else
@@ -194,9 +191,15 @@ export class GlyphRendererView extends RendererView
 
     if not (selected.length and @have_selection_glyphs())
         trender = Date.now()
-        glyph.render(ctx, indices, @glyph)
-        if @hover_glyph and inspected.length
-          @hover_glyph.render(ctx, inspected, @glyph)
+        if @glyph instanceof LineView
+          if @hover_glyph and inspected.length
+            @hover_glyph.render(ctx, @model.view.convert_indices_from_subset(inspected), @glyph)
+          else
+            glyph.render(ctx, @all_indices, @glyph)
+        else
+          glyph.render(ctx, indices, @glyph)
+          if @hover_glyph and inspected.length
+            @hover_glyph.render(ctx, inspected, @glyph)
         dtrender = Date.now() - trender
 
     else
@@ -209,14 +212,16 @@ export class GlyphRendererView extends RendererView
       # intersect/different selection with render mask
       selected = new Array()
       nonselected = new Array()
-      for i in indices
-        # now, selected is changed to subset space, except for Line glyph
-        if @glyph instanceof LineView
+
+      # now, selected is changed to subset space, except for Line glyph
+      if @glyph instanceof LineView
+        for i in @all_indices
           if selected_mask[i]?
             selected.push(i)
           else
             nonselected.push(i)
-        else
+      else
+        for i in indices
           if selected_mask[@all_indices[i]]?
             selected.push(i)
           else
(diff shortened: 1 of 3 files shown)


clairetang6:
Script I used to test handling of Line and glyphs like Rect:
```python
from math import pi
from bokeh.models import CDSView, IndexFilter, ColumnDataSource, Line
from bokeh.layouts import row

source = ColumnDataSource({'x': [1, 2, 3, 4, 5], 'y': [1, 2, 3, 4, 5]})
view = CDSView(source=source, filters=[IndexFilter([1, 2, 3, 4])])

p1 = figure(plot_width=400, plot_height=400, tools=['box_select', 'hover', 'reset'], x_range=(0, 6), y_range=(0, 6))
p1.rect(x='x', y='y', width=0.2, height=40, color="#CAB2D6", angle=pi/3, height_units="screen", source=source, view=view)

r = p1.line(x='x', y='y', source=source, view=view, line_width=6)
r.hover_glyph = Line(line_color="red", line_width=2)

p2 = figure(plot_width=400, plot_height=400, tools=['box_select', 'tap', 'reset'], x_range=(0, 6), y_range=(0, 6))
p2.rect(x='x', y='y', width=0.2, height=40, color="#CAB2D6", angle=pi/3, height_units="screen", source=source, view=view)

r = p2.line(x='x', y='y', source=source, view=view, line_width=6)
r.hover_glyph = Line(line_color="red", line_width=2)

show(row(p1, p2))
```

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
```diff
@@ -137,8 +137,11 @@ export class GlyphRendererView extends RendererView
     dtmap = Date.now() - t0
 
     tmask = Date.now()
-    # all_indices and indices are in cdsview subset space
+    # all_indices is in full data space, indices is converted to subset space
+    # either by mask_data (that uses the spatial index) or manually
     indices = @glyph.mask_data(@all_indices)
+    if '' + indices == '' + @all_indices
+      indices = [0...@all_indices.length]
```
bryevdv: Is this testing equality by converting to strings? Wouldn't s length comparison suffice? They can't be equal unless they are the same length and (assuming duplicate entries are not possible) they will only be the same length when they are equal.

clairetang6: Yes, it is meant to see if mask_data was a no-op. So, I think the length comparison would result in the correct behavior, but the logic would be a little bit wrong. Right now, all of the glyphs that implement _mask_data use the spatial index, which will return the correctly masked data in subset space. So [1, 20, 45] -> [0, 1, 2]. Here the length is the same, but we don't need to convert the indices to subset space. That said, the conversion would simply result in the same [0, 1, 2]. If mask_data does return a shorter array, then it will correctly not run the additional conversion ([1, 20, 45] -> [0, 2] -/> [0, 1], that last arrow would be wrong to do). This code is not ideal because it's not clear that mask_data should convert from full data space to subset space, it just happens to do so.

bryevdv: I guess I am worried about performance implications. For scenarios that are currently just on the edge of possible (let's say 50k pts for a concrete plausible number), this means serializing two 50k element lists to strings and comparing on every render? I think that may push "currently just possible" cases into "no longer reasonable" territory. 

When you say "So, I think the length comparison would result in the correct behavior, but the logic would be a little bit wrong" do you mean that the logic is simply indirect and possibly confusing to anyone looking at it? Or that it only happens to be correct in this case, but will be wrong in other cases? If the former, I think that is solved with a comment explaining why the code is doing the comparison

bryevdv:
@clairetang6 is this `ready` ?

bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
clairetang6: Yes, it would be the former, so I agree the length comparison would be the better way to go given the performance cost.

clairetang6:
I was planning to add some tests to this PR, but I just created a new issue for them, so this PR is ready.

PR #6981, opened 2017-09-28
Rewrite client/* and protocol/* in typescript

mattpap:
addresses #6481


mattpap:
@bryevdv, I will need you to verify types in `client/connection.ts`, especially around promises and resolve/reject, as this caused me a lot of work.

bryevdv:
@mattpap The tests for client and protocol are not quite extensive, so I'd say get the last remaining tests passing then let's schedule a call to go over it together (mostly for my benefit to ask some TS syntax questions)

bryevdv:
@mattpap do you have a handle on what it would take just to get the tests passing?

bryevdv:
@mattpap the old tests checked the state handlers to return `null` which is not important, just what they happened to do. I changed the test to test for `undefined` returns.

bryevdv:
I tried out several apps they all seem to function as expected.

mattpap:
> I tried out several apps they all seem to function as expected.

That's good. Still we should go through those types, perhaps after tomorrow's meeting.

bryevdv:
Sounds good.

Issue #6986, opened 2017-09-30
MercatorTicker behavior poorly defined for ranges exceeding mercator bounds


philippjfr:
When using the MercatorTicker with a tile source you can often get a range that exceeds the valid range of mercator coordinates, which results in ill defined behavior. Either the axis ranges disappear completely or you end up with a partial set of ticks. This is for one of two reasons, certain coordinates will trip proj4.js up because they are out of bounds, and others will wrap around when projected causing the ``proj_low`` to be higher than ``proj_high``, which ends up tripping up how ticks are drawn. The simplest solution to address this is simply to clip the mercator coordinates before projecting to the allowed bounds by the mercator projection as defined [here](https://epsg.io/3857).

Here's a simple example:

```python
from bokeh.io import show
from bokeh.plotting import figure
from bokeh.resources import INLINE
from bokeh.tile_providers import STAMEN_TONER
from bokeh.models import MercatorTicker, MercatorTickFormatter

bound = 20000000 # meters
fig = figure(tools='pan, wheel_zoom', plot_width=800, x_range=(-bound, bound), y_range=(-bound, bound))
fig.xaxis[0].ticker = MercatorTicker(dimension='lon')
fig.yaxis[0].ticker = MercatorTicker(dimension='lat')
fig.xaxis[0].formatter = MercatorTickFormatter(dimension='lon')
fig.yaxis[0].formatter = MercatorTickFormatter(dimension='lat')
fig.add_tile(STAMEN_TONER)
show(fig)
```

<img width="777" alt="screen shot 2017-09-30 at 1 04 08 pm" src="https://user-images.githubusercontent.com/1550771/31045625-e3fab2fa-a5df-11e7-8cad-f038f921057d.png">

As you can see there are no ticks to begin with and the console shows the following error:

```
VM6513:362 Uncaught TypeError: Cannot read property 'x' of null
    at n (eval at append_javascript (outputarea.js:747), <anonymous>:362:9723)
    at Object.forward (eval at append_javascript (outputarea.js:747), <anonymous>:362:9991)
    at e.r.MercatorTicker.e.get_ticks_no_defaults (eval at append_javascript (outputarea.js:747), <anonymous>:254:31257)
    at e.r.Ticker.e.get_ticks (eval at append_javascript (outputarea.js:747), <anonymous>:255:1848)
    at e.r.Axis.e._tick_coords (eval at append_javascript (outputarea.js:747), <anonymous>:247:4150)
[...]

PR #6987, opened 2017-09-30
Clip MercatorTicker coordinates to officially defined bounds

philippjfr:
This PR fixes issues with poorly defined behavior of the MercatorTicker when the axis ranges exceed the officially defined bounds of the pseudo-mercator projection (as described in 
https://github.com/bokeh/bokeh/issues/6986). The coordinates the ticks are drawn between are clipped to the officially defined bounds defined [here](https://epsg.io/3857), resulting in consistent behavior when zoomed out.

screen shot 2017-09-30 at 1 13 25 pm


bryevdv:
@philippjfr this is failing due to whitespace:
```
  where 1 = len(['File contains trailing whitespace: bokehjs/src/coffee/core/util/proj4.coffee, line 28.'])
```

However, there should also be simple unit tests for `clip_mercator` and `mercator_bounds` and a unit test exercising the ticker with out of bounds values.

jbednar:
Seems like a reasonable approach, thanks!

philippjfr:
>However, there should also be simple unit tests for clip_mercator and mercator_bounds and a unit test exercising the ticker with out of bounds values. I think you'll need to add new test files for this.

Sounds good, I'll add those shortly.

philippjfr:
@bryevdv I'm pretty confused, I've been trying to write tests for the MercatorTicker when set to ``dimension='lat'`` but keep running into this error:

>TypeError: Cannot read property 'x' of null

I've even tried copying the exact arguments that get passed into the method when I interactively zoom in the notebook where I don't see errors, but when run as a test it always breaks. Is it possible that the browser does something differently?

philippjfr:
Nevermind figured it out.

philippjfr:
The issue was that for small ``desired_n_ticks`` the computed ticks would also sometimes fall outside the bounds resulting in errors while projecting. I'm now also dropping ticks which are not within the bounds, but that only seems to occur for ``num_ticks=2`` when you're zoomed out beyond the bounds.

jbednar:
Looks good to me; eager to see it merged!

bryevdv:
@mattpap this jusst failed when merged to master:
```
    @pytest.mark.quality
    def test_files():
        errors = collect_errors()
E         File contains trailing whitespace: bokehjs/src/coffee/core/util/proj4.coffee, line 28.
E       assert 1 == 0
E        +  where 1 = len(['File contains trailing whitespace: bokehjs/src/coffee/core/util/proj4.coffee, line 28.'])
```
 But was green before then, what's going on?

PR #6988, opened 2017-09-30
Fix bug : no URL in IE

### LINKED ISSUE #6982: Bugfix: Bokeh-Server: IE fails with "Object doesn't support this action"
Dear Bokeh coworkers!

I am Volker liking to contribute to Bokeh.

I found a severe bug in 
https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/embed.coffee
targeting the IE.
**It uses the URL() object that is not available in IE.**

_ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)_
Server: Bokeh 0.12.9, Python 2.7, Debian. HTTPS/WSS.
bokeh serve --num-procs=2 --port 5006 --address 127.0.0.1 --allow-websocket-origin=bs_plone.inqbus.de profileview.py 
Living behind nginx
Clients IE (all other OS/browsers not affected).

_Description of expected behavior and the observed behavior_
IE shows a red box with "Object doesn't support this action" instead of page.

The problematic code is
```
  if absolute_url?
    loc = new URL(absolute_url)
  else
loc = window.location
```
The URL object is not available in IE. https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

BugFix (stolen at https://gist.github.com/jlong/2428561):
```
  if absolute_url?
    loc = document.createElement('a')
    loc.href = absolute_url
  else
    loc = window.location
```
I have patched my Bokeh.js and it worked like a charm!

volkerjaenisch:
issues: fixes #6982

no URL in IE

### DIFF
--- a/bokehjs/src/coffee/embed.coffee
+++ b/bokehjs/src/coffee/embed.coffee
@@ -168,7 +168,8 @@ export embed_items = (docs_json, render_items, app_path, absolute_url) ->
     protocol = 'wss:'
 
   if absolute_url?
-    loc = new URL(absolute_url)
+    loc = document.createElement('a')
+    loc.href = absolute_url
   else
     loc = window.location
 


bryevdv:
Thanks @volkerjaenisch !

PR #6994, opened 2017-10-01
canonicalize bokeh.colors

bryevdv:
* adds API declarations
* splits named colors into bokeh.colors.named
* adds bokeh.colors.groups
* adds docstrings and missing docs from refguid
* sphinx extension for dispaying colors in docs
* 100% test coverage for the package


bryevdv:
After a little more experience with the API declarations I do want to go back and make one change before proceeding further. Explicitly marking class methods public/internal can intersect pretty weirdly with inheritance. I don't want to chase that rabbit hole and don't think its necessary: method are simply visible or not (i.e. underscore or not) and have the API level of the class they live on. Just need to add somethign to mark the initial version but not API level.

bryevdv:
@bokeh/dev I'd like to merge this tonight or early tomorrow and get a dev build out

bryevdv:
@bokeh/dev FYI with this PR, Bokeh (python) test coverage has finally attained 90%. I'd like to proposed adding the option to specify this as a min test coverage amount on CI runs (else a failure)

PR #6997, opened 2017-10-01
add Scalar versions of property mixins

### LINKED ISSUE #5648: Line renderer raises the "without value specification" when a ColumnDataSource is used
bokeh 0.12.3

Line renderer raises the JS error, while Circle renderer works as expected. There's probably a workaround to define a separate source column for each value of width (highly inefficient).

```python
from bokeh.io import output_notebook, push_notebook

output_notebook(hide_banner=True)

source = ColumnDataSource(dict(x=[1.2, 3.4], y=[3.2, 4.5], sz=[20.5, 40.1], lw=[1.1, 2.2]))

plot = figure()
plot.circle('x', 'y', source=source, size='sz') # This line works fine
plot.line('x', 'y', source=source, line_alpha2='sz') # This line raises 
# Uncaught Error: attempted to retrieve property value for property without value specification
show(plot)
```

bryevdv:
- [x] issues: fixes #5648
- [x] tests added / passed

@mattpap if you agree with this general approach, I will:

* add a real `FontSize` prop instead of `String`
* apply these scalar props other places to reduce code dupe (some annotations)
* add more tests
* general clean up

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -312,6 +312,19 @@ class String(PrimitiveProperty):
     '''
     _underlying_type = string_types
 
+class FontSize(String):
+
+    _font_size_re = re.compile(r"^[0-9]+(.[0-9]+)?(%|em|ex|ch|ic|rem|vw|vh|vi|vb|vmin|vmax|cm|mm|q|in|pc|pt|px)$", re.I)
+
+    def validate(self, value):
+        super(FontSize, self).validate(value)
+
+        if isinstance(value, string_types):
+            if len(value) == 0:
+                raise ValueError("empty string is not a valid font size value")
+            elif self._font_size_re.match(value) is None:
+                raise ValueError("%r is not a valid font size value" % value)
+
 class Regex(String):
     ''' Accept strings that match a given regular expression.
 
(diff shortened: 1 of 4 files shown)


bryevdv:
FWIW I think we should just go this route rather than try to add a flag to make dataspecs scalar or vector, they are already too complicated, and also have different documentation requirements. Just cleaving by making scalar versions of mixins covers all the cases I can think of without disturbing existing working but complicated code.

mattpap:
In need to look closer at this, but this is pretty much how I would (or will) do this in typescript. One shouldn't view this kind of code as duplication, but as precise specification.

bryevdv:
> I need to look closer at this, but this is pretty much how I would (or will) do this in typescript. One shouldn't view this kind of code as duplication, but as a precise specification.

I think I communicated poorly, I agree with this 100%. My comment about duplication was that some annotations currently do this "by hand" so adding these scalar versions of mixing will reduce duplication there.

bryevdv:
In any case since I have your tentative approval I will finish things up

bryevdv:
Going to save more work for later PRs this is `ready` for review

bryevdv:
I'll merge this this afternoon unless there is objection

PR #7008, opened 2017-10-03
Allow toolbar in side panels

### LINKED ISSUE #1651: Grid Plot with row of all None behaves badly
See #1640 for more details.

### LINKED ISSUE #4574: Plot border gets covered up by things stuck on edge
Easiest place to see it is on burtin example in this branch: https://github.com/bokeh/bokeh/pull/4576

screen shot 2016-06-19 at 6 19 43 am

When the frame has a border it overlaps with the content on the outside.

### LINKED ISSUE #4874: Toolbar css class occasionally is `bk-toolbar-null`
See notes on #4813 and #4856

### LINKED ISSUE #6065: Touch events fail in device mode under chrome devtools
The following code in `core/ui_events.coffee`:
```
_bokify_hammer: (e, extras={}) ->
    if e.pointerType == 'mouse'
      x = e.srcEvent.pageX
      y = e.srcEvent.pageY
    else
      x = e.pointers[0].pageX
      y = e.pointers[0].pageY
```
fails with "can't read property pageX of undefind". `e.pointers` is empty. However, `e.changedPointers` isn't and contains valid coordinates. Just to note, `e.srcEvent.{pageX,pageY}` are valid as well. I'm not sure if this is a bug in hammer or we use hammer improperly.

@bryevdv, do you know why this code is written like this?

### LINKED ISSUE #6466: Rendering error with toolbar_sticky=False and sizing_mode='scale_width'
After upgrading to bokeh 0.12.6, I now have a javascript rendering issue when using both ``toolbar_sticky=False`` when making a figure and ``sizing_mode='scale_width'`` for a column layout. It seems to work for different sizing modes or if toolbar sticky is True. I think the bug was introduced from 0.12.3rc3 -> 0.12.6rc4 and is still present on ``master``.

Here is a minimal example that I use the bokeh server to run with ``bokeh serve demo.py``

```python
from bokeh.io import curdoc  

p = figure(toolbar_location='above',  toolbar_sticky=False)

l0 = p.line(x=[0, 1], y=[1, 1]) 
l = column([p], sizing_mode='scale_width') 
curdoc().add_root(l) 
```

With bokeh==0.12.6rc3, the result is
image

And with bokeh>0.12.6rc3 (both 0.12.6rc4 and 0.12.6),
image

If the javascript error isn't clear it is
```
Error rendering Bokeh items  Error: unsatisfiable constraint
    at t.addConstraint (bokeh.min.js?v=ef5547c…:85)
    at t.add_constraint (bokeh.min.js?v=ef5547c…:2)
    at e.r.CanvasView.e.set_dims (bokeh.min.js?v=ef5547c…:25)
    at e.r.PlotCanvasView.e.render (bokeh.min.js?v=ef5547c…:31)
    at e.r.PlotCanvasView.e._layout (bokeh.min.js?v=ef5547c…:31)
    at e.r.LayoutDOMView.e._layout (bokeh.min.js?v=ef5547c…:29)
    at e.r.LayoutDOMView.e._layout (bokeh.min.js?v=ef5547c…:29)
    at e.r.LayoutDOMView.e._do_layout (bokeh.min.js?v=ef5547c…:29)
    at e.r.LayoutDOMView.e.layout (bokeh.min.js?v=ef5547c…:29)
    at e.r.DOMView.e.renderTo (bokeh.min.js?v=ef5547c…:1)
```

There are no errors from the bokeh server.

Versions:
```
bkcharts                  0.2                      py36_0  
bokeh                     0.12.6                   py36_0  
jinja2                    2.9.6                    py36_0  
markupsafe                0.23                     py36_2  
mkl                       2017.0.1                      0  
numpy                     1.13.0                   py36_0  
openssl                   1.0.2l                        0  
pandas                    0.20.2              np113py36_0  
pip                       9.0.1                    py36_1  
python                    3.6.1                         2  
python-dateutil           2.6.0                    py36_0  
pytz                      2017.2                   py36_0  
pyyaml                    3.12                     py36_0  
readline                  6.2                           2  
requests                  2.14.2                   py36_0  
setuptools                27.2.0                   py36_0  
six                       1.10.0                   py36_0  
sqlite                    3.13.0                        0  
tk                        8.5.18                        0  
tornado                   4.5.1                    py36_0  
wheel                     0.29.0                   py36_0  
xz                        5.2.2                         1  
yaml                      0.1.6                         0  
zlib                      1.2.8                         3 

Fedora 25 Linux 4.8.6-201.fc24.x86_64
Chrome Version 58.0.3029.81 (64-bit)
```

### LINKED ISSUE #6990: Allow toolbar in side panels and drop sticky property
Although `toolbar_sticky` is the default, it rarely works properly, e.g. overlapping with axes, unless you stick to default configuration. Allowing toolbars in side panels would allow us to properly maintain relationship between toolbar and axes, annotations, etc. I'm not sure about the details right now, because toolbar is a `LayoutDOM`, so there may be some deeper changes needed. Those changes may be necessary anyway as we want to have off-canvas legends (situation isn't exactly the same, but it's close). This could also allow us to merge `Plot` and `PlotCanvas` again, if desired. It would also, in a longer term, allow us to have multiple toolbars (e.g. main split or having separate animation toolbar) and sharing one panel between a toolbar and e.g. a legend.

### LINKED ISSUE #7075: Changing Title attributes does not work as expected in Bokeh server
To reproduce. A title is firstly aligned left. After a change of the user it shall be positioned centred. 
test.py
```
from bokeh.io import curdoc
from bokeh.models.widgets import TextInput

text = TextInput(title="title", value='my sine wave')

p = figure(plot_width=400, plot_height=400, title='Hello')

p.circle([1, 2, 3, 4, 5], [2, 5, 8, 2, 7], size=10)

p.title.align='left'
                    
# Set up callbacks
def update_title(attrname, old, new):
    p.title.text = text.value
    p.title.align = "center"

text.on_change('value', update_title)

root = column(text, p)
doc = curdoc()
doc.add_root(root)
```
```
bokeh  serve --port=5000 test.py
```
Enter text. The text is aligned at the center, but starting right of the center.
Same holds for font size and probably other parameters.

mattpap:
This is early work in progress. ~~Currently tooltips and detached toolbars don't work~~. I removed all special code handling toolbars and allowed them in panels. Styling is the same as before, but there are no more overlapping toolbars with other entities.

toolbars

fixes #6990
fixes #4874
fixes #1651
fixes #6466 
fixes #6065 
fixes #4574 
fixes #7075


mattpap:
@bryevdv, fortunately it's sufficient to prefix all flex code with `-webkit` to make it work in phantomjs.

bryevdv:
I generally like the approach here but I would not consider a toolbar to be an annotation. Is there a technical difficulty with a different hierarchy?

mattpap:
It can be either an annotation or a renderer. Otherwise, there will be issues with class hierarchy. I don't consider it as an annotation, but currently there isn't much I can do without further changes. I will have to do those anyway, but not here, as there is already too many changes.

bryevdv:
OK fair enough, we should try to find a new name for it before release, or just decided to live with that name.

mattpap:
Initially I didn't intend to bring it to python, so it wouldn't be a problem, but server didn't like this, because this annotation needs to be added to `Plot.{above,below,left,right}` which get synchronized and must contain only public models. We could implement the hack we use for `renderers`, but that isn't much better. The best solution would be to cleanly separate model's and view's subcomponents, but that's a lot of work.

bokehjs/src/coffee/models/annotations/index.ts
```diff
@@ -1,20 +1,21 @@
...
+export {ColorBar}          from "./color_bar"
+export {Label}             from "./label"
+export {LabelSet}          from "./label_set"
+export {Legend}            from "./legend"
+export {LegendItem}        from "./legend_item"
+export {PolyAnnotation}    from "./poly_annotation"
+export {Span}              from "./span"
+export {TextAnnotation}    from "./text_annotation"
+export {Title}             from "./title"
+export {ToolbarAnnotation} from "./toolbar_annotation"
```
bryevdv: I'm OK with `Toolbar` inheriting from `Annotation` if it must be (tho I'd love to fix it sometime), but we should just call it `Toolbar` not `ToolbarAnnotation` even so

bokeh/core/has_props.py
```diff
@@ -259,8 +259,9 @@ def __setattr__(self, name, value):
 
         props = sorted(self.properties())
         deprecated = getattr(self, '__deprecated_attributes__', [])
+        descriptor = getattr(self.__class__, name, None)
 
-        if name in props or name in deprecated:
+        if name in props or name in deprecated or (descriptor is not None and descriptor.fset is not None):
             super(HasProps, self).__setattr__(name, value)
```
bryevdv: Can you provide context for this change

bokeh/models/plots.py
```diff
@@ -195,6 +197,16 @@ def tools(self):
...
+    @property
+    def toolbar_sticky(self):
+        deprecated("Plot.toolbar_sticky property is no longer needed, and its use is deprecated."\
+                   "In the future, accessing Plot.toolbar_sticky will result in an AttributeError.")
+        return True
+
+    @toolbar_sticky.setter
+    def toolbar_sticky(self, val):
+        deprecated("Plot.toolbar_sticky property is no longer needed, and its use is deprecated."\
+                   "In the future, accessing Plot.toolbar_sticky will result in an AttributeError.")
```
bryevdv: Will need a release note about this as well

bokeh/models/tools.py
```diff
@@ -138,35 +133,20 @@ class Toolbar(ToolbarBase):
...
-    toolbar_location = Enum(Location, default='right', help="""
-        Should the toolbar be presented as if it was stuck to the `above`, `right`, `left`, `below`
-        edge of a plot. Default is `right`.
-    """)
 
-    tools = List(Instance(Tool), help="""
-    A list of tools to add to the plot.
-    """)
-
-    merge_tools = Bool(default=True, help="""
```
bryevdv: Also need a release note about this if we are removing it. How does a user control merged vs non-merged intent?

mattpap: This actually was never used. There is still `merge_tools` option to `gridplot()` which governs whether a grid plot will have a shared toolbar or not.

bokehjs/src/coffee/models/annotations/index.ts
mattpap: I need to figure out how to do side panels anyway, and thus annotations, so this will be done. However, I could rename it now to e.g. `ToolbarPanel` (after all, there is already `ToolbarBox` for detached toolbars). Renaming to just `Toolbar` would mean renaming actual `Toolbar` to something else. Currently we can't have two models with the same name.

bokeh/core/has_props.py
mattpap: This isn't currently strictly necessary, because I removed commits using this (at least for now). This change is necessary when you have a native property with a setter which isn't deprecated. This `name in deprecated` clause is more of a hack and could be removed in the light of this change. In particular I had a change like this:
```
+    @property
+    def toolbar_location(self):
+        return self.toolbar.location
+
+    @toolbar_location.setter
+    def toolbar_location(self, location):
+        self.toolbar.location = location
```
and setting `plot.toolbar_location = whatever` would result in `AttributeError`.

mattpap: Obviously I need to add a test for this.

mattpap:
There is one more thing I need to figure out in PR. I somewhere a long the way I broke layouting of markup widgets.

bokeh/models/tools.py
bryevdv: OK

bokeh/core/has_props.py
```diff
@@ -258,9 +257,9 @@ def __setattr__(self, name, value):
             return
 
         props = sorted(self.properties())
-        deprecated = getattr(self, '__deprecated_attributes__', [])
+        descriptor = getattr(self.__class__, name, None)
```
bryevdv: It might be nice to make `lookup` return `None` if there is no descriptor and use that: https://github.com/bokeh/bokeh/blob/810c35d059d4d3fac03a1a6f1f1254d2fc7ccf49/bokeh/core/has_props.py#L391

mattpap:
So, actually I didn't break any anything, just my other fixes revealed that layout around widgets was broken all along. I think I will leave this for another PR next week.

bokeh/core/has_props.py
mattpap: This may affect how other code uses `lookup()`, so I wouldn't proceed with this in this PR.

bryevdv:
I will look at this now and plan to merge today.

bryevdv:
@mattpap what's going on here:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/models/file/glyphs.png

bryevdv:
Also this one, the table is covering up some text I believe:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/models/file/widgets.png

bryevdv:
This one dropped the space between the custom divs now they overlap:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/plotting/file/css_classes.png

bryevdv:
OK, now we come to the crux of why sticky existed. This example looks much worse now:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/plotting/file/elements.png

A "long" plot, with a title on the top left and the tool bar on the top right, *both* abutting the plot area is something we want to be able to support.

mattpap:
> OK, now we come to the crux of why sticky existed. This example looks much worse now:

This can be fairly easily restored by allowing multiple annotations/panels in one side panel, which I plan to do at some point (perhaps sooner than later).

bryevdv:
There are no toolbars in this example, so I'd like to at least understand what is differing:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/howto/layouts/manual_grid0.12.10dev2-ff66b42-diff.png

bryevdv:
> This can be fairly easily restored by allowing multiple annotations/panels in one side panel, which I plan to do at some point (perhaps sooner than later). 

OK, we need at least to figure out a plan sooner, and I don't want to make a release with this broken, so an option would be a quicker `0.12.10` without this, then merge, or figuring it out sooner. Here is another example that looks much worse:

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/plotting/server/selection_histogram.png

bryevdv:
In general, (text?) divs seem  to have less padding, resulting in thing shifting up?

https://s3.amazonaws.com/bokeh-travis/0.12.10dev2-7e0beeb/app/movies0.12.10dev2-ff66b42-diff.png

mattpap:
> what's going on here:

All there comments are related to what I wrote in https://github.com/bokeh/bokeh/pull/7008#issuecomment-334829706 and https://github.com/bokeh/bokeh/pull/7008#issuecomment-334873840. The implementation of widget layout is completely wrong and previously worked mostly by chance, and improvements to layout in general reveal those issues. I actually panned to fix this, but not necessarily now, but now it's unavoidable.

bryevdv:
OK here are the things that seem to need attention:

* figure out why some divs are taking up less vertical space
* glyphs/tabs example
* plan for title and toolbar to share space

mattpap:
> OK, we need at least to figure out a plan sooner, and I don't want to make a release with this broken, so an option would be a quicker 0.12.10 without this, then merge, or figuring it out sooner. Here is another example that looks much worse:

All examples with this particular set of panels will look worse than before. I think that postponing merge of this PR is a good idea, especially in the light of problems with widget layout, which I didn't anticipate will need to be fixed right now.

bryevdv:
OK, in that case please prioritize known regressions and we can target late next week for a release (without this PR)

bryevdv:
I don’t support deprecating `align`. `text_align` refers to the low level canvas anchor point of the text bounding box, not a position inside a larger div. it would make sense to have both with the latter controlling where the former is positioning. Or `text_align` can be set internally to by a heuristic to be appropriate for  the `align` value, which is actually my preference

mattpap:
@bryevdv, I'm not sure what's up with python unit tests, e.g. https://travis-ci.org/bokeh/bokeh/jobs/290778851#L1263. This doesn't fail locally for me.

bryevdv:
This is failing for me locally I will see what I can find out

bryevdv:
OK the first thing I can note is that the tests only fail on this branch, the pass on master, to sit would in fact seem to be an issue with the changes in the branch. I don't know what the effect of the extra self-closing `<g/>` tag is. If it's harmless we could consider just updating the test.

canavandl:
I'm not sure why the SVG repr of the plot has a new, empty g tag, I assume it's related to how we're creating plot areas (frames and side bars). The failing test isn't a great test anyway, I wouldn't be opposed to stripping all of the ``g`` tags from the tested output and comparing just the svg body.

bryevdv:
@mattpap I would support either solution, pending a quick visual inspection of an actually-generated svg

mattpap:
I wasn't able to reproduce this before, because I don't generate minified js files during development, `test_export` requires such, contrary to everything else that uses non-minified version (e.g. examples' tests).

mattpap:
I updated those tests. Stripping empty paths should be canvas2svg's job.

mattpap:
> I don’t support deprecating align.

I retracted from that.

mattpap:
Implementation of multiple annotations per side panel is more complex that I initially thought. I would like this PR get merged, and I will continue fixing this and widget layouts in following PRs.

bryevdv:
> Implementation of multiple annotations per side panel is more complex that I initially thought.

Unfortunately we have been in this situation before, of merging work that was very strongly contingent on other work later and then having releases held up as a result, so I am -1 on merging this until the functionality is restored:

* I'd like to release again in ~1 week or thereabouts
* But 100% won't release while top titles and toolbar capability is missing

So we will just have to take more care to avoid conflicts. Some other possibilities: 

* If there are smaller chunks of this that can be broken off and merged sooner and independently, that seems reasonable

* If you are wanting to isolate the work for organizational purposes, then branching a new PR to continue off this one seems prudent

bryevdv:
OK the examples report looks generally good [1], I'd like to spend some time running code from this branch and also going over the changes diff. Let's plan on trying to get this merged tomorrow. 

[1] only diff I noticed about current sticky toolbars is that they are slightly offset a few pixels from the inner plot frame. I can live with this for now if it's fixed later.

bokeh/models/annotations.py
```diff
@@ -879,9 +879,12 @@ class Title(TextAnnotation):
...
 
-    align = Enum(TextAlign, default='left', help="""
-    Location to align the title text.
+    vertical_align = Enum(VerticalAlign, default='bottom', help="""
+    Vertical aligment of the text.
+    """)
 
+    align = Enum(TextAlign, default='left', help="""
+    Horizontal aligment of the text.
     """)
```
bryevdv: We might want to mark this for re-visting before release. I think it might be better to always just have one "align" which is agnostic to orientation. It is always relative *to the direction of the text*, whether horizontal or vertical. Then the setting makes sense even if you switch from horizontal to vertical (this way does not)

bryevdv:
I've run a majority of examples, the only issues I've run into (neither showstoppers)


toolbars3 seems messed up:

screen shot 2017-10-24 at 19 03 18

sliders overlap plot in dashboard.py (I think this might be an old issue):

screen shot 2017-10-24 at 18 44 32

mattpap:
> toolbars3 seems messed up:

This is now fixed.

> sliders overlap plot in dashboard.py (I think this might be an old issue):

Yes, sliders were always too wide (widget padding is not accounted for when sizing sliders).

mattpap:
> only diff I noticed about current sticky toolbars is that they are slightly offset a few pixels from the inner plot frame.

Any particular example? Currently sticky toolbars don't overlap frame border, so depending on colors involved, it may be an optical illusion that there's a 1px offset. I though the same about e.g. plotting/file/elements.py, but under zoom there is no visible gap.

bokeh/models/annotations.py
mattpap: This docstring is actually wrong. This `align` relates to the "major" dimension, so horizontal if panel is above or below, and vertical if left or right. `vertical_align` acts on the orthogonal dimension.

bryevdv:
@mattpap I updated the help text in the GH UI, will merge when green.

mattpap:
> I updated the help text in the GH UI, will merge when green.

OK.

I verified all iris examples, and iris_splom has a gap between toolbar and plots, but it uses a grid, so this is how it worked before. This can be changed in future work.

PR #7013, opened 2017-10-04
Raise error when jupyter notebook is not installed or is obsolete

### LINKED ISSUE #7005: output_notebook() fails with TypeError: OutputArea.prototype.mime_types is not a function
Hello,

I am running Chrome Version 61.0.3163.100 (Official Build) (64-bit) on Win-7 64-Bit 

with Bokeh 0.12.7 Juptyer Notebook 5.0.0 ,and Py 3.6.2, output_notebook() works, while

with Bokeh 0.12.9, Juptyer Notebook 4.0.6, and Py 2.7.10, output_notebook() fails with the follow result

```
Loading BokehJS ...
TypeError: OutputArea.prototype.mime_types is not a function
```
```
TypeError: OutputArea.prototype.mime_types is not a function
    at eval (eval at append_javascript (outputarea.js:625), <anonymous>:130:30)
    at eval (eval at append_javascript (outputarea.js:625), <anonymous>:266:2)
    at OutputArea.append_javascript (outputarea.js:625)
    at OutputArea.append_mime_type (outputarea.js:569)
    at OutputArea.append_display_data (outputarea.js:533)
    at OutputArea.append_output (outputarea.js:319)
    at OutputArea.handle_output (outputarea.js:233)
    at output (codecell.js:351)
    at Kernel._handle_output_message (kernel.js:1012)
    at i (jquery.min.js:4)
```
I did get a CORs policy notice on the console initially about the font in the developer tools but upon restart that error disappeared.

mattpap:
fixes #7005

### DIFF
--- a/bokeh/io/state.py
+++ b/bokeh/io/state.py
@@ -55,10 +55,13 @@ from six import string_types
 # Bokeh imports
 from ..document import Document
 from ..resources import Resources
+from ..util.dependencies import import_optional
 
 #-----------------------------------------------------------------------------
 # Globals and constants
 #-----------------------------------------------------------------------------
+
+notebook = import_optional("notebook")
 
 #-----------------------------------------------------------------------------
 # Public API
@@ -189,6 +192,11 @@ class State(object):
             None
 
         '''
+        if not notebook:
+            raise RuntimeError("jupyter notebook is not installed")
+        if notebook.version_info < (5, 0, 0):
+            raise RuntimeError("jupyter notebook 5.0 or higher is required")
+
         self._notebook = True
         self.notebook_type = notebook_type
 


bokeh/io/state.py
```diff
@@ -189,6 +192,11 @@ def output_notebook(self, notebook_type='jupyter'):
             None
 
         '''
+        if not notebook:
+            raise RuntimeError("jupyter notebook is not installed")
+        if notebook.version_info < (5, 0, 0):
+            raise RuntimeError("jupyter notebook 5.0 or higher is required")
+
```
bryevdv: will this work with Jupyter Lab?

mattpap: I was about to say no, however jupyter notebook is listed as a dependency in jupyterlab's conda package and in prerequisites section in jupyterlab's README, so I presume it should work, though I will need to test this. However, what's confusing is that their minimal version number is 4.3.

mattpap: If this approach won't work, then this PR doesn't make sense, because not knowing what environment bokeh runs in, we can't have separate checks for notebook and lab.

bryevdv: I think the JS approach might be safer

mattpap: Yeah, that will be better.

PR #7021, opened 2017-10-08
Canonicalize more top level modules

bryevdv:
This PR makes a few minor improvements to the testing machinery (re-use `verify_all` code, have `verify_api` return a class to simplify boilerplate) and also brings the following modules under API test and canonical module structure:

* `bokeh.__init__`
* `bokeh.__main__`
* `bokeh.driving`
* `bokeh.tile_providers`
* `bokeh.transform`

This PR also makes the `_force` function in `bokeh.driving` be public, `force`


bryevdv:
OK, this actually should not be controversial in any way. It's blocking another PR I am working on so I am going to go ahead and merge. @mattpap if you have any concerns after the fact please let me know.

PR #7022, opened 2017-10-09
Clean up sampledata

bryevdv:
This PR is an extensive cleanup of `bokeh.sampledata`:

* package local data files moved to `sampledata/_data`
* factored much code to new `bokeh.util.sampledata`
* added test file for ever sample data module
* applied module file template
* fixed comment length issue across all boilerplate comments


bryevdv:
I am going to add more tests in the morning but this is otherwise ready.

bryevdv:
@mattpap I will plan to merge this in a few hours unless there is comment.

bryevdv:
Note: I added tests for all the sample data files to check basic type. I only added "detail" tests for the packaged data, not the external ones. Although the external data should not really ever change, I don't want to get in to a situation where if it did, some versions of bokeh tests could now no loner ever pass again. 

We could also consider adding an pytest mark for `sampledata` tests so that they are not run as part of the standard unit tests.

bryevdv:
I went ahead and added a `sampledata` mark, so these tests will not be include in `unit` tests. I have them running in the "codebase" stage.

bryevdv:
OK I also protected the sample data module imports so that e.g. `py.test -m unit` will always work whether sample data is installed or not.

PR #7025, opened 2017-10-09
call formatter for major categorical ticks

### LINKED ISSUE #7015: FuncTickFormatter broken with Categorical Axis
Bokeh v. 0.12.10dev2

In attempting to customize axis tick labels, there are two ways to achieve this (python function and javascript).
```python
from bokeh.models FuncTickFormatter
from bokeh.plotting import figure, output_notebook
output_notebook()
p = figure(plot_width=500, plot_height=500)
p.circle([0, 2, 4, 6, 8, 10], [6, 2, 4, 10, 8, 0], size=30)
p.yaxis.formatter = FuncTickFormatter(code="""
    return 'foo'
""")
def ticker():
    return "bar"
p.xaxis.formatter = FuncTickFormatter.from_py_func(ticker)
show(p)
```

However, attempting to do so on a categorical plot does not work (on the categorical axis):
```python
from bokeh.models import ColumnDataSource, FuncTickFormatter
from bokeh.palettes import Spectral5
from bokeh.transform import factor_cmap

df.cyl = df.cyl.astype(str)
group = df.groupby('cyl')

source = ColumnDataSource(group)

cyl_cmap = factor_cmap('cyl', palette=Spectral5, factors=sorted(df.cyl.unique()))

p = figure(plot_height=350, x_range=group, title="MPG by # Cylinders",
           toolbar_location=None, tools="")

p.vbar(x='cyl', top='mpg_mean', width=1, source=source,
       line_color=cyl_cmap, fill_color=cyl_cmap)

p.y_range.start = 0
p.xaxis.axis_label = "some stuff"
p.xaxis.major_label_orientation = 1.2
p.outline_line_color = None

# TICKER FUNCTIONS

def ticker():
    return "bar"
# CATEGORICAL TICKS SHOULD ALL BE  'bar' (BROKEN)
p.xaxis.formatter = FuncTickFormatter.from_py_func(ticker) # CATEGORICAL AXIS

# LINEAR TICKS ARE ALL 'foo' (WORKS)
p.yaxis.formatter = FuncTickFormatter(code="""
    return 'foo'
""")

show(p)
```

bryevdv:
- [x] issues: fixes #7015
- [ ] tests added / passed

A small change applies formatting to the major ticks adjacent to to the axis:

screen shot 2017-10-09 at 13 59 14

However, not in the nested case, it does *not* apply formatting to the higher level:

screen shot 2017-10-09 at 13 58 56

@hiramf @hafen is this OK for now? Another change *can* produce this:

screen shot 2017-10-09 at 13 58 39

but it's not ideal &mdash; the only info available to the (one single) formatter is the local tick value, not what level it is for, or a complete compound category value. I think this is not ideal so I don't want to push it now. It might be better to have separate formatters for mid or top level ticks, or possibly there could be a way to pass more info to the formatter, so that a single formatter can intelligently switch on the category level. 

If this seems OK for the immediate issue I will figure out some tests and merge.

### DIFF
--- a/bokehjs/src/coffee/models/axes/categorical_axis.coffee
+++ b/bokehjs/src/coffee/models/axes/categorical_axis.coffee
@@ -83,15 +83,16 @@ export class CategoricalAxisView extends AxisView
     info = []
 
     if range.levels == 1
-      info.push([ ticks.major, coords.major, @model.major_label_orientation, @visuals.major_label_text ])
+      labels = @model.formatter.doFormat(ticks.major, @)
+      info.push([ labels, coords.major, @model.major_label_orientation, @visuals.major_label_text ])
 
     else if range.levels == 2
-      labels = (x[1] for x in ticks.major)
+      labels = @model.formatter.doFormat((x[1] for x in ticks.major), @)
       info.push([ labels,     coords.major, @model.major_label_orientation, @visuals.major_label_text ])
       info.push([ ticks.tops, coords.tops,  'parallel',                     @visuals.group_text       ])
 
     else if range.levels == 3
-      labels = (x[2] for x in ticks.major)
+      labels = @model.formatter.doFormat((x[2] for x in ticks.major), @)
       mid_labels = (x[1] for x in ticks.mids)
       info.push([ labels,     coords.major, @model.major_label_orientation, @visuals.major_label_text ])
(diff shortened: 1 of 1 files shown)


bryevdv:
OK well I will merge this for now, and when/if there is a call for more capability we can figure out the best approach then.

PR #7046, opened 2017-10-12
ensure Document modules are properly disposed of

### LINKED ISSUE #7044: bokeh server sessions not released correctly
I am running the following on python 2.7, bokeh 0.12.9. 

```
from bokeh.io import curdoc

# eating some memory
GB = 1024*1024*1024

# take 500MB
largevar = "a" * int(0.5*GB)

# minimal setup
p = figure()
l = layout([p])
curdoc().add_root(l)
curdoc().title = 'minimal'
```
When I open multiple tabs of this app, each claim 500MB. After closing the tabs the memory consumption remains the same. I am not sure what the timeout is but I have waited at least 10 minutes without any change.

bryevdv:
- [x] issues: fixes #7044
- [x] tests added / passed

This PR adds a detailed check in `Document.delete_modules` that saved modules only have the expected set of referrers. The method `delete_modules` is made responsible for clearing the internal list of modules so that they may be correctly reaped. Additionally some extra logging has been added to trace the session destruction codepath.

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -299,7 +299,7 @@ from os import getpid
 from ..subcommand import Subcommand
 from ..util import build_single_handler_applications, die, report_server_init_errors
 
-LOGLEVELS = ('debug', 'info', 'warning', 'error', 'critical')
+LOGLEVELS = ('trace', 'debug', 'info', 'warning', 'error', 'critical')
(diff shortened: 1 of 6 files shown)


bryevdv:
@mattpap I'd appreciate if you can give this a quick look, ASAP i would like to get it in a dev build this afternoon.

PR #7047, opened 2017-10-12
Properly maintain active tools when proxying toolbars

### LINKED ISSUE #6921: Shared drag tools in grid plots only work on the last plot
Panning/Box Zoom on any of the first plots is ignored.

mattpap:
This makes shared tools initialize properly, but this doesn't fix the overall mess that tool/toolbar initialization is. I will address the broader issue after PR #7008.

fixes #6921


bryevdv:
OK tested this locally on several examples and thing are working so merging.

PR #7049, opened 2017-10-12
sort the results of _mask_data so patch draw order is stable

### LINKED ISSUE #3601: PatchesView._mask_data() changes the draw order.
I think `patches()` should draw the patches in the order of the data.But it seems that `PatchesView._mask_data()` will changes the draw order randomly. Maybe it should return a sorted array .

Here is an example that demonstrate the problem:

https://ruoyu0088.gitbooks.io/book/content/evolution.html

if you slide the slider, it will add or remove triangles, finally it will become a firefox logo. But because the draw order is random, the triangles blink randomly.

run `fix()` that disable `_mask_data()` in console fix the problem.

bryevdv:
issues: fixes #3601

I tested this with existing choropleth examples and it seems fine. 

@philippjfr I can't get HV to work at all right now, and there is no MRE in the issue, so someone else needs to confirm that it functions as desired with overlapping patches.

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/patches.coffee
+++ b/bokehjs/src/coffee/models/glyphs/patches.coffee
@@ -71,7 +71,10 @@ export class PatchesView extends GlyphView
     [y0, y1] = [yr.min, yr.max]
 
     bbox = hittest.validate_bbox_coords([x0, x1], [y0, y1])
-    return @index.indices(bbox)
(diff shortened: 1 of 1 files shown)


philippjfr:
@bryevdv It's working perfectly for me, matches matplotlib output exactly.

bryevdv:
Great, merging now.

Issue #4965, opened 2016-08-16
DatePicker errors on input from Chrome on Windows 8.1


rheineke:
Using Chrome 52 on Windows 8.1, I find that DatePicker throws an error on any value chosen from the dropdown calendar. For same Chrome version on Mac OS X, I find that it always works.

I am using bokeh 0.12.0-1.

I have a bokeh server with the following code in app/main.py:

```
from bokeh.io import curdoc
from bokeh.models import DatePicker

curdoc().add_root(DatePicker())
```

and run with the command

`bokeh serve --log-level debug app`

The exception is:

```
2016-08-16 15:16:51,365 error handling message Message 'PATCH-DOC' (revision 1): ValueError('Unknown string format',)
2016-08-16 15:16:51,365   message header {'msgid': '8B5BE6EBF7044F8793268769745D6C7A', 'msgtype': 'PATCH-DOC'} content {'events': [{'new': 'Mon Aug 01 2016 00:00:00 GMT-0400 (Eastern Daylight Time)', 'model': {'type': 'DatePicker', 'id': '5f8ecfc4-a06e-428f-b21d-6ed1a34cefff'}, 'attr': 'value', 'kind': 'ModelChanged'}], 'references': []}
Traceback (most recent call last):
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/bokeh/server/protocol/server_handler.py", line 38, in handle
    work = yield handler(message, connection)
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/tornado/gen.py", line 1008, in run
    value = future.result()
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/tornado/concurrent.py", line 232, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/tornado/gen.py", line 1017, in run
    yielded = self.gen.send(value)
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/bokeh/server/session.py", line 45, in _needs_document_lock_wrapper
    result = yield yield_for_all_futures(func(self, *args, **kwargs))
  File "/Users/rheineke/anaconda/lib/python3.5/site-packages/bokeh/server/session.py", line 217, in _handle_patch
    message.apply_to_document(self.document)
[...]

PR #7050, opened 2017-10-13
fixes: Datepicker returns one day earlier than picked in IE (#7048)

### LINKED ISSUE #7048: Datepicker returns one day earlier than picked in IE
Using datepicker in FF and IE revelas. The returned date from IE is one day earlier than selected.
Example. Selected is 23.09.2017T00:00:00GMT+2: -> FF 23.09.2017, IE 22.09.2017.
The date.ToISOString() functions behaves differently on FF and IE. In FF only the date part is used -> "22.09.2017". IE takes into account the timezone (in this case Europe/Berlin GMT+2) and shifts the time to UTC before calculating the ISO-String. So 23.09.2017:00:00:00GMT+2 becomes 22.09.2017 22:00:00UTC -> "22.09.2017".
Fix will be to change : 
```
  _on_select: (date) =>
    @model.value = date.toISOString()
    @change_input()
```
to  :
```
  _on_select: (date) =>
    @model.value = date.getFullYear().toString()+'-'+date.getMonth().toString()+'-'+date.getDate().toString() 
    @change_input()
```
If this is OK for everybody then I will prepare a PR

volkerjaenisch:
- [ ] issues: fixes #7048
Replaces my former PR for #4965.

### DIFF
--- a/bokehjs/src/coffee/models/widgets/date_picker.coffee
+++ b/bokehjs/src/coffee/models/widgets/date_picker.coffee
@@ -70,7 +70,10 @@ export class DatePickerView extends InputWidgetView
     return @
 
   _on_select: (date) =>
-    @model.value = date.toString()
+    # Always use toDateString()!
+    # toString() breaks the websocket #4965.
+    # toISOString() returns the wrong day (IE on day earlier) #7048
+    @model.value = date.toDateString()
     @change_input()
 
 export class DatePicker extends InputWidget


bryevdv:
@volkerjaenisch There is some issue with this branch, can you rebase on master and resolve the conflict?

bryevdv:
@mattpap if you concur I'd like to merge this then cut RC1

PR #7068, opened 2017-10-18
slice indices must be non-negative, not positive

### LINKED ISSUE #6545: Patch bug: patch slices must have positive (start, stop, step) values, got slice(0, 200, 50)
In [.../0.12.6/examples/howto/patch_app.py](https://github.com/bokeh/bokeh/blob/0.12.6/examples/howto/patch_app.py) at line 59, changing `slice(50, 151, 20)` to `slice(0, 200, 50)`generates the following error: `ValueError('Patch slices must have positive (start, stop, step) values, got slice(0, 200, 50)',)`. 

As far as I know, `0` is supposed to be a valid slice index. Only negative values are prohibited.

bryevdv:
- [x] issues: fixes #6545
- [x] tests added / passed

@nleclercq can you test this PR to confirm that it matches your expectations?

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -589,10 +589,10 @@ class ColumnDataSource(ColumnarDataSource):
 def _check_slice(s):
     if (s.start is not None and s.stop is not None and s.start > s.stop):
         raise ValueError("Patch slices must have start < end, got %s" % s)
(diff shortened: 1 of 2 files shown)


bryevdv:
Have not heard back from OP, but this should be uncontroversial, I will plan to merge at the end of the day

PR #7073, opened 2017-10-19
compute 'nearest' to span only in one dimension

### LINKED ISSUE #6174: HoverTool vline/hline models for Line glyph (_hit_span method) causes non-intuitive tooltips to appear
HoverTool vline/hline models for Line glyph (_hit_span method) causes non-intuitive tooltips to appear. I would expect the same tooltip behavior as the commented-out scatter:

If I hold the cursor at (3.9, 3.5)

screen shot 2017-04-20 at 1 13 19 pm

Otherwise, If I hold the cursor at (4.1, 3.5)

screen shot 2017-04-20 at 1 13 15 pm

At no point along y=3.5 will the top point be inspected. The user has to move the mouse up to y=4. Is this the intended behavior?

Code to reproduce.
```python

p = figure(plot_width=400, plot_height=400, tools='')
# p.scatter(x=[1,2,3,4,5], y=[4,2,4,5,3])
p.line(x=[1,2,3,4,5], y=[4,2,4,5,3])

h = HoverTool(mode='vline')
p.add_tools(h)
show(p)
```

Bokeh 0.12.6dev2
Chrome 58
Python 3.5

bryevdv:
issues: fixes #6174

When computing the nearest point for determining where to position a tooltip, when the hit geometry is a span, only compute the distance along the transverse dimension.

This PR also:

* deletes a duplicative "color to hex" function in favor of one in utils
* factors out some common code

In general this is all too complicated and untested. A larger refactor should be considered, perhaps in conjunction with selection refactor work, or immediately after.

### DIFF
--- a/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
+++ b/bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
@@ -6,21 +6,41 @@ import * as hittest from "core/hittest"
 import {replace_placeholders} from "core/util/templating"
 import {div, span} from "core/dom"
 import * as p from "core/properties"
+import {color2hex} from "core/util/color"
 import {values, isEmpty} from "core/util/object"
 import {isString, isFunction} from "core/util/types"
 import {build_views, remove_views} from "core/build_views"
 
-_color_to_hex = (color) ->
-  if (color.substr(0, 1) == '#')
-      return color
-  digits = /(.*?)rgb\((\d+), (\d+), (\d+)\)/.exec(color)
+export _nearest_line_hit = (canvas, i, geometry, sx, sy, dx, dy) ->
+  d1x = dx[i]
+  d1y = dy[i]
 
-  red = parseInt(digits[2])
-  green = parseInt(digits[3])
-  blue = parseInt(digits[4])
+  d2x = dx[i+1]
+  d2y = dy[i+1]
 
-  rgb = blue | (green << 8) | (red << 16)
-  return digits[1] + '#' + rgb.toString(16)
+  if geometry.type == "span"
+    switch geometry.direction
+      when "h"
+        dist1 = Math.abs(d1x-sx)
+        dist2 = Math.abs(d2x-sx)
+      when "v"
+        dist1 = Math.abs(d1y-sy)
+        dist2 = Math.abs(d2y-sy)
+  else
+    dist1 = hittest.dist_2_pts(d1x, d1y, sx, sy)
+    dist2 = hittest.dist_2_pts(d2x, d2y, sx, sy)
+
+  if dist1 < dist2
+    rx = canvas.sx_to_vx(d1x)
+    ry = canvas.sy_to_vy(d1y)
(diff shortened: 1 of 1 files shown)


bokehjs/src/coffee/models/tools/inspectors/hover_tool.coffee
```diff
@@ -175,112 +195,77 @@ export class HoverToolView extends InspectToolView
...
-            [sdatax, sdatay] = [d2x, d2y]
-            ii = i+1
-
-          data_x = renderer_view.glyph._x[i]
-          data_y = renderer_view.glyph._y[i]
-          rx = canvas.sx_to_vx(sdatax)
-          ry = canvas.sy_to_vy(sdatay)
+          [[rx, ry], ii] = _nearest_line_hit(canvas, i, geometry, sx, sy, glyph.sx, glyph.sy)
+          data_x = glyph._x[i]
+          data_y = glyph._y[i]
```
clairetang6: These should use the returned `ii` for indexing, so that the data x and data y are correct.

bryevdv: Thanks for the catch @clairetang6 I will make the change and merge

bryevdv: Should this similarly be `jj`? 
```
          when "nearest"
              [[rx, ry], jj] = _nearest_line_hit(canvas, j, geometry, sx, sy, glyph.sxs[i], glyph.sys[i])
              data_x = glyph._xs[i][j]
              data_y = glyph._ys[i][j]
```

clairetang6: Yeah, that should also be `jj`. 

Also, from testing with multiline, I found that there is an extra loop in the code for the 2d selections. The outer loop is through the 1d indices, then for multilines, it loops again through the 2d indices, but shouldn't. The line `for i, [j] of indices['2d'].indices` could be changed to `j = indices['2d'].indices[i]` with the `i` being defined by the outer loop already. Or the loop for the multiline could be pulled out of the loop for the 1d indices.

bryevdv:
@clairetang6 thatnks for the catch, I have made the requested changes

clairetang6:
Oops actually I think the line should be j = indices['2d'].indices[i][0].

bryevdv:
@clairetang6 actually I think the intent of the intent of the second loop was to actually loop over the sublists, so I don't think `[i][0]` seems correct.

clairetang6:
@bryevdv, right, I see. There could be the situation where you hit multiple points of one line in a multiline. The `[i][0]` is just how the current loop works, giving you a tooltip for the first point hit, but you get multiple of the same tooltip if multiple lines were hit currently. Whether or not you decide to change the loop for now, I'll keep this in mind while working on selections.

bryevdv:
@clairetang6 I have restored the old loop, given how bad the test coverage is over this tool (my fault) I think its best to keep the changes minimal

PR #7086, opened 2017-10-22
factor rect hit-testing against an index out to glyph view

### LINKED ISSUE #5298: box select does not work with vbar but tap tool does
The following code illustrates the issue when run with bokeh serve

```

import logging
logger = logging.getLogger('test')

def on_selection_change(attr, old, new):
    logger.info('selection change')

source = ColumnDataSource(data=dict(bins=[0, 1, 2], counts=[1, 10, 20]))
doc = curdoc()

p = figure(tools="box_select,tap")
source.on_change('selected', on_selection_change)

b = p.vbar(x = 'bins', bottom=0, width=1, top='counts' , source=source)
# c = p.circle(x='bins', y='counts', source=source)

doc.add_root(p)
```

Using vbar tap tool fires selection change but box select does not.
Using circle both tools fire selection change.

Python version      :  3.5.1 |Continuum Analytics, Inc.| (default, Feb 16 2016, 09:49:46) [MSC v.1900 64 bit (AMD64)]
IPython version     :  5.1.0
Bokeh version       :  0.12.2

bryevdv:
- [x] issues: fixes #5298
- [x] tests added / passed

In the interest of remaining consistent with `rect` behaviour, and also re-using code, I have factored the rect-rect index hit testing up to glyph to also be used by `Box` subclasses. Will add test before merging.

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/glyph.coffee
+++ b/bokehjs/src/coffee/models/glyphs/glyph.coffee
@@ -175,9 +175,9 @@ export class GlyphView extends View
   _hit_rect_against_index: (geometry) ->
     [x0, x1] = @renderer.xscale.v_invert([geometry.vx0, geometry.vx1])
     [y0, y1] = @renderer.yscale.v_invert([geometry.vy0, geometry.vy1])
-    bbox = hittest.validate_bbox_coords([x0, x1], [y0, y1])
+    bb = hittest.validate_bbox_coords([x0, x1], [y0, y1])
     result = hittest.create_hit_test_result()
-    result['1d'].indices = @index.indices(bbox)
+    result['1d'].indices = @index.indices(bb)
     return result
 
   set_data: (source, indices, indices_to_update) ->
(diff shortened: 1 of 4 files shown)


bryevdv:
Not a great test but better than nothing

bokehjs/src/coffee/models/glyphs/glyph.coffee
```diff
@@ -171,6 +172,14 @@ export class GlyphView extends View
 
     return result
 
+  _hit_rect_against_index: (geometry) ->
+    [x0, x1] = @renderer.xscale.v_invert([geometry.vx0, geometry.vx1])
+    [y0, y1] = @renderer.yscale.v_invert([geometry.vy0, geometry.vy1])
+    bb = hittest.validate_bbox_coords([x0, x1], [y0, y1])
```
bryevdv: Without renaming this to `bb` (instead of `bbox`) another test fails with `bbox.empty` is not a function in `log_bounds` I don't understand why that should be the case, though.

mattpap: This is an effect of coffeescript's scoping rules. Don't reuse global names inside functions. Here https://github.com/bokeh/bokeh/pull/7008/commits/e865559a31ee7c3b132e7a235e5ecf3e8349b579 you have another example.

bryevdv: Well, that's garbage

PR #7099, opened 2017-10-24
compute rect dists more carefully

### LINKED ISSUE #6583: rect not behaving correctly with zero height
Python 2.7.12
Bokeh 0.12.6

```
from bokeh.plotting import figure, show, output_file, 
output_file('test.html')

a = figure(plot_width=400, plot_height=400)
a.rect(range(4),[0.0, -5.0, -4.0, 5.0],
        [0.5]*4,[0.0, 10.0, 8.0, 10.0])
show(a)
```

I would expect this piece of code to plot 3 rectangles which y coordinates would be [0,-10], [0,-8],[0,+10]. 
At the place I am getting 3 rectangles with the following y coordinates : [-10,-20],[-8,-16],[0,-10]

n.b.

it works perfectly fine if I remove the coordinate of the empty rectangle at the beginning ...

```
a = figure(plot_width=400, plot_height=400)
a.rect(range(3),[-5.0, -4.0, 5.0],
        [0.5]*3,[10.0, 8.0, 10.0])
show(a)

```

bryevdv:
- [x] issues: fixes #6583
- [x] tests added/passed

@clairetang6 this is your suggested function only modified to always set `vpt_corner` to something. If you have suggestions for additional/better tests please let me know (or just add them). TBH I am still not 100% clear on `sx0` and `sy1` but digging into the details will have to be another day.

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/rect.coffee
+++ b/bokehjs/src/coffee/models/glyphs/rect.coffee
@@ -137,10 +137,18 @@ export class RectView extends XYGlyphView
     vpt1 = scale.v_compute(pt1)
     sside_length = @sdist(scale, pt0, side_length, 'edge', @model.dilate)
     if dim == 0
-      vpt_corner = if vpt0[0] < vpt1[0] then vpt0 else vpt1
+      vpt_corner = vpt0
+      for i in [0...vpt0.length]
+        if vpt0[i] != vpt1[i]
+          vpt_corner = if vpt0[i] < vpt1[i] then vpt0 else vpt1
+          break
       return [sside_length, canvas.v_vx_to_sx(vpt_corner)]
     else if dim == 1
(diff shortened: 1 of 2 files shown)


bryevdv:
@clairetang6 if you'd like time to look at this first, please just comment and I will wait to merge this. Otherwise I will plan to merge this in an hour or so

clairetang6:
@bryevdv LGTM. 

I would need to dig in a little bit to remember exactly which corners of the rect vpt0 and vpt1 represent, but the idea is just that we need the screen coordinates of the upper left corner to give to the canvas context rect function, and the upper left would be smallest x value (x0) and the largest y value (y1) (Just looking up the canvas rect function let me reason out the corner logic). Then because the scales can be reverse scales, you need to check to make sure that the smaller of vpt0 and vpt1 is the vptcorner for the x-dimension and vice versa for the y-dim.

PR #7105, opened 2017-10-26
Fixes issue #7103 by updating documentation on BokehJS standalone installation guidelines

### LINKED ISSUE #7103: Update BokehJS standalone installation guidelines
As follow-up to this [question](https://stackoverflow.com/questions/46935288/bokehjs-basic-example-error-cannot-read-property-linspace-of-undefined) on stackoverflow,
I would like to suggest an update of section [https://bokeh.pydata.org/en/latest/docs/installation.html#bokehjs-standalone](https://bokeh.pydata.org/en/latest/docs/installation.html#bokehjs-standalone)
to include [http://cdn.pydata.org/bokeh/release/bokeh-api-x.y.z.min.js](http://cdn.pydata.org/bokeh/release/bokeh-api-x.y.z.min.js)

Also, it would be nice to first list css files and then js, just in case the new user is in a hurry and does not notice that it is important that css should be listed before js in his html.

In any case, congratulations for all the effort put into Bokeh, this is a most needed library and I'm looking forward to the day v1.0 is released!

Cheers,
A

mattbest:

- [ ] issues: fixes #7103 
- [ ] tests added / passed

### DIFF
--- a/sphinx/source/docs/installation.rst
+++ b/sphinx/source/docs/installation.rst
@@ -154,17 +154,18 @@ two easy ways to get any published release.
 First, released versions of BokehJS is available for download from CDN at
 pydata.org, under the following naming scheme::
 
-    http://cdn.pydata.org/bokeh/release/bokeh-x.y.z.min.js
-    http://cdn.pydata.org/bokeh/release/bokeh-widgets-x.y.z.min.js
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @mattbest

PR #7107, opened 2017-10-26
7091 how to keep state in bokeh server across more than one worker

### LINKED ISSUE #7091: How to keep state in bokeh server across more than one worker
Use Case: A bokeh view needs to know the IP the client requests (HTTP/WS) are coming from to make decisions.
Our View class uses the following code to determine if an incoming IP (as HTTP-Param "remote_ip"  in the request to autoload.js) is in our IP database:
```
class ProfileView(LayoutDOM):
...
    def __init__(self):
         self.set_is_privileged()

    def set_is_privileged(self):
        ip = None
        doc = curdoc()
        if doc.session_context:
            args = doc.session_context.request.arguments

            try:
                ip = args.get('remote_ip')[0]
                logger.info('%s: comes with IP %s' % (self._id, ip))
            except :
                logger.info('Exception: %s' % sys.exc_info()[0])
                logger.info('traceback: %s' % tb.format_tb(sys.exc_info()[2]))
                logger.info('%s: No ip set' % self._id)
                raise( Exception('No ip set'))
        if ip:
            ip_address = ipaddress.ip_address(ip.decode('utf8'))
            for network_entry in self._provider.privileged_ips():
                network = ipaddress.ip_network(network_entry.ip)
                if ip_address in network:
                    self._is_privileged = True
                    logger.info('%s: User is privileged.' % self._id)
                    return
        logger.info('%s: User NOT privileged.' % self._id)
        self._is_privileged = False
```
Herby self._is_privileged is the state that has to be assured over the whole session the client is connected and using the application (an interactive map and a plot).

Our network setup is not simple. We have a plone instance acting as the front portal and a Bokeh instance acting as the backend portal. 
Plone Portal:
```
server {

    location / {
        proxy_pass
        http://127.0.0.1:8083/VirtualHostBase/https/e-profile.inqbus.de:443/Plone/VirtualHostRoot/;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_http_version 1.1;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_buffering off;
    }
}
```

Bokeh portal:
```
server {
    location / {
        proxy_pass http://127.0.0.1:5006;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_http_version 1.1;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_buffering off;
        }
}
```
The code works as well as the nginx config. 
```
2017-10-24 00:27:18,544 INFO     3a8734ad-506b-4ffa-8ef3-991e4bd6ad46: comes with IP b'95.91.249.0'
2017-10-24 00:27:18,547 INFO     3a8734ad-506b-4ffa-8ef3-991e4bd6ad46: User is privileged.
2017-10-24 00:27:18,561 INFO     Send JS call: map.set_all_marker
2017-10-24 00:27:33,248 INFO     Send JS call: map.set_all_marker
2017-10-24 00:27:33,309 INFO     date changed to 2017-10-24
2017-10-24 00:27:33,310 INFO     Send JS call: this.aod_layer_set_date(2017-10-24)
2017-10-24 00:27:33,313 INFO     Send JS call: map.set_all_icons
2017-10-24 00:27:33,374 INFO     Send JS call: map.set_all_icons
2017-10-24 00:27:33,377 INFO     Map zoomed to level 5
```

But only if we do not use more than one bokeh worker process. And here begins the mystery.

```
2017-10-24 00:23:47,189 INFO     71378899-4b02-4a6c-adbd-ad32730f5b9e: comes with IP b'95.91.249.0'
2017-10-24 00:23:47,192 INFO     71378899-4b02-4a6c-adbd-ad32730f5b9e: User is privileged.
2017-10-24 00:23:47,205 INFO     Send JS call: map.set_all_marker
2017-10-24 00:24:01,762 INFO     traceback: ['  File "/home/graphdemoplone/workspace/eprofile/eprofile/src/eprofile/bokeh/profile/profile_layout.py", line 133, in set_is_privileged\n    ip = args.get(bcfg.IP_PARAMETER)[0]\n']247-a8e4-ae3695225d42: No ip set
2017-10-24 00:24:01,765 INFO     57c100ec-c43b-4247-a8e4-ae3695225d42: User NOT privileged.
2017-10-24 00:24:01,832 INFO     Send JS call: map.set_all_marker
2017-10-24 00:24:03,278 INFO     Send JS call: map.set_all_marker
2017-10-24 00:24:03,342 INFO     date changed to 2017-10-24
2017-10-24 00:24:03,342 INFO     Send JS call: this.aod_layer_set_date(2017-10-24)
2017-10-24 00:24:03,345 INFO     Send JS call: map.set_all_icons
2017-10-24 00:24:03,404 INFO     Send JS call: map.set_all_icons
2017-10-24 00:24:03,407 INFO     Map zoomed to level 5
```

In the case where we use more than one worker we have not repeatable issues with a second  session (self._id of the view). This means that our view class is instantiated more than once in a "user session".
This behaviour is repeatable but complex. Today I had a case where of four computers with same IP (same flat) three were accepted and the fourth not, persistently. Restarting the bokeh-Process does not change anything. Restarting nginx does not help either. As I said it is complex with to many degrees of freedom to track easily. 

Since we have no loadbalancing on the nginx (yet), all the traffic lands at the tornado instance. We have not much knowledge how tornado works. Also we have no idea what Bokeh does with sessions.

Probably we are doing something stupid and someone can point us simply to the right solution.

From a developer perspective it would be very useful to have some informations how bokeh sessions, tornado and the bokeh views interact. 

To sum up. We assume for our code that the following statements hold:
1) In a Bokeh view class e.g. derived from LayoutDOM a property like self._is_privileged keeps its state during the whole time a user is connected.
2) The state of self._is_privileged is shared across the tornado workers.

I am quite sure that 1 hold but 2 not.

Please shed some light on this topic.

sandrarum:
PROTOTYPE for adding session storage

- [ ] issues: fixes #7091 
- [ ] tests added / passed


bryevdv:
@sandrarum Thanks for sharing this prototype. I think some discussion/iteration will be needed. The changes as they are now are somewhat more invasive that I would think would be necessary. What I had imagined was a base class for encapsulating the responsibility for persisting session state, that users could subclass. I think it needs a more descriptive name than `Session` since that is already used:
```
class PersistentSessionState(object):

    def save(self, session_id, request):
       # subclasses must implement

    def restore(self, session_id):
        # subclasses must implement

class NullSessionState(PersistentSessionState)
    ''' This is the default '''
    
    def save(self, session_id, request): 
        pass

    def restore(self, session_id): 
        pass

class RedisSessionState(PersistentSessionState)
    ''' User might create this '''
    
    def save(self, session_id, request):
         # save whatever to redis

    def restore(self, session_id): 
        # load whatever to redis 

```

Then in the HTTP views, something like this would happen:
```
    self.state.save(session.id, request)
```

and in the WS handler something like this would happen:
```
    # need to figure out what to do with stuff? attach to Session or SessionContext maybe?
    stuff = self.state.load(session_id)
```
The "state persister" could probably be an optional argument to `Server` so that if you are running the server programmatically it is easy to provide your own custom `PersistentSessionState` subclass. I believe this would keep things nicely decoupled, with very few changes required to existing code.

bryevdv:
I guess to put thing another way: The code in this PR seems to be geared towards persisting actual sessions, themselves. This is not the approach I would favor. I would rather keep the existing session machinery as-is (it is well tested and understood) but provide a small encapsulated, extensible, policy-agnostic mechanism for saving and attaching arbitrary state to sessions between HTTP and WS connection times.

bryevdv:
@sandrarum do you expect or intend to do further work on this? If not we can close and I will look into the other approach I described when I am able.

bryevdv:
Absent response I am going to go ahead and close this.

PR #7108, opened 2017-10-26
Switch layout to use screen coordinates

### LINKED ISSUE #6447: Switch layout's coordinate system from view to screen

mattpap:
~~Very early work-in-progress.~~

fixes #6447


bryevdv:
I know this is `WIP` but I'd appreciate some comment about the intended direction. By "switch to screen coordinates" I would assume all the functions like `v_vy_to_sy` (and all usage of "vx" type coordinates) would go away entirely.

mattpap:
The main goals are to move canvas layout's origin from (left, bottom) to (left, top) and align it with DOM layout (although not fully yet), and then to make all positioning computations done in screen coordinate system (absolute or panel/frame-relative). This should remove the bulk of `v[xy]_to_s[xy]`. However, not all, because view coordinates (absolute and relative) were introduced to annotations' APIs, by allowing things like `x_units = "screen"`.  I'm not sure what to do about this. I would like to just fix it and make sure that there is exactly one understating of screen coordinates. This would be a breaking change, but we use wrong `screen` units only in annotations and mostly internally in tools' overlays. In examples and docs we only use `[xy]_units` with `Label`.

bokehjs/src/coffee/models/scales/scale.coffee
```diff
@@ -2,22 +2,29 @@ import {Transform} from "../transforms"
...
+    if @target_range.is_reversed
+      return [@compute(x1), @compute(x0)]
+    else
+      return [@compute(x0), @compute(x1)]
+
+  r_invert: (sx0, sx1) ->
+    if @target_range.is_reversed
+      return [@invert(sx1), @invert(sx0)]
+    else
+      return [@invert(sx0), @invert(sx1)]
```
bryevdv: I don't really like these functions. It's potentially very confusing to have the order of the parameters coupled implicitly in a hidden way to a property of some other model. It also muddies the conceptual clarity and simplicity of "compute" which really should just be "compute the transform of this input". The mapping of `x0` is what it is, it should not be conflated with a mapping for a completely different value in some implicit way.

bryevdv: I also don't understand why this would be needed? If this is just to try and keep the "order" or two arbitrary inputs in the same order as a range, we already have "bbox" function to make sure start < end, etc. that are used where appropriate.

mattpap: I think it would be very confusing if I implemented `v_invert/v_compute` like this. This is why I introduced `r_` variant for pairs of coordinates. `r_` variant simply maintains spatial order. If you don't want this, then you use `v_` variant or multiple calls to `invert/compute`. I tried multiple solutions so far, and this one is the cleanest and allows for very straightforward and explicit maintenance of transition between data and screen space ranges. Previously there were multiple ways to achieve spatial order, now there is one. No implicit conversions are necessary, no variable swapping, no reversed inequalities. I presume this change may make scales less pure, but it makes everything else much easier to work with.

bryevdv: I was definitely not suggesting `v_invert/v_compute` be implemented this way. I was suggesting not having this at all, and making callers explicitly responsible for caring (or not) about order. 

Because this does not prevent misuse. What if I call `r_compute(x1, x0)` then then values get swapped when they shouldn't be. And it doesn't help at all if I need to convert more than two values for some reason. Isn't the actual intent of this function simply that the  results be sorted in the same order as the target range? If so, I'd propose adding a `sort_by_target` flag to `v_compute` and does nothing if false (current behavior), or sorts ascending or descending based on `is_reversed` if true. That has semantics that don't force an artificial start/end concept on input data that does not have it, works correctly regardless of what order args are passed in, and has semantics that make more sense (to me).

bryevdv: Alternatively `sorted_compute`, `v_sorted_compute` but personally I'd prefer the flag

bryevdv: I guess my takeaway is: actually do a real sort instead of the fake-sort-by-swapping. Because the "swapping" only works coincidentally, under a specific subset of circumstances, whereas doing a real sort will always work and also make the semantics consistent and sensible.

mattpap: >  but personally I'd prefer the flag

Abstracting from this PR, adding a flag is almost always a very bad idea, and a terrible one in JS, where you don't have named arguments. The two variations must be easily distinguishable and searchable, so there must be different names. Internally one can be implemented in terms of the other.

bryevdv: I agree JS makes it worse, so as I say I'd be fine with `sorted_compute, v_sorted_compute` then

mattpap: There is no point in having `v_` sorted variant, because the only usage of the existing `r_` variant is for pairs of coordinates. That's why this function takes two scalar arguments and returns a 2-tuple. Previously this was emulated with `v_compute([arg0, arg1])`, but that's not semantically equivalent, and it will be better visible when I add typings for scales.

mattpap: > I guess my takeaway is: actually do a real sort instead of the fake-sort-by-swapping.

I disagree that what's needed here is sorting. The goal of `r_invert` is to maintain the order of output coordinates the same as the order of inputs. However, you may be right, but I need to give this some more thought and look at this from a fresh perspective.

bryevdv: The output of this function is highly dependent on the order of the arguments, but there is indication in the function name that this is the case, or further, what that order needs to be. If I have two parameters `xfoo` and `xbar` there is nothing to tell me whether to call `func(xfoo, xbar)` or to call `func(xbar, xfoo)`. The only way I can be sure to call things correctly is to know to know something about the function (that it might swap the order) and to then put the args in the right order by inspecting them and taking that into account. If you ask me to say what this function *does* I can't say, swapping unconditionally based on the range is nonsensical in many situations that seem easy to be in.

bryevdv:
custom label overlaps title:

https://s3.amazonaws.com/bokeh-travis/298542324/examples.html#models/file/external_resources

Perhaps the example needs an update?

bryevdv:
Any idea what's going on with the band annotation? 

https://s3.amazonaws.com/bokeh-travis/298542324/examples.html#integration/annotations/test_band

Their does not appear to be any diff of the frame outline so all the changes must be internal positioning?

bryevdv:
Maybe unrelated but what's going on with the number formatting in the displacement column here??

https://s3.amazonaws.com/bokeh-travis/298542324/examples.html#models/server/data_tables

mattpap:
`external_resources` and all `annotations` are all instances of the same issue. Previously some annotations used canvas-relative positioning and some panel/frame relative. Now it's uniformly the later. Thus differences. I didn't update `annotations` because they use "whatever" location on frame. I tried to update `external_resources`, but that example attempts to position a label relative to left-top corner, and this is currently not possible, at least without fiddling with pixels, which still can be wrong.

mattpap:
> Maybe unrelated but what's going on with the number formatting in the displacement column here??

Unrelated and pops up from time to time (I think since upgrading slickgrid).

mattpap:
I plan to fix the problem with label positioning later on (add anchor points as legend has).

bryevdv:
> Previously some annotations used canvas-relative positioning and some panel/frame relative. 

Just to be clear, the *user* code specified (i.e. the code in the example) specifies to use screen coords? If so that's probably not so common and so OK if it changes.

mattpap:
> Just to be clear, the user code (i.e. the code in the example) specifies to use screen coords? If so that's probably not so common and so OK if it changes.

It still uses `view` coordinates, just panel/frame-relative. This is the last place where view coordinates are used. I would like to changes this, especially that this code uses `screen` units when it means something different, but I thought it will be too much in this PR.

bryevdv:
OK let's get this in a dev build

PR #7126, opened 2017-10-30
Add JS script to correct element when showing plot

### LINKED ISSUE #7125: Only first plot in jupyter notebook cell is shown
*(Note that I have a proposed fix for this Issue. I'll open a PR.)*

#### Software version info
*Bokeh:* 0.12.10
*Python:* 3.6.1 (default, Apr  4 2017, 09:40:21) 
    [GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)]
*Notebook:* Jupyter notebook (5.2.0)
*OS:* MacOS 10.12.6
*Browser:* Google chrome version 61.0.3163.100 (Official Build) (64-bit)

#### Expected behavior and the observed behavior

Expected:
When multiple bokeh plots are defined and shown in a notebook code cell, I expect all plots to appear in the cell's output.

Observed:
Only the first plot is output. The subsequent plots are absent.

```
import bokeh.plotting
import bokeh.io

bokeh.io.output_notebook()

print("Plot 1:")
p1 = bokeh.plotting.figure(plot_height=200)
p1.line([1,2,3],[1,2,3])
bokeh.plotting.show(p1)

print("Plot 2:")
p2 = bokeh.plotting.figure(plot_height=200)
p2.line([1,2,3],[1,4,9])
bokeh.plotting.show(p2)
```
(Note that, the resulting .ipynb file contains all the necessary plot data and there are no JavaScript errors in the browser console.)

#### Screenshot
image

plazar:
This PR makes it possible to correctly display multiple bokeh plots in a single jupyter notebook cell.

The array index was hard-coded as 0, meaning only first plot was properly added to notebook cell. Now, add JS script to more recently added bokeh element, so that multiple plots in the same cell will be embedded.

_I'm not sure how to add a test for this fix. I'd appreciate any help to add the appropriate tests._

- [x] issues: fixes #7125 
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/_templates/autoload_nb_js.js
+++ b/bokeh/core/_templates/autoload_nb_js.js
@@ -63,7 +63,7 @@
     var toinsert = output_area.element.find(`.${CLASS_NAME.split(' ')[0]}`);
 
     if (output.metadata[EXEC_MIME_TYPE]["id"] !== undefined) {
-      toinsert[0].firstChild.textContent = output.data[JS_MIME_TYPE];
+      toinsert[toinsert.length - 1].firstChild.textContent = output.data[JS_MIME_TYPE];
       // store reference to embed id on output_area
       output_area._bokeh_element_id = output.metadata[EXEC_MIME_TYPE]["id"];
     }
--- a/bokeh/core/tests/test_templates.py
+++ b/bokeh/core/tests/test_templates.py
@@ -1,0 +1,23 @@
+import hashlib
+from os.path import abspath, join, split, pardir
+
+TOP_PATH = abspath(join(split(__file__)[0], pardir))
+
+
+def test_autoload_template_has_changed():
+    """This is not really a test but a reminder that if you change the 
+    autoload_nb_js.js template then you should make sure that insertion of
+    plots into notebooks is working as expected. In particular, this test was
+    created as part of https://github.com/bokeh/bokeh/issues/7125.
+    """
+    with open(join(TOP_PATH, '_templates/autoload_nb_js.js'), mode='rb') as f:
+        # compute a hash of the template and compare it against the
+        # last known hash
+        hash = hashlib.sha224(f.read()).hexdigest()
+        assert  hash == \
+        'dda19c5cb89b1ae6f10cdc24b052fa53738a1c31e654e5232e60ea59', \
+        """It seems that the template autoload_nb_js.js has changed. 
+        If this is voluntary and that proper testing of plots insertion 
+        in notebooks has been completed successfully, update this test
+         with this new file hash {}""".format(hash)
+
--- a/bokeh/core/tests/test_templates.py
+++ b/bokeh/core/tests/test_templates.py
@@ -1,22 +1,204 @@
-import hashlib
 from os.path import abspath, join, split, pardir
+
 
 TOP_PATH = abspath(join(split(__file__)[0], pardir))
 
+pinned_template = """{% extends "autoload_js.js" %}
 
+{% block register_mimetype %}
+
+  {%- if register_mime -%}
+  var JS_MIME_TYPE = 'application/javascript';
+  var HTML_MIME_TYPE = 'text/html';
+  var EXEC_MIME_TYPE = 'application/vnd.bokehjs_exec.v0+json';
+  var CLASS_NAME = 'output_bokeh rendered_html';
+
+  /**
+   * Render data to the DOM node
+   */
+  function render(props, node) {
+    var script = document.createElement("script");
+    node.appendChild(script);
+  }
+
+  /**
+   * Handle when an output is cleared or removed
+   */
+  function handleClearOutput(event, handle) {
+    var cell = handle.cell;
+
+    var id = cell.output_area._bokeh_element_id;
+    var server_id = cell.output_area._bokeh_server_id;
+    // Clean up Bokeh references
+    if (id !== undefined) {
+      Bokeh.index[id].model.document.clear();
+      delete Bokeh.index[id];
+    }
+
+    if (server_id !== undefined) {
+      // Clean up Bokeh references
+      var cmd = "from bokeh.io.state import curstate; print(curstate().uuid_to_server['" + server_id + "'].get_sessions()[0].document.roots[0]._id)";
+      cell.notebook.kernel.execute(cmd, {
+        iopub: {
+          output: function(msg) {
+            var element_id = msg.content.text.trim();
+            Bokeh.index[element_id].model.document.clear();
+            delete Bokeh.index[element_id];
+          }
+        }
+      });
+      // Destroy server and session
+      var cmd = "import bokeh.io.notebook as ion; ion.destroy_server('" + server_id + "')";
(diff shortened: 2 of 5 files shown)


bryevdv:
Unfortunately automated testing of the notebook is a continuing challenge. Probably the only reasonably quick thing that could be done here that would not involve building out some infrastructure would be to simply lock down the contents of the template in a unit test so that at least it does not change accidentally or without consideration. 

That said, someone will need to test this by hand with JupyterLab before it can be accepted. I am not sure I'd expect any problem but I do want to be clear that going forward, nothing will be accepted that does not also also work for JupyterLab. For purposes of validation, I'd request screenshots of the functionality being exercised from both the classic notebook and JupyterLab, appended to this PR. 

If this is going to become an officially supported usage, then it's probably worth a note in both the docstring, and the user's guide. The ReST source for the relevant docs is here: https://raw.githubusercontent.com/bokeh/bokeh/master/sphinx/source/docs/user_guide/notebook.rst

Finally we need to give some consideration to whether this introduces any performance or resource issues. cc @canavandl I think this should be ok but your input would be appreciated.

canavandl:
This appears to work in the classic notebook:

screen shot 2017-10-30 at 10 24 54 pm

And brings the classic notebook support up to matching the current JupyterLab behavior:

screen shot 2017-10-30 at 10 31 51 pm

A quick sentence in the user docs and/docstring saying this is possible would be nice. In terms of testing, it'd be nice to have a screenshot but that's blocked by #6854 . It'd also accept a followup issue to add a test in the future.

bryevdv:
@plazar do you anticipate being able to make the docs changes and adding a basic comparison unit test for the template?

plazar:
@bryevdv I'd be happy to update the docs. Where's the preferred place to document this behaviour (`show` or `output_notebook`)?

@guyd will follow up about adding tests for this PR.

guyd:
@bryevdv I will give some thinking on how we could test that. Regarding doing a comparison check on the template, isn't this going to create a lot of headaches moving forward. Unless I misunderstood your suggestion, this would make the test fail for each new commit.

bryevdv:
@plazar I would say add a short note to both docstrings, as well as to the Sphinx docs:

https://raw.githubusercontent.com/bokeh/bokeh/master/sphinx/source/docs/user_guide/notebook.rst

Just something along the lines of "repeated calls to `show` will cause objects to be displayed in order in the output cell"

@guyd Yes but making changes be a headache is the intent, this file should not change much at all. Making an annoying test will force any changes to at least be considered more carefully, which is probably the best we can do until image diff tests for notebooks can be restored.

plazar:
@bryevdv, @canavandl, thanks for the help. I've added some documentation to  `output_notebook` and `show` docstrings, as well as to `notebook.rst`. Please let me know what else I can do to help get this merged into master. :)

bryevdv:
@plazar have you built the docs? I am not sure that
```
:func:`show`
```
is sufficient. I usually have to put the full path, with a tilde in front to cause just the last part (the function name) to be displayed

bryevdv:
This does not seem to be working for me when I try it rebased on current master. Also seeing errors in the `handleClearOutput` function:

> Exception with thrown value: TypeError: undefined is not an object (evaluating 'Bokeh.index[id].model')

bryevdv:
scratch that, I guess I had not rebuilt when I though I had.

bryevdv:
@guyd are you amenable to adding a template test? I do think there is a substantial risk of this functionality regressing at some point without it. Additionally a comment in the template source code about the intentions with that index will help anyone looking at that code in the future.

guyd:
@bryevdv yes, I will do that over the weekend. I was thinking of adding a test that would compare a freshly computed file hash of the template with a hash of the file containing the fix. I will also add an _explicit_ comment to the template. And sorry for the extra delay.

guyd:
@bryevdv I've added the test over the weekend to find out that it fails on your Travis build. I first thought that it might be related to a difference in carriage returns on the build system and my mac. Not sure that this is the case anymore. First, the build runs on Ubuntu so there should be no differences from a Mac. Second, I tested on another Ubuntu laptop and the test is passing. So, my question to you, is there anything that touches the 'bokeh/core/_templates/autoload_nb_js.js' before it is being hashed by the test? If this is not the case then I am clueless. :-(

bryevdv:
@guyd I'm not sure offhand but I probably should have been clearer that I was suggesting a direct comparison of the file contents, not a hash. That will be much more informative (by providing a diff) in case there is a change or some problem.

guyd:
Unfortunately, making a file diff will not fix the build issue. I will push anyway another commit to see what is the diff in the file as you suggested.

bryevdv:
> Unfortunately, making a file diff will not fix the build issue. I will push anyway another commit to see what is the diff in the file as you suggested.

Right, but it will hopefully point out what the source of the problem is. Thanks for the updates

guyd:
@bryevdv The build step in Travis failed when fetching npm package. This is hardly related to this PR. Can someone restart the build to see if it passes?

bryevdv:
I've restarted a few times. NPM seems to be having persistent issues today, hopefully they will resolve soon.

bryevdv:
I've filed an issue: https://github.com/npm/registry/issues/248

bryevdv:
@guyd OK the NPM issue has been resolved. Here is the diff:
```
E           assert '{% extends "...endblock %}\n' == '{% extends "a...endblock %}\n'
E             Skipping 1930 identical leading characters in diff, use -v to show
E             Skipping 4310 identical trailing characters in diff, use -v to show
E             - ment.find(`.${CLASS_NAME.split(\' \')[0]}`);
E             ?           ^ ^^                          --
E             + ment.find("." + CLASS_NAME.split(\' \')[0]);
E             ?           ^ ^^^^
```
I can't say I have any idea what could cause this offhand, though.

mattpap:
This is a difference between ES6 and ES5 string interpolation syntax. The former is used in autoload_nb_js (which it should btw, because it's not supported by IE). I don't know what's the logic behind this test, but it looks like raw code is compared with its compiled version.

bryevdv:
@guyd @plazar I am not sure what is going on with the test and unfortunately I don't have bandwidth to devote to investigating. As an alternative, I would propose this: delete the template comparison text, and simple update some of the existing notebooks in the repo to exercise this capability. That's not great, but perhaps the best we can do at the moment to improve the odds that this feature is kept maintained.

guyd:
@bryevdv I like this approach. I will update the PR shortly.

nickleman:
Don't we need to change line 75 of autoload_nb_js.js as well? I suspect it is a corner case that just hasn't shown up yet.

bryevdv:
@guyd did you still intend to update the PR? If not we should close this out.

bryevdv:
@nickleman that seems plausible, if showing multiple apps from a single cell is desired

bryevdv:
OK given that the original contributors are no longer responding, I feel this must be closed. Perhaps we can revisit it at a different time.

PR #7140, opened 2017-10-31
Make slider handles updatable from arrow keys

### LINKED ISSUE #6862: Slider keyboard arrows no longer work 0.12.7
The Slider in v0.12.7 no longer responds to keyboard arrows. Confirmed with 0.12.6 it was working.
System: Mac OSX, El Capitan 10.11.6, Chrome 60.0 64-bit

agoodm:
- [x] issues: fixes #6862 

This was adapted from this [example](https://refreshless.com/nouislider/examples/#section-keyboard) in the nouislider documentation.

Tested this with the sliders example app and it works well, but I have yet to try this for sliders with multiple handles.

image

### DIFF
--- a/bokehjs/src/coffee/models/widgets/abstract_slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/abstract_slider.coffee
@@ -68,6 +68,27 @@ export class AbstractSliderView extends WidgetView
       @sliderEl.noUiSlider.on('slide',  (_, __, values) => @_slide(values))
       @sliderEl.noUiSlider.on('change', (_, __, values) => @_change(values))
 
+      # Add keyboard support
+      keypress = (e) =>
+        value = Number(@sliderEl.noUiSlider.get())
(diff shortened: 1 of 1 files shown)


bryevdv:
cc @mattpap

bryevdv:
@agoodm what happens if the slider has focus and you press something besides an arrow? At a glance, it looks like the property setting and callback code will get called unconditionally. Perhaps the `keypress` function should first check that `e` is ` left or right arrow, and if not, return immediately.

agoodm:
@bryevdv From the python side I couldn't see callbacks being fired from non-arrow keypresses. Nonetheless I think your suggestion is sensible, so I have implemented it.

bryevdv:
@agoodm thanks, there was a linter issue with trailing whitespace, I should have fixed it with a quick edit. 

@mattpap I will plan to merge this tomorrow before a dev build unless you object

PR #7143, opened 2017-10-31
note start/end of interactive periods on document

### LINKED ISSUE #5937: LOD in linked plots
When I have two linked plots (via `x_range=plot1.x_range`)  and move them only the plot I'm moving by hand gets LOD mode applied. This effectively makes the whole moving operation slow again. LOD should be applied to all plot being moved.

Version: 0.12.4

bryevdv:
- [x] issues: fixes #5937
- [x] tests added / passed

This PR adds support for "linked" LOD by making the document centrally responsible to note the start/end/duration of interactive periods. This simplifies the code at the point of usage and allows for some simple unit tests to be added. 

With this work, LOD start/end events are only triggered on the plot that originated the interactive period. 

Here is an example script that can be used:
```
import numpy as np

from bokeh.layouts import row

N = 4000
colors = [
]

p0 = figure()

p0.scatter(x, y, radius=radii,
          line_color=None)

colors = [
]

p1 = figure(x_range=p0.x_range, y_range=p0.y_range)

p1.scatter(x, y, radius=radii,
          line_color=None)

show(row(p0, p1))  # open a browser
```

There are further refinements that could be made in the future. In particular the same simple LOD threshold applies for all plots but this is not always ideal. In particular default settings do not trigger LOD on the second plot in the `vector.py` example, because the data size for the `multi_line` is small (92) even though the total number of points that make up those lines is much larger.

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -1,6 +1,7 @@
 import {Models} from "./base"
 import {version as js_version} from "./version"
 import {logger} from "./core/logging"
+import {LODStart, LODEnd} from "core/bokeh_events"
 import {HasProps} from "./core/has_props"
 import {Signal} from "./core/signaling"
 import {is_ref} from "./core/util/refs"
@@ -121,6 +122,9 @@ export class Document
     @idle = new Signal(this, "idle") # <void, this>
     @_idle_roots = new WeakMap() # TODO: WeakSet would be better
 
+    @_interactive_timestamp = null
+    @_interactive_plot = null
+
   Object.defineProperty(@prototype, "layoutables", {
     get: () -> (root for root in @_roots when root instanceof LayoutDOM)
   })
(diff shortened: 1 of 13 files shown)


bryevdv:
Still some tweaks to make sure lod events are handled reasonably

bryevdv:
This is seeming to fail due to two JS examples not working? Not sure what's going on there yet.

bryevdv:
Also three JS tests I thought were passing I had accidentally just commented out. Some problem with sinon/spy but I have not been able to to figure it out. @mattpap if nothing jumps out at you we can just close this PR and reject the issue with `wontfix` it's not that important.

mattpap:
@bryevdv, `gulp examples` fails to compile. I see this is in other PRs as well, so it's unrelated. TypeScript 2.6 was released yesterday, so I presume this is the reason. I will fix this in the morning.

bryevdv:
OK I found the issue, the mocks for `trigger_event` were expecting `calledOnce` to be true, but it's not. However, `called` is, and I believe this is correct. In addition to the actual UI event, e.g. pan or wheel, there is also LODStart. So I think it should expect more than one call, which is what the test now reflects.

bryevdv:
This should be pretty uncontroversial I will merge in the morning and cut a dev build unless there is comment.

bokehjs/src/coffee/document.coffee
```diff
@@ -148,6 +152,23 @@ export class Document
...
       @_pop_all_models_freeze()
 
+  interactive_start: (plot) ->
+    if @_interactive_plot == null
+      @_interactive_plot = plot
+      @_interactive_plot.trigger_event(new LODStart({}))
+    @_interactive_timestamp = Date.now()
+
+  interactive_stop: (plot) ->
+    if @_interactive_plot?.id == plot.id
```
mattpap: @bryevdv, what's the reason for comparing ids instead of just references (`@_interactive_plot == plot`)?

Issue #6370, opened 2017-06-01
Proposing a Bokeh Draw tool


caffeine-potent:
Code  would work like this   
 
```
ds = ColumnDataSource({'x': [], 'y': []})  
tool = DrawTool(source = ds, x = "x", y = 'y') 
my_plot.add_tools(tool)  

```  
The draw tool draws points on your plot.  A column data source of your choice is updated with the position of these dots. The point placement/min-resolution is dependent on your zoom level.    
  
An example that can be used in the documentation would be drawing points and fitting lines/curves to those points.   
  
Does anyone here have the patience to walk me through the development of this tool? I would like to start contributing other tools in the future.

bryevdv:
ping @philippjfr @jlstevens

jlstevens:
@caffeine-potent Definitely sounds very similar to a tool we are currently working on. @philippjfr from the description, do you think this fits into our plans for editing related tools?

philippjfr:
Yes, we're building a tool just like that, which will be part of a bigger suite of drawing and editing tools. Our DrawTool will probably be slightly more general letting us draw both points and lines. We'll make sure to keep you updated on our progress.

caffeine-potent:
@philippjfr   
Thanks for the response.  
Do you have any notes on your plans for the drawing or editing tools?

jlstevens:
@caffeine-potent The general plan is to write a number of tools that are broken up along two axes: a specification of *what* you are editing and then the type of action (insert, draw, move vertex etc). Then we have a plan to group these tools together into more powerful and general tools that merge the individual functionalities together.

caffeine-potent:
@jlstevens   
Thanks for the response! I'm really looking forward to this! 

 As for this issue. I'm not certain yet about GIT etiquette yet. Should this remain an open issue until the feature in question is released?

bryevdv:
Yes I think so, @jlstevens @philippjfr maybe a wiki page working document to discuss an outline requirements would be good? 

FYI @caffeine-potent just to set expectations, both Bokeh and HV are in the middle of getting releases out next week so there probably won't be much activity (even discussion) until after then.

jlstevens:
> Yes I think so, @jlstevens @philippjfr maybe a wiki page working document to discuss an outline requirements would be good?

Good idea!

> FYI @caffeine-potent just to set expectations, both Bokeh and HV are in the middle of getting releases out next week so there probably won't be much activity (even discussion) until after then.

[...]

PR #7149, opened 2017-11-01
Add support for multi-gesture tools

philippjfr:
All the current tools in bokeh are tied to a particular gesture, e.g. pan, box_zoom, box_select all make use of the pan gesture, however more complex tools like the draw tools suggested in (https://github.com/bokeh/bokeh/issues/6370) will require responding to multiple gestures to work effectively. This PR makes it possible to create a multi tool by declaring multiple ``event_types`` as an array rather than a single string. The tool will appear in its own section of the the toolbar and toggle all other tools sharing one of the same event types as inactive when it is activated.

In the example below I've added a few multi-tools (currently they don't have their own distinct icons), and have activated the pan and tap tools. When I select a multi-tool (the one with the lasso icon) which uses the ``pan`` and  ``tap`` events, both other tools are deactivated, and when I do the inverse and select pan and tap tools the multi-tool is also inactivated:

multi_tool

This PR simply provides the support for adding multi-gesture tools but does not yet add any concrete instance of such a tool. I'm therefore not quite sure how best to test this yet. Also let me know if you want any documentation entries for this. Hopefully this will allow us to eventually merge such tools back into bokeh (if deemed general enough). 

- [x] tests added / passed


philippjfr:
I'll look into those test errors.

mattpap:
> I'll look into those test errors.

@philippjfr, https://github.com/bokeh/bokeh/pull/7143#issuecomment-341270563.

philippjfr:
Ah, great thanks!

philippjfr:
@bryevdv What would it take for you to be happy to merge this? I still haven't worked out a good way to test it without a concrete multi-gesture tool implementation in bokeh itself, but it would be really nice if we didn't need to rely on a custom bokeh build for the continued development of the draw tools.

bryevdv:
> I still haven't worked out a good way to test it without a concrete multi-gesture tool implementation in bokeh itself

Unit tests on `UIEvent` seem feasible? Also you could define a multi-tool in the test to test with? It doesn't have to be in BokehJS itself.

philippjfr:
Sounds good, I'll get on that today.

philippjfr:
Just added some unit tests which define a multi-gesture tool and test that other tools are toggled active and inactive correctly. I'm not entirely sure what you mean by adding ``UIEvent`` tests for this.

bryevdv:
> I'm not entirely sure what you mean by adding UIEvent tests for this.

`register_tool` does different things if the parameter are lists vs strings, etc. Suggesting a unit test to cover the different cases.

philippjfr:
Got you, doing that now.

philippjfr:
Okay, I've now added a unit test that checks that multi-gesture tools receive all the gesture events they subscribe to.

philippjfr:
Thanks @bryevdv!

PR #7153, opened 2017-11-02
Update TypeScript to 2.6.1 and fix examples

mattpap:
Contrary to my previous understanding, "^2.6.1" means that only first number (major version) can't change. Given that TypeScript introduces breaking changes in minor versions, I locked its version to 2.6.1 and will update when new compilers are available.


mattpap:
Now notebooks started to fail in examples' tests.

mattpap:
I'm going to move ahead to merge this to not block other work.

PR #7157, opened 2017-11-02
Add new widget NumberInput for floating point numbers.

### LINKED ISSUE #6173: Add numeric input widget

herm:
Add a new widget type: NumberInput.
This widget is useful for cases where using a slider is to imprecise or the range of possible values is large. Just like a slider it updates it's value while it is being changed. The code enforces that all values passed to the callback are inside the range defined by start and end. 

Supersedes #6237

- [X] issues: fixes #6173
- [X] tests passed


bryevdv:
@mattpap any thoughts? seems ok at a glance tho wonder if more code re-use is possible

bryevdv:
ping @mattpap

bryevdv:
@herm My apologies for the delayed response. After some internal discussion, we have decided not to accept this PR, or two reasons:

* some of the features of number inputs are still supported inconsistently across browsers (so it's a good candidate for people to use as a custom extension when they have some idea what browsers might be used)

* We'd prefer to find a path that makes the existing input Model parametrizable, rather than adding another new Model entirely (we could discuss that on the issue if you are interested in pursuing it)

Definitely appreciate the PR and issue, I hope you'll consider contributing again to Bokeh in the future.

PR #7158, opened 2017-11-02
Upgrade to CoffeeScript 2

### LINKED ISSUE #7154: Investigate CoffeeScript 2 and its benefits for transition to TypeScript
CoffeeScript 2 generates ES6 (clases, et al.), so it should greatly simplify transition to TypeScript on syntactical level.

mattpap:
fixes #7154


mattpap:
So close yet far from working. Everything works perfectly with new coffeescript version, except that it wraps classes in closures, which prevents typescript from generating declaration files (`d.ts`), which in turn doesn't allow to run unit tests.

mattpap:
I hacked around this.

mattpap:
@bryevdv, I would like this to be merged just after release. I have big plans to build on top of this.

bryevdv:
Sounds like a good timeline for me.

mattpap:
@bryevdv, as we are postponing the release, lets have this merged earlier. This will allow to straighten a few typescript bits in PR #7108.

bryevdv:
OK I will take a look this afternoon.

bokehjs/src/coffee/models/ranges/range1d.coffee
```diff
@@ -18,14 +18,6 @@ export class Range1d extends Range
...
       @setv({bounds: [min, max]}, {silent: true})
 
-  constructor: () ->
-    # new Range1d({start: start, end: end}) or Range1d(start, end)
-    if this instanceof Range1d
-      return super(arguments...)
-    else
-      [start, end] = arguments
-      return new Range1d({start: start, end: end})
-
```
bryevdv: I am not in favor of removing this, but if you insist, since this will break people's code a migration note is in order

mattpap: This shorthand syntax was possible in ES5, but it's not possible in ES6. ES6 classes can't be initialized without `new` and this trick doesn't work anymore. TypeScript will reject this as well. This can be restored in future by implementing an API layer on top of models, but that will require much more work than could spend in this PR.

bryevdv: no problem, as I said it just needs a migration note

mattpap: I added a migration note.

mattpap:
I will go ahead and merge this. Migration note is in place.

PR #7162, opened 2017-11-03
Fix documentation for properties.RGB

herm:
Fix copy&paste error in documentation.


bryevdv:
Thanks for the catch and fix @herm, will merge when green

PR #7171, opened 2017-11-05
Correction to example

### LINKED ISSUE #7170: Missing Application variable in example/howto/server_embed/flask_embed.py
```
server = Server({'/bkapp': modify_doc}, allow_websocket_origin=["localhost:8000"])
```

`modify_doc` should be used by a handler to modify the document inside an application
This line result in a _static_path error.

PR will follow

ChristianTremblay:
It was lacking the "application" variable and was pointing only to the modify_doc function

Signed-off-by: Christian Tremblay <christian.tremblay@servisys.com>

- [ ] issues: fixes #7170

Please forget the /scripts file modified here...


ChristianTremblay:
The PR includes modification to files inside /scripts which are making trouble with the build. Please forgive my poor git skills and just look at the example .py

bryevdv:
Thanks for the PR, however the reason the example is not working for you is because it is using a new feature  that will be in the next release. There is always a risk of this if you run examples from master instead of a releas  tag or tarball

PR #7175, opened 2017-11-06
add init and validation check for map ranges

### LINKED ISSUE #7028: Gmaps incompatible with DataRange1d, make error louder/earlier
Runtime Info:

notebook server is 5.0.0
Python 3.6.1 |Continuum Analytics, Inc.| (default, May 11 2017, 13:09:58)
IPython 6.0.0
bokeh '0.12.9'
ipywidgets '7.0.1'
ubuntu 16

When i run the gmaps demo with push notebook it displays the dots in the wrong location until i interact with he map.

```python
from bokeh.models import (
  GMapPlot, GMapOptions, ColumnDataSource, Circle, DataRange1d, PanTool, WheelZoomTool, BoxSelectTool
)

map_options = GMapOptions(lat=30.29, lng=-97.73, map_type="roadmap", zoom=11)

plot = GMapPlot(
    x_range=DataRange1d(), y_range=DataRange1d(), map_options=map_options
)
plot.title.text = "Austin"


plot.api_key = API_KEY

source = ColumnDataSource(
    data=dict(
        lat=[],
        lon=[],
    )
)

circle = Circle(x="lon", y="lat", size=15, fill_color="blue", fill_alpha=0.8, line_color=None)
plot.add_glyph(source, circle)

plot.add_tools(PanTool(), WheelZoomTool(), BoxSelectTool())
h = show(plot, notebook_handle=True)
```

```python
source.data = dict(
        lat=[30.29, 30.20, 30.29],
        lon=[-97.70, -97.74, -97.78],
    )
push_notebook(handle=h)
```

pre interaction
image

post interaction
image

bryevdv:
- [x] issues: fixes #7028 
- [ ] tests added / passed

Not super happy with this but it will do for now. Nothing is done with "extra" ranges but I don't think that's common enough to bother with.

### DIFF
--- a/bokeh/models/map_plots.py
+++ b/bokeh/models/map_plots.py
@@ -39,16 +39,16 @@ class MapPlot(Plot):
     def __init__(self, *args, **kw):
         from ..models.ranges import Range1d
         for r in ('x_range', 'y_range'):
-            if not isinstance(kw.get(r, None), Range1d):
+            if r in kw and not isinstance(kw.get(r), Range1d):
                 raise ValueError('Invalid value for %r, MapPlot ranges may only be Range1d, not data ranges' % r)
         super(MapPlot, self).__init__(*args, **kw)
 
     @error(INCOMPATIBLE_MAP_RANGE_TYPE)
     def _check_incompatible_map_range_type(self):
         from ..models.ranges import Range1d
-        if not isinstance(self.x_range, Range1d):
+        if self.x_range is not None and not isinstance(self.x_range, Range1d):
             return "%s.x_range" % str(self)
-        if not isinstance(self.y_range, Range1d):
+        if self.y_range is not None and not isinstance(self.y_range, Range1d):
             return "%s.y_range" % str(self)
 
 class GMapOptions(MapOptions):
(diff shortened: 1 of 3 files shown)


bokeh/models/map_plots.py
```diff
@@ -36,6 +36,21 @@ class MapPlot(Plot):
 
     '''
 
+    def __init__(self, *args, **kw):
+        from ..models.ranges import Range1d
+        for r in ('x_range', 'y_range'):
+            if not isinstance(kw.get(r, None), Range1d):
+                raise ValueError('Invalid value for %r, MapPlot ranges may only be Range1d, not data ranges' % r)
```
mattpap: Why is this necessary? I would expect this should be handled solely by validation.

mattpap: I presume this is because validation happens a bit late.

bryevdv: It happens late, and also we have not (at least not yet) made validation errors raise real exceptions (they just output the message and proceed)

bryevdv:
this caught a bad example in the user guide docs, pushing fix soon

PR #7187, opened 2017-11-07
Permit ColumnsStreamedEvent to accept DataFrames

### LINKED ISSUE #7184: Support stream and update with pandas dataframes
When calling `ColumnDataSource.update` or `.stream` with Pandas dataframes I get JSON serialization errors

```python
source.stream(df)  # this fails
```

I find that I need to explicitly unpack the arrays beforehand

```python
d = {c: df[c] for c in df.columns}
source.stream(d)
```

Interestingly passing dataframes does work on the first iteration, but fails on every subsequent iteration.

bryevdv:
When calling CDS stream with DataFrames, the DataFrame would make it all the way to the `ColumnsStreamedEvent`, which presented it as-is for JSON encoding. This fails because DataFrames are not directly JSON serializable.

For compartmentalization and ease of testing, the issue is resolved by teaching `ColumnsStreamedEvent` how to convert DataFrames to the appropriate serializable dict itself. Other possible solutions not pursued at this time:

* ensure `ColumnsStreamedEvent` only ever receives a dict (higher level)
* teach the Bokeh JSON encoder how to handle DataFrames (lower level)

- [x] issues: fixes #7184
- [x] tests added / passed

### DIFF
--- a/bokeh/document/events.py
+++ b/bokeh/document/events.py
@@ -7,6 +7,10 @@ for :ref:`bokeh.events`.
 '''
 from __future__ import absolute_import
 
+from ..util.dependencies import import_optional
+
+pd = import_optional('pandas')
+
 class DocumentChangedEvent(object):
     ''' Base class for all internal events representing a change to a
     Bokeh Document.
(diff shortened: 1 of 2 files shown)


bryevdv:
cc @mrocklin @p-himik

PR #7188, opened 2017-11-08
Change css_classes property type to List

### LINKED ISSUE #6841: RangeSlider css_classes not being updated anymore
It seems in 0.12.7 the `css_classes` attribute of the RangeSlider is not being updated anymore by the callbacks. It was working in 0.12.6.
In the example below, both the text input and the slider starts with the `initial` class as expected. When the slider changes, both should gain the `sent-later` class, but only the text input is gaining it...

```
from bokeh.io import curdoc
from bokeh.models import TextInput, RangeSlider
from bokeh.models import Div

div1 = Div(text="""
        <style>
        .sent-later {
            border-radius: 0.5em;
            padding: 1em;
            border: 3px solid #2397D8;
        }
        </style>
""")


def on_slider_change(attr, old, new):
    slider.css_classes.append('sent-later')
    text.css_classes.append('sent-later')
    print(slider.css_classes)
    print(text.css_classes)


text = TextInput(
        value=str(2),
        title='Population Beta:')

slider = RangeSlider(
        start=0,
        end=10,
        value=(0, 2),
        step=1,
        title='Population Beta')

slider.css_classes = []
slider.css_classes.append('initial')
text.css_classes = []
text.css_classes.append('initial')

slider.on_change('value', on_slider_change)

curdoc().add_root(column(div1, widgetbox(text, slider)))

```

bryevdv:
In order for in place updates such as `.append` to properly trigger events, the property type must be `List` (so that `PropertyValueList` is used to wrap the value). This PR changes the type to `List`, but also adds an `accept` clause so that setting from tuples, etc. still functions.

- [x] issues: fixes #6841
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tests/test_layouts.py
+++ b/bokeh/models/tests/test_layouts.py
@@ -1,6 +1,6 @@
 import pytest
 from bokeh.plotting import Figure
-from bokeh.models.layouts import Row, Column, WidgetBox
+from bokeh.models.layouts import Row, Column, WidgetBox, LayoutDOM
 from bokeh.models.widgets import Slider
 from bokeh.models.sources import ColumnDataSource
 
@@ -76,3 +76,11 @@ def check_widget_box_children_prop(layout_callable):
 def test_WidgetBox():
     check_props(WidgetBox())
     check_widget_box_children_prop(WidgetBox)
+
+def test_LayoutDOM_css_classes():
+    m = LayoutDOM()
+    assert m.css_classes == []
+    m.css_classes = ['foo']
+    assert m.css_classes == ['foo']
+    m.css_classes = ('bar', )
+    assert m.css_classes == ['bar']
--- a/bokeh/models/layouts.py
+++ b/bokeh/models/layouts.py
@@ -58,10 +58,14 @@ class LayoutDOM(Model):
 
     """)
 
-    css_classes = Seq(String, help="""
+    # List in order for in-place changes to trigger changes, ref: https://github.com/bokeh/bokeh/issues/6841
+    css_classes = List(String, help="""
     A list of css class names to add to this DOM element. Note: the class names are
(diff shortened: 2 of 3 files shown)


bokeh/models/tests/test_layouts.py
```diff
@@ -76,3 +76,11 @@ def check_widget_box_children_prop(layout_callable):
 def test_WidgetBox():
     check_props(WidgetBox())
     check_widget_box_children_prop(WidgetBox)
+
+def test_LayoutDOM_css_classes():
+    m = LayoutDOM()
+    assert m.css_classes == []
+    m.css_classes = ['foo']
+    assert m.css_classes == ['foo']
+    m.css_classes = ('bar', )
```
mattpap: I don't see a point in supporting tuples here or in general for `Seq`s.

bryevdv: I can easily imagine defining a sequence of classes in tuples rather than lists, being able to assign from that directly is a nicety. But in this case also avoids changing the previous semantics.

mattpap: > But in this case also avoids changing the previous semantics.

That's definitively convenient. Let's leave it as-is and later change the semantics altogether, especially if we plan to maintain a decent level of compatibility between bokehjs typescript APIs and Python.

bryevdv: What has to remain compatible and consistent are the property types and their default values. Adding language-appropriate convenience on top of that, e.g. accepting tuples, which are common in Python, for direct assignment, is completely orthogonal to Python/BokehJS compatibility.

bryevdv: @mattpap are you requesting a change or are you ok with merging?

mattpap: No, OK with this.

PR #7192, opened 2017-11-08
Reduce (multi) plot initialization time

### LINKED ISSUE #6294: Browser rendering extremely slow when many figures in a gridplot
Hi all,

when I have many figures in a gridplot, browsers take a very very long time to render the html file, even if the data is very small. 
For instance, when I run
```
import bokeh.plotting as bplt
import bokeh.io as bio
n_rows = 10
n_columns = 6
data_x = [1, 2, 3]
data_y = [1, 2, 3]
grid = []
for __ in range(n_rows):
    row = []
    for __ in range(n_columns):
        sub_fig = bplt.figure(width=200, height=100)
        sub_fig.line(data_x, data_y)
        row.append(sub_fig)
    grid.append(row)

fig = bplt.gridplot(grid)

bplt.reset_output()
bio.output_file('test.html')
bplt.save(fig)
```
and then open the 'test.html' file in a browser (firefox 53.0.2 or chromium 58.0.3029.96, on ubuntu 16.04 LTS), it takes 40 seconds to render, with 100 % use of one of my CPU (cf below)
cpu_bokeh

Moreover, when the html file is open, the memory used increases by 400 Mo, even if the html file only weights 380 ko.

Do you know if there is a workaround for this problem? Maybe, this slowness is due to one special feature of bokeh, I could bypass?
Otherwise, is there a chance that bokeh will soon optimize the rendering of multiple plots? 
I would love to use bokeh in my project, but I need to make this kind of plots and the rendering time is two long.

Thank you for your work!

EDIT:  bokeh==0.12.5

mattpap:
This so far reduces initialization time of the 60 plot example from 37 to 22 seconds on my hardware. Further optimization is possible, but not as cheap as this. I may add some more if time permits. Otherwise I will look into this more after release.

addresses #6294

### DIFF
--- a/bokehjs/src/coffee/document.coffee
+++ b/bokehjs/src/coffee/document.coffee
@@ -119,6 +119,7 @@ export class Document
 
   constructor: () ->
     documents.push(this)
+    @_init_timestamp = Date.now()
     @_title = DEFAULT_TITLE
     @_roots = []
     @_all_models = {}
@@ -149,6 +150,7 @@ export class Document
     @_idle_roots.set(model, true)
 
     if @is_idle
+      logger.info("document idle at #{Date.now() - @_init_timestamp} ms")
       @idle.emit()
 
   clear : () ->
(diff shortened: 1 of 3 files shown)


mattpap:
I will go ahead and merge this, as this interferes with my other work.

PR #7195, opened 2017-11-09
Don't require CSS width/height for responsive sizing modes

### LINKED ISSUE #7014: The inherited width should be set to 100% and up to the user to implement margin and padding restrictions
**Here is a link to the line that is in contention:**
https://github.com/bokeh/bokeh/blob/08020a31393630a2bda1bb083564492d14669264/bokeh/core/_templates/file.html#L33

When inheriting a sizing parameter it should always be 100% up to the user to be in control of the restrictions. This should be done entirely via the users own HTML setup of the layout or via a set of control options in the doc object, I would think.

mattpap:
fixes #7014

### DIFF
--- a/bokeh/core/_templates/file.html
+++ b/bokeh/core/_templates/file.html
@@ -24,17 +24,6 @@ that accepts these same parameters.
         <title>{{ title|e if title else "Bokeh Plot" }}</title>
         {{ bokeh_css }}
         {{ bokeh_js }}
-        <style>
-          html {
-            width: 100%;
-            height: 100%;
-          }
-          body {
-            width: 90%;
-            height: 100%;
-            margin: auto;
-          }
-        </style>
     </head>
     <body>
         {{ plot_div|indent(8) }}
--- a/bokehjs/src/coffee/core/dom.ts
+++ b/bokehjs/src/coffee/core/dom.ts
@@ -152,6 +152,18 @@ export function parent(el: HTMLElement, selector: string): HTMLElement | null {
   return null
 }
 
+export type Sizing = {top: number, bottom: number, left: number, right: number}
+
+export function margin(el: HTMLElement): Sizing {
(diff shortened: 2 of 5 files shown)


bryevdv:
FYI if you run the one failing example `dashboard.py` locally, this is the error:

> unsatisfiable constraint: SidePanel(5784ED1C1A3346DDABFBCF2CCD7794AE, above).height - 28.49453125 >= 0

mattpap:
Yeah, but this isn't layout's fault, though it could handle this situation more gracefully. The problem is that I underestimated stupidity of CSS. There's a good reason why this `width` and especially `height` is there. CSS4 will close this gap, but maybe something can be done in the meantime, because it's so close to working.

mattpap:
In fact there is a solution, though more a hack. Not rendering widgets in `initialize` fixes this. We shouldn't be doing any rendering in init anyway, so this would be a win-win.

bryevdv:
testing now

bryevdv:
@mattpap is there an easy way to expose this to be configurable, so .e.g our existing examples could be made to look like they did before? I agree that not-setting is the best default.

bryevdv:
@mattpap any comment on the question above?

mattpap:
>  is there an easy way to expose this to be configurable,

I would do this via resources by adding a raw CSS, but I'm not sure if this is currently actually feasible or not. It definitively would be feasible if I finally dig out code I wrote a few months ago, which was supposed to address such situation (abstract resource vs resource bundle separation).

bryevdv:
I'd say we need to hold off on this until there is some proven mechanism to get back things the way they were if desired, and also a migration note would be appropriate.

bryevdv:
@mattpap if you add a migration note now I'll merge and cut a dev build

mattpap:
@bryevdv, I added migration note.

PR #7196, opened 2017-11-09
Make extensions_ion_range_slider work with CoffeeScript 2

### LINKED ISSUE #7193: Ion slider extension example broken
The Ion slider extension example here (when deployed with `0.12.11dev3` is broken:

https://bokeh.pydata.org/en/dev/docs/user_guide/extensions_gallery/widget.html#userguide-extensions-examples-widget

The issue appears to be in the Ion code itself, perhaps we are misusing it and a change uncovered this, or perhaps something changed to actively break it:

screen shot 2017-11-08 at 12 52 38

cc @mattpap

mattpap:
fixes #7193

### DIFF
--- a/sphinx/source/docs/releases/0.12.11.rst
+++ b/sphinx/source/docs/releases/0.12.11.rst
@@ -15,8 +15,20 @@ Migration Guide
 NOTE: the 0.12.x series is the last planned release series before a version
 1.0 release. For more information see the `project roadmap`_.
 
-Breaking changes
-~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


bryevdv:
Should probably add a "Switch to CS2" migration note, e.g. "To help with the eventual migration to TS, we have switched to the CS2 compiler. This should be completely transparent to standard users but custom extensions may need small changes in some specific cases, for example..."

bryevdv:
I added the note will get this merged and a dev build out later today for @hsparra

Issue #6932, opened 2017-09-13
0.12.9 misplaced figure title


rocheseb:
You changed ResetToolView.prototype.doit with that line :
`return this.plot_model.plot.trigger_event(new bokeh_events_1.Reset());`
I am not sure if that is the reason but the reset tool no longer fixes the position of updated titles in plots.
See this example:

```
from bokeh.io import show, output_file
from bokeh.plotting import figure
from bokeh.models import CustomJS, Button
from bokeh.layouts import gridplot

output_file("test.html")
fig = figure()
fig.line(x=[1,2,3], y=[1,2,3])

button = Button(label="add title",callback=CustomJS(args={'title':fig.title},code="title.text='new title'"))

show(gridplot([[fig,button]]))
```
After pushing the button the title will be inside the plot instead of above it, this used to be fixed when pushing the reset button.
This gets fixed when you drag the webpage tab and then let it go, or when you open the console (checked in firefox and chrome).

If an original title is given like 
`fig = figure(title='original title')`
then the title will be correctly placed above the plot, and the updated title too

bryevdv:
@rocheseb I'm not clear, when you say:

 > After pushing the button the title will be inside the plot instead of above it, this used to be fixed when pushing the reset button.

Do you mean it has *always* been the case that changing the title has misplaced the title, and that hitting reset was your workaround to fix it? If so, that was purely an accident. I don't think the change has anything to do with this, there was other layout work that might have had an impact. Regardless, if the title is being misplaced, we should concentrate on fixing that core issue, not on an accidental and unintentional  bandaid workaround.

rocheseb:
@bryevdv yes agreed the misplaced title is the issue and it has always happened when a title was not originally given.

mattpap:
This is "by design" and limited by the fact that updating certain parts of layout isn't possible. I will be working on this broad subject in my next round of layout improvements.

mattpap:
@rocheseb, for now you can use ugly hack from https://github.com/bokeh/bokeh/issues/6748#issuecomment-321014827 to force re-layout in your custom js.

bryevdv:
@mattpap after the recent changes, this code produces the toolbar above the button, what's going on?
<img width="798" alt="screen shot 2017-10-25 at 19 10 40" src="https://user-images.githubusercontent.com/1078448/32029155-45931e76-b9b8-11e7-8bc7-01d5b68c9a5c.png">

bryevdv:
[...]

PR #7197, opened 2017-11-09
Allow to recompute layout when panels' dimensions change

### LINKED ISSUE #4394: Axis label size change does not trigger a re-alignment
When @bryevdv was testing the new layout, he was surprised that when he interacted with a grid of plots and zoomed to the point where axis labels were longer than before that other plots didn't update their alignment.

Should we implement / provide an option for that - I worry that it might decrease performance.

### LINKED ISSUE #6356: Colorbar tick labels truncated
Tick labels on the colorbar are truncated.

bokeh_tick_label_truncation

This only occurs for some lengths of tick label, changing the data column which controls the color of the glyphs changes the layout of the plot and the tick labels are no longer truncated.

bokeh_tick_trunc_gif

Code to reproduce the top image:
```python
from bokeh.models import ColumnDataSource, LinearColorMapper, Select, ColorBar
from bokeh.palettes import Viridis256
from bokeh.sampledata import iris

df = iris.flowers

data_options = sorted(df.columns)[:-1]

p = figure(plot_height=400, plot_width=500, toolbar_location=None, id='fig')

# Select widgets
color_select = Select(title='Color', options=data_options, value=data_options[0])


def update_color(attr, old, new):
    source.data.update({'color': df[color_select.value]})
    color_mapper.high = df[color_select.value].max()
    color_mapper.low = df[color_select.value].min()


# set the on_change functions
color_select.on_change('value', update_color)


color_mapper = LinearColorMapper(low=df[color_select.value].min(),
                                 high=df[color_select.value].max(),
                                 palette=Viridis256,
                                 id='cmpr')

source = ColumnDataSource({'x': df['petal_length'],
                           'y': df['petal_width'],
                           'color': df[color_select.value]})
 

# add circle glyphs
r = p.circle('x', 'y',
             size=8,
             color={'field': 'color', 'transform': color_mapper},
             source=source, id='circles')

color_bar = ColorBar(color_mapper=color_mapper,
                     label_standoff=5,
                     location=(10, 0))

# put the pieces together
p.add_layout(color_bar, 'right')
controls = widgetbox([color_select],
                     width=200)
layout = row(controls, p)

curdoc().add_root(layout)
```

bokeh version: 0.12.5
python: 3.5

Reproduced in chrome, firefox, and the notebook.

Discussed in https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/lZKC74AR6qo

### LINKED ISSUE #6676: Axes' tick labels are trimmed when moving plot
According to the output of `Bokeh.version` in the JS console, the page at https://demo.bokehplots.com/apps/sliders uses Bokeh 0.12.3. If you move the plot around (say, move it up to see negative values on the Y axis), you'll see that the plot width changes to make all axes' tick labels visible.

Now, if you serve the same app with Bokeh 0.12.6, you'll notice that the behavior is different - the width is not changed and the labels get trimmed.

Probably it's normal as many people desire a constant size of the plot itself. In this case, is there any way to make sure that the axes' tick labels are not trimmed at least within some range?

mattpap:
fixes #4394
fixes #6676 
~~~fixes #6356~~~

### DIFF
--- a/bokehjs/src/coffee/core/layout/side_panel.coffee
+++ b/bokehjs/src/coffee/core/layout/side_panel.coffee
@@ -135,14 +135,20 @@ _align_lookup_positive = {
   right  : LEFT
 }
 
+export _view_sizes = new WeakMap() # <View, number>
+
 export update_panel_constraints = (view) ->
+  visible = not view.model.props.visible? or view.model.visible
+  size = if visible then Math.round(view._get_size()) else 0
+
   s = view.solver
 
   if view._size_constraint? and s.has_constraint(view._size_constraint)
+    if _view_sizes.get(view) == size
+      return
     s.remove_constraint(view._size_constraint)
 
-  visible = not view.model.props.visible? or view.model.visible
-  size = if visible then view._get_size() else 0
+  _view_sizes.set(view, size)
 
   view._size_constraint = GE(view.model.panel._size, -size)
   s.add_constraint(view._size_constraint)
--- a/bokehjs/src/coffee/models/layouts/layout_dom.coffee
+++ b/bokehjs/src/coffee/models/layouts/layout_dom.coffee
@@ -90,6 +90,7 @@ export class LayoutDOMView extends DOMView
       @_solver.add_constraint(EQ(variables.height, @_root_height))
 
     @_solver.update_variables()
+    @_solver_inited = true
 
   _suggest_dims: (width, height) ->
     variables = @model.get_constrained_variables()
@@ -111,14 +112,20 @@ export class LayoutDOMView extends DOMView
     else
       @_do_layout(false, width, height)
 
+  partial_layout: () ->
+    if not @is_root
+      @root.partial_layout()
+    else
+      @_do_layout(false)
+
   layout: (full=true) ->
     if not @is_root
-      @root.layout(full)
+      @root.layout()
     else
-      @_do_layout(full)
+      @_do_layout(true)
 
   _do_layout: (full, width=null, height=null) ->
-    if full
+    if not @_solver_inited or full
       @_solver.clear()
       @_init_solver()
 
--- a/bokehjs/src/coffee/models/plots/plot.coffee
+++ b/bokehjs/src/coffee/models/plots/plot.coffee
@@ -37,7 +37,7 @@ export class PlotView extends LayoutDOMView
     @plot_canvas_view.save(name)
 
   @getters {
-    plot_canvas_view: () -> (view for view in values(@child_views) when view instanceof PlotCanvasView)[0]
+    plot_canvas_view: () -> @child_views[@model._plot_canvas.id]
   }
 
 export class Plot extends LayoutDOM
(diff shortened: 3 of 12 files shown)


mattpap:
This is useful for testing:
```python
import numpy as np

from bokeh.models import NumeralTickFormatter

def plot(n):
    x = np.linspace(0.1, 5, 100)

    p = figure(title="log axis example: %s" % n, y_axis_type="log",
               y_range=(0.001, 10**22))

    p.yaxis.formatter = NumeralTickFormatter(format="0,0")

    p.line(x, np.sqrt(x), legend="y=sqrt(x)",
           line_color="tomato", line_dash="dotdash")

    p.line(x, x, legend="y=x")
    p.circle(x, x, legend="y=x")

    p.line(x, x**2, legend="y=x**2")
    p.circle(x, x**2, legend="y=x**2",
             fill_color=None, line_color="olivedrab")

    p.line(x, 10**x, legend="y=10^x",
           line_color="gold", line_width=2)

    p.line(x, x**x, legend="y=x^x",
           line_dash="dotted", line_color="indigo", line_width=2)

    p.line(x, 10**(x**2), legend="y=10^(x^2)",
           line_color="coral", line_dash="dashed", line_width=2)

    p.xaxis.axis_label = 'Domain'
    p.yaxis.axis_label = 'Values (log scale)'
    return p

output_file("logplot.html", title="log plot example")

show(column(plot(1), plot(2)))  # open a browser
```

bryevdv:
Guessing you already know, but`iris_splom.py` becomes entirely unresponsive to pan/zoom with just this change

bryevdv:
Just as a point of comparison, `0.12.4` is very responsive to pan/zoom, but also does make room for axis labels:

screen shot 2017-11-10 at 17 26 33

Are we over-computing layout now, or has the number of constraints increased?

mattpap:
> Are we over-computing layout now, or has the number of constraints increased?

Both. I will be looking into this.

mattpap:
> > Are we over-computing layout now, or has the number of constraints increased?

> Both. I will be looking into this.

Though this is true, the real reason the original approach didn't work is because it didn't make sense. `_get_size()` gives the preferred size of a panel, but actual `_{width,height}.value` may be different, e.g. a title and toolbar sharing a panel.

bryevdv:
@mattpap this does not actually fix the color bar case #6356 It does fixes the main plot axes labels when resizing. It's possible there is just another event handler to hook up, the colorbar code is somewhat duplicative of axis. I will take a quick look

bryevdv:
@mattpap well its not something I will be able to figure out in short order. My inclination is to merge this now and move the color bar issue to the next milestone. Do you have any thoughts?

mattpap:
#6356 seems to be a different issue after all. I think this will get fixed together with issues like #6932.

mattpap:
I tested this with my previous approach and it worked, but the reason was that I enabled full relayout by mistake. Obviously full relayout will make examples like iris_splom unresponsive.

bryevdv:
OK I will merge this as is then and move the last issue to the next milestone, then cut an RC

PR #7199, opened 2017-11-10
Missing import of logger in toolbar.coffee

busino:
Missing import of logger.
https://github.com/bokeh/bokeh/edit/master/bokehjs/src/coffee/models/tools/toolbar.coffee#L35
Error is logged if event_type does not match and therefore a Reference error was raised because of the missing import.


bryevdv:
Thanks for the catch @busino

PR #7201, opened 2017-11-10
multi_line() Documentation Addition

### LINKED ISSUE #7198: Request for improvement to ColumnDataSource documentation in reference to multi_line()
Hello,

I just wanted to comment on the mutli_line behavior. Several parameters are NumberSpec's which seemed to be ignored silently and the default values used due to the structure of the column datasource required for multi_view. In addition, the unique column data source consisting of a list of lists prevents the filtering of the data set using CDSView. In addition, if I use a CDS I cannot use an iterable for the other parameters such as color (passing in a source and a list of colors throws and error below).

A brief comment on my use case. I have two plots, one shows a scatter plot of averages. The other shows a plot of frequency vs time. I am using a hovertool JS callback to highlight the respective circle and line(s).  I had hoped to pass the multi_line CDS to the JS callback arg to change the alpha values.

In my prior code I used a for loop to plot out many lines. I had wanted to pass a nested dictionary of models to the JS callback but the JS call back expects models or strings only in a dictionary. I could go back and use a dictionary comprehension but that seems a bit messy.

```
plot.multi_line(xs='dates',ys='kde',
                                color = '#c50000',
                                line_alpha='alpha',
                                line_width = 'linewidth',
                                source =source)
```

```
_process_sequence_literals(glyphclass, glyph_ca, source, is_user_source)
  File "C:\WinPython-64bit-2.7.10.3\python-2.7.10.amd64
\lib\site-packages\bokeh\plotting\helpers.py", line 296, in _process_sequence_li
terals
    raise RuntimeError(_GLYPH_SOURCE_MSG)
RuntimeError:
Supplying a user-defined data source AND iterable values to glyph methods is
not possibe. Either:

Pass all data directly as literals:

    p.circe(x=a_list, y=an_array, ...)

Or, put all data in a ColumnDataSource and pass column names:

    source = ColumnDataSource(data=dict(x=a_list, y=an_array))
    p.circe(x='x', y='x', source=source, ...)
```

David-Herman:
Adding in additional Information regarding  CDS data structure used in multi_line() to clarify usage.

- [ X] issues: fixes #7198
- [ ] tests added / passed

### DIFF
--- a/sphinx/source/docs/user_guide/plotting.rst
+++ b/sphinx/source/docs/user_guide/plotting.rst
@@ -93,9 +93,9 @@ accomplished with the |multi_line| glyph method:
 .. note::
     This glyph is unlike most other glyphs. Instead of accepting a one
     dimensional list or array of scalar values, it accepts a "list of lists"
-    for x and y positions of each line, parameters xs and ys. multi_line  
-    also expects a scalar value or a list of scalers per each line for 
-    parameters such as color, alpha, linewidth, etc. Similarily, a 
+    for x and y positions of each line, parameters xs and ys. multi_line
+    also expects a scalar value or a list of scalers per each line for
+    parameters such as color, alpha, linewidth, etc. Similarily, a
     ColumnDataSource may be used consisting of a "list of lists" and a
     lists of scalars where the length of the list of scalars and length of
     lists must match.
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks @David-Herman ! There is a similar note for `patches`, which is also a "list of lists" glyph. Would you mind making this update there as well (in the same .rst file).

David-Herman:
I added the matching note and found a copy and paste error where the multi_line text was copied over for patches but did not make literal sense. I would also recommend adding a code example of use of a CDS that mirrors one or both examples

bryevdv:
Awesome thanks @David-Herman

PR #7203, opened 2017-11-10
Encode JSON into application/json script for standalone HTML

### LINKED ISSUE #7039: Clarify Some Security Considerations
A user noted this sample code:
```
import sys

import bokeh.embed
import bokeh.plotting
import bokeh.resources

figure = bokeh.plotting.figure()
figure.text(0, 0, "</script><script>alert('BAD')</script>")

sys.stdout.write(bokeh.embed.file_html(figure, bokeh.resources.CDN))
```

This could be happen in two cases:

*  A users does this to themself (not an issue, it is not possible to exploit oneself by definition)

* A user accepts untrusted user input and passes it without first sanitizing it

WRT the second scenario, it is worth reminding that like other web app frameworks and servers, Bokeh is explicitly purposed with transmitting unescaped HTML and JS code to browsers. This means, as with any other web development, that any untrusted user inputs should be scrubbed before using as outputs (e.g. in this case before passing to Bokeh).

That said, it is never appropriate for a `</script>` tag to be passed on, so this issue proposes to:

* clarify and restrict and clearly document exactly which Bokeh functions are expected to pass HTML and JS unescaped, and remind users that that inputs to these functions should *always* be sanitized first if they are coming from untrusted sources.

* automatically scrub any use of  a closing script tag `</script>` in any Bokeh output

cc @bokeh/dev @pzwang @mattpap

mattpap:
This splits off docs JSON from javascript script into a HTML-safe encoded application/json script in standalone HTML output mode. Otherwise it uses HTML-safe encoded JavaScript string with JSON content. As long as one uses `script_for_render_items()` as the **only** way to generate bokeh script, then vulnerability described in #7039 is fixed. 

fixes #7039

### DIFF
--- a/bokeh/core/_templates/doc_js.js
+++ b/bokeh/core/_templates/doc_js.js
@@ -2,10 +2,8 @@
   function embed_document(root) {
     var docs_json = {{ docs_json }};
     var render_items = {{ render_items }};
-
     root.Bokeh.embed.embed_items(docs_json, render_items{%- if app_path -%}, "{{ app_path }}" {%- endif -%}{%- if absolute_url -%}, "{{ absolute_url }}" {%- endif -%});
   }
-
   if (root.Bokeh !== undefined) {
     embed_document(root);
   } else {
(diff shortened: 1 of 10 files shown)


bryevdv:
Thank you for this including the test, looks like a good approach

bryevdv:
tested apps, file, and notebook with inline and server resources

cc @philippjfr @jlstevens there were some internal API changes here, not sure if they will affect your not but a head's up just in case.

Issue #7189, opened 2017-11-08
HoverTool data not completely matching when using CDSView filters in bokeh 0.12.11dev2


Zyell:
When using the HoverTool with a custom tooltip in conjunction with a CDSView of any kind, the data does not match up.  If you use the default tooltips, you will find that the index matches appropriately to the glyph being hovered over.  However, when you use a specific column from the data source in a tootip, it will not match the glyph being hovered over.

This was tested with python 3.6 and bokeh versions: 0.12.10, 0.12.11dev2 in a notebook and embedded in an html page.

```python
import pandas as pd
import numpy as np
from bokeh.models import CDSView, BooleanFilter, ColumnDataSource, HoverTool
from bokeh.plotting import figure, show
from bokeh.io import output_notebook

output_notebook()


df = pd.DataFrame([{'xs': [1, 3, 2], 'ys': [2, 1, 4], 'name':'a'},
                   {'x': 5, 'y': 2, 'xs': np.array([np.nan]), 'ys': np.array([np.nan]), 'name':'b'},
                   {'xs':[3, 7, 6, 6], 'ys': [3, 3, 4, 4], 'no_hover': True, 'name':'c'},
                   {'xs':[3, 4, 6, 6], 'ys': [4, 7, 8, 5], 'name':'d'},
                   {'x': 7, 'y': 5, 'xs': np.array([np.nan]), 'ys': np.array([np.nan]), 'no_hover': True, 'name':'e'},
                   {'x': 2, 'y': 7, 'xs': np.array([np.nan]), 'ys': np.array([np.nan]), 'name':'f'}
                   ])
source = ColumnDataSource(data=df)
p = figure(plot_width=400, plot_height=400, tools='hover')

view1 = CDSView(source=source, filters=[BooleanFilter(list((df['x'].notnull()) & (df['no_hover'] != True)))])
view2 = CDSView(source=source, filters=[BooleanFilter(list((df['x'].isnull()) & (df['no_hover'] != True)))])
view3 = CDSView(source=source, filters=[BooleanFilter(list((df['x'].isnull()) & (df['no_hover'] == True)))])
view4 = CDSView(source=source, filters=[BooleanFilter(list((df['x'].notnull()) & (df['no_hover'] == True)))])
a = p.circle(x='x', y='y', source=source, view=view1, radius=0.5)
p.circle(x='x', y='y', source=source, view=view4, radius=0.5, color='red')
[...]

PR #7204, opened 2017-11-11
fix tooltip rendering with CDSview

Zyell:
fixes issue #7189


bryevdv:
Thanks @Zyell I am going to hold off on this until @clairetang6 is able to comment. At a minimum there are a couple of other places in the same file that this change should be made. Thanks for your patience

clairetang6:
@bryevdv @Zyell, this fixes the tooltips for all of the non-line and non-multilines including patches, but the issue is there for multi-lines. To fix that, the following lines should be placed at line 236 after the `switch @model.line_policy` block. 

```python
          if renderer_view.model instanceof GlyphRenderer
            index = renderer_view.model.view.convert_indices_from_subset([i])[0]
          else
            index = i
```
Then, the next two lines should be changed to use `index` instead of `i`, in the same way. 
```python
          vars = {index: index, segment_index: jj, x: x, y: y, sx: sx, sy: sy, data_x: data_x, data_y: data_y}

          tooltip.add(rx, ry, @_render_tooltips(ds, index, vars))
```

bryevdv:
@Zyell FYI if you can make these changes in the next day or two, I can include them in the release Monday. Otherwise it will be the following release.

Zyell:
@bryevdv & @clairetang6 I will make these changes today. I'm traveling home now, so it will be later this evening. Thanks!

Zyell:
@bryevdv & @clairetang6 I have made the requested changes.  Let me know if there is anything else.  Thanks!

bryevdv:
I am going to go ahed and merge this but we definitely need to find a way to maintain these many different cases better under test. @clairetang6 we should discuss testing strategy as part of your current work on refactoring selection

bryevdv:
Thanks for the PR @Zyell !

PR #7207, opened 2017-11-13
add kwargs documentation for DirectoryHandler

n-west:
- [ ] issues: none
- [x] tests added / passed: none

Added so that the required filename keyword is exposed to docs in https://bokeh.pydata.org/en/latest/docs/reference/application/handlers/directory.html#bokeh.application.handlers.directory.DirectoryHandler.__init__

### DIFF
--- a/bokeh/application/handlers/directory.py
+++ b/bokeh/application/handlers/directory.py
@@ -12,7 +12,7 @@ The directory may also optionally contain:
 * A ``theme.yaml`` file containing a Bokeh theme to automatically apply to
   all new documents.
 
-* A ``tempates`` subdirectory containing templates for app display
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @n-west, as long as you are updating this file, could you add the optional `argv` keyword arg as well?

n-west:
Sure. What's your preference for commits here? Just add a new one or squash them together?

BTW, bokeh has been mostly great to work with-- really nice job by the dev team and contributors :-)

bryevdv:
@n-west we ultimately do squash merges on our end, so we are pretty flexible with however people want to submit. This PR is fine now as is so I plan to merge when CI goes green

bryevdv:
Also, thanks!

PR #7208, opened 2017-11-13
Rewrite models/tickers/* in TypeScript

mattpap:
addresses #6481


bryevdv:
I'd like to get the tests working before merging.

mattpap:
I rewrote tests in TypeScript as well, and this revealed a few issues. This will need further work.

bokehjs/src/coffee/models/ranges/factor_range.coffee
```diff
@@ -70,6 +70,13 @@ export map_three_levels = (factors, outer_pad, inner_pad, factor_pad, offset=0)
 export class FactorRange extends Range
   type: 'FactorRange'
 
+  `
+  factors: string[]
+
+  mids: string[] | undefined
+  tops: string[] | undefined
+  `
```
bryevdv: @mattpap What going on here with the backticks?

mattpap: Those backticks are CoffeeScript's syntax for injecting raw JavaScript. I hijacked this to inject TypeScript, to fill-in missing type information. Otherwise modules being rewritten would require many more instances of `any` type sprinkled around.

PR #7215, opened 2017-11-15
Allow ES6 output from bokehjs' build

### LINKED ISSUE #7214: Allow ES6 output from bokehjs' build
ES6 isn't something we would switch to right now, but it's good to work with to future proof bokehjs. It will also allow us to work as close to TypeScript code as possible for development, without need for generating expensive source maps (currently we generate only partial sources maps by default).

mattpap:
The default is still ES5. To generate ES6 issue `gulp build --es6`. Everything else works the same. Nice side effect is size reduction of bokeh.min.js from 651 to 580 kB.

fixes #7214


bokehjs/gulp/linker.ts
```diff
@@ -342,7 +351,52 @@ export class Module {
...
+      ],
+      body: {
+        type: "BlockStatement",
+        body: Array.from(this.ast.body) as Statement[],
+      }
+    }
+    const comment: Comment = {type: "Block", value: this.canonical}
+    wrapper.leadingComments = [comment]
+    this.ast.body = [wrapper]
+  }
```
bryevdv: Presumably this can go away when the port is fully completed?

mattpap: No, this has nothing to do with coffeescript or typescript. Given that module ids aren't human readable (or readable at all in the case of the main bundle), this is a quick way to ease reading bundle contents.

bryevdv: You mentioned having to work around something about how CS2 converted  classes to closures (?) where is that? (just curious)

mattpap: That's a different piece of code:
https://github.com/bokeh/bokeh/blob/b8b7559cd67da9a43ec950b2fa090959ba9e619b/bokehjs/gulp/tasks/scripts.ts#L31-L57

mattpap: Yeah, the above garbage will go away as soon as transition is done.

PR #7233, opened 2017-11-17
Rewrite core in TypeScript

mattpap:
This transitions most of core (excluding visuals), document and model to TypeScript.

addresses #6481


mattpap:
@bryevdv, this is done. Better code hub fails, but some of its expectation are just plain ridiculous.

PR #7251, opened 2017-11-26
Fix incorrect values of all-integers Sliders

### LINKED ISSUE #7250: Slider callback doesn't occur on particular values
bokeh/0.12.10 and bokeh/0.12.11rc2
Chrome/62.0.3202.94 and Firefox/57.0

A registered slider callback isn't executed on particular slider values

```
from bokeh.io import curdoc
from bokeh.models import Slider

def slider_test_callback(attr, old, new):
    print(new)

slider_test = Slider(start=0, end=9, value=0, step=1, title="Test")
slider_test.on_change('value', slider_test_callback)

curdoc().add_root(slider_test)
```

Stopping on every value, while moving the slider left to right all the way, produces the following output in a console (note missing 3 and 6):
```
1
2
4
5
7
8
9
```

However, changing a number of possible slider values either to 9 (```end=8```), or to 11 (```end=10```) makes a slider to work correctly. Also, switching to floating point values, e.g. ```start=0, end=0.9, value=0, step=0.1``` works fine:
```
0.1
0.2
0.3
0.4
0.5
0.6
0.7
0.8
0.9
```

p-himik:
Fixes #7250

### DIFF
--- a/bokehjs/src/coffee/models/widgets/slider.coffee
+++ b/bokehjs/src/coffee/models/widgets/slider.coffee
@@ -14,7 +14,7 @@ export class SliderView extends AbstractSliderView
 
   _calc_from: ([value]) ->
     if Number.isInteger(@model.start) and Number.isInteger(@model.end) and Number.isInteger(@model.step)
-      return value|0
+      return Math.round(value)
     else
       value
 


bryevdv:
Thanks @p-himik !

PR #7253, opened 2017-11-26
Add MetricTickFormatter and tests

### LINKED ISSUE #7252: Feature offer: metric prefix ticker
type: feature

Hi,

I made a TickFormatter that uses [Metric Prefixes](https://en.wikipedia.org/wiki/Si_prefixes) like k, M, G, T. This is very useful in electronics, where everything is expressed using these prefixes. People reading exponential notation have to mentally convert to the nearest metric prefix.

This is how it looks:

image

This is how it handles large/small values:

image

ignamv:
- [x] issues: fixes #7252
- [x] tests added / passed


bokehjs/src/coffee/models/formatters/metric_tick_formatter.ts
```diff
@@ -0,0 +1,70 @@
...
+  const min_tick: number = Math.min(...absticks);
+  const max_tick: number = Math.max(...absticks);
+  const prefixes: Prefix[] = (max_tick / min_tick < 10) ?
+    // One metric prefix that makes the largest value pretty
+    repeat(metric_prefix(Math.max(...ticks.map(Math.abs))), ticks.length) :
+    // Different prefixes that make each value pretty
+    ticks.map(metric_prefix);
+  const scaledticks: [number, string][] = prefixes.map(
+    ([name, value], idx) => [ticks[idx] / value, name] as [number, string]);
+  const formatter = new Intl.NumberFormat('en-us', {
```
mattpap: Very interesting. This may come in handy in a few places. It's even supported by IE.

p-himik: I'd argue that `'en-us'` shouldn't be hard-coded. As far as I can tell, there are no other places where Bokeh explicitly hard-codes locale.
I think an empty array can be passed there so the default locale is selected.

bryevdv:
@ignamv This a great PR, I really appreciate the tests and clean code. I do think a couple of small change would make this even more useful to more people:

* an optional `unit` property, so that it can display e.g. `1.2km` (in this case `unit='m'`)
* an optional separator, so that you can have `1.2km` or `1.2 km`

Additionally, I'd ask that the `repeat` function be moved to `core/util/array` and also get tests (since it is isolated and easy to test it definitely should get tests)

mattpap:
Before we get into technical details, we need to discuss merit of this addition, because similar functionality can be achieved with `NumeralTickFormatter`. I would prefer to implement missing bits there instead, especially we already pay not insignificant price for having it (numbro) on-board.

bryevdv:
I don't see anything in Numbro that would do this currently. Are you suggesting further patching our fork of numbro? I am not sure I like that idea, also I am not sure how to add the `unit` I mentioned above in the context of numbro. That addition makes this feature extremely useful to a lot of users.

mattpap:
The logic is there, under `0 a` or `0 b` syntax, just needs adjustment to match this use case. `unit` is not possible currently, but seems like a trivial addition. This can be implemented in the fork, but I can safely presume that this would be a welcome addition upstream.

I'm not discussing functionality here, but the implementation. We already have almost everything we need to implement this, so adding another model seems like a stretch. Going this road we would have to consider percentage or currency formatters as well, but again, those can be achieved with `NumeralTickFormatter`. The only reason I see for adding another model is if there were additional properties needed, that would be cumbersome or impossible to express in a formatting language (`MercatorTickFormater` comes to my mind in this context).

ignamv:
Thanks! I won't have much time to work on this during the week so feel free to make changes.

ignamv:
I've added metric prefixes to numbro. Problem is that numbro does not support exponential notation.

https://github.com/BenjaminVanRyseghem/numbro/pull/316

bryevdv:
@ignamv I am not sure how actively maintained numbro is. You might have to make a PR against our fork https://github.com/bokeh/numbro Given that, you could also potentially implement exponential notation there. 

Alternatively, we could imagine putting some or all of the implementation in the existing `NumeralTickFormatter`:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/formatters/numeral_tick_formatter.coffee

Which avoid having to add a separate new model.

bryevdv:
This seems to have stagnated and now has conflicts. Absent response from OP, I feel this must be closed especially since this sort of thing can be accomplished with a custom extension.

PR #7257, opened 2017-11-27
Don't cut corners and implement proper get_size() method

### LINKED ISSUE #7255: Zooming performance regression in 0.12.11rc
I was just playing around with the bokeh 0.12.11rc2 release and noticed a pretty large performance regression when zooming on a number of linked plots when compared to bokeh 0.12.10.

You can see a comparison of the same grid plot generated using bokeh 0.12.10 and 0.12.11rc2 here:

[0.12.10](http://assets.holoviews.org/tmp/test_0.12.10.html)

[0.12.11rc2](http://assets.holoviews.org/tmp/test_0.12.11rc2.html)

The 0.12.11rc2 version will appear to do nothing for a while, but eventually it does zoom with a lot of stutter.

mattpap:
Performance is still worse than before, but that's expected. This should improve with further refinements to layout and rendering pipeline.

fixes #7255

### DIFF
--- a/bokehjs/src/coffee/core/layout/side_panel.coffee
+++ b/bokehjs/src/coffee/core/layout/side_panel.coffee
@@ -138,8 +138,7 @@ _align_lookup_positive = {
 export _view_sizes = new WeakMap() # <View, number>
 
 export update_panel_constraints = (view) ->
-  visible = not view.model.props.visible? or view.model.visible
-  size = if visible then Math.round(view._get_size()) else 0
+  size = view.get_size()
 
   s = view.solver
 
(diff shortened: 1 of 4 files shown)


bryevdv:
Works well in some spot testing, @philippjfr confirms performance improvement in his specific example, so merging now. 

@mattpap can you explain the changes, offhand they just look to push the same computation out to other classes (but not to change the computation)

mattpap:
@bryevdv, the problem was (mostly) with hidden renderers, specifically axes, which reported non-zero size, but got assigned 0 due to their invisibility. This resulted in `needs_layout()` always returning `true`, and thus performance regression was observed.

PR #7265, opened 2017-11-30
update for jinja exposed an issue with Options processing

bryevdv:
fix formatting for new jinja2 version


bryevdv:
@mattpap you'll want to merge/rebase master in to any current work

PR #7275, opened 2017-12-04
Rewrite axes in TypeScript

mattpap:
This rewrites axes, ranges, some of renderers and last bits of the core (visuals, mixins and layout).

addresses #6481


bokehjs/src/coffee/models/ranges/data_range1d.ts
```diff
@@ -0,0 +1,262 @@
...
+  end: number
+  range_padding: number
+  range_padding_units: PaddingUnits
+  flipped: boolean
+  follow: StartEnd
+  follow_interval: number
+  default_span: number
+  bounds: [number, number] | "auto"
+  min_interval: any // XXX: what's this?
+  max_interval: any // XXX: what's this?
```
bryevdv: These are numbers

bokehjs/src/coffee/models/ranges/factor_range.ts
```diff
@@ -0,0 +1,233 @@
...
+  factor_padding: number
+  subgroup_padding: number
+  group_padding: number
+  range_padding: number
+  range_padding_units: PaddingUnits
+  start: number
+  end: number
+  bounds: [number, number] | "auto"
+  min_interval: any // XXX: what's this
+  max_interval: any // XXX: what's this
```
bryevdv: also numbers

mattpap: Yeah, I actually answered this to my self when rewriting another type of range. I will have to revisit this anyway, because inheritance needs to be set up properly for these and several other properties.

bokehjs/src/coffee/core/layout/side_panel.ts
```diff
@@ -0,0 +1,271 @@
...
+    let baseline: string
+    let align: string
+
+    if (isString(orient)) {
+      baseline = _baseline_lookup[side][orient]
+      align = _align_lookup[side][orient]
+    } else {
+      if (orient === 0) {
+        baseline = "whatever" // XXX: _baseline_lookup[side][orient]
+        align = "whatever" // XXX: _align_lookup[side][orient]
```
mattpap: @bryevdv, this is something to review. The original code (in comments) doesn't make much sense.

bryevdv: It's been this way (which is broken, I think) as far back as 0.5 at least. Here is what should happen when angle is 0: the labels should be incrementally different from when orient==0.1 So this branch should do whatever makes that happen. It will probably take a nested if-block, since things are probably different based on the side.

mattpap:
I'm going ahead and merge this as this blocks further work on transition to TypeScript. I opened issue #7296 for the `orient == 0` bug.

PR #7277, opened 2017-12-04
Add proper error message when invalid BOKEH_LOG_LEVEL (#7276)

### LINKED ISSUE #7276: Unclear Exception when BOKEH_LOG_LEVEL env variable is set to a wrong value
I was developing a bokeh server application, and my application was using the BOKEH_LOG_LEVEL env variable to control the log level. Apparently the same env variable is being used by bokeh server itself, and is expected to be something in `['trace', 'debug', 'info', 'warn', 'error', 'fatal']`. As I was passing something different as an env variable ("DEBUG" for the record) bokeh server was failing with the following error:

```
ValueError: ("Uknown log level '%s', valid levels are :%s:, "['trace', 'debug', 'info', 'warn', 'error', 'fatal']'")
```
It would be easier if I could see the wrong value which was supplied and caused the error. e.g I would expect the ValueError as:

```
ValueError: ("Uknown log level 'DEBUG', valid levels are ['trace', 'debug', 'info', 'warn', 'error', 'fatal']")
```

sestus:
- [x] issues: fixes #7276

### DIFF
--- a/bokeh/resources.py
+++ b/bokeh/resources.py
@@ -193,7 +193,7 @@ class BaseResources(object):
             "trace", "debug", "info", "warn", "error", "fatal"
         ]
         if not (level is None or level in valid_levels):
-            raise ValueError("Unknown log level '%s', valid levels are: %s", str(valid_levels))
+            raise ValueError("Unknown log level '{}', valid levels are: {}".format(level, str(valid_levels)))
         self._log_level = level
 
     @property


bryevdv:
Thanks for the PR @sestus Test failure seems to be spurious, restarting.

bryevdv:
Thanks for the PR @sestus !

PR #7282, opened 2017-12-05
add missing return statements to get_indices TypeScript version

bryevdv:
Function always returned `undefined` due to missing `return` statements, caused `OpenURL` to fail.


bryevdv:
@mattpap I will merge this for `0.12.12` as well

mattpap:
TypeScript's pretty dumb allowing this, even with `any` return type. This deserves at least a warning, e.g. "expression in statement context". Anyway, LGTM.

mattpap:
On the other hand, knowing that any piece of code in ES can bring side effects, this could have been a valid implementation.

Issue #6655, opened 2017-07-21
x/y_range not properly set with tile source


lgasnier:
Initial extent can't be set as expected by setting ranges as explained in http://bokeh.pydata.org/en/latest/docs/user_guide/plotting.html#setting-ranges.

However, extent can be well adjust manually with Wheel Zoom and Box Zoom.

See script below.
If we comment the two lines :
```
wms = BBoxTileSource(url=url)
p.add_tile(wms)
```
initial extent is properly set.

Note that url called by BBoxTileSource is a not tiled web map service (WMS getmap()).
The same behavior occurs with Tiled WMS (called with WMTSTileSource).

Code example:
Map extent should be equal to the colored square.

```
from bokeh.plotting import figure
from bokeh.models.tiles import BBoxTileSource, WMTSTileSource
from bokeh.io import show
from bokeh.io import output_file

output_file("test_getmap.html")

url = ("http://ws.carmencarto.fr/WMS/74/smartphone_mdo_nationale2016?"
        "&REQUEST=getmap&layers=mdo&BBOX={XMIN},{YMIN},{XMAX},{YMAX}"
        "&SRS=epsg:3857&FORMAT=image/png&WIDTH=700&HEIGHT=700")

xmin = 200000
xmax = 200000 + 1000000
ymin = 5000000
ymax = 5000000 + 1000000

p = figure(width=700, height=700, x_range=(200000, 200000 + 1000000),
           y_range=(5000000, 5000000 + 1000000))
p.axis.visible = False
#wmts = WMTSTileSource(url="http://c.tile.openstreetmap.org/{Z}/{X}/{Y}.png")
#p.add_tile(wmts)
wms = BBoxTileSource(url=url)
p.add_tile(wms)
p.quad(top=[ymax], bottom=[ymin], left=[xmin],
       right=[xmax], color="#B3DE69", alpha=0.5)
show(p)
```

nanodan:
Is there a current workaround for this?

lgasnier:
Sorry, I didn't find a workaround.

TiemenSch:
Perhaps this is more broad than the tile? 

I'm having issues with a plot to which I add two sets of rects using p.rect().

At first, the axis behave as normal and leave a little whitespace around all rects. 

After the second one though, the whitespace on the beginning of the axis becomes way too much.

This only happens when the rects are somewhat "near" the (0,0) value, but still farther away than the initial set. This would leave me to believe that the whitespace need not increase. 

[...]

Issue #5126, opened 2016-09-08
Use decimal zoom level for WMTSTileSource


jbednar:
When a Bokeh plot includes both a WMTSTileSource and an another plot of some kind, such as the ImageRGBA plot in this example taken from https://anaconda.org/jbednar/landsat/notebook:

![image](https://cloud.githubusercontent.com/assets/1695496/18366496/82017f9a-75dc-11e6-8bf8-8014e06ec8d7.png)

the plot is very different than it would be had there been no tile source:

![image](https://cloud.githubusercontent.com/assets/1695496/18366409/2bef43ee-75dc-11e6-9443-a85d772928ea.png)

I.e., including the tile source changes both the aspect ratio and the bounding region of the plot (in data coordinates).  Changing the aspect ratio is arguably appropriate, though it would be much nicer just to be able to declare that on the data itself so that it would correctly have square pixels with and without the tile source.  But changing the bounding region is very rude; I want to show the data, not a huge area of tiles with my data in one tiny bit!

According to @brendancol, what's happening is that the tile source is forcing the display to an integer zoom level at the start, which is how some tile sources usually work online (not being able to re-render to levels in between).  But that doesn't seem appropriate here, because the wheel zoom and box zoom tools allow users to select arbitrary zoom levels.  So, can we please remove the snapping to an integer zoom level, at least by default, so that the explicitly requested bounding box in data coordinates can be respected?  

@brendancol suggests that it would not be difficult to snap to some specified number of decimal points instead of integer zoom levels, which would at least help alleviate the issue, though it wouldn't solve it for all possible bounding boxes.

The biggest practical problem with the current approach is that when notebooks are rendered by anaconda.org, they use the initial extent only, regardless of whether one has subsequently zoomed in, and so there's no way to upload notebooks that show the actual data; currently they only show this vast area of uninformative tile data.

It would also be nice to allow the requirement for a square aspect ratio to be expressed at the ImageRGBA or Scatter levels, so that geographic data will never get squashed whether or not there are underlying tiles, but that's not quite as urgent to fix.

jbednar:
@brendancol, any hope this can be addressed soon?  I keep having to work around it when posting static notebooks to anaconda.org, by choosing very precise plot sizes that default to a reasonable zoom level, which is not something I want anyone to have to keep doing!

bryevdv:
ping @brendancol any comments on this?

bryevdv:
[...]

PR #7294, opened 2017-12-07
Added snap_to_zoom option for MercatorTileSource

philippjfr:
As discussed in https://github.com/bokeh/bokeh/issues/6655 and https://github.com/bokeh/bokeh/issues/5126, when using a map tile source the plot will snap to the nearest zoom level, which is not always desirable and often makes it very difficult to initialize the plot to a reasonable zoom range.

This PR adds an option called ``snap_to_zoom`` to the ``MercatorTileSource`` which disables this behavior and instead simply tries to maintain the appropriate aspect ratio. It does this by finding the supplied range which is closest to the next zoom level and then adjusts the other range by scaling it as a fraction of the range of the closest zoom level. In my testing this produces much better results overall, but I've left the option set to True by default to maintain backward compatibility.

Here is a before and after comparison. I've added two points to the plot which also correspond to the lower and upper bounds for each axis. As you can see the default (and current) behavior result in a suboptimal plot:

screen shot 2017-12-07 at 7 36 43 pm

When we disable the ``snap_to_zoom`` option we can see that it still maintains the aspect but now tries to respect the extents as much as possible:

screen shot 2017-12-07 at 7 36 34 pm

Before going further down this path it would be great to get your feedback @bryevdv. Personally I think this behavior is much preferable over the default behavior but I didn't want to unilaterally change default behavior. I also think this is preferable over respecting user supplied ranges even if they distort aspect because that's never a good thing to do. However we could consider allowing that by adding a ``snap_extents`` parameter instead, with 'zoom', 'aspect' and ``None`` options. 

- [x] issues: fixes #xxxx
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tiles.py
+++ b/bokeh/models/tiles.py
@@ -68,6 +68,10 @@ class MercatorTileSource(TileSource):
 
     initial_resolution = Override(default=156543.03392804097)
 
+    snap_to_zoom = Bool(default=True, help="""
+    Enables snapping of extents to the closest zoom level.
+    """)
+
     wrap_around = Bool(default=True, help="""
     Enables continuous horizontal panning by wrapping the x-axis based on
     bounds of map.
--- a/bokeh/models/tiles.py
+++ b/bokeh/models/tiles.py
@@ -68,9 +68,8 @@ class MercatorTileSource(TileSource):
 
     initial_resolution = Override(default=156543.03392804097)
 
-    snap_to_zoom = Bool(default=True, help="""
-    Enables snapping of extents to the closest zoom level.
-    """)
+    snap_to_zoom = Bool(default=False, help="""
+    Forces initial extents to snap to the closest larger zoom level.""")
 
     wrap_around = Bool(default=True, help="""
     Enables continuous horizontal panning by wrapping the x-axis based on
--- a/bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee
+++ b/bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee
@@ -5,7 +5,7 @@ export class MercatorTileSource extends TileSource
   type: 'MercatorTileSource'
 
   @define {
-    snap_to_zoom:       [ p.Bool,   true               ]
+    snap_to_zoom:       [ p.Bool,   false              ]
     wrap_around:        [ p.Bool,   true               ]
   }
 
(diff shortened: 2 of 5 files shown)


jbednar:
Thanks so much for this fix!

> I think this behavior is much preferable over the default behavior but I didn't want to unilaterally change default behavior.

Personally, I would consider the current behavior to be a longstanding bug, so I would vote for setting snap_to_zoom=False by default.  It is true that it will change current results for some people, but (a) almost certainly for the better, and (b) there are probably far more people who will get correct plots in the future when it is set to False than will be annoyed by their existing plots changing.

> I also think this is preferable over respecting user supplied ranges even if they distort aspect because that's never a good thing to do, however we could consider allowing that by adding a snap_extents parameter instead, with 'zoom', 'aspect' and None options.

Yes, I could imagine supporting an aspect-changing range, but it should definitely not be the default, and does not seem like an urgent thing to add.

bryevdv:
I'm OK with changing the default as long as there is a migration note. I would be -1 an any aspect-changing features in this PR (and probably in general)

bokeh/models/tiles.py
```diff
@@ -68,6 +68,10 @@ class MercatorTileSource(TileSource):
 
     initial_resolution = Override(default=156543.03392804097)
 
+    snap_to_zoom = Bool(default=True, help="""
+    Enables snapping of extents to the closest zoom level.
```
bryevdv: "Forces initial extents to snap to the closest larger zoom level"

bokehjs/src/coffee/models/tiles/mercator_tile_source.coffee
```diff
@@ -111,12 +112,20 @@ export class MercatorTileSource extends TileSource
...
-    [xmin, ymin, xmax, ymax] = extent
+    if !@snap_to_zoom
+      xscale = (xmax-xmin)/desired_x_delta
+      yscale = (ymax-ymin)/desired_y_delta
+      if xscale > yscale
+        desired_x_delta = (xmax-xmin)
+        desired_y_delta = desired_y_delta*xscale
+      else
+        desired_x_delta = desired_x_delta*yscale
+        desired_y_delta = (ymax-ymin)
```
bryevdv: There may be re-usable code around `match_aspect` code paths. In any case anything that is a simple mathematical computation would be good to factor out since it can then be tested with simple unit tests.

philippjfr: I had a look and couldn't find anything directly relevant. I think the problem is a bit better constrained than in the ``match_aspect`` case anyway.

philippjfr:
Okay, great! I'll change the default then, add some tests and update existing tests if needed.

bryevdv:
Just needs migration note in the release notes and we can merge.

philippjfr:
Done, let me know if that migration note looks okay.

PR #7295, opened 2017-12-07
Removed a broken link from server.rst

biggstd:
Removed a broken link. No changes made to code.


bryevdv:
Thanks for the catch and the PR @biggstd !

PR #7300, opened 2017-12-08
Make it possible to specify fixed attachment to HoverTool

### LINKED ISSUE #7292: Make it possible to create HoverTool with attachment set to explicit left, right, above, below
Bokeh 0.12.13

Right now, the only possible values for the `attachment` attribute of `HoverTool` are `horizontal` and `vertical`. I think it would be great to allow other values that `Tooltip` supports - `left, `right`, `above`, `below`.

p-himik:
Fixes #7292

### DIFF
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@ -191,6 +191,10 @@ VerticalLocation = enumeration("above", "below")
 #: Specify a horizontal location in plot layouts
 HorizontalLocation = enumeration("left", "right")
 
+#: Specify an attachment for tooltips
+TooltipAttachment = enumeration("horizontal", "vertical",
+                                "left", "right", "above", "below")
+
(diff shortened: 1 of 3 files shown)


mattpap:
Looks good.

mattpap:
@p-himik, there is some trivia to fix (`test_enums_contents`).

p-himik:
What's the point of `test_enums_contents`? It looks like a test for its own sake, without any substance. The `any changes to contents of enums.py easily trackable here` comment doesn't really make any sense to me either - the changes should be easily trackable with `git log` and `git blame`.

mattpap:
@p-himik, I completely agree with your assessment. However, there is an ongoing discussion / disagreement regarding how to increase stability of bokeh's APIs, and this is one of proposed approaches.

bryevdv:
Thanks @p-himik !

PR #7305, opened 2017-12-09
Using 'utf-8' codec to open scripts in windows

### LINKED ISSUE #7272: Codec error on windows
### Versions

    os: Windows 10 x64 Professional with Chinese language
    python: 3.6.3
    bokeh: 0.12.10
    tornado: 4.5.2

#### Error 1: 

On windows system, if a script file with some charactor other than ascii is saved in utf-8 codec, and then run `bokeh serve script.py`, an error will be given like:

    D:\>bokeh serve --show script-utf8.py
    ERROR:'gbk' codec can't decode byte 0xad in position 201: illegal multibyte sequence

There is a similar [SO question](https://stackoverflow.com/questions/46187520/python-trouble-with-encoding-on-windows-when-run-bokeh-serve-bokeh-plotting-li). I've figured this out in the answer. It's because the [line of code](https://github.com/bokeh/bokeh/blob/f3d13963a36130ec126f0a1e87d3abcf8e5870f7/bokeh/application/handlers/script.py#L53) as mentioned open the file in the default codec ('gbk' in this case).

#### Error 2
If the script file is changed to gbk codec, then the file can be opened properly. But another kind of error will be given:

```
D:\>bokeh serve --show script-gbk.py
2017-12-03 09:50:29,811 Starting Bokeh server version 0.12.10 (running on Tornado 4.5.2)
2017-12-03 09:50:29,816 Bokeh app running at: http://localhost:5006/script
2017-12-03 09:50:29,817 Starting Bokeh server with process id: 1544
--- Logging error ---
--- Logging error ---
--- Logging error ---
Exception ignored in: <bound method Future.__del__ of <tornado.concurrent.Future object at 0x0000025F39BE77F0>>
  File "C:\Anaconda3\lib\site-packages\tornado\concurrent.py", line 341, in __del__
    tb = traceback.format_exception(*self._exc_info)
  File "C:\Anaconda3\lib\traceback.py", line 117, in format_exception
    type(value), value, tb, limit=limit).format(chain=chain))
  File "C:\Anaconda3\lib\traceback.py", line 510, in __init__
    self._load_lines()
  File "C:\Anaconda3\lib\traceback.py", line 522, in _load_lines
    self.__context__._load_lines()
  File "C:\Anaconda3\lib\traceback.py", line 522, in _load_lines
    self.__context__._load_lines()
  File "C:\Anaconda3\lib\traceback.py", line 522, in _load_lines
    self.__context__._load_lines()
  [Previous line repeated 3 more times]
  File "C:\Anaconda3\lib\traceback.py", line 520, in _load_lines
    frame.line
  File "C:\Anaconda3\lib\traceback.py", line 282, in line
    self._line = linecache.getline(self.filename, self.lineno).strip()
  File "C:\Anaconda3\lib\linecache.py", line 16, in getline
    lines = getlines(filename, module_globals)
  File "C:\Anaconda3\lib\linecache.py", line 47, in getlines
    return updatecache(filename, module_globals)
  File "C:\Anaconda3\lib\linecache.py", line 137, in updatecache
    lines = fp.readlines()
  File "C:\Anaconda3\lib\codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd6 in position 199: invalid continuation byte
```

I haven't located the source code of this error. But I managed to give a simplest example, and it has something to do with `gbk` charactor in the pandas data file directory.

### Example code:

Codes of this example: 
```python
import pandas as pd
from bokeh.io import curdoc

# Set up data
d = pd.read_csv('d:/中文.csv', index_col=0)
source = ColumnDataSource(data=dict(x=d.x, y=d.y))

# Set up plot
plot = figure(plot_height=400, plot_width=400, title="my sine wave",
              tools="crosshair,pan,reset,save,wheel_zoom",
              x_range=[0, 12], y_range=[-2.5, 2.5])
curdoc().add_root(plot)
```
The codes is saved in `gbk` codec named `script-gbk.py`, and `utf-8` codec named `script-utf8.py` seperately.

### Data used in this example:

```python
import numpy as np
import pandas as pd

N = 200
y = np.sin(x)
d = pd.DataFrame(dict(x=x, y=y))
d.to_csv('d:/中文.csv')
```

gepcel:
Using 'utf-8' codec to open scripts in bokeh app in windows

- [x] issues: fixes #7272

### DIFF
--- a/bokeh/application/handlers/script.py
+++ b/bokeh/application/handlers/script.py
@@ -29,6 +29,7 @@ Documents by adding an empty plot with a title taken from ``args``.
 from __future__ import absolute_import, print_function
 
 from .code import CodeHandler
+import io
 
 class ScriptHandler(CodeHandler):
     ''' Modify Bokeh documents by executing code from Python scripts.
@@ -50,7 +51,7 @@ class ScriptHandler(CodeHandler):
             raise ValueError('Must pass a filename to ScriptHandler')
         filename = kwargs['filename']
 
-        with open(filename, 'r') as f:
+        with io.open(filename, 'r', encoding='utf-8') as f:
             kwargs['source'] = f.read()
 
         super(ScriptHandler, self).__init__(*args, **kwargs)


gepcel:
I noted that there was an error in `bokeh/application/handlers/tests/test_script.py:35: AssertionError`. I've runned the script in my system, and couldn't find why.

bryevdv:
@gepcel The issue is specific to Python 2.7 AFAICT. I appears things do not work when specifying an encoding in `io.open` and in  the file, because the python parser complains when there is an encoding in an already unicode string. Here is a minimal example:
```
In [9]: import io, ast

In [10]: code = io.open("enc.py", encoding="utf8").read()

In [11]: code
Out[11]: u'# -*- coding: utf-8 -*-\n\nimport os\n'

In [12]: nodes = ast.parse(code, "path")
  File "path", line unknown
SyntaxError: encoding declaration in Unicode string
```

So this actually makes me take a step back. Was the original example tried by providing an encoding commend as the first line? It's possible that would solve the problem, in which case I'd say this is maybe a documentation issue.

gepcel:
@bryevdv, This a getting too complicated for me, I don't know much about python2. Maybe you can take this over from here? If there's anything I can help, please tell.

bryevdv:
@gepcel Can you try running your problem example with a line like this at the very top:
```
# -*- coding: utf-8 -*-
```
(possibly updated to the actually encoding you are needing if `utf-8` is not it? The python interpreter will detect such comments and adjust the encoding based on it. That might be the solution, rather than any changes to bokeh itself.

gepcel:
In an environment like:

    python: 3.6.2
    bokeh: 0.12.10

A script file saved in 'utf-8 without BOM' format, with `# -*- coding: utf-8 -*-` as the first line, runned with `bokeh serve --show main.py`, gives the following error message ('gbk' maay be the default codec in a Chinese language Windows sytem):

    ERROR:'gbk' codec can't decode byte 0xb4 in position 1004: illegal multibyte sequence

Script file with or without `# -*- coding: utf-8 -*-`, gives the same error message.

bryevdv:
> Maybe you can take this over from here? 

@gepcel OK well this is out of my depth at the moment, then. It will require some investigation since the `coding` comment did not work. But I'm not sure when I will be able to turn to it. So I will leave the issue open, but close the PR since it is not functional

gepcel:
@bryevdv Some more tests shows that change [this line](https://github.com/bokeh/bokeh/blob/f3d13963a36130ec126f0a1e87d3abcf8e5870f7/bokeh/application/handlers/script.py#L53) to `with open('filename, 'rb') as f:` may also solve this specific problem, but some tests won't pass. Also if the script file is indeed saved in 'utf-8' codec with charactors other than ascii, then in py27:

1. `# -*- coding: utf-8 -*-` is needed in the script file.
2. `from __future__ import unicode_literals` may also be needed.

PR #7306, opened 2017-12-09
Rewrite layouts and plots in TypeScript

mattpap:
This rewrites `models/{layouts,plots,canvas,grids,scales}`.

addresses #6481


mattpap:
I intended to rewrite `plot_canvas*` in this PR as well, but that's a lot of work, so I will do it in a future one.

Issue #7298, opened 2017-12-08
Remove dependency on PhantomJS when other webdriver is used


ghost-of-turek:
### The problem
Recently I made a switch to Headless Chrome, as PhantomJS is basically dead software at this point.

I manually create a Chrome webdriver with some options and pass it as an argument to export_svgs. But despite that, the function `detect_phantomjs()` is still called and raises an exception. I only noticed it because I cleaned up the container and it doesn't have PhantomJS installed anymore. This forces me to still have PhantomJS installed in the container, even though it is not ever used.

### Suggested resolution:
In `bokeh.io.export` remove:
```
detect_phantomjs()
```
change:
```
web_driver = driver if driver is not None else webdriver.PhantomJS(service_log_path=devnull)
```
to
```
web_driver = driver if driver is not None else get_default_webdriver()
```
and let the `get_default_webdriver()` method call `detect_phantomjs()` and then create a new instance of `webdriver.PhantomJS`.

bryevdv:
@ghost-of-turek This seems reasonable, would you mind submitting the small PR to implement this?

ghost-of-turek:
Sure, I will try to get around to it this weekend.

PR #7311, opened 2017-12-11
Remove dependency on PhantomJS when the default webdriver is not used

ghost-of-turek:
Resolves [#7298](https://github.com/bokeh/bokeh/issues/7298).


bryevdv:
Thanks for the PR @ghost-of-turek !

PR #7315, opened 2017-12-13
Rewrite embed in TypeScript

mattpap:
addresses #6481


bokehjs/src/coffee/embed.ts
```diff
@@ -0,0 +1,336 @@
...
+declare var Jupyter: Jupyter | undefined
+
+// Matches Bokeh CSS class selector. Setting all Bokeh parent element class names
+// with this var prevents user configurations where css styling is unset.
+export const BOKEH_ROOT = "bk-root"
+
+function _handle_notebook_comms(this: Document, receiver: Receiver, msg?: Message | null): void {
+  if (msg != null) {
+    if (msg.buffers.length > 0)
+      receiver.consume((msg.buffers[0] as any).buffer) // XXX: buffers[i] is a tuple
```
mattpap: @bryevdv, can you look at this function and this line in particular, because either this code or my understanding of involved types is wrong.

bryevdv: @mattpap the message is a jupyter comms message which wraps another layer around things:
screen shot 2017-12-13 at 10 57 25

This code is to discern between json vs binary messages given how jupyter comms messages are structured.

bryevdv: when we send a binary array it is a single `DataView` in the `.buffers` list:
screen shot 2017-12-13 at 11 01 57

PR #7316, opened 2017-12-13
Sort app names

### LINKED ISSUE #7289: Sort application names in index page
When I create a Bokeh server application I provide a dictionary of named applications:

```python
Server({'foo': ..., 'bar': ...})
```

When I go to the index.html page of my bokeh server it would be convenient to have these sorted alphabetically.  That does not appear to currently be the case.

image

Is there a way to acheive this?

Maggie-M:
issues: fixes #7289


mrocklin:
Thanks @Maggie-M !

PR #7322, opened 2017-12-14
make docs version warning banner stick to the top always

### LINKED ISSUE #7318: Make Old/Dev docs warning banner always float
The banner only stays at the very top and many people miss it if linked in from external sites.

bryevdv:
issues: fixes #7318
 tests added / passed


bryevdv:
Can currently be seen on 0.11 docs, e.g.

https://bokeh.pydata.org/en/0.11.0/docs/user_guide/charts.html#box-plots

Other versions will show up when CF static cache clears tonight.

PR #7324, opened 2017-12-15
Bryanv/more cleanup

bryevdv:
This PR renames and applies the API decorators `general` and `dev` for public APIs across the entire codebase. Also updates the `bokeh.application` subpackage to have consistent module template and testing conventions.


bryevdv:
cc @mattpap I'll probably merge this tonight unless there is objection.

Issue #7319, opened 2017-12-14
Themes not applying in the notebook


jlstevens:
Using current bokeh master, I have noticed that themes do not seem to apply in the notebook environment. After defining the following theme in the notebook:

```python
theme_json = {
'attrs' : {
    'Figure' : {
        'background_fill_color': '#2F2F2F',
        'border_fill_color': '#2F2F2F',
        'outline_line_color': '#444444',
    },
    'Grid': {
        'grid_line_dash': [6, 4],
        'grid_line_alpha': .3,
    },
    'Title': {
        'text_color': 'white'
    }
}
}
```

the following example does not apply the theme:

```python
import numpy as np
from bokeh.themes.theme import Theme

from bokeh.layouts import gridplot
from bokeh.plotting import figure, show, output_notebook, curdoc
output_notebook()

x = np.linspace(0, 4*np.pi, 100)
y = np.sin(x)

TOOLS = "pan,wheel_zoom,box_zoom,reset,save,box_select"

p1 = figure(title="Legend Example", tools=TOOLS)

p1.circle(x,   y, legend="sin(x)")
p1.circle(x, 2*y, legend="2*sin(x)", color="orange")
p1.circle(x, 3*y, legend="3*sin(x)", color="green")

p2 = figure(title="Another Legend Example", tools=TOOLS)

p2.circle(x, y, legend="sin(x)")
p2.line(x, y, legend="sin(x)")

p2.line(x, 2*y, legend="2*sin(x)", line_dash=(4, 4), line_color="orange", line_width=2)

p2.square(x, 3*y, legend="3*sin(x)", fill_color=None, line_color="green")
p2.line(x, 3*y, legend="3*sin(x)", line_color="green")


curdoc().theme = Theme(json=theme_json)
show(gridplot(p1, p2, ncols=2, plot_width=400, plot_height=400))  # open a browser
```

![image](https://user-images.githubusercontent.com/890576/34005594-4fb5568c-e0f3-11e7-9a76-90021768e20a.png)

If it is agreed that this is unexpected behavior (i.e a bug), I have found a small diff that fixes the issue. I am happy to submit a PR with what I believe to be the fix which then results in the expected result:

![image](https://user-images.githubusercontent.com/890576/34005657-8a1668ca-e0f3-11e7-9efc-5f6ed9fce9ab.png)

jlstevens:
With the PR merged I am happy to see this issue closed.

PR #7325, opened 2017-12-15
Fix for themes in notebooks

jlstevens:
This PR proposes a small change that seems to fix the issue described in #7319 where bokeh themes fail to apply in the notebook environment. 

I've had this diff outstanding for a while in my local copy of bokeh and I have not yet noticed any issues in other contexts (e.g bokeh server). That said, here might be other implications/issues that I am currently unaware of so maybe this simple fix isn't sufficient.


bryevdv:
This seems correct to me. `doc` is only determined in order to restore it after the "model in new empty doc" context manager is done. The theme should really be applied to that new doc.

jlstevens:
Thanks!

PR #7334, opened 2017-12-19
Compute correct attachment side when value = "vertical"

### LINKED ISSUE #7291: HoverTool with attachment="vertical" incorrectly positions tooltips
Bokeh 0.12.13

I think it happened after the recent change of the coordinates origin.
When you create `HoverTool` with `attachment="vertical"`, the tooltip is supposed to be above a point when the point is below the center of the plot, and below the point when it's above the center of the plot. Right now, the directions are swapped.

bryevdv:
Recent change to remove "view" coordinates and use screen coordinates throughout did not fix this codepath up.

- [x] issues: fixes #7291
- [x] tests added / passed


bryevdv:
This should be uncontroversial, I am going to go ahead and merge now

bokehjs/src/coffee/models/annotations/tooltip.coffee
```diff
@@ -2,6 +2,16 @@ import {Annotation, AnnotationView} from "./annotation"
 import {div, show, hide, empty} from "core/dom"
 import * as p from "core/properties"
 
+export compute_side = (attachment, sx, sy, hcenter, vcenter) ->
+  switch attachment
+    when "horizontal"
+      side = if sx < hcenter then 'right' else 'left'
+    when "vertical"
+      side = if sy > vcenter then 'below' else 'above'
```
mattpap: This doesn't look correct to me, but maybe I'm misunderstanding what `above` and `below` actually mean here.

bryevdv:
Actually, the meaning of "side" seems to be inconsistent now. "left" means left of the point being pointed to, but "below" means below to tooltip (i.e. above the point pointed to)

bokehjs/src/coffee/models/annotations/tooltip.coffee
bryevdv: It wasn't, I just pushed the real fix

bokehjs/src/coffee/models/annotations/tooltip.coffee
```diff
@@ -70,11 +75,11 @@ export class TooltipView extends AnnotationView
         @el.classList.add("bk-right")
         left = sx - @el.offsetWidth - arrow_size
         top = sy - @el.offsetHeight/2
-      when "above"
+      when "below"
         @el.classList.add("bk-above")
         top = sy + (@el.offsetHeight - @el.clientHeight) + arrow_size
         left = Math.round(sx - @el.offsetWidth/2)
-      when "below"
+      when "above"
```
bryevdv: @mattpap this was the real problem. With this change, side is always relative to the point being annotated

bokehjs/src/coffee/models/annotations/tooltip.coffee
```diff
@@ -61,6 +64,8 @@ export class TooltipView extends AnnotationView
 
     show(@el) # XXX: {offset,client}Width() gives 0 when display="none"
 
+    # slightly confusing: side "left" is relative to point that is being annotated
+    # but CS class "bk-left" is relative to the tooltip itself
```
mattpap: That's because `bk-left` relates to the arrow, so it's reversed in meaning to that of `attachment`'s `left`. Anyway, it's confusing.

PR #7339, opened 2017-12-20
remove bokeh.client 'app' examples

### LINKED ISSUE #6887: Remove all server examples that use session.loop_until_closed
Another user having issues using `bokeh.client` to run apps "outside" a Bokeh server. Text of mailing list response:

> There are some current issues with using bokeh.client (i.e. push_session and loop_until_closed) with some features (e.g. events and streaming)
>
> However, even without these current bugs, I cannot stress enough (and evidently have not been able to stress enough, despite trying to) that this usage of bokeh.client is absolutely, 100% NOT ADVISED. Using bokeh.client in this way (with the app running "outside" the server) has inherent drawbacks that will never be overcome (described in the user's guide).
>
> Whether you are running "bokeh serve app.py" or using a FunctionHander and embedding a Bokeh server programmatically in a script or a notebook, it is always, without fail, better to run the app IN a Bokeh server.
>
> The *only* use I would ever advise using bokeh.client for is to pull specific session for an app running normally a Bokeh server, for the purpose of "tweaking it". This might be useful e.g. for serving customized versions of an app to different users looking at a Flask or Django page. 
>
> FYI I am basically almost to the point where I am going to remove every example in the repo that uses "session.loop_until_closed" under the premise that ever demonstrating it in the first place was a great mistake. 

I'd like to discuss that last point. Specifically, I'd like to:

* completely remove `examples/plotting/server` and `examples/model/server`
* demonstrate a good "tweak a session" use of `bokeh.client`
* make `session.loop_until_closed` print a loud warning that it is not advised for any real (non-toy) usage

cc @bokeh/dev

bryevdv:
issues: fixes #6887


examples/app/population.py
```diff
@@ -0,0 +1,80 @@
...
+locations = sorted(df.Location.unique())
+groups = list(df.AgeGrp.unique())
+
+# pyramid plot of age groups by gender
+
+ages = ColumnDataSource(data=dict(AgeGrp=[], Sex=[], Value=[]))
+
+gender_transform = CustomJSTransform(args=dict(source=ages), func="", v_func="""
+    var val = new Float64Array(xs.length)
+    for (i = 0; i < xs.length; i++) {
```
mattpap: `var i`.

Issue #7327, opened 2017-12-15
output_png and output_svgs fails in the Settings.phantomjs_path( ) function.  


SweetsRepo:
#### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)

Bokeh:
  bokeh==0.12.13
Related Dependencies:
  Pillow==4.3.0
  selenium==3.8.0
  phantomjs-prebuilt==2.1.16
Python:
  python==3.6.3
OS:
  Windows 7 and Windows 10
Run in: 
  Cmder and Command Prompt
 
#### Description of expected behavior and the observed behavior
Expected Behavior:
  When calling the export_png function the plot should be exported to the filename given. 

Observed Behavior 1:
  Function fails because BOKEH_PHANTOMJS_PATH environment variable does not get set on install of 
  bokeh or phantomjs. Related traceback is shown in Traceback 1 below. 

  Manually set variable to the install location of phantomjs:
  set BOKEH_PHANTOMJS_PATH=C:\Users\Christopher\AppData\Roaming\npm\phantomjs.exe

Observed Behavior 2:
  Function fails due to Popen([phantomjs_path, '--version'], stdout=PIPE, stderr=PIPE) returning OSError: 
  [WinError 193]. Related traceback is shown in Traceback 2 below. 

#### Complete, minimal, self-contained example code that reproduces the issue
```
from bokeh.io import export_png
from bokeh.plotting import figure, output_file, show

# prepare some data
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]

# output to static HTML file
output_file("lines.html")

# create a new plot with a title and axis labels
p = figure(title="simple line example", x_axis_label='x', y_axis_label='y')

# add a line renderer with legend and line thickness
p.line(x, y, legend="Temp.", line_width=2)

export_png(p, filename="plot.png")
```
#### Stack traceback and/or browser JavaScript console output
Traceback 1:
D:\PythonWorkspace\bokeh_bug>python bug.py
Traceback (most recent call last):
  File "C:\Users\Christopher\AppData\Local\Programs\Python\Python36-32\lib\site-packages\bokeh\util\dependencies.py", line 62, in detect_phantomjs
    proc = Popen([phantomjs_path, "--version"], stdout=PIPE, stderr=PIPE)
  File "C:\Users\Christopher\AppData\Local\Programs\Python\Python36-32\lib\subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
[...]

PR #7340, opened 2017-12-20
Sweet phantomjs dependency

SweetsRepo:
- [#7327 ] issues: fixes 
- [NA] tests added / passed - Can add, just need to know preferred test directory I should add to.
- [NA] release document entry (if new feature or API change)

### DIFF
--- a/bokeh/util/dependencies.py
+++ b/bokeh/util/dependencies.py
@@ -3,6 +3,8 @@
 '''
 from importlib import import_module
 import logging
+import shutil
+import sys
 from subprocess import Popen, PIPE
 
 from ..settings import settings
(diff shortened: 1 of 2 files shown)


bokeh/util/dependencies.py
```diff
@@ -53,14 +54,20 @@ def import_required(mod_name, error_msg):
...
-        phantomjs_path = "phantomjs"
+
+    try:
+        phantomjs_path = shutil.which('phantomjs')
+    # Python 2 relies on Environment variable in PATH
+    except AttributeError:
+        if settings.phantomjs_path() is not None:
+            phantomjs_path = settings.phantomjs_path()
+        else:
+            phantomjs_path = "phantomjs"
```
bryevdv: I think the env var setting should always be available as an explicit override, even in Python 3. Otherwise if it is `None`, then use `shutil.which`

SweetsRepo: I'm amicable to that. I would anticipate it looking as follows:
```python 
 def detect_phantomjs():
      '''Detect if PhantomJS is avaiable in PATH.'''
      if settings.phantomjs_path() is not None:
           phantomjs_path = settings.phantomjs_path()
      else:
          try:
               phantomjs_path = shutil.which('phantomjs')
           # Python 2 relies on Environment variable in PATH - attempt to use as follows
          except AttributeError:
               phantomjs_path = "phantomjs"
```

Does that seem good to you?

bryevdv: Yes that looks perfect

SweetsRepo:
I corrected a trailing whitespace issue flagged by TravisCI. On the next test the following error was given.

> 
The command "scripts/ci/install:test" failed and exited with 1 during .

bryevdv:
That usually indicates a spurious network install failure. I've restarted the affected build

bryevdv:
@SweetsRepo did you want to change to using `communicate` instead of `wait` in this PR as well?

SweetsRepo:
@bryevdv No, I looked into it further and don't believe it's an issue. I see no reason why checking the version of phantomjs would ever hang. Popen.communicate( ) is more useful for calls to processes which have a potential to hang or long operation times.

bryevdv:
@SweetsRepo Sounds good, will plan to get this merged today then. Thanks for the contribution!

PR #7345, opened 2017-12-26
Fix minor typos in bokeh server docs

### LINKED ISSUE #7344: Minor typos in bokeh server docs code example
The code example in section "[Updating From Threads](http://bokeh.pydata.org/en/latest/docs/user_guide/server.html#updating-from-threads)" in the bokeh server docs contain a couple of typos:

- "Bokeh session allback"
- "[...] all threads see then same document."

I'll submit a PR to fix these in a minute.

maxalbert:
Fixes #7344

### DIFF
--- a/sphinx/source/docs/user_guide/server.rst
+++ b/sphinx/source/docs/user_guide/server.rst
@@ -455,11 +455,11 @@ below:
 
     from tornado import gen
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the catch and fix @maxalbert !

PR #7352, opened 2017-12-29
Rewrite models/tools (except toolbars) in TypeScript

mattpap:
I don't want to even count the time I burned on this.

addresses #6481


bokehjs/src/coffee/models/tools/actions/reset_tool.ts
```diff
@@ -0,0 +1,30 @@
...
+  doit(): void {
+    this.plot_view.clear_state()
+    this.plot_view.reset_range()
+    this.plot_view.reset_selection()
+    this.plot_model.plot.trigger_event(new Reset())
+  }
+}
+
+export class ResetTool extends ActionTool {
+  reset_size: boolean
```
bryevdv: as an aside this could probably go away, since the resize tool went away

bokehjs/src/coffee/models/tools/actions/save_tool.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export class SaveTool extends ActionTool {
+  tool_name = "Save"
+  icon = "bk-tool-icon-save"
+}
+
+SaveTool.prototype.type = "SaveTool"
+
+SaveTool.prototype.default_view = SaveToolView
```
bryevdv: All of these actions are very similar I wonder if they could be created generatively like markers

bokehjs/src/coffee/models/tools/actions/zoom_out_tool.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+ZoomOutTool.prototype.type = "ZoomOutTool"
+
+ZoomOutTool.prototype.default_view = ZoomOutToolView
+
+ZoomOutTool.define({
+  factor:     [ p.Percent,    0.1    ],
+  dimensions: [ p.Dimensions, "both" ],
+})
```
bryevdv: zoom in/out only tools only differ by their icon, and a minus sign :)

bokehjs/src/coffee/models/tools/actions/undo_tool.ts
```diff
@@ -0,0 +1,28 @@
...
+  icon = "bk-tool-icon-undo"
+}
+
+UndoTool.prototype.type = "UndoTool"
+
+UndoTool.prototype.default_view = UndoToolView
+
+UndoTool.override({
+  disabled: true,
+})
```
bryevdv: FWIW undo/redo don't seem to be working for me on `color_scatter.py`

bokehjs/src/coffee/models/tools/actions/zoom_out_tool.ts
mattpap: This is how it's implemented. It's not my concern in this PR, unless it would affect types.

bokehjs/src/coffee/models/tools/actions/save_tool.ts
mattpap: It can be shortened, but without macros there isn't much that can be done here. Though I would consider getting rid of `default_view` (at least by default) and deriving it in `base`.

bokehjs/src/coffee/models/tools/actions/undo_tool.ts
mattpap: Yeah, it kind of works, but also doesn't (maybe one step works).

bryevdv:
I did not go line by line over entire PR but I gave it a decent scan and ran about 30 tool heavy examples locally. Examoples report also looks perfect, so I'm inclined to merge this and get it into a dv build sooner rather than later.

bokehjs/src/coffee/models/tools/actions/reset_tool.ts
mattpap: I have a feeling that I removed this before, but maybe it was something else related to `ReiszeTool`. Anyway I started #7354 for this.

bokehjs/src/coffee/models/tools/actions/undo_tool.ts
mattpap: I will look into this, #7355.

mattpap:
> I did not go line by line over entire PR (...)

This may be a good thing, because this was the worse rewrite I had so far. I will need to revise this when I get to `core/ui_events` and then when I rewrite glyph/graph renderers (there is a lot of `any`s due to lack of typings).

PR #7361, opened 2018-01-02
Fix errant scroll to end on DataTable with no selection

### LINKED ISSUE #7139: DataTable automatically scrolls down when app loads
Hi!
I recently tried to update my bokeh version to 0.12.10 (I was still in 0.12.4) and I noticed something that looks to me like a regression. When the app loads or updates, DataTables automatically scroll down (which was not the case in 0.12.4).
You will find attached a minimal example for this behavior.
Thanks (again) for the great library!

Julien
[minimal_example.zip](https://github.com/bokeh/bokeh/files/1430247/minimal_example.zip)

bryevdv:
This problem was caused by applying the scroll computation in the case where there was an empty selection. In this instance, the scroll index was computed using `Math.min.apply(null, [])` which is `Infinity`, resulting in a scroll to the very end. The moves the logic for computing the index in to a method on the model (with the fix for empty selections) and places it under test.

- [x] issues: fixes #7139
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/tables/data_table.coffee
+++ b/bokehjs/src/coffee/models/widgets/tables/data_table.coffee
@@ -129,9 +129,10 @@ export class DataTableView extends WidgetView
     # the selection is already in the viewport so that selecting from the
     # datatable itself does not re-scroll.
     cur_grid_range = @grid.getViewport()
-    if @model.scroll_to_selection and not any(permuted_indices, (i) -> cur_grid_range.top <= i <= cur_grid_range.bottom)
-      min_index = Math.max(0, Math.min.apply(null, permuted_indices) - 1)
-      @grid.scrollRowToTop(min_index)
+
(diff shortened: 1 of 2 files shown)


bokehjs/src/coffee/models/widgets/tables/data_table.coffee
```diff
@@ -230,3 +231,12 @@ export class DataTable extends TableWidget
   @internal {
     default_width:        [ p.Number, 600   ]
   }
+
+  get_scroll_index: (grid_range, selected_indices) ->
+    if not @scroll_to_selection or selected_indices.length == 0
+      return null
+
+    if not any(selected_indices, (i) -> grid_range.top <= i <= grid_range.bottom)
+      return Math.max(0, Math.min.apply(null, selected_indices) - 1)
```
mattpap: This could be simplified to `Math.min(selected_points...)` in CS and `Math.min(...selected_points)` in TS.

bryevdv:
@mattpap made your suggested change

PR #7363, opened 2018-01-02
Support updates to factor range factors

### LINKED ISSUE #7083: FactorRange regression: plot does not get updated when factors change.
#### Software version info (bokeh, python, notebook, OS, browser, any other relevant packages):

Bokeh (affected) 0.12.10, 0.12.9, 0.12.7, 0.12.10rc2 
Bokeh (working as expected): 0.12.6

OS: docker image [lukauskas/bokeh](https://hub.docker.com/r/lukauskas/bokeh/)
Python: 3.6

Browser: Chrome 61.0.3163.100 

Since 0.12.7 onwards, when updating factors of a FactorRange, the plot does not get updated any more.

In the example below, I create a plot with a factor x axis containing factors ['a', 'b', 'c'].
This plot has a circle drawn at the location x='a', y=1.
When the button is pressed, the factors on the axis are changed to ['d', 'c', 'b', 'a'].
I would expect the circle to move within the image to the new location of the factor 'a'.
Instead, the circle stays in the same position, but the factors update. So now the circle is effectively at the location x='d', y=1.

Here's the code for a minimal example to reproduce it (run it via `bokeh serve`).

```
from bokeh.models import FactorRange, Button

button = Button()
x_range = FactorRange(factors=['a', 'b', 'c'])
fig = figure(x_range=x_range)
fig.circle(['a'], [1], size=20, color='navy')

def change_range():
    print('Changing range')
    x_range.factors = ['d', 'c', 'b', 'a']

button.on_click(change_range)

l = layout([[button],
            [fig]]
)
curdoc().add_root(l)
```

If you paste this code to a file `app/main.py` then you can use the following `Dockerfile` to reproduce everything:

```
ARG VERSION=0.12.10
FROM lukauskas/bokeh:${VERSION}
COPY ./app /app
```

To run it:
```
docker build --build-arg VERSION="0.12.10" -t bokeh .
docker run --rm -p 5100:5100 --name bokeh_cont bokeh
```

Access the plot at `http://127.0.0.1:5100`.
Replace the version number if needed.

After button click in 0.12.10: 

image

After button click in 0.12.6 (expected behaviour):

image

### LINKED ISSUE #7084: FactorRange regression: BokehJS crashes if data has values not in FactorRange factors.
This is probably related to  #7083 but at the same time sufficiently different, hence new issue number.

Bokeh (crashing) 0.12.10, 0.12.7
Bokeh (not crashing): 0.12.6

OS: docker image [lukauskas/bokeh](https://hub.docker.com/r/lukauskas/bokeh/)
Python: 3.6

Browser: Chrome 61.0.3163.100

Since 0.12.7 onwards, bokehjs would crash if asked to plot something on a factor axis, when given a value that is not in `factors`.
Specifically, in the example below I create a figure with a FactorRange having factors `'a'`, `'b'` and `'c'`.

I then proceed to draw a circle from `ColumnDataSource`. Now the tricky bit is that my ColumnSource also has an item at the location `x='d'`, `y=2`. 

Now I am not sure what the expected behaviour should be, but is definitely not to crash silently.
The options are:

1) Display the plot without the data that is not in factors (this would be my preferred option, also how it worked in 0.12.6). When combined with  #7083 this would allow to 'zoom in' into factor ranges without changing the source.

or, at least:

2) Throw an useful error message so it is easier to find what the problem is.

```
from bokeh.models import FactorRange, Button, ColumnDataSource

button = Button()
x_range = FactorRange(factors=['a', 'b', 'c'])

source = ColumnDataSource({'x': ['a', 'd'], 'y': [1, 2]})

fig = figure(x_range=x_range)
fig.circle(x='x', y='y', size=20, color='navy', source=source)

l = layout([[fig]])
curdoc().add_root(l)
```

To reproduce it either run the code directly via `bokeh serve`, or place it `app/main.py` and use the aforementioned docker image with this `Dockerfile`:

```
ARG VERSION=0.12.10
FROM lukauskas/bokeh:${VERSION}
COPY ./app /app
```

```
docker build --build-arg VERSION="0.12.10" -t bokeh .
docker run --rm -p 5100:5100 --name bokeh_cont bokeh
```

#### browser JavaScript console output (bokeh=0.12.10)
```
Error rendering Bokeh items  TypeError: Cannot read property 'value' of undefined
    at e.r.FactorRange.e._lookup (factor_range.js:288)
    at e.r.FactorRange.e.synthetic (factor_range.js:223)
    at e.<anonymous> (factor_range.js:239)
    at e.r.FactorRange.e.v_synthetic (factor_range.js:242)
    at e.r.GlyphView.r.set_data (glyph.js:301)
    at e.r.GlyphRendererView.e.set_data (glyph_renderer.js:160)
    at e.r.GlyphRendererView.e.initialize (glyph_renderer.js:87)
    at e.t (view.js:42)
    at e [as constructor] (dom_view.js:21)
    at e [as constructor] (renderer.js:25)
```

In short `this._mapping['d']` returns `undefined` (as it is not in the factors list), which then does not have a `value` property and hence the `TypeError`.

bryevdv:
- [x] issues: fixes #7084 fixes #7083
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/ranges/factor_range.ts
+++ b/bokehjs/src/coffee/models/ranges/factor_range.ts
@@ -128,12 +128,12 @@ export class FactorRange extends Range {
   initialize(attrs: any, options: any): void {
     super.initialize(attrs, options)
     this._init()
-    this.connect(this.properties.factors.change, () => this._init())
-    this.connect(this.properties.factor_padding.change, () => this._init())
-    this.connect(this.properties.group_padding.change, () => this._init())
-    this.connect(this.properties.subgroup_padding.change, () => this._init())
-    this.connect(this.properties.range_padding.change, () => this._init())
-    this.connect(this.properties.range_padding_units.change, () => this._init())
+    this.connect(this.properties.factors.change, () => this.reset())
+    this.connect(this.properties.factor_padding.change, () => this.reset())
+    this.connect(this.properties.group_padding.change, () => this.reset())
+    this.connect(this.properties.subgroup_padding.change, () => this.reset())
+    this.connect(this.properties.range_padding.change, () => this.reset())
+    this.connect(this.properties.range_padding_units.change, () => this.reset())
   }
 
   reset(): void {
--- a/bokehjs/src/coffee/models/ranges/factor_range.ts
+++ b/bokehjs/src/coffee/models/ranges/factor_range.ts
@@ -142,13 +142,25 @@ export class FactorRange extends Range {
   }
 
   protected _lookup(x: any): number {
-    if (x.length == 1)
-      return (this._mapping as L1Mapping)[x[0]].value
-    else if (x.length == 2)
-      return (this._mapping as L2Mapping)[x[0]].mapping[x[1]].value
-    else if (x.length == 3)
-      return (this._mapping as L3Mapping)[x[0]].mapping[x[1]].mapping[x[2]].value
-    else
+    if (x.length == 1) {
+      const m = this._mapping as L1Mapping
+      if (!m.hasOwnProperty(x[0])) {
+        return NaN
+      }
+      return m[x[0]].value
+    } else if (x.length == 2) {
+      const m = this._mapping as L2Mapping
+      if (!m.hasOwnProperty(x[0]) || !m[x[0]].mapping.hasOwnProperty(x[1])) {
+        return NaN
+      }
+      return m[x[0]].mapping[x[1]].value
+    } else if (x.length == 3) {
+      const m = this._mapping as L3Mapping
+      if (!m.hasOwnProperty(x[0]) || !m[x[0]].mapping.hasOwnProperty(x[1]) || !m[x[0]].mapping[x[1]].mapping.hasOwnProperty(x[2]))  {
+        return NaN
+      }
+      return m[x[0]].mapping[x[1]].mapping[x[2]].value
+    } else
       return undefined as never
   }
 
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.coffee
@@ -75,6 +75,12 @@ export class GlyphRendererView extends RendererView
     @connect(@model.properties.view.change, () -> @set_data())
     @connect(@model.view.change, () -> @set_data())
 
+    for name, rng of @plot_model.frame.x_ranges
+      @connect(rng.change, () -> @set_data())
+
+    for name, rng of @plot_model.frame.y_ranges
+      @connect(rng.change, () -> @set_data())
+
     @connect(@model.glyph.transformchange, () -> @set_data())
 
   have_selection_glyphs: () -> @selection_glyph? && @nonselection_glyph?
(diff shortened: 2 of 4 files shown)


bryevdv:
cc @hsparra @philippjfr @jlstevens  if you guys have "updating factor range" examples (that currently error because of invalid mappings) it would be helpful to test them with this PR. It might "solve" things entirely, or or might solve things but with a quick flicker during update.

bokehjs/src/coffee/models/ranges/factor_range.ts
```diff
@@ -142,14 +142,19 @@ export class FactorRange extends Range {
...
+      if (x.length == 1)
+        return (this._mapping as L1Mapping)[x[0]].value
+      else if (x.length == 2)
+        return (this._mapping as L2Mapping)[x[0]].mapping[x[1]].value
+      else if (x.length == 3)
+        return (this._mapping as L3Mapping)[x[0]].mapping[x[1]].mapping[x[2]].value
+      else
+        return undefined as never
+    }
+    catch (e) {
```
mattpap: At minimum this needs a comment regarding what this code guards against. However, this code just asks for causing problems in future. I would handle missing factors explicitly at each step.

bryevdv:
Actually a full solution probably involves fixing #7083 as well (has a proposed diff in the issue)

bryevdv:
OK I pushed another commit to also make sure that when `FactorRange` factors are changed, that the range emits a `change` signal, and I also hooked up the glyph renderer to trigger a call to `set_data` (to remap screen coords) whenever a range `change` happens.

bokehjs/src/coffee/models/ranges/factor_range.ts
```diff
@@ -142,13 +142,25 @@ export class FactorRange extends Range {
...
+        return NaN
+      }
+      return m[x[0]].mapping[x[1]].value
+    } else if (x.length == 3) {
+      const m = this._mapping as L3Mapping
+      if (!m.hasOwnProperty(x[0]) || !m[x[0]].mapping.hasOwnProperty(x[1]) || !m[x[0]].mapping[x[1]].mapping.hasOwnProperty(x[2]))  {
+        return NaN
+      }
+      return m[x[0]].mapping[x[1]].mapping[x[2]].value
+    } else
```
bryevdv: @mattpap I've updated not to rely on exceptions at all. I am assuming that `||` is guaranteed to short-circuit? (since the tests still pass)

mattpap: > || is guaranteed to short-circuit? 

Yes.

mattpap: > `hasOwnProperty`

Very pedantic, but you are right. JavaScript is so gross.

philippjfr:
Just tested it, I do see some flicker but it's not too noticeable and definitely preferable over a temporarily broken plot. Just to be clear what is the recommended way of updating a ``FactorRange`` now? Should you update the data first and then update the factors or the inverse?

bryevdv:
I would say with this PR that the order is irrelevant

PR #7371, opened 2018-01-04
Connect GraphRenderer to range change events

### LINKED ISSUE #7370: Regression in Graph rendering during zoom
The changes introduced in https://github.com/bokeh/bokeh/pull/7363 seem to have caused a regression when rendering graphs. The issue can be reproduced using the code below (which has been changed only slightly from the documentation example):

```python
import math

N = 8
node_indices = list(range(N))

plot = figure(title="Graph Layout Demonstration", x_range=(-1.1,1.1), y_range=(-1.1,1.1))

graph = GraphRenderer()

    index=node_indices,
    fill_color=Spectral8)

    start=[0]*N,
    end=node_indices)

### start of layout code

plot.renderers.append(graph)

output_file("graph.html", mode='inline')
show(plot)
```

The bug looks like this in action:

graph_odd_bug

When zooming all the nodes snap to the origin and any subsequent mouse move then rerenders them in the correct location (but only if a hover tool is enabled).

Versions:

bokeh==0.12.14.dev2+5.g2b7ec420b

philippjfr:
This PR fixes a regression when zooming on plots with a GraphRenderer. As described in #7370 zooming causes the nodes to snap to the origin. By connecting the ``GraphRenderer.set_data`` method to range change events just as was done for the ``GlyphRenderer`` in #7363 the issue is resolved.

- [x] issues: fixes #7370
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/renderers/graph_renderer.coffee
+++ b/bokehjs/src/coffee/models/renderers/graph_renderer.coffee
@@ -31,6 +31,12 @@ export class GraphRendererView extends RendererView
     @connect(@model.edge_renderer.data_source.inspect, () -> @set_data())
     @connect(@model.edge_renderer.data_source.change, () -> @set_data())
 
+    for name, rng of @plot_model.frame.x_ranges
+      @connect(rng.change, () -> @set_data())
+
+    for name, rng of @plot_model.frame.y_ranges
+      @connect(rng.change, () -> @set_data())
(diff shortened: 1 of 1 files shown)


philippjfr:
Based on some quick JS profiling zoom events on a large (5000 node/edge) graph go from ~60ms to about ~80ms with this change.

PR #7379, opened 2018-01-07
Bryanv/6681 df index streaming

### LINKED ISSUE #6681: Streaming numpy datetime64 data does not work
Related to #6667 - but in addition to that issue the datetime data also needs special treatment. 

My comment on #6667 contains a complete example and also the workaround, but I thought a ticket explicitly mentioning the datetime64 problem would be useful.

bryevdv:
This PR allows for `CDS.stream` to accept more data frames. In particular, DataFrames with unnamed indices would previously fail, since the index would be automatically captured at CDF creation, but would *not* be captured by ``stream``.  Other notes:

* This moves all pandas processing into `CDS._stream` and makes the interface for `PropertyValueDict` always assume a plain dict of named columns (lists/arrays/series/etc). This was necessary to properly process indices, but also makes the `PropertyValueDict` simpler. However, some changes to tests were necessary.

* Although it is not new, it's worth pointing out that with the conversion to CDS, there is no guarantee that streaming a new data frame will preserve the "index-ness" of the CDS index column (see comment below in test). This was true before, but now there are more opportunities to stream from dataframes. Users who need a valid index in the CDS at all times when streaming will have to take care to construct the stream dataframe index correctly (see example below)

- [x] issues: fixes #6681
- [x] tests added / passed

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -6,6 +6,7 @@ from ..core.has_props import abstract
 from ..core.properties import Any, Bool, ColumnData, Dict, Enum, Instance, Int, JSON, List, Seq, String
 from ..model import Model
 from ..util.dependencies import import_optional
+from ..util.serialization import convert_datetime_array
 from ..util.warnings import BokehUserWarning
 
 from .callbacks import Callback
@@ -171,7 +172,6 @@ class ColumnDataSource(ColumnarDataSource):
 
         '''
         _df = df.copy()
-        index = _df.index
         tmp_data = {c: v.values for c, v in _df.iteritems()}
 
         new_data = {}
@@ -180,15 +180,8 @@ class ColumnDataSource(ColumnarDataSource):
                 k = "_".join(k)
             new_data[k] = v
 
-        if index.name:
-            new_data[index.name] = index.values
-        elif index.names:
-            try:
-                new_data["_".join(index.names)] = index.values
-            except TypeError:
-                new_data["index"] = index.values
-        else:
-            new_data["index"] = index.values
+        index_name = ColumnDataSource._df_index_name(df)
+        new_data[index_name] = _df.index.values
         return new_data
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


bryevdv:
With this PR, the modified example from the original issue now works:
```
import six
from bokeh.io import curdoc

import pandas as pd
import numpy as np

def modify_doc(doc):
    """

    :param Document doc:
    :return:
    """
    df = pd.DataFrame(
        index=pd.date_range('now', periods=30, freq='T'),
        columns=['A', 'B', 'C'],
        data=np.cumsum(np.random.standard_normal((30,3)), axis=0)
    )

    cds = ColumnDataSource(df)
    figure = Figure(x_axis_type="datetime")
    figure.line("index", "A", source=cds)
    figure.line("index", "B", source=cds)
    figure.line("index", "C", source=cds)

    doc.add_root(figure)

    store = {'df': df}
    def callback():
        df = store['df']
        new_df = pd.DataFrame(
            index=df.index + pd.to_timedelta('30m'),
            columns=df.columns,
            data=np.random.standard_normal((30,3))
        )
        cds.stream(new_df)        # This is what now works
        store['df'] = new_df

    doc.add_periodic_callback(callback, 1 * 1000)

modify_doc(curdoc())
```
screen shot 2018-01-07 at 16 10 45

Note that in constructing `new_df` care was taken to make new index values that were actually unique wrt to the previously streamed data.

bokeh/models/tests/test_sources.py
```diff
@@ -289,15 +323,17 @@ def stream_wrapper(*args, **kwargs):
...
-                                                c=np.array([30])))
+                                                c=np.array([30]),
+                                                index=np.array([0])))
 
         self._assert_equal_dicts_of_arrays(dict(ds.data),
                                            dict(a=np.array([10, 11]),
                                                 b=np.array([20, 21]),
-                                                c=np.array([30, 31])))
+                                                c=np.array([30, 31]),
+                                                index=np.array([0, 0])))
```
bryevdv: Here is an example of an index column in a CDS getting duplicates. This is not new, and I don't think there is anything that can be done about it. If users want the CDS index column to really be an index with unique valuea at all times, that is up to them to enforce. Though perhaps I could add documentation support to call out the possibility explicitly.

bokeh/models/sources.py
```diff
@@ -208,6 +201,38 @@ def _data_from_groupby(group):
...
+        Returns:
+            str
+
+        '''
+        if df.index.name:
+            return df.index.name
+        elif df.index.names:
+            try:
+                return "_".join(df.index.names)
+            except TypeError:
```
mattpap: Under what circumstances this happens?

bryevdv: I call it "unnamed multi index" not sure if there is some other proper name for it. There's a test for it in any case: https://github.com/bokeh/bokeh/pull/7379/files#diff-16c9ca614fbc1d71df3e42fa8df21265R187

bryevdv: screen shot 2018-01-08 at 09 32 53

mattpap: I see. OK.

bryevdv:
OK to merge?

mattpap:
@bryevdv, sure.

PR #7383, opened 2018-01-08
Decaffeinate bokehjs

mattpap:
This uses `decaffeinate` to rewrite bokehjs (*.coffee) in ES6, which is the syntactical basis for TypeScript. The translation isn't fully automatic, but is fast enough. The output is much better than from CoffeeScript's compiler and thus usable for further translation to TypeScript. Generated files are checked by TS for syntactical compliance, but no typechecking is performed (so far).

addresses #6481


mattpap:
Keeping Better Code Hub aside, this actually works. There are no compilation errors, all tests pass and there are no image differences. The unfortunate part is that TypeScript doesn't support generating declaration files for *.js files (even under `checkJs`), so I have to use the old good trick with renaming files to *.ts under tree_ts and ignoring certain classes of errors (can be viewed anyway behind `--filter`).

This way we have about 115 files transitioned to ES6 and partially (marginally) typed. This leaves us with api (which I'm 75% done rewriting), widgets (100% ES6 and about 15% typed) and tables (nothing done yet). Given good progress in this PR, I will finish api, rename *.ts -> *.js in widgets, so that we have it done similarly to what we have here, and use the approach of this PR in tables. This way we can have `src/` coffee-free next week. Still we will have about 150 files to type, but this can be done much more gradually when syntactical changes are not in the picture. Then I can use the approach of this PR in tests. Typing tests will fully make sense only when source code is typed.

mattpap:
@philippjfr, can you test this PR whether changes of #7392 were correctly transferred?

philippjfr:
@mattpap Just tried it, seems to work just fine.

mattpap:
@philippjfr, OK. Thanks.

mattpap:
I renamed all *.js files to *.ts, to simplify future merges. This way we will have a single file extension across the repository, even though previously *.js files aren't fully typed. To distinguish fully typed from partially typed files, former *.js got `/* XXX: partial */` marker as the first line of a file. This way type errors in those files are filtered out, though syntactical and other errors are kept. After a partial file is fully typed, this marker will be removed.

PR #7395, opened 2018-01-10
Rewrite api in TypeScript

mattpap:
Rewrites api/charts in TypeScript and api/plotting in ES6. Also simplifies the build by removing coffeescript from the pipeline.

addresses #6481


mattpap:
I will go ahead and merge this to get build changes in.

PR #7396, opened 2018-01-10
add polyfill for Number.isInteger

### LINKED ISSUE #7373: "Number.isinteger()" javascript error of Slider in Internet Explorer
I've an bokeh app runs by `bokeh serve app`.  It works under Chrome, and Edge, but not Internet Explorer. Do anybody know why?

In chrome or edge, the plot is like this:
<p><image src="https://user-images.githubusercontent.com/4179106/34599026-599f20b2-f22b-11e7-8163-2fa636570c27.png" width=400/></p>

But in Internet Explorer, it is like (no points showed up):
<p><image src="https://user-images.githubusercontent.com/4179106/34599054-7115be04-f22b-11e7-861f-06c53a436bfa.png" width=400/></p>

And the console keeps getting error messages like:
```cmd
2018-01-05 15:13:12,991 received invalid integer in pong b''
  File "C:\Anaconda3\lib\site-packages\bokeh\server\views\ws.py", line 161, in on_pong
    self.latest_pong = int(codecs.decode(data, 'utf-8'))
ValueError: invalid literal for int() with base 10: ''
2018-01-05 15:13:12,995 received invalid integer in pong b''
  File "C:\Anaconda3\lib\site-packages\bokeh\server\views\ws.py", line 161, in on_pong
    self.latest_pong = int(codecs.decode(data, 'utf-8'))
ValueError: invalid literal for int() with base 10: ''
2018-01-05 15:13:50,016 received invalid integer in pong b''
  File "C:\Anaconda3\lib\site-packages\bokeh\server\views\ws.py", line 161, in on_pong
    self.latest_pong = int(codecs.decode(data, 'utf-8'))
ValueError: invalid literal for int() with base 10: ''
```

Anybody knows why?

System and versions:
   
    python: 3.6.2
    bokeh: 0.12.10
    os: Windows 10 x64 Version 1709 - Chinese Language
    Internet Explorer: 11 (11.125.16299.0)
    tornado: 4.5.2

## Later Update:

With help from many kind people, there seems to be two problems. The first one is about a javascript method `Number.isinteger()` not supported in IE. The second one is about an inconsistent lengths error. Since @mattpap has tagged the first one a bug, I think it's better to make this issue about error 1. I'll open another issue with an minimum reproducable example.

bryevdv:
issues: fixes #7373

@mattpap I believe this is the required fix but please take a look (doing some cargo-cult TS learning)

### DIFF
--- a/bokehjs/src/coffee/polyfill.ts
+++ b/bokehjs/src/coffee/polyfill.ts
@@ -8,6 +8,14 @@ if (typeof Set !== "function") {
   require("es6-set/implement")
 }
 
+// ref: https://github.com/bokeh/bokeh/issues/7373
+const Number_proto = Number.prototype as any
+if (!Number_proto.isInteger) {
+  Number_proto.isInteger = function(value: number): boolean {
+    return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
+  }
+}
+
 const String_proto = String.prototype as any
 if (!String_proto.repeat) {
   String_proto.repeat = function(count: number) {


bokehjs/src/coffee/polyfill.ts
```diff
@@ -8,6 +8,14 @@ if (typeof Set !== "function") {
   require("es6-set/implement")
 }
 
+// ref: https://github.com/bokeh/bokeh/issues/7373
+const Number_proto = Number.prototype as any
+if (!Number_proto.isInteger) {
+  Number_proto.isInteger = function(value: Number): boolean {
```
mattpap: Use lower-case types for primitives.

mattpap:
I would prefer to actually not use polyfillls like this, but provide a utility function in `core/util`. However, I will be cleaning this module up anyway, so we can proceed with this for now.

PR #7406, opened 2018-01-11
Cleaned up tool icons

### LINKED ISSUE #7405: Icons have various visual artifacts
Issue https://github.com/bokeh/bokeh/issues/3250 is nominally about icon size, but it mentions that some of the icons are very pixellated.  There are also other problems with the icons.  E.g. the zoom icons have clear visual artifacts:

image

I'll open a PR to fix these issues.

jbednar:
As shown in issue https://github.com/bokeh/bokeh/issues/7405 and mentioned in issue https://github.com/bokeh/bokeh/issues/3250 the current icons have various visual artifacts or pixellation problems.  This PR doesn't address making them be a uniform size, but it does remove the artifacts and smooths out the pixellation problems. It may be hard to see that in the github diffs, but if you zoom in the differences should be clear on any tool with an arrow and on the wheelzoom tools; some of the other icons have only very minor differences. 

I'd be happy to make them a uniform size as well if someone tells me what size they should fit into.

Fixes #7405.


jbednar:
BTW, the obvious size to shoot for would be 32x32, as they are already all 32 pixels high, so please let me know if you want me to scale them into that size.

mattpap:
32x32 is fine given existing dimensions.

jbednar:
Ready for review.

PR #7409, opened 2018-01-13
Allow to accumulate CSS classes from parent views

### LINKED ISSUE #5420: View.className doesn't respect class structure
Every time `className` is set, it overrides previous classes. CSS classes should accumulate across class structure. Even worse part is that you can provide CSS class name as `attributes: {class: "name"}`, which results is `className` of the parent class being used instead. This is the reason why `DataTable` is messed up under bootstrap, even when there is CSS code to remedy this. Simply `DataTable`'s CSS class is not applied. This was broken for a very long time, but surfaced when `bk-widget` was added to `DataTable`'s parent class. Backbone was never good with class inheritance and this is yet another instance of this.

mattpap:
fixes #5420

### DIFF
--- a/bokehjs/src/coffee/core/dom_view.ts
+++ b/bokehjs/src/coffee/core/dom_view.ts
@@ -5,7 +5,6 @@ import * as DOM from "./dom"
 export class DOMView extends View {
 
   tagName: keyof HTMLElementTagNameMap
-  className: string | null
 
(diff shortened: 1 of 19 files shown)


bryevdv:
@mattpap Should add a migration note to point folks at in case they have an issue with a custom model they've created, etc

PR #7413, opened 2018-01-15
Declare variables in widgets SelectView constant

### LINKED ISSUE #7412: Variables in widget SelectView undeclared
While converting to TypeScript there must have been some issues declaring all variables correctly. I suspect there are probably more cases but for now this code is causing errors:

```javascript
build_options(values): HTMLElement[] {
    return values.map((el) => {
      if (isString(el))
        value = _label = el
      else
        [value, _label] = el

      const selected = this.model.value == value
      return option({selected: selected, value: value}, _label)
    })
  }
```

    ReferenceError: _label is not defined

or

    ReferenceError: value is not defined

This is easily fixed by declaring the variables constant.

philippjfr:
This fixes some variables which were not declared properly in the widget SelectView. I suspect there's a few more of these so as I discover issues I'll add more.

- [x] issues: fixes #7412
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/selectbox.ts
+++ b/bokehjs/src/coffee/models/widgets/selectbox.ts
@@ -22,9 +22,9 @@ export class SelectView extends InputWidgetView {
   build_options(values): HTMLElement[] {
     return values.map((el) => {
       if (isString(el))
-        value = _label  = el
+        const value = _label = el
       else
-        [value, _label] = el
+        const [value, _label] = el
 
       const selected = this.model.value == value
       return option({selected: selected, value: value}, _label)


bokehjs/src/coffee/models/widgets/selectbox.ts
```diff
@@ -22,9 +22,9 @@ export class SelectView extends InputWidgetView {
   build_options(values): HTMLElement[] {
     return values.map((el) => {
       if (isString(el))
-        value = _label  = el
+        const value = _label = el
       else
-        [value, _label] = el
+        const [value, _label] = el
```
mattpap: This is incorrect. `value` and `_label` are local to `if-else` branches with this setup. You need to declare them with `let` before `if-else`. Given that this is a partially typed *.ts, use `gulp dev-build --include widgets/selectbox` to see all compilation errors (by default only syntactical and a few other are shown).

philippjfr: Thanks, that's really helpful! I still have a lot to learn about ES6 and TypeScript.

mattpap:
If you want to find more errors of this kind, you can issue `gulp dev-build --include TS2304`. Just note that I'm the process of gradually fixing stuff, so lets figure out if or what other errors you may want to fix, so that we don't duplicate work.

philippjfr:
You're probably a lot more efficient at finding and resolving these errors, so I'd be happy to defer these fixes to you and close this PR. Due to recent changes I have to work with bleeding edge bokeh for the moment, and will mostly just be fixing bugs as I come across them in my work, which is not a very systematic way of going about this.

mattpap:
OK, I will take care of those errors today. I expect more issues to appear, because right now there are over 5000 filtered out errors across 150 files, so even trivial errors like undeclared variables can slip through.

PR #7261, opened 2017-11-29
Selection model and selection architecture refactor

clairetang6:
This PR is an overhaul of the selection architecture. The goal is to clarify the roles of each piece of the selection machinery and also to create a new models to represent selections and inspections. 

There's more information about the goals of the selection architecture refactor and notes on the selection machinery on [this page](https://github.com/bokeh/bokeh/wiki/Selection-Architecture-Refactor) in the wiki. 

- [x] issues: fixes #6845 #7416 #7054
- [ ] tests added / passed
- [ ] release document entry (if new feature or API change)

PR #7415, opened 2018-01-15
Add drawing tools

philippjfr:
This PR adds a range of draw tools, which allow adding and editing points, multi-lines, boxes and polygons, addressing https://github.com/bokeh/bokeh/issues/6370. For a demo of each of the tools go here (currently these use HoloViews but I'm working on pure bokeh examples now): https://anaconda.org/philippjfr/drawing_tools/notebook

## BoxDrawTool

Allows drawing a rectangular bounding box, which will persist.

## PointDrawTool

Allows adding points, dragging points and deleting points:

* Add:     Click anywhere on the plot
* Drag:    Click on the point and drag it
* Delete: Tap on point (or use box_select tool) and hit ``delete``

## PolyDrawTool

The tool handles drawing multiple polygons/multi-lines. Holding shift will allow editing the previously drawn polygon. Without holding shift a new polygon is added.

* Add polygon/multi-line: Drag from one point to another
* Add vertex to polygon: Hold shift then click and drag to new vertex position
* Delete: Tap on polygon/multi-line and hit ``delete``

## PolyEditTool

The vertex editing tools allow double clicking on a polygon/polyline and edit its vertices.

* Double Tap: Selects a polygon/polyline for vertex editing
* Drag:       Clicking and dragging a vertex adjusts the polygon/polyline
* Delete:     Tap on a vertex and press ``delete``

## API

The API generally is very simple, pass the renderers to be edited to each of the tools. The VertexEditTool also requires a separate ``vertex_renderer`` to draw the vertices.

## ToDo

- [x] Agree on naming
- [x] Refactor a bit to reduce code duplication
- [x] Add docstrings
- [x] Add documentation and examples 
- [x] Resolve remaining TypeScript issues
- [x] tests added / passed


bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+BoxDrawTool.prototype.type = "BoxDrawTool"
+
+BoxDrawTool.prototype.default_view = BoxDrawToolView
+
+BoxDrawTool.define({
+  source: [ p.Instance ],
+  dimensions: [ p.Dimensions, "both"],
+  x: [ p.String, 'x' ],
+  y: [ p.String, 'y' ]
+})
```
jbednar: Needs a newline?

bokehjs/src/coffee/models/tools/draw/point_draw_tool.ts
```diff
@@ -0,0 +1,120 @@
...
+
+PointDrawTool.prototype.type = "PointDrawTool"
+
+PointDrawTool.prototype.default_view = PointDrawView
+
+PointDrawTool.define({
+  source: [ p.Instance ],
+  x:      [ p.String, 'x' ],
+  y:      [ p.String, 'y' ]
+})
```
jbednar: Newline?

bokehjs/src/less/icons.less
```diff
@@ -78,3 +78,27 @@
 .bk-tool-icon-wheel-zoom {
   .tool-icon("WheelZoom");
 }
+
+.bk-tool-icon-wheel-zoom {
+  .tool-icon("WheelZoom");
+}
```
mattpap: Duplicate.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+    const renderers_by_source = this._computed_renderers_by_data_source();
+    for (let _ in renderers_by_source) {
+      var sm;
+      const renderers = renderers_by_source[_];
+      if (this.model.source) {
+        sm = this.model.source.selection_manager;
+      } else {
+        sm = this.model.renderers[0].data_source.selection_manager;
+      }
+      const r_views = (Array.from(renderers).map((r) => this.plot_view.renderer_views[r.id]));
```
mattpap: I presume original this is product of decaffeinate? If so, all those `Array.from()` can be removed.

philippjfr: Yes, I used decaffeinate, I cleaned up most of the stuff it added but there's still a few traces.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+  keyCode: number
+  timeStamp: number
+}
+
+
+export class BoxDrawToolView extends EditToolView {
+  model: BoxDrawTool
+  _basepoint: number[]
+
+  _pan_start(e: BkEv) {
```
mattpap: All methods, including inherited members, must have return types declared. I need finally write a "style" guide.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  _pan_end(e: BkEv) {
+    let emit;
+    const {sx, sy} = e.bokeh;
+    const curpoint = [sx, sy];
+    const { frame } = this.plot_model;
+    const dims = this.model.dimensions;
+
+    const [sxlim, sylim] = this.model._get_dim_limits(this._basepoint, curpoint, frame, dims);
+    this._set_extent(sxlim, sylim, (emit=true));
```
mattpap: There are no named arguments in ES6/TypeScript. Thus, unfortunately, this will have to be rewritten to `this._set_extent(sxlim, sylim, true)`.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+  _pan_start(e: BkEv) {
+    const {sx, sy} = e.bokeh;
+    this._basepoint = [sx, sy];
+  }
+
+  _set_extent(...args: any[]) {
+    const [sx0, sx1] = args[0],
+      [sy0, sy1] = args[1],
+      val = args[2],
+      emit = val != null ? val : false;
```
mattpap: All this code can be rewritten as:
```typescript
_set_extent([sx0, sx1]: [number, number], [sy0, sy1]: [number, number], emit: boolean = false): void {
  ...
}
```

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+    const r = this.computed_renderers[0];
+    const ds = this.model.source;
+    const { frame } = this.plot_model;
+    const xscale = frame.xscales[r.x_range_name];
+    const yscale = frame.yscales[r.y_range_name];
+    const [x0, x1] = xscale.r_invert(sx0, sx1);
+    const [y0, y1] = yscale.r_invert(sy0, sy1);
+    const new_data = {};
```
mattpap: `const new_data: {[key: string]: any[]} = {}`.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -0,0 +1,96 @@
...
+    shiftKey?: boolean
+  }
+  keyCode: number
+  timeStamp: number
+}
+
+
+export class BoxDrawToolView extends EditToolView {
+  model: BoxDrawTool
+  _basepoint: number[]
```
mattpap: Should be a tuple, so `_basepoint: [number, number]`.

mattpap: Value-level syntax is the same, thus `[1, 1]` can be `number[]` as well as `[number, number]`, so explicit type annotations may be necessary later in the code to differentiate between those two cases.

philippjfr:
There's some discussion here about the API. Currently you supply a data source and the columns corresponding to the x- and y-values to determine which columns to edit, but it's probably easier to simply supply a renderer from which the data source and columns can be inferred. I think at the same time we could generalize most tools so they work on multiple renderers.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+    return [x, y];
+  }
+
+  _select_event(e: BkEv, append: boolean) {
+    let did_hit;
+    const {sx, sy} = e.bokeh;
+    const geometry = {
+      type: 'point',
+      sx,
+      sy
```
mattpap: Another style preference is to always add trailing comma. TypeScript will clean this up to generate valid JavaScript.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+    };
+    const renderers_by_source = this._computed_renderers_by_data_source();
+    for (let _ in renderers_by_source) {
```
mattpap: `_` should be replace with a more meaningful name.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
philippjfr: Thanks, I puzzled over this one briefly, I really appreciate your help in bringing me up to speed with the syntax.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+      var sm;
```
mattpap: `var` should be changed to `let`.

bokehjs/src/coffee/models/tools/draw/point_draw_tool.ts
```diff
@@ -0,0 +1,120 @@
...
+export class PointDrawView extends EditToolView {
+  model: PointDrawTool
+
+  _tap(e: BkEv) {
+    const ds = this.model.source;
+    const indices = ds.selected['1d'].indices.slice(0);
+    const append = e.srcEvent.shiftKey != null ? e.srcEvent.shiftKey : false;
+    const did_hit = this._select_event(e, append);
+    if (did_hit) {
+      return null;
```
mattpap: I presume just `return` will suffice (given `void` return type).

bokehjs/src/coffee/models/tools/draw/point_draw_tool.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+    for (let k in ds.data) {
+      let v = ds.data[k];
+      if (![xs, ys].includes(v)) {
+        if ((v.push == null)) {
+          ds.data[k] = (v = Array.prototype.slice.call(v));
+        }
+        v.push('');
+      }
+    }
```
mattpap: What is going on here?  My wild guess is that this has something to do with typed arrays. Typing `xs` and `ys` would be a start to clean this up. Additionally, bokehjs' `includes()` should be used instead of `Array.includes()` (not available in IE). However, given that `v` is an array, then this code likely wouldn't work anyway, because JavaScript supports only reference equality for non-primitive objects. If you want universal equality you have to use bokehjs' `isEquals()`.

bokehjs/src/coffee/models/tools/draw/poly_draw_tool.ts
```diff
@@ -0,0 +1,105 @@
...
+    const indices = ds.selected['1d'].indices.slice(0);
+    const append = e.srcEvent.shiftKey != null ? e.srcEvent.shiftKey : false;
+    did_hit = this._select_event(e, append);
+  }
+
+  _keyup(e: BkEv) {
+    if ((e.keyCode === 8) && this.model.active) {
+      const ds = this.model.source;
+      const { indices } = ds.selected['1d'];
+      indices.sort((a: number, b: number) => a-b);
```
mattpap: This `cmp` function seems unnecessary. `indices.sort()` will achieve the same thing.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
mattpap: Additionally, `_` should be used as a throwaway placeholder as in Python.

bokehjs/src/coffee/models/tools/draw/vertex_edit_tool.ts
```diff
@@ -0,0 +1,157 @@
...
+    // If a Point is selected drag it
+    const [x, y] = this._map_drag(e.bokeh.sx, e.bokeh.sy);
+    const index = ds.selected['1d'].indices[0];
+    ds.data[this.model.x][index] = x;
+    ds.data[this.model.y][index] = y;
+    ds.change.emit(undefined);
+    this.model.line_source.change.emit(undefined);
+  }
+
+  _pan_end(e: BkEv) {
```
mattpap: If an argument isn't used in the body of a function/method, then prefix it with `_`. This will prevent TypeScript from emitting TS6133 error.

bokehjs/src/coffee/models/tools/draw/point_draw_tool.ts
philippjfr: Right, this code converts typed arrays to non-typed arrays, so that data can be added more easily added and removed. When inserting a point x, y value all other columns need to be filled with empty data to avoid the ColumnDataSource column length warnings and maintain consistency. Instead of an empty string I'll likely change the empty value to NaN though.

>JavaScript supports only reference equality for non-primitive objects

Reference equality is all that's needed but I'll make the equality checks explicit anyway.

bokeh/models/tools.py
```diff
@@ -865,3 +866,89 @@ class RedoTool(Action):
...
+    """)
+
+    renderers = List(Instance(Renderer), help="""
+    An explicit list of renderers corresponding to scatter glyphs
+    that may be edited.
+    """)
+
+    x = String('x', help="The name of the x-value column in the supplied source")
+
+    y = String('y', help="The name of the y-value column in the supplied source")
```
mattpap: Add an abstract base class, even if the sole purpose is to group models by functionality. However, given that `x`, `y` are repeated in all new models, you could factor out those two properties to the base class.

mattpap: `source` and `renderers` could be factored out as well.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+        sm = this.model.source.selection_manager;
+      } else {
+        sm = this.model.renderers[0].data_source.selection_manager;
+      }
+      const r_views = (renderers.map((r) => this.plot_view.renderer_views[r.id]));
+      did_hit = sm.select(r_views, geometry, true, append);
+    }
+    return did_hit;
+  }
+}
```
mattpap: Define model/view pairs. This greatly simplifies code analysis and will match (to be) Python's class structure.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+    for (let source in renderers_by_source) {
+      let sm;
+      const renderers = renderers_by_source[source];
+      if (this.model.source) {
+        sm = this.model.source.selection_manager;
+      } else {
+        sm = this.model.renderers[0].data_source.selection_manager;
+      }
+      const r_views = (renderers.map((r) => this.plot_view.renderer_views[r.id]));
+      did_hit = sm.select(r_views, geometry, true, append);
```
mattpap: This considers only last source to renderers mapping. Shouldn't this be `did_hit ||= ...` or something.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+  _select_event(e: BkEv, append: boolean): boolean {
+    let did_hit;
+    const {sx, sy} = e.bokeh;
+    const geometry = {
+      type: 'point',
+      sx,
+      sy,
+    };
+    const renderers_by_source = this._computed_renderers_by_data_source();
+    for (let source in renderers_by_source) {
```
mattpap: `source` can be const (if in doubt, see https://github.com/bokeh/bokeh/pull/7261#discussion_r155922592 for details).

bokehjs/src/coffee/models/tools/draw/line_edit_tool.ts
```diff
@@ -0,0 +1,78 @@
...
+    for (let i = 0; i < ys.length; i++) {
+      xs[i] = xs[i]+dx;
+      ys[i] = ys[i]+dy;
+    }
+    ds.change.emit(undefined);
+  }
+
+  _pan_end(_e: BkEv): void {
+    this.model.source.selected['1d'].indices = [];
+    this._basepoint = null;
```
mattpap: Given this assignment earlier definition of `_basepoint` should be changed to `_basepoint: [number, number] | null`.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
philippjfr: Yes, I mentioned above that there's still some work to do on generalizing these tools to multiple renderers. I'll be looking at that next. That said the behavior for tools that insert data rather than merely editing it (i.e. PointDrawTool, PolyDrawTool, BoxDrawTool), isn't well defined for multiple renderers.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export abstract class EditToolView extends SelectToolView {
+
+  // this is executed when the pan/drag event starts
+  _map_drag(sx: number[], sy: number[]): [number, number] {
+    const frame = this.plot_model.frame;
+
+    if (!frame.bbox.contains(sx, sy)) {
+      return
```
mattpap: The return type of this method is `[number, number]`, but you are returning (implicitly) `undefined`, which would imply `[number, number] | undefined`. However, the more idiomatic approach is to return `null`, to indicate return of a proper value, and so the return type should be changed to `[number, number] | null`.

philippjfr: Thanks, that makes sense.

bokehjs/src/coffee/models/tools/draw/line_edit_tool.ts
philippjfr: Thanks, btw, should all private variables be declared/typed in this way? I tried to look at other classes for examples, but the conversion is probably not complete yet so not all classes are doing it.

bokehjs/src/coffee/models/tools/draw/box_draw_tool.ts
```diff
@@ -20,27 +20,24 @@ export interface BkEv {
...
-    const { frame } = this.plot_model;
+    const frame = this.plot_model.frame;
     const xscale = frame.xscales[r.x_range_name];
     const yscale = frame.yscales[r.y_range_name];
     const [x0, x1] = xscale.r_invert(sx0, sx1);
     const [y0, y1] = yscale.r_invert(sy0, sy1);
-    const new_data = {};
-    new_data[this.model.x] = [[x0, x0, x1, x1]];
+    const new_data: {[key: string]: any[]} = {}
+	new_data[this.model.x] = [[x0, x0, x1, x1]];
```
philippjfr: Odd spacing.

bokehjs/src/coffee/models/tools/draw/edit_tool.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    const x = frame.xscales['default'].invert(sx);
+    const y = frame.yscales['default'].invert(sy);
+    return [x, y];
+  }
+
+  _select_event(e: BkEv, append: boolean): boolean {
+    let did_hit;
```
mattpap: Every variable declaration must have type annotation. Otherwise compiler will complain with "variable has implicit type any", which indeed it currently has. In some cases this can be correctly inferred, but I still prefer to have either a declaration with type annotation or a definition (immediate assignment). Though one thing to keep in mind, is that even if a value is immediately assigned, the inferred type may be incorrect (e.g. `let x = [1, 2]` will be inferred to `number[]`, but you may want `[number, number]`).

bokehjs/src/coffee/models/tools/draw/line_edit_tool.ts
mattpap: This is unrelated to whether a variable is private or not, but to make sure that the type reflects all allowed values. If this line wouldn't exist, then it would be sufficient to use the existing type. We use almost the strictest possible settings (almost because there are a few new options which we didn't exercise yet), which don't allow implicit `null` or `undefined` values. Every use of these must be reflected in the type of a variable, signature, etc. One gray and unexplored area so far is analysis whether an instance variable, like `_basepoint`, were initialized or not. This can result in a behavior that would indicate its type to be `something | undefined` (`[number, number] | undefined` in this case), which is not. I will enable feature checking initialization status at some point, but for now it would only add another 1000 or so errors to the existing long list.

mattpap: I'm actually getting ahead of myself, because was speaking about features not released yet. This will be available in TypeScript 2.7 (sometime this month).

philippjfr:
Okay I've now pushed a refactor for the tool API, instead of defining a ``source`` and the ``x`` and ``y`` coordinate columns each of the tools now processes the ``renderers``. This generalizes the tools quite a bit in that the PointDrawTool now allows dragging and deleting points on multiple renderers, the VertexEditTool allows adding, deleting and dragging vertices on multiple renderers. That said certain operations are only well defined for a single renderer, i.e. there's no way of telling the PointDraw, PolyDraw and BoxDraw tools to add a new point, polygon/multi_line, or box to a specific renderer or data source. Currently these operations simply add the data to the first renderer/data source they find. Similarly the VertexEditTool only allows editing one polygon/multi_line at a time so when you select multiple it once again picks the first.

I think this behavior is all reasonable as long as its documented properly but its definitely still worth discussing.

bokehjs/src/coffee/models/tools/edit/point_draw_tool.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    this._basepoint = null;
+  }
+}
+
+
+export class PointDrawTool extends EditTool {
+  add: boolean
+  drag: boolean
+  renderers: GlyphRenderer[] & Array<HasXYCDS>
```
philippjfr: @mattpap Just wanted to check with you whether this is the proper way to handle this. Each tool can only work with renderers which render certain glyph types and use a ColumnDataSource so I've declared an intersection interface to type that (note that as I discussed with Bryan, Python will also validate this during serialization). This gets rid of all the TypeScript errors warning about ``.data`` not existing on a ``DataSource``. The only errors then remaining are these:
```
[13:14:39] src/coffee/models/tools/edit/point_draw_tool.ts(43,33): error TS2339: Property 'x' does not exist on type 'Glyph & XYGlyph'.
```

This is because the x and y properties are not yet typed on the glyphs. Advice on handling that would be very helpful too.

mattpap: That's a lot to unpack here. Some errors went away because there's a syntax error just before definition of `HasCDS`. The type should be `(GlyphRenderer & HasXYCDS)[]`, though maybe they are equivalent (I need check the rules for intersection types). I will be able to say more when the basics are fixed.

busino:
Hi Philipp,

Just tested your notebook.
Pressing the backspace button in FF goes one page back!
Is there a preventDefault missing somewhere.

Thanks for implementing the drawtool.

Will do some more testing.

Raphael

philippjfr:
@busino I need to update that notebook, the actual tools have changed a bit since I uploaded it and the examples in sphinx/source/docs/user_guide/examples are probably better things to try. 

>Pressing the backspace button in FF goes one page back!

I've now changed deletion to shift+backspace, which does not have the FF issue. That said maybe it's still to easy to make a mistake and go back a page.

busino:
I implemented my own draw tool with custom tool and used the `delete` button.
One of the users had the issue on Mac OS that the `delete` button does a page back!!!
I'm not sure if a preventDefault or something could be useful.

philippjfr:
>I would prefer using the delete button instead of the backspace.

I'd be okay with that, on a small OSX keyboard ``delete`` is triggered using  ``fn``+``backspace``, which doesn't seem too bad. ``backspace`` on its own as I did initially is definitely not a good idea.

philippjfr:
Now updated to use delete, and here are some basic examples to try:

[Box Draw](http://philippjfr.com/ipynbs/box_draw.html)
[Point Draw](http://philippjfr.com/ipynbs/point_draw.html)
[Poly Draw](http://philippjfr.com/ipynbs/poly_draw.html)
[Vertex Edit](http://philippjfr.com/ipynbs/vertex_edit.html)

bokehjs/src/coffee/models/tools/edit/point_draw_tool.ts
philippjfr: Thanks, fixed the syntax now, the errors didn't come back. That said those definitely are not equivalent and those checks are only passing because my type definition gets interpreted as "a list of GlyphRenderers **or** a list of things that have a XY CDS", rather than your more correct suggestion that it should be "a list of GlyphRenderers that have an XY CDS".

I'll have to do some more reading.

bokehjs/test/models/tools/edit/point_draw_tool.coffee
```diff
@@ -0,0 +1,205 @@
+{expect} = require "chai"
```
mattpap: Are there any reasons not to write this in TypeScript?

philippjfr: No, happy to rewrite it, I thought you might be sticking with CoffeeScript in tests for now since I could not find any tests written in TypeScript. Guess it's just one of the last things you'll convert.

mattpap: There aren't many, but should be sufficient for inspiration:
```
$ git ls-files bokehjs/test/ | grep 'ts$'
test/client/index.ts
test/core/has_props.ts
test/core/index.ts
test/core/selector.ts
test/core/signaling.ts
test/core/util/array.ts
test/core/util/refs.ts
test/defaults.ts
test/models/tickers/adaptive_ticker.ts
test/models/tickers/categorical_ticker.ts
test/models/tickers/composite_ticker.ts
test/models/tickers/continuous_ticker.ts
test/models/tickers/fixed_ticker.ts
test/models/tickers/index.ts
test/models/tickers/log_ticker.ts
test/models/tickers/mercator_ticker.ts
test/models/tickers/single_interval_ticker.ts
test/protocol/index.ts
test/size.ts
```
Unfortunately I can't yet rewrite the rest of the tests in bulk, only those which have solid typings.

philippjfr: Great thanks, I did an initial conversion but will use those to make the style a bit more consistent.

bokehjs/test/models/tools/edit/point_draw_tool.ts
```diff
@@ -0,0 +1,212 @@
...
+const {Keys} = utils.require("core/dom");
+const {create_1d_hit_test_result, create_hit_test_result} = utils.require("core/hittest");
+
+const {PointDrawTool, PointDrawToolView} = utils.require("models/tools/edit/point_draw_tool");
+const {Range1d} = utils.require("models/ranges/range1d");
+const {Circle} = utils.require("models/glyphs/circle");
+const {GlyphRenderer} = utils.require("models/renderers/glyph_renderer");
+const {ColumnDataSource} = utils.require("models/sources/column_data_source");
+const {Plot} = utils.require("models/plots/plot");
+const {Toolbar} = utils.require("models/tools/toolbar");
```
mattpap: You will see this in other files I listed, but those can be rewritten as imports, so e.g.:
```typescript
import * as sinon from "sinon"
import {PointDrawTool, PointDrawToolView} from "models/tools/edit/point_draw_tool"
```
Helper function `util.require()` isn't necessary anymore, because typescript does module mapping and knows how to resolve bokehjs' modules (`paths` in `tests/tsconfig.json` governs this).

bokehjs/test/models/tools/edit/point_draw_tool.coffee
mattpap: btw. Tests are dynamically compiled, so you have to actually run tests to see any compilation errors.

bokehjs/test/models/tools/edit/point_draw_tool.ts
philippjfr: Working on this now, just had to add the sinon typings to the requirements.

mattpap: That's good.

bokehjs/test/models/tools/edit/box_draw_tool.ts
```diff
@@ -62,50 +62,44 @@ describe("BoxDrawTool", () =>
...
-      this.hit_test_stub.returns(create_1d_hit_test_result([[1, 0]]));
-      const draw_tool = new BoxDrawTool({renderers: [this.glyph_renderer]});
+      const draw_tool = new BoxDrawTool({
+        active: true,
+        renderers: [this.glyph_renderer],
+        empty_value: "Test"
+      });
       this.plot.add_tools(draw_tool);
-      const draw_tool_view = this.plot_canvas_view.tool_views[draw_tool.id];
+      this.draw_tool_view = this.plot_canvas_view.tool_views[draw_tool.id];
```
mattpap: My preference is not to set on `this`, but to create a function that will spawn a new instance for every test. In fact, this code shouldn't compile, because implicit `this` for the enclosing function doesn't declare this attribute. Unless of course `mocha` is doing something nasty with types.

mattpap: Obviously, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mocha/index.d.ts#L139.

mattpap: I don't have polite words to describe this.

philippjfr: >My preference is not to set on this, but to create a function that will spawn a new instance for every test.

Happy to do that, I'd have the function return an object containing a bunch of objects for easy access. The tools interacts with a lot of different models and views, which I need to access.

>I don't have polite words to describe this.

Nothing surprises me in the world of JS and its derivatives.

mattpap: > The tools interacts with a lot of different models and views, which I need to access.

You can define an interface for this purpose and return a literal object conforming to it. This is what you would do anyway to properly type the current approach, if `@types/mocha` didn't do its magic. You would have to have `function(this: mocha callback arg type & interface with your custom attributes)` everywhere. This is why I hate so much tricks that `@types/mocha` uses for users' convenience, because it tricked you into thinking that you're writing type safe code, whereas you were writing completely dynamically typed code. Note that `any` means anything can happen, and there is no type checking as soon as `any` appears in the code.

mattpap: Stylistic issues aside, not sharing state between tests is actually pretty important for correctness of tests, because you have to acknowledge that there are side effects of bokehjs' code you run. Unless you want to test those side effects, you should be always starting a test with a fresh set of objects. Of course, one can stub out certain classes of functionality to prohibit those side effects from happening, but stubbing code out isn't without its issues.

philippjfr: >Stylistic issues aside, not sharing state between tests is actually pretty important for correctness of tests, because you have to acknowledge that there are side effects of bokehjs' code you run.

Isn't the point of the beforeEach block that it is rerun before each individual test which means none of the state is actually shared between tests?

philippjfr:
@bryevdv you mentioned you wanted gifs to demonstrate the functionality. I'll have a go at recording some and then overlaying some text to show the key combinations at each point. How would you structure it? Currently I'm imagining something like this:

```
Tool Name
~~~~~~~~~

High-level description

Gif

Description of supported tool actions

Embedded bokeh plot example
```

Secondly once I have the gifs, where should I upload them?

bryevdv:
@philippjfr sounds good. Put them somewhere I can access them and I will upload them to the static dir on pydata.org

philippjfr:
I'm tempted to rename the ``VertexEditTool`` to ``PolyEditTool`` because it does more than just editing vertices. Otherwise I'm happy with the naming. If anyone has other suggestions speak up now.

jbednar:
I'm happy with PolyEditTool.

bokehjs/test/models/tools/edit/box_draw_tool.ts
philippjfr: Still happy to split it out of course, if only to define proper types, just trying to clarify my understanding.

philippjfr:
@mattpap I'm still puzzling over how to properly type the renderers and glyph types. On the Python side there's now validation that the renderers must have the appropriate glyph types for the tool and in the TypeScript I'm also validating through interfaces like this:

```javascript
export interface HasXYCDS {
  data_source: ColumnDataSource
  glyph: XYGlyph
}

renderers: (GlyphRenderer & HasXYCDS)[]
```

Unfortunately the dataspecs I access on the Glyphs don't seem to be typed so I'm getting warnings like this:

```
[19:10:24] src/coffee/models/tools/edit/point_draw_tool.ts(43,33): error TS2339: Property 'x' does not exist on type 'Glyph & XYGlyph'.
[19:10:24] src/coffee/models/tools/edit/point_draw_tool.ts(43,48): error TS2339: Property 'y' does not exist on type 'Glyph & XYGlyph'.
```

bokehjs/src/coffee/models/tools/edit/poly_edit_tool.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  _doubletap(e: BkEv): void {
+    // Perform hit testing
+    const renderers = this._select_event(e, false, this.model.renderers);
+
+    // If we did not hit an existing line, clear node CDS
+    const vertex_renderer = this.model.vertex_renderer;
+    const point_ds = vertex_renderer.data_source;
+    const point_glyph = vertex_renderer.glyph;
+    const [pxkey, pykey] = [point_glyph.x.field, point_glyph.y.field];
```
mattpap: `field` is not a required member of a `NumberSpec`, so unless you guard against this somewhere else, this code may fail.

mattpap:
> Property 'x' does not exist on type 'Glyph & XYGlyph'.

Those are not yet typed. You have to options, either add those members to `XYGlyph` or just view `point_glyph` as `any` in those code (for now) (which I think I would prefer, because I'm filling in those gaps now).

bokehjs/src/coffee/models/tools/edit/poly_edit_tool.ts
philippjfr: Thanks I did mean to guard against that everywhere but seemingly missed this file.

philippjfr:
@bryevdv How are these gifs? I've overlaid the key presses and taps/drags are highlighted through circles around the cursor to show what's going on. 

[Box Draw Tool](https://assets.holoviews.org/tmp/box_draw_keyboard_optimized.gif)
[Point Draw Tool](https://assets.holoviews.org/tmp/point_draw_keyboard_optimized.gif)
[Poly Draw Tool](https://assets.holoviews.org/tmp/poly_draw_keyboard_optimized.gif)
[Poly Edit Tool](https://assets.holoviews.org/tmp/poly_draw_keyboard_optimized.gif)

philippjfr:
@mattpap Sorry to keep bugging you, could you have a brief look at my [last commit](https://github.com/bokeh/bokeh/pull/7415/commits/f0c03ebef8cbb29e5dd30c8a4f2152ab45ef85c9) and confirm that's what you were imagining when you suggested refactoring the tests.

bryevdv:
@philippjfr those look great (FYI the last two both link to poly draw tool but changing the URL worked) How big are the images?

philippjfr:
Fixed the link and added file sizes to the comment above.

bryevdv:
Yah I guess that's too much to just check in. I can put them in e.g. the same place as this:

https://bokeh.pydata.org/static/stocks.png

Does that work for writing docs?

philippjfr:
>Does that work for writing docs?

Sure, sounds good. Of course it means an offline build won't include them, but that's okay I guess.

bryevdv:
@philippjfr uploaded, e.g.

http://bokeh.pydata.org/static/poly_edit_keyboard_optimized.gif

philippjfr:
Thanks, I'll add those to the documentation now and then add a release notes entry. Apart from that I think this PR is ready for final review.

bokehjs/test/models/tools/edit/poly_draw_tool.ts
```diff
@@ -4,165 +4,221 @@ import * as sinon from "sinon"
...
 const utils = require("../../../utils")
 
+export interface PolyDrawTestCase {
+  data: {[key: string]: (number[] | null)[]}
+  data_source: ColumnDataSource
+  draw_tool_view: PolyDrawToolView
+  glyph_view: PatchesView
+}
+
+export interface BkEv {
```
mattpap: I presume this could be just imported from appropriate tool's module?

philippjfr: Yes, I actually wanted to ask you about that, these are currently duplicated all over the place. Is there a central place where we could define these or should I just eliminate the definitions in the tests for now?

mattpap: I did it on purpose to learn what all tools actually use (it's a mess with all sorts of behavior with hammer, native events and bokehjs' customizations). I've started typing ui_events and when done, event interfaces will be defined there. But for now just follow the existing approach.

philippjfr:
@bryevdv Looks like Sphinx isn't happy about the non-local gifs:

```
/home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/tools.rst:None: WARNING: nonlocal image URI found: http://bokeh.pydata.org/static/box_draw_keyboard_optimized.gif
```

mattpap:
@philippjfr, with the next batch of changes, can you rebase on top of `master` and fix any emerging errors in this code, including from `gulp scripts:tslint`.

philippjfr:
Okay, will push those fixes in a minute. Almost entirely missing trailing commas.

mattpap:
@philippjfr, `suppress_warnings = ['image.nonlocal_uri']` should silence sphinx.

bryevdv:
Don't do that tho. Use `raw:: html` blocks as done elsewhere

philippjfr:
Thanks @bryevdv, will push that shortly.

philippjfr:
@bryevdv Ready for review whenever you get a chance.

bryevdv:
I will build it and try things out today

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -14,6 +14,7 @@ interactions:
 
 * :ref:`userguide_tools_actions`
 * :ref:`userguide_tools_inspectors`
+* :ref:`userguide_tools_edit`
```
bryevdv: Text above this says 'There are three categories of tool interactions:' which is no longer consistent with list below. I think this intro could be improves a bit with some fairly minor changes, so I will just push an update the the branch  in a little while.

philippjfr: Thanks, should have looked more closely.

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -561,6 +562,218 @@ shown below:
...
+data.
+
+All the edit tools share a small number of key bindings:
+
+* SHIFT: Allows adding to a selection or continue drawing
+* DELETE: Deletes the selected glyphs
+* ESC: Clear the selection
+
+.. note::
+
```
bryevdv: no blank line after admonition marker

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -561,6 +562,218 @@ shown below:
...
+   On some shortened keyboards (including MacBooks) the DELETE key is
+   missing and can be toggled by pressing FN + BACKSPACE
```
bryevdv: I've never seen a macbook without a delete key (a key literally physically labeled "delete", am looking at one right now) , but I can't find a backspace on this MPB, so this note seems very confusing

bryevdv: Right, so on my MPB I have to hit FN and the key literally labeled "delete" so I think this will be very confusing as is

philippjfr: Worth clarifying this since it is confusing, the key labeled ``delete`` on a MacBook actually maps to the backspace key code, when you press Fn+ the key labeled "delete" it maps to the "delete" key code.

bryevdv:
Regarding `BoxDrawTool`, it lets you draw new boxes while there is a selection active. Is this desirable? I was thinking this tool would let me drag the selected rects to move them, but that just added new boxes, which was annoying.

sphinx/source/docs/user_guide/tools.rst
bryevdv: Most users are not going to care or know about keycodes. Is there not a way re-map the key code to something uniform so that users can be spared from having to consider this at all?

philippjfr:
>I was thinking this tool would let me drag the selected rects to move them, but that just added new boxes, which was annoying.

Good point, should be pretty straightforward to improve. Originally you could only draw a single box but now that you can draw multiple I like your suggestion. Happy to do that now.

sphinx/source/docs/user_guide/tools.rst
bryevdv: Because this will definitely generate repeated bug reports from mac users.

bryevdv: I'd even suggest something else entirely, e.g. "x" to delete, if delete can't be "just made to work" on all platforms

philippjfr: Any suggestion for a key? I originally simply had it mapped to backspace, but since key events are processed as long as the tool is active and backspace is a very common key to press it was too easy to accidentally delete your selection. How about shift-backspace?

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -561,6 +562,218 @@ shown below:
...
+
+* name: ``'box_draw'``
+* menu icon: |box_draw_icon|
+
+The BoxDrawTool allows drawing, dragging and deleting ``Rect`` glyphs
+on one or more renderers by editing the underlying
+``ColumnDataSource`` data. Like other drawing tools, the renderers
+that are to be edited must be supplied explicitly as a list. When
+drawing a new box the data will always be added to the
+``ColumnDataSource`` on the first supplied renderer.
```
bryevdv: it's never demonstrated anywhere what "supplying as a list" looks like. This needs an example code snippet to show what configuring one of these tools looks like in some way. Additionally there should at least be a link to the code for the live demos. If it is too cumbersome to put directly inline, then a "The code for this example is at..." would do. Or there is also a collapsible code block directive to make it hidden by default.

sphinx/source/docs/user_guide/tools.rst
```diff
@@ -561,6 +562,218 @@ shown below:
...
+  SHIFT key while tapping.
+
+* Delete point: Select one or more points then press DELETE key.
+
+* Move point: Click and hold the mouse down point, drag it, then
+  release the mouse to complete the action. To move multiple points
+  select them and hold SHIFT key while dragging one of them.
+
+.. bokeh-plot:: docs/user_guide/examples/tools_point_draw.py
+    :source-position: none
```
bryevdv: In this and the other examples, the tool is not active by default. I tried to just click to add points right away and nothing happened. Immediate thought was "what's broken". Suggest making the relevant edit tools active by default in these examples.

bryevdv: Also in this specific example I think the table should go above, or the plot has to be made much narrower

sphinx/source/docs/user_guide/tools.rst
jbednar: The same key does "just work" to delete backwards on all platforms, it's just that Mac decided to call it Delete, PCs decided to call it Backspace, and even worse, PCs use the label Delete for an entirely different key.  The key that deletes backwards is what deletes selected stuff in any drawing program or editor I've ever seen across platforms, so it's the key people will be trying regardless of any docs we write. Thus even though it's confusing to describe, I think it's much more discoverable than 'x' or any other key, so I think it should still work as it does now, just described as best we can.

sphinx/source/docs/user_guide/tools.rst
bryevdv: screen shot 2018-01-25 at 12 08 20

sphinx/source/docs/user_guide/tools.rst
philippjfr: I think the PolyEdit example does demonstrate it. Should I just modify each example to use multiple renderers?

>Or there is also a collapsible code block directive to make it hidden by default.

I'll give that a go.

bryevdv:
Regarding point edit tool: wouldn't it be simpler for both users and the code, if dragging just always dragged whatever is selected? It seems very complicated to have to hit shift to multi-select, then hit shift again while dragging or else it will move something different than what I just selected. I made the effort to make a multi-selection, it seems natural that's what I'd want to move as a result. 

Additionally, so far I think it's a bit confusing to include "select point" as an action because selecting doesn't really do anything useful or interesting on its own, it is a precursor to doing a move or a delete. My preference would be a narrative more like:

> The point edit tool can ADD, DELETE, or MOVE points on plots. 

> To ADD new points, simply click anywhere on the plot
>
> Additionally, when the point select tool is active, points may be selected (or multi-selected by holding shift) in the usual way. Once a points are selected, the following actions are possible:
> 
> * DELETE the selection - ...
> * MOVE the selection - ...

philippjfr:
>It seems very complicated to have to hit shift to multi-select, then hit shift again while dragging or else it will move something different than what I just selected.

I think I agree with you here, happy to change that as well. The box tool will also work like that in that case.

bryevdv:
Also a general comment. I find bullet lists list this very hard to visually parse quickly:

screen shot 2018-01-25 at 12 20 25

the colon separator is hidden in a dense block so its hard to tell the first part that is the thing being defined. A simple fix is to highlight the text before the colon somehow (emphasis or bold) though I think I prefer to use ReST definition lists for this kind of thing.

sphinx/source/docs/user_guide/tools.rst
bryevdv: No I think the example is fine, the snippet does not have to relate to the example. Just a few lines showing e.g.
```
r1 = plot.rect(...)
r2 = plot.rect(...)

plot.add_tool(BoxDrawTool(renderers=(r1, r2))
```
slightly more fleshed out

sphinx/source/docs/user_guide/tools.rst
bryevdv: > The key that deletes backwards is what deletes selected stuff in any drawing program or editor I've ever seen across platforms, so it's the key people will be trying regardless of any docs we write. 

I agree! But that key on mac is just "delete", *not* "FN + delete". So I am suggesting to find a way to make "delete" work on mac, and everything the same on other platforms.

philippjfr:
>Regarding point edit tool: wouldn't it be simpler for both users and the code, if dragging just always dragged whatever is selected? It seems very complicated to have to hit shift to multi-select, then hit shift again while dragging or else it will move something different than what I just selected.

Quick question about this, what it currently does is that if you try to drag a point by clicking on it and then dragging it will selected that. If we decide to move to dragging the current selection we have three options:

1. It will drag only the previously selected points regardless of whether you're attempting to click on and drag a specific point.

2. It will drag the union of previously selected points and the specific point you clicked on and dragged.

3. If you happen to click on and drag a specific point it drags only that point (despite what was previously selected), but if you drag anywhere else it will drag the current selection.

sphinx/source/docs/user_guide/tools.rst
philippjfr: >So I am suggesting to find a way to make just "delete" work on mac

I initially had it mapped to the backspace key code but didn't like it because that's such a common key to press and it's too easy to accidentally delete the current selection while editing a cell. Having it behave that way for OSX would still have that problem.

bryevdv:
I would say for the above question, 3 is most intuitive. click any thing and dragging it drags that thing. But if the thing is part of a selection, the entire selection should move. 

tangentially related: selections should be cleared after a deletion. If I add 5 lines with poly draw tool then select and delete the third, the selection magically jumps to the next line (the new number three) after the deletion and that seems undesirable.

philippjfr:
>tangentially related: selections should be cleared after a deletion.

Need to add tests for this, that was definitely the intended behavior.

philippjfr:
>tangentially related: selections should be cleared after a deletion.

Actually I can't reproduce this and it appears to be tested.

sphinx/source/docs/user_guide/tools.rst
jbednar: Is that a big problem if undo works well?

bryevdv:
@philippjfr not sure, it happens every time for me:

screenflow

philippjfr:
Changed the PointDrawTool dragging behavior to match suggestion 3. above. Should the same dragging behavior apply to the BoxDrawTool? If I allow click and drag on a specific box (without applying a selection first), then you won't be able to start drawing a new box inside an existing one (which might not be a big issue since you can subsequently drag it around).

philippjfr:
>@philippjfr not sure, it happens every time for me:

Thanks, it seems it does happen for multi_line glyphs but not for patches. Will investigate.

bryevdv:
> Should the same dragging behavior apply to the BoxDrawTool? 

Depends on which action should get priority, moving or adding? I guess I'd suggest that the cleanest thing might be for bare dragging only be for moving, and to add you have to shift-drag

philippjfr:
>I guess I'd suggest that the cleanest thing might be for bare dragging only be for moving, and to add you have to shift-drag

I'd be happy with that.

bryevdv:
also about the poly draw tool, I've never experience shift-drag to add a single point. It's a really disorienting mechanic. I would expect shift-click to immediately drop a new vertex. Not having the vertex add until some movement happens is very unexpected.

philippjfr:
>also about the poly draw tool, I've never experience shift-drag to add a single point. It's a really disorienting mechanic.

Seems reasonable, I'm not quite clear when hammer.js emits a ``tap`` vs. a ``_pan_start`` event but it should be pretty straightforward.

sphinx/source/docs/user_guide/tools.rst
bryevdv: > I initially had it mapped to the backspace key code but didn't like it because that's such a common key to press and it's too easy to accidentally delete the current selection while editing a cell. Having it behave that way for OSX would still have that problem.

How is that not a problem on other platforms then? It's physically the same upper right key (called "delete" on mac and called "backspace" on others). If hitting bare "delete" on osx is to problematic, then it must also be the case that hitting bare "backspace" on others is also problematic for the exact same reason. If this is an issue because key bindings are global we can and probably should definitely check that the cursor is inside the plot before deleting anything.

philippjfr: >If hitting bare "delete" on osx is to problematic, then it must also be the case that hitting bare "backspace" on others is also problematic for the exact same reason.

That's true, I thought you were suggesting using the delete key on OSX and the entirely different delete key on other machines.

>. If this is an issue because key bindings are global we can and probably should definitely check that the cursor is inside the plot before deleting anything.

If there's an easy way to do that I have no issue at all with using the backspace key code.

bryevdv:
> Seems reasonable, I'm not quite clear when hammer.js emits a tap vs. a pan_start event but it should be pretty straightforward.

It would actually be better if there was someway to have "adding a poly" be a more continuous and defined process. I mean, once started I just click to drop new points where I want them then a new "next point" follows my mouse until i click again. If  don't another next point, I hit esc and the poly is completed. maybe the initial line drag coudl start this state?

philippjfr:
>I mean, once started I just click to drop new points where I want them then a new "next point" follows my mouse until i click again.

That's a very intriguing suggestion and I'll have a go at it. My only worry is how it would interact with selection if you're trying to draw overlapping polygons. Would you also want to tap to start drawing a new polygon or would you leave that is it is and shift+drag to initiate the draw mode?

sphinx/source/docs/user_guide/tools.rst
bryevdv: > That's true, I thought you were suggesting using the delete key on OSX and the entirely different delete key on other machines.

Just to make sure I'm not being amibiguous, I am suggesting using the physically most upper, most right key, where "upper" is bounded by the row of number keys. 

> If there's an easy way to do that I have no issue at all with using the backspace key code.

I think we'd have to toggle a flag on canvas mouse enter/mouse leave events. I still don't understand how it can be too easy to hit the most upper right key on a mac (necessitating adding FN) but it's somehow not too easy to hit the same most upper right key on all other platforms? That logic of that does not make sense to me.

philippjfr: >I still don't understand how it can be too easy to hit the most upper right key on a mac (necessitating adding FN) but it's somehow not too easy to hit the same most upper right key on all other platforms?

This is all very confusing. Fn+delete on OSX is equivalent to (i.e. emits the same key code as)  the actual dedicated delete key (located under the insert key) on full keyboards. That was a safer choice when it was capturing events globally because on OSX systems you'd have to press Fn+delete and on other systems the dedicated delete key (not the backspace). If we do detect whether you're hovering over the plot then backspace is fine.

bryevdv:
>  Would you also want to tap to start drawing a new polygon or would you leave that is it is and shift+drag to initiate the draw mode?

I made some edits but did not call them out. I think maybe double tap would be best

philippjfr:
Okay great, I like it. If we're using double tap to start a new poly, how about letting you end it with a double tap too (ESC can still be supported)?

philippjfr:
Great, starting on that in a minute. I've done the PointDraw tool, am just finishing off the improved BoxDrawTool. Once all those are done the PolyEdit tool can probably use some of the same improvements, i.e. the improved vertex drag behavior from the PointDraw tool and the doubletap + draw on mouse move behavior from the PolyDraw tool.

philippjfr:
Okay, I've been experimenting with your suggestion on the poly tool and it works fairly well. There is a small issue due to the fact that the tap event has a timeout window in which it briefly waits for a double tap to be emitted. By that point the user will have already moved their mouse cursor somewhere else generating move events in the process. What ends up happening is that you'll see a brief flicker between the time you tap your mouse and when the tap event actually arrives.

philippjfr:
I've pushed the changes now if you want to have a go to see what I mean. I still think it's an improvement overall. If you're happy with those changes I'll get all the documentation updated in the morning.

sphinx/source/docs/user_guide/tools.rst
philippjfr: >I think we'd have to toggle a flag on canvas mouse enter/mouse leave events.

That worked, it now deletes if you press backspace (or the OSX key labelled "delete") while the mouse is within the frame.

sphinx/source/docs/user_guide/tools.rst
philippjfr: Sounds good, should I do this for each tool?

jlstevens:
> There is a small issue due to the fact that the tap event has a timeout window in which it briefly waits for a double tap to be emitted. 

I noticed this when using the HoloViews [Tap](https://github.com/ioam/holoviews/blob/master/holoviews/plotting/bokeh/callbacks.py#L622) stream which tries to combine tap and double tap events. It sounds like both these drawing tools and the ``Tap`` stream would benefit if bokeh could offer some sort of 'raw' tap events i.e an event that can represent all taps (whether single or double) without delay.

bryevdv:
> It sounds like both these drawing tools and the Tap stream would benefit if bokeh could offer some sort of 'raw' tap events i.e an event that can represent all taps (whether single or double) without delay. 

Definitely something for a later day, if ever. Might be worth seeing if Hammer exposes any tunable parameters around this, first.

bryevdv:
@philippjfr point and poly edit tools seems much better to me now, thanks. A few other last comments:

* I still think the tools in the docs examples should be enabled/on by default
* With so few widely separated points in the point edit examples, the colors become hard  to distinguish (e.g. dark blue vs black especially). Maybe a grey background, so that some lighter colors like orange/yellows could be available for use?

philippjfr:
>I still think the tools in the docs examples should be enabled/on by default

Already changed that, haven't pushed yet.

>Maybe a grey background, so that some lighter colors like orange/yellows could be available for use?

Sounds like a good idea. I'll also have new gifs for you to upload in a minute.

philippjfr:
New gifs here:

[Box Draw Tool](http://assets.holoviews.org/tmp/box_draw_keyboard_optimized.gif) [700 KB]
[Point Draw Tool](http://assets.holoviews.org/tmp/point_draw_keyboard_optimized.gif) [600 KB]
[Poly Draw Tool](http://assets.holoviews.org/tmp/poly_draw_keyboard_optimized.gif) [1.2 MB]
[Poly Edit Tool](http://assets.holoviews.org/tmp/poly_edit_keyboard_optimized.gif) [1.7 MB]

philippjfr:
>I still think the tools in the docs examples should be enabled/on by default

This does not yet work in the point tool example due to an issue with proxy tools I was going to resolve in another PR. Can try to fix that here.

philippjfr:
Apart from that I think this is now ready for another round of review.

bryevdv:
I will start checking it out now

bryevdv:
@philippjfr I think the docs and interactions are all much improved. I still owe a commit to update the section heading, I will do that shortly. My only comments are these:

* the data table no longer updates while a move is "in flight" I'm not sure if I have a preference, tho only updating at the "end"(the new current behaviour) is probably better for performance in server apps. Just pointing out in case it was not intentional or expected. 

* when ending adds, double tap and ESC behave differently. Double tap drops the last point, but ESC does not. That's nice in a certain way, but it does mean mobile does not have access to identical functionality. Wondering if maybe double tap should do the same as ESC (just end, without dropping the current position as a last point) 

I have *not* done a thorough review of the code, because it looks like @mattpap and you have both iterated extensively already.

bryevdv:
@philippjfr I have made the commit to update the intro

philippjfr:
>the data table no longer updates while a move is "in flight" I'm not sure if I have a preference, tho only updating at the "end"(the new current behaviour) is probably better for performance in server apps. Just pointing out in case it was not intentional or expected.

Assume that would require a ``ds.properties.data.change.emit(undefined)`` which I purposefully avoided to stop all the data being synced on every move event. I would actually except the table to update on a simple ``ds.change.emit()`` so maybe we can fix that. If not I think the performance concerns are still to great to enable it.

>when ending adds, double tap and ESC behave differently. Double tap drops the last point, but ESC does not. That's nice in a certain way, but it does mean mobile does not have access to identical functionality. 

Right that seemed like the most intuitive behavior. I'd be okay making them identical but it would be a bit unfortunate.

>I have not done a thorough review of the code, because it looks like @mattpap and you have both iterated extensively already.

That's true but I think that was mostly focused on TypeScript issues.

bokeh/models/tools.py
```diff
@@ -865,3 +872,230 @@ class RedoTool(Action):
...
+
+    dimensions = Enum(Dimensions, default="both", help="""
+    Which dimensions the box drawing is to be free in. By default,
+    users may freely draw boxes with any dimensions. If only "width"
+    is supplied, the box will be constrained to span the entire
+    vertical space of the plot, only the horizontal dimension can be
+    controlled. If only "height" is supplied, the box will be
+    constrained to span the entire horizontal space of the plot, and
+    the vertical dimension can be controlled.
+    """)
```
bryevdv: @philippjfr suddenly wonder if we should call this `BoxEditTool` to leave open the future possibility of additional modes to edit existing boxes (by dragging their edges to expend contract, etc)

bryevdv: As it is this tool is *almost* useful for a somewhat often asked for capability: having a "zoom window" on one plot that controls the range on another.

bokeh/models/tools.py
```diff
@@ -865,3 +872,230 @@ class RedoTool(Action):
...
+
+    @error(INCOMPATIBLE_BOX_DRAW_RENDERER)
+    def _check_compatible_renderers(self):
+        incompatible_renderers = []
+        for renderer in self.renderers:
+            if not isinstance(renderer.glyph, Rect):
+                incompatible_renderers.append(renderer)
+        if incompatible_renderers:
+            glyph_types = ', '.join([type(renderer.glyph).__name__
+                                     for renderer in incompatible_renderers])
```
bryevdv: I'd personally prefer this on one line.  I don't think it would exceed our linter limit (160 IIRC)

bokeh/models/tools.py
```diff
@@ -865,3 +872,230 @@ class RedoTool(Action):
...
+
+    @error(INCOMPATIBLE_POINT_DRAW_RENDERER)
+    def _check_compatible_renderers(self):
+        incompatible_renderers = []
+        for renderer in self.renderers:
+            if not isinstance(renderer.glyph, XYGlyph):
+                incompatible_renderers.append(renderer)
+        if incompatible_renderers:
+            glyph_types = ', '.join([type(renderer.glyph).__name__
+                                     for renderer in incompatible_renderers])
```
bryevdv: ditto

bokeh/models/tools.py
philippjfr: Sounds good to me, I did consider adding that while adding drag capability.

bokehjs/src/coffee/models/tools/edit/poly_edit_tool.ts
```diff
@@ -0,0 +1,244 @@
...
+    this._select_event(e, append, this.model.renderers);
+  }
+
+  _remove_vertex(emit: boolean = true): void {
+    if (!this._drawing || !this._selected_renderer) { return; }
+    const renderer = this.model.vertex_renderer;
+    const ds = renderer.data_source;
+    // Type once dataspecs are typed
+    const glyph: any = renderer.glyph;
+    const index = ds.selected['1d'].indices[0];
```
bryevdv: just noting that this and presumably others will need updates for @clairetang6 PR

philippjfr: Yes, I think there are probably a fair number of changes needed, but I've only scanned the PR. I'd be happy to help resolve those issues after a rebase.

bryevdv: I think we merge this first, then fix up with your help in the other PR

bokeh/models/tools.py
bryevdv: OK well I think after that change we merge? And get a dev build out

philippjfr: You mean adding the expand/contract mode? Only way I can currently think of is adding a vertex_renderer just like the PolyEditTool. Does that seem sensible?

bryevdv: No, just changing the name now, to leave open the possibility of adding the editing features later without needing a new separate tool.

bryevdv:
OK, I am going to go ahead and merge this, and cut a dev build. Let's ping the issue and mailing list to get folks to try it out.

philippjfr:
@bryevdv Would you mind updating the gifs? I noticed it's still using the old ones.

[Box Edit Tool](http://assets.holoviews.org/tmp/box_draw_keyboard_optimized.gif) [700 KB]
[Point Draw Tool](http://assets.holoviews.org/tmp/point_draw_keyboard_optimized.gif) [600 KB]
[Poly Draw Tool](http://assets.holoviews.org/tmp/poly_draw_keyboard_optimized.gif) [1.2 MB]
[Poly Edit Tool](http://assets.holoviews.org/tmp/poly_edit_keyboard_optimized.gif) [1.7 MB]

Just make sure to rename box_draw_keyboard_optimized.png to box_edit_keyboard_optimized.png.

PR #7421, opened 2018-01-17
Remove gloo2.js and add a dependency on gloo2

### LINKED ISSUE #7420: Move gloo2 to its own repository

mattpap:
fixes #7420


bryevdv:
I'd like to replace gloo with ReGL entirely some day. If we this, we are either stuck keeping `bokeh/gloo` around forever, or stuck with some revisions of bokeh being unbuildable some day. So I'm -1 on this.

mattpap:
@bryevdv, this change actually makes it easier to replace gloo with another library. This is also needed to simplify our build.

> If we this, we are either stuck keeping bokeh/gloo around forever, or stuck with some revisions of bokeh being unbuildable some day

That same you can say about any other library we depend on on github and npm, which btw. isn't immutable, so most bokehjs reversions will be unbuildable at some point, unless you provide a complete, reproducible build environment.

mattpap:
> I'd like to replace gloo with ReGL entirely some day

Knowing the resources we have available, that's not going to happen anytime soon, if ever.

mattpap:
Alternatively I can push a package to npm, but once again, that would have to be done for other packages as well, to make things consistent.

bryevdv:
Sigh. Ok

PR #7425, opened 2018-01-18
Add support for path_args and path_kwargs in bokeh server

### LINKED ISSUE #7424: Add support for path_args and path_kwargs in bokeh server
Given that ```session_context.request``` only supports ```arguments```, it is not possible to get fragments in urls with regex patterns, which is supported in Tornado as ```RequestHandler.path_args``` and ```RequestHandler.path_kwargs```. Hope that bokeh can add these two properties to ```session_context``` for REST-style urls.

For instance, with this functionality, we can handle urls like ```http://localhost:5000/graphs/34512``` by code below

```python
def handler_fun(doc):
    # if request url is http://localhost:5000/graphs/34512,
    # doc.session_context.path_kwargs will be {'graph_id': '34512'}
    doc.title = 'Graph ' + doc.session_context.path_kwargs['graph_id']
    # handle request for graph `graph id`

server = Server({'/graphs/(?P<graph_id>[^/]+)', Application(FunctionHandler(handler_fun))}, port=5000)
```

siyuwj:
- [x] issues: fixes #7424
- [x] tests added

This PR adds two properties, ```path_args``` and ```path_kwargs``` to ```BokehSessionContext``` to store capturing groups in Tornado ```RequestHandler```. Test cases has been added.

### DIFF
--- a/sphinx/source/docs/user_guide/server.rst
+++ b/sphinx/source/docs/user_guide/server.rst
@@ -394,7 +394,7 @@ group ``graph_id``:
 
   # if the url requested is http://localhost:5006/graphs/345126,
   # graph_id should be `345126`
-  graph_id = kwargs.get('345126', '')
+  graph_id = kwargs.get('graph_id', '')
 
 .. _userguide_server_applications_callbacks:
 
--- a/sphinx/source/docs/user_guide/server.rst
+++ b/sphinx/source/docs/user_guide/server.rst
@@ -377,6 +377,25 @@ plot):
   writing directly to ``request.connection`` is unsupported and will result in
   undefined behavior.
 
+When capturing groups are defined in regular expressions for application paths
+for a Bokeh application, you can get captured values in ``curdoc().session_context``
+via ``session_context.path_args`` and ``session_context.path_kwargs``.
+
+As an example, the following code will access the value of capturing groups
+for the ID of the graph, which is defined in the url pattern as named capturing
+group ``graph_id``:
+
+.. code-block:: python
+
+  # the url pattern of the application is defined as `/graphs/(?P<graph_id>[^/]+)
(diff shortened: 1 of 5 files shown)


sphinx/source/docs/user_guide/server.rst
```diff
@@ -377,6 +377,25 @@ plot):
...
+.. code-block:: python
+
+  # the url pattern of the application is defined as `/graphs/(?P<graph_id>[^/]+)
+  # and urls accepted by this pattern can be http://localhost:5006/graphs/345126
+
+  kwargs = curdoc().session_context.path_kwargs
+
+  # if the url requested is http://localhost:5006/graphs/345126,
+  # graph_id should be `345126`
+  graph_id = kwargs.get('345126', '')
```
bryevdv: is this meant to be `kwargs.get('graph_id', '')` ?

bryevdv:
@siyuwj This seems interesting. Can you provide a small complete example that demonstrates this in use? It's not clear to me how these regex patterns are to be supplied by the user. Does this only work when embedding a Bokeh server as a library (since then a user has direct access to the Tornado handers)? Or can it be applied to the more typical `bokeh serve app.py` case? if not we might want to consider a new command line arg to pass the regex along so it can work in all cases.

Additionally we will need to determine that this works in the presence of `--num-threads` and load balancers. That was the issue with trying to provide access to the full request (the originating HTTP request and the subsequent websocket upgrade are not guaranteed to go to the same process)

bryevdv:
@siyuwj will you be able to provide a small complete example that exercises this functionality?

siyuwj:
I decide to withdraw this PR because I found it hard to deal with static paths in BokehJS and currently I have few time to deal with it.

PR #7432, opened 2018-01-18
Remove attrs from HasProps.initialize()

### LINKED ISSUE #7431: Remove attrs from HasProps.initialize()
There is only one case where `attrs` are actually used, and it can be rewritten differently. Given this, I think, that they aren't necessary anymore. Removing them would also greatly simplify typing of models.

mattpap:
fixes #7431


mattpap:
I will go ahead and merge this, as this conflicts with other TypeScript-related work I had planned.

PR #7435, opened 2018-01-18
allow push_notebook with empty events

### LINKED ISSUE #7426: allow push_notebook() to run when no change has occurred
As first mentioned in #7268, and discussed on gitter with @bryevdv:

A typical usage of an animated bokeh figure inside a jupyter notebook, when using the native ipywidgets - think `interact()` - would be to:

* first create a figure with dummy/default settings using `output_notebook()` to create a handle
* then call `interact()` to let the user change these settings, associated with a function that does `push_notebook(handle)` 

In many cases this implies that the very first call made by interact() will come with no change from the initialized figure; and even if not, nothing prevents the user from tweaking with her settings but releasing the mouse while no actual change was made; `interact()` would not - and should not - filter it out.

I hope this is a convincing argument to allow for `push_notebook()` to be called when no change has occurred since last time the figure was refreshed. 

To repeat some of the material from #7268:
Right now, with bokeh 0.12.13 and several ancestors that I tried as well, I get the following message upon the very first call to `push_notebook()`

    ValueError: PATCH-DOC message requires at least one event

bryevdv:
- [x] issues: fixes #7426
- [x] tests added / passed

### DIFF
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -250,6 +250,13 @@ def push_notebook(document=None, state=None, handle=None):
         return
 
     events = list(handle.doc._held_events)
+
+    # This is to avoid having an exception raised for attempting to create a
+    # PATCH-DOC with no events. In the notebook, we just want to silenty
+    # ignore calls to push_notebook when there are no new events
+    if len(events) == 0:
+        return
+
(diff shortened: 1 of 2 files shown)


bryevdv:
Should be uncontroversial

PR #7440, opened 2018-01-20
Fixed regression in notebook plot embedding

### LINKED ISSUE #7439: Regression displaying embedded bokeh plots outside the notebook
Using current bokeh master there appears to be a regression in embedded bokeh plots, which causes them to fail to render outside a notebook context (and probably also in JupyterLab). The error is the following:

```
Uncaught ReferenceError: Jupyter is not defined
    at r 
    at Object.i.embed_items_notebook 
    at embed_document 
```

which corresponds to this code in the embed.ts file:

```javascript
function _init_comms(target: string, doc: Document): void {
  if (Jupyter != null && Jupyter.notebook.kernel != null) {
    logger.info(`Registering Jupyter comms for target ${target}`)
```

which appears to be have been incorrectly converted from the original typescript:

```coffeescript
_init_comms = (target, doc) ->
  if Jupyter? and Jupyter.notebook.kernel?
    logger.info("Registering Jupyter comms for target #{target}")
```

If ``Jupyter`` is not defined in the namespace the new JS/TS code breaks while the ``Jupyter?`` check guarded against the non-existence of the variable.

philippjfr:
Simple fix for the regression described in #7439. Trying to open a notebook with embedded bokeh plots outside the notebook itself would cause it to fail to render the plots because an error occurred when trying to set up a Jupyter Comm. It now checks whether the Jupyter variable is defined.  

- [x] issues: fixes #7439

### DIFF
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -71,7 +71,7 @@ function _update_comms_callback(target: string, doc: Document, comm: Comm): void
 }
 
 function _init_comms(target: string, doc: Document): void {
-  if (Jupyter != null && Jupyter.notebook.kernel != null) {
+  if (typeof Jupyter !== 'undefined' && Jupyter.notebook.kernel != null) {
     logger.info(`Registering Jupyter comms for target ${target}`)
     const comm_manager = Jupyter.notebook.kernel.comm_manager
     const update_comms = (comm: Comm) => _update_comms_callback(target, doc, comm)


philippjfr:
Not sure what Better Code Hub is complaining about, I can't view it without giving them all kinds of permissions.

mattpap:
> Not sure what Better Code Hub is complaining about, (...)

Nothing significant. It doesn't consider what you actually changed, but the entire file, and points out some obvious unrelated rubbish.

bokehjs/src/coffee/embed.ts
```diff
@@ -71,7 +71,7 @@ function _update_comms_callback(target: string, doc: Document, comm: Comm): void
 }
 
 function _init_comms(target: string, doc: Document): void {
-  if (Jupyter != null && Jupyter.notebook.kernel != null) {
+  if (typeof Jupyter !== 'undefined' && Jupyter != null && Jupyter.notebook.kernel != null) {
```
mattpap: You could remove `Jupyter != null`, because the `undefined` case of this check is already considered by the new clause, and `null` case shouldn't happen, nor is permitted by the declaration of `Jupyter` (at the beginning of the file).

PR #7459, opened 2018-01-24
Update version number in package-lock.json

### LINKED ISSUE #7458: Update version number in package-lock.json

mattpap:
fixes #7458


bryevdv:
Seems like it should work like the others. I'm going to merge so we can just test in the next dev build

PR #7463, opened 2018-01-25
Define attributes and interfaces for all models

mattpap:
addresses #6481


mattpap:
@bryevdv, this applies only to the type system, so it should be safe to merge.

PR #7473, opened 2018-01-29
transform Date correctly on windows

### LINKED ISSUE #7460: bokeh.core.properties.Date#transform fails on Windows
Bokeh 0.12.14dev4

`bokeh.core.properties.Date#transform` catches `ValueError`, but it doesn't work on Windows.
On windows, the exception for invalid values is `OSError`.

bryevdv:
- [x] issues: fixes #7460
- [x] tests added / passed

This adds a catch for `OSError` to `Date.transform` which is the proper way to catch the exception on some platforms/versions. 

Note that `test_validate_milliseconds` was verified to fail without the associated code change, and is passing with the code change (on windows)

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -1071,7 +1071,7 @@ class Date(Property):
         if isinstance(value, (float,) + bokeh_integer_types):
             try:
                 value = datetime.date.fromtimestamp(value)
-            except ValueError:
+            except (ValueError, OSError):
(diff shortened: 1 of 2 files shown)


bryevdv:
@p-himik if you can look at/test this out it would be appreciated.

p-himik:
@bryevdv Checked it, it works.

bryevdv:
Thanks @p-himik

PR #7479, opened 2018-01-30
Fix regression in rendering (non)selection glyphs

### LINKED ISSUE #7478: Selection and nonselection glyph properties ignored
When setting style properties on the selection and nonselection glyphs these options are now ignored on latest master. It doesn't matter whether these options are set through the constructor or directly on the (non)selection glyphs.

Here is a simple example using the circle and line glyph but as far as I can tell all glyphs are affected:

```python

output_file('glyph_properties.html')

p = figure(x_range=(0, 10), y_range=(0, 10), width=600, height=600, tools=['tap'])
rc1 = p.circle([1, 2, 3], [1, 2, 3], color='green', selection_color='red',
               selection_alpha=0.2, nonselection_color='blue', size=20)
rc2 = p.line([3, 4, 5], [3, 4, 5], line_width=5, selection_color='red')

print(rc1.selection_glyph.fill_color)
print(rc1.nonselection_glyph.fill_color)
show(p)
```

As far as I can tell this regression was introduced in https://github.com/bokeh/bokeh/commit/264a1d31fdc1903c739d455ad17662ee034e45d6.

mattpap:
fixes #7478

### DIFF
--- a/examples/integration/glyphs/selection_nonselection.py
+++ b/examples/integration/glyphs/selection_nonselection.py
@@ -1,0 +1,22 @@
+from __future__ import absolute_import
+
+from bokeh.io import save
+from bokeh.plotting import figure
+
+p = figure(x_range=(0, 6), y_range=(0, 7), width=600, height=600, toolbar_location=None)
+
+def mkglyph(x, y, selection):
+    r = p.circle(
+        x, y,
+        color='green',
+        size=40,
+        selection_color='red',
+        selection_alpha=0.2,
+        nonselection_color='black')
+    r.data_source.selected['1d'] = dict(indices=selection)
+    return r
+
+mkglyph([1, 2, 3, 4, 5], [1, 2, 3, 4, 5], selection=[1, 3, 4])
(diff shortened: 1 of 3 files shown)


examples/integration/glyphs/selection_nonselection.py
```diff
@@ -0,0 +1,22 @@
...
+
+def mkglyph(x, y, selection):
+    r = p.circle(
+        x, y,
+        color='green',
+        size=40,
+        selection_color='red',
+        selection_alpha=0.2,
+        nonselection_color='black')
+    r.data_source.selected['1d'] = dict(indices=selection)
```
bryevdv: This will not work after @clairetang6 selection PR, it would need to use the new `Selection` model

mattpap: That PR isn't merged yet and I don't expect this PR being merged after.

bryevdv: I'm saying that you should put a link in the other PR with a note to make sure that updating it gets done.

PR #7485, opened 2018-02-01
only trigger set_data from FactorRange changes

### LINKED ISSUE #7475: setting range end or start through CustomJS not working
Setting the range start or end through a CustomJS callback is not functioning in bokeh 0.12.4dev5, python 3.6 on Linux or Windows in the chrome browser.  I found this when upgrading from 0.12.10, where this worked, to test some functionality.  The following code will illustrate the issue:

```python
from bokeh.models import CustomJS, Slider

output_file('testing_range_setting.html')

p = figure()
p.line(x=[1, 2, 3, 4, 5], y=[1, 2, 3, 4, 5])

callback = CustomJS(args=dict(x=p.x_range), code="""
var end = cb_obj.value;
x.end = end;
    """)

sld = Slider(start=0.1, end=10, value=1, step=.1,
             title="X-range", callback=callback)

show(column(sld, p))
```
I haven't been able to isolate the issue, but no exceptions are thrown in the browser.

### LINKED ISSUE #7484: Cannot select tools in certain examples
In current bokeh master selecting tools does not work in certain examples, specifically the problem can be seen in ``./examples/apps/selection_histogram.py`` when trying to select a tool other than the default pan tool.

I've tried bisecting the history to figure out what was causing it and the bisect points to this commit: https://github.com/bokeh/bokeh/commit/2b7ec420b091987840d8bfbde38af7e13cac1661 I currently have no idea how that commit could possibly be responsible for this issue but  so I double checked that the commit before it indeed works fine.

bryevdv:
- [x] issues: fixes #7475 fixes #7484
- [ ] tests added / passed

@philippjfr @mattpap this seems to fix both the issues above, as well as keep the fix for #7083 but more testing appreciated.

My best guess is the the more general original code resulted in some kind of perpetual event ping-pong that made UI interactions difficult

### DIFF
--- a/bokehjs/src/coffee/models/renderers/glyph_renderer.ts
+++ b/bokehjs/src/coffee/models/renderers/glyph_renderer.ts
@@ -11,6 +11,7 @@ import {difference, includes, range} from "core/util/array";
 import {extend, clone} from "core/util/object"
 import {Context2d} from "core/util/canvas"
 import {SelectionManager} from "core/selection_manager"
+import { FactorRange } from '../ranges/factor_range';
 
 export class GlyphRendererView extends RendererView {
   model: GlyphRenderer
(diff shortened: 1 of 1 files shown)


bryevdv:
OK I'm going to go ahead and merge this to get a dev build out

Issue #7488, opened 2018-02-02
renderers/glyph_renderer contains invalid syntax


mattpap:
https://github.com/bokeh/bokeh/blob/b93e359d067fa44ea184cb9fde331b940e055c8a/bokehjs/src/coffee/models/renderers/glyph_renderer.ts#L99-L105

Note that bokehjs still works, because TypeScript is able to recover from this, however there are compilation errors.

PR #7489, opened 2018-02-02
Improve state management for implicit toolbar panel

### LINKED ISSUE #7461: Unable to reuse sessions of a server application
Bokeh 0.12.14dev4

```python

def setup_plot(doc):
    p = figure()
    p.line([1, 2, 3], [3, 2, 1])
    doc.add_root(p)


io_loop = IOLoop.current()

server = Server(setup_plot, io_loop=io_loop)

server.start()
io_loop.start()
```

1. Start the code
2. Go to http://localhost:5006/?bokeh-session-id=1
3. Open JS console, check that there're no errors
4. Refresh the webpage
5. Behold `Error rendering Bokeh items  TypeError: Cannot read property '_left' of undefined` in the JS console

Preliminary analysis shows that the issue is most likely caused by `Plot._init_toolbar_panel` - it adds a new panel each time the page is refreshed. The old panel is not being removed properly at the top of the function because `this._toolbar_panel` is not set by the document JSON.

At this point, this stuff is rather beyond my expertise with Bokeh, so any pointers would be appreciated.

mattpap:
fixes #7461, #7488

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot.ts
+++ b/bokehjs/src/coffee/models/plots/plot.ts
@@ -4,7 +4,7 @@ import * as p from "core/properties"
 import {Color} from "core/types"
 import {LineJoin, LineCap} from "core/enums"
 import {Place, Location, OutputBackend} from "core/enums"
-import {find, removeBy} from "core/util/array"
+import {find, removeBy, includes} from "core/util/array"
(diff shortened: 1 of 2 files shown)


bryevdv:
Thanks @mattpap

PR #7495, opened 2018-02-05
Upgrade TypeScript to version 2.7.1

mattpap:
addresses #6481


bryevdv:
@mattpap since #6481 is so big but does not have sub-issues, let's make these PRs "issue PRs"

PR #7502, opened 2018-02-07
Resolve compilation errors around Number.isInteger

mattpap:
The previous fix was incorrect on the type-level, because `Number.isInteger` is not a part of the standard library (es5 + es2015.collections) that we use. This isn't needed for the release.

addresses #7373, PR #7493

### DIFF
--- a/bokehjs/src/coffee/polyfill.ts
+++ b/bokehjs/src/coffee/polyfill.ts
@@ -9,8 +9,8 @@ if (typeof Set !== "function") {
 }
 
 // ref: https://github.com/bokeh/bokeh/issues/7373
-if (!Number.isInteger) {
-  Number.isInteger = function(value: number): boolean {
+if (!(Number as any).isInteger) {
+  (Number as any).isInteger = function(value: number): boolean {
     return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
   }
 }


bryevdv:
@mattpap I guess I need to  figure out what about my process is keeping me from seeing these errors

mattpap:
> I guess I need to figure out what about my process is keeping me from seeing these errors

You should see them locally from any gulp task that involves `scripts:ts`. However, even if there are errors, this doesn't imply that the build fails. There are two reasons for this. One is gulp and its design. The other is that we use `noEmitOnError: false`, which allows to proceed with emitting js files even if there are errors, because you don't have to resolve all errors, especially type errors, to have correctly working code (for faster iteration between changes). This means that compilation errors won't trigger build failure on travis ci. You need to click on the first task and investigate manually, similarly to how image diff works. This is both good and bad. Bad because it can lead to situations like this, but good, because one can iterate in presence of errors (like in selection PR).

bryevdv:
I use the `setup.py develop --build-js` option, perhaps some output is getting swallowed (or maybe just missed amongst other)

mattpap:
> I use the setup.py develop --build-js option, perhaps some output is getting swallowed (or maybe just missed amongst other)

That's a possibility.

PR #7508, opened 2018-02-08
Fix typo in docs

victor-o-silva:


bryevdv:
Thanks for the catch and fix @victor-o-silva !

PR #7514, opened 2018-02-08
Notify the user if package.json changed and `npm install` is needed

### LINKED ISSUE #7513: Check if package.json changed before proceeding with build
At minimum this should notify the user that `package.json` changed and one needs to run `npm install`, or this should run npm automatically.

mattpap:
For now this only notifies the user. It's not possible to run `npm install` and keep `gulp task_name` for running the build. This could be achieved in future by adding a layer above `gulp` or dropping `gulp` altogether (which I've been contemplating for some time).

fixes #7513

### DIFF
--- a/bokehjs/package.json
+++ b/bokehjs/package.json
@@ -114,6 +114,7 @@
     "slickgrid": "https://github.com/bokeh/SlickGrid.git#6c1327b"
   },
   "scripts": {
+    "prepare": "node ./prepare.js",
     "build": "gulp build",
     "dev-build": "gulp dev-build"
   },
--- a/bokehjs/.gitignore
+++ b/bokehjs/.gitignore
@@ -10,3 +10,4 @@
 *.DS_Store
 npm-debug.log
 /test/.generated_defaults/
+/.package.json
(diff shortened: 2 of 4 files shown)


bryevdv:
@mattpap does this need to add `.package.json` to git ignore?

mattpap:
> does this need to add .package.json to git ignore?

Yes. Any other options here?

bryevdv:
> Yes. Any other options here? 

Not sure what you are asking? Adding it to `.gitignore` seems straightforward and appropriate

mattpap:
> Not sure what you are asking? Adding it to .gitignore seems straightforward and appropriate

I was just confused by your question.

bokehjs/package.json
```diff
@@ -114,6 +114,7 @@
     "slickgrid": "https://github.com/bokeh/SlickGrid.git#6c1327b"
   },
   "scripts": {
+    "prepare": "node ./prepare.js",
```
canavandl: Is this documented anywhere? I had to find and read this PR in order to build bokeh locally. @mattpap

PR #7522, opened 2018-02-13
Fix typo

BioGeek:
type: bug
tag: docs


bryevdv:
Thanks for the catch and the fix @BioGeek

PR #7533, opened 2018-02-16
Fixed FutureWarning in json serialization of floating values

### LINKED ISSUE #7526: JSONEncoder warning when using latest NumPy
When using the latest version of NumPy (1.14.0) the JSON encoder generates warnings during the serialization of float arrays.

```
/Users/philippjfr/miniconda/envs/intake/lib/python3.6/site-packages/bokeh/core/json_encoder.py:80: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  elif np.issubdtype(type(obj), np.float):
```

The fix should be simple, replacing ``np.float`` with ``np.floating``.

philippjfr:
The most recent version of NumPy raises a FutureWarning during serialization because the np.issubdtype expects a scalar baseclass instead of a specific dtype. Replacing np.float with np.floating addresses this issue.

- [x] issues: fixes #7526

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -77,7 +77,7 @@ class BokehJSONEncoder(json.JSONEncoder):
             return dict(start=obj.start, stop=obj.stop, step=obj.step)
 
         # NumPy scalars
-        elif np.issubdtype(type(obj), np.float):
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good; thanks for the fix!

PR #7539, opened 2018-02-19
Clean up tmp files and processes used by bokeh.io.export

### LINKED ISSUE #7537: export_png does not clean up temporary files before exit
**ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)**

'vanilla' Python installation.
No Anaconda, no virtualenv.
Packages installed with pip (9.0.1)

Python (3.5.0)

bokeh (0.12.13)

DateTime (4.2)
Jinja2 (2.10)
Pillow (5.0.0)
PyYAML (3.12)
requests (2.18.4)
selenium (3.8.1)
six (1.11.0)
tornado (4.5.3)

PhantomJS (2.1.1) - only version that will work on my platform.

PuppyLinux 5.6.1 (32-bit, Ubuntu "Precise", kernel 3.2.44)

**Description of expected behaviour and the observed behaviour**

_Expected:_

export_png creates a temporary .html file in the following code fragment, and I would expect either to be able to call export_png with a parameter that allows one to save or delete that temporary file after it is used, or deletes the temporary file by default. 

(I don't know enough bokeh to know whether the temporary file might be re-used after export_png has been called and completed.)

```
def save_layout_html(obj, resources=INLINE, **kwargs):
    '''

    '''
    resize = False
    if kwargs.get('height') is not None or kwargs.get('width') is not None:
        # Defer this import, it is expensive
        from ..models.plots import Plot
        if not isinstance(obj, Plot):
            warn("Export method called with height or width kwargs on a non-Plot layout. The size values will be ignored.")
        else:
            resize = True
            old_height = obj.plot_height
            old_width = obj.plot_width
            obj.plot_height = kwargs.get('height', old_height)
            obj.plot_width = kwargs.get('width', old_width)

    html_path = NamedTemporaryFile(suffix=".html").name
    save(obj, filename=html_path, resources=resources, title="")

    if resize:
        obj.plot_height = old_height
        obj.plot_width = old_width

    return html_path
```
This fn is called from
```
 def get_screenshot_as_png(obj, driver=None, **kwargs):
  '''

    '''
    webdriver = import_required('selenium.webdriver',
                                'To use bokeh.io.export_png you need selenium ' +
                                '("conda install -c bokeh selenium" or "pip install selenium")')

    Image = import_required('PIL.Image',
                            'To use bokeh.io.export_png you need pillow ' +
                            '("conda install pillow" or "pip install pillow")')
    # assert that phantomjs is in path for webdriver
    detect_phantomjs()

    html_path = save_layout_html(obj, **kwargs)

    web_driver = driver if driver is not None else webdriver.PhantomJS(service_log_path=devnull)

    web_driver.get("file:///" + html_path)
    web_driver.maximize_window()

    ## resize for PhantomJS compat
    web_driver.execute_script("document.body.style.width = '100%';")

    wait_until_render_complete(web_driver)

    png = web_driver.get_screenshot_as_png()

    b_rect = web_driver.execute_script(_BOUNDING_RECT_SCRIPT)

    if driver is None: # only quit webdriver if not passed in as arg
        web_driver.quit()

    image = Image.open(io.BytesIO(png))
    cropped_image = _crop_image(image, **b_rect)

    return cropped_image
```
_Observed:_

Each time export_png is called, it creates a new uniquely-named file in my /tmp directory. 
As my application (a Twitter 'bot' that replies to enquiries and includes 2x bar charts as png Tweet attachments) runs 24/7, the number or tempfiles grows incrementally. Each file (in this use case) is ~667Kb. Unless manually deleted, there appears no limit to the number of files that might be found.

I only noticed this shortcoming when looking for reasons why I intermittently get an (OSError [Errno 9] Bad file descriptor) when export_png is called. 

It's easy enough to add a line of code in either my app or to customise my copy of export_png to remove the temporary file afterwards, but it feels like a shortcoming in the module as well and thought it worth flagging up.

(Though bokeh in truth is overkill for my needs in this use instance and I would prefer something much lighter)

**Complete, minimal, self-contained example code that reproduces the issue:**

```

p.vbar(x=[1, 2, 3], width=0.5, bottom=0,
       top=[1.2, 2.5, 3.7], color="firebrick")
       
export_png(p, filename="plot.png")
```
**Stack traceback and/or browser JavaScript console output**

None

**Screenshots or screencasts of the bug in action**

capture30828

mattpap:
fixes #7537, #6525

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -25,6 +25,7 @@ from bokeh.util.api import general, dev ; general, dev
 
 # Standard library imports
 import io
+import signal
 from os.path import abspath, devnull
 from tempfile import NamedTemporaryFile
 from warnings import warn
@@ -179,7 +180,7 @@ def get_screenshot_as_png(obj, driver=None, **kwargs):
         b_rect = web_driver.execute_script(_BOUNDING_RECT_SCRIPT)
 
         if driver is None: # only quit webdriver if not passed in as arg
-            web_driver.quit()
+            terminate_web_driver(web_driver)
 
     image = Image.open(io.BytesIO(png))
     cropped_image = _crop_image(image, **b_rect)
@@ -204,7 +205,7 @@ def get_svgs(obj, driver=None, **kwargs):
         svgs = web_driver.execute_script(_SVG_SCRIPT)
 
         if driver is None: # only quit webdriver if not passed in as arg
-            web_driver.quit()
+            terminate_web_driver(web_driver)
 
     return svgs
 
@@ -256,6 +257,11 @@ def wait_until_render_complete(driver):
         severe_errors = [l for l in browser_logs if l.get('level') == 'SEVERE']
         if len(severe_errors) > 0:
             log.warn("There were severe browser errors that may have affected your export: {}".format(severe_errors))
+
+@dev((1,0,0))
+def terminate_web_driver(driver):
+    driver.service.process.send_signal(signal.SIGTERM)
+    driver.quit()
 
 #-----------------------------------------------------------------------------
 # Private API
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -160,24 +160,26 @@ def get_screenshot_as_png(obj, driver=None, **kwargs):
                             'To use bokeh.io.export_png you need pillow ' +
                             '("conda install pillow" or "pip install pillow")')
 
-    html_path = save_layout_html(obj, **kwargs)
+    with _tmp_file() as tmp:
+        html_path = tmp.name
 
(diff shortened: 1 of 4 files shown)


bokeh/io/export.py
```diff
@@ -222,15 +227,12 @@ def save_layout_html(obj, resources=INLINE, **kwargs):
             obj.plot_height = kwargs.get('height', old_height)
             obj.plot_width = kwargs.get('width', old_width)
 
-    html_path = NamedTemporaryFile(suffix=".html").name
     save(obj, filename=html_path, resources=resources, title="")
```
bryevdv: I think I'd like this better if `save` could just accept the named tmpfile object directly, but not worth changing here.

PR #7542, opened 2018-02-19
Fix typo in Figure.vbar_stack docstring

victor-o-silva:


bryevdv:
Thanks for the catch and the fix @victor-o-silva !

PR #7552, opened 2018-02-21
Bryanv/2507 grid tick bounds

### LINKED ISSUE #2507: Have Grid that respects bounds of axis
In `examples/glyph/iris.py` the bounds parameter of LinearAxis is demonstrated.

Grids are also added to the plot.  On zoom out, the axis does not gain any extra tick marks, but the grid lines do continue.

iris_grid_bounds

bryevdv:
- [x] issues: fixes #2507
- [x] tests added / passed

With axis bounds but auto grid bounds:

screen shot 2018-02-21 at 12 09 09

With axis bounds and different explicit grid bounds:

screen shot 2018-02-21 at 12 09 20


bryevdv:
My only question @bokeh/dev: With this change it is hard to get the original behaviour of "full visible range for grid" if axis bounds are set. Should we add a new non-default `"range"` or `"full"` option for grid bounds in addition to `"auto"` ?

mattpap:
@bryevdv, there are compilation errors (https://travis-ci.org/bokeh/bokeh/jobs/344500810#L676-L683).

bokehjs/src/coffee/models/tickers/continuous_ticker.ts
```diff
@@ -80,11 +80,17 @@ export abstract class ContinuousTicker extends Ticker<number> {
       const minor_interval = interval / num_minor_ticks
       const minor_offsets = range(0, num_minor_ticks).map((i) => i*minor_interval)
       for (const x of minor_offsets.slice(1)) {
-        minor_ticks.push(ticks[0] - x)
+        mt = ticks[0] - x
```
mattpap: `const`.

bokehjs/src/coffee/models/grids/grid.ts
```diff
@@ -194,6 +195,11 @@ export class Grid extends GuideRenderer {
       //  end = null
     } else {
       [start, end] = range_bounds
+      for (const axis of this.plot.select(Axis)) {
```
mattpap: For now you have to assert the expected type with `for (const axis of this.plot.select(Axis) as Axis[])`.

bryevdv:
is it possible to make compile errors fail `gulp build`?

mattpap:
> is it possible to make compile errors fail gulp build?

I will look into this. The thing is that we don't want to fail `scripts:ts` task, because that wouldn't allow to e.g. generate js bundles even if errors are present, but fail just `build` task if an sub-task reported failure. That's tricky with gulp.

PR #7556, opened 2018-02-22
More generic sphinx doc sorting.

### LINKED ISSUE #7505: Sphinx Bokeh Extension Script Directory Sorting
When adding specified directories to the `bokeh_plot_pyfile_include_dirs` the sphinx sorting of the source docs does not properly sort based on the directories or python files containing bokeh.

[This line](https://github.com/bokeh/bokeh/blob/master/bokeh/sphinxext/bokeh_plot.py#L268) seems to be catered specifically to the bokeh doc structure with the 'examples' directory. 

Ideally the sorting should put any directories specified in the sphinx configuration variable `bokeh_plot_pyfile_include_dirs` to be parsed first. I ran into the issue where not all of the python files containing bokeh code would be parsed before the `bokeh-plot` directive in a source doc RST was found which would result in a sphinx error when trying to build.

I was able to get my structure working which has two bokeh source directories under my normal RST source directory so it is something like:

Top level doc repo
--> source
|----> bokeh_dir_1
|----> Rest
|----> Of
|----> RSTs
|----> Z_bokeh_dir_2

I swapped out the previously linked line with:

```python
docnames.sort(key=lambda x: 0 if x.startswith(tuple(env.app.config.bokeh_plot_pyfile_include_dirs)) else 1)
```
This sorting method probably needs improvement, I don't think it will work if the desired directories are in lower depths of directories.

hallyourbase:
Addresses #7505 

- [x] issues: fixes #7505
- [ ] tests added / passed


bryevdv:
@hallyourbase thanks for the PR! We are having some issue on Travis CI with builds outside @bokeh/dev so I will check this out locally on the bokeh docs then merge later today assuming no issues.

bryevdv:
OK unfortunately this will need some iteration. Thinking more about this code, I recall that the reason it does what it does is that the custom sphinx parser needs to be able to see those files first, as later steps depend on the output. So this change breaks the project docs. I'd say this state of coupling points to a deeper design issue, but it's not one I will be able to dig into. In any case, using `bokeh_plot_pyfile_include_dirs` for sorting is not correct.  What could probably work as a band aid solution is to add a new config option to specify how the sorting should be done. There are probably many ways this might be done, for the projects needs we just need to be able to spell a config that causes `examples` to be sorted first as they currently are.

bryevdv:
actually the Bokeh docs build with this diff, which also removes all sorting altogether:
```
diff --git a/bokeh/sphinxext/bokeh_plot.py b/bokeh/sphinxext/bokeh_plot.py
index 9b04e2796..4e083c693 100644
--- a/bokeh/sphinxext/bokeh_plot.py
+++ b/bokeh/sphinxext/bokeh_plot.py
@@ -238,7 +238,10 @@ class BokehPlotDirective(Directive):
             # handle examples external to the docs source, e.g. gallery examples
             else:
                 app.debug("[bokeh-plot] handling external example in %r: %s", env.docname, self.arguments[0])
-                source = open(self.arguments[0]).read()
+                try:
+                    source = open(self.arguments[0]).read()
+                except FileNotFoundError:
+                    source = open(join('source', self.arguments[0])).read()
                 source = decode_utf8(source)
                 docname = env.docname.replace("/", "-")
                 js_name = "bokeh-plot-%s-external-%s.js" % (docname, uuid4().hex)
@@ -265,7 +268,6 @@ class BokehPlotDirective(Directive):
         return result

 def env_before_read_docs(app, env, docnames):
-    docnames.sort(key=lambda x: 0 if x.startswith(tuple(env.app.config.bokeh_plot_pyfile_include_dirs)) else 1)
     for name in [x for x in docnames if env.doc2path(x).endswith(".py")]:
         if not name.startswith(tuple(env.app.config.bokeh_plot_pyfile_include_dirs)):
             env.found_docs.remove(name)
```
Does this work for your docs? If so there are two options:

* merge this as a temp fix with comment and new issue to investigate further later, or 

*  Merge just the second chunk that removes sorting, but update Bokeh `.rst` files to change things that look like this:
    ```
    .. bokeh-plot:: docs/user_guide/examples/plotting_title_basic.py
       :source-position: above
    ```
    to look lik this:
    ```
    .. bokeh-plot:: source/docs/user_guide/examples/plotting_title_basic.py
        :source-position: above
    ```

Thoughts @hallyourbase ?

bryevdv:
The mass update can be made with a command like this:
```
find . -name '*.rst' -print0 | xargs -0 perl -pi -e 's/bokeh-plot:: docs/bokeh-plot:: source\/docs/g'
```
If you do that, and delete the `docnames.sort` line altogether, then the Bokeh docs build. If that also works for your docs build, I would say that is the best way forward.

hallyourbase:
@bryevdv Just the second change with the updated paths seems to work after a quick test rebuild of our docs. I agree, I think that is probably the best way forward. Thanks for taking a look.

bryevdv:
@hallyourbase OK if you push those changes I will go ahead and merge

bryevdv:
We might also want to add a migration note regarding this change. I don't think there are alot of people using these sphinx extensions but I've been wrong about that before, and in any case it's worth having migration instructions to point at.

hallyourbase:
Can do on the docs update. Please expand on the migration note comment so I can make sure to do it correctly. I am not sure I follow exactly what is needed for that in terms of is it an RST doc modification or something else.

bryevdv:
@hallyourbase brilliant, for the release not, you'd just edit the file I linked above to add a new subsection, e.g. 

```
Sphinx Plot Extension
~~~~~~~~~~~~~~~~~~~~~

The `bokeh-plot` Sphinx extension previously assumed the location of scripts was relative 
to a `source` subdirectory of the Sphinx project. This is no longer the case. Now, the full 
relative path from the Sphinx project (i.e. from the `Makefile`) should be provided.
```

Feel free to add/edit that as desired.

bryevdv:
@hallyourbase I think maybe you were working off an older revision of master? You could try merging current master in but it might be simplest just to make a new PR off the current master

hallyourbase:
@bryevdv yep sorry about that, still new to working with forks and PRs. I think I fixed it?

sphinx/source/docs/releases/0.12.15.rst
```diff
@@ -21,14 +21,33 @@ Grid Bounds
...
+
+.. code-block:: none
+
+    .. bokeh-plot:: docs/user_guide/examples/plotting_title_basic.py
+
+Should now be:
+
+.. code-block:: none
+
+    .. bokeh-plot:: source/docs/user_guide/examples/plotting_title_basic.py
```
bryevdv: The part after "So what was once" is only true for a particular Sphinx project layout, which is common, but not required. I would just leave that part off.

bryevdv:
@hallyourbase yes looks much better now! I had one comment but otherwise ready to merge when tests are green.

sphinx/source/docs/releases/0.12.15.rst
hallyourbase: Fair enough, I was just trying to provide an example of what an example change might have to be.

bryevdv: If you want to keep it we'd put need to be more specific, e.g. "For example, if your sphinx project has a layout like...."

hallyourbase: If you don't think it is necessary then removing it is fine.

bryevdv:
Thanks @hallyourbase !

hallyourbase:
@bryevdv It seems like there may be some unintended effects as a result of this change. Now it appears the bokeh plots that are in the `source` directory in my docs are not passing [this check](https://github.com/bokeh/bokeh/blob/master/bokeh/sphinxext/bokeh_plot.py#L234) and as such are being treated as external plots.  I only noticed this because my Emacs Flycheck was giving an error about not finding the plot py files even though they are in source. I think this might be due to the addition of the `source/` prefix to the paths in the directive calls. Worth noting that my docs still build fine and all of the plots appear to be unchanged from how they were before. Thoughts?

bryevdv:
Ah, we will probably have to find a way to make things work with the old directive paths then. Otherwise it's probably duplicating every plot since the custom py parser will process them too

bryevdv:
@hallyourbase I'm going to go ahead and revert this PR (while it's still possible before any other newer PRs cause a conflict) I will see if I can find another way forward today.

bryevdv:
Well never mind, GH already won't revert it automatically. I will try to have a new PR soon.

bryevdv:
@hallyourbase I believe in fact that changing the original sort statement to this:
```
# sort to make sure the custom Python file parser gets to process any plot
# scripts first, since it will save the plot output for use later. 
docnames.sort(key=lambda x: 0 if env.doc2path(x).endswith(".py") else 1)
```
will accomplish the real priority (making sure the py files come first) without relying on specifics of the Bokeh docs layout. Can you test it this right quick? If it works do you want to make a PR or if not I am happy to.

hallyourbase:
@bryevdv That change and reverting the plot paths used with the directive to omit the `source/` works great. That is a good solution which you are correct in the statement that was the original priority for the issue and the PR. I can make a PR later today but likely not until 7PM EST. So it is up to you if you would like to take care of it before then. Thanks for taking a look.

bryevdv:
No hurry, I'd rather you have the opportunity. Thanks for pushing on this!

hallyourbase:
@bryevdv  [New PR is up](https://github.com/bokeh/bokeh/pull/7588). Let me know if I missed anything.

PR #7560, opened 2018-02-23
Fail travis ci tests when TS errors are present

### LINKED ISSUE #7559: Fail travis ci tests when there are TS compilation errors
For developer's convenience allow to run all tests, but fail the entire run at the end.

mattpap:
fixes #7559

### DIFF
--- a/_setup_support.py
+++ b/_setup_support.py
@@ -275,10 +275,8 @@ def build_js():
     sys.stdout.flush()
     os.chdir('bokehjs')
 
-    if sys.platform != "win32":
-        cmd = [join('node_modules', '.bin', 'gulp'), 'build']
-    else:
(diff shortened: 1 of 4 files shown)


mattpap:
This approach allows to run all tests even if there are TS errors present. This allows for continuous testing without the need for fixing all of non-critical errors (type errors). Perhaps `--emit-error` could be used with `setup.py` workflow.

mattpap:
An alternative solution would be to store compilation outcome in `build` stage, and use this information to fail `deploy` stage.

bryevdv:
@mattpap I think I'd prefer if this was wrapped into the existing `codebase` build group since that's conceptually the same category of checking.

Issue #7563, opened 2018-02-24
FuncTickFormatter option to receive all ticks at once


kernc:
`FuncTickFormatter`'s `code` snippet only formats a single tick at a time. But sometimes more context is needed (i.e. all ticks at once) to format them correctly. 

As requested on the mailing list:
* [how many ticks will be for the graph in FuncTickFormatter?](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/1AoZUO4hLKs)
* [Datetime axis, missing values skipped, adaptive formatting](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/t3HkalO4TGA)

PR #7564, opened 2018-02-24
Add `pass_all_ticks` argument to FuncTickFormatter

kernc:
Was unsure of the argument name. `array` seemed more explicit than `vectorized`. Open to suggestions.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/7563
- [x] tests added / passed

### DIFF
--- a/bokeh/models/formatters.py
+++ b/bokeh/models/formatters.py
@@ -322,9 +322,11 @@ class FuncTickFormatter(TickFormatter):
     """)
 
     code = String(default="", help="""
-    A snippet of JavaScript code that reformats a single tick to the desired
-    format. The variable ``tick`` will contain the unformatted tick value and
+    A snippet of JavaScript code that reformats ticks to the desired format.
+    The variable ``tick`` will contain the unformatted tick value and
     can be expected to be present in the code snippet namespace at render time.
+    If ``pass_all_ticks`` is ``True``, ``tick`` will contain an array of ticks,
+    with array of strings expected in return.
 
     Example:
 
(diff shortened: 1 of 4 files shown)


bryevdv:
Hi @kernc Thanks for for PR! FYI we are currently experiencing some build issues with PRs from forks on TravisCI which may hold things up a bit. I am generally in favor of the implementation and change (especially the test, thank you!) My only comment is that I would prefer a property name more descriptive than `array`. Perhaps `pass_all_ticks` with a default value of `False`. Do you have any thoughts?

kernc:
My thoughts are fully aligned in that the more explicit the identifiers, while still relatively short, so much the better. :+1: Amended.

bryevdv:
@kernc great, getting a bit even more in the weeds, what d you think about defining the function differently based on the property, so that the parameter is `tick` in the default case, and `ticks` in the `pass_all_ticks=True` case?  I would not demand this change, but I think it might be a little nicer for users (the help string would need to be very clear on the two modes of operation though)

kernc:
No strong opinions. Implementation-wise, the change is negligible. Did my best with the strings, but feel free to reword.

mattpap:
From the issue:

> But sometimes more context is needed (i.e. all ticks at once) to format them correctly.

This is very useful to have, however, I will argue against this implementation. It's not a good design to have a class work completely differently by a flip of a switch. I would rather go with a well established pattern in JS world and have a formatter function get current tick, index and array of all ticks, like `Array.map()` does (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map). The only argument for the current approach is if there was need to cache stuff or otherwise pass data between consecutively formatted ticks, though this could be achieved with the "map" approach as well. However, there are no new tests or examples that would actually use contextual information, nor the more advanced use case, so I'm not sure if that was intended to be supported or not.

kernc:
Right. Proposed alternative in https://github.com/bokeh/bokeh/pull/7581.

PR #7565, opened 2018-02-25
Issue #7474 addressed: image alpha implemented

### LINKED ISSUE #7474: Alpha not working on images
bokeh: 0.12.10
python: 2.7
notebook: 5.0.0
jupyter: 1.0.0
jupyter_client 5.1.0
jupyter_console: 5.2.0
jupyter_core: 4.3.0
ipython: 5.4.1
ipython_genutils: 0.2.0
OS: ubuntu 14.04
broswer: Safari Version 11.0.1 (13604.3.5)

Setting the alpha of an image should change its transparency, but setting alpha has no effect (I tried setting to 0,.5,1,128,255).

```
from skimage import data
image = data.coins() #Any image will do, but I didn't find a sample image in bokeh's sample data
fig_overlay = figure(x_range=(0, image.shape[1]), y_range=(0, image.shape[0]),
                     plot_width=image.shape[1]*1, plot_height=image.shape[0]*1)
bkimg1 = fig_overlay.image(image=[image], x=0, y=0, dw=image.shape[1], dh=image.shape[0], alpha=.5) 
bkimg2 = fig_overlay.image(image=[image], x=100, y=100, dw=image.shape[1], dh=image.shape[0], alpha=.5) #changing alpha has no effect on the resulting figure.
show(fig_overlay)
```

marcuskd:
New ColorMapper attribute 'alpha', can be assigned directly or via 'global_alpha' Image kwarg.
Example provided in image_alpha.py.

- [x] issues: fixes #7474
- [x] tests added / passed

### DIFF
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -324,9 +324,10 @@ class Image(XYGlyph):
         if 'palette' in kwargs and 'color_mapper' in kwargs:
             raise ValueError("only one of 'palette' and 'color_mapper' may be specified")
         elif 'color_mapper' not in kwargs:
-            # Use a palette (given or default)
+            # Use a palette and alpha value (given or default)
             palette = kwargs.pop('palette', 'Greys9')
-            mapper = LinearColorMapper(palette)
+            alpha = kwargs.pop('global_alpha', 1.0)
+            mapper = LinearColorMapper(palette=palette, alpha=alpha)
             kwargs['color_mapper'] = mapper
 
         super(Image, self).__init__(**kwargs)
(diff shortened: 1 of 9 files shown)


bryevdv:
@marcuskd Thanks for the PR! I will have to consider this, as it's not the approach I had in mind, which was:

* add a real `Bool` Bokeh property `global_alpha` to the `Image` class
* have the BokehJS image renderer apply the global alpha right before drawing
* add some convenience on python side to convert `alpha` to `global_alpha` in the  `image` glyph method

marcuskd:
@bryevdv It's possible I've over-complicated this: I thought the alpha needed to belong to the ColorMapper, but having looked more closely at the ImageURL glyph I can see how your suggested approach is probably neater. Hang fire on this PR, I'll have a look at the other approach...

marcuskd:
Closing: produced alternative solution.

PR #7567, opened 2018-02-25
Add support for JupyterLab comms

### LINKED ISSUE #7566: Add support for push_notebook in JupyterLab
In previous releases support for displaying bokeh plots in JupyterLab was added but due to difficulties handling Jupyter Comms in JupyterLab support for ``push_notebook`` was not yet enabled. It would be good to expose this functionality by extending the [JupyterLab bokeh extension](https://github.com/bokeh/jupyterlab_bokeh) and then adding support in ``BokehJS.embed._init_comms``.

philippjfr:
This PR adds support for using ``push_notebook`` in JupyterLab when combined with https://github.com/bokeh/jupyterlab_bokeh/pull/20. The way it works is that it defines a ``Bokeh.embed.kernels`` registry which the JupyterLab extension registers kernels with on display. This allows the embed functions to register the Jupyter comm to send data from Python to JS.

I'd be happy to hear a different approach, and or make ``Bokeh.embed.kernels`` private to avoid cluttering the namespace. I've also made the ``Bokeh.embed.Receiver`` object public to allow other extensions (e.g. HoloViews) to register their own message handlers on platforms other than Jupyter.

- [x] issues: fixes #7566
- [ ] tests added / passed

@canavandl @bryevdv

### DIFF
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -119,9 +119,7 @@ function _init_comms(target: string, doc: Document): void {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
   } else {
-    console.warn('Jupyter notebooks comms not available. push_notebook() will not function.'
-                 'If running JupyterLab ensure the latest jupyterlab_bokeh extension is installed.'
-				 'In an exported notebook this warning is expected.');
+    console.warn(`Jupyter notebooks comms not available. push_notebook() will not function. If running JupyterLab ensure the latest jupyterlab_bokeh extension is installed. In an exported notebook this warning is expected.`);
   }
 }
 
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -33,7 +33,7 @@ declare interface CommMessage {
 declare interface Comm {
   target_name: string
   on_msg: (msg: CommMessage) => void
-  onMsg: Function
+  onMsg: (this: Document, receiver: Receiver, comm_msg: CommMessage) => void
 }
 
 declare interface Jupyter {
@@ -51,6 +51,7 @@ declare var Jupyter: Jupyter | undefined
 
 export const kernels: {[key: string]: any} = {}
 
+// Exported to allow external libraries to define custom message handlers
 export {Receiver}
 
 // Matches Bokeh CSS class selector. Setting all Bokeh parent element class names
@@ -72,8 +73,10 @@ function _update_comms_callback(target: string, doc: Document, comm: Comm): void
   if (target == comm.target_name) {
     const r = new Receiver()
     if (comm.on_msg) {
+      // Classic Notebook
       comm.on_msg(_handle_notebook_comms.bind(doc, r))
     } else {
+      // JupyterLab
       comm.onMsg = _handle_notebook_comms.bind(doc, r)
     }
   }
@@ -97,7 +100,7 @@ function _init_comms(target: string, doc: Document): void {
     } catch (e) {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
-  } else if (doc.roots()[0].id in kernels)) {
+  } else if (doc.roots()[0].id in kernels) {
     logger.info(`Registering JupyterLab comms for target ${target}`)
     const kernel = kernels[doc.roots()[0].id]
     for (const id in kernel._comms) {
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -33,6 +33,7 @@ declare interface CommMessage {
 declare interface Comm {
   target_name: string
   on_msg: (msg: CommMessage) => void
+  onMsg: Function
 }
 
 declare interface Jupyter {
@@ -47,6 +48,10 @@ declare interface Jupyter {
 }
 
 declare var Jupyter: Jupyter | undefined
+
+export const kernels: {[key: string]: any} = {}
+
+export {Receiver}
 
 // Matches Bokeh CSS class selector. Setting all Bokeh parent element class names
 // with this var prevents user configurations where css styling is unset.
@@ -66,15 +71,19 @@ function _handle_notebook_comms(this: Document, receiver: Receiver, comm_msg: Co
 function _update_comms_callback(target: string, doc: Document, comm: Comm): void {
   if (target == comm.target_name) {
     const r = new Receiver()
-    comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    if (comm.on_msg) {
+      comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    } else {
+      comm.onMsg = _handle_notebook_comms.bind(doc, r)
+    }
   }
 }
 
 function _init_comms(target: string, doc: Document): void {
+  const update_comms = (comm: Comm) => _update_comms_callback(target, doc, comm)
   if (typeof Jupyter !== 'undefined' && Jupyter.notebook.kernel != null) {
     logger.info(`Registering Jupyter comms for target ${target}`)
     const comm_manager = Jupyter.notebook.kernel.comm_manager
-    const update_comms = (comm: Comm) => _update_comms_callback(target, doc, comm)
     for (const id in comm_manager.comms) {
       const promise = comm_manager.comms[id]
       promise.then(update_comms)
@@ -88,8 +97,25 @@ function _init_comms(target: string, doc: Document): void {
     } catch (e) {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
-  } else
+  } else if (doc.roots()[0].id in kernels)) {
+    logger.info(`Registering JupyterLab comms for target ${target}`)
+    const kernel = kernels[doc.roots()[0].id]
+    for (const id in kernel._comms) {
+      const promise = kernel._comms[id]
+      promise.then(update_comms)
+    }
+    try {
+      kernel.registerCommTarget(target, (comm: Comm) => {
+        logger.info(`Registering JupyterLab comms for target ${target}`)
+        const r = new Receiver()
+        comm.onMsg = _handle_notebook_comms.bind(doc, r)
+      })
+    } catch (e) {
+      logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
+    }
+  } else {
     console.warn('Jupyter notebooks comms not available. push_notebook() will not function');
+  }
 }
 
 function _create_view(model: HasProps): DOMView {
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -56,9 +56,6 @@ declare var Jupyter: Jupyter | undefined
 
 export const kernels: {[key: string]: Kernel} = {}
 
-// Exported to allow external libraries to define custom message handlers
-export {Receiver}
-
 // Matches Bokeh CSS class selector. Setting all Bokeh parent element class names
 // with this var prevents user configurations where css styling is unset.
 export const BOKEH_ROOT = "bk-root"
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -119,7 +119,9 @@ function _init_comms(target: string, doc: Document): void {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
   } else {
-    console.warn('Jupyter notebooks comms not available. push_notebook() will not function');
+    console.warn('Jupyter notebooks comms not available. push_notebook() will not function.'
+                 'If running JupyterLab ensure the latest jupyterlab_bokeh extension is installed.'
+				 'In an exported notebook this warning is expected.');
   }
 }
 
--- a/bokehjs/src/coffee/embed.ts
+++ b/bokehjs/src/coffee/embed.ts
@@ -38,10 +38,8 @@ export declare interface Comm {
 
 export declare interface Kernel {
   comm_manager: {
-    comms: {[key: string]: Promise<Comm>},
     register_target: (target: string, fn: (comm: Comm) => void) => void,
   },
-  _comms: {[key: string]: Promise<Comm>},
   registerCommTarget: (target: string, fn: (comm: Comm) => void) => void,
 }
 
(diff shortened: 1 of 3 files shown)


bokehjs/src/coffee/embed.ts
```diff
@@ -33,6 +33,7 @@ declare interface CommMessage {
 declare interface Comm {
   target_name: string
   on_msg: (msg: CommMessage) => void
+  onMsg: Function
```
canavandl: should be ``(this: Document, receiver: Receiver, comm_msg: CommMessage) => void`` instead of Function. Or (better yet), define an interface and share it here and with ``_handle_notebook_comms``.

bokehjs/src/coffee/embed.ts
```diff
@@ -48,6 +49,10 @@ declare interface Jupyter {
 
 declare var Jupyter: Jupyter | undefined
 
+export const kernels: {[key: string]: any} = {}
+
+export {Receiver}
```
canavandl: Is this for Holoviews? AFAICT jupyterlab_bokeh doesn't require it.

bokehjs/src/coffee/embed.ts
```diff
@@ -66,15 +71,19 @@ function _handle_notebook_comms(this: Document, receiver: Receiver, comm_msg: Co
 function _update_comms_callback(target: string, doc: Document, comm: Comm): void {
   if (target == comm.target_name) {
     const r = new Receiver()
-    comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    if (comm.on_msg) {
+      comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    } else {
+      comm.onMsg = _handle_notebook_comms.bind(doc, r)
```
canavandl: I'd add inline comments explain the difference here (classic notebook vs JLab)

bokehjs/src/coffee/embed.ts
philippjfr: That's right, HoloViews uses it to make a generic message handler which should work outside Jupyter as well. Would be really helpful if this can be public.

bokehjs/src/coffee/embed.ts
```diff
@@ -88,8 +97,25 @@ function _init_comms(target: string, doc: Document): void {
     } catch (e) {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
-  } else
+  } else if (doc.roots()[0].id in kernels)) {
```
canavandl: ``} else if (doc.roots()[0].id in kernels)) {`` there's an extra ")" here

canavandl:
I'm getting a ``Uncaught Error: property ColumnDataSource.selection_policy wasn't declared`` JS exception on ``show`` when I try to run the "Basic Usage.ipynb" example with this PR, so I'm unable to test.

philippjfr:
>I'm getting a Uncaught Error: property ColumnDataSource.selection_policy wasn't declared JS exception on show when I try to run the "Basic Usage.ipynb" example with this PR, so I'm unable to test.

Strange, just tried it with your new JupyterLab PR and it works fine for me. Only had to switch to INLINE resources.

canavandl:
>I'm getting a Uncaught Error: property ColumnDataSource.selection_policy wasn't declared JS exception on show when I try to run the "Basic Usage.ipynb" example with this PR, so I'm unable to test.

I had a stupid path issue. Nvm

bokehjs/src/coffee/embed.ts
```diff
@@ -48,6 +49,11 @@ declare interface Jupyter {
 
 declare var Jupyter: Jupyter | undefined
 
+export const kernels: {[key: string]: any} = {}
```
mattpap: Declare an interface for `Kernel` type, the same as it was done for `Comm` and `Jupyter`.

philippjfr:
From my perspective this is ready to merge once @canavandl is happy with and has merged https://github.com/bokeh/jupyterlab_bokeh/pull/21. It's not clear to me how I'd go about adding tests for this code, but it seems there are currently no unit tests for ``embed.ts`` anyway.

canavandl:
This works for me locally and seems reasonable-ish. It's definitely a JupyterLab anti-pattern to attach the kernel to window like were are doing, but I'm not sure how else to add this support without breaking classic notebook support or sending two copies of the document over the wire.

I've merged the change into jupyterlab_bokeh and will make a point release when this has merged.

jlstevens:
> I've merged the change into jupyterlab_bokeh and will make a point release when this has merged.

That would be great! I would love to test this ASAP.

bokehjs/src/coffee/embed.ts
```diff
@@ -88,8 +105,25 @@ function _init_comms(target: string, doc: Document): void {
...
+      kernel.registerCommTarget(target, (comm: Comm) => {
+        logger.info(`Registering JupyterLab comms for target ${target}`)
+        const r = new Receiver()
+        comm.onMsg = _handle_notebook_comms.bind(doc, r)
+      })
+    } catch (e) {
+      logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
+    }
+  } else {
     console.warn('Jupyter notebooks comms not available. push_notebook() will not function');
```
bryevdv: What would lead to this branch? (Can the error message be improved)

bokehjs/src/coffee/embed.ts
```diff
@@ -23,31 +23,42 @@ export interface RenderItem {
...
   }
 }
 
+
 declare var Jupyter: Jupyter | undefined
 
+export const kernels: {[key: string]: Kernel} = {}
+
+// Exported to allow external libraries to define custom message handlers
+export {Receiver}
```
bryevdv: Is this necessary? It's already exported from `protocol` ?

bokehjs/src/coffee/embed.ts
philippjfr: This branch would be hit if you're exporting the HTML, or if you haven't installed the jupyterlab extension. No way of knowing which really.

bokehjs/src/coffee/embed.ts
```diff
@@ -66,15 +77,21 @@ function _handle_notebook_comms(this: Document, receiver: Receiver, comm_msg: Co
...
   if (target == comm.target_name) {
     const r = new Receiver()
-    comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    if (comm.on_msg) {
+      // Classic Notebook
+      comm.on_msg(_handle_notebook_comms.bind(doc, r))
+    } else {
+      // JupyterLab
+      comm.onMsg = _handle_notebook_comms.bind(doc, r)
+    }
```
bryevdv: Amazing, this spelling difference is a back door way to distinguish classic notebook from jlab?

bokehjs/src/coffee/embed.ts
bryevdv: OK, I'd suggest mentioning those possibilities either in the message or in some docs somewhere.

bokehjs/src/coffee/embed.ts
philippjfr: Afaik, ``protocol`` is not exported however. It's really not needed for Bokeh itself but it would be really great if we could make it available to other libraries (e.g. HoloViews) somehow. Don't mind at all where, as long as I can access it via the ``Bokeh`` object somehow. I'm planning to write a small example using flask and socketIO to demonstrate how to use it independently of the notebook and bokeh server contexts.

bryevdv:
> It's definitely a JupyterLab anti-pattern to attach the kernel to window like were are doing, but I'm not sure how else to add this support without breaking classic notebook support or sending two copies of the document over the wire.

Well we will do whatever we need, but out of curiousity where is this discussed or documented as an anti-pattern? Would it be profitable to engage with the jlab team to see if there is any better approach?

bokehjs/src/coffee/embed.ts
philippjfr: Yes, there's a bunch of API differences you could use but this seemed simplest.

bokehjs/src/coffee/embed.ts
bryevdv: FWIW I'd rather export `protocols` so that people can grab it from there, than rely on a transitive import like this

bokehjs/src/coffee/embed.ts
philippjfr: Will do.

bokehjs/src/coffee/embed.ts
philippjfr: Sounds good, I do think these are useful.

bokehjs/src/coffee/embed.ts
```diff
@@ -88,8 +102,27 @@ function _init_comms(target: string, doc: Document): void {
...
+        const r = new Receiver()
+        comm.onMsg = _handle_notebook_comms.bind(doc, r)
+      })
+    } catch (e) {
+      logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
+    }
+  } else {
+    console.warn('Jupyter notebooks comms not available. push_notebook() will not function.'
+                 'If running JupyterLab ensure the latest jupyterlab_bokeh extension is installed.'
+				 'In an exported notebook this warning is expected.');
```
mattpap: This isn't valid syntax. Also, indentation is off.

philippjfr:
>Would it be profitable to engage with the jlab team to see if there is any better approach?

The way I understand it is that we could do it all properly and move all the rendering logic into the JupyterLab extension and hook up the comms there, but that would basically require reworking the entire machinery around embedding plots and also sacrifice compatibility for classic notebooks.

In the long term I can see that as a viable strategy but it's very early days for JupyterLab and plenty of people are still using classic notebooks. For now this provides a pretty simple path forward and I don't see many issues with the approach. JupyterLab, by its nature lets you execute arbitrary code, the JupyterLab MIME types already allow arbitrary JS execution, so exposing the kernel does not seem to me to open up any meaningful additional security holes. I'll let @canavandl have the final say here though.

bokehjs/src/coffee/embed.ts
philippjfr: Thanks, I did wonder, but how come this compiles and runs okay?

mattpap: It doesn't compile:
```
[18:21:20] src/coffee/embed.ts(123,18): error TS1005: ',' expected.
[18:21:20] src/coffee/embed.ts(124,6): error TS1005: ',' expected.
```
Our build currently doesn't fail if there are compilation errors. This will change with PR #7560.

mattpap: It also doesn't run correctly, at least according to chrome, and fails with:
```
SyntaxError: missing ) after argument list
```

philippjfr: Great, good to know, thanks! Easy to overlook errors locally, so I'm looking forward to having it error.

mattpap: You can use string template literals (backticks instead of quotes) for multiline strings, but that doesn't work nicely with indented text, so usually is pretty useless without additional processing.

philippjfr: >SyntaxError: missing ) after argument list

Suspect it must have cached an old version when I tried it.

canavandl:
> Well we will do whatever we need, but out of curiosity where is this discussed or documented as an anti-pattern? 
I would have to hunt around for design docs, but intentional removal of the kernel from ``window.Jupyter`` was to prevent people from running arbitrary code on public notebook servers. I think that's a pretty uncontroversial summary of the reasoning.

> Would it be profitable to engage with the jlab team to see if there is any better approach?
Our problem is having access to the client-side Document object from our renderer extension, so that we can attach the comm.onMsg callback to it. Instead of receiving the doc JSON and building the document inside of the extension, we're receiving a script tag that contains this logic and directly embedding it in the output.

I attempted to implement things by passing the document id as part of the mime message and then selecting the doc out of Bokeh.index and adding the callback, but I didn't quite get that working. It's presumably possible though? If somebody wants to take a crack at it, please be my guest.

> JupyterLab, by its nature lets you execute arbitrary code, the JupyterLab MIME types already allow arbitrary JS execution, so exposing the kernel does not seem to me to open up any meaningful additional security holes. I'll let @canavandl have the final say here though.

The problem isn't arbitrary JS but submitting arbitrary python code to be executed on the server. What we're doing is enabling somebody to run:
```
window.Bokeh.kernels[id].requestExecute("import os; os.rmdir('~/.')")
```
This is the problem with public classic notebooks.

philippjfr:
>The problem isn't arbitrary JS but submitting arbitrary python code to be executed on the server. What we're doing is enabling somebody to run:

True, I had assumed you could hack that anyway by editing the cell from JS and then sending keyboard events to execute it, but it seems that is handled. If we're really worried about this we could add a ``KernelProxy`` object to the extension which just exposes the comm methods without exposing ``requestExecute``.

canavandl:
> True, I had assumed you could hack that anyway by editing the cell from JS and then faking virtual keyboard events to execute it, but it seems that is handled. If we're really worried about this we could add a KernelProxy object to the extension which just exposes the comm methods without exposing requestExecute.

ooooh I like that idea. I think we should do that.

philippjfr:
>ooooh I like that idea. I think we should do that.

Great, I agree. This seems to be sufficient:

```javascript
const kernel_proxy: any = {_comms: kernel._comms, registerCommTarget: kernel.registerCommTarget, _targetRegistry: kernel._targetRegistry};
```

mattpap:
> Great, I agree. This seems to be sufficient:

Except the dreadful `any`.

philippjfr:
>Except the dreadful any.

I'll get to the point where I type things properly from the start soon, I promise 😄

If we agree that's a reasonable approach, I'll make a PR with that change, properly typed, on the extension repo.

bokehjs/src/coffee/embed.ts
```diff
@@ -88,8 +102,25 @@ function _init_comms(target: string, doc: Document): void {
     } catch (e) {
       logger.warn(`Jupyter comms failed to register. push_notebook() will not function. (exception reported: ${e})`)
     }
-  } else
-    console.warn('Jupyter notebooks comms not available. push_notebook() will not function');
+  } else if (doc.roots()[0].id in kernels) {
+    logger.info(`Registering JupyterLab comms for target ${target}`)
+    const kernel = kernels[doc.roots()[0].id]
+    for (const id in kernel._comms) {
```
canavandl: What does this loop do? This relies on a private attribute on ``kernel`` for in JupyterLab, so I'm hesitant to include it.

philippjfr: I think it updates the onMsg handler on comms that have already been set up. It's not quite clear to me in what scenario the comm would already have been registered though.

canavandl: I see the same logic for classic notebooks (which presumably has a reason? I don't know how to git blame the previous embed.coffee), but I can't think of how it'd happen in a JLab context. I'd proposed deleting it.

philippjfr: Looks like it was me 2 years in https://github.com/bokeh/bokeh/pull/5014 😞 

I'll have a go at removing it in classic notebooks and jupyterlab to see what happens.

philippjfr: Here's what I said about it, I don't think any of this is true now, but I'd have to double check:

>Because now the template registers a comms target with Jupyter immediately, so it's not dependent on BokehJS being loaded. When bokeh does load it will first look if a comm has already been initialized for the specified target and attach itself with a callback using Comm.on_msg. Hope that makes sense. It feels a little bit dirty but ensures the comm target is always registered in time for the Comm to be created and initialized in Python.

philippjfr: Okay I'm now fairly confident this is not needed anymore.

canavandl: Great - thanks for finding that out.

canavandl:
I'm very pleased with the final form of this. It seems minimally hacky and we avoid exposing the Jupyter kernel unnecessarily. I'm +1 on merging

bokehjs/src/coffee/embed.ts
mattpap: > git blame the previous embed.coffee

Unfortunately due to how git handles renames and the fact that we use squash merges, this is far from being easy. You will have to find a commit or a tag  (`<rev>`) where rename happened and then use `git blame <rev>~1 -- file_name`.

PR #7570, opened 2018-02-25
Alpha for images

marcuskd:


bryevdv:
@marcuskd thanks for the new PR! This looks more like the approach I imagined. We are currently having some issues on TravisCI with PRs from forks. I will have to try things out locally. It might be good to add a but more to the help string, e.g. to state that this alpha is applied on top of any alpha from an explicit color mapper (i.e. they will combine, in similar way to the chart here: https://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#specifying-colors)

marcuskd:
Hi @bryevdv. Help string updated as suggested, and I've also updated the example comment to make the same point. Being British my biggest issue was remembering to spell colour without a u! :-)

bryevdv:
@marcuskd fantastic, I think our CI issue may be resolved soon so I will hold off a bit and restart the tests, but hope to merge this today

marcuskd:
Great news, thanks Bryan.

bryevdv:
Tests now green, output image for change documentation:

screen shot 2018-02-27 at 13 09 40

Thanks for the PR @marcuskd !

bryevdv:
@marcuskd if you'd still like to make the convenience `alpha` parameter to glyph methods work (i.e. by setting `global_alpha` appropriately when the glyph is `Image` please open a new issue. Additionally, I think these same changes would be useful for `ImageRGBA`, if you have the opportunity  to make a similar PR for `ImageURL` we'd be happy to accept it!

marcuskd:
Thanks for the merge @bryevdv! I'll have a look at the convenience alpha parameter and ImageRGBA class, I hope to have an opportunity to address those.

PR #7575, opened 2018-02-27
Add node and npm version requirement specification

### LINKED ISSUE #7571: Enforce usage of npm >= 5.6
Because of #7514 , I had to run ``npm run prepare`` in order to generate a .package.json file (containing a hash of package.json) in order for gulp build to run.

I would suggest two things:
* Better message for when .package.json doesn't exist - currently it says ``package.json has changed. Run `npm install`.`` but suggesting running ``npm run prepare`` in this case would be clearer.
* Document in the dev guide the need to run the ``npm run prepare`` command

bokeh=0.12.14dev4

mattpap:
fixes #7571

### DIFF
--- a/bokehjs/gulpfile.js
+++ b/bokehjs/gulpfile.js
@@ -1,5 +1,22 @@
 const crypto = require("crypto")
+const semver = require("semver")
+const cp = require("child_process")
 const fs = require("fs")
+
+const {engines} = require("./package.json")
+
+const node_version = process.version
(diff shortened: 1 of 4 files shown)


mattpap:
So far tests will fail, because an older version of npm is shipped with node. Somewhere, I presume in a few places given the spread of our build, there needs to be the following code run:
```sh
npm install -g npm@'>=5.6 <6.0'
```
(though `npm install -g npm` should be fine most of the time).

PR #7578, opened 2018-02-27
Fix type errors in core

mattpap:
This is a bit chaotic PR, because it started as an attempt to fix all TS2339 errors, but then I managed proceed with core/ui_events rewrite (which was split off as PR #7555), and continued with fixing type errors in last bits of the core.

addresses #6481


mattpap:
I'm almost done with `core/util/serialization`, which will make rewrite of `core` complete.

bokehjs/src/coffee/core/util/serialization.ts
```diff
@@ -1,215 +1,260 @@
...
+
+  let dtype: DType
+  if (name in DTYPES)
+    dtype = DTYPES[name as ArrayName]
+  else
+    throw new Error(`unknown array type: ${name}`)
 
-export const encode_base64 = function(array, shape) {
-  const b64 = arrayBufferToBase64(array.buffer);
-  const dtype = DTYPES[array.constructor.name];
```
mattpap: How this didn't fail before in phantomjs? `TypedArray.name` is not supported in phantomjs. Currently this fails with:
```
[JS] Error: unknown array type: undefined
[JS] [bokeh] Failed to repull session Error: unknown array type: undefined
```
in all server examples.

bryevdv: I have no idea offhand, @canavandl any thoughts?

mattpap:
@bryevdv, unless there are any objections, I would like to get this merged today.

bryevdv:
I had a look over the non-trivial changes and nothign stood out. Let's merge now.

bokehjs/src/coffee/core/util/serialization.ts
canavandl: I have no idea. Sorry

PR #7579, opened 2018-02-28
Bryanv/7518 ajax data source endpoint undefined

### LINKED ISSUE #3723: Write unittest for AjaxDataSource
Need to write unittests for AjaxDataSource which was added in PR #3682.

### LINKED ISSUE #7518: Is Ajaxdatasource losing the endpoint?
I reported this issue in Stackoverflow, because there the code is easier to read: https://stackoverflow.com/questions/48747524/bokeh-0-12-14-flask-broken-library-or-new-syntax-for-ajaxdatasource

I am reading some data using AjaxDataSource.
With Bokeh 0.12.13 it works, while with Bokeh 0.12.14 it doesn't: it looks like the endpoint is lost after the first data refresh. 

The JS error is "`bokeh-0.12.14.min.js:1 undefined http://0.0.0.0:5011/undefined 404 (NOT FOUND)`". 
The endpoint (that works when called from the browser directly) should be instead `http://0.0.0.0:5011/data`.

If it is not a problem for you, I would avoid reposting the same code here.

bryevdv:
- [x] issues: fixes #7518 fixes #3723
- [x] tests added / passed

This PR fixes the regression introduced in the TypeScript conversion, and also factors `AjaxDataSource` into more testable chunks

### DIFF
--- a/bokehjs/test/models/sources/ajax_data_source.coffee
+++ b/bokehjs/test/models/sources/ajax_data_source.coffee
@@ -1,0 +1,87 @@
+{expect} = require "chai"
+utils = require "../../utils"
+
+sinon = require 'sinon'
+
+{AjaxDataSource} = utils.require("models/sources/ajax_data_source")
+
+describe "ajax_data_source module", ->
+
+  describe "AjaxDataSource", ->
+
+    beforeEach ->
+      global.XMLHttpRequest = sinon.useFakeXMLHttpRequest();
+      requests = this.requests = [];
+
+      global.XMLHttpRequest.onCreate = (xhr) ->
+        requests.push(xhr)
+
+    afterEach ->
+      global.XMLHttpRequest.restore()
+
+    describe "do_load method", ->
+
+      it "should replace", ->
+        s = new AjaxDataSource({data_url: "http://foo.com"})
+        expect(s.data).to.be.deep.equal {}
+
+        xhr = s.prepare_request()
+        xhr.respond(200, {}, '{"foo": [10, 20], "bar": [1, 2]}')
+        s.do_load(xhr, "replace", 10)
+        expect(s.data).to.be.deep.equal {"foo": [10, 20], "bar": [1, 2]}
+
+        xhr = s.prepare_request()
+        xhr.respond(200, {}, '{"foo": [100, 200], "bar": [1.1, 2.2]}')
+        s.do_load(xhr, "replace", 10)
+        expect(s.data).to.be.deep.equal {"foo": [100, 200], "bar": [1.1, 2.2]}
+
+        # max size ignored when replacing
+        xhr = s.prepare_request()
+        xhr.respond(200, {}, '{"foo": [1000, 2000], "bar": [10.1, 20.2]}')
+        s.do_load(xhr, "replace", 1)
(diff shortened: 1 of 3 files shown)


bryevdv:
Here is the updated test code, to work, `bokeh static` must also be running

```
# -*- coding: utf-8 -*-

from flask import Flask, jsonify, render_template_string

from bokeh.embed import components
from bokeh.models.sources import AjaxDataSource

import pandas as pd
import numpy as np

#########################
BOKEH_VERSION = "0.12.13"   # <--- DON'T FORGET TO CHANGE HERE IF USING ANOTHER VERSION
#########################

html_template="""
    <!DOCTYPE html>
    <html>
        <head>
            <meta charset="utf-8">
            <title>Example</title>
            <link href="http://localhost:5006/static/css/bokeh.css" rel="stylesheet" type="text/css">
            <script src="http://localhost:5006/static/js/bokeh.js"></script>
        </head>
    """.format(bokeh_ver=BOKEH_VERSION)

html_template += """
     <body>
            <h1>Working in Bokeh 0.12.13<br>and not working in 0.12.14</h1>
            <div id="example_plot"></div>

            {{ plot_div.plot_grid | safe}}
            {{ plot_script | safe}}

        </body>
    </html>
    """

# ------------------------------------------------------------------------------

app = Flask(__name__)

# ------------------------------------------------------------------------------

class ExamplePlot:

    def __init__(self, app):

        self._data_source = None
        self.data_source_update()   # create the first set of dummy data for self.data_source

        self._ajax_data_source = AjaxDataSource(
            data_url="/data",
            data=self._data_source,
            polling_interval = 1000,
        )

        self.figure = figure()

        self.figure.line(
            source=self._ajax_data_source,
            x="index",
            y="value",
        )

    def data_source_update(self):
        self._data_source = {
            "index": list(range(180)),
            "value": list(np.random.randint(1,100,size=180))
        }

# ------------------------------------------------------------------------------

example_plot = ExamplePlot(app)

@app.route("/")
def main_page():

    script, div = components({'plot_grid': example_plot.figure})

    html = render_template_string(
        html_template,
        plot_script = script,
        plot_div = div,
    )
    return html


@app.route("/data", methods=['GET', 'OPTIONS', 'POST'])
def serve():
    example_plot.data_source_update()

    return jsonify(
        index=example_plot._data_source["index"],
        value=[int(x) for x in example_plot._data_source["value"]],
    )


if __name__ == '__main__':
    app.run(host="0.0.0.0", port=5011)
```

bokehjs/test/models/sources/ajax_data_source.coffee
```diff
@@ -0,0 +1,87 @@
+{expect} = require "chai"
```
mattpap: This is a new file, so it must be written in TypeScript.

bryevdv: I was thinking about that. Putting aside that this was never a decision that was actually discussed or taken for all files, what is the specific value of *tests* being in TS? The biggest value of tests is that they be as quick and easy and painless to write as possible, so that more people are encouraged to write more of them. This class for instance has been untested, and has suffered as a result, for years. There is more value in there being tests at all, than not having them, regardless of language.

bryevdv: recording decision to merge this now and update directories of files at time rather than mix sources together.

PR #7581, opened 2018-02-28
FuncTickFormatter: expose `index` and `ticks`

kernc:
Alternative to https://github.com/bokeh/bokeh/pull/7564.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/7563
- [x] tests added / passed


bryevdv:
Agreed this is much preferable, I did not realize this was a common approach. Thanks for the suggestion @mattpap

PR #7583, opened 2018-02-28
Switched from RBush to FlatBush

### LINKED ISSUE #7582: Consider replacing RBush spatial index with flatbush
Recently the creator of RBush has released a new library for static spatial indexes called [flatbush](https://github.com/mourner/flatbush), which promises to be a lot faster than RBush both for index creation and for actual querying. Since the promised improvements are quite large and the API looks quite similar it would be worth investigating whether we could upgrade the spatial indexes in bokehJS to use flatbush.

philippjfr:
As suggested in https://github.com/bokeh/bokeh/issues/7582, this PR attempts to switch out RBush for flatbush to perform the spatial indexing. As flatbush is very new is no ``@types/flatbush`` package yet, so I haven't typed everything properly yet. Before proceeding with this I'd like to find a principled way to perform some benchmarks. So far I've only seen noticeable improvements for very large number of points. 

- [x] issues: fixes #7582
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/core/util/spatial.ts
+++ b/bokehjs/src/coffee/core/util/spatial.ts
@@ -1,4 +1,5 @@
-import * as rbush from "rbush"
+/// <reference types="@types/rbush" />
+import * as flatbush from "flatbush"
 
 export type Rect = {minX: number, minY: number, maxX: number, maxY: number}
 export type IndexedRect = Rect & {i: number}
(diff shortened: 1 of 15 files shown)


bryevdv:
> So far I've only seen noticeable improvements for very large number of points.

Other factors may dominate for smaller data sizes. But if this improves things for only larger sets and nothing else that's still a win. There are some `trace` log statements that report timing for various render steps currently, you could add some specifically to report index creation time. Lookups might require adding logs to more places.

bokehjs/src/coffee/core/util/spatial.ts
```diff
@@ -1,37 +1,45 @@
...
 
   indices(rect: Rect): number[] {
-    const points = this.search(rect)
-    const n = points.length
-    const indices = new Array<number>(n)
-    for (let j = 0; j < n; j++) {
-      indices[j] = points[j].i
-    }
-    return indices
+    return this.index.search(rect.minX, rect.minY, rect.maxX, rect.maxY);
```
mattpap: This doesn't seem right. `this.index.search()` will give you indices into `this.index.data`, but those don't have to match `i`'s 1-1.

philippjfr: That's true I suppose, in practice all the ``i``s are straightforward ascending indexes as far as I can tell, but I suppose this should be general.

philippjfr:
Just for future reference and before I lose it, here's the result of my profiling of the spatial indexing using 1000-500k points: 

rbush

mourner:
Note that there was a bug in Flatbush that made search slower than it should be — fixed now in v1.3.1.

bryevdv:
Thanks for the info @mourner 

@philippjfr whenever you get a chance can you re-run your benchmark with 1.3.1?

mourner:
@bryevdv FYI since then I released a few new versions that push the search performance further. The latest one is v2.0.4. It has a breaking change though (`flatbush(...)` -> `new Flatbush(...)`).

mattpap:
There are two image differences where z-order is affected:
- https://s3.amazonaws.com/bokeh-travis/358399990/examples.html#plotting/file/custom_tooltip
- https://s3.amazonaws.com/bokeh-travis/358399990/examples.html#plotting/file/graphs

bryevdv:

FYI I am fine with those two differences, any obejections to merging this? @philippjfr can you re-run the benchmark?

PR #7584, opened 2018-02-28
don't repeat base install for external PRs

bryevdv:
The previous CI scripts would unconditionally re-install miniconda on top of an previous install, when a PR from an outside fork was made. With the latests miniconda installers this causes conda errors. This PR changes the CI scripts to only perform the base install a single time in all cases.


bryevdv:
Builds are passing with `MINICONDA_FILENAME` changed back to the latest installer. Will merge once tests are all green.

PR #7598, opened 2018-03-05
Remove opts argument from models' constructors

mattpap:
Previously we had `new Model(attrs, opts)`, now there is `new Model(attrs)`. `opts` was only used for deferred initialization. The best solution would be to have multiple constructors, but that's not supported by TypeScript. This approach seems a bit hackinsh, but at least we don't have to define `Opts` interface for all models. Also, having models configurable in general wasn't the best idea from the API design standpoint.

addresses #6481

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/xy_glyph.ts
+++ b/bokehjs/src/coffee/models/glyphs/xy_glyph.ts
@@ -27,16 +27,14 @@ export namespace XYGlyph {
     x: NumberSpec
     y: NumberSpec
   }
-
-  export interface Opts extends Glyph.Opts {}
 }
 
 export interface XYGlyph extends XYGlyph.Attrs {}
 
 export abstract class XYGlyph extends Glyph {
 
-  constructor(attrs?: Partial<XYGlyph.Attrs>, opts?: XYGlyph.Opts) {
-    super(attrs, opts)
+  constructor(attrs?: Partial<XYGlyph.Attrs>) {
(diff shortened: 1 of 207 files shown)


bokehjs/src/coffee/models/glyphs/xy_glyph.ts
```diff
@@ -27,16 +27,14 @@ export namespace XYGlyph {
...
 
 export interface XYGlyph extends XYGlyph.Attrs {}
 
 export abstract class XYGlyph extends Glyph {
 
-  constructor(attrs?: Partial<XYGlyph.Attrs>, opts?: XYGlyph.Opts) {
-    super(attrs, opts)
+  constructor(attrs?: Partial<XYGlyph.Attrs>) {
+    super(attrs)
   }
```
bryevdv: If `constructor` only calls `super` and nothing else, is it needed at all?

bryevdv:
Definitely a big hack but it's livable/an improvement

bryevdv:
Merging now to get out of the way

bokehjs/src/coffee/models/glyphs/xy_glyph.ts
mattpap: It's needed, because it asserts its type. Otherwise you will be using the type of the parent constructor, and in general `HasProps`', which allows any attributes. It TypeScript had a stronger type system (supported class type members), then this may not be needed.

PR #7599, opened 2018-03-05
Fix type errors in models/sources

mattpap:
This is another painful rewrite (after ui_events and serialization), because `data_source.data` can mix arrays with typed arrays, but most code doesn't bother to account for that, and when it does, it does this in atrocious ways.

addresses #6481


bryevdv:
OK will merge this and get a dev build out

PR #7601, opened 2018-03-06
7600 image rgba alpha

### LINKED ISSUE #7600: Global alpha for image_rgba plot
It's currently not possible to set a global alpha for image_rgba plots. 

Example code:

```
import numpy as np

N = 20
img = np.empty((N, N), dtype=np.uint32)
for i in range(N):
    for j in range(N):
        view[i, j, 0] = int(i/N*255)
        view[i, j, 1] = 158
        view[i, j, 2] = int(j/N*255)
        view[i, j, 3] = 255

# Solid line to show effect of alpha...
p.line([0, 10], [10, 0], color='red', line_width=2)
# Attempt to set global alpha of 0.5...
p.image_rgba(image=[img], x=0, y=0, dw=10, dh=10, alpha=0.5)

show(p)
```

The red line from top left to bottom right is not visible behind the image.

Software versions...
Bokeh 0.12.14
Python 3.6.4
Firefox 58.0.2 (64-bit)
Ubuntu MATE 16.04.4 LTS

NB: I'm working on a solution following conclusion of PR#7570.

marcuskd:
- [x] issues: fixes #7600 
- [x] tests added / passed

Summary:
1. global_alpha kwarg added to ImageRGBA
2. convenience alpha kwarg now sets global_alpha for all three image types (Image, ImageRGBA, ImageURL)
3. example file image_alpha.py updated to include examples for all three image types


bokehjs/package-lock.json
```diff
@@ -1136,7 +1136,6 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
-        "fsevents": "1.1.3",
```
mattpap: @marcuskd, please revert changes to this file. You will also need to update npm to version >= 5.6 to avoid such issues (anything below results in undeterministic module resolution). This will be enforced on build-level after PR #7575.

marcuskd: Hi @mattpap. Changes reverted and pushed a moment ago. I didn't notice I'd done this, will update npm as suggested.

bokeh/models/glyphs.py
```diff
@@ -402,7 +406,15 @@ class ImageRGBA(XYGlyph):
...
-    _args = ('image', 'x', 'y', 'dw', 'dh', 'dilate')
+    _args = ('image', 'x', 'y', 'dw', 'dh', 'global_alpha', 'dilate')
+
+    # a hook to specify any additional kwargs handled by an initializer
+    _extra_kws = {
+        'alpha': (
+            'float',
+            'a convenience parameter to set the global_alpha'
+        )
+    }
```
bryevdv: I'm not sure the `_extra_kws` additions are needed. The convenience `alpha` parameter is (or should be) documented as a universally available kw arg to all glyph methods (just up until now it was a no-op on images). IIRC `_extra_kws` is for things like palette that are convenience args specific to one glyph

bokeh/plotting/helpers.py
```diff
@@ -724,6 +724,10 @@ def func(self, **kwargs):
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
 
+        # Assign global_alpha from alpha if glyph type is an image
+        if 'alpha' in kwargs and 'Image' in glyphclass.__view_model__:
+            kwargs['global_alpha'] = kwargs['alpha']
+
```
bryevdv: I think I'd be happier with this with explicit instance checks. Alternatively a generic way to do this would be to delegate to the glyphclasses to handle directly somehow but that does not need to happen right now. 
```

**Comment by bryevdv** [comment]
@marcuskd Thanks for the PR! I only had a couple of small comments

File: bokeh/models/glyphs.py (hunk 2 above)
Comment (marcuskd) [reply bokeh/models/glyphs.py]: No problem, will remove the _extra_kws. I wasn't sure about them but thought I'd add them for completion.

File: bokeh/plotting/helpers.py (hunk 3 above)
Comment (marcuskd) [reply bokeh/plotting/helpers.py]: OK, I'll change to explicit instance checks. The __view_model__ approach seemed neat but I guess it's not foolproof against 'Image' being in an unsuitable view model in the future.

**Comment by marcuskd** [comment]
Hi @bryevdv. All done!

File: bokehjs/src/coffee/models/glyphs/image_rgba.ts [hunk 4]
```diff
@@ -8,6 +8,11 @@ import {Context2d} from "core/util/canvas"
 export class ImageRGBAView extends XYGlyphView {
   model: ImageRGBA
 
+  initialize(options: any): void {
+    super.initialize(options);
+    this.connect(this.model.properties.global_alpha.change, () => this.renderer.request_render());
+  }
+
```
bryevdv: I am not sure this is necessary, I would expect it to be covered by this:

https://github.com/marcuskd/bokeh-fork/blob/b6c21f35195ec6c7a9c826400a9598034eb77a37/bokehjs/src/coffee/models/renderers/glyph_renderer.ts#L118

@marcuskd do you know for certain that changing alpha does not result in an update without this line? If not can you test with it removed?

marcuskd: I added this to Image (in PR#7570) and ImageRGBA as it was already in ImageURL so I assumed it was necessary. I'll remove it and test.

bryevdv: Ok thanks! It may be needed, but it's worth double checking.

marcuskd: @bryevdv no, that line was not necessary. I'll remove it from all three image types and push the changes shortly...

bryevdv: Thanks!

bryevdv: @marcuskd Hrmm I just realized why those are there, the other line I linked does a full `set_data` on changes, but the line just requests a render. In theory that should be a performance optimization. It might be worth just keeping.

marcuskd: @bryevdv is the 'initialize' block (see below) needed? I'm not sure what it does but it was already in ImageURL, and my current PR will add it to ImageRGBA.

  initialize(options: any): void {
    super.initialize(options);
  }

bryevdv: @marcuskd yes apparently the initialize blocks are necessary for TypeScript reasons.

marcuskd: OK, will change back shortly...

bryevdv: great sorry for the back and forth and thanks for your patience!

bryevdv:
Thanks @marcuskd !

PR #7605, opened 2018-03-07
Fix WebDriverException when using export_png

### LINKED ISSUE #7168: WebDriverException when using export_png
Bokeh v0.12.10

When I try to use `export_png` in my application, I receive (pretty-printed):
```
{
  "errorMessage": "'undefined' is not an object (evaluating 'window.Bokeh.documents')",
  "request": {
    "headers": {
      "Accept": "application\/json",
      "Accept-Encoding": "identity",
      "Connection": "close",
      "Content-Length": "349",
      "Content-Type": "application\/json;charset=UTF-8",
      "Host": "127.0.0.1:32956",
      "User-Agent": "Python-urllib\/3.5"
    },
    "httpVersion": "1.1",
    "method": "POST",
    "post": "{\"sessionId\": \"ae4dd6f0-c218-11e7-8813-fba0ae2a2a1f\", \"script\": \"\\n\/\/ add private window prop to check that render is complete\\nwindow._bokeh_render_complete = false;\\nfunction done() {\\n  window._bokeh_render_complete = true;\\n}\\n\\nvar doc = window.Bokeh.documents[0];\\n\\nif (doc.is_idle)\\n  done();\\nelse\\n  doc.idle.connect(done);\\n\", \"args\": []}",
    "url": "\/execute",
    "urlParsed": {
      "anchor": "",
      "query": "",
      "file": "execute",
      "directory": "\/",
      "path": "\/execute",
      "relative": "\/execute",
      "port": "",
      "host": "",
      "password": "",
      "user": "",
      "userInfo": "",
      "authority": "",
      "protocol": "",
      "source": "\/execute",
      "queryKey": {
        
      },
      "chunks": [
        "execute"
      ]
    },
    "urlOriginal": "\/session\/ae4dd6f0-c218-11e7-8813-fba0ae2a2a1f\/execute"
  }
}
```

Apparently, `Bokeh.documents` can still be uninitialized when the page is loaded by Selenium.
Maybe `wait_until_render_complete` should use `WebDriverWait` to also check that `Bokeh.documets` exists?

p-himik:
Fixes #7168

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -13,6 +13,7 @@
 # Boilerplate
 #-----------------------------------------------------------------------------
 from __future__ import absolute_import, division, print_function, unicode_literals
+from six import raise_from
 
 import logging
 log = logging.getLogger(__name__)
@@ -240,6 +241,17 @@ def wait_until_render_complete(driver):
     '''
     from selenium.webdriver.support.ui import WebDriverWait
     from selenium.common.exceptions import TimeoutException
+
+    def is_bokeh_loaded(driver):
+        return driver.execute_script('''
+            const b = window.Bokeh;
+            return b && b.documents && b.documents.length > 0;
+        ''')
+
+    try:
+        WebDriverWait(driver, 5, poll_frequency=0.1).until(is_bokeh_loaded)
+    except TimeoutException as e:
+        raise_from(RuntimeError('Bokeh was not loaded in time. Something may have gone wrong.'), e)
 
     driver.execute_script(_WAIT_SCRIPT)
 


bryevdv:
@p-himik This looks good, as long as you are touching this function, what do you think about making the two hard-coded 5 second timeouts be configurable parameters to the function (with defaults of 5)?

bokeh/io/export.py
```diff
@@ -241,6 +241,14 @@ def wait_until_render_complete(driver):
     from selenium.webdriver.support.ui import WebDriverWait
     from selenium.common.exceptions import TimeoutException
 
+    def is_bokeh_loaded(driver):
+        return driver.execute_script('''
+            const b = window.Bokeh;
+            return b && b.documents && b.documents.length > 0;
+        ''')
+
+    WebDriverWait(driver, 5, poll_frequency=0.1).until(is_bokeh_loaded)
```
bryevdv: Should this be in a `try/except` block, that converts the timeout exception into a `RuntimeError` with a more Bokeh-specific message?

p-himik:
@bryevdv What for? It's a dev-level API, so no user is expected to touch it, and since it hasn't been done yet, no other function needs these parameters.
With that being said, I would either not add the parameters at all, or add them to all relevant public functions. But the latter may be the case for a different PR.

bryevdv:
@p-himik a general aversion to hard coded magic numbers, plus knowing that some users do more sophisticated things and do use `dev` APIs. The `dev` decorator is not a prohibition against using things, just an indicator that things are not needed for typical usage.

This is not a sticking point for me though, so I will merge now.

PR #7609, opened 2018-03-08
Datetime stream fix

### LINKED ISSUE #7587: Regression streaming datetime columns
The changes in https://github.com/bokeh/bokeh/pull/7379 introduced a regression streaming datetime columns expressed as a dictionary of arrays. This is because the PR introduced special handling for datetime types into the ColumnDataSource._stream method:

```python
        # slighgly awkward that we have to call converty_datetime_array here ourselves
        # but the downstream code expects things to already be ms-since-epoch
        for key in new_data:
            new_data[key] = convert_datetime_array(new_data[key])  
```

This causes issues when the data held by the ColumnDataSource has not also had the same transformation applied, triggering errors with type promotion when concatenating the two arrays:

```
  File "/Users/philippjfr/miniconda/envs/pyviz/lib/python3.6/site-packages/bokeh/core/property/containers.py", line 384, in _stream
    data = np.append(self[k], new_data[k])
  File "/Users/philippjfr/miniconda/envs/pyviz/lib/python3.6/site-packages/numpy/lib/function_base.py", line 5166, in append
    return concatenate((arr, values), axis=axis)
TypeError: invalid type promotion
```

Here is a simple reproducible example:

```python
from bokeh.io import curdoc

import pandas as pd
import numpy as np

def modify_doc(doc):
    """

    :param Document doc:
    :return:
    """
    df = pd.DataFrame(
        index=pd.date_range('now', periods=30, freq='T'),
        columns=['A', 'B', 'C'],
        data=np.cumsum(np.random.standard_normal((30,3)), axis=0)
    )

    cds = ColumnDataSource({'index': df.index, 'A': df.A})
    figure = Figure(x_axis_type="datetime")
    figure.line("index", "A", source=cds)

    doc.add_root(figure)

    def callback():
        new_df = pd.DataFrame(
            index=df.index + pd.to_timedelta('30m'),
            columns=df.columns,
            data=np.random.standard_normal((30,3))
        )
        cds.stream({'index': new_df.index, 'A': new_df.A})

    doc.add_periodic_callback(callback, 1 * 1000)

modify_doc(curdoc())

philippjfr:
Fixes two regressions in datetime streaming, the first applies to attempts to stream pandas Index or Series objects, which are currently not appropriately transformed. The second deals with a difference between server and notebook streaming. When streaming data on the server the data is transformed from datetime64 types to integers, which ends up being synced. Since JS->Python is not synced in the notebook this does not occur there. To avoid this problem we only transform the data during streaming if the column it is being streamed to has also been transformed.

The two added tests cover these two conditions, one where the data has not been transformed (emulating the notebook) and one where the data has been transformed (emulating bokeh server).

- [x] issues: fixes #7587
- [x] tests added / passed

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -423,9 +423,8 @@ class ColumnDataSource(ColumnarDataSource):
             else:
                 raise ValueError("Must stream updates to all existing columns (extra: %s)" % ", ".join(sorted(extra)))
 
+        import numpy as np
         if needs_length_check:
-            import numpy as np
-
             lengths = set()
             arr_types = (np.ndarray, pd.Series) if pd else np.ndarray
             for k, x in new_data.items():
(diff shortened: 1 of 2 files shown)


bryevdv:
OK I tested with the code from #7379 as well as both examples from #7587 and they are all working. I would say if there are any other combinations/cases that could be put under explicit test it would be good to do

bryevdv:
@philippjfr if you get a chance can you see if we need a similar fix for `patch`? (and link the issue her if so)

philippjfr:
Okay, I just checked and patch seems to work fine.

bryevdv:
Pleasant surprise!

PR #7610, opened 2018-03-08
Fix type errors in models/glyphs

mattpap:
addresses #6481

### DIFF
--- a/bokehjs/src/coffee/core/has_props.ts
+++ b/bokehjs/src/coffee/core/has_props.ts
@@ -19,6 +19,10 @@ export module HasProps {
     id: string
   }
 
+  export interface Props {
+    id: p.Any
+  }
+
   export interface SetOptions {
     check_eq?: boolean
     silent?: boolean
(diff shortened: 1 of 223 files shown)


bryevdv:
Is there a plan to unify e.g. this:
```
  export interface Attrs extends Annotation.Attrs, Mixins {
    x_start: NumberSpec
    y_start: NumberSpec
    start_units: SpatialUnits
    start: ArrowHead | null
    x_end: NumberSpec
    y_end: NumberSpec
    end_units: SpatialUnits
    end: ArrowHead | null
    source: DataSource
    x_range_name: string
    y_range_name: string
  }

  export interface Props extends Annotation.Props {
    x_start: p.NumberSpec
    y_start: p.NumberSpec
    start_units: p.Property<SpatialUnits>
    start: p.Property<ArrowHead | null>
    x_end: p.NumberSpec
    y_end: p.NumberSpec
    end_units: p.Property<SpatialUnits>
    end: p.Property<ArrowHead | null>
    source: p.Property<DataSource>
    x_range_name: p.Property<string>
    y_range_name: p.Property<string>
  }
```
 Or are we going to have to live with this duplication?

mattpap:
> Or are we going to have to live with this duplication?

No. In fact I only added props where needed to satisfy the type checker. Originally I was hoping for generating props from attrs, but it looks like I will have to approach this the other way round. This will happen in a later PR.

bryevdv:
OK, merging for a dev build

PR #7612, opened 2018-03-09
have one unit test job use tornado 4.5

### LINKED ISSUE #7483: Add Tornado 5 to CI testing
As a follow on to #7329 a new build should be adde that installs and tests with Torndao 5 once a conda build of it is available.

bryevdv:
- [x] issues: fixes #7483
- [x] tests added / passed

Rather than burden our already long build with a new job, this PR just makes one of the existing unit test jobs use Tornado 4.5.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,7 @@ jobs:
       env:
       - GROUP=unit
       - PYTHON=3.5
+      - TORNADO=4
 
     - install: scripts/ci/install:test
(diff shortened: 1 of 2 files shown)


bryevdv:
install from 2.7 (tornado 5) job:

screen shot 2018-03-08 at 17 06 45

install from 3.5 (tornado 4) job:

screen shot 2018-03-08 at 17 07 05

PR #7614, opened 2018-03-09
Fix io_loop paramenter in examples/howto/server_embed/flask_embed.py

volartem:
Added parameter io_loop in bk_worker() at examples/howto/server_embed/flask_embed.py for correct working with new version tornado.


bryevdv:
Thanks @volartem ! Noting for posterity that it is the example that has to be updated due to changes in Tornado, there is no way to make it work and keep the example as it was (but fortunately the new version also works with Tornado < 5)

PR #7617, opened 2018-03-11
Set a data source's inspected taking multiple renderers into account

### LINKED ISSUE #7397: hover_color only works for the last GlyphRenderer

Bokeh  0.12.13, Python 3.5, notebook 5.2.1, windows 10, Chrome

```
output_notebook()

source = ColumnDataSource(data=dict(x=[1, 2, 3], y0=[1, 2, 3], y1=[-1, -2, -3]))
fig = figure(tools="hover", plot_width=500, plot_height=300, title=None)
c1 = fig.circle('x', 'y0', size=10, hover_color="firebrick", source=source)
c2 = fig.circle('x', 'y1', size=10, hover_color="firebrick", source=source)
show(fig)
```

when mouse is over c1:

chrome_2018-01-11_12-53-43

when mouse is over c2:

chrome_2018-01-11_12-54-18

clairetang6:
This PR addresses the issue where a data source's `inspected` property gets overwritten by each renderer's inspection hit test. To fix the issue, the selection manager now calls `inspect` with a list of renderer views instead of a single renderer view, and `inspected` is set using the `UnionRenderers` policy. 

- [x] issues: fixes #7397
- [ ] tests added / passed


bokehjs/src/coffee/models/selections/interaction_policy.ts
```diff
@@ -25,6 +25,20 @@ export abstract class SelectionPolicy extends Model {
       return !source.selected.is_empty()
     }
   }
+
+  do_inspection(hit_test_result: HitTestResult, source: ColumnarDataSource): boolean {
+    if (hit_test_result === null) {
```
mattpap: Unless you distinguish `undefined` from `null` in this code, it's preferred to use `==` and `!=` operators.

bokehjs/src/coffee/core/selection_manager.ts
```diff
@@ -28,8 +28,8 @@ export class SelectionManager extends HasProps {
     this.inspectors = {}
   }
 
-  select(renderer_views: RendererView[], geometry: Geometry, final: boolean, append: boolean = false): boolean {
-    // divide renderers into glyph_renderers or graph_renderers
+  _split_renderer_views_by_type(renderer_views: RendererView[]): {[key: string]: RendererView[]} {
+    // TODO: return type, object values should be array of one type of renderer view
```
mattpap: You should also return an interface instead of a mapping, unless you can't (or won't be in future) predict what keys are returned.

bokehjs/src/coffee/core/selection_manager.ts
```diff
@@ -41,6 +41,11 @@ export class SelectionManager extends HasProps {
         }
       }
     }
+    return {'glyph_renderer_views': glyph_renderer_views, 'graph_renderer_views': graph_renderer_views}
```
mattpap: Drop quotes. This will signify that you work with interface members and not arbitrary keys.

bokehjs/src/coffee/models/tools/inspectors/hover_tool.ts
```diff
@@ -105,6 +105,25 @@ export class HoverToolView extends InspectToolView {
     return renderers
   }
 
+  protected _computed_renderers_by_data_source(): {[key: string]: DataRenderer[]} {
+    const renderers_by_source: {[key: string]: DataRenderer[]} = {}
+    for (const r of this.computed_renderers) {
+      let source_id: string
+      // XXX: needs typings for renderers
```
mattpap: `models/renderers` are typed as of last week.

mattpap:
@clairetang6, just rebase on top of `origin/master` to resolve file renaming conflicts.

mattpap:
Also, what's the status of this PR?

clairetang6:
Hi, I probably could have changed the status to paused at some point, but I can finish this up this weekend.

bokehjs/src/lib/core/selection_manager.ts
```diff
@@ -26,8 +25,7 @@ export class SelectionManager extends HasProps {
     this.inspectors = {}
   }
 
-  select(renderer_views: RendererView[], geometry: Geometry, final: boolean, append: boolean = false): boolean {
-    // divide renderers into glyph_renderers or graph_renderers
+  _split_renderer_views_by_type(renderer_views: RendererView[]): {glyph_renderer_views: GlyphRendererView[], graph_renderer_views: GraphRendererView[]} {
```
clairetang6: @mattpap left this comment on the last iteration: "You should also return an interface instead of a mapping, unless you can't (or won't be in future) predict what keys are returned." Was this what you meant by returning an interface?

bryevdv:
@clairetang6 I merged master to resolve the new conflict. I am pretty sure it was trivial enough that I couldn't mess it up, but please take a look when you get a chance (even after merge). I did test with the OP issue code and it works, so I will probably merge later today in any case.

bokehjs/src/lib/core/selection_manager.ts
bryevdv: I believe so, yes

bryevdv:
I wish m maybe I had suggested `HitTestPolicy` or something, so the selection and inspection policies could both accept a hit test policy (instead of both accepting selection policy which seems a bit confusing)

bryevdv:
Reverted this for now

PR #7628, opened 2018-03-14
Fix type errors in models/formatters

mattpap:
addresses #6481


bryevdv:
Looks good after a guick pass over, merging for dev build

PR #7635, opened 2018-03-15
Clean up all rendering levels in PlotCanvasView.build_levels()

### LINKED ISSUE #7476: PlotCanvasView#build_levels removes only glyphs when renderers change
Bokeh 0.12.14dev4

Brought to light by https://groups.google.com/a/continuum.io/forum/?utm_medium=email&utm_source=footer#!topic/bokeh/wsiTB4r38OM

If you add some non-glyph renderer (e.g. a BoxAnnotation) and then remove it by changing `plot.renderers` list, the renderer will still be displayed because of https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/plots/plot_canvas.ts#L563-L565

### LINKED ISSUE #7497: Disappearing Toolbar
In the example below (the plot appears after the button is pressed), the toolbar disappears when the plot (or toolbar) is tapped. The problem appeared in 0.12.11 onwards.

Software versions...
Bokeh 0.12.13
Python 3.5.2
Firefox 57.0.4 (64 bit)
```

from bokeh.models.layouts import Column, WidgetBox
from bokeh.models.widgets import Button, Div
from bokeh.io import curdoc
from bokeh.events import Tap

fig = Figure()
line_src = ColumnDataSource(data={'x': [0, 1], 'y': [0, 1]})
line = fig.line('x', 'y', source=line_src)


def update_data(event):
    line_src.data['x'][1] = event.x
    line_src.data['y'][1] = event.y
    fig.renderers.remove(line)
    fig.renderers.append(line)

fig.on_event(Tap, update_data)


def update_plot():
    gui.children[1] = fig

btn = Button(label='Click to plot')
btn.on_click(update_plot)
gui = Column(WidgetBox(btn), Div())

curdoc().add_root(gui)

```

### LINKED ISSUE #7590: Adding new renderers using push_notebook triggers errors
In earlier versions of bokeh (specifically 0.12.10) it was possible to add new bokeh renderers to a plot and then use ``push_notebook`` to update the plot. This still mostly works fine using recent versions (0.12.13+) but the JS console shows an error during rendering, which appears to be related to the ``ToolbarPanelView`` being deleted:

```javascript
    const toolbar = this._toolbar_views[this.model.toolbar.id];
    toolbar.render();

VM995:11109 Uncaught TypeError: Cannot read property 'render' of undefined  
```

This is because ``build_views`` appears to delete the view just before rendering:

```javascript
 for (const model_id of to_remove) {
  view_storage[model_id].remove()
  delete view_storage[model_id]
}
```

It is worth noting that this works fine on bokeh server so it likely has something to do with some state not being synced in the notebook. I realize that this may not be officially supported but since this used to work it would be very nice if it could be made to work again.

Here is a small example that demonstrates the issue:

```python
from bokeh.plotting import figure, output_notebook
from bokeh.io import  push_notebook

# output to static HTML file
output_notebook()

handle = show(p, notebook_handle=True)

p.circle([6, 7, 2, 4, 5], [1, 2, 3, 4, 5], size=20, color="red", alpha=0.5)
push_notebook(handle=handle)
```

mattpap:
~~The problem outlined in the issues is very general. Here I was able to get around it, but in the longer term, we will have to rethink document access in models and state management, especially related to models initialized in bokehjs.~~

fixes ~~#7497 (works around)~~
fixes #7476

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot_canvas.ts
+++ b/bokehjs/src/coffee/models/plots/plot_canvas.ts
@@ -614,8 +614,10 @@ export class PlotCanvasView extends DOMView {
     const new_renderer_views = build_views(this.renderer_views, renderer_models, this.view_options()) as RendererView[]
     const renderers_to_remove = difference(old_renderers, renderer_models.map((model) => model.id))
 
-    for (const id_ of renderers_to_remove) {
-      delete this.levels.glyph[id_]
+    for (const level in this.levels) {
+      for (const id of renderers_to_remove) {
+        delete this.levels[level][id]
+      }
     }
 
     for (const view of new_renderer_views) {


mattpap:
@philippjfr, can you try this out?

philippjfr:
Trying it now, will report back shortly.

mattpap:
Without a server, can Python objects be changed in a notebook by bokehjs?

philippjfr:
>Without a server, can Python objects be changed in a notebook by bokehjs?

No, in a notebook there is zero communication from JS -> Python.

mattpap:
So, this is PR is partly rubbish. So much for workarounds, this will have to be fixed properly.

mattpap:
Perhaps we could go forward with this despite it doesn't fix the push_notebook issue.

philippjfr:
I think that's reasonable, it's certainly not the main use case for bokeh users I imagine.

mattpap:
Well, tests started to fail.

bryevdv:

mattpap:
I repurposed this PR to fix #7476 only. I will get back to #7497 (and others) in another PR.

PR #7649, opened 2018-03-18
Bug fix: slider keyboard navigation

### LINKED ISSUE #7622: Slider violates start and end when using keyboard navigation
You can see this behavior on the official example:

[http://bokeh.pydata.org/en/latest/docs/gallery/slider.html](http://bokeh.pydata.org/en/latest/docs/gallery/slider.html)

Drag the amplitude slider with the mouse, and it varies from 0.1 to 10.  Use the keyboard instead and the slider appears to vary from 0 to 10.1 exceeding the start bound of 0.1 and the end bound 10.  This happens in Bokeh 0.12.14.

The fix appears to be:

```
--- bokeh-widgets-0.12.14.js
+++ bokeh-widgets-0.12.14.js
@@ -225,11 +225,11 @@ function _(require, module, exports) {
                     var value = Number(_this.sliderEl.noUiSlider.get());
                     switch (e.which) {
                     case 37: {
-                            value -= step;
+                            value = Math.max(value - step, start);
                             break;
                         }
                     case 39: {
-                            value += step;
+                            value = Math.min(value + step, end);
                             break;
                         }
                     default:
```

gbrunick:
Small bug fix.  This fix could potentially hid some deeper problem; I don't have a very complete understanding of the javascript code.

- [x] issues: fixes #7622
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/abstract_slider.ts
+++ b/bokehjs/src/coffee/models/widgets/abstract_slider.ts
@@ -93,11 +93,11 @@ export abstract class AbstractSliderView extends WidgetView {
         let value = Number(this.sliderEl.noUiSlider.get())
         switch (e.which) {
           case 37: {
-            value -= step
+            value = Math.max(value - step, start)
             break
           }
           case 39: {
-            value += step
+            value = Math.min(value + step, end)
             break
           }
           default:


gbrunick:
Unfortunately, the TravisCI build failed, but I'm don't know how to debug the issue.  To test this change on my machine, I used setup.py to rebuild the javascript and then confirmed that the slider in example in examples/app worked correctly.

mattpap:
@gbrunick, the failure seems spurious. I restarted the job.

bryevdv:
Thanks for the PR @gbrunick !

Issue #3886, opened 2016-02-17
Feature Request: Add hover inspection Image glyphs


ofirsa:
Hi, I hope this the right place to ask my question.
is it possible to implement the Hover Tool on image in such a way that when I moving my mouse on the image pixels, I get and information on the pixel below the mouse, like the pixel coordinates and pixel value.
I would be glad to get an example.
Thanks
Saar

birdsarah:
Hi @ofirsa, our issue tracker is becoming too large to manage easily. We have to reserve it for feature requests and bug reports only, to have any hope of taming it. This kind of general support question should go to the general public mailing list:

https://groups.google.com/a/continuum.io/forum/#!forum/bokeh

jbednar:
Sorry to revisit this old issue, but as far as I can tell it is still valid.  Sarah closed it as a support question, but to me it sounds like either a feature request (can images please support hover information like other glyphs do) or a bug report (images don't support hover information like they should), as requested.

In either case, what's the answer?  Is there some fundamental reason why Image shouldn't support hovering to reveal the pixel coordinates and the value of the image at that point?  We will need this in our work, and can start working on a PR if that's appropriate.

jbednar:
There's some relevant discussion on the Google Group, but nothing that indicates to me that this feature request is somehow invalid or not feasible:

https://groups.google.com/a/continuum.io/forum/#!searchin/bokeh/hover$20tool$20image|sort:relevance/bokeh/W-n34ElNlRo/O7_eut-5DwAJ

https://groups.google.com/a/continuum.io/forum/#!searchin/bokeh/Hover$20image/bokeh/fOlnivwLwvw/miZDTdA8CQAJ

bryevdv:
Not sure why this was closed, I thought there was another issue for image hit-testing but I can't find it offhand. Doing this will required setting a "2d" selection, i.e, will want to know which `i, j` pixel in the image was hit. That's not common, but there should already be kind-of similar example code for `MultiLine`. There's also the question of what to do when the image is big and a single pixel on the screen spans more than one pixel in the image (e.g. 1000x1000 image in a 500x500 canvas). 

Certainly possible in principle, but mostly just needs someone to spend some time thinking about cases like the above, and defining and documenting the exact behaviors. Will need tests too.

bryevdv:
To be clear: would be a welcome addition, just needs a little upfront thought and design.

jbednar:
[...]

PR #7651, opened 2018-03-19
Hover tool support for revealing values in image glyphs

jlstevens:
This PR aims to address issue #3886 by showing image values using the hover tool.

Although this PR is still WIP (e.g no typing yet), it is sufficiently complete to work with this example, demonstrating hover over an image containing integers and one containing floats:

image_hover_demo

## Notes:

* I have only considered getting hover information working on the JS side and have not considered how image indexing might be useful or represented syncing with Python.
* At this time, I have only added support for the special variables described [here](https://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#hovertool).
* The ``_lrtb`` method is used both for finding the integer indexing and for the hit testing with ``RBush`` and was inspired by what I saw in ``quad.ts``.
* Typed arrays seem to need a linear index which I compute in the ``_hit_point`` method.
* I use ``data_source._shapes`` to find the columns that correspond to image values in `` replace_placeholders``. The alternative seems to be to get this information from the glyph at the hover tool level, but this means passing this information through several levels to get to ``replace_placeholders``.

## Outstanding items

- [ ] Updated/add typing information.
- [ ] Update docs
- [ ] Tests (though I don't know what sort of testing is recommended here)

It is likely that we will need some discussion around the representation used for image indexing and this PR should help illustrate what is required, at least in the context of implementing the (often requested!) image hover functionality.

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/image.ts
+++ b/bokehjs/src/coffee/models/glyphs/image.ts
@@ -48,6 +48,14 @@ export class ImageView extends XYGlyphView {
     }
   }
 
+  _lrtb(i) {
+    const l = this._x[i]
+    const r = l + this._dw[i]
+    const b = this._y[i]
+    const t = b + this._dh[i]
+    return [l, r, t, b]
+  }
+
   protected _set_data(): void {
     if (this.image_data == null || this.image_data.length != this._image.length)
       this.image_data = new Array(this._image.length)
--- a/bokehjs/src/coffee/models/glyphs/image.ts
+++ b/bokehjs/src/coffee/models/glyphs/image.ts
@@ -8,6 +8,9 @@ import {max, concat} from "core/util/array"
 import {Context2d} from "core/util/canvas"
 import {Rect} from "core/util/spatial"
 import {RBush} from "core/util/spatial"
+import * as hittest from "core/hittest";
+import {Selection} from "../selections/selection";
+
 
 // XXX: because ImageData is a global
 export interface _ImageData extends XYGlyphData {
@@ -27,6 +30,7 @@ export interface _ImageData extends XYGlyphData {
 }
 
 export interface ImageView extends _ImageData {}
+
 export class ImageView extends XYGlyphView {
   model: Image
   visuals: Image.Visuals
@@ -67,6 +71,35 @@ export class ImageView extends XYGlyphView {
     const t = b + this._dh[i]
     return [l, r, t, b]
   }
+
+  _image_index(index, x,y) {
+    let [l,r,t,b] = this._lrtb(index);
+    let width = this._width[index]
+    let height = this._height[index]
+    let dx = (r - l) / width;
+    let dy = (t - b) / height;
+    let dim1 = Math.floor((x-l) / dx)
+    let dim2 = Math.floor((y-b) / dy)
+    return [index, dim1,  dim2, (dim2 * width) + dim1]
+  }
+
+  _hit_point(geometry) : Selection {
+    let {sx, sy} = geometry;
+    const x = this.renderer.xscale.invert(sx);
+    const y = this.renderer.yscale.invert(sy);
+    const bbox = hittest.validate_bbox_coords([x, x], [y, y]);
+    const candidates = this.index.indices(bbox);
+    const result = hittest.create_empty_hit_test_result()
+
+    result['im2d']['indices'] = []
+    for (let index of candidates) {
+      if ((sx != Infinity) && (sy != Infinity)) {
+        result['im2d']['indices'].push(this._image_index(index, x,y))
+      }
+    }
+    return result;
+  }
+
 
   protected _set_data(): void {
     if (this.image_data == null || this.image_data.length != this._image.length)
--- a/bokehjs/src/coffee/models/glyphs/image.ts
+++ b/bokehjs/src/coffee/models/glyphs/image.ts
@@ -7,6 +7,7 @@ import * as p from "core/properties"
 import {max, concat} from "core/util/array"
 import {Context2d} from "core/util/canvas"
 import {Rect} from "core/util/spatial"
+import {RBush} from "core/util/spatial"
 
 // XXX: because ImageData is a global
 export interface _ImageData extends XYGlyphData {
@@ -26,7 +27,6 @@ export interface _ImageData extends XYGlyphData {
 }
 
 export interface ImageView extends _ImageData {}
-
 export class ImageView extends XYGlyphView {
   model: Image
   visuals: Image.Visuals
@@ -46,6 +46,18 @@ export class ImageView extends XYGlyphView {
       this._set_data()
       this.renderer.plot_view.request_render()
     }
+  }
+
+  _index_data(): RBush {
+    const points = [];
+    for (let i = 0, end = this._x.length; i < end; i++) {
+      const [l, r, t, b] = this._lrtb(i);
+      if (isNaN(l+r+t+b) || !isFinite(l+r+t+b)) {
+        continue;
+      }
+      points.push({minX: l, minY: b, maxX: r, maxY: t, i});
+    }
+    return new RBush(points);
   }
 
   _lrtb(i) {
--- a/bokehjs/src/coffee/models/glyphs/image.ts
+++ b/bokehjs/src/coffee/models/glyphs/image.ts
@@ -64,7 +64,7 @@ export class ImageView extends XYGlyphView {
     return new RBush(points);
   }
 
-  _lrtb(i) {
+  _lrtb(i: any) {
     const l = this._x[i]
     const r = l + this._dw[i]
     const b = this._y[i]
@@ -72,7 +72,7 @@ export class ImageView extends XYGlyphView {
     return [l, r, t, b]
   }
 
-  _image_index(index, x,y) {
+  _image_index(index : any, x: any, y : any) {
     let [l,r,t,b] = this._lrtb(index);
     let width = this._width[index]
     let height = this._height[index]
@@ -83,7 +83,7 @@ export class ImageView extends XYGlyphView {
     return [index, dim1,  dim2, (dim2 * width) + dim1]
   }
 
-  _hit_point(geometry) : Selection {
+  _hit_point(geometry: any) : Selection {
     let {sx, sy} = geometry;
     const x = this.renderer.xscale.invert(sx);
     const y = this.renderer.yscale.invert(sy);
--- a/bokehjs/src/coffee/core/util/templating.ts
+++ b/bokehjs/src/coffee/core/util/templating.ts
@@ -39,8 +39,18 @@ export function replace_placeholders(str: string, data_source: ColumnarDataSourc
       value = special_vars[name.substring(1)]
     else {
       const column = data_source.get_column(name)
-      if (column != null)
-        value = column[i]
+      let [ind, dim1, dim2, linind] = Array.isArray(i) ? i : [i, null, null, null]
+      if ((data_source._shapes[name] != undefined) && (column != null)) {
+        if (ArrayBuffer.isView(column[ind])) { // Typed arrays use the linear index
+          value = column[ind][linind]
+        }
+        else {
+          value = column[ind][dim2][dim1]
+        }
+      }
+      else if (column != null) {
+        value = column[ind]
+      }
     }
 
     let replacement = null
--- a/bokehjs/src/coffee/core/util/templating.ts
+++ b/bokehjs/src/coffee/core/util/templating.ts
@@ -25,7 +25,8 @@ function _format_number(num: string | number): string {
     return `${num}`  // get strings for categorical types
 }
 
-export function replace_placeholders(str: string, data_source: ColumnarDataSource, i: number,
+
+export function replace_placeholders(str: string, data_source: ColumnarDataSource, i : any,
     formatters: {[key: string]: "numeral" | "printf" | "datetime"} | null = null, special_vars: {[key: string]: any} = {}): string {
 
   str = str.replace(/(^|[^\$])\$(\w+)/g, (_match, prefix, name) => `${prefix}@$${name}`)
--- a/bokehjs/src/coffee/models/selections/selection.ts
+++ b/bokehjs/src/coffee/models/selections/selection.ts
@@ -59,6 +59,7 @@ export class Selection extends Model {
                   'get_view': () => null}
     this['2d'] = {'indices': {}}
     this['1d'] = {'indices': this.indices}
+    this['im2d'] = {'indices': []}
 
     this.get_view = () => null
 
@@ -100,6 +101,7 @@ export class Selection extends Model {
       this.selected_glyphs = selection.selected_glyphs
       this.get_view = selection.get_view
       this.multiline_indices = selection.multiline_indices
+      this['im2d']['indices'] = selection['im2d']['indices']
     }
   }
 
@@ -113,7 +115,7 @@ export class Selection extends Model {
   }
 
   is_empty (): boolean {
-    return this.indices.length == 0 && this.line_indices.length == 0
+    return this.indices.length == 0 && this.line_indices.length == 0 && this['im2d']['indices'] == 0
   }
 
   update_through_union(other: Selection): void {
--- a/bokehjs/src/coffee/models/sources/cds_view.ts
+++ b/bokehjs/src/coffee/models/sources/cds_view.ts
@@ -100,6 +100,7 @@ export class CDSView extends Model {
     selection_full.update_through_union(selection_subset)
     const indices_1d = (selection_subset.indices.map((i) => this.indices[i]))
     selection_full.indices = indices_1d
+    selection_full['im2d']['indices'] = selection_subset['im2d']['indices']
     return selection_full
   }
 
@@ -108,6 +109,7 @@ export class CDSView extends Model {
     selection_subset.update_through_union(selection_full)
     const indices_1d = (selection_full.indices.map((i) => this.indices_map[i]))
     selection_subset.indices = indices_1d
+    selection_subset['im2d']['indices'] = selection_full['im2d']['indices']
     return selection_subset
   }
 
--- a/bokehjs/src/coffee/models/tools/inspectors/hover_tool.ts
+++ b/bokehjs/src/coffee/models/tools/inspectors/hover_tool.ts
@@ -359,7 +359,7 @@ export class HoverToolView extends InspectToolView {
     }
   }
 
-  _render_tooltips(ds: ColumnarDataSource, i: number, vars: any): HTMLElement {
+  _render_tooltips(ds: ColumnarDataSource, i: any, vars: any): HTMLElement {
     const tooltips = this.model.tooltips
     if (isString(tooltips)) {
       const el = div()
--- a/bokehjs/src/coffee/models/tools/inspectors/hover_tool.ts
+++ b/bokehjs/src/coffee/models/tools/inspectors/hover_tool.ts
@@ -255,6 +255,12 @@ export class HoverToolView extends InspectToolView {
       tooltip.add(rx, ry, this._render_tooltips(ds, ii, vars))
     }
 
+    for (const [ind, dim1, dim2, linind] of indices['im2d']['indices']) {
+      const vars = {index:ind, x:x, y:y, sx: sx, sy: sy}
+      let rendered = this._render_tooltips(ds, [ind, dim1, dim2, linind], vars)
+      tooltip.add(sx, sy, rendered)
+    }
+
     for (const i of indices.indices) {
       // multiglyphs set additional indices, e.g. multiline_indices for different tooltips
       if (!isEmpty(indices.multiline_indices)) {


jbednar:
Looks great!  Thanks for doing this. I'm mainly interested in this for use with datashader, for which I have three questions:

1. Datashader supports simultaneously measuring an average, count, max, min, stddev, etc, per pixel, but the image itself can only easily show one of those.  Will it be simple to make the hover show these other values, which will be in separate arrays? 

2. If 1 is supported, for datashader the values will usually have the same shape as the one being displayed, so that the same row and column lookup will work on them too, but there are similar cases where the other arrays can have different resolutions.  E.g. while looking at a visible image of LANDSAT data, you might want the hover to [show the values for all spectral bands](http://datashader.org/topics/landsat.html) , but some bands have higher spatial resolution than others.  Thus the values in such arrays might need to be indexed in data coordinates; will that work?

3. Datashader can be used either in a mode where BokehJS does the colormapping, in which case the underlying data value is available directly in JS, or where Datashader does the colormapping, in which case only the RGBA values are in the array passed to BokehJS.  Will this code be able to accept the original array as well so that the values will be meaningful in both these cases?  This seems like a special case of 1., but deserves calling out because it is the typical case.

bokehjs/src/coffee/models/glyphs/image.ts
```diff
@@ -48,6 +52,55 @@ export class ImageView extends XYGlyphView {
...
+      points.push({minX: l, minY: b, maxX: r, maxY: t, i});
+    }
+    return new RBush(points);
+  }
+
+  _lrtb(i: any) {
+    const l = this._x[i]
+    const r = l + this._dw[i]
+    const b = this._y[i]
+    const t = b + this._dh[i]
```
jbednar: Maybe switch the lines here so that they are in lrtb order to avoid cognitive dissonance. :-)

jlstevens: Sure.

jlstevens:
> Datashader supports simultaneously measuring an average, count, max, min, stddev, etc, per pixel, but the image itself can only easily show one of those. Will it be simple to make the hover show these other values, which will be in separate arrays?

This code just displays what is available in the columns of the column data source. If an image can have multiple columns for different types of value (something I haven't checked), I don't see why it couldn't work.

> ... but there are similar cases where the other arrays can have different resolutions. 

I'm not sure how that would work though. The indexing is computed once at the ``ImageView`` level for all the columns that become part of the hover information.

> ... Will this code be able to accept the original array as well so that the hover values will be meaningful in both these cases? 

Right now I think this only applies to the image glyph where the color mapping is done in JS. I haven't thought about RGB images but if they can have columns containing the original values, the same principles should apply for generating hover information.

jbednar:
Sounds good. Sounds like case 2 is infeasible, but that's ok, as we can presumably handle it by regridding any non-matching arrays.  Please consider the other two cases as feature requests, to be covered in documentation and tests even if they don't require new underlying code.

bryevdv:
ping @clairetang6 will be good to have your input

@jlstevens looks like a good start, a couple of immediate comments:

* the `indices['1d']` , is not purely for back compat. Since this never existed before, we don't need to (and shouldn't) set anything like this. Instead `Selection` should get a new real property, `image_indices` or similar

* for implementing other `@foo` fields, we should explore what to support, some ideas:

 - if the column is 1d of scalars, then the index of the entire image is used 

 - if the column is 1d of 1d arrays, then the linearindex of the particular image is used

 - if the column is 1d array of 2d arrays, then the (i,j) index is used

bokehjs/src/coffee/models/glyphs/image.ts
jlstevens: Actually, I undid that commit as ``t`` is defined using ``b`...

jbednar: Ah, good point!

bryevdv:

jlstevens:
@bryevdv 

> * the indices['1d'] etc. , is now purely for back compat. Since this never existed before, we don't need to (and shouldn't) set anything like this. Instead Selection should get a new real property, image_indices or similar

I have now removed this as requested. That said, this is what I am pushing into ``image_indices``: ``[index, dim1,  dim2, (dim2 * width) + dim1]``. The last field is the 'linear index' which makes indexing into the typed arrays easier. Here is how it is used in ``templating.ts``:

```javascript
      const [ind, dim1, dim2, linind] = Array.isArray(i) ? i : [i, null, null, null]
      if ((data_source._shapes[name] != undefined) && (column != null)) {
        if (ArrayBuffer.isView(column[ind])) { // Typed arrays use the linear index
          value = column[ind][linind]
        }
        else {
          value = column[ind][dim2][dim1]
        }
```

In order to compute this linear index I need the image ``width`` which isn't immediate accessible if I only push ``[index, dim1,  dim2]``. Any suggestions?

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -25,7 +25,8 @@ function _format_number(num: string | number): string {
     return `${num}`  // get strings for categorical types
 }
 
-export function replace_placeholders(str: string, data_source: ColumnarDataSource, i: number,
+
+export function replace_placeholders(str: string, data_source: ColumnarDataSource, i : any,
```
philippjfr: Going to make a stab at suggesting appropriate typings (but I'm definitely not an expert).

``i: number | [number, number, number, number]`` should be sufficient

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -48,6 +52,55 @@ export class ImageView extends XYGlyphView {
...
+      const [l, r, t, b] = this._lrtb(i);
+      if (isNaN(l+r+t+b) || !isFinite(l+r+t+b)) {
+        continue;
+      }
+      points.push({minX: l, minY: b, maxX: r, maxY: t, i});
+    }
+    return new SpatialIndex(points);
+  }
+
+  _lrtb(i: any) : any{
```
philippjfr: ``i: number`` is not sufficient here, certainly looks like it should be based on the way it is used.

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -48,6 +52,55 @@ export class ImageView extends XYGlyphView {
...
+    const width = this._width[index]
+    const height = this._height[index]
+    const dx = (r - l) / width;
+    const dy = (t - b) / height;
+    const dim1 = Math.floor((x-l) / dx)
+    const dim2 = Math.floor((y-b) / dy)
+    return [index, dim1,  dim2, (dim2 * width) + dim1]
+  }
+
+  _hit_point(geometry: any) : Selection {
```
philippjfr:
```
import {PointGeometry} from "core/geometry"

geometry: PointGeometry
```

jlstevens:
@philippjfr Thanks for the hints about typing! I'll have a go now...

bokehjs/src/lib/core/util/templating.ts
jlstevens: @mattpap I tried Philipp suggestion above (which looks right to me) without any luck. I get:

```
TS2538: Type 'null' cannot be used as an index type.
```

I've tried a number of variants, including some that allow ``null`` but none of them work. This column data source index is now the only new thing introduced in this PR with the ``any`` type.

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -48,6 +52,55 @@ export class ImageView extends XYGlyphView {
...
+  _image_index(index : number, x: number, y : number) : [number, number, number, number] {
+    const [l,r,t,b] = this._lrtb(index);
+    const width = this._width[index]
+    const height = this._height[index]
+    const dx = (r - l) / width;
+    const dy = (t - b) / height;
+    const dim1 = Math.floor((x-l) / dx)
+    const dim2 = Math.floor((y-b) / dy)
+    return [index, dim1,  dim2, (dim2 * width) + dim1]
+  }
```
bryevdv: This should get a comment about the 4-tuple return at least. Alternatively would it be feasible to return a structure?

bryevdv:
@jlstevens This should get some examples and sphinx docs, esp around different ways that indexing works. I think the release will not be until early next week at this point so there should be time.

bokehjs/src/lib/models/glyphs/image.ts
jlstevens: Something like:

 ```javascript
{'index':index, 'dim1':dim1, 'dim2':dim2, 'flat_index':dim2 * width) + dim1}
```?

Sure, I think that makes sense and will work and that having explicit keys isn't a bad idea. I guess it means the property definition will need to be updated accordingly.

**Comment by jlstevens** [comment]
I'm happy to add examples once we are happy with the code. There is one thing you mentioned earlier that we haven't decided on yet:

> * for implementing other `@foo` fields, we should explore what to support, some ideas:
> 
>   - if the column is 1d of scalars, then the index of the entire image is used 
>
>   - if the column is 1d of 1d arrays, then the linearindex of the particular image is used
>
>   - if the column is 1d array of 2d arrays, then the (i,j) index is used

Is the idea to have a plan to support such fields or to actually add this kind of support in this PR?

**Comment by bryevdv** [comment]
> Is the idea to have a plan to support such fields or to actually add this kind of support in this PR?

I just think we should decide and implement whatever we ultimately plan to support, and document it, in this PR. If we don't want to support all that, that's fine, we just need to decided what will be supported.

File: bokehjs/src/lib/models/glyphs/image.ts (hunk 5 above)
Comment (jlstevens) [reply bokehjs/src/lib/models/glyphs/image.ts]: Now using a structure in 4e9e2041b

File: bokehjs/src/lib/core/util/templating.ts (hunk 2 above)
Comment (jlstevens) [reply bokehjs/src/lib/core/util/templating.ts]: Defining ``ImageIndex`` helped solve this problem for ``_render_tooltips`` but I am still having trouble using it in ``replace_placeholders``....

**Comment by jlstevens** [comment]
@mattpap 

The test that is currently failing describes a mismatch between JS and Python. I assume that is to do with the property declaration. In JS I currently have:

```javascript
image_indices:     [ p.Array,   [] ]
```

And in Python:
```python
image_indices = Seq(Dict(keys_type=String, values_type=Int), help="""    """)
```

What do I need to do to get them matching and the test passing?

bryevdv:
> Maybe this is about what happens when you have extra columns and reference them?

@jlstevens yes, exactly this is for other columns

jlstevens:
In 5569b5cdd, I added handling for other CDS column names in images. Here is an example:

```python

import numpy as np

output_notebook()

list_of_lists1 = [[1,2,3,4,5,6,7,8,9,10] for i in range(5)]
list_of_lists2 = [[1,2,3,4,5] for i in range(10)]

img1 =  np.arange(50).reshape(5,10).astype(float)
img2 = np.random.rand(10, 5)

data = dict(image=[img1, img2], x=[0, 10], y=[0, 10], dw=[10, 5], dh=[5, 10], 
            urls=['url1', 'url2'], 
            times_100=[img1*100, img2*100],
            list_of_lists=[list_of_lists1, list_of_lists2])

cds = ColumnDataSource(data=data)
hover = HoverTool(tooltips=[
    ('index', "$index"),
    ("x", "$x"),
    ("y", "$y"),
    ('urls', '@urls'),
    ('times_100', '@times_100'),
    ('list_of_lists', '@list_of_lists'),
    ("image", "@image")])

p = figure(plot_width=400, plot_height=400, tools=[hover])
p.image(source=cds, image='image', x='x', y='y', dw='dw', dh='dh', palette="Spectral11")
show(p)
```

bokeh_hover_cols

This commit aims to separate the image code path using ``ImageIndex`` from the original code path. In the next commit, I'll pull out the shared lines to avoid redundancy but I'm happy to use whatever you think is clearest.

jlstevens:
I think templating.ts is clearer and less redundant in 27df2d9 even if it is a little tricker to compare to the original code path.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -36,8 +39,28 @@ export function replace_placeholders(str: string, data_source: ColumnarDataSourc
...
+          }
+          else {
+            value = data[i['flat_index']]
+          }
+        }
+        else {
+          value = data
+        }
+      }
+    }
```
bryevdv: This function is now pushing 100 lines, it would be good to split it up in to some sub-functions that are also amendable to unit tests

bryevdv: Additionally, and in particular, the code to apply a formatter should apply to images as well, so that could be pulled out and put under test, then each image/non-image path could also probably be its own helper function

philippjfr:
I'd love to see such an example too, I just tried implementing it for geoviews without much luck.

bryevdv:
Can you elaborate?

philippjfr:
Since there are no columns in the dataframe for the x and y values, ``@x{special}'``  does not seem to resolve giving me '???' and if try formatting the special var instead with ``$x{special}`` I get hover tool  entries like ``35489th``.

bryevdv:
What is the contents of your `CustomJSHover`? Looking at the code, it looks like special vars such as `$x` should have formatters applied like anything else

bryevdv:
@philippjfr can you try to construct a complete pure Bokeh example? If it's not being called we will need to run real code to debug (also just to check: you also set `formatters` on the tool?)

jlstevens:
Thanks. Hopefully we can now determine whether something needs fixing in this PR to support the custom hover or not.

bryevdv:
I think the code above works if you change to 
```
formatters = {'$x': formatter}
```

screen shot 2018-05-07 at 07 25 49

If that's the case, then I guess what I might like to do is make it so that `@x` and `@{a b}` also work in the formatters dict, so that the keys in the formatters dict exactly match whats in the tooltips. Then we'd change the examples and docs to use the "more explicit" versions. We'd leave `x` working for compatibility, but encourage the other way.

jlstevens:
> If that's the case, then I guess what I might like to do is make it so that @x and @{a b} also work in the formatters dict, so that the keys in the formatters dict can exactly match whats in the tooltips.

That sounds reasonable to me although it does sound like a change unrelated to the specifics of image hover. Is that right? I'm planning to work on the docs today and hope this PR can be merged soon.

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -48,6 +60,56 @@ export class ImageView extends XYGlyphView {
...
+
+  _image_index(index : number, x: number, y : number) : ImageIndex {
+    const [l,r,t,b] = this._lrtb(index);
+    const width = this._width[index]
+    const height = this._height[index]
+    const dx = (r - l) / width;
+    const dy = (t - b) / height;
+    const dim1 = Math.floor((x-l) / dx)
+    const dim2 = Math.floor((y-b) / dy)
+    return {index:index, dim1:dim1, dim2:dim2, flat_index :((dim2 * width) + dim1)}
```
mattpap: Before this is merged, please clean up the new or modify code. I think I wouldn't be far off saying, that there are all combinations of whitespace involved with `:`, `,`, `-` and other symbols. Also, please remove trailing `;` and double empty lines.

bryevdv:
> That sounds reasonable to me although it does sound like a change unrelated to the specifics of image hover. Is that right? I'm planning to work on the docs today and hope this PR can be merged soon.

It is tangentially related. What I am trying to avoid is having any state of the repo where we simultaneously document `$x` for some cases (matching the tooltips), but `x` for others (not matching to tooltips)

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,146 @@ import * as Numbro from "numbro"
...
+export type FormatterSpec = CustomJSHover | FormatterType
+export type Formatters = {[key: string]: FormatterSpec} | null
+export type FormatterFunc = (value:any, format:string, special_vars: Vars) => string
+export type Index = number | ImageIndex
+export type Vars = {[key: string]: any}
+
+export const DEFAULT_FORMATTERS = {
+  "numeral"  : function(value:any, format:string, special_vars: Vars) { special_vars; return Numbro.format(value, format) },
+  "datetime" : function(value:any, format:string, special_vars: Vars) { special_vars; return tz(value, format)            },
+  "printf"   : function(value:any, format:string, special_vars: Vars) { special_vars; return sprintf(format, value)       },
```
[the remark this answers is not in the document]
mattpap: Precede an argument name with `_`, so `_special_vars`. This will silence the compiler regarding unused function arguments. Note this apply only to function arguments.

bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
```diff
@@ -337,7 +344,7 @@ export class HoverToolView extends InspectToolView {
     }
   }
 
-  _render_tooltips(ds: ColumnarDataSource, i: number, vars: any): HTMLElement {
+  _render_tooltips(ds: ColumnarDataSource, i: number | ImageIndex, vars: any): HTMLElement {
```
mattpap: `Vars` was introduced earlier, so perhaps it could be used here.

jlstevens: Addressed in 0ac58b267

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,145 @@ import * as Numbro from "numbro"
...
+    return column[i]
+
+  // image index
+  const data = column[i['index']]
+  if (isTypedArray(data) || isArray(data)) {
+
+    // inspect array of arrays
+    if (isArray(data[0])) {
+      let row: any
+      row = data[i['dim2']]
```
mattpap: Unnecessary `let`.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,145 @@ import * as Numbro from "numbro"
...
+      return row[i['dim1']]
+    }
+
+    // inspect flat array
+    else
```
mattpap: I presume some people view this as more readable, but I can't disagree more. What I see when reading this code, is an `if` statement, a gap and then a dangling `else`. Although this is valid syntax, this breaks the flow of the code.

bokehjs/src/lib/core/util/templating.ts
jlstevens: If I remove the line with ``let row: any`` and simply use ``let row = data[i['dim2']]`` instead, I get:

> error TS7017: Element implicitly has an 'any' type because type '{}' has no index signature.

What do you suggest?

mattpap: `const row: any = ...`. The point of this error message is to force you to make an explicit type annotation, so that `any` doesn't slip in silently.

jlstevens: Thanks for the explanation. Addressed in 0457337df.

mattpap: This is due to `noImplicitAny: true` in `tsconfig`. Otherwise you could just write `const row = ...` and won't be aware that from this point on any usage of `row` is essentially untyped. Whenever you use `any` (especially implicit `any`), you are writing nothing more than javascript.

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -48,6 +60,55 @@ export class ImageView extends XYGlyphView {
...
+
+  _image_index(index : number, x: number, y : number) : ImageIndex {
+    const [l,r,t,b] = this._lrtb(index)
+    const width = this._width[index]
+    const height = this._height[index]
+    const dx = (r - l) / width
+    const dy = (t - b) / height
+    const dim1 = Math.floor((x - l) / dx)
+    const dim2 = Math.floor((y - b) / dy)
+    return {index: index, dim1: dim1, dim2: dim2, flat_index: ((dim2 * width) + dim1)}
```
mattpap: There is shorter `{index, dim1, dim2, flat_index: dim2*width + dim1}` syntax.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,144 @@ import * as Numbro from "numbro"
...
+    return column[i]
+
+  // image index
+  const data = column[i['index']]
+  if (isTypedArray(data) || isArray(data)) {
+
+    // inspect array of arrays
+    if (isArray(data[0])) {
+      const row: any = data[i['dim2']]
+      return row[i['dim1']]
```
mattpap: `i` is of type `ImageIndex` at this point, which is an interface, so it's better to use `i.dim1` syntax instead. Otherwise you indicate that `i` is a map/dict or something else with arbitrary keys.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,144 @@ import * as Numbro from "numbro"
...
+
+}
+
+export function get_value(name: string, data_source: ColumnarDataSource, i: Index, special_vars: Vars) {
+
+  if (name[0] == "$")
+    if (name.substring(1) in special_vars)
+      return special_vars[name.substring(1)]
+    else
+      throw new Error(`Unknown special variable '${name}'`)
```
mattpap: This needs to be either flattened or put in curly brackets.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,144 @@ import * as Numbro from "numbro"
...
+      key = name
+    const formatter = formatters[key]
+
+    if (isString(formatter)) {
+      if (formatter in DEFAULT_FORMATTERS)
+        return DEFAULT_FORMATTERS[formatter]
+      else
+        throw new Error(`Unknown tooltip field formatter type '${formatter}'`)
+    }
+    return function(value:any, format:string, special_vars: Vars) : string { return formatter.format(value, format, special_vars) }
```
mattpap: This line should be split. Also whitespace around `:`.

bokehjs/src/lib/core/util/templating.ts
```diff
@@ -3,84 +3,144 @@ import * as Numbro from "numbro"
...
+  // format spec in the formatters dict, use that
+  if (formatters != null && (name in formatters || raw_spec in formatters)) {
+
+    // some day (Bokeh 2.0) we can get rid of the check for name, and just check the raw spec
+    // keep it now for compatibility but do not demonstrate it anywhere
+    let key: string
+    if (raw_spec in formatters)
+      key = raw_spec
+    else
+      key = name
```
mattpap: Could be just `const key = raw_spec in formatters ? raw_spec : name`. Less vertical space wasted and no mutability allowed.

bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
```diff
@@ -240,6 +241,12 @@ export class HoverToolView extends InspectToolView {
       tooltip.add(rx, ry, this._render_tooltips(ds, ii, vars))
     }
 
+    for (const struct of indices.image_indices) {
+      const vars = {index: struct['index'], x: x, y: y, sx: sx, sy: sy}
```
mattpap: Same comment here regarding interfaces.

bokehjs/src/lib/models/glyphs/image.ts
jlstevens: Addressed in 6f2740d85

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -27,6 +31,14 @@ export interface _ImageData extends XYGlyphData {
 
 export interface ImageView extends _ImageData {}
 
+
+export interface ImageIndex {
+  index: number,
+  dim1: number,
+  dim2: number,
+  flat_index: number
```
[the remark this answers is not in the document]
jlstevens: Addressed in 6f2740d85

bokehjs/src/lib/core/util/templating.ts
jlstevens: Addressed in 6f2740d85

bokehjs/src/lib/core/util/templating.ts
jlstevens: Addressed in 6f2740d85

bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
jlstevens: Addressed in 6f2740d85

bokehjs/src/lib/core/util/templating.ts
jlstevens: Addressed in d0b6c9cc5

bokehjs/src/lib/core/util/templating.ts
jlstevens: Addressed in d0b6c9cc5

bokehjs/src/lib/core/util/templating.ts
jlstevens: I have no preference and I am willing to change it as suggested. I just want to check first with both you and @bryevdv that this style of comment would be acceptable for these conditionals:

```javascript
    if (isArray(data[0])) {     // inspect array of arrays
      const row: any = data[i.dim2]
      return row[i.dim1]
    }
    else                        // inspect flat array
      return data[i.flat_index]
  }
```

Or should I only put the comments on the same line for the branches after the initial ``if``?

bokehjs/src/lib/core/util/templating.ts
mattpap: Actually I was speaking only about the outer block and not the inner blocks. There isn't usually value in using curly brackets with one liners (maybe besides loops; I'm not consistent with this across all classes of statements).

jlstevens: Now the braces have been added only for the outermost block in ced4d9adc

bokehjs/src/lib/core/util/templating.ts
bryevdv: It's definitely my preference, I find things much more readable when whitespace divides them into small chunks, but I won't object to changing it.

PR #7653, opened 2018-03-19
Restore code handling empty array shape tuples

### LINKED ISSUE #7652: Regression when plotting images with non-binary array types
Using bokeh master and the most recent dev releases there appears to be a regression plotting array types which cannot be serialized as binary data, i.e. the list of lists image format and dtypes that are not supported in JS (such as int64) are affected:

```
VM30002:100 Uncaught DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The source width is 0.
```

Here is a small reproducible example:

```python
import numpy as np

N = 500
x = np.linspace(0, 10, N)
y = np.linspace(0, 10, N)
xx, yy = np.meshgrid(x, y)
d = (np.sin(xx)*np.cos(yy)*100).astype('int64')
# Also reproducible using list-of-lists format
# d = [[0, 1, 2, 3], [4, 5, 6, 7]]

p.image(image=[d], x=0, y=0, dw=10, dh=10, palette="Spectral11")

output_file("image.html", title="image.py example")

show(p)  # open a browser
```

My guess is that some change was made to image rendering in JS, which didn't take into account the fact that the image data can be a 1D typed array or a 2D regular array type. If I had to guess I think the changes to global alpha could be causing this but I haven't yet had time to do a bisection.

mattpap:
fixes #7652

### DIFF
--- a/bokehjs/src/coffee/models/glyphs/image.ts
+++ b/bokehjs/src/coffee/models/glyphs/image.ts
@@ -60,7 +60,7 @@ export class ImageView extends XYGlyphView {
 
     for (let i = 0, end = this._image.length; i < end; i++) {
       let img: Arrayable<number>
-      if (this._image_shape != null) {
+      if (this._image_shape != null && this._image_shape[i].length > 0) {
         img = this._image[i] as Arrayable<number>
         const shape = this._image_shape[i]
         this._height[i] = shape[0]
--- a/bokehjs/src/coffee/models/glyphs/image_rgba.ts
+++ b/bokehjs/src/coffee/models/glyphs/image_rgba.ts
@@ -51,7 +51,7 @@ export class ImageRGBAView extends XYGlyphView {
         continue
 
       let buf: ArrayBuffer
-      if (this._image_shape != null) {
+      if (this._image_shape != null && this._image_shape[i].length > 0) {
         buf = (this._image[i] as TypedArray).buffer
         const shape = this._image_shape[i]
         this._height[i] = shape[0]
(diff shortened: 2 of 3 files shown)


philippjfr:
Can confirm that this appears to fix the issue.

bryevdv:
@mattpap is there any reasonable tests we can add around this?

mattpap:
All those variations could be tested under "integration tests" part of examples tests. Maybe a "unit" test could do, just for making sure this code runs. However, I would like to do is to get rid of this case altogether. Handling of shapes drove me crazy when I was fixing type errors in glyphs and serialization code.

bryevdv:
> However, I would like to do is to get rid of this case altogether. 

I mean if there is a better/simpler way to preserve all existing capability, I'm all for it. But we have to handle numpy array, non numpy arrays, etc.

philippjfr:
I guess what we could do is to always send a flattened array whether it is typed or not, that way at least the shape would be consistent between the two cases. Currently the non-typed array case contains arrays-of-arrays, while the typed array case is just a flat typed array.

bryevdv:
I think the issue is not changing the data underneath the user. If they  set a list of lists and read it back, they should get a list of lists. I there is a performant was to do a late transformation then the user never sees, then that would be ok. I'd say let's get this merged  now and make another issue about trying to unify these cases though.

mattpap:
I'm not sure about the details right now, but I would like to do something about this and other issues that were brought to my attention recently, e.g. the role of `DataSource` and `ColumnarDataSource`, mutability of sources (so essentially the role of `ColumnarDataSource`), special requirements on arrays in edit tools, etc. This is subject for a broader discussion later on.

mattpap:
For now I would just add an integration test with the code from the issue. Adding such tests I usually pretty cheap. It adds maybe 2-3 seconds to examples' tests run time.

bryevdv:
@mattpap OK, great, please add that and then we will merge this.

PR #7655, opened 2018-03-19
Bryanv/deprecate row header

### LINKED ISSUE #7654: DataTable index column enhancements
The index header name and width are not configurable. Additionally, whether to have an index column or not is controlled by a mis-named property `row_header` that should be deprecated.

bryevdv:
- [x] issues: fixes #7654
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/coffee/models/widgets/tables/data_table.ts
+++ b/bokehjs/src/coffee/models/widgets/tables/data_table.ts
@@ -193,8 +193,20 @@ export class DataTableView extends WidgetView {
       columns.unshift(checkboxSelector.getColumnDefinition())
     }
 
-    if (this.model.row_headers) {
-      columns.unshift(this.newIndexColumn())
+    if (this.model.index_position != null) {
+      const index_position = this.model.index_position
+      const index = this.newIndexColumn()
+      // This is to be able to provide negative index behaviour that
+      // matches what python users will expect
+      if (index_position == -1) {
+        columns.push(index)
+      }
+      else if (index_position < -1) {
+        columns.splice(index_position+1, 0, index)
+      }
+      else {
+        columns.splice(index_position, 0, index)
+      }
     }
 
     let { reorderable } = this.model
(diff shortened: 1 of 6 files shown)


bokehjs/src/coffee/models/widgets/tables/data_table.ts
```diff
@@ -193,8 +194,20 @@ export class DataTableView extends WidgetView {
...
+      // matches what python users will expect
+      if (index_position == -1) {
+        columns.push(index)
+      }
+      else if (index_position < -1) {
+        columns.splice(index_position+1, 0, index)
+      }
+      else {
+        columns.splice(index_position, 0, index)
+      }
```
bryevdv: Hrm I was wrong about this behavior in python, I will have to think how to let users put the index at the end.

bryevdv: well, actually,  on the assumption that this is specifying what the final index of the index column *should be* then this seems to be OK. Thoughts @mattpap ?

bryevdv: Specifically, this is so `index_column = -1` will make the index column the last column.

mattpap: I wouldn't use `splice()` directly for this purpose, because its handling of negative indices is different than in Python and it's further complicated by the removal action, which can easily fool one to think that this works as expected. I would handle negative indices explicitly.

bryevdv: I tested with values -1, -N less than length, and -N greater than length, and they all worked as intended. Do you have a specific other suggestion?

mattpap: Heh, I though Python's indexing worked differently, but it looks like it's the same:
```
$ ipython
In [1]: a = [1, 2, 3]; a.insert(-1, 4); a
Out[1]: [1, 2, 4, 3]

$ node
> a = [1, 2, 3], a.splice(-1, 0, 4), a
[ 1, 2, 4, 3 ]
```
In that case you can proceed safely.

bryevdv: Right that was my comment above. In any case merging now.

PR #7658, opened 2018-03-20
add support for a hex scatter marker

### LINKED ISSUE #7638: Add a hex scatter marker
Just an x,y, size, angle scatter marker that draws little hexagons, *not* a non-overlapping tiling as described in #4786 

previous work: https://github.com/bokeh/bokeh/pull/4827

bryevdv:
- [x] issues: fixes #7638
- [x] tests added / passed

Had this lying around mostly ready to go

Works with canvas and webgl:

screen shot 2018-03-20 at 08 56 28

screen shot 2018-03-20 at 08 56 39


bryevdv:
Note that there is a separate issue for making `HexTile` glyph that will use a hex coord system on a fixed size hex tile grid (and also scale under zoom, etc)

PR #7659, opened 2018-03-20
Add example of using Bokeh server with another Tornado service

mrocklin:
@bryevdv asked for a small example using an asynchronous data source.  This is such an example.  I do not expect this to be merged, but thought that submitting it as a PR would be a good way to have a conversation around it.  There are some thoughts on current issues and opportunities for improvement at the bottom.

### DIFF
--- a/examples/app/asynchronous/main.py
+++ b/examples/app/asynchronous/main.py
@@ -1,0 +1,77 @@
+from tornado.locks import Condition
+from tornado import gen
+import time
+import random
+from collections import deque
+
+# Lets say we have some asynchronous process generating data
+# and placing that data into a fixed-length buffer
+
+buffer = deque(maxlen=10)
+condition = Condition()
+
+@gen.coroutine
+def f():
+    while True:
+        yield gen.sleep(random.random() / 10)  # sleep for a random amount of time
+        data = random.random()                 # random data
+        now = time.time()
+        buffer.append((now, data))
+        condition.notify_all()                 # inform everyone that new data has arrived
+
+# Lets run this coroutine on a Tornado event loop
+# We won't start the event loop yet though,
+# We'll add Bokeh to the same event loop in a momnet
+
+from tornado.ioloop import IOLoop
+
+io_loop = IOLoop.current()
+io_loop.add_callback(f)
+
+
+# Now we make a Bokeh Server application to render this data
+
+from bokeh.plotting import figure, ColumnDataSource
+from bokeh.server.server import Server
+
+def make_bokeh_page(doc):
+    source = ColumnDataSource({'time': [], 'value': []})
+
+    fig = figure(x_axis_type='datetime')
+    fig.line(x='time', y='value', source=source)
+
+    doc.add_root(fig)
+
+    @gen.coroutine
+    def update():
+        last_time = 0
+        while True:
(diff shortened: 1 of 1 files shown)


bryevdv:
How hard would it be to make the producer a separate process, and send data over e.g. zmq. Thinking really like a toy version of what you have in dask where data is streaming in from out of process.

mrocklin:
Not particularly hard (though I would avoid ZMQ myself) but the parts that are relevant to Bokeh are the same.  Somewhere you're going to have a tornado coroutine getting data from something asynchronously.  Here that's a while-sleep loop.  You could get data from an IOStream or something else if you wanted, but you'll have to build that something else.

mrocklin:
Maybe we can find some web source with a real-time stream of data and get async updates from it instead?  I'd be happy to help people walk through using `tornado.httpclient.AsyncClient`, I probably won't find a dataset and construct a story around it though.

bryevdv:
> Maybe we can find some web source with a real-time stream of data and get async updates from it instead?

That would be fantastic. I don't know of one off hand, but I can poke around, unless you have some immediate ideas.

mrocklin:
I looked before and didn't find much.  Ideally stock stuff, but none of the feeds are particularly accessible.  There's weather, but it updates slowly.  

In personal work I've used my network card as a source of streaming data, but this requires sudo access and involves a fair amount of parsing: https://youtu.be/yI_yZoUaz60?t=295

bryevdv:
What about taking the existing OHLC simulation code and just putting it in another process that pushes out updates? I mentioned zmq because thats what I know about in passing, but happy to use  another mechanism

mrocklin:
One could do that, yes.  You would have to ask and answer a number of questions about how you wanted to serve, serialize, frame, and move that data between processes.  The approach we would take in this example is unlikely to be related to how a user is actually going to be getting data.  

My gut reaction is that this adds a non-trivial amount of complexity to the example that is unlikely to be of use and fairly likely to be distracting.

However, that gut reaction may also just be because I'll probably be asked to help with all that :)

bryevdv:
>My gut reaction is that this adds a non-trivial amount of complexity to the example that is unlikely to be of use and fairly likely to be distracting.

Well,  for context I need to work up some content for a bokeh workshop focused specifically on streaming. There's plenty of simple lead in to do that I can put together to cover most of the course. But it seems natural to try and end with something at least tangentially approaching "real world use case". 

Absent a free publicly available streaming data source, my intent was to build some kind of toy server to fill that role. For the "real world" part, using something like zmq, protobufs, kafka... But I am only passingly familiar with any of those things. I was hoping someone with more experience could plumb together a minimal example without to much of an effort, and then I could hang the more bokeh-specific pieces off of.  I do think an end to end demonstration of streaming data from out of process would be valuable to maybe people. 

However, if the effort in that is more than I had imagined, I understand.

ivan-usov:
@bryevdv , I uploaded a very basic zmq streaming example here 
https://github.com/ivan-usov/bokeh-streaming-example
It has a data sender as a separate process and the receiving part is on the bokeh server (data is shared between all clients).

bryevdv:
@ivan-usov thanks for the link I will check that out as I put materials together

PR #7660, opened 2018-03-20
manage checkbox active state more carefully

### LINKED ISSUE #7562: CheckboxButtonGroup active list broken in 0.12.14
### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)
Bokeh 0.12.14
Windows 10
Firefox

### Complete, minimal, self-contained example code that reproduces the issue
```
from bokeh.models.widgets import CheckboxButtonGroup
from bokeh.io import curdoc

def cbg_callback(active):
    print(active)

cbg = CheckboxButtonGroup(labels=['foo', 'bar', 'baz'], active=[0, 1, 2])
cbg.on_click(cbg_callback)

layout = column(cbg)
curdoc().add_root(layout)
```

### Description of expected behavior and the observed behavior
When clicking on button 0 'foo' four times, we get the following print:
#### Expected (output in 0.12.13)
```
[1, 2]
[0, 1, 2]
[1, 2]
[0, 1, 2]
```

#### Observed in 0.12.14
```
[1, 2]
[1]
[]
[0]
```
As you see, the active list status behaves strangely. 
Pressing only button 2 'baz' repeatedly actually produces the expected output, but the second click on 1 deselects 2 as well.
A little more testing with longer lists of buttons seems to indicate that each button deselects the ones that have a higher index, one after the other.

Pressing button 0 in a five button group (all start active):
```
[1, 2, 3, 4]
[1, 2, 3]
[1, 2]
[1]
[]
[0]
```
Pressing button 2 in a five button group (all start active):
```
[0, 1, 3, 4]
[0, 1, 3]
[0, 1]
[0, 1, 2]
```

I hope there is a simple fix!

### Edit
You don't even need to run the code above. Just go to the examples to see the issue:
https://bokeh.pydata.org/en/latest/docs/user_guide/interaction/widgets.html#checkbox-button-group

Maybe it has to do with this commit?
https://github.com/bokeh/bokeh/commit/75460498f061aa1ae1b1f4aab2a27eb3890782f6#diff-cf1ec3cccc01a186330ccab9f26b5542

bryevdv:
- [x] issues: fixes #7562
- [x] tests added / passed


bryevdv:
@mattpap looking elsewhere it looks like you have a convention with methods named `change_input` and `super`. It's not used here, and I very much prefer having the code on the model so it is easily maintained under test. Should I rename it to something else?

mattpap:
>  looking elsewhere it looks like you have a convention with methods named change_input and super (...)

If there is something like that, it's most likely coincidental.

PR #7664, opened 2018-03-21
Add support for doubletap drawing on BoxEditTool

### LINKED ISSUE #7553: Make BoxEditTool match PolyDrawTool UI
The new [drawing tools in Bokeh 0.12.14](http://bokeh.pydata.org/en/latest/docs/user_guide/tools.html) are great, but there seems to be one inconsistency: BoxEditTool uses a shift tap to start a new shape, while PolyDrawTool uses double tap.  In both cases the user is starting an operation that will involve collecting multiple taps at different locations, so it seems to me that this should be the same user action in both cases. To make the interfaces more consistent, and possibly to make box drawing more feasible on tablets, can BoxEditTool be changed to start with a double tap instead of a shift tap?

philippjfr:
This PR adds support to the ``BoxEditTool`` to draw boxes by doubletapping on the canvas, moving the mouse, and then doubletapping again to finish drawing. This makes it more consistent with the ``PolyDrawTool``, while retaining the old behavior to draw on shift+drag.

- [x] issues: fixes #7553
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -909,7 +909,9 @@ class BoxEditTool(EditTool, Drag, Tap):
 
     The supported actions include:
 
-    * Add box: Hold shift then click and drag anywhere on the plot.
+    * Add box: Hold shift then click and drag anywhere on the plot or
+      double tap once to start drawing, move the mouse and double tap
+      again to finish drawing.
 
     * Move box: Click and drag an existing box, the box will be
       dropped once you let go of the mouse button.
--- a/bokehjs/src/coffee/models/tools/edit/box_edit_tool.ts
+++ b/bokehjs/src/coffee/models/tools/edit/box_edit_tool.ts
@@ -14,6 +14,7 @@ export interface HasRectCDS {
 
 export class BoxEditToolView extends EditToolView {
   model: BoxEditTool
+  _draw_basepoint: [number, number] | null
 
   _tap(ev: TapEvent): void {
     const append = ev.shiftKey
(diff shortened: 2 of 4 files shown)


bryevdv:
What happens if you double tap, then start panning (and subsequently stop panning)?

philippjfr:
Good point, currently it simply aborts drawing and deletes the box. I suppose it should probably ignore the pan events if it's already drawing?

philippjfr:
Actually it seems to have created really tiny boxes so definitely wrong.

philippjfr:
>Yes. One way might be to have two separate base points, one for each kind of interaction. The all pan callbacks can return immediately if a tap base point is set, and vice versa

Would be happy to add it but after fixing another bug the behavior seems a lot more reasonable. Since it only draws on shift-pan, you have to be pretty deliberate about it, just panning does nothing.

bryevdv:
Looks like you've already decided to do it :) I'd only suggest "x base point" and "y base point" as poised to "base point" and "something else base point".  Maybe `_tap_basepoint` and `_pan_basepoint` ?

philippjfr:
>Looks like you've already decided to do it :) I'd only suggest "x base point" and "y base point" as opposed to "base point" and "something else base point". Maybe _tap_basepoint and _pan_basepoint ?

I actually just fixed the other bug, box draw and box drag actions on pan weren't sufficiently isolated, which caused some weird issues. I've left the basepoint for the box drag action as ``_basepoint`` since it's inherited from the baseclass and used by various methods there.

Still happy to also isolate the pan-based and doubletap-based box draw action by adding distinct basepoints for those. In practice that doesn't feel like a huge problem though.

bryevdv:
OK, I will defer to you on that, happy to merge when green unless you have more to add

philippjfr:
Yes, I think it's good to merge. Doubletap is so much more intuitive I doubt many people end up using the old way anyway, your last minute suggestion on the original PR really saved the tools from being a usability nightmare.

philippjfr:
Actually, I'm going to push one more fix. It currently allows selecting while drawing and dragging which is just confusing.

PR #7669, opened 2018-03-21
allow fow simpler mercator axis creation

### LINKED ISSUE #7642: Add MercatorAxis class
Should add a `MercatorAxis` class that automatically configured mercator tickers and formatters. Additionally supper axis type `"mercator"` to `figure` for easy creation. E..g this should function:

```
from bokeh.tile_providers import CARTODBPOSITRON

output_file("tile.html")

# create plot and add tools
p = figure(x_range=(-2000000, 2000000), y_range=(-1000000, 5000000),
           x_axis_type="mercator", y_axis_type="mercator")
p.add_tile(CARTODBPOSITRON)

show(p)
```

bryevdv:
- [x] issues: fixes #7642
- [x] tests added / passed


bokeh/plotting/helpers.py
```diff
@@ -468,19 +470,11 @@ def _tool_from_string(name):
...
     if axiscls:
 
-        if axiscls is LogAxis:
-            if dim == 0:
-                plot.x_scale = LogScale()
-            elif dim == 1:
-                plot.y_scale = LogScale()
-            else:
-                raise ValueError("received invalid dimension value: %r" % dim)
-
```
bryevdv: removed this because it seems to be duplicative of `_get_scale`

bokehjs/src/coffee/models/axes/mercator_axis.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export interface MercatorAxis extends MercatorAxis.Attrs {}
+
+export class MercatorAxis extends LinearAxis {
+
+  properties: MercatorAxis.Props
+
+  ticker:    MercatorTicker
+  formatter: MercatorTickFormatter
```
mattpap: This shouldn't be necessary. `export interface MercatorAxis` makes sure that `Attrs` are both part of the interface and are "implemented". There are a few rare cases where this doesn't wok though.

bryevdv: Doesn't this specialize them to be required to be mercator ticker and formatter specifically?

mattpap: Yes, but you already did this in `Attrs` and by using `export interface MercatorAxis extends MercatorAxis.Attrs {}`. Those two lines would be necessary, if `export interface MercatorAxis (...)` wasn't there. This is a way to avoid duplication.

mattpap: Also, you don't get any errors, because what you do is a valid (yet trivial) override of already defined members.

bryevdv: If i remove those lines VSCode reports an error:

> file: 'file:///Users/bryanv/work/bokeh/bokehjs/src/coffee/models/axes/mercator_axis.ts'
severity: 'Error'
message: 'Interface 'MercatorAxis' cannot simultaneously extend types 'LinearAxis' and 'Attrs'.
  Named property 'formatter' of types 'LinearAxis' and 'Attrs' are not identical.'
at: '19,18'
source: 'ts'
code: '2320'

bryevdv: did you also mean remove the lines in `Attrs` as well?

mattpap: I see, so this is the one case where this is necessary. Just leave it as it is now.

bryevdv: OK I rebased to removed the commit

PR #7673, opened 2018-03-21
Enable user_guide examples in image diff tests

### LINKED ISSUE #2644: Enable image diff testing for User's Guide source example
ref: #2625 

Would be good to activate image diff testing for all the source examples for the user's guide, and potentially enable overall test pass/failure for this subset as well.

mattpap:
fixes #2644


bryevdv:
What is the rationale for removing `output_file`, etc? My intent with any code that is displayed along side live plots is that it be generally cut and pastable and runnable as is.

mattpap:
Most `output_file` instances had file name not matching *.py file, very often duplicated. This wouldn't work with image diff. The fastest approach was to remove it, though I left a few cases, where it made sense (e.g. using custom title). This can be resolved differently if needed.

mattpap:
However, personally I view `output_file()` more as an annoyance than a feature. I won't count how many times I've been looking for a wrong HTML file, because someone misspelled file name in `output_file()`.

mattpap:
This only adds 5 minutes to overall run time of image diff tests (though that's additional 140 examples) and already found one broken example.

bryevdv:
I definitely am in favor of generating output for all these examples, but I am still -1 on removing output file, and also -1 for importing `show` from `bokeh.plotting`. I have been trying to be fastidious about importing from the "correct" location and not relying on transitive imports for a long time now (at yoru urging, IIRC)

mattpap:
I can restore `output_file` and fill in file names. Just it isn't going to look any pretty with those overly long names, though this can be improved at later time.

mattpap:
> importing show from bokeh.plotting

As far I can tell I didn't add new instances, though I may be off by a positive number.

bryevdv:
Can't we just monkeypatch `output_file` to be a no-op when we run these? Then they will always work for testing regardless of the filename.

mattpap:
This is an idea worth exploring. After all, we already configure examples quite a bit to get them running.

PR #7680, opened 2018-03-23
Propagate settings from Bokeh Server to TornadoApplication

### LINKED ISSUE #6562: Propagate settings to tornado application from bokeh server
`examples/howto/server_embed/tornado_embed.py` suggests using bokeh server for serving web requests for other requests that don't involve bokeh. The upside is that you have to allocate only one port and specifying `allow_websocket_origin` is unneeded. The downside is that it's impossible to alter default tornado app settings.
E.g. I need to use `static_url` in my tornado template, and that requires `static_path` setting.

There's `tornado_server_kwargs` kwargs in `bokeh.server.server.Server.__init__`.
Does adding `tornado_application_settings` to the server kwargs and passing it through `bokeh.server.tornado.BokehTornado` make sense?

p-himik:
Fixes #6562

### DIFF
--- a/bokeh/server/tornado.py
+++ b/bokeh/server/tornado.py
@@ -124,6 +124,7 @@ class BokehTornado(TornadoApplication):
             If there are multiple Bokeh applications configured, this option
             has no effect.
 
+    Any additional keyword arguments are passed to ``tornado.web.Application``.
     '''
 
     def __init__(self,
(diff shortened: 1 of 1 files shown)


bryevdv:
@p-himik This looks good can you just add a note to the docstring that any additional keyword params are passed on to `TornadoApplication`

bryevdv:
Thanks @p-himik !

PR #7682, opened 2018-03-23
Hex Tiling and Binning

### LINKED ISSUE #4786: Add a hex tiling glyph
hex binning is useful (https://cran.r-project.org/web/packages/hexbin/vignettes/hexagon_binning.pdf) and I've seen a lot of cool hexmaps recently: http://www.theguardian.com/politics/ng-interactive/2016/jun/23/eu-referendum-live-results-and-analysis

screen shot 2016-07-11 at 1 34 17 pm

bryevdv:
WIP but checking in for anyone that wants to get an early look. 

(Currently) Supports:

* point top or flat top orientation
* individual hex tile scaling
* point and line hit testing

Does *not* support WebGL, and I doubt I will be capable of adding that in this round. 

- [x] issues: fixes #4786
- [x] tests added / passed


bryevdv:
@mattpap I am little hazy on when the any `self` is needed so comments welcome.

bryevdv:
Here is a sample script to play around with:

```
import numpy as np
import pandas as pd

from bokeh.models import ColumnDataSource, HoverTool
from bokeh.models.glyphs import HexTile
from bokeh.palettes import Viridis256
from bokeh.transform import linear_cmap

np.random.seed(0)
n = 10000
x = 2 + 3*np.random.standard_normal(n)
y = 2 + 3*np.random.standard_normal(n)

size=0.5

x = (x - size) / (2*size);

t1 = y / size
t2 = np.floor(x + t1)
r = np.floor((np.floor(t1 - x) + t2) / 3)
q = np.floor((np.floor(2 * x + 1) + t2) / 3) - r

df = pd.DataFrame(dict(r=r.astype(int), q=q.astype(int)))
bins = df.groupby(['q', 'r']).size().reset_index(name='counts')

plot = figure(title=None, match_aspect=True, tools="tap")

    q=bins.q,
    r=bins.r,
    c=bins.counts
))

plot.hex_tile(q="q", r="r", size=size, fill_color=linear_cmap('c', Viridis256, 0, max(bins.counts)),
              line_color=None, source=source, hover_fill_color="firebrick", hover_fill_alpha=0.2)

hover = HoverTool(tooltips=None,# [
    #     ("index", "$index"),
    #     ("(x,y)", "($x, $y)"),
    #     ("(sx,sy)", "($sx, $sy)"),
    # ],
    mode="mouse")
plot.add_tools(hover)

output_file("foo.html")

show(plot)
```

mattpap:
> I am little hazy on when the any self is needed so comments welcome.

You need to fill in all the type-related stuff for coords (`q` and `r`). You can just copy code from `XYGlyph` and rename members accordingly.

philippjfr:
Just a quick comment, I couldn't find a way to draw unequally sized hexagon using the current implementation. Hex bins/tiles are frequently used as a density map on data that has very unequal scales, so it would be very nice if that could be supported. Here's an example:

image

bryevdv:
I want to add more docs but I think they will involve some re-organization so I would like to merge this and get a dev build out, then follow up with another docs PR. @philippjfr @mattpap can you look at this first thing tomorrow?

I think we can release as soon as this work is done.

bokeh/models/glyphs.py
```diff
@@ -305,6 +306,59 @@ class HBar(Glyph):
...
+
+    q = NumberSpec(help="""
+    The "column" axial coordinates of the tile centers.
+    """)
+
+    scale = NumberSpec(1.0, help="""
+    A scale factor for individual tiles.
+    """)
+
+    orientation = String("pointytop", help="""
```
mattpap: This should be an enum.

bokeh/models/glyphs.py
```diff
@@ -305,6 +306,59 @@ class HBar(Glyph):
...
+    __example__ = "examples/reference/models/HexTile.py"
+
+    # a canonical order for positional args that can be used for any
+    # functions derived from this class
+    _args = ('q', 'r')
+
+    size = Float(1.0, help="""
+    The radius (in data space units) of the hex tiling.
+
+    The radius is always measured along the cartesian y-axis for "pointy_top"
```
mattpap: `pointytop`

bokehjs/src/coffee/models/glyphs/hex_tile.ts
```diff
@@ -0,0 +1,279 @@
...
+export class HexTileView extends GlyphView {
+  model: HexTile
+  visuals: HexTile.Visuals
+
+  scenterx(i: number): number { return this.sx[i] }
+
+  scentery(i: number): number { return this.sy[i] }
+
+  protected _set_data(): void {
+    const self = this as any
```
mattpap: Those `any`s have to be removed. Adding `_q` and `_r` (`Arrayable<number>`) to `HexTileData` should be sufficient.

bokehjs/src/coffee/models/glyphs/hex_tile.ts
```diff
@@ -0,0 +1,279 @@
...
+        continue
+      points.push({minX: x-xsize, minY: y-ysize, maxX: x+xsize, maxY: y+ysize, i})
+    }
+    return new RBush(points)
+  }
+
+  // overriding map_data instead of _map_data because the default automatic mappings
+  // for other glyphs (with cartesian coordinates) is not useful
+  map_data(): void {
+    const self = this as any
```
mattpap: Same here.

bokehjs/src/coffee/models/glyphs/hex_tile.ts
```diff
@@ -0,0 +1,279 @@
...
+      const svy = [0, -h,  -h,   0,  h,  h ]
+
+      return [svx, svy]
+    }
+
+  }
+
+  protected _render(ctx: Context2d, indices: number[], {sx, sy, scale}: HexTileData): void {
+
+    const [svx, svy] = this._get_unscaled_vertices()
```
mattpap: This was already computed in `map_data()`.

bokehjs/src/coffee/models/glyphs/hex_tile.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  protected _hit_point(geometry: PointGeometry): Selection {
+    const {sx, sy} = geometry
+    const x = this.renderer.xscale.invert(sx)
+    const y = this.renderer.yscale.invert(sy)
+
+    const candidates = this.index.indices({minX: x, minY: y, maxX: x, maxY: y})
+
+    const hits = Array()
```
mattpap: Idiomatic code should read `const hits = []`. This isn't incorrect, but it catches the eye (maybe the author missed an initializer, etc.).

bokeh/util/hex.py
```diff
@@ -0,0 +1,254 @@
...
+
+    '''
+    HEX_FLAT = [2.0/3.0, 0.0, -1.0/3.0, np.sqrt(3.0)/3.0]
+    HEX_POINTY = [np.sqrt(3.0)/3.0, -1.0/3.0, 0.0, 2.0/3.0]
+
+    coords = HEX_FLAT if orientation == 'flattop' else HEX_POINTY
+
+    x =  x / size * (aspect_scale if orientation == "pointytop" else 1)
+    y = -y / size / (aspect_scale if orientation == "flattop" else 1)
+    print(y)
```
mattpap: ^^^

bokehjs/src/coffee/models/glyphs/xy_glyph.ts
```diff
@@ -7,6 +7,8 @@ export interface XYGlyphData extends GlyphData {
   _x: Arrayable<number>
   _y: Arrayable<number>
 
+  scale: Arrayable<number>
```
mattpap: This isn't being used anywhere.

bryevdv: it's passed to `_render`

mattpap: Sure, but that's `HexTileData.scale` and this `XYGlyph.scale`. `HexTile` doesn't extend `XYGlyph`.

bokehjs/src/coffee/models/glyphs/hex_tile.ts
bryevdv: It's not available to hover, selection glyphs, and they don't work.

mattpap: So, please add this as a comment, because it's likely I will be tripping over this sooner than later.

bryevdv: same thing with `scale` IIRC, for now it has to be passed in to `_render` (or computed in render) for alt glyphs to work. In the future, the `GlyphRenderer` should take responsibility for these computed quantities so that all glyphs can see it.

bryevdv: I realized I could pass it to render by just putting it in the argument that unpacks the view attributes

bryevdv:
@mattpap I have pushed a commit to address your comments

mattpap:
Looks good.

PR #7683, opened 2018-03-23
Add maintain_focus parameter to WheelZoomTool

### LINKED ISSUE #5821: Zooming out with bound ranges doesn't zoom all the way
Bokeh version: 0.12.4

1. set `bounds='auto'` on `x_range` and `y_range` range
2. zoom in with mouse pointer near the lower left corner of the plot
3. zoom out with mouse pointer near the upper right corner of the plot

Observed behavior:
Plot is zoomed in just fine, but zooming out doesn't bring the plot to the same state as before apparently because zooming in doesn't try to move plot at all

Expected behavior:
The plot moves as to both be inside the specified bounds and to be zoomed out up to the initial state when zooming out is no longer possible without breaking bounds contstraint.

p-himik:
Fix #5821

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -218,6 +218,12 @@ class WheelZoomTool(Scroll):
     vertically across the height of the plot.
     """)
 
+    maintain_focus = Bool(default=True, help="""
+    Whether or not zooming tool maintains its focus position. Setting it
+    to False results in a more "gliding" behavior, allowing one to
+    zoom out more smoothly, at the cost of losing the focus position.
+    """)
+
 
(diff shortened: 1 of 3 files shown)


bryevdv:
LGTM @mattpap do you have any comments?

mattpap:
Looks good.

bryevdv:
Thanks @p-himik !

cjw296:
In relation to https://github.com/bokeh/bokeh/issues/11294, `maintain_focus = False` makes the experience a little better for zooming in, but still has the same problem for zooming out.

PR #7686, opened 2018-03-23
bokeh server support for running in JupyterHub. (#7685)

### LINKED ISSUE #7685: bokeh server compatibility with running behind jupyterhub
Feature Request: Allow bokeh server to run behind JupyterHub (and other reverse proxies)

Right now, bokeh supports being run in a JupyterLab notebook using the appropriate plugin (https://github.com/bokeh/jupyterlab_bokeh).  This plugin is great, and allows for bokeh graphs to be rendered in JupyterLab notebooks.

For JupyterHub, there is an additional level of proxing to get to your JupyterLab environment (and the bokeh server running therein), which is provided by JupyerHub.  This maps URLs like https://jupyterhub.example.com/user/cbanek to the URLs on the JupyterLab environment.

When running bokeh server in a Jupyter notebook, bokeh emits javascript that points the browser to the URL of the bokeh server.  In this case, bokeh doesn't know that the server is behind a reverse proxy (JupyterHub) and therefore emits a URL that cannot be accessed (http://localhost:12345)

There are two components to getting this scenario to work:
1. Proxying URL requests to the correct port on the correct container via URL.
2. Giving bokeh the public URL that uses the proxy to emit into the javascript.  Bokeh can't know the proper URL, as is the case in many reverse proxy situations.

To solve this, I've combined nbserverproxy (https://github.com/jupyterhub/nbserverproxy) to provide a URL that gets me to the bokeh server port given a well formed URL.  This URL is of the form https://jupyterhub.example.com/user/xxx/proxy/port/ .  This means that HTTP requests received on this URL are then sent to the JupyterLab instance on the port provided in the URL.  This allows for a browser given the right URL to forward an HTTP request (or websocket) to any port on JupyterLab.

bokeh doesn't know how to generate these URLs, as bokeh isn't aware of the public URL of the proxy you are connecting to.  To solve this, I will be submitting an associated pull request that allows for the injection of a function to help define this URL, after the port has been bound (and therefore known) but before bokeh emits the URL into the notebook cell.  I am using environment variables to help construct the URL in this injected function, though different cases might need different logic, hence the function.

I've seen only one other person attempting something similar in my research:
https://stackoverflow.com/questions/48625893/using-bokeh-server-in-jupyter-notebook-behind-proxy-jupyterhub

cbanek:
- [x] issues: fixes #7685
- [x] tests added / passed
- [x] release document entry (if new feature or API change) - new docstrings added.

### DIFF
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -430,7 +430,7 @@ def publish_display_data(*args, **kw):
     return publish_display_data(*args, **kw)
 
 @dev((1,0,0))
-def show_app(app, state, notebook_url, port=0):
+def show_app(app, state, notebook_url, port=0, proxy_url_func=None):
     ''' Embed a Bokeh serer application in a Jupyter Notebook output cell.
 
     Args:
@@ -443,6 +443,9 @@ def show_app(app, state, notebook_url, port=0):
         port (int) : the port the embedded server will listen on. By default
         the port is 0, which results in the server listening on a random dynamic port.
 
+        proxy_url_func (function(string,int)): function that takes the notebook_url
+        and server port and returns the server url to connect to.
+
     Returns:
         None
 
@@ -454,6 +457,9 @@ def show_app(app, state, notebook_url, port=0):
 
     loop = IOLoop.current()
 
+    if not proxy_url_func:
+        proxy_url_func = _server_url
+
     origin = _origin_url(notebook_url)
     server = Server({"/": app}, io_loop=loop, port=port,  allow_websocket_origin=[origin])
 
@@ -461,10 +467,13 @@ def show_app(app, state, notebook_url, port=0):
     curstate().uuid_to_server[server_id] = server
 
     server.start()
-    url = _server_url(notebook_url, server.port)
+    url = proxy_url_func(notebook_url, server.port)
+
+    logging.debug("Server URL is %s" % url)
+    logging.debug("Origin URL is %s" % origin)
 
     from ..embed import server_document
-    script = server_document(url)
+    script = server_document(url, resources=None)
 
     publish_display_data({
         HTML_MIME_TYPE: script,
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -434,22 +434,28 @@ def show_app(app, state, notebook_url, port=0):
     ''' Embed a Bokeh serer application in a Jupyter Notebook output cell.
 
     Args:
-        app (Application) :
-
+        app (Application or callable) :
+            A Bokeh Application to embed inline in a Jupyter notebook.
+            
         state (State) :
+            ** Unused **
 
-        notebook_url (str or function) :
-        If notebook_url is a string, this string is parsed to construct
-        the origin and server URLs for the server.
-        If notebook_url is a function, it takes one parameter, the port.
-        If the port is present, generate the server URL, otherwise
-        generate the origin URL for the server.
+        notebook_url (str or callable) :
+            The URL of the notebook server that is running the embedded app.
+            
+            If ``notebook_url`` is a string, the value string is parsed to 
+            construct the origin and full server URLs.
 
-        port (int) : the port the embedded server will listen on. By default
-        the port is 0, which results in the server listening on a random dynamic port.
+            If notebook_url is a callable, it must accept one parameter,
+            which will be the server port, or None. If passed a port,
+            the callable must generate the server URL, otherwise if passed
+            None, it must generate the origin URL for the server.
 
-        proxy_url_func (function(string,int)): function that takes the notebook_url
-        and server port and returns the server url to connect to.
+        port (int) : 
+            A port for the embedded server will listen on.
+
+            By default the port is 0, which results in the server listening
+            on a random dynamic port.
 
     Returns:
         None
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -436,14 +436,14 @@ def show_app(app, state, notebook_url, port=0):
     Args:
         app (Application or callable) :
             A Bokeh Application to embed inline in a Jupyter notebook.
-            
+
         state (State) :
             ** Unused **
 
         notebook_url (str or callable) :
             The URL of the notebook server that is running the embedded app.
-            
-            If ``notebook_url`` is a string, the value string is parsed to 
+
+            If ``notebook_url`` is a string, the value string is parsed to
             construct the origin and full server URLs.
 
             If notebook_url is a callable, it must accept one parameter,
@@ -451,7 +451,7 @@ def show_app(app, state, notebook_url, port=0):
             the callable must generate the server URL, otherwise if passed
             None, it must generate the origin URL for the server.
 
-        port (int) : 
+        port (int) :
             A port for the embedded server will listen on.
 
             By default the port is 0, which results in the server listening
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -430,7 +430,7 @@ def publish_display_data(*args, **kw):
     return publish_display_data(*args, **kw)
 
 @dev((1,0,0))
-def show_app(app, state, notebook_url, port=0, proxy_url_func=None):
+def show_app(app, state, notebook_url, port=0):
     ''' Embed a Bokeh serer application in a Jupyter Notebook output cell.
 
     Args:
@@ -438,7 +438,12 @@ def show_app(app, state, notebook_url, port=0, proxy_url_func=None):
 
         state (State) :
 
-        notebook_url (str) :
+        notebook_url (str or function) :
+        If notebook_url is a string, this string is parsed to construct
+        the origin and server URLs for the server.
+        If notebook_url is a function, it takes one parameter, the port.
+        If the port is present, generate the server URL, otherwise
+        generate the origin URL for the server.
 
         port (int) : the port the embedded server will listen on. By default
         the port is 0, which results in the server listening on a random dynamic port.
@@ -457,17 +462,22 @@ def show_app(app, state, notebook_url, port=0, proxy_url_func=None):
 
     loop = IOLoop.current()
 
-    if not proxy_url_func:
-        proxy_url_func = _server_url
+    if callable(notebook_url):
+        origin = notebook_url(None)
+    else:
+        origin = _origin_url(notebook_url)
 
-    origin = _origin_url(notebook_url)
     server = Server({"/": app}, io_loop=loop, port=port,  allow_websocket_origin=[origin])
 
     server_id = uuid4().hex
     curstate().uuid_to_server[server_id] = server
 
     server.start()
-    url = proxy_url_func(notebook_url, server.port)
+
+    if callable(notebook_url):
+        url = notebook_url(server.port)
+    else:
+        url = _server_url(notebook_url, server.port)
 
     logging.debug("Server URL is %s" % url)
     logging.debug("Origin URL is %s" % origin)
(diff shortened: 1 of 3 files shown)


bokeh/io/notebook.py
```diff
@@ -454,17 +457,23 @@ def show_app(app, state, notebook_url, port=0):
 
     loop = IOLoop.current()
 
+    if not proxy_url_func:
+        proxy_url_func = _server_url
+
```
bryevdv: Instead of this check, couldn't `_server_url` just be the default arg value (instead of `None`)?

bokeh/io/notebook.py
```diff
@@ -454,17 +457,23 @@ def show_app(app, state, notebook_url, port=0):
...
     server.start()
-    url = _server_url(notebook_url, server.port)
+    url = proxy_url_func(notebook_url, server.port)
+
+    logging.debug("Server URL is %s" % url)
+    logging.debug("Origin URL is %s" % origin)
 
     from ..embed import server_document
-    script = server_document(url)
+    script = server_document(url, resources=None)
```
bryevdv: Why are you passing `resources=None` here? AFAIK this will cause normal usage to break.

bryevdv:
Hi @cbanek thanks for your interest in Bokeh and the PR, I left a couple of questions above.

bokeh/io/notebook.py
cbanek: Excellent point, but the definition of the function is above that of _server_url.  It seems like in the code you generally want to keep the private functions at the bottom, but if the definition of _server_url was moved above, I can do that.  I've had notoriously bad luck passing complex default parameters in python (like the dreaded default list argument)

bokeh/io/notebook.py
cbanek: I thought normal bokeh server code doesn't go through this pathway - only the notebook loading.  For our case, and I thought was the general case, output_notebook is always assumed to be called and has the javascript.  This seemed to try to load the javascript on each cell, which seemed excessive.  Also, the URLs there were munged in a way that had them 404.  I could do more work on server_document to fix more of the URL parsing, but was also afraid of breaking things.  If memory serves me correctly, it assumes again that all the JS being served is right off the host, and doesn't have any additional path items from the URL.

cbanek:
Thanks @bryevdv for the feedback.  If you think this is the right way with the bokeh architecture, I'm happy to keep working on this to get it in.  If you think there's a better way, I'm also happy to learn that.

Other ideas for expanding this PR and fixing a few more things are the server_document comment above, also if you try to run bokeh server on https, the default _server_url implementation will also produce a bad URL.  (Could be considered a separate issue/PR?)

bokeh/io/notebook.py
bryevdv: I would actually suggest this:

* un-private it, i.e. rename to `default_server_url` and mark `@dev`
* make it a required parameter to `show_app`
* import `default_server_url` in `bokeh.io.showing` and make it the default arg to `show`

I also think `proxy_url_func` should be renamed to `server_url_function` since there may be cases when this hook is useful that do not involve proxies.

cbanek: Sounds good.  Naming is one of the hardest things in CS, so typically I let other people come up with the names.

bryevdv: Actually, would it be feasible to let `notebook_url` accept a callable in addition to a string? In which event the callable is responsible for generating the right URL. If this is possible, I think I'd prefer it to adding yah another parameter to `show` that only applies in a subset of use scenarios.

bryevdv: i.e., `show(app, notebook_url=my_url_function)`  If a string is passed, the current code path with `_server_url` is used, if a function is passed, it is delegated to entirely, and `_server_url` is not called.

cbanek: Sure that seems reasonable, I think it's mostly up to style (some people really don't like overloading their parameters too much).

bryevdv: If JupyterHub follows a common format, we could provide a `jupyter_hub` convenience wrapper that returns a url function that is properly configured:
```
show(app, notebook_url=jupyter_hub("https://foo.com"))
```

bryevdv: OK, let's see what that looks like. I definitely agree that in most cases overloading is to be avoided, but I think there is a balance here to not add more parameters to `show`

bokeh/io/notebook.py
bryevdv: Hrmm, I guess I will have to test it out. I had not considered that the resources loaded by `output_notebook` would be accessible to a notebook app, but perhaps that is in fact the case.

bryevdv:
> if you try to run bokeh server on https, the default _server_url implementation will also produce a bad URL. 

Can you elaborate? Also yes let's make a separate issue for that. Thanks for working on these issues!

bokeh/io/notebook.py
cbanek: Totally up to you.

As for providing the function, you can see what my definition is here:
https://github.com/lsst-epo/hr-diagram-investigations/blob/0148deaefe038dd3d43b5e6c980d8a5fee4661ec/notebooks/hr/__init__.py#L24

I'm not sure how many of those environment variables are automatically put there by jupyter (I believe this might be up to configuration).  Also it requires installing nbserverproxy (a notebook extension).

I'm happy to provide my proxy function though, it wouldn't even need to take a parameter, which is good, because partial functions make me crazy.  The reason why it has to be a function rather than a static URL is because the port number is going to end up in the URL and isn't known when calling show if port=0.  You want to pick port=0 though, so you don't have to worry about the port conflicts / being already bound from a previous run of the cell.  This means that you have to generate the URL after binding to the port, hence the function call.

bryevdv: OK if it's not known to be fully general (i.e. because env vars may or may not be there) then we might instead add docs/examples that demonstrate your jupyterhub function, instead of adding it directly into the core library.

athornton: Jupyter sets JUPYTERHUB_SERVICE_PREFIX, but EXTERNAL_URL is something that I'm setting in my OAuth spawner configuration (and which @cbanek is inheriting from me):

`pod_env['EXTERNAL_URL'] = oauth_callback[:-len(endstr)]`

cbanek:
As for the problem I mentioned with _server_url - I have filed #7701 to track this issue.

cbanek:
I've gone ahead and done what we've talked about with respect to overloading notebook_url to take a function as opposed to a string.  I think it has worked out pretty well.

I've expanded the scope a bit to include the _origin_url path as well, as that might need to be set correctly depending on the proxy settings / network configuration that bokeh isn't aware of.

bokeh/io/showing.py
```diff
@@ -38,7 +38,7 @@
 #-----------------------------------------------------------------------------
 
 @general((1,0,0))
-def show(obj, browser=None, new="tab", notebook_handle=False, notebook_url="localhost:8888"):
+def show(obj, browser=None, new="tab", notebook_handle=False, notebook_url="localhost:8888", proxy_url_func=None):
```
bryevdv: `proxy_url_func` can be removed now?

bryevdv:
@cbanek looks good, I only had one comment. Do you mind if I make some style edits on the docs? I think it will be quicker if I just make them, rather than trying to describe them. I'ld like to get this merged tonight so that it can be in the upcoming release for you.

cbanek:
Pushed, thanks for spotting the parameter still there.

I'm also hoping to look a little more closely tomorrow at the resources=None line, and how much work it might be to fix whatever was behind that (or at least report an issue).

Thanks for helping me out with docs.

bryevdv:
Thanks @cbanek !

rsignell-usgs:
@bryevdv, I'm struggling to understand how this feature works.    I've got a [simple bokeh interaction notebook](https://gist.github.com/rsignell-usgs/dc4de1264b85ca762571e976f70b74e5) that captures the coordinates  the user draws, and it works in Jupyter just fine:

but in JupyterHub the `show(modify_doc)` doesn't produce a plot, and the developer console shows that JS failed to load response data from this request:
```
http://localhost:43474/autoload.js?bokeh-autoload-element=111c97fa-dbc8-437c-9770-471dc23fb13f&bokeh-absolute-url=http://localhost:43474&resources=none
```
Since I access the Dask Dashboard on my jupyterhub at 
http://pangeo.esipfed.org/user/rsignell-usgs/proxy/8787
I though maybe something like this could work:
```
show(modify_doc, notebook_url='pangeo.esipfed.org/user/rsignell-usgs/proxy')
```
but that generated this URL:
```
http://pangeo.esipfed.org/user/rsignell-usgs/proxy:34560/autoload.js?bokeh-autoload-element=870004ec-7366-4b38-b20f-2119e2b52327&bokeh-app-path=/user/rsignell-usgs/proxy:34560&bokeh-absolute-url=http://pangeo.esipfed.org/user/rsignell-usgs/proxy:34560&resources=none
```
that gave a 404 error. 

I gather that I should be able to specify `notebook_url` to be a function to make this work, but I can't figure it out.

Is it clear what I'm doing wrong, or is there an example somewhere of how this works?

bryevdv:
@rsignell-usgs I have never personally used Jupyter Hub, and didn't author the PR, so I can't really offer any specific guidance. Perhaps @cbanek can chime in. Otherwise, I'd suggest taking support questions to the mailing list, or to stack overflow .

cbanek:
@rsignell-usgs It sounds like you are on the right track.  You can't use the notebook_url with a string and have it work.  This is because you don't know what ephemeral port Tornado will assign to the bokeh server.  This is why if you look at your URL there's that nasty :34560 in there, which is the port number.

You need to provide a function as the notebook_url parameter.  like I have here:
https://github.com/lsst-epo/vela/blob/1b870efe10c81f30765991475722e356b65d7f39/astropixie-widgets/astropixie_widgets/config.py#L10

This function returns the URL for the origin header, and what URL to inject into the javascript that is put into the notebook cell to connect to.  The function that I provide as an example has a few environment variables that are referenced, JUPYTERHUB_SERVICE_PREFIX which jupyterhub will inject into your environment, and EXTERNAL_URL, which you need to set yourself.  EXTERNAL_URL should be something like http://pangeo.esipfed.org for you, and JUPYTERHUB_SERVICE_PREFIX should be set by jupyter to /user/rsignell-usgs.  By using JUPYTERHUB_SERVICE_PREFIX this allows for any user on your jupyterhub environment to load a bokeh graph running from their own notebook (it handles the username part of the URL).

You also need to install the nbserverproxy extension to your notebook (https://github.com/jupyterhub/nbserverproxy).  This will allow a URL to route to any port on the container hosting your jupyterlab server, and allows it all to use port 8888 to talk with the hub and your browser. (Otherwise you'd have to worry about setting up additional routes for the bokeh server port, which again, you don't know up front).

For the javascript loading, I've simply called output_notebook() before, which contains all the JS you need to connect to the bokeh server. Once you pass in notebook_url as a function, it shouldn't try to download the JS from the Tornado server any more and will rely on the JS from a previously called output_notebook() call.

Hope this helps.

rsignell-usgs:
@cbanek ,  fantastic!  My [notebook updated to use your `remote_jupyter_proxy_url` function](https://nbviewer.jupyter.org/gist/rsignell-usgs/dc4de1264b85ca762571e976f70b74e5) works perfectly!


Thank you so much!

bryevdv:
This seems like useful and important information. Would one of you mind making a PR to update the file linked below?

https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/notebook.rst

I'd suggest a new section heading "JupyterHub Configuration" or something like that that describes when and how to use the `notebook_url` parameter with a function.

rsignell-usgs:
@bryevdv , I'm hoping @cbanek can do this as she is the expert, but if she doesn't have time, I can try.

PR #7694, opened 2018-03-26
Make all tests compatible with xdist

### LINKED ISSUE #7679: Make it possible to use pytest-xdist for all tests
Currently, running something like `pytest -n auto` results in an error - different nodes collect different test.

p-himik:
Fixes #7679


bryevdv:
@p-himik FWIW I am willing to entertain the notion of scrapping the API spec. I guess we don't have to do that now, but in any case why is the `OrderedDict` only needed in one place but not everywhere?

p-himik:
@bryevdv Yeah, working around unordered members of a class was rather simple.
Ah, good catch - that's just me being sloppy.

bryevdv:

bryevdv:
FYI I've resolved the merge conflict.

bryevdv:
I'm going to go ahead and close this, given the conflict and inactivity. @p-himik please feel free to open a new PR for this anytime. Also note the API tests have been removed, in case that makes things simpler

PR #7695, opened 2018-03-26
Force "use strict" in CustomJS and related models

### LINKED ISSUE #6306: Force strict mode for CutomJS, etc. written in pure JS
We use strict mode everywhere beside this case. For the sake of consistency we should change that.

mattpap:
fixes #6306

### DIFF
--- a/bokeh/models/callbacks.py
+++ b/bokeh/models/callbacks.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 from types import FunctionType
 
 from ..core.has_props import abstract
-from ..core.properties import Dict, Instance, String
+from ..core.properties import Dict, Instance, String, Bool
 from ..model import Model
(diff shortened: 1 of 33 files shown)


bryevdv:
We have to do this in stages: 

* update all the examples
* add strict as an optional mode (off by default)
* add a validation warning if strict mode is off
* turn strict mode on

The first three can happen all at once, but turning it on (and breaking every js callback out in the wild) has to happen later.

bryevdv:

mattpap:
Given the extent of this change so far, I would like to get this merge, and I will add deprecation in another PR.

bryevdv:
> Given the extent of this change so far, I would like to get this merge, and I will add deprecation in another PR. 

OK, should get a release note in that PR too

Issue #4749, opened 2016-07-05
Custom CSS in Jupyter Notebooks


AlbertDeFusco:
I was attempting to use `bokeh.resrouces.Resources` to create a Resources object with custom CSS, particularly for the HoverTool. I order to do this I had to subclass Resources and write a method to add my own CSS. It would be really great if something like this was available.

``` python
from bokeh.resources import Resources

class myCDN(Resources):
    def add_css(self, input_css):
        _, self.extra_css = self._resolve('css')
        self.extra_css.append(input_css)

    @property
    def css_raw(self):
        _, raw = self._resolve('css')
        raw.extend(self.extra_css)
        return raw
```

An [example notebook](https://anaconda.org/defusco/custom-tooltips/notebook) is on my Anaconda Cloud channel.

bryevdv:
@mattpap is working on similar things as part of other issues, so I will close this.

PR #7708, opened 2018-03-27
Allow easy customization of HTML templates

### LINKED ISSUE #4986: Allow elements to be placed in server templates
In https://github.com/dask/distributed/pull/434 I want to add a button to a bokeh server application that uses the `stretch_both` sizing scheme.  Putting the button in the normal layout causes some dead space due to the current 50/50 split scheme.  A workaround would be to place the button into the template.  

@birdsarah suggests that while this is possible normally, it's not yet possible for server templating.

mattpap:
Early work in progress.

fixes #4986


bryevdv:

mattpap:
This PR is a good opportunity to make CSSResources obsolete, unless there are compelling reasons for keeping it (there is not a single examples showing its usefulness).

mattpap:
Also #5021 won't be needed anymore, because it will be easy to add custom resources without touching bokeh's resources.

mattpap:
Another similar issue is #4749. Instead of defining custom resources, one will be able to just add those to a template.

bryevdv:
@mattpap  when I run this `main.py`:
```
rom bokeh.io import curdoc

p0 = figure(x_range=(0,10), y_range=(0,10))

p1 = figure(sizing_mode="scale_width", x_range=(0,10), y_range=(0,10))

curdoc().add_root(p0)
curdoc().add_root(p1)
```

Then this template, with just `embed(roots[0])` displays both plots:
```
{% extends base %}

{% block contents %}
  {{ embed(roots[0]) }}

  {{ plot_script | indent(8) }}
{% endblock %}
```
And if I try a template with `{{ embed(roots[1]) }}` then I get a Jinja exception that this list is too short:

> jinja2.exceptions.UndefinedError: list object has no element 1

What am I missing?

bryevdv:
Maybe the double duplication had something to do with template inheritance? I thought not calling `super()` would suppress the base `content` block but evidently not? This template with no inheritance just displays both plots once (but I maintain it should only display the *first* plot, not both):

```
{% macro embed(root) -%}
<div class="bk-root">
    <div class="bk-plotdiv" id="{{ root.elementid }}"></div>
</div>
{%- endmacro %}

<!DOCTYPE html>
<html lang="en">
    <head>
      <meta charset="utf-8">
      <title>{% block title %}{{ title | e if title else "Bokeh Plot" }}{% endblock %}</title>
      {% block resources %}
        {% block js_resources %}
          {{ bokeh_css | indent(8) if bokeh_css }}
        {% endblock %}
        {% block css_resources %}
          {{ bokeh_js | indent(8) if bokeh_js }}
        {% endblock %}
      {% endblock %}
    </head>
    <body>
      {{ embed(roots[0]) }}
      {{ plot_script | indent(8) }}
    </body>
</html>
```

bryevdv:
digging more, if I run the code above, and put a `print` in `PULL-DOC-REPLY`, then I see:
```
         'root_ids':[  
            'b37c81a2-04e6-4807-8aad-f1799a449fdc',
            '553214a1-0ef3-41a2-b9c1-c659652bcfb9'
         ]
```
But if I view page source for the page that just opens, then I see:
```
var render_items = [{"elementid":"9fed516c-fb36-42d8-8b42-3b03b0d9699b","sessionid":"yEzQ1bqFfbmLxliiZ1B9fa8AyQzWFndpPuAWp0Mc9xV7","use_for_title":true}];
```
So that would appear to explain things, I believe `render_items` needs to specify the model ids, not just a session (which default so showing everything)

Tangentially, I noticed that after the `PULL-DOC-REPLY` finishes, it appears that the server redundantly patches lots of attributes:

```
2018-05-06 10:07:41,118 Sending pull-doc-reply from session 'AA99nBWH69yrZ57fHGCCAiEmyJnOqdSO4SGaUn7l1SsF'
--------- print after pull-doc-reply
2018-05-06 10:07:41,204 Patching attribute 'plot' of LinearAxis(id='c51f46f0-8692-4293-bbfe-b04b2c5da060', ...) with {'type': 'Plot', 'id': 'bbd7596b-8d71-49a3-9a6e-22fcf10671be', 'subtype': 'Figure'}
2018-05-06 10:07:41,204 Patching attribute 'ticker' of LinearAxis(id='c51f46f0-8692-4293-bbfe-b04b2c5da060', ...) with {'type': 'BasicTicker', 'id': 'c558ff61-b755-4803-9193-92b4eb9952ea'}
2018-05-06 10:07:41,205 Patching attribute 'formatter' of LinearAxis(id='c51f46f0-8692-4293-bbfe-b04b2c5da060', ...) with {'type': 'BasicTickFormatter', 'id': '2a4fe606-6512-400d-aa47-2ebb58f2e5f3'}

...... lots more
```

bryevdv:
Also other comments for now: I like the `embed` macro syntax well enough. I do think we will need to document ways for people to take some arbitrary big template they already have, say, and insert Bokeh components into it, and not just expecting that everyone can or will inherit from our base template and try to fit their content into its form.

Karel-van-de-Plassche:
This update looks great! Can this also be used in conjunction with `bokeh serve`/`curdoc.add_root` paradigm?

mattpap:
> Can this also be used in conjunction with bokeh serve/curdoc.add_root paradigm?

Not yet. This is a major issue to resolve, to get this PR merged.

examples/integration/embed/css_grid_simple_no_extend.py
```diff
@@ -0,0 +1,60 @@
...
+  <body>
+    <div class="grid">
+      {% for root in roots %}
+        {{ embed(root) }}
+      {% endfor %}
+    </div>
+    {{ plot_script }}
+  </body>
+</html>
+""")
```
mattpap: @bryevdv, this shows how to use a custom template that doesn't extend the base template.

bryevdv:
I will proceed with the release without this PR, it's just too risky this soon before a release. Apart from the breaking tests, it does not seems to functioning yet:

screen shot 2018-05-11 at 19 56 51

bryevdv:
OK, can confirm the example above works after a clean build:

screen shot 2018-05-23 at 12 06 18

I'll play around with this some more, and perhaps add some docs.

Karel-van-de-Plassche:
Would be great to have a 'dashboard' example. E.g. one or two plots in one (rescalable?) div, and widgets/controls (sliders, buttons) in another one

bryevdv:
@mattpap should this be expected to work yet with "no extend" templates?

bryevdv:
OK here is a marginally working example extending the default template and using keen.io dash:

```
from bokeh.io import curdoc

p0 = figure(sizing_mode="scale_width", plot_height=300)
p0.circle(0, 0, radius=1, color="red")
p1 = figure(sizing_mode="scale_width")
p1.circle(0, 0, radius=1, color="blue")
p2 = figure(sizing_mode="scale_width")
p2.circle(0, 0, radius=1, color="green")
p3 = figure(sizing_mode="scale_width")
p3.circle(0, 0, radius=1, color="orange")
p4 = figure(sizing_mode="scale_width")
p4.circle(0, 0, radius=1, color="grey")

curdoc().template = """
{% block preamble %}
  <title>Keen Bokeh Dashboard</title>
  <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no' />

  <!-- Demo Dependencies -->
  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.3/jquery.min.js" type="text/javascript"></script>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.2.0/js/bootstrap.min.js" type="text/javascript"></script>
  <link href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.2.0/css/bootstrap.min.css" rel="stylesheet" type="text/css" />

  <!-- Dashboard -->
  <style>
    .keen-dashboard {
    background: #f2f2f2;
    font-family: 'Gotham Rounded SSm A', 'Gotham Rounded SSm B', 'Helvetica Neue', Helvetica, Arial, sans-serif;
    }

    .keen-dataviz {
    background: #fff;
    border: 1px solid #e7e7e7;
    border-radius: 2px;
    box-sizing: border-box;
    }
    .keen-dataviz-title {
    border-bottom: 1px solid #e7e7e7;
    border-radius: 2px 2px 0 0;
    font-size: 13px;
    padding: 2px 10px 0;
    text-transform: uppercase;
    }
    .keen-dataviz-stage {
    padding: 10px;
    }

    .keen-dataviz-notes {
    background: #fbfbfb;
    border-radius: 0 0 2px 2px;
    border-top: 1px solid #e7e7e7;
    font-family: 'Helvetica Neue', Helvetica, sans-serif;
    font-size: 11px;
    padding: 0 10px;
    }

    .keen-dataviz .keen-dataviz-metric {
    border-radius: 2px;
    }

    .keen-dataviz .keen-spinner-indicator {
    border-top-color: rgba(0, 187, 222, .4);
    }

    .keen-dashboard .chart-wrapper {
    background: #fff;
    border: 1px solid #e2e2e2;
    border-radius: 3px;
    margin-bottom: 10px;
    }
    .keen-dashboard .chart-wrapper .chart-title {
    border-bottom: 1px solid #d7d7d7;
    color: #666;
    font-size: 14px;
    font-weight: 200;
    padding: 7px 10px 4px;
    }

    .keen-dashboard .chart-wrapper .chart-stage {
    overflow: hidden;
    padding: 5px 10px;
    position: relative;
    }

    .keen-dashboard .chart-wrapper .chart-notes {
    background: #fbfbfb;
    border-top: 1px solid #e2e2e2;
    color: #808080;
    font-size: 12px;
    padding: 8px 10px 5px;
    }

    .keen-dashboard .chart-wrapper .keen-dataviz,
    .keen-dashboard .chart-wrapper .keen-dataviz-title,
    .keen-dashboard .chart-stage .chart-title {
    border: medium none;
    }

    .body {

    }
  </style>
{% endblock %}

{% block contents %}
  {{ embed(doc) }}

  <div class="container-fluid">
    <div class="row">

      <div class="col-sm-8">
        {{ embed(doc.roots[0]) }}
        <br>
      </div>

      <div class="col-sm-4">
        {{ embed(doc.roots[1]) }}
        <br>
      </div>

    </div>


    <div class="row">

      <div class="col-sm-4">
        {{ embed(doc.roots[2]) }}
        <br>
      </div>

      <div class="col-sm-4">
        {{ embed(doc.roots[3]) }}
        <br>
      </div>

      <div class="col-sm-4">
        {{ embed(doc.roots[4]) }}
        <br>
      </div>

    </div>
    <hr>
    <p class="small text-muted">Built with &#9829; by <a href="https://keen.io">Keen IO</a></p>
  </div>

{% endblock %}


"""

curdoc().add_root(p0)
curdoc().add_root(p1)
curdoc().add_root(p2)
curdoc().add_root(p3)
curdoc().add_root(p4)
```

The results are this:

screen shot 2018-05-23 at 14 03 00

several other sizing mode combinations do not work. In particular, I think `scale_both` is currently completely broken. It infinite expands on any resize. Without the `plot_height` on the one plot, it makes things too tall for the browser usually.

bryevdv:
Another issue currently, the template does not afford access to the `body` tag, and really to use the keen.io stuff, need to add a class to `body`

bryevdv:
Here is what resizing that example looks like:

screenflow

Seems fairly OK, pretty much exactly how the examples from the keen docs behave. But as mentioned various other combinations of sizing modes are not viable. We will need to develop some guidance around this.

bryevdv:
also just noting: had to inline keen dash CSS, can't find it on CDN

bryevdv:
With some widgets and `scale_width`

screen shot 2018-05-23 at 14 34 53
screen shot 2018-05-23 at 14 35 06

mattpap:
> Another issue currently, the template does not afford access to the body tag, and really to use the keen.io stuff, need to add a class to body

I added support for this. See `examples/plotting/file/custom_layout.py`.

mattpap:
> `{{ embed(doc) }}`

This isn't necessary anymore. `elementid` is now optional. If `elementid` not provided, or `embed(doc)` wasn't added to a template, then roots that weren't given explicit locations in a template, will be added to `<body>` and wrapped in `bk-root`.

mattpap:
> ```
> {% block preamble %}
>  <title>Keen Bokeh Dashboard</title>
>  ...
> ```

Preferably it should be `{% block title}Keen Bokeh Dashboard{% endblock %}`.

mattpap:
I still need to update autoload and notebook code.

bryevdv:
@mattpap do you have an updated ETA for this PR?

mattpap:
I'm tentatively marking this PR as ready. All features are implemented at this point and tests should be all green (though I had to disable a few for now; in the last commit). The main concern is making sure that what was supposed to work before, still works. Especially autoload, which works, but not all cases may be usable. I specifically didn't allow filling in missing members of a render item from a script tag, which didn't seem to make sense most of the time. However, I may be missing some rare use cases. It would be very useful to me get through those and point out what's broken, what should be restored on top of this PR.

bryevdv:
OK I will test locally with it now and try to merge this afternoon.

bryevdv:
@mattpap I built the docs and the first gallery example gives:

> Error: Error rendering Bokeh model: could not find tag with id: undefined

bryevdv:
The proximate issue is that items in `render_items` no longer have an `elementid` and so `_get_element` fails:

screen shot 2018-06-01 at 09 58 23

bokeh/embed/util.py
```diff
@@ -212,37 +288,36 @@ def standalone_docs_json_and_render_items(models):
...
-            modelid = p._id
-        docid = None
-        for key in docs_by_id:
-            if docs_by_id[key] == doc:
-                docid = key
-        if docid is None:
-            docid = make_id()
-            docs_by_id[docid] = doc
-
-        elementid = make_id()
```
bryevdv: presumably this is the reason why the `RenderItem` does not have an `elementid`. Is this intentional?

bryevdv:
@mattpap I have pushed a quick fix that gets `autoload_static` (and hence the docs) working. YOu should review it since you may prefer to leave the element id only in `RenderItem.roots` and extract it later?

bokeh/embed/util.py
mattpap: Yes. By default, `render_item.roots` tells where to put particular roots. `render_item.elementid` is meant for the rare situation where pre-mapping roots wasn't possible, e.g. when adding roots after creation and initial rendering of a document. Thus it should be used only when really necessary, e.g. with the server.

bryevdv:
OK that's frustrating, the `_get_element` was a red-herring. It seems a `node make clean` was needed to clear out the old `embed.js`. The server app issues are real though.

bokeh/embed/util.py
bryevdv: This was fixed by a `node make clean` see below

bryevdv:
@mattpap so there is no way to avoid a breaking change to app templates?

mattpap:
>  It seems a node make clean was needed to clear out the old embed.js.

This is the same problem as earlier in this PR. In fact I already devised a fix to our linker, but for now it interferes with actually valid patterns in node module resolution.

> The server app issues are real though.

I updated examples/app templates.

bryevdv:
> I updated examples/app templates.

See my question above. If we can't maintain compatibility this needs a migration note,  but ideally some kind of detection to provide a more specific error message that describes corrective action. I also think at least some of the examples need to be bare templates that don't extend `base` (I think you are underestimating how much people will want to avoid that and just have a "plain HTML" template)

mattpap:
>  If we can't maintain compatibility this needs a migration note, but ideally some kind of detection to provide a more specific error message that describes corrective action. 

For now I just want to make sure everything is updated and works with the new code.

> I also think at least some of the examples need to be bare templates that don't extend base (...)

Sure. In fact it would be best to have the same example in both versions to show precisely the difference between both approaches.

> (I think you are underestimating how much people will want to avoid that and just have a "plain HTML" template)

Possibly, but I also want to make sure people are directed to the new approach by default, in which is much harder to do big mistakes (e.g. forgetting the doctype, which is very common and results in nasty and subtle errors).

mattpap:
btw. I figured out how to improve module resolution when both `mod.js` and `mod/index.js` exist, so now you will get linker error, instead of the build continuing with the wrong module being picked.

bryevdv:
> For now I just want to make sure everything is updated and works with the new code.

I'd *much* prefer to make sure the PR is complete without breaking changes before merging, so that it's one unit of work together, I can't tell if you are suggesting otherwise

> Sure. In fact it would be best to have the same example in both versions to show precisely the difference between both approaches.

I agree but I'd probably leave that for the user guide where they can intentionally be presented side by side, it's been my observation that many people don't know about the examples

mattpap:
>  I can't tell if you are suggesting otherwise

I agree, I just want to make sure that all functionality is in place and works correctly, and all other code uses the new functionality before we add a compatibility layer. Otherwise, given the tooling we have, we ask ourselves for even more headache than this PR already brings.

mattpap:
TODO:

- [ ] fix `components()` when given a `dict`
- [ ] add full template example
- [ ] add compatibility layer for `plot_div`

bryevdv:
> I agree, I just want to make sure that all functionality is in place and works correctly,

OK, that's fine, I just wasn't sure what you intended. My testing the other day seemed to have everything working as expected except what I noted (and that you have since updated) I will run through things again today.

bryevdv:
@mattpap `examples/embed/server_session` does not seem to be working

mattpap:
> examples/embed/server_session does not seem to be working

It seems it has been always broken, or maybe it was broken during some restructuring (`git log --follow` doesn't give much output).

> examples/embed/arguments too, maybe just a general problem with server_session and server_document

`server_document()` is indeed broken. I will finish fixing this tomorrow.

bryevdv:
OK apps, embed, standalone, notebooks all generally seem to be working now. I will keep testing more but I would like to get this merged ASAP

mattpap:
Speaking about breakage, I'm still working on `components()`, to bring back it's previous output behavior. Though that will be ready tomorrow.

mattpap:
This works for me. It's likely a setup issue. In all honesty, I don't know what `setup.py develop` exactly does. I just use a `pth` file that links to bokeh's repository. This way there is no room for confusion between files installed and files in the repository (which I think get mixed up, because it seems like you run new code but have old templates, or some variation on this).

Karel-van-de-Plassche:
@mattpap You were right, thanks. I deleted all bokeh-related files from all directories on my filesystem, re-installed with `python setup.py develop` and it worked. I'll have a go at using this new template feature for my own code soon.

Karel-van-de-Plassche:
I'm trying to create an app with python callbacks with the new HTML templates. Although it is mentioned in the [README](https://github.com/bokeh/bokeh/tree/1d917cfd92f314bb07a5f9c35b610b0d70dc70eb/examples/plotting), there are no server examples in `bokeh/examples/plotting/`. I'm trying to manually place my plots in specific divs. Minimal example:

``` jinja
{% extends base %}
{% block preamble %}
<style>
.plots { display: flex; flex-direction: row; }
.p0, .p1{ width: 300px; padding: 50px; }
.p0 { background-color: red;   }   
.p1 { background-color: green; }
</style>
{% endblock %}
{% block contents %}
<div class="plots">
    <div class="p0">
        {{ embed(roots.p0) }}
    </div>
</div>
{% endblock %}
```
``` python
from bokeh.io import curdoc
from IPython import embed

# prepare some data
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]

# output to static HTML file
output_file("lines.html", title="line plot example")

p0 = figure(title="simple line example", x_axis_label='x', y_axis_label='y', name='p0')

p0.line(x, y, legend="Temp.", line_width=2)

# show the results
curdoc().add_root(p0)
```
error:
``` javascript
dom.js:19 Uncaught Error: Error rendering Bokeh model: could not find #5a6446f1-fd92-44b4-a08c-668775a3d2fb HTML tag
    at o (dom.js:19)
    at Object.n._resolve_element (dom.js:33)
    at index.js:38
    at index.js:25
```

Please also let me know if you think these tests by a newb are not necessary (yet)

Karel-van-de-Plassche:
In that case the README should be adjusted. All these examples use `{{ super() }}` to include the plots, with the aforementioned problem. Maybe I'm just misunderstanding how this manual placement works?

mattpap:
What's missing in the template is `{{ embed(doc) }}`. This is currently required to allow adding roots after document is initially rendered. I think I will lift this requirement, because neither this is convenient nor consistent with other styles of embedding, and adding roots after initial rendering is pretty rare.

Karel-van-de-Plassche:
How would we handle dynamic templates. For example, one could loop over all roots with:

``` jinja2
{% for plot in roots %}
```

But how would we handle different classes of models? I'd like to do something like:

``` jinja2
{% if roots.name.startswith('flux') %}
```

But I think roots just has the element-ids of the roots it contains.

Ideally there would be a way to collect related models in some kind of ModelList, without any formatting implied. Something like row/column without the formatting, so one can do:

``` python
fig0 = Figure(name='flux0')
fig1 = Figure(name='flux1')
fig2 = Figure(name='freq0')
fluxfigs = modelList(fig0, fig1, name='fluxfigs')
curdoc().add_root(fluxfigs)
curdoc().add_root(fig2)
```
``` jinja2
{% for plot in roots.fluxfigs %}
  {# Some stuff #}
  {{ embed(plot) }}
{% endfor %}
 {# Some other stuff #}
 {{ embed(fig2) }}
```

Maybe this is already implemented, but I'm not sure how to do this.

NB. How to easily debug jinja-related errors in bokeh? I now tried putting `{{ variable | pprint }}` in the template, is this the way to go?

mattpap:
@Karel-van-de-Plassche, see the example in the most recent commit.

bryevdv:
@mattpap  is this ready-ready?

mattpap:
@bryevdv, yes.

bryevdv:
ok I am going to go ahead and merge and get a dev build out soon then

bryevdv:
@chesucr @mrocklin @Karel-van-de-Plassche @crashMOGWAI @jlstevens @philippjfr This has been merged and will be in a dev build later today. Please try things out if you have a chance. More docs are coming in a few days but there are some examples in this PR.

bokeh/core/_templates/autoload_tag.html
```diff
@@ -21,9 +21,4 @@
     This script injects a ``<div>`` in place, so must be placed under ``<body>``.
 
 #}
-<script
-    src="{{ src_path }}"
-    id="{{ elementid }}"
-    data-bokeh-model-id="{{ modelid }}"
```
complyue: This breaks the API specification [server_session](http://bokeh.pydata.org/en/latest/docs/reference/embed.html#bokeh.embed.server_session), breaks my code that spread a few figures/tables to respective divs inside my html template. Now every such div contains all figures from the document!

bryevdv: @complyue please join the discussion at https://github.com/bokeh/bokeh/issues/8499 to have a voice in getting this kind of functionality added back in a supported, maintainable way.

PR #7730, opened 2018-04-03
temp delete custom repr so that ipython pprint does not use it

### LINKED ISSUE #7729: Update of ipython to 6.3.0 breaks unit tests
CI unit test build suddenly failing after recent update to ipython 6.3:
```
______________________________ test_Model_pretty _______________________________
    def test_Model_pretty():
>       class Foo1(Model):
bokeh/tests/test_model.py:8: 
meta_cls = <class 'bokeh.model.MetaModel'>, class_name = 'Foo1'
bases = (<class 'bokeh.model.Model'>,)
class_dict = {'__container_props__': set(), '__module__': 'bokeh.tests.test_model', '__properties__': set(), '__properties_with_refs__': set(), ...}
        '''
    
            Raises:
                Warning
    
            '''
    
        # use an explicitly provided view model name if there is one
        if "__view_model__" not in class_dict:
            class_dict["__view_model__"] = class_name
    
        # call the parent metaclass to create the new model type
        newcls = super(MetaModel, meta_cls).__new__(meta_cls, class_name, bases, class_dict)
    
        # update the mapping of view model names to classes, checking for any duplicates
        # and handling any subtype relationships or custom implementations
        entry = class_dict.get("__subtype__", class_dict["__view_model__"])
        if entry in MetaModel.model_class_reverse_map and not hasattr(newcls, "__implementation__"):
            raise Warning("Duplicate __view_model__ or __subtype__ declaration of '%s' for " \
                          "class %s.  Previous definition: %s" % \
                          (entry, class_name,
>                          MetaModel.model_class_reverse_map[entry]))
E           Warning: Duplicate __view_model__ or __subtype__ declaration of 'Foo1' for class Foo1.  Previous definition: <class 'bokeh.tests.test_model.test_Model_pretty.<locals>.Foo1'>

```
confirmed locally that this package difference triggers the failure.

One option is to pin the version but would like to identify the root cause as this connection is surprising and unexpected.

bryevdv:
- [x] issues: fixes #7729
- [x] tests added / passed

Can't say I am happy with this but don't have any better solution.

### DIFF
--- a/bokeh/core/has_props.py
+++ b/bokeh/core/has_props.py
@@ -273,7 +273,9 @@ class HasProps(with_metaclass(MetaHasProps, object)):
     def __str__(self):
         return "%s(...)" % self.__class__.__name__
 
-    __repr__ = __str__
+    # we need to define the __repr__ this way because of https://github.com/bokeh/bokeh/issues/7729
+    def __repr__(self):
+        return self.__str__()
 
     def equals(self, other):
         ''' Structural equality of models.
@@ -661,7 +663,14 @@ class HasProps(with_metaclass(MetaHasProps, object)):
         else:
             stream = StringIO()
             printer = _BokehPrettyPrinter(stream, verbose, max_width, newline)
+
+            # temporarily deleting the old repr to make ipython not use it, see:
+            # https://github.com/ipython/ipython/issues/11066
+            old_repr = self.__repr__
+            self.__repr__ = None
             printer.pretty(self)
+            self.__repr__ = old_repr
+
             printer.flush()
             return stream.getvalue()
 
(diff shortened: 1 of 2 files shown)


bryevdv:
As an aside, fairly sure if any model defines `__repr__` on itself, it will break pretty printing for all its descendents.

bokeh/core/has_props.py
```diff
@@ -661,7 +663,14 @@ def __init__(self, output, verbose=False, max_width=79, newline='\n'):
...
             stream = StringIO()
             printer = _BokehPrettyPrinter(stream, verbose, max_width, newline)
+
+            # temporarily deleting the old repr to make ipython not use it, see:
+            # https://github.com/ipython/ipython/issues/11066
+            old_repr = self.__repr__
+            self.__repr__ = None
             printer.pretty(self)
+            self.__repr__ = old_repr
+
```
mattpap: `try-finally` would be in order here.

mattpap:
This is a fine workaround.

PR #7735, opened 2018-04-03
don't use --rerun

### LINKED ISSUE #7734: remove pytest-rerunfailures from CI tests
Tests are more stable now, and removing will allow deletion of the package from the bokeh channel as well.

bryevdv:
- [x] issues: fixes #7734
- [x] tests added / passed

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -68,7 +68,6 @@ test:
     - pytest-html
     - selenium >=3
     - pytest-xdist
-    - pytest-rerunfailures
     - requests >=1.2.3
     - beautifulsoup4
     - scipy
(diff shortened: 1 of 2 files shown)


bryevdv:
@mattpap the image export tests still seem to fail with some regularity on travis with a "Bad file descriptor" message that re-running seems to fix.

mattpap:
Well, lets investigate. I can do it later.

bryevdv:
FWIW seems to be the py2.7 job, at least that's the one I've seen fail twice in a couple of days now

bryevdv:
I'd be ok to merge this and investigate later if it keeps happening enough. I haven't seen it in last several jobs so maybe it's not as common as I imagined. If it happens again we can capture the log output and open a new issue. Sound OK?

mattpap:
Yes, sounds good.

mattpap:
Also, I'm pretty sure we will have to increase timeouts in JS tests (I've been noticing this sporadically locally, so it should happen eventually on travis ci).

PR #7743, opened 2018-04-04
add option to disable wheel zoom on axes

### LINKED ISSUE #2427: Make axis wheel zoom configurable
This would be nice for figures where aspect is important, but zooming is also desired. Could either turn off the individual axis zoom interactions (so only scrolling in the figure will zoom), or "link" the zooming on the axis so that zooming on x moves the y axis as well.

### LINKED ISSUE #3125: Expose the wheel zoom speed in Python

bryevdv:
- [x] issues: fixes #2427 fixes #3125

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -229,6 +229,10 @@ class WheelZoomTool(Scroll):
     zoom that dimension.
     """)
 
+    speed = Float(default=1/600, help="""
+    Speed at which the wheel zooms. Default is 1/600. Optimal range is between
+    0.001 and 0.09. High values will be clipped. Speed may very between browsers.
+    """)
 
 class SaveTool(Action):
     ''' *toolbar icon*: |save_icon|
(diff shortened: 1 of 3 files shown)


bryevdv:
Also added wheel zoom speed option

chilly-and-momo:
Was this implemented already? I'm trying to limit the wheel zoom to just the plot itself, and would like to disable the x and y axis zoom on the axis or canvas area outside the plot. I've looked through the current documentation on the bokeh user and reference guides, but haven't found an obvious solution.

bryevdv:
@chilly-and-momo 

https://bokeh.pydata.org/en/latest/docs/reference/models/tools.html#bokeh.models.tools.WheelZoomTool.zoom_on_axis

Also, if you click on "Files Changed" above you can see where this property was added.

chilly-and-momo:
Thank you so much!

PR #7747, opened 2018-04-04
fix error with link highlighting and external links

### LINKED ISSUE #7744: Docs - JSON Prototype does not expand
See, for example, https://bokeh.pydata.org/en/latest/docs/reference/models/sources.html#bokeh.models.sources.ColumnDataSource

Clicking on JSONPrototype it does not expand.

Tested on chrome and firefox on linux.

bryevdv:
- [x] issues: fixes #7744


bryevdv:
CI failing with `OSError` file handle issue, but master is also already failing with it so I will merge this now.

PR #7748, opened 2018-04-04
make ClientSession a context manager

### LINKED ISSUE #7724: Plots not reacting to sliders inside flask.
Hello,

Using Bokeh 0.12.15 with flask, via "bokeh serve". My bokeh dashboard features sliders to dynamically update the plots. When I visit the site hosted by flask, the plot responds to the sliders for the few second or so, and then stops reacting, until I refresh the page. In contrast, if I directly visit the page hosted by the Bokeh server (i.e. localhost:5100/my_bokeh_page), then no problems arise; the sliders are fully functional throughout the entire session.

This problem seems to have arisen sometime after Bokeh 0.12.13. Reverting to 0.12.13 fixes the problem entirely. (Haven't tried 0.12.14)

bryevdv:
- [x] issues: fixes #7724
- [x] tests added / passed

### DIFF
--- a/bokeh/embed/server.py
+++ b/bokeh/embed/server.py
@@ -107,7 +107,7 @@ def server_document(url="default", relative_urls=False, resources="default", arg
     return encode_utf8(tag)
 
 @general((1,0,0))
-def server_session(model, session_id, url="default", relative_urls=False, resources="default", arguments=None):
+def server_session(model=None, session_id=None, url="default", relative_urls=False, resources="default", arguments=None):
     ''' Return a script tag that embeds content from a specific existing session on
     a Bokeh server.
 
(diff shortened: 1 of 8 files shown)


bokeh/embed/server.py
```diff
@@ -170,6 +170,9 @@ def server_session(model, session_id, url="default", relative_urls=False, resour
             probably not desired.
 
     '''
+    if session_id is None:
+        raise ValueError("Must supply a session_id")
+
```
bryevdv: This would be great with py3 keyword only args, some day...

Also I'd like to make `server_session(session=mysession)` possible, it will complicate arg handling but I think it is the natural API we missed along the way. Ideally it would have been `server_session(mysession)` but that ship has sailed

PR #7753, opened 2018-04-04
Revert "don't use --rerun"

bryevdv:
Reverts bokeh/bokeh#7735

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -68,6 +68,7 @@ test:
     - pytest-html
     - selenium >=3
     - pytest-xdist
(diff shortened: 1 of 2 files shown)


mattpap:
What happened to this commit history?

bryevdv:
> What happened to this commit history? 

Looks like what's been merged since the original PR, this is from the "Revert" button on the GH web UI.  Not sure it actually matters, tho. I'm planning to merge this when green, unless you have other immediate ideas.

mattpap:
> (...) unless you have other immediate ideas.

See PR #7754.

PR #7757, opened 2018-04-05
add reset signal to plots

### LINKED ISSUE #5071: Add a method to "reset" a figure
I want the ability to reset a figure via code in the same way as the figure reset tool.

I asked a related question on Stack Overflow, and it was recommended that I open an issue to track the request.

http://stackoverflow.com/questions/39278110/whats-the-command-to-reset-a-bokeh-plot

bryevdv:
- [x] issues: fixes #5071

example:

```py
from bokeh.io import show
from bokeh.models import Button, CustomJS

p = figure(tools="reset,pan,wheel_zoom,lasso_select")
p.circle(list(range(10)), list(range(10)))

b = Button()
b.js_on_click(CustomJS(args=dict(p=p), code="""
    p.reset.emit()
"""))

show(column(p, b))
```

### DIFF
--- a/bokehjs/src/coffee/models/plots/plot.ts
+++ b/bokehjs/src/coffee/models/plots/plot.ts
@@ -2,6 +2,7 @@ import {EQ, Constraint, Variable} from "core/layout/solver"
 import {logger} from "core/logging"
 import * as visuals from "core/visuals"
 import * as p from "core/properties"
+import {Signal0} from "core/signaling"
 import {Color} from "core/types"
 import {LineJoin, LineCap} from "core/enums"
 import {Place, Location, OutputBackend} from "core/enums"
@@ -159,6 +160,8 @@ export namespace Plot {
 export interface Plot extends Plot.Attrs {}
 
 export class Plot extends LayoutDOM {
+
+  reset = new Signal0(this, "reset")
 
   properties: Plot.Props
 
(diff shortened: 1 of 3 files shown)


bokehjs/src/coffee/models/plots/plot.ts
```diff
@@ -160,6 +161,8 @@ export interface Plot extends Plot.Attrs {}
 
 export class Plot extends LayoutDOM {
 
+  reset = new Signal0(this, "reset")
```
mattpap: Only declaration can be here. Definition must be moved to `initialize()`. This is because `initialize()` doesn't work particularly well with this style of member initialization. This will work by default though (in deferred mode), but not when used directly in bokehjs.

mattpap: Sure, I will finish this. I wasn't consistent myself with this. I will need to review how classes are initialized to make sure that there is one, correct way.

bryevdv:
I fixed this up, the issue was earlier I had added to `constructor`, not to `initialize`

mattpap:
> I fixed this up, the issue was earlier I had added to constructor, not to initialize

Adding to constructor (after super, but TS won't let you add before) is equivalent to your original approach.

bryevdv:
> Adding to constructor (after super, but TS won't let you add before) is equivalent to your original approach. 


Weird, not sure why it didn't seem to be working then. In any case I think this addresses your comment, so I will go ahead and merge

PR #7765, opened 2018-04-06
added 'pan' to the hexbin example

afonit:
When I was playing with this example I found myself wanting to pan after zooming in.
I thought this might be a sane default to add.  Granted others could add that in later but I thought it would be nice to have right off the bat.

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Seems reasonable,thanks @afonit !

PR #7769, opened 2018-04-07
factor and improve compute_renderers for tools

### LINKED ISSUE #3442: `BoxSelectTool` should accept renderers = None
currently defaults to `[]` which is expanded all available renderers. Need a way to specify both "all" and "none".

bryevdv:
- [x] issues: fixes #3442
- [x] tests added / passed


bryevdv:
I'd like to make it so hover tools can display some info regardless of whether any glyphs were hit, so this leads into that.

bryevdv:
I will do some more hand testing with various hover and selection examples but would definitely comments.

bokeh/models/tools.py
```diff
@@ -704,8 +705,8 @@ class HoverTool(Inspection):
     have a matching value for their ``name`` attribute will be used.
     """)
 
-    renderers = List(Instance(Renderer), help="""
-    An explicit list of renderers to hit test again. If unset,
+    renderers = Either(Auto, List(Instance(Renderer)), help="""
```
clairetang6: The default value of "auto" isn't set here, or the help text is wrong if it's meant to default to no renderers.

sphinx/source/docs/releases/0.12.16.rst
```diff
@@ -35,4 +35,14 @@ report the current column names automatically. Setting ``column_names`` is
 no longer supported, however as this possibility has never been officially
 documented or demonstrated, it is not expected to affect any standard usage.
 
+Tools ``renderers`` property
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Previously a the default value, an empty list, corresponded to automatically
```
clairetang6: Typo at "Previously a the"

sphinx/source/docs/releases/0.12.16.rst
```diff
@@ -35,4 +35,14 @@ report the current column names automatically. Setting ``column_names`` is
...
+choosing all available renderers for hit testing. Now, a new default value
+``"auto"`` is used for that, so that an empty list can mean "hit test no
+renderers". A value of ``None`` may also be used to indicate no hit testing
+should occur. Since there is typically no reason to explicity set a property
+to its default value, it is not expected that his change impacts any standard
```
clairetang6: typo at "his change"

bokehjs/src/coffee/models/tools/util.ts
```diff
@@ -0,0 +1,33 @@
...
+export type RendererSpec = DataRenderer[] | "auto" | null
+
+export function compute_renderers(renderers: RendererSpec, all_renderers: Renderer[], names: string[]): DataRenderer[] {
+    if (renderers == null) {
+      return []
+    }
+
+    if (renderers.length == 0) {
+      return []
+    }
```
mattpap: Can't we just have one value for "empty" behavior (drop `null`)?

bokehjs/src/coffee/models/tools/util.ts
```diff
@@ -0,0 +1,33 @@
...
+      result = all_renderers.filter((r): r is DataRenderer => {
+        return r instanceof GlyphRenderer || r instanceof GraphRenderer
+      })
+    }
+
+    if (names.length > 0)
+      result = result.filter((r) => includes(names, r.name))
+
+    return result
+
```
mattpap: Unnecessary whitespace.

bokehjs/src/coffee/models/tools/util.ts
```diff
@@ -0,0 +1,33 @@
...
+export function compute_renderers(renderers: RendererSpec, all_renderers: Renderer[], names: string[]): DataRenderer[] {
+    if (renderers == null) {
+      return []
+    }
+
+    if (renderers.length == 0) {
+      return []
+    }
+
+    let result = renderers as DataRenderer[]
```
mattpap: You can't just assume `renderers` are `DataRenderer`s. The original code correctly filtered `all_renderers` to assert that.

mattpap: In fact the problem starts with`renderers = Either(Auto, List(Instance(Renderer)))` and not using `DataRenderer` (doesn't exist at this point, but I have intended for a while to add one to clarify things).

bokehjs/src/coffee/models/tools/util.ts
bryevdv: Do we have a technical way to validate against `None` values? If not I'd rather have it do something defined and documented (which is the only reason I added a path for it)

bokehjs/src/coffee/models/tools/util.ts
bryevdv: The code immediately below does the filtering (and is identical to the previous code). If you have a concrete suggestion, I am happy to implement it.

bryevdv:
@clairetang6 thanks I have made fixes for your comments

@mattpap let me know what you want to do for `result`

bokehjs/src/coffee/models/tools/util.ts
mattpap: Ah yes, I see it now. This is even worse that I initially viewed it. This is how the code should look like:
```
  if (renderers == null)
      return []

  let result: DataRenderer[]
  if (renderers == 'auto') {
    result = all_renderers.filter((r): r is DataRenderer => {
      return r instanceof GlyphRenderer || r instanceof GraphRenderer
    })
  } else
    result = renderers

  if (names.length > 0)
    result = result.filter((r) => includes(names, r.name))

  return result
```

mattpap: Also, with the original code:
```ts
   if (renderers.length == 0) {
      return []
    }
```
without proper type inspection, you are effectively checking length of a string when `"auto"`. This isn't incorrect, but definitively strange. TypeScript doesn't help here, because `lenght` is a valid member on string, even if it's a literal string. However, as with tuples, this is likely to get improved and length with be fixed (to 4 in this case), so this code may start failing compilation at some point.

bryevdv: I've updated to start `result` as an empty list and push appropriate things to it.

bryevdv:
fixed one last typo in a default value, I guess it surprises me that this did not trigger an error or test failure of some sort

PR #7776, opened 2018-04-10
rip out API decorators and spec tests

### LINKED ISSUE #7774: Make API decorators more debugging friendly
Currently we have the following code in `core.util.api`:
https://github.com/bokeh/bokeh/blob/87a2e9a93b8e5d87ddef7de6a08b50810301512f/bokeh/util/api.py#L182-L194
This makes every API call go through `wrapper`, which is incredibly annoying during debugging. I'm not sure why wrapping is needed for functions and methods. I removed the wrapper and used the same code for classes and functions/methods, and there are no test failures. Is this a Py2/3 thing or else?

bryevdv:
- [x] issues: fixes #7774
- [x] tests added / passed

@mattpap @p-himik unless there is objection I'd like to merge this now and we can figure out the replacement convention separately.


bryevdv:
I did keep the checks on `__all__` for top level packages. These are much more conceptually simple and explicit.

mattpap:
This is a bit radical approach to fix #7774. I would leave decorators, even if they are phony, because this give us classification into dev and user APIs (?).

bryevdv:
> This is a bit radical approach to fix #7774. I would leave decorators, even if they are phony, because this give us classification into dev and user APIs 

The reason I decided to do this is that the dev/general classification intersects pretty poorly with classes and inheritance. Does it make sense for dev classes to have general methods, etc. It's also not possible without the wrapper for py2 which makes debugging crappy, and although I haven't actually tested I worried about the import overhead of wrapping every single function and method in the library.

mattpap:
OK, sounds good.

PR #7782, opened 2018-04-11
Add docstring to get_screenshot_as_png

### LINKED ISSUE #7781: Missing docstring for get_screenshot_as_png

*ISSUES THAT DO NOT CONTAIN NECESSARY INFORMATION MAY BE CLOSED, IMMEDIATELY*

The issue tracker is NOT the place for general support. For questions and
technical assistance, come ask the [Bokeh mailing list](https://groups.google.com/a/continuum.io/forum/#!forum/bokeh) or join the chat on [Gitter](https://gitter.im/bokeh/bokeh). For feature requests, please provide a detailed description or proposal of the new capability or behavior.

bokeh 0.12.16dev

Missing docstring for `bokeh.io.exports` function `get_screenshot_as_png`.

```
# code goes here between backticks

```

Seth-Rothschild:
This PR adds a docstring to `get_screenshot_as_png` in `bokeh.io.exports` as mentioned [on SO](https://stackoverflow.com/questions/49744664/how-to-generate-static-bokeh-plots-in-jupyter-notebook/49755653#49755653). Since it's the internal function to `export_png`, I treated it as though the docstring would be similar. Happy to modify it or any of the associated github text to suit project preferences.

- [ ] issues: fixes #7781

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -149,8 +149,28 @@ def export_svgs(obj, filename=None, height=None, width=None, webdriver=None):
 # Dev API
 #-----------------------------------------------------------------------------
 
+
 def get_screenshot_as_png(obj, driver=None, **kwargs):
-    '''
+    ''' Get a screenshot of a LayoutDOM object.
+
+    Args:
+        obj (LayoutDOM or Document) : a Layout (Row/Column), Plot or Widget
+            object or Document to export.
+
+        driver (selenium.webdriver) : a selenium webdriver instance to use
+            to export the image.
+
+    Returns:
+        cropped_image (PIL.Image.Image) : a pillow image loaded from PNG.
+
+    .. warning::
+        Responsive sizing_modes may generate layouts with unexpected size and
+        aspect ratios. It is recommended to use the default ``fixed`` sizing mode.
+
+    .. warning::
+        Glyphs that are rendered via webgl won't be included in the generated PNG.
+
+
 
     '''
     Image = import_required('PIL.Image',


bokeh/io/export.py
```diff
@@ -149,8 +149,28 @@ def export_svgs(obj, filename=None, height=None, width=None, webdriver=None):
...
+
+    Returns:
+        cropped_image (PIL.Image.Image) : a pillow image loaded from PNG.
+
+    .. warning::
+        Responsive sizing_modes may generate layouts with unexpected size and
+        aspect ratios. It is recommended to use the default ``fixed`` sizing mode.
+
+    .. warning::
+        Glyphs that are rendered via webgl won't be included in the generated PNG.
```
mattpap: This depends if the driver supports webgl. This will definitively work with chromedriver.

Seth-Rothschild: Sure. This section is an exact copy of the docstring from `export_png`, so the change to chromedriver will also affect that. Feel free to use this PR to add the changes to the other docstring, or to copy the docstring found here into a different commit when the internals are updated.

bryevdv: Chrome is already an optional option that can be used. Could you update both doc strings? Basically to the effect that webgl may not render with some webdrivers such as phantom JS.

mattpap: On a second thought, this warning is unnecessary. bokehjs falls back to canvas renderer when webgl isn't available or fails for some reason, so all glyphs will be always present in the output.

PR #7787, opened 2018-04-11
Rename bokehjs/src/{coffee->lib}

### LINKED ISSUE #7786: Rename bokehjs/src/coffee
Given that we don't use CoffeeScript anymore, this should be finally renamed. This change must not affect existing PRs.

mattpap:
Do **NOT** squash merge.

fixes #7786


bryevdv:
This should get a mention in the release notes as well

PR #7794, opened 2018-04-14
Make handling of UI events more robust

### LINKED ISSUE #7627: Selected points get lost with Tap Tool
When I select some points with some selection tools and you change the tool you are working with, the current selection should be kept. But when I use Tap tool and I change, enable or disable some other tool the selection is lost. Is this an issue or a expected behaviour?

```
from bokeh.models.callbacks import CustomJS
from bokeh.io import curdoc

plot = figure(
    width=600,
    height=600,
    tools='lasso_select,pan,box_select,tap'
)

x = list(range(-20, 21))
y = [xx**2 for xx in x]
source = ColumnDataSource(data=dict(x=x,y=y))

cr = plot.circle(
    x='x', y='y',
    source=source, size=10, color="navy", alpha=0.5
)

curdoc().add_root(column([plot]))
```

This could be related with this [other issue](https://github.com/bokeh/bokeh/issues/2328)

mattpap:
So far this fixes `z-index` mess and doesn't allow propagation of events from toolbar to plot area. This makes, however, the appearance of special overlays dependent on the order of insertion of tools.

fixes #7627

### DIFF
--- a/bokehjs/src/less/main.less
+++ b/bokehjs/src/less/main.less
@@ -2,5 +2,4 @@
   position: absolute;
   display: block;
   border: 1px dashed green;
-  z-index: 100;
 }
--- a/bokehjs/src/lib/core/ui_events.ts
+++ b/bokehjs/src/lib/core/ui_events.ts
@@ -227,8 +227,13 @@ export class UIEvents {
     return this.plot_view.frame.bbox.contains(sx, sy)
   }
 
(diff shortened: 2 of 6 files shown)


mattpap:
The order of overlays is still an issue, however, this was always a problem and ordering of tools allows to mitigate this. I still would like to clarify this situation, though in another PR.

PR #7799, opened 2018-04-16
Make defaults match between bokeh and bokehjs

### LINKED ISSUE #7750: Investigate output from test_defaults
JS tests give this output:
```
Band.source: defaults are out of sync between Python and bokehjs
    Band.source.selected: bokehjs defaults to {"type":"Selection"} but python defaults to null
    Band.source.selection_policy: bokehjs defaults to {"type":"UnionRenderers"} but python defaults to null
LabelSet.source: defaults are out of sync between Python and bokehjs
    LabelSet.source.selected: bokehjs defaults to {"type":"Selection"} but python defaults to null
    LabelSet.source.selection_policy: bokehjs defaults to {"type":"UnionRenderers"} but python defaults to null
[bokeh] LogTickFormatter not configured with a ticker, using default base of 10 (labels will be incorrect if ticker base is not 10)
MercatorAxis.ticker: defaults are out of sync between Python and bokehjs
    MercatorAxis.ticker.dimension: bokehjs defaults to null but python defaults to "lat"
MercatorAxis.formatter: defaults are out of sync between Python and bokehjs
    MercatorAxis.formatter.dimension: bokehjs defaults to null but python defaults to "lat"
Whisker.source: defaults are out of sync between Python and bokehjs
    Whisker.source.selected: bokehjs defaults to {"type":"Selection"} but python defaults to null
    Whisker.source.selection_policy: bokehjs defaults to {"type":"UnionRenderers"} but python defaults to null
Python/bokehjs matching defaults problems: 0
```

### LINKED ISSUE #7803: py.test -m unit doesn't run pytest.mark.parametrize tests
Parametrized tests apparently aren't unit tests by default. This is because parametrization is done using `mark` API, and `unit` marker is applied automatically only to unmarked test cases.

mattpap:
fixes #7750
fixes #7803


bokeh/core/property/bases.py
```diff
@@ -61,9 +61,6 @@ def __init__(self, default=None, help=None, serialized=True, readonly=False):
         self.alternatives = []
         self.assertions = []
 
-        # "fail early" when a default is invalid
-        self.validate(self._raw_default())
```
mattpap: This is unfortunate, but it's the more reliable way to get around Python's recursive imports. On the bright side, removing this makes `from bokeh.models import *` 20% faster. To make sure all defaults are of the correct type, a test can be added.

bryevdv: I do agree that this kind of testing should happen ex-runtime, so that user's don't have to pay for it.

mattpap:
@bryevdv, can you comment on the failing tests? E.g:
```
            # fake self not great but much easier than setting up a real view
>           ret = WSHandler.send_message("self", ExcMessage())
E           TypeError: unbound method send_message() must be called with WSHandler instance as first argument (got str instance instead)
```
Python 2 doesn't like this hack.

bryevdv:
@mattpap Well, that sucks. I think you'd have to create an actual `WSHandler` instance and invoke the method on that in py2, but that's kind of a pain to set up. I would say just skipping the test on py2 would be ok with me.

mattpap:
OK, I will proceed this way.

mattpap:
Looks like another case of "bad file descriptor", just completely unrelated to previous ones.

bryevdv:
Never seen before, restarted in case it was spurious. Maybe we should raise an issue with Travis support

bryevdv:
Still failing, I will try restarting an old successful job and see if that also now fails.

bryevdv:
Still failing same place, something about this PR specifically?

mattpap:
I'm not sure right now. I will look closer at this failing code.

bryevdv:
@mattpap I only just realized this is happening in the exception codepath or `run_bokeh_server`. This is probably just some race condition, if the process is already closed that explains the bad FD on `stdout`. I would propose this: always print the actual exception in that codepath, and try to print `stdout`, but don't complain if it doesnt work.

bryevdv:
Went ahead and pushed that idea to see what happens, if it still fails at least now we should see the exception

mattpap:
In fact documentation of Popen states that code written in this style (PIPE + wait) may deadlock.

bryevdv:
Well I am going to assume that is ok then merge/cut a dev build later tonight

PR #7800, opened 2018-04-16
Add relative path option and handling to sphinx ext.

### LINKED ISSUE #5955: sphinxext.bokeh_plot generates script but doesn't render plot locally
When using the sphinxext.bokeh_plot directive with the example from the [docs](http://bokeh.pydata.org/en/latest/docs/reference/sphinxext.html#module-bokeh.sphinxext.bokeh_plot), the sphinx build executes successfully (after following #5951). The js script is generated and copied into `SPHINX_DOCS_ROOT_PATH/build/html/scripts/bokeh-plot-index-inline-5.js`. 

However when the file `SPHINX_DOCS_ROOT_PATH/build/html/index.html` is opened locally in chrome, the plot doesn't render (although the code-block does), and the console gets the following error: 
`Failed to load resource: net::ERR_FILE_NOT_FOUND` with a link to `file:///scripts/bokeh-plot-index-inline-5.js`.

When I edited line 122 of `PYTHON_LIB_PATH/bokeh/sphinxext/bokeh_plot.py` from `script_path = join("/scripts", js_name)` to `script_path = join("scripts", js_name)`, the plot appeared. I don't know a lot about web development, so I don't know if removing the `/` would break documentation hosted on a server.

Version Info:

Python version      :  3.6.0 | packaged by conda-forge | (default, Jan 14 2017, 03:13:00) 
IPython version     :  5.1.0
Bokeh version       :  0.12.4
BokehJS static path :  /Users/Atom/.anaconda/envs/py3/lib/python3.6/site-packages/bokeh/server/static
node.js version     :  v6.3.1
npm version         :  3.10.3
jupyter --version     :  4.2.1
jupyter notebook --version    : 4.3.1
os version : macOS 10.12.3
browser verision: Google Chrome Version 56.0.2924.87 (64-bit)
sphinx version: 1.5.3

This is a really awesome sphinx extension!

hallyourbase:
Add 'bokeh_plot_use_relative_paths' sphinx config option which places
relative path strings in the generated html to render bokeh plots served
statically.

Default is false which leaves the paths unchanged as they currently are.

-  partially addresses #5955

### DIFF
--- a/bokeh/sphinxext/bokeh_plot.py
+++ b/bokeh/sphinxext/bokeh_plot.py
@@ -123,7 +123,7 @@ class PlotScriptError(SphinxError):
 
     category = 'PlotScript error'
 
-def _process_script(source, filename, auxdir, js_name):
+def _process_script(source, filename, auxdir, js_name, use_relative_paths=False):
     # This is lame, but seems to be required for python 2
     source = CODING.sub("", source)
 
@@ -139,7 +139,12 @@ def _process_script(source, filename, auxdir, js_name):
     if c.error:
         raise PlotScriptError(c.error_detail)
 
-    script_path = join("/scripts", js_name)
+    script_path = None
+    if(use_relative_paths):
+        script_path = join("$REL_PATH$", js_name)
+    else:
+        script_path = join("/scripts", js_name)
+
     js_path = join(auxdir, js_name)
     js, script = autoload_static(d.roots[0], resources, script_path)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
I will try to look at this this week.

bryevdv:
@hallyourbase So I did go ahead and confirm that the project docs build as expected, so that is good. I guess I'd be OK merging this as-is, tho I do think I would like to revisit this more broadly at some point, I just don't have any idea when that might be. But just please be aware that there is some chance that there may be breaking changes if/when I do (i.e. this parameter may go away it's no longer needed)

bryevdv:
Also @mattpap interested to merge this working PR and see if the bad description build fail shows up on master...

bryevdv:
@hallyourbase definitely, if/when I do look at making all the file locations more flexible I will also try to ping you on here so that you can help with testing during development, if you have the bandwidth.

bryevdv:
So also, aspirationally, I am going to leave the original issue open as well

PR #7804, opened 2018-04-17
Handle cases where TouchEvent.touches may be empty

### LINKED ISSUE #7801: Bokeh 0.12.15 throws TypeError on Touch Devices when dragging
* python: 2.7.14
* Bokeh 0.12.15
* Browser: Mobile Safari 11.3/11.4 on iPad or iPhone

We're rendering a basic line plot with bokeh. This worked fine in 0.12.14 and before. After upgrading to 0.12.15, it starts throwing errors when I touch the graph:

```TypeError: undefined is not an object (evaluating 'e.pageX')```

The code responsible was introduced in 675aacf72a:
```js
  protected _get_sxy(event: TouchEvent | MouseEvent | PointerEvent): {sx: number, sy: number} {
    // XXX: jsdom doesn't support TouchEvent constructor
    function is_touch(event: TouchEvent | MouseEvent | PointerEvent): event is TouchEvent {
      return typeof TouchEvent !== "undefined" && event instanceof TouchEvent
    }
    const {pageX, pageY} = is_touch(event) ? event.touches[0] : event // XXX: assuming at last one touch
...
```

@dwt debugged this down to the event being a `TouchEvent`, but its `touches` array has length 0. A safer way to access the page offsets can be seen in [Waves.js](https://github.com/fians/Waves/blob/master/src/js/waves.js#L129)

mattpap:
This needs to be verified on an actual mobile device.

fixes #7801

### DIFF
--- a/bokehjs/src/lib/core/ui_events.ts
+++ b/bokehjs/src/lib/core/ui_events.ts
@@ -319,7 +319,7 @@ export class UIEvents {
     function is_touch(event: TouchEvent | MouseEvent | PointerEvent): event is TouchEvent {
       return typeof TouchEvent !== "undefined" && event instanceof TouchEvent
     }
-    const {pageX, pageY} = is_touch(event) ? event.touches[0] : event // XXX: assuming at last one touch
+    const {pageX, pageY} = is_touch(event) ? (event.touches.length != 0 ? event.touches : event.changedTouches)[0] : event
     const {left, top} = offset(this.hit_area)
     return {
       sx: pageX - left,


bryevdv:
I'd say just merge it, and get it into  dev build, then we can test it (and have the OP test it) on the dev docs site.

mattpap:
OK, sounds like a plan. I know that it's possible to set up remote debugging with chrome's devtools with a physical mobile device over USB, I just need to figure out details to make it work in a VM.

dwt:
Can't reproduce the issue with this change applied.

bryevdv:
Thanks for confirming @dwt

PR #7808, opened 2018-04-18
Less strict Callback.args and improve webgl/line_compare example

### LINKED ISSUE #7807: line_compare.py example is not working anymore
The widget controls are not working anymore, I think. At least on Windows 10 and Chrome 65

https://github.com/bokeh/bokeh/blob/93b6d16747366f7165bad74a679a39ff14bb347f/examples/webgl/line_compare.py

```
""" Compare WebGL, SVG with canvas line.
"""
import numpy as np

from bokeh.models import Slider, Dropdown, CustomJS

p1 = figure(title="Canvas", output_backend="canvas")

p2 = figure(title="SVG", output_backend="svg")

p3 = figure(title="WebGL", output_backend="webgl")

ys = 10  # yscale, to increase anisotropy

lines = []
for p in (p1, p2, p3):

    t = np.linspace(0, 2 * np.pi, 50)
    x = np.sin(t) * 10
    y = np.cos(t) * 10
    l1 = p.line(x, y * ys, color="#2222aa",
                line_width=6, line_cap='butt',
                line_join='round', line_dash=(10, 6, 3, 6, 3, 6))

    t = np.arange(10)
    t = np.linspace(0, 4 * np.pi, 150)
    x = t - 5
    y = (t + 1) * ((t % 2) * 2 - 1)
    y = np.sin(t) + 5
    l2 = p.line(x, y * ys, color="#22aa22",
                line_width=6, line_cap='butt', line_join='round')

    t = np.arange(10)
    x = t - 5
    y = 0.3 * (t + 1) * ((t % 2) * 2 - 1) - 6
    l3 = p.line(x, y * ys, color="#aa2222",
                line_width=6, line_cap='butt',
                line_join='round', line_dash=(10, 10))
    l4 = p.line(y, x * ys, color="#aa2222",
                line_width=6, line_cap='butt',
                line_join='round', line_dash=(10, 10))

    lines.extend([l1, l2, l3, l4])

def add_callback(widget, prop):
    widget.callback = CustomJS(args=dict(widget=widget), code="""
        for (var i = 0; i < %s; i++) {
            var g = eval('line' + i).get('glyph');
            g.set('%s', widget.get('value'));
            window.g = g;
        }
        """ % (len(lines), prop))
    for i, line in enumerate(lines):
        widget.callback.args['line%i' % i] = line

def make_slider(prop, start, end, value):
    slider = Slider(title=prop, start=start, end=end, value=value)
    add_callback(slider, prop)
    return slider

def make_dropdown(prop, menu):
    dropdown = Dropdown(label=prop, menu=menu)
    add_callback(dropdown, prop)
    return dropdown

sliders = [
    make_slider('line_width', start=0.2, end=16, value=5),
    make_slider('line_dash_offset', start=0, end=100, value=1),
    make_dropdown('line_cap', [("butt", "butt"), ("round", "round"), ("square", "square")]),
    make_dropdown('line_join', [("miter", "miter"), ("round", "round"), ("bevel", "bevel")]),
]

sliders = column(*sliders)

output_file("line_compare.html", title="line_compare.py example")

show(row(sliders, p1, p2, p3))
```

mattpap:
I'm not sure why we did insist on having `args = Dict(String, Instance(Model))`, which is very limiting. `Dict(String, Any)` allows e.g. having a list of models or a constant (instead of using Python's string formatting).

fixes #7807

### DIFF
--- a/bokeh/models/callbacks.py
+++ b/bokeh/models/callbacks.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 from types import FunctionType
 
 from ..core.has_props import abstract
-from ..core.properties import Dict, Instance, String, Bool
+from ..core.properties import Dict, String, Bool, Any
 from ..model import Model
 from ..util.dependencies import import_required
 from ..util.compiler import nodejs_compile, CompilationError
@@ -71,10 +71,10 @@ class CustomJS(Callback):
         else:
             return cls(code=compiled.code, args=args)
 
-    args = Dict(String, Instance(Model), help="""
-    A mapping of names to Bokeh plot objects. These objects are made
-    available to the callback code snippet as the values of named
-    parameters to the callback.
+    args = Dict(String, Any, help="""
+    A mapping of names to Python objects. In particular those can be bokeh's models.
+    These objects are made available to the callback's code snippet as the values of
+    named parameters to the callback.
     """)
 
     code = String(default="", help="""
--- a/bokeh/models/filters.py
+++ b/bokeh/models/filters.py
@@ -4,7 +4,7 @@ import inspect
 from textwrap import dedent
 from types import FunctionType
 
-from ..core.properties import Bool, Dict, Either, Instance, Int, Seq, String
+from ..core.properties import Bool, Dict, Either, Int, Seq, String, Any
 from ..model import Model
 from ..util.dependencies import import_required
 from ..util.compiler import nodejs_compile, CompilationError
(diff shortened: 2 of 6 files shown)


bryevdv:
> I'm not sure why we did

AFAIK because it only worked that way, at least without more effort. The test failures here are real adn reproducible. Something about this change causes `Document._resolve_refs` to puke

bryevdv:
The issue is that e.g. the linear interpolator in the example, which is only in the doc as an `arg` to CustomJS is now for some reason not in `new_references` any longer:

screen shot 2018-04-18 at 11 30 43

bryevdv:
Seems like the Python side code could automatically add any `Instance(Model)` to the references when serializing, but now with `Any`, model instances in `args` are getting missed.

mattpap:
@bryevdv, I introduced `AnyRef` to resolve this issue. This way uses of `Any` aren't affected (possible performance degradation, e.g. sources) and we have an easy way to allow any values and still have reference discovery.

bryevdv:
Seems great, this is actually a very nice feature change.

Issue #7720, opened 2018-03-29
Add "downstream" CI build 


bryevdv:
ref: https://github.com/bokeh/bokeh/issues/7719

In an effort to ensure better API stability, add a new CI build that runs tests suites of prominent downstream projects. 

cc @mrocklin @jlstevens @philippjfr

PR #7813, opened 2018-04-19
add downstream test report

bryevdv:
fixes: #7720 

For now at least this test does not fail, it only generates a report. Want to avoid a situation where one or other project is intentionally out out of sync (major revision) and then the test can't be made to pass at all in any way for some interval

### DIFF
--- a/scripts/ci/test:downstream
+++ b/scripts/ci/test:downstream
@@ -1,11 +1,12 @@
 #!/bin/bash
 
-set -e # exit on error
+# disable set -e, just run to generate a report
+#set -e # exit on error
+
 set -x # echo commands
 
 pushd `python -c "import site; print(site.getsitepackages()[0])"`
 py.test distributed/bokeh
 popd
 
-# just run to generate a report
 exit 0
--- a/scripts/ci/test:downstream
+++ b/scripts/ci/test:downstream
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e # exit on error
+set -x # echo commands
+
+pushd `python -c "import site; print(site.getsitepackages()[0])"`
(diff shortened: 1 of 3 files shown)


bryevdv:
@philippjfr @jlstevens are there any tests you would suggest adding?

cc @mrocklin

philippjfr:
Apologies, I meant to reply to this ages ago. I think running our bokeh plotting unit tests would be sensible, I don't think you'd ever want to run our notebook tests due to the complexity and run times involved. Not sure what the best way for you to do so is though since we don't ship the unit tests inside the holoviews package.

mrocklin:
> Not sure what the best way for you to do so is though since we don't ship
the unit tests inside the holoviews package.

Are the tests very large?  If not then you might choose to change this
policy.

philippjfr:
>Are the tests very large?  If not then you might choose to change this policy.

They aren't and I've argued this point a few times. I'll restart that discussion.

bryevdv:
Ok well we should merge this now then, and circle back after HV has a release with tests

bryevdv:
@mattpap do you have any comments?

scripts/ci/test:downstream
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e # exit on error
+set -x # echo commands
+
+pushd `python -c "import site; print(site.getsitepackages()[0])"`
+py.test distributed/bokeh
```
mattpap: Given `-e` flag, this will fail this job when test fail (or am I missing something?).

bryevdv: Ah, yes apparently so. I thought it meant something slightly different.

PR #7818, opened 2018-04-20
fix missing python and data table updates

### LINKED ISSUE #7597: Bug: DataTable View doesn't rerender on table.source.change.emit()
DataTable View doesn't rerender on table.source.change.emit()

Is currently:
```
this.connect(this.model.source.properties.data.change, () => this.updateGrid());
```

Should be:
```
this.connect(this.model.source.change, () => this.updateGrid());
```

Warning - I hand-modified the BokehJS source and making this change appeared to dramatically hurt render performance. I would take care to check that this problem isn't introduced by a PR.

To reproduce (the right table should updated after a selection on the left-side plot:

```
from random import random

from bokeh.layouts import row
from bokeh.models import CustomJS, ColumnDataSource
from bokeh.models.widgets import DataTable, TableColumn

output_file("test.html")

x = [random() for x in range(500)]
y = [random() for y in range(500)]

s1 = ColumnDataSource(data=dict(x=x, y=y))
p1 = figure(plot_width=400, plot_height=400, tools="lasso_select, box_select", title="Select Here")
p1.circle('x', 'y', source=s1, alpha=0.6)

s2 = ColumnDataSource(data=dict(x=[], y=[]))
p2 = figure(plot_width=400, plot_height=400, x_range=(0, 1), y_range=(0, 1),
tools="", title="Watch Here")
p2.circle('x', 'y', source=s2, alpha=0.6)

columns = [TableColumn(field="x", title="X"),
TableColumn(field="y", title="Y"), ]
dt = DataTable(source=s2, width=400, height=280, columns=columns)

s1.callback = CustomJS(args=dict(s2=s2), code="""
var inds = cb_obj.selected['1d'].indices;
var d1 = cb_obj.data;
var d2 = s2.data;
d2['x'] = []
d2['y'] = []
for (i = 0; i < inds.length; i++) {
d2['x'].push(d1['x'][inds[i]])
d2['y'].push(d1['y'][inds[i]])
}
# s2.properties.data.change.emit(); # causes table update
s2.change.emit(); # doesn't update table
""")

layout = row(p1, p2, dt)

show(layout)
```

### LINKED ISSUE #7745: PointDrawTool : no python side update of the data source
In Bokeh: 0.12.15,  python 3.6.4 & firefox 59.0.2

I tried the [PointDrawTool](https://bokeh.pydata.org/en/latest/docs/user_guide/examples/tools_point_draw.html)  example modifying it for use with the bokeh server and removing the table widget. See code below

I find that the source is updated on the javascript side but **not** on the python server side: the python data change event is never triggered. On the other hand, the selection change event is triggered on both sides but without the source's data being updated in python.

A workaround is to add the widget again, possibly with a height and width set to 0. 

The code:
```
from bokeh.plotting import figure, Column, curdoc
from bokeh.models import DataTable, TableColumn, PointDrawTool, ColumnDataSource, CustomJS

p = figure(x_range=(0, 10), y_range=(0, 10), tools=[""],
           title='Point Draw Tool')
p.background_fill_color = 'lightgrey'

source = ColumnDataSource({
    'x': [1, 5, 9], 'y': [1, 5, 9], 'color': ['red', 'green', 'yellow']
})

renderer = p.circle(x='x', y='y', source=source, color='color', size=10)
columns = [TableColumn(field="x", title="x"),
           TableColumn(field="y", title="y"),
           TableColumn(field='color', title='color')]
table = DataTable(source=source, columns=columns, editable=True, height=0, width = 0)
source.on_change("selected", lambda attr, old, new: print("python changed selected", new))
source.on_change("data", lambda attr, old, new: print("python changed data", new))
source.js_on_change("selected", CustomJS(code = 'console.log("javascript: change selected", cb_obj)'))
source.js_on_change("data", CustomJS(code = 'console.log("javascript: change data", cb_obj)'))

draw_tool = PointDrawTool(renderers=[renderer], empty_value='black')
p.add_tools(draw_tool)
p.toolbar.active_tap = draw_tool

curdoc().add_root(Column(p))

```

python console:
```
2018-04-04 10:14:00,904 Starting Bokeh server version 0.12.15 (running on Tornado 4.5.2)
2018-04-04 10:14:00,909 Bokeh app running at: http://localhost:5006/toto
2018-04-04 10:14:00,909 Starting Bokeh server with process id: 30817
2018-04-04 10:14:01,588 200 GET /toto (127.0.0.1) 220.16ms
2018-04-04 10:14:01,972 101 GET /toto/ws?bokeh-protocol-version=1.0&bokeh-session-id=7YdOmfAtA3Jn38ry9QkTNoWpu2EOC1CiFRKYH81ubuKi (127.0.0.1) 1.34ms
2018-04-04 10:14:01,973 WebSocket connection opened
2018-04-04 10:14:01,974 ServerConnection created
python changed selected Selection(id='B7B1F1C26934499D8C622D851B9CD564', ...)
python changed selected Selection(id='E989B4D359664AE3924315EAD66C998C', ...)
```

browser console:
```
[bokeh] setting log level to: 'info'
logging.js:131
[bokeh] Websocket connection 0 is now open
connection.js:194
[bokeh] document idle at 302 ms
document.js:188
Bokeh items were rendered successfully
embed.js:262
javascript: change selected 
Object { _subtype: undefined, document: {…}, destroyed: {…}, change: {…}, transformchange: {…}, attributes: {…}, properties: {…}, _set_after_defaults: {…}, _pending: false, _changing: false, … }
bokeh.min.js:3:1
javascript: change selected 
Object { _subtype: undefined, document: {…}, destroyed: {…}, change: {…}, transformchange: {…}, attributes: {…}, properties: {…}, _set_after_defaults: {…}, _pending: false, _changing: false, … }
bokeh.min.js:3:1
javascript: change data 
Object { _subtype: undefined, document: {…}, destroyed: {…}, change: {…}, transformchange: {…}, attributes: {…}, properties: {…}, _set_after_defaults: {…}, _pending: false, _changing: false, … }
```

bryevdv:
- [x] issues: fixes #7745
- [x] issues: fixes #7597 
- [ ] tests added / passed

This fixes both issues that got lumped into #7745

The part missing for python updates was a self-assign to `.data`, so that the doc notification machinery is invoked. The part missing for the table updates was a more explicit change signal on `.data`. Not sure why it worked before, though. Data tables have required that signal for a long time AFAICT. 

@philippjfr are there other places that the "self-assign" should go, e.g. for other tools, so that python gets updates at the appropriate times?

@mattpap any comments welcome

We do need to document that Python side updates will only happen at the end of these edit operations and not continuously throughout.

### DIFF
--- a/bokehjs/src/lib/models/tools/edit/edit_tool.ts
+++ b/bokehjs/src/lib/models/tools/edit/edit_tool.ts
@@ -75,7 +75,8 @@ export abstract class EditToolView extends GestureToolView {
       }
     }
     for (const renderer of renderers) {
-      renderer.data_source.change.emit();
+      renderer.data_source.change.emit()
+      renderer.data_source.properties.data.change.emit()
     }
     this._basepoint = [ev.sx, ev.sy];
   }
--- a/bokehjs/src/lib/models/tools/edit/point_draw_tool.ts
+++ b/bokehjs/src/lib/models/tools/edit/point_draw_tool.ts
@@ -64,8 +64,12 @@ export class PointDrawToolView extends EditToolView {
     if (!this.model.drag) { return; }
     this._pan(ev);
     for (const renderer of this.model.renderers) {
-      renderer.data_source.selected.indices = [];
-      renderer.data_source.properties.data.change.emit();
+      renderer.data_source.selected.indices = []
+
+      // This is only needed to call @_tell_document_about_change()
+      renderer.data_source.data = renderer.data_source.data
+
+      renderer.data_source.properties.data.change.emit()
     }
     this._basepoint = null;
   }
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -950,16 +950,22 @@ class PointDrawTool(EditTool, Drag, Tap):
     ''' *toolbar icon*: |point_draw_icon|
 
     The PointDrawTool allows adding, dragging and deleting point-like
-    glyphs (of ``XYGlyph`` type) on one or more renderers by editing
-    the underlying ``ColumnDataSource`` data. Like other drawing
-    tools, the renderers that are to be edited must be supplied
-    explicitly as a list. Any newly added points will be inserted on
-    the ``ColumnDataSource`` of the first supplied renderer.
+    glyphs (of ``XYGlyph`` type) on one or more renderers by editing the
+    underlying ``ColumnDataSource`` data. Like other drawing tools, the
+    renderers that are to be edited must be supplied explicitly as a list.
+    Any newly added points will be inserted on the ``ColumnDataSource`` of
+    the first supplied renderer.
 
     The tool will automatically modify the columns on the data source
-    corresponding to the ``x`` and ``y`` values of the glyph. Any
-    additional columns in the data source will be padded with the
-    declared ``empty_value``, when adding a new point.
+    corresponding to the ``x`` and ``y`` values of the glyph. Any additional
+    columns in the data source will be padded with the given ``empty_value``
+    when adding a new point.
+
+    .. note::
+        The data source updates will trigger data change events continuously
+        throughout the edit operations on the BokehJS side. In Bokeh server
+        apps, the data source will only be synced once, when the edit operation
+        finishes.
 
     The supported actions include:
 
--- a/bokehjs/src/lib/models/widgets/tables/data_table.ts
+++ b/bokehjs/src/lib/models/widgets/tables/data_table.ts
@@ -119,13 +119,14 @@ export class DataTableView extends WidgetView {
   connect_signals(): void {
     super.connect_signals()
     this.connect(this.model.change, () => this.render())
-    this.connect(this.model.source.properties.data.change, () => this.updateGrid())
     this.connect(this.model.source.streaming, () => this.updateGrid())
     this.connect(this.model.source.patching, () => this.updateGrid())
-    this.connect(this.model.source.change, () => this.updateSelection())
+    this.connect(this.model.source.change, () => this.updateGrid(true))
+    this.connect(this.model.source.properties.data.change, () => this.updateGrid())
+    this.connect(this.model.source.selected.change, () => this.updateSelection())
   }
 
-  updateGrid(): void {
+  updateGrid(from_source_change=false): void {
     // TODO (bev) This is to enure that CDSView indices are properly computed
     // before passing to the DataProvider. This will result in extra calls to
     // compute_indices. This "over execution" will be addressed in a more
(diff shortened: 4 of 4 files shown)


bokehjs/src/lib/models/tools/edit/edit_tool.ts
```diff
@@ -75,7 +75,8 @@ export abstract class EditToolView extends GestureToolView {
       }
     }
     for (const renderer of renderers) {
-      renderer.data_source.change.emit();
+      renderer.data_source.change.emit()
+      renderer.data_source.properties.data.change.emit()
```
bryevdv: without the first, the drag does not update the glyph in a server app, without the second, a data table does not update in a standalone doc

bokehjs/src/lib/models/tools/edit/point_draw_tool.ts
```diff
@@ -64,8 +64,9 @@ export class PointDrawToolView extends EditToolView {
     if (!this.model.drag) { return; }
     this._pan(ev);
     for (const renderer of this.model.renderers) {
-      renderer.data_source.selected.indices = [];
-      renderer.data_source.properties.data.change.emit();
+      renderer.data_source.selected.indices = []
+      renderer.data_source.data = renderer.data_source.data
+      renderer.data_source.properties.data.change.emit()
     }
```
bryevdv: self assign required to notify documemt (to sync with python)

philippjfr: Thanks, I had previously assumed ``renderer.data_source.properties.data.change.emit()`` was sufficient.

bryevdv: it's definitely a weird edge, should probably be documented in the dev guide somewhere

bokehjs/src/lib/models/tools/edit/edit_tool.ts
mattpap: Sure, `GlyphRenderer` depends on the former and `DataTable` depends on the later. If you are emitting change signals manually, you have to do this in pairs, to cover all cases.

mattpap:
The next step should be really to figure out how to do this properly, so that we don't have to remember how many signals are necessary and when `data = data` is needed. Anyway, for now LGTM.

bryevdv:
> The next step should be really to figure out how to do this properly, so that we don't have to remember how many signals are necessary and when data = data is needed. Anyway, for now LGTM.

The asymmetry here is, I think, the fact that when setting up a callback, it has to be on a property, so, on `data:
```
source.on_change('data', ...)
```
But in CustomJS callbacks, it's simpler to trigger a change on the source:
```
source.change.emit() // vs source.properties.data.emit()
```
That's why I connected both, but I agree this can be cleaned up.

PR #7820, opened 2018-04-22
Add support for center origin to box select/zoom tools

### LINKED ISSUE #3126: Add box zoom out tool
It is a common practice to have both zoom in and zoom out tools in various visualization applications, such as image preview, map, and PDF readers. However, current Bokeh (0.10.0) does not provide a box zoom out tool, which is sometimes inconvenient to users. It would be better to have this tool available.

mattpap:
fixes #3126

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -373,7 +373,7 @@ DEFAULT_BOX_OVERLAY = lambda: BoxAnnotation(
     line_color="black",
     line_alpha=1.0,
     line_width=2,
-    line_dash=[4, 4]
+    line_dash=[4, 4],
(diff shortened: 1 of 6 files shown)


mattpap:
Originally I intended this feature to be a part of a bigger change, where I would allow to easily configure tools from within bokeh's UI. Currently this PR is a nice addition, but a bit pointless if user has to change code to get it working. However, configurability I have in mind is a bit more involved to implement, so there is no point in stalling this PR.

bryevdv:
I think it's useful even programmatically, so happy to merge (some applications always benefit from this kind of zoom).

PR #7827, opened 2018-04-23
support axes at fixed locations

### LINKED ISSUE #113: Axis should have option to have a fixed orientation
Right now axes determine the orientation of the "In" vector based on which side of the plot they're on.  For axis locations of "min" or "max", that's OK, but when a plot has a fixed position in dataspace, then it should be able to have a fixed orientation, i.e. tick labels are always above/below or always left/right of the axis.

bryevdv:
- [x] issues: fixes #113
- [x] tests added / passed

This implements the plan outlined in #113 and produces output such as the following:

screen shot 2018-04-23 at 12 59 01

Including working with categorgical coordinates:

screen shot 2018-04-23 at 12 59 27

It's worth noting here that at least for now, axes labels are suppressed when a fixed location is provided. 

@mattpap I'd like to add some more docs and examples (and a few tests) but before going further I'd like your comments.

### DIFF
--- a/bokeh/models/axes.py
+++ b/bokeh/models/axes.py
@@ -146,6 +146,16 @@ class Axis(GuideRenderer):
     main plot area.
     """)
 
+    fixed_location = Either(String, Float, help="""
+    Set to specify a fixed coordinate location to draw the axis. The direction
+    of ticks and major labels is determined by the side panel that the axis
+    belongs to.
+
+    .. note::
+        Axes labels are suppressed when axes are positioned at fixed locations
+        inside the central plot area.
+    """)
+
 @abstract
 class ContinuousAxis(Axis):
     ''' A base class for all numeric, non-categorical axes types.
--- a/bokeh/models/axes.py
+++ b/bokeh/models/axes.py
@@ -146,7 +146,7 @@ class Axis(GuideRenderer):
     main plot area.
     """)
 
-    fixed_location = Either(Float, String, Tuple(String, String), Tuple(String, String, String), help="""
+    fixed_location = Either(Float, String, Tuple(String, String), Tuple(String, String, String), default=None, help="""
     Set to specify a fixed coordinate location to draw the axis. The direction
     of ticks and major labels is determined by the side panel that the axis
     belongs to.
--- a/bokeh/models/axes.py
+++ b/bokeh/models/axes.py
@@ -146,7 +146,7 @@ class Axis(GuideRenderer):
     main plot area.
     """)
 
-    fixed_location = Either(String, Float, help="""
+    fixed_location = Either(Float, String, Tuple(String, String), Tuple(String, String, String), help="""
     Set to specify a fixed coordinate location to draw the axis. The direction
     of ticks and major labels is determined by the side panel that the axis
     belongs to.
--- a/bokehjs/src/lib/models/axes/axis.ts
+++ b/bokehjs/src/lib/models/axes/axis.ts
@@ -77,7 +77,7 @@ export class AxisView extends GuideRendererView {
     return this._tick_extent() + this._tick_label_extent() + this._axis_label_extent()
   }
 
-  get needs_clip() : boolean {
+  get needs_clip(): boolean {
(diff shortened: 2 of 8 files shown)


bryevdv:
The categorical example above was generated with:
```

p = figure(title = "Iris Morphology", x_range=['setosa', 'versicolor', 'virginica'])

p.circle(flowers["species"], flowers["petal_width"],
p.yaxis.fixed_location = "versicolor"

output_file("iris.html", title="iris.py example")

show(p)

```

bokeh/models/axes.py
```diff
@@ -146,6 +146,16 @@ class Axis(GuideRenderer):
     main plot area.
     """)
 
+    fixed_location = Either(String, Float, help="""
```
mattpap: `String`?

mattpap: Also, there should be an explicit default given that the default in bokehjs is `null`.

bokehjs/src/lib/models/axes/axis.ts
```diff
@@ -70,9 +70,16 @@ export class AxisView extends GuideRendererView {
   }
 
   protected _get_size(): number {
+    if (this.model.fixed_location != null) {
+      return 0
+    }
     return this._tick_extent() + this._tick_label_extent() + this._axis_label_extent()
   }
 
+  get needs_clip() : boolean {
```
mattpap: Whitespace.

bokehjs/src/lib/models/axes/axis.ts
```diff
@@ -380,6 +387,7 @@ export namespace Axis {
     major_tick_out: number
     minor_tick_in: number
     minor_tick_out: number
+    fixed_location: any
```
mattpap: It would have been useful to have this filled in at this point (`float | null`).

bokeh/models/axes.py
bryevdv: String is for categorical coordinates. It should probably accept various kinds of tuples for nested categorical support, but OK to start with this.

bokehjs/src/lib/models/axes/axis.ts
bryevdv: good point, the `p.Any` threw me off (tho `float | string | null`) I think

mattpap: So, based on another comment, if `string` is for categoricals, then this should be `Factor` (from `models/ranges/factor_range`) instead.

mattpap: So, `number | Factor | null`.

mattpap: I'm not sure where I took `float` from.

bryevdv: unfortunately so far  trying to change this goes down a rabbit hole of type problems.

bokehjs/test/models/axes/axis.coffee
```diff
@@ -32,6 +34,43 @@ describe "Axis", ->
...
+  it "loc should return numeric fixed_location", ->
+    doc = new Document()
+    p = new Plot({
+      x_range: new Range1d({start: 0, end: 10})
+      y_range: new Range1d({start: 0, end: 10})
+    })
+    doc.add_root(p)
+    ticker = new BasicTicker()
+    formatter = new BasicTickFormatter()
+    axis = new Axis({
```
mattpap: You are constructing an abstract class here. This is btw. why I want to have all new tests in typescript, to avoid such mistakes.

bryevdv: I am not opposed, I can just only handle so much at a time. In any case given how much implementation is inside Axis I think it needs to be concrete, just so we can test it. Are you proposing any specific change here right now or is this good to merge?

bryevdv:
Absent other comment I'll go ahead and merge this,  the comment above can be addressed when the test are converted to TS (will need that anyway, other other tests there all do the same thing)

PR #7829, opened 2018-04-24
Update for npm 6

bryevdv:
Other PRs will need to rebase/merge master after this. 

cc @mattpap

### DIFF
--- a/bokehjs/package-lock.json
+++ b/bokehjs/package-lock.json
@@ -10,8 +10,8 @@
       "integrity": "sha1-iQrnxdjId/bThIYCFazp1+yUW9o=",
       "dev": true,
       "requires": {
-        "normalize-path": "2.1.1",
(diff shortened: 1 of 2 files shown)


mattpap:
Even if everything works with new version (which isn't necessarily clear from just tests being all green), `package-lock.json` is now quite different from the one generated by npm 5.x. This means that we can't support both 5.x and 6.x, because people using different versions will submit PRs modifying this file back-and-forth. It was a mistake to use `npm install -g npm`, i.e. without version range, which is the cause of this issue.

bryevdv:
AFAICT things are working. The s3 reports look good and things worked with a random sample of local testing. I'm happy with going back to `<6.0` with a version specified in the install, or staying at `>=6.0`. Maybe +0.5 for the latter but I leave it up to you

PR #7830, opened 2018-04-24
Accommodate more color and size options in Crossfilter Example

### LINKED ISSUE #7725: Increase `size` and `color options for Crossfilter Example
Looking at the [`crossfilter` example](https://github.com/bokeh/bokeh/blob/master/examples/app/crossfilter/main.py), I wanted to use something similar for my own data with a small adjustment, I wanted to include more data features in the `color` and `size` options.  Rather than only allowing those features that are "quantile-able", I used the `continuous` list and changed [line 46](https://github.com/bokeh/bokeh/blob/master/examples/app/crossfilter/main.py#L46) to be

```
n_groups = len(set(df[size.value]))
if n_groups > len(SIZES):
    groups = pd.qcut(df[size.value].values, len(SIZE)
else:
    groups = pd.Categorical(df[size.value])
```

I applied essentially the same changes to [line 51](https://github.com/bokeh/bokeh/blob/master/examples/app/crossfilter/main.py#L51) for the colors.  

This makes the example able to handle a bit more.

StevenCHowell:
These changes accommodate including data features with less than 20 distinct values in the "Color" and "Size" options.   

- [ ] issues: fixes #7725

### DIFF
--- a/examples/app/crossfilter/main.py
+++ b/examples/app/crossfilter/main.py
@@ -19,7 +19,6 @@ del df['name']
 columns = sorted(df.columns)
 discrete = [x for x in columns if df[x].dtype == object]
 continuous = [x for x in columns if x not in discrete]
-quantileable = [x for x in continuous if len(df[x].unique()) > 20]
 
 def create_figure():
     xs = df[x.value].values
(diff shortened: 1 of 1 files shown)


bryevdv:
@StevenCHowell can you estimate how much work is left for this PR?

StevenCHowell:
@bryevdv I think it's good to go.  The bulk of the changes were in the first commit.

While doing this I realized the changes I made primarily apply to datasets where some of the data features have more possibilities.  In an effort to demonstrate this I thought it would be interesting to adapt it to show more of the demo datasets within Bokeh, as a separate drop-down.  I did not have much time to toy around with it.  It seemed I would need to figure out how to update the list entries for all the other lists based on the data selected.  Regardless, this would be better as a separate Issue and PR.  Though it less significant for the cars dataset, the changes I made would allow it to include features with less than the previous 20-cutoff applied for the "color" and "size" options.

PR #7839, opened 2018-04-29
Add Range Tool

### LINKED ISSUE #7795: Feature-request: Selection-Bar
This is a feature-request for a Selection-Bar. You can see an example about Selection-Bar [here](https://webapps.qlik.com/masters/index.html#/2018-stats/). I am interested to help to do this feature but I need to understand a few things.

**How would you implement Selection-Bar on Bokeh? With Extension or build it to the base package?**

bryevdv:
Super WIP

- [x] issues: fixes #7795
- [x] tests added / passed


bryevdv:
Probably will not have *all* the bells and whistles that would be nice, but:

screenflow

bryevdv:
@mattpap if there are some `util` functions that would obviate some of the code in the new range tool file, please let me know. Also, do you now why the `set_cursor` calls would behaving no effect?

bryevdv:
Well, the cursor is not working because the ui events code is setting things one move, and since this does not plug into that machinery, it gets set to "default". So I guess the best thing to do is make `BoxAnnotation` optionally handle supplying a bbox and setting a `cursor` property itself.

bryevdv:
OK, got this working with cursors, which is the minimum I consider necessary for this to be usable:

ezgif-4-9563c346f0

I still am going to add some tests and docs and examples, but the implementation should be close. @mattpap can you take a look? I am not 100% happy but it seems reasonable for now. Basically I had to make `cursor` be a function accepting `sx` and `sy` so that different cursors could be returned in different situations. Also I made `bbox` into `bbox_contains` which returns a bool. Was getting invalid bboxes with non-rangetool box annotations, and wanted to find a way to short circuit the work in the standard case anyway.

bryevdv:
still needs docs but otherwise ready for review

bokehjs/src/lib/models/tools/gestures/range_tool.ts
```diff
@@ -0,0 +1,245 @@
...
+    level: "overlay",
+    render_mode: "css",
+    fill_color: "lightgrey",
+    fill_alpha: {value: 0.5},
+    line_color: {value: "black"},
+    line_alpha: {value: 1.0},
+    line_width: {value: 0.5},
+    line_dash: [2, 2],
+    })
+  }
```
mattpap: Indentation is off here.

bokehjs/src/lib/models/annotations/box_annotation.ts
```diff
@@ -111,6 +119,25 @@ export class BoxAnnotationView extends AnnotationView {
...
+
+  bbox_contains(sx: number, sy: number): boolean {
+    if (this.model.in_cursor == null)
+      return false
+    const bbox = new BBox({x0: this.sleft-3, y0: this.stop-3, x1: this.sright+3, y1: this.sbottom+3})
+    return bbox.contains(sx, sy)
+  }
+
+  cursor(sx: number, sy: number): string | null {
+    const tol = 3
```
mattpap: I presume this number and the above are the same. This has to be extracted. Is this somehow related to border width (line width)? If so, lets use that instead.

bokehjs/src/lib/models/annotations/legend.ts
```diff
@@ -131,9 +131,10 @@ export class LegendView extends AnnotationView {
     return {x: sx, y: sy, width: legend_width, height: legend_height}
   }
 
-  bbox(): BBox {
```
mattpap: Having access to the bbox is actually useful to me.

bryevdv: OK well the issue was that the `BBox` was being used to determine whether to engage an interactive feature or not, and since there is no "null bbox" there was no clean way to turn that off under certain conditions. I will: 

* restore an `interactive_bbox` function, that just unconditionally returns a bbox that describes what the interactive hit area should be, if interactions are turned on

* rename `bbox_contains` to `interactive_hit` and have it use the above and be responsible for the on/off policy setting

bokehjs/src/lib/models/annotations/box_annotation.ts
bryevdv: It's just to make the UI landing area slightly bigger for users, `line_width+1` is probably reasonable.

bokehjs/src/lib/models/tools/gestures/range_tool.ts
bryevdv: fixed in #74e294f

bokehjs/src/lib/models/annotations/box_annotation.ts
bryevdv: fixed in 2cb8e88

bokehjs/src/lib/models/annotations/legend.ts
bryevdv: fixed in https://github.com/bokeh/bokeh/pull/7839/commits/adfa5840cee963976f522a471c18b3eaab404e71

bokehjs/src/lib/models/annotations/box_annotation.ts
```diff
@@ -120,10 +122,20 @@ export class BoxAnnotationView extends AnnotationView {
     ctx.restore()
   }
 
-  bbox_contains(sx: number, sy: number): boolean {
+  interactive_bbox() {
```
mattpap: `tslint` will complain. Every function and method must have explicit return type.

bryevdv: right just an oversight

bryevdv: fixed in https://github.com/bokeh/bokeh/pull/7839/commits/e7d2ef600b0a95c4bd67af7d3473a1aba7830c2a

bryevdv:
@mattpap I'm going to do a larger docs PR this weekend. Any objections to merging this now?

mattpap:
Well, I have some reservations, but that can be addressed later. LGTM.

bryevdv:
OK I will go ahead and merge, if you can state those briefly here it would be helpful to me.

mattpap:
To be specific, e.g. I'm not sure if `BoxAnnotation` is still an annotation after those changes. Though this isn't an uncommon situation with annotations.

jbednar:
Is there an icon to enable and disable the range?  I'm not saying there necessarily needs to be; just curious because the box select tool that we've previously used for a similar purpose has an icon in the toolbar.

mattpap:
@jbednar, https://github.com/bokeh/bokeh/blob/dacc684abdb3384fa5c786ba53267522c5fa5708/sphinx/source/_images/icons/Range.png.

bryevdv:
I'm not sure deactivating hides the overlay. I was thinking if probably should, if it doesn't already I will do that in a follow on

bryevdv:
As for the icon I am happy to replace it with anything better

jbednar:
The icon looks fine to me, thanks!

PR #7844, opened 2018-04-30
Clean up package.json

### LINKED ISSUE #7841: Drop dependency on webpack

mattpap:
Removes unnecessary dependencies from `package.json`.

fixes #7841


bokehjs/webpack.config.js
```diff
@@ -1,34 +0,0 @@
...
-  },
-  plugins: [
-    new webpack.optimize.CommonsChunkPlugin({
-      name: "bokeh",
-    }),
-    new webpack.IgnorePlugin(/moment$/)
-  ]
-};
-
-module.exports = config;
```
mattpap: This could be moved to a wiki if we want to keep it, as long as this still works.

bryevdv: I've added it here for now:

https://github.com/bokeh/bokeh/wiki/webpack.config.js

But we should find a way to support this in a more integrated way (i.e. make sure it works, and have a documentation section on it) since it seems to come up reasonably often.

Issue #7788, opened 2018-04-11
DateRangeSlider error not logged nor shown in CLI


hyamanieu:
I noticed the following bug:  If you have a DateRangeSlider of which max and min are the same date, the browser shows a blank page client side, and the CLI and error loggers don't show anything unusual.

It took me a while to find the origin of the problem due to the lack of logging.

Conditions hwere this bug was found:

- Tested on both Windows 10 and Ubuntu 16.04
- Python 3.6, tornado 4.5.2, bokeh 0.12.15 (on both systems)
- Firefox (on both systems)

How to reproduce: 
- Put this file in a local folder (e.g. softfocus/): https://github.com/hyamanieu/softfocus/blob/master/softfocus/main.py
- change lines 224 to 227 from (after the bug fix)
```
last_date = self.df['last modification'].max() 
first_date = self.df['last modification'].min()
if last_date == first_date:
    last_date = first_date + timedelta(days=1)
```

to these two lines:
```
last_date = self.df['last modification'].max() 
first_date = last_date
```
With your CLI, cd into the parent folder and execute
`bokeh serve softfocus --show --log-level debug`

Output: blank page on your browser, no errors/warnings in your CLI
Expected output: a working webpage with a frozen DateRangeSlider.

bryevdv:
Is there any JS console output in the browser?

hyamanieu:
Nothing.

bryevdv:
OK, well we will have to try and take a closer look, but there are many issues, so I am not sure offhand when that may be.

hyamanieu:
People with the blank page problem will hopefully find out about it, so it's not a priority.

hyamanieu:
Erratum: I just discovered the javascript console in the browser, I thought it would come in the CLI. 
I don't have time to check this issue again for the moment, but I will report back asap.

hyamanieu:
Here's the JS Console rendering error message:

```
Error rendering Bokeh items  
Error: noUiSlider (10.1.0): 'range' 'min' and 'max' cannot be equal.
Stack trace:
y@http://localhost:5006/static/js/bokeh-widgets.min.js?v=a2dd57b513ce6b5a818fc187fe161f28:2:3016
O/<@http://localhost:5006/static/js/bokeh-widgets.min.js?v=a2dd57b513ce6b5a818fc187fe161f28:2:8470
[...]

PR #7850, opened 2018-05-01
Fixed name collision in ion_range_slider example

### LINKED ISSUE #7849: IonRangeSlider example fails when adding more than one slider.
Using the code as specified in https://bokeh.pydata.org/en/latest/docs/user_guide/extensions_gallery/widget.html#userguide-extensions-examples-widget. The local variable 'input' overwrites the 'input' from core/dom if more than one IonRangeSlider is added. For example:

``` Python
ion_range_slider = IonRangeSlider(start=0.01, end=0.99, step=0.01, range=(min(x), max(x)),
    title='Ion Range Slider - Range', callback=callback_ion, callback_policy='continuous')
ion_range_slider2 = IonRangeSlider(start=0.01, end=0.99, step=0.01, range=(min(x), max(x)),
    title='Ion Range Slider - Range', callback=callback_ion, callback_policy='continuous')
```
Raises:
``` console
Error rendering Bokeh items  TypeError: dom_1.input is not a function
    at IonRangeSliderView.exports.IonRangeSliderView.IonRangeSliderView.template (ionrangeslider:68478)
    at IonRangeSliderView.exports.IonRangeSliderView.IonRangeSliderView.initialize (ionrangeslider:68498)
    at IonRangeSliderView.View [as constructor] (ionrangeslider:9175)
    at IonRangeSliderView.DOMView [as constructor] (ionrangeslider:3839)
    at IonRangeSliderView.LayoutDOMView [as constructor] (ionrangeslider:21889)
    at IonRangeSliderView.WidgetView [as constructor] (ionrangeslider:49492)
    at IonRangeSliderView.InputWidgetView [as constructor] (ionrangeslider:48486)
    at new IonRangeSliderView (ionrangeslider:68460)
    at Object.build_views (ionrangeslider:3636)
    at WidgetBoxView.LayoutDOMView.build_child_views (ionrangeslider:22073)
```

bokeh == 0.12.15
python == 3.6.5
browser == google chrome Version 64.0.3282.119 (Official Build) (64-bit)

Karel-van-de-Plassche:
- [x] issues: fixes #7849
- [x] tests added / passed

### DIFF
--- a/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider.coffee
+++ b/sphinx/source/docs/user_guide/examples/extensions_ion_range_slider.coffee
@@ -76,8 +76,8 @@ export class IonRangeSliderView extends InputWidgetView
       onFinish: (data) => @slidestop(data),
     }
 
-    input = @$el.find('.slider')[0]
-    slider = jQuery(input).ionRangeSlider(opts)
+    inp = @$el.find('.slider')[0]
+    slider = jQuery(inp).ionRangeSlider(opts)
     range = [from, to]
     @$el.find( "##{ @model.id }" ).val( range.join(' - '))
     @$el.find('.bk-slider-parent').height(@model.height)


bryevdv:
Thanks for the PR @Karel-van-de-Plassche! 

@mattpap  I think converting this to TypeScript would allow for the correct scoping to leave the name as is? However, I am OK with this quick fix until the example can be ported. Do you have any objection?

Karel-van-de-Plassche:
Maybe this needs its own issue, but are there any plans to 'officially' support IonRangeSlider? I'd add my voice to making this the default slider, as per https://github.com/bokeh/bokeh/issues/7788. Unfortunately I'm not a JS dev, otherwise I'd have a stab at it.. Might have a go at it either way ;) Right now I just use my own IonRangeSlider implementation exposing more features that I keep semi-up-to-date at https://github.com/Karel-van-de-Plassche/bokeh_ion_range_slider.

mattpap:
> I think converting this to TypeScript would allow for the correct scoping to leave the name as is? However, I am OK with this quick fix until the example can be ported. Do you have any objection?

It would, however we need to test that coffeescript works in custom models, as long as we support it. So I'm fine with this change.

mattpap:
> Maybe this needs its own issue, but are there any plans to 'officially' support IonRangeSlider?

What are the benefits of IonRangeSlider over range sliders we have implemented in bokeh (based on noUiSlider)? It is more likely that we would implement missing features, rather than add another implementation. Additionally, IonRangeSlider depends on jquery, which makes it a very unlikely candidate for becoming a part of bokehjs.

bryevdv:
I would say ionRangeSlider is definitely a better slider in almost all respects than noUiSlider, in terms of appearance, configurability, and features, with the sole exception of lacking a vertical orientation. Unfortunately I don't really think we can replace it as the default slider at this point, but I would not be against adding at as another optional bundle so that users don't have to resort to copy and pasting a custom extension just to use it.

Karel-van-de-Plassche:
To be honest I have not looked at it for a while (pre bokeh 0.12.5). The feature I missed in an older version were discrete values, custom strings as labels, formatting of numbers in labels and ticks. I don't know how much can be done in the current bokeh slider. Also, I think the IonRangeSlider looks way nicer, but maybe this can be achieved with a skin.

Karel-van-de-Plassche:
> but I would not be against adding at as another optional bundle
I would be happy to supporting this as a bundle, I'll read up on the developer notes tomorrow.

bryevdv:
@Karel-van-de-Plassche I doubt there is much by way of documentation on this, it's somewhat tied to our build/release process which has mostly only involved a couple of people. That said perhaps this is an opportunity to improve the dev docs. In terms of previous models, you can look to how the bokeh tables widgets bundle was split off.

bryevdv:
Well, that happened in this PR, which was unfortunately rather large, but individual commits may point to how it worked to split things up: https://github.com/bokeh/bokeh/pull/6335

mattpap:
If anything, this may be a good opportunity to figure out **reusable** custom models. So far we only focused to provide support for custom models at all, for individual use. To get started with that, one would have to move the custom model code to a new repository, rewrite the code in TypeScript (otherwise maintenance burden would be to high), setup CI, and we will see then what else needs to be done (implementation- or process-wise). As long as this bundle isn't a core bokehjs bundle, then the dependency on jquery doesn't have to be that strict, although the approach we use in bokeh-tables is still preferred.

PR #7851, opened 2018-05-01
Improve speed of bokehjs' build

### LINKED ISSUE #7848: Improve speed of bokehjs' build
The build should be incremental and only process (scan for dependencies, minify) files that changed. We should also build `*.min.js` files by minifying and bundling individual small files instead of minifying the generated bundle.

mattpap:
- [x] minify individual files
- [ ] incremental compilation

fixes #7848


mattpap:
I will have to start this from scratch.

Issue #7852, opened 2018-05-02
Create optional bundle for IonRangeSlider from the examples


Karel-van-de-Plassche:
Create an optional bundle for the IonRangeSlider, as discussed in https://github.com/bokeh/bokeh/pull/7850#issuecomment-385771880. Relevant snips:

> I would say ionRangeSlider is definitely a better slider in almost all respects than noUiSlider, in terms of appearance, configurability, and features, with the sole exception of lacking a vertical orientation. Unfortunately I don't really think we can replace it as the default slider at this point, but I would not be against adding at as another optional bundle so that users don't have to resort to copy and pasting a custom extension just to use it.

> I doubt there is much by way of documentation on this, it's somewhat tied to our build/release process which has mostly only involved a couple of people. That said perhaps this is an opportunity to improve the dev docs. In terms of previous models, you can look to how the bokeh tables widgets bundle was split off.

> Well, that happened in this PR, which was unfortunately rather large, but individual commits may point to how it worked to split things up: #6335

Karel-van-de-Plassche:
I'm calling it a day for today.

Spend today reading and looking into the bokeh structure. I've never written TypeScript before (also no JavaScript..) so I'm sure my code will need a lot of checking and refactoring. I've initialized a folder and file structure in the 'contrib' style, but a permanent solution will have to be found in #7853.

I tried to make it a bit stand-alone, so I did not extend `AbstractSlider`. Instead, I copied heavily from it and mashed it together with my https://github.com/Karel-van-de-Plassche/bokeh_ion_range_slider. The result can be found here https://github.com/Karel-van-de-Plassche/bokeh/commit/07a10ab0b30f3206a29f66b28314ddebd58b3ac2.

My current example builds (`python setup.py develop --build-js`) and the JS tests pass (`py.test -m js`) , but `python ./examples/models/file/ion_range_sliders.py` currently show a blank screen. Note to future me:

```
Uncaught Error: Model 'IonRangeSlider' does not exist. This could be due to a widget
                     or a custom model not being registered before first usage.
    at Object.i.Models (ion_range_sliders.html:26)
    at Function.t._instantiate_object (ion_range_sliders.html:26)
    at Function.t._instantiate_references_json (ion_range_sliders.html:26)
    at Function.t.from_json (ion_range_sliders.html:26)
    at ion_range_sliders.html:26
    at ion_range_sliders.html:26
```

bryevdv:
[...]

PR #7855, opened 2018-05-02
Added how-to set up conda environment using fish shell

Karel-van-de-Plassche:
- [x] issues: Related to #7852
- [ ] tests added / passed


bryevdv:
@Karel-van-de-Plassche thanks for the PR, I was going to make a request to to bring the line lengths into consistency with the rest of the doc, but it was simple to do from the web UI, so for expediency I just made the edits here. I will merge when Travis goes green

Issue #7864, opened 2018-05-08
Rewrite bokehjs' tests in TypeScript


mattpap:
A follow up on #6481. Writing tests in TypeScript is an important part of verifying that our typings actually work, especially parts that can't be verified within the lib itself. This is also an opportunity to clean up the atrocious mess that bokehjs' unit tests are.

Rewrite progress:

- [x] client
- [x] core
- [x] document
- [x] models
- [x] protocol

PR #7865, opened 2018-05-08
Rewrite test/core in TypeScript

mattpap:
addresses #7864


mattpap:
This doesn't fully rewrite core yet. `properties` and `ui_events` need to be rewritten, but this will happen in another PR due to their size. This also removes `utils.require`, so there is now one way to imports modules in tests. Additionally, this replaces ts-node with upfront compilation. This reduces the overall time of compilation and test time from 34 s to 20 s (7 + 14)  (on my computer) and allows to see all typescript errors upfront.

bryevdv:
@mattpap please note I just added a bunch of tests to the old `templating.coffee` in the image hover PR

mattpap:
I noticed this as well. Garbled output is most likely due to upgrade of mocha. Unknown edit variable is a result of bokehjs' unit tests being utter crap. I also changed the way plot canvas is obtained across all tests, so that plot canvas created by a plot is used instead of new plot canvas being created. This was the right thing to do (we are now testing something that is more like bokehjs in its behavior), but this may have exacerbated the fact that tests depend on getting the timing exactly right. I don't know right now, I will be investigating, because I can't make tests all green in my other PRs.

mattpap:
> Garbled output is most likely due to upgrade of mocha.

This may also be the reason why I didn't upgrade mocha for so long.

bryevdv:
we can downgrade if need be, it seemed to be working fine earlier. Or possibly there is a different reporter we can use

mattpap:
> we can downgrade if need be, it seemed to be working fine earlier. Or possibly there is a different reporter we can use

It worked smoothly, because there were no errors. This doesn't depend on the reporter of choice, it just kind of quits when the first error occurs, but also keeps going.

bryevdv:
The problem appears to stem from the ""interacting with GlyphViews" test in visuals.ts, which is run right before the document test. It has that subbing code, which seems to cause some delayed error.

mattpap:
> However, I think it was better to be able to collect all the errors at the end.

Yes, this is what I would like to preserve.

mattpap:
> The problem appears to stem from the ""interacting with GlyphViews" test in visuals.ts, which is run right before the document test. It has that subbing code, which seems to cause some delayed error.

I'm not sure about the source, but the problem in general is as follows. A test consists of:

- before block, which usually constructs objects and stubs out problematic code (solver, canvas)
- the test
- after block, which almost never destroys anything, but restores stubbed out code
- delayed side effects caused by all three steps before (mostly the second)

Delayed (or deferred; it's pretty much the same concept) side effects are caused by `setInterval`, `setTimeout`, `setImmediate`, `delayAnimationFrame` (also our wrappers `defer`, `delay`) (maybe others). Signals are not the sources here, because connected slots are invoked immediately at emit. I'm not sure about the ordering between after block and delayed side effects, but in general I wouldn't count too much on any specific order here. Instead I would make sure to either cancel any possible side effects in before block or make sure that the test block doesn't finish before all side effects finished. The main source of delayed side effects is `request_render`, though those are also present in tiles, image_url and possibly elsewhere. I think, though I'm not sure, that subbing out `request_render` instead of the solver should bring immediate relief to the discussed class of issues with our unit tests.

mattpap:
> Adding --exit will make it stop and report correctly after the first failure

If I understand this correctly, this only applies to situation when there is an actual failure in the tested library or test code, and not a test failure (test not passing) per se. What we observe in our tests is a delayed function failing, detached from any particular test. At this point we don't know if the library and/or tests are in a consistent state or not, so it seems logical to terminate test run at this point. Though lets make sure that this doesn't terminate a test run after a first test failure.

bryevdv:
Yes that is consistent with my understanding and experience. AFAIK only a fairly fraction of our tests have the delayed/deferred style issues, but obviously they all need to be re-worked. Anyway I'd propose merging the rollback PR for now.

PR #7871, opened 2018-05-08
use whole doc for saving

### LINKED ISSUE #7621: `save` creates non-working files when used with widgets
Bokeh v0.12.15dev2

```python
from bokeh.plotting import figure, save, output_file

fig = figure()
fig.x([0], [0])

button = Button(label="Button")

d = Document()
d.add_root(fig)
d.add_root(button)

output_file('fig.html')
save(fig)
```

Now, if you open `fig.html` in your browser, you should see `Uncaught Error: Model 'Button' does not exist.`

The reason is that `bokeh.embed.bundle.bundle_for_objs_and_resources` checks just the passed models for widgets (`fig` in this case), but `bokeh.embed.util.standalone_docs_json_and_render_items` saves the whole document JSON.

It's hard for me to say how it should work ideally. Most likely, there should be some common function used by `save` that extracts only relevant models from a document, and only those models are checked for widgets and bundled as a JSON.

bryevdv:

- [x] issues: fixes #7621
- [x] tests added / passed

### DIFF
--- a/bokeh/embed/standalone.py
+++ b/bokeh/embed/standalone.py
@@ -275,10 +275,10 @@ def file_html(models,
     '''
     models = check_models_or_docs(models)
 
-    with _ModelInDocument(models, apply_theme=theme):
+    with _ModelInDocument(models, apply_theme=theme) as doc:
         (docs_json, render_items) = standalone_docs_json_and_render_items(models)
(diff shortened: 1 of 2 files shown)


bryevdv:
Merging, since this was previously discussed

PR #7874, opened 2018-05-09
make grid bands handle ends correctly

### LINKED ISSUE #6698: Bands wrongly coloured
Following the examples in the User Guide, bands are wrongly coloured when panned up and down. 
http://bokeh.pydata.org/en/latest/docs/user_guide/styling.html#bands
Try to pan (slowly) up and down to see the effect.

There are probably 2 errors involved: 
1) the bottom line, when incomplete, has the same color of the line above. When incomplete, the top line looks always light blue; 
2) the line count to determine the color is based on the lower major tick on y: when you pan up/down colors are no more consistent and change during movement. It would be nice to start the color line count on the real "0" instead of on the lower major tick on y so on pan the bands would keep their own color. Of course the same problem applies when zooming in/out, but there is nothing that can be done about it. 
bokeh_plot

bryevdv:
- [x] issues: fixes #6698
- [x] tests added / passed

Does not fix the flickering issue, going to punt on that. But makes sure the ends are always alternating from the first full band:

screen shot 2018-05-08 at 21 53 09

### DIFF
--- a/bokehjs/src/lib/models/grids/grid.ts
+++ b/bokehjs/src/lib/models/grids/grid.ts
@@ -229,6 +229,14 @@ export class Grid extends GuideRenderer {
     const cmax = cross_range.max
 
     const coords: [number[][], number[][]] = [[], []]
+
+    if (!exclude_ends) {
+      if (ticks[0] != min)
+        ticks.splice(0, 0, min)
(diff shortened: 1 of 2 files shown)


bryevdv:
Trivial change only affects one call to `grid_coords` (which is now under test), merging

PR #7881, opened 2018-05-09
Image hover documentation and tests

jlstevens:
This PR follows on from #7651 to add an example, an entry in the user guide, other miscellaneous documentation (e.g a pointer from the image glyph section) and unit tests.

In the first commit a new ``tools_hover_tooltips_image.py`` example is added which looks like this:

</img>

I'm open to suggestions on how to improve this example. In particular, @bryevdv suggested that the bit mask would look better if the two colors contrasted less.

### DIFF
--- a/bokehjs/test/core/util/templating.ts
+++ b/bokehjs/test/core/util/templating.ts
@@ -113,6 +113,13 @@ describe("templating module", () => {
 
     const source = new ColumnDataSource({data: {foo: [10, 1.002], bar: ["a", "<div>b</div>"], baz: [1492890671885, 1290460671885]}})
 
+    const imsource = new ColumnDataSource({data:
+                                           {arrs: [[[0, 10, 20], [30, 40, 50]]],
+                                            floats: [[0.,1.,2.,3.,4.,5.]],
+                                            labels: ['test label']}})
+    const imindex1 = {index: 0, dim1: 2, dim2: 1, flat_index: 5}
+    const imindex2 = {index: 0, dim1: 1, dim2: 0, flat_index: 1}
+
     it("should return $ values from special_vars", () => {
       const v = tmpl.get_value("$x", source, 0, {"x": 99999})
       expect(v).to.be.equal(99999)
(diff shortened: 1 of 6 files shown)


jlstevens:
I've rebased so I can add some unit tests to ``templating.ts`` instead of ``templating.coffee`` now that #7865 has been merged.

jlstevens:
@bryevdv The new unit tests I've added here are passing locally (using ``gulp test:core``). I assume the test failures are due to recent changes to the test infrastructure?

bryevdv:
sort of, but yes you should rebase on master as of this morning

jlstevens:
The one commit missing was ' Work around various test stupidity (#7882)' so that is probably it. Anyway, I've rebased again so I expect the tests to go green this time.

jlstevens:
@bryevdv Now ready for review and further suggestions for improving the docs/tests.

bryevdv:
Is there an interesting but somewhat simple example we could add to the gallery?

jlstevens:
Would it be worth enabling hover to [this example](https://bokeh.pydata.org/en/latest/docs/gallery/image.html)?

bryevdv:
Sure, I think it would be easy to just read off the image scalar values?

jlstevens:
Here is what the updated gallery example looks like:



```python
import numpy as np

N = 500
x = np.linspace(0, 10, N)
y = np.linspace(0, 10, N)
xx, yy = np.meshgrid(x, y)
d = np.sin(xx)*np.cos(yy)

p = figure(x_range=(0, 10), y_range=(0, 10),
           tools=[HoverTool(tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")])])

p.image(image=[d], x=0, y=0, dw=10, dh=10, palette="Spectral11")

output_file("image.html", title="image.py example")

show(p)  # open a browser
```

jlstevens:
I've added a simple tornado version check to ``docserver.py`` as it hangs when using tornado 4 (which is still needed by the latest version of classic notebook). Here is what happens if you have tornado 4:

```
BOKEH_DOCS_CDN=local  GOOGLE_API_KEY="junk" make serve
python docserver.py
docserver.py script requires tornado 5 or higher
make: *** [serve] Error 1
```

Issue #7880, opened 2018-05-09
Upgrade npm packages 


mattpap:
Use `npm audit` and `npm outdated` to get feedback on which packages need to be updated and what versions are available.

bryevdv:
After #7879 `npm audit` reports some vulnerabilities, but only with respect to some dev-dependencies (no runtime dependencies):

```
(base) ❯ npm audit

                       === npm audit security report ===



# Run `npm install gulp@4.0.0` to resolve 4 vulnerabilities
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimatch                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ gulp [dev]                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ gulp > vinyl-fs > glob-stream > minimatch                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/118                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimatch                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ gulp [dev]                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ gulp > vinyl-fs > glob-watcher > gaze > globule > glob >     │
│               │ minimatch                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/118                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimatch                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ gulp [dev]                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
[...]

PR #7882, opened 2018-05-10
Work around various test stupidity

bryevdv:
ref: #7865

@mattpap this is a minimal PR that seems to get tests running again on mocha 4, I agree BokehJS unit tests need a top-to-bottom overhaul but that will have to be after this release.

### DIFF
--- a/bokeh/command/subcommands/tests/test_serve.py
+++ b/bokeh/command/subcommands/tests/test_serve.py
@@ -9,6 +9,7 @@ import sys
 
 import pytest
 import requests
+import six
 
 import bokeh.command.subcommands.serve as scserve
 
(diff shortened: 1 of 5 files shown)


bryevdv:
Ok well now some later tests are failing on Travis due to edit variables (they don't fall locally)

bryevdv:
OK changed to just restore the old mocha version. Unless you come up with some different fix I think you should merge this first thing in the morning. @mattpap

bryevdv:
FFS this is back
```
     IOError: [Errno 9] Bad file descriptor
```

bryevdv:
I guess we can skip that test on py2.7, it's not super important.

bryevdv:
@mattpap any comments or should I merge this?

mattpap:
Looks good (I mean it looks as good as it gets with the tooling at our disposal).

mattpap:
Anyway, as I mentioned in the other PR, we should try out `--exit` with mocha 5 when proceeding with issue #7880 , because it seems like it should do the right thing.

bryevdv:
I agree, and in fact I think `--exit` was the previous default, if I read the mocha issue thread correctly

PR #7889, opened 2018-05-10
Don't use encode_utf8(), keep unicode and encode when writing

### LINKED ISSUE #7885: Exporting DataTable to png broken
Using bokeh master it appears that exporting a DataTable to png is broken. This appears to be a regression in 0.12.16 as it's working in 0.12.15. Here is a minimal reproducible example:

```python
from datetime import date
from random import randint

data = dict(
    )

columns = [
    ]

export_png(data_table, filename="table.png")
```

>UnicodeEncodeError: 'latin-1' codec can't encode character '\ufffd' in position 879408: ordinal not in range(256)

### LINKED ISSUE #7901: Drop encode_utf8() and decode_utf8()
This is a follow up on issue #7885, where a particular case was fixed.

mattpap:
This is a bit excessive to fix #7885, however, I was curious how badly I will break bokeh when I remove all `encode_utf8()` calls. The only case where I needed to add `.encode("utf-8")` (in both 2 and 3) was when comparing HMAC. Otherwise it's sufficient to make sure files are open with `io.open(..., "w", encoding="utf-8")`. Lets see if tests are all green (I verified unit tests locally). This is a breaking change, I didn't update docstrings, etc. Some removed tests could be updated and restored.

fixes #7901

### DIFF
--- a/bokeh/command/subcommands/file_output.py
+++ b/bokeh/command/subcommands/file_output.py
@@ -7,8 +7,6 @@ from abc import abstractmethod
 import argparse
 import io
 
-from bokeh.util.string import decode_utf8
-
 from ..subcommand import Subcommand
 from ..util import build_single_handler_applications, die
 
@@ -129,10 +127,10 @@ class FileOutputSubcommand(Subcommand):
         '''
         contents = self.file_contents(args, doc)
         if filename == '-':
-            print(decode_utf8(contents))
+            print(contents)
         else:
             with io.open(filename, "w", encoding="utf-8") as file:
-                file.write(decode_utf8(contents))
+                file.write(contents)
         self.after_write_file(args, filename, doc)
 
     # can be overridden optionally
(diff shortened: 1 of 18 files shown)


bryevdv:
Definitely worried about breaking any users of these functions.

mattpap:
On Python 3 everything stays the same. It's Python 2 that gets the hit.

mattpap:
This is one approach. Alternatively we could go all in on returning bytes everywhere and in both Pythons. I didn't try this and I'm not sure which I would actually prefer (assuming both work reliably).

mattpap:
Also, I'm not pushing this for the current release. We can just patch the single case reported in #7885. Though in the long term, I think this may be the right thing to do.

bryevdv:
I definitely agree that long term it is the thing to do. I think returning bytes everywhere would be a mistake, and if there is path that keeps compatibility for python 3 throughout that is the path to prioritize.

mattpap:
I'm going to close this PR. Given that we will be dropping support for Python 2 in "near" future, there is no point in investing time in fixing things.

PR #7892, opened 2018-05-12
correctly export unicode content with unminified resources

### LINKED ISSUE #7885: Exporting DataTable to png broken

bryevdv:
- [x] issues: fixes #7885
- [x] tests added / passed

cc @mattpap @philippjfr

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -36,6 +36,7 @@ from tempfile import mkstemp
 from ..embed import file_html
 from ..resources import INLINE
 from ..util.dependencies import import_required, detect_phantomjs
+from ..util.string import decode_utf8
 from .util import default_filename
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bryevdv:
The new tests fail with the old code and pass with the new.

bryevdv:
FYI my best guess is that one of the BokehJS dependencies has some unicode comments that get stripped out during minification.

bryevdv:
OK well this is failing on py2 pretty hard. I thought I had run locally but now I'm guessing I accidentally used ` py.test` from a different conda env. 

The failures seems to be what the `encode_utf8` function was for: to obtain something that could be written as unicode by `write` on either python 2 or python 3... it just has a misleading name

bryevdv:
OK green now. @philippjfr please check this out if you get a chance. @mattpap any objections? This is the most minimally invasive solution I could find at this point.

mattpap:
> The failures seems to be what the decode_utf8 function was for

It works now, because you are cancelling out both transforms on py 2. Looks good. Obviously PR #7889 when done, will make all this unnecessary.

PR #7895, opened 2018-05-14
fix typos

foobarbecue:
containtain --> contained
embeeded --> embedded

Issue #7896


bryevdv:
Thanks for the catches and fixes @foobarbecue !

PR #7907, opened 2018-05-15
Fix release month for 0.12.16 in docs

nasedil:
Does it makes sense to change Apr to May for 0.12.16 release?  Because the final version was released on 14 of May.

And should I create an issue for such minor change?

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


mattpap:
> And should I create an issue for such minor change?

It's preferred to have an issue for a PR, but in such case we can skip that. Anyway, thanks for the fix.

PR #7910, opened 2018-05-15
Bryanv/test build cleanup

bryevdv:
WIP work toward #7880

cc @mattpap just FYI


bokehjs/package.json
```diff
@@ -24,96 +24,97 @@
...
+    "merge2": "^1.2.2",
     "logtrap": "^0.0.1",
     "mocha": "^2.2.5",
-    "run-sequence": "^1.0.0",
+    "run-sequence": "^2.2.1",
     "sinon": "^1.15.4",
-    "vinyl-source-stream": "^1.1.0",
-    "yargs": "^3.11.0",
+    "vinyl-source-stream": "^2.0.0",
+    "yargs": "^11.0.0",
```
mattpap: I recently tried to upgrade yargs, and it completely broke compiler's build. Here it just works.

mattpap:
> chai                        3.5.0   3.5.0    4.1.2  bokehjs
> chai-as-promised            5.3.0   5.3.0    7.1.1  bokehjs

Minor test failures.

> jsdom                      9.12.0  9.12.0  11.10.0  bokehjs

Requires major change in how we initialize and interact with jsdom.

> mocha                       2.5.3   2.5.3    5.1.1  bokehjs

`--exit` should be sufficient to make it work.

> nouislider                 10.1.0  10.1.0   11.1.0  bokehjs

Just doesn't work.

> proj4                      2.3.17  2.3.17    2.4.4  bokehjs

Requires changes to build our pipeline. Alternatively we could request proj4/dist to contain all compilation products, not just js bundles.

> sinon                      1.17.7  1.17.7    5.0.7  bokehjs

A lot of API changes. `sinon.test` was split off into a separate package.

bokehjs/package.json
mattpap: After all it's not a mystery. I missed that you added `spawn-sync` to dependencies in "update flatbush". Reordering of items in `dependencies` didn't help to catch this, especially given that `spawn-sync` isn't mentioned anywhere in this PR. Will later inspect why this is needed.

PR #7912, opened 2018-05-15
Make bokehjs' unit tests more robust

### LINKED ISSUE #7911: Make bokehjs' unit tests more robust

mattpap:
fixes #7911


mattpap:
This PR removes the bulk of the stubs. There are still some particular ones that will have to be evaluated on case by case basis.

Issue #7908, opened 2018-05-15
Tile source attribution font size


jbednar:
Would it be possible to reduce the default font size on the [attributions for tile sources](https://github.com/bokeh/bokeh/blob/a0572fa40c3a59b72ba56a8c300e0ad06a044a23/bokeh/models/tiles.py)?

![image](https://user-images.githubusercontent.com/1695496/40062037-e05bceca-581f-11e8-87b6-75dac06cfc3f.png)

Seems like it just needs to be barely readable, while the current size is overwhelming for smaller plots.  Maybe 30% of the tick label size would work?

bryevdv:
30% of the *default* tick label size? I don't think the font should vary based on user-configured tick label size. In any case sure, happy to merge a PR.

jbednar:
Looks like it's set to 9pt right now:

https://github.com/bokeh/bokeh/blob/a0572fa40c3a59b72ba56a8c300e0ad06a044a23/bokehjs/src/coffee/models/tiles/tile_renderer.coffee#L50

If that's 9pt, then maybe 4pt?

bryevdv:
I don't really have a strong opinion here, so am happy to change to whatever works for you.

PR #7914, opened 2018-05-17
Made tile attributions less distracting

jbednar:
As described in https://github.com/bokeh/bokeh/issues/7908, the attribution text for tile sources is currently quite distractingly large, particularly for small plots:

image

After applying this PR, the text should still be readable, but is less distracting:

image

Open issue: Is there a way to avoid coloring the link text blue?  It's very difficult to read in certain cases (see the black and white Stamen example above), and it's hard to make a background color that works well for both black and bright blue text when combined with various elements that might be in the map tiles.  I can see how to change any CSS elements that apply to the general text here, but not for link anchors.

### DIFF
--- a/bokehjs/src/less/main.less
+++ b/bokehjs/src/less/main.less
@@ -3,3 +3,7 @@
   display: block;
   border: 1px dashed green;
 }
+
+.bk-tile-attribution a {
+  color: black;
+}
--- a/bokehjs/src/lib/models/tiles/tile_renderer.ts
+++ b/bokehjs/src/lib/models/tiles/tile_renderer.ts
@@ -100,9 +100,10 @@ export class TileRendererView extends RendererView {
             right: `${right}px`,
             'max-width': `${max_width}px`,
             padding: "2px",
-            'background-color': 'rgba(255,255,255,0.8)',
-            'font-size': '9pt',
+            'background-color': 'rgba(255,255,255,0.5)',
+            'font-size': '7pt',
             'font-family': 'sans-serif',
+            'line-height': '1.05',
           },
         })
 
--- a/bokehjs/src/lib/models/tiles/tile_renderer.ts
+++ b/bokehjs/src/lib/models/tiles/tile_renderer.ts
@@ -85,10 +85,10 @@ export class TileRendererView extends RendererView {
   }
 
   protected _add_attribution(): void {
-    const { attribution } = this.model.tile_source
+    const {attribution} = this.model.tile_source
 
-    if (isString(attribution) && (attribution.length > 0)) {
-      if ((this.attributionEl == null)) {
+    if (isString(attribution) && attribution.length > 0) {
+      if (this.attributionEl == null) {
         const right = this.plot_model.canvas._right.value - this.plot_model.frame._right.value
         const bottom = this.plot_model.canvas._bottom.value - this.plot_model.frame._bottom.value
         const max_width = this.map_frame._width.value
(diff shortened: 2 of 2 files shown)


jbednar:
@mattpap, any idea how to control link text color within the attribution?

mattpap:
You will have to add `.bk-tile-attribution * { color: black; }` (or replace `*` with `a`, if you want to target only links) to some file under `src/lib/less` (`main.less` seems fine for this).

jbednar:
Perfect, thanks!  Much more readable now, and no longer distracting in most cases:

image

Ready for review/merge.

mattpap:
@jbednar, please rebase on top of `master` to resolve test failures. I will cancel the ongoing build.

mattpap:
Maybe we should allow only one line for attribution, and if it doesn't fit, then show full content in a tooltip?

bryevdv:
> Maybe we should allow only one line for attribution, and if it doesn't fit, then show full content in a tooltip?

I expect most real usage of maps will have bigger maps than those above. But in any case it's unclear that that a tooltip would satisfy all terms of service (possible they dictate that attributions be visible at all times). I'd rather just do the safe thing that also requires no additional work.

jbednar:
Personally, I'd vote for @mattpap's suggestion, because that's what Esri does: 

https://server.arcgisonline.com/ArcGIS/rest/services/NatGeo_World_Map/MapServer?f=jsapi

I think Esri sets the standard in this area.  It's rare that maps are this small, and also rare that attributions are this long, but it would be nice for it not to matter when that happens...

bryevdv:
Is there some fancy CSS way to do this without any code? I'd be much more amenable if so.

jbednar:
Looks great!

image

I'm very happy with this now.

mattpap:
@jbednar, what browser do you use? For me, on Chrome and Firefox, those attributions are much smaller and with less padding.

jbednar:
I'm using Chrome, and that example is using GeoViews master with https://github.com/ioam/geoviews/blob/master/examples/gallery/bokeh/tile_sources.ipynb

I tried running the Bokeh tile sources example before making any changes to Bokeh, but always got an html file that showed no output in my browser, so I gave up on that.

mattpap:
> Note that my original commit had 'line-height': '1.05',, which reduces the padding,

I removed it because it doesn't do anything on linux, so I assumed this was a leftover. Note that padding is controlled by `padding: 2px`. It would be good to understand what is going on here. What version of Chrome do you use?

jbednar:
The defaults already conflate line height and padding, in that those two add up to determine the observed spacing.  You could potentially separate those two concerns by using line height to determine the top and bottom padding, then make the padding specification only apply padding to the left and right of the text, leaving `padding` controlling the spacing to the sides, and `line-height` controlling spacing above and below.

I don't think the other approach to making these settings independent would work, i.e. to set `line-height` to 1.0 and then control the spacing entirely using `padding`, because then in the hover there would be no space between multiple lines (unless there were yet again different settings for the hover).  

In any case, I don't think there's much danger in the previous approach (small `padding` along with a `line-height` of 1.05, or the equivalent expressed in pixels), but you may find a better approach.

jbednar:
I much preferred the line spacing tight, the way I had it originally:
 
image

to the wide padding that got merged:

image

I found the tighter spacing much less distracting, and more similar to how Esri and Google Maps do it:

image

@mattpap was concerned about how the tighter spacing was achieved, but I don't think doing it that way was worse than having it vary by browser and platform, so it seems unequivocally better to specify it more precisely and repeatably in this way.  Ok if I make a tiny (one line) follow-on PR, unless Mateusz is already doing that?

mattpap:
I restored the initial `line-height` in this PR.

PR #7918, opened 2018-05-17
update to gulp 4

bryevdv:
- [x] tests added / passed

@mattpap I know I said I'd hold off on this, but the changes were minimal, and get audit  results down to:

```
[!] 4 vulnerabilities found [7118 packages audited]
    Severity: 1 Low | 3 Moderate
    Run `npm audit` for more detail
```

Additionally I'd like to consider the custom build question more. There are definitely benefits to owning everything, but also drawbacks, namely the ability to attract and integrate other JS contributors if everything we do is unique and bespoke.


mattpap:
`run-sequence` and `@types/run-sequences` could be removed from dependencies.

bryevdv:
Ah yes I mean to and forgot, I will push that now.

bryevdv:
not sure why the one codebase test is failing:

https://travis-ci.org/bokeh/bokeh/jobs/380352779

That command works fine when run locally

bryevdv:
> It fails for me as well. There's a conflict between type declarations. I will resolve it. 

OK my last commit can probably be reverted then

mattpap:
> OK my last commit can probably be reverted then

I presume this deals with `anonymous` tasks? If so, it should stay. In fact, there are more cases to fix.

bryevdv:
> I presume this deals with anonymous tasks? If so, it should stay. In fact, there are more cases to fix.

IIUC this actually *makes* a new anonymous task. There was another issue that required that kind of change to fix an "async completion" error, so I was giving it a shot.

mattpap:
There is another problem with this upgrade. gulp 4 doesn't support task dependencies. Previously if multiple tasks depended on another task, then that task was run only once. Now it will run as many times as is depended on. I didn't yet check if we are affected or not.

bryevdv:
closing, given #7920

PR #7920, opened 2018-05-18
Replace gulp with a more lightweight solution

### LINKED ISSUE #7919: Replace gulp with a more lightweight solution

mattpap:
This drops gulp, all its plugins and a few other modules, reducing the size of our dependency tree by 2/3.

To run the build at this stage issue:
```sh
node make build # or other task name
```
To preserve the old workflow, e.g. in bash, you can add:
```
alias gulp='node make'
```
to `~/.bash_aliases`.

Issues to resolve:

- [ ] ~~make `gulp task-name` still work (assuming gulp is installed with `npm install -g gulp`)~~

This won't work, because even if gulp (or to be specific gulp-cli) is installed globally, it still requires 
`gulp` to be installed locally, which defeats the purpose of this PR. One can make an alias to `node gulpfile.js` to get around this.

- [x] rename `gulpfile.js` and `gulp/` to make it clear we don't use gulp anymore

`gulpfile.js` is gone and `gulp/` was renamed to `make/`.

- [ ] fix source map support for `*.min.js` files

This will be fixed in PR #7851

- [ ] ~~restore support for watch tasks~~

fixes #7919


mattpap:
For `git bisect`, the most universal way to build bokehjs is to use `npm run build`. This runs `gulp build` before this PR and `node make build` since this PR. This command was added in 2016 and covers entire 0.12.x line.

sphinx/source/docs/dev_guide/testing.rst
```diff
@@ -355,7 +351,7 @@ separate ``test:<GROUP>`` script for each of the following test groups:
     a report is uploaded to see the detailed results.
 
 ``js``
-    This job runs all the JavaScript unit tests (i.e. ``gulp test``)
+    This job runs all the JavaScript unit tests (i.e. ``node make test``)
```
bryevdv: I think the setup section of the dev guide also references gulp

bryevdv: I thought there was mention of it in the install section, but apparently I was mistaken.

PR #7921, opened 2018-05-20
Bryanv/usability

bryevdv:
@pzwang suggested trying to get more in the "user mind" this PR is an attempt to make some common things easier than they currently are. In particular:

* add `$name` special variable to shoe glyph renderer `name`
* adds `tooltips` convenience arg to `Figure`
* adds `Plot.hover` accessor

all of which make simple hover tool cases much easier to get up and running. Also update various examples to pass dataframes or dicts directly as `source` where possible.  

I'd like to try and solve at least one more issue, which is there is not a good way to add a single hover tool to a `vbar_stack` etc that covers the stacked quantities. 

cc @mattpap for any early thoughts. 

- [x] tests added / passed

### DIFF
--- a/bokeh/plotting/figure.py
+++ b/bokeh/plotting/figure.py
@@ -822,6 +822,10 @@ Examples:
             stackers (seq[str]) : a list of data source field names to stack
                 successively for ``left`` and ``right`` bar coordinates.
 
+                Additionally, the ``name`` of the renderer will be set to
+                the value of each successive stacker (this is useful with the
+                special hover variable ``$name``)
+
         Any additional keyword arguments are passed to each call to ``hbar``.
         If a keyword value is a list or tuple, then each call will get one
         value from the sequence.
@@ -843,8 +847,8 @@ Examples:
 
             .. code-block:: python
 
-                p.hbar(bottom=stack(),       top=stack('2016'),         x=10, width=0.9, color='blue', source=source)
-                p.hbar(bottom=stack('2016'), top=stack('2016', '2017'), x=10, width=0.9, color='red', source=source)
+                p.hbar(bottom=stack(),       top=stack('2016'),         x=10, width=0.9, color='blue', source=source, name='2016')
+                p.hbar(bottom=stack('2016'), top=stack('2016', '2017'), x=10, width=0.9, color='red',  source=source, name='2017')
 
         '''
         result = []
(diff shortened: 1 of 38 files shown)


bryevdv:
Oh, I also disabled `_repr_pretty` it made the console `ipython` unusable.

bryevdv:
OK, I found a slight special-casey solution, but it is trivial to implement, and reasonable to explain, which is to add support for `@$name` specifically, with the meaning `@` (lookup a column) but use special var `$name` for getting the column name. This lets a single hover tool work great for stacked bar charts:

```
from bokeh.core.properties import value

output_file("stacked.html")

fruits = ['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries']
years = ["2015", "2016", "2017"]
colors = ["#c9d9d3", "#718dbf", "#e84d60"]

data = {'fruits' : fruits,
        '2015'   : [2, 1, 4, 3, 2, 4],
        '2016'   : [5, 3, 4, 2, 4, 6],
        '2017'   : [3, 2, 4, 4, 5, 3]}

p = figure(x_range=fruits, plot_height=250, title="Fruit Counts by Year",
           toolbar_location=None, tools="hover", tooltips="$name @fruits: @$name")

p.vbar_stack(years, x='fruits', width=0.9, color=colors, source=data,
             legend=[value(x) for x in years])

p.y_range.start = 0
p.x_range.range_padding = 0.1
p.axis.minor_tick_line_color = None
p.outline_line_color = None
p.legend.orientation = "horizontal"

show(p)
```
screen shot 2018-05-20 at 17 49 25

bryevdv:
I put `_repr_pretty_` back for now, but we have to do something about it. IPython console is currently unusable due to super long reprs spamming the console.

mattpap:
> I put _repr_pretty_ back for now, but we have to do something about it.

This behavior isn't unexpected. We just need to revisit https://github.com/bokeh/bokeh/issues/7729#issuecomment-378247775.

bryevdv:
@mattpap this is `ready`, I will plan to merge later today unless you indicate you want additional time to comment.

bokeh/plotting/figure.py
```diff
@@ -83,6 +83,10 @@ class FigureOptions(Options):
     The type of the y-axis.
     """)
 
+    tooltips = Either(String, List(Tuple(String, String)), help="""
+
```
mattpap: Missing tooltip.

mattpap:
Changes to bokeh's plotting API should be mirrored in bokehjs. Though given the bokehjs API is behind anyway, this could be done in another PR, filling in all the gaps.

bokeh/plotting/figure.py
bryevdv: Thanks, I have added it and will merge when green.

PR #7938, opened 2018-05-27
Addresses #7937 Add support for ajax data source in TableWidget

### LINKED ISSUE #7937: Add support in TableWidget for AjaxDataSource
bokeh 0.12.16, python 3.6.5, linux, chrome

plugging source=AjaxDataSource... in DataTable does not populate the table with query result

In [ajax_source.py](https://github.com/bokeh/bokeh/blob/master/examples/howto/ajax_source.py) replace lines 15-20 and 27 with creating table 
```
from bokeh.models.widgets import DataTable, TableColumn
colx = TableColumn(field="x", title="x")
coly = TableColumn(field="y", title="y")
table = DataTable(source=source, columns=[colx, coly])

#line 27
show(table)
```
Here is the full code in a [gist](https://gist.github.com/2torus/36605e7c2e7a6ef672e4b345ad2c88d7)

No data:
image

2torus:
Invokes source.setup() if the source has such a method, similar to [GlyphRenderer.ts#106](https://github.com/bokeh/bokeh/blob/965989c1335d64ea9742560c15feabfe72f6c1d9/bokehjs/src/lib/models/renderers/glyph_renderer.ts#L106)

issues: fixes #7937


bryevdv:
@2torus I believe you should be acting on `this.source`(which there should always be) and not `this.view.source` otherwise this won't work if a view is not defined.

bryevdv:
@2torus do you think you can try the idea suggested by @mattpap above?

2torus:
@bryevdv @mattpap I created setup as an abstract method of RemoteDataSource. My thought is that anybody who wants to RemoteDataSource must implement setup().

bryevdv:
thanks @2torus !

PR #7942, opened 2018-05-29
add a private _clear_extensions method that the docs can use for now

### LINKED ISSUE #7905: All gallery examples bundle unnecessary extension code
All the gallery examples have the code for ion range slider, surface3d, etc. Perhaps we will need to run every export for the gallery in a separate process to avoid this. Alternatively, is there an easy way to clear out any custom models that the sphinx extension could do between exports?

cc @mattpap

bryevdv:
- [x] issues: fixes #7905
- [ ] tests added / passed

@mattpap if you think this (or something like it) should be made public please let me know and I will add tests and docs. Otherwise I think this is just a stopgap measure until something more comprehensive can be undertaken.

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -561,6 +561,13 @@ class Model(with_metaclass(MetaModel, HasProps, PropertyCallbackManager, EventCa
         self._document = doc
         self._update_event_callbacks()
 
+    @staticmethod
+    def _clear_extensions():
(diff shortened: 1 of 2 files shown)


mattpap:
> if you think this (or something like it) should be made public (...)

Lets keep it as it is in this PR for now.

Issue #7936, opened 2018-05-26
ResetTool doesn't restore initial plot when using WMTSTileSource and x/y_range


lgasnier:
#### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)
bokeh 0.12.16
python 3.6.4
OS linux
browser Firefox, Chrome

#### Description of expected behavior and the observed behavior
ResetTool doesn't restore initial plot when using WMTSTileSource and x/y_range.
Therefore, the match aspect is not preserved.

#### Complete, minimal, self-contained example code that reproduces the issue

```
from bokeh.plotting import figure
from bokeh.models.tiles import WMTSTileSource
from bokeh.io import show
from bokeh.io import output_file

output_file("test_tilesource_reset.html")

xmin = 200000
xmax = 200000 + 1000000
ymin = 5000000
ymax = 5000000 + 1000000

p = figure(width=500, height=700,
           x_range=(xmin, xmax),
           y_range=(ymin, ymax),
           tools="pan,wheel_zoom,reset")
p.axis.visible = False

wmts = WMTSTileSource(url="http://c.tile.openstreetmap.org/{Z}/{X}/{Y}.png")
p.add_tile(wmts)

p.quad(top=[ymax], bottom=[ymin], left=[xmin],
       right=[xmax], color="#B3DE69", alpha=0.5)
show(p)
```

philippjfr:
Thanks for filing this, this was on my to-do list to file as well. The cause of this is that the tile source enforces an equal aspect ratio and if your initial extent does not match it will reset to the originally supplied extent not the extent that the tile source set. I would argue that the tile source should update the initial extents to avoid this issue, but @bryevdv should chime in too.

bryevdv:
Sounds good to me if that's what you'd like to implement.

philippjfr:
>Sounds good to me if that's what you'd like to implement.

Okay great, I'll do that.

PR #7948, opened 2018-06-02
Exposed Range1d initial start/end properties

philippjfr:
This PR does two things, first of all it exposes the ``initial_start`` and ``initial_end`` attributes as properties on the ``Range1d`` model. Secondly it ensures that tile sources override these properties since they sometimes adjust the initial ranges on initialization to enforce the correct aspect ratio.

Setting the initial ranges is generally useful since you often want to toggle between different views and currently the reset tool is not very useful in these scenarios. The remaining question is whether to expose these properties on the ``DataRange1d`` model as well, I'm leaning towards not exposing them but I'm not entirely sure.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/6320 and https://github.com/bokeh/bokeh/issues/7936
- [x] tests added / passed


bryevdv:
They should be called `reset_start` in order to avoid confusion. For users, `initial_start` sounds like something that makes something happen when the plot is first rendered. What about my other concerns in https://github.com/bokeh/bokeh/pull/6321 ? How does this interact with bounds, following, or data ranges?

philippjfr:
>They should be called reset_start in order to avoid confusion.

Sounds good.

>How does this interact with bounds, following, or data ranges?

I assume they would have to be validated against the bounds, I'll make sure to add that. There is currently no interaction with following or data ranges since this is currently not implemented for ``DataRange1d``, would be happy to add that if you think it's a good idea though. I'm not entirely sure what the correct behavior would be though, I would guess it should update the values for both ``following`` and ``data_range`` options.

bryevdv:
Let's leave it off data ranges for now

philippjfr:
>Let's leave it off data ranges for now

To be clear, do you mean for me to implement it for ``DataRange1d`` but not update the ``reset_start`` and ``reset_end`` on data_range changes or do you mean that we shouldn't implement it for ``DataRange1d`` at all for now?

bryevdv:
Don't implement anything for data ranges

philippjfr:
I think there is no special handling needed for the ``bounds``, as far as I could tell the ranges ``start`` and ``end`` values are only clipped on the plot canvas so the clipping would still apply.

bryevdv:
> I think there is no special handling needed for the bounds, as far as I could tell the ranges start and end values are only clipped on the plot canvas so the clipping would still apply.

Fair enough, I was thinking that might be confusing to people but if thats already the precedent we should just go with that. It might do to make an explicit note of that in the properties help strings.

I think this change is also amenable to some unit JS unit tests.

philippjfr:
I've added a very simple unit test. I was hoping to add more, e.g. for the tile source update to the range, but I'm having quite some trouble setting that test up, because I have to create a plot, plot view, plot canvas, some ranges and more to even set it up. I also noticed that very few tests do that kind of thing. Would it make sense to have some testing utilities which create a fully configured plot instance for you?

bryevdv:
> Would it make sense to have some testing utilities which create a fully configured plot instance for you?

Probably, and it's probably partly done in various places. We don't have to do that here, my main interest now was in testing `Range1d` which can be accomplished easily. It's maybe worth adding another test that makes sure that changes to those properties after init also have the desired effect. 

For the general issue what we should really do is factor out "view models" so that views that require lots of logic or state can be tested more thoroughly by testing the view model (which would be creatable and testable in isolation without a ton of scaffolding)

philippjfr:
>For the general issue what we should really do is factor out "view models" so that views that require lots of logic or state can be tested more thoroughly by testing the view model (which would be creatable and testable in isolation without a ton of scaffolding)

This sounds great, although it seems quite difficult. Definitely would make a difference though, I just couldn't figure setting up the views correctly at all.

>It's maybe worth adding another test that makes sure that changes to those properties after init also have the desired effect.

Will do.

bryevdv:
> This sounds great, although it seems quite difficult. Definitely would make a difference though, I just couldn't figure setting up the views correctly at all.

I expect it's mostly just a large amount of work as opposed to being difficult work, in and of itself. But really it's something we have to get to at some point, BokehJS is not nearly well tested enough

PR #7956, opened 2018-06-05
Changed default line_join from miter to bevel

### LINKED ISSUE #7951: Line plots jaggy when overplotting noisy data with `line_join='miter'`
## versions
- bokeh: `0.12.16`
- python: 
  ```
  3.6.5 (default, Apr  5 2018, 23:36:52) 
  [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
  ```
- OS: `Darwin Kernel Version 17.5.0: Fri Apr 13 19:32:32 PDT 2018; root:xnu-4570.51.2~1/RELEASE_X86_64`
- browser: Chrome `Version 66.0.3359.181 (Official Build) (64-bit)`

## Description of expected behavior and the observed behavior

When plotting noisy time series with low screen resolution but high sampling rate (overplotting), I get a sort of "noise amplification" effect, which is apparently linked to the way in which `line_join='miter'` is implemented.

## Complete, minimal, self-contained example code that reproduces the issue

```

output_notebook(verbose=True, notebook_type='jupyter')
import numpy as np

fs = 10**4
x = np.linspace(0, 1, num=fs)
# signal
y = x**2
# gaussian noise
wgn = np.random.randn(*x.shape)
# noisy signal, should be visually indistinguibile from original one
a = 1e-3
yn = y + a*wgn

p1 = figure(title="miter",plot_width=450, plot_height=250)
p2 = figure(title="bevel",plot_width=450, plot_height=250,
            x_range=p1.x_range, y_range=p1.y_range,)

p1.line(x, yn, line_join='miter')
p1.line(x, y, color='red')
p2.line(x, yn, line_join='bevel')
p2.line(x, y, color='red')

grid = gridplot([[p1,p2],])
show(grid)
```

I get the following images:

bokeh_plot
bokeh_plot 1

Here you can see that with `line_join='miter'` the noise appears much greater as it really is: the correct representation is obtained with `line_join='bevel'`. This effect disappears when zooming in:

bokeh_plot 2
bokeh_plot 3

## Discussion

I know that the correct way of representing data with high sample rate with respect to screen resolution should be via datashader and friends, however sometimes a simpler approach is more desirable (here I have about 20 data points per pixel, which should be acceptable, imho).

The bokeh default (`line_join='miter'`) gives a visually incorrect result, which was for me at a first glance incomprehensible until I traced it down to the miter joint thing.

I'm not sure if this is a bug or a feature (high frequency noise makes the miter joint overshoot the data envelope) but the result is surprising. May be the `line_join='bevel'` or `line_join='round'` should be the default.

jbednar:
- [x] issues: fixes #7951

In every case where I can detect a difference from the current value, the change is an improvement.  E.g. from a [TriMesh example](https://holoviews.org/reference/elements/bokeh/TriMesh.html), before:
image
and after:
image
(see all the spurious corners along the top edge of the image above, and how they disappear on the bottom image.)
Thanks to @bryevdv for the code.

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -467,7 +467,7 @@ class Serve(Subcommand):
         ('--num-procs', dict(
             metavar='N',
             action='store',
-            help="Number of worker processes for an app. Default to one. Using "
+            help="Number of worker processes for an app. Using "
                  "0 will autodetect number of cores (defaults to 1)",
             default=1,
             type=int,
--- a/bokeh/command/subcommands/tests/test_serve.py
+++ b/bokeh/command/subcommands/tests/test_serve.py
@@ -167,7 +167,7 @@ def test_args():
         ('--num-procs', dict(
              metavar='N',
              action='store',
-             help="Number of worker processes for an app. Default to one. Using "
+             help="Number of worker processes for an app. Using "
                   "0 will autodetect number of cores (defaults to 1)",
              default=1,
              type=int,
--- a/bokeh/core/property_mixins.py
+++ b/bokeh/core/property_mixins.py
@@ -98,7 +98,7 @@ Stroke width in units of pixels.
 """
 
 class BaseLineProps(HasProps):
-    line_join = Enum(LineJoin, help="""
+    line_join = Enum(LineJoin, default='bevel', help="""
     How path segments should be joined together.
 
     Acceptable values are:
--- a/bokeh/models/tests/utils/property_utils.py
+++ b/bokeh/models/tests/utils/property_utils.py
@@ -33,7 +33,7 @@ def check_line_properties(model, prefix="", line_color=Color.black, line_width=1
     assert getattr(model, prefix + "line_color") == line_color
     assert getattr(model, prefix + "line_width") == line_width
     assert getattr(model, prefix + "line_alpha") == line_alpha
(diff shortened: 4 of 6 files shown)


bryevdv:
@jbednar sorry I forget out the test here:
```
bokeh/models/tests/utils/property_utils.py:
   36 :     assert getattr(model, prefix + "line_join") == LineJoin.miter
```

Change that to `LineJojn.bevel` and it should pass. I would recommend a subsection in the Migration notes as well at
```
./sphinx/source/docs/releases/0.13.0.rst
```

jbednar:
I've changed the test, thanks.  I'm confused about the migration notes; do you mean an additional change beyond what is already in this PR?

bryevdv:
Ah sorry I missed it, but I would change the header. The whole "migration guide" super section is basically deprecations and compatibility, so I'd be specific and title this subsection "Line Join Default" or something.

jbednar:
@bryevdv, the tests are passing now, but I don't know where to find the image diffs to check on those.

bryevdv:
The URL is at the end of the first test, e.f. 

https://s3.amazonaws.com/bokeh-travis/388077384/examples.html

I will merge this when I get back

jbednar:
There are a few diffs that seem relevant, e.g.:
image
image
image
But there are various other diffs that seem unrelated to this, in that they show changes not just at line joins; not sure what those are about.

PR #7961, opened 2018-06-05
add a cumsum transform to cumulatively sum a single column

bryevdv:
I ran across the SO question **[Create pie-chart using bokeh with dictionary
](https://stackoverflow.com/questions/50697847/create-pie-chart-using-bokeh-with-dictionary)** and thought I woudl see if it could be done with `Stack` but soon realized that it could not, since `Stack` stacks multiple entire columns element-wise. It occured to me that a cummulative sum was a very useful but missing piece that would also be trivial to implement, so I threw together this PR. 

I will add docs and tests in subsequent commits.


bryevdv:
With this PR the SO question can be achieved much more cleanly with:
```

from collections import Counter
from math import pi

import pandas as pd

from bokeh.palettes import Category20c
from bokeh.transform import cumsum

x = Counter({
    'United States': 157,
    'United Kingdom': 93,
    'Japan': 89,
    'China': 63,
    'Germany': 44,
    'India': 42,
    'Italy': 40,
    'Australia': 35,
    'Brazil': 32,
    'France': 31,
    'Taiwan': 31,
    'Spain': 29
})

data = pd.DataFrame.from_dict(dict(x), orient='index').reset_index().rename(index=str, columns={0:'value', 'index':'country'})
data['angle'] = data['value']/sum(x.values()) * 2*pi
data['color'] = Category20c[len(x)]

p = figure(plot_height=350, title="Pie Chart", toolbar_location=None,
           tools="hover", tooltips=[("Country", "@country"),("Value", "@value")])

p.wedge(x=0, y=1, radius=0.4, 
        start_angle=cumsum('angle', include_zero=True), end_angle=cumsum('angle'),
        line_color="white", fill_color='color', legend='country', source=data)

p.axis.axis_label=None
p.axis.visible=False
p.grid.grid_line_color = None

show(p)
```

screen shot 2018-06-05 at 11 31 30

bokehjs/src/lib/models/expressions/cumsum.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  _v_compute(source: ColumnarDataSource): Arrayable<number> {
+    const result = new Float64Array(source.get_length() || 0)
+    const col = source.data[this.field]
+    const offset = this.include_zero ? 1 : 0
+    if (this.include_zero)
+        result[0] = 0
+    else
+        result[0] = col[0]
+    debugger
```
mattpap: ^^^

bokeh/models/expressions.py
```diff
@@ -46,6 +46,34 @@ class Expression(Model):
...
+    Whether to include zero at the start of the result. Note that the length
+    of the result is always the same as the input column. Therefore if this
+    property is True, then the last value of the column will not be included
+    in the sum.
+
+    .. code-block:: python
+
+        source = ColumnDataSource(data=dict(foo=[1, 2, 3, 4]))
+
+        CumSum('foo')
```
philippjfr: Do ``Expressions`` actually allow positional arguments? Might be the case but maybe you confused it with the cumsum helper function signature?

bokehjs/src/lib/models/expressions/cumsum.ts
bryevdv: We should really just make a code quality check that looks for this

bokehjs/src/lib/models/expressions/cumsum.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  _v_compute(source: ColumnarDataSource): Arrayable<number> {
+    const result = new Float64Array(source.get_length() || 0)
+    const col = source.data[this.field]
+    const offset = this.include_zero ? 1 : 0
+    if (this.include_zero)
+        result[0] = 0
+    else
+        result[0] = col[0]
```
mattpap: This could have been reduced to one line as on the line before.

bokeh/models/expressions.py
mattpap: No, this code is incorrect.

mattpap: Also, it would be good to use actual Python shell syntax (`>>>`).

bokehjs/src/lib/models/expressions/cumsum.ts
mattpap: Sure, there's a built-in rule for this (https://palantir.github.io/tslint/rules/no-debugger/).

bokehjs/src/lib/models/expressions/cumsum.ts
bryevdv: fixed in https://github.com/bokeh/bokeh/pull/7961/commits/0c3eaa9beabbab60dbd1d7d99571f2c7b7cbb1fb

bokeh/models/expressions.py
bryevdv: fixed in https://github.com/bokeh/bokeh/pull/7961/commits/7f2a0dc151138aec7add74bebf5249a1d40740bb

bokehjs/src/lib/models/expressions/cumsum.ts
bryevdv: fixed in https://github.com/bokeh/bokeh/pull/7961/commits/c59175147ff6c61c95692f1ac13223b61a7161b6

mattpap: If your are at fixing things, you can add `"no-debugger": true` to `bokehjs/tslint.json`.

bryevdv: I could try to add it to `quality` but it doesn't currently actually fail when there is a violation:
```
~/work/bokeh/bokehjs bryanv/cumsum_transform*
(base) ❯ node make tslint
[13:32:35] Starting 'tslint:lib'...
[13:32:36] /Users/bryanv/work/bokeh/bokehjs/src/lib/models/expressions/cumsum.ts:35:5
[13:32:36] ERROR: 35:5  no-debugger  Use of debugger statements is forbidden
[13:32:36] Finished 'tslint:lib' after 1.34 s
[13:32:36] Starting 'tslint:test'...
[13:32:36] Finished 'tslint:test' after 116 ms
[13:32:36] Starting 'tslint:examples'...
[13:32:36] Finished 'tslint:examples' after 32 ms
[13:32:36] Finished 'tslint'

~/work/bokeh/bokehjs bryanv/cumsum_transform*
(base) ❯ echo $?
0
```

mattpap: It a part of `codebase` job on travis ci. Otherwise you have to run `node make tslint`. Perhaps we could make it part of `quality` tests at some point.

mattpap: >  but it doesn't currently actually fail when there is a violation:

It was supposed to get `--emit-error`, the same as `scripts:compile` and `test:compile` have. Must have forgotten about this or maybe there is a regression since then new build is in place.

bryevdv: OK I added it, tho as I said it's not clear to me that a linter failure results in a test failure since the exit always seems to be 0

bryevdv: I tried adding `--emit-error` locally, that still did not result in a nonzero exit code when a lint violation was present

mattpap: In fact this was done on purpose, because initially tslint reported a lot of errors and it was tedious dealing with this. Now all were resolved, so failing `tslint` build task on error should be implemented. I will do it with the next round of build improvements.

PR #7968, opened 2018-06-06
Typo fixed in image w parameter (height instead of width)

### LINKED ISSUE #7967: Typo in the image_url documentation
In the `image_url` documentation, both the `h` and `w` parameters are documented as the `height` where the image will be located. For the `w` parameter should be the `width`.

datapythonista:
- [X] issues: fixes #7967
- [x] tests added / passed


bryevdv:
Thanks for the catch and correction @datapythonista

PR #7975, opened 2018-06-08
cache jitter for identical input lengths

### LINKED ISSUE #7915: re-jitter on tap is unexpected
This may be intended behavior. But if so, it would be good to provide documentation of a workaround because it surprised me, and I would think others too

This only a slightly tweaked version of the categorical jitter example in gallery, that adds a tap tool.

You can also see it here: http://nbviewer.jupyter.org/github/birdsarah/bokeh-miscellany/blob/master/jitter%20on%20tap.ipynb

```python
from bokeh.sampledata.commits import data
from bokeh.transform import jitter

output_notebook()

DAYS = ['Sun', 'Sat', 'Fri', 'Thu', 'Wed', 'Tue', 'Mon']

p = figure(plot_width=800, plot_height=300, y_range=DAYS, x_axis_type='datetime', tools='tap, save',
           title="Commits by Time of Day (US/Central) 2012-2016")

p.circle(x='time', y=jitter('day', width=0.6, range=p.y_range),  source=source, alpha=0.3, 
         selection_color='magenta', selection_alpha=1, size=8)

p.xaxis[0].formatter.days = ['%Hh']
p.x_range.range_padding = 0
p.ygrid.grid_line_color = None

show(p)
```

On tap, I expect the point under my mouse to highlight.

What happens is that a repaint occurs causing a fresh jitter so the higlighted point has moved. This is particularly clear if you tap on two points that were on top of each other but have now moved

bokeh_plot_two

Maggie-M:
issues: fixes #7915

### DIFF
--- a/bokehjs/src/lib/models/transforms/jitter.ts
+++ b/bokehjs/src/lib/models/transforms/jitter.ts
@@ -13,7 +13,7 @@ export namespace Jitter {
     width: number
     distribution: Distribution
     range: Range
-    previous_values: Float64Array
+    previous_values: Arrayable<number>
   }
 
   export interface Props extends Transform.Props {}
--- a/bokehjs/src/lib/models/transforms/jitter.ts
+++ b/bokehjs/src/lib/models/transforms/jitter.ts
@@ -13,6 +13,7 @@ export namespace Jitter {
     width: number
     distribution: Distribution
     range: Range
+    previous_values: Float64Array
   }
 
   export interface Props extends Transform.Props {}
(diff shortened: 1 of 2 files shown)


bryevdv:
@birdsarah @mattpap does this need to be configurable?

bokehjs/src/lib/models/transforms/jitter.ts
```diff
@@ -13,6 +13,7 @@ export namespace Jitter {
     width: number
     distribution: Distribution
     range: Range
+    previous_values: Float64Array
```
mattpap: This should be `Arrayable<number>`.

bokehjs/test/models/transforms/jitter.coffee
```diff
@@ -43,6 +43,18 @@ describe "Jitter transform module", ->
       # We can set this deterministically because we've stubbed random
       expect(thediff).to.equal 0
 
+    it "should cache values for identical input lengths", ->
+      N = 100
+      val1 =  Array.apply(null, Array(N)).map ->
+                5
```
mattpap: Use `repeat(5, N)` (from `core/util/array`). Do similarly in other instances.

Maggie-M: Thanks, @mattpap!  I've made these changes.

PR #7979, opened 2018-06-08
Rewrite bokehjs' tests in TypeScript (part 2)

mattpap:
addresses #7864


bokehjs/src/lib/models/sources/column_data_source.ts
```diff
@@ -193,7 +193,7 @@ export class ColumnDataSource extends ColumnarDataSource {
     this.streaming.emit()
   }
 
-  patch(patches: [Index, any][]): void {
+  patch(patches: {[key: string]: [Index, any][]}): void {
```
mattpap: One reason why tests must be written in TypeScript.

PR #7980, opened 2018-06-08
don't generate detail validation stings for purely internal checks

### LINKED ISSUE #7977: High (probably unnecessary) CPU load due to is_valid calls.
python 3.6 on linux 4.16, tested with the latest (as of 2018-06-07) bokeh dev version and with 0.12.16

This example exhibits bokeh slowness and high CPU usage:

``` python
from bokeh.models import Div, CustomJS
from bokeh.io import curdoc
import os

# prepare some data
x = range(0,100)
y = range(0,100)
os.environ['BOKEH_VALIDATE_DOC'] = "0" # This does nothing for this example

def create_plot():

    return p

plots = []

for i in range(0,50):
    plots.append(create_plot())

l = layout(plots)

curdoc().add_root(l)
```

The high CPU usage is partially in the browser, but that only lasts a few seconds, however in the server, it uses *100% CPU for up to 20 seconds* (on an 8-th gen Intel CPU) even after the page seems fully rendered in the browser.

`bokeh serve --show example.py`

[**Flamechart**](https://gistpreview.github.io/?f1fa42b4aeea8f0e3f42b43863980221/bokeh_flame.html
)

Most of the time seems to be spent doing validations with `self.validate()` which throws an expensive (there is a lot of string concatenating/formatting) error the message of which is then [ignored in `is_valid`](https://github.com/bokeh/bokeh/blob/16cfb18ab72c35bb6e6c59888545b509cedd0b9f/bokeh/core/property/bases.py#L256).

I am not sure whether is_valid should not be called here or if it should be cheaper to call.

Setting [BOKEH_VALIDATE_DOC](https://bokeh.pydata.org/en/latest/docs/dev_guide/env_vars.html#bokeh-validate-doc) to false does not seem to help as it hits no code path which checks that variable. It is checked here:
```
$ git grep 'perform_document_validation'
bokeh/application/application.py:        if settings.perform_document_validation():
bokeh/embed/notebook.py:    if settings.perform_document_validation():
bokeh/embed/standalone.py:    if settings.perform_document_validation():
bokeh/settings.py:    def perform_document_validation(self, default=True):
```

The flamechart has been created using [pyflame](https://github.com/uber/pyflame):
`sudo pyflame -s 20 -p {PID} -o prof.txt`

and [FlameGraph](https://github.com/brendangregg/FlameGraph):
`flamegraph.pl prof.txt > _bokeh_flame.html`

bryevdv:
- [x] issues: fixes #7977
- [x] tests added / passed

Will add tests after it is determined that this is of reasonable benefit.

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -613,7 +613,7 @@ class Interval(ParameterizedProperty):
     def validate(self, value, detail=True):
         super(Interval, self).validate(value, detail)
 
-        if not (value is None or self.interval_type.is_valid(value, False) and value >= self.start and value <= self.end):
+        if not (value is None or self.interval_type.is_valid(value) and value >= self.start and value <= self.end):
             msg = ""
             if detail:
                 msg = "expected a value of type %s in range [%s, %s], got %r" % (self.interval_type, self.start, self.end, value)
@@ -710,7 +710,7 @@ class Either(ParameterizedProperty):
     def validate(self, value, detail=True):
         super(Either, self).validate(value, detail)
 
-        if not (value is None or any(param.is_valid(value, False) for param in self.type_params)):
+        if not (value is None or any(param.is_valid(value) for param in self.type_params)):
             msg = ""
             if detail:
                 msg = "expected an element of either %s, got %r" % (nice_join(self.type_params), value)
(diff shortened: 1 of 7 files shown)


bryevdv:
@mattpap any thoughts welcome

PR #7983, opened 2018-06-11
Support line annotations like span but with slope

### LINKED ISSUE #6386: Support line annotations like span but with slope
In addition to the vertical and horizontal spans annotations, it would be nice to have a line annotation that can be sloped and can be specified by a point and a slope (or maybe optionally two sets of points which might be more useful if an axis is categorical), where the line does not terminate but continues across the extent of the canvas.

The reason for this request is that it is often desired to place a reference line or a fitted regression line that is not horizontal or vertical, and it is desired for this line to continue beyond the span of the data, or the span of the data is not known at the time of specifying the line, or it is not convenient to specify the line in terms of points to connect at the extent of the data. Especially in the case of superposing a linear regression line, a reference line is easiest specified by a y-intercept and a slope.

anthonydouc:
Adds support for a sloped line annotation, https://github.com/bokeh/bokeh/issues/6386. 

Initial implementation takes user supplied gradient and y-intercept to render the line. No support worked in for screen units and css rendering, but happy to work these in if they are useful and can make sense.

- [X] issues: fixes #6386 
- [] tests added / passed
- [] release document entry (if new feature or API change)


bokehjs/src/lib/models/annotations/slope.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  initialize(options: any): void {
+    super.initialize(options)
+  }
+
+  connect_signals(): void {
+    super.connect_signals()
+    this.connect(this.model.change, () => this.plot_view.request_render())
+    this.connect(this.model.properties.gradient.change, () => this.plot_view.request_render())
+    this.connect(this.model.properties.y_intercept.change, () => this.plot_view.request_render())
```
bryevdv: I don't think the last two are necessary (they are covered by `model.change`)

bokehjs/src/lib/models/annotations/slope.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  render(): void {
+    if (!this.model.visible)
+      return
+    this._draw_span()
+  }
+
+  protected _draw_span(): void {
```
bryevdv: Let's call this `_draw_slope`

bokehjs/src/lib/models/annotations/slope.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    y_start = yscale.invert(sy_start)
+    y_end = yscale.invert(sy_end)
+
+    let sx_start, sx_end, x_start, x_end
+    x_start = (y_start - y_intercept) / gradient
+    x_end = (y_end - y_intercept) / gradient
+
+    sx_start = xscale.compute(x_start)
+    sx_end = xscale.compute(x_end)
```
bryevdv: None of these eight values appear to be re-assigned, so they can all eight just be `const` instead of `let`

bryevdv:
ping @hafen do you have any input here?

anthonydouc:
Current example - the slope is for a linear equation with an intercept of 10, gradient of 2. Some random points are sampled and plotted alongside the equation.

slope example

bryevdv:
@mattpap any comments? Otherwise I will merge for this release later today

Issue #2080, opened 2015-03-17
HoverTool panel for annular wedge displays at center coordinates, not over wedge.


cyrusmaher:
There may also be an issue with the hit detection. I end up with the annotation for all wedges being displayed in the center of the concentric wedges.

fpliger:
hi @cyrusmaher , could you paste an example code here so we can easily reproduce the issue (or maybe a screenshot or both :-) )? Thanks

bryevdv:
There is a parameter to snap to the glyph center that defaults to true. I'm in my phone or I'd look it up but it's in the reference guide. 

Bryan

cyrusmaher:
Thanks Bryan!! That's just what I needed. In future versions, Would it be possible (pretty please) to make the snap_to_data default coordinates the middle of the annular wedge patch? 

Also, in the process of putting together the example code below, I encountered a perhaps mildly interesting bug. For the `annular_wedge` call, if you swap "starts" and "stops" and switch the direction to "clock" instead of "anticlock", it seems to raise an issue with the hit detection even though what's displayed looks identical. Am missing something there or is that actually weird?

```python
import bokeh.plotting as bk
import pandas as pd
import math
from bokeh.models import HoverTool, ColumnDataSource

rad2deg = (180 / math.pi)
deg2rad = 1 / rad2deg
        
y_mins = [1] * 3
y_maxes = [2] * 3
start_angles = pd.Series([30, 45, 90])
stop_angles = pd.Series([40, 80, 120])

start_angles_2 = start_angles + 180
stop_angles_2 = stop_angles + 180

df = pd.DataFrame(
                    {
                        'x': 0,
                        'y': 0,
                        'ymin': y_mins, 
                        'ymax': y_maxes,
                        'starts': start_angles * deg2rad,
                        'stops': stop_angles * deg2rad
                    }
                )

df_s = ColumnDataSource(df)
TOOLS = "pan,wheel_zoom,box_zoom,reset,hover"
bk.output_file("test.html")
plot = bk.figure(width=1000, height=1000, tools=TOOLS)


plot.annular_wedge(x='x', y='y', 
                    inner_radius='ymin',
                    outer_radius='ymax',
                    start_angle='stops', # change to "starts"
                    end_angle='starts', # change to "stops"
                    direction="clock", # change to "anticlock"...now it works
                    source=df_s
                    )


hover = plot.select(dict(type=HoverTool))
hover.tooltips = [
    # add to this
    ("index", "$index"),
[...]

PR #7999, opened 2018-06-16
Fix hit testing issues for wedge and annular wedge

### LINKED ISSUE #7428: Correctly showing Wedge Hover Tooltips with "clock" orient

Python 3.6.3
Bokeh 0.12.10

The hover shows information of all the slices except the one you are pointing to, it should work the other way around.

Example code taken from [https://stackoverflow.com/questions/48305017/correctly-showing-hover-tooltips-on-bokeh-wedge-chart](url)

```
import numpy as np
from bokeh.models import HoverTool, ColumnDataSource
from math import pi



percents = [0, 5/143, 51/143, 88/143, 108/143, 141/143, 1.0]
category = ['A ', 'B ', 'C ', 'D ', 'E ', 'F']
counts = [5, 46, 37, 20, 33, 2]
starts = [1/2*pi-(p*2*pi) for p in percents[:-1]]
ends = [1/2*pi-(p*2*pi) for p in percents[1:]]
colors = ['#889dba', '#1f356f', '#1e92b8', '#33748a', '#a5d3e3', '#bbc2d4']
# create source
source = ColumnDataSource(
    data=dict(
        x=[0 for x in percents],
        y=[0 for x in percents],
        radius = [0.5 for x in percents],
        percents=percents,
        category= category,
        starts=starts,
        colors=colors,
        ends=ends,
        counts = counts
    )
)

TOOLS = "hover"

p = figure(plot_width = 500, plot_height = 500, x_axis_label = None, y_axis_label = None,
title = 'Type', tools = TOOLS)

p.title.align = 'center'
p.title.text_font = 'arial narrow'

p.wedge(x='x', y='y',  radius = 'radius', direction="clock",
                start_angle='starts', end_angle='ends', color='colors', source=source)

hover = p.select(dict(type=HoverTool))
hover.tooltips = [
    ('category', '@category'),
    ('percents','@counts')
]


p.axis.visible = False
p.ygrid.visible = False
p.xgrid.visible = False

output_file(pie.html')
show(p)

```

The hover works correctly if the user inverts the data and removes the clockwise statement as shown in this answer of the Stackoverflow question.
[https://stackoverflow.com/questions/48305017/correctly-showing-hover-tooltips-on-bokeh-wedge-chart/48306227#48306227](url)

anthonydouc:
Corrects the issues discussed in #7428 and #2080. Namely the incorrect behaviour when the direction is set to clock, and also improving hover tool tip display position for wedge glyphs.

Might be some changing required depending on the exact intention of using angle_between.

- [X] issues: fixes #7428 and #2080

### DIFF
--- a/bokehjs/src/lib/core/util/math.ts
+++ b/bokehjs/src/lib/core/util/math.ts
@@ -1,3 +1,5 @@
+const enum Direction {clock, anticlock}
+
 export function angle_norm(angle: number): number {
   while (angle < 0) {
     angle += 2*Math.PI
@@ -12,12 +14,11 @@ export function angle_dist(lhs: number, rhs: number): number {
   return Math.abs(angle_norm(lhs-rhs))
 }
 
-export function angle_between(mid: number, lhs: number, rhs: number, direction: 1 | 0): boolean {
-  // 1 = anticlock, 0 = clock
+export function angle_between(mid: number, lhs: number, rhs: number, direction: Direction): boolean {
   const norm_mid = angle_norm(mid)
   const d = angle_dist(lhs, rhs)
   const cond = angle_dist(lhs, norm_mid) <= d && angle_dist(norm_mid, rhs) <= d
-  if (direction == 1)
+  if (direction == Direction.anticlock)
     return !cond
   else
     return cond
--- a/bokehjs/src/lib/core/util/math.ts
+++ b/bokehjs/src/lib/core/util/math.ts
@@ -12,14 +12,15 @@ export function angle_dist(lhs: number, rhs: number): number {
   return Math.abs(angle_norm(lhs-rhs))
 }
 
-export function angle_between(mid: number, lhs: number, rhs: number, direction: "anticlock" | "clock"): boolean {
+export function angle_between(mid: number, lhs: number, rhs: number, direction: 1 | 0): boolean {
+  // 1 = anticlock, 0 = clock
   const norm_mid = angle_norm(mid)
   const d = angle_dist(lhs, rhs)
   const cond = angle_dist(lhs, norm_mid) <= d && angle_dist(norm_mid, rhs) <= d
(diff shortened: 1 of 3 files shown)


bokehjs/src/lib/core/util/math.ts
```diff
@@ -12,14 +12,15 @@ export function angle_dist(lhs: number, rhs: number): number {
   return Math.abs(angle_norm(lhs-rhs))
 }
 
-export function angle_between(mid: number, lhs: number, rhs: number, direction: "anticlock" | "clock"): boolean {
+export function angle_between(mid: number, lhs: number, rhs: number, direction: 1 | 0): boolean {
+  // 1 = anticlock, 0 = clock
```
mattpap: Use an enum to make the API more meaningful (https://www.typescriptlang.org/docs/handbook/enums.html).

anthonydouc: So its preferable to pass in a string with one of "anticlock" or "clock"? In that case I can change how the value is being passed into "angle_between"

bryevdv: @anthonydouc not strings, TypeScript supports actual enum variables, see e.g. https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/tools/gestures/range_tool.ts#L11

anthonydouc:
I must of missed that, that issue does not occur in every example depending on the width and height of the figure.

bryevdv:
> I must of missed that, that issue does not occur in every example depending on the width and height of the figure. 

Yes, I noticed it even in the example above, different wedges would respond differently to hit testing. Easy to miss, and an example of why more eyes is good for OSS :D

bryevdv:
brilliant, thanks @anthonydouc I am very happy to get this in the next release, it complements the `cumsum` transform addition, and together both make donut charts much more feasible

PR #8001, opened 2018-06-17
Fixes #7401: set range with pandas Series

### LINKED ISSUE #7401: Support passing a pandas.Series as x_range for Figure
Bokeh 0.12.14dev3

Currently, running this code (taken from https://stackoverflow.com/questions/48152251/unable-to-create-charts-using-bokeh/48170111):
```python
import pandas as pd

car = pd.DataFrame([l.split() for l in """Jan 1808
Feb 1251
Mar 3023""".splitlines()], columns=['Month', 'Sales'])

p = figure(title="Bokeh Scatter Plot Example", x_axis_label='Month',
           y_axis_label='Sales', x_range=car['Month'])

p.circle(car['Month'], car['Sales'], size=5, alpha=0.5)

show(p)
```
results in 
```python
  File "/home/p-himik/dev/git/bokeh/test.py", line 29, in <module>
    y_axis_label='Sales', x_range=car['Month'])
  File "/home/p-himik/dev/git/bokeh/bokeh/plotting/figure.py", line 793, in figure
    return Figure(**kwargs)
  File "/home/p-himik/dev/git/bokeh/bokeh/plotting/figure.py", line 140, in __init__
    self.x_range = _get_range(opts.x_range)
  File "/home/p-himik/dev/git/bokeh/bokeh/plotting/helpers.py", line 355, in _get_range
    raise ValueError("Unrecognized range input: '%s'" % str(range_input))
ValueError: Unrecognized range input: '0    Jan
1    Feb
2    Mar
Name: Month, dtype: object'
```

The workaround is to replace `x_range=car['Month']` with `x_range=list(car['Month'])`.

Karel-van-de-Plassche:
Access underlying `np.ndarray` when passing `pd.Series`. Also added tests, and made the pandas dependency optional (this is good-form, right?)

- [x] issues: fixes #7401
- [x] tests added / passed

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -346,6 +346,8 @@ def _get_range(range_input):
         return FactorRange(factors=sorted(list(range_input.groups.keys())))
     if isinstance(range_input, Range):
         return range_input
+    if pd and isinstance(range_input, pd.Series):
+        range_input = range_input.values
     if isinstance(range_input, (Sequence, np.ndarray)):
         if all(isinstance(x, string_types) for x in range_input):
             return FactorRange(factors=list(range_input))
--- a/bokeh/plotting/tests/test_helpers.py
+++ b/bokeh/plotting/tests/test_helpers.py
@@ -289,7 +289,7 @@ def test__get_range_with_ndarray():
 
 def test__get_range_with_too_long_ndarray():
     with pytest.raises(ValueError):
-        r = _get_range(np.array([10, 20, 30]))
+        _get_range(np.array([10, 20, 30]))
 
 def test__get_range_with_ndarray_factors():
     f = np.array(["Crosby", "Stills", "Nash", "Young"])
@@ -307,7 +307,7 @@ def test__get_range_with_series():
 @skipIf(not is_pandas, "pandas not installed")
 def test__get_range_with_too_long_series():
     with pytest.raises(ValueError):
-        r = _get_range(pd.Series([20, 30, 40]))
+        _get_range(pd.Series([20, 30, 40]))
 
 def test__get_range_with_string_seq():
     f = ["foo" ,"end", "baz"]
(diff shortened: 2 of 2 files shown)


bokeh/plotting/tests/test_helpers.py
```diff
@@ -1,7 +1,13 @@
...
-
-import pandas as pd
+from unittest import skipIf
+
+import numpy as np
+try:
+    import pandas as pd
+    is_pandas = True
+except ImportError as e:
+    is_pandas = False
```
bryevdv: FWIW I'd personally be fine just stipulating that pandas is a required *test* dependency and not complicating the test files like this but I don't have super strong feelings about it, either

bokeh/plotting/helpers.py
```diff
@@ -346,6 +346,8 @@ def _get_range(range_input):
         return FactorRange(factors=sorted(list(range_input.groups.keys())))
     if isinstance(range_input, Range):
         return range_input
+    if pd and isinstance(range_input, pd.Series):
+        range_input = range_input.values
```
bryevdv: We probably need to validate that the length of the series is exactly two and raise an informative `ValueError` if not

bokeh/plotting/tests/test_helpers.py
Karel-van-de-Plassche: For me it is also fine either way. This was the style in test_sources.py, which is why I copied it. I think you should just decide on one style, and then I can change it everywhere.

bokeh/plotting/helpers.py
Karel-van-de-Plassche: Actually I just expose the underlying numpy array, and let this be handled by the next 'if' statement. I've added tests that reflect this.

bryevdv:
looks just like pyflakes fails at this point:
```
E         ./bokeh/plotting/tests/test_helpers.py:292:9: F841 local variable 'r' is assigned to but never used
E         ./bokeh/plotting/tests/test_helpers.py:310:9: F841 local variable 'r' is assigned to but never used
E         
E       assert 1 == 0
```

Karel-van-de-Plassche:
Whoops, that's what you get for trying to fix something quickly. Fixed!

bryevdv:

PR #8002, opened 2018-06-17
Added workaround when initializing CDS with column named index

### LINKED ISSUE #6770: Task: Add user warn/error to ColumnDataSource.from_df method about DataFrame containing "index" column
There's some weirdness around using pandas DataFrames to create ColumnDataSources for graph visualization in Bokeh. Currently Bokeh requires the GraphRenderer.node_renderer.data_source to have a column named "index". This column is the reference for the edge's start and end values.

The issue is that creating a ColumnDataSource with a pd.DataFrame with:
* an unnamed Index
* a column named "index"

The resulting CDS will have a "index" column containing the dataframe (capital-i) Index and not the column. This is a gotcha in pandas where ``df.index`` refers to the Index but ``df["index"]`` refers to a column named "index".

Bokeh 0.12.7dev16

Karel-van-de-Plassche:
Was not 100% sure of the course of action here. I think that the user expects the column name to be the same, and cares less about the specific name of the index. Which is why I opted to adjust the index name, and not the column name. The warning should help either way for this edge case.

- [x] issues: fixes #6770
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -159,6 +159,11 @@ class ColumnDataSource(ColumnarDataSource):
             new_data[k] = v
 
         index_name = ColumnDataSource._df_index_name(df)
+        if index_name in new_data:
+            old_index_name = index_name
+            index_name = old_index_name + '_index'
+            warnings.warn('Column named %s. Renaming index to %s' %
+                          (old_index_name, index_name))
         new_data[index_name] = _df.index.values
         return new_data
 


bryevdv:
cc @mattpap @canavandl 

Unsure about this as well, just picking an arbitrary name + `_index` risks the same problem in cases where  name + `_index` also already exists. So this seems like a band aid. I guess I'm -0 on it, if others think it's a net improve then I'm fine to have it merged. But the docstring at least needs to be updated to explicitly describe what happens in this situation.

Karel-van-de-Plassche:
It's indeed a band-aid, although I strongly question the sanity of naming a column `index_index` ;) Anyhow, I feel it would be better to use [df.reset_index](https://pandas.pydata.org/pandasdocs/version/0.21/generated/pandas.DataFrame.reset_index.html) here. I didn't want to change the behaviour, but now that I think about it, isn't putting every index in the multiindex in its own column (as `reset_index` does), not more intuative than the current solution anyway? A column of tuples is a bit weird. If not, I can just copy the naming behaviour of `reset_index` and implement that.

Karel-van-de-Plassche:
We then could document that it uses `reset_index` under the hood, and even let the user pass kwargs to it.

bryevdv:
Well, we shouldn't tamper with or modify any DataFrame that a user passes in, so unless I'm misunderstanding, I don't think we should call `df.reset_index`. I think I might be misunderstanding the proposal tho, so maybe being able to look at the proposed code change would help clarify.

Karel-van-de-Plassche:
`reset_index()` makes a copy by default, so the original would stay untouched. I'll make a proof of concept later this week.

Karel-van-de-Plassche:
Superseded by #6770

PR #8003, opened 2018-06-17
Updated docs for two-way Jupyter communication

Karel-van-de-Plassche:
As discussed on gitter


bryevdv:

Issue #8000, opened 2018-06-16
Small documentation error for BoxSelectTool


kevinsa5:
The documentation for `BoxSelectTool` says that to constrain selection to a single axis, you pass `width` or `height` as a string in a single-element list, but this is incorrect.  Passing just the string itself works.  Relevant docs section [here](http://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#specifying-tools)

```
(venv) kevin@onion-run:~/iplot$ python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import bokeh
>>> print bokeh.__version__
0.12.16
>>> 
>>> from bokeh.models import BoxSelectTool
>>> from bokeh.plotting import figure
>>> 
>>> plot = figure(tools="pan,wheel_zoom,box_zoom,reset")
>>> plot.add_tools(BoxSelectTool(dimensions=["width"]))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/model.py", line 219, in __init__
    super(Model, self).__init__(**kwargs)
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/core/has_props.py", line 236, in __init__
    setattr(self, name, value)
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/core/has_props.py", line 263, in __setattr__
    super(HasProps, self).__setattr__(name, value)
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/core/property/descriptors.py", line 503, in __set__
    self._internal_set(obj, value, setter=setter)
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/core/property/descriptors.py", line 724, in _internal_set
    value = self.property.prepare_value(obj, self.name, value)
  File "/home/kevin/iplot/venv/local/lib/python2.7/site-packages/bokeh/core/property/bases.py", line 278, in prepare_value
    raise e
ValueError: expected a value of type basestring, got ['width'] of type list
>>> plot.add_tools(BoxSelectTool(dimensions="width"))
>>> 

```

bryevdv:
@kevinsa5 are you interested in submitting a PR to fix this up?

PR #8004, opened 2018-06-17
Corrected docs for `dimensions` parameter to BoxZoomTool

kevinsa5:
Closes https://github.com/bokeh/bokeh/issues/8000.  Nothing fancy, just cleaning up some docs.  Current version lives [here](bokeh.pydata.org/en/latest/docs/user_guide/tools.html)

### DIFF
--- a/sphinx/source/docs/user_guide/tools.rst
+++ b/sphinx/source/docs/user_guide/tools.rst
@@ -181,7 +181,7 @@ BoxSelectTool
 The box selection tool allows the user to define a rectangular selection
 region by left-dragging a mouse, or dragging a finger across the plot area.
 The box select tool may be configured to select across only one dimension by
-setting the ``dimensions`` property to ``width`` or ``height`` instead of the 
+setting the ``dimensions`` property to ``width`` or ``height`` instead of the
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @kevinsa5 only one small comment, the value can also be `"both"`

bryevdv:
Awesome, thanks for the catch and the fix @kevinsa5 !

kevinsa5:
@bryevdv happy to help!  Thank you for your tireless efforts here -- we have several "business-critical" tools implemented in Bokeh at my workplace in the clean energy industry.  Wouldn't have happened without you guys.

bryevdv:
@kevinsa5 that's a nice testimonial to hear! If you ever have something you can publicly share or showcase we are always happy to have guest blogs on https://bokeh.github.io (or tweet out links to write-ups elsewhere) I do understand things are often confidential/proprietary though. Alternatively if your company is in a position to directly support OSS, Bokeh is now a NumFocus project and donations can be made to NF in the projects name. 

In any event thanks again for the PR, we are having some spurious TravisCI fails but we will get this merged when things are green!

PR #8005, opened 2018-06-18
bokeh server: reference leak fix

### LINKED ISSUE #7992: memory leak on add_next_tick_callback
ubuntu/18.04
python/3.6
bokeh/0.12.15dev8 onwards

The following code, which is derived from [this example](https://bokeh.pydata.org/en/latest/docs/user_guide/server.html#updating-from-unlocked-callbacks) (the size of an object being passed to `add_next_tick_callback` is enlarged for emphasis), continuously drains memory starting from bokeh/0.12.15dev8 (in particular, https://github.com/bokeh/bokeh/commit/de32960039d19599f03e32ab0aedc090918b4602)

```python
from functools import partial

from concurrent.futures import ThreadPoolExecutor
from tornado import gen
import numpy as np

from bokeh.document import without_document_lock

source = ColumnDataSource(data=dict(x=[0], y=[0], color=["blue"]))

i = 0

doc = curdoc()

executor = ThreadPoolExecutor(max_workers=2)

def blocking_task():
    # receive/generate new data
    return np.random.rand(1024, 1024)

@gen.coroutine
def locked_update(res):
    # keep it simple as in example, but practically the plot may be updated with res values
    source.stream(dict(x=[source.data['x'][-1]+1], y=[i], color=["blue"]))

@gen.coroutine
@without_document_lock
def unlocked_task():
    global i
    i += 1
    res = yield executor.submit(blocking_task)
    doc.add_next_tick_callback(partial(locked_update, res=res))

p = figure(x_range=[0, 100], y_range=[0, 20])
l = p.circle(x='x', y='y', color='color', source=source)

doc.add_periodic_callback(unlocked_task, 1000)
doc.add_root(p)
```
Seems that, the `partial` objects passed to 'add_next_tick_callback' are piling up in memory and never get released/garbage collected.

Here is a minimal reproducible example:
```python
from functools import partial

import numpy as np
from tornado import gen

doc = curdoc()

@gen.coroutine
def locked_update(res):
    pass

@gen.coroutine
def update():
    res = np.random.rand(1024, 1024)
    doc.add_next_tick_callback(partial(locked_update, res))

doc.add_periodic_callback(update, 1000)
```

ivan-usov:
The #7992 issue originates from the fact that `_callback_objs_by_callable` in document.py and `_next_tick_removers_by_callable` etc. in tornado.py are retaining references, as dictionary keys, to callable objects (and their wrapped versions, correspondingly), passed to `add_next_tick_callback` etc. document methods. I guess, the intention was to group multiple next tick or timeout callbacks for the same callback function in `defaultdict(set)`s. Though, the wrapped callbacks of the same callback function are already different objects and "won't stack" in `_next_tick_removers_by_callable`, while passing different callback functions will aggregate in both location, thus doubling the captured reference count.

This PR is fixing it by removing the `key, value` pairs completely, once the `value` is an empty set. This is, probably, the least invasive way to fix the issue, though, not the most performant. Arguably, a simple dict `{callback_obj: callback}` in document.py and `{callback_id: wrapped_callback}` in tornado.py might be better, and can also eliminate the separation between different types of document callbacks in different dictionaries(?). Not sure about the latter in a view of deprecated ways of adding/removing callbacks.

Maybe a couple of tests should be added to verify that the bokeh server doesn't hold (directly and indirectly) any refs to callback functions after their execution.

- [x] issues: fixes #7992 
- [ ] tests added / passed

### DIFF
--- a/bokeh/document/document.py
+++ b/bokeh/document/document.py
@@ -1006,9 +1006,11 @@ class Document(object):
             else:
                 callback_objs = [callback_obj]
                 self._session_callbacks.remove(callback_obj)
-                for cb_objs in self._callback_objs_by_callable[originator].values():
+                for cb, cb_objs in list(self._callback_objs_by_callable[originator].items()):
                     try:
                         cb_objs.remove(callback_obj)
+                        if not cb_objs:
+                            del self._callback_objs_by_callable[originator][cb]
                     except KeyError:
                         pass
         except KeyError:
--- a/bokeh/util/tornado.py
+++ b/bokeh/util/tornado.py
@@ -167,20 +167,20 @@ class _CallbackGroup(object):
         return remover
 
     def _execute_remover(self, callback_id, removers):
-            try:
-                with self._removers_lock:
-                    if callable(callback_id):
-                        remover = self._create_deprecated_remover(callback_id, removers)
-                    else:
-                        remover = removers.pop(callback_id)
-                        for cb_ids in self._get_removers_ids_by_callable(removers).values():
-                            try:
-                                cb_ids.remove(callback_id)
-                            except KeyError:
-                                pass
-            except KeyError:
-                raise ValueError("Removing a callback twice (or after it's already been run)")
-            remover()
+        try:
+            with self._removers_lock:
+                if callable(callback_id):
+                    remover = self._create_deprecated_remover(callback_id, removers)
+                else:
+                    remover = removers.pop(callback_id)
+                    for cb_ids in self._get_removers_ids_by_callable(removers).values():
+                        try:
+                            cb_ids.remove(callback_id)
+                        except KeyError:
+                            pass
+        except KeyError:
+            raise ValueError("Removing a callback twice (or after it's already been run)")
+        remover()
 
     def add_next_tick_callback(self, callback, callback_id=None):
         """ Adds a callback to be run on the next tick.
--- a/bokeh/util/tornado.py
+++ b/bokeh/util/tornado.py
@@ -173,9 +173,11 @@ class _CallbackGroup(object):
                     remover = self._create_deprecated_remover(callback_id, removers)
                 else:
                     remover = removers.pop(callback_id)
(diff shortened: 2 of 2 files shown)


bryevdv:
> Maybe a couple of tests should be added to verify that the bokeh server doesn't hold (directly and indirectly) any refs to callback functions after their execution.

The way we handled this with document modules that were leaking was to add a runtime check, that the number of referrers was the expected number, and then a test that maintained that the error was logged if there were extra referrers:

https://github.com/bokeh/bokeh/blob/master/bokeh/document/tests/test_document.py#L107
https://github.com/bokeh/bokeh/blob/master/bokeh/document/document.py#L458
 
But I'm open to just having unit tests if that seems reasonable.

bryevdv:
@ivan-usov I would also be ok with merging this now and leaving a new issue to add clean up (add tests, and we could also remove the deprecated code paths for removing callbacks by value)

If you can post some benchmarks or measurements here that demonstrate the improved change, I think that would be fine for now, so this can go in the imminent release.

ivan-usov:
@bryevdv This PR, pretty much, recovers numbers from bokeh/0.12.14 on the same test cases mentioned in the issue,
<details>
  <summary>e.g. a case of the same callback function</summary>

```
import gc

from tornado import gen

doc = curdoc()

def func(): pass

@gen.coroutine
def update():
    doc.add_next_tick_callback(func)

    gc.collect()
    print(len(gc.get_referrers(func)))

doc.add_periodic_callback(update, 1000)
```
```
4
4
4
4
```
</details>

 </br> 
Awesome! Great, that it's been merged before 0.13.0.

bryevdv:
> This PR, pretty much, recovers numbers from bokeh/0.12.14 on the same test cases mentioned in the issue,

Those don't seem sufficient though, right? They didn't alert to this situation, after all, and that's because they only fire when a session ends and the module is reclaimed incompletely. This issue was related to problems while a session was active and ongoing. To guarantee there is not a regression on this in the future there probably need to be new tests.

PR #7494, opened 2018-02-05
Specify compilation inputs in tsconfig.json

mattpap:
This is a further simplification of our build. This also makes bokehjs development with vscode finally possible.

addresses #6481

PR #8006, opened 2018-06-18
Ignore .vscode folder

ivan-usov:
Ignore `.vscode` folder (similar to e.g. `.idea`), especially, in a view of #7494 .

Shall I also create an issue for that?

- [ ] issues: fixes #xxxx
- [ ] tests added / passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,7 @@ bokeh.server.*
 bokeh.sets.*
 remotedata
 target/
+/.vscode/
 /.idea/
 /.c9/
 


mattpap:
There are better ways to achieve this. One can have either a global `~/.gitignore` (needs an entry in `gitconfig`) or per repository exclusions (in `repo_name/.git/info/exclude`). This way you can configure each repository you work with with your specific setup. In fact, I want to clean up bokeh's `.gitignore` and remove most entries.

ivan-usov:
Also, `~/.config/git/ignore` would work without an entry in `gitconfig` (as a default location for the global ignore file). Fair enough, I guess, who needs it, can set it up locally...

Issue #799, opened 2014-07-08
Templates Not Included When Compiling With Py2exe


bhicks92:
I am attempting to use py2exe to build an executable from a python script that uses bokeh. The entire templates folder is not included in the compiled bokeh library, and the executable therefore cannot create plots.

Traceback (most recent call last):
  File "myscript.py", line 624, in myfunction
  File "bokeh\plotting.pyc", line 358, in show
  File "bokeh\session\html_file_session.pyc", line 194, in save
  File "bokeh\session\html_file_session.pyc", line 150, in dumps
  File "bokeh\session\base_html_session.pyc", line 67, in _load_template
IOError: [Errno 2] No such file or directory: 'C:\dist\library.zip\bokeh\session\..\templates\plots.js'

bryevdv:
No response, `py2exe` seems like a fairly marginal use case, closing. OP please re-open if you have more information to provide regarding the new information above.

FabeG:
I succeeded in building an executable with ``py2exe`` from a script using ``bokeh``.

My first try wasn't very promising (using the last ``bokeh`` version 0.12.3). It failed with:

Traceback (most recent call last):
  File "test_bokeh_simple.py", line 8, in <module>
  File "bokeh\io.pyc", line 29, in <module>
  File "bokeh\core\state.pyc", line 44, in <module>
  File "bokeh\document.pyc", line 34, in <module>
  File "bokeh\core\templates.pyc", line 22, in <module>
  File "jinja2\environment.pyc", line 812, in get_template
  File "jinja2\environment.pyc", line 774, in _load_template
  File "jinja2\loaders.pyc", line 235, in get_source
jinja2.exceptions.TemplateNotFound: js_resources.html

It seems that all the template files contained in the `bokeh/core/_templates` should be included in the package created by ``py2exe``.
Using the ``data_files`` option of ``py2exe``  doesn't solve the problem, because these template files have to be in the `bokeh/core/_templates` directory.
Since the ``bokeh`` module is included in the ``library.zip`` file created by ``py2exe``, the idea was to add the ``_templates`` directory into this zip file. 

Here is the minimal ``setup.py`` file that worked for me (`test_simple_bokeh.py` being the file to convert into an executable):

```python
import zipfile
import os
import sys
from setuptools import setup

import bokeh.core
import py2exe
from matplotlib import get_py2exe_datafiles

current_dir = os.path.dirname(os.path.realpath(__file__))
[...]

PR #8008, opened 2018-06-18
Make Bokeh work in frozen scripts

severinsimmler:
This PR fixes a bug concerning the interaction between Bokeh and PyInstaller by using Jinja2's `FileSystemLoader` instead of `PackageLoader` (which is currently not supported by PyInstaller), and adjusting the path to the `_templates` folder. It should also fix issues with other freezing tools (like py2exe in #799). However, PyInstaller is probably the most widely used freezing tool and is still actively maintained (unlike e.g. py2exe, whose last release is almost four years old).

Since the changes are in an `if getattr(sys, 'frozen', False)` block (as described in the [PyInstaller documentation](http://pyinstaller.readthedocs.io/en/stable/runtime-information.html)), this PR shouldn't break anything for the "live" use.

PyInstaller itself does not copy the `bokeh/core/_templates` folder (which contains files to be loaded by `bokeh/core/templates.py`). When a bundled app starts up, the bootloader stores the absolute path to the bundle folder in `sys._MEIPASS` – and this is where I tell Jinja2 to look for the `_templates` folder.

If you now freeze the script with PyInstaller, the `_templates` folder must be explicitly copied with `--add-data _templates:_templates` (or in the `.spec` file), but that's something the PyInstaller people should mention in their documentation (if my PR is merged, I can get this going).

I'm not sure how to test this (for continuous integration). Creating a bundled app and check if it works seems to be a bit too much for this, well, pretty special use case?

The changes are tested locally on my machine with Bokeh 0.12.16 and PyInstaller 3.3.1 on Ubuntu 18.04 and Python 3.6.5. This is the minimal example that worked:

```python

# prepare some data
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]

# output to static HTML file
output_file("lines.html", title="line plot example")

# show the results
show(p)
```

If there is anything else I can do, let me know!

### DIFF
--- a/bokeh/core/templates.py
+++ b/bokeh/core/templates.py
@@ -16,12 +16,22 @@ models in various ways.
 
 '''
 from __future__ import absolute_import
-
 import json
+from jinja2 import Environment, Markup, FileSystemLoader, PackageLoader
+import sys
+import os
 
-from jinja2 import Environment, PackageLoader, Markup
+def get_env():
+    ''' Get the correct Jinja2 Environment, also for frozen scripts.
+    '''
+    if getattr(sys, 'frozen', False):
+        templates_path = os.path.join(sys._MEIPASS, '_templates')
+        return Environment(loader=FileSystemLoader(templates_path))
+    else:
+        return Environment(loader=PackageLoader('bokeh.core', '_templates'))
 
-_env = Environment(loader=PackageLoader('bokeh.core', '_templates'))
+
+_env = get_env()
 _env.filters['json'] = lambda obj: Markup(json.dumps(obj))
 
 JS_RESOURCES = _env.get_template("js_resources.html")


bryevdv:
> Creating a bundled app and check if it works seems to be a bit too much for this

Thanks for the PR @severinsimmler  I'm going to go ahead and merge this now, since I'd like it to be in the release coming up, but I do think making a bundled app and testing it would be the best approach. That kind of test could go in the informational `downstream` build. I'd definitely be happy to accept a PR for tests later and I think that would be the best guarantee that this stays maintained.

axil:
I've found and fixed two more issues when freezing the bokeh with pyinstaller, see bokeh PR8038 and pyinstaller PR3607.

lukasz-migas:
Sadly this broke the not-maintained `py2exe `compiler which does not use `sys._MEIPASS`

bryevdv:
@lukasz-migas if you have a way to make it work with both, I'm happy to accept a PR. Otherwise this is a very low on the priority already, and if we have to choose between an supporting an unmaintained tool, and a maintained on, we will go with the latter.

PR #8016, opened 2018-06-22
test with minimal packages

### LINKED ISSUE #2596: CI test to make sure bokeh works without heavy dependencies
There should be a separate test that makes sure that bokeh works without ~~~numpy~~~, scipy, pandas, etc. The minimal environment should contain python (obviously), six and preferably nothing more. Make sure to test not only if bokeh imports like this, but test as many plotting APIs as feasible 

*removed comments about charts*

bryevdv:
This adds a `MINIMAL=1` flag to one unit test build that results in these dependencies being uninstalled:

> pandas scipy notebook scikit-learn sympy flask

As a side effect, all reliance on `unittest.TestCase` has been removed, all tests are "pytest" style. As a further result, all use of pandas in unit tests can now be moderated through a `pd` pytest fixture that will automatically skip a test if pandas is not installed, with one exception: The modules in `bokeh.sampledata` heavily depend on pandas and throw a runtime exception stating to install pandas even on just an import . To deal with this the unit test script adds `"and not sampledata"` when `MINIMAL` is set.

- [x] issues: fixes #2596
- [x] tests added / passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,7 @@ jobs:
       - GROUP=unit
       - PYTHON=3.5
       - TORNADO=4
+      - MINIMAL=1
 
     - install: scripts/ci/install:test
       script: scripts/ci/test
--- a/bokeh/core/property/tests/test_bases.py
+++ b/bokeh/core/property/tests/test_bases.py
@@ -2,9 +2,9 @@ from mock import patch
 import pytest
 
 import numpy as np
-import pandas as pd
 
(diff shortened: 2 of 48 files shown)


bryevdv:
I'd add I'm not 100% happy about the way sampledata is handled, but it's ok for now. There doesn't seem to be a great way to have those also skip automatically, but perhaps they could be split off to a non-unit test

PR #8017, opened 2018-06-22
Added short usage instructions for examples/app/dash

numpynewb:
Added short usage instructions as per @bryevdv (#8011 comment: https://github.com/bokeh/bokeh/pull/8011#issuecomment-399509585)


bryevdv:
Thanks for the contribution, @numpynewb !

PR #8025, opened 2018-06-24
Ellipse and oval hittesting

### LINKED ISSUE #6985: Oval selection does not update color (does not support hit testing)
When using an oval for a graph node, the oval's color is not updated when selected.

bokeh: 0.12.9

Expected behavior is that the fill_color would be updated like it is with a Circle glyph. Uncomment the Circle glyph lines in the code to see the expected behavior.

```
import math

from bokeh.models import GraphRenderer, StaticLayoutProvider, Oval, TapTool, Circle
output_notebook()

N = 8
node_indices = list(range(N))

graph = GraphRenderer()

graph.node_renderer.glyph = Oval(height=0.1, width=0.2, fill_color="red")
graph.node_renderer.selection_glyph = Oval(height=0.1, width=0.2, fill_color="blue")
# graph.node_renderer.glyph = Circle(size=15, fill_color="red")
# graph.node_renderer.selection_glyph = Circle(size=15, fill_color="blue")
    index=node_indices)

    start=[0]*N,
    end=node_indices)

### start of layout code

plot.renderers.append(graph)
plot.add_tools(TapTool())

# output_file("graph.html")
show(plot)

```

anthonydouc:
WIP to add in hit testing for ovals and ellipse.  Will need some further direction to complete this PR

- [X] issues: fixes #6985
- [X] tests added / passed


anthonydouc:
The initial additions I have made allow for hover to work, however making a selection seems to be bugged. So some further guidance as to what needs to be changed to allow that would be good.

anthonydouc:
Here is example code that shows hover working, but fails on tap selection. Tap produces the error "Uncaught TypeError: Cannot read property '0' of undefined"

```
import numpy as np

from bokeh.models import ColumnDataSource, DataRange1d,LinearAxis, Grid
from bokeh.models.glyphs import Ellipse

N = 9
x = np.linspace(-2, 2, N)
y = x**2
mult = 1*100
w = mult*(x/15.0 + 0.3)
h = mult*(y/20.0 + 0.3)

source = ColumnDataSource(dict(x=x, y=y, w=w, h=h))

xdr = DataRange1d()
ydr = DataRange1d()

plot = figure(
    title=None, x_range=xdr, y_range=ydr, plot_width=300, plot_height=300,
    h_symmetry=False, v_symmetry=False, min_border=0, tools='tap,hover')

plot.ellipse(x="x",y="y",width="w",height="h",angle=-0.7, source=source, height_units='screen', width_units='screen')


show(plot)
```

bryevdv:
This is a bug in the existing ellipse implementation. For reasons that are somewhat obscure, it this diff is the correct code:
```
diff --git a/bokehjs/src/lib/models/glyphs/ellipse.ts b/bokehjs/src/lib/models/glyphs/ellipse.ts
index 5c85a289a..73b8055e5 100644
--- a/bokehjs/src/lib/models/glyphs/ellipse.ts
+++ b/bokehjs/src/lib/models/glyphs/ellipse.ts
@@ -54,13 +54,13 @@ export class EllipseView extends XYGlyphView {
       this.sh = this._height
   }

-  protected _render(ctx: Context2d, indices: number[], {sx, sy, sw, sh}: EllipseData): void {
+  protected _render(ctx: Context2d, indices: number[], {sx, sy, sw, sh, _angle}: EllipseData): void {
      for (const i of indices) {
-       if (isNaN(sx[i] + sy[i] + sw[i] + sh[i] + this._angle[i]))
+       if (isNaN(sx[i] + sy[i] + sw[i] + sh[i] + _angle[i]))
          continue

        ctx.beginPath()
-       ctx.ellipse(sx[i], sy[i], sw[i]/2.0, sh[i]/2.0, this._angle[i], 0, 2 * Math.PI)
+       ctx.ellipse(sx[i], sy[i], sw[i]/2.0, sh[i]/2.0, _angle[i], 0, 2 * Math.PI)

        if (this.visuals.fill.doit) {
          this.visuals.fill.set_vectorize(ctx, i)
```

anthonydouc:
Ok that seems to address the issue. Does that change also need to apply for hit testing? Seems to work keeping this._angle

anthonydouc:
Added in oval hit test code, seems to work well for oval glyphs. When looking at a graph layout, the hover tooltip seems to have two arrows for some glyphs.

graph_hover

bryevdv:
@anthonydouc looks good I think the hist testing helper functions should be moved to this module:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/core/hittest.ts

Then, if you are up for it, it would be good to add a few unit tests as well. Let me know your thoughts.

anthonydouc:
Aside from any other issues, not sure if this PR should be merged until we know whats behind the double arrow displayed on hover for graph renderers.

bryevdv:
> until we know whats behind the double arrow displayed on hover for graph renderers.

I thought I replied to this, but not sure where it went... graph renderers have two sub-renderers, one for the nodes and one for the edges. By default without any configuration a hover tool applies to all available renderers, I expect this is what you are seeing. The specific result and appearance is a little strange and possibly worth investigation but it should not hold up this PR

bokehjs/src/lib/core/hittest.ts
```diff
@@ -24,6 +24,26 @@ export function point_in_poly(x: number, y: number, px: number[], py: number[]):
...
+
+export function point_in_oval(x: number, y: number, angle: number, b: number, a: number, x0: number, y0: number): boolean {
+  // width of oval from cubic bezier = 3/4 * width ellipse
+  a = a * 0.75
+  const A = ((Math.cos(angle) / a) ** 2 + (Math.sin(angle) / b) ** 2)
+  const B = 2 * Math.cos(angle) * Math.sin(angle) * ((1 / a) ** 2 - (1 / b) ** 2)
+  const C = ((Math.cos(angle) / b) ** 2 + (Math.sin(angle) / a) ** 2)
+  const eqn = A * (x - x0) ** 2 + B * (x - x0) * (y - y0) + C * (y - y0) ** 2
+  const inside = eqn <= 1
+  return inside
```
bryevdv: These two functions look almost identical, couldn't this just be:
```typescript
// width of oval from cubic bezier = 3/4 * width ellipse
export function point_in_oval(x: number, y: number, angle: number, b: number, a: number, x0: number, y0: number): boolean {
    return point_in_ellipse(x, y, angle, b, 0.75*x, x0, y0)
}

bryevdv:
@anthonydouc I have one small comment about the code organization, otherwise looks good

bokehjs/src/lib/core/hittest.ts
anthonydouc: Yes I agree it is largely redundant, left them seperate incase that was preferable for others.

bryevdv: Less code to test and maintain is always preferable :)

bryevdv:
Thanks @anthonydouc !

PR #8027, opened 2018-06-25
handle Pandas PeriodIndex and Timedelta properly

### LINKED ISSUE #2266: Pandas periods are not recognised
When I try to use pandas periods (generated with pandas.period_range) as x axis, I get `ms` instead of dates.
Code example:

``` python
import pandas as pd
import numpy as np

dates = index=pd.period_range('1900-01-01','2000-01-01', freq='A')
ts = np.random.randn(100)

p = figure(x_axis_type="datetime")
p.line(dates,ts)
show(p)
```

If I use:

``` python
dates = index=pd.date_range('1900-01-01','2000-01-01', freq='A')
```

I get proper dates.

### LINKED ISSUE #5426: It is not possible to set a Range1d with datetime.timedelta value
I am plotting a series of datetime.timedelta values on one of the axes of my chart. I am trying to constraint the axes minimum and maximum by: 

    yrange1 = Range1d(start=datetime.timedelta(minutes=1,seconds=45),end=datetime.timedelta(minutes=3,seconds=30))

plot.y_range = yrange1

Got the following: 

Exception Type:	ValueError
Exception Value:	
expected an element of either Float, Datetime or Int, got datetime.timedelta(0, 210)

Bokeh version 0.12.3
Python version 2.7.12

bryevdv:
- [x] issues: fixes #2266 fixes #5426
- [x] tests added / passed

Noticed quite a bit of silliness in working on those two issues, so added some other cleanup. This PR:

* adds proper handling for Pandas `Period` and `PeriodIndex`, also making sure the binary protocol is used for Pandas series when it should be

* clarifies internally that time delta values are not datetimes

* allows Range1d to accept time delta values

* allows DataRange1d to also accept datetimes and time delta values (cannot think of any reason they should not)

* fixes the bizarre `url = NumberSpec` type on `ImageURL`

Thanks to @miccoli for discussion and prompting

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -46,7 +46,7 @@ import numpy as np
 
 from ..settings import settings
 from ..util.dependencies import import_optional
-from ..util.serialization import convert_datetime_type, is_datetime_type, transform_series, transform_array
+from ..util.serialization import convert_datetime_type, convert_timedelta_type, is_datetime_type, is_timedelta_type, transform_series, transform_array
 
 pd = import_optional('pandas')
 rd = import_optional("dateutil.relativedelta")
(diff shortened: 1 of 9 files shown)


bryevdv:
With this sample code:
```

import datetime as dt

p = figure(y_range=(0, 10),
           x_range=(-dt.timedelta(seconds=5), dt.timedelta(seconds=3)) )
show(p)
```

screen shot 2018-06-25 at 17 16 12

Issue #7959, opened 2018-06-05
Image hover broken on axes with inverted range


philippjfr:
When plotting an image on an axis with an inverted range the hover tool behaves incorrectly. This example reproduces the issue (note the ``y_range=(10, 0)``):

```python
import numpy as np
from bokeh.plotting import figure, output_file, show, ColumnDataSource
from bokeh.models import HoverTool

output_file("tools_hover_tooltip_image.html")

ramp = np.array([np.linspace(0, 10, 200)]*20)

data = dict(
    image=[ramp], squared=[ramp**2], pattern=['smooth ramp'],
    x=[0], y=[5], dw=[20], dh=[10]
)

cds = ColumnDataSource(data=data)
hover = HoverTool(tooltips=[
    ('index', "$index"),
    ('pattern', '@pattern'),
    ("x", "$x"),
    ("y", "$y"),
    ("value", "@image"),
    ('squared', '@squared')], )

p = figure( x_range=(0, 20), y_range=(10, 0), tools=[hover, 'wheel_zoom'])
p.image(source=cds, image='image', x='x', y='y', dw='dw', dh='dh', palette="Inferno256")
show(p)
```

<img width="554" alt="screen shot 2018-06-05 at 1 28 44 pm" src="https://user-images.githubusercontent.com/1550771/40975933-6b140f50-68c4-11e8-84e9-726593daa003.png">

Version: Bokeh 0.13.0dev5-2-g0504e1fa8

@jlstevens

mattpap:
Is this really a problem with hover or the image is in a wrong location? Changing `y_range` doesn't affect the placement of the image.

philippjfr:
>Shouldn't the image be flipped when an axis is inverted?

While it might be argued that the image should flip automatically, this has never been the case, so I'd argue it's going to cause major compatibility issues to switch this behavior now. If you want to properly position the image on a flipped axis you currently have to change the x/y coordinate of the image. It may have been less confusing if I had done that in the example above, which would look like this:

```python
data = dict(
    image=[ramp], squared=[ramp**2], pattern=['smooth ramp'],
    x=[0], y=[10], dw=[20], dh=[10]
)
```

This also does not work when hovering.

philippjfr:
@jlstevens Any chance you can look into this since you're most familiar with the code?

jlstevens:
I'll try to have a look at this soon although I don't know messy/hard it will be to fix.

PR #8031, opened 2018-06-27
Image hover fix for inverted ranges

jlstevens:
Simple PR to fix the ``ImageView._lrtb`` method, ensuring it returns the correct bounds when the ranges are inverted. Addresses issue #7959.

Here is an example where the y range starts at ten and ends at zero (see the screenshot in #7959 as an example of the incorrect behavior):

image

The code is relatively straightforward although I am not sure I've used the best approach for detecting the inverted ranges.

### DIFF
--- a/bokehjs/src/lib/models/glyphs/image.ts
+++ b/bokehjs/src/lib/models/glyphs/image.ts
@@ -73,14 +73,13 @@ export class ImageView extends XYGlyphView {
   }
 
   _lrtb(i: number) : [number, number, number, number]{
-    const xflip = (this.renderer.xscale.source_range.start >
-                   this.renderer.xscale.source_range.end)
-    const yflip = (this.renderer.yscale.source_range.start >
-                   this.renderer.yscale.source_range.end)
+    const xr = this.renderer.xscale.source_range
     const x1 = this._x[i]
-    const x2 = xflip ? x1 - this._dw[i] : x1 + this._dw[i]
+    const x2 = xr.is_reversed ? x1 - this._dw[i] : x1 + this._dw[i]
+
+    const yr = this.renderer.yscale.source_range
     const y1 = this._y[i]
-    const y2 = yflip ? y1 - this._dh[i] : y1 + this._dh[i]
+    const y2 = yr.is_reversed ? y1 - this._dh[i] : y1 + this._dh[i]
     const [l,r] = x1 < x2 ? [x1,x2] : [x2,x1]
     const [b,t] = y1 < y2 ? [y1,y2] : [y2,y1]
     return [l, r, t, b]
--- a/bokehjs/src/lib/models/glyphs/image.ts
+++ b/bokehjs/src/lib/models/glyphs/image.ts
@@ -73,10 +73,16 @@ export class ImageView extends XYGlyphView {
   }
 
   _lrtb(i: number) : [number, number, number, number]{
-    const l = this._x[i]
-    const r = l + this._dw[i]
-    const b = this._y[i]
-    const t = b + this._dh[i]
+    const xflip = (this.renderer.xscale.source_range.start >
+                   this.renderer.xscale.source_range.end)
+    const yflip = (this.renderer.yscale.source_range.start >
+                   this.renderer.yscale.source_range.end)
+    const x1 = this._x[i]
+    const x2 = xflip ? x1 - this._dw[i] : x1 + this._dw[i]
+    const y1 = this._y[i]
+    const y2 = yflip ? y1 - this._dh[i] : y1 + this._dh[i]
+    const [l,r] = x1 < x2 ? [x1,x2] : [x2,x1]
+    const [b,t] = y1 < y2 ? [y1,y2] : [y2,y1]
     return [l, r, t, b]
   }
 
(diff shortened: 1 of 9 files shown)


bryevdv:
@jlstevens a very little more work will improve this I think. There is an `is_reversed` property on `Range1d`. I'd propose adding that as a required interface to `Range` and then:

* for categorical ranges it just always returns `False`
* for data range it returns the value of `this.flipped`

Then:

* it's easy to add some simple unit tests for these 
* the code in this PR can be simplified to:
  ```
  const xr = this.renderer.xscale.source_range
  const x1 = this._x[i]
  const x2 = xr.is_reversed ? x1 - this._dw[i] : x1 + this._dw[i]
  ```
* this spot can also be fixed up:
  ```
  ./models/plots/plot_canvas.ts:
    467 :       const is_reversed = (rng.start > rng.end)
    485 :         if (is_reversed) {
  ```

bryevdv:
@jlstevens I think that line noted above in `models/plots/plot_canvas.ts` can be fixed up now to use `is_reversed` as well?

bokehjs/src/lib/models/glyphs/image.ts
```diff
@@ -73,10 +73,15 @@ export class ImageView extends XYGlyphView {
...
-    const r = l + this._dw[i]
-    const b = this._y[i]
-    const t = b + this._dh[i]
+    const xr = this.renderer.xscale.source_range
+    const x1 = this._x[i]
+    const x2 = xr.is_reversed ? x1 - this._dw[i] : x1 + this._dw[i]
+
+    const yr = this.renderer.yscale.source_range
+    const y1 = this._y[i]
+    const y2 = yr.is_reversed ? y1 - this._dh[i] : y1 + this._dh[i]
```
bryevdv: style nit: would say either add a blank line after these three "y" lines of code, or don't add one after the three "x" lines of code

jlstevens: Addressed in 9fd3fd32d

jlstevens:
@bryevdv I've made the suggested changes including the simplification in ``plot_canvas.ts``.

bryevdv:
Looks great!

bryevdv:
Well, actually, do you mind adding a few tiny unit tests for `is_reversed` ?

jlstevens:
I've added some unit tests for ``is_reversed`` though I just want to check with you what you expect the behavior to be in this case:

```coffeescript
 describe "explicit inverted bounds=(20,10) creation", ->
   r = new DataRange1d({start: 20, end:10})

   it "should be reversed", ->
     expect(r.is_reversed).to.be.equal true
```

This unit test fails because the ``flipped`` property is false even though the start value is greater than the end. Maybe the correct thing to do is to compare the start and end values *after* flipped is potentially applied instead of just returning ``this.flipped``?

bryevdv:
@jlstevens `flipped` is something a user configures, to say "I want auto-range to put end < start" I guess the situation you see in the failed test can arise (and only arise) in the case that a user *initialized* a `DataRange1d` that way (because if the initial start/end values are None, then auto-ranging happens thereafter). 

I guess this simplest and most correct thing to do is put a single `start > end` test on the `Range` and nothing on the subclasses.

jlstevens:
> I guess this simplest and most correct thing to do is put a single start > end test on the Range and nothing on the subclasses.

I agree. It is good to keep the semantics as simple as possible.

I've implemented this suggestion in d8d8e1b and 94f2ccc.

bokehjs/test/models/ranges/data_range1d.coffee
```diff
@@ -67,6 +67,12 @@ describe "datarange1d module", ->
     it "should have max = 20", ->
       expect(r.max).to.be.equal 20
 
+ describe "explicit inverted bounds=(20,10) creation", ->
+   r = new DataRange1d({start: 20, end:10})
+
+   it "should be reversed", ->
+     expect(r.is_reversed).to.be.equal true
+
```
bryevdv: Just noting here that ideally we'd also have a test that initializes with `flipped=True` and lets `start` and `end` be filled in by auto-ranging, but that would be a pain to do unfortunately

psyrocloud:
I'm using bokeh 1.0.4 in anaconda, when inverse the y-axis and show a up-down flipped image, the hover tool is showing @image value incorrectly. The value of current pixel is the y-mirror position's pixel value. The example on top of this page cannot test out my problem, because the value of image pixel is same along y axis.

thoth291:
I second the comment above.
```python
from matplotlib import pyplot as plt
from skimage.data import astronaut
c = astronaut().mean(axis=-1)[::2,:]
plt.imshow(c)
plt.show()
```
Gives me this
Screen Shot 2019-03-21 at 10 52 08 PM

While this bokeh (1.0.4) code
```python
scale = 1.0
p = figure(x_range=(1, c.shape[1]), y_range=(c.shape[0],1),
           plot_width=int(c.shape[1]*scale), plot_height=int(c.shape[0]*scale),
           tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")],
           tools=["hover"])
print(c[89:91,399:401])
p.image(image=[np.flipud(c)],
        x=1, y=c.shape[0],
        dw=c.shape[1], dh=c.shape[0])
show(p)
```
gives me this
Screen Shot 2019-03-21 at 10 59 39 PM

As you can see - image value is not correct - it seems to be reading it from different region of the image.
Hope this reproducer is clear enough to help dig into it.
Please note that it crucially important to keep Y axis increasing from top to bottom.

bryevdv:
Well, I guess this was fixed, then somehow broken again @thoth291 at this point please make a new issue with all that information, so we can track and prioritize this.

PR #8033, opened 2018-06-27
warning for large grouped legends

### LINKED ISSUE #5365: Plot legend shows value instead of text specified
On a Plot legend (bokeh 0.12.3), if the name for the legend is the same as the name for the column in the data source, the value will be displayed int he legend instead of the name. Is this expected behavior?

Example:

```
from bokeh import plotting
import pandas as pd

data = [{"right1": 10, "right2": 14, "right3": 17,
         "right4": 22, "right5": 37, "right6": 44,
         "right7": 49, "right8": 59}]

source = pd.DataFrame(data)
source = ColumnDataSource(source)

fig = plotting.figure(width=1200, height=900)

fig.hbar(y=1, right="right1", left=0, height=0.5, fill_color="blue", source=source, legend="right1")
fig.hbar(y=1, left="right1", right="right2", height=0.5, fill_color="green",
         line_color="black", source=source, legend="right2")
fig.hbar(y=1, left="right2", right="right3", height=0.5, fill_color="yellow",
         line_color="black", source=source, legend="right3")
fig.hbar(y=1, left="right3", right="right4", height=0.5, fill_color="orange",
         line_color="black", source=source, legend="right4")
fig.hbar(
    y=1, left="right4", right="right5", height=0.5, fill_color="red", line_color="black", source=source,
    legend="right 5")
fig.hbar(
    y=1, left="right5", right="right6", height=0.5, fill_color="purple", line_color="black", source=source,
    legend="right 6")
fig.hbar(y=1, left="right6", right="right7", height=0.5,
         fill_color="grey", line_color="black", source=source, legend="7.")
fig.hbar(
    y=1, left="right7", right="right8", height=0.5, color="black", line_color="black", source=source, legend="*8.",
    fill_alpha=.12, line_alpha=.12)

curdoc().add_root(fig)

show(fig)
```

As you can see, in the legend for the first 4 hbar(s), the value for that hbar is displayed instead of the text specified in `legend=`

bryevdv:
- [x] issues: fixes #5365
- [x] tests added / passed

OK @bokeh/dev @birdsarah @philippjfr @jlstevens here is an implementation of an approximate  validation warning for when a user does `legend="foo"` and gets a grouped legend without intending that. Do we want to do this? I guess I am questioning about performance concerns. Thoughts:

* This does not affect any case with explicit `value` or with a legend that does not match a column (or for that matter, a bad column name)
* For cases where grouping will happen, if the user to changes to `value` as a result of the warning then there is no ultimate cost once the change is made
 
So the cost (of uniquing the columns) is borne by anyone who *does* want grouped legends. Is this reasonable? If we do, is 7 the right heuristic cutoff to warn?


bryevdv:
ping @mrocklin in case you have thoughts as well

bryevdv:
I guess I am inclined not to do this after all. Apart from the performance, there is no good way to silence the warning in cases where the warning fires, but a large legend *is* intended.

PR #8036, opened 2018-06-28
Added x and y values to the geometries property of cb_data when tapto…

### LINKED ISSUE #8013: Expose the more detailed geometry data to the tap tool
When a callback is written for the tap_tool, it exposes the screen geometry of the point that was tapped, available as 
```
cb_data["geometries"]["sx"]
cb_data["geometries"]["sy"]
```
However, to be truly useful it should also provide the converted data coordinates as 
```
cb_data["geometries"]["x"]
cb_data["geometries"]["y"]
```
 This would emulate the hover tool functionality. (Apart from calling one geometry and the other geometries)

There are many cases where it is useful to know exactly what the data values are when a point is tapped. On a line, the point tapped may be somewhere between two distant points, and we need the extrapolated values for that exact point along the line.

Currently, the only way to get these values is to hack an additional callback on the hover tool, and save the values in a CDS so we can interrogate them during the tap tool callback. This is very expensive as we are using the hover event to continually store this data even if we never hit the tap tool.

Below is a current example of this hack:
```
# Create a CDS so we can store the hover x and y coordinates.
locator = ColumnDataSource(name="pred_locator", data=dict(x=[0], y=[0]))

# Define a call back for the hover tool
callback_hover = CustomJS(args=dict(locator=locator), code="""
         geometry = cb_data["geometry"];
         locator.data["x"][0] = geometry["x"];
         locator.data["y"][0] = geometry["y"];
         """)

# define a callback for the tap tool
callback_tap = CustomJS(args=dict(locator=locator), code="""
         geometry = cb_data["geometries"];

         var y_data_val = locator.data["y"][0];
         var y_screen_val = geometry["sy"]; 
         alert("Tapped at " + y_screen_val + ", actual value is " + y_data_val);
         """)


plot.add_tools(HoverTool(callback=callback_hover))
plot.add_tools(TapTool(callback=callback_tap))

```
This is related to the original request to expose the geometry to the tap tool.

Expose Geometries to TapTool Callbacks #4082

euandekock:
Addes two additional variables to the cb_data.geometries object when a callback is called from the taptool.

x, and y provide values for the calculated data values for the x and y screen coordinates sx and sy

- [x] issues: fixes #8013 
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -394,6 +394,13 @@ class TapTool(Tap):
     its ``.source`` attribute and the selection geometry as its
     ``.geometries`` attribute.
 
+    The ``.geometries`` attribute has 5 members.
+    ``.type`` is the geometry type, which always a ``.point`` for a tap event.
+    ``.sx`` and ``.sy`` are the screen X and Y coordinates where the tap occurred.
+    ``.x`` and ``.y`` are the converted data coordinates for the item that has
+    been selected. The ``.x`` and ``.y`` values are based on the axis assiged
+    to that glyph.
+
     .. note::
         This callback does *not* execute on every tap, only when a glyphs is
         "hit". If you would like to execute a callback on every mouse tap,
--- a/bokehjs/src/lib/models/tools/gestures/tap_tool.ts
+++ b/bokehjs/src/lib/models/tools/gestures/tap_tool.ts
@@ -2,7 +2,7 @@ import {SelectTool, SelectToolView} from "./select_tool"
 import * as p from "core/properties"
 import {TapEvent} from "core/ui_events"
 import {isFunction} from "core/util/types"
-import {Geometry, PointGeometry} from "core/geometry"
+import {PointGeometry} from "core/geometry"
 import {DataSource} from "../../sources/data_source"
 
 export class TapToolView extends SelectToolView {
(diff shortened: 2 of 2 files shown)


bryevdv:
@euandekock looking at this I think the proper way to do things here is to change the definition of the `PointGeometry` type in `core/geometries` to contain `x` and `y`. And then to be consistent, we should update the other `Geometry` types to contain similar data space coords as well, and make similar conversions where they are used. Is that a task you are willing to take on?

euandekock:
@bryevdv I'm about to start on a new project from Monday and won't be using any charting for the next month or so, but I'll see if I can work on it in the evenings for a few days.

I agree that the proper way is to extend all the Geometry classes, but I suspect this is going to be quite a deep dive into the guts of Bokeh's typescript. I'll try to implement the changes for Tap and Hover tools initially and will commit to completing the rest after I've got a better feel for the size of the task.

Are you happy to keep the property names as x and y, or should we call them dx and dy to differentiate between sx = screen X, dx = data X ?

bryevdv:
@euandekock so a point in favor of adding the values to the geometry classes is that they will otherwise not be available to  Bokeh server apps (the information needed to compute the data values is only available in the client).  I suppose a very correct way to go about is to have a set of internal types (the current ones) that only have screed coords, and then have a different set of types for the user-visible values, which might have data space transformed coords (maybe configurable).  However, as you note that is some amount of work, and I have not perceived a large user demand for this, so if you are content to leave things as they are, I am fine with that and we can close this PR and the issue.  Perhaps if more people request it we can revisit later. 

Regarding the idea to rename `geometries` why don't we discuss on the gitter dev channel sometime and see if there is any action that can be taken: https://gitter.im/bokeh/bokeh-dev

PR #8037, opened 2018-06-29
DOC typo (funtion -> function)

lukemcraig:


bryevdv:
Thanks for the catch and fix @lukemcraig

PR #8038, opened 2018-06-29
improved support for pyinstaller

axil:
A better treatment for freezing bokeh with pyinstaller.

The `_templates` dir should better stay in its original subdirectory than in the project root as implemented in #8008, as it better fits pyinstaller directory layout: as documented [here](https://pythonhosted.org/PyInstaller/hooks.html), `collect_data_files()` as opposed to `--add-data` doesn't support the destination path.

See my [PR3607](https://github.com/pyinstaller/pyinstaller/pull/3607) with the associated pyinstaller hook file which uses the new _templates location.

### DIFF
--- a/bokeh/core/templates.py
+++ b/bokeh/core/templates.py
@@ -25,7 +25,7 @@ def get_env():
     ''' Get the correct Jinja2 Environment, also for frozen scripts.
     '''
     if getattr(sys, 'frozen', False):
-        templates_path = os.path.join(sys._MEIPASS, '_templates')
+        templates_path = os.path.join(sys._MEIPASS, 'bokeh', 'core', '_templates')
         return Environment(loader=FileSystemLoader(templates_path))
     else:
         return Environment(loader=PackageLoader('bokeh.core', '_templates'))


bryevdv:
@axil Just to be clear, this will require minimal version of pyinstaller (i.e. the next release after that other PR is merged)? Is that something we can document?

axil:
This PR is more crucial than the pyinstaller one: once this PR is merged, the pyinstaller hook I've written can be used in a standalone fashion even before it is merged into pyinstaller:

Copy [hook-bokeh.py](https://github.com/axil/pyinstaller/blob/1d2c0babb701e2b731e1ff7a65b983e781f8fbad/PyInstaller/hooks/hook-bokeh.py) file into the project dir root (say, where `mybokehscript.py` resides) and run

    pyinstaller mybokehscript.py --additional-hooks-dir=.

Here're some more details on the subject: https://github.com/pyinstaller/pyinstaller/issues/1898

bryevdv:
> On the other hand, the pyinstaller PR won't work without merging this PR first.

Right I figured that, I'm just asking if there is already an idea of what the future pyinstaller version version will be, if so we could pro-actively document it.

axil:
Not really sure. I'm not affiliated with them in any way.

bryevdv:
OK well let's go ahead and merge this, then.

axil:
Adapted "sliders.py" example to be packaged with pyinstaller: https://github.com/axil/frozen_sliders

axil:
The hook got merged into PyInstaller devel, milestone 3.4 was assigned.

PR #8055, opened 2018-07-05
remove all deprecations

bryevdv:
Removes all deprecations for upcoming 1.0 release.


bryevdv:
One example test failing due to passing `logo` to `figure`. I removed some of that code as well. Do we want to keep it, or change the example?

mattpap:
> Do we want to keep it, or change the example?

Lets change the example. Changing the logo isn't a particularly often need.

PR #8056, opened 2018-07-05
remove code dependent on IPython pprint machinery

bryevdv:
IPython pretty printer won't work without `_repr_pretty`, having `_repr_pretty` spams the ipython console making it unusable. I am tired of try to work around this, so just removing the whole mess. Perhaps we can add back our own implementation that does not depend on IPython at all in the future. 

cc @mattpap


bryevdv:
@mattpap any objections to merging this today?

mattpap:
Lets merge.

PR #8057, opened 2018-07-06
Added x and y values to the geometries property of cb_data for TapTool

### LINKED ISSUE #8013: Expose the more detailed geometry data to the tap tool

euandekock:
The Tap Tool doesn't expose any direct access to the calculated range-specific x and y data coordinates. This is available in all the Select tools, and in the Hover tool. 

The method used here is similar to the approach used in the Select and Hover tools in that we step outside the Geometry classes and define a naked object to hold all the values. Most user-side js code will not reference or understand these internal classes and will access the values directly.

- [x] issues: fixes #8013 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/tools/gestures/tap_tool.ts
+++ b/bokehjs/src/lib/models/tools/gestures/tap_tool.ts
@@ -2,7 +2,7 @@ import {SelectTool, SelectToolView} from "./select_tool"
 import * as p from "core/properties"
 import {TapEvent} from "core/ui_events"
 import {isFunction} from "core/util/types"
-import {Geometry, PointGeometry} from "core/geometry"
+import {PointGeometry} from "core/geometry"
 import {DataSource} from "../../sources/data_source"
 
 export class TapToolView extends SelectToolView {
@@ -22,14 +22,6 @@ export class TapToolView extends SelectToolView {
   _select(geometry: PointGeometry, final: boolean, append: boolean): void {
     const callback = this.model.callback
 
-    const cb_data: {
-      geometries: Geometry,
-      source: DataSource | null,
-    } = {
-      geometries: geometry,
-      source: null,
-    }
-
     if (this.model.behavior == "select") {
       const renderers_by_source = this._computed_renderers_by_data_source()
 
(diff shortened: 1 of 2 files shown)


euandekock:
@mattpap, please can you review this as per the request from @bryevdv . I'd like to get it in the code to at least provide feature parity between TapTool and the Select and Hover tools, but I'm also open to looking at how we can extend the event subsystem to replace the callback functions.

euandekock:
@mattpap and @brydev FYI I cannot access the gitter discussion forum at work due to network restrictions, so can we keep the discussions in the issue or in the PR. Thanks.

euandekock:
@mattpap , @bryevdv Just following up to see if there is any progress on this? I'd really like to get it completed as I've finished my project that justified my work on it for now.

I'm happy to revisit the broader events subsystem when my new work reaches a stage where I I am back on visualisation work.

bokehjs/src/lib/models/tools/gestures/tap_tool.ts
```diff
@@ -40,7 +32,16 @@ export class TapToolView extends SelectToolView {
...
         if (did_hit && callback != null) {
-          cb_data.source = sm.source
+          const frame = this.plot_model.frame
+          const xscale = frame.xscales[renderers[0].x_range_name]
+          const yscale = frame.yscales[renderers[0].y_range_name]
+          const x = xscale.invert(geometry.sx)
+          const y = yscale.invert(geometry.sy)
+          const g = {...geometry, x, y}
+          const cb_data: {
+              geometries: { type: "point", sx: number, sy: number, x: number, y: number }
```
mattpap: This would be more meaningful if the type was `PointGeometry & {x: number, y: number}`.

euandekock:
OK, thanks for the suggestion - my understanding of TypeScript is limited, so I didn't realise you could do that. Changes made as requested.

PR #8059, opened 2018-07-06
fix annulus hit testing with screen units

### LINKED ISSUE #2022: Not correct behaviour of HoverTool with annular glyph and units='screen'
Hi guys!
I think Ive found a bug,

Add 2 types of annulus glyphs, first - inner and outer radius sizes in 'screen' units, second - in 'data' units.

``` python
import pandas as pd
import numpy as np
import bokeh.plotting as bk
from bokeh.plotting import show, output_file
from bokeh.models.formatters import DatetimeTickFormatter

TOOLS = "pan,wheel_zoom,box_zoom,reset,previewsave,hover"
rng = pd.date_range('1/1/2011', periods=72, freq='H')
np.random.seed(1)
df = pd.DataFrame([[np.random.randint(2), np.random.randint(2)] for _ in range(len(rng))],
                      index=rng,
                      columns=['y_axis_values', 'chart_group'])
fig = bk.figure(title='Test Annulus', tools=TOOLS, x_axis_type='datetime')
fig.annulus(
        x=df[df['chart_group']==0].index,
        y=df[df['chart_group']==0]['y_axis_values'],
        color='blue',
        inner_radius=1,
        outer_radius=3,
        inner_radius_units="screen",
        outer_radius_units="screen")

fig.annulus(
        x=df[df['chart_group']==1].index,
        y=df[df['chart_group']==1]['y_axis_values'],
        color='red',
        inner_radius=500000,
        outer_radius=1500000)
hover = fig.select(dict(type=HoverTool))

hover.tooltips = [
        ("index", "$index"),
    ]

show(fig)
```

If you try to hover red and blue glyphs you can see popups dont appear at blue ones (with "screen" units).

anthonydouc:
Corrects annulus hit testing where outer diameter is specified in screen units. 

Coincidently I noticed you can specify an annulus with an inner diameter greater than the outer diameter, in that case hover fails.

- [X] issues: fixes #2022
- [] tests passed

### DIFF
--- a/bokehjs/src/lib/models/glyphs/annulus.ts
+++ b/bokehjs/src/lib/models/glyphs/annulus.ts
@@ -84,12 +84,25 @@ export class AnnulusView extends XYGlyphView {
   protected _hit_point(geometry: PointGeometry): Selection {
     const {sx, sy} = geometry
     const x = this.renderer.xscale.invert(sx)
-    const x0 = x - this.max_outer_radius
-    const x1 = x + this.max_outer_radius
-
     const y = this.renderer.yscale.invert(sy)
-    const y0 = y - this.max_outer_radius
-    const y1 = y + this.max_outer_radius
+
+    let x0: number, y0: number
+    let x1: number, y1: number
+    if (this.model.properties.outer_radius.units == "data") {
+      x0 = x - this.max_outer_radius;
+      x1 = x + this.max_outer_radius;
+
+      y0 = y - this.max_outer_radius;
+      y1 = y + this.max_outer_radius;
+    } else {
+      const sx0 = sx - this.max_outer_radius;
+      const sx1 = sx + this.max_outer_radius;
+      [x0, x1] = this.renderer.xscale.r_invert(sx0, sx1);
+
+      const sy0 = sy - this.max_outer_radius;
+      const sy1 = sy + this.max_outer_radius;
+      [y0, y1] = this.renderer.yscale.r_invert(sy0, sy1);
+    }
 
     const hits: [number, number][] = []
 


bokehjs/src/lib/models/glyphs/annulus.ts
```diff
@@ -84,12 +84,25 @@ export class AnnulusView extends XYGlyphView {
...
+
+      y0 = y - this.max_outer_radius;
+      y1 = y + this.max_outer_radius;
+    } else {
+      const sx0 = sx - this.max_outer_radius;
+      const sx1 = sx + this.max_outer_radius;
+      [x0, x1] = this.renderer.xscale.r_invert(sx0, sx1);
+
+      const sy0 = sy - this.max_outer_radius;
+      const sy1 = sy + this.max_outer_radius;
```
bryevdv: isn't `this.max_outer_radius` in data units? I would expect this distance would need to be converted to screen units before adding or subtracting to `sx` or `sy`

bryevdv: Or better, to use a max of `souter_radius` since that's the value that is actually used in the call to `ctx.arc`

bryevdv: Or alternatively: for annulus we could just skip the spatial indexing altogether and make things simpler. i.e. just iterate over and test all the annuli always, instead of trying to get a list of candidates from the index first. I doubt users are trying sho show thousands of annulus values

anthonydouc: this.max_outer_radius seems to be in the same units as what is specified. This follows a similar approach to annular wedge, https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/glyphs/annular_wedge.ts#L96.

bryevdv: It appears to be being added to "x" (data units) in the first block and "sx" (screen units) in the second. If so, those can't both be correct

PR #8062, opened 2018-07-07
feature: add built in themes: dark_minimal and light_minimal

### LINKED ISSUE #99: Making predefined themes available by calling a single function
for now accept "bokeh" and "ggplot"

ahuang11:
https://github.com/bokeh/bokeh/issues/99
- [x] issues: fixes #99
- [x] tests added / passed

Examples of it applied:
Without:
without_theme
Lite:
with_lite
Dark Lite:
with_dark_lite

Without:
without_theme_img
Lite:
with_lite_img
Dark Lite
with_dark_lite_img

W/o
without_fourier
lite
with_lite_fourier
dark_lite
without_fourier

From hvplot:
w/o
image
w/
image

More examples of how it compares to
https://pyviz.github.io/hvplot/user_guide/Gridded_Data.html
https://anaconda.org/ahuang11/hvplot_theme/notebook
https://anaconda.org/ahuang11/hvplot_dark_theme

Issues noticed:
the title doesn't update according to the theme (holoviews specific)
the settings set by the theme do not seem to be overwriteable (holoviews specific?)
the theme doesn't scale well for subplots and categorical x (font size is too large, and get's crowded)
box whisker's are black in the dark theme

### DIFF
--- a/bokeh/themes/dark_lite.json
+++ b/bokeh/themes/dark_lite.json
@@ -1,0 +1,66 @@
+{
+    "attrs": {
+        "Figure" : {
+            "background_fill_color": "#20262B",
+            "border_fill_color": "#15191C",
+            "outline_line_color": "#E0E0E0",
+        },
+
+        "Axis": {
+            "major_tick_line_alpha": 0,
+            "major_tick_line_color": "#E0E0E0",
+
+            "minor_tick_line_alpha": 0,
+            "minor_tick_line_color": "#E0E0E0",
+
+            "axis_line_alpha": 0,
+            "axis_line_color": "#E0E0E0",
+
+            "major_label_text_color": "#E0E0E0",
+            "major_label_text_font": "Helvetica",
+            "major_label_text_font_size": "1.05em",
+
+            "axis_label_standoff": 10,
+            "axis_label_text_color": "#E0E0E0",
+            "axis_label_text_font": "Helvetica",
+            "axis_label_text_font_size": "1.35em",
+            "axis_label_text_font_style": "normal",
+        },
+
+        "Legend": {
+            "spacing": 8,
+            "glyph_width": 15,
+
+            "label_standoff": 8,
+            "label_text_color": "#E0E0E0",
+            "label_text_font": "Helvetica",
+            "label_text_font_size": "1.05em",
+
+            "border_line_alpha": 0,
+            "background_fill_alpha": 0.25,
+            "background_fill_color": "#20262B",
+        },
+
+        "ColorBar": {
+            "title_text_color": "#E0E0E0",
+            "title_text_font": "Helvetica",
+            "title_text_font_size": "1.05em",
+            "title_text_font_style": "normal",
+
+            "major_label_text_color": "#E0E0E0",
+            "major_label_text_font": "Helvetica",
+            "major_label_text_font_size": "1.05em",
+
+            "background_fill_color": "#15191C",
+            "major_tick_line_alpha": 0,
+            "bar_line_alpha": 0,
+
+        },
+
+        "Title": {
+            "text_color": "#E0E0E0",
+            "text_font": "Helvetica",
+            "text_font_size": "1.15em",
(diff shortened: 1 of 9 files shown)


bryevdv:
Only a couple of comments: 

* I think the grid lines on the dark theme are too heavy/bright. Perhaps a smaller width, an alpha, a more muted color, or some combination of these

* "Dark Lite" is not a good name, I couldn't figure out at first what it would mean for something to be dark and light at the same time, until it occurred that "lite" refers (I presume) to a notion of "visual weigh" I think name has to change. If you want this to be a pair, "X" and "Dark_X" then "X" needs to be something besides "lite". But I think it might be even better to make the names symmetric, i.e. "LIght_X" and "Dark_X" where "X" is something new.

bokeh/themes/dark_lite.json
```diff
@@ -0,0 +1,66 @@
+{
+    "attrs": {
+        "Figure" : {
+            "background_fill_color": "#20262B",
+            "border_fill_color": "#15191C",
+            "outline_line_color": "#E0E0E0",
```
mattpap: JSON specification doesn't allow for trailing commas. Perhaps we use an extended (aka. non-compliant) parser, but still it would be good to have a *.json file contain valid JSON.

mattpap: You can run `python -m json.tool some_file.json` to verify if `some_file.json` parses correctly.

ahuang11:
Examples of the newer dark_minimal https://anaconda.org/ahuang11/hvplot_dark_theme

bryevdv:
@ahuang11 let's not worry about the HTML in this PR. However, to close #99 let's make it simpler to set themes. I would propose making a small change to the `Document.theme` setter:

https://github.com/bokeh/bokeh/blob/master/bokeh/document/document.py#L142-L152

so that it can also accept strings, in which case it tries to load the theme from `bokeh.themes` using that name. Does that sound reasonable? A simple test would be warranted as well.

jlstevens:
@ahuang11 Looks good! I like the dark theme as shown in the last notebook but personally I find the text (e.g axis ticks) are a bit too large...

philippjfr:
I like the fact that the ticks are slightly larger than default but I agree that their size could be reduced slightly. Also agree on merging sooner rather than later, there should be plenty of time to make adjustments before a 1.0 release if we find cases where the themes aren't resulting in nice behavior. The best way to find those will be to merge and test them as much as possible.

bryevdv:
Ok I think there is a general preference to reduce the tick label size slightly @ahuang11 do you mind doing that? Also I think there are some globals like `FP_FMT` and `THIS_DIR` that could be made "private" by putting an underscore in front. If you make these changes I think we are ready to merge!

jlstevens:
@bryevdv @ahuang11 Thinking about it, wouldn't it be good to have a dark theme that matches 'dark lite' with *only* the color changes? This would be an additional theme as dark lite also looks useful!

jbednar:
I agree with reducing the size and with having a separate theme that is simply the color changes and not the size changes.  Also, the name `dark_lite` is a bit confusing, which could perhaps be addressed at the same time as having such a separate theme (each named to keep them all clearly distinct from each other)?

jbednar:
Oops; looks like it's now dark_minimal, which is clearer, but if there's a separate dark theme that only has color changes and not size changes, then that one would be more minimal in terms of changes, so it's confusing...

bryevdv:
The themes were renamed already

ahuang11:
Let's end this PR with just the the two themes with reduced sizing?

bryevdv:
@ahuang11 awesome thanks I will merge this when green!

bryevdv:
Thanks @ahuang11 !

philippjfr:
@ahuang11 Thanks for all your effort on these themes, excited to start using them myself. Would you mind filing an issue in holoviews describing any issues you encountered when applying these themes?

PR #8067, opened 2018-07-09
add header_row property to support hiding table headers

### LINKED ISSUE #6454: Add an option to disable column titles in DataTable

This issue is being reported for bokeh version 0.12.6.

DataTable widgets do not include an option to hide column titles. It would be nice to have that option.  

Coincidentally, the `row_headers` property is poorly named. It turns on/off the index column, but its name seems to suggest it would toggle column titles. Perhaps that option should be redefined.

```
data = dict(
    )

columns = [
    ]

data_table = DataTable(source=source, columns=columns, row_headers=False, editable=True)
```

screenshot

bryevdv:
- [x] issues: fixes #6454
- [ ] tests added / passed

This is not an especially important feature in and of itself, but I plan to use this issue to figure out how to ad back some selenium tests on Travis.

### DIFF
--- a/bokeh/models/widgets/tables.py
+++ b/bokeh/models/widgets/tables.py
@@ -571,4 +571,8 @@ class DataTable(TableWidget):
     in the viewport.
     """)
 
+    header_row = Bool(True, help="""
+    Whether to show a header row with column names at the top of the table.
+    """)
+
     height = Override(default=400)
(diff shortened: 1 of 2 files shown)


bryevdv:
Actual test overhaul will be a bigger PR, merging thins now

PR #8068, opened 2018-07-09
clamp bar endpoints to viewport

### LINKED ISSUE #8060: Inaccurate rendering around 0 on tall bars
I have a distribution with very tall bars, and a few really short bars.

When I plot a close up to just see the short bars, you can see that the tall bars do not correctly start at 0, they start just above 0.

bokeh_plot

bokeh_plot

You can keep zooming in to the bottom of those tall bars and it will still always render just above 0:

bokeh_plot

Data: https://github.com/birdsarah/bokeh-miscellany/blob/master/all_counts.csv
Code: https://github.com/birdsarah/bokeh-miscellany/blob/master/all_counts.py
Rendered html: http://htmlpreview.github.io/?https://github.com/birdsarah/bokeh-miscellany/blob/master/all_counts.html

```python
import pandas as pd

data = pd.Series.from_csv('all_counts.csv')
p_all = figure(tools='save', title='all')
p_all.vbar(data.index, top=data.values, width=0.9)

p_close = figure(tools='save, box_zoom, reset', title='closeup', x_range=(0, 250), y_range=(0, 20))
p_close.vbar(data.index, top=data.values, width=0.9)

l = layout([[p_all], [p_close]])
output_file('all_counts.html')
show(l)
```

bryevdv:
- [x] issues: fixes #8060
- [ ] tests added / passed

OP code now generates:

screen shot 2018-07-09 at 12 29 57

Tested manually with `hbar`. Need to verify hover, etc is not impacted. Also need to add at minimum, unit tests for the viewport clamp function, but would like to add some selenium tests on travis directly.

### DIFF
--- a/bokehjs/src/lib/models/glyphs/box.ts
+++ b/bokehjs/src/lib/models/glyphs/box.ts
@@ -63,6 +63,20 @@ export abstract class BoxView extends GlyphView {
     }
   }
 
+  // We need to clamp the endpoints inside the viewport, because various browser canvas
+  // implementations have issues drawing rects with enpoints far outside the viewport
+  protected _clamp_viewport(): void {
+    const hr = this.renderer.plot_view.frame.bbox.h_range
+    const vr = this.renderer.plot_view.frame.bbox.v_range
+    const n = this.stop.length
+    for (let i = 0; i < n; i++) {
+      this.stop[i] = Math.max(this.stop[i], vr.start)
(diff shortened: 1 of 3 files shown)


bryevdv:
Actual test overhaul will be a bigger PR, merging thins now

birdsarah:
Thanks for the quick turnaround on this! 

Would it also not apply to rect and quad? Any other glyphs?

PR #8069, opened 2018-07-09
Update download_sample_data.py

lancelot1969:
It doesn't reference the missing global variable. Instead it treats the string 'file' as part of the path.

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
@lancelot1969 I am not sure what usage issue you are running into, but I have run this script many times over the years (and it has not been changed in years) and it functions exactly as intended. I can also state without doubt that passing the literal string `"__file__"` instead of `__file__` is not a correct or useful thing to do. So, while we appreciate your interest in contributing, I will close PR. If you are having issues with using this script I'd encourage you to go to the public mailing list for help/support.

PR #8083, opened 2018-07-17
Support for text glyph hit testing

### LINKED ISSUE #7618: Feature Request: support hit-testing for text glyphs
Text glyphs do no currently support hit-testing, so hover or selections (or callbacks based on those) do not work with text.

See: https://groups.google.com/a/continuum.io/forum/#!msg/bokeh/g1zm_uBJFKk/b9y-CtIPAgAJ

anthonydouc:
Added some initial work to support text glyph hit testing. 

- [ ] issues: fixes #7618
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/glyphs/text.ts
+++ b/bokehjs/src/lib/models/glyphs/text.ts
@@ -1,17 +1,23 @@
 import {XYGlyph, XYGlyphView, XYGlyphData} from "./xy_glyph"
 import {NumberSpec, StringSpec, AngleSpec} from "core/vectorization"
 import {TextMixinVector} from "core/property_mixins"
+import {PointGeometry} from "core/geometry";
+import * as hittest from "core/hittest";
 import {Arrayable} from "core/types"
 import * as visuals from "core/visuals"
 import * as p from "core/properties"
 import {get_text_height} from "core/util/text"
 import {Context2d} from "core/util/canvas"
+import {Selection} from "../selections/selection"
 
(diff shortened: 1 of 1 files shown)


anthonydouc:
Two key areas which we may want to improve before this is merged:
Currently there is no generation of candidates for hit testing, and all glyphs are checked. Additionally the scenterx and y methods have been fixed to the centre of the first line of text.

Additionally there may be some scope for code re use from the rect hit testing methods, as the approaches are somewhat similar.

bryevdv:
> Currently there is no generation of candidates for hit testing, and all glyphs are checked.  

I think this is fine in general, I do not anticipate that users would (or should) ever try to display e.g. thousands of text lines. 

> Additionally the scenterx and y methods have been fixed to the centre of the first line of text.

I'm also unconcerned about this at all. User's have asked about clicking on test, I don't really see the point of a hover tool over text. 

My only suggestion would be a an example that demonstrates this new capability

anthonydouc:
```

bryevdv:
Actually I have some ideas about incorporating this into some existing examples but that can happen later. I will plan to look at the PR code more closely this afternoon and get this merged later today.

bryevdv:
At first glance, I think it's a bit novel having hit test information computed during `render` but given how text measurements (sort-of) work on canvas, it makes sense. Thanks for the PR @anthonydouc

anthonydouc:
Yes it is not exactly preferable. It also means the calculations are occurring regardless of whether they are required or not.

PR #8093, opened 2018-07-21
Fixes #6770. Flatten pandas columns and index explicitly on CDS creation

### LINKED ISSUE #6770: Task: Add user warn/error to ColumnDataSource.from_df method about DataFrame containing "index" column

Karel-van-de-Plassche:
Both df.index and df.columns can be a pd.MultiIndex. Bokeh
expects a 'flat' table. This change uses the pandas API to
flatten, solving some issues with naming + being more consistent
with pandas

Supersedes https://github.com/bokeh/bokeh/pull/8002

- [x] issues: fixes #6770
- [x] tests added / passed

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -65,10 +65,15 @@ class ColumnDataSource(ColumnarDataSource):
           source = ColumnDataSource(df)
 
       In this case the CDS will have columns corresponding to the columns of
-      the ``DataFrame``. If the ``DataFrame`` has a named index column, then
-      CDS will also have a column with this name. However, if the index name
-      (or any subname of a ``MultiIndex``) is ``None``, then the CDS will have
-      a column generically named ``index`` for the index.
+      the ``DataFrame``. If the ``DataFrame`` columns have multiple levels,
+      they will be flattend using an underscore (e.g. level_0_col_level_1_col).
+      The index of the DataFrame will be flattened to an ``Index`` of tuples
+      if it's a ``MultiIndex``, and then reset using ``reset_index``. The result
+      will be a column with the same name if the index was named, or
+      level_0_name_level_1_name if it was a named ``MultiIndex``. If the
+      ``Index`` did not have a name or the ``MultiIndex`` name could not be
+      flattened/determined, the ``reset_index`` function will name the index column
+      ``index``, or ``level_0`` if the name ``index`` is not available.
 
     * A Pandas ``GroupBy`` object
 
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -158,7 +158,11 @@ class ColumnDataSource(ColumnarDataSource):
 
         # Flatten columns
         if isinstance(df.columns, pd.MultiIndex):
-            _df.columns = ['_'.join(col) for col in _df.columns.values]
+            try:
+                _df.columns = ['_'.join(col) for col in _df.columns.values]
+            except TypeError:
+                raise TypeError('Could not flatten MultiIndex columns. '
+                                'use string column names or flatten manually')
         # Flatten index
         index_name = ColumnDataSource._df_index_name(df)
         if index_name == 'index':
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -150,16 +150,24 @@ class ColumnDataSource(ColumnarDataSource):
 
         '''
         _df = df.copy()
+
+        # Flatten columns
+        if isinstance(df.columns, pd.MultiIndex):
+            _df.columns = ['_'.join(col) for col in _df.columns.values]
+        # Flatten index
+        index_name = ColumnDataSource._df_index_name(df)
+        if index_name == 'index':
+            _df.index = pd.Index(_df.index.values)
+        else:
+            _df.index = pd.Index(_df.index.values, name=index_name)
+        _df.reset_index(inplace=True)
+
         tmp_data = {c: v.values for c, v in _df.iteritems()}
 
         new_data = {}
         for k, v in tmp_data.items():
-            if isinstance(k, tuple):
-                k = "_".join(k)
             new_data[k] = v
 
-        index_name = ColumnDataSource._df_index_name(df)
-        new_data[index_name] = _df.index.values
         return new_data
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


bokeh/models/tests/test_sources.py
```diff
@@ -53,6 +53,18 @@ def test_init_dataframe_data_kwarg(self, pd):
...
+        data = dict(a=[1, 2], b=[2, 3], index=[4, 5])
+        df = pd.DataFrame(data)
+        ds = ColumnDataSource(data=df)
+        assert set(df.columns).issubset(set(ds.column_names))
+        for key in data.keys():
+            assert isinstance(ds.data[key], np.ndarray)
+            assert list(df[key]) == list(ds.data[key])
+        assert isinstance(ds.data['level_0'], np.ndarray)
+        assert [0, 1] == list(ds.data['level_0'])
+        assert set(ds.column_names) - set(df.columns) == set(["level_0"])
```
bryevdv: Where is `"level_0"` coming from. I have run some of this code by hand in ipython and don't understand this at all.

bryevdv: Oh I guess it comes from `_df.reset_index(inplace=True)`,that's really confusing

bryevdv:
@Karel-van-de-Plassche I think we need to add some documentation support for this PR, if only to have concrete guidance/information to point people to when there are questions. I think the docstring should be expanded, as well as this user's guide section: 

https://bokeh.pydata.org/en/latest/docs/user_guide/data.htm

There's probably enough gotchas with CDS init from Pandas that that could be an entire subsection under "ColumnDataSource". There's already some information about Pandas there, but I think a dedicated subsection with examples of what inputs generate what kind of outputs would be warranted.

Karel-van-de-Plassche:
I'd be happy to write some docs.

I agree that it's somewhat confusing, but reset_index is the pandas way (tm) to handle converting indices to columns, so I'd rather copy that approach, as it would be what pandas users would expect.

bokeh/models/tests/test_sources.py
Karel-van-de-Plassche: Indeed. From the pandas docs:

For DataFrame with multi-level index, return new DataFrame with labeling information in the columns under the index names, defaulting to ‘level_0’, ‘level_1’, etc. if any are None. For a standard index, the index name will be used (if set), otherwise a default ‘index’ or ‘level_0’ (if ‘index’ is already taken) will be used.

Karel-van-de-Plassche:
Also about the way `MultiIndex` is flattened now: the pandas way would be to add each level of the `MultiIndex` in its own column, not create a column of tuples. This might be a backwards-compatibility breaking change though, so I wasn't sure if we'd like to implement that.

bryevdv:
> This is a backwards-compatibility breaking change though, so I wasn't sure if we'd like to implement that.

I don't think so, at least not now

bryevdv:
> I agree that it's somewhat confusing, but reset_index is the pandas way (tm) to handle converting indices to columns, so I'd rather copy that approach, as it would be what pandas users would expect. The is no 'standard' way to flatten the column index, but the one I implemented matches your old implementation and I've seen this way appear more often in pandas projects.

Right I am not suggesting any particular changes, just noting that it's tricky enough that we need to be clear about expectations and behavior, and have examples ready to point people to in SO, gitter, or ML questions.

bryevdv:
> Although I don't know if this is the clearest way to display this info.

I would suggest making two subsections in the Pandas section, e.g. "Named Indices" and "Multi-indices" or something like that, and then show snippets of input/output rather than an entire script. I think in this context the rest is just noise. Some day I'd like to make nice sphinx extension for displaying data frames in the bokeh docs, but that will have to wait till later.

bokeh/models/sources.py
```diff
@@ -150,16 +155,24 @@ def _data_from_df(df):
 
         '''
         _df = df.copy()
+
+        # Flatten columns
+        if isinstance(df.columns, pd.MultiIndex):
+            _df.columns = ['_'.join(col) for col in _df.columns.values]
```
pratapvardhan: A note: This will work nicely with string names. However would break for integer-like column names.

```
pd.Index([('a', 'b'), ('c', 2), ('c', 1), ('d', 'f')])
```

Which I think, is okay.

Also, in pandas you could use `.map` like 

```
_df.columns = _df.columns.map('_'.join)
```

Karel-van-de-Plassche: Thanks for your comments. 

> However would break for integer-like column names.

Indeed. I could cast with `str` here, but that might give trouble with other classed names, so it might be better to just only accept strings here implicitly. @bryevdv ?

> Also, in pandas you could use .map like, although a bit less readable perhaps.

I'd rather go for readability here using the list comprehension.

bryevdv: AFAIK it is somewhat conventionally discouraged to have pandas column names that are numbers, am I wrong in that impression? if it is, then I agree it's not worth worrying about

Karel-van-de-Plassche: By convention 'names' are 'strings' I think. I'm 50/50 on just using`str()` and let the user deal with the fallout. I've added a note to the docs on how bokeh expects a flattened frame.

bryevdv: @Karel-van-de-Plassche I leave it up to you, I'd only ask that if you change then a new test get added to cover the new case. Either way I'm happy to merge as soon as you indicate `ready`

Karel-van-de-Plassche: I noticed the non-multiIndex method only accepts strings anyway. I've opted for re-raising the TypeError with a clearer error message instead and added tests for both the MultiIndex and single indexed column cases. If the tests pass, It'll get a ready from me.

bryevdv:

PR #8094, opened 2018-07-21
Fixes 7890: Do not draw line if value is NaN in step

Karel-van-de-Plassche:
Copied over the code in `line.ts`. Not sure if this maybe can/should be generalized and/or extended to the other line-like glyphs, e.g. arc and bezier.

- [x] issues: fixes #xxxx
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/glyphs/step.ts
+++ b/bokehjs/src/lib/models/glyphs/step.ts
@@ -16,6 +16,9 @@ export class StepView extends XYGlyphView {
   visuals: Step.Visuals
 
   protected _render(ctx: Context2d, indices: number[], {sx, sy}: StepData): void {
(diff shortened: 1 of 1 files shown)


Karel-van-de-Plassche:
It copies the way `line` plots.

Example with line:
bokeh_plot 9
Example with step:
bokeh_plot 10
Script (based on the script in the issue):
``` python

# add a line renderer with a NaN
nan = float('nan')
data = [[1, 2, 3, nan, nan, 8, 9, 10],
        [6, 7, 2, nan, nan, 4, 5, 7]] 

p.step(*data, line_width=2)
#p.line(*data, line_width=2)

show(p)
```

bryevdv:

PR #8095, opened 2018-07-21
Fix Karel/7788 rangeslider minmax

### LINKED ISSUE #7788: DateRangeSlider error not logged nor shown in CLI

Karel-van-de-Plassche:
- [x] issues: fixes #7788
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/validation/errors.py
+++ b/bokeh/core/validation/errors.py
@@ -71,6 +71,9 @@ validation checks.
 1020 *(BAD_EXTRA_RANGE_NAME)*
     An extra range name is configued with a name that does not correspond to any range.
 
+1021 *(EQUAL_START_END)*
+    noUiSlider most have a nonequal start and end.
+
 9999 *(EXT)*
     Indicates that a custom error check has failed.
 
@@ -97,6 +100,7 @@ codes = {
     1018: ("NO_RANGE_TOOL_RANGES",                               "RangeTool must have at least one of x_range or y_range configured"),
     1019: ("DUPLICATE_FACTORS",                                  "FactorRange must specicy a unique list of categorical factors for an axis"),
     1020: ("BAD_EXTRA_RANGE_NAME",                               "An extra range name is configued with a name that does not correspond to any range"),
+    1021: ("EQUAL_START_END",                                    "Slider 'start' and 'end' cannot be equal"),
     9999: ("EXT",                                                "Custom extension reports error"),
 }
 
--- a/bokeh/core/validation/errors.py
+++ b/bokeh/core/validation/errors.py
@@ -71,7 +71,7 @@ validation checks.
 1020 *(BAD_EXTRA_RANGE_NAME)*
     An extra range name is configued with a name that does not correspond to any range.
 
-1021 *(EQUAL_START_END)*
+1021 *(EQUAL_SLIDER_START_END)*
     noUiSlider most have a nonequal start and end.
 
 9999 *(EXT)*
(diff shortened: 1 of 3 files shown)


Karel-van-de-Plassche:
What do you mean with validation check, do you have an example somewhere in bokeh? I was not exactly sure how to hook into the bokeh way of defining attributes to the javascript <> python interface.

Karel-van-de-Plassche:
Also, is it possible to raise an Exception from TypeScript? Wasn't sure.

bryevdv:
Here is an example:

https://github.com/bokeh/bokeh/blob/9bf4ff0b47734b2d3d184c9c7690e8b013893a36/bokeh/models/tickers.py#L171-L174

You can decorate a method with `error` or `warning` decorator (and the validation code it checks) and then at serialization time the check is performed. This is the only way to catch some problems that are a result of e.g. how the entire collection of Bokeh models is arranged or configured. In this case, the current check will not catch:

```
s = RangeSlider(start=1, end=2)
s.start = 2
```

But a validation check will. I would suggest having both. Might as well fail early when we can. 

The method should return `None` if there is no problem, or a descriptive string if there is a problem. We don't currently raise hard exceptions on ERROR conditions but we might in the future. Presently we just try as hard as possible to make the messages loud an unavoidable.

Karel-van-de-Plassche:
> Certainly, or we can log a console error... most people just don't know how to access this information in their browser though, so it's of limited use. Best to try and catch known error conditions on the python side when at all possible.

A console error is already logged by noUiSlider itself, so don't think anything is needed there. 

> Also, is it possible to raise an Exception from TypeScript? Wasn't sure.

AH, sorry, I meant is it possible to raise a bokehjs error to Python? I guess this wouldn't work anyhow if no server is running, so the validation check you suggested is better

Karel-van-de-Plassche:
I have added the validation checks. Unfortunately, I did not manage to write tests for this. I tried copying `bokeh/application/tests/test_application.py`, but no luck. Example:

``` python
import bokeh.models.widgets.sliders as mws 

from datetime import datetime
import logging
import pytest

from bokeh.util.serialization import convert_datetime_type
from bokeh.util.logconfig import basicConfig

# needed for caplog tests to function
basicConfig()

#### SNIP ####

def test_rangeslider_equal_start_end_validation(caplog):
    start = 0 
    end = 10
    s = mws.RangeSlider(start=start, end=end)
    #with caplog.at_level(logging.ERROR, logger='bokeh.core.validation.check'):
    with caplog.at_level(logging.ERROR):
        assert len(caplog.records) == 0
        s.end = 0 
        assert len(caplog.records) == 1
```

Karel-van-de-Plassche:
And shameless self-plug: ionRangeSlider does support equal start and end https://github.com/Karel-van-de-Plassche/bokeh-ion-rangeslider!

bryevdv:
I would switch to ion in a heartbeat if they would add a vertical orientation option.

bryevdv:
@Karel-van-de-Plassche validation does not happen  every time a property is set, that would be too expensive. It only happens once, right when the document is serialized. For tests you will need to call `validate` manually.

bryevdv:
I would also suggest making the name more specific, e.g. `EQUAL_SLIDER_START_END`

bryevdv:

Issue #7499, opened 2018-02-06
Allow defining custom tooltip for certain tools


philippjfr:
When using some of the more specific tools, e.g. the drawing tools, you might want to add multiple copies of the same tool and assign different semantics to each tool instance. As a simple example I wrote a demo that allows drawing two types of polygons to annotate the foreground and background of an image. In this case it is difficult to distinguish the two tool instances. It would therefore be nice if it were possible to assign the tools custom tooltips to distinguish between them.

Currently the tooltip simply reflects the ``tool_name`` on the model but it would be great if that could be overridden by another property. I think this only makes sense for certain types of tools, most likely the draw tools and maybe also the select tools.

jbednar:
It would also be great if the icon could be overridden, for the same reason.  It might be difficult to support (and to provide) arbitrary new icons for each new usage, but at least some of the drawing tools can be used for quite different things and we could provide a selection of such icons to be chosen when instantiating the tool...

mattpap:
> Currently the tooltip simply reflects the `tool_name`

No, tooltip is generated by a getter, which you can override. This is done in e.g. pan tool, which comes in three variants, pan, x-pan and y-pan, and all have different tooltips. 

> It would also be great if the icon could be overridden, for the same reason.

It can be. `icon` is defined as class attribute, but one can override it with a getter. Once again pan tool is a great example, as it has different icons for its each variant.

philippjfr:
Thanks @mattpap, great to know this is already overridable at the JS level.

jbednar:
But it sounds like there has to be a subclass for each variant?  Here it's about a Python user wanting to instantiate the same tool multiple times with different info each time.

philippjfr:
No I was wrong about that before, at the JS level you can define a getter, for the pan tool it looks like this:

```javascript

  get tooltip(): string {
    return this._get_dim_tooltip("Pan", this.dimensions)
  }

  get icon(): string {
    switch (this.dimensions) {
      case "both":   return "bk-tool-icon-pan"
      case "width":  return "bk-tool-icon-xpan"
      case "height": return "bk-tool-icon-ypan"
    }
  }
```

jbednar:
Ok, I guess then we can put in a few canned variants to choose from...

philippjfr:
[...]

PR #8099, opened 2018-07-21
Add CustomAction toolbar action

bryevdv:
This is a slight diversion, but I intend to use this targetable toolbar button that executes callback code as a simple and standard means to inject JS code into selenium tests. Additionally, it seems like quite a useful feature for users in its own right. 

This PR also adds an `Image` property, that can be set from RGB(A) NumPy arrays, PIL image objects, or from filenames of images in disk. The transformed value is always a Base64 `data\image` string. In this way the icon for the toolbar tool may be set to an arbitrary choice by the user. 

This requires adding Pillow as a dependency, however Pillow is easily obtainable. Additionally, there was Gitter discussion about adding PNG, TIFF, etc. image glyphs, and Pillow will be needed for that work as well.

### DIFF
--- a/bokehjs/src/lib/models/tools/actions/custom_action.ts
+++ b/bokehjs/src/lib/models/tools/actions/custom_action.ts
@@ -1,0 +1,66 @@
+import {ActionTool, ActionToolView, ActionToolButtonView} from "./action_tool"
+import * as p from "core/properties"
+import {isFunction} from "core/util/types"
+
+export class CustomActionButtonView extends ActionToolButtonView {
+  model: CustomAction
+
+  css_classes(): string[] {
+    return super.css_classes().concat("bk-toolbar-button-custom-action")
+  }
+
+}
+
+export class CustomActionView extends ActionToolView {
+  model: CustomAction
+
+  doit(): void {
+    const callback = this.model.callback
+    if (isFunction(callback))
+      callback(this, {})
+    else
+      callback.execute(this, {})
+    }
+}
+
+export namespace CustomAction {
+  export interface Attrs extends ActionTool.Attrs {
+    action_tooltip: string
+    callback: any // XXX
+    icon: string
+  }
+
+  export interface Props extends ActionTool.Props {}
+}
+
+export interface CustomAction extends CustomAction.Attrs {}
+
+export class CustomAction extends ActionTool {
+
+  properties: CustomAction.Props
+
+  constructor(attrs?: Partial<CustomAction.Attrs>) {
+    super(attrs)
+  }
+
+  static initClass(): void {
+    this.prototype.type = "CustomAction"
+    this.prototype.default_view = CustomActionView
+
+    this.define({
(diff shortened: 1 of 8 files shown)


bokehjs/src/lib/models/tools/button_tool.ts
```diff
@@ -20,7 +21,12 @@ export abstract class ButtonToolButtonView extends DOMView {
 
   render(): void {
     empty(this.el)
-    this.el.classList.add(this.model.icon)
+    if (isString(this.model.icon)) {
+      if (this.model.icon.startsWith("bk-tool-icon-"))
+        this.el.classList.add(this.model.icon)
+      else
+        this.el.style.backgroundImage = "url('data:image/png;base64," + this.model.icon + "')"
```
bryevdv: Just occurs to me this is problematic, as ostensibly `Image` will accept other valid PIL image formats. But I'm not sure data URI's can actually be more than PNGs. Might end up simply restricting what `Image` accepts.

bokehjs/src/lib/models/tools/actions/custom_action.ts
```diff
@@ -0,0 +1,66 @@
+import {ActionTool, ActionToolView, ActionToolButtonView} from "./action_tool"
+import * as p from "core/properties"
+import {isFunction} from "core/util/types"
+
+export class CustomActionButtonView extends ActionToolButtonView {
+  model: CustomAction
+
+  css_classes(): string[] {
+    return super.css_classes().concat("bk-toolbar-button-custom-action")
```
bryevdv: This is to provide a known target for Selenium test to be able to click this button.

bryevdv:
Well, this is causing real test failures for `bokeh png` and `bokeh svg` tho for the life of me I have no idea why.

bryevdv:
Apparently PhantomJS does not support the `startsWith` method, that was a fun few hours...

mattpap:
> Apparently PhantomJS does not support the startsWith method, that was a fun few hours...

Yep, that's why we have `startsWith` in `core/util/string`.

bryevdv:
@mattpap thanks, I'd forgotten. Any other comments before merging?

bokehjs/src/lib/models/tools/button_tool.ts
```diff
@@ -20,7 +22,13 @@ export abstract class ButtonToolButtonView extends DOMView {
 
   render(): void {
     empty(this.el)
-    this.el.classList.add(this.model.icon)
+    const icon = this.model.icon
+    if (isString(icon)) {
+      if (startsWith(icon, "bk-tool-icon"))
+        this.el.classList.add(icon)
+      else
+        this.el.style.backgroundImage = "url('data:image/png;base64," + icon + "')"
```
mattpap: I would change (reverse) this logic, especially in light of #7499. Anything that starts with `data:` should be treated as a CSS icon and everything else should be a CSS class. This also means that `icon` should contain `data:image/png;base64,` prefix. This way we would allow for other formats and encodings.

PR #8101, opened 2018-07-22
REF: Use generators inplace of lists #8092

### LINKED ISSUE #8092: Use generators instead of lists
For built-in Python functions use generators instead of lists.

For example in  `./bokeh/core/enums.py:121`

```
attrs = dict([ (value, value) for value in values ])
```

can be

```
attrs = dict((value, value) for value in values)
```

Similarly, there are other instances of these `list` creations -- should we clean these up to use generators instead?

If the dev's think so too, I'll submit a PR.

pratapvardhan:
Use generators inplace of lists

- [ ] issues: fixes #8092

### DIFF
--- a/_setup_support.py
+++ b/_setup_support.py
@@ -292,9 +292,9 @@ def build_js():
     if result != 0:
         indented_msg = ""
         outmsg = proc.stdout.read().decode('ascii', errors='ignore')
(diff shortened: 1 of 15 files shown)


pratapvardhan:
Error on one of this jobs seems unrelated to this one.

bryevdv:
Thanks @pratapvardhan

PR #8102, opened 2018-07-22
WIP: Point hit-test for step

### LINKED ISSUE #7419: Hover tooltip not showing for step line
Python 3.6
Bokeh 0.12.13

Using the default hover tool I do not see the tooltip when using the step line. Switching to circle or line will result to the correct behaviour.

```

# output to static HTML file
output_file("test.html")

p = figure(plot_width=400,
           plot_height=400,
           tools='hover')

# p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5],
#          size=20,
#          color="navy",
#          alpha=0.5,
#          )

p.step([1, 2, 3, 4, 5], [6, 7, 2, 4, 5],
         color="navy",
         alpha=0.5,
         )

# p.line([1, 2, 3, 4, 5], [6, 7, 2, 4, 5],
#        color="navy",
#        alpha=0.5,
#        )

# show the results
show(p)
```

Karel-van-de-Plassche:
I was working on hit testing for `step`, but something weird is going on. Even though the resulting `Selection` has the right `line_indices`, the wrong glyph is still selected. See this GIF. Even though `result.line_indices` is 2, glyph 3 is selected. @mattpap you have any idea what is going on?

ezgif com-video-to-gif

``` python

# output to static HTML file
output_file("test.html")

p = figure(plot_width=400,
           plot_height=400,
           tools='hover')

p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], 
          size=20,
          color="navy",
          alpha=0.5,
          )   

p.step([1, 2, 3, 4, 5,6,7,8], [6, 7, 2, 4, 5, 7, 2, 3], 
         color="navy",
         alpha=0.5,
         line_width=7,
         mode='before'
         )   

#p.line([1, 2, 3, 4, 5],
#        [6, 7, 2, 4, 5],
#        color="navy",
#        alpha=0.5,
#        )

# show the results
show(p)
```

- [ ] issues: fixes #7419 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/glyphs/step.ts
+++ b/bokehjs/src/lib/models/glyphs/step.ts
@@ -4,8 +4,12 @@ import {LineMixinVector} from "core/property_mixins"
 import {Line} from "core/visuals"
 import {StepMode} from "core/enums"
 import * as p from "core/properties"
+import * as hittest from "core/hittest"
+import {PointGeometry} from "core/geometry"
+import {Arrayable} from "core/types"
 import {IBBox} from "core/util/bbox"
 import {Context2d} from "core/util/canvas"
+import {Selection} from "../selections/selection"
 
 export interface StepData extends XYGlyphData {}
 
@@ -14,6 +18,32 @@ export interface StepView extends StepData {}
 export class StepView extends XYGlyphView {
   model: Step
   visuals: Step.Visuals
+
+  protected _get_points(i: number, sx: Arrayable<number>, sy: Arrayable<number>): number[][] {
+    let x1: number, x2: number
+    let y1: number, y2: number
+    switch (this.model.mode) {
+      case "before": {
+        ;[x1, y1] = [sx[i-1], sy[i]]
+        ;[x2, y2] = [sx[i],   sy[i]]
+        break
+      }
+      case "after": {
+        ;[x1, y1] = [sx[i], sy[i-1]]
+        ;[x2, y2] = [sx[i], sy[i]  ]
+        break
+      }
+      case "center": {
+        const xm = (sx[i-1] + sx[i])/2
+        ;[x1, y1] = [xm, sy[i-1]]
+        ;[x2, y2] = [xm, sy[i]  ]
+        break
+      }
+      default:
+        throw new Error("unexpected")
+    }
+    return [[x1, y1], [x2, y2]]
+  }
 
   protected _render(ctx: Context2d, indices: number[], {sx, sy}: StepData): void {
     let drawing = false
@@ -28,29 +58,11 @@ export class StepView extends XYGlyphView {
     ctx.beginPath()
     ctx.moveTo(sx[0], sy[0])
 
+    let x1: number, x2: number
+    let y1: number, y2: number
     for (const i of indices) {
-      let x1: number, x2: number
-      let y1: number, y2: number
-      switch (this.model.mode) {
-        case "before": {
-          ;[x1, y1] = [sx[i-1], sy[i]]
-          ;[x2, y2] = [sx[i],   sy[i]]
-          break
-        }
-        case "after": {
-          ;[x1, y1] = [sx[i], sy[i-1]]
-          ;[x2, y2] = [sx[i], sy[i]  ]
-          break
-        }
-        case "center": {
-          const xm = (sx[i-1] + sx[i])/2
-          ;[x1, y1] = [xm, sy[i-1]]
-          ;[x2, y2] = [xm, sy[i]  ]
-          break
-        }
-        default:
-          throw new Error("unexpected")
-      }
+      [[x1, y1], [x2, y2]] = this._get_points(i, sx, sy)
+
 
       if (drawing) {
         if (!isFinite(sx[i] + sy[i])) {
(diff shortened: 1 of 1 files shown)


bryevdv:
The default `line_policy` for hover tools is "nearest":

https://github.com/bokeh/bokeh/blob/master/bokeh/models/tools.py#L1122-L1123

But the code that computes "nearest" was only ever written with linear segments in mind. My guess is that it falls over for step, and that making it work reasonably with step is going to be non-trivial. We will probably have to move the "nearest" computation on to the glyph itself, and let the hover tool delegate computing the nearest index. Same story with "interp" that definitely assumes linear segments and will not function well, I am sure. 

Just for the info, you might try setting the other line policies to see if they behave better.

bryevdv:
To be clear, making this work will require new computations for nearest and interp, and I am not sure what exactly should happen in all cases. E.g. when you have "vline" selection mode, and you hover right over a vertical segment in a step, what should get displayed? the "prev" index? the "next" index? But then where? In the middle of the the vertical segment? Fair warning, I expect this is gong to be one of those tasks that is a little hard, not because of some inherent difficulty, but rather because there are going to be lots of annoying fiddly cases to keep track of

Karel-van-de-Plassche:
@bryevdv  Thanks for your explanation. Did I misinterpret what `_hit_point` returns? I thought `Selection.line_indices` would be the index(es) of the glyphs that were hit, and as such the points where the hover tooltip should appear.

bokehjs/src/lib/models/glyphs/step.ts
```diff
@@ -15,6 +19,32 @@ export class StepView extends XYGlyphView {
   model: Step
   visuals: Step.Visuals
 
+  protected _get_points(i: number, sx: Arrayable<number>, sy: Arrayable<number>): number[][] {
```
mattpap: `[[number, number], [number, number]]` should be the return type.

bryevdv:
>  I thought Selection.line_indices would be the index(es) of the glyphs that were hit

Yes, but that's only part of the story. If I hover over the middle of a line segment, what should get displayed? The nearest point to the cursor? The next point? The previous point? Should the x,y positions be interpolated? All of these are valid options, depending on circumstance.  

You can see how the line indices interact with all these options here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/tools/inspectors/hover_tool.ts#L210-L257

The good news is it looks like interpolation hits are already delegated, so that's a bit less work up front. The Step glyph will need a version of that method appropriate for step glyphs. However, there we call `_nearest_line_hit` unconditionally, and that won't work. We will need to delegate to a method on the glyph so that Line and Step can do their separate thing.

bokehjs/src/lib/models/glyphs/step.ts
```diff
@@ -83,6 +95,45 @@ export class StepView extends XYGlyphView {
...
+    */
+    const result = hittest.create_empty_hit_test_result()
+    const point = {x: geometry.sx, y: geometry.sy}
+    const threshold = Math.max(2, this.visuals.line.line_width.value() / 2)
+
+    let x1: number, x2: number
+    let y1: number, y2: number
+    let p0 = {x: NaN, y: NaN}
+    for (let i = 0, end = this.sx.length-1; i < end; i++) {
+      [[x1, y1], [x2, y2]] = this._get_points(i, this.sx, this.sy)
```
mattpap: Given you don't reassign those, you could just put `const` in front of this line and remove declarations above.

bokehjs/src/lib/models/glyphs/step.ts
```diff
@@ -83,6 +95,45 @@ export class StepView extends XYGlyphView {
...
+    that describes the hit result:
+      Args:
+        * geometry (object): object with the following keys
+          * sx (float): screen x coordinate of the point
+          * sy (float): screen y coordinate of the point
+          * type (str): type of geometry (in this case it's a point)
+      Output:
+        Object with the following keys:
+          * 0d (bool): whether the point hits the glyph or not
+          * 1d (array(int)): array with the indices hit by the point
```
mattpap: This is redundant. Types should explain arguments and products of this function. If there is anything here that's not covered by types, then it can stay (mostly description).

bokehjs/src/lib/models/glyphs/step.ts
```diff
@@ -28,29 +58,11 @@ export class StepView extends XYGlyphView {
...
-        case "center": {
-          const xm = (sx[i-1] + sx[i])/2
-          ;[x1, y1] = [xm, sy[i-1]]
-          ;[x2, y2] = [xm, sy[i]  ]
-          break
-        }
-        default:
-          throw new Error("unexpected")
-      }
+      [[x1, y1], [x2, y2]] = this._get_points(i, sx, sy)
```
mattpap: Given lack of reassignments, `const`.

Karel-van-de-Plassche:
Thanks for all the explaining and hints. Seems I might've bitten off more than I can chew, but I'll have a go at it later this week.

bryevdv:
@Karel-van-de-Plassche just checking in, do you expect to return to this any time soon? If not we can close it for now.

Karel-van-de-Plassche:
Bit swamped, sorry. Let's close it for now

PR #8105, opened 2018-07-24
REF: Use .loc instead of .ix in box_annotation example. Fixes #8104

### LINKED ISSUE #8104: box_annotation example .ix method is deprecated
In https://bokeh.pydata.org/en/latest/docs/gallery/box_annotation.html

`.ix` is deprecated, use `.loc`

pratapvardhan:
REF: Use .loc instead of .ix in box_annotation example.

- [ ] issues: fixes #8104


bryevdv:
Thanks for the catch @pratapvardhan

PR #8107, opened 2018-07-25
REF: Use .loc instead of .ix in examples

pratapvardhan:
In continuation to #8104 and #8105

### DIFF
--- a/examples/howto/layouts/words_and_plots.py
+++ b/examples/howto/layouts/words_and_plots.py
@@ -47,8 +47,8 @@ def scatter():
 
 
 def hover_plot():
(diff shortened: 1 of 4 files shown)


bryevdv:
Thanks @pratapvardhan !

bryevdv:
@pratapvardhan BTW I definitely appreciate the cleanup PRs. If you do intend to keep contributing on a (semi) regular basis please let me know and we can talk about adding you to @bokeh/dev at some point.

pratapvardhan:
@bryevdv -- Sure, would you want to talk over gitter, I could join there, or else?

PR #8110, opened 2018-07-26
Restore selenium tests

### LINKED ISSUE #6495: Task: pytest marked "selenium" tests should assert no browser console errors.
Task: pytest marked "selenium" tests should assert no browser console errors.

bryevdv:
- [x] issues: fixes #6495 
- [x] tests added / passed


bryevdv:
cc @philippjfr @jlstevens it will take a little doing but it should be possible to make real integration tests for the edit tools with this framework, if you'd like to think about tests for those tools it would be helpful

bryevdv:
@mattpap I discovered this during writing selenium tests. With this code:
```
    source = ColumnDataSource(dict(x=[1, 2], y=[1, 1]))
    plot = Plot(plot_height=400, plot_width=430,x_range=Range1d(0, 1), y_range=Range1d(0, 1), min_border=0)
    plot.add_glyph(source, Rect(x='x', y='y', width=0.9, height=0.9))
    plot.add_tools(tool)
    code = RECORD("xrstart", "p.x_range.start") + RECORD("xrend", "p.x_range.end") + RECORD("yrstart", "p.y_range.start") + RECORD("yrend", "p.y_range.end")
    plot.add_tools(CustomAction(callback=CustomJS(args=dict(p=plot), code=code)))
    plot.toolbar_sticky = False
    return plot
```
The toolbar takes up part of the 400 plot width:

screen shot 2018-07-26 at 20 28 57

Is this expected?

bryevdv:
Just to note, this issue with Selenium appears to be causing intermittent failures: https://github.com/SeleniumHQ/selenium/issues/5296

mattpap:
@bryevdv, it's expected. What you see is a toolbar panel. Also, `toolbar_sticky` was removed months ago.

bryevdv:
> it's expected. What you see is a toolbar panel. 

That seems unfortunate, that means there is no way for me to specify a canvas size explicitly in the presence of a toolbar without fudge like `Plot(min_border_right=50, plot_width=size_i_want+50, ...)

> Also, toolbar_sticky was removed months ago.

@mattpap That does not seem to be entirely true 

https://github.com/bokeh/bokeh/search?q=toolbar_sticky&unscoped_q=toolbar_sticky

bryevdv:
> I don't know what you want to achieve. 

I'd like for
```
Plot(min_border=0, plot_width=400, plot_height=400) 
```
to yield a square plot area. I would expect the toolbar to abut the central plot area, which means being on top of the canvas if there is border space, or being outside the canvas if there is not border space. 

 > I can only guess that either you have to hide the toolbar

Can't hide to toolbar, extensively using selenium to activate `CustomAction` from the toolbar

> create a layout with an external toolbar.

Is there an example of this? I did not think it was feasible yet. 

> Right, there is still some backwards compatibility code.

I didn't see deprecations, does it need to be deleted?

mattpap:
>  I would expect the toolbar to abut the central plot area (...)

It seems to me that you want to have the frame of certain dimensions, not the canvas. Unfortunately existing layout doesn't give explicit control over frame dimensions. Toolbar (panels) are not rendered over the central area, but over side panels. So, until the new layout is in place, the best option is to detach toolbars from plots, then you can make canvas == frame to control frame's dimensions.

> Can't hide to toolbar, extensively using selenium to activate CustomAction from the toolbar

You could enable a tool with the API, though using click events is more idiomatic with selenium.

> Is there an example of this? I did not think it was feasible yet.

Every grid plot uses this. There is `examples/models/files/toolbars3` that shows this feature outside grid plot scope.

> I didn't see deprecations, does it need to be deleted?

It's not technically deprecated, but it should go when layout will allow to express the underlying features without hacks.

bryevdv:
> It seems to me that you want to have the frame of certain dimensions, not the canvas. Unfortunately existing layout doesn't give explicit control over frame dimensions.

Yes and no, I do want a square frame, but absent and independent of anything else, I think 
```
Plot(min_border=0, plot_width=400, plot_height=400) 
```
*should result* in a square plot frame. The current behaviour makes `min_border` inconsistent. It is supposed to be the min border between the central frame and the canvas perimeter, but if there is a toolbar, that is no longer the case. I am saying it should behave like this:

screen shot 2018-07-27 at 16 12 13

bryevdv:
Maybe that's a better way of putting it: 

* I don't think the presence or absence of a toolbar should have any input into the frame dimensions.
* toolbars should always abut the frame, wherever it happens to be

mattpap:
I see that you are viewing toolbars as something special in a plot. They aren't, they are just panels. You can replace a toolbar with a title (or a label, legend, etc.) and you will get exactly the same behavior. Toolbars used to be special, though in a different way from what you describe, but that din't exactly work well. Anyway, lets move this to a more appropriate place for discussion.

bokehjs/src/lib/testing.ts
```diff
@@ -0,0 +1,35 @@
...
+// results structure. This function deletes/creates a div that can act as
+// a semaphore. Tests should wait for the previous div to be stale, then
+// find the new div. At that point the results should be available
+function _update_test_div(): void {
+  const body = document.getElementsByTagName("body")[0]
+  const col = document.getElementsByClassName("bokeh-test-div");
+
+  if (col.length == 1)
+    body.removeChild(col[0])
+    delete col[0]
```
mattpap: `{ }`

bokehjs/src/lib/testing.ts
```diff
@@ -0,0 +1,35 @@
+// Just a dumb key/value record for collecting arbitrary info for tests
```
mattpap: This this file has to be under `lib/`, when it's not part of the library?

bryevdv: It is part of the library, the test suite calls this little function to be able to perform tests without a ton of boilerplate

mattpap: Then it's a part of testing code and shouldn't affect the library. Do you want every user pay a price for testing code? I presume it's a matter of convenience to put this code here, but it should be placed elsewhere (`bokeh._testing.util.selenium`).

bryevdv: > Do you want every user pay a price for testing code? 

For 30 lines? Yes, I am absolutely, 100% ok with that.

philippjfr:
>cc @philippjfr @jlstevens it will take a little doing but it should be possible to make real integration tests for the edit tools with this framework, if you'd like to think about tests for those tools it would be helpful

I'm now back from holiday and will be revisiting the edit tools asap. There are various issues I have to investigate and putting them under test will help a lot. One question I have is whether we can have fully integrated tests, i.e. launching a bokeh server, using selenium to interact with the plot and then checking those interactions have been processed appropriately server-side.

bryevdv:
OK there is still plenty more to do, including some of the original selenium tests to restore or decide to discard, but this is enough for now. Intend to merge when green

PR #8114, opened 2018-07-27
Add Dash marker glyph

### LINKED ISSUE #8109: Feature Request: VerticalLine marker (equivalent to marker='|' in matplotlib)
There is an extensive list of markers available in bokeh: https://bokeh.pydata.org/en/latest/docs/reference/models/markers.html

What I am missing is an equivalent for the `marker='|'` in matplotlib: https://matplotlib.org/examples/lines_bars_and_markers/marker_reference.html

My usecase is to denote different types of events on a timeline, so that it looks like this:
image

A circle or asterisk is not fine grained enough for this as the horizontal space taken by the marker is too broad so that the actual center of different markers is hard to compare visually.

gregmuellegger:
A PR for #8109. This is currently still work in progress.

issues: fixes #8109 

- [ ] ~~~Adjust shader code for dash marker~~~
- [x] tests added / passed

### DIFF
--- a/examples/reference/models/Dash.py
+++ b/examples/reference/models/Dash.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 from bokeh.models import ColumnDataSource, DataRange1d, Plot, LinearAxis, Grid
-from bokeh.models.markers import VerticalLine
+from bokeh.models.markers import Dash
 from bokeh.io import curdoc, show
 
 N = 9
@@ -18,7 +18,7 @@ plot = Plot(
     title=None, x_range=xdr, y_range=ydr, plot_width=300, plot_height=300,
     h_symmetry=False, v_symmetry=False, min_border=0, toolbar_location=None)
 
-glyph = VerticalLine(x="x", y="y", size="sizes", line_color="#99d594", line_width=2, fill_color=None)
+glyph = Dash(x="x", y="y", size="sizes", line_color="#3288bd", line_width=1, fill_color=None)
 plot.add_glyph(source, glyph)
 
 xaxis = LinearAxis()
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -1097,8 +1097,9 @@ class Wedge(XYGlyph):
 
 # XXX: allow `from bokeh.models.glyphs import *`
 from .markers import (Asterisk, Circle, CircleCross, CircleX, Cross, Diamond, DiamondCross,
-                      Hex, InvertedTriangle, Marker, Square, SquareCross, SquareX, Triangle, X)
+                      Hex, InvertedTriangle, Marker, Square, SquareCross, SquareX, Triangle,
+                      VerticalLine, X)
 
 # Fool pyflakes
 (Asterisk, Circle, CircleCross, CircleX, Cross, Diamond, DiamondCross,
-Hex, InvertedTriangle, Marker, Square, SquareCross, SquareX, Triangle, X)
+Hex, InvertedTriangle, Marker, Square, SquareCross, SquareX, Triangle, VerticalLine, X)
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -1096,10 +1096,10 @@ class Wedge(XYGlyph):
     """)
 
 # XXX: allow `from bokeh.models.glyphs import *`
-from .markers import (Asterisk, Circle, CircleCross, CircleX, Cross, Diamond, DiamondCross,
-                      Hex, InvertedTriangle, Marker, Square, SquareCross, SquareX, Triangle,
-                      VerticalLine, X)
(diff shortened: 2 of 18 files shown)


bryevdv:
@gregmuellegger This is a great throrough first PR! I only have a couple of comments:

* What do you think about `Pipe` (analogous to "unix pipe", i.e. `|`) as a shorter name? 

* I'm not sure what the GL fragment needs to be, @mattpap ?

* The python unit tests are failing with just a missing import:
    ```
        def test_VerticalLine():
    >       marker = VerticalLine()
    E       NameError: name 'VerticalLine' is not defined
    bokeh/models/tests/test_glyphs.py:562: NameError  
    ```

* Similar for codebase checks, just a typo:
    > [14:17:44] src/lib/api/typings/plotting.d.ts(98,31): error TS2304: Cannot find name 'VerticalLignAttrs'.

* The docs build only needs the section header underline length to match (sphinx is very picky):
  
  
  >/home/travis/build/bokeh/bokeh/sphinx/source/docs/reference/models/markers/vertical_line.rst:4:Title underline too short.

gregmuellegger:
Thanks for the first review! 

> What do you think about Pipe (analogous to "unix pipe", i.e. |) as a shorter name?

I dislike the name Pipe as it conveys semantics, like that it “pipes“ data (read as verb). I would prefer the more descriptive VerticalLine or maybe more concise: VLine.

That also lends itself more towards a corresponding HLine, which is harder to name should the vertical one be called Pipe.

bryevdv:
Well apart from the name just being long, I also am not fond of "line", as every other use in Bokeh connotes a multi-line, i.e. something with extension and connected topology, and this is definitely not like that kind of line. I'll keep thinking,

bryevdv:
Well here is an idea: all markers can be rotated it's easy to make a "vline" from "hline", or vice versa. So instead of making both, which seems rather duplicative, how about making at horizontal by default and calling it `Dash`

gregmuellegger:
> how about making this horizontal by default and calling it Dash

I like that! I changed the name in the code accordingly and fixed the errors that you pointed out.

The shader code is currently missing. Do you know where I could learn to make the code do what it is supposed to do?

Karel-van-de-Plassche:
Looks like something is going wrong in the documentation string of `Dash`:

```
.. |dash|       `      replace:: :func:`~bokeh.plotting.figure.Figure.dash`
Warning, treated as error:
/home/travis/build/bokeh/bokeh/sphinx/source/docs/user_guide/plotting.rst:481:Substitution definition "dash" empty or invalid.
.. |dash|       `      replace:: :func:`~bokeh.plotting.figure.Figure.dash`
```
and
```
E         File ends with more than 1 empty line: sphinx/source/docs/reference/models/markers/dash.rst, line 9
E       assert 1 == 0
E        +  where 1 = len(['File ends with more than 1 empty line: sphinx/source/docs/reference/models/markers/dash.rst, line 9'])
```

bryevdv:
> Do you know where I could learn to make the code do what it is supposed to do?

It's a GL shader, in case that is a helpful lead. I don't know much more beyond that offhand. It's possible that @mattpap can offer an implementation, or else we an simply proceed and document that this marker is not supported by webgl mode.

gregmuellegger:
I fixed the failing build. I'm happy to wait a bit if @mattpap gets around to respond regarding the GL shader.

If there is no response in about a week, I'll be adjusting the documentation accordingly.

examples/reference/models/Dash.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
```
pratapvardhan: Just confirming, do we intend to have python files with upper-case characters -- **D**ask.py?

gregmuellegger: It's consistent with the rest of the directory. Usually I would advise not to use uppercase module names. But these files are standalone and not part of  package.

The names describe class names, so I prefer the filenames to match the exact classname here.

bryevdv: Yes in this specific instance these are essentially test files an the file names refer to corresponding class names

bryevdv:
@gregmuellegger just checking in, do you expect to be able to make the final changes?

gregmuellegger:
Yes, probably this week. What I don't know how to do is writing the release
notes. Will you add them? Or is there a guide?

Bryan Van de Ven <notifications@github.com> schrieb am Di., 7. Aug. 2018,
16:47:

bryevdv:
I'll have to edit the release notes extensively later regardless, so I will take care of mentioning this work at that time. For now I have removed the dummy fragment shader and edited the webgl users guide to reflect that not every marker supports webgl, so that this can be merged today. Thanks @gregmuellegger

gregmuellegger:
Thanks @bryevdv for guidance, cleaning it up and merging!

Sorry that I didn't find the time to polish it up.

PR #8125, opened 2018-07-31
Fixed Exception raising tests of test_sources

Karel-van-de-Plassche:
As soon as an exception is caught, the enclosing with context is broken. So, the assertions are never reached. You can test this by putting assert 0 somewhere

- [ ] tests added / passed


bryevdv:
Thanks for the catch @Karel-van-de-Plassche !

Issue #6822, opened 2017-08-27
AutocompleteInput does not complete word when option is clicked


tkchris93:
I am using MacOS, Bokeh 0.12.7rc3, python 2.7.13, and Google Chrome to view a static html file generated by Bokeh. Clicking on a word in the autocomplete dropdown only sometimes (~25%) completes the word in the input box. I have not been able to see any pattern as to when it works and when it doesn't. There is no additional information to share from the javascript console.

```
from bokeh.plotting import show, output_file
from bokeh.models.widgets import AutocompleteInput

output_file('autocomplete_input.html')

input_widget = AutocompleteInput(completions=['pok', 'wer', 'foo', 'bar', 'baz'], title='test')
show(input_widget)
```

mattpap:
This is a bizarre bug.

guojunchi11:
This happens to me, too.  
Also, I can not use the down arrow on the keyboard to select, have to use the mouse

RafaelPais:
Same behavior in bokeh 0.12.13, any update about this?

bryevdv:
@RafaelPais  sorry, unfortunately not. Generally speaking, there is currently a large excess of issues relative to the number of core contributors, meaning some things will simply not be able to get done for a long time, unless new contributors get involved. If you have the interest and ability, I'm happy to point you in the right direction to get started.

RafaelPais:
@bryevdv Many thanks for the prompt reply. Yes, I'm interested in contributing. I'll take note of the documentation available in the Getting Set Up section on DEVELOPER GUIDE. However, I will be very pleased if you can direct me to code related to autocomplete. I have a special interest in solving this problem.

bryevdv:
@RafaelPais Absolutely, the python model is defined here:

https://github.com/bokeh/bokeh/blob/master/bokeh/models/widgets/inputs.py#L65

I doubt there is much to see or do there, but just for reference. All the work is done in the JS implementation, which is here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/widgets/autocomplete_input.ts

DanielaBreitman:
This issue still occur seems to occur in bokeh==0.13.0. Is there any workaround?

magee256:
I'm still seeing this in bokeh==1.0.1, [this pull request](https://github.com/bokeh/bokeh/pull/8126) references a layout rework that should fix the issue. Is that rework still ongoing?

linusmarco:
[...]

PR #8126, opened 2018-07-31
fix for https://github.com/bokeh/bokeh/issues/6822

AaronWatters:

- [ ] issues: https://github.com/bokeh/bokeh/issues/6822
- [ ] tests passed

### DIFF
--- a/bokehjs/src/lib/models/widgets/autocomplete_input.ts
+++ b/bokehjs/src/lib/models/widgets/autocomplete_input.ts
@@ -51,6 +51,7 @@ export class AutocompleteInputView extends TextInputView {
       const el = event.target as HTMLElement
       const text = el.dataset.text!
       this.model.value = text
+      this.render()
       //this.inputEl.value = text
     }
   }
--- a/bokehjs/src/lib/models/widgets/text_input.ts
+++ b/bokehjs/src/lib/models/widgets/text_input.ts
@@ -16,7 +16,6 @@ export class TextInputView extends InputWidgetView {
 
   connect_signals(): void {
     super.connect_signals()
-    // this.connect(this.model.change, () => this.render())
   }
 
   css_classes(): string[] {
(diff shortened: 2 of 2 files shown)


bryevdv:
@mattpap do you have any thoughts? Offhand it seems like this change would break standard text input property events. If this change fixes autocomplete then perhaps auto complete view should not inherit from text input view. 

In any case @AaronWatters FYI any code that is no longer needed or appropriate should be deleted, not commented out.

AaronWatters:
Closing with intent to reopen because the travis tests timed out...

AaronWatters:
reopening to restart travis build

bryevdv:
ping @mattpap

AaronWatters:
If I understand it correctly the problem is that the render() operation sometimes happens before the value is updated due to a race condition in the event handlers.  The render() has the effect of cancelling the other event since it destroys the object associated with the event (I think). Any fix that guarantees the render() happens after the value update should work.  I was not certain that the proposed changes did not adversely effect other subclasses of text_input, but I couldn't find any problems.

bryevdv:
@AaronWatters thanks for the edit, @mattpap and I are both heads down in other work over the next week or so but we would both like to get a fix in for autocomplete before the next release (especially now that it is possible to create some selenium tests that will fully exercise these widgets under automated test). I appreciate your patience and your contribution, we will get back to you as soon as we are able.

bryevdv:
Just as an update, waiting to look at this until after the current big WIP PR for layout is done/merged

mattpap:
@bryevdv, given we are waiting here for the layout PR, which rewrites autocomplete widget almost from scratch, fixing isssue #6822 as a side effect, I think we can close this PR (unless we want to change the order of merges).

bryevdv:
I'm fine with that, @AaronWatters thank you very much for taking the time to submit this. Given the other large ongoing PR work around all widgets and layout, I am going to go ahead and close this since it will be addressed in time by that work.

Issue #7987, opened 2018-06-12
Restrict the drawing tools to draw a single glyph


philippjfr:
The drawing tools such as the PointDrawTool, PolyDrawTool and BoxEditTool currently insert a new point, polygon, or box every the user initiates a draw action. In certain scenarios it is useful to only allow drawing a single glyph and override it every time a new draw action is executed. It would therefore be helpful if an option was added to these tools that allows declaring whether to limit the number of drawn glyphs, this could take the form of an option to limit the number of drawn glyphs or simply to toggle between drawing a single or multiple glyphs.

PR #8127, opened 2018-08-01
Allow ability to overwrite existing glyphs in drawing tools

philippjfr:
This PR addresses a feature requested from the existing drawing tools, which is to overwrite the last drawn glyph. This is useful when you only want to draw a single box or polygon.

The spelling to enable this behavior is to set the new property ``overwrite=True``, suggestions for alternative spelling would be welcome (@jbednar?).

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/7987
- [x] tests added / passed

### DIFF
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -1251,6 +1251,11 @@ class BoxEditTool(EditTool, Drag, Tap):
     the vertical dimension can be controlled.
     """)
 
+    overwrite = Bool(default=False, help="""
+    Whether the tool should overwrite the existing box, if there is one,
+    or whether it should add a new box on every draw event (the default).
+    """)
+
     @error(INCOMPATIBLE_BOX_EDIT_RENDERER)
     def _check_compatible_renderers(self):
         incompatible_renderers = []
@@ -1348,6 +1353,11 @@ class PolyDrawTool(EditTool, Drag, Tap):
     drag = Bool(default=True, help="""
     Enables dragging of existing patches and multi-lines on pan events.""")
 
+    overwrite = Bool(default=False, help="""
+    Whether the tool should overwrite the existing box, if there is one,
+    or whether it should add a new box on every draw event (the default).
+    """)
+    
     @error(INCOMPATIBLE_POLY_DRAW_RENDERER)
     def _check_compatible_renderers(self):
         incompatible_renderers = []
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -1254,7 +1254,8 @@ class BoxEditTool(EditTool, Drag, Tap):
     num_objects = Int(default=0, help="""
     Defines a limit on the number of boxes that can be drawn. By
     default there is no limit on the number of objects, but if enabled
-    the oldest drawn box will be overwritten when the limit is reached.
+    the oldest drawn box will be dropped to make space for the new box
+    being added.
     """)
 
     @error(INCOMPATIBLE_BOX_EDIT_RENDERER)
@@ -1311,7 +1312,8 @@ class PointDrawTool(EditTool, Drag, Tap):
     num_objects = Int(default=0, help="""
     Defines a limit on the number of points that can be drawn. By
     default there is no limit on the number of objects, but if enabled
-    the oldest drawn point will be overwritten when the limit is reached.
+    the oldest drawn point will be dropped to make space for the new
+    point.
     """)
 
     @error(INCOMPATIBLE_POINT_DRAW_RENDERER)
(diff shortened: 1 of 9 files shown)


bokeh/models/tools.py
```diff
@@ -1348,6 +1353,11 @@ class PolyDrawTool(EditTool, Drag, Tap):
     drag = Bool(default=True, help="""
     Enables dragging of existing patches and multi-lines on pan events.""")
 
+    overwrite = Bool(default=False, help="""
+    Whether the tool should overwrite the existing box, if there is one,
+    or whether it should add a new box on every draw event (the default).
+    """)
```
jbednar: This one isn't a box, so "existing Poly"?  Or "existing object" (here and above)?

philippjfr:
>I don't suppose it's easy to make it support a fixed number of objects, instead?

Seems like a reasonable suggestion but I do wonder a bit about the usability, say you set n=3, now you draw 3 things. It would now always overwrite the last thing you drew. You could of course delete one of the earlier drawn objects manually so it's not totally useless, but just a bit weird I suppose.

If we decide to do that what would that parameter be called since ``n`` on its own wouldn't be very descriptive? ``draw_limit`` or something?

jbednar:
My assumption was that it would be a fixed-size queue, pushing a new item onto the top of the queue and popping the oldest one off the other end.  E.g. for this example:
image
you could set a limit of 5, and then it would show the last 5 ROIs you drew.   Here, with an overlay, it might not matter if the number of ROIs goes up forever, but if they were laid out in separate subplots, pretty soon the plot would get unusable.  So e.g. one could set up a fixed layout with three subplots on the right, and the last three ROIs would map to those.

`draw_limit` or `num_objects` or something like that, sure.

bokeh/models/tools.py
philippjfr: Right, should be polygon/object here will fix when I switch to the ``num_objects`` suggestion.

bokehjs/src/lib/models/tools/edit/box_edit_tool.ts
```diff
@@ -140,6 +147,7 @@ export class BoxEditToolView extends EditToolView {
 export namespace BoxEditTool {
   export interface Attrs extends EditTool.Attrs {
     dimensions: Dimensions
+    num_objects: Number
```
mattpap: `number`.

bokehjs/src/lib/models/tools/edit/poly_draw_tool.ts
```diff
@@ -194,6 +201,7 @@ export class PolyDrawToolView extends EditToolView {
 export namespace PolyDrawTool {
   export interface Attrs extends EditTool.Attrs {
     drag: boolean
+    num_objects: Number
```
mattpap: Same here.

bokehjs/src/lib/models/tools/edit/poly_draw_tool.ts
```diff
@@ -219,6 +227,7 @@ export class PolyDrawTool extends EditTool {
 
     this.define({
       drag: [ p.Bool, true ],
+      num_objects: [ p.Int, null ]
```
mattpap: The default in Python is `0`.

bryevdv:
> the oldest drawn box will be overwritten when the limit is reached.

when I read this my mind went to "last drawn" I think these docs would be clearer with verbiage that explicitly mentions queue behaviour, i.e. popping off the oldest, to make room for a new one, rather than describing as "overwriting"

philippjfr:
Agreed, "overwriting" is confusing in this context. Will fix shortly.

bokehjs/src/lib/models/tools/edit/edit_tool.ts
```diff
@@ -53,6 +53,17 @@ export abstract class EditToolView extends GestureToolView {
     cds.selection_manager.clear();
   }
 
+  _pop_glyphs(cds: ColumnarDataSource, num_objects: number, xkey: string | null, ykey: string | null) {
+    // Pops rows in the CDS until only num_objects are left
+    while (num_objects &&
+           ((xkey && cds.get_array(xkey).length >= num_objects) ||
+            (ykey && cds.get_array(ykey).length >= num_objects))) {
+      for (const column of cds.columns()) {
+        cds.get_array(column).splice(0, 1)
```
bryevdv: Why the outer loop? wouldn't this work
```
N = cds.get_array(xkey).length = num_objects
cds.get_array(column).splice(0, N)
```
Also AFAIK Typed arrays do not have a `splice` merthod

philippjfr: > Why the outer loop? Can't you just compute the number to delete up front?

Good point.

>Also AFAIK Typed arrays do not have a splice method

This is true, the other code already converts typed arrays in some cases, probably have to do that here too.

philippjfr:
Ready for another round of review and hopefully merging.

PR #8132, opened 2018-08-02
Updating the docstring for the dimensions property for Band, Span and Whisker

### LINKED ISSUE #8131: Docstrings not clear for dimension property for Band, Span and Whisker class
In using Whisker to plot error bars on figures there is a lack of information associated with the ``dimension`` property - it is not clear how you should use it. 

The correct method is to specify ``dimension=height`` for error bars aligned in the ``y``-direction and ``dimension=weight`` for error bars aligned in the ``x``-direction. 

The docstrings should be updated for this property to help users on how to use it.

cao-jacky:
Providing a clearer explanation on the usage of the ``dimensions`` property in the classes Band, Span, and Whisker.

- [x] issues: fixes #8131, documentation update 
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -528,7 +528,8 @@ class Band(Annotation):
     """)
 
     dimension = Enum(Dimension, default='height', help="""
-    The direction of the band.
+    The direction of the band can be specified by setting this property 
+    to ``height`` (``y`` direction) or ``dimension`` (``x`` direction).
     """)
 
     source = Instance(DataSource, default=lambda: ColumnDataSource(), help="""
@@ -856,7 +857,8 @@ class Span(Annotation):
     """)
 
     dimension = Enum(Dimension, default='width', help="""
-    The direction of the span.
+    The direction of the band can be specified by setting this property 
+    to ``height`` (``y`` direction) or ``dimension`` (``x`` direction).
     """)
 
     x_range_name = String('default', help="""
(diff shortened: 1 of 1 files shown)


bokeh/models/annotations.py
```diff
@@ -1019,7 +1021,8 @@ class Whisker(Annotation):
     """)
 
     dimension = Enum(Dimension, default='height', help="""
-    The direction of the band.
+    The direction of the band can be specified by setting this property 
+    to ``height`` (``y`` direction) or ``dimension`` (``x`` direction).
```
bryevdv: @cao-jacky as long as we are updating this, I notice it looks like the text has a copy paste error and should be "direction of the *whisker*" would you mind fixing that here as well?

bokeh/models/annotations.py
```diff
@@ -528,7 +528,8 @@ class Band(Annotation):
...
```
bryevdv: typo, should be "or ``width``" (same comment for others as well)

bokeh/models/annotations.py
cao-jacky: Oh oops, will do

bokeh/models/annotations.py
cao-jacky: Will change

bryevdv:
@cao-jacky Thanks for the PR! I noted a couple of typos an another fix that makes sense to make here. What do you think about using standard quotes instead of backticks? I think that would make e.g. ipython output more readable, also be clear that the values to set are strings.

cao-jacky:
Yeah, I think standard quotes would clear the confusion between whether you should set a numerical value or use a string for the property.

bryevdv:
@cao-jacky our linter is picky about trailing whitespace, I have take the liberty to edit the fixes inline on GH UI. Look forward to merging when tests are green

bryevdv:
Thanks @cao-jacky !

cao-jacky:
No problem @bryevdv, I'm glad to have helped!

Issue #8032, opened 2018-06-27
Draw and Edit tools event handling on server broken


philippjfr:
Playing around with the draw and edit tools on bokeh server that even after https://github.com/bokeh/bokeh/issues/7745 was resolved many actions don't seem to trigger Python side events. Specifically drawing a new point using PointDrawTool and drawing box using the BoxEditTool seems to trigger no event at all on the Python end, and the PolyEditTool seems to cause an endless chain of events being sent. The fix in https://github.com/bokeh/bokeh/pull/7818/files addressed the missing updates using on point drag events by explicitly setting the data:

```python
      // This is only needed to call @_tell_document_about_change()
      renderer.data_source.data = renderer.data_source.data
```

I'd be happy to put together a PR to fix this in other cases, but I still don't quite understand why this is necessary. Shouldn't property change events be sufficient?

```python
data_source.properties.data.change.emit()
```

bryevdv:
I'm not sure, but there is no point in making PRs for these things until we can find a way to maintain them under test.

PR #8137, opened 2018-08-06
Ensure draw/edit tools sync data on bokeh server

philippjfr:
This PR ensures that the drawing/editing tools trigger server side events by explicitly setting the data when necessary. I'll also add some selenium tests to this PR to avoid regressions.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/8032
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/tools/edit/box_edit_tool.ts
+++ b/bokehjs/src/lib/models/tools/edit/box_edit_tool.ts
@@ -67,6 +67,7 @@ export class BoxEditToolView extends EditToolView {
     ds.change.emit();
     if (emit) {
       ds.properties.data.change.emit();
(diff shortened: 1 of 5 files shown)


bryevdv:
I'm still working on a server test workflow for selenium, hoping I can knock something out today.

bryevdv:
Offhand my only comment about the tests is they the might be more robust either with an approximate fp comparison, or with a test situation engineered to yield round numbers

philippjfr:
>Offhand my only comment about the tests is they the might be more robust either with an approximate fp comparison

I think that might be the only option. Since I'm generating the events in screen space I doubt I'll be able to generate round numbers very easily.

PR #8144, opened 2018-08-07
Add context manager and decorator to control property validation

### LINKED ISSUE #8138: Turn off self.validate checks during update
When running a Bokeh server with streaming updates in my application I often find that bokeh takes up 30% of the CPU time, about half of which is consumed by the following self.validate checks:

```diff
diff --git a/bokeh/core/property/bases.py b/bokeh/core/property/bases.py
index d28aa2b..a50ed9e 100644
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -260,7 +260,7 @@ class Property(PropertyDescriptorFactory):
 
         '''
         try:
-            self.validate(value, False)
+            # self.validate(value, False)
         except ValueError as e:
             return False
         else:
@@ -275,7 +275,7 @@ class Property(PropertyDescriptorFactory):
 
     def prepare_value(self, obj_or_cls, name, value):
         try:
-            self.validate(value)
+            # self.validate(value)
         except ValueError as e:
             for tp, converter in self.alternatives:
                 if tp.is_valid(value):
```

I would like some way to disable these, preferably without clobbering with other users' Bokeh experience (my bokeh server application is often embedded within other Jupyter sessions, where the user may use Bokeh on their own).

So some questions:

1.  Does the current environment variable solution with `BOKEH_VALIDATE_DOC=0` disable these checks?
2.  Is there some way I can turn them off programmatically,rather than tweaking an environment variable?  
3.  Is there some safe way that I can mark all of the properties that I care about as being safe and not needing validation, while not affecting other users' experience?

Some thoughts:

1.  Can we construct a function that propagates through a figure, CDS, or doc and marks all properties as not needing to be validated?
2.  Can we temporarily set a global before and after calling update/stream/patch that signals bokeh to not validate during that period but unset it directly afterwards?

    ```python
    with bokeh.no_validate:
        source.update(...)
    ```

Follows on from https://github.com/bokeh/bokeh/issues/6042

mrocklin:
This allows users to control whether or not they want validate checks when
updating properties.

Examples
--------

    with validate(False):
        ...

or

    @without_property_validation
    def update(attr, old, new):
        ...

Fixes #8138

- [x] issues: fixes #8138
- [x] tests added / passed

### DIFF
--- a/bokeh/core/property/tests/test_bases.py
+++ b/bokeh/core/property/tests/test_bases.py
@@ -148,31 +148,8 @@ def test_property_matches_dicts_with_index_values(capsys, pd):
     out, err = capsys.readouterr()
     assert err == ""
 
-
-def test_validate():
+def test_validation_on():
+    assert pb.Property._should_validate == True
     assert pb.validation_on()
-    with pb.validate(False):
-        assert not pb.validation_on()
-    assert pb.validation_on()
-
-    with pb.validate(False):
-        assert not pb.validation_on()
-        with pb.validate(True):
-            assert pb.validation_on()
-        assert not pb.validation_on()
-    assert pb.validation_on()
-
-    pb.validate(False)
+    pb.Property._should_validate = False
     assert not pb.validation_on()
-    pb.validate(True)
-    assert pb.validation_on()
-
-
-def test_without_property_validation():
-    @pb.without_property_validation
-    def f():
-        assert not pb.validation_on()
-
-    assert pb.validation_on()
-    f()
-    assert pb.validation_on()
--- a/bokeh/core/property/tests/test_bases.py
+++ b/bokeh/core/property/tests/test_bases.py
@@ -147,3 +147,22 @@ def test_property_matches_dicts_with_index_values(capsys, pd):
     assert p.matches(d1.index, 10) is False
     out, err = capsys.readouterr()
     assert err == ""
+
+
+def test_validate():
+    assert pb.validation_on()
+    with pb.validate(False):
+        assert not pb.validation_on()
+    assert pb.validation_on()
+
+    with pb.validate(False):
+        assert not pb.validation_on()
+        with pb.validate(True):
+            assert pb.validation_on()
+        assert not pb.validation_on()
+    assert pb.validation_on()
+
+    pb.validate(False)
+    assert not pb.validation_on()
+    pb.validate(True)
+    assert pb.validation_on()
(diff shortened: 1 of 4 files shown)


mrocklin:
Very rough draft.  Thoughts on this @bryevdv ?

mrocklin:
I'll take all the approval I can get, misdirected or otherwise :)

I've handled the first few points you raised above.  I think I'd prefer to defer to your judgment on where to place this within the Bokeh API.

bryevdv:
`bokeh.properties` seems like a reasonable place, though I think we would not want to actually move things there because that would set up a circular import with `bokeh.property.bases`. So instead we would need to transitively import these functions from `bokeh.property.bases` into `bokeh.properties` and then fix up the ref guide to display them in the "proper" place. 

Alternatively, if you want to move the actual implementations to `bokeh.properties` then going back to the original module global (in `bokeh.property.bases`, but set from `bokeh.properties`) would probably work.

bryevdv:
Just some linter errors at the moment:
```
E         ./bokeh/core/property/bases.py:482:6: F821 undefined name 'functools'
E         ./bokeh/core/property/bases.py:482:22: F821 undefined name 'input_function'
E         ./bokeh/core/property/bases.py:485:20: F821 undefined name 'input_function'
E         ./bokeh/core/property/bases.py:486:12: F821 undefined name '_'
```

mrocklin:
The linter errors caught real issues.  I've resolved them and added a proper test for the behavior.  

If you don't mind I think I'd prefer to leave the problem of where to place these within the API to you.  Would that be ok?

bryevdv:
Sure I will take a look at it tomorrow

bryevdv:
@mrocklin please look over my changes and confirm that everything is acceptable

bokeh/core/property/tests/test_bases.py
```diff
@@ -147,3 +147,9 @@ def test_property_matches_dicts_with_index_values(capsys, pd):
     assert p.matches(d1.index, 10) is False
     out, err = capsys.readouterr()
     assert err == ""
+
+def test_validation_on():
+    assert pb.Property._should_validate == True
+    assert pb.validation_on()
+    pb.Property._should_validate = False
+    assert not pb.validation_on()
```
mrocklin: You might want to clean up `_should_validate` to the previous value after this test.

mrocklin:
Other than a small comment everything here looks great to me.  Thank you
for finishing up.

PR #8145, opened 2018-08-07
Jsnsross/8142 span datetime

### LINKED ISSUE #8142: Span location cannot accept datetime whereas Labels can deal with it.
OK I hope this is an easy one. I'm working with spans and labels for the spans and I see that the Span object requires an Integer location meaning you have to do a little bit of conversion to your datetime data whereas Labels can accept a datetime object directly for the x argument and automatically converts it to microseconds since epoch. So when I make a span and a label for the span next to it, I'm coding the x-axis location differently (saving a bit of code by not converting the location for the Label)

Since the formulation for automatic conversion of datetimes to integer exists for Labels, it would be nice to use the same thing for Span's and thus relax the integer type requirement for the location argument.

jasonsross:
- [x] issues: fixes #8142 
- [x] tests added / passed

resolved issue with Span not accepting a datetime by copying formulation used for x and y arguments in Label. Also added tests for Span.location and Label.x / Label.y to ensure that datetime is properly being converted to milliseconds-since-epoch

### DIFF
--- a/bokeh/models/tests/test_annotations.py
+++ b/bokeh/models/tests/test_annotations.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 import mock
-
+import pandas as pd
 from bokeh.core.properties import field, value
 from bokeh.core.validation import check_integrity
 from bokeh.models.annotations import (
@@ -340,6 +340,15 @@ def test_Span():
         "render_mode"
     ], LINE)
 
+def test_Span_accepts_datetime_location():
+    obj = Span(location = pd.to_datetime('2018-8-7'))
+    assert obj.location == 1533600000000.0
+
+def test_Label_accepts_datetime_location():
+    obj = Label(x = pd.to_datetime('2018-8-7'),
+                y = pd.to_datetime('2018-8-7'))
+    assert obj.x == 1533600000000.0
+    assert obj.y == 1533600000000.0
 
 def test_Title():
     title = Title()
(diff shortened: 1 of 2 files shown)


bokeh/models/tests/test_annotations.py
```diff
@@ -340,6 +345,9 @@ def test_Span():
         "render_mode"
     ], LINE)
 
+def test_Span_accepts_datetime_location():
+    obj = Span(location = pd.to_datetime('2018-8-7'))
+    assert obj.location == 1533600000000.0
 
 def test_Title():
     title = Title()
```
bryevdv: @jasonsross one small change is needed. We have a test that makes sure that basic things work without pandas installed. So the test should not import pandas directly. There are two options:

* add an argument `pd` to the two test functions, e.g `def test_Label_accepts_datetime_location(pd)`. We have a pytest fixture that automatically will provide pandas if it installed or skip the test if not

* use the built in `datetime` module, or `numpy` datetime functions, instead of pandas

Either is ok with me!

jasonsross: @bryevdv sounds good I've dropped pandas for datetime so if pandas is not installed, we don't lose this test.

bryevdv:
Great first PR @jasonsross, thanks for the contribution!

jasonsross:
@bryevdv Thanks for the help getting this experience. I look forward to finding more to contribute to.

Issue #8133, opened 2018-08-03
Freehand drawing tool


rsignell-usgs:
It would be very useful to have a freehand drawing tool in Bokeh, as this would allow us to run our workflow for pre-trained deep convolutional neural network image segmentation developed by @dbuscombe-usgs.    

The workflow is that the user is presented with part of the image, and cycles through the different classes (e.g. sand, water, vegetation, etc) scribbling on each one to indicate the regions they see that match the class.   A 2 minute video is here: https://www.youtube.com/watch?v=0sXYKmbLz_s, and an image is shown below where the user has already indicated sand regions and has moved on to water:
![2018-08-02_21-00-12](https://user-images.githubusercontent.com/1872600/43618831-a6f4d606-9698-11e8-85b9-447064867ff6.jpg)

It would be awesome to be able to have this interaction with an image in a notebook cell, returning the freehand line coordinates to python.

PR #8147, opened 2018-08-08
Added FreehandDrawTool

philippjfr:
This PR adds a FreehandDrawTool which draws on click and drag events and also allows deleting previously lines by selecting them and hitting backspace.

It also needs an icon (ping @jbednar), I'm imagining something like [this](https://upload.wikimedia.org/wikipedia/commons/thumb/9/91/Inkscape_icons_draw_freehand.svg/2000px-Inkscape_icons_draw_freehand.svg.png) taken from Inkscape. 

Small demo:

freehand

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/8133
- [x] tests added / passed


jbednar:
How about:

image

philippjfr:
Perfect thanks. Now added unit tests and integration tests, just need to add docs and an example now.

philippjfr:
Here's what the icon looks like in the toolbar:

screen shot 2018-08-08 at 2 01 39 pm

jbednar:
Seems clear enough.  Any other icons you need for other new tools while I'm at it?

philippjfr:
>Any other icons you need for other new tools while I'm at it?

See https://github.com/pyviz/EarthSim/issues/130

philippjfr:
Here is the gif that should be uploaded to http://bokeh.pydata.org/static/freehand_draw_keyboard_optimized.gif:

freehand_draw_keyboard_optimized

philippjfr:
Ready to review now.

tests/integration/tools/test_freehand_draw_tool.py
```diff
@@ -0,0 +1,163 @@
...
+        assert 'active' in button.get_attribute('class')
+
+        assert page.has_no_console_errors()
+
+    def test_can_be_deselected_and_selected(self, single_plot_page):
+        plot = _make_plot()
+
+        page = single_plot_page(plot)
+
+        target = 'bk-tool-icon-freehand-draw'
```
bryevdv: use `get_toolbar_button` for this 

https://github.com/bokeh/bokeh/blob/master/bokeh/_testing/plugins/bokeh.py#L199-L200

bokehjs/src/lib/models/tools/edit/freehand_draw_tool.ts
```diff
@@ -0,0 +1,110 @@
...
+import {isArray} from "core/util/types"
+import {EditTool, EditToolView} from "./edit_tool"
+
+export class FreehandDrawToolView extends EditToolView {
+  model: FreehandDrawTool
+
+  _draw(ev: UIEvent, mode: string, emit: boolean = false): void {
+    if (!this.model.active) { return; }
+
+    const renderer = this.model.renderers[0];
```
bryevdv: If the tool can only operate on one renderer (which seems sensible, offhand) then the API should reflect that

bryevdv:
comments left

bokehjs/src/lib/models/tools/edit/freehand_draw_tool.ts
philippjfr: I wanted to maintain consistency with other drawing tools, which generally allow drawing to only the first renderer but do allow deleting and moving glyphs on multiple renderers. In this particular case you can only delete items on multiple renderers so I'd definitely be happy to change it.

tests/integration/tools/test_freehand_draw_tool.py
philippjfr: Will do.

philippjfr:
Thanks, will address them shortly.

bokehjs/src/lib/models/tools/edit/freehand_draw_tool.ts
philippjfr: After investigating this I think we should either stay consistent and keep this as a list or change all the editing tools, except for the PolyEditTool, to accept only a single renderer. That would be an API change and I'd have to do a fair bit of refactoring to make it work, so if you think we should go ahead with that I'd probably prefer merging this as is and doing the refactoring of all the tools at once in a new PR.

philippjfr: Just to be clear, it's not a difficult change, I would just want to do it all at once since it's tied up with changes to the base class.

bryevdv: Changing everything seems disruptive at this point, so let's just make sure the documentation is excessively clear about expected behaviours

philippjfr: Okay fine with me. I'm expecting to make two more drawing tool related PRs after this one:

1. Add support for vertex snapping to the poly draw and edit tools
2. Final cleanup and tests for all edit tools

I'd propose that we merge this PR and I add that those clarifications in 2.

UConnPhysicsLabTech-ZT:
Hello everyone. I am the Laboratory Technician for Undergraduate Physics at UConn. We are looking into converting our lab notebooks from Mathematica to Python and this drawing feature helps us move away from the drawing tools in Mathematica. We are asking students to draw predictions of position, velocity, and acceleration plots of a cart accelerating down a ramp. We want to use this freehand draw tool inside of our Goolge Colab notebook. We are able to get the code to generate the drawing environment however, it is a bit difficult for our introductory students to call and upload the saved png from their Drive folders (we don't want to ask the students to mount the Google Drive everytime). What we want to do is have the students be able to save the drawings they made to the output file so that when their TA opens the file after not having run the cells, can still view the students answer.

tl;dr
Is there a way to save the drawing created on the output of the FreehandDrawingTool directly inside the output html file so that it can be seen when the file is reopened by someone else? I do not want to use the export png feature, since it gets too complicated for students to mount the Google Drive and load the file.

bryevdv:
@UConnPhysicsLabTech-ZT please take support questions to the Project Discourse, not the issue tracker: https://discourse.bokeh.org

UConnPhysicsLabTech-ZT:
> @UConnPhysicsLabTech-ZT please take support questions to the Project Discourse, not the issue tracker: https://discourse.bokeh.org

Thanks. Just posted there.

https://discourse.bokeh.org/t/updating-save-freehanddraw-figure-after-drawing-on-output/3976/1

Issue #8172, opened 2018-08-17
Make image smoothing configurable on TileRenderer


philippjfr:
Usually you want web map tiles to be rendered with smoothing enabled because they are often low resolution and would look very pixelated otherwise. However in certain cases you do want the pixelated look because the smoothing blurs the data across boundaries, which can give misleading impressions or simply for aesthetic reasons. One such example is a set of tiles generated from the census dataset, here seen with smoothing enabled:

![screen shot 2018-08-17 at 1 47 02 pm](https://user-images.githubusercontent.com/1550771/44266866-1171e780-a224-11e8-9b2b-989469fdc69e.png)

And here without the smoothing:

![screen shot 2018-08-17 at 1 46 52 pm](https://user-images.githubusercontent.com/1550771/44266861-0e76f700-a224-11e8-9f75-dfc3d36bcb97.png)

PR #8173, opened 2018-08-17
Make TileRenderer smoothing configurable

philippjfr:
As the title says, exposes control over canvas image smoothing on the TileRenderer.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/8172
- [x] tests added / passed


philippjfr:
Not sure how I can test this, since the smoothing is only disabled temporarily.

bokeh/models/renderers.py
```diff
@@ -53,6 +53,10 @@ class TileRenderer(DataRenderer):
     tile opacity 0.0 - 1.0
     """)
 
+    smoothing = Bool(default=True, help="""
+    Enable image smoothing for the rendered tiles
```
mattpap: A trailing period would be nice.

mattpap:
> Not sure how I can test this, since the smoothing is only disabled temporarily.

I would just create a static integration example with the same tile source side by side with smoothing enabled and disabled. If something get broken, we will see thin in image diff.

bryevdv:
@philippjfr does the suggested test sound reasonable for now? Do you have any questions about implementing it?

philippjfr:
>does the suggested test sound reasonable for now? Do you have any questions about implementing it?

Yes, had half forgotten about this after I addressed the flicker issue. If you could point me to where the example would live I'll add it tonight. Not sure how noticeable the difference between smoothed and non-smoothed will be but presumably an image diff will definitely pick up on it.

philippjfr:
Let me know if the example looks okay.

philippjfr:
Ah just realized I put it in tests/integration and you wanted it in examples/integration, moving it now.

mattpap:
Looks good. Difference in smoothness is clearly visible.

PR #8184, opened 2018-08-23
Bryanv/8179 show accumulation

### LINKED ISSUE #8179: show() accumulates objects
Starting in 0.13, this script:

```

p = figure()
show(p)

p2 = figure()
show(p2)
```
generates two plots in the output instead of one.

bryevdv:
- [x] issues: fixes #8179
- [x] tests added / passed

This PR makes the following changes:

* `show` only adds content to `curdoc` when showing in the notebook. It is necessary in the notebook that *all* output be in the same global document, but this is not necessary or especially useful for typical usage outside the notebook. 

* Consolidates ill-specified and undocumented `_ModelInDocument` and `_ModelInEmptyDocument` into a documented context manager `OutputDocumentFor` that is comprehensively tested.

The old context managers seemed to try to handle cases that do no seem important. e.g. mixing documents and models together, making their behaviour overly complicated. 

`OutputDocumentFor` tries to just re-use an existing document where possible. If there is no existing document, it will create one, and make it the permanent new document (i.e. it will call `add_root` on them). If there are too many documents, or if we only want to consider a *subset of roots*.  then `OutputDocumentFor` creates a "temp doc" that effectively re-homes the models, to use inside the with-block

The last bit is what could be considered bit kludgy. Specifically, because the "temp" doc is not completely initialized, but is useful enough to use for the serialization cases we currently have. The *implementation* could be perhaps improved with a better internal API. Or perhaps the "temp doc" concept could be explicitly realized in a new type. Comments here are welcome. 

However, I believe the outward facing *behaviour* of `OutputDocumentFor` is fairly close to what it needs to be. I'd like to come to explicit agreement on this point. Then as long as those behaviours are fully codified in tests we can improve implementation now or later. 

cc @mattpap 

One thing to note also is that `OutputDocumentFor` is somewhat faster now, and especially with validation turned off, only takes on the order of miroseconds on my latptop.

### DIFF
--- a/bokeh/embed/util.py
+++ b/bokeh/embed/util.py
@@ -45,9 +45,13 @@ from ..util.serialization import make_id
 # Dev API
 #-----------------------------------------------------------------------------
 
-# Note: Comms handling relies on the fact that the new_doc returned
-# has models with the same IDs as they were started with
 
+class FromCurdoc(object):
+    ''' This class merely provides a non-None default value for ``theme``
+    arguments, since ``None`` itself is a meaningful value for users to pass.
+
+    '''
+    pass
 
 @contextmanager
 def OutputDocumentFor(objs, apply_theme=None, always_new=False):
@@ -101,6 +105,9 @@ def OutputDocumentFor(objs, apply_theme=None, always_new=False):
         Document
 
     '''
+    # Note: Comms handling relies on the fact that the new_doc returned
+    # has models with the same IDs as they were started with
+
     if not isinstance(objs, Sequence) or len(objs) == 0 or not all(isinstance(x, Model) for x in objs):
         raise ValueError("OutputDocumentFor expects a sequence of Models")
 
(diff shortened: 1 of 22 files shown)


bryevdv:
@jlstevens @philippjfr if you have any comments as well please chime in.

bryevdv:
I should add, I have tested this manually:

* bokeh serve apps
* notebook including push_notebook and embedded apps
* ditto jupyter-lab
* components
* autoload_static
* built all the docs

bokeh/embed/util.py
```diff
@@ -49,6 +50,149 @@
...
+def _create_temp_doc(models):
+    doc = Document()
+    for m in models:
+        doc._all_models[m._id] = m
+        m._temp_document = doc
+        for ref in m.references():
+            doc._all_models[ref._id] = ref
+            ref._temp_document = doc
+    doc._roots = models
+    return doc
```
bryevdv: These four functions could be considered "quick and dirty". If you have preferences for internal-ish APIs to implement this, please speak up @mattpap

bryevdv:
I think we could also consider simplifying this to only accept `Seq[Model]`?  I am not sure what the use case for passing in a `Document` is at all, and I could live with `OutputDocumentFor([m])` instead of `OutputDocumentFor(m)`

bryevdv:
OK I went ahead and made those simplifying changes. I also split up `bokeh.embed.util` because it was a huge tangled mess. There is more I want to do but I will continue with cleanup in the "embed json" PR

bryevdv:
@mattpap I"m going to go ahead and merge this in the next hour or so. I think there is plenty of room for more improvement on the implementation, but I think the highest level behavior of `OutputDocumentFor`  is now reasonable, documented, and tested, so implementation improvements can follow on.

mattpap:
Go ahead, I don't have capacity right now for a meaningful review of a PR of this size anyway.

mattpap:
>  think we could also consider simplifying this to only accept Seq[Model]? I am not sure what the use case for passing in a Document is at all, and I could live with OutputDocumentFor([m]) instead of OutputDocumentFor(m)

Sure, the simpler the better. Supporting a document was, I presume, a clever thing to do at the time. If there is no immediate point for supporting it, lets remove this.

bryevdv:
> Sure, the simpler the better. Supporting a document was, I presume, a clever thing to do at the time. If there is no immediate point for supporting it, lets remove this. 

Already done, will merge now

Issue #8190, opened 2018-08-27
Pandas > 0.23.0 & < 0.23.4 Breaks boxplot.py Example


HuntJSparra:
Pandas changed the behavior of what empty DataFrame/Series Columns return when indexed. Previously, it returned an empty column, but since 0.23.0 it now returns a KeyError. This was unintentionally "fixed" in 0.23.4, but will be removed again in 0.24.0. See: https://github.com/pandas-dev/pandas/issues/22221

The issue with the boxplot.py example is that it can generate empty columns and then index them. Thus, there is a chance that a KeyError occurs while generating the plot's outliers. This can also cause issues with generating the Bokeh documentation locally.

PR #8191, opened 2018-08-27
Fix Past & Future Issue with boxplot.py & Pandas 0.23.x

HuntJSparra:
Minor fix for https://github.com/bokeh/bokeh/issues/8190


bryevdv:
Thanks @HuntJSparra will merge when green

PR #8085, opened 2018-07-17
Improve the layout subsystem

mattpap:
addresses #8084

fixes #4407 (new feature)
fixes #4643 (irrelevant)
fixes #5169 (fixed, needs a test)
fixes #5572 (fixed)
fixes #6294 (fixed)
fixes #6417 (fixed, needs a test)
fixes #6768 (fixed, needs a test)
fixes #7120 (fixed, needs a test)
fixes #7185 (irrelevant)
fixes #7497 (fixed)
fixes #7590 (fixed) [already closed?]
fixes #7771 (fixed)
fixes #8227 (fixed)
fixes #8229 (new feature, needs tests, examples)
fixes #6822 (fixed)
fixes #4395 (fixed)

fixes #8391 (fixed, needs test)
fixes #7260 (fixed)
fixes #7454 (fixed, needs test)
fixes #5749 (fixed)
fixes #6259 (fixed)
fixes #4608 (fixed)
fixes #8611 (fixed)

PR #8193, opened 2018-08-27
Bryanv/5231 json embeds

### LINKED ISSUE #5231: Save Bokeh Plot as Bokeh Plot
I would like to save a snapshot of Bokeh plot created by a Bokeh server application in all of the following ways:
1.  Download the contents of the ColumnDataSource as JSON/msgpack/whatever
2.  Download a static HTML file of the Bokeh plot
3.  Publish that bokeh plot as a gist
### Dask's use case

For example with Dask.distributed, the current best way to have a discussion about performance is by referring to our task-stream diagnostic Bokeh plot.  This occurs when I write blogposts discussing algorithms and when users have performance questions.  If there was a Bokeh Tool that let users publish a static view of their task-stream plot as a gist from within the Dask Dashboard then it would elevate the level of conversation significantly (and produce a lot of cool looking Bokeh images).  As a pleasant side effect, I would start including Bokeh plots way more often when writing blogposts and documentation.
### How to do this?

When I briefly mentioned this to @birdsarah she recommended building a custom tool like the current `SaveTool` for Dask which would dump to one of the above forms rather than to png.  However, looking at this a bit more I don't think that this functionality necessarily needs to be Dask-specific.  This may be of general utility and would, I think, encourage people to embed Bokeh plots more often within broadcast publications.  For example if they built a Bokeh plot in a notebook but wanted to include it in a blogpost then this might help.

bryevdv:
- [x] issues: fixes #5231
- [x] tests added / passed

So far this PR adds the ability to make a simple JSON export of a Bokeh root model from Python and then embed it in JS using `Bokeh.embed.embed_item` (see example below). Notes:

* It is not yet possible to export JSON from a JS doc and also pass that to `embed_item`. It seems that `to_json` includes a lot of layout related attrs that are not present in the python export and this confuses re-constitution of the document. I think #8085 may solve this issue since layout is moved off of models there. 

* currently the target ID is specified in `json_item(p, "myplot")` and them embedded with `embed_item(item)` on the JS side. But maybe the converse is better for specifying the target? e.g `json_item(p)` on the export side with the target specified in JS `embed_item(item, "myplot")`

* currently can only export a single root this way, perhaps would be good to be able to embed multiple roots at once but open to not-doing that now as long as a reasonable pathway later is not precluded. 

This PR also switches to simple IDs by default.

### DIFF
--- a/bokeh/embed/util.py
+++ b/bokeh/embed/util.py
@@ -24,6 +24,7 @@ log = logging.getLogger(__name__)
 # Standard library imports
 from collections import Sequence, OrderedDict
 from contextlib import contextmanager
+import uuid
 
 # External imports
 
@@ -31,7 +32,6 @@ from contextlib import contextmanager
 from ..document.document import Document
 from ..model import Model, collect_models
 from ..settings import settings
-from ..util.serialization import make_id
 
 #-----------------------------------------------------------------------------
 # Globals and constants
@@ -160,7 +160,7 @@ class RenderItem(object):
         if roots is None:
             roots = OrderedDict()
         elif isinstance(roots, list):
-            roots = OrderedDict([ (root, make_id()) for root in roots ])
+            roots = OrderedDict([ (root, str(uuid.uuid4())) for root in roots ])
 
         self.docid = docid
         self.sessionid = sessionid
@@ -233,6 +233,13 @@ class RenderRoots(object):
     def to_json(self):
         return OrderedDict([ (root._id, elementid) for root, elementid in self._roots.items() ])
 
+def standalone_docs_json(models):
+    '''
+
+    '''
+    docs_json, render_items = standalone_docs_json_and_render_items(models)
+    return docs_json
+
 def standalone_docs_json_and_render_items(models, suppress_callback_warning=False):
     '''
 
@@ -259,15 +266,15 @@ def standalone_docs_json_and_render_items(models, suppress_callback_warning=Fals
                 raise ValueError("to render a model as HTML it must be part of a document")
 
         if doc not in docs:
-            docs[doc] = (make_id(), OrderedDict())
+            docs[doc] = (str(uuid.uuid4()), OrderedDict())
 
         (docid, roots) = docs[doc]
 
         if model is not None:
-            roots[model] = make_id()
+            roots[model] = str(uuid.uuid4())
         else:
             for model in doc.roots:
-                roots[model] = make_id()
+                roots[model] = str(uuid.uuid4())
 
     docs_json = {}
     for doc, (docid, _) in docs.items():
--- a/bokeh/embed/util.py
+++ b/bokeh/embed/util.py
@@ -263,7 +263,7 @@ def standalone_docs_json_and_render_items(models, suppress_callback_warning=Fals
             doc = model.document
 
             if doc is None:
-                raise ValueError("to render a model as HTML it must be part of a document")
+                raise ValueError("A Bokeh Model must be part of a Document to render as standalone content")
 
         if doc not in docs:
             docs[doc] = (make_globally_unique_id(), OrderedDict())
--- a/bokeh/embed/util.py
+++ b/bokeh/embed/util.py
@@ -24,7 +24,6 @@ log = logging.getLogger(__name__)
 # Standard library imports
 from collections import Sequence, OrderedDict
 from contextlib import contextmanager
-import uuid
 
 # External imports
 
@@ -32,6 +31,7 @@ import uuid
 from ..document.document import Document
 from ..model import Model, collect_models
 from ..settings import settings
+from ..util.serialization import make_globally_unique_id
 
 #-----------------------------------------------------------------------------
 # Globals and constants
@@ -160,7 +160,7 @@ class RenderItem(object):
         if roots is None:
             roots = OrderedDict()
         elif isinstance(roots, list):
-            roots = OrderedDict([ (root, str(uuid.uuid4())) for root in roots ])
+            roots = OrderedDict([ (root, make_globally_unique_id()) for root in roots ])
 
         self.docid = docid
         self.sessionid = sessionid
@@ -266,15 +266,15 @@ def standalone_docs_json_and_render_items(models, suppress_callback_warning=Fals
                 raise ValueError("to render a model as HTML it must be part of a document")
 
         if doc not in docs:
-            docs[doc] = (str(uuid.uuid4()), OrderedDict())
+            docs[doc] = (make_globally_unique_id(), OrderedDict())
 
         (docid, roots) = docs[doc]
 
         if model is not None:
-            roots[model] = str(uuid.uuid4())
+            roots[model] = make_globally_unique_id()
         else:
             for model in doc.roots:
-                roots[model] = str(uuid.uuid4())
+                roots[model] = make_globally_unique_id()
 
     docs_json = {}
     for doc, (docid, _) in docs.items():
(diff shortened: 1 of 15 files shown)


philippjfr:
>currently the target ID is specified in json_item(p, "myplot") and them embedded with embed_item(item) on the JS side. But maybe the converse is better for specifying the target? e.g json_item(p) on the export side with the target specified in JS embed_item(item, "myplot")

Fully agree that specifying the target on the JS end would be more flexible.

bryevdv:
> Fully agree that specifying the target on the JS end would be more flexible.

For the single case, I agree 100%. What about we wish to allow embedding multiple roots? That co)uld be handled on the python side somewhat transparently, but making the JS called reponsible for mapping root id's to div ids seems untenable. Do we not allow that at all? (Then you'd have to resort to `components` for linked selection etc) or do we make it implicit in the order, e.g.
```
json_item(p1, p2, p3)
```

and then on the JS side:

```
embed_item(item, "p1div", "p2div", "p3div)"
```
?
 I suppose that could work tho the name probably needs to be pluralized or changed.

bokeh/embed/util.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, docid=None, sessionid=None, elementid=None, roots=None, use_f
         if roots is None:
             roots = OrderedDict()
         elif isinstance(roots, list):
-            roots = OrderedDict([ (root, make_id()) for root in roots ])
+            roots = OrderedDict([ (root, str(uuid.uuid4())) for root in roots ])
```
mattpap: May I propose `make_unique_id()` to assert in all cases what a unique ID is and make future changes easier.

bokehjs/src/lib/embed/index.ts
```diff
@@ -15,6 +15,21 @@ export {add_document_from_session} from "./server"
...
+type JsonItem = {doc: DocJson, root_id: string, target_id: string}
+interface Roots {[index: string]: string}
+
+export function embed_item(item: JsonItem) {
+  const docs_json: DocsJson = {}
+  const doc_id = uuid4()
+  docs_json[doc_id] = item.doc
+
+  const roots: Roots = {}
+  roots[item.root_id] = item.target_id
```
mattpap: `const roots: Roots = {[item.root_id]: item.target_id}` (aka. computed properties in ES2015).

bokehjs/src/lib/embed/index.ts
```diff
@@ -15,6 +15,21 @@ export {add_document_from_session} from "./server"
 export {embed_items_notebook, kernels} from "./notebook"
 export {BOKEH_ROOT, inject_css, inject_raw_css} from "./dom"
 
+type JsonItem = {doc: DocJson, root_id: string, target_id: string}
+interface Roots {[index: string]: string}
+
+export function embed_item(item: JsonItem) {
```
mattpap: All types of exported members must be exported. Also, explicit return type is needed. Also, you could use either `type` or `interface` consistently. It (almost) doesn't matter which one is used (until we get into inheritance, which isn't in the picture her).

bokeh/embed/util.py
bryevdv: Sure that seems reasonble

bokehjs/src/lib/embed/index.ts
bryevdv: Thanks!

bokehjs/src/lib/embed/index.ts
bryevdv: will update

bryevdv:
Silly me, it's easy to just accept both:
```
import json

from bokeh.embed import json_item

from flask import Flask
from jinja2 import Template

app = Flask(__name__)

page = Template("""
<!DOCTYPE html>
<html lang="en">
<head>
  <link href="http://localhost:5006/static/css/bokeh.css" rel="stylesheet" type="text/css">
  <script src="http://localhost:5006/static/js/bokeh.js"></script>
</head>

<body>
  <div id="myplot"></div>
  <div id="myplot2"></div>
  <script>
  fetch('/plot')
    .then(function(response) { return response.json(); })
    .then(function(item) { Bokeh.embed.embed_item(item); })
  </script>
  <script>
  fetch('/plot2')
    .then(function(response) { return response.json(); })
    .then(function(item) { Bokeh.embed.embed_item(item, "myplot2"); })
  </script>
</body>
""")

def make_plot(x, y):
    p = figure(title = "Iris Morphology", sizing_mode="fixed")
    p.xaxis.axis_label = x
    p.yaxis.axis_label = y
    p.circle(flowers[x], flowers[y], color=colors, fill_alpha=0.2, size=10)
    return p

@app.route('/')
def root():
    return page.render()

@app.route('/plot')
def plot():
    p = make_plot('petal_width', 'petal_length')
    return json.dumps(json_item(p, "myplot"))

@app.route('/plot2')
def plot2():
    p = make_plot('sepal_width', 'sepal_length')
    return json.dumps(json_item(p))

if __name__ == '__main__':
    print("\n\n *** Run 'bokeh static --port 5006` in a separate window ***\n\n")
    app.run())

```

bryevdv:
This just needs a few tests for `json_item` I will wrap things up tomorrow and get merged.

bryevdv:
I have no idea what's going on with the one example test failure on the 2.7 job. It runs fine locally (on 2.7) and it ran fine on Travis before the last commit which can't possibly have any relevance.

mattpap:
In https://s3.amazonaws.com/bokeh-travis/437227313/examples.log:
```
2018-10-04 20:22:34,987 error handling message Message 'PUSH-DOC' (revision 1): RuntimeError("Somehow we didn't detach DataTable(id=u'1598', ...)",)
```

bryevdv:
Right, I've seen that error message (and there has been a different one as well). I have no idea why its happening on travis, but not locally, and no idea why it just started after otherwise unrelated commits.

bryevdv:
kicking the build with new unit tests. Perhaps it's some unexplained issue affecting just that build. Otherwise, the only other thing I can think of is that the rebase on master brought in some problem, except master is passing so that does not make much sense either.

bryevdv:
Still failing, this is the other error that shows up

> /home/travis/build/bokeh/bokeh/bokeh/client/connection.py:190: RuntimeError: Failed to push document: RuntimeError(u"Models must be owned by only a single document, BasicTickFormatter(id='1598', ...) is already in a doc",)

My best current suspicious is that the new simple id's somehow interact poorly with the persistent bokeh server we use to run all the tests on in the CI infrastructure. Not sure why it would so specifically affect this one example and not others, though, or why it did not fail in earlier builds.

mattpap:
This seems to be a sound explanation for those issues. Maybe this will fix the problem with unwanted axes being shown in the layout PR.

bryevdv:
OK switching to UUIDs fixed things. I am a little piqued about this, but:

* it only happens on python 2
* only on one example
* only on TravisCI
* only when using the server in a way that is not advised

so I'm cautiously OK with proceeding with the switch to simple ids

bryevdv:
> This seems to be a sound explanation for those issues. Maybe this will fix the problem with unwanted axes being shown in the layout PR.

It might be, I'd like to re-do our server tests to not use `push_session` at all. It does not represent typical or advised usage. For now, I've updated to use simple ids only on py2.7 server tests.

mattpap:
It looks like some sort of ordering or collision issue which can happen only in one branch of Python. Travis CI usually only makes things worse (I presume due to limited performance), so it often exposes such problems. I've been seeing such failures for some time, but I'm not sure what to attribute them to. I'm not sure if those were present before switching to simple ids. This solution seems fine, especially Python 2 is on the way out. Though having server tests run the same way we run server examples manually, would be a good thing to have.

bryevdv:
well,  now it failed for python 3 after being merged to master....

PR #8202, opened 2018-09-01
Using 'utf-8' codec to open python 3 scripts in bokeh app

### LINKED ISSUE #682: Trouble with encoding on Windows
From SO: http://stackoverflow.com/questions/24075840/python-trouble-with-encoding-on-windows-bokeh-plotting-library

Proposed solution, add "encoding" parameter defaulting to UTF-8.

lgasnier:
Using 'utf-8' codec to open python 3 scripts in bokeh app.

- [x] issues: fixes #682 
- [ ] tests added / passed

### DIFF
--- a/bokeh/application/handlers/script.py
+++ b/bokeh/application/handlers/script.py
@@ -49,6 +49,7 @@ log = logging.getLogger(__name__)
 # Standard library imports
 
 # External imports
+import six
 
 # Bokeh imports
 from .code import CodeHandler
@@ -81,8 +82,12 @@ class ScriptHandler(CodeHandler):
             raise ValueError('Must pass a filename to ScriptHandler')
         filename = kwargs['filename']
 
-        with open(filename, 'r') as f:
-            kwargs['source'] = f.read()
+        if six.PY3:
+            with open(filename, 'r', encoding='utf-8') as f:
+                kwargs['source'] = f.read()
+        else:
+            with open(filename, 'r') as f:
+                kwargs['source'] = f.read()
 
         super(ScriptHandler, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
@lgasnier thanks for the PR. Passing existing tests is an encouraging step. Is it possible to reproduce the original problem on Linux? Then we could potentially add a new CI test to ensure that this capability is thoroughly maintained.

bryevdv:
@lgasnier OK, see, we could add a windows-specific test, but even then, this seems like it will be difficult to test. Can you add an explanatory comment above the `if` block that explains this difference? Otherwise I think there is a risk someone might try to "simplify" it in the future and inadvertently cause a regression. 

As an aside, this does look like a mess:

https://docs.python.org/3/library/locale.html#locale.getpreferredencoding

Ideally, I would also like to include here a link to documentation that specifically explains when and how the locale encoding overrides the `coding` comment line. Do you happen to have a link to anything like that? 

cc @mattpap just in case you have any comments

lgasnier:
@bryevdv After a google search, I found only these links that could explain the PR:

- https://docs.python.org/3.7/library/functions.html#open says that, for `open` statement,  *the default encoding is platform dependent (whatever `locale.getpreferredencoding()` returns)*.
- https://ncoghlan-devs-python-notes.readthedocs.io/en/latest/python3/text_file_processing.html#files-with-a-reliable-encoding-marker explains that `tokenize.open()` can be used to open a file with PEP 263 encoding marker. So, it confirms that, with Python 3, `open` ignores PEP 263 encoding marker.  

Do I add these links to the explanations for the PR (as comment lines)?

bryevdv:
@lgasnier I think in the comment just a link to this PR will suffice, thanks!

bryevdv:
failure is unrelated and will go away upon merge to master

PR #8204, opened 2018-09-03
Add balanced theme

ahuang11:
I'm not sure why it's showing commits from https://github.com/bokeh/bokeh/pull/8062; the only changes should be in __init__.py where I add 'balanced' to the built-in theme and a balanced.json file

(do I have to add an issue and tests for this?)
- [ ] issues: fixes #xxxx
- [ ] tests added / passed

This theme was inspired by plots seen around in https://serialmentor.com/dataviz/time-series.html

Here's the comparison between balanced and light_minimal
image

Theme applied on hvplot documentation: https://anaconda.org/ahuang11/hvplot_balanced_theme/notebook


mattpap:
> I'm not sure why it's showing commits from #8062;

The reason is that (most likely) you started new development on top of your old branch and then updated it with `master`. Most of the time we use squash merges, so original commits do not appear on `master` after a merge. Here you are pulling previous and now unrelated changes once again. The diff is fine, but the commit history already got messy. I would strongly recommend you always start your new development from up-to-date `master` branch, unless you have some continued development going on and/or dependencies on other topic branches. You can use `git rebase` (in particular with `--onto` option) to clean up the commit history.

PR #8206, opened 2018-09-03
Add another theme: balanced

ahuang11:
I accidentally removed all commits (which automatically closed the other PR when I rebased) so here it is again...

(do I have to add an issue and tests for this?)
- [ ] issues: fixes #xxxx
- [ ] tests added / passed

This theme was inspired by plots seen around in https://serialmentor.com/dataviz/time-series.html

Here's the comparison between balanced and light_minimal
image

Theme applied on hvplot documentation: https://anaconda.org/ahuang11/hvplot_balanced_theme/notebook


bryevdv:
@ahuang11 Thanks for the PR. Is there perhaps a different name we could consider? I think theme names should either be:

* descriptive (and "balanced" does not seem to be that, to  me. Balanced against what?) or
* just an arbitrary name, e.g Monokai or Durant

ahuang11:
I was thinking balanced between minimalistic and embellished, but I'm open to name suggestions! Maybe 'just_right'

bryevdv:
> Maybe 'just_right'

Well that's another tick against "balanced" for me, it seems to imply a judgment against other themes.  :)

I notice that the new theme uses Calibri, how about an otherwise meaningless play on that, e.g. "Caliber" or "Calibrese" ?

BTW if you are up for a small amount more work, we could add to the `bokeh.themes` docstring, to inline small plot examples of every theme so that users can more easily compare themes.

ahuang11:
Well, there are only two other themes, both by me, so I'm just judging my previous works :P

Caliber sounds pretty cool.

Adding small different themed plots to the docstrings is a good idea!

ahuang11:
Question @bryevdv would I output a static jpg into some save directory and then somehow read it in within the docstring or do I write code that generates the plot on the go?

bryevdv:
@ahuang11 no need for that, you can use the `:bokeh-plot:` directive to generate plots directly inline in the generated docs:

https://bokeh.pydata.org/en/latest/docs/reference/sphinxext.html#module-bokeh.sphinxext.bokeh_plot

I'd recommend setting the `source-position` to `none` and just making the plot title be the name of the theme.

bryevdv:
@ahuang11  Sorry I should have mentioned, there can only be one call to `show` per directive. So you could: 

* put all three plots in a column, and show that, or
* make three directives (each will need complete code, imports, etc)

Either is fine with me. FYI I don't think this approach scale so well but I think it would be easy to make a new sphinx extension e.g.  `.. bokeh-theme:: default` later, I just don't have time at this moment to create it. We can lok at it as more themes are added

bryevdv:
Thanks @ahuang11 !

PR #8210, opened 2018-09-04
Convert NetworkX node/edge attributes to Bokeh node_renderer/edge_renderer data source. (#8183)

### LINKED ISSUE #8183: [Feature Request] In the function ‘from_networkx’, convert NetworkX node/edge attributes to Bokeh node_renderer/edge_renderer data_source.
In the data structure of NetworkX, nodes and edges can have attributes.
However, node/edge attributes are not converted to Bokeh data source in `from_networkx`.

[In]
```Python
import matplotlib.pyplot as plt
import networkx as nx
# Prepare Data
node_datasource_for_nx = [(1, {'tag': 'feature', 'status': 'open', 'status_color': 'cyan', 'pr': 1}),
                          (2, {'tag': 'feature', 'status': 'close', 'status_color': 'gray'}),
                          (3, {'tag': 'bug', 'status': 'open', 'status_color': 'cyan'})]

edge_datasource_for_nx = [(1, 2, {'association_type': 'duplicated'}),
                          (1, 3, {'association_type': 'impact', 'importance': 5})]


G = nx.Graph()
G.add_nodes_from(node_datasource_for_nx)
G.add_edges_from(edge_datasource_for_nx)

print(G.nodes(data=True))
print(G.edges(data=True))

fig = plt.figure()
pos = nx.spring_layout(G)
nx.draw_networkx_nodes(G, pos=pos, node_color=[x[1]['status_color'] for x in node_datasource_for_nx], alpha=0.5)
nx.draw_networkx_labels(G, pos=pos)
nx.draw_networkx_edges(G, pos=pos)
plt.show()
```

[Out]
prepare_data

[In]
```Python
from bokeh.models.graphs import from_networkx
graph_renderer = from_networkx(G, nx.spring_layout)

graph_renderer.node_renderer.data_source
```

[Out]
node_datasource_no_attribute

[In]
```Python
graph_renderer.edge_renderer.data_source
```

[Out]
edge_datasource_no_attribute

I currently solve this in the following way.

```Python
from bokeh.models.sources import ColumnDataSource, CDSView
# Convert: NetworkX node attributes -> ColumnDataSource
node_dict = dict()
node_dict['index'] = list(G.nodes())

node_attr_keys = [attr_key for node in list(G.nodes(data=True)) for attr_key in node[1].keys() ]
node_attr_keys = list(set(node_attr_keys))

for attr_key in node_attr_keys:
    node_dict[attr_key] = [node_attr[attr_key] if attr_key in node_attr.keys() else None 
                           for node_key, node_attr in list(G.nodes(data=True))]
node_source = ColumnDataSource(node_dict)

# Set node attribute to Bokeh DataSource
graph_renderer.node_renderer.data_source = node_source
graph_renderer.node_renderer.view = CDSView(source=node_source)

# Convert: NetworkX edge attributes -> ColumnDataSource
edge_dict = dict()
edge_dict['start'] = [x[0] for x in G.edges(data=True)]
edge_dict['end'] = [x[1] for x in G.edges(data=True)]

edge_attr_keys = [attr_key for edge in list(G.edges(data=True)) for attr_key in edge[2].keys() ]
edge_attr_keys = list(set(edge_attr_keys))

for attr_key in edge_attr_keys:
    edge_dict[attr_key] = [edge_attr[attr_key] if attr_key in edge_attr.keys() else None 
                           for _, _, edge_attr in list(G.edges(data=True))]
edge_source = ColumnDataSource(edge_dict)
    
# Set edge attribute to Bokeh DataSource
graph_renderer.edge_renderer.data_source = edge_source
graph_renderer.edge_renderer.view = CDSView(source=edge_source)

graph_renderer.node_renderer.data_source
```

[Out]
node_datasource_with_attributes

[In]
```Python
graph_renderer.edge_renderer.data_source
```

[Out]
edge_datasource_with_attributes

However, it is painful to write the above code every time.
I think that it is general case to specify Bokeh's hover information and drawing style by using node/edge attributes  (for example, specifying node colors with node attributes).
It is useful to be able to use the node/edge  attributes of NetworkX also in Bokeh.

[In]
```Python
from bokeh.models import Plot, Range1d, MultiLine, Circle, HoverTool

plot = Plot(plot_width=300, plot_height=300,
            x_range=Range1d(-1.1, 1.1), y_range=Range1d(-1.1, 1.1))

# !!! Specify colors with node attributes !!!
graph_renderer.node_renderer.glyph = Circle(size=15, 
                                            fill_color='status_color') 
graph_renderer.edge_renderer.glyph = MultiLine(line_color="#CCCCCC", 
                                               line_alpha=0.8, 
                                               line_width=1)

# !!! Hover the node attributes !!! 
node_hover = HoverTool(tooltips=[('Issue', '@index'),
                                 ('Tag', '@tag'),
                                 ('Status','@status')],)
plot.add_tools(node_hover)
plot.renderers.append(graph_renderer)

output_notebook()
show(plot)
```

[Out]
node_hover

# Feature request
## Expected result
In `from_networkx`, convert the attributes of nodes/edges as follows.   
1. NetworkX: [Graph.nodes](https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.Graph.nodes.html) data -> Bokeh: [GraphRenderer](https://bokeh.pydata.org/en/stable/docs/reference/models/renderers.html).node_renderer.data_source.data
2. NetworkX: [Graph.edges](https://networkx.github.io/documentation/latest/reference/classes/generated/networkx.Graph.edges.html) edge attribute -> Bokeh: [GraphRenderer](https://bokeh.pydata.org/en/latest/docs/reference/models/renderers.html).edge_renderer.data_source.data

I think that this function can be realized by modifying the code as follows.

**[before]**   
https://github.com/bokeh/bokeh/blob/master/bokeh/models/graphs.py#L35-L77

**[after]**
```Python
def from_networkx(graph, layout_function, **kwargs):
    '''
        ...
    '''

    # inline import to prevent circular imports
    from bokeh.models.renderers import GraphRenderer
    from bokeh.models.graphs import StaticLayoutProvider

    # Handles nx 1.x vs 2.x data structure change
    # Convert node attributes
    node_dict = dict()
    node_dict['index'] = list(graph.nodes())
        
    node_attr_keys = [attr_key for node in list(graph.nodes(data=True)) for attr_key in node[1].keys()]
    node_attr_keys = list(set(node_attr_keys))


    for attr_key in node_attr_keys:
        node_dict[attr_key] = [node_attr[attr_key] if attr_key in node_attr.keys() else None 
                               for node_key, node_attr in list(graph.nodes(data=True))]
        
    # Convert edge attributes
    edge_dict = dict()
    edge_dict['start'] = [x[0] for x in graph.edges(data=True)]
    edge_dict['end'] = [x[1] for x in graph.edges(data=True)]

    edge_attr_keys = [attr_key for edge in list(graph.edges(data=True)) for attr_key in edge[2].keys()]
    edge_attr_keys = list(set(edge_attr_keys))

    for attr_key in edge_attr_keys:
        edge_dict[attr_key] = [edge_attr[attr_key] if attr_key in edge_attr.keys() else None 
                                   for _, _, edge_attr in list(graph.edges(data=True))]

    node_source = ColumnDataSource(data=node_dict)
    edge_source = ColumnDataSource(data=edge_dict)

    graph_renderer = GraphRenderer()
    graph_renderer.node_renderer.data_source.data = node_source.data
    graph_renderer.edge_renderer.data_source.data = edge_source.data

    graph_layout = layout_function(graph, **kwargs)
    graph_renderer.layout_provider = StaticLayoutProvider(graph_layout=graph_layout)

    return graph_renderer
```

**Note:**   
The above code is still incomplete. I think that it is necessary to consider how to handle it in the following case. (raising exceptions, etc.)
1. The case that nodes have a attribute named 'index'
2. The case that edges have a attribute named 'start' or 'end'

komo-fr:
- [x] issues: fixes #8183
- [x] tests added / passed
    - Is it necessary to change the document?

In the code that I pull requested this time, it doesn't correspond to the following case.
I will implement it if necessary.

#8183
> I think that it is necessary to consider how to handle it in the following case. (raising exceptions, etc.)
> 1. The case that nodes have a attribute named 'index'
> 2. The case that edges have a attribute named 'start' or 'end'
>

### DIFF
--- a/bokeh/models/graphs.py
+++ b/bokeh/models/graphs.py
@@ -55,17 +55,37 @@ def from_networkx(graph, layout_function, **kwargs):
         from ..models.graphs import StaticLayoutProvider
 
         # Handles nx 1.x vs 2.x data structure change
-        nodes = list(graph.nodes())
-        edges = list(graph.edges())
+        # Convert node attributes
+        node_dict = dict()
+        node_dict['index'] = list(graph.nodes())
 
-        edges_start = [edge[0] for edge in edges]
-        edges_end = [edge[1] for edge in edges]
+        node_attr_keys = [attr_key for node in list(graph.nodes(data=True))
+                          for attr_key in node[1].keys()]
+        node_attr_keys = list(set(node_attr_keys))
 
-        node_source = ColumnDataSource(data=dict(index=nodes))
-        edge_source = ColumnDataSource(data=dict(
-            start=edges_start,
-            end=edges_end
-        ))
+        for attr_key in node_attr_keys:
+            node_dict[attr_key] = [node_attr[attr_key]
+                                   if attr_key in node_attr.keys() else None
+                                   for node_key, node_attr
+                                   in list(graph.nodes(data=True))]
+
+        # Convert edge attributes
+        edge_dict = dict()
+        edge_dict['start'] = [x[0] for x in graph.edges(data=True)]
+        edge_dict['end'] = [x[1] for x in graph.edges(data=True)]
+
+        edge_attr_keys = [attr_key for edge in list(graph.edges(data=True))
+                          for attr_key in edge[2].keys()]
+        edge_attr_keys = list(set(edge_attr_keys))
+
+        for attr_key in edge_attr_keys:
+            edge_dict[attr_key] = [edge_attr[attr_key]
+                                   if attr_key in edge_attr.keys() else None
+                                   for _, _, edge_attr
+                                   in list(graph.edges(data=True))]
+
+        node_source = ColumnDataSource(data=node_dict)
+        edge_source = ColumnDataSource(data=edge_dict)
 
         graph_renderer = GraphRenderer()
         graph_renderer.node_renderer.data_source.data = node_source.data
--- a/bokeh/models/tests/test_graphs.py
+++ b/bokeh/models/tests/test_graphs.py
@@ -33,3 +33,22 @@ def test_from_networkx_method_with_kwargs():
     gl = renderer.layout_provider.graph_layout
     assert set(gl.keys()) == set([0,1,2,3])
     assert_allclose(gl[0], np.array([2., 0.]), atol=1e-7)
+
+def test_from_networkx_with_attributes():
+    G = nx.Graph()
+    G.add_nodes_from([(0, {"attr_1": "a", "attr_2": 10}),
+                      (1, {"attr_1": "b"}),
+                      (2, {"attr_1": "c", "attr_2": 30})])
+    G.add_edges_from([(0, 1, {'attr_1': 'A'}),
+                      (0, 2, {'attr_1': 'B', 'attr_2': 10})])
+
+    renderer = from_networkx(G, nx.circular_layout)
+
+    assert renderer.node_renderer.data_source.data["index"] == [0, 1, 2]
+    assert renderer.node_renderer.data_source.data["attr_1"] == ["a", "b", "c"]
+    assert renderer.node_renderer.data_source.data["attr_2"] == [10, None, 30]
+
+    assert renderer.edge_renderer.data_source.data["start"] == [0, 0]
+    assert renderer.edge_renderer.data_source.data["end"] == [1, 2]
+    assert renderer.edge_renderer.data_source.data["attr_1"] == ["A", "B"]
+    assert renderer.edge_renderer.data_source.data["attr_2"] == [None, 10]


bokeh/models/graphs.py
```diff
@@ -55,17 +55,37 @@ def from_networkx(graph, layout_function, **kwargs):
...
+
+        node_attr_keys = [attr_key for node in list(graph.nodes(data=True))
+                          for attr_key in node[1].keys()]
+        node_attr_keys = list(set(node_attr_keys))
+
+        for attr_key in node_attr_keys:
+            node_dict[attr_key] = [node_attr[attr_key]
+                                   if attr_key in node_attr.keys() else None
+                                   for node_key, node_attr
+                                   in list(graph.nodes(data=True))]
```
bryevdv: I don't think there is any need for `list` inside the comprehension

bokeh/models/graphs.py
```diff
@@ -55,17 +55,37 @@ def from_networkx(graph, layout_function, **kwargs):
...
+
+        edge_attr_keys = [attr_key for edge in list(graph.edges(data=True))
+                          for attr_key in edge[2].keys()]
+        edge_attr_keys = list(set(edge_attr_keys))
+
+        for attr_key in edge_attr_keys:
+            edge_dict[attr_key] = [edge_attr[attr_key]
+                                   if attr_key in edge_attr.keys() else None
+                                   for _, _, edge_attr
+                                   in list(graph.edges(data=True))]
```
bryevdv: same comment about `list` inside a comprehension

bokeh/models/graphs.py
```diff
@@ -55,17 +55,37 @@ def from_networkx(graph, layout_function, **kwargs):
...
+        node_dict = dict()
+        node_dict['index'] = list(graph.nodes())
+
+        node_attr_keys = [attr_key for node in list(graph.nodes(data=True))
+                          for attr_key in node[1].keys()]
+        node_attr_keys = list(set(node_attr_keys))
+
+        for attr_key in node_attr_keys:
+            node_dict[attr_key] = [node_attr[attr_key]
+                                   if attr_key in node_attr.keys() else None
```
bryevdv: prefer if conditional expressions don't span multiple lines (our linter line limit is 140)

bokeh/models/graphs.py
```diff
@@ -55,17 +55,37 @@ def from_networkx(graph, layout_function, **kwargs):
...
+        edge_dict['start'] = [x[0] for x in graph.edges(data=True)]
+        edge_dict['end'] = [x[1] for x in graph.edges(data=True)]
+
+        edge_attr_keys = [attr_key for edge in list(graph.edges(data=True))
+                          for attr_key in edge[2].keys()]
+        edge_attr_keys = list(set(edge_attr_keys))
+
+        for attr_key in edge_attr_keys:
+            edge_dict[attr_key] = [edge_attr[attr_key]
+                                   if attr_key in edge_attr.keys() else None
```
bryevdv: same comment about making conditional expression be all on one line

bryevdv:
@komo-fr Thanks for the PR, I made a couple of very minor comments. Now with code in front of me I see the two questions to decide on definitively are:

* what to do with attributes labeled `"index"` 
* what to do with missing attributes

You are currently returning `None` for the second case. That's probably fine but I'd like to give it a little thought. 

For the first case, the options seem to be:

* ignore/discard the attribute, with a warning
* relabel in some prescribed way, with or without a warning
* raise an error

I don't really have a strong opinion. What are your thoughts? cc @bokeh/dev in case others have thoughts

birdsarah:
Been doing a lot of networkx plotting lately - very excited about this.

My vote for the index attribute question would be (1) ignore/discard the attribute, with a warning. I like it because:
a) unlikely to cause surprises - what if you have another column called index_attr or whatever we decide etc.
b) warning provides notice, but keeps you moving forward. As opposed to error - which stops you in your track. My experience, given the size of graphs I'm working with, is that the layout step takes a while. If I've got all the way through the networkx layout step and then bokeh errors out (because i'm likely to do these two steps in one method / cell) then that's just going to annoy me. 
c) given that this is a convenience function, i think its okay to err on the side of making decisions for users rather than being ultra explicit.

bryevdv:
I'm fine with (1) especially given we could always later add options for explicit re-mapping if users wanted to respond to the warning.

bryevdv:
@komo-fr please also merge/rebase on master when you have a chance.

komo-fr:
@bryevdv
Thank you for reviewing.
I fixed the point you pointed out.
In addition, I unified the double quotes and the single quotes in the test code I wrote.

@bryevdv  @birdsarah
Thank you for your opinion.
I have not come up with a lot of ideas, so it was very helpful.
I agree with (1).

bryevdv:
@komo-fr it sounds like we all agree (1) is a good way to go. Can you implement that?  Ideally we would test that the warning is emitted as well, there are examples of testing warnings in other tests in the repo, I can point them out if necessary. 

Otherwise, my only last comment is that this block: 
```
for attr_key in node_attr_keys:
            node_dict[attr_key] = [attr[attr_key] if attr_key in attr.keys() else None for _, attr in graph.nodes(data=True)]
``` is basically repeated verbatim for both nodes and edges. Perhaps factor it out to a helper function `_rows_to_columns` or something? That could get its own small test as well. 

This will be a great new feature, thanks for working on this!

**Comment by komo-fr** [comment]
@bryevdv 
Thank you, I will try them this weekend!  
I will ask here if there is anything I do not know.

- [ ] implement (1) (ignore/discard the attribute “index”/”start”/“end”, with a warning)
- [ ] add test for (1)
- [ ] factor it out to a helper function (code converting node/edge attributes to dict)
- [ ] add test for the helper function
- [ ] merge/rebase

**Comment by komo-fr** [comment]
@bryevdv
I fixed and pushed ... but the result of Travis CI told me that my code fails on testing in Python 2.7 environment.

> - [ ] factor it out to a helper function (code converting node/edge attributes to dict)

I tried to solve the difference between `graph.nodes(data=True) `and `graph.edges(data=True)` with `* _`, but this syntax seems to have been added in Python 3.0 ([PEP 3132](https://www.python.org/dev/peps/pep-3132/)).
I'm looking for another solution, but I'm not familiar with the syntax of Python 2.7, so I would be happy if you advice me if there is any other good way.

**Comment by birdsarah** [comment]
So the problem code is this:

```python
    attr_dict = {}
    for attr_key in attr_keys:
        attr_dict[attr_key] = [attr[attr_key] if attr_key in attr.keys() else None for *_, attr in source]
    return attr_dict
```

To make this py2/3 compatible, the trick to be a little more explicit/verbose - which I think will make it clearer. I did this here:
* py2 - http://nbviewer.jupyter.org/github/birdsarah/bokeh-miscellany/blob/master/rows%20to%20columns%20-%20py2_3%20-%20py2.ipynb
* py3 - http://nbviewer.jupyter.org/github/birdsarah/bokeh-miscellany/blob/master/rows%20to%20columns%20-%20py2_3%20-%20py3.ipynb

HOWEVER. I would not do this. To my mind the above code makes it clear you're dealing with two different APIs / inputs.

Looking back at your code, I would revert the following commit. 

https://github.com/bokeh/bokeh/pull/8210/commits/ae825402a6eea1d358fe2c57bf28b7b632a2f43f

The code that you "one-lined" in that commit was clearer I think.

I believe that @bryevdv may have missed the extra `,_` once it was one-lined and though it was duplicate code and that's what prompted the refactor suggestion (just a guess, apologies if I'm wrong @bryevdv).  

The code you removed in the above commit makes it clear you're dealing with first node, then edge, data and to my eyes at least the API difference pops out.

bryevdv:
I did miss the extra `_`, apologies for the noise, please disregard my suggestion.

birdsarah:
For what it's worth @komo-fr, I generally try and un-one-line code. 

Folks who know me, will know I'm verbose :D, and this is a lot of personal taste.  But, for me, not one-line is often clearer and easier to parse for others coming in to reading code - because it ends up being really step-by-step - which is helpful when looking at code for the first time / quickly.

birdsarah:
On the plus side, I had totally forgot about python 3's *_ extended unpacking syntax, so it was awesome to be reminded of that.

komo-fr:
@bryevdv @birdsarah
Thank you for taking your time to advice me.

I may not understand your comments correctly. 
Therefore, I would like to make sure of a few points to proceed the next step without confusion.

**1. 423dd36 Convert node/edge attributes to dict in common code**
OK, I will revert this commit because @bryevdv said

> I did miss the extra _, apologies for the noise, please disregard my suggestion.

**2. ae82540 Make conditional expression be all on one line and change the variable name to fit within 140.**
I don’t understand well how to deal with this commit.

I fixed it for the review comments of @bryevdv below.
HOWEVER, I looked back this comment and noticed that I have misunderstood. I apologize. 

> prefer if conditional expressions don't span multiple lines (our linter line limit is 140)

The scope of this comment was only  _"conditional expression"_

```Python
node_dict[attr_key] = [node_attr[attr_key]
　　　　　　　　　　      if attr_key in node_attr.keys() else None
```

However, I misunderstood that it is the whole _“List comprehension”_ .

```Python
node_dict[attr_key] = [node_attr[attr_key]
　　　　　　　　　　      if attr_key in node_attr.keys() else None
　　　　　　　　　　      for node_key, node_attr
　　　　　　　　　　      in list(graph.nodes(data=True))]
```

Therefore, I have changed the variable names in order to make the whole less than 140. (etc, node_attr -> attr)
I guess that it became difficult to understand that nodes and edges had different APIs.

Now, I'd like to reflect the review comments from @bryevdv and @birdsarah without misunderstanding.
**What should I do next?**

**(A) Revert only 423dd36, but keep ae82540  as is.** 

**(B) Also revert ae82540,  fix it by other ways, and commit 0e82929 ~ 6600ef8 again.**
There are several “other ways” as written in (C) below.

**(C) Revert only 423dd36, keep ae82540  as is, and fix by other ways as new commit.**
　　(I seem that this is easier than (B))
There are several “other ways” as below.
I own like (i), but I would like to follow it if bokeh has a policy in such a case.

(i) Make it explicit that **_they are unused variable_** using **_a single underscore_** `_`.
```Python
node_dict[attr_key] = [node_attr[attr_key] if attr_key in node_attr.keys() else None
　　　　　　　　　　      for _, node_attr
　　　　　　　　　　      in graph.nodes(data=True)]
```

```Python
edge_dict[attr_key] = [edge_attr[attr_key] if attr_key in edge_attr.keys() else None
　　　　　　　　　　      for _, _, edge_attr
　　　　　　　　　　      in graph.edges(data=True)]
```

(ii) Make it explicit that **_node and edge have different APIs_** by using **_descriptive names_** even if the are unused variables, 
```Python
node_dict[attr_key] = [node_attr[attr_key] if attr_key in node_attr.keys() else None
　　　　　　　　　　      for node_key, node_attr
　　　　　　　　　　      in graph.nodes(data=True)]
```

```Python
edge_dict[attr_key] = [edge_attr[attr_key] if attr_key in edge_attr.keys() else None
　　　　　　　　　　      for start_node_key, end_node_key, edge_attr
　　　　　　　　　　      in graph.edges(data=True)]
```

(iii) The others

bryevdv:
@komo-fr I think C (i) is advised. If variables are unused they need to be underscores (or start with underscores) other wise the linter will complain.

komo-fr:
@bryevdv
Thank you for reply.
I reverted and fixed it with C(i)

bokeh/models/graphs.py
```diff
@@ -48,24 +48,56 @@ def from_networkx(graph, layout_function, **kwargs):
...
+            edge_dict[attr_key] = [edge_attr[attr_key] if attr_key in edge_attr.keys() else None
+                                   for _, _, edge_attr
+                                   in graph.edges(data=True)]
+
+        if 'start' in edge_attr_keys or 'end' in edge_attr_keys:
+            from warnings import warn
+            warn("Converting edge attributes labeled 'start' or 'end' are skipped. "
+                 "If you want to convert these attributes, please re-label them with other names.")
+
+        edge_dict['start'] = [x[0] for x in graph.edges(data=True)]
```
birdsarah: I do not believe you need `data=True` here. I don't think this matters particularly as it should be performant either way, but I would expect without data to be marginally faster.

bokeh/models/graphs.py
```diff
@@ -48,24 +48,56 @@ def from_networkx(graph, layout_function, **kwargs):
...
+        edge_dict['end'] = [x[1] for x in graph.edges(data=True)]
+
+        node_source = ColumnDataSource(data=node_dict)
+        edge_source = ColumnDataSource(data=edge_dict)
```
birdsarah: This is all soo pretty. Great job. I think it reads really nicely.

birdsarah:
I haven't tested this, but the code looks great.

birdsarah:
@bryevdv, was watching CI and noticed that while py27 passed, it ended very strangely - https://travis-ci.org/bokeh/bokeh/jobs/428423175#L1727-L1745 (this may be true on other runs, that just happened to be the one i was looking at). Also guessing that the test failure is due to lack of permissions?

bokeh/models/graphs.py
komo-fr: Thank you for pointing it out.
I fixed it and confirmed that there is no problem with both NetworkX 1 and 2.

komo-fr:
@bryevdv @birdsarah
I fixed and pushed it again.
The following error occurred in the previous push, but it seems that it doesn’t occur now.
(I have not changed anything other than code🤔）

> [FAIL] Upload was requested but could not connect to S3.
> [FAIL] This is expected if you are an external contributor submitting a PR to Bokeh.
> [FAIL] This could also happen if S3 credentials are not available on the machine where this test is running.

bryevdv:
@birdsarah @komo-fr those  failures are purely due to the PR coming from a fork, and are expected. TravisCI does not supply encrypted credentials except to builds for PR branches off the main repo. It just means the image diff report is not saved, and the build also takes a bit longer due to having to build BokehJS from scratch on every job.

komo-fr:
@bryevdv @birdsarah
Thank you very much for your kind review!
This PR became a very good experience for me. 🎉

@bryevdv 
>  If you have time and are inclined, another small PR with an example (or updating an existing example) to demonstrate this would be useful

OK, I will send a PR again when I have time.
(Recently I've been using `from_networkx` a lot)

birdsarah:
Brilliant. Just tested it and was able to delete this line from my code.

```python
graph_renderer.edge_renderer.data_source.data['p_id'] = [edge[2]['p_id'] for edge in G.edges(data=True)]
```
Hurrah!

This line took a while to work out, and felt very fragile. This PR is great! Thanks @komo-fr.

PR #8211, opened 2018-09-04
8196 Add .copy method to ColumnDataSource

guzwaatensen:
Made the necessary changes to implement my own feature request [8196].
Added `__copy__` and `__deepcopy__` hooks to `PropertyValueColumnData` and changed `ColumnDataSource` to make a shallow copy by default as requested by @bryevdv.

I noticed that if the CDS is initialized from a DataFrame `index` is now the first key instead of the last, but I assume this is the result of other changes to the repo?


bokeh/models/sources.py
```diff
@@ -58,6 +59,11 @@ class ColumnDataSource(ColumnarDataSource):
 
           source = ColumnDataSource(data)
 
+    .. note::
+        ``ColumnDataSource`` creates a shallow copy of the passed ``dict``. Use
```
bryevdv: Sphinx is very finicky, this will fail the docs build, it just needs a blank line after the `.. note::`

bokeh/core/property/containers.py
```diff
@@ -318,6 +319,12 @@ class PropertyValueColumnData(PropertyValueDict):
     def __setitem__(self, i, y):
         return self.update([(i, y)])
 
+    def __copy__(self):
+        return PropertyValueColumnData(copy.copy(dict(self)))
```
bryevdv: I don't think the `copy.copy` is needed here, just calling the `dict` initializer creates a shallow copy

guzwaatensen: I wasn’t sure if the result is exactly the same, but I’m hindsight you are likely right. With larger dicts it might make a difference in performance, I’ll remove the call to `copy.copy`.

bokeh/models/sources.py
guzwaatensen: At the bottom of the very same docstring there is:
```
.. note::
        There is an implicit assumption that all the columns in a given
        ``ColumnDataSource`` all have the same length at all times. For this
        reason, it is usually preferable to update the ``.data`` property
        of a data source "all at once".
```
So I'm assuming it would work, but I will make the change as sphinx' documentation clearly states that a blank line after the directive is the intended use.

bryevdv: Oh, that's a mistake then. Can you fix that one as well, as long as we are editing this docstring?

bryevdv: Perhaps this has changed, though? I know for certain it used to be required but it seems to be rendering correctly, and it's not causing any warnings (which are considered errors). Still, I'd prefer if they were made consistent with everywhere else in the codebase.

bryevdv:
@guzwaatensen Can you also add a few tests methods here:

https://github.com/bokeh/bokeh/blob/master/bokeh/models/tests/test_sources.py
https://github.com/bokeh/bokeh/blob/master/bokeh/core/property/tests/test_containers.py

to exercise this new code?

bokeh/models/sources.py
```diff
@@ -132,6 +139,8 @@ def __init__(self, *args, **kw):
                 raw_data = self._data_from_groupby(raw_data)
             else:
                 raise ValueError("expected a dict or pandas.DataFrame, got %s" % raw_data)
+        else:
+            raw_data = copy.copy(raw_data)
```
bryevdv: This makes copies of plain `dict` as well, which may slightly change the semantics (and is not really necessary). I wonder if it would be better to  check specifically for `PropertyValueColumnData` instances, and only call `copy` in that case?

bryevdv:
@guzwaatensen I left one more comment and request for a couple of tests. Thanks for iterating on this, I think it's nice addition.

bokeh/models/sources.py
guzwaatensen: I though that this would at least be consistent as it just makes copies of any ``dict`` subclass passed.
I can implement the typecheck of course, but maybe I could also just remove these two lines altogether, as the functionality I requested is covered by the ``__deepcopy__`` hook and in my opinion making a shallow copy only has limited use cases anyway. Then I would slightly change the docstring ``.. note::`` wording to inform users about the need to invoke ``deepcopy()`` on the data object should they require an independent copy.

I'm fine either way, though this is all the time I have today, so I will implement the tests and any other changes another time.

guzwaatensen:
I've removed the conditional test from `ColumnDataSoure.__init__` as I realized during creation of the corresponding test that the initializer calls `dict.update(data)` which essentially already creates a shallow copy.

So that leaves the changes to the docstring and the hooks for `PropertyValueColumnData`. I've added corresponding tests and I left the test for making sure there is actually a shallow copy being made when initializing a `ColumnDataSource` with a `PropertyValueColumnData` in. If you don't think it's necessary I can remove it again. Not sure if the tests are up to your standards, I rarely do test driven development.

bryevdv:
@guzwaatensen looks great, will merge when green

PR #8214, opened 2018-09-05
Bryanv/misc cleanup

bryevdv:
This is just some misc. code cleanup/gardening. It brings a number of modules up to 100% coverage (some important features had no coverage), as well as refactors some duplicate code in places, and simplifies some dynamic docstrings to static ones.


bryevdv:
cc @HuntJSparra FYI in case this small PR intersects any files you are currently working on.

Issue #2603, opened 2015-07-23
Add a SizeBar guide to label relative scatter glyph sizes


hafen:
When size is being varied in a plot according to some variable in the data, it is useful to make a legend that ties the size of the glyph to some numeric value.  Currently size in a legend is fixed.  Here's an example:

![rplot](https://cloud.githubusercontent.com/assets/1275592/8843362/95990f70-30d5-11e5-9862-6afb151a3dfc.png)

In this case in bokeh, I would just create some dummy glyphs with varying sizes and use those for the legend.

bryevdv:
@birdsarah do you have any input on the questions above?

birdsarah:
ooohhhh....I feel like a kid in a candy store. Very excited about getting this implemented.

For this particular use case, I think:
* the main thing you want is for the legend size to definitely be the matching size from the plot
* you may want to give that label a different name e.g. you want to pick out the 1, 10, 100 size bubbles for the legend, but you want the legend labels to be something like "1 (no risk)", "100 (certain death)"

Am I understanding the second option correctly:
* `index` parameter - pick the glyph representation for the rendered glyph at that index
* `use_glyph_size` - would normally default to false, and when you request true it becomes the same size as the rendered glyph

If yes, then I kind of like option 2. 

It would be nice to have a syntactic sugar version where, just like on an axis, you ask it to intelligently 
pull 3, or 5 round number values.

I would consider a name for `use_glyph_size` something in the other direction e.g. `normalize_size` and it would typically be True and then all glyphs in the legend end up the same size. And then when you manually set it to False it pulls in the rendered glyph size.

My main lingering concern is that I think a regular occurrence is where your data is: [2, 54, 22, 89, 105, .....] But you want the legend to show 1, 10, 100 - nice, clean round numbers that represent the range of your data, but don't actually appear in your data. If we're relying on the exact value we want to represent being in the index it's going to be cumbersome.

ganiatgithub:
Hi,

I seem to understand partly from @bryevdv 's post. I think the following is what I want.

```
legend = Legend(items=[
    LegendItem(label="10", renderers=[r], index=0, use_glyph_size=True),
    LegendItem(label="20", renderers=[r], index=1, use_glyph_size=True),
    LegendItem(label="30", renderers=[r], index=2, use_glyph_size=True),
]) 
```

My code is:
```
import pandas as pd
import numpy as np
from bokeh.palettes import Blues
[...]

PR #8218, opened 2018-09-06
support specifying representative point from legend items

bryevdv:

This was raised on Gitter and falls under the category of low hanging but valuable fruit. This PR allows users to specify an explicit `index` on legend items, so that e.g. annotating each line in a multi_line is possible. With a little more follow-on work, #2603 would be within reach. Later work might also add a convenience API to `multi_line` and `patches` to make creating the legend items less manual. 

Unfortunately I can't think of a good direct test for this with the current view architecture, but I have added an expample that will show up under image diff tests. 

```
from bokeh.models import Legend, LegendItem

p = figure()
r = p.multi_line([[1,2,3], [1,2,3]], [[1,3,2], [3,4,3]],
                 color=["orange", "red"], line_width=4)

legend = Legend(items=[
    LegendItem(label="orange", renderers=[r], index=0),
    LegendItem(label="red", renderers=[r], index=1),
])
p.add_layout(legend)

output_file("multi_legend.html")

show(p)
```

screen shot 2018-09-06 at 11 41 10

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -81,6 +81,18 @@ class LegendItem(Model):
     then all data_sources of renderers must be the same.
     """)
 
+    index = Int(default=None, help="""
+    The column data index to use for drawing the representative items.
+
+    If None (the default), then Bokeh will automatically choose an index to
+    use. If the label does not refer to a data column name, this is typically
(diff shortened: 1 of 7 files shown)


bryevdv:
This should be uncontroversial, I am going to go ahead and merge.

PR #8219, opened 2018-09-06
don't report unversioned python in deps.py output

bryevdv:
Build is currently failing due to an apparent python 3.7 issue, which has uncovered that the current method of installation will always use the most recent python, instead of what is explicitly specified and intended. 

The issue is having an unversioned "python" in the conda install line:

screen shot 2018-09-06 at 14 05 33

This PR addresses that by removing "python" if it is present (it has to remain in the recipe, so our only option is to filter it out)

### DIFF
--- a/scripts/deps.py
+++ b/scripts/deps.py
@@ -27,6 +27,9 @@ spec = []
 for name in sys.argv[1:]:
     spec += section[name]
 
+# bare python unpins python version causing upgrade to latest
+if 'python' in spec: spec.remove('python')
+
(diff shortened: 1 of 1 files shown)


bryevdv:
Further digging suggests this problem only affects "outside" PRs. The unintended upgrade happened when installing `BUILD_DEPS` and that only happens for PRs from forks.

bryevdv:
Actually, it seemed to happen on main repo PRs too (master just failed)

Issue #8207, opened 2018-09-03
Adding/Updating Boilerplate Code


HuntJSparra:
As part of the NumFocus small grant, we are funded to "Update all Bokeh source files with common boilerplate and header format (started but not yet finished)". The work should cover the following directories:

- [x] Top Level (#8220)
- [x] _testing
- [x] application (#8245)
- [x] client (#8276)
- [x] colors (#8293)
- [x] command (#8313)
- [x] core (#8341)
- [x] document
- [x] embed
- [x] io
- [x] models
- [x] plotting
- [x] protocol
- [x] sampledata
- [x] server
- [x] sphinxext
- [x] tests
- [x] themes
- [x] util

Non-directory:
- [x] Copyright
- [x] Update __all__ for transitive imports

Note that test module boilerplate varies from the standard.

bryevdv:
BTW as long as we are doing this work we should also update the copyright statement. Since there has never been any CLA or copyright assignment agreement put in place it should now properly read:

> Copyright (c) 2012 - 2018, Anaconda, Inc., and Bokeh Contributors. All rights reserved.

I'd say let's make one PR at the end that just does this with some automated search/replace script in one swoop.

HuntJSparra:
Do you want the "boxes" extended or the ``All rights reserved.`` moved down a line?

It currently looks a bit odd (in at least my opinion) as:
```
#-----------------------------------------------------------------------------
# Copyright (c) 2012 - 2018, Anaconda, Inc., and Bokeh Contributors. All rights reserved.
#
# Powered by the Bokeh Development Team.
#
# The full license is in the file LICENSE.txt, distributed with this software.
#-----------------------------------------------------------------------------
```

bryevdv:
If we are going to add a line, let's subtract others:

```
#-----------------------------------------------------------------------------
# Copyright (c) 2012 - 2018, Anaconda, Inc., and Bokeh Contributors. 
# All rights reserved.
#
# The full license is in the file LICENSE.txt, distributed with this software.
#-----------------------------------------------------------------------------
```
Definitely wait to do this until the end, though, so this can be automated

jbednar:
"Powered by the Bokeh Development Team"?

I'm not sure what this line adds...

bryevdv:
@HuntJSparra I am pretty sure `_testing` is already done, but let me know when you think this is complete and can be closed.

HuntJSparra:
[...]

PR #8220, opened 2018-09-07
Boilerplate for toplevel bokek/bokeh

HuntJSparra:
- [ ] issues: adresses part of #8207 
- [ ] tests added / passed

Adds the boilerplate code for top level files in bokeh/bokeh.


HuntJSparra:
@bryevdv 
Some code placement questions:

1) (events.py) class `_MetaEvent` should be in the Private API, but must be before `Event`. Do you want it left under General API or, at the very least, have both moved to Dev API? Do you have a different solution?

2) (_version.py) I am likely missing something, but the variables `LONG_VERSION_PY` and `HANDLERS` seem to go unused after initialization. Should this be moved from being sandwiched between a class and various functions under Dev API to Code?

3) (model.py) Similar to (1), the constant '_HTML_REPR' must be used before Private API. Would you suggest it be moved to Constants, or be left before the class that uses it?

4) (settings.py) 2 things:
    4.1) There is a block of comments at the bottom that describes the Settings object, but is not a docstring. Do you have any preference on where this goes? Maybe at the top as a "A note to developers"?

    4.2) `logging` is imported to set log levels (via `Settings`) in this file. Do you want the import under the Boilerplate section, or moved to the imports section?


Thanks!

bryevdv:
Hi @HuntJSparra great start, a few comments:

1) Right, sometimes language constraints conflict with our goals. But not too often, so it's generally fine to make exceptions in those cases. That said, in this case, I think moving both under "dev" API might make sense (And rename to `MetaEvent` with no underscore). Users would not use `Event` directly unless they are making custom extension events, which is not common. 

2) `_version.py` is an auto-generated file from versioneer. I should have mentioned explicity, but we should just leave it exactly as it is

3) I think in cases like this I am fine leaving a "private" global constant that is closely associated with a particular class or function, right before the class or function. 

4.1) Just FYI those `#:` comments  are ones that are picked up by our docs build so that they get rendered in the docs. However, in this case I don't see why all of that couldn't just go in the docstring at the top, instead of these comments, so I would suggest doing that.

4.2) let's leave it under the standard boilerplate location. That it happens to get used in more ways than in other modules is not really relevant in my mind

bryevdv:
> MetaEvent and Event are going to have to be before the other events due to inheritance if I am not mistaken. Do you want to just leave everything under General API?

Sure, that's fine. Another solution would be to make `events` a module (directory) instead of a single package, but that seems like overkill. 

> Also, from __future__ import unicode_literals had to be left out of model.py due to issues it caused in a Python 2.7 example.

Can you elaborate?

bryevdv:
> For some reason, my adjusting of the code had caused self.instance_type to become unicode type (instead of a bokeh.models.CustomJS object)

That's incredibly weird and unexpected. It's possible that some other problem is being masked, but I would like to investigate and get to the bottom of it because adding `unicode_literals` should definitely be fine. I will poke at this in the next day or two, I think it would be OK for you to move on to the next chunk you want to look at in the mean time.

bryevdv:
@HuntJSparra This was in fact pointing to a bug somewhere else. I have made https://github.com/bokeh/bokeh/pull/8230 to fix it. Once that is merged, you will have to merge it into this PR branch (there may be conficts to resolve)

bryevdv:
@HuntJSparra the other PR is merged, I added `unicode_literals`, it's possible a `git merge master` will merge cleanly, but if not hopefully the conflicy is straightforward to resolve. Let me know any questions.

bryevdv:
@HuntJSparra OK I think this is just about ready. After a little more consideration, my last comments are:

* Keeping in line with 

    >  think in cases like this I am fine leaving a "private" global constant that is closely associated with a particular class or function, right before the class or function.

   I think we should leave `_MetaEvent` and `_CONCRETE_EVENT_CLASSES` as private, and just keep them above `Event` Users should never touch these and we should make that clear with the names

* I think we could make the `bokeh.settings` class definition private, i.e. `_Settings` the user facing API is the actual settings instance that is created at the end. 

* I'd like to try and clean up/alphabetize imports as part of this work as well. So in `bokeh.model`, this:

    ```
    from .core.json_encoder import serialize_json
    from .core.properties import Any, Dict, Instance, List, String
    from .core.has_props import HasProps, MetaHasProps
    from .core.query import find
    from .themes import default as default_theme
    from .util.callback_manager import PropertyCallbackManager, EventCallbackManager
    from .util.future import with_metaclass
    from .events import Event
    ```
     could become 
    ```
    from .core.json_encoder import serialize_json
    from .core.properties import Any, Dict, Instance, List, String
    from .core.has_props import HasProps, MetaHasProps
    from .core.query import find
    from .events import Event
    from .themes import default as default_theme
    from .util.callback_manager import EventCallbackManager, PropertyCallbackManager
    from .util.future import with_metaclass
    ```

bryevdv:
@HuntJSparra OK awesome, two last comments (I promise) and then I think things are sufficently calibrated to merge this and move on!

* the contents of `__all__` should be put in alphabetic order

* in `model.py` the `get_class` and `collect_models` funcs are super low level, they should go in "dev" API.

HuntJSparra:
ping @bryevdv

PR #8226, opened 2018-09-07
Ci windows appveyor pr

### LINKED ISSUE #8225: Appveyor build and test for windows
Add continous integration for windows using appveyor

xavArtley:
- [ ] issues: fixes #8225

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,5 +22,5 @@ install:
   - cmd: conda install jinja2 pyyaml
   - ps: $BUILD_DEPS = python scripts/deps.py build run test
   - ps: $BUILD_DEPS
-  - cmd: conda install --yes $BUILD_DEPS
+  - ps: conda install --yes $BUILD_DEPS
   
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,5 +21,6 @@ install:
   - cmd: conda update -q conda
   - cmd: conda install jinja2 pyyaml
   - ps: $BUILD_DEPS = python scripts/deps.py build run test
-  - cmd: echo %BUILD_DEPS%
+  - ps: $BUILD_DEPS
+  - cmd: conda install --yes $BUILD_DEPS
   
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,25 @@
+build: false
+
+platform:
+  - x64
+
+environment:
+  matrix:
+    # Python 3.6
+    - MINICONDA: C:\Miniconda36-x64
+
+init:
+  - cmd: set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%MINICONDA%\\Library\\bin;%PATH%
+  - cmd: echo %path%
+  
+install:
+  - cmd: conda config --set auto_update_conda off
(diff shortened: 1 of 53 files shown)


scripts/deps.py
```diff
@@ -31,9 +31,17 @@ def _setup(**kw): data.update(kw)
 if 'python' in spec: spec.remove('python')
 
 deps = ""
-deps += " ".join(s for s in spec)
+if sys.platform == "win32":
+    deps += '"'
+    deps += '" "'.join(s for s in spec)
+    deps += '"'
+else:
+    deps += '" "'.join(s for s in spec)
```
bryevdv: this doesn't seem correct in the non-windows case. It looks like it would produce:
```
foo" "bar" "baz
```
with no quotes at the beginning. I think it needs to be `deps += " ".join(...)` as before. 

However, is it possible to a avoid changing this script at all, and put the necessary logic to handle this output in the windows scripts? AFAIK this works on windows with the original script:
```
for /F "delims=" %i in ('python scripts\deps.py build run') do (conda install %i)
```

bryevdv:
@xavArtley is it possible to rebase on lastest master? I think this is branched from an older commit and that is why there is exrta cruft in the commit log

scripts/deps.py
bryevdv: Tho perhaps that windows command is much slower (I think it does a separate install for each pkg?)

.appveyor.yml
```diff
@@ -0,0 +1,23 @@
+build: false
+
+platform:
+  - x64
+
+environment:
+  matrix:
+    # Python 3.6
```
bryevdv: I think the latest or possibly upcoming miniconda is 3.7 I would say leave the comment out since it will soon be outdated if it is not already

README.md
```diff
@@ -88,6 +88,15 @@ Bokeh
...
 </tr>
+<tr>
+	<td>AppVeyor</td>
+	<td>
+	<a href="https://ci.appveyor.com/project/xavArtley/bokeh">
+		<img src=https://ci.appveyor.com/api/projects/status/github/xavArtley/bokeh
+			alt="Current AppVeyor build status" />
+	</a>
+	</td>
+</tr>
```
bryevdv: We will need to make a bokeh project account and have the link here refer to that. For now can you just leave this out? I need to re-edit the README anyway, I will re-organize that table to accommodate this badge at that point.

bryevdv: OK we already have a bokeh project on Appveyor, linked under the "bokeh-integrations" GitHub account. If you let me know any specific configuration necessary, I can make any changes.

scripts/deps.py
xavArtley: On windows i never succeed to run the command of the documentation directly because of node <>= ineed to use quotes. With powershell or dos

README.md
xavArtley: I just sign in on appveyor with my github account and add the repo to projects. Now each time a modification is done its trigger a build

xavArtley:
I close this PR, I made a cleaner branch

bryevdv:
OK, look forward to it, let me know if I can help with anything

PR #8238, opened 2018-09-12
add clear method to CDS

### LINKED ISSUE #7990: Add clear tool to clear out one or more datasources
The draw and edit tools are very useful to annotate a plot, however often you want to be able to both add and delete these annotations without having to select each drawn glyph and deleting it manually. It would therefore be useful to add a clear tool which can be given one or more data sources to clear. The implementation would be quite straightforward since the tool would simply have to iterate over the supplied CDS columns and set each to an empty array.

bryevdv:
This PR adds a `clear` method to BokehJS CDS. Additionally:

* removes duplicative `column_names` property (use `columns()` consistently)
* fixes a bug preventing direct init of CDS from typed arrays

- [x] issues: fixes #7990
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/lib/api/plotting.ts
+++ b/bokehjs/src/lib/api/plotting.ts
@@ -501,8 +501,8 @@ export class Figure extends Plot {
     if (legend != null) {
       if (isString(legend)) {
         legend_item_label = { value: legend }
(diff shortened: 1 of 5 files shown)


bryevdv:
Seems uncontroversial and has tests, merging.

PR #8245, opened 2018-09-13
Boilerplate for bokeh/bokeh/application

HuntJSparra:
- [ ] issues: adresses part of #8207 
- [ ] tests added / passed
Adds the boilerplate code for files in bokeh/bokeh/application.

Note:
The majority of the boilerplate in bokeh/bokeh/application was added to master prior to this PR


HuntJSparra:
@bryevdv 
I moved all the handlers under Dev API since they did not seem like things the average user would interact with. Does that make sense? Should they be moved back into General API since anyone poking around there would be wanting to access those anyhow?

Also, 'ServerContext' and 'SessionContext' in 'application.py' are labeled as not being of interest to the general user. Do you want those left under Dev API, or moved under Private API with the necessary changes being made to related files?

Finally, the boilerplate for test scrips in the directory is slightly different ('pytest' imported instead of 'logger' and Setup replacing Globals and Constants).

bokeh/application/tests/test___init__.py
```diff
@@ -40,3 +40,15 @@
...
+# Private API
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Dev API
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Code
+#-----------------------------------------------------------------------------
```
bryevdv: Private/Dev sections out of order?

bryevdv:
@HuntJSparra Looks good, I only had one small comment

> Also, 'ServerContext' and 'SessionContext' in 'application.py' are labeled as not being of interest to the general user. Do you want those left under Dev API, or moved under Private API with the necessary changes being made to related files?

Leave under "Dev"

> Finally, the boilerplate for test scrips in the directory is slightly different ('pytest' imported instead of 'logger' and Setup replacing Globals and Constants).

Yes that sounds correct

bokeh/_version.py
```diff
@@ -524,3 +557,11 @@ def get_versions():
...
             "dirty": None,
             "error": "unable to compute version", "date": None}
+
+#-----------------------------------------------------------------------------
+# Private API
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Code
+#-----------------------------------------------------------------------------
```
bryevdv: `_version.py` should be left as-si (it is an auto-generated vendored file)

HuntJSparra:
Ping @bryevdv

bryevdv:
@HuntJSparra I left a couple of comments above

bryevdv:
@HuntJSparra there were still some changes showing up in `_version.py`, in the interest of merging this today, I just copied the current `_version.py` explicitly from master, then copied it back into the PR branch. Will merge when green!

PR #8251, opened 2018-09-17
update requests min version

bryevdv:
@mattpap this appears to resolve the CVE noted by GH regarding the `cryptiles` package. By updating `requests` we no longer pull in `cryptiles` at all. Is this the best way? I guess my only question is about adding it to "dependencies" since AFAIK it is a second order dependency.


mattpap:
It isn't the best way, but it doesn't do harm either. Preferably we should upgrade jsdom and less, but upgrading jsdom may be a bigger task (I recall having issues with an upgrade attempt previously), and it's prohibitive until layout PR is merged. In longer term I would drop both dependencies, because they are currently the biggest contributors of pointless dependencies. I'm fine with merging this, but I will keep it on my radar to get those updated when feasible.

PR #8255, opened 2018-09-19
Adding autoreload based on tornado

### LINKED ISSUE #4236: Implement --develop mode
Also document what `--develop` does.

`tornado` does not seem to do a good job in tracking dependencies all the way to the bokeh app, and does not restart when the bokeh app source changes.

PS Why was `tornado` chosen as a web server in the first place? E.g. `flask` does a brilliant job at serving simple WSGI apps, doesn't to require the use of non-blocking event-loop paradigms, whereas `tornado` can be shown to becomes unstable when these paradigms are violated.

jensschroer:
- [ ] issues: fixes #4236
- [ ] tests added / passed

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -309,7 +309,7 @@ from bokeh.util.string import nice_join, format_docstring
 from bokeh.server.tornado import DEFAULT_WEBSOCKET_MAX_MESSAGE_SIZE_BYTES
 from bokeh.settings import settings
 
-import tornado.autoreload as tar
+from tornado.autoreload import watch
 import os
 from fnmatch import fnmatch
 
@@ -485,11 +485,17 @@ class Serve(Subcommand):
         )),
 
         ('--dev', dict(
-            action='store_true',
-            help="Enable live reloading during app development."
-                 "NOTE: This setting has effect ONLY for Tornado>=5.0"
-                 "and only works with a single app."
-                 "It also restricts the number of processes to 1.",
+            metavar ='FILES-TO-WATCH',
+            action  ='store',
+            default = None,
+            type    = str,
+            nargs   = '*',
+            help    =   "Enable live reloading during app development."
+                        "By default it watches all *.py *.html *.css *.yaml files"
+                        "in the app directory tree. Additional files can be passed"
+                        "as arguments."
+                        "NOTE: This setting only works with a single app."
+                        "It also restricts the number of processes to 1.",
         )),
     )
 
@@ -579,18 +585,26 @@ class Serve(Subcommand):
                     if (fnmatch(name, '*.html') or
                         fnmatch(name, '*.css') or
                         fnmatch(name, '*.yaml')):
-                        print("Watching: " + os.path.join(path, name))
-                        tar.watch(os.path.join(path, name))
+                        log.info("Watching: " + os.path.join(path, name))
+                        watch(os.path.join(path, name))
+
+        def add_optional_autoreload_files(file_list):
+            for filen in file_list:
+                if os.path.isdir(filen):
+                    log.warn("Cannot watch directory " + filen)
+                    continue
+                log.info("Watching: " + filen)
+                watch(filen)
 
         if server_kwargs['autoreload']:
             if len(applications.keys()) != 1:
                 die("--dev can only support a single app.")
             if server_kwargs['num_procs'] != 1:
-                log.info("Running in --dev mode. Will limit processes to 1.")
+                log.info("Running in --dev mode. --num-procs is limited to 1.")
                 server_kwargs['num_procs'] = 1
 
             find_autoreload_targets(args.files[0])
-
+            add_optional_autoreload_files(server_kwargs['autoreload'])
 
 
         with report_server_init_errors(**server_kwargs):
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -482,6 +482,11 @@ class Serve(Subcommand):
             type=int,
         )),
 
+        ('--dev', dict(
+            action='store_true',
+            help="Enable live reloading during app development."
+                 "NOTE: This setting has effect ONLY for Tornado>=5.0",
+        )),
     )
 
 
@@ -558,6 +563,8 @@ class Serve(Subcommand):
 
         server_kwargs['use_index'] = not args.disable_index
         server_kwargs['redirect_root'] = not args.disable_index_redirect
+        
+        server_kwargs['autoreload'] = args.dev
 
         with report_server_init_errors(**server_kwargs):
             server = Server(applications, **server_kwargs)
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -573,7 +573,7 @@ class Serve(Subcommand):
 
         server_kwargs['use_index'] = not args.disable_index
         server_kwargs['redirect_root'] = not args.disable_index_redirect
-        server_kwargs['autoreload'] = args.dev
+        server_kwargs['autoreload'] = args.dev is not None
 
         def find_autoreload_targets(app_path):
             path = os.path.abspath(app_path)
@@ -604,7 +604,7 @@ class Serve(Subcommand):
                 server_kwargs['num_procs'] = 1
 
             find_autoreload_targets(args.files[0])
-            add_optional_autoreload_files(server_kwargs['autoreload'])
+            add_optional_autoreload_files(args.dev)
 
 
         with report_server_init_errors(**server_kwargs):
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -563,7 +563,6 @@ class Serve(Subcommand):
 
         server_kwargs['use_index'] = not args.disable_index
         server_kwargs['redirect_root'] = not args.disable_index_redirect
-        
         server_kwargs['autoreload'] = args.dev
 
         with report_server_init_errors(**server_kwargs):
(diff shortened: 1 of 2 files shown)


rs2:
:+1:

bryevdv:
This is interesting. AFAIK this option did not exist a few years ago when last looked, so we were looking at implementing our own watchdog, adding a C code build step to vendor `watchdog`, or get `watchdog` available and installable on more platforms. 

However, there are still some questions to answer:

* Why does this work? The tornado docs say it automatically watches all imported modules, but Bokeh *reads* module code as text to create modules programmatically (so that each session is isolated). If there is some assumption or coincidence this relies on, I want to make sure we don't break it in the future

* It does not work with theme files or templates in directory style apps. Changing these has no effect. We can either be clear about this limitation and document it so there is a place to point people when the inevitable support questions come in, or we an try to use the Tornado auto-reload API to track these things. I don't have a preference, so it's up to you.

* Tornado docs also state that this option does not work well with `--num-threads`. We should catch this combination and warn, error out with a better error message or ignore one of them when they conflict. Again, whichever you want to do is fine with me.

Thanks for the PR!

birdsarah:
Thanks for getting the ball moving @jensschroer, this is great. Let us know any support you need to get this through to completion.

For @bryevdv's second point, I'm okay with including this feature without watching theme or template files and then waiting for demand for a more complete solution. This certainly moves things forward in a positive way.

birdsarah:
Source: https://www.tornadoweb.org/en/stable/autoreload.html

```
tornado.autoreload.watch(filename)
    Add a file to the watch list.
    All imported modules are watched by default.
```

Maybe it would be straightforward-ish to add all the files in an app directory. Or, I'd also be 100% fine with a non-automagic solution where you do something like:

```
app/
- main.py
- templates/
  - index.html
  - style.css
- misc.py
- theme.yaml
```

```sh
bokeh serve app --dev app/main.py app/theme.yaml app/style.css
```

So I've written out all the files I'm interested in watching.

jensschroer:
@bryevdv Thank you for the feedback.
Regarding your questions

1. Why does it work - `tornado.autoreload` is checking two pieces: `sys.modules` and an internal watch list, which is updated by the function `tornado.autoreload.watch(filename)` as @birdsarah (thank you!) mentioned. However it finds the bokeh app via the `sys.modules` monitor, as `CodeHandler`  in `bokeh.application.handlers.code` adds the loaded module to `sys.modules` in its `modify_document()` method. The remaining dependencies are pulled in when `CodeRunner`  in `bokeh.application.handlers.code_runner`  calls `exec()` in its `run()` method. So as long as bokeh keeps on adding the at runtime created temporary module to `sys.modules` the autoreload should work ( assuming that tornado does not change its implementation)
2. While working out why the file tracking works, I realised that using the `--dev` option only makes sense if bokeh is started with a single app as argument, as otherwise a file change in any app will restart all apps. So --dev forces to start bokeh with a single app only. However if it recognises that the parameter is an app directory, it will also add all `*.yaml`, `*.css` and `*.html` files to the watch list.
3. I did not find a thread option for bokeh serve but the `--num-proc` option. I have limited this one to 1. Is there another way of passing threads to tornado that I am not aware of?

bokeh/command/subcommands/serve.py
```diff
@@ -558,6 +567,31 @@ def invoke(self, args):
...
+            path = os.path.abspath(app_path)
+            if not os.path.isdir(path):
+                return
+
+            for path, subdirs, files in os.walk(path):
+                for name in files:
+                    if (fnmatch(name, '*.html') or
+                        fnmatch(name, '*.css') or
+                        fnmatch(name, '*.yaml')):
+                        print("Watching: " + os.path.join(path, name))
```
bryevdv: Use a `log.debug` or `log.info` instead of a bare print

bokeh/command/subcommands/serve.py
```diff
@@ -482,6 +484,13 @@ class Serve(Subcommand):
             type=int,
         )),
 
+        ('--dev', dict(
+            action='store_true',
+            help="Enable live reloading during app development."
+                 "NOTE: This setting has effect ONLY for Tornado>=5.0"
```
bryevdv: I think I may have just missed this module previously, it looks like `tornado.autoreload` is in Tornado 4.x? https://www.tornadoweb.org/en/branch4.5/autoreload.html If so, this line about >=5.0 can be removed.

bokeh/command/subcommands/serve.py
```diff
@@ -309,7 +309,9 @@
 from bokeh.server.tornado import DEFAULT_WEBSOCKET_MAX_MESSAGE_SIZE_BYTES
 from bokeh.settings import settings
 
-from os import getpid
+import tornado.autoreload as tar
```
bryevdv: `from tornado.autoreload import watch` would would be more consistent with the rest of the codebase

bryevdv:
@jensschroer Thank you for the detailed explanation. I have left a few final comments then this should be ready to go.

Re: `--num-threads` I simply mis-spoke, I meant `--num-procs`

birdsarah:
Firstly, I am using this PR today, as I'm working on bokeh app. It's already useful. So a huge hurray, and I feel more confident in my first notion that even the minimal add is a useful step forward.

Second, I think the limitation that you can only work on one app at a time is perfectly reasonable. I don't recall ever *developing* multiple apps simultaneously.

In addition to manually adding non-python files, my ultimate solution would have the browser auto-refresh.

For reference to what I currently do now.

I have this file `watch.sh`

```sh
#!/bin/sh

sigint_handler()
{
  kill $PID
  exit
}

trap sigint_handler SIGINT

while true; do
  $@ &
  PID=$!
  inotifywait -e modify -e move -e create -e delete -e attrib -r `pwd`
  kill $PID
done
```

In my template, I have the following line which I uncommend

```html
    <!--UNCOMMENT FOR DEV<script> 
    document.write('<script src="http://' + (location.host || 'localhost').split(':')[0] +
    ':35729/livereload.js?snipver=1"></' + 'script>')
    </script>-->
```

And then I run the following from the command line:

```sh
$ conda install nodejs
$ npm install
$ ./node_modules/livereload/bin/livereload.js . -- extraExts py -d -w 5000
$ ./watch.sh "bokeh serve ." . 
```

But all this is a pain to setup and a lot of times I don't bother and just waste time manually refreshing over and over again.

bryevdv:
@birdsarah is this a platform thing? If I just add `server_kwargs['autoreload'] = True` to my local install as a test, then on OSX/Safari the page reloads by itself when I save an app file.

jensschroer:
@birdsarah are there any other files that you would like me to add to the watch list? *.js for example?

birdsarah:
@jensschroer 

> @birdsarah are there any other files that you would like me to add to the watch list? *.js for example?

I think I'd like all the files in my app directory to be added to the watch list. The known files would be theme.yaml and templates/index.html. But I typically also have one or more css files in my templates directory too that I'd like to watch.

As I said earlier, I would be fine with manually passing a list of files:

```
bokeh serve app --dev app/theme.yaml app/templates/style.css app/templates/index.html
```

Everything after --dev is the files I want watched.

---

@bryevdv 

> @birdsarah is this a platform thing? If I just add server_kwargs['autoreload'] = True to my local install as a test, then on OSX/Safari the page reloads by itself when I save an app file.

WAAATTTT. Am jealous. I am Firefox on Fedora. I will investigate.

birdsarah:
Okay. Update. I've cleaned-up my livereload workflow:

1. One off install of live reload in my whole system

    $ npm install -g livereload

2. One off install of live reload FF addon

    https://addons.mozilla.org/en-US/firefox/addon/livereload-web-extension/?src=search

3. Dev mode:
    
    $ livereload
    $ bokeh serve . --dev

My browser is auto-reloading on all file changes (because livereload watches everything). But bokeh doesn't restart on non-python file changes (as expected).

birdsarah:
(deleted my last message, this isn't working as well as I thought)

bryevdv:
@birdsarah I'd suggest we make a separate issue to investigate forcing and auto-refresh once this PR is done. I had actually originally imagined a protocol message that would trigger a `window.reload`, but I'd want to make sure that it's entirely disabled inside BokehJS except unless `--dev` is active, and we should make sure doing that unconditionally doesn't cause problems on platforms where auto-refresh already happens.

birdsarah:
In poking at this I discovered that tornado doesn't complain if you add file to watch that don't exist. So it should be straight forward to add the default files that bokeh automagically recognizes and does stuff with.

@bryevdv protocol message is a great idea.

jensschroer:
@bryevdv Thank you for the review. I addressed your change requests. I also added your wish @birdsarah . Take a look. 

I am not sure why the  AppVeyor build is failing or what do about it. Any help appreciated.

birdsarah:
@jensschroer, with your latest code updates, what is the way to invoke watching files?

jensschroer:
The way you described in your post @birdsarah. However all existing files
in the app subdirectories ending with .css .html .yaml and .py are watched
by default .

birdsarah:
Then I think it's not working for me anymore.

If I just type `bokeh serve . --dev` I am not getting restarts (I was before)

birdsarah:
Yes, I've confirmed. Once you supply an argument it's working, but with no arguments it's currently not working - because you're saving None in that case?

bryevdv:
The appveyor failure appears to be spurious, another push should restart it.

jensschroer:
Yes that how I've used it so far. I also get the new tab but that was good
enough for me.

bryevdv:
Is it possible to make things work with just 
```
bokeh serve . --show --dev?
```

Adding `main.py` at the end seems arbitrary and hard to discover

jensschroer:
I think so. Just need to fix my code - Sorry for the bug. Haven't had the
time to look at it yet. But latest Monday I should be able to.

jensschroer:
Triggering build

bryevdv:
Thanks @jensschroer this looks good to me, but I have not had a chance to pull the branch and test things out locally. 

@birdsarah I am conferencing/travelling for several days. If you get a chance to pull the branch and test out more, please merge if all is working as you expect.

birdsarah:
@bryevdv - the master build failed on integration tests. feel sure it's not this PR but wanted to check in. It's been a while since I merged anything.

bryevdv:
@birdsarah nope it's fine, we just have one single flaky integration test that has not been fixed/skipped yet

PR #8256, opened 2018-09-19
add plumbing to make legends update changes

### LINKED ISSUE #589: Update Legend when legends or other attributes change

bryevdv:
- [x] issues: fixes #589

Re: testing, there is not any good current way. This needs an integration/selenium test since it involves executing an interaction to make an update to legend state. It's not merely enough to inspect the state however, the test needs to confirm the render (that was the actual issue to fix). That will require "image-diff" integration tests, which we do not have yet. I will add an issue to add those, and start a list linking back to specific issues like this that need one. 

Test script:
```
from bokeh.io import curdoc

x = [1,2,3,4,5, 10]
y = [5,6,2,3,4,10]

p = figure()
r = p.line(x, y, legend="foo")

def cb():
    p.circle(x, y, legend="foo")
    r.glyph.line_dash = "dotted"
    r.glyph.line_color = "black"
    p.line([1,4,8], [2, 12, 6], line_color="red", legend="bar")
    p.legend.background_fill_color = "blue"
    p.legend.background_fill_alpha = 0.2

button = Button()
button.on_click(cb)

curdoc().add_root(column(p, button))
```

Before:

screen shot 2018-09-19 at 11 06 48

After:

screen shot 2018-09-19 at 11 06 56

### DIFF
--- a/bokehjs/src/lib/models/annotations/legend.ts
+++ b/bokehjs/src/lib/models/annotations/legend.ts
@@ -6,6 +6,7 @@ import {Line, Fill, Text} from "core/visuals"
 import {FontStyle, TextAlign, TextBaseline, LineJoin, LineCap} from "core/enums"
 import {Orientation, LegendLocation, LegendClickPolicy} from "core/enums"
 import * as p from "core/properties"
+import {Signal0} from "core/signaling"
 import {get_text_height} from "core/util/text"
 import {BBox} from "core/util/bbox"
 import {max, all} from "core/util/array"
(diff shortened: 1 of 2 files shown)


bryevdv:
cc @mattpap in case you have any comments. I don't think this is perfect, but I think it is actually much more simple and robust than trying to connect/disconnect things dynamically.

bryevdv:
Going to merge as the change is pretty trivial. We can make changes later after an integration image test is added. 

Noting for the future: It might be better to have legend items have their own view to draw themselves, then the signalling can be simplified as the items could request a re-render directly. Additionally it seems like there might be value in distinguishing expicitly between "simple" items, and items for "groupby" legends that might generate multiple "simple" items on the client side.

PR #8274, opened 2018-09-24
Added ability to define custom tooltip and icon on EditTools

### LINKED ISSUE #7499: Allow defining custom tooltip for certain tools

philippjfr:
Adds a ``custom_tooltip`` property to all tools, which if defined overrides the ``tool_name``. This allows telling apart multiple tools added for different purposes, which is mostly useful for the editing tools. If we think it's preferable to limit this to only those specific tool I'd be happy to remove it from other tools.

- [x] issues: fixes #7499
- [ ] tests added / passed

### DIFF
--- a/bokehjs/test/models/tools/edit/box_edit_tool.ts
+++ b/bokehjs/test/models/tools/edit/box_edit_tool.ts
@@ -79,6 +79,17 @@ const make_testcase = function(): BoxEditTestCase {
 
 describe("BoxEditTool", () =>
 
+  describe("Model", function(): void {
+
+    it("should create proper tooltip", function(): void {
+      const tool = new BoxEditTool({})
+      expect(tool.tooltip).to.be.equal('Box Edit Tool')
+
+      const custom_tool = new BoxEditTool({custom_tooltip: 'Box Edit Custom'})
+      expect(custom_tool.tooltip).to.be.equal('Box Edit Custom')
+	});
+  }),
+
   describe("View", function(): void {
 
     it("should select rect on tap", function(): void {
--- a/bokehjs/test/models/tools/edit/box_edit_tool.ts
+++ b/bokehjs/test/models/tools/edit/box_edit_tool.ts
@@ -77,7 +77,7 @@ const make_testcase = function(): BoxEditTestCase {
   }
 }
 
-describe("BoxEditTool", () =>
+describe("BoxEditTool", function(): void {
 
   describe("Model", function(): void {
 
@@ -253,6 +253,6 @@ describe("BoxEditTool", () =>
       const drag_event = make_gesture_event(300, 300, true);
       testcase.draw_tool_view._doubletap(drag_event);
       expect(testcase.draw_tool_view._draw_basepoint).to.be.equal(undefined);
-    });
-  }),
-);
+    })
+  })
+});
--- a/bokehjs/test/models/tools/edit/box_edit_tool.ts
+++ b/bokehjs/test/models/tools/edit/box_edit_tool.ts
@@ -82,12 +82,12 @@ describe("BoxEditTool", () =>
   describe("Model", function(): void {
 
     it("should create proper tooltip", function(): void {
-      const tool = new BoxEditTool({})
+      const tool = new BoxEditTool()
       expect(tool.tooltip).to.be.equal('Box Edit Tool')
 
       const custom_tool = new BoxEditTool({custom_tooltip: 'Box Edit Custom'})
       expect(custom_tool.tooltip).to.be.equal('Box Edit Custom')
-	});
+    });
   }),
 
   describe("View", function(): void {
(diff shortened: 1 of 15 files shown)


bryevdv:
@philippjfr instead of always having to make sure every tool does this, it seems like you should be able to hook in a single place, e.g. here

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/tools/button_tool.ts#L32

I'm also not super happy with the way this is duplicative of `help_tooltip` and `action_tooltip`

philippjfr:
>@philippjfr instead of always having to make sure every tool does this, it seems like you should be able to hook in to override things in a single place, e.g. here

Thanks, the problem with that is that it won't combine nicely with the dimension based formatters, i.e. ones which append (x-axis) or (y-axis). That said maybe that's actually the behavior we want. Either way I'm sure this could be refactored to be less duplicative.

>I'm also not super happy with the way this is duplicative of help_tooltip and action_tooltip

Yes, I noticed those after doing this. Maybe it is better to only add the custom tooltips to certain tools, which can be used in different ways, i.e. mostly the edit tools.

bryevdv:
> That said maybe that's actually the behavior we want. 

I think so, I would make this an unconditional override, then users can put whatever they want which, is simple to explain and document  Otherwise the behavior is quite a bit more complicated with different cases and that's harder to explain. 

> Yes, I noticed those after doing this. Maybe it is better to only add the custom tooltips to certain tools, which can be used in different ways, i.e. mostly the edit tools.

It's not a blocker for me, but if you can think of any simple improvements or changes that would not be hugely invasive I think they'd be worth doing. Worst case we add them everywhere and deprecate `help_tooltip` and `'action_tooltip` in favor of `custom_tooltip` sometime in the future

philippjfr:
Okay I've now made `custom_tooltip` a hard override for the tooltip getter as you suggested. Certainly much simpler as far as the implementation goes.

>Otherwise, worst case we add them everywhere and deprecate `help_tooltip` and `action_tooltip` in favor of custom_tooltip sometime in the future

That seems fairly sensible to me.

philippjfr:
I've now decided to isolate these changes on the `EditTool` baseclass. This way there are no duplication issues with `help_tooltip` and `action_tooltip` and the edit tools are the ones that benefit most from this customizability.

One problem I'm still having is that since I've made icon a property it is now complaining that the default icon value on the JS end does not match the Python default, which is expected but the tests will error because of it. Since `icon` is of `Image` type on the Python end I also can't set a default.

philippjfr:
What I could do is rename the `icon` property to `custom_icon`, create an `icon` getter and define the default as `default_icon`. Does that sound reasonable?

bryevdv:
What about leaving the public API unchanged (i.e. "icon") and use something new internally for the property (i.e. "computed_icon")

bokehjs/test/models/tools/edit/box_edit_tool.ts
```diff
@@ -79,6 +79,17 @@ const make_testcase = function(): BoxEditTestCase {
...
 
+  describe("Model", function(): void {
+
+    it("should create proper tooltip", function(): void {
+      const tool = new BoxEditTool({})
+      expect(tool.tooltip).to.be.equal('Box Edit Tool')
+
+      const custom_tool = new BoxEditTool({custom_tooltip: 'Box Edit Custom'})
+      expect(custom_tool.tooltip).to.be.equal('Box Edit Custom')
+	});
```
mattpap: Bad indentation (same in other test files). Looks like a tab (?). I've been seeing this in other places in recent commits.

philippjfr: Thanks, missed this one, I messed up my emacs config and haven't had time to fix it.

bokehjs/test/models/tools/edit/freehand_draw_tool.ts
```diff
@@ -75,6 +75,17 @@ const make_testcase = function(): FreehandDrawTestCase {
 
 describe("FreehandDrawTool", (): void => {
 
+  describe("Model", function(): void {
+
+    it("should create proper tooltip", function(): void {
+      const tool = new FreehandDrawTool({})
```
mattpap: There's no need for `{}`, because `attrs` are optional.

philippjfr:
>What about leaving the public API unchanged (i.e. "icon") and use something new internally for the property (i.e. "computed_icon")

Sounds fine with me but you would still have `icon` as a property on the Python side and a `icon` default on the JS side which does not match. Therefore I'd still have to call the property `custom_icon`. It does avoid introducing `default_icon` though, which is probably a good thing. Happy to make that change, just wanted to be clear.

philippjfr:
Ready to merge if you agree with the latest changes.

PR #8278, opened 2018-09-26
Remove global event listeners when deleting views to prevent memory leaks

### LINKED ISSUE #5355: No way to cleanup used memory
When you embed bokeh plots by manually inserting div and script you get from `components` there is no obvious way to clean up the memory when you no longer need plots. This results into memory leaking.

Example code can be found here: https://github.com/sashabaranov/bokeh-leak-example

Also this issue on stackoverflow: http://stackoverflow.com/questions/39895717/remove-bokeh-plot-without-leaking-memory

```
% bokeh info
Python version      :  2.7.12 (default, Sep 30 2016, 15:42:31)
Bokeh version       :  0.12.3
BokehJS static path :  /Users/sashab1/.virtualenvs/bokeh-leak-example/lib/python2.7/site-packages/bokeh/server/static
```

dracorpg:
As promised in the [memory leak issue discussion](https://github.com/bokeh/bokeh/issues/5355), here is a PR for properly removing event listeners registered by BokehJS views on **window** and **document** global objects.
Because they constitute an active reference to the view, these listeners prevent garbage collection of most (all ?) of a Bokeh document's memory footprint after it has been entirely removed from the DOM.

I don't know what kind of testing is expected from such a PR. All I can say is these changes don't evidently break bokehjs' behavior and now re-running again and again the same cell in Jupyter Notebook shows the memory still increasing short-term, but being properly garbage-collected afterwards.

- [x] issues: fixes #5355 
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/core/ui_events.ts
+++ b/bokehjs/src/lib/core/ui_events.ts
@@ -83,6 +83,34 @@ export class UIEvents {
               readonly hit_area: HTMLElement,
               readonly plot: Plot) {
     this._configure_hammerjs()
+    
+    // Mouse & keyboard events not handled through hammerjs
+
+    // We can 'add and forget' these event listeners because this.hit_area is a DOM element
+    // that will be thrown away when the view is removed
+    this.hit_area.addEventListener("mousemove", (e) => this._mouse_move(e))
+    this.hit_area.addEventListener("mouseenter", (e) => this._mouse_enter(e))
+    this.hit_area.addEventListener("mouseleave", (e) => this._mouse_exit(e))
+    this.hit_area.addEventListener("wheel", (e) => this._mouse_wheel(e))
+
+    // But we MUST remove listeners registered on document or we'll leak memory: register
+    // 'this' as the listener (it implements the event listener interface, i.e. handleEvent)
+    // instead of an anonymous function so we can easily refer back to it for removing
+    document.addEventListener("keydown", this)
+    document.addEventListener("keyup", this)
+  }
+
+  destroy(): void {
+    this.hammer.destroy()
+    document.removeEventListener("keydown", this)
+    document.removeEventListener("keyup", this)
+  }
+
+  handleEvent(e: KeyboardEvent): void {
+    if (e.type == "keydown")
+      this._key_down(e)
+    else if (e.type == "keyup")
+      this._key_up(e)
   }
 
   protected _configure_hammerjs(): void {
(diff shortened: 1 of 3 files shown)


bokehjs/src/lib/core/ui_events.ts
```diff
@@ -109,17 +137,8 @@ export class UIEvents {
...
-    this.hit_area.addEventListener("mouseenter", (e) => this._mouse_enter(e))
-    this.hit_area.addEventListener("mouseleave", (e) => this._mouse_exit(e))
-
-    this.hit_area.addEventListener("wheel", (e) => this._mouse_wheel(e))
-
-    document.addEventListener("keydown", (e) => this._key_down(e))
-    document.addEventListener("keyup", (e) => this._key_up(e))
   }
-
+
```
mattpap: You need to remove trailing whitespace here and possibly some other places to fix `codebase` tests.

bokehjs/src/lib/models/layouts/layout_dom.ts
```diff
@@ -242,12 +242,21 @@ export abstract class LayoutDOMView extends DOMView {
...
     this.connect(this.model.properties.sizing_mode.change, () => this.layout())
   }
 
+  handleEvent(): void {
+    this.resize()
+  }
+
+  disconnect_signals(): void {
+    super.disconnect_signals()
+    window.removeEventListener("resize", this)
```
mattpap: It doesn't matter here, but this should be done in reverse.

dracorpg: You mean the super call should go at the end of the method? I checked and there should be no impact to the ordering indeed, so I followed the pattern in connect_signals with super call at the top. But thinking about it again, it seems logical to mirror it.

dracorpg:
@mattpap do I need to do anything else regarding this PR?

bryevdv:
Regarding testing, I"m not sure what good options would be. We could use mocks to verify that the chain of calls starting will `PlotCanvasView.remove` ends up with the expected `removeEventListener` invocations.

bryevdv:
After discussion with @mattpap we will go ahead and merge this. We would like to make a more comprehensive and focused effort, but the useful comments you left in the code should be an adequate signpost to anyone working on this codepath in the future to take care about removing handlers.

bryevdv:
Also to note: I will kick off a dev build later today

bokehjs/src/lib/models/layouts/layout_dom.ts
```diff
@@ -242,12 +242,21 @@ export abstract class LayoutDOMView extends DOMView {
...
 
     if (this.is_root)
-      window.addEventListener("resize", () => this.resize())
+      window.addEventListener("resize", this)
 
     // XXX: this.connect(this.model.change, () => this.layout())
     this.connect(this.model.properties.sizing_mode.change, () => this.layout())
   }
 
+  handleEvent(): void {
```
mattpap: @dracorpg, one thing to consider in future is to make the class implementing this method declare the interface being implemented (`EventListenerObject` in this case). Given that we use snake-case naming convention, one could be eager to rename this method accordingly, or even remove it, as it isn't referenced (at least directly) anywhere. I implemented this in https://github.com/bokeh/bokeh/pull/8085/commits/9e776b491d1eda1c9fdf907d5f630d57e3357ea9.

PR #8285, opened 2018-09-28
Bryanv/5884 scatter

### LINKED ISSUE #5884: Create Marker class that encompasses all markers and allow specific marker type to be specified from data

bryevdv:
- [x] issues: fixes #5884
- [x] tests added / passed

```
from bokeh.transform import factor_cmap, factor_marker

from bokeh.models import CategoricalMarkerMapper

SPECIES = ['setosa', 'versicolor', 'virginica']
MARKERS = ['circle', 'diamond', 'triangle']

p.scatter("petal_length", "petal_width", source=flowers, legend="species", 
          fill_alpha=0.4, size=12,
          marker=factor_marker('species', MARKERS, SPECIES),
          color=factor_cmap('species', 'Category10_3', SPECIES))
          
show(p)
```

screen shot 2018-09-27 at 19 54 48


bryevdv:
I'd like to merge this tomorrow

bryevdv:
@mattpap I will go ahead and merge this with green, I am working on another PR that builds on it. It has fairly extensive docs and tests. If there is anything you'd amend about the BokehJS changes, we can circle back later.

PR #8293, opened 2018-10-01
bokeh/colors boilerplate

HuntJSparra:
Adds boilerplate for bokeh/colors

- [x] issues: partially addresses #8207 
- [ ] tests added / passed


bokeh/colors/groups.py
```diff
@@ -148,3 +150,32 @@ class black(ColorGroup):
...
+    green,
+    orange,
+    pink,
+    purple,
+    red,
+    white,
+    yellow,
+)
+
+for color_group in color_groups:
```
HuntJSparra: @bryevdv 
1) Does the reorganization of this file to have the ColorGroups under General API instead of constants seem reasonable?
2) Does moving the docstring modifications to the Code section for said groups also seem reasonable?
3) I have changed the docstring modification to use a for loop instead of manually doing it for each ColorGroup. Would you rather have it this way or the original way (commented out below)?
4) Brown previously did not have any docstring modification. Was this intentional or an oversight?

bryevdv: @HuntJSparra I am not sure `color_groups` was ever more than some internal bookkeeping. That said, as we approach 1.0 I have decided to make an effort to reduce dynamic generation of docstrings. I'd actually prefer you replace the templated docstring with a fully rendered static version and not use `format_docstring` at all. 

Brown missing was just an oversight I imagine

PR #8294, opened 2018-10-02
add minor ticks "minor_ticks=" option in FixedTicker

### LINKED ISSUE #8295: FixedTicker does not accept minor ticks
This is a feature request. 

The current default implementation of FixedTicker ignores minor ticks, so is only able to accept a list of values that correspond to major tick locations. The desire to have minor ticks stems from enhancing FixedTicker into a more flexible custom ticker, particularly in the realm of custom axes transformations, where it is helpful to use a combination of major/minor ticks along with label overrides to visually observe the non-intuitive effects of the dimension transform. 

Because FixedTicker is based off of ContinuousTicker, the option to handle minor ticks is already mostly there. I couldn't see a strong reason not to enable the minor_ticks option. Hence, laying out the thought process here for a PR.

### Existing behavior
To stay backwards compatible, users should still be able to call FixedTicker one of two ways: 

Shorthand way ([simply calls FixedTicker](https://github.com/bokeh/bokeh/blob/13baccbb2835f7d5831c22b5cf68214e89c02618/bokeh/models/axes.py#L63))
```
p.axis.ticker = [0, 5, 10]
```
Long way
```
p.axis.ticker = FixedTicker(ticks=[0, 5, 10])
```

### Requested behavior

```
p.axis.ticker = FixedTicker(ticks=[0, 5, 10], minor_ticks=[1,2,3,4, 6,7,8,9])

#by default, minor_ticks=[] would not have to be specified
```
the choice of `minor_ticks` as the attribute spelling follows the javascript side convention, as used on the [bokehjs side, here](https://github.com/bokeh/bokeh/blob/f9a4013a3640da3c73fc49ed906f6d202a30a489/bokehjs/src/lib/models/tickers/continuous_ticker.ts#L80)
however, in the bokeh folder, the tickers do not specify "minor_ticks" as it is relegated for the javascript side to handle, but for completeness sake, need to doublecheck whether this would be a different convention as "minor_ticks" is used elsewhere on the [python side, here](https://github.com/bokeh/bokeh/blob/99632902efdfe628809ee02b8fc1abd829c280b7/bokeh/plotting/helpers.py#L475)

jfx319:
This PR attempts to add an optional minor_ticks option for FixedTicker. 

Subsequently, one should be able to pass FixedTicker a list of minor tick values analogous to how one would pass a list of major tick values. 
```python
p.axis.ticker = FixedTicker(ticks=[0, 5, 10], minor_ticks=[1,2,3,4, 6,7,8,9])
```
past behavior should still work because `minor_ticks=[]` by default

I am new to typescript, so please doublecheck the work! Not sure where else to update (e.g. in [tests](https://github.com/bokeh/bokeh/blob/master/bokehjs/test/models/tickers/fixed_ticker.ts) or in docs)

- [ ] issues: fixes #8295
- [ ] tests added / passed

## Tested example for reproducibility
```python
import numpy as np

from bokeh.io import show, output_file, output_notebook
from bokeh.models.tickers import FixedTicker

#data (will not be plotted!)
major_data = [0.1, 1, 10, 100]
minor_data = [0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9] + [2, 3, 4, 5, 6, 7, 8, 9] + [20, 30, 40, 50, 60, 70, 80, 90]

#original data will be viewed in log10 space
minor_ticks = np.log10(minor_data).tolist()
major_ticks = np.log10(major_data).tolist()
major_data_labels = ["10⁻¹", "10⁰", "10¹", "10²"]

#workaround to handle whole number floats being treated differently in override process
    #https://github.com/bokeh/bokeh/issues/8166
major_ticks_converted = [int(item)  if float(item).is_integer()  else item  for item in major_ticks]

major_label_overrides = dict(zip(major_ticks_converted, major_data_labels))
```
#### Default behavior
```python
p = figure(x_range=(-2,3))
p.circle(x=[-1,0,1,2,2], y=[1,1,1,1,1.5])

#p.axis.ticker = major_ticks  #to verify shorthand compatibility
p.xaxis.ticker = FixedTicker(ticks=major_ticks)
p.xaxis.major_label_overrides = major_label_overrides
 
#output_notebook()
output_file('FixedTicker_default.html')
show(p)
```
image

#### New behavior after recompiling bokehjs and activating with `BOKEH_RESOURCES=inline`
```python
p = figure(x_range=(-2,3))
p.circle(x=[-1,0,1,2,2], y=[1,1,1,1,1.5])

p.xaxis.ticker = FixedTicker(ticks=major_ticks, minor_ticks=minor_ticks)
p.xaxis.major_label_overrides = major_label_overrides

#output_notebook()
output_file('FixedTicker_new.html')
show(p)
```
image

### DIFF
--- a/bokeh/models/tickers.py
+++ b/bokeh/models/tickers.py
@@ -40,8 +40,10 @@ class ContinuousTicker(Ticker):
 
 class FixedTicker(ContinuousTicker):
     ''' Generate ticks at fixed, explicitly supplied locations.
+
     .. note::
         The ``desired_num_ticks`` property is ignored by this Ticker.
+
     '''
 
     ticks = Seq(Float, default=[], help="""
--- a/bokeh/models/tickers.py
+++ b/bokeh/models/tickers.py
@@ -47,8 +47,8 @@ class FixedTicker(ContinuousTicker):
     ticks = Seq(Float, default=[], help="""
     List of major tick locations.
     """)
-	
-	minor_ticks = Seq(Float, default=[], help="""
+
+    minor_ticks = Seq(Float, default=[], help="""
     List of minor tick locations.
     """)
 
--- a/bokeh/models/tickers.py
+++ b/bokeh/models/tickers.py
@@ -40,14 +40,16 @@ class ContinuousTicker(Ticker):
 
 class FixedTicker(ContinuousTicker):
     ''' Generate ticks at fixed, explicitly supplied locations.
-
     .. note::
         The ``desired_num_ticks`` property is ignored by this Ticker.
-
     '''
 
     ticks = Seq(Float, default=[], help="""
-    List of tick locations.
+    List of major tick locations.
+    """)
+	
+	minor_ticks = Seq(Float, default=[], help="""
+    List of minor tick locations.
     """)
 
 class AdaptiveTicker(ContinuousTicker):
--- a/bokehjs/src/lib/models/tickers/fixed_ticker.ts
+++ b/bokehjs/src/lib/models/tickers/fixed_ticker.ts
@@ -5,6 +5,7 @@ import * as p from "core/properties"
 export namespace FixedTicker {
   export interface Attrs extends ContinuousTicker.Attrs {
     ticks: number[]
+    minor_ticks: number[]
   }
 
   export interface Props extends ContinuousTicker.Props {}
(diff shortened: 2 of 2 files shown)


bokeh/models/tickers.py
```diff
@@ -40,14 +40,16 @@ class ContinuousTicker(Ticker):
 
 class FixedTicker(ContinuousTicker):
     ''' Generate ticks at fixed, explicitly supplied locations.
-
     .. note::
         The ``desired_num_ticks`` property is ignored by this Ticker.
-
```
bryevdv: @jfx319 this whitespace is necessary, and is what is currently failing the build (Sphinx is very picky)

bryevdv:
@jfx319 Thanks for the PR, looks good I only had one comment.

bryevdv:
Thanks @jfx319 !

PR #8297, opened 2018-10-02
use standard args for gridplot

### LINKED ISSUE #8273: Unhelpful error when passing invalid argument to gridplot
I often find myself assembling a list of plots and then writing:

    $ show(gridplot(plots, n_cols=4))

This will give the following error:

```python
TypeError                                 Traceback (most recent call last)
<ipython-input-25-656b46f60be9> in <module>()
----> 1 show(gridplot(plots, n_cols=4, toolbar_location=None))

~/Dev/bokeh/bokeh/bokeh/layouts.py in gridplot(*args, **kwargs)
    320         row_tools = []
    321         row_children = []
--> 322         for item in row:
    323             if merge_tools:
    324                 if item is not None:

TypeError: 'Figure' object is not iterable
```

The problem is that I wrote `n_cols` instead of `ncols`. At this point I've done this dozens and dozens of times. Every time I come back to bokeh after a gap, this always trips me up and I forget the ncols/n_cols thing and think there's some more fundamental mistake.

Can we validate the input to gridplot better?

One of my favorite features about bokeh is the helpful suggestions when you mess up a parameter. I know this is completely different, and nothing to do with the property machinery, but none-the-less a similar experience would be awesome.

bryevdv:
This PR removes unnecessary arbitrary arg/kwarg processing in `gridplot` in favor of standard Python optional args 

- [x] issues: fixes #8273
- [x] tests added / passed

Now get for unknown arg:

```
TypeError: gridplot() got an unexpected keyword argument 'n_cols'
```

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -231,7 +231,7 @@ def layout(*args, **kwargs):
     # Make the grid
     return _create_grid(children, sizing_mode)
 
-def gridplot(*args, sizing_mode='fixed',  toolbar_location='above', ncols=None,
+def gridplot(children, sizing_mode='fixed',  toolbar_location='above', ncols=None,
              plot_width=None, plot_height=None, toolbar_options=None, merge_tools=True):
     ''' Create a grid of plots rendered on separate canvases.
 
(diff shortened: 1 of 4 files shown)


bryevdv:
We could make this:
```
def gridplot(children, sizing_mode='fixed',  toolbar_location='above', ncols=None,
             plot_width=None, plot_height=None, toolbar_options=None, merge_tools=True):
```

which would work, but disallow 
```
gridplot(p0, p1, p2...)
```
I don't think we ever demo that anywhere, so it'd be ok with me. @birdsarah ? @mattpap ?

bryevdv:
I've gone ahead and made that change, if there is no disagreement today-ish, I will merge.

bryevdv:
Evidently we did have some examples that used that syntax. I am going to close this PR, this is not worth all the trouble.

PR #8306, opened 2018-10-04
Fix tiny typo in the document about GraphRenderer

### LINKED ISSUE #8299: A tiny typo in the document about GraphRenderer
I found a tiny typo in the document about `GraphRenderer` while working on other PRs.

**attr y_range_name**
https://bokeh.pydata.org/en/latest/docs/reference/models/renderers.html#bokeh.models.renderers.GraphRenderer.y_range_name
> A particular (named) y-range to use for computing screen locations when rendering graphs on the plot. If unset, use the default -range.

It seems that it is '"y-range", not "-range”.

I think that this line is to be modified.
If there is not problem, I will send a PR.
https://github.com/bokeh/bokeh/blob/a01ae6ecb6726cb97ae1997e57d62f0cdf6adacc/bokeh/models/renderers.py#L205-L206

komo-fr:
fixes #8299


mattpap:
I canceled the build. No point in wasting resources on a typo.

PR #8309, opened 2018-10-05
Bryanv/reduce import code

bryevdv:
This PR claims some low hanging fruit for reducing Bokeh import times:

* Don't use `PackageLoader` for loading Bokeh Jinja templates (required `pkg_resources` import is very expensive)

* defer non-stdlib imports in sampledata modules (downloading sampledata is rare usage)

* reduce some dynamic docstring manipulations

All told this shaves ~200ms off `import bokeh.plotting` on my laptop. The results using:
```
python3.7 -X importtime -c "import bokeh.plotting" 2> bokeh3.log
tuna bokeh3.log
``` 
are:

screen shot 2018-10-05 at 15 16 30

About half of that ~600ms appears to be NumPy and Pandas, which is borne out by this very rough timing:

```
In [1]: import pandas, numpy

In [2]:

In [2]: %time import bokeh.plotting
CPU times: user 190 ms, sys: 40.8 ms, total: 231 ms
Wall time: 309 ms
```

It's worth noting that 60-70 ms for computing `bokeh.__version__` disappears in real release packages where the version string is hardcoded. 

We can't do much about the NumPy/Pandas burden, but there are still some things we can do to reduce things on our end later:

* defer loading default temlate yaml files until requested (this will take a little more work/care)
* remove more dynamic module code (e.g dynamic glyph method construction)
* defer loading Jinja templates until needed
* move submodules of `bokeh.models` to `bokeh._modules` so that individual models can be imported internally without importing everything in `bokeh.models.__init__.py`

I would estimate ~150ms (relative reference on this laptop) is probably a floor for `import bokeh.plotting`

### DIFF
--- a/bokeh/core/templates.py
+++ b/bokeh/core/templates.py
@@ -17,19 +17,20 @@ models in various ways.
 '''
 from __future__ import absolute_import
 import json
-from jinja2 import Environment, Markup, FileSystemLoader, PackageLoader
+from os.path import dirname, join
 import sys
-import os
+
+from jinja2 import Environment, Markup, FileSystemLoader
 
 def get_env():
     ''' Get the correct Jinja2 Environment, also for frozen scripts.
     '''
     if getattr(sys, 'frozen', False):
-        templates_path = os.path.join(sys._MEIPASS, 'bokeh', 'core', '_templates')
+        templates_path = join(sys._MEIPASS, 'bokeh', 'core', '_templates')
         return Environment(loader=FileSystemLoader(templates_path))
     else:
-        return Environment(loader=PackageLoader('bokeh.core', '_templates'))
-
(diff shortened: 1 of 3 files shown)


bryevdv:
Additionally, in 1.0 Bokeh is switching to "simple ids" by default, i.e. a simple monotonically increasing sequence of integers. This is ~2x improvement but since you mention separate processes I will note just for completeness: if you are somehow constructing models for a single document across multiple processes (I *very much* doubt you or anyone is ever doing this), that's the one scenariou where you'd need to set `BOKEH_SIMPLE_IDS=no` to return to uuids.

bryevdv:
image report looks good, changes are small and should be uncontroversial so merging now

PR #8310, opened 2018-10-08
DataTable data model fixes

### LINKED ISSUE #7116: Patch/Stream message to the ColumnDataSource for DataTable widget on front-end triggers response containing entire data source
I've been struggling to optimize my application that is using `DataTable` widget on the front-end to show to the user a catalog of information updated in the real time. Originally I've used data source trigger to sent information to the front-end. This become quite quickly unfeasible solution, because number of information is quite large. For this reason I've tried to update data source via stream and patch operations. First problem that I've come across was that patch/stream operation takes progressively more time as the size of data source grows. I've raised issue [#7072](https://github.com/bokeh/bokeh/issues/7072).

Even after my monkey patch solution of previous problem the situation didn't improve much. Using Chrome DevTools I've monitored communication over websocket and found that on each `ColumnsStreamed` message sent to front-end, it responds with `ModelChanged` message that contains entire data source. This is hugely ineffective. Because when the data source grows these messages get big very quickly especially when `DataTable` has multiple columns. Same situation is of course with patch operation as well.

Strange thing is that my application also contains plot widgets which are not exhibiting this behavior. Therefore my conclusion is that this is caused by `DataTable` widget and not by `ColumnDataSource`.

# Expected behavior

By default Stream and Patch operation that are applied to `ColumnDataSource` of a `DataTable` widget will not trigger any response from front-end.

# System Info

Since this is only very general issue I'm presuming that these are the only relevant system information:

    bokeh: 0.12.10
    python: 3.5.2 (default, Nov 17 2016, 17:05:23)
    system: Linux-4.4.0-97-generic-x86_64-with-Ubuntu-16.04-xenial

# Example

I've put together minimal example that shows this behavior. It creates page with one button and one `DataTable`. On button click back-end starts sending table rows with `ColumnDataSource` stream method. It was implemented by modification of [export\_csv](https://github.com/bokeh/bokeh/tree/master/examples/app/export_csv) example app.

```python
import time
import random

from bokeh.io import curdoc
from bokeh.models.widgets import DataTable, TableColumn, Button

class TestSource:
  def __init__(self):
    self.source = ColumnDataSource(data = self._empty_source_dict())

  def _empty_source_dict(self):
    return {'x': [], 'quantity': []}

  def add(self, record):
    record_list = {key: [value] for key, value in record.items()}
    self.source.stream(record_list)

def get_record(index):
  return {
    'x': index,
    'quantity': random.randint(0, 500)
  }

def callback():
  for i in range(30):
    record = get_record(i)
    test_source.add(record)
    time.sleep(1)

test_source = TestSource()
columns = [TableColumn(field="x", title="x"), TableColumn(field="quantity", title="quantity")]
table = DataTable(source = test_source.source, columns = columns, width = 400)
button = Button()

button.on_click(callback)

curdoc().add_root(column(button, table))
curdoc().title = "Test of patch and stream"
```

To illustrate the issue here is last message being sent to front-end:

    {
      "references": [],
      "events": [
        {
          "column_source": {
            "id": "84c57a47-12e0-43bb-92d6-09a62dbf4707",
            "type": "ColumnDataSource"
          },
          "rollover": null,
          "kind": "ColumnsStreamed",
          "data": {
            "quantity": [41],
            "x": [29]
          }
        }
      ]
    }

On which front-end response with this one:

    {
      "events": [
        {
          "kind": "ModelChanged",
          "model": {
            "type": "ColumnDataSource",
            "id": "84c57a47-12e0-43bb-92d6-09a62dbf4707"
          },
          "attr": "data",
          "new": {
            "quantity": [419, 251, 99, 6, 44, 65, 97, 156, 43, 457, 387, 338, 34, 390, 287, 53, 181, 397, 439, 250, 154, 36, 386, 103, 300, 165, 23, 321, 446, 41],
            "x": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29]
          }
        }
      ],
      "references": []
    }

# Conclusion

My hypothesis is that `DataTable` is sending the source data back to bokeh because the widget provides the possibility of editable tables. Which would explain the need to send update back to back-end. If this is the case then I would propose an addition of parameter that would disable this behavior and possibly disable by default unless `DataTable` parameter `editable` is set to `True`.

I would appreciate some pointers on how to disable it the meantime. I starting to understand the bokeh server code base pretty well, but my knowledge is severally lacking in the front-end area.

### LINKED ISSUE #7417: DeserializationError when editing Data Table
Current environment:
- Bokeh 0.12.14dev3 (although I think the same error occurs in 0.12.13)
- Python 3.6
- Windows 10
- Google Chrome

When editing or clicking on cells in a data table bokeh will display a `DeserializationError`:
image

This can be tested with the [data_table.py](https://github.com/bokeh/bokeh/blob/53cd93bea4168f87d296ad8a0b9bcd7d18e886ae/examples/models/server/data_tables.py) example on the bokeh github. Open that server up and then click on a cell and the error should appear. Not sure where this is coming from, but I do believe it happened somewhere after 0.12.10 (which doesn't produce the error).

### LINKED ISSUE #8021: Datatable and Glyph selections not linking properly
One ref is: https://github.com/bokeh/bokeh/issues/2323#issuecomment-395032022

Alternatively, run `models/file/data_tables.py` The first selection of glyphs updates the table selection. No subsequent selection does. 

Possibly related to #8014 as well, the issue may be that there is inconsistency in when and whether `selected` is changed vs `selected.indices`

ANY solution to this must provide the following

* documention (in a wiki, docs, somewhere) of all known expected requirements of tables. 
* provide runnable examples that can be used to exercise all the requirements

Additionally, this would be a good chance to get some kind of selenium tests up and running, either run ourselves on travis or run again on sauce labs. The whack-a-mole with table regressions has to stop.

### LINKED ISSUE #8040: Editing filtered DataTable cell impacts wrong row
bokeh 0.13.0, python 3.6.5, macOS High Sierra 10.13.4

Editing filtered DataTable cell impacts wrong row. Impacted cell is below targeted one when rows above it are filtered (DTINDEX_NAME/index_position confusion).

```python
from datetime import date
from random import randint

from bokeh.models import CDSView, ColumnDataSource, CustomJS, CustomJSFilter
from bokeh.models.widgets import DataTable, DateFormatter, NumberEditor, Slider, TableColumn

threshold = Slider(title='Threshold', value=30, start=0, end=100, step=10)

source = ColumnDataSource(dict(
))

filter_ = CustomJSFilter(args=dict(threshold=threshold), code='''
return source.data.downloads.map(x => x >= threshold.value)
''')

view = CDSView(source=source, filters=[filter_])

columns = [
    TableColumn(field="downloads", title="Downloads", editor=NumberEditor()),
]

table = DataTable(source=source, view=view, columns=columns, width=400, height=280, editable=True)

threshold.js_on_change('value', CustomJS(args=dict(view=view), code='''
view.compute_indices(view.source);
view.source.properties.data.change.emit()
'''))

show(widgetbox(threshold, table))
```
Increase value of a cell where index_position > current row, hit enter, and observe your new value appearing below the cell you edited.

No error or warning in console.

### LINKED ISSUE #8201: DataTable does not stay sorted when the data is updated

bokeh: 0.13.0
python: 3.6.5
OS: Ubuntu 16.04
Browser: Firefox and Chromium

This is based on the example at https://demo.bokehplots.com/apps/export_csv. Click on the "Experience" column to sort by Experience. Then move the slider on the left hande side that defines the salary range.
* actual behavior: the data is not sorted anymore by Experience
* expected behavior: the table stays sorted by Experience when you move the slider

This is a copy of paste of the example at https://demo.bokehplots.com/apps/export_csv except with 10 data points.

```py
from os.path import dirname, join

import pandas as pd

from bokeh.models import ColumnDataSource, CustomJS
from bokeh.models.widgets import RangeSlider, Button, DataTable, TableColumn, NumberFormatter
from bokeh.io import curdoc

df = pd.read_csv(join(dirname(__file__), 'salary_data.csv')).iloc[:10]

source = ColumnDataSource(data=dict())

def update():
    current = df[(df['salary'] >= slider.value[0]) & (df['salary'] <= slider.value[1])].dropna()
    source.data = {
        'name'             : current.name,
        'salary'           : current.salary,
        'years_experience' : current.years_experience,
    }

slider = RangeSlider(title="Max Salary", start=10000, end=110000, value=(10000, 50000), step=1000, format="0,0")
slider.on_change('value', lambda attr, old, new: update())

button = Button(label="Download", button_type="success")
button.callback = CustomJS(args=dict(source=source),
                           code=open(join(dirname(__file__), "download.js")).read())

columns = [
    TableColumn(field="name", title="Employee Name"),
    TableColumn(field="salary", title="Income", formatter=NumberFormatter(format="$0,0.00")),
    TableColumn(field="years_experience", title="Experience (years)")
]

data_table = DataTable(source=source, columns=columns, width=800)

controls = widgetbox(slider, button)
table = widgetbox(data_table)

curdoc().add_root(row(controls, table))
curdoc().title = "Export CSV"

update()
```

Side-comment: I noticed that in a dask context, I wanted the workers sorted by CPU usage (similar to htop) in the dask dashboard but the sorting disappears on the first data update.

bryevdv:
- [x] issues: fixes #7116
- [x] issues: fixes #7417
- [x] issues: fixes #8021
- [x] issues: fixes #8201 
- [x] issues: fixes #8040
- [x] tests added / passed

This PR cleans up the data model around DataTables:

* data updates, including patch and stream, do not result in an unnecessary full update back to the server
* as a collateral result, DataTables do not now try to send back Typed Arrays, in particular
* cell editors use an efficient patch call to update underlying table date sources

   *UPDATE: for now, also set `.data` to goose update msg back to server*
* updating the grid always re-sorts the columns according to the latest sort
* filtered data sources are handled correctly by cell editors
* selenium tests are added

Things that are now under test:

* #7116

    test_server_source_patch_does_not_duplicate_data_update_event
    test_server_source_stream_does_not_duplicate_data_update_event
    test_server_source_update_does_not_duplicate_data_update_event
    test_server_edit_does_not_duplicate_data_update_event

* #8021 

    test_glyph_selection_updates_table

* #8201

    test_server_sorted_after_data_update

* general DataTable tests:

    test_server_basic_selection 
    test_server_basic_mulitselection 
    test_server_source_updated_after_edit
    test_server_source_callback_triggered_after_edit 

These tests about auto-maintaing sorting (#8201) are present but skipped (they work on all browsers I have tried locally but somer magic incantation is missing to have them work properly on selenium):

test_server_sorted_after_patch
test_server_sorted_after_stream 
test_server_sorted_after_edit

## Future work:

in order to make sure edits propogate changes and change events to the server, `setField` currently does a `source.data = data` type operation after the patch. A much bette thing would be to send a real patch event back to the server. That is better left to a separate PR, but the tests put in place for this PR should/must continue to work unchanged, which will help when that work is done. 

**UPDATE** this was also done

cc @mattpap


bryevdv:
I think the code paths were to support cell editing, but I'd say that the concerns are in the wrong place. Ideally, a cell editor might result in a call to patch on the data source. That would be appropriate and efficient since a cell editor only edits one cell. Then we would on need to make sure that calling patch from BokehJS *does* trigger a server update, and that table views respond appropriately to the path (instead of trying to update the data the way they currently do)

bryevdv:
patching actually seems easy to do, and works well. I will udpate the PR soon.

bryevdv:
OK I think this now also fixes #8040 and #8021 

@mattpap I have made `DataTable` now only accept `ColumnDataSource` (i.e not `ColumnarDataSource`) in order to support patching from cell editors. Is this a practical issue? I don't ever recall promising e.g. that `GeoJsonDataSource` would work with tables.

Additionally, I changed things to update selection properties and not replace `Selection` objects wholesale. Fairly sure this gets us into lots of trouble. I'd like to ensure we never do this in the codebase and can then make clear guarantees about what the "right" thing to watch for changes is in callbacks. That said, the process right now is kludgy, I will clean it up behind a (tested) API.

I will also add more selenium tests for all of these issues and as many other common use cases as i can think ok.

mattpap:
> Is this a practical issue?

I don't think so. I wouldn't concern ourselves at this point with such hypothetical scenarios.

> (...) changed things to update selection properties and not replace Selection objects wholesale. 

In fact, I've been considering making properties holding `Selection` objects readonly (write once). That would make sure that illegal usage is caught quickly.

bryevdv:
`readonly` seems like a good idea, I will do that here as well and summarize everything in the release notes.

bryevdv:
Now tables never need to set `source.data = data` themselves, which also happens to fix #7417 

This PR is kind of a jumble (because the code and issues are an intertwined jumble) I want to keep adding more selenium tests before merging this.

bryevdv:
OK one more push to also fix #8201 before grinding out tests

bryevdv:
I may add some additional data table examples, but the implementation for this is done, unless I run into any unexpected big issues.

bryevdv:
Update: after looking a little bit the "Future Work" was straightforward to implement (including a test to maintain no full patch updates for edits) as well as nicely valuable, so I went ahead and added it to this PR. Patch/Stream operations on the client now actually generate patch/stream event messages back to the server. As part of this, I also split up `document.ts` similar to the python side.

philippjfr:
I played with a bunch of tables today and the issues with editing and updating selections appear to be gone. I've also played with using patch events to communicate drawing actions inside the drawing tools, which also worked great. So from my perspective this is ready.

PR #8317, opened 2018-10-11
Only deduplicate some tool types

### LINKED ISSUE #8074: Overzealous deduplication of tools for merge_tools
Bokeh has a handy `merge_tools` option, which merges all tools from across plots into a single toolbar, removing any resulting duplicate tools.  That's great, but because it both merges the toolbars and removes duplicate tools, it can't be used when multiple instances of the same tool have been added. For instance, in the left-hand plot here there are multiple PolygonDraw tools:

image

but with `merge_tools=True`, only one of them is retained in the final plot:

image

Is there any way that Bokeh could detect that the two original tools are *not* duplicates, and in fact have been configured differently and thus shouldn't be combined together?  Or maybe some tools like the drawing tools should just never be combined when merging?  I'm not sure of precisely the correct solution, but it seems like the current behavior could be improved.

philippjfr:
Bokeh helpfully supports the option of deduplicating tools in a layout by creating tool proxies for any tool type that is added more than once. This is appropriate for tool types such as pan/zoom/save/help etc. but not for others such as the drawing tools.

For now this PR explicitly skips multi-gesture tools which is the ``event_type`` used by the drawing tools and ``CustomAction`` tools. This may not be the best approach and/or may not cover all the tool types that should be skipped during deduplication. 

I don't think you will generally want multiple navigation tools (BoxZoomTool, PanTool, WheelZoomTool, WheelPanTool, ZoomInTool), select tools (BoxSelectTool, LassoSelectTool, PolySelectTool , TapTool), inspection tools (HoverTool, CrosshairTool), or action tools (SaveTool, UndoTool, RedoTool, ResetTool). The only other tool I can imagine wanting multiple entries for is the ``CustomActionTool`` since each instance can trigger a different action and can have a different tooltip and icon.

Therefore the current check I'm doing to skip deduplication is ``(event_type == 'multi' || tool_type == "CustomAction")``. If someone can suggest a more principled way of determining which tools not to deduplicate I'd be happy to hear it.

- [x] issues: fixes #8074
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/tools/toolbar_box.ts
+++ b/bokehjs/src/lib/models/tools/toolbar_box.ts
@@ -145,7 +145,7 @@ export class ProxyToolbar extends ToolbarBase {
         const tools = gestures[event_type][tool_type]
 
         if (tools.length > 0) {
-          if (event_type == 'multi') {
+          if (event_type == 'multi' || tool_type == "CustomAction") {
             for (const tool of tools) {
               const proxy = make_proxy([tool])
               gesture.tools.push(proxy as any)
--- a/bokehjs/src/lib/models/tools/toolbar_box.ts
+++ b/bokehjs/src/lib/models/tools/toolbar_box.ts
@@ -145,9 +145,17 @@ export class ProxyToolbar extends ToolbarBase {
         const tools = gestures[event_type][tool_type]
 
         if (tools.length > 0) {
-          const proxy = make_proxy(tools)
-          gesture.tools.push(proxy as any)
-          this.connect(proxy.properties.active.change, this._active_change.bind(this, proxy))
+          if (event_type == 'multi') {
+            for (const tool of tools) {
+              const proxy = make_proxy([tool])
+              gesture.tools.push(proxy as any)
+              this.connect(proxy.properties.active.change, this._active_change.bind(this, proxy))
+            }
+          } else {
+            const proxy = make_proxy(tools)
+            gesture.tools.push(proxy as any)
+            this.connect(proxy.properties.active.change, this._active_change.bind(this, proxy))
+          }
         }
       }
     }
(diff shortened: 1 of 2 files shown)


jbednar:
I think your list of tools to deduplicate seems fine for now, though I could imagine there being support for multiple selection tools someday, each e.g. selecting in a different color in a way that is reflected across multiple plots.  But I assume that if something like that is ever implemented, the appropriate clause can be added to your test above so that those tools avoid deduplication.

bryevdv:
This seems reasonable to me for now.

philippjfr:
Okay thanks (and sorry for the wrong tag), I'll add some tests now then.

philippjfr:
I'll also fix up these existing tests that were commented out:

```coffeescript
describe "ToolbarBox", ->

  it "should be an instance of LayoutDOM", ->
    box = new ToolbarBox()
    expect(box).to.be.an.instanceof(LayoutDOM)

  ### TODO
  it "should correctly merge multiple actions", ->
    reset1 = new ResetTool()
    reset2 = new ResetTool()
    save1 = new SaveTool()
    save2 = new SaveTool()
    box = new ToolbarBox({tools: [reset1, reset2, save1, save2]})
    expect(box._toolbar.actions.length).equal 2

  it "should correctly merge multiple inspectors", ->
    hover1 = new HoverTool()
    hover2 = new HoverTool()
    crosshair1 = new CrosshairTool()
    crosshair2 = new CrosshairTool()
    box = new ToolbarBox({tools: [hover1, hover2, crosshair1, crosshair2]})
    expect(box._toolbar.inspectors.length).equal 2
  ###
```

bryevdv:
@philippjfr can you merge/rebase on master?

philippjfr:
Just did, there was something odd going on with table tests.

PR #8318, opened 2018-10-11
Autohide toolbar

### LINKED ISSUE #8284: Autohide toolbar
Right now, when a Bokeh plot has a toolbar, the toolbar is always visible.  This helps people see the tools that are available, but it is distracting when one is primarily reading through a long document, looking at several plots and not wishing to interact with any one of them.  

To support this usage, I'd prefer that the default behavior is that the tools are visible only when the mouse is actually inside the plot area, and otherwise they are silently hidden (which would then also apply on export, providing a cleaner exported figure). Presumably the tools could be set to be invisible (alpha=0) whenever the mouse is outside the plot area?  

An implementation of this would be to add an option `autohide` to `Toolbar`, and once supported we could debate whether it should be True or False by default.

jsignell:
- [x] issues: fixes #8284
- [x] tests added / passed - need help

### Demo
autohide

### To see it working - if you have built the js for this branch:

```html
<!doctype html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Complete Example</title>
<link rel="stylesheet" href="./bokehjs/build/css/bokeh.min.css" type="text/css" />
<script type="text/javascript" src="./bokehjs/build/js/bokeh.min.js"></script>
<script type="text/javascript" src="./bokehjs/build/js/bokeh-api.min.js"></script>
<!-- The order of CSS and JS imports above is important. -->
</head>

<body>

<div>

// arrays to hold data
var source = new Bokeh.ColumnDataSource({
    data: { x: [1,2,3,4], y: [0,5,7,2] }
});

// make the plot and add some tools
var tools = "pan,crosshair,wheel_zoom,box_zoom,reset,save";

var plot = Bokeh.Plotting.figure({title:'Example of Random data', tools: tools, height: 300, width: 300});

plot.toolbar.autohide = true;

var scatterData = plot.line({ field: "x" }, { field: "y" }, {
    source: source,
    line_width: 2
});

// Show the plot, appending it to the end of the current
// section of the document we are in.
Bokeh.Plotting.show(plot,document.currentScript.parentElement);

</script>
</div>
</body>
</html>

### CODE REVIEW

**Comment by philippjfr** [comment]
This is really great! I wonder if we could achieve a smoother transition by declaring CSS classes with transition attributes, something like (my CSS skills are horrible so apologies if this doesn't work):

```css
.bk-toolbar-hidden {
    visibility:hidden;
    opacity:0;
    transition:visibility 0.3s linear,opacity 0.3s linear;
}

.bk-toolbar-visible {
    visibility:visible;
    opacity:1;
}
```

jsignell:
Not sure if I have hooked up the python correctly and whether the functionality should live under `toolbar_options`, or up a level and be called something like `toolbar_autohide` (following the `toolbar_location` pattern).

bokehjs/test/models/tools/toolbar.coffee
```diff
@@ -46,6 +46,16 @@ describe "Toolbar", ->
...
 
+    it "should have autohide as False by default", ->
+      toolbar = new Toolbar({tools:[@hover_1, @hover_2, @hover_3]})
+      expect(toolbar.autohide).to.be.false
+      expect(toolbar.visibility).to.equal("visible")
+
+    it "should have autohide as False by default", ->
+      toolbar = new Toolbar({tools:[@hover_1, @hover_2, @hover_3], autohide=true})
+      expect(toolbar.autohide).to.be.true
+      expect(toolbar.visibility).to.equal("hidden")
```
jsignell: need to add a test where given a certain screen positition of the cursor, we see whether the toolbars is hidden or visible

bryevdv:
@jsignell the example should go under `examples/plotting/file` or perhaps `examples/howto`. The `examples/model` directory is for things that don't use `bokeh.plotting` (i.e. start from a bare `Plot`)

jsignell:
> the example should go under examples/plotting/file
sounds good.

@bryevdv - do you have suggestions for adding to the tests?

bryevdv:
FYI @jsignell if you expect to make more PRs in the next 6-12 month we should add you back to @bokeh/dev I recently cleared the group out some in accordance with the new BEP-4

jbednar:
I'd be happy to have her on @bokeh/dev, as I do expect she'll be making more Bokeh PRs (e.g. she's about to tackle "polygons with holes"...)

bryevdv:
sure, per BEP-4, users must also affirm their own interest/willingness to be added :)

bokehjs/test/models/tools/toolbar.coffee
```diff
@@ -46,6 +46,16 @@ describe "Toolbar", ->
       expect(@hover_2.active).to.be.false
       expect(@hover_3.active).to.be.false
 
+    it "should have autohide as False by default", ->
+      toolbar = new Toolbar({tools:[@hover_1, @hover_2, @hover_3]})
+      expect(toolbar.autohide).to.be.false
+      expect(toolbar.visible).to.be.true
+
+    it "should have autohide as False by default", ->
```
bryevdv: is this test name intended to be identical to the previous one?

bokehjs/src/lib/models/plots/plot_canvas.ts
```diff
@@ -653,6 +657,7 @@ export class PlotCanvasView extends DOMView {
 
     this.connect(this.model.plot.properties.renderers.change, () => this.build_levels())
     this.connect(this.model.plot.toolbar.properties.tools.change, () => { this.build_levels(); this.build_tools() })
+    this.connect(this.model.plot.toolbar.properties.visible.change, () => { this.build_levels(); this.build_tools() })
```
bryevdv: What is the need to rebuild the tools here? Does a request to render not suffice?

bokehjs/src/lib/models/tools/toolbar_base.ts
```diff
@@ -120,8 +136,10 @@ export class ToolbarBase extends Model {
     this.prototype.default_view = ToolbarBaseView
 
     this.define({
-      tools: [ p.Array,    []       ],
-      logo:  [ p.String,   'normal' ], // TODO (bev)
+      tools:      [ p.Array,      []       ],
+      logo:       [ p.String,     'normal' ], // TODO (bev)
+      visible:    [ p.Bool,       true ],
+      autohide:   [ p.Bool,       false ],
```
bryevdv: Prefer to align the closing braces too

bryevdv:
@jsignell this looks great in action. I only had a few final minor comments.

philippjfr:
What does the CSS transition look like? (If it's too much of a hassle to make a quick gif I'll just check it out and try for myself).

bryevdv:
> What does the CSS transition look like? (If it's too much of a hassle to make a quick gif I'll just check it out and try for myself).

screenflow

philippjfr:
Woah, thanks, I love it!

jsignell:
@bryevdv, thanks for the review! I'll make the little changes and see if I can get a selenium test working.

I'd be happy to be added back to bokeh dev. I just noticed on this PR that I didn't have it anymore :)

bryevdv:
@jsignell invite pending!

For selenium tests, here is a good example to start from: 

https://github.com/bokeh/bokeh/blob/master/tests/integration/tools/test_tap_tool.py

bokehjs/src/lib/models/plots/plot_canvas.ts
jsignell: I was pretty all over the place at the beginning of this PR. This isn't needed.

bokehjs/src/lib/models/tools/toolbar_base.ts
jsignell: done

bokehjs/test/models/tools/toolbar.coffee
jsignell: no...

bryevdv:
There seems to be some issue running the codebase tests from a fork:
```
~/build/bokeh/bokeh/bokehjs ~/build/bokeh/bokeh
+node make scripts:compile --emit-error
[17:22:50] Starting 'scripts:ts'...
[17:23:02] Finished 'scripts:ts' after 12.44 s
[17:23:02] Finished 'scripts:compile'
+node make test:compile --emit-error
[17:23:06] Starting 'test:compile'...
[17:23:15] failed: unexpected =
```
Not sure what is going on there yet, cc @mattpap

bryevdv:
Another possible addition: A short docs section in the users guide:

https://github.com/bokeh/bokeh/blob/master/sphinx/source/docs/user_guide/styling.rst

bryevdv:
@jsignell I pointed out the TS syntax error in the test above, other failures are spurious/intermittent

@jbednar this is why I would suggest `authohide=True` as a good default for HV, but not Bokeh

bryevdv:
@jsignell I went ahead and fixed the syntax issue on the GH UI, I'd like to merge this and get it in a dev build

jbednar:
> @jbednar this is why I would suggest authohide=True as a good default for HV, but not Bokeh

I'm happy to have autohide by default for HV immediately, but I also think it's a good choice for Bokeh, just not immediately; only after resolving any issues like how to generate the website...

jsignell:
> Have you tested this with Bokeh PNG export? 

I don't see any tools *ever* (autohide=True and autohide=False) with bokeh PNG exports. It wouldn't make any sense to include them since they wouldn't work.

jbednar:
The last I recall from several months back, I was always getting tools on a PNG export, which was not what I wanted, but I'm happy to hear that doesn't happen now or that I was somehow confused.

bryevdv:
@jsignell now that the JS tests are running they report a property mismatch:
```
  1438 passing (10s)
  8 pending
  2 failing
  1) Toolbar
       _init_tools method
         should have set visible to false when autohide is true:
      AssertionError: expected true to be false
      + expected - actual
      -true
      +false
      
      at Context.<anonymous> (build/test/models/tools/toolbar.js:88:43)
  2) Defaults
       match between Python and bokehjs:
      AssertionError: expected 2 to equal 0
      + expected - actual
      -2
      +0
      
      at Context.<anonymous> (build/test/defaults.js:231:38)

```

bryevdv:
Better info:
```
GMapPlot.toolbar: python is missing some properties found in bokehjs
    GMapPlot.toolbar.visible: bokehjs defaults to true but python has no such property
GMapPlot.x_range: defaults are out of sync between Python and bokehjs
    GMapPlot.x_range.reset_start: bokehjs defaults to 0 but python defaults to null
    GMapPlot.x_range.reset_end: bokehjs defaults to 1 but python defaults to null
GMapPlot.y_range: defaults are out of sync between Python and bokehjs
    GMapPlot.y_range.reset_start: bokehjs defaults to 0 but python defaults to null
    GMapPlot.y_range.reset_end: bokehjs defaults to 1 but python defaults to null
[bokeh] LogTickFormatter not configured with a ticker, using default base of 10 (labels will be incorrect if ticker base is not 10)
ProxyToolbar: python is missing some properties found in bokehjs
    ProxyToolbar.visible: bokehjs defaults to true but python has no such property
ProxyToolbar
[]
Toolbar: python is missing some properties found in bokehjs
    Toolbar.visible: bokehjs defaults to true but python has no such property
Toolbar
[]
Python/bokehjs matching defaults problems: 2
``` 
Either visible needs to be added to `ToolbarBase` on the python side, or it needs to be marked `this.internal` on the JS side.

jsignell:
I don't think `visible` should be exposed on the python side. So I will make it internal.

jsignell:
I'm getting 

```
  2) Toolbar
       _init_tools method
         should have set visible to false when autohide is true:

      AssertionError: expected true to be false
      + expected - actual

      -true
      +false

      at Context.<anonymous> (build/test/models/tools/toolbar.js:88:43)

**Comment by jsignell** [comment]
Ah I see you have that above. Any idea why that test doesn't work?

File: bokehjs/src/less/toolbar.less [hunk 5]
```diff
@@ -10,6 +10,12 @@
   padding: 0;
 }
 
+.bk-toolbar-hidden {
+    visibility: hidden;
+    opacity: 0;
+    transition: visibility 0.3s linear,opacity 0.3s linear;
```
mattpap: Bad indentation (should be 2 spaces).

bokehjs/src/lib/core/ui_events.ts
```diff
@@ -262,6 +266,11 @@ export class UIEvents implements EventListenerObject {
 
         const active_inspectors = this.toolbar.inspectors.filter(t => t.active)
         let cursor = "default"
+        let toolbar_is_visible: boolean =  false
```
mattpap: Type annotation isn't necessary for local variables with primitive types. Also, bad whitespace.

bokehjs/src/lib/models/plots/plot_canvas.ts
```diff
@@ -221,6 +221,10 @@ export class PlotCanvasView extends DOMView {
     this.canvas_view.el.style.cursor = cursor
   }
 
+  set_toolbar_visibility(visible: boolean = true): void {
+    this.model.toolbar.visible = this.model.toolbar.autohide ? visible : true;
```
mattpap: We don't use semicolons in typescript.

bokehjs/src/lib/models/tools/toolbar_base.ts
```diff
@@ -41,11 +43,22 @@ export class ToolbarBaseView extends DOMView {
     build_views(this._tool_button_views, tools, {parent: this}, (tool) => tool.button_view)
   }
 
+  protected _set_visibility(): void {
+    const hidden_class = "bk-toolbar-hidden"
+
+    if (this.el.classList.contains(hidden_class) && (this.model.visible)) {
```
mattpap: Unnecessary parenthesis.

bokehjs/test/models/tools/toolbar.coffee
```diff
@@ -46,6 +46,16 @@ describe "Toolbar", ->
       expect(@hover_2.active).to.be.false
       expect(@hover_3.active).to.be.false
 
+    it "should have autohide as False by default and set visible to false", ->
+      toolbar = new Toolbar({tools:[@hover_1, @hover_2, @hover_3]})
```
mattpap: Missing whitespace after colon. Same below.

bokehjs/test/models/tools/toolbar.coffee
```diff
@@ -46,6 +46,16 @@ describe "Toolbar", ->
       expect(@hover_2.active).to.be.false
       expect(@hover_3.active).to.be.false
 
+    it "should have autohide as False by default and set visible to false", ->
```
bryevdv: This test seems mis-named (visible is True in this case)

bryevdv:
> Ah I see you have that above. Any idea why that test doesn't work?

I don't offhand, actually, but I will try to investigate some.

examples/plotting/file/toolbar_autohide.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+from bokeh.layouts import row, column
```
bryevdv: `column` is unused (failing linter test)

jsignell:
Actually @bryevdv I think it is because visible has a default value that doesn't get over-ridden until the view gets initialized here https://github.com/jsignell/bokeh/blob/autohide/bokehjs/src/lib/models/tools/toolbar_base.ts#L25

mattpap:
Based on the discussion in the issue, I thought the autohidden toolbar shouldn't participate in layout in full, i.e. only for location purpose, but shouldn't occupy any space in the dimension orthogonal to the dominant dimension. Given this PR doesn't make any changes to toolbar layout, then there will be space visible when toolbar is hidden, which will be easily visible when plot and document backgrounds differ in color.

bryevdv:
> I don't think that's practical -- there then won't be any space for the toolbar when it's visible, right? I don't want the tools visible, but it still seems like there needs to be space for them to be visible regardless.

Right, in case it is not evident, the expressly *desired behaviour* is that the toolbar simply wink in and out of visibility, changing nothing else (the plot should *not* resize itself into the space occupied by a now-invisible toolbar, for instance, nor should content outside the plot next to the toolbar shift over) What is in the GIF above is exactly the behavior that is wanted. Different background/plot colors is not really relevant, that's always an issue with any plot with a border whether there is a toolbar or not. It's up to users to contend with that however is appropriate for their use case. 

@jsignell is there a reason to defer setting `visible` until the view? It seems it could be done on the model init, which presumably would also fix the test.

jsignell:
> @jsignell is there a reason to defer setting visible until the view? It seems it could be done on the model init, which presumably would also fix the test.

I just couldn't figure out how to set it on the model using autohide which is also set on the model.

mattpap:
> I don't think that's practical -- there then won't be any space for the toolbar when it's visible, right? I don't want the tools visible, but it still seems like there needs to be space for them to be visible regardless.

In my view there are two states. First the current state of affairs (or `autohide=False`), where toolbar fully participates in layout. If toolbar is programmatically hidden, then layout is recomputed. The other state is the new feature  (or `autohide=True`), where toolbar participates in layout essentially only for positioning, allocating 0 pixels in the orthogonal dimension. When toolbar is showed due to user interaction, then the toolbar is displayed floating (I think this is what plotly does, but I may be mistaken).

bryevdv:
@mattpap IIUC what you are saying is that you would like it to be the case that if (in the future) a user sets `toolbar.visible = False` than *that* explicit operation should result in the toolbar not taking up space, correct? If, so;

* I think that's fine and reasonable, but also
* I think it is not precluded by this work, since it does not expose `visible` to users at all

we can still make an explicit `toolbar.visible = False` action "remove space" in future work, so I propose to merge this as it is.

mattpap:
> Different background/plot colors is not really relevant, that's always an issue with any plot with a border whether there is a toolbar or not.

Another way to look at this is to use `export_png()`, which crops HTML page to plot dimensions, which will produce an image with 30px more space than necessary in place where toolbar was allocated by layout. I'm pretty sure the first user that exports an image with `autohide=True` and the current design, will complain that the result is far from what he would expect.

bryevdv:
> Another way to look at this is to use export_png(), which crops HTML page to plot dimensions, which will produce an image with 30px more space than necessary in place where toolbar was allocated by layout. I'm pretty sure the first user that exports an image with autohide=True and the current design, will complain that the result is far from what he would expect.

Then (in the future) they can explicitly set `toolbar.visble = False`

If I wanted to suggest a full possible set of states (for the future) I'd posit:
```
visible = False # set by user, toolbar gone and takes up no space, autohide irrelevant
visible = True # set by user, toolbar always present autohide irrelevant

visible = "auto" # default, shows based on autohide value, but always takes up space
```


```

File: bokehjs/src/lib/core/ui_events.ts [hunk 12]
```diff
@@ -262,6 +266,11 @@ export class UIEvents implements EventListenerObject {
 
         const active_inspectors = this.toolbar.inspectors.filter(t => t.active)
         let cursor = "default"
+        let toolbar_is_visible = false
+
+        if (this._hit_test_canvas(e.sx, e.sy)) {
+          toolbar_is_visible = true
+        }
```
mattpap: Is there a reason for this kind of code instead of just `const toolbar_is_visible = this._hit_test_canvas(e.sx, e.sy)`, which is shorter and doesn't introduce a mutable variable?

bryevdv:
@mattpap please signal if you agree with https://github.com/bokeh/bokeh/pull/8318#issuecomment-429132138 If not, let's continue this discussion back on the issue, rather than the PR.

mattpap:
> please signal if you agree with #8318 (comment)

I think I don't, but I need to properly read through it and give it thought, and it's a bit late in the day (night) for that. I will let you know tomorrow.

bokehjs/src/lib/models/tools/toolbar_base.ts
```diff
@@ -120,8 +136,9 @@ export class ToolbarBase extends Model {
     this.prototype.default_view = ToolbarBaseView
 
     this.define({
-      tools: [ p.Array,    []       ],
-      logo:  [ p.String,   'normal' ], // TODO (bev)
+      tools:      [ p.Array,      []       ],
+      logo:       [ p.String,     'normal' ], // TODO (bev)
+      autohide:   [ p.Bool,       false    ],
```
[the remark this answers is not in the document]
bryevdv: Personally I am willing to stipulate that autohide does not apply to grid plot toolbars, tho I don't see why the proxy toolbar could not delegate to all the merged toolbars (e.g. if they all have `autohide=True` then the merged toolbar does too).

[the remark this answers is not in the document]
jbednar: Delegating sounds reasonable to me.

jsignell:
I think this is done unless @mattpap has any comments.

jsignell:
There is still a minor issue where the toolbar remains visible if you let your mouse drift slowly off the edge, but I showed it to @bryevdv and he agreed that it shouldn't be blocking.

slowly

mattpap:
Given that everyone except me agrees on this design, I'm not going to block this.

mattpap:
> There is still a minor issue where the toolbar remains visible if you let your mouse drift slowly off the edge (...)

This doesn't seem that minor. Let's have this fixed now, because otherwise we will end up with another issue in the "short term" milestone.

bryevdv:
> This doesn't seem that minor. Let's have this fixed now, because otherwise we will end up with another issue in the "short term" milestone.
 
Having tried things out locally, you have to move extremely and unnaturally slowly to have this happen. I agree for the need to make another issue, but what is important at the moment is getting in the functionality (working, for any realistic usage) and the API. We do not have the luxury of time at the moment.

PR #8321, opened 2018-10-12
Add adapter property to AjaxDataSource

bryevdv:
This small PR adds an `adapter` property to `AjaxDataSource` that accepts a `customJS` that will convert any arbitrary JSON response into Bokeh CDS data format. The `ajax_source.py` is updated and simplified to demonstrate this new fearture.

### DIFF
--- a/bokeh/models/sources.py
+++ b/bokeh/models/sources.py
@@ -9,7 +9,7 @@ from ..util.dependencies import import_optional
 from ..util.serialization import convert_datetime_array
 from ..util.warnings import BokehUserWarning
 
-from .callbacks import Callback
+from .callbacks import Callback, CustomJS
 from .filters import Filter
 from .selections import Selection, SelectionPolicy, UnionRenderers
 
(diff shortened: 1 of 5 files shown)


bryevdv:
cc @bokeh/dev Although this change makes `AjaxDataSource` vastly more useful, it only required a small, isolated batch of work. I think things should be uncontroversial, and I'd like to merge this tonight, but any quick comments are welcome.

bokeh/models/sources.py
```diff
@@ -708,6 +716,18 @@ class AjaxDataSource(RemoteSource):
     replace existing data entirely.
     """)
 
+    adapter = Instance(CustomJS, help="""
+    A JavaScript callback to adapt raw JSON responses to Bokeh ColumnDataSource
+    format.
+
+    If provided, this callback is executes imediately after the JSON data is
+    received, but before apprending or replacing data in the data source. The
+    CustomJS callback will receiece the AjaxDataSource as ``cb_obj`` and will
```
canavandl: "receiece" (sp)

bokeh/models/sources.py
```diff
@@ -708,6 +716,18 @@ class AjaxDataSource(RemoteSource):
     replace existing data entirely.
     """)
 
+    adapter = Instance(CustomJS, help="""
+    A JavaScript callback to adapt raw JSON responses to Bokeh ColumnDataSource
+    format.
+
+    If provided, this callback is executes imediately after the JSON data is
```
canavandl: "imediately" (sp)

Issue #8319, opened 2018-10-12
Examples Punch List


bryevdv:
Our examples are a bit long in the tooth. There are *lots* more improvements and additions that can be made, but at a minimum these should be passed over for changes describe below:

# TO-DO

## howto

* categorical: pandas deprecation, add "sorting factors" example

## plotting

* add sorted bar example
* add grouped and stacked example

# Done

## apps

* ~~~gapminder: has pandas deprecation issues~~~ #8322
* ~~~pivot: complicated and unmaintained, remove?~~~
* ~~~populatation: legend position bad~~~ #8326
* ~~~selection_histogram: needs update for selection changes~~~ #8325
* ~~~simple_hdf5: install h5py with test deps, more interesting/attractive, auto generate data~~~ #8328

## howto

* ~~~linked panning: add explanation text~~~
* ~~~hover callback: add explanation text~~~
* ~~~prevent_scrolling: LOD off when can't pan?~~~
* ~~~hold_app: better title, remove toolbar~~~
* ~~~range_updates: clean up JS, in general~~~
* ~~~notebook embed: add comment about setting notebook url~~~
* ~~~marriage/divorces nb: bad responsive param, remove notebook or .py example? rename?~~~
* ~~~ipywidgets: add ipywidgets to test deps install, make prettier~~~
* ~~~dashboard.py: deprecation~~~
* ~~~gunicorn embed: add gunicorn test dep~~~

## custom

~~~need READMEs~~~
~~~add other examples from docs~~~

## embed

* ~~~autoload_static: change default port~~~
* ~~~embed_multiple: better arrangement~~~ #8331
* ~~~embed themed: more interesting~~~
* ~~~json_item: run `bokeh static` automatically~~~
* ~~~simple: remove? improve?~~~
* ~~~arguments: readme, shell script to start, reverts to sin batch on slider use~~~

## models

* ~~~grid: prettier~~~
* ~~~dateaxis: more interesting~~~
* ~~~custom: add explanation~~~
* ~~~choropleth: colorbar~~~
* ~~~anscombe: add text explanation~~~
* ~~~colorspec.py remove~~~
* ~~~customjs: title instruct to click~~~
* ~~~daylight: styling~~~
* ~~~external_resources: rename? better styling~~~
* ~~~glyphs: add new markers~~~
* ~~~guages: no border/title~~~
* ~~~glyph1: remove/rename~~~
* ~~~imageurl: match aspect? or note data extents in explanation~~~
[...]

PR #8322, opened 2018-10-13
Pandas warning removed in Gapminder example

xavArtley:
#8319 :

> gapminder: has pandas deprecation issues


bryevdv:
Thanks @xavArtley I wasn't sure how to get rid of `Panel` so this is very helpful

PR #8324, opened 2018-10-13
Compat: Get correct env when freezing with cx_Freeze

### LINKED ISSUE #8323: Compat: Get correct env when freezing with cx_Freeze
Anaconda Python 3.6 bokeh 0.13.0
Pip cx-Freeze 5.1.1
OS Windows

#### Expected behaviour
Freezing with cx-Freeze works for bokeh.

#### Actual behaviour
Exception is raised
```
AttributeError: module 'sys' has no attribute '_MEIPASS'
```

#### Bug description
When freezing with cx_Freeze `sys.frozen` is `True`. It does not define `sys._MEIPASS` though. `sys._MEIPASS` is only used by the PyInstaller freezer.

I prepared a patch in #8324.

fleimgruber:
Fixes #8323

### DIFF
--- a/bokeh/core/templates.py
+++ b/bokeh/core/templates.py
@@ -26,8 +26,10 @@ def get_env():
     ''' Get the correct Jinja2 Environment, also for frozen scripts.
     '''
     if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
+        # PyInstaller uses _MEIPASS and only works with jinja2.FileSystemLoader
         templates_path = join(sys._MEIPASS, 'bokeh', 'core', '_templates')
     else:
+        # Non-frozen Python and cx_Freeze can use __file__ directly
         templates_path = join(dirname(__file__), '_templates')
 
     return Environment(loader=FileSystemLoader(templates_path))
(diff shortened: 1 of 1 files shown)


bryevdv:
@fleimgruber is the import of `bokeh.core` necessary? It seems that 
```
templates_path = join(dirname(__file__), '_templates')
``` 
ought to work in all cases when `sys._MEIPASS` is no present?

fleimgruber:
See description in #8323: The issue is that the attribute is accessed and exception is raised.

bryevdv:
> Ah, just saw your edit... ok I will try with that change. 

Sorry I'm on the site all the time, so I forget many folks interact with GH messages through the original emails the GH sends out, makes me terrible about stealth edits. :)

bryevdv:
@fleimgruber do you have the ability to quickly test with the other freeze tool as well? I've never used either so it's a bit or a learning curve for me to even try

bryevdv:
In any case, since there are not automated tests, it would be advised to add a comment about the conditional to explain what the different cases are for.

bryevdv:
@axil can you confirm that this change to support cx_Freeze does not disturb the pyinstaller use case?

bryevdv:
OK there is a separate (new) issue with pyinstaller not loading `bokeh.themes` automatically, which it probably needs to do as of recent developements. But after copying `bokeh.themes` in manually, the bundle worked, i.e. it was able to find and load the templates correctly. Since that is all the scope of this issue is concerned with, I will go ahead and merge.

PR #8326, opened 2018-10-13
population example legend position -> "bottom_right"

xavArtley:
#8319

> populatation: legend position bad

### DIFF
--- a/examples/app/population.py
+++ b/examples/app/population.py
@@ -51,8 +51,8 @@ population.line("x", "y", color="violet", line_width=2, line_dash="dashed", sour
 
 population.xaxis.major_label_orientation = pi/4
 population.xgrid.grid_line_color = None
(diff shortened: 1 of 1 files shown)


bryevdv:
How does `top_left` look across a sampling of the data sets?

xavArtley:
image

bryevdv:
Hrm not great, I wonder if we could make extra space for it on the right by explicitly setting the x_range.end to 2200 (or whatever works) then putting the legend `center_right` ?

xavArtley:
image

bryevdv:
Better I think? If the legend orientation is changed to vertical (default) then it can take up less horizontal space

xavArtley:
image

bryevdv:
@xavArtley looks much better! Thanks, merging now

PR #8327, opened 2018-10-14
load themes from python modules instead of json files

bryevdv:
This PR moves the YAML theme definitions into strings in python modules, rather than external files. This is to support better integration with "freeze" tools like cx_freeze and pyinstaller. As a side effect it probably shaves a small amount off general import times 

This PR also pins sphinx to 1.7 as it was noticed 1.8 results in broken static image links.


bryevdv:
cc @axil @fleimgruber FYI

PR #8328, opened 2018-10-14
Simple hdf5 example improvement

xavArtley:
#8319

> simple_hdf5: install h5py with test deps, make more interesting/attractive, auto generate data

I had h5py to examples dependencies instead of tests
I tried to make the example more intersting using a MultiSelect widget and a multi_line glyph
The auto generate data are done by importing create_hdf5.py, not sure if it's the more pythonic way

### DIFF
--- a/examples/app/simple_hdf5/create_hdf5.py
+++ b/examples/app/simple_hdf5/create_hdf5.py
@@ -11,7 +11,11 @@ def generate_data(path):
                      'Exponential': {'options': dict(loc=-0.5, scale=1),
                                      'name': 'expon'},
                      'Chi Square': {'options': dict(loc=-0.5, df=1),
-                                    'name': 'chi2'}
+                                    'name': 'chi2'},
+                     'Alpha': {'options': dict(a=3, loc=-0.5),
+                               'name': 'alpha'},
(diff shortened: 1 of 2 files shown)


bryevdv:
@xavArtley Thanks for looking in to these, a couple of ideas for this one:

* make a function in `create_hdf5.py` that does the work to create the file, then have a "main check" that calls the function if the module is run as as script. Otherwise, the function can be imported and called, e.g. in the app itself, which is more typical usage (I agree it's a little weird to just import the module to have to perform real work)

* I think the original simple line+Select is fine and that we should perhaps keep that. The main purpose of this example is to show how to use hdf5 within a Bokeh app so we should not get too fancy with other things. But the actual data is really uninteresting. It would be nice to generate some more interesting data sets. 

* The visual appearance is also pretty bland. Maybe the plot could get light grey background similar to the `selection_histogram.py` app, with a thicker, more colorful line?

xavArtley:
Concerning datasets I could generate some examples of probabiltity distributions
https://en.wikipedia.org/wiki/List_of_probability_distributions
For example
- Beta distribution
- Chi-squared distribution
- Gamma distribution
- Gaussian distribution
- ...
I don't know if it's more interesting

PR #8329, opened 2018-10-14
add webdriver_control and default to reuse

bryevdv:
Adds a module level object  to control and manage webdriver creation, defaults to re-using selenium webdrivers. 

- [x] tests added / passed


bryevdv:
Not sure what's up with the Appveyor build for this branch. The tests finish and pytest exits but the job does not finish.

PR #8331, opened 2018-10-15
embed_multiple: better arrangement

xavArtley:
#8319 

> embed_multiple: better arrangement


bryevdv:
@xavArtley can you post a screen shot?

xavArtley:
image

PR #8333, opened 2018-10-15
Initial implementation of loading compiled CustomModel implementation from files

### LINKED ISSUE #5345: Custom models distribution without compiler on the server side
Running Bokeh in the cloud environment (like Google App Engine) gives this error:

```
An internal error occurred:
node.js is needed to allow compilation of custom models ("conda install -c bokeh nodejs" or follow https://nodejs.org/en/download/)
```

The setup, it is working fine on my local computer, but on cloud version looks like in the latest version 0.12.3. some extra dependency is added to node.js that was not in the previous version.

It will be nice if the error message is giving more hints about which custom model is creating all the issues. In my case I have several Bokeh plots (and some of them have a custom model like https://github.com/bokeh/bokeh/issues/4690#issuecomment-229694803).

This compilation problem creates challenges for distribution when the package is only managed by pip install -r requirements.txt and no option to add conda install -c bokeh nodejs.

The dependency is very subtle and not well documented.

philippjfr:
This PR adds a new subclass of ``Implementation`` which allows loading a pre-compiled CustomModel implementation from file. The ``Compiled`` implementation takes two arguments, a path or dictionary containing the compiled code in a JSON format, and the original uncompiled ``Implementation``. The code then checks whether the ``hash`` stored inside the compiled JSON matches the hash of the uncompiled code. If the hashes do not match a warning is raised which will suggest the code needs to be recompiled.

Currently the code can be recompiled using the ``compile_models`` method and then saved out to a json file manually, however I'd be happy to add convenience functions around this and even a ``bokeh compile`` entrypoint if that is desirable.

In practice it looks something like this:

```python
class VegaPlot(LayoutDOM):
    """
    A Bokeh model that wraps around a Vega plot and renders it inside
    a Bokeh plot.
    """

    __implementation__ = Compiled(os.path.join(os.path.dirname(__file__), 'compiled', 'panel_vega_VegaPlot.json'),
                                  os.path.join(os.path.dirname(__file__), 'models', 'vega.ts'))
```

and I've used the following code to generate the compiled JSON:

```python
for k, v in compile_models([VegaPlot]).items():
    with open(os.path.join('../../panel/compiled/', k.replace('.', '_')+'.json'), 'w') as f:
        json.dump(v, f)
```

I have also not fully worked out how the bokeh version comes into it. It would be a shame if the compiled code is only useable with a very specific version of bokeh.

- [x] issues: fixes #5345
- [ ] tests added / passed

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -277,22 +277,17 @@ class Compiled(Implementation):
         has been pre-compiled.
     '''
 
-    def __init__(self, compiled, implementation):
-        if isinstance(compiled, six.string_types):
-            path = compiled
-            compiled = None
-        elif not isinstance(compiled, dict):
-            raise ValueError('Compiled model must be defined as a path '
-                             'or a dictionary.')
+    def __init__(self, path, implementation):
+        if not (isinstance(path, six.string_types) and path.endswith('json')):
+            raise ValueError('Compiled implementation must define json '
+                             'file to load implementation from.')
 
-        if compiled is None:
-            if not os.path.isfile(path):
-                self.compiled = None
-            else:
-                with io.open(path, encoding="utf-8") as f:
-                    self.compiled = AttrDict(json.loads(f.read()))
+        if not os.path.isfile(path):
+            self._compiled = None
         else:
-            self.compiled = AttrDict(compiled)
+            with io.open(path, encoding="utf-8") as f:
+                self._compiled = json.loads(f.read())
+        self.file = path
 
         if implementation is None:
             raise ValueError('Compiled code must declare uncompiled code')
@@ -308,12 +303,18 @@ class Compiled(Implementation):
 
     @property
     def outdated(self):
+        if self._compiled is None or __version__ not in self._compiled:
+            return True
         sig = hashlib.sha256(self.implementation.code.encode('utf-8')).hexdigest()
-        return sig != getattr(self.compiled, 'hash', None)
+        return sig != self._compiled.get('hash')
+
+    @property
+    def compiled(self):
+        return AttrDict(self._compiled.get(__version__, {}))
 
     @property
     def code(self):
-        return self.compiled.code
+        return self.compiled.get('code', None)
 
     @property
     def lang(self):
@@ -427,7 +428,7 @@ def _get_custom_models(models):
         return None
     return custom_models
 
-def compile_models(models):
+def _compile_models(models, recompiling=False):
     """Returns the compiled implementation of supplied `models`. """
     custom_models = _get_custom_models(models)
     if custom_models is None:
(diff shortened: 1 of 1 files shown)


philippjfr:
I suppose the JSON file could hold multiple versions of the compiled code indexed by bokeh version.

bryevdv:
> It would be a shame if the compiled code is only useable with a very specific version of bokeh.

We are not yet in a position to make guarantees about BokehJS API stability, unfortunately, so I think versions have to be pinned together, at least for now.

cc @mattpap This is needed for some client work to it would be nice to get in before the buzzer.

bryevdv:
I do think this is to much work for users:
```
for k, v in compile_models([VegaPlot]).items():
    with open(os.path.join('../../panel/compiled/', k.replace('.', '_')+'.json'), 'w') as f:
        json.dump(v, f)
```
so some kind of much simpler user-facing API is needed

philippjfr:
>I do think this is to much work for users:

Definitely, if we say that the compiled code always needs to be loaded from a JSON file (rather than being provided as an inline dictionary), it could be as simple as:

```bash
bokeh compile some_file.py
```

since the ``Compiled`` implementation would be declaring the expected path for the file. It would simply look for all custom models with a ``Compiled`` ``__implementation__`` and then save the compiled code to that file.

philippjfr:
I've now modified the implementation in such a way that, a ``Compiled`` implementation must always reference a JSON file (which does not have to exist). The JSON file has the following structure:

```
{bokeh.__version__: {'code': ..., 'deps': ...}, 'hash': ...}}
```

The compilation will check whether the implementation.json file exists, whether it contains the current bokeh version and whether the hash which is computed from the non-compiled implementation matches. If all of these conditions are met it will use the pre-compiled code.

To generate and/or update the implementation.json files, you call:

```python
from bokeh.util.compiler import save_compiled_models
save_compiled_models(custom_models, append=True)
```

This will compile the models and save it to the JSON file, if the file already exists, ``append=True`` and the hash matches it will simply add the compiled code to the existing JSON bundle otherwise it will overwrite the existing bundle.

If we think this is all a good idea I can add a new entrypoint to allow:

```
bokeh compile some_file.py
```

bryevdv:
I'm less interested in the entry point (though I don't object if it's a priority for you), but these JSON files are on a per-custom-model basis? I think users might want some control over the filename generated for each one

philippjfr:
>I'm less interested in the entry point (though I don't object if it's a priority for you), but these JSON files are on a per-custom-model basis?

That's right, do you think we should allow bundling them also?

>I think users might want some control over the filename generated for each one

Right, in the current approach the filename would be defined by the author of the custom model in the ``Compiled()`` definition. That would allow a bokeh extension library to have a compile step where all the models are compiled against different versions of bokeh and then included in the distributed package. ``python setup.py develop`` would put them in your local copy. But it also means you'd have to dynamically swap out the path on the ``Compiled`` instance on the ``NewModel.__implementation__`` to change the json path, which seems like a hassle.

Instead we could go with a global cache and provide a helper function to load the json files with the compiled code and populate the cache. That way this kind of thing would make sense:

```python
save_custom_model_bundle(custom_models, './bundle.json', append=True)
```

```python
load_custom_model_bundle('./bundle.json')
```

I'm not sure which approach is preferable.

bryevdv:
superceded by #8336

PR #8336, opened 2018-10-16
Allow loading cached compiled models using hook

### LINKED ISSUE #5345: Custom models distribution without compiler on the server side

philippjfr:
Simplified implementation for loading cached compiled models until we settle on a public API. Works by allowing users to set a cache hook which should return the compiled code given the uncompiled implementation. Offers public methods to get and set this hook, available as:

```
bokeh.util.compiler.get_cache_hook
bokeh.util.compiler.set_cache_hook
```

Also splits the compilation code a bit so that it is easier for an external library to pre-compile a model and load it in the cache hook it registers.

- [x] issues: fixes #5345
- [ ] tests added / passed

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -277,22 +277,17 @@ class Compiled(Implementation):
         has been pre-compiled.
     '''
 
-    def __init__(self, compiled, implementation):
-        if isinstance(compiled, six.string_types):
-            path = compiled
-            compiled = None
-        elif not isinstance(compiled, dict):
-            raise ValueError('Compiled model must be defined as a path '
-                             'or a dictionary.')
+    def __init__(self, path, implementation):
+        if not (isinstance(path, six.string_types) and path.endswith('json')):
+            raise ValueError('Compiled implementation must define json '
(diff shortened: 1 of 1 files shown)


bryevdv:
Merging this scaled back PR with only minor dev API hooks added. With some more experience of real world usage we can come back to adding a more public capability later.

PR #8343, opened 2018-10-18
Add log1p option to LogScale

### LINKED ISSUE #6536: Patching regression in 0.12.6
Here is the screenshot of 1D histogram, drawn with `multi_line` and `y_axis_type="log"` in bokeh==0.12.4. Background was done using `plot.patch`

screen shot 2017-06-26 at 12 20 45

Same plot using 0.12.6:

screen shot 2017-06-26 at 12 23 16

Also in 0.12.6, `.patch` is used for plotting blue background

screen shot 2017-06-26 at 12 27 27

hsparra:
Add option to `LogScale` to use `math.log1p` for situations where there is a zero value. Ex. A histogram with empty bins.

- [x] issues: fixes #6536
- [ ] tests added / passed

image

### DIFF
--- a/bokeh/models/scales.py
+++ b/bokeh/models/scales.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 from ..core.has_props import abstract
+from ..core.properties import String
 from .transforms import Transform
 
 @abstract
@@ -40,7 +41,21 @@ class LogScale(Scale):
     '''
 
     '''
-    pass
+
+    log_type = String(help="""
+    An optional value specifying the type of log. Value values are:
+
+        log (default value)
+        log1p
+
+    """)
+    def __init__(self, *arg, **kw):
+        print(" in LogScale __init__:", kw)
+        super(LogScale, self).__init__(*arg, **kw)
+        if "log_type" in kw:
+            self.log_type = kw['log_type']
+        else:
+            self.log_type = 'log'
 
 class CategoricalScale(LinearScale):
     '''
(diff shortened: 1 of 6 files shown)


bokehjs/src/lib/models/scales/log_scale.ts
```diff
@@ -1,8 +1,12 @@
 import {Scale} from "./scale"
 import {Arrayable} from "core/types"
 
+import * as p from "core/properties"
+
 export namespace LogScale {
-  export interface Attrs extends Scale.Attrs {}
+  export interface Attrs extends Scale.Attrs {
+    log_type: string
```
mattpap: This should be a union type, i.e. `"log" | "log1p"`.

bokehjs/src/lib/models/scales/log_scale.ts
```diff
@@ -19,6 +23,20 @@ export class LogScale extends Scale {
...
   static initClass(): void {
     this.prototype.type = "LogScale"
+    this.define({
+      log_type: [ p.String, "log" ],
+    })
+  }
+
+  calculate_log(x: number): number {
+    let value: number
+    console.log(`log_type: ${this.log_type}`)
```
mattpap: This should be removed. In general, if you really need to log something, you should use bokehjs' logger.

bokeh/models/scales.py
```diff
@@ -40,7 +41,20 @@ class LogScale(Scale):
...
+        log (default value)
+        log1p
+
+    """)
+    def __init__(self, *arg, **kw):
+        super(LogScale, self).__init__(*arg, **kw)
+        if "log_type" in kw:
+            self.log_type = kw['log_type']
+        else:
+            self.log_type = 'log'
```
mattpap: All this should be replaced by `log_type = Enum("log", "log1p", default="log", help=...)`.

PR #8350, opened 2018-10-20
Initial boilerplate for bokeh/document

HuntJSparra:
- [x] issues: Partially addresses #8207 
- [x] tests added / passed

Adds boilerplate for bokeh/document


bokeh/document/locking.py
```diff
@@ -1,10 +1,50 @@
...
+
+#-----------------------------------------------------------------------------
+# General API
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Dev API
+#-----------------------------------------------------------------------------
+
 def without_document_lock(func):
```
bryevdv: This is general API, users need to use it with callback threads in server apps

Issue #8332, opened 2018-10-15
Autohide toolbar quirks


jsignell:
### Does not become hidden when mouse moves slowly off canvas
When `toolbar.autohide == True`, the toolbar remains visible if you let your mouse drift slowly off the edge.

![slowly](https://user-images.githubusercontent.com/4806877/46954334-4a7cdb80-d05e-11e8-9935-8720b99c47cb.gif)

### Interacting with the plot causes the toolbar to disappear
It reappears when the mouse is moved. 

**Proposal:** I think this behavior could be resolved by keeping track of visible state either on the model or on the renderer.

![pan_wheel_autohide](https://user-images.githubusercontent.com/4806877/47460453-89243b80-d7ac-11e8-8020-984a2373e66d.gif)

```python
from bokeh.plotting import figure, show

plot = figure(width=400, height=400, title='Toolbar Autohide')
plot.line([1,2,3,4,5], [2,5,8,2,7])
plot.toolbar.autohide = True

show(plot)
```

<details>
<summary>Version</summary>
<br>

```
Python version      :  3.6.6
IPython version     :  5.8.0
Tornado version     :  4.5.3
Bokeh version       :  1.0.0dev11
node.js version     :  v8.10.0
npm version         :  6.4.1
```
</details>

jsignell:
We should probably move it out of the milestone. If we don't catch the mouse leaving the area I don't see how we could know that the toolbar should be hidden.

philippjfr:
I would suggest either reopening this issue or I could file a new one. In many, many cases the autohide behavior does not work for me, specifically in the single plot case it does not work for me in a notebook and when using a ProxyToolbar I could not get it to work at all.

bryevdv:
Let's open a fresh issue.

mattpap:
I tested `autohide` in notebook and jlab on 3.0-dev.7 and everything seem to work fine. Please start an issue with specifics, so that I know how the environment may affect bokehjs.

PR #8363, opened 2018-10-25
Fixing autohide pan and wheel stutter

jsignell:
- [x] issues: partially #8332 
- [x] tests added / passed

pan_works

```python

output_file("test0.html")

p = figure(plot_width=300, plot_height=300)
p.scatter([1, 2, 3], [1, 2, 3])

p.toolbar.autohide = True

show(p)
```

### DIFF
--- a/bokehjs/src/lib/models/annotations/toolbar_panel.ts
+++ b/bokehjs/src/lib/models/annotations/toolbar_panel.ts
@@ -15,6 +15,8 @@ export class ToolbarPanelView extends AnnotationView {
     this.plot_view.canvas_events.appendChild(this.el)
     this._toolbar_views = {}
     build_views(this._toolbar_views, [this.model.toolbar], {parent: this})
+    const toolbar_view = this._toolbar_views[this.model.toolbar.id]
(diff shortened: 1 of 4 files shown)


bryevdv:
If possible, at test that it is visible after clicking the reset tool would also be valuable I think

bryevdv:
Also, one benefit of View Models is that they are often easier to test than views. Here e.g. we could also add a unit test to the method on the view model.

jsignell:
> add a unit test to the method on the view model

You mean the getter for visible right?

bryevdv:
> You mean the getter for visible right?

Yes it has internal logic that should be straightforward to directly maintain under test

bryevdv:
TS linter failure, not exactly sure what it means:

> [21:41:26] build/js/types/models/tools/toolbar_base.d.ts(24,1): error TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.

jsignell:
ok unit test added :)

bryevdv:
Thanks for the first PR of the post 1.0 era @jsignell :)

PR #8368, opened 2018-10-26
Improved handling of DocumentLifeCycleHandler callbacks

### LINKED ISSUE #8367: DocumentLifeCycleHandler should catch exception and clean up callbacks
The DocumentLifeCycleHandler allows adding on_session_destroyed callbacks to a Document, which works pretty well. However if a single callback fails for whatever reason other callbacks won't be called. Additionally the references to the callbacks themselves may be prevent references from being collected, therefore the handler should delete the callbacks once it is done with them.

philippjfr:
Handles exceptions in Document.on_session_destroyed callbacks and discards them once they have been processed.

- [x] issues: fixes #8367
- [x] tests added / passed

### DIFF
--- a/bokeh/application/handlers/document_lifecycle.py
+++ b/bokeh/application/handlers/document_lifecycle.py
@@ -62,14 +62,23 @@ def _on_session_destroyed(session_context):
     '''
     Calls any on_session_destroyed callbacks defined on the Document
     '''
-    for callback in session_context._document._session_destroyed_callbacks:
+    callbacks = list(session_context._document._session_destroyed_callbacks)
+    session_context._document._session_destroyed_callbacks = {}
+    for callback in callbacks:
         try:
             callback(session_context)
         except Exception as e:
             log.warn('DocumentLifeCycleHandler on_session_destroyed '
-                     'callback failed on %s callback with following error: %s'
+                     'callback %s failed with following error: %s'
                      % (callback, e))
-    session_context._document._session_destroyed_callbacks = {}
+    if callbacks:
+        # If any session callbacks were defined garbage collect after
+        # deleting all references
+        del callback
+        del callbacks
+
+        import gc
+        gc.collect()
 
 #-----------------------------------------------------------------------------
 # Code
--- a/bokeh/application/handlers/document_lifecycle.py
+++ b/bokeh/application/handlers/document_lifecycle.py
@@ -62,8 +62,8 @@ def _on_session_destroyed(session_context):
     '''
     Calls any on_session_destroyed callbacks defined on the Document
     '''
-    callbacks = list(session_context._document._session_destroyed_callbacks)
-    session_context._document._session_destroyed_callbacks = {}
+    callbacks = session_context._document.session_destroyed_callbacks
+    session_context._document.session_destroyed_callbacks = {}
     for callback in callbacks:
         try:
             callback(session_context)
--- a/bokeh/application/handlers/document_lifecycle.py
+++ b/bokeh/application/handlers/document_lifecycle.py
@@ -72,8 +72,7 @@ def _on_session_destroyed(session_context):
                      'callback %s failed with following error: %s'
                      % (callback, e))
     if callbacks:
-        # If any session callbacks were defined garbage collect after
-        # deleting all references
+        # If any session callbacks were defined garbage collect after deleting all references
         del callback
         del callbacks
 
--- a/bokeh/application/handlers/document_lifecycle.py
+++ b/bokeh/application/handlers/document_lifecycle.py
@@ -63,7 +63,7 @@ def _on_session_destroyed(session_context):
     Calls any on_session_destroyed callbacks defined on the Document
     '''
     callbacks = session_context._document.session_destroyed_callbacks
-    session_context._document.session_destroyed_callbacks = {}
+    session_context._document.session_destroyed_callbacks = set()
     for callback in callbacks:
         try:
             callback(session_context)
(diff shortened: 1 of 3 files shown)


bryevdv:
Looks good but should be possible to add tests

philippjfr:
@bryevdv Would you object to adding a ``gc.collect()`` call at the end of this, otherwise the last session isn't necessarily cleaned up immediately. In my case the on session destroy callbacks themselves hold references, which means I can't simply add a gc.collect call inside one of them, since they will still be referenced by the document.

bryevdv:
>  Otherwise the last session isn't necessarily cleaned up immediately

Is it cleaned up eventually? If so I don't think this is necessary?

philippjfr:
>Is it cleaned up eventually? If so I don't think this is necessary?

I'll try again, but I've only been able to get to collect the references if another session is opened and then closed, which means that it will hold two sessions in memory even if only one is running.

The two profiles aren't exactly comparable but in both I've consecutively opened and closed 10 sessions, then waited and then repeated.

Here's without the extra gc.collect:

screen shot 2018-10-26 at 11 26 45 pm

Here's with it:

screen shot 2018-10-26 at 11 42 30 pm

I'll run a longer running example to see if the extra 50 MB eventually gets collected or not.

philippjfr:
A longer experiment without gc.collect:

screen shot 2018-10-27 at 12 02 07 am

bryevdv:
Are the top images swapped? The one without looks "better". 

If we add this here, can we remove this one here:

https://github.com/bokeh/bokeh/blob/master/bokeh/document/document.py#L472

It seems that we should only need one call per session, as late as possible. Perhaps the one in `Document.destroy` is just happening too soon.

philippjfr:
>Are the top images swapped? The one without looks "better".

Yes, sorry I fixed the comment once I noticed.

bryevdv:
> private API though:

We could add API?  or perhaps this removal could happen in `Document.destroy` ?

philippjfr:
>It seems that we should only need one call per session, as late as possible. Perhaps the one in Document.destroy is just happening too soon.

Good point.

philippjfr:
I could move the gc.collect here:

https://github.com/bokeh/bokeh/blob/master/bokeh/server/contexts.py#L245

That's after the Document and session have been destroyed and the lifecycle hooks have been called, but I don't know if anyone ever calls ``Document.destroy`` and expects gc.collect to happen.

philippjfr:
>We could add API? or perhaps this removal could happen in Document.destroy ?

It can't unfortunately since the lifecycle hooks are called after ``Document.destroy``.

philippjfr:
At minimum I could skip the second gc.collect if there are no Document.on_session_destroyed callbacks defined, leaving almost everyone unaffected.

bryevdv:
> At minimum I could skip the second gc.collect if there are no Document.on_session_destroyed callbacks defined, leaving almost everyone unaffected.

perfect

bokeh/application/handlers/document_lifecycle.py
```diff
@@ -62,8 +62,23 @@ def _on_session_destroyed(session_context):
     '''
     Calls any on_session_destroyed callbacks defined on the Document
     '''
-    for callback in session_context._document._session_destroyed_callbacks:
-        callback(session_context)
+    callbacks = list(session_context._document._session_destroyed_callbacks)
+    session_context._document._session_destroyed_callbacks = {}
```
bryevdv: Let's try to define some minimal API so that it can be put under test

bokeh/application/handlers/document_lifecycle.py
```diff
@@ -62,8 +62,23 @@ def _on_session_destroyed(session_context):
...
+    for callback in callbacks:
+        try:
+            callback(session_context)
+        except Exception as e:
+            log.warn('DocumentLifeCycleHandler on_session_destroyed '
+                     'callback %s failed with following error: %s'
+                     % (callback, e))
+    if callbacks:
+        # If any session callbacks were defined garbage collect after
+        # deleting all references
```
bryevdv: one line for this comment

bokeh/application/handlers/document_lifecycle.py
philippjfr: I've amended the existing test to check they are in fact cleared, do you want me to add ``Document.clear_on_session_destroyed`` or similar?

bryevdv: how about just a (Python) property getter/setter for `_session_destroyed_callbacks` ?

philippjfr: Sure.

bokeh/document/document.py
```diff
@@ -132,6 +132,17 @@ def session_callbacks(self):
         '''
         return list(self._session_callbacks)
 
+    @property
+    def session_destroyed_callbacks(self):
+        ''' A list of all the on_session_destroyed callbacks on this document.
+
+        '''
+        return list(self._session_destroyed_callbacks)
```
bryevdv: Why the list asymmetry here? If you want to return a copy, why not call `dict` and return that?

philippjfr: Yeah, no good reason, I was copying `session_callbacks` but that is actually stored as a set.

bryevdv: I think setting it to `{}` above is a mistake too, then, you'd need to assign it `set()` I think it should be consistent throughout, whatever it is

philippjfr: Yes, I'll store these callbacks in a set too.

PR #8370, opened 2018-10-26
Make BokehJS prop.spec Private

### LINKED ISSUE #8356: DataTable crashes layout if 'field' is key in ColumnDataSource data
Bokeh 0.13.0
Mac OS 10.13.3
python 2.7.10
Safari 11.0.3

If 'field' is a key in the data of ColumnDataSource, bokeh serve cannot generate a layout including a DataTable using this ColumnDataSource.  This only fails if the DataTable is in the layout, the DataTable may exist if not in the layout.

The following code will fail.  If you replace 'field' in variables with almost anything else, the code works.

```
from bokeh.models.widgets import TableColumn, DataTable
from bokeh.io import curdoc


variables = ['x', 'y', 'field']
source = ColumnDataSource(data={v: [i] for i, v in enumerate(variables)})
table = DataTable(source=source, columns=[TableColumn(field=v, title=v) for v in variables])
curdoc().add_root(column(table))
```

bryevdv:
- [x] issues: fixes #8356
- [x] tests added / passed

This PR addresses #8356, however after a brief closer look it became evident that with a little more work than the one-line fix, some serious tech debt around BokehJS properties could be paid down. 

This PR:

* makes the `spec` field on `Property` be private, i.e. `prop._spec`
* restricts all direct access to `_spec` to inside `properties.ts`
* ensures that only actual "DataSpec" properties (i.e. only those with `prop.dataspec = True`) have a `._spec` present
* some cosmetic variable renames to aid with confirming `spec` not used outside `properties.ts`

Some new methods provide uniform access to the fields inside `_spec`. These first four only work on DataSpec properties and will raise an error on non-dataspecs:

* `get_expr()` -- access a dataspec expression
* `get_field()` -- access a dataspec field
* `get_transform()` -- access a dataspec transform
* `array(...)` -- return CDS array for field (unchanged from existing)

These last two work uniformly on any property dataspec or nor:

* `value_optional()` -- return undefined if no value set
* `value()` -- raise error if no value set (unchanged from existing)

I don't think the naming is ideal, but `transform` already refers to the "intrinsic" transform a property can have, and I do not want to rename `value` to `get_value` e.g. because that breaks compatibility on something that might possibly reasonably be used by users [1]

This PR is currently WIP. further tasks:

* [ ] release notes
* [ ] more tests (explicit test for #8356 and others)
* [ ] convert test to TS

---

[1] Note that this is a "breaking" change to BokehJS, but does not contravene the ***Specific Commitment for BokehJS*** made in the announcement (all existing CustomJS examples and BokehJS models and properties  and events will stay compatible) Direct usage of `.spec` is not something we can reasonably support (and I have no reason to believe *anyone* has tried to use this)

cc @bokeh/core

### DIFF
--- a/bokehjs/src/lib/core/has_props.ts
+++ b/bokehjs/src/lib/core/has_props.ts
@@ -520,7 +520,7 @@ export abstract class HasProps extends Signalable() {
       // store under the canonical field name, e.g. _image_shape, even if the column name is "foo"
       if (prop.spec.field != null && prop.spec.field in source._shapes)
         data[`_${name}_shape`] = source._shapes[prop.spec.field]
-      if (prop instanceof p.Distance)
+      if (prop instanceof p.DistanceSpec)
         data[`max_${name}`] = max(data[`_${name}`])
     }
     return data
--- a/bokehjs/src/lib/core/has_props.ts
+++ b/bokehjs/src/lib/core/has_props.ts
@@ -204,7 +204,7 @@ export abstract class HasProps extends Signalable() {
     for (const name in this.properties) {
       const prop = this.properties[name]
       prop.update()
-      if (prop.spec.transform != null)
+      if (prop.dataspec && prop.spec.transform != null)
         this.connect(prop.spec.transform.change, () => this.transformchange.emit())
     }
 
--- a/bokehjs/src/lib/core/has_props.ts
+++ b/bokehjs/src/lib/core/has_props.ts
@@ -512,7 +512,7 @@ export abstract class HasProps extends Signalable() {
       if (!prop.dataspec)
         continue
       // this skips optional properties like radius for circles
-      if (prop.optional && prop.spec.value == null && !(name in this._set_after_defaults))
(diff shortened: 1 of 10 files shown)


bryevdv:
This PR uncovered an issue with mis-matched column lengths in `donut.py`. It "runs" locally, but hangs the examples test. Running locally does show console errors in the browser (and does with 1.0 too) so I don't know why it has not been failing the examples test earlier.

bryevdv:
Well, this PR also uncovered that several `TextAnnotation` subclasses have vectorized mixins when they shouldn't. I think fixing that is OK, since there was no meaningful way to use things except setting single values, and that will continue to function just as before. But that will take some more work.

bryevdv:
Scalar visual props on the JS side is too big of a lift for me right now. I think doing that in isolation first will be a better approach. I will close this PR and submit a smaller one with the fix for the original linked issue and as much other small cleanup as can be made easily.

PR #8372, opened 2018-10-27
Extended BAD_COLUMN_NAME error 

mborus:
Extended BAD_COLUMN_NAME error to include suggestions and 
show key that caused error.

### DIFF
--- a/bokeh/core/validation/errors.py
+++ b/bokeh/core/validation/errors.py
@@ -109,7 +109,7 @@ log = logging.getLogger(__name__)
 #-----------------------------------------------------------------------------
 
 codes = {
-    1001: ("BAD_COLUMN_NAME",                                    "Glyph refers to nonexistent column name. This could either be due to a misspelling or typo, or due to an expected column being missing. "),
+    1001: ("BAD_COLUMN_NAME",                                    "Glyph refers to nonexistent column name. This could either be due to a misspelling or typo, or due to an expected column being missing. "), # NOQA
     1002: ("MISSING_GLYPH",                                      "Glyph renderer has no glyph set"),
     1003: ("NO_SOURCE_FOR_GLYPH",                                "Glyph renderer has no data source"),
     1004: ("REQUIRED_RANGE",                                     "A required Range object is missing"),
--- a/bokeh/core/validation/errors.py
+++ b/bokeh/core/validation/errors.py
@@ -109,7 +109,7 @@ log = logging.getLogger(__name__)
 #-----------------------------------------------------------------------------
 
 codes = {
-    1001: ("BAD_COLUMN_NAME",                                    "Glyph refers to nonexistent column name"),
+    1001: ("BAD_COLUMN_NAME",                                    "Glyph refers to nonexistent column name. This could either be due to a misspelling or typo, or due to an expected column being missing. "),
     1002: ("MISSING_GLYPH",                                      "Glyph renderer has no glyph set"),
     1003: ("NO_SOURCE_FOR_GLYPH",                                "Glyph renderer has no data source"),
     1004: ("REQUIRED_RANGE",                                     "A required Range object is missing"),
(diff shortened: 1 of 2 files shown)


bryevdv:
@mborus can you include some example output as a comment in this PR? Or, if you are up for it, it would be even better to add a test to confirm this functioning. I'm happy to advise if you would like to do that.

mborus:
@bryevdv Can you let me know where in the codebase the test of this error is found (or should be added)? I will see if I can add a test.

For the code:

    from bokeh.io import show
    p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5],
         size=15,
         line_color='navy',
         fill_color='xx',
         fill_alpha=0.5
        )
    show(p)

The extended error is:

ERROR:C:\Python36-32\lib\site-packages\bokeh\core\validation\check.py:E-1001 (BAD_COLUMN_NAME): Glyph refers to nonexistent column name: key "fill_color" value "xx" (closest match: "x") [renderer: GlyphRenderer(id='c92deafa-6139-4711-8844-3c83b7ecf336', ...)]

A misspelled color like `fill_color='orrange'` will at least tell you that the error is caused by the `fill_color`
field.

ERROR:C:\Python36-32\lib\site-packages\bokeh\core\validation\check.py:E-1001 (BAD_COLUMN_NAME): Glyph refers to nonexistent column name: key "fill_color" value "orrange" [renderer: GlyphRenderer(id='53867fda-5ed5-4d29-af39-6933568668cc', ...)]

bryevdv:
@mborus The place for those tests is here:

https://github.com/bokeh/bokeh/blob/master/bokeh/models/tests/test_renderers.py

Sadly, I would regard the existing tests as pretty sub-adequate. They check that an error happens at all, or not, but do not confirm the actual message contents. And looking closer it seems like there is not actually any tests at all for `_check_bad_column_name` specifically. So, if you are able, adding some tests that execute `_check_bad_column_name` with some good and bad configurations, similar to the existing tests there would be a great start. Ideally, all the tests in that file could be refined to actually compare the error message more carefully.

bryevdv:
@mborus one other idea: Since, as you point out, the problem can either be:

* a misspelled column name, or
* a misspelled fixed value ("orrange") 

Maybe the message can be tweaked to suggest both possibilities?

mborus:
I discussed this with @mattpap in person when he suggested to fix this at pyconDE saturday sprint day.
He pointed out that if there's a list of column names to test against, all errors needed reporting in one go.

Currently, the beginning "Glyph refers to nonexistent column name" and end "[renderer: GlyphRenderer(id='53867fda-5ed5-4d29-af39-6933568668cc', ...)]" are given. The code is supposed to match candidates against multiple column names and multiple keys (it doesn't have to be "fill_color")

At this point, the code can't see the previous check against other fixed values like color words and I don't know yet, where this check is run in other scenarios.

My hope is that the new error message at least names the key "fill_color" as the problem, so if should be helpful even if mentioning column names are still confusing.

I'll have a look at the test during the week. I'll also rewrite my list comprehension, because I found a simpler way to get the first element of a list of lists that allows empty lists and works in python2.7 and 3.4+, so I can get rid of that double comprehension.

bryevdv:
> At this point, the code can't see the previous check against other fixed values like color words and I don't know yet, where this check is run in other scenarios.

right, just to be clear I was not suggesting doing this, only mentioning the possibility that a misspelling could either be a column name misspelling, or a fixed value misspelling. I'd still suggest updating the message to read:

> Glyph refers to nonexistent column name: key "fill_color" value "orrange" This could either be due to a misspelling or typo, or due to an expected column not being missing. Closest existing column names are: ...

Which is just a change to the text, not involving any extra or different checks than are there now

bryevdv:
@mborus I am back from my break, are you able to continue work on this PR? I think the message should get a tweak and a basic text would also be appreciated

mborus:
Am 2018-11-27 00:07, schrieb Bryan Van de Ven:
> @mborus [1] I am back from my break, are you able to continue work on
> this PR? I think the message should get a tweak and a basic text would
> also be appreciated

Thanks for the reminder. I'm on home office next week, I will try to 
complete it then.

bryevdv:
@mborus any update?

bryevdv:
In the interest of moving this forward I have made a minimal change to the error message to make it more verbose from the GH web UI. Will merge when green.

PR #8373, opened 2018-10-27
minor dataspec cleanup

### LINKED ISSUE #8356: DataTable crashes layout if 'field' is key in ColumnDataSource data

bryevdv:
- [x] issues: fixes #8356
- [x] tests added / passed

Much more modest PR. Makes the simple one line change to fix the linked issue:

screen shot 2018-10-27 at 10 08 12

and also adds a unit test to explicitly check that `{field: "field"}` works.

Also cherry picks a few of the smaller cleanup commits from the other PR just so that they are done/out of the way for later.


bryevdv:
I also went ahead and converted the test file `properties.coffee` to TypeScript.  There appeared to be some broken tests, and a broken validator for `Color` when the input was not a string.

mattpap:
It looks good, but I will have a closer look when I'm done with PyCon DE sprint.

bryevdv:
Great, there are some tslint errors about const that I already have a commitment for. I will push it later when I get home.

bryevdv:
@mattpap I'd like to merge this today. 

FYI I would also like to try and have no more than 10-15 small PRs per point release, so I'd like to cut 1.0.1 before I take time off.

bryevdv:
I'm going to go ahead and merge this, all existing tests were preserved plus some new ones were added. The meat of this PR is pretty trivial, mostly the only place that could use improvement would be the test file itself, but that can happen later when scalar props on the JS side happens.

PR #8378, opened 2018-10-29
Allow OutputDocumentFor to handle children with docs

### LINKED ISSUE #8362: Combination of "export_png()" and "show()" causes error
Hi there!
It pains me to bring up these export_png-related issues... but here I go again. This time it is an error that shows up when using `export_png()` and `show()` together:

In 0.13.0, the following works just fine:
```python
from bokeh.io import export_png, show, output_file

source = ColumnDataSource({'x': [2, 3, 4], 'y': [0.5, 1.5, 2.5]})
fig = figure(plot_width=1000, plot_height=300)
fig.hbar(y='y', right='x', height=0.4, source=source)

export_png(fig, filename="export fig.png")

doc_layout = layout([fig])

output_file('export_test.html', title='export test')
show(doc_layout)
```
...but in 1.0.0, it throws the following error for the line `show(doc_layout)`:
`RuntimeError: Models must be owned by only a single document, HelpTool(id='1026', ...) is already in a doc`

Some observations:
- I wonder if it has to do with the changes that allowed to fix #8020.
- Not using either `export_png` or `show` works.
- Using `show(fig)` instead of `show(doc_layout)` works, as well (but that is not really a workaround). 
- Using `export_png` **after** `show` works, too. This could be a workaround in most use cases, but is quite inconvenient in mine.

Many thanks for the 1.0 release, anyway. I read about it in the change log, and it really feels like the export function is now faster for multiple exports!

bryevdv:
- [x] issues: fixes #8362
- [x] tests added / passed

### Acceptable change argument:

This PR fixes #8362 by allowing the dev API context manager `OutputDocumentFor` to correctly handle cases where children models have different models that documents roots (specifically a case where a child model has a document, but its root parents have None). Prior to this PR, this situation was **explicitly** an error condition. This change is deemed acceptable, because this PR only *relaxes the condition*, i.e. all previous situations are still handled exactly as before with no change, but new acceptable cases (which *should* be acceptable) are now possible. 

In practice, almost no users will use `OutputDocumentFor` directly, and of those very few, arguably any that do would expect the behavor in this PR. The practical outcome is simply that cases that currently fail, that should succeed (i.e. #8363), will now succeed. 

Additionally, defensive code was added to the private function `_unset_temp_theme`. Specifically, it is now idempotent. Previously it would fall over if called repeatedly, or called on a document that had not previously been acted on by `_create_temp_doc`. This restriction seems unnecessary, and as the function is private, any change is acceptable. 

cc @bokeh/core re: change argument

### DIFF
--- a/bokeh/embed/tests/test_util.py
+++ b/bokeh/embed/tests/test_util.py
@@ -136,7 +136,7 @@ class Test_OutputDocumentFor_general(object):
                 pass
         assert str(e).endswith(_ODFERR)
 
-    def test_with_doc_in_child_raises_error(self):
+    def test_with_doc_in_child_sets_temp_doc(self):
         doc = Document()
         p1 = Model()
         p2 = SomeModelInTestObjects(child=Model())
@@ -144,10 +144,14 @@ class Test_OutputDocumentFor_general(object):
         assert p1.document is None
         assert p2.document is None
         assert p2.child.document is doc
-        with pytest.raises(RuntimeError) as e:
-            with beu.OutputDocumentFor([p1, p2]):
-                pass
-            assert "already in a doc" in str(e)
+        with beu.OutputDocumentFor([p1, p2]) as d:
+            assert d is not doc
+            assert p1.document is d
+            assert p2.document is d
+            assert p2.child.document is d
+        assert p1.document is None
+        assert p2.document is None
+        assert p2.child.document is doc
 
     @patch('bokeh.document.document.check_integrity')
     def test_validates_document_by_default(self, check_integrity, test_plot):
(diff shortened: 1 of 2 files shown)


bryevdv:
Other possible solutions:

* Change the default for `always_new` in `OutputDocumentFor` to True 

* leave the default but always set `always_new=True` in internal usage

I'm -1 on the first (do not want to change any argument default). For the second I believe it might have some small adverse performance penalty, additionally it would be a bit odd to have a default that is no longer the generally recommended usage. I prefer the solution in the PR to this option, but if there are objections to the PR we can look at this route.

bryevdv:
I'll plan to merge this tomorrow morning, absent any feedback.

bryevdv:
OK, I remembered that the reason for only checking roots previously was that computing `.references` for everything all the time was expensive. I have made the private API function exit early, as soon as two docs are found. This puts an upper bound on how many times `.references` may be called. This makes a common case of a single plot somewhat worse (~600ns -> 1.5ms) but prevents things from spiraling completely out of hand in more complicated cases:

```
In [45]: %timeit _compute_current_docs([p, p2, p3])
4.65 ms ± 128 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [46]: %timeit _compute_up_to_two_current_docs([p, p2, p3])
1.05 µs ± 4.82 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```


I've also changed things so that the doc computation is skipped entirely if `always_new` is True.

bryevdv:
Hrm, I just realized that makes the common single plot case 3000x worse, not 3x worse. Starting to wonder if we should just always unconditionally re-home the models in a new temp doc.

Issue #8379, opened 2018-10-30
Export CSV example more useful if you don't have to hard code headers in download.js


GenevieveBuckley:
The example code at bokeh/examples/app/export_csv would likely be more useable if you didn't have to specify in advance what the datatable column headings were in download.js

@jni recently went to some effort to work out a nicer way of doing this for one of our own projects, and we thought we'd share the solution here for others to benefit from as well. 

#### Description of expected behavior and the observed behavior
Expected behavior: I want to be able to pass in any kind of table, with arbitrary numbers of columns, and arbitrary column headings.
Actual behavior: It doesn't work unless you edit the download.js file which contains hard-coded column header names.

GenevieveBuckley:
@jni - you don't seem to have a fork of the bokeh repo. I've got a branch with the changes here https://github.com/GenevieveBuckley/bokeh/tree/export-csv-example

Do you want to fork that branch and put a PR in yourself? It's your hard work!

bryevdv:
Certainly happy to consider a PR to make the example more general

PR #8381, opened 2018-10-30
Bokeh export csv example now discovers column heading names

GenevieveBuckley:
Export csv now discovers column heading names in data table before downloading csv file.

@jni went to some effort to work out how we could do this for datatables with arbitrary numbers of columns and column headings. It's a very neat solution.

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/8379
- [ ] tests added / passed (not necessary, no new functionality added)
- [ ] release document entry (not necessary, not a new feature)

### DIFF
--- a/examples/app/export_csv/download.js
+++ b/examples/app/export_csv/download.js
@@ -1,12 +1,12 @@
 function table_to_csv(data_table) {
-    var columns = Object.keys(data_table)
-    var nrows = data_table[columns[0]].length
-    var lines = [columns.join(',')]
+    const columns = Object.keys(data_table)
+    const nrows = data_table[columns[0]].length
+    const lines = [columns.join(',')]
 
     for (var i = 0; i < nrows; i++) {
         var row = [];
         for (var j = 0; j < columns.length; j++) {
-            var column = columns[j]
+            const column = columns[j]
             row.push(data_table[column][i].toString())
         }
         lines.push(row.join(','))
@@ -16,16 +16,16 @@ function table_to_csv(data_table) {
 }
 
 
-var data = source.data
-var filetext = table_to_csv(data)
-var filename = 'data_result.csv'
-var blob = new Blob([filetext], { type: 'text/csv;charset=utf-8;' })
+const data = source.data
+const filename = 'data_result.csv'
+filetext = table_to_csv(data)
+const blob = new Blob([filetext], { type: 'text/csv;charset=utf-8;' })
 
 //addresses IE
 if (navigator.msSaveBlob) {
     navigator.msSaveBlob(blob, filename)
 } else {
-    var link = document.createElement('a')
+    const link = document.createElement('a')
     link.href = URL.createObjectURL(blob)
     link.download = filename
     link.target = '_blank'
--- a/examples/app/export_csv/download.js
+++ b/examples/app/export_csv/download.js
@@ -1,35 +1,35 @@
 function table_to_csv(data_table) {
     var columns = Object.keys(data_table)
     var nrows = data_table[columns[0]].length
-    var lines = [columns.join(',')];
+    var lines = [columns.join(',')]
 
     for (var i = 0; i < nrows; i++) {
         var row = [];
         for (var j = 0; j < columns.length; j++) {
-            var column = columns[j];
-            row.push(data_table[column][i].toString());
+            var column = columns[j]
+            row.push(data_table[column][i].toString())
         }
-        lines.push(row.join(','));
+        lines.push(row.join(','))
     }
-    filetext = lines.join('\n').concat('\n');
+    filetext = lines.join('\n').concat('\n')
     return filetext
 }
 
 
-var data = source.data;
-var filetext = table_to_csv(data);
-var filename = 'data_result.csv';
-var blob = new Blob([filetext], { type: 'text/csv;charset=utf-8;' });
+var data = source.data
+var filetext = table_to_csv(data)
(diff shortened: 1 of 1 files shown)


examples/app/export_csv/download.js
```diff
@@ -1,14 +1,23 @@
...
+        }
+        lines.push(row.join(','));
+    }
+    filetext = lines.join('\n').concat('\n');
+    return filetext
 }
 
+
+var data = source.data;
+var filetext = table_to_csv(data);
```
jni: Another thing I changed is below, the second `link = document.createElement` is redundant and can be removed.

examples/app/export_csv/download.js
```diff
@@ -1,14 +1,23 @@
-var data = source.data;
-var filetext = 'name,income,years_experience\n';
-for (var i = 0; i < data['name'].length; i++) {
-    var currRow = [data['name'][i].toString(),
-                   data['salary'][i].toString(),
-                   data['years_experience'][i].toString().concat('\n')];
+function table_to_csv(data_table) {
+    var columns = Object.keys(data_table)
+    var nrows = data_table[columns[0]].length
```
bryevdv: CDS has a `get_length` method, I think it would be better to use that, though it would require `table_to_csv` to accept `source` and not `source.data`

examples/app/export_csv/download.js
```diff
@@ -1,14 +1,23 @@
...
+    var lines = [columns.join(',')];
```
bryevdv: Should consistently not use semicolons throughout. Also should use `const` and `let` instead of `var` anywhere

bryevdv:
Thanks for the PR, I have only a few small comments/changes

examples/app/export_csv/download.js
GenevieveBuckley: Ok, I think this is fixed now.

There is one line `filetext = table_to_csv(data)` that doesn't have `const`, `let`, or `var` now (and it doesn't seem to happen inside the function either). It still works, but if it's bad style can you let me know what I should replace this with? Many thanks!

examples/app/export_csv/download.js
```diff
@@ -1,26 +1,34 @@
...
+    for (let i = 0; i < nrows; i++) {
+        let row = [];
+        for (let j = 0; j < columns.length; j++) {
+            const column = columns[j]
+            row.push(data_table[column][i].toString())
+        }
+        lines.push(row.join(','))
+    }
+    filetext = lines.join('\n').concat('\n')
+    return filetext
```
bryevdv: these two lines can be combined:
```
return lines.join('\n').concat('\n')
```

examples/app/export_csv/download.js
```diff
@@ -1,26 +1,34 @@
...
 }
 
-var filename = 'data_result.csv';
-var blob = new Blob([filetext], { type: 'text/csv;charset=utf-8;' });
+
+const data = source.data
+const filename = 'data_result.csv'
+filetext = table_to_csv(data)
```
bryevdv: first and third lines can be combined:
```
const filetext = table_to_csv(source.data)
```

bryevdv:
@GenevieveBuckley I left two small comments however it seems there is more work to do. The code as written does not work for me on Safari/OSX:

screen shot 2018-10-30 at 21 59 23

examples/app/export_csv/download.js
GenevieveBuckley: Using `const` (or `let` for that matter) here gives me the javascript error: `Uncaught ReferenceError: filetext is not defined`

GenevieveBuckley:
> The code as written (neither the original version in this PR, nor the latest) does not work for me on Safari/OSX

Hm, that's odd. I'll have to look into this.

examples/app/export_csv/download.js
GenevieveBuckley: I must have missed that, I don't see anything about a ColumnDataSource get_length method [in the docs](https://bokeh.pydata.org/en/latest/docs/reference/models/sources.html)?

It's implemented now.

bryevdv:
>  It doesn't look like a new bug we've introduced - I get the same Safari error when I run the master branch code at

OK, I am back from my November break and confirmed that this is the case. Perhaps Safari has made some change that will have to be dealt with. I'm inclined then to go ahead and merge this and make a separate issue to investigate. But for good measure can you try out and report result with your new version on some variety of platforms/browsers?

cc @GenevieveBuckley

GenevieveBuckley:
MacOS Mojave (version 10.14.1):
- Chrome Version 70.0.3538.102 (Official Build) (64-bit). Works as expected.
- Firefox 63.0.3 (64-bit). Works as expected.
- Safari Version 12.0.1 (14606.2.104.1.1). Bug as reported above, when you try and download the csv "Safari can't open the page".

@jni, @VolkerH can you do me a favour? @bryevdv would like it if we could test this on Linux (Firefox, Chrome) and Windows (Internet Explorer 11, Edge, Firefox, Chrome). Maybe we have a bit of time for this on Wednesday.

jni:
I have tested and confirm that the example works on FF and Chromium on Ubuntu 18.04.

VolkerH:
Tested working on Windows 10 with current versions of Firefox and Edge.

bryevdv:
OK I think that seems sufficient, thanks @GenevieveBuckley @jni @VolkerH 

Please make a separate issue about the Safari problem whenever you have time,

PR #8385, opened 2018-10-30
Add figure's plot id in DOM structure as well

### LINKED ISSUE #8384: Feature: Add Plot's root id to DOM so to allow API access to the object
I'm using Bokeh in Standalone mode, embedding a plot in a page like so:

```py
    script_tag, html = components(plot)
```

Unfortunately, the `html` does not contain any reference to the plot's ID (`root.id`), it only contains the element id (`root.elementid`), but the plot id is necessary if I want to interact with the plot through the API, for example calling:

```js
        Bokeh.index[plotId].save(plotTitle)
```

A workaround is to render the plot manually and communicate a reference to the plot ID out of band, like so:

```py
    script_tag, plot_info = components(plot, wrap_plot_info=False)
    root_el = ROOT_DIV.render(root=plot_info, macros=MACROS)
    # put plot_info.id somewhere in the dom
```

I'll push a PR for this.

rbu:

- [x] issues: fixes #8384
- [x] tests added / passed

### DIFF
--- a/bokeh/core/_templates/macros.html
+++ b/bokeh/core/_templates/macros.html
@@ -1,3 +1,3 @@
 {% macro embed(obj) -%}
-<div class="bk-root" id="{{ obj.elementid }}"></div>
+<div class="bk-root" id="{{ obj.elementid }}"{% if obj.id %} data-root-id="{{ obj.id }}"{% endif %}></div>
 {%- endmacro %}
--- a/bokeh/embed/tests/test_standalone.py
+++ b/bokeh/embed/tests/test_standalone.py
@@ -125,6 +125,7 @@ class Test_components(object):
         assert len(scripts) == 1
         assert scripts[0].attrs == {'type': 'text/javascript'}
 
+    @patch('bokeh.embed.util.make_globally_unique_id', new=stable_id)
     def test_div_attrs(self, test_plot):
         script, div = bes.components(test_plot)
         html = bs4.BeautifulSoup(div, "lxml")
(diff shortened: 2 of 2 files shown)


rbu:
@bryevdv I added a test for the functionality and fixed the other test that got broken. Unfortunately, the travis build is red due to a timeout in the selenium/chrome tests. Is that related to my change or just a hickup?

bryevdv:
@rbu apologies for the long delay, I took all of November off from OSS work, and my December has been a little personally hectic so far. The test failure was spurious and everything is green now.  I don't have any objections to adding this since there is a test for it. But we should give @mattpap a chance to chime in in case there are any considerations I may not be aware of.

bryevdv:
My only other comment is that I would not want to encourage folks to use private attrs like `plot._id` so we might want to consider adding a read-only getter for that.

bryevdv:
Actually, there is one other route we could consider. Models have some other user accessible fields:

* `obj.name` - can be set to an arbitrary string
* `obj.tags` - can be set to an arbitrary list of values

The `name` field is only used lightly by Bokeh, i.e. if you supply a  name it is available for templating the object by name. And the `tags` field not at all (there is a function that can query the object graph for objects with matching tags, but that is it). We could consider using these properties as a basis for adding DOM attrs instead of the model id. That way, users have explicit control over what the values will be and know them ahead of time. Thoughts?

bryevdv:
@mattpap is there any particular objection to adding 
```
@property
def id(self): return self._id
```
to `Model`?

mattpap:
> `@property`
> `def id(self): return self._id`

Not at all. It's peculiar to me, why this wasn't in the first place. When this is done, we should update the code base to use `id` instead of `_id`.

bryevdv:
OK, @rbu has not responded to recent comments, so I will merge this as-is and make a follow-up PR to add the getter and make the other changes.

rbu:
@bryevdv, thank you for merging. The changes you proposed were perfectly fine, I simply took time away from the project myself as well.

PR #8386, opened 2018-10-31
Bryanv/8362 new temp doc for export

bryevdv:
- [x] issues: fixes #xxxx
- [x] tests added / passed

This is a simpler PR for #8362 that forces all export functions to set `aways_new=True` when generating their file output. This alleviates the problem of exporting a child of some layout before showing the layout causing a "multiple document" error, but does no change any semantics of `OutputDocumentFor` (not even just by expanding acceptable usage) and does not cause a costly references traversal.

### DIFF
--- a/bokeh/io/tests/test_export.py
+++ b/bokeh/io/tests/test_export.py
@@ -26,6 +26,7 @@ import re
 from PIL import Image
 
 # Bokeh imports
+from bokeh.layouts import row
 from bokeh.models.plots import Plot
 from bokeh.models.ranges import Range1d
 from bokeh.io.webdriver import webdriver_control, terminate_webdriver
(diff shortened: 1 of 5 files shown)


bryevdv:
cc @mattpap

bokeh/io/tests/test_export.py
```diff
@@ -170,6 +171,22 @@ def test_get_layout_html_resets_plot_dims():
...
+    bie.get_layout_html(layout)
+
+def test_layout_html_on_parent_first():
+    p = Plot(x_range=Range1d(), y_range=Range1d())
+
+    layout = row(p)
+    bie.get_layout_html(layout)
+
+    bie.get_layout_html(p, height=100, width=100)
+
```
bryevdv: verified these failed before the change in the PR

PR #8398, opened 2018-11-05
Fix typo and grammar mistakes

dnnr:

- [ ] issues: fixes #xxxx
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/annotations.py
+++ b/bokeh/models/annotations.py
@@ -1002,7 +1002,7 @@ class Tooltip(Annotation):
     """)
 
     show_arrow = Bool(default=True, help="""
-    Whether tooltip's arrow should be showed.
+    Whether tooltip's arrow should be shown.
     """)
 
 # This only exists to prevent a circular import.
--- a/bokeh/models/tools.py
+++ b/bokeh/models/tools.py
@@ -994,7 +994,7 @@ class HoverTool(Inspection):
     have a matching value for their ``name`` attribute will be used.
     """)
 
-    renderers = Either(Auto, List(Instance(Renderer)), defatult="auto", help="""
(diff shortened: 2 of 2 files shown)


dnnr:
Not sure if there's any point in opening an issue for that, as the PR message template suggests...

bokeh/models/tools.py
```diff
@@ -994,7 +994,7 @@ class HoverTool(Inspection):
     have a matching value for their ``name`` attribute will be used.
     """)
 
-    renderers = Either(Auto, List(Instance(Renderer)), defatult="auto", help="""
+    renderers = Either(Auto, List(Instance(Renderer)), default="auto", help="""
```
mattpap: Keyword arguments strike again. It's interesting that this wasn't picked up by any tests.

mattpap: Actually `"auto"` is the intrinsic default value and specifying `default` manually is just either pedantry or is supposed to aid documentation.

bokeh/models/annotations.py
```diff
@@ -1002,7 +1002,7 @@ class Tooltip(Annotation):
     """)
 
     show_arrow = Bool(default=True, help="""
-    Whether tooltip's arrow should be showed.
+    Whether tooltip's arrow should be shown.
```
mattpap: As far I can tell (not a native speaker), both are correct, just, I suppose, "shown" is the preferred one.

mattpap:
> Not sure if there's any point in opening an issue for that, as the PR message template suggests...

It's advised, but we usually skip this in such trivial cases.

mattpap:
Test failures are unrelated and will be fixed in PR #8406.

PR #8407, opened 2018-11-07
Omit colon in hover tooltips if label is empty

### LINKED ISSUE #8399: Omit colon in hover tooltips if first tuple entry is empty
This is a suggestion for a very tiny improvement (but still...) of the presentation of HoverTool's tooltips.

Currently, something like this:
```
tooltips=[
    ('Date', '@date{%F}'),
    ('Balance', '@hours{(0.[00])} hours'),
    ('', '@days{0.[0]} days'),
],
```
...gets rendered like this:
image

I'm suggesting to remove the colon from the third row by omitting it whenever the first string in a tuple is empty.

It seems that for more sophisticated presentations, people are supposed to use custom tooltips. So there are probably no plans to add any complexity to the default tooltips.

However, I'd say this small adjustment makes the default behavior considerably more useful/powerful without making things worse for any use-case (...that I could think of).

dnnr:
- [x] issues: fixes #8399 
- [ ] ~~tests added / passed~~
- [ ] ~~release document entry (if new feature or API change)~~

### DIFF
--- a/bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
+++ b/bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
@@ -397,7 +397,7 @@ export class HoverToolView extends InspectToolView {
 
         let cell: HTMLElement
 
-        cell = div({style: {display: "table-cell"}, class: 'bk-tooltip-row-label'}, label.length != 0 ? "${label}: " : "")
+        cell = div({style: {display: "table-cell"}, class: 'bk-tooltip-row-label'}, label.length != 0 ? `${label}: ` : "")
         row.appendChild(cell)
 
         cell = div({style: {display: "table-cell"}, class: 'bk-tooltip-row-value'})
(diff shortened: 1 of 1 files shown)


bokehjs/src/lib/models/tools/inspectors/hover_tool.ts
```diff
@@ -397,7 +397,7 @@ export class HoverToolView extends InspectToolView {
 
         let cell: HTMLElement
 
-        cell = div({style: {display: "table-cell"}, class: 'bk-tooltip-row-label'}, `${label}: `)
+        cell = div({style: {display: "table-cell"}, class: 'bk-tooltip-row-label'}, label.length != 0 ? "${label}: " : "")
```
mattpap: Actually my suggestion in the issue was slightly wrong. It should be ``${label}: `` (backticks instead of quotes, the same as above).

dnnr: I was already wondering about that... I gotta admit that I didn't try it out myself because setting up the dev environment seemed to be more effort than the proverbial `pip install .`.

dnnr: Fixed now.

mattpap: > (...) setting up the dev environment seemed to be more effort than the proverbial pip install 

Setting up a dev environment is indeed a pain in the bottom currently.

PR #8419, opened 2018-11-10
Skip rendering legend glyph for empty scatter

### LINKED ISSUE #8395: Legend breaks plot when plotting empty scatter glyph
When plotting an empty scatter glyph, which is often useful when streaming data, the legend breaks because it cannot determine the correct marker to render the legend items. A basic example to reproduce 

```python

p = figure(x_range=(0, 5), y_range=(0, 3))
scatter = p.scatter([], [], legend='Some legend entry')
```

```
VM1330:130 Uncaught TypeError: _.marker_funcs[a[u]] is not a function
    at e._render (<anonymous>:130:341219)
    at e.draw_legend_for_index (<anonymous>:130:341597)
    at t.draw_legend (<anonymous>:130:394695)
    at e (<anonymous>:130:159349)
    at t._draw_legend_items (<anonymous>:130:159662)
    at t.render (<anonymous>:130:158255)
    at t._paint_levels (<anonymous>:130:371299)
    at t.paint (<anonymous>:130:370650)
    at t.repaint (<anonymous>:130:369420)
    at t.<anonymous> (<anonymous>:130:366711)
```

screen shot 2018-11-03 at 12 33 52 pm

This is reproducible using bokeh 1.0.0 and bokeh 1.0.1 and likely originates in this PR https://github.com/bokeh/bokeh/pull/8285 generalizing the scatter glyph.

philippjfr:
This skips rendering the legend glyph entry if the scatter glyph is empty. Alternatively we could simply default to a circle but that may be misleading and the legend does update if you subsequently set or stream data to the scatter.

Here is an image of a plot with two scatter glyphs each with empty data:

screen shot 2018-11-10 at 9 23 22 pm

and then I set some data:

screen shot 2018-11-10 at 9 23 30 pm

Not sure what the best way is to test actual rendering of plots.

- [x] issues: fixes #8395
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/markers/scatter.ts
+++ b/bokehjs/src/lib/models/markers/scatter.ts
@@ -16,7 +16,7 @@ export class ScatterView extends MarkerView {
 
   protected _render(ctx: Context2d, indices: number[], {sx, sy, _size, _angle, _marker}: ScatterData): void {
     for (const i of indices) {
-      if (isNaN(sx[i] + sy[i] + _size[i] + _angle[i]))
+      if (isNaN(sx[i] + sy[i] + _size[i] + _angle[i]) || _marker[i] == null)
         continue
 
       const r = _size[i]/2


bokehjs/src/lib/models/markers/scatter.ts
```diff
@@ -53,6 +53,8 @@ export class ScatterView extends MarkerView {
 
     const marker: string[] = new Array(len)
     marker[index] = this._marker[index]
+    if (marker[index] === undefined)
+      return
```
mattpap: I suppose this would do the job, but the problem is really with `_render`, which should skip rendering if `_marker[index]` isn't defined. I would rather proceed with the following change:
```diff
diff --git a/bokehjs/src/lib/models/markers/scatter.ts b/bokehjs/src/lib/models/markers/scatter.ts
index 9f4950d..9a9498a 100644
--- a/bokehjs/src/lib/models/markers/scatter.ts
+++ b/bokehjs/src/lib/models/markers/scatter.ts
@@ -16,7 +16,7 @@ export class ScatterView extends MarkerView {
 
   protected _render(ctx: Context2d, indices: number[], {sx, sy, _size, _angle, _marker}: ScatterData): void {
     for (const i of indices) {
-      if (isNaN(sx[i] + sy[i] + _size[i] + _angle[i]))
+      if (isNaN(sx[i] + sy[i] + _size[i] + _angle[i]) || _marker[i] == null)
         continue
 
       const r = _size[i]/2
```

philippjfr: Okay, I'll update the PR first thing in the morning.

PR #8434, opened 2018-11-14
Clean up .gitignore files

### LINKED ISSUE #8433: Clean up .gitignore files across the repository
Currently `.gitignore` files, especially the root one, contain a lot of garbage that often is counter productive hiding files from developer's eyes.

mattpap:
My goal is to have `.gitignore` files reduced to the bare minimum. Anything that's not project related (so e.g. editor/IDE temporary files) should be configured locally by the user. All entries (or groups of entries) must have an annotation why we ignore or what generate those (a good example is `.cache/`). If entries are specific to a subdirectory, they should be moved to a specific `.gitignore` file.

fixes #8433

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,113 +1,42 @@
-# Editor temporary/working/backup files #
-#########################################
-.#*
-[#]*#
-*~
-*$
-*.bak
-*.diff
-*.org
-.project
-.pydevproject
-*.rej
-.settings/
-.*.sw[nop]
-.sw[nop]
-*.tmp
-*.coverage.*
-**/screenshots/current*
-**/screenshots/diff*
-
-# Compiled source #
-###################
-*.a
-*.com
-*.class
-*.dll
-*.exe
-*.o
 *.py[ocd]
-*.so
 
-# Packages #
-############
-# it's better to unpack these files and commit the raw source
-# git has its own built in compression methods
-*.7z
-*.bz2
-*.bzip2
-*.dmg
-*.gz
-*.iso
-*.jar
-*.rar
-*.tar
-*.tbz2
-*.tgz
-*.zip
-
-# Python files #
-################
 # setup.py working directory
 /build
+
 # sphinx build directory
 /_build
+
 # setup.py dist directory
 /dist
 /doc/build
 /doc/cdoc/build
 /MANIFEST
+
 # Egg metadata
-*.egg-info
-# The shelf plugin uses this dir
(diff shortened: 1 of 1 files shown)


bryevdv:
> (so e.g. editor/IDE temporary files) should be configured locally by the user. 

What's the rationale for this? They should never be checked in, and doing this only pushes busy work out to potential contributors and increases the risk of accidental check ins (which ideally are caught in review, of course, but then require extra back-and-forth and this increases the friction of getting PRs merged)

mattpap:
We can't include every type of file in our gitignore that users' software may produce, so I would prefer to have none and educate users users what they should do instead. That's one line in the documentation (possibly a link to https://git-scm.com/docs/gitignore), next to `git clone`. The checked-in gitignore files should include only files that are produced by tools we use in the repository.

bryevdv:
Ok it's not my preference but you seem to feel strongly about it. So I'd request those updated docs be added to this PR, including a full example of how to ignore some common editor file junk.

bryevdv:
@mattpap I'd like for this not to linger, but I'm not super happy about it as evidently some other people are not. I'd propose the following, which keeps common editor cruft ignored but still significantly cleans up the file:
```
*.py[ocd]

# setup.py working directory
/build

# sphinx build directory
/sphinx/build

# setup.py dist directory
/dist
/doc/build
/doc/cdoc/build
/MANIFEST

# Egg metadata
/*.egg-info

# coverage reports
/.coverage

# ipython files
**/.ipynb_checkpoints

# tests/examples
report.html
examples.log

# conda build
/linux-64
/osx-64
/win-64

# pytest
/.cache/
/.pytest_cache/

# auto-generated
/bokeh/LICENSE.txt

# dev files
*.patch
*.diff
*.rej

# common editor cruft
.#*
[#]*#
*~
*$
*.bak
.project
.pydevproject
.*.sw[nop]
.sw[nop]
*.tmp

# compressed / binary files
*.bz2
*.bzip2
*.dmg
*.gz
*.iso
*.tar
*.tgz
*.zip
```

bryevdv:
Edit: tho what are:
```
/doc/build
/doc/cdoc/build
```

bryevdv:
@mattpap I'd like to make the change proposed above and merge this

bryevdv:
@mattpap ^^^

bryevdv:
Hopefully an acceptable compromise to all

PR #8435, opened 2018-11-15
Fix text_align in table cell formatters

### LINKED ISSUE #5721: text_align attribute in NumberFormatter not doing anything
When making a `datatable`, I want to right align numerical values in the table, but when I set the `text_align` attribute in `NumberFormatter`, the values continue to remain left aligned.  Here is my enviornment:
* Python 2.7.12 :: Anaconda 4.1.1 (64-bit)
* numpy==1.11.1
* pandas==0.18.1
* bokeh==0.12.4
* Windows 7, Chrome

And here is a code snippet:
```
import pandas as pd
import numpy as np
from bokeh.models.widgets import DataTable, NumberFormatter, TableColumn
from bokeh.plotting import show

df = []
for ii in range(1, 11):
    df.append({'x': ii, 'y': 1000 * np.random.rand()})
df = pd.DataFrame(df)

source = ColumnDataSource(data=df)

columns = [
    TableColumn(field='x', title='Col 1'),
    TableColumn(field='y', title='Col 2',
                formatter=NumberFormatter(format='$0,0.00',
                                          text_align='right')),
]

dt = DataTable(source=source, columns=columns, width=500, height=200, row_headers=False)

show(dt)
```

Here is the output I am getting in my Jupyter Notebook:
untitled

I would expect that the dollar amounts in `Col 2` would be right aligned, but they aren't.

konmeo:
The text_align attribute of table cell text's span is set but does not take effect because it does not fill the container width of table cell's div.  This trivial fix is in the base StringFormatter which also addresses NumberFormatter, allowing numeric columns/cells to be defined with proper right-alignment.
- [x] issues: fixes #5721

### DIFF
--- a/bokehjs/src/lib/models/widgets/tables/cell_formatters.ts
+++ b/bokehjs/src/lib/models/widgets/tables/cell_formatters.ts
@@ -79,6 +79,8 @@ export class StringFormatter extends CellFormatter {
     if (text_color != null)
       text.style.color = text_color
 
+    text.style.display = "block"
+
     return text.outerHTML
   }
 }
--- a/bokehjs/src/lib/models/widgets/tables/cell_formatters.ts
+++ b/bokehjs/src/lib/models/widgets/tables/cell_formatters.ts
@@ -3,7 +3,7 @@ import compile_template = require("underscore.template")
 import tz = require("timezone")
 
 import * as p from "core/properties"
-import {span, i} from "core/dom"
+import {div, i} from "core/dom"
 import {Color} from "core/types"
 import {FontStyle, TextAlign, RoundingFunction} from "core/enums"
 import {isString} from "core/util/types"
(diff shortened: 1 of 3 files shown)


bokehjs/src/lib/models/widgets/tables/cell_formatters.ts
```diff
@@ -79,6 +79,8 @@ export class StringFormatter extends CellFormatter {
     if (text_color != null)
       text.style.color = text_color
 
+    text.style.display = "block"
```
mattpap: It would be better to change `span` to `div` on line 67.

mattpap:
There is `examples/integration/widgets/data_table_customization` that can be extended to test this (e.g. you can add another column with an appropriately configured string formatter).

Issue #626, opened 2014-05-14
Make radius dimension configurable for circles


bryevdv:
The notion of a "circle with radius" implicitly assumes 1-1 aspect ratio. If the aspect is not 1-1, the "proper" thing would be to draw ellipses. However, it is often desired to draw circles anyway, even with non 1-1 aspect As a practical matter we have to pick one dimension to measure screen distance along. Currently this dimension is fixed to be the `x` axis. Propose making the dimension configurable with a `radius_dim` parameter that takes the following values:
* `"x"` measure screen distance along x-axis
* `"y"` measure screen distance along y-axis
* `"min"` use min screen distance
* `"max"` use max screen distance

damianavila:
status: low priority

bryevdv:
Accidental push to master. Screenshot attached below (would have been in PR)

![screen shot 2015-02-09 at 11 18 41 am](https://cloud.githubusercontent.com/assets/1078448/6111346/c7f8b8f2-b04d-11e4-8525-892108b0d2f6.png)

Venedictos:
Hello 

Please i intend to produce a bubble graph in Bokeh, but do not seem to find a way of supplying different values for size. A CDS with x,y and z columns. Have used p.circle(x,y, size=z, source=source), radius and radius_dimensions do not take a column. Any help please! Thanks
![image](https://user-images.githubusercontent.com/13404635/58475013-b66f0880-8144-11e9-878a-91ccddf9153e.png)

bryevdv:
@Venedictos There are several ways to have many circles with different sized, but GitHub is not the appropriate place to ask suppose questions.  Please ask again on the [Bokeh Discourse site](https://discourse.bokeh.org)

Venedictos:
@bryevdv thanks. I will. Cheers

PR #8440, opened 2018-11-15
Add min/max radius dimensions

0x5e5e:
Fixes: #626

This PR completes the functionality proposed in #626 which was partially addressed in commit ee8de73.

**Description:**
This PR adds two additional dimensions to radius_dimension, 'min' and 'max'

'min' dynamically selects either the x and y dimension off which to calculate the radius value, choosing whichever results in a smaller size, while 'max' does the same thing but chooses the maximum.

### DIFF
--- a/bokehjs/src/lib/models/glyphs/circle.ts
+++ b/bokehjs/src/lib/models/glyphs/circle.ts
@@ -45,14 +45,16 @@ export class CircleView extends XYGlyphView {
             break
           }
           case "max": {
-            let sradius_x = this.sdist(this.renderer.xscale, this._x, this._radius)
-            let sradius_y = this.sdist(this.renderer.yscale, this._y, this._radius)
+            const sradius_x = this.sdist(this.renderer.xscale, this._x, this._radius)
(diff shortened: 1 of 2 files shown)


bokehjs/src/lib/models/glyphs/circle.ts
```diff
@@ -44,6 +44,16 @@ export class CircleView extends XYGlyphView {
             this.sradius = this.sdist(this.renderer.yscale, this._y, this._radius)
             break
           }
+          case "max": {
+            let sradius_x = this.sdist(this.renderer.xscale, this._x, this._radius)
+            let sradius_y = this.sdist(this.renderer.yscale, this._y, this._radius)
```
mattpap: Should be `const`. The same below.

bokehjs/src/lib/models/glyphs/circle.ts
```diff
@@ -44,6 +44,16 @@ export class CircleView extends XYGlyphView {
...
+            this.sradius = map(sradius_x, (s, i) => Math.max(s, sradius_y[i]))
+          }
```
mattpap: Missing `break` statement. The same below.

PR #8445, opened 2018-11-16
Add support for multiline axis labels

### LINKED ISSUE #8169: Line break in categorical axis labels
Feature request (similar to #7317)

Please introduce some straightforward way to include line breaks (e.g. "\n" or "\<br>") in labels of categorical axes (currently these are ignored):

```
factors = ["a very long label \n that needs a break", 
           "a very long label \\n that needs a break",  
           "a very long label <br> that needs a break", 
          ]
x = [50, 40, 65,]
p = figure(y_range=factors)
p.circle(x, factors, size=15, fill_color="orange", line_color="green", line_width=3)
show(p)
```

image

mattpap:
image

At this point align and baseline can be configured (overriding the heuristic), but label positioning will be wrong. It would be nice to fix this here, but this will conflict greatly with the layout PR, so I suppose I will do it later. Also, baseline isn't exactly correct in this setup and I will need to introduce vertical align instead.

fixes #8169

### DIFF
--- a/bokehjs/src/lib/models/axes/axis.ts
+++ b/bokehjs/src/lib/models/axes/axis.ts
@@ -10,6 +10,7 @@ import {Side, TickLabelOrientation, SpatialUnits} from "core/enums"
 import {Text, Line} from "core/visuals"
 import {SidePanel, Orient} from "core/layout/side_panel"
 import {Context2d} from "core/util/canvas"
+import {get_text_height} from "core/util/text"
 import {sum} from "core/util/array"
 import {isString, isArray, isNumber} from "core/util/types"
 import {Factor, FactorRange} from "models/ranges/factor_range"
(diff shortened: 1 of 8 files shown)


bryevdv:
@mattpap I think this would make a nice "centerpiece" feature for a 1.0.2 how much effort is left?

bokehjs/src/lib/models/axes/axis.ts
```diff
@@ -287,20 +326,25 @@ export class AxisView extends GuideRendererView {
     const c = Math.cos(angle)
```
mattpap: @bryevdv, what's the reason to have a different scaling factor depending on the type of the orientation? I'm speaking about `hscale`, but github doesn't allow me to place this comment in the right place.

bryevdv: My only recollection is that it is a fudge factor that more or less lets things work when the orientation was angled at all. The real problem could be in `get_text_height` which is just an approximating function we created, IIRC, since browsers inexplicably don't report text height reliably. I.e. maybe `get_text_height` underestimates consistently, and things look OK for orthogonal orientations, but bad for angled ones, and if it computed a better value `hscale` would be unneeded. Could try setting `hscale` to 1.5 unconditionally and see if things look ok in both orthogonal/angled cases (that would simulate correcting for text height underestimate)

bryevdv:
@mattpap this has conflict now, should we close this in favor of a new pr later?

mattpap:
I will revisit this in the context of future work.

Issue #8448, opened 2018-11-19
Documentation spelling and formatting


mbrown1413:
There are various typos scattered throughout the documentation.

mattpap:
First batch of fixes was merged in PR #8449. I will leave this issue open to get some more fixes in. Spellchecking shouldn't be limited solely to (sphix) documentation. There are also docstrings, comments in code, strings (e.g. validation messages, exception texts), etc. In the longer term, it would be good to set up spell checking as a part of codebase tests in CI (I will start another issue for that).

mbrown1413:
After taking another look at the `bokeh.core.validation` module docs, I realized there are two ways classes and modules and such are currently referenced in the docs:

1. Using double backticks:
```rst
``ColumnDataSource``
```

2. Using pipes: `|ColumnDataSource|`, then replaced (using [Sphinx Substitutions](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#substitutions)) at the bottom of the rst file like this:
```rst
.. |ColumnDataSource| replace:: :class:`~bokeh.models.sources.ColumnDataSource`
```

Right now there is a mix of the two in the docs. Everything should be changed to use method 2, since it actually links to the reference for the class/module. 

The downside is that it will be a pain to write all of the substitutions needed for every file. So I'm thinking of standardizing things by putting substitutions for all bokeh classes and modules in the Sphinx `rst_prolog` (not sure about functions, there could be namespace collisions there). There are already a few global substitutions for classes and functions in there.

It would be nice to automate this, instead of manually updating the substitutions every time the bokeh code changes. Writing a simple Sphinx extension to automatically add substitutions would probably be the cleanest way to do this. I haven't looked too hard yet at what it would take to write such an extension.

Let me know if I'm over-engineering things here. It probably makes sense to do this step-by-step anyways, so I'll definitely wait on the automation.

Of course I used method 1 in my pull requests above, so I'd at least like to fix those before closing this issue.

bryevdv:
I think putting things in `rst_prolog` is definitely the way to go. I would not be opposed to just using a one-time script to convert mentions of `thing` into `|thing|`, especially since we could update docstrings this way, too (it would be quite a bit harder to make an extension to update docstrings)

mbrown1413:
[...]

PR #8453, opened 2018-11-21
Documentation cleanup

mbrown1413:
I'm glad Bokeh is really well documented, but there's a lot more documentation than I thought!

This fixes lots of typos and places where classes/code were referenced without using double backticks. A few docstring rewordings to make them "verb first" style. Some minor formatting fixes too.

Also fixed some formatting typos where `.. code-block::` was used with one colon instead of two. It would be nice if Sphinx/rst would error when that happens, instead of silently displaying nothing.

- [ ] issues: addresses #8448

### DIFF
--- a/bokeh/application/application.py
+++ b/bokeh/application/application.py
@@ -78,7 +78,7 @@ class Application(object):
                 The URL is taken from the first one only.
 
         Keyword Args:
-            metadata (dict): abitrary user-supplied JSON data to make available
+            metadata (dict): arbitrary user-supplied JSON data to make available
                 with the application.
 
                 The server will provide a URL ``http://applicationurl/metadata``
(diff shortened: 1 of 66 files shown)


mattpap:
Thanks @mbrown1413.

bryevdv:
Just FYI I think some of the back-ticking was a bit agressive. The validation messages, e.g. are only ever displayed in the console, there is no point to add backticks there (thanks @mbrown1413 !)

mattpap:
Changes to validation module are limited to its docstring, which may appear in documentation (I'm not sure if it actually does) and already has some other minor formatting. Or maybe you were referring to something different, because I didn't see any excessive changes when reviewing this?

mbrown1413:
I think the confusion is just that there are two descriptions for each error code: the message itself displayed on the console, and description in the docstring, which appears in the documentation for [bokeh.core.validation](https://bokeh.pydata.org/en/latest/docs/reference/core/validation.html).

PR #8460, opened 2018-11-27
fix a typo

kyang922:
sr.data_source -> cr.data_source

### DIFF
--- a/examples/howto/Hover callback.ipynb
+++ b/examples/howto/Hover callback.ipynb
@@ -46,7 +46,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "callback = CustomJS(args={'title': p.title, 'segment': sr.data_source}, code=\"\"\"\n",
+    "callback = CustomJS(args={'title': p.title, 'segment': cr.data_source}, code=\"\"\"\n",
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the catch and fix @kyang922

kyang922:
@bryevdv 
Sorry, it seems not to be a typo. I found the original there
https://bokeh.pydata.org/en/latest/docs/user_guide/interaction/callbacks.html#customjs-for-hover

the 'sr' is used to draw the link line. but in this notebook version, here is no those lines. so it may not be necessary to write this arg here.

kyang922:
@bryevdv 
by the way, I found interaction selection examples in the user guide not works.
https://bokeh.pydata.org/en/latest/docs/user_guide/interaction/callbacks.html#customjs-for-selections

I tried it in my Firefox(63.0.3) and Chrome(70.0.3538), neither worked. I also tried those code in my jupyter with bokeh 1.0.1, same result.

bryevdv:
@kyang922 Yes the example needs to be updated (there is already another open issue for it)

kyang922:
@bryevdv oh, I just saw that issue. Thanks a lot.

Issue #8458, opened 2018-11-26
Simplify the JSON representation


mattpap:
This is a follow up on the idea proposed in issue #8163. This will be resolved in several steps:

1. Simplify the existing representation, which currently is too opaque and contains a lot of unnecessary content. However, this representation, with its flaws fixed, is perfectly suited for automatic processing, server setup, etc., and should be kept as the reference representation. Fixing it will include, possibly among other things:

- [x] don't include `type` in object references (in fact disambiguate refs, because they refer to at least two distinct concepts) (see mattpap/8458_ref_type) (this was done in PR #9333)
- [ ] avoid referencing the same object multiple times (e.g. in layout's `left`, `right`, `above` or `below`, and `renderers`) (in this particular case done in PR #8085)
- [x] remove other unnecessary references, like `plot`  (done in PR #8085)
- [ ] don't include unnecessary default values (investigate unstable defaults, which break symmetry between bokeh and bokehjs)
- [x] don't include `subtype`,  just `type` should be sufficient to describe a model. `subtype` is an internal thing and shouldn't be part of the JSON. (fixed in PR #11665)

2. A secondary, simplified representation should be added. Representations will be disambiguated by either using a JSON field or a different API. Details of the representation will have to be established in this issue. At minimum, a nested structure should be allowed to eliminate most of the references. In case case of referencing the same object multiple times, either references by id or relative references (think XPath as an example) (or both) should be allowed. I suppose more changes will be needed to make this actually user friendly and editable by hand, but this depends on how much we want do deviate from the representation being a simple attr -> value mapping. There is also the matter of supporting Vega(-lite) (issue #4844), which I presume matters in the context of this issue.

bryevdv:
Major parts done, remaining work will be put in follow on issues

PR #8461, opened 2018-11-27
Don't include type when referencing an object

mattpap:
addresses #8458

### DIFF
--- a/bokeh/core/json_encoder.py
+++ b/bokeh/core/json_encoder.py
@@ -243,7 +243,7 @@ class BokehJSONEncoder(json.JSONEncoder):
         elif isinstance(obj, collections.deque):
             return list(map(self.default, obj))
         elif isinstance(obj, Model):
-            return obj.ref
+            return obj.ptr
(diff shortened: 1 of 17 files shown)


bokeh/core/json_encoder.py
```diff
@@ -243,7 +243,7 @@ def default(self, obj):
         elif isinstance(obj, collections.deque):
             return list(map(self.default, obj))
         elif isinstance(obj, Model):
-            return obj.ref
+            return obj.ptr
```
bryevdv: This kind of change needs to occur under an optional feature flag that is off by default. We could imagine flipping it on for 2.0

bryevdv: Surely Python API stability means that Python function calls continue to return the same results on the same inputs, over time. I would not be surprised if there are people who are testing their usage of Bokeh by validating JSON output.

mattpap:
I will close this PR, but preserve the branch for later use.

PR #6162, opened 2017-04-18
2828 multi line input

mas-dse-pjmulroo:
All pull requests must have an associated issue in the issue tracker. If there
isn't one, please go open an issue describing the defect, deficiency or desired
feature. You can read more about our issue and PR processes in the
[wiki](https://github.com/bokeh/bokeh/wiki/BEP-1:-Issues-and-PRs-management).

- [ X ] issues: fixes #2828
- [ - ] tests added / passed
- [ X ] release document entry (if new feature or API change)

Didn't get a chance to write the tests yet. It will take me some time to get to them.

PR #8464, opened 2018-11-28
textarea input widget

### LINKED ISSUE #2828: Multi-line TextInput box?
In general I've found it very difficult to find information on Bokeh widgets in the documentation, so maybe this is already possible, but is there a way to specify a paragraph-sized TextInput box?  I'm using one for the user to be able to edit some code in a domain-specific language, which is awkward to do in a single-line box.  

If multiple lines are not possible, is there at least a way to change the size of the box?

hhuuggoo:

- [x] issues: fixes #2828
- [x] tests added / passed (will follow up, not sure what the right thing is here)


hhuuggoo:
This is a resurrection of #6162

bryevdv:
@hhuuggoo looks like some property defaults mismatch between python/JS:

>TextAreaInput: defaults are out of sync between Python and bokehjs
    TextAreaInput.cols: bokehjs defaults to 200 but python defaults to 20
    TextAreaInput.rows: bokehjs defaults to 5 but python defaults to 2
    TextAreaInput.max_length: bokehjs defaults to 1000 but python defaults to null

bryevdv:
@hhuuggoo FYI I resolves the small conflict I will look at any test failures when these builds complete

bryevdv:
I also added `TextAreaInput` to `__all__` which was also necessary, but did not show up in the conflict

hhuuggoo:
@bryevdv  thanks!

bokeh/models/widgets/inputs.py
```diff
@@ -97,6 +98,37 @@ class TextInput(InputWidget):
...
+    """)
+
+    callback = Instance(Callback, help="""
+    A callback to run in the browser whenever the user unfocuses the TextInput
+    widget by hitting Enter or clicking outside of the text box area.
+    """)
+
+    placeholder = String(default="", help="""
+    Placeholder for empty input field
+    """)
```
bryevdv: @hhuuggoo it looks like `value`, `callback` and `placeholder` are shared with `TextInput`, what do you think about factoring into a `TextInputBase` or similar? I think it would be fine if this is just on the Python side, but it could also be done on the JS side for a small reduction in code size.

hhuuggoo: sure I can do that - I'm also not sure about callback, I saw some initial discussions about removing it but I left it in cause TextInput still has it.

bryevdv: Oh that's a good point, I would say it should be omitted in new classes, and many be deprecated for old ones at some future point (no set plans tho). IN that case the base could just factor `value and `placeholder`

bryevdv:
@hhuuggoo as you do there small refactor, there was also one linter issue (also fallout from the conflict I think):

> E         ./tests/integration/widgets/test_textarea_input.py:27:1: F811 redefinition of unused 'TextAreaInput' from line 26

bryevdv:
cc @mattpap

hhuuggoo:
but any ideas why this won't work?

https://github.com/bokeh/bokeh/blob/11095002b5639cdd66a16f3fda2aefd182294f23/tests/integration/widgets/test_textarea_input.py#L88

looks like the python callback fires ok, but it doesn't propagate back the the js data source (even though the code is almost identical to test_text_input.py

bryevdv:
@hhuuggoo not offhand, I will have to check out the PR branch and run things locally. I can try to do that tonight.

hhuuggoo:
I'll try to repro the test in a bokeh server example first.  Maybe that'll clarify things

hhuuggoo:
HA - so what happens is that `enter_text_in_element` defaults to hitting enter after the string, which serves to unfocus and fire a change event on a text input.  But enter doesn't have the same behavior on textareas since textareas are multiline.  If I modify the test to send a TAB afterwards, then the original logic works fine.

I'll make the recommended changes and push tomorrow

bryevdv:
@hhuuggoo I'd like to cut a micro release soon, any update on this?

hhuuggoo:
@bryevdv sorry been distracted.  fixing remaining issues right now

hhuuggoo:
I eneded up making TextAreaInput inheriting from TextInput (though have to pull out TextInputView into a separate base class.  Should be good to go once CI passes

hhuuggoo:
made fixes w.r.t your comments @bryevdv.  Looking into the import errors now

bryevdv:
@mattpap I'd like to have another micro-release before 1.1 to isolate other small change. So I would like to merge this today I presume that will necessitate some light additions to the layout PR?

bryevdv:
@hhuuggoo can you post an example script and screen shot here?

hhuuggoo:
image

I added a script here:

https://github.com/hhuuggoo/bokeh/blob/wip_textrea/sphinx/source/docs/user_guide/examples/interaction_textarea.py

(but if you want to run it, you'd have to change the mode to 'inline', since the CDN bokehjs does not have the relevant model)

bryevdv:
@mattpap I intend to merge this today and cut 1.0.3 tomorrow, unless there is objection

mattpap:
@bryevdv, sure. This will have to be redone on top of the layout PR, mostly because of removal of bootstrap, but that shouldn't be a big deal.

bryevdv:
Thanks @hhuuggoo

PR #8465, opened 2018-11-28
Bryanv/8457 bkroot for embed

### LINKED ISSUE #8457: embeds with json_item missing toolbar/interactivity
Running this example:

https://github.com/bokeh/bokeh/blob/master/examples/embed/json_item.py

Will display plots, but the toolbars are missing, and none of the interactivity (e.g. panning which should be the active tool) responds either.

@mattpap have you seen this in any other situations, or do you have and ideas what might cause that? The embed goes through nearly the same code path, only a small wrapper around `Bokeh.embed._embed_items` There are no messages in the console log and nothing really jumps out. FWIW the toolbar button divs are present with the correct CSS classes but have 0 height. 

If you can identify a root cause I can take a closer look. This is a somewhat important priority, I have a prototype serverless-based "imgur for Bokeh plots" working, but it's not useful to deploy until this is resolved.

bryevdv:
- [x] issues: fixes #8457
- [x] tests added / passed

cc @mattpap I definitely think this is the appropriate immediate-term solution. If/when we can remove `bk-root` altogether, there will be no change from a user usage perspective.

### DIFF
--- a/bokeh/_testing/plugins/bokeh.py
+++ b/bokeh/_testing/plugins/bokeh.py
@@ -71,6 +71,20 @@ def output_file_url(request, file_server):
 
     return file_server.where_is(url)
 
+@pytest.fixture
+def test_file_path_and_url(request, file_server):
+    filename = request.function.__name__ + '.html'
+    file_obj = request.fspath.dirpath().join(filename)
+    file_path = file_obj.strpath
+    url = file_path.replace('\\', '/')  # Windows-proof
+
+    def tear_down():
+        if file_obj.isfile():
+            file_obj.remove()
+    request.addfinalizer(tear_down)
+
+    return file_path, file_server.where_is(url)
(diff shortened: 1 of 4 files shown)


mattpap:
One thing to note is that there's already code for adding `bk-root` in:

https://github.com/bokeh/bokeh/blob/64bf0ba60660a9abe7a46fff95cc8b362932883b/bokehjs/src/lib/embed/standalone.ts#L27-L30

For some reason it doesn't trigger. I presume the logic embed is still too convoluted.

bryevdv:
Given you want to get rid of `bk-root` entirely some day I'm still inclined to just merge this as-is unless you are objecting.

mattpap:
Given this affects only `embed_item()` (not `embed_items()`), then sure, go ahead.

PR #8466, opened 2018-11-28
Update lasso example documentation (#8393)

### LINKED ISSUE #8393: "CustomJS for Selections" Example in Docs Broken
In the latest version of the docs, it appears [this example]( https://bokeh.pydata.org/en/latest/docs/user_guide/interaction/callbacks.html#customjs-for-selections ) is broken. This is also true of the example in the Bokeh 1.0.0 docs. Selecting points in the plot on the left does not result in points being shown in the right plot. Compare this to [the same plot using Bokeh 0.13.0]( https://bokeh.pydata.org/en/0.13.0/docs/user_guide/interaction/callbacks.html#customjs-for-selections ), which seems to work without issues.

jlcoto:
Updates documentation example of lasso selection to make it work
with bokeh version 1.0.1. Thank you @philippjfr for pointing to/adding solution.
Ref: https://github.com/bokeh/bokeh/issues/8393

Tested new example and produces expected figure:

screen shot 2018-11-28 at 11 03 36 pm

- [x] issues: fixes #8393

### DIFF
--- a/sphinx/source/docs/user_guide/examples/interaction_callbacks_for_selections.py
+++ b/sphinx/source/docs/user_guide/examples/interaction_callbacks_for_selections.py
@@ -18,9 +18,9 @@ p2 = figure(plot_width=400, plot_height=400, x_range=(0, 1), y_range=(0, 1),
             tools="", title="Watch Here")
 p2.circle('x', 'y', source=s2, alpha=0.6)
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @jlcoto !

Issue #8444, opened 2018-11-16
Feature Request: Expose sort_columns in JS DataTable object


jgriffi6:
Hi,

I would like the ability to get the current sort columns and sortAsc of a DataTable within JS.  The motivation is that currently when I sort, then make a selection on another widget that triggers a call back that may result in a change in the ColumnDataSource, then the table is no longer sorted.  What I would like to be able to do, is to collect the sort columns in the callback that changes the data.  Then, I can have the server pre-sort the data and send it back to my application.  In this way the sort will essentially be preserved.  

I have downloaded the trunk, and made some changes in https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/widgets/tables/data_table.ts

[data_table.ts.txt](https://github.com/bokeh/bokeh/files/2590733/data_table.ts.txt)

I've added an array property to DataTable: sort_columns, which is updated in DataTableView.render() specifically this.grid.onSort.subscribe(...).

I think I would have preferred a dynamic function that returns the sortColumns, but could not figure out how to access the 'grid' object from within the DataTable object.

Thanks,
Justin

bryevdv:
> I think I would have preferred a dynamic function that returns the sortColumns, but could not figure out how to access the 'grid' object from within the DataTable object.

It's' not going to be accessible, at least as things are currently. The grid belongs to the view, and models do not have forward references to any views. Agree however that a real model property is not a great solution, as it might seem as though it is user-settable, which is not the case. How about setting a private attribute on the model, and having a public getter (or method) that returns its value?

jgriffi6:
Thanks @bryevdv for the suggestion.  I will submit a PR once I feel comfortable with the update (and hopefully others too once I get a better handle on js/ts!).

About the private suggestion, if I declare a private attribute to the model, then I still need a public setter so that the view, which has the grid, would be able to update the value.  So I think it is better than having a public model property, but there is then still a public method that one could assume they could call unless I am missing something?
[data_table.ts.diff.txt](https://github.com/bokeh/bokeh/files/2627522/data_table.ts.diff.txt)

Thanks,
Justin

PR #8470, opened 2018-11-29
add method to retrieve sorted columns on DataTable in js

jgriffi6:
This is a pull request to implement the feature request documented below.

- [x] issues: feature #8444 
- [ ] tests added / passed
Let me know if you want more.  Attaching compilation log, and a python script that tests the altered DataTable in js.
[build.log](https://github.com/bokeh/bokeh/files/2629793/build.log)

[test_sort_table.py.txt](https://github.com/bokeh/bokeh/files/2629797/test_sort_table.py.txt)

The above script will create a static page which if you then open you can test that we can indeed see which columns are sorted.  sort columns, then press the button. Note, the script will write files into the [bokeh]/tests directory.

Please let me know if the tests should be done differently.

Adding a function 'get_sort_columns()' to the DataTable object

Thanks,
Justin


bryevdv:
@jgriffi6 I was not as verbose in my description as I could have been, but what I was thinking of was this pattern:
```
get sort_columns(): any[] { return this._sort_columns }
```

And then the view would set the private `_sort_columns`. I like this because it has a true property getter (and only a getter) so there is no way to even try to set the value (will raise a compile error)

bryevdv:
@jgriffi6 do you have thoughts on the idea above?

jgriffi6:
@bryevdv 

Yeah I like the idea.  Having unrelated problems now simply setting up the bokehjs compiler.  Will update once I've figured out.

Justin

jgriffi6:
@bryevdv 

I've attempted to follow your suggestion.  One thing I am not clear on: are we trying to make it so that the '_sort_columns' member is unsettable at compile time? Or also at runtime as well?  In my testing script, I create a page, and am able to manipulate the _sort_columns property in the console and w/i js callbacks.  However, I am unable to affect the property through the sort_columns getter.  

Is this desired behavior?

Thanks,
Justin

bryevdv:
> are we trying to make it so that the '_sort_columns' member is unsettable at compile time? Or also at runtime as well? 

From my POV we are trying to give as clear an indication as possible that the value is not intended to be set. Hopefully the leading underscore makes it clear that it is not meant for users to interact with, and if they do try, well, we have done our due diligence.

bryevdv:
Thanks @jgriffi6 !

PR #8472, opened 2018-11-30
Boilerplates for bokeh/models

### LINKED ISSUE #8401: Incorrect Image import from bokeh.models
The content of `bokeh/models/__init__.py` indicates the intention that all model classes should also be importable from `bokeh.models`. However, importing `Image` directly from `bokeh.models` brings into scope a class outside of `bokeh.models` package probably due to name collision with other indirect imports.
```
>>> from bokeh.models.glyphs import Image as I1
>>> from bokeh.models import Image as I2
>>> I1
<class 'bokeh.models.glyphs.Image'>
>>> I2
<class 'bokeh.core.property.visual.Image'>
```
I would expect either consistency of imports, or `bokeh.models` should not provide this interface at all, so that only the first line is a valid option.

bokeh/1.0.1 has this issue
bokeh/0.13.0 doesn't have this issue and correctly imports identical classes

HuntJSparra:
- [x] issues: fixes #8401; partially addresses #8207 
- [ ] tests added / passed

Adds boilerplate for bokeh/models directory


mattpap:
> In order for the module to pass tests, tables.py needs to have "Model" listed in its __all__

That's completely wrong. The rest is Python telling you, in a fairly convoluted way, that imports are recursive. It would be more productive to avoid trying to hack this around and just show us how things fail without this.

bokeh/models/sources.py
```diff
@@ -757,3 +789,23 @@ class AjaxDataSource(RemoteSource):
...
+# Private API
+#-----------------------------------------------------------------------------
+
+def _check_slice(s):
+    if (s.start is not None and s.stop is not None and s.start > s.stop):
+        raise ValueError("Patch slices must have start < end, got %s" % s)
+    if (s.start is not None and s.start < 0) or \
+       (s.stop  is not None and s.stop < 0) or \
+       (s.step  is not None and s.step < 0):
+        raise ValueError("Patch slices must have non-negative (start, stop, step) values, got %s" % s)
```
bryevdv: @HuntJSparra I think the meaning of the sections should be clarified for test file. A section "Private API" here means *tests for private API functions*, not just a place for underscore functions in the test module itself. (In fact, I'm not sure why anything should be underscore in a test module since users don't interact with them anyway, but that's a separate question) I think helper functions like this in test files should go at the top in the "setup" section, or else just nearby where they are used

HuntJSparra: bokeh/models/sources.py is not a test module though, correct?

bryevdv: It's not, my apologies, obviously trying to do too many things at once this morning...

PR #8481, opened 2018-12-05
DOC: remove extra "to execute" from embed

mikofski:
in the introductory paragraphs of ["Embedding Plots and Apps"](https://bokeh.pydata.org/en/latest/docs/user_guide/embed.html#embedding-plots-and-apps) on the distinction between bokeh apps and standallone HTML, IMHO the text reads better if "to execute that execute" is changed to just "that execute":

>These are Bokeh documents that are backed by a Bokeh Server. In addition to all the features of standalone documents, it is also possible to connect events and tools to real Python callbacks, ~to execute~ that execute in the Bokeh server.

- [ ] issues: fixes #xxxx
- [ ] tests added / passed


bryevdv:
Thanks for the catch @mikofski !

PR #8485, opened 2018-12-06
update sampledata bucket URL

bryevdv:
This changes the S3 bucket to use one under the control of the Bokeh AWS account. 

I had hoped to make a new bucket, rename the old `bokeh_data` bucket , then change the new bucket name to `bokeh_data`, so that the URL would not be different. However, this is no longer possible. AWS has restricted the namespace for buckets, new bucket names may no longer contain underscores at all. 

I have not changed or deleted the old bucket, and have no plans to, but cannot guarantee it will exist in perpetuity.

### DIFF
--- a/bokeh/util/sampledata.py
+++ b/bokeh/util/sampledata.py
@@ -53,7 +53,7 @@ def download(progress=True):
     data_dir = external_data_dir(create=True)
     print("Using data directory: %s" % data_dir)
 
-    s3 = 'https://s3.amazonaws.com/bokeh_data/'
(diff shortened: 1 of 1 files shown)


bryevdv:
This is failing on Appveyor (only) with 
```
ERROR: <urlopen error [Errno 11001] getaddrinfo failed>
```
No idea why, will have to look into later.

bryevdv:
I've emailed appveyor support as I can't find any good reason this should not work.

bryevdv:
Failing due to assembling the full URL with `join` instead of `urljoin`, but no idea in that case, why it was not failing all along.

Issue #7762, opened 2018-04-05
Copy & paste from datatable


rupgr:
I don't believe it is possible already, but could you add the ability to copy & paste from the datatable.
My use case would be for copying and pasting between Bokeh and Excel.
Many thanks!

bryevdv:
It appears there is nothing built into SlickGrid, but there may be some hope to follow this example:

http://6pac.github.io/SlickGrid/examples/example-excel-compatible-spreadsheet.html

Nicoowr:
I took a look at the code in the link and it appears we need the SlickGrid object to set up the copy/paste tool : 
      `grid = new Slick.Grid("#myGrid", data, columns, options);`
      `grid.setSelectionModel(new Slick.CellSelectionModel());`
      `grid.registerPlugin(new Slick.AutoTooltips());`
Do you know how we could get it from bokeh? So far, using customJS, I have only been able to get the "div" object containing the table. I could recreate the table from it in JS but this seems an absurd way to do it...

PR #8504, opened 2018-12-14
Enable copy paste from Datatable

alhoo:
Add support for copy pasting data from datatables

Very simple implementation of copy pasting. Probably needs some more work and no unit tests with this commit, but hopefully gives a quick start to finalizing the rest. Fixes issue #7762.


alhoo:
The resulting paste is not ordered correctly and/but the paste also contains an additional column with the original line number :D

alhoo:
Apparently the slickgrid copy manager captures/modifies some click events and causes some of the current unit tests to break (atleast tests/integration/widgets/tables/test_cell_editors.py::Test_*Editor::test_editing_updates_source).

With this copy-paste-plugin I need to use a triple-click with the mouse to edit the values (tested manually with firefox and chrome) where the current master only needs 2 clicks. We could either accept the new behaviour and change our unit tests until the plugin gets fixed by slickgrid or then the slickgrid plugin needs to be updated. Slickgrid has atleast the following known issues which might be related: #1101 and #994.

As for the incorrectly ordered rows: the copy manager seems to get the badly ordered list from the selection tool so the problem might be in the RowSelectionModel, so once again not a problem in the Bokeh code.

alhoo:
I got to agree with @bryevdv that the changed cell editing logic is annoying to use. I disabled the copy-paste feature from editable datatables, which is also a bit annoying but I guess it is less annoying for the average user.

I guess some users would like to have the copy-paste also with editable fields even if that breaks/changes the editing gesture. Maybe we could add a [force] copy_paste=True [and change editing gesture... sorry about that] -flag for that?

bryevdv:
@alhoo we consider putting it under a feature flag. My main concern is that when/if it is fixed upstream, we would be left with a no-longer-useful flag that would need to be kept around indefinitely deprecated in order to no break anyone's code. Before doing that, I would like to have at least engaged minimally with the devs working on the fork, to make sure there is not some simple fix that could be applied, or that it's not something they plan to imminently work on. Can you open up an issue on the issue tracker for the 6pac fork to ask about this? (Please link back to this issue)

alhoo:
@bryevdv I tried to replicate the problem with slickgrid, but apparently the problem does not occur in the 6pac.github.io (master).

alhoo:
I made some more testing and the triple-click problem seems to be fixed starting with the SlickGrid commit 9463a06 (release v2.3.18 - update github.io).

bokehjs/src/lib/models/widgets/tables/data_table.ts
```diff
@@ -255,6 +256,21 @@ export class DataTableView extends WidgetView {
...
+        const pluginOptions = {
+          dataItemColumnValueExtractor: function(val: any, col: any) {
+            let value: any = val[col.field]
+            if ( typeof value === 'string'){
+              value = value.replace(/\n/g, "\\n")
+            }
+            return value
+          },
+          includeHeaderWhenCopying : false,
+        }
```
mattpap: This code needs a cleanup. Please fix whitespace around `(`, `{` and `:`. `typeof` should be replaced with `isString()` function (from `core/util/types`). `any` types need to be replaced with more specific types or, if there is no other better choice, there needs to be a comment explaining why `any` is being used, including the best approximation of a possible type.

alhoo: @mattpap I was unable get any output from tslint, and py.test -m codebase doesn't seem to test the typescript syntax. I could not find instructions on the "Getting Set Up" -guide. I tried to fix the issues by guessing what exactly you meant. Hopefully that is close enough.

alhoo: Argh. Line 260 still needs a space before `{`.

bryevdv:
> I made some more testing and the triple-click problem seems to be fixed starting with the SlickGrid commit 9463a06 (release v2.3.18 - update github.io).

That's good news, I will try updating our fork but FYI it will probably not be until late this week before I can get to it.

bryevdv:
OK just as a quick test I tried to merge the upstream master and as expected our changes do not apply cleanly. @mattpap is probably the best person to take a look at this, since he made the changes to our fork. Ideally we can reduce/eliminate the differences between our fork and the upstream.

alhoo:
I found the [problematic line](https://github.com/alhoo/SlickGrid/commit/361041e0409e445d3bcc7d5393ed030a40471c71) from the bokeh/slickgrid, but it would be nice to get the full upgrade to slickgrid.

mattpap:
>  @mattpap is probably the best person to take a look at this

Yeah, but not earlier than the layout is done.

bryevdv:
@alhoo this is pretty fantastic. Can you comment on how far `d193f8a` is from the current HEAD of `6pac/slickgrid`? It looks like you have consolidated all the commits from our for into one patch, would it be possible to split them up? Currently they are also failing out linter due to trailing whitespace. If that's an issue we could exclude the patch file. 

@mattpap can you comment? It looks like the layout PR has almost no changes to anything data table related, so it seems like this work would not conflict. I'd be interested to get this feature in and turn out a 1.0.4 release next week. 

Are our changes something that the upstream might consider if we made a PR?

alhoo:
Oh and d193f8a is the current HEAD.

bryevdv:
@alhoo I have merged the PR in `bokeh/SlickGrid` can you update`package.json` to build from the new head, and do some minimal testing, then submit a new PR?

PR #8506, opened 2018-12-16
Boilerplate for bokeh/plotting directory

HuntJSparra:
- [ ] issues: addresses #8207
- [x] tests added / passed

Adds boilerplate to the modules and tests under bokeh/bokeh/plotting and its subdirectories.


bryevdv:
@HuntJSparra it's debatable whether the transitive imports in `bokeh.plotting` should have been allowed/demonstrated in the first place, but it's definitely the case that they are part of the official API now. In general, if we intend a transitive import to be part of the official API, we do something like 
```
from .foo import bar; bar
```
in order to make an explicit indication about it. So anything you see imported this way can safely be considered part of the public API (and should be included in `__all__`

PR #8510, opened 2018-12-18
monitor JS package license usage

### LINKED ISSUE #8495: Document Licenses of included projects
We're trying to be diligent about (checking/) showing all the licenses of the dependencies we use in our project, and to this end, it proves rather difficult to extract the licenses of the bundled js dependencies that bokeh ships.

To my understanding it is not enough to show for bokeh, that it is just under the new BSD-Clause, as all the bundled dependencies licenses would also need to be shown (as we are in effect using them too).

My workaround for now is to manually install bokeh from source, build BokehJS and then inspect the built up NPM-repo for all the installed tools, it would be much better however if the Bokeh-project has an always up to date document (on the website or in the repository) that lists all the licenses that it contains - bonus points if it's a plaintext file.

What do you think? Can you provide such a document?

bryevdv:
- [x] issues: fixes #8495
- [x] tests added / passed

cc @dwt @mattpap

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -26,3 +26,25 @@ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
 CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 THE POSSIBILITY OF SUCH DAMAGE.
+
+
+
+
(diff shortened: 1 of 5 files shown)


bryevdv:
@mattpap this seems like it should be uncontroversial and as we are more on a trajectory of reducing dependencies rather than adding them, I don't expect it will trip up any dev work frequently.

dwt:
I like this, it's a great step in the right direction.

bryevdv:
As it turns out putting things in one file thwarts the GH license detection, which I rather liked having. So I will probably circle back on this at some point soon and move the summary info in the license file to the Documentation somewhere, and generate a separate file at build/package time that lists all the BokehJS licenses explicitly.

PR #8522, opened 2018-12-26
Enable copy paste for datatables

### LINKED ISSUE #7762: Copy & paste from datatable

alhoo:
Use slickgrid external copy paste manager to add copy paste support for bokeh datatables.

- [x] issues: fixes #7762
- [x] tests added / passed

### DIFF
--- a/bokehjs/package-lock.json
+++ b/bokehjs/package-lock.json
@@ -1933,6 +1933,43 @@
       "integrity": "sha1-m387DeMr543CQBsXVzzK8Pb1nZQ=",
       "dev": true
     },
+    "patch-package": {
+      "version": "5.1.1",
+      "resolved": "http://registry.npmjs.org/patch-package/-/patch-package-5.1.1.tgz",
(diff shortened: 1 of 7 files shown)


bryevdv:
This worked well with local testing and has good new tests. I can't see any reason not to merge. 

cc @mattpap the 6pac repo seems at least moderately active, I'd suggest we propose upstream changes to get rid of the need for our fork

bokehjs/package-lock.json
```diff
@@ -1933,6 +1933,43 @@
       "integrity": "sha1-m387DeMr543CQBsXVzzK8Pb1nZQ=",
       "dev": true
     },
+    "patch-package": {
```
mattpap: Where this come from? It isn't in `package.json`.

mattpap: `npm install` needs to be run and changes to this file commited.

mattpap:
> I'd suggest we propose upstream changes to get rid of the need for our fork

We can try, but that may be a hard sale given that we use completely different module and packaging scheme.

bokehjs/package-lock.json
bryevdv: ah missed that, it's cruft from the other version of the PR. I will fix it up in another small PR

PR #8526, opened 2018-12-28
Don't resize window when capturing test images

### LINKED ISSUE #8525: Don't resize window when running images tests
This is inefficient and results in incorrect images, which is particularly problematic with baseline tests from PR #8085.

mattpap:
fixes #8525

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ jobs:
       addons:
         chrome: stable
       before_script:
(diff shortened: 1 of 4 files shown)


bryevdv:
@mattpap do you want this in before 1.0.3 or after? I have some small docs and release notes to write, so will make the release Monday.

mattpap:
After is fine. This is mostly relevant to the layout PR.

mattpap:
Unfortunately this doesn't play well with custom templates.

PR #8530, opened 2018-12-30
Boilerplate for bokeh/sphinxext and bokeh/template

HuntJSparra:
- [x] issues: addresses #8207 
- [x] tests added / passed

Adds boilerplate for bokeh/sphinxext and bokeh/template, the latter being included due to the fact it is only 5 files.


HuntJSparra:
@bryevdv 
bokeh/sphinxext has no tests as far as I can tell. Do you have any advice for how to check things on my end? I am planning to locally build the docs on the 2nd once I am on a computer with more memory. Would that be reasonable?

bryevdv:
@HuntJSparra yes the only tests now are just to build the docs, it's rather unwieldy to test sphinx extensions directly unfortunately.

HuntJSparra:
I was able to locally build successfully and everything looks in order.

Issue #3529, opened 2016-01-04
Unify dev and non-minified resources


mattpap:
There was a distinction when we used requirejs (loading individual files), but now it's only confusing. By default all resources should point to minified versions and under `dev` should point to non-minified.

bryevdv:
@mattpap does your new PR deal with this?

PR #8537, opened 2019-01-03
Birdsarah/8534 docs tweak

### LINKED ISSUE #8534: Docs tweak to add note about BOKEH_DEV and apps
Make clearer need to change BOKEH_RESOURCES too.

birdsarah:
Fixes #8534

Here is a screenshot of the docs fix rendered

screenshot from 2019-01-03 09-58-35

To Do:
- [ ] bokeh/settings.py docstring update
- [ ] note about notebook not working with bokeh dev (check first)

Not doing:

* After checking with @bryevdv, I'm just going to leave it to be "server" as the recommendation and maybe we can clean-up all these options another time.

### DIFF
--- a/sphinx/source/docs/dev_guide/env_vars.rst
+++ b/sphinx/source/docs/dev_guide/env_vars.rst
@@ -34,7 +34,6 @@ This is a meta variable equivalent to the following environment variables:
 - ``BOKEH_PRETTY=true``
 - ``BOKEH_PY_LOG_LEVEL=debug``
 - ``BOKEH_RESOURCES=absolute-dev``
-- ``BOKEH_SIMPLE_IDS=true``
 
 Accepted values are ``yes``/``no``, ``true``/``false`` or ``0``/``1``.
 
--- a/sphinx/source/docs/dev_guide/env_vars.rst
+++ b/sphinx/source/docs/dev_guide/env_vars.rst
@@ -38,6 +38,11 @@ This is a meta variable equivalent to the following environment variables:
 
 Accepted values are ``yes``/``no``, ``true``/``false`` or ``0``/``1``.
 
+.. note::
+    When running server examples, the ``BOKEH_RESOURCES`` setting that
+    ``BOKEH_DEV`` sets will cause the page to stop rendering. So you 
+    will need to manually also set ``BOKEH_RESOURCES=server``.
+
 ``BOKEH_DOCS_CDN``
 --------------------
 What version of BokehJS to use when building sphinx docs.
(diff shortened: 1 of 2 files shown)


sphinx/source/docs/dev_guide/env_vars.rst
```diff
@@ -34,10 +34,14 @@ This is a meta variable equivalent to the following environment variables:
...
 - ``BOKEH_PY_LOG_LEVEL=debug``
 - ``BOKEH_RESOURCES=absolute-dev``
-- ``BOKEH_SIMPLE_IDS=true``
 
 Accepted values are ``yes``/``no``, ``true``/``false`` or ``0``/``1``.
 
+.. note::
+    When running server examples, the ``BOKEH_RESOURCES`` setting that
+    ``BOKEH_DEV`` sets will cause the page to stop rendering. So you 
+    will need to manually also set ``BOKEH_RESOURCES=server``.
```
mattpap: Given this is a dev guid, this should be `server-dev`.

bryevdv: `server-dev` just sets to use unminified resources, which`BOKEH_DEV` already does? All things being otherwise equal, less typing is better

birdsarah: Nice catch - thanks @mattpap

mattpap: I overlooked BOKEH_MINIFIED, though without looking at the code I'm not sure what takes precedence. In fact I intended to fix this for a while (#3529).

bryevdv:
@birdsarah there is a big docstring:

https://github.com/bokeh/bokeh/blob/master/bokeh/settings.py

It's probably worth updating that to match. I thought the docs just rendered the docstring but I guess that changed for some reason.

sphinx/source/docs/dev_guide/env_vars.rst
mattpap: Additionally, I think notebook also needs some help to work width BOKEH_DEV (setting BOKEH_RESOURCES=inline I suppose).

bryevdv: `BOKEH_DEV=true BOKEH_RESOURCES=server bokeh serve --show sliders.py` uses unminified resources, which is also what I would expect.

mattpap:
Tests fail due to `pull_session()` issues (simple IDs on the server I presume) (https://s3.amazonaws.com/bokeh-travis/474916271/examples.log).

bryevdv:
Annoying, but only a spurious failure it seems. There's only just the codebase issue about trailing whitespace.

Issue #8513, opened 2018-12-18
Strip out IPython magics when serving notebooks


jlstevens:
When nbconvert is used in `NotebookHandler` to convert notebooks into vanilla Python code ( supporting `bokeh serve notebook.ipynb`) any IPython magics are converted into a call involving `get_ipython` which is then undefined and causes a failure.

This means that any notebook using one of the common magics such as `%time` or `%%timeit` cannot be served with bokeh server. Supporting magics in bokeh server probably isn't technically possible (and typically wouldn't even make much sense) so it would probably be better to strip them out with an nbconvert preprocessor to allow the notebook to be served in some form (instead of always resulting in an error).

jbednar:
This functionality would be really useful to me when writing serveable notebooks.  It might be helpful to print a warning message saying that IPython-specific code XXX has been stripped out in case any one is confused.

bryevdv:
Happy to have a PR for this.

PR #8539, opened 2019-01-04
Preprocessor to strip out IPython magics from notebooks

jlstevens:
First cut at addressing #8513.

Using this tiny notebook called 'Test.ipynb':

image

On master, running `bokeh serve --show Test.ipynb` gives you:

image

With this PR:

image

Right now this is a first cut. Things to think about:

- [ ] Is this the best way to detect magics? Probably not as a line of valid Python could start with a  bare `%` e.g an expression in parentheses or an expression using `\` to break across lines. Checking for an alphanumeric identifier immediately after `%` doesn't really help as this is valid Python too:

   ```python
   foo = 2
   4 \
   %foo
   ```

  I suppose we could make an explicit blacklist of the most common magics to be safe?

- [ ] Should we warn when magics are detected? (probably)
- [ ] Avoid inlining the `StripMagicsProcessor` class (currently done to use bokeh's approach to importing `nbconvert` inline).

### DIFF
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -75,11 +75,42 @@ class NotebookHandler(CodeHandler):
 
         if 'filename' not in kwargs:
             raise ValueError('Must pass a filename to NotebookHandler')
+
+
+        class StripMagicsProcessor(nbconvert.preprocessors.Preprocessor):
+            """
+            Preprocessor to convert notebooks to Python source while stripping
+            out all magics (i.e IPython specific syntax).
+            """
+
+            def strip_magics(self, source):
+                """
+                Given the source of a cell, filter out all cell and line magics.
+                """
+                filtered=[]
+                for line in source.splitlines():
+                    if not line.startswith('%'):
+                        filtered.append(line)
+                return '\n'.join(filtered)
+
+            def preprocess_cell(self, cell, resources, index):
+                if cell['cell_type'] == 'code':
+                    cell['source'] = self.strip_magics(cell['source'])
+                return cell, resources
+
+            def __call__(self, nb, resources):
+                return self.preprocess(nb,resources)
+
+        preprocessors=[StripMagicsProcessor()]
         filename = kwargs['filename']
 
         with open(filename) as f:
             nb = nbformat.read(f, nbformat.NO_CONVERT)
             exporter = nbconvert.PythonExporter()
+
+            for preprocessor in preprocessors:
+                exporter.register_preprocessor(preprocessor)
+
             source, _ = exporter.from_notebook_node(nb)
             kwargs['source'] = source
 
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -92,12 +92,12 @@ class NotebookHandler(CodeHandler):
                 Given the source of a cell, filter out all cell and line magics.
                 """
                 filtered=[]
-                for num, line in enumerate(source.splitlines()):
+                for line in source.splitlines():
                     if self._magic_pattern.match(line) is None:
                         filtered.append(line)
                     else:
-                        msg = 'Stripping out IPython magic detected on line {num} of cell {cell}: {line}' #%r' % line
-                        message = msg.format(num=num, cell=self._cell_counter, line=repr(line))
+                        msg = 'Stripping out IPython magic detected in code cell {cell}: {line}' #%r' % line
+                        message = msg.format(cell=self._cell_counter, line=repr(line))
                         log.warn(message)
                 return '\n'.join(filtered)
 
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -85,7 +85,7 @@ class NotebookHandler(CodeHandler):
             """
 
 
-            _magic_pattern = re.compile('^\s*%+\w\w')
+            _magic_pattern = re.compile('^\s*%+\w\w+($|(\s+))')
 
             def strip_magics(self, source):
                 """
--- a/bokeh/application/handlers/notebook.py
+++ b/bokeh/application/handlers/notebook.py
@@ -84,8 +84,7 @@ class NotebookHandler(CodeHandler):
             out all magics (i.e IPython specific syntax).
             """
 
-
-            _magic_pattern = re.compile('^\s*%+\w\w+($|(\s+))')
+            _magic_pattern = re.compile('^\s*(?P<magic>%+\w\w+)($|(\s+))')
 
             def strip_magics(self, source):
                 """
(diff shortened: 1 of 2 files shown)


jbednar:
Based on https://ipython.readthedocs.io/en/stable/interactive/magics.html, rather than a blacklist I'd vote for checking for a two-character or more alphanumeric string after the leading %, and for warning when one is skipped.  Alternatively, would it be possible to ignore them later more safely by simply defining get_ipython to be a no-op?  Also, could there be an option to actually run the magics with ipython instead, so that they would do whatever they were meant to do (if optionally enabled)?

bryevdv:
I'm incredulous that python does not provide a safe API for stripping magics.

jbednar:
From https://github.com/ipython/ipython/issues/3707/ ; it doesn't seem like anything suitable is available.

jlstevens:
I would also love there to be a suitable API to do this but I haven't seen one yet (and the issue Jim links to suggests there isn't one). The thing is, even *if* there were such an API, it would probably have to involve `IPython` machinery in bokeh (as opposed to just `nbformat` and `nbconvert`) which at minimum would make `IPython` an extra optional dependency. For instance, I can imagine trying to find the names of all the magics registered to try to strip them out safely, but that would need `get_ipython` to exist.

One thought might be to run the code in an IPython context (if IPython is available) in which case `get_ipython` would exist and no stripping of magics would be required.

At any rate, I'm happy to use any method that works reliably but currently I don't know of a better way that this type of source transformation. I suspect that simply stripping out the magics in this way is the simplest approach and will work in the vast majority of cases. If we also add warnings when stripping out the magics, it would also give an indication to the user when the magic detection heuristic fails.

bryevdv:
As all of this involves and is related to notebook things, I am happy to have `ipython` be imported as long as it is inside the relevant functions and not at the top level.

bryevdv:
Well, I'll defer to you guys. If there's a way to raise the issue again upstream, it seems advised.

jlstevens:
I think it is also worth reiterating that even an imperfect heuristic to strip magics (that works most of the time) might be better than the current situation where any notebook using any magic will definitely fail.

jbednar:
I propose we allow magics to be stripped out entirely for now, while imagining that at some point one could set it up to run using IPython as an option.

jlstevens:
> I propose we allow magics to be stripped out entirely for now, while imagining that at some point one could set it up to run using IPython as an option.

That seems reasonable. In the long term I can imagine something like this:

1) The `CodeHandler` class takes the `CodeRunner` class as an optional argument in the constructor. The default would be the existing `CodeRunner` class to avoid breaking things, keeping everything unchanged.
2) `NotebookHandler` (a subclass of `CodeHandler`) changes this argument to a new `IPyRunner` class if IPython is available.

This much refactoring would be fairly easy - the tricky bit is writing `IPyRunner` to do everything `CodeRunner` does but in an IPython context.

jlstevens:
The regexp I propose to match magics is:

```
pattern = re.compile('^\s*%+\w\w')
```

i.e start of the line, followed by any amount of whitespace, followed by one or more % symbols, followed by two alphanumeric characters.

jlstevens:
Here is an example of a warning due to a detected magic:

image

jbednar:
Looks good!  I guess the line number can't easily follow the word line in the error message? Ok either way.

jlstevens:
I think this regexp is slightly more restrictive (a magic name must be an identifier afaik):

```
pat = re.compile('^\s*%+\w\w+($|(\s+))')
```

>  I guess the line number can't easily follow the word line in the error message?

I'm not quite sure what that means in a notebook context: does that include markdown? What I could do is report the code cell number and then the line number in that cell. Could be something like:

```
Stripping out IPython magic detected on line 3 of code cell 4: '%%timeit'
```

Would that still be useful?

jbednar:
I think just the code cell number is sufficient, personally; "detected in code cell 4".

jlstevens:
Ok, here is an example of the new warning:

```
WARNING:bokeh.application.handlers.notebook:Stripping out IPython magic detected in cell 3: '%%timeit'
```

jlstevens:
Example of the message since the last commit:

```
WARNING:bokeh.application.handlers.notebook:Stripping out IPython magic %%timeit detected in code cell 2
```

jlstevens:
Slightly shortened:

```
WARNING:bokeh.application.handlers.notebook:Stripping out IPython magic %%timeit in code cell 2
```

jbednar:
Looks good to me, thanks! Ready to merge?

jlstevens:
I think I'm done with the code. The question now is about tests: I'm not sure what tests would be suitable (if any). I believe the code is quite straightforward so I'm not sure what kinds of tests would be most useful. I suppose the most annoying problem setting up tests would be creating notebook files for testing or having to inline the notebook JSON somewhere...

bryevdv:
> I suppose the most annoying problem setting up tests would be creating notebook files for testing or having to inline the notebook JSON somewhere...

Or you can just create a temp notebook programmatically as the notebook handler tests do:

https://github.com/bokeh/bokeh/blob/master/bokeh/application/handlers/tests/test_notebook.py

philippjfr:
So I've only partially followed the discussion here, but I had one quick question/comment, which you can feel free to shoot down immediately. Basically I was wondering whether you couldn't let nbconvert handle the conversion as normal, and instead of adding a pre-processor you would instead add a post-processor, which removes any lines containing ``get_ipython()`` which is what nbconvert uses to replace the magics by default.

bryevdv:
>  which removes any lines containing get_ipython() which is what nbconvert uses to replace the magics by default.

I suppose if users are using `get_ipython` directly themselves then things are not going to work anyway. But to be careful I think it would need to handle
```
x = 10; get_ipython(); y =20
```

jlstevens:
@bryevdv The two unit tests I attempted to added in the last commit seem to be passing for all travis builds except for Python 2.7. Any idea why `handler._runner.failed` indicates that the code failed to execute but only for 2.7?

bryevdv:
Probably a str/unicode issue, as a first guess

bryevdv:
@jlstevens if I add a failing check on the value of `error_detail` it shows:
```
>       assert handler._runner.error_detail is 20
E       assert 'Traceback (most recent call last):\n  File "/Users/bryanv/work/bokeh/bokeh/application/handlers/code_runner.py", line...\n  File "/Users/bryanv/work/bokeh/build/tmp/tmpxNUM5H", line 0\nSyntaxError: encoding declaration in Unicode string\n' is 20
E        +  where 'Traceback (most recent call last):\n  File "/Users/bryanv/work/bokeh/bokeh/application/handlers/code_runner.py", line...\n  File "/Users/bryanv/work/bokeh/build/tmp/tmpxNUM5H", line 0\nSyntaxError: encoding declaration in Unicode string\n' = <bokeh.application.handlers.code_runner.CodeRunner object at 0x112fe0bd0>.error_detail
E        +    where <bokeh.application.handlers.code_runner.CodeRunner object at 0x112fe0bd0> = <bokeh.application.handlers.notebook.NotebookHandler object at 0x10344f110>._runner

```

bryevdv:
FYI That error means that there is a line like `# -*- coding: utf-8 -*-` in the code that is exec'ed but I am not sure where it would come from. I suppose there is an outside chance that `bokeh serve foo.ipynb` does not work at all currently on python 2

bryevdv:
Hrm, running by hand I also get:

```
In [13]: handler.error
Out[13]: u'Invalid syntax in "tmpgy83FB" on line 0:\nNone'
```

jlstevens:
I wonder where that `None` comes from!

PR #8545, opened 2019-01-07
remove warning message aboug gulp

### LINKED ISSUE #8533: Remove warning about `gulp build` in prepare.js
I was confused when running through the Getting Set Up instructions that I got a warning about using `gulp build`, which I wasn't aware I'd tried to do. It turned out I hadn't that's just printed to the console everytime `prepare.js` is run.

If I have a clean setup gulp isn't even installed.

I think it would be good to remove this warning to reduce confusion.

bryevdv:
- [x] issues: fixes #8533

I realize this is marked `good first issue` but I want to clear down the tracker, also FWIW I think good first issues should have *slightly* more teeth

### DIFF
--- a/bokehjs/prepare.js
+++ b/bokehjs/prepare.js
@@ -10,5 +10,3 @@ function write_hash(file) {
 }
 
 write_hash("package.json")
-
-const chalk = require("chalk")
--- a/bokehjs/prepare.js
+++ b/bokehjs/prepare.js
@@ -12,7 +12,3 @@ function write_hash(file) {
 write_hash("package.json")
 
 const chalk = require("chalk")
-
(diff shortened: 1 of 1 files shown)


bokehjs/prepare.js
```diff
@@ -12,7 +12,3 @@ function write_hash(file) {
 write_hash("package.json")
 
 const chalk = require("chalk")
```
mattpap: This import should be removed as well.

bryevdv: done in https://github.com/bokeh/bokeh/pull/8545/commits/d037b17ab39d22b79b7b8c00e9080f20f0010891

PR #8546, opened 2019-01-07
Simple approach to replacing line and cell magics

jlstevens:
This PR is a simpler and safer alternative to #8539 - it lets nbconvert try to convert the magics and *then* removes `run_cell_magic` and `run_line_magic` calls.

Current it doesn't issue warnings but warnings are now less important now that false positives are far more unlikely.


jlstevens:
I realize that this won't work well for cell magics. I'll close this PR and fold the approach for handling line magics back into #8539.

Issue #8548, opened 2019-01-07
Add small documentation to slider callback_policy which only apply to CustomJS


slamer59:
The callback_policy throttling options only apply to CustomJS callbacks, not python/server callbacks.

Some comments will be added and warning. 

This is expected behavior, discussed here: 
https://github.com/bokeh/bokeh/issues/6897, 
https://github.com/bokeh/bokeh/issues/4540
https://github.com/bokeh/bokeh/pull/3958

PR #8549, opened 2019-01-07
Add documentation on callback policy for slider

slamer59:

- [x] issues: fixes https://github.com/bokeh/bokeh/issues/8548

### DIFF
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -15,6 +15,8 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import logging
+
+
 log = logging.getLogger(__name__)
 
 #-----------------------------------------------------------------------------
@@ -31,8 +33,9 @@ import numbers
 from ...core.has_props import abstract
 from ...core.properties import Bool, Int, Float, String, Date, Enum, Tuple, Instance, Color, Override
 from ...core.enums import SliderCallbackPolicy
-from ...core.validation import error
+from ...core.validation import error, warning
 from ...core.validation.errors import EQUAL_SLIDER_START_END
+from ...core.validation.warnings import SLIDER_CALLBACK_APPLY_TO_CUSTOM_JS
 from ..callbacks import Callback
 from .widget import Widget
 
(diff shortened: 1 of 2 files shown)


bokeh/models/widgets/sliders.py
```diff
@@ -110,6 +116,13 @@ def _check_missing_dimension(self):
             if self.start == self.end:
                 return '{!s} with title {!s}'.format(self, self.title)
 
+
+    @warning(SLIDER_CALLBACK_APPLY_TO_CUSTOM_JS)
+    def _check_callback_policy_called_with_js_custom(self):
+        if hasattr(self, 'callback_policy') and not hasattr(self, 'callback'):
+            return 'Callback policy currently apply to JS callbacks'
```
bryevdv: Hi @slamer59 as I mentioned on gitter, just checking for `callback` is too restrictive. There is also now `js_on_change('value', ...)` which is actually what I would regard as current best practice. Those are stored here in the `. js_property_callbacks` property on all models:

https://github.com/bokeh/bokeh/blob/master/bokeh/model.py#L335

That would need to be checked for a key `'value'`.

However, I also in looking at this that that is not right either. Because it should be harmless to set a callback policy with *no callbacks* but that is not the case. What really would need to happen is that the Model needs to be checked for the positive presence of a Python callback on `"value"` That's unfortunately a bit down in the weeds to accomplish. I am happy to try and provide more guidance if you want to pursue it, but also happy for a PR that just updates the docs and nothing else.

bryevdv: And thinking about it even more, I am not sure it is appropriate to warn at all. There can well be valid use cases for having a callback policy to affect a JS callback and at the same time having a python callback that will be known and expected to run every change. I think for now just a documentation update is the best course.

bokeh/models/widgets/sliders.py
```diff
@@ -99,6 +102,9 @@ def __init__(self, **kwargs):
     * "mouseup": the callback will be executed only once when the slider is released.
 
     The "mouseup" policy is intended for scenarios in which the callback is expensive in time.
+
+    .. warning::
+    Callback policy currently apply to JS callbacks
```
bryevdv: I believe this will need to be indented under `.. warning::` but compare to other instances to double check

bokeh/models/widgets/sliders.py
slamer59: Oh. I totally miss your point on this. My intention was more to write something to check if callback is a CustomJS instance:

```
if hasattr(self, 'callback_policy') and not isinstance(callback, CustomJS):
```

I'm not sure to understand all machinery and will just update documentation consequently.

Thanks

bokeh/models/widgets/sliders.py
slamer59: I check the proper indentation

bryevdv:
Thanks @slamer59 I hope we are able to implement a proper solution for python callbacks in the not too distant future.

PR #8555, opened 2019-01-08
Unicode fix when serving notebooks on Python 2

### LINKED ISSUE #8554: Bokeh cannot serve notebooks on Python 2
Using a notebook containing a single code cell (simply  `1+1`) in a Python 2 environment: 

`bokeh serve --show Test.ipynb`

Results in the following error:

```
SyntaxError: encoding declaration in Unicode string
```

Note that this is why the unit tests added in #8546 skip for Python 2.

jlstevens:
Quick PR to try to fix #8554.

### DIFF
--- a/bokeh/application/handlers/code_runner.py
+++ b/bokeh/application/handlers/code_runner.py
@@ -73,6 +73,7 @@ class CodeRunner(object):
         self._code = None
 
         try:
+            source = str(source) if sys.version_info.major == 2 else source
             nodes = ast.parse(source, path)
             self._code = compile(nodes, filename=path, mode='exec', dont_inherit=True)
         except SyntaxError as e:
--- a/bokeh/application/handlers/code_runner.py
+++ b/bokeh/application/handlers/code_runner.py
@@ -73,7 +73,6 @@ class CodeRunner(object):
         self._code = None
 
         try:
-            source = str(source) if sys.version_info.major == 2 else source
             nodes = ast.parse(source, path)
             self._code = compile(nodes, filename=path, mode='exec', dont_inherit=True)
         except SyntaxError as e:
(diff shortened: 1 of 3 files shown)


bryevdv:
OK now that it is local to notebook handler I will plan to merge for 1.0.4. I assume you have tested manually on both 2/3?

jlstevens:
> I assume you have tested manually on both 2/3?

Yes, using a very simple notebook.

There is one issue I am not sure how to fix: if your notebook contains unicode characters, e.g:

```
θ=3
```

then you'll get:

``ERROR: 'ascii' codec can't encode character u'\u03b8' in position 78: ordinal not in range(128)``

This isn't surprising given the currently suggested fix and I'm trying to figure out how to get `ast.parse` to work when unicode is present.

bryevdv:
OK well I am happy to merge whenever, since it is so localized, but I will wait for your `ready`

bryevdv:
The uncommented test are failing on python 2.7

jlstevens:
Actually, I've just read something suggesting that `θ=3` would always be invalid for python 2. In which case, I'll put the unicode in a string and test that instead.

bryevdv:
Still failing Python 2.7 I would recommend running all unit tests locally before pushing, there is a huge backlog/slowdown on Travis CI and limited workers available.

jlstevens:
Had to update the earlier unit test: all the tests on Travis are passing now. 

In the end, this PR simply removes the line `# coding: utf-8` on Python 2 as it cannot be processed when in a unicode string. In addition, line magics in the older version of IPython are now handled by removing `get_ipython().magic` - this older version is the last version that will support Python 2.

jlstevens:
@bryevdv Ready to review/merge.

PR #8556, opened 2019-01-08
Correct three minor typos

EJEP:
Correct three minor typos in` tools.py` comments.


bryevdv:
Thanks for the catch and the fix @EJEP

PR #8558, opened 2019-01-09
Safer alternative fix for unicode notebook issue in Python 2

jlstevens:
This PR is an alternative fix to the one implemented in #8555. Based on a suggestion from @philippjfr, it seems like it might be best to always encode to utf8 as nbconvert is templated to always use that encoding.


bryevdv:
Python 2 tests failing. 

```
>       assert result['handler']._runner.source == expected_source
E       AssertionError: assert '\n# coding: utf-8\n' == '\n\n'
E         Strings contain only whitespace, escaping them using repr()
E         - '\n# coding: utf-8\n'
E         + '\n\n'
```

PR #8565, opened 2019-01-13
Ensure compilation of custom model works with bokehjs compiled on win

xavArtley:
If windows compile bokehjs => the path separator in bokeh.json is `\\` instead of `/`
Then when a custom module extension with lines like :
`import * as p from "core/properties"` 
is compiled it failed because available module in bokeh.json is `core\\properties`

### DIFF
--- a/bokehjs/make/linker.ts
+++ b/bokehjs/make/linker.ts
@@ -328,7 +328,7 @@ export class Module {
     for (const base of this.linker.bases) {
       const path = relative(base, this.file)
       if (!path.startsWith("..")) {
-        return path.replace(/\.js$/, "").replace(new RegExp(/\\/, 'g'), "/")
+        return path.replace(/\.js$/, "").replace(/\\/g, "/")
       }
     }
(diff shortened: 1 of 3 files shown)


bokehjs/make/linker.ts
```diff
@@ -328,7 +328,7 @@ export class Module {
     for (const base of this.linker.bases) {
       const path = relative(base, this.file)
       if (!path.startsWith("..")) {
-        return path.replace(/\.js$/, "")
+        return path.replace(/\.js$/, "").replace(new RegExp(/\\/, 'g'), "/")
```
mattpap: `new RegExp(/\\/, 'g')` could be replaced with simpler `/\\/g`.

bryevdv:
@xavArtley @mattpap is there a simple unit test we could add that would exercise this? Then things would be maintained for windows on Appveyor tests

bryevdv:
Thanks @xavArtley

PR #8569, opened 2019-01-15
Prototype - Feedback requested

birdsarah:
This is not for merging it's a quick knock up of a prototype for issue #6536.

Some questions:

1) API

  * What do we want the properties to be called. @bryevdv you had previously spelled out positive_infinity and negative_infinity as seperate properties, is that still something you want to do. ~~Personally I am leaning towards just `not is_finite` so "infinity." But if there's a compelling case of when you'd want to set both for a range I'm all ears. I have not yet looked into what other libraries do in this regard.~~ I have thought about this more, see later comment.
  * What do we want the enums to be. The more I think about it the less I like "screen_min" and "screen_max" they're so "insider" / implementation specific. I think that should be "scale_min", "scale_max" 

2) Default

  * I actually think we should set the default to be "scale_min" (or whatever we call it) so that log plots just start working for new plot writers. I wouldn't regard it as an API change as this property didn't exist before and most people who have a currently functioning log plot would probably have set bottom (as I have done everywhere) and so this change seems unlikely to affect them.....it could affect some people who've used the disappearing nature of log(0) as a feature.....feels less common than people being confused by their bar charts not working

3) Other changes? 

  * Are there other changes that need to be made along with this PR in terms of syntactic magic etc.
  * Should this just be on LogScale or on all scales for consistency?

This is not intended to be complete or review-worthy, just me getting my head around the basic mechanics. I will be starting a new branch with tests etc.

The functional test cases I plan to make sure I've covered are:

  *  y and x axes in both directions
  *  circles, lines, vbars, patches (see old issue)
  *  double axes
  *  colorbar
  * server driven property change

screenshot from 2019-01-15 10-20-22

### DIFF
--- a/bokehjs/src/lib/models/scales/scale.ts
+++ b/bokehjs/src/lib/models/scales/scale.ts
@@ -40,6 +40,14 @@ export abstract class Scale extends Transform {
 
   abstract v_invert(sxs: Arrayable<number>): Arrayable<number>
 
+  screen_min(): number {
+    return this.target_range.start
+  }
+
+  screen_max(): number {
+    return 0
+  }
+
   r_compute(x0: number, x1: number): [number, number] {
     if (this.target_range.is_reversed)
       return [this.compute(x1), this.compute(x0)]
--- a/bokeh/core/enums.py
+++ b/bokeh/core/enums.py
@@ -353,6 +353,7 @@ VerticalAlign = enumeration("top", "middle", "bottom")
 #: Specify a vertical location in plot layouts
 VerticalLocation = enumeration("above", "below")
 
+InfinityPosition = enumeration("screen_min", "screen_max")
 #-----------------------------------------------------------------------------
 # Private API
 #-----------------------------------------------------------------------------
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -60,7 +60,7 @@ log = logging.getLogger(__name__)
 # External imports
 
 # Bokeh imports
-from ..core.enums import Anchor, Direction, StepMode
+from ..core.enums import Anchor, Direction, InfinityPosition, StepMode
 from ..core.has_props import abstract
 from ..core.properties import (AngleSpec, Bool, DistanceSpec, Enum, Float, String,
                                Include, Instance, Int, NumberSpec, Override, StringSpec)
@@ -1131,6 +1131,8 @@ class VBar(Glyph):
     x = NumberSpec(help="""
     The x-coordinates of the centers of the vertical bars.
     """)
+
+    inf = Enum(InfinityPosition, default="screen_min")
 
     width = NumberSpec(help="""
     The widths of the vertical bars.
--- a/bokeh/models/scales.py
+++ b/bokeh/models/scales.py
@@ -13,6 +13,9 @@
 #-----------------------------------------------------------------------------
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from ..core.enums import InfinityPosition
+from ..core.properties import Bool, Dict, Either, Enum, Float, Instance, Seq, String, AnyRef
+
 import logging
 log = logging.getLogger(__name__)
 
@@ -77,7 +80,9 @@ class LogScale(Scale):
     '''
 
     '''
-    pass
+    inf = Enum(InfinityPosition, default=None, help="""
+    screen_min, max
+    """)
 
 class CategoricalScale(LinearScale):
     '''
--- a/bokeh/models/scales.py
+++ b/bokeh/models/scales.py
@@ -80,9 +80,6 @@ class LogScale(Scale):
     '''
 
     '''
-    inf = Enum(InfinityPosition, default=None, help="""
-    screen_min, max
-    """)
 
 class CategoricalScale(LinearScale):
     '''
--- a/bokeh/models/scales.py
+++ b/bokeh/models/scales.py
@@ -13,9 +13,6 @@
 #-----------------------------------------------------------------------------
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from ..core.enums import InfinityPosition
-from ..core.properties import Bool, Dict, Either, Enum, Float, Instance, Seq, String, AnyRef
-
 import logging
 log = logging.getLogger(__name__)
 
@@ -80,6 +77,7 @@ class LogScale(Scale):
     '''
 
     '''
+    pass
 
 class CategoricalScale(LinearScale):
     '''
--- a/bokehjs/src/lib/core/enums.ts
+++ b/bokehjs/src/lib/core/enums.ts
@@ -106,3 +106,6 @@ export const HTTPMethod: HTTPMethod[] = ["POST", "GET"]
 
 export type Logo = "normal" | "grey"
 export const Logo: Logo[] = ["normal", "grey"]
+
+export type InfinityPosition = "screen_min" | "screen_max"
+export const InfinityPosition: InfinityPosition[] = ["screen_min", "screen_max"]
--- a/bokehjs/src/lib/core/properties.ts
+++ b/bokehjs/src/lib/core/properties.ts
@@ -256,6 +256,8 @@ export class PaddingUnits extends enum_prop("PaddingUnits", enums.PaddingUnits) 
 
 export class StartEnd extends enum_prop("StartEnd", enums.StartEnd) {}
 
+export class InfinityPosition extends enum_prop("InfinityPosition", enums.InfinityPosition) {}
+
 //
 // Units Properties
 //
--- a/bokehjs/src/lib/models/glyphs/vbar.ts
+++ b/bokehjs/src/lib/models/glyphs/vbar.ts
@@ -1,6 +1,7 @@
 import {Box, BoxView, BoxData} from "./box"
 import {Scale} from "../scales/scale"
 import {Arrayable} from "core/types"
+import {InfinityPosition} from "core/enums"
 import {NumberSpec, DistanceSpec} from "core/vectorization"
 import * as p from "core/properties"
 import {SpatialIndex} from "core/util/spatial"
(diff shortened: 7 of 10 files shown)


birdsarah:
For what it's worth, I have some strong preferences on some of the above questions, based on my own daily run-ins with this issue.

For the past ~6 months I've been using bokeh daily while working as a researcher. I build multiple histograms a day, most of which end up having a log version, which adds up to hundreds and I bump into this issue all the time. I use the plotting interface exclusively (how times have changed :D). There are three common scenarios:

1) I know I want to build a log plot, in which case it's just annoying to have to select a "bottom" value that might be appropriate before I've run the cell, and it's just extra typing
2) I build a linear plot, and then realize I need a log plot. I now need to make two changes - one to add the `axis_type`, the second to add a bottom value
3) I build a log plot, and then realize I want a linear plot instead or alongside. I now need to make two changes. While technically removing bottom isn't really necessary, I find it irksome having it there as it's not technically accurate, and it's cruft.

This leads me to the following suggestions:

1) The default should be set to `scale_min` so the common case just works
2) The infinity property should be available on all scales so that switching between plot types does not result in a property error / extra typing even though the use-case for infinity on a LinearScale is much less common

birdsarah:
Regarding `positive_infinity` and `negative_infinity` my thinking has evolved to this. Let's have both of those, but lets also have `infinite` or `inf` as syntactic sugar to set both.

bryevdv:
> The default should be set to scale_min so the common case just works

I'm maybe ok with this but not sure Currently, if there are `inf` values, that will just result in nothing being rendered for that point. Are there reasonable cases where people would explicitly rely on and expect this behavior? Imagine a scatter with some infs, now all those markers just show up on the edge of the plot. Is that desired?

A less invasive default change would be to have the scale report what the `inf` policy value is, but still actually return the `inf` values to renderers in mapped data. Then, when the renderer encounters `inf` values it can choose what to do individually. E.g. the line renderer could draw those vertical segments, and the `bar` renderer can ask the scale for the policy value and use that. Maybe that would be more flexible and decoupled? Under this scenario I'd be happy to have bar renderers "just do the right thing" because nothing else would be impacted. 

Also under this design change, 2) above also make more sense. 

In my head the suggested change is a small delta  from what you have now, so let me know if my explanation does not make sense.

HuntJSparra:
1) I talked to a few people that I know who are newer to data visualization, and they unanimously told me that "scale_min" and "scale_max" made more sense to them (and thus I presume new users). I don't know what more experienced users might prefer.

2) As far as I know, matplotlib and seaborn essentially have "scale_min" as their default. I have seen very few examples that use "scale_max" instead of "scale_min".

A semi-related comment:
#8343 was (indirectly) addressing the handling of log(0). @bryevdv proposes there for axes to have a policy for handling non-real values such as -inf and +inf using names such as "screen_max" (just like is done here). My concern is that while "displaying 0 on a log axis" and "handling the log of 0" are very different problems, the two sound confusingly similar (to an inexperienced person such as myself, at least). If, at a later date, "handling the log of 0" was added, whoever is working on it should be careful to make its documentation distinct from this.

bryevdv:
Oh I forgot to comment, re: naming. `"scale_min"` seems fine by me. Tho if we adopt the proposal change I suggested above, I am not sure we need any properties at all aymore? The scales always compute provide the scale_min, etc values as things that renderers can ask for e.g. ` scale.negative_inf_value()` so we don't need a user tweakable property (at least not on Scale, individual renderers might want to expose options for how to handle `inf` but not Bar, which we jsut make "do the right thing")

hsparra:
How would this work from a user's perspective with a histogram. Ex.

bryevdv:
> How would this work from a user's perspective with a histogram.

I'm not sure how this is different from bars that have `bottom=0` (in fact, that's exactly what the histogram example has) But to be clear, the goal is just for that code above to work, as is.

hsparra:
bryevdv, it is the same situation, just a different starting point. If the goal is for that code above to just work as-is then I support that fully. I guess upon further reflection my question is around what spelling the user will use, if any different. With the talk of infinity I wanted to make sure I understood the implications. I can open another issue to discuss how histograms that need to use a log axis should be "done" by a user. Thanks!

birdsarah:
@bryevdv - I think I understand your suggestion. 

The idea is to put the control back in the hands of the renderer.

The main con I can see is that we're going to get in a position of having lots of implementations - like for hover. But at the same time, your suggestion is a more gradual case and I think covers the main pain points.

I'm going to try making a prototype update and then see what you think.

birdsarah:
As I dug into this more I'm slightly confused, but working on a prototype.

As I think about it more, and your way round, I would argue that the `inf` property is a property of the renderer / glyph (I will need guidance on the specifics of what's in the renderer what's in the glyph model / view). 

It's a property of the glyph / renderer because it's a choice of how the *renderer* decides to show itself on a log scale, rather than being an opinion of the scale. The scale should just report back "infinite" if that's what it generated. I think we will need a new enum / const for this because NaN is already used for blank data so it's good to keep it specific.

Let me know if this sounds right.

bryevdv:
>  I think we will need a new enum / const for this because NaN is already used for blank data so it's good to keep it specific.

Yes it could be, I mentioned that above but perhaps not clearly. But it's possibly completely different for every renderer (might be multiple ways to behave, some may not apply at all to other glyphs) or, there might not need to be one at all. I was suggesting in this case that bars could just always do the right thing, i,e. that there is no need for a renderer specific config for bar in particular, because the only alternative is just "it's broken completely when there are infs"

birdsarah:
As I dug into this more I'm slightly confused, but working on a prototype.

As I think about it more, and your way round, I would argue that the `inf` property is a property of the renderer / glyph (I will need guidance on the specifics of what's in the renderer what's in the glyph model / view). 

It's a property of the glyph / renderer because it's a choice of how the *renderer* decides to show itself on a log scale, rather than being an opinion of the scale. The scale should just report back "infinite" if that's what it generated. I think we will need a new enum / const for this because NaN is already used for blank data so it's good to keep it specific.

Let me know if this sounds right.

birdsarah:
@bryevdv, can you comment on this:

> The main con I can see is that we're going to get in a position of having lots of implementations - like for hover. But at the same time, your suggestion is a more gradual case and I think covers the main pain points.

birdsarah:
Logic on separate implementations makes sense. Thanks for clarifying. I definitely agree this is a renderer choice.

I've pushed the latest prototype. Below is spelling and a screenshot of it working. 

Obviously the code isn't pretty or dry, but can you take a look at where properties are etc to see if this is shaping up as you expected.

```python
p_test_vbar_x_a = figure(y_axis_type='log', title="screen_min")
p_test_vbar_x_a.vbar(
    x=[1, 2, 3], 
    top=[1, 10, 100], 
    width=0.9,
    inf="screen_min",
)

p_test_vbar_x_b = figure(y_axis_type='log', title="screen_max")
p_test_vbar_x_b.vbar(
    x=[1, 2, 3], 
    top=[1, 10, 100], 
    width=0.9,
    inf="screen_max",
)

p_test_vbar_x_c = figure(y_axis_type='log', title="none")
p_test_vbar_x_c.vbar(
    x=[1, 2, 3], 
    top=[1, 10, 100], 
    width=0.9,
    inf=None,
)

p_test_circle = figure(y_axis_type='log', title="infinity circle")
p_test_circle.circle(x=[1, 2, 3, 4], y=[0, 1, 0, 2], size=10)
```

screenshot from 2019-01-18 11-03-42

The most surprising thing about this so far is that I learned that we declare x and y for every glyph, and so by that token will need to add inf for every glyph that wants to support this. Having said that, this means that this PR can be relatively small and only support vbar and we can add others with the same template later.

bryevdv:
@birdsarah I think that looks fine, my only question is whether "scale_max" and "none" are really use cases that would be useful/used? If not, I think we could dispense with a user-configurable option (dispense with just specifically for bar) and use "scale_min" all the time.

bokehjs/src/lib/models/scales/scale.ts
```diff
@@ -40,6 +40,14 @@ export abstract class Scale extends Transform {
 
   abstract v_invert(sxs: Arrayable<number>): Arrayable<number>
 
+  screen_min(): number {
+    return this.target_range.start
+  }
+
+  screen_max(): number {
+    return 0
```
bryevdv: Thinking out loud, this may need to be a bit more sophisticated. E.g. if the range is flipped with `start > end` then I think you'd want `target_range.end`. Maybe a better long plain description for "screen_min" is "the screen (target) coordinate corresponding to the min data (source) coordinate in the range"

birdsarah: totally agree. lots of edge cases to think about.

bryevdv: What do you think about `screen_for_data_min`, `screen_for_data_max` as more descriptive names?

birdsarah:
> my only question is whether "scale_max" and "none" are really use cases that would be useful/used? If not, I think we could dispense with a user-configurable option (dispense with just specifically for bar) and use "scale_min" all the time.

Stream of consciousness:

I think it would be good to implement a "model implementation" and API that are robust for inclusion on other glyphs as that comes up - e.g. the patches issue that someone had previously.

I think we want `positive_infinity` and `negative_infitinty` as the properties with the possible values `scale_max`, `scale_min`, and None (current behavior). 

I think `inf` is just a plotting API short-hand for setting both.

...On the flip side....

There is an obvious, and intuitive, mapping of `positive_infinity -> scale_max` and `negative_infinity -> scale_min`.

Maybe we want a property like `display_infinity` / `render_infinity` or something like that, that is False by default for most things - which is the current behavior. If True, different glyphs will render appropriately for the glyph (as previously discussed). What would that property be called?

bryevdv:
Offhand I am just not sure theres one consistent API across all glyphs, except possibly to offer the same options 

* "do nothing / don't display" 
* "so something (glyph dependent) / try to display"

I'm not opposed to that, but it does give up some possibility of specificity per glyph regarding policy.

bryevdv:
It seems like this has stalled out a bit, and there are now conflicts as well, so I am going to go ahead and close (but leave the branch)

birdsarah:
I already have an updated branch without conflicts going. Sorry, should have close this.

bryevdv:
Can this branch be deleted then?

birdsarah:
can i delete it when i finish the next pr

birdsarah:
(my laptop died so i've been slowed down)

PR #8574, opened 2019-01-17
Bryanv/20 deprecations

### LINKED ISSUE #8560: Deprecations for 2.0?
Given that we have a little under ~1 year until making a 2.0 release that drops Python 2 support, it seems like we should start thinking of what else we might want to drop, and start deprecating those things soon. Off the top of my head:

* CoffeeScript support

  It's is a complete dead end at this point and the machinery to support extensions and CustomJS code in CoffeeScript could be stripped out.

* `CustomJS.from_py_func`, etc. 

  There's actually a somewhat large of amount of almost-duplicated code to support this. I am not sure of the status of `pscript` going forward and don't want to be on the hook maintaining support for it forever. I don't think this feature is widely used (we make a point not to highlight it at all). It was a very useful stopgap yeas ago but I think the need/justification for it is far less now.  

For both of the above I think the deprecation warning message could emit the actual code users would need to replace with. 

Thoughts? What else could we consider?

cc @bokeh/dev

bryevdv:
Needs some documentation support, also to make sure there are not places that have been overlooked, or other things we can slate for 2.0 deprecation. 

- [x] issues: fixes #8560
- [x] tests added / passed


bryevdv:
I will go ahead and merge this when green. There are still other things to consider but we can make separate issues.

PR #8580, opened 2019-01-22
Add a color picker and spinbox to widgets

xavArtley:
- [x] Add 2 new widgets:
- A SpinBox widget for numerical inputs
- A Color picker to select colors

- [x] 2 examples of use case are added

- [ ] Need to add tests


bokehjs/src/lib/models/widgets/color_picker.ts
```diff
@@ -0,0 +1,83 @@
...
+    const labelEl = label({for: this.model.id}, this.model.title)
+    this.el.appendChild(labelEl)
+
+    this.inputEl = input({
+      class: "bk-widget-form-input",
+      id: this.model.id,
+      name: this.model.name,
+      value: this.model.color,
+      disabled: this.model.disabled,
+      type: "color",
```
mattpap: This isn't supported in IE and possibly Safari (sources vary).

jbednar: It would be nice if there was a simple fallback that could be supported by those browsers.

xavArtley: Ok but is it ok for the spinner?

mattpap: Yeah, I think that `type="number"` is supported well enough. On a side note, going through other types, it's interesting to me that `type="range"` is so well supported.

bryevdv:
@xavArtley do you have thoughts about the browsers that are unsupported? I mistakenly assumed when you mentioned this PR that you were wrapping/including some lightweight cross-platform color picker. I think that would be the only workable approach for actually adding this to the core libarary.

xavArtley:
I tested colorpicker on different browsers this is the results:

- internet explorer (a simple textinput with hexadecimal as text)
color_picker_ie
- opera seems ok
color_picker_opera
- chrome seems ok
color_picker_chrome
- edge a different color picker and the result is displayed as hexadecimal string
color_picker_edge
- firefox seems ok except the color in the input miss padding above and bellow
color_picker_firefox
I can't test on safari

I will remove the color picker from the PR

mattpap:
If it works that much (at least doesn't fail), then we may proceed this way. Personally I can't care less about IE and I would be glad to drop support for it altogether. @jbednar, do you have any clients explicitly asking for IE support? In any case, I proposed dropping IE in https://github.com/bokeh/bokeh/issues/8560#issuecomment-456110652, so lets discuss there.

jbednar:
Oh yes, I deal with lots of clients stuck on IE.  But they are accustomed to hardship. :-) They have low expectations, and are happy when things work at all, even if they aren't fancy.

xavArtley:
Ok so I keep it :)
However I don't understand the error on travis

mattpap:
> However I don't understand the error on travis

```
ColorPicker: python is missing some properties found in bokehjs
    ColorPicker.callback: bokehjs defaults to null but python has no such property
SpinBox: python is missing some properties found in bokehjs
    SpinBox.callback: bokehjs defaults to null but python has no such property
```

The problem is that in bokehjs `InputWidget` defines `callback` property, but bokeh's InputWidget doesn't. Instead, each model in Python defines callback individually. For now just follow how other `InputWidget`s are implemented in Python.

bryevdv:
@xavArtley Let's also add a comment in the helpsstring about the reduced functionality on IE

There are also some other unrelated build issues currently, on both Appveyor and Travis. I hope to have them all resolved in the next few hours.

bryevdv:
@xavArtley also the builds should be working now if you want to rebase/merge with any future commits.

bryevdv:
@xavArtley I just noticed faa3242 I think these should be omitted. The `callback` property dates back to before `on_change` an `js_on_change` existed. But those are the methods people should use in new code. I don't think `callaback` should be added to anything new at this point.

bokeh/models/widgets/inputs.py
```diff
@@ -208,6 +215,54 @@ class DatePicker(InputWidget):
...
+    """)
+
+    def __init__(self, **kwargs):
+        if 'color' in kwargs:
+            if hasattr(named, kwargs['color']):
+                kwargs['color'] = getattr(named, kwargs['color']).to_hex()
+        super(ColorPicker, self).__init__(**kwargs)
+
+
+class SpinBox(InputWidget):
```
mattpap: Lets rename this to just `Spinner`.

bryevdv: Agree

mattpap:
> But those are the methods people should use in new code. I don't think callaback should be added to anything new at this point.

Here it just fulfills an existing interface, so nothing new is being added. But in the long term the situation with callbacks will have to be resolved.

xavArtley:
Concerning the color picker I made a custom extension 
https://xavartley.github.io/bokeh/test_color_pick.html
To integrate it it will require more work and for the moment it's limited to select a color with cursors (no box to specify hsl or rgb)
I was wondering if it's worth to integrate it in bokehJS (will be longer and need more review from your part) or if I keep going with the actual ColorPicker and add this custom one in examples

bryevdv:
I definitely love the simplicity and uniformity of it, though for some use cases being able to specify e.g. the hex RGB values may be important. So that would be a reason to keep the native colorpicker option. 

I guess what I might suggest is actually to merge the current PR. Then later we could make the color picker configurable, e.g. between "native" (which has warnings about not working on IE) and "basic" (which is your nice uniform version). We would make the default be "native" so that there are not breaking changes along the way. How does that sound?

bokeh/models/widgets/inputs.py
```diff
@@ -208,6 +217,73 @@ class DatePicker(InputWidget):
...
+            elif isinstance(potential_color, tuple) and len(potential_color) == 3:  # tuple
+                color = RGB(*potential_color).to_hex()
+
+            if color is None:
+                raise ValueError("expected and element of either String, HSL, RGB or Tuple(r,g,b). " + \
+                                 "String must be a valid hexadecimal color or a named color from this list: " + \
+                                 f"{list(NamedColor)}. got {potential_color}")
+            else:
+                kwargs['color'] = color
+        super(ColorPicker, self).__init__(**kwargs)
```
bryevdv: @xavArtley what's preventing from just letting the `Color` property handle all this validation? I would expect not needing to have an `__init__` at all.

xavArtley: Actually in typescript I don't know how to handle Color other than hexadecimal string

bryevdv: I see. There's probably a way we can handle things at the property level, such that serialization always happens to RGB hex values. Let me think on it.

mattpap:
This will have to be updated on top of `master` when PR #8085 is merged (hopefully tomorrow). @xavArtley, I will help with the necessary changes.

bryevdv:
Noting this will need a merge/rebase on master to accommodate new conda/conda-build versions now used on TravisCI

mattpap:
I canceled the build. When I'm done with my work, I will restart it.

bryevdv:
FYI @xavArtley  this PR will have to be rebased on master, after #8662

bryevdv:
Ok there are some linter issues, and also two examples failures because the new examples need baselines added:
```
[FAIL] tests/baselines/sphinx/source/docs/user_guide/examples/interaction_spinner baseline doesn't exist
```
@mattpap can you instruct @xavArtley either how to ad new baselines, or how to skip these two for now

xavArtley:
I need to continue to workon it but i have not much time for the moment

bryevdv:
@xavArtley OK, I think you can just list the two new examples to be skipped in this file:

https://github.com/bokeh/bokeh/blob/master/examples.yaml

Otherwise the other option would be to close this PR and re-open it on the main repo, then @mattpap or I could directly push changes to handle the last few tasks

mattpap:
There's a third rarely explored option: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/.

bryevdv:
> There's a third rarely explored option
 
I am aware of that, however having tried a couple of times and never been able to actually make it work, I don't suggest it anymore.

mattpap:
I see, I haven't tried it myself.

mattpap:
I finished this PR in #8678, so that we can get this merged for 1.1.

bryevdv:
Thanks @xavArtley !

xavArtley:
Thx for finishing my PR

PR #8584, opened 2019-01-24
add option to set websocket_max_message_size from notebook

### LINKED ISSUE #8583: Allow to pass websocket_max_message_size when embedding bokeh server in a notebook
When bokeh server is embedded in a notebook (like here https://github.com/bokeh/bokeh/blob/master/examples/howto/server_embed/notebook_embed.ipynb) there are connection errors when plots are large, like in #7374 (and smaller plots work just fine).  I see from the docs I can use `--websocket-max-message-size` argument or `websocket_max_message_size` kwarg with a large enough value for a standalone bokeh server.

However, it is not currently possible to pass that argument to the `Server` class from notebook when using `bokeh.io.show(app, notebook_url)`:
https://github.com/bokeh/bokeh/blob/d260c429451a04b1fc00f593ef98f10ce1392e57/bokeh/io/notebook.py#L475

The solution would be to pass something like `server_kwargs` to `show()` so that they go through to the `Server.__init__`.

An example notebook to test the functionality can be constructed from the following code:
```python
import numpy as np

from bokeh.palettes import Inferno

output_notebook()

def modify_doc(doc):
    size = 3000
    a = np.broadcast_to((np.arange(size, dtype=np.float64))[np.newaxis, :], (size, size))

    plot = figure(x_range=(0, 100), y_range=(0, 100))
    plot.image(
        image=[a], x=[0], y=[0], dw=[100], dh=[100],
        palette=Inferno[256]
    )
    
    doc.add_root(plot)

show(modify_doc)
```

nasedil:
- [*] issues: fixes #8583 
- [ ] tests added / passed

### DIFF
--- a/bokeh/io/notebook.py
+++ b/bokeh/io/notebook.py
@@ -429,7 +429,7 @@ def publish_display_data(*args, **kw):
     from IPython.display import publish_display_data
     return publish_display_data(*args, **kw)
 
-def show_app(app, state, notebook_url, port=0):
+def show_app(app, state, notebook_url, port=0, server_kwargs=None):
     ''' Embed a Bokeh serer application in a Jupyter Notebook output cell.
 
     Args:
@@ -456,6 +456,11 @@ def show_app(app, state, notebook_url, port=0):
             By default the port is 0, which results in the server listening
             on a random dynamic port.
 
+        server_kwargs (dict, optional) :
+            Extra arguments passed to :func:`bokeh.server.Server.__init__`
+
+            If None, no extra arguments are passed (default: None)
+
     Returns:
         None
 
@@ -472,7 +477,9 @@ def show_app(app, state, notebook_url, port=0):
     else:
         origin = _origin_url(notebook_url)
 
-    server = Server({"/": app}, io_loop=loop, port=port,  allow_websocket_origin=[origin])
+    if server_kwargs is None:
+        server_kwargs = {}
+    server = Server({"/": app}, io_loop=loop, port=port,  allow_websocket_origin=[origin], **server_kwargs)
 
     server_id = uuid4().hex
     curstate().uuid_to_server[server_id] = server
--- a/bokeh/io/showing.py
+++ b/bokeh/io/showing.py
@@ -43,7 +43,7 @@ __all__ = (
 # General API
 #-----------------------------------------------------------------------------
 
-def show(obj, browser=None, new="tab", notebook_handle=False, notebook_url="localhost:8888"):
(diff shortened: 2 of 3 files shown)


bokeh/io/notebook.py
```diff
@@ -456,6 +456,11 @@ def show_app(app, state, notebook_url, port=0):
             By default the port is 0, which results in the server listening
             on a random dynamic port.
 
+        server_kwargs (dict, optional) :
+            Extra arguments passed to :func:`bokeh.server.Server.__init__`
+
+            If None, no extra arguments are passed (default: None)
+
```
bryevdv: Now that we are past 1.0 we should try to start noting when new parameters and functions are added, I think. Can you make a note here "Added in 1.1" (that is the next expected release version)

bryevdv: It's probably also worth discussing whether it's better to pass an explicit dict, or better to accept `**kwargs`. the slight drawback to accepting an explicit dict is that `port` is already a named parameter so it leaves open a possibility of supplying `port` twice, as a parameter and as a value in the dict, which would be inconsistent.

bryevdv:
@nasedil sorry for the delayed response, I've been out this week. Thanks for the PR. I had one small comment to discuss but we should be able to merge this soon.

bokeh/io/notebook.py
nasedil: @bryevdv I've added the note about version.

Regarding the `**kwargs`, I thought about that, and my reasoning for using `server_kwargs` was that `**kwargs` may be passed to different functions called from from `show()` and `show_app()`, and here I only pass it to the `Server`.  But I'm not sure which way is better.

bryevdv: @nasedil but `show` and `show_app` don't currently accept `**kw`? so then it's not the case that extra kwargs are currently being passed. I think it would be ok to add `**kw` and state in the docs that any additional keyword args are passed to initialize the server in `output_notebook` mode.

nasedil: @bryevdv I've changed `server_kwargs` to `**kw`.

bryevdv:
@nasedil awesome that seems great. Currently the docs build is failing because Sphinx can be very finicky. HOw about instead of this two line note:

```
Extra keyword arguments are passed to :func:`~bokeh.server.Server.__init__`
            ``**kw`` is added in 1.1
```

we go with a simpler one-line version:

```
Any additional keyword arguments are passed to :class:`~bokeh.server.Server` (added in version 1.1)
```

nasedil:
Oh yes, sorry about the docstrings.  I've fixed the docs with the line you provided.

bryevdv:
Thanks @nasedil !

PR #8586, opened 2019-01-26
Possible typo in documentation for bokeh.client.session. 

dankovacek:
In [documentation for bokeh.client.session](https://bokeh.pydata.org/en/latest/docs/reference/client/session.html) under ClientSession object, mysession and session objects should maybe be the same?

```
with pull_session(url=app_url) as session:
    # customize session here
    script = server_session(session_id=mysession.id, url=app_url)
    return render_template("embed.html", script=script, template="Flask")
```

### DIFF
--- a/bokeh/client/session.py
+++ b/bokeh/client/session.py
@@ -230,13 +230,13 @@ class ClientSession(object):
 
     .. code-block:: python
 
-        with pull_session(url=app_url) as session:
+        with pull_session(url=app_url) as mysession:
             # customize session here
             script = server_session(session_id=mysession.id, url=app_url)
             return render_template("embed.html", script=script, template="Flask")
 
     If you do not use ``ClientSession`` in this way, it is up to you to ensure
-    that ``session.close()`` is called.
+    that ``mysession.close()`` is called.
 
     '''
 


bryevdv:
@dankovacek thanks for making a PR. A reminder that test failures are due to an unrelated Pandas issue. I think the edit did not make it took your branch tho since there are no changes showing up here? You should only need to commit the change on your branch, and git push for it to show up here.

dankovacek:
Thanks for the additional help, and sorry for making more work for you than this is worth!  In order to make the edit, I thought I had to complete the build process for the html pages (`>make clean html`), find the corresponding html file, make the change, and commit.  I had to revert pandas to 0.23.4 to get the build to work, then there was another error regarding a google maps API key which I got around without changing any files.  I found two folder locations `/sphinx/_build/html/docs/reference/client/session.html` and `/sphinx/_build/html/modules/bokeh/client/session.html` and make changes, but it seems these changes aren't tracked, git status says 'nothing to commit'.   I understand these html files were generated from a `make` command, so it makes sense that changing the html files themselves isn't what is required.  There were no changes to my branch earlier because I committed with --allow-empty, because I didn't understand there was actually an edit to make somewhere.  Where should I be looking in the file tree?  Thanks.

bryevdv:
@dankovacek Ah, yes that's a reasonable thing to think, we could probably make the docs chapter of the dev guide more clear. Those docs are generated automatically from docstrings in the source code. This bit in particular is located here:

https://github.com/bokeh/bokeh/blob/master/bokeh/client/session.py#L235

So if you edit that line you should be able to commit and push the change

dankovacek:
@bryevdv thanks, I found it.  Do I need to undo the original commit that didn't have any changes, or is it OK to leave as is?

bryevdv:
it should be fine to leave the other commit. I'll get this merged tomorrow

bryevdv:
Thanks for your contribution @dankovacek !

Issue #8589, opened 2019-01-29
[DOCS] Bad number in webpage patch method documentation


BrettMontague:
https://bokeh.pydata.org/en/latest/docs/reference/models/sources.html

In the little example blurb for the patch method, it reads:

```
The following example shows how to patch entire column elements. In this case,


source = ColumnDataSource(data=dict(foo=[10, 20, 30], bar=[100, 200, 300]))

patches = {
    'foo' : [ (slice(2), [11, 12]) ],
    'bar' : [ (0, 101), (2, 301) ],
}

source.patch(patches)
After this operation, the value of the source.data will be:

dict(foo=[11, 22, 30], bar=[101, 200, 301])

```

the last line should read:

```
dict(foo=[11, 12, 30], bar=[101, 200, 301])
```

PR #8590, opened 2019-01-29
#8589 Fixes docstring typo in patch method of sources.py

BrettMontague:
Fixed a docstring typo as raised in #8589


bryevdv:
Thanks for the catch and the fix @BrettMontague !

PR #8593, opened 2019-01-30
Refactor tile_providers to handle multiple documents more intuitively…

### LINKED ISSUE #8566: Tile provider causes "Model must be owned by only a single document" error
**System**:
Bokeh 1.0.4
python 3.6
Ubuntu 18.04

**Expected:** When using a `tile_provider` such as `CARTODBPOSITRON`, the object that is imported globally can be attached to more than one document.
**Actual:** A RuntimeError is raised, because the tile provider already is part of a document.

```python
from bokeh.tile_providers import CARTODBPOSITRON
from bokeh.io import show


def test():
    fig = figure(plot_width=1000, plot_height=1000,
                 active_scroll="wheel_zoom",
                 x_range=(49., 49.5), y_range=(30., 30.5),
                 x_axis_type="mercator", y_axis_type="mercator",
                 )

    fig.add_tile(CARTODBPOSITRON)
    show(fig)

test()
test()
```

Traceback:

```
  File "/home/veith/.PyCharmCE2018.3/config/scratches/scratch.py", line 18, in <module>
    test()
  File "/home/veith/.PyCharmCE2018.3/config/scratches/scratch.py", line 15, in test
    show(fig)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/io/showing.py", line 142, in show
    return _show_with_state(obj, state, browser, new, notebook_handle=notebook_handle)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/io/showing.py", line 180, in _show_with_state
    _show_file_with_state(obj, state, new, controller)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/io/showing.py", line 163, in _show_file_with_state
    filename = save(obj, state=state)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/io/saving.py", line 86, in save
    _save_helper(obj, filename, resources, title, template)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/io/saving.py", line 148, in _save_helper
    html = file_html(obj, resources, title=title, template=template)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/embed/standalone.py", line 307, in file_html
    with OutputDocumentFor(models, apply_theme=theme, always_new=_always_new) as doc:
  File "/home/veith/Tools/pyenv/versions/3.6.5/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/embed/util.py", line 137, in OutputDocumentFor
    doc.add_root(model)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/document/document.py", line 304, in add_root
    self._pop_all_models_freeze()
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/document/document.py", line 1018, in _pop_all_models_freeze
    self._recompute_all_models()
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/document/document.py", line 1041, in _recompute_all_models
    a._attach_document(self)
  File "/home/veith/vscen/lib/python3.6/site-packages/bokeh/model.py", line 609, in _attach_document
    raise RuntimeError("Models must be owned by only a single document, %r is already in a doc" % (self))
RuntimeError: Models must be owned by only a single document, WMTSTileSource(id='1001', ...) is already in a doc
```

Moving the import into the function solves the issue:

```python
from bokeh.io import show


def test():
    from bokeh.tile_providers import CARTODBPOSITRON  # Importing here works

    fig = figure(plot_width=1000, plot_height=1000,
                 active_scroll="wheel_zoom",
                 x_range=(49., 49.5), y_range=(30., 30.5),
                 x_axis_type="mercator", y_axis_type="mercator",
                 )

    fig.add_tile(CARTODBPOSITRON)

    show(fig)

test()
test()
```

This is neither very intuitive, nor mentioned in the [documentation](https://bokeh.pydata.org/en/latest/docs/reference/tile_providers.html) (as far as I have seen).

RunOrVeith:
Deprecate functionality to retrieve tile providers as already instanciated objectes,
which was confusing for multiple document use-cases.
Instead, add a new function `get_provider` that returns a provider and is thus more clear.

 -   Implement get_provider. It accepts the provider name as a string (case does not matter), or an
    element of a new enum Provider. This function creates the tile provider accordingly and returns the new instance of it. 
 -   Issue deprecations in the existing properties for the providers
 -  Update the test cases to assert the deprecation warning, and add a new testcase that tests `get_provider`. (The functionality of the other test cases is not duplicated, so that has to be done when the real deprecation happens).
 -   Update the docstring of tile_providers
 -   Update all examples (that I could find).

Example snippets:

```python
from bokeh.tile_providers import CARTODBPOSITRON

# Works, but raises BokehDeprecationWarning:
# "BokehDeprecationWarning: CARTODBPOSITRON was deprecated in Bokeh 2.0.0 and will be removed, use get_provider(Provider.CARTODBPOSITRON) instead."

from bokeh.tile_providers import get_provider, Provider

# Use the new enum for tile types
p1 = get_provider(Provider.CARTODBPOSITRON)

# Also works with string names (upper and lower case)
p2 = get_provider("CARTODBPOSITRON")

# Also works by passing in the "old" provider. Not sure about this feature
p3 = get_provider(CARTODBPOSITRON)
```

There are a few things which I am not sure about:

-    Whether I used the right deprecation protocol (I used bokeh.util.deprecation.deprecated)
-    Whether I found all examples
-    Whether the docstring is what you expect/the right format. I am not sure how to document the enum values
-    Whether I found all examples

Please let me know if anything is not what you expected/should be changed!

- [x] issues: fixes #8566 
- [x] tests added / passed: Yes
- [ ] release document entry (if new feature or API change): Not sure where to add

### DIFF
--- a/bokeh/tile_providers.py
+++ b/bokeh/tile_providers.py
@@ -1,48 +1,50 @@
-# -----------------------------------------------------------------------------
+#-----------------------------------------------------------------------------
 # Copyright (c) 2012 - 2019, Anaconda, Inc., and Bokeh Contributors.
 # All rights reserved.
 #
 # The full license is in the file LICENSE.txt, distributed with this software.
-# -----------------------------------------------------------------------------
+#-----------------------------------------------------------------------------
 ''' Pre-configured tile sources for common third party tile services.
 
 
 Attributes:
 
-    Provider
-        Enum listing all available tile services
+    ThirdParty
+        Enum listing all available preconfigured tiling services
 
     get_provider
-        Use this function to retrieve any tile provider.
+        Use this function to retrieve an instance of a predefined tile provider.
 
         Args:
-            provider_name (Union[str, Provider])
+            provider_name (Union[str, ThirdParty])
                 Name of the tile provider to supply.
-                Use tile_providers.Provider enum, or the name of a provider as string
+                Use tile_providers.ThirdParty enum, or the name of a provider as string
+                
 
         Returns:
             WMTSTileProviderSource: The desired tile provider instance
 
         Raises:
             ValueError, if the provider can not be found
+            ValueError, if the attribution for that provider can not be found
 
 
         Example:
 
-        .. code-block:: python
+            .. code-block:: python
 
-                >>> from bokeh.tile_providers import get_provider, Provider
-                >>> get_provider(Provider.CARTODBPOSITRON)
-                <class 'bokeh.models.tiles.WMTSTileSource'>
-                >>> get_provider('CARTODBPOSITRON')
-                <class 'bokeh.models.tiles.WMTSTileSource'>
+                    >>> from bokeh.tile_providers import get_provider, ThirdParty
+                    >>> get_provider(ThirdParty.CARTODBPOSITRON)
+                    <class 'bokeh.models.tiles.WMTSTileSource'>
+                    >>> get_provider('CARTODBPOSITRON')
+                    <class 'bokeh.models.tiles.WMTSTileSource'>
 
 
     CARTODBPOSITRON
         Tile Source for CartoDB Tile Service
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -52,7 +54,7 @@ Attributes:
         Tile Source for CartoDB Tile Service (tiles at 'retina' resolution)
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -62,7 +64,7 @@ Attributes:
         Tile Source for Stamen Terrain Service
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -72,7 +74,7 @@ Attributes:
         Tile Source for Stamen Terrain Service
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -82,7 +84,7 @@ Attributes:
         Tile Source for Stamen Toner Service
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -92,7 +94,7 @@ Attributes:
         Tile Source for Stamen Toner Background Service which does not include labels
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
@@ -102,7 +104,7 @@ Attributes:
         Tile Source for Stamen Toner Service which includes only labels
 
         Warns:
-            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
+            BokehDeprecationWarning: Deprecated in bokeh 1.1.0. Use get_provider instead
 
         .. raw:: html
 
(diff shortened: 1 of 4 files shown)


bokeh/tile_providers.py
```diff
@@ -1,59 +1,109 @@
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
```
bryevdv: The standardized boilerplate across the project does not have a space between the `#` and the first `-`, so these changes should be removed throughout.

bokeh/tile_providers.py
```diff
@@ -1,59 +1,109 @@
...
+
+        Example:
+
+        .. code-block:: python
+
+                >>> from bokeh.tile_providers import get_provider, Provider
+                >>> get_provider(Provider.CARTODBPOSITRON)
+                <class 'bokeh.models.tiles.WMTSTileSource'>
+                >>> get_provider('CARTODBPOSITRON')
+                <class 'bokeh.models.tiles.WMTSTileSource'>
```
bryevdv: I *think* this entire block, starting with `.. code-block` will need to be indented one more level under `Example` in order for Sphinx to render it correctly. We will have to locate some similar examples in the codebase to compare.

bokeh/tile_providers.py
```diff
@@ -1,59 +1,109 @@
...
+
+
     CARTODBPOSITRON
         Tile Source for CartoDB Tile Service
 
+        Warns:
+            BokehDeprecationWarning: Deprecated in bokeh 2.0.0. Use get_provider instead
```
bryevdv: The deprecation will be executed when 2.0 happens (i.e. that's when breaking changes will happen) but the deprecation will start from 1.1 (presumably) so we should state that throughout.

bokeh/tile_providers.py
```diff
@@ -65,46 +115,48 @@
...
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 # Imports
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
 # Standard library imports
 import sys
 import types
+from enum import Enum
```
bryevdv: We still support Python 2 at least through the end of the year, so unfortunately we cannot use the new enum feature in core Python 3. For now we will have to use our hand-rolled `Enum` class in `bokeh.core.enums`:

https://github.com/bokeh/bokeh/blob/master/bokeh/core/enums.py

bokeh/tile_providers.py
```diff
@@ -117,10 +169,45 @@ class _TileProvidersModule(types.ModuleType):
...
+            url=url,
+            attribution=attribution
+        )
+
     # Properties --------------------------------------------------------------
 
     @property
     def CARTODBPOSITRON(self):
+        from bokeh.util.deprecation import deprecated
+        deprecated(since_or_msg=(2, 0, 0), old='CARTODBPOSITRON', new='get_provider(Provider.CARTODBPOSITRON)')
```
bryevdv: `since_or_msg=(1,1,0)` throughout

bokeh/tile_providers.py
```diff
@@ -117,10 +169,45 @@ class _TileProvidersModule(types.ModuleType):
         'under %s.'
     )
 
+    class Provider(Enum):
```
bryevdv: I'm lukewarm about the name `Provider`, both because `tile_providers.Provider` seems redundant/clunky, but also because I think it might be nice to convey that these are the standard/built-in ones that Bokeh makes available. I haven't thought much about alternatives yet but interested if you have any ideas.

bryevdv:
@RunOrVeith Thanks for the PR! I have left some initial review comments. You are correct to use our `deprecated` function, that will handle issuing the `DeprecationWarning` (and hopefully in a way that is not muted by default)

bokeh/tile_providers.py
RunOrVeith: Oh damn I guess that was my auto-pep8. Will do.

bokeh/tile_providers.py
bryevdv: I have not thought at all about how using our `Enum` might be different (hopefully not much) but let me know if you run into any questions/problems.

examples/integration/tiles/tile_smoothing.py
```diff
@@ -14,9 +14,10 @@
 p2.y_range = p1.y_range
 
 # Add smoothed tiles (the default)
-r1 = p1.add_tile(CARTODBPOSITRON)
+tile_provider = get_provider(Provider.CARTODBPOSITRON)
```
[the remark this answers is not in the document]
RunOrVeith: Yeah with the current version, it could be done as simply as
`CARTODBPOSITRON = Provider.CARTODBPOSITRON` once the deprecation has happened,
and then it should work. I personally like the enum more than strings, because it prevents miss-spelling and the like, but that could also stay undocumented until 2.0 I guess.

bokeh/tile_providers.py
RunOrVeith: I also wasn't too hot on it. Possibly an alternative: `Supplier` or `Maps` or  `Vendor` or `Source` or `TileSource`. (or other combinations like `TileSupplier`). What do you think?

RunOrVeith:
I'll work on this more in the next few days!

bokeh/tile_providers.py
bryevdv: of those, I think I like `Vendor` or `Source` the most, but I will think on it some more. Another possible concept to possibly try and get across is "preconfigured"

bryevdv:
> I'll work on this more in the next few days! 

Brilliant, thanks so much for working this!

bokeh/tile_providers.py
RunOrVeith: I went for `ThirdParty`not, how do you think that sounds? I think it gets across  that it is preconfigured and a default tile source

examples/integration/tiles/tile_smoothing.py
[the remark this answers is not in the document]
bryevdv: I think preserving the deprecation warning is more important, we should be as rigorous as possible about telegraphing breaking changes that will occur in the future

bokeh/tile_providers.py
bryevdv: `ThirdParty` is a bit over-wordy for my taste, I'd try to keep it to a single word rather than two. I think so far `Vendor` or `Vendors` is OK with me

examples/integration/tiles/tile_smoothing.py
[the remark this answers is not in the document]
RunOrVeith: It is preserved. I just mean when you use `get_provider(CARTODBPOSITRON)` (not the string version), you will receive a deprecation warning even though this exact same behavior will work in 2.0, when the properties are changed to constants. This is due to the fact that the import of CARTODBPOSITRON triggers the warning already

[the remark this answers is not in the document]
bryevdv: Ah, I see. I would say that's OK then, as long as we don't demonstrate that usage anywhere right now, since the converse is also true: normal usage of supported features should not raise deprecations. We can document and convert examples to this style at 2.0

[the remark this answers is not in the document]
RunOrVeith: Would you remove it from the test case also (due to "demonstrated usage")?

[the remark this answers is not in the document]
bryevdv: I think it's OK to leave the test, it's the docs/examples that I think are relevant here (most users don't read tests AFAIK)

philippjfr:
Seems very reasonable to me.

bryevdv:
Thanks @RunOrVeith !

mattpap:
@bryevdv, @RunOrVeith , after this PR, `py.test -m unit` is littered with:
```
bokeh/tests/test_tile_providers.py::Test___all__::()::test_contents[CARTODBPOSITRON]
  BokehDeprecationWarning: CARTODBPOSITRON was deprecated in Bokeh 1.1.0 and will be removed, use get_provider(Vendors.CARTODBPOSITRON) instead.
```
Looking at the code, it looks like measures were taken to avoid this, but it seems it doesn't work.

PR #8594, opened 2019-01-30
Invalidate TileSource cache on property changes

### LINKED ISSUE #8164: TileRenderer cache is not invalidated when URL changes
When updating a TileSource URL the TileRenderer does not invalidate the existing cache which means as you zoom you will now get a mixture of old and new tiles. It seems the TileRenderer has an internal caching scheme that considers only the zoom level and location of the tile (i.e. in most cases the X/Y/Z 
 parameters of the tile source), the cache keys should probably also include the tile source URL to ensure that old tiles are not fetched from the cache when the URL has changed.

philippjfr:
Upgrades ``TileSource`` attributes to properties and add events to clear cache and rerender when a property changes. Let me know if my assumption that attributes need to be properties to trigger model change events is wrong.

geoviews_tilesources_small

- [x] issues: fixes #8164
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/tiles/tile_renderer.ts
+++ b/bokehjs/src/lib/models/tiles/tile_renderer.ts
@@ -52,6 +52,7 @@ export class TileRendererView extends RendererView {
   connect_signals(): void {
     super.connect_signals()
     this.connect(this.model.change, () => this.request_render())
+    this.connect(this.model.tile_source.change, () => this.request_render())
   }
 
   get_extent(): Extent {
--- a/bokehjs/src/lib/models/tiles/tile_source.ts
+++ b/bokehjs/src/lib/models/tiles/tile_source.ts
@@ -20,7 +20,17 @@ export namespace TileSource {
     initial_resolution: number
   }
 
-  export interface Props extends Model.Props {}
+  export interface Props extends Model.Props {
+    url: p.Property<string>
+    tile_size: p.Property<number>
+    max_zoom: p.Property<number>
+    min_zoom: p.Property<number>
+    extra_url_vars: p.Property<{[key: string]: string}>
+    attribution: p.Property<string>
+    x_origin_offset: p.Property<number>
+    y_origin_offset: p.Property<number>
+    initial_resolution: p.Property<number>
+  }
 }
 
 export interface TileSource extends TileSource.Attrs {}
@@ -60,6 +70,11 @@ export abstract class TileSource extends Model {
     this._normalize_case()
   }
 
+  connect_signals(): void {
+    super.connect_signals()
+    this.connect(this.change, () => this._clear_cache())
+  }
+
   string_lookup_replace(str: string, lookup: {[key: string]: string}): string {
     let result_str = str
     for (const key in lookup) {
@@ -83,6 +98,10 @@ export abstract class TileSource extends Model {
       .replace('{xmax}','{XMAX}')
       .replace('{ymax}','{YMAX}')
     this.url = url
+  }
+
+  protected _clear_cache(): void {
+    this.tiles = {}
   }
 
   tile_xyz_to_key(x: number, y: number, z: number): string {
--- a/bokehjs/test/models/tiles/tile_renderer.coffee
+++ b/bokehjs/test/models/tiles/tile_renderer.coffee
@@ -131,6 +131,15 @@ describe "tile sources", ->
         expect(t[0]).to.be.within(3, 4)
         expect(t[1]).to.be.within(3, 4)
 
+    it "should invalidate cache on property change", ->
+      tile_options =
+        url : 'http://mock/{x}/{y}/{z}.png'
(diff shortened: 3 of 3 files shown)


bryevdv:
> Let me know if my assumption that attributes need to be properties to trigger model change events is wrong.

This is true AFAIK @mattpap can comment if I am mistaken

bokehjs/src/lib/models/tiles/tile_renderer.ts
```diff
@@ -52,6 +52,7 @@ export class TileRendererView extends RendererView {
   connect_signals(): void {
     super.connect_signals()
     this.connect(this.model.change, () => this.request_render())
+    this.connect(this.model.tile_source.change, () => this.request_render())
```
bryevdv: So, what happens here if the tile source itself is swapped out?

bryevdv: Hrm I am not actually sure it is handled currently though I thought it was in the past. I would be OK with merging this and making a follow on issue to investigate more thorough signal disconnect/reconnect (in general, for glyphs and this and anything else)

bokehjs/src/lib/models/tiles/tile_source.ts
```diff
@@ -20,7 +20,17 @@ export namespace TileSource {
...
+    url: p.Property<string>
+    tile_size: p.Property<number>
+    max_zoom: p.Property<number>
+    min_zoom: p.Property<number>
+    extra_url_vars: p.Property<{[key: string]: string}>
+    attribution: p.Property<string>
+    x_origin_offset: p.Property<number>
+    y_origin_offset: p.Property<number>
+    initial_resolution: p.Property<number>
+  }
```
mattpap: > Let me know if my assumption that attributes need to be properties to trigger model change events is wrong.

Yes and no. If you want automation, then yes. You may try to use `setv()` directly, but that will fail, if "properties" weren't defined. You can, however, emit change signal manually if you want (e.g. from a setter). I'm writing this in the context of the above change, because it isn't necessary in this PR (though useful in general). From runtime perspective only `this.define(...)` matters. Also note that `define()` is untyped currently, so you can put whatever you want there (which obviously I don't recommend).

bokehjs/src/lib/models/tiles/tile_renderer.ts
philippjfr: +1

Would be neat if the property could detect a model was being replaced and then dis- and reconnect any signals that were attached to the model being replaced, instead of having to handle this reconnecting logic all over the place. I don't know enough to determine whether that's feasible though.

bokehjs/src/lib/models/tiles/tile_source.ts
philippjfr: Thanks for the clarification, I think in this case it is needed because we want the model to be responsive to any change events, no matter how they were triggered.

mattpap: Sure, the default is to use properties, especially as the part of the API. However, I just want to make sure that it's understood that the code above doesn't have any run time implications. Any interfaces or types (or anything else that is part of TypeScript and not JavaScript) are fully erased at compilation time. The only thing that matters at runtime is `define()`, and that was fully specified already. But as I said, it doesn't harm to add this code (in fact I have a WIP PR that fills all those in), just I want to make sure that be benefits and consequences of such change are fully understood.

philippjfr: Ah okay, I did misunderstand you initially then. Thanks for clarifying, the only thing I'm missing is why attributes and properties seem to be duplicating the same type information. Wouldn't it be sufficient to declare them once as either attributes or as properties?

mattpap: > Wouldn't it be sufficient to declare them once as either attributes or as properties?

Yes (\*), however, limitations of the language when we started the transition didn't allow for this. Currently I can use mapped and conditional types, to generate attributes from properties (this direction, because properties carry more information than attributes). I have a WIP PR that fixes that (after layout). However, (*) attributes and properties may, and in fact do in case of specs, carry different types, or even input (constructor) and output (model interface) types may differ. Altogether this can get pretty tricky, but with some type-level magic it can be handled without duplication.

bryevdv: > Wouldn't it be sufficient to declare them once as either attributes or as properties?

Just to add more color: the duplication is entirely because we developed a properties system in coffee script, if we had had the current version of typescript at the time we probably could have done else. Now we have the task of unifying our old sort-of-types with TypieScript's actual language capabilities, and that is an onerous task.

philippjfr: Thanks for the background, I really appreciate it.

bryevdv:
No I will merge now, please make a follow up issue about the other.

PR #8602, opened 2019-01-31
remove extraneous ctx.fill for bands

### LINKED ISSUE #8599: grid bands broken when cross grid lines disabled
This code:
```

output_file("gridbands.html")

p.circle([1,2,3,4,5], [2,5,8,2,7], size=10)

# change just some things about the x-grid

# change just some things about the y-grid
p.ygrid.band_fill_alpha = 0.1
p.ygrid.band_fill_color = "navy"

show(p)
```
results in this output:

screen shot 2019-01-30 at 20 21 04

If the xgrid setting is commented out, it displays correctly.

bryevdv:
There was an unnecessary `ctx.fill()` after the `ctx.fillRect(...)` to draw the bands. When the cross grid line color is `None` this actually supresses stroking and resetting some paths, and presumably this allowed a bad path accumulate for the `ctx.fill()`.

- [x] issues: fixes #8599

### DIFF
--- a/bokehjs/src/lib/models/grids/grid.ts
+++ b/bokehjs/src/lib/models/grids/grid.ts
@@ -226,6 +226,7 @@ export class Grid extends GuideRenderer {
 
     const cmin = cross_range.min
     const cmax = cross_range.max
+
     const coords: [number[][], number[][]] = [[], []]
 
     if (!exclude_ends) {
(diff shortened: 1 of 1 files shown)


bokehjs/src/lib/models/grids/grid.ts
```diff
@@ -227,7 +226,6 @@ export class Grid extends GuideRenderer {
 
     const cmin = cross_range.min
     const cmax = cross_range.max
-
```
mattpap: I would prefer not to remove this line.

bryevdv:
sure, done

PR #8604, opened 2019-02-01
remove python 3.5 from matrix

bryevdv:
In order to update to the latest conda and conda-build we would need to drop Python 3.5 from our test matrix. This PR is to prove out that things will indeed work, and have a discussion about whether to actually do it. 

I am personally not happy about how aggressively conda has dropped support for Python 3.4 and 3.5 but here we are. I am +0 on this change. FWIW builds do seem to go faster which is certainly a benefit to devs. 

@bokeh/dev thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ jobs:
         - google-chrome-stable --no-sandbox --headless --hide-scrollbars --remote-debugging-port=9222 &
       env:
       - GROUP=examples
-      - PYTHON=3.5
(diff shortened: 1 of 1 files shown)


bryevdv:
Thinking more, I guess the actual obstacle is installing conda and conda-build in root python 3.5 envs, in which case it might be possible to create non-root envs to run the tests in. But that will require a non-trivial re-tooling of some of our CI scripts.

PR #8615, opened 2019-02-01
Remove duplication of types between attributes and properties

### LINKED ISSUE #8612: Remove type duplication between models' attributes and properties
Currently type of every attribute of a model is provided twice, however this can be reduced to one with mapped and conditional types.

### LINKED ISSUE #8616: Finalize typing plotting API and remove old declarations
PR #8085 allowed for removal of old type declarations, however I kept them because they still have valid type information (though incomplete and a bit outdated) regarding plotting API. When this API is fully typed (this may rely on issue #8615 and possibly other work), then old declarations can be finally removed. Just to note, currently we automatically generate declarations from *.ts files.

mattpap:
fixes #8612 
fixes #8616


bokehjs/src/lib/core/properties.ts
```diff
@@ -196,49 +233,49 @@ export class Direction extends Enum<enums.Direction> {
...
+export const RadiusDimension = Enum(enums.RadiusDimension)
+export const Sort = Enum(enums.Sort)
+export const ButtonType = Enum(enums.ButtonType)
+export const HexTileOrientation = Enum(enums.HexTileOrientation)
+export const TapBehavior = Enum(enums.TapBehavior)
+export const BoxOrigin = Enum(enums.BoxOrigin)
+export const HoverMode = Enum(enums.HoverMode)
+export const PointPolicy = Enum(enums.PointPolicy)
+export const LinePolicy = Enum(enums.LinePolicy)
+export const MarkerType = Enum(enums.MarkerType)
```
bryevdv: These should all be alphabetized .

mattpap: This maintains order of core/enums, so I updated both. Though, I think, this code will be removed at some point in favor of inline enums.

bryevdv:
I will start 1.1dev3 soon

PR #8620, opened 2019-02-02
Make custom extensions compiler work on Windows

### LINKED ISSUE #8613: Inline TS code failed at compilation on windows (post #8085)
I tried to run this code :
```python
from bokeh.io import show
from bokeh.models import TickFormatter
from bokeh.util.compiler import TypeScript

TS_CODE = """
import {TickFormatter} from "models/formatters/tick_formatter"
import {AxisView} from "models/axes/axis"

export class MyFormatter extends TickFormatter {
    type: "MyFormatter"

    // TickFormatters should implement this method, which accepts a lisst
    // of numbers(ticks) and returns a list of strings
    doFormat(ticks: string[] | number[], _axis_view: AxisView): string[] {
        // format the first tick as- is
        const formatted = [`${ticks[0]}`]
        for (let i = 1, len = ticks.length; i < len; i++) {
            formatted.push(`+${(Number(ticks[i]) - Number(ticks[0])).toPrecision(2)}`)
        }
        return formatted
    }
}
"""


class MyFormatter(TickFormatter):

    __implementation__ = TypeScript(TS_CODE)


p = figure()
p.circle([1, 2, 3, 4, 6], [5, 7, 3, 2, 4])

p.xaxis.formatter = MyFormatter()

show(p)
```

And I get this error:
```cmd
    custom_impls = _compile_models(custom_models)
  File "c:\users\xavier\repos\bokeh\bokeh\util\compiler.py", line 615, in _compile_models
    raise CompilationError(compiled.error)
bokeh.util.compiler.CompilationError:
?[91merror?[0m?[90m TS6053: ?[0mFile 'c:/Users/Xavier/Repos/bokeh_custom_ext/Extensions/Ticker/extensions_example_ticking.py:MyFormatter.ts' not found.
```

If I put the TS_CODE in a separate file everything seems OK

mattpap:
fixes #8613

### DIFF
--- a/scripts/ci/install.unit
+++ b/scripts/ci/install.unit
@@ -5,7 +5,7 @@ set -x # echo commands
 
 # install NPM dependencies
 pushd bokehjs
-npm install --no-save --no-progress phantomjs
+npm install --no-save --no-progress
 popd
 
 if  [[ ! -z "${MINIMAL}" ]]; then
--- a/scripts/ci/install.unit
+++ b/scripts/ci/install.unit
@@ -5,7 +5,7 @@ set -x # echo commands
 
 # install NPM dependencies
 pushd bokehjs
-npm install --no-save --no-progress
+npm install --no-save --no-progress phantomjs
 popd
 
 if  [[ ! -z "${MINIMAL}" ]]; then
--- a/scripts/ci/install.unit
+++ b/scripts/ci/install.unit
@@ -5,7 +5,7 @@ set -x # echo commands
 
 # install NPM dependencies
 pushd bokehjs
-npm install --no-save --no-progress phantomjs typescript
+npm install --no-save --no-progress
 popd
 
 if  [[ ! -z "${MINIMAL}" ]]; then
--- a/scripts/ci/install.unit
+++ b/scripts/ci/install.unit
@@ -5,7 +5,7 @@ set -x # echo commands
 
 # install NPM dependencies
 pushd bokehjs
-npm install --no-save --no-progress
+npm install --no-save --no-progress phantomjs typescript
 popd
 
 if  [[ ! -z "${MINIMAL}" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,11 @@ notifications:
     secure: UYjR/NEuFdHEXwYxaDR3DxQ6dFrHG4vziqp1Fs+EAoI21bo4r/0/ywTU6+jedlUCiahiFrUqDcX1tdM4s5Y/tnatAwZtEaHdVYTNtislN+f/EuutlE/TUeo8VliK4P47rxDGfdLBxCLes6X80q8QEaOFeIyCauoouQ/keRbrLNs=
 
 before_install:
+  # path order is important here, need to find nodejs from testenv
   - export PATH="$HOME/miniconda/bin:$PATH"
-  - export PATH="$HOME/miniconda/envs/testenv/bin:$PATH"
   - export PATH="$HOME/build/bokeh/bokeh/bokehjs/node_modules/phantomjs-prebuilt/lib/phantom/bin:$PATH"
   - export PATH="$HOME/bin:$PATH"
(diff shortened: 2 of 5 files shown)


bryevdv:
Is there some kind of test we could add to run on Appveyor?

bryevdv:
Test passed on windows but failed on linux!

mattpap:
The compiler didn't find TS stdlib, so it's a problem with installation. Given that examples' tests work, some installation step must be missing from unit tests' jobs.

bryevdv:
OK, I will take a look later tonight

bryevdv:
Well, I'm not sure. I changed to a full `npm install` and it still failed. The other day I did make the tests start running out of a non-root conda-env but I am not sure how/why that should matter since `npm` is installed globally and `nodejs` is installed in the same testenv (and that testenv is first on the path)

bryevdv:
Oh, I think there are some other things first in the path. That might be it.

bryevdv:
OK well I tried putting the conda env first on the path and a full npm install and it is still failing. I will have to try things out locally in the docker image tomorrow.

bryevdv:
I think I know why this is failing. I expect that the tarball I make and upload to S3 to share across jobs does not include these files.

bryevdv:
well, not quite. The build does this: 
```
# upload the noarch directory for later jobs in this build
tar czf conda-bld-noarch.tgz "miniconda/conda-bld/noarch"
artifacts upload                        \
    --bucket "bokeh-travis"             \
    --target-paths "${TRAVIS_BUILD_ID}" \
    conda-bld-noarch.tgz

# upload the BokehJS build directory for later jobs in this build
tar czf bokehjs-build.tgz "build/bokeh/bokeh/bokehjs/build"
artifacts upload                        \
    --bucket "bokeh-travis"             \
    --target-paths "${TRAVIS_BUILD_ID}" \
    bokehjs-build.tgz
```
does the conda package not include these files? should it?

bryevdv:
Well, they appear to be in the built conda package:
```
~/anaconda/conda-bld/noarch/site-packages/bokeh/server/static/lib master
(base) ❯ ls
lib.dom.d.ts  lib.es2015.core.d.ts  lib.es2015.promise.d.ts  lib.es5.d.ts
```
so I don't really know

scripts/ci/install.unit
```diff
@@ -5,7 +5,7 @@ set -x # echo commands
 
 # install NPM dependencies
 pushd bokehjs
-npm install --no-save --no-progress phantomjs
+npm install --no-save --no-progress phantomjs typescript
```
mattpap: This should install typescript version as specified in package.json. Otherwise, in the window between new version is out and we update, this particular test job may fail compilation.

mattpap: The syntax is `typescript@3.3.1` (for the currently used version).

bryevdv: `npm install` doesn't use package.json?

bryevdv: I definitely don't want the burden of manually keeping that in sync. I'd just change to `npm install` (with no arguments) instead

mattpap: `npm install` does use package.json, but `npm install package_name` does not.

bryevdv: of course it does....

mattpap: That's new to me. I'm pretty sure it didn't work like this before.

bryevdv: Sorry, I was being sarcastic and complaining that of course it behaves in this inconsistent way re: ignoring package.json in some cases but not others

mattpap: One thing to keep in mind is that `--no-save` is important here, because otherwise npm would overwrite package.json. One it will make a mess, but more importantly it actually changes version specifications, so in our case it would replace `3.3.1`, with `^3.3.1`, which as minor it is, it could lead to issues.

bryevdv: noted, thanks

PR #8629, opened 2019-02-05
Daterangeslider value as date

### LINKED ISSUE #8628: DateRangeSlider returns datetime instead of date
DateRangeSlider has two options, returning a tuple of UNIX epochs or conveniently returning a tuple of datetimes, as explained in #6895.

I think it would be logical to have an option to return a tuple of dates, with the same time resolution as the actual DateRangeSlider. 

This could be fixed by adding a second convenience function very similar to the one defined in #6895.

YMRTN:
- [x] issues: fixes #8628 
- [x] tests added / passed

### DIFF
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 #-----------------------------------------------------------------------------
 
 # Standard library imports
-from datetime import datetime
+from datetime import datetime, date
 import numbers
 
 # External imports
@@ -200,6 +200,25 @@ class DateRangeSlider(AbstractSlider):
             d1 = v1
         if isinstance(v2, numbers.Number):
             d2 = datetime.utcfromtimestamp(v2 / 1000)
+        else:
+            d2 = v2
+        return d1, d2    \
+
+    @property
+    def value_as_date(self):
+        ''' Convenience property to retrieve the value tuple as a tuple of
+        datetime objects.
+
+        '''
+        if self.value is None:
+            return None
+        v1, v2 = self.value
+        if isinstance(v1, numbers.Number):
+            d1 = date.fromtimestamp(v1 / 1000)
+        else:
+            d1 = v1
+        if isinstance(v2, numbers.Number):
+            d2 = date.fromtimestamp(v1 / 1000)
         else:
             d2 = v2
         return d1, d2
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -209,6 +209,7 @@ class DateRangeSlider(AbstractSlider):
         ''' Convenience property to retrieve the value tuple as a tuple of
         date objects.
 
+        Added in version 1.1
         '''
         if self.value is None:
             return None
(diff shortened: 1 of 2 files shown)


YMRTN:
I didn't expect that all new commits appear separately here. Apologies for the mess.

bokeh/models/widgets/sliders.py
```diff
@@ -202,6 +202,25 @@ def value_as_datetime(self):
             d2 = datetime.utcfromtimestamp(v2 / 1000)
         else:
             d2 = v2
+        return d1, d2    \
+
+    @property
+    def value_as_date(self):
+        ''' Convenience property to retrieve the value tuple as a tuple of
+        date objects.
+
```
bryevdv: We should add a note here that states that this function was added in version 1.1

bryevdv:
@YMRTN This looks good, I only had one small comment

bokeh/models/widgets/sliders.py
YMRTN: Can this literally be 'Added in version 1.1'? I can't immediately find an example when looking through previous releases.

bryevdv: That's because this is one of the first additions since 1.0 :) I think that text on a line by itself would be fine.

YMRTN:
@bryevdv I added the version number. Should be good to go.

bryevdv:
Thanks, @YMRTN !

bryevdv:
@YMRTN I am afraid there is a snag. These tests fail for me locally:
```
>       assert s.value_as_date == (start, end)
E       assert (datetime.dat...e(2017, 8, 9)) == (datetime.date...(2017, 8, 10))
E         At index 0 diff: datetime.date(2017, 8, 8) != datetime.date(2017, 8, 9)
E         Full diff:
E         - (datetime.date(2017, 8, 8), datetime.date(2017, 8, 9))
E         ?                         ^                          ^
E         + (datetime.date(2017, 8, 9), datetime.date(2017, 8, 10))
E         ?                         ^                          ^^

bokeh/models/widgets/tests/test_slider.py:83: AssertionError
________________________________________________________ test_daterangeslider_value_as_date_when_set_mixed ________________________________________________________

    def test_daterangeslider_value_as_date_when_set_mixed():
        start = date(2017, 8, 9)
        end = date(2017, 8, 10)
        s = mws.DateRangeSlider(start=start, end=end,
                value=(start, convert_datetime_type(end)))
>       assert s.value_as_date == (start, end)
E       assert (datetime.dat...e(2017, 8, 9)) == (datetime.date...(2017, 8, 10))
E         At index 1 diff: datetime.date(2017, 8, 9) != datetime.date(2017, 8, 10)
E         Full diff:
E         - (datetime.date(2017, 8, 9), datetime.date(2017, 8, 9))
E         ?                                                    ^
E         + (datetime.date(2017, 8, 9), datetime.date(2017, 8, 10))
E         ?                                                    ^^

bokeh/models/widgets/tests/test_slider.py:90: AssertionError
```
And I expect it has to do with the fact that the conversion does not assume UTC (the value_as_datetime does this, with `datetime.utcfromtimestamp` but the `date` does not have a similar method). 

Do you think you can look into this? If not I think we will have to revert this PR before the next release, until it can be addressed.

bryevdv:
Actually I think it has something to do with `convert_datetime_type`, those are the values that are wrong. Perhaps it is not doing the correct thing for dates (not datetimes)

bryevdv:
fixed in https://github.com/bokeh/bokeh/pull/8653

PR #8636, opened 2019-02-08
Add js_link convenience method

bryevdv:
Add a method to enable simple linkages between JS attributes without explicitly having to import and create a CustomJS object:

```
select.js_link('value', plot, 'sizing_mode')
```
- [x] tests added / passed

Have had this in mind for a long time just had never added it. Seems fairly straightforward and uncontroversial, but cc @bokeh/dev for any thoughts.


philippjfr:
Seems like a great idea. In ``panel`` we have the same thing except the signature is slightly different:

```python
select.jslink(plot, value='sizing_mode')
```

bryevdv:
@philippjfr Right so the difference is that you are implicitly encoding what attribute is linked from, i.e. `select.jslink` always links the `value` property and nothing else. The Bokeh mechanism is a little more general but also a little more explicit, but that seems appropriate, in fact.

philippjfr:
Not quite, we are simply defining the mapping(s) using keywords, while you use two separate args. So both are general, just different spellings.

bryevdv:
Ah, I see. I'm trying to decide if I'd like that more here or not, I'll have to think on it.

bryevdv:
@mattpap is it documented somewhere how to add new baselines?

```
[FAIL] tests/baselines/sphinx/source/docs/user_guide/examples/interaction_linked_properties baseline doesn't exist
[WARN] reference image https://s3.amazonaws.com/bokeh-travis/1.1.0dev3/sphinx/source/docs/user_guide/examples/interaction_linked_properties.png doesn't exist
FAILED
```

mattpap:
I need to document this, but the procedure is to run tests locally (not all, use `-k` option to select only newly added examples). This will fail, but also will generate respective baseline files. Commit those files (only committed files are considered valid baselines) and rerun the test to make sure everything works (this helps to guard against any immediate non-deterministic nature of tests, though baselines currently only test layout and this is usually not affected by such issues).

mattpap:
I'm using this bash alias to not have to remember the testing command:
```sh
alias test_examples="time py.test -s -v -m examples --tb line --diff-ref origin/master --self-contained-html --report-path=~/tmp/examples.html
```
Then I would run:
```
test_examples -k new_example_name
```

bryevdv:
@mattpap when I run that with *examples/plotting/file/sizing_mode.py* it dumps a *sizing_mode.html* in the current directory. Where does that need to be checked in?

Also, when I run it for *sphinx/source/docs/user_guide/examples/interaction_linked_properties.py* it just displays the example in a browser, and no new file is created.

bryevdv:
@mattpap well, after adding, they are still failing on travis:
```
[FAIL] BASELINE DOESN'T MATCH (make sure to update baselines before running tests):
[FAIL] diff --git a/tests/baselines/examples/plotting/file/sizing_mode b/tests/baselines/examples/plotting/file/sizing_mode
[FAIL] index ca4ed87..e2c86b0 100644
[FAIL] --- a/tests/baselines/examples/plotting/file/sizing_mode
[FAIL] +++ b/tests/baselines/examples/plotting/file/sizing_mode
[FAIL] @@ -1,6 +1,6 @@
[FAIL] -Column bbox=[0, 0, 600, 657]
[FAIL] -  Select bbox=[5, 5, 590, 47]
[FAIL] -  Plot bbox=[0, 57, 600, 600]
[FAIL] +Column bbox=[0, 0, 600, 656]
[FAIL] +  Select bbox=[5, 5, 590, 46]
[FAIL] +  Plot bbox=[0, 56, 600, 600]
[FAIL]      LinearAxis bbox=[27, 576, 543, 24]
[FAIL]      LinearAxis bbox=[0, 5, 27, 571]
[FAIL]      Title bbox=[27, 5, 543, 0]
[WARN] generated and reference images differ: 26.97% of pixels
```

```
[FAIL] BASELINE DOESN'T MATCH (make sure to update baselines before running tests):
[FAIL] diff --git a/tests/baselines/sphinx/source/docs/user_guide/examples/interaction_linked_properties b/tests/baselines/sphinx/source/docs/user_guide/examples/interaction_linked_properties
[FAIL] index 1651075..434dd78 100644
[FAIL] --- a/tests/baselines/sphinx/source/docs/user_guide/examples/interaction_linked_properties
[FAIL] +++ b/tests/baselines/sphinx/source/docs/user_guide/examples/interaction_linked_properties
[FAIL] @@ -1,7 +1,7 @@
[FAIL] -Column bbox=[0, 0, 400, 441]
[FAIL] +Column bbox=[0, 0, 400, 440]
[FAIL]    Plot bbox=[0, 0, 400, 400]
[FAIL]      LinearAxis bbox=[18, 376, 352, 24]
[FAIL]      LinearAxis bbox=[0, 5, 18, 371]
[FAIL]      Title bbox=[18, 5, 352, 0]
[FAIL]      ToolbarPanel bbox=[370, 5, 30, 371]
[FAIL] -  Slider bbox=[5, 405, 390, 31]
[FAIL] +  Slider bbox=[5, 405, 390, 30]
[WARN] reference image https://s3.amazonaws.com/bokeh-travis/1.1.0dev3/sphinx/source/docs/user_guide/examples/interaction_linked_properties.png doesn't exist
FAILED
```
Is a manual edit the correct approach here? I am guessing it is subtle platform differences between OSX headless chrome and linux headless chrome

bryevdv:
I've gone ahead and pushed hand edits. This does suggest to me we should have a flag to easily turn these off, if running them locally on every platform is not going to be workable.

mattpap:
This looks like a bug to me (rounding error), but it also may be a very subtle platform difference. Before we jump to conclusions, we need to investigate the difference, because it really shouldn't be there as we control pretty much all aspects of rendering.

bryevdv:
@mattpap OK well FWIW it does pass on TravisCI linux now

bryevdv:
Just running all tests locally and several discrepancies pop up almost immediately, that appear to be more than rounding errors:
```
tests/test_examples.py::test_file_examples[/Users/bryanv/work/bokeh/examples/models/file/anscombe.py-example0-config0]
[INFO] Example run in 0.973s
[INFO] Example rendered in 1.180s
[JS] {log} file:///Users/bryanv/work/bokeh/bokeh/server/static/js/bokeh.js:3481:21 [bokeh] setting log level to: 'info'
[JS] {info} file:///Users/bryanv/work/bokeh/bokeh/server/static/js/bokeh.js:7592:34 [bokeh] document idle at 92 ms
[FAIL] BASELINE DOESN'T MATCH (make sure to update baselines before running tests):
[FAIL] diff --git a/tests/baselines/examples/models/file/anscombe b/tests/baselines/examples/models/file/anscombe
[FAIL] index b6f4e621e..6d0b5a762 100644
[FAIL] --- a/tests/baselines/examples/models/file/anscombe
[FAIL] +++ b/tests/baselines/examples/models/file/anscombe
[FAIL] @@ -1,6 +1,6 @@
[FAIL] -Column bbox=[0, 0, 984, 871]
[FAIL] -  Div bbox=[5, 5, 974, 61]
[FAIL] -  GridBox bbox=[0, 71, 984, 800]
[FAIL] +Column bbox=[0, 0, 984, 923]
[FAIL] +  Div bbox=[5, 5, 974, 65]
[FAIL] +  GridBox bbox=[0, 123, 800, 800]
[FAIL]      Plot bbox=[0, 0, 400, 400]
[FAIL]        LinearAxis bbox=[24, 376, 371, 24]
[FAIL]        LinearAxis bbox=[0, 29, 24, 347]
[WARN] generated and reference images differ: 51.23% of pixels
```
Given that, I plan to merge this PR when green (since that has nothing to do with this) and make a separate issue to investigate the OSX/linux differences

PR #8644, opened 2019-02-11
Rewrite test/models in TypeScript

### LINKED ISSUE #7864: Rewrite bokehjs' tests in TypeScript

mattpap:
After this PR `test/index`, `test/document/document`, `test/core/ui_events` and `test/client/connection` will still need a rewrite.

addresses #7864


bokehjs/test/models/mappers/categorical_color_mapper.ts
```diff
@@ -0,0 +1,342 @@
...
+            end: j,
+          })
+
+          const vals0 = cm0.v_compute(["a"])
+          expect(vals0).to.be.deep.equal(["gray"])
+
+          /* XXX: gives ["red"] because original code used [i, j] = [3, 4] due to lack of lexical scoping
+          const vals1 = cm0.v_compute([["a", "1"]])
+          expect(vals1).to.be.deep.equal(["gray"])
+          */
```
mattpap: @bryevdv, the original code always used `[i, j] = [3, 4]` due to CoffeeScript's messed up scoping rules (`for` loop without `do` statement). This now fails with `["red"]` result, but I thing this is actually correct behavior.

bokehjs/test/models/mappers/categorical_marker_mapper.ts
```diff
@@ -0,0 +1,360 @@
...
+            markers,
+            factors: factors.map((x) => x.slice(0, 2) as F2),
+            start: i,
+            end: j,
+          })
+
+          const vals0 = cm0.v_compute(["a"])
+          expect(vals0).to.be.deep.equal(["circle"])
+
+          /* XXX: gives ["hex"] because original code used [i, j] = [3, 4] due to lack of lexical scoping
```
mattpap: @bryevdv, similarly here, just fails with `["hex"]`, which I think is correct.

bokehjs/test/protocol/message.ts
```diff
@@ -0,0 +1,179 @@
...
+        expect(m.buffers).to.be.deep.equal([])
+      })
+    })
+
+    describe("assemble_buffer method", () => {
+      const m = Message.create("FOO", {bar:2}, {baz:3})
+      m.header.num_buffers = 2
+
+      it("should append a new buffer", () => {
+        m.assemble_buffer(1, 2)
```
mattpap: @bryevdv, I suppose that arguments to `assemble_buffer()` are just whatever/throwaway values (they don't match types I figured out in `src/lib/protocol/message.ts`)?

bokehjs/test/models/mappers/categorical_color_mapper.ts
bryevdv: I concur it should be `"red"`

bokehjs/test/models/mappers/categorical_marker_mapper.ts
bryevdv: Yes, i think this specific one should actually map to `hex` as correct behavior

bokehjs/test/protocol/message.ts
bryevdv: the values here in the test are just junk, yes

Issue #8645, opened 2019-02-11
Use cross-browser compatible CSS for testing


bryevdv:
Detected during #8636 

Running on OSX, differences pop up frequently (but not all the time):

```
tests/test_examples.py::test_file_examples[/Users/bryanv/work/bokeh/examples/models/file/anscombe.py-example0-config0]
[INFO] Example run in 0.973s
[INFO] Example rendered in 1.180s
[JS] {log} file:///Users/bryanv/work/bokeh/bokeh/server/static/js/bokeh.js:3481:21 [bokeh] setting log level to: 'info'
[JS] {info} file:///Users/bryanv/work/bokeh/bokeh/server/static/js/bokeh.js:7592:34 [bokeh] document idle at 92 ms
[FAIL] BASELINE DOESN'T MATCH (make sure to update baselines before running tests):
[FAIL] diff --git a/tests/baselines/examples/models/file/anscombe b/tests/baselines/examples/models/file/anscombe
[FAIL] index b6f4e621e..6d0b5a762 100644
[FAIL] --- a/tests/baselines/examples/models/file/anscombe
[FAIL] +++ b/tests/baselines/examples/models/file/anscombe
[FAIL] @@ -1,6 +1,6 @@
[FAIL] -Column bbox=[0, 0, 984, 871]
[FAIL] -  Div bbox=[5, 5, 974, 61]
[FAIL] -  GridBox bbox=[0, 71, 984, 800]
[FAIL] +Column bbox=[0, 0, 984, 923]
[FAIL] +  Div bbox=[5, 5, 974, 65]
[FAIL] +  GridBox bbox=[0, 123, 800, 800]
[FAIL]      Plot bbox=[0, 0, 400, 400]
[FAIL]        LinearAxis bbox=[24, 376, 371, 24]
[FAIL]        LinearAxis bbox=[0, 29, 24, 347]
[WARN] generated and reference images differ: 51.23% of pixels
```

Chrome was run with
```
/Applications/Google\ Chrome\ Canary.app/Contents/MacOS/Google\ Chrome\ Canary --headless --remote-debugging-port=9222 --disable-gpu https://chromium.org
```

Version : `Google Chrome 69.0.3489.0 canary`

And tests were run with:
```
py.test -s -v -m examples --tb line --diff-ref origin/master -k examples
```

bryevdv:
[...]

PR #8646, opened 2019-02-12
isolate instantiation/initialization from json better

### LINKED ISSUE #8627: issue with bokeh.client after layout PR
@mattpap can you fill in the details here? (How to reproduce, etc). It's been awhile and this keeps falling off my radar because there has not been an issue.

bryevdv:
- [x] issues: fixes #8627
- [ ] tests added / passed

@mattpap if you think this seems reasonable I will add some tests and merge.


bryevdv:
OK app tests run but need baselines
```
tests/test_examples.py::test_server_examples[/home/travis/build/bokeh/bokeh/examples/app/apply_theme.py-example0-config0] 
[INFO] Example rendered in 1.060s
[JS] {log} http://localhost:5006/static/js/bokeh.min.js?v=8ec2289e1baa0c3e0c0b9e018735cea5:31:52635 [bokeh] setting log level to: 'info'
[JS] {info} http://localhost:5006/static/js/bokeh.min.js?v=8ec2289e1baa0c3e0c0b9e018735cea5:31:4611 [bokeh] Websocket connection 0 is now open
[JS] {info} http://localhost:5006/static/js/bokeh.min.js?v=8ec2289e1baa0c3e0c0b9e018735cea5:31:118316 [bokeh] document idle at 65 ms
[JS] {log} http://localhost:5006/static/js/bokeh.min.js?v=8ec2289e1baa0c3e0c0b9e018735cea5:31:132810 Bokeh items were rendered successfully
[FAIL] tests/baselines/examples/app/apply_theme baseline doesn't exist
[WARN] reference image https://s3.amazonaws.com/bokeh-travis/1.1.0dev3-876d72e/examples/app/apply_theme.png doesn't exist
FAILED
```

bryevdv:
Ok I added the baselines, we will see if they match on Travis. However, even after adding them, at least locally it says they are not found (and hence still fail):

> [FAIL] tests/baselines/examples/app/apply_theme baseline doesn't exist

(it definitely does exist)

bryevdv:
OK, it is finding the baselines, but they don't match. I'd kind of like to get this merged, without holding up for #8645 to be fixed. Maybe you can generate baselines on linux?

mattpap:
`app/faces` uses `cv2` module which isn't specified in test dependencies. Also, exceptions in server examples don't make tests fail, because bokeh.application just logs them. I think that at least for testing we should allow exceptions to fail a program.

mattpap:
Actually, `cv2` is installed by `opencv`.

bryevdv:
> I think that at least for testing we should allow exceptions to fail a program.

Certainly seems like the correct thing to me

PR #8662, opened 2019-02-16
use ci.bokeh.org for artifact storage

bryevdv:
Migrate CI test artifact storage to ci.bokeh.org

PR #8664, opened 2019-02-17
add support for server-sent events

### LINKED ISSUE #8663: Add support for data source using server-sent events
Currently bokeh uses Ajax polling or web sockets to receive data from a backend server. It would be nice to use server-sent events in some situations, since they can simplify the backend-code and are a good option for one way data streaming.

alhoo:

- [x] issues: fixes #8663
- [ ] tests added / passed


bryevdv:
Since server-sent events are incompatible with the `polling_interval` I think it might be worth considering making a new separate class that avoids that situation by not having a `polling interval`. It's always hard/confusing to document options that only apply to some situations, but not others. The `load_data` function could still be factored out and shared by both `AjaxDataSource` an the the one (`ServerSentDataSource` ?)

alhoo:
Good points! Currently the polling_interval is unfortunately defined in the `RemoteDataSource`, so the `ServerSentDataSource` would need to have its parent class above it. Should I add another abstract class between the `ColumnDataSource` and the `RemoteDataSource` and move the `url` and `load_data` there, or can I break the `RemoteDataSource` and move the polling_interval to `AjaxDataSource`?

bryevdv:
Let's add a new parent class for now, and I will make a note in #8666 to fix up the hierarchy (with a breaking change) for a 2.0 release.

bryevdv:
@alhoo I should have mentioned earlier, I think this new class can also be a base for `RemotaDataSource`, so that `adapter`, `data_url` could be factored in to one place. Does that seem reasonable?

alhoo:
I thought it would be easiest to make a new "WebDataSource" which contains the url, adapter, mode and the data loading function, but we might not want to move the AjaxDataSource to not use the RemoteDataSource. I'm not sure what will break if we do this?

alhoo:
We could of course still put the `RemoteDataSource` between the `WebDataSource` and `AjaxDataSource`, and that would contain the `polling_interval`.

bryevdv:
Actually, sorry I think there is still a change I would make, I think it should be:

```
class WebDataSource(ColumnDataSource)

class ServerSentDataSource(WebDataSource)

class RemoteSource(WebDataSource)

class AjaxDataSource(RemoteSource)
```
In your code you have more or less done away with usage of RemoteDataSource, but I think we should keep it in the class hierarchy, to avoid disturbing any users that might be doing `isinstance` checks, etc.  Then in 2.0 we can get rid of or rename `RemoteSource` as appropriate

alhoo:
Yeah I'll do that.

bryevdv:
@alhoo brilliant thank you for working on this it seems like a very nice and useful feature that had not been on our radar. If there is any public source that an example could be connected to that would also be great, or if it's simple to make an example similar to the current flask based ajax data source, I think that would be a great help to users in using this feature

alhoo:
I noticed that I of course needed to add the new server sent data source to the python models list. There is some documentation that might be wanted to written better. I added a small example to test the server sent events similar to the ajax module.

alhoo:
I'll do some clean up tomorrow. There is at least a couple of things still that needs refining.

bryevdv:
@alhoo I think all the duplicate properties like `http_headers` can be put on `WebSource` now? Anything that applies to both

bryevdv:
> I'll do some clean up tomorrow. There is at least a couple of things still that needs refining.

OK, great! Thanks!

bryevdv:
FYI @alhoo this PR will have to be rebased on master now, after #8662

bryevdv:
@alhoo not sure what's happened to the revision history here, but it seems like it might just be simpler to open a new PR at this point

PR #8672, opened 2019-02-19
open bokeh serve notebooks with utf8

### LINKED ISSUE #8625: Use utf-8 encoding to open notebook file.
Currently,  `ipynb` file is opened by following code:

```
        with open(filename) as f:
            pass
```

https://github.com/bokeh/bokeh/blob/master/bokeh/application/handlers/notebook.py#L117

I suggest to open the file with `encoding="utf-8"`, otherwise the serve command will raise error on Windows system.

bryevdv:
- [x] issues: fixes #8625

### DIFF
--- a/bokeh/_testing/util/examples.py
+++ b/bokeh/_testing/util/examples.py
@@ -240,7 +240,7 @@ class Example(object):
 
     @property
     def ref_url_path(self):
-        return join(self._diff_ref, self.relpath_no_ext) + '.png'
+        return join("travis", "image_refs", __version__, self._diff_ref, self.relpath_no_ext) + '.png'
 
     @property
     def diff_url_path(self):
--- a/bokeh/_testing/util/examples.py
+++ b/bokeh/_testing/util/examples.py
@@ -236,7 +236,7 @@ class Example(object):
 
     @property
     def img_url_path(self):
-        return join("travis", JOB_ID, __version__, self.relpath_no_ext) + '.png'
+        return join("travis", "image_refs", __version__, self.relpath_no_ext) + '.png'
 
     @property
     def ref_url_path(self):
@@ -244,7 +244,7 @@ class Example(object):
 
     @property
     def diff_url_path(self):
-        return join("travis", JOB_ID, __version__, self.relpath_no_ext) + self._diff_ref + '-diff.png'
+        return join("travis", "image_refs", __version__, self.relpath_no_ext) + self._diff_ref + '-diff.png'
 
     @property
     def has_ref(self):
--- a/bokeh/_testing/util/examples.py
+++ b/bokeh/_testing/util/examples.py
@@ -240,7 +240,7 @@ class Example(object):
 
     @property
     def ref_url_path(self):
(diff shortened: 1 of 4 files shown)


bryevdv:
@mattpap here is a PR we can iterate on for CI issues

bryevdv:
@mattpap OK I am still not seeing any images in the report:

https://ci.bokeh.org/travis/495611698/examples.html#bokehjs/examples/hover/hover

screen shot 2019-02-19 at 11 23 23

 But the reports have looked exactly like this to me for a long time so I do not know if this is correct. The "link" there just goes to itself. not to the images.

bryevdv:
Also, as earlier, the logs are super truncated:

https://ci.bokeh.org/travis/495611698/examples.log

mattpap:
> OK I am still not seeing any images in the report:

You have to click "show all" to see current images of successful tests.

bryevdv:
@mattpap ah OK I missed that change. There is no diff/ref in this build then:

screen shot 2019-02-19 at 11 29 14

mattpap:
That aside, things still don't work, because I don't see how we can reference an image from a different job. Previously it worked, because version number was the only discriminant.

```
diff --git a/bokeh/_testing/util/examples.py b/bokeh/_testing/util/examples.py
index 302563d..52c9a4a 100644
--- a/bokeh/_testing/util/examples.py
+++ b/bokeh/_testing/util/examples.py
@@ -240,7 +240,7 @@ class Example(object):
 
     @property
     def ref_url_path(self):
-        return join(self._diff_ref, self.relpath_no_ext) + '.png'
+        return join("travis", ??????, self._diff_ref, self.relpath_no_ext) + '.png'
 
     @property
     def diff_url_path(self):
````

**Comment by bryevdv** [comment]
Fair enough, if we need to move out of `JOB_ID` so be it, but just needs to be under some subdirectory of the top level

**Comment by bryevdv** [comment]
@mattpap in any case what do you propose for next steps? Are you going to make a follow on OR or do you want me to take a look?

**Comment by mattpap** [comment]
> Are you going to make a follow on PR or do you want me to take a look?

I would prefer if you would look into this, I need to get some layout work done.

**Comment by bryevdv** [comment]
OK I will just keep working on this PR for now

**Comment by bryevdv** [comment]
@mattpap OK  I merged this and now an master it appears to be looking for refs based on the version that is actually cloned in the job:

https://travis-ci.org/bokeh/bokeh/jobs/495712201
```
$ git checkout -qf cf6c97e35301adb15dd69eb24f4f8af5e75d64e1

[WARN] reference image https://ci.bokeh.org/travis/image_refs/1.1.0dev6-cf6c97e/bokehjs/examples/burtin/burtin.png doesn't exist
```

bryevdv:
on the PR job, had ` 'TRAVIS_COMMIT': '69ae3ec9096dc813ea1d39d05a546f547bd85e37',` so it looks like the job stored ref images (for future jobs to use) correctly for its own hash. So then it appears when the diff happens, it's looking in the wrong place? I could not find where/how `_diff_ref` is computed, though.

PR #8685, opened 2019-02-25
Observe changes to offset parent and relayout

### LINKED ISSUE #8655: Loading plot in separate JupyterLab tab collapses layout
As I originally outlined in https://github.com/bokeh/bokeh/issues/8084 since the layout PR you can get into situation where a layout collapses completely when reloading a notebook window. Since then I've further narrowed it down, and figured out that this only occurs in JupyterLab when the plot is in a different tab than the current one being displayed.

Here is a screencast demonstrating the issue, in it I have two notebooks open each with a bokeh layout. When I reload the notebook (by refreshing the page) the layout of widgets in the second notebook, which is not visible during the reload, collapses.

screencast_jlab

My guess is that this occurs because non-active JupyterLab tabs render but do not correctly report their size. I'm not really sure if this will be an easy thing to fix and it may have to happen in the JupyterLab extension telling bokeh to rerender the plots in the newly activated tabs.

mattpap:
This fixes the problem universally (not only in jupyter). This doesn't affect performance (computing `offsetParent` is cheap, compared to other DOM APIs).

fixes #8655

### DIFF
--- a/bokehjs/src/lib/models/layouts/layout_dom.ts
+++ b/bokehjs/src/lib/models/layouts/layout_dom.ts
@@ -27,6 +27,10 @@ export abstract class LayoutDOMView extends DOMView {
 
   protected _on_resize?: () => void
 
+  protected _offset_parent: Element | null = null
+
+  protected _parent_observer?: number
+
   protected _viewport: Partial<Size> = {}
 
   layout: Layoutable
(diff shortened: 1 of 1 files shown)


bryevdv:
> This doesn't affect performance (computing `offsetParent` is cheap, compared to other DOM APIs).

@mattpap can you provide an updated ETA on the performance work?  Other projects depending on it are starting to be exposed to schedule risk

philippjfr:
Just wanted to confirm that this indeed fixes the issue I reported in #8655 for me.

bryevdv:
@philippjfr thanks for testing things out (hopefully we can make some automated notebook testing a priority later this year)

mattpap:
> can you provide an updated ETA on the performance work?

@bryevdv, tomorrow. I think I finally figured out all the details, I just need to finish the implementation.

Issue #8084, opened 2018-07-17
Improve reliability, performance and feature coverage of the layout


mattpap:
This is a meta issue addressing specific issues with the layout (`tag: layout`). 

* [x] words_and_plots.py
    blank until browser resize, DOM text on top of iris plot after

* [x] models/glyphs.py
    tab bars truncated

* [x] dash
    range plot on top of (covering) the line plot

* [x] gapminder
    widgets on top of scatter plot
    plot can overlap top of text below
    Play/Pause does not work

* [x] movies
    FIXED_SIZING_MODE warning
    huge gap between top text and bokeh content, until browser resize, then snaps to place

* [x] OHLC
    would be nice if sliders and button were vertically aligned

* [ ] spectrogram (issue #8710)

    - [x] slider label overlaps text above
    - [ ] would be nice if right plot top/bottom border aligned with tops/bottoms of plots to the left

* [ ] integration/glyphs/hex_tile (issue #8711)

    Last column of plots has too narrow left panel (minus sign is cut). This isn't strictly an issue with the new layout, but a problem with the range-based aspect management scheme, which doesn't work well with old layout either, causing infinite update loops (e.g. plotting/file/aspect).

* [ ] add `GridPlot` model and make handling of a toolbar an implementation detail

* [x] finalize implementation of flow layout (`HTMLBox`, `Div`, some other widgets)

* [x] add support for tabs scrolling (or some other mechanism, e.g. a dropdown)

- [x] Fix performance for larger dashboards with many widgets, divs etc. (see https://github.com/bokeh/bokeh/pull/8085#issuecomment-459345949 for an example app)

- [x] Ensure that changes to width/height and other layout related properties trigger a new layout

- [x] Avoid slider widgets cutting off the handle and text (see https://github.com/bokeh/bokeh/pull/8085#issuecomment-459207880)

mattpap:
(based on https://github.com/bokeh/bokeh/pull/8085#issuecomment-451315916)

@mattpap I went though maybe ~80% of examples. Things are generally very good but I wanted to leave some observations for you before merging, at least 1 or 2 need some attention:

* [x] ~~line_compare.py~~
    line join/end options have no effect

* [ ] marker_compare.py
    Tabs run off screen

* [x] ~~sizing_mode.py~~
    changes to dropdown have no effect at all
    FIXED_SIZING_MODE warning

* [ ] events_app.py,  js_events.py
[...]

PR #8682, opened 2019-02-22
Make autoload_js load CSS before JavaScript

mattpap:
I tested this manually in Jupyter notebook and lab.

fixes #8681

PR #8689, opened 2019-02-26
Improve performance of widget heavy layouts

### LINKED ISSUE #8622: Input Widget heights behavior
I tried to fix height for widgets of type input however the input element does not fit the height specification

```python
from bokeh.models import Slider, TextAreaInput

from bokeh.io import show, output_notebook, output_file
from bokeh.layouts import column, widgetbox
output_file('test.html')

text = TextAreaInput(value='test', height=100, rows=50, sizing_mode='fixed')
slid = Slider(value=5, start=0, end=10)
show(widgetbox(text, slid))
```
image

I would have expected this output:
image

mattpap:
addresses #8084
fixes #8622


philippjfr:
Apologies if this is way too premature but I'm way too excited about this. First of all to say that this solves all the performance issues, thing are super snappy now. This whole layout refactor is huge step forward and I'm super excited about it.

In case it is at all helpful, my initial attempts did raise a few errors, e.g. about Select not being a valid model and this error I haven't looked into further:

```
Uncaught TypeError: (intermediate value)(...) is not a function
    at inline_js (<anonymous>:40284:7)
```

The only other issue is centering of widget titles:

screen shot 2019-02-26 at 6 53 44 pm

I'll do some more thorough testing once you've removed the WIP label.

mattpap:
> `Uncaught TypeError: (intermediate value)(...) is not a function
    at inline_js (<anonymous>:40284:7)`

This looks like a side effect of PR #8682.

mattpap:
@philippjfr, this is ready. Also, how did you end up with the above error?

mattpap:
> The only other issue is centering of widget titles:

This should be fixed now. What browser do you use? I used a wrong value for alignment and both FF and Chrome rendered it correctly anyway.

philippjfr:
I was using Chrome on OSX but it is indeed fixed now. I think the other errors must have been due to an old bokehjs sticking around or something similar because I can't reproduce them anymore. If it turns out that I can reproduce them later I'll just file an issue.

bryevdv:
Unless there is objection I'll merge this and cut a dev build in the morning

PR #8697, opened 2019-03-01
Compute content size with subpixel precision

### LINKED ISSUE #8654: Text wrapping in Div after layout PR
On current master after the layout PR the sizing behavior for ``Div`` is hugely improved however there is one new behavior which is quite frustrating. By default it seems the Div now wraps any text in it depending on the length of the string. Here's a reproducible example of the issue:

```python 
tinput = TextInput(value='Some text')
div = Div(text='Some Text')

tinput.js_link('value', div, 'text')

Row(children=[tinput, div])
```

div_text

Since Divs can be quite helpful for adding labels, titles and short snippets of instructions this can be quite annoying. Even if there was a convenient workaround here that would be very helpful.

mattpap:
@philippjfr, can you try this out? This should resolve issue with unnecessary wrapping.

fixes #8654

### DIFF
--- a/bokehjs/src/lib/core/dom.ts
+++ b/bokehjs/src/lib/core/dom.ts
@@ -213,14 +213,14 @@ export function outer_size(el: HTMLElement): Size {
 }
 
 export function content_size(el: HTMLElement): Size {
+  const {left, top} = el.getBoundingClientRect()
   const {padding} = extents(el)
-  const left = Math.ceil(padding.left)
-  const top = Math.ceil(padding.top)
   let width = 0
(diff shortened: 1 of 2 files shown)


bryevdv:
Setting up new email accounts and notifs, this is a test

bryevdv:
I'll plan to merge this Monday unless I hear otherwise

philippjfr:
I'll test this shortly.

philippjfr:
I can confirm this fixes the text wrapping issues I was seeing.

Issue #8708, opened 2019-03-04
handlebars security vulnerability - npm won't install


birdsarah:
On running `npm install --no-save`

I get the response

```
> bokehjs@1.1.0-dev8 prepare /home/bird/Dev/bokeh/bokeh/bokehjs
> node ./prepare.js

audited 738 packages in 1.96s
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
```

Results of `npm audit`

```
                                                                    
                       === npm audit security report ===                        
                                                                                
# Run  npm update handlebars --depth 2  to resolve 1 vulnerability
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ handlebars                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ istanbul [dev]                                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ istanbul > handlebars                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/755                             │
└───────────────┴──────────────────────────────────────────────────────────────┘


found 1 high severity vulnerability in 738 scanned packages
  run `npm audit fix` to fix 1 of them.
```

PR #8709, opened 2019-03-04
After running npm audit fix

birdsarah:
Fixes: #8708

Not sure if this is all that needs to be done, but this is what updated after running `npm audit fix`

### DIFF
--- a/bokehjs/package-lock.json
+++ b/bokehjs/package-lock.json
@@ -192,23 +192,12 @@
         "json-schema-traverse": "^0.3.0"
       }
     },
-    "align-text": {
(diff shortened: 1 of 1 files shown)


bryevdv:
The only thing that caught my eye is that this seems to bring in `lodash` but maybe we will have to live with that

mattpap:
This is introduced by a dev dependency (istanbul), so it shouldn't be a big concern.

mattpap:
Unforunately, appveyor tests are failing all day everywhere.

bryevdv:
Thanks @birdsarah

PR #8714, opened 2019-03-05
Run inline js in individual functions in autoload_js

### LINKED ISSUE #8713: Displaying plot in notebook quickly after output_notebook raises JS error
Since https://github.com/bokeh/bokeh/pull/8682 was merged displaying a plot in a notebook environment raises a JS error if it is run before ``output_notebook`` has had time to initialize Bokeh properly. A simple way to reproduce this is by simply putting ``output_notebook`` and ``show`` in the same cell in a clean notebook and running it:

```python

output_notebook(resources=INLINE)

p = figure()
p.circle([1, 2, 3], [1, 2, 3])

show(p)
```

>Javascript error adding output!
>TypeError: Cannot read property 'set_log_level' of undefined
>See your browser Javascript console for more details.

mattpap:
fixes #8713

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -125,7 +125,7 @@ calls it with the rendered model.
       {{ js|indent(6) }}
     },
     {% endfor -%}
-    function(Bokeh) {}
+    function(Bokeh) {} // ensure no trailing comma for IE
   ];
 
   function run_inline_js() {
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -115,16 +115,17 @@ calls it with the rendered model.
   var css_urls = {{ css_urls|json }};
 
   var inline_js = [
+    {%- for css in css_raw %}
     function(Bokeh) {
-      {%- for css in css_raw %}
-        inject_raw_css({{ css }});
-      {%- endfor %}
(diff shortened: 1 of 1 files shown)


bokeh/core/_templates/autoload_js.js
```diff
@@ -115,16 +115,17 @@ calls it with the rendered model.
...
+    {%- for js in js_raw %}
     function(Bokeh) {
-      {%- for js in js_raw %}
-        {{ js|indent(6) }}
-      {% endfor -%}
-    }
+      {{ js|indent(6) }}
+    },
+    {% endfor -%}
+    function(Bokeh) {}
```
bryevdv: What's the purpose of this last no-op function?

mattpap: To make sure there's no trailing comma (for IE). Otherwise I would have to use some jinja's magic loop variables.

bryevdv: Comment to that effect would be helpful I think

philippjfr:
I can confirm that this fixes the error I was seeing.

gkanarek:
test comment requested by @bryevdv

gkanarek:
another test comment requested by @bryevdv

bryevdv:
only failure is appveyor, just merging on top of the appveyor fix PR

PR #8722, opened 2019-03-07
darken and desat disabled buttons

### LINKED ISSUE #8717: Disabled buttons should be gray by default
```
Python version      :  3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
IPython version     :  7.1.1
Tornado version     :  5.1.1
Bokeh version       :  1.1.0dev8
BokehJS static path :  /Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/server/static
node.js version     :  v7.10.0
npm version         :  4.2.0
```
Disabled buttons should be gray by default so that it is clear that they are disabled. Now it seems that button color gets muted when disabled which is fine unless the button background is white which it is by default. 

I first brought up this issue here: https://github.com/pyviz/panel/issues/275 where it's been noted that other widget behavior is similar.

```python
from bokeh.io import show

button = Button(label="Foo", disabled=True)

show(widgetbox(button))
```

screen shot 2019-03-05 at 11 50 37 am

bryevdv:
- [x] issues: fixes #8717

Updated test script to cover all buttons
```
from bokeh.io import show
from bokeh.core.enums import ButtonType

buttons = [Button(label=x, disabled=True, button_type=x) for x in ButtonType]

show(widgetbox(*buttons))
```

Current output:

screen shot 2019-03-06 at 3 58 22 pm

cc @mattpap @jsignell I think this looks pretty reasonable but open to suggestion

### DIFF
--- a/bokehjs/src/less/buttons.less
+++ b/bokehjs/src/less/buttons.less
@@ -48,7 +48,7 @@
 
   &[disabled] {
     &, &:hover, &:focus, &:active, &.bk-active {
-      background-color: @background;
+      background-color: darken(desaturate(@background, 30%), 10%);
       border-color: @border;
     }
   }


jbednar:
Looks good!  Does it also work well with a dark theme?

bryevdv:
> Does it also work well with a dark theme? 

not sure someone will have to try it

PR #8729, opened 2019-03-07
add phantomjs to test deps

### LINKED ISSUE #8571: `scripts/deps.py` does not install phantom_js
Have been running through a fresh install and build and followed the dev instructions here: https://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html

To install dependencies I ran:  
    
    conda install `python scripts/deps.py build run test` 

After doing this if you run `py.test -m unit` or `py.test -m "unit and not selenium"`, the tests will fail because phantom_js is not installed.

We should either update the documentation or the script to install phantom_js while it's a dependency.

bryevdv:
- [x] issues: fixes #8571 

I think things were working on TravisCI by coincidence they they have a pre-installed PhantomJS

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -68,6 +68,7 @@ test:
     - nodejs >=8.8
     - numba
     - pandas
+    - phantomjs
     - pillow
     - pytest >=3.3.0
     - pytest-cov >=1.8.1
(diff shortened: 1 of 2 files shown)


bryevdv:
Well, `phantomjs` is not available on linux and our configured conda channels.

PR #8732, opened 2019-03-08
Add support for closable tabs

### LINKED ISSUE #7498: Bokeh Panel(closable=True) doesn't make 'x' appear on the Panel title and closable by user
Hi, 
I am running your example for Panel and Tabs in bokeh. The 'closable=True' not working as expected.
Tested on Ubuntu 16.04, Chrome and Firefox browsers.
Using Bokeh 0.12.13 server and Python 3.6. 
Nothing special showed in the debugging.

Thanks

```
from bokeh.models.widgets import Panel, Tabs
from bokeh.io import curdoc


output_file("slider.html")

p1 = figure(plot_width=300, plot_height=300)
p1.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], size=20, color="navy", alpha=0.5)
tab1 = Panel(child=p1, title="circle")

p2 = figure(plot_width=300, plot_height=300)
p2.line([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], line_width=3, color="navy", alpha=0.5)
tab2 = Panel(child=p2, title="line", closable=True)

tabs = Tabs(tabs=[ tab1, tab2 ])

curdoc().add_root(tabs)
```

mattpap:
image

fixes #7498

### DIFF
--- a/bokeh/models/layouts.py
+++ b/bokeh/models/layouts.py
@@ -412,6 +412,12 @@ class Panel(Model):
     The child widget. If you need more children, use a layout widget, e.g. a ``Column``.
     """)
 
+    closable = Bool(False, help="""
+    Whether this panel is closable or not. If True, an "x" button will appear.
+
+    Closing a panel is equivalent to removing it from its parent container (e.g. tabs).
+    """)
+
 class Tabs(LayoutDOM):
     ''' A panel widget with navigation tabs.
 
--- a/bokehjs/src/less/tabs.less
+++ b/bokehjs/src/less/tabs.less
@@ -66,6 +66,10 @@
       background-color: white;
       border-color: lighten(black, 90%);
     }
+
+    .bk-close {
+      margin-left: 10px;
+    }
   }
 
   &.bk-above .bk-tab {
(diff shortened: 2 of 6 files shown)


mattpap:
Currently if you remove a tab after scrolling headers, the view will jump to the beginning. I will fix this at some point under the umbrella of incremental DOM changes.

mattpap:
Or I just scroll active header into view, which would be useful in general.

bryevdv:
I don't have strong opinions about this feature @philippjfr @mattpap you two should decide when this has the most desirable behavior and is ready to merge.

bokeh/models/layouts.py
```diff
@@ -412,6 +412,10 @@ class Panel(Model):
     The child widget. If you need more children, use a layout widget, e.g. a ``Column``.
     """)
 
+    closable = Bool(False, help="""
+    Whether this panel is closeable or not. If True, an "x" button will appear.
+    """)
```
bryevdv: We should be explicit here that the value of `tabs` will be mutated, and the panel removed

philippjfr:
This is looking good to me, I do agree that the docstring and docs should be very explicit that closing a tab will modify the tabs property.

bryevdv:
@mattpap do you plan to add the docs, or do you want someone else to pick it up?

PR #8734, opened 2019-03-09
Make GMapPlotCanvas work in strict mode

### LINKED ISSUE #8733: GMap plots not working in the notebook and INLINE resources
The following code: 
```
from bokeh.models import GMapOptions
from bokeh.plotting import gmap

output_notebook()

map_options = GMapOptions(lat=30.2861, lng=-97.7394, map_type="roadmap", zoom=13)

# replace with your google api key
p = gmap("API-KEY", map_options)

show(p)
```
Generates no output, and the following console error:

> ReferenceError: Can't find variable: _bokeh_gmaps_callback

This is on master but evidently at least on 1.0.4 as well

mattpap:
fixes #8733

### DIFF
--- a/bokehjs/src/lib/models/plots/gmap_plot_canvas.ts
+++ b/bokehjs/src/lib/models/plots/gmap_plot_canvas.ts
@@ -4,12 +4,16 @@ import {Context2d} from "core/util/canvas"
 import {GMapPlot} from "./gmap_plot"
 import {PlotView, RangeInfo, FrameBox} from "./plot_canvas"
 
-declare let _bokeh_gmaps_callback: () => void
+declare global {
+  interface Window {
+    _bokeh_gmaps_callback: () => void
+  }
+}
 
 const gmaps_ready = new Signal0({}, "gmaps_ready")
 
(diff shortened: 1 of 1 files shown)


bokehjs/src/lib/models/plots/gmap_plot_canvas.ts
```diff
@@ -4,12 +4,16 @@ import {Context2d} from "core/util/canvas"
...
+  interface Window {
+    _bokeh_gmaps_callback: () => void
+  }
+}
 
 const gmaps_ready = new Signal0({}, "gmaps_ready")
 
 const load_google_api = function(api_key: string): void {
-  _bokeh_gmaps_callback = () => gmaps_ready.emit()
+  window._bokeh_gmaps_callback = () => gmaps_ready.emit()
```
mattpap: In strict mode naked (no `var`, `let`, `const` keyword) variables aren't allowed, resulting in `ReferenceError` exception. It's interesting why it started to fail just now, because this code was changed during transition to TypeScript which was like a year ago. Or maybe it was broken all along, just we didn't test this scenario.

PR #8736, opened 2019-03-10
Note for windows user for server --dev

### LINKED ISSUE #8529: Autoreloader cannot find bokeh on reload
I type command to serve my project:
`bokeh serve --show proj.py --dev proj.py`
By now everything works - I see correct results in my browser.

Later after changing and saving file proj.py bokeh tries to reload page and it crashes with message:

> can't open file 'D:\Anaconda\envs\pyt27\Scripts\bokeh': [Errno 2] No such file or directory

I have such file (precisely it's bokeh.exe seeing details of the file).
Also I have D:\Anaconda\envs\pyt27\Scripts in PATH environmental variable.
My environment:

bokeh: 1.0.2
python: 2.7.15
windows 7

xavArtley:
- [x] issues: fixes #8529
- [x] tests passed

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -521,15 +521,17 @@ class Serve(Subcommand):
             default = None,
             type    = str,
             nargs   = '*',
-            help    =   "Enable live reloading during app development."
-                        "By default it watches all *.py *.html *.css *.yaml files"
-                        "in the app directory tree. Additional files can be passed"
-                        "as arguments."
-                        "NOTE: This setting only works with a single app."
-                        "It also restricts the number of processes to 1.",
+            help    = "Enable live reloading during app development. "
+                      "By default it watches all *.py *.html *.css *.yaml files "
+                      "in the app directory tree. Additional files can be passed "
+                      "as arguments. "
+                      "NOTE: This setting only works with a single app. "
+                      "It also restricts the number of processes to 1. "
+                      "NOTE FOR WINDOWS USERS : this option must be invoked using "
+                      "'python -m bokeh'. If not Tornado will fail to restart the "
+                      "server",
         )),
     )
-
 
     def invoke(self, args):
         '''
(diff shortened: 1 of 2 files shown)


bokeh/command/bootstrap.py
```diff
@@ -96,6 +99,15 @@ def main(argv):
         prog=argv[0],
         epilog="See '<command> --help' to read about a specific subcommand.")
 
+    if sys.platform == 'win32':
+        """
+        windows hack
+        """
+        argv0_ext = os.path.splitext(argv[0])[-1]
+        if argv0_ext != '.py':
```
bryevdv: The comment should be a comment (rather than an string) and it would be good to reference the actual issue this clunky code is for.  But also this logic seems like it would not work well with directory style apps? I'm much less inclined to resort to a tacky solution if it only works part of the time, in that case I'd rather just document the necessary invocation for windows users and move on.

xavArtley: I'll change my pull request for a doc entry.
Actually it's a hack to work around a tornado issue
When invoking the autoreload with bokeh or bokeh.exe, on tornado side it tries to reconstruct cmd invocation but it ends by calling :
`python.exe -m bokeh.[exe] serve ... `
Which is not a valid command

xavArtley: Where should I put the doc?
Directly in the __init__.py of commands?
However it does not seems to be referenced by the .rst

bryevdv: @xavArtley I think a note should definitely go here:

https://github.com/bokeh/bokeh/blob/master/bokeh/command/subcommands/serve.py#L518

And, as long as you are touching that, can you fix up the broken missing spaces between sentences? I didn't realize those help strings used implicit string concat (I don't like that personally but if that's what will be used then spaces need to be inserted appropriately). The text in the corresponding test file will need to be updated as well. 

The docs there will automatically be added to the reference guide too, tho I don't know that those are used much. If you want to also add some narrative docs in User's Guide, then I think a new small section for "dev mode" would be appropriate here:

https://raw.githubusercontent.com/bokeh/bokeh/master/sphinx/source/docs/user_guide/server.rst

As an aside `server.rst` is getting big enough that we could probably consider splitting it up but that can happen separately.

PR #8738, opened 2019-03-10
kill node.exe and phantom phanomjs.exe before quitting selenium webdr…

### LINKED ISSUE #8737: Windows phantomjs not killed on selenium termination
I reinstalled a fresh python environment on windows with python 3.7 and pythonenv
I use only pip for package dependencies
When I tried to run tests some failed because temp files were locked.
win32error
test_failures

When driver terminate phantomjs is not correctly killed:
proc_pantomjs

xavArtley:
- [x] issues: fixes #8737

### DIFF
--- a/bokeh/io/webdriver.py
+++ b/bokeh/io/webdriver.py
@@ -28,10 +28,9 @@ import warnings
 from os.path import devnull
 
 # External imports
-import psutil
 
 # Bokeh imports
-from ..util.dependencies import import_required, detect_phantomjs
+from ..util.dependencies import import_required, detect_phantomjs, import_optional
 
 #-----------------------------------------------------------------------------
 # Globals and constants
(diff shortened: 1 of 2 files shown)


bryevdv:
@xavArtley we would need to add `psutil` as a dependency for this, so first we would need to make sure it is available both pip and conda installable on all platforms.  Or possibly we could protect the import inside `kill_proc_tree` and conditionally only add the dependency on windows (then still need to verify package availability on windows). Or, if there is some way to do this without `psutil` that's probably ideal.

xavArtley:
I'm not able to retrieve childs pid without `psutils`.
Seems a tricky problem on windows
I'll make a test on `psutils` in sys.modules

bryevdv:
OK, to be clear I am not opposed to just adding `psutils`, as long as it is easily obtainable. On the one hand we already have some optional dependencies for image export, so that'd be OK as well, we just need to document them and use `import_optional` appropriately

bryevdv:
Just for reference, Bokeh's current dependency list is fairly light by my standards, and `six` will disappear later this year

https://github.com/bokeh/bokeh/blob/7e2bff4a66b22ed4743edee365d4f9d7da46283a/setup.py#L89-L98

philippjfr:
I'd certainly prefer it if it was a optional dependency, ``psutils`` is easy to install on conda but it doesn't seem trivial to install when you're using pip, unless I'm mistaken.

bryevdv:
Though, I only see wheels on PyPI for windows:

https://pypi.org/project/psutil/#files

So I suppose optional dependency (i.e. use `import_optional` inside the function) may be the way to go

bryevdv:
@xavArtley this LGTM is it `ready`?

xavArtley:
yes

bryevdv:
Thanks!

PR #8739, opened 2019-03-11
Make warnings configurable

### LINKED ISSUE #8725: Don't warn on empty layouts
```
Python version      :  3.6.7 | packaged by conda-forge | (default, Nov 20 2018, 18:37:09)
IPython version     :  7.2.0
Tornado version     :  5.1.1
Bokeh version       :  1.1.0dev9+2.g1138433
BokehJS static path :  /Users/jsignell/conda/envs/bokeh-dev/lib/python3.6/site-packages/bokeh-1.1.0.dev9+2.g1138433-py3.6.egg/bokeh/server/static
node.js version     :  v11.6.0
npm version         :  6.5.0
```

I think that layouts should be allowed to be empty without raising a warning. There is a legitimate use of an empty layout object to create an empty area which can later be appended to or extended. This is especially useful in panel (see issue: https://github.com/pyviz/panel/issues/285) where the layout object will re-render if the children change. 

I understand that it could potentially be confusing if someone were to create an empty layout object and then think that it is just not rendering - but it doesn't seem like that would be terribly common (if it has been in the past, then maybe this warning shouldn't be removed).

```python

output_notebook()

col = column([])
show(col)
```
warning will be displayed

```python
# add a button
button = Button(label='Click me!')
col.children.append(button)
show(col)
```
column will render properly

This is the relevant warning:

https://github.com/bokeh/bokeh/blob/113843391ae24a685c12f04bc57cb2404df56155/bokeh/core/validation/warnings.py#L13-L14

This is the warning:

screen shot 2019-03-07 at 8 57 03 am

And this is the column rendering properly after adding an item to it:
screen shot 2019-03-07 at 8 58 31 am

jsignell:
Implementing configurable warning silencer as suggested by @bryevdv: https://github.com/bokeh/bokeh/issues/8725#issuecomment-470586707.

- [x] issues: fixes #8725
- [x] tests added / passed

## To try it

```python
import bokeh
from bokeh.layouts import row

output_notebook(verbose=True)
# create an empty layout object
empty_row = row([])

# turn the warning off
bokeh.core.validation.silence('EMPTY_LAYOUT')
# see that there is no warning
show(empty_row) 

# turn the warning back on
bokeh.core.validation.silence('EMPTY_LAYOUT', False)
# see that there is a warning
show(empty_row)
```

## Screenshot of use case
Screen Shot 2019-03-11 at 1 46 18 PM

### DIFF
--- a/bokeh/core/validation/check.py
+++ b/bokeh/core/validation/check.py
@@ -48,7 +48,7 @@ def silence(name, silence=True):
         silence (bool) : Whether or not to silence the warning
 
     Returns:
-        set conatining names of silenced warnings
+        A set containing the names of all silenced warnings
 
     This function adds or removes names from a set of silencers which
     is refered to when running ``check_integrity``. If a warning with a particular
--- a/bokeh/core/validation/check.py
+++ b/bokeh/core/validation/check.py
@@ -29,14 +29,49 @@ log = logging.getLogger(__name__)
 #-----------------------------------------------------------------------------
 # Globals and constants
 #-----------------------------------------------------------------------------
+__silencers__ = set()
 
 __all__ = (
     'check_integrity',
+    'silence'
 )
 
 #-----------------------------------------------------------------------------
 # General API
 #-----------------------------------------------------------------------------
+
+def silence(name, silence=True):
+    ''' Silence named warning on all Bokeh models.
+
+    Args:
+        name (str) : Names of warnings to silence
+        silence (bool) : Whether or not to silence the warning
+
+    Returns:
+        set conatining names of silenced warnings
+
+    This function adds or removes names from a set of silencers which
+    is refered to when running ``check_integrity``. If a warning with a particular
+    name is added to the silencers - then it will never be raised.
+
+    .. code-block:: python
+
+        >>> silence('EMPTY_LAYOUT', True)
+        {'EMPTY_LAYOUT'}
+
+        >>> empty_row = Row
+
+        >>> check_integrity([empty_row])
+
+    '''
+    if isinstance(name, int):
+        raise ValueError('Input to silence should be name of warning - not code')
+    if silence:
+        __silencers__.add(name)
+    else:
+        __silencers__.remove(name)
+    return __silencers__
+
 
 def check_integrity(models):
     ''' Apply validation and integrity checks to a collection of Bokeh models.
@@ -75,7 +110,9 @@ def check_integrity(models):
         log.error("E-%d (%s): %s: %s" % msg)
 
     for msg in sorted(messages['warning']):
-        log.warning("W-%d (%s): %s: %s" % msg)
+        code, name, desc, obj = msg
+        if name not in __silencers__:
+            log.warning("W-%d (%s): %s: %s" % msg)
 
     # This will be turned on in a future release
     # if len(messages['error']) or (len(messages['warning']) and settings.strict()):
--- a/bokeh/core/validation/check.py
+++ b/bokeh/core/validation/check.py
@@ -74,7 +74,7 @@ def silence(warning, silence=True):
                          '- not of type {}'.format(type(warning)))
     if silence:
         __silencers__.add(warning)
-    else:
+    elif warning in __silencers__:
         __silencers__.remove(warning)
     return __silencers__
 
--- a/bokeh/core/validation/check.py
+++ b/bokeh/core/validation/check.py
@@ -41,10 +41,10 @@ __all__ = (
 #-----------------------------------------------------------------------------
 
 def silence(warning, silence=True):
-    ''' Silence warning on all Bokeh models.
+    ''' Silence a particular warning on all Bokeh models.
 
     Args:
-        warning (Warning) : Warning to silence
+        warning (Warning) : Bokeh warning to silence
         silence (bool) : Whether or not to silence the warning
 
     Returns:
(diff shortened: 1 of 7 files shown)


jbednar:
Looks good to me!

bokeh/core/validation/check.py
```diff
@@ -29,15 +29,50 @@
...
 
+def silence(name, silence=True):
+    ''' Silence named warning on all Bokeh models.
+
+    Args:
+        name (str) : Names of warnings to silence
+        silence (bool) : Whether or not to silence the warning
+
+    Returns:
+        set conatining names of silenced warnings
```
philippjfr:
```suggestion
        A set containing the names of all silenced warnings
```

bryevdv:
@jsignell The warnings are available as actual module attrs, we should use those instead of raw strings:

```
In [1]: import bokeh.core.validation.warnings as m

In [2]: m.EMPTY_LAYOUT
Out[2]: 1002
```

jsignell:
Does this look right to you now @bryevdv - and if so do you think we need docs?

bryevdv:
I would be fine with just some docs in the `bokeh.validation` docstring: 

https://github.com/bokeh/bokeh/blob/master/bokeh/core/validation/__init__.py

@mattpap this seems like a useful thing in any case, and as-is preserves existing behavior, so I propose to merge and defer other questions about defaults until later.

jsignell:
Ok I just added some docs. Here they are rendered:

#### At the top of http://bokeh.pydata.org/en/latest/docs/reference/core/validation.html
Screen Shot 2019-03-12 at 10 23 08 AM

#### In http://bokeh.pydata.org/en/latest/docs/reference/core/validation.html#helper-functions
Screen Shot 2019-03-12 at 10 23 00 AM

jsignell:
Should I merge this? @bryevdv @mattpap

bokeh/core/validation/check.py
```diff
@@ -29,15 +29,56 @@
...
+        set()
+
+    '''
+    if not isinstance(warning, int):
+        raise ValueError('Input to silence should be a warning object '
+                         '- not of type {}'.format(type(warning)))
+    if silence:
+        __silencers__.add(warning)
+    else:
+        __silencers__.remove(warning)
```
bryevdv: It's harmless to add the same warning multiple times, should we catch key errors here and make repeated removals similarly harmless?

bryevdv:
@jsignell I had one late question

bokeh/core/validation/check.py
jsignell: I guess so.

bryevdv: I'm not saying it definitely should be, it only occurred to me that it might be reasonable. Do you have a preference?

bokeh/core/validation/check.py
```diff
@@ -29,15 +29,56 @@
...
+        >>> bokeh.core.validation.silence(EMPTY_LAYOUT, False)
+        set()
+
+    '''
+    if not isinstance(warning, int):
+        raise ValueError('Input to silence should be a warning object '
+                         '- not of type {}'.format(type(warning)))
+    if silence:
+        __silencers__.add(warning)
+    else:
```
jsignell:
```suggestion
    elif warning in __silencers__:
```

bokeh/core/validation/check.py
jsignell: No I think you are right. It is the kind of thing that might get set somewhere where the user doesn't control how many times it is called.

bryevdv: Ok, let's merge after that change then.

bokeh/core/validation/check.py
bryevdv: some day if we wanted to to be fancy we could collect all the defined error and warning values, then we could raise an error of the number is not a valid error or warning. But that can happen some other time.

jsignell:
I have some more tests to push

bryevdv:
@jsignell (cc @mattpap) a new version of `flake8` just landed in defaults and some codebase tests are failing now as a result:
```
E         ./bokeh/models/tests/test_callbacks.py:47:12: F632 use ==/!= to compare str, bytes, and int literals
E         ./bokeh/models/tests/test_callbacks.py:67:12: F632 use ==/!= to compare str, bytes, and int literals
E         ./bokeh/colors/util.py:77:12: F632 use ==/!= to compare str, bytes, and int literals
E         ./bokeh/tests/test_transform.py:104:16: F632 use ==/!= to compare str, bytes, and int literals
E         
```
Do you want to just fix this here? Tests will fail from now on until it is fixed

jsignell:
> Do you want to just fix this here? 

I can easily do it here if that is helpful

bryevdv:
I think that would be quickest/easiest if you don't mind

mattpap:
That's a very nice addition. Using `is` like this isn't the best idea (the only exception is with `None`).

jsignell:
I cancelled the outdated Travis builds, but can't see how to do the same for appveyor. I fixed the flake8 errors and am all done with this PR.

bryevdv:
I have cancelled all the extra jobs @jsignell if you would like to be added as an admin for appveyor, let me know (that is needed to cancel jobs)

jsignell:
@bryevdv sure - I'd like to be added as an appveyor admin. I often create more jobs than I mean to :)

bryevdv:
@jsignell invite is by email, can you email me from the account you'd like me to use?

PR #8740, opened 2019-03-11
Miscellaneous improvements to bokehjs

mattpap:
A few changes that accumulated recently in my git stash.


mattpap:
Looks like one or more of the upgraded dependencies brought a few transitive dependencies along, increasing the size of `package-lock.json` by 2600 lines. That's atrocious. As soon as I learn which one is it, I will let it go.

mattpap:
That was `mocha` effectively doubling our dependency footprint. I restored the previous version.

Issue #8724, opened 2019-03-07
LogColorMapper maps to wrong values


poplarShift:
Possibly related: #8061
Definitely related: #5020

TL;DR I'm happy to make a PR to use `log` for the `LogColorMapper` instead of `log1p` but wanted to hear opinions first.

As of now, LogTicker and LogTickFormatter seem to work with a simple log(x) transform, whereas LogColorMapper does a log(1+x) transform (`log1p`), which makes the colorbar map to the wrong values. The documentation does not mention the occurence of `log1p` at all.

So one of these two options has to be used for both cases.

Scientifically, `log1p` makes absolutely no sense because it makes the transform depend on the unit of the measurement, but in #5020 the reason given was that small values smaller than 10-ish are "usually below the range of values that log is used for", which I don't understand at all.

I get the concern that for 0-values, the log scale will blow up, but that's a problem of choosing the right transform for your data and should be handled by the user, not by bokeh. One could always add separate `Log1pColorMapper`, `Log1pTickFormatter`, etc. or something like it if users really need this behaviour for their dimensionless values.

philippjfr:
Unfortunately at this point I think maintaining backwards compatibility for `LogColorMapper` would be necessary, so I'd suggest that we could simply add an option to switch between ``log``, ``log1p`` and ``log10`` and make it very clear in the docs that the default is ``log1p``. The default could then be reconsidered for bokeh 2.0.

bryevdv:
> I get the concern that for 0-values, the log scale will blow up, but that's a problem of choosing the right transform for your data and should be handled by the user, not by bokeh.

This is easy to say when you aren't on the receiving end of support requests and bug reports. :) Users expect these things to "just work" and complain when they don't. Certainly other libraries, e.g. MPL make log scale usage "including zero" work in various ways, we just have not had the resources to devote to doing the same as of yet. 

> Unfortunately at this point I think maintaining backwards compatibility for LogColorMapper would be necessary,

I could be swayed that the current implementation is approximately useful/correct for "large N" but broken (i.e. in need of a *fix*) for "small N", and that therefore solution that works for both regimes by default is permissible.

philippjfr:
Fair enough, I agree that this could be considered broken particularly since none of the docs ever mentioned the fact that it was actually using log1p.

jbednar:
>  in #5020 the reason given was that small values smaller than 10-ish are "usually below the range of values that log is used for", which I don't understand at all.

[...]

PR #8743, opened 2019-03-14
Refactored continuous color mappers with .scan and .cmap

jlstevens:
This PR implements a refactor of continuous colormappers that I think makes the structure of the code clearer and will make it easier to add new colormappers to Bokeh (addressing https://github.com/pyviz/datashader/issues/609). I will submit another PR based off of this one shortly adding these new colormappers, starting with a cubic colormapper. Working on colormappers will also let me address https://github.com/bokeh/bokeh/issues/8724.

The core insight here is that that all continuous colormappers do at least two scans of the data 1) an initial scan to collect information about the data being colormapped - this can be as simple as finding the min/max of the data 2) a loop over the data that actually applies the colormapping based on the value being colormapped and the information collected in step 1.

These two steps are implemented by a `scan` method which does the initial collection of the information needed to do the colormapping and a `cmap` method which uses this information to colormap a specific value.

Other than making things clearer, one of the advantages of splitting the color mapping process into two methods is that you'll be able to implement new colormappers just by implementing a new `scan` method while inheriting a generic `cmap`method from a base class. This isn't demonstrated in this PR but will be demonstrated in the follow up PR that adds new colormappers.

This PR keeps the existing core machinery of both the linear and log color mappers intact. In the next PR it will be possible to use a more generic approach throughout.

### DIFF
--- a/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
@@ -46,7 +46,7 @@ export abstract class ContinuousColorMapper extends ColorMapper {
 
       if (isNaN(d))
         values[i] = nan_color
-      else 
+      else
         values[i] = this.cmap(d, palette, low_color, high_color, scan_data)
     }
   }
--- a/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
@@ -33,7 +33,7 @@ export abstract class ContinuousColorMapper extends ColorMapper {
     })
   }
 
-  protected abstract scan<T>(data: Arrayable<number>, palette: Arrayable<T>) : any
+  protected abstract scan<T>(data: Arrayable<number>, palette: Arrayable<T>): unknown
 
   protected _v_compute<T>(data: Arrayable<number>, values: Arrayable<T>,
     palette: Arrayable<T>, colors: {nan_color: T, low_color?: T, high_color?: T}): void {
--- a/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
@@ -45,10 +45,9 @@ export abstract class ContinuousColorMapper extends ColorMapper {
     for (let i = 0, end = data.length; i < end; i++) {
       const d = data[i]
 
-      if (isNaN(d)) {
+      if (isNaN(d))
         values[i] = nan_color
-        continue
-      }
+      else 
         values[i] = this.cmap(d, palette, low_color, high_color, scan_data)
     }
   }
--- a/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
@@ -33,6 +33,26 @@ export abstract class ContinuousColorMapper extends ColorMapper {
     })
   }
 
+  protected abstract scan<T>(data: Arrayable<number>, palette: Arrayable<T>) : any
+
+
+  protected _v_compute<T>(data: Arrayable<number>, values: Arrayable<T>,
+    palette: Arrayable<T>, colors: {nan_color: T, low_color?: T, high_color?: T}): void {
+
+    const {nan_color, low_color, high_color} = colors
+    const scan_data = this.scan(data, palette)
+
+    for (let i = 0, end = data.length; i < end; i++) {
+      const d = data[i]
+
+      if (isNaN(d)) {
+        values[i] = nan_color
+        continue
+      }
+        values[i] = this.cmap(d, palette, low_color, high_color, scan_data)
+    }
+  }
+
   protected _colors<T>(conv: (c: Color) => T): {nan_color: T, low_color?: T, high_color?: T} {
     return {
       ...super._colors(conv),
@@ -40,8 +60,7 @@ export abstract class ContinuousColorMapper extends ColorMapper {
       high_color: this.high_color != null ? conv(this.high_color) : undefined,
     }
   }
+  protected abstract cmap<T>(d : number,  palette: Arrayable<T>,  low_color: T, high_color: T, scan_data: any) : any
 
-  protected abstract _v_compute<T>(data: Arrayable<number>, values: Arrayable<T>,
-    palette: Arrayable<T>, colors: {nan_color: T, low_color?: T, high_color?: T}): void
 }
 ContinuousColorMapper.initClass()
--- a/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
@@ -35,7 +35,6 @@ export abstract class ContinuousColorMapper extends ColorMapper {
 
   protected abstract scan<T>(data: Arrayable<number>, palette: Arrayable<T>) : any
 
-
   protected _v_compute<T>(data: Arrayable<number>, values: Arrayable<T>,
     palette: Arrayable<T>, colors: {nan_color: T, low_color?: T, high_color?: T}): void {
 
--- a/bokehjs/src/lib/models/mappers/log_color_mapper.ts
+++ b/bokehjs/src/lib/models/mappers/log_color_mapper.ts
@@ -43,21 +43,18 @@ export class LogColorMapper extends ContinuousColorMapper {
 
   protected cmap<T>(d : number,  palette: Arrayable<T>,  low_color: T,
                     high_color: T, scan_data: LogScanData) : T {
-    // This handles the edge case where d == high, since the code below maps
-    // values exactly equal to high to palette.length, which is greater than
-    // max_key
     const max_key = palette.length - 1
 
     if (d > scan_data.high) {
       return high_color != null ? high_color : palette[max_key]
     }
-    if (d == scan_data.high) {
+    // This handles the edge case where d == high, since the code below maps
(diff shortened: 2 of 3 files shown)


bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
```diff
@@ -33,15 +33,34 @@ export abstract class ContinuousColorMapper extends ColorMapper {
...
+    const scan_data = this.scan(data, palette)
+
+    for (let i = 0, end = data.length; i < end; i++) {
+      const d = data[i]
+
+      if (isNaN(d)) {
+        values[i] = nan_color
+        continue
+      }
+        values[i] = this.cmap(d, palette, low_color, high_color, scan_data)
```
bryevdv: indentation here is off, but with only two 1-line branches I would prefer an `if-else` rather than the `continue`

bokehjs/src/lib/models/mappers/log_color_mapper.ts
```diff
@@ -26,53 +34,43 @@ export class LogColorMapper extends ContinuousColorMapper {
...
+    // This handles the edge case where d == high, since the code below maps
+    // values exactly equal to high to palette.length, which is greater than
+    // max_key
+    const max_key = palette.length - 1
 
-      values[i] = palette[key]
-    }
+    if (d > scan_data.high) {
+      return high_color != null ? high_color : palette[max_key]
+     }
```
bryevdv: indentation seems to be off starting here for the entire block below

bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
jlstevens: Addressed in e008d63

bokehjs/src/lib/models/mappers/log_color_mapper.ts
jlstevens: Fixed in 2565beb92. @bryevdv is there some auto-formatter program I can run to automatically fix this sort of formatting issue?

jbednar:
Just for some more motivation about why we're doing this: Currently if we want to use any of the more esoteric colormappers available in Datashader but not Bokeh, such as histogram equalization, we have to render the aggregated data array all the way into an RGBA image in datashader, which means giving up on hovering and colorbars in Bokeh. That's a major problem because it makes the interpretation of the visible images very difficult.  This refactoring will avoid that problem by making it possible to provide the aggregated array directly to BokehJS and letting BokehJS support histogram equalization (and any other nonlinear colormapping).

bryevdv:
> which means giving up on hovering and colorbars in Bokeh.

Well, hover capability should be added to `ImageRGBA`, which would also address your need. But this seems like useful work regardless.

@philippjfr please take the task to final review/merge this PR when you are satisfied

jbednar:
> hover capability should be added to ImageRGBA, which would also address your need

Wouldn't that only show the RGB values, not the underlying data values before colormapping (which is what is desired here)?

bokehjs/src/lib/models/mappers/log_color_mapper.ts
```diff
@@ -26,53 +32,43 @@ export class LogColorMapper extends ContinuousColorMapper {
...
     const scale = n / (log1p(high) - log1p(low))  // subtract the low offset
+
+    return {high, low, scale}
+  }
+
+  protected cmap<T>(d : number,  palette: Arrayable<T>,  low_color: T,
+                    high_color: T, scan_data: LogScanData) : T {
+    // This handles the edge case where d == high, since the code below maps
+    // values exactly equal to high to palette.length, which is greater than
+    // max_key
```
philippjfr: This comment in the wrong place now.

bokehjs/src/lib/models/mappers/log_color_mapper.ts
```diff
@@ -26,53 +32,43 @@ export class LogColorMapper extends ContinuousColorMapper {
...
+      return low_color != null ? low_color : palette[0]
+    }
+
+     // Get the key
+     const log = log1p(d) - log1p(scan_data.low)  // subtract the low offset
+     let key = Math.floor(log * scan_data.scale)
+
+     // Deal with upper bound
+     if (key > max_key) {
+       key = max_key }
```
philippjfr: Inconsistent bracketing here.

bokehjs/src/lib/models/mappers/log_color_mapper.ts
```diff
@@ -26,53 +32,43 @@ export class LogColorMapper extends ContinuousColorMapper {
...
-
-      // Deal with upper bound
-      if (key > max_key)
-        key = max_key
-
-      values[i] = palette[key]
+    if (d > scan_data.high) {
+      return high_color != null ? high_color : palette[max_key]
+    }
+    if (d == scan_data.high) {
```
philippjfr: Not a big deal but I'd either keep the spacing between these ``if`` branches consistent or merge them into ``else if``

bryevdv:
> Wouldn't that only show the RGB values, not the underlying data values before colormapping (which is what is desired here)?

You can always add a column(s) for any extra data you want just to drive hovers

bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
```diff
@@ -33,15 +33,32 @@ export abstract class ContinuousColorMapper extends ColorMapper {
     })
   }
 
+  protected abstract scan<T>(data: Arrayable<number>, palette: Arrayable<T>) : any
```
mattpap: Use `unknown` instead of `any`. The API will be similarly usable, but at least won't turn into a dynamically typed one.

philippjfr:
All very minor style related comments from me. It might also be nice to see separate tests for ``scan`` and ``cmap``. My TypeScript is not sufficiently advanced to suggest a better way of typing ``scan_data`` unfortunately.

bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
mattpap: Also, please fix whitespace (no space before colon in type position).

mattpap: Alternatively maybe a more specific type could be provided, if there is common ground between types used by more specific classes.

jbednar:
> You can always add a column(s) for any extra data you want just to drive hovers

Sure, but that means doubling the data sent the browser, which isn't good for datashader since it's sending images that are the full resolution of the plot, so it's not something we can enable by default.  So people have to choose between being fast or having hover, and even then still don't get a colorbar that they can use to make the plot publishable in static form.  Bokeh doing the colormapping addresses all those issues without increasing the data size at all.

bryevdv:
> All very minor style related comments from me. It might also be nice to see separate tests for scan and cmap. 

+1

bokehjs/src/lib/models/mappers/log_color_mapper.ts
jlstevens: Addressed in d2c9e2f5e

bokehjs/src/lib/models/mappers/log_color_mapper.ts
jlstevens: Addressed in d2c9e2f5e

bokehjs/src/lib/models/mappers/log_color_mapper.ts
jlstevens: Addressed in d2c9e2f5e

bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
jlstevens: Partially addressed in cb2acb9e2. I think the issue about a more specific type will be clearer in the more experimental PR I am going to branch off this one.

jlstevens:
I think a lot of the discussion in this PR will be clarified by the prototype PR I'll be pushing shortly.

bryevdv:
@jlstevens I'd like to move on these two PRs very soon. Can you remind what is left to be done?

jlstevens:
@bryevdv I am looking for feedback on the general approach in #8765 (certainly not *detailed* feedback at this point!). That will help inform some of the decision in this PR. After that, there are some decisions that need to be made about colorbars.

bryevdv:
@jlstevens I'm not entirely sure what specific questions you want addressed. IIUC there is still a `v_compute` in a base class that is the entry point for most outside code to use. Then in a subclass `BinnedColorMapper`, you propose to implement `v_compute` in terms of new methods `scan` and `cmap`, and then finally you propose that many existing mappers can become subclasses of the new `BinnedColorMapper`. Is that accurate? That all seems fine to me. All the mappers are still subclasses of `ContinuousColorMapper` and they all still expose `v_compute` so there is no disruption to outside use AFAICT.  

The only change I might suggest is that it be something like `BinningColorMapper` or `ScanningColorMapper` avoid a connotation that it operates on discrete/categorical data.

jlstevens:
> Is that accurate?

Yes.

> That all seems fine to me. 

Ok great! That establishes a baseline agreement before I present how I prototyped colorbar support. This bit is more tricky and I wanted to make sure the core changes are sound before tackling that (and I'll try to update the PR with that tonight). This also helps address @mattpap's concern regarding the return type of `scan`: it can be `unknown` for now but in future it could be a numeric array type if the existing colormappers are converted to `BinnedColorMapper` subclasses.

> The only change I might suggest is that it be something like BinningColorMapper or ScanningColorMapper avoid a connotation that it operates on discrete/categorical data.

Very happy to change the name. ``ScanningColorMapper`` sounds good to me.

mattpap:
Superseded by PR #8922.

Issue #8747, opened 2019-03-15
Custom extensions cannot import models/widgets/widget


philippjfr:
As part of panel we are shipping various custom models, generally that works perfectly but it seems when I export a plot containing custom models to a static HTML file it fails to resolve certain modules, specifically this line raises errors:

          var widget_1 = require("models/widgets/widget"); 

while other imports seem to work just fine, e.g.:

          var p = require("core/properties");

I'm really not quite sure how best to provide a minimal reproducible example, so [here](http://assets.holoviews.org/temp/Panel.html) is the exported HTML file, [here](https://github.com/pyviz/panel/blob/master/panel/models/audio.ts) is the custom model definition and here is the traceback:

```
Uncaught Error: Cannot find module 'models/widgets/widget'
    at s (Panel.html:93)
    at Object.custom/panel.models.widgets.audio (Panel.html:358)
    at s (Panel.html:93)
    at Object.custom/main (Panel.html:147)
    at s (Panel.html:93)
    at Object.t.Bokeh.r.register_plugin (Panel.html:93)
    at outer (Panel.html:137)
    at Panel.html:142
    at Panel.html:132
    at Panel.html:133
```

philippjfr:
This may or may not be related to this error I get when I try to import the Widget model like this: 

```
import {Widget} from "models/widgets"
```

```
Compilation failed:

../../../bokeh/bokeh/server/static/js/types/models/widgets/date_picker.d.ts:3:26 - error TS2307: Cannot find module 'pikaday'.

3 import * as Pikaday from "pikaday";
```

philippjfr:
cc @mattpap

mattpap:
> `import {Widget} from "models/widgets"`

This is an unfortunate limitation, you will have to use the canonical module path `models/widgets/index` for now. This should have been fixed awhile ago, because I've been fining my self tripping over this.

mattpap:
>  error TS2307: Cannot find module 'pikaday'.

This is a bug. All those external libraries shouldn't be part of the public API. I though I already fix all instances.

philippjfr:
>This is an unfortunate limitation, you will have to use the canonical module path models/widgets/index for now. This should have been fixed awhile ago, because I've been fining my self tripping over this.

Thanks, that makes sense, I think it's reasonable to have to import a specific module. I only tried this because I was hoping it would be able to resolve the original issue, but even ``models/widgets/index`` won't work due to the pikaday issue.

[...]

PR #8749, opened 2019-03-15
Selectively bundle custom models if possible

mattpap:
This is a crude but working change. In the long term this should be handled by resources (or resource bundles, yet to be implemented), but that's quite some work given the current state of affairs.

addresses #8747


mattpap:
@philippjfr, please test this.

philippjfr:
Working perfectly for me.

PR #8750, opened 2019-03-15
Allow to resolve both core/layout and core/layout/index

### LINKED ISSUE #8747: Custom extensions cannot import models/widgets/widget

mattpap:
fixes #8747

### DIFF
--- a/bokeh/util/compiler.py
+++ b/bokeh/util/compiler.py
@@ -597,7 +597,9 @@ def _bundle_models(custom_models):
                     extra_modules[sig] = True
                     modules.append((sig, code, deps_map))
             else:
-                raise RuntimeError("no such module: %s" % module)
+                index = module + ("" if module.endswith("/") else "/") + "index"
+                if index not in known_modules:
+                    raise RuntimeError("no such module: %s" % module)
 
         return resolved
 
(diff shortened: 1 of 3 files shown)


mattpap:
@philippjfr, another one to test.

philippjfr:
Seems to be working for me as well.

PR #8751, opened 2019-03-15
Update MAINTAINERS list

philippjfr:
Adding myself and updating @bryevdv's email


philippjfr:
@canavandl Just wanted to note that it's still listing your old email here. I'll leave it and let you decide whether you want it to be updated or removed.

Issue #4867, opened 2016-07-22
Don't make space for element if not visible


birdsarah:
While working on dask webui with @mrocklin, we wanted to squish two plots together by removing the xaxis. Setting `p.xaxis.visible = False` visually removed the xaxis but there was still a space for it (This is because the layout and the axis rendering happen separately).

I think the expected behavior is that if something is not there things squash up appropriately.

Would this also extend to DataRanges changing if a glyph became not visible?

birdsarah:
@pzwang's two cents on this issue was (paraphrasing): things should squish up by default, but only if there's an option to have things not squish when you don't want them to.

mattpap:
Fixed in PR #8754.

PR #8754, opened 2019-03-15
Make layout respect visible property

### LINKED ISSUE #6713: add visible property to models
Would it be possible to add a "visible" property to bokeh models ? 
When set to False It would just add 'display: none' to the style of the div containing the model; or even replace the style of the div with just 'display: none' which would also remove the space between that element and others divs.
When set to True the div containing the model would go back to its default style.
This would allow to quickly show/hide plots, widgets, or Row / Column layouts

mattpap:
fixes #6713, #4867

### DIFF
--- a/bokehjs/src/lib/core/dom.ts
+++ b/bokehjs/src/lib/core/dom.ts
@@ -327,14 +327,16 @@ export function unsized<T>(el: HTMLElement, fn: () => T): T {
 }
 
 export function sized<T>(el: HTMLElement, size: Partial<Size>, fn: () => T): T {
-  const {width, height, position} = el.style
+  const {width, height, position, display} = el.style
   el.style.position = "absolute"
+  el.style.display = ""
   el.style.width = size.width != null && size.width != Infinity ? `${size.width}px` : "auto"
   el.style.height = size.height != null && size.height != Infinity ? `${size.height}px` : "auto"
   try {
     return fn()
   } finally {
     el.style.position = position
+    el.style.display = display
     el.style.width = width
     el.style.height = height
   }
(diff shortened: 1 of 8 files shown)


bryevdv:
Setting visible to False initially works, setting it to True later in a `CustomJS` looks like this:

Screen Shot 2019-03-15 at 9 03 32 PM

MRE:

```
import numpy as np

from bokeh.models import CustomJS, Slider
from bokeh.plotting import figure, output_file, show, ColumnDataSource

x = np.linspace(0, 10, 500)
y = np.sin(x)

plot = figure(y_range=(-10, 10), plot_width=400, plot_height=400)

callback = CustomJS(args=dict(source=source), code="""
    var data = source.data;
    var A = amp.value;
    var k = freq.value;
    var phi = phase.value;
    var B = offset.value;
    var x = data['x']
    var y = data['y']
    for (var i = 0; i < x.length; i++) {
        y[i] = B + A*Math.sin(k*x[i]+phi);
    }
    source.change.emit();
    phase.visible = true
""")

amp_slider = Slider(start=0.1, end=10, value=1, step=.1,
                    title="Amplitude", callback=callback)
callback.args["amp"] = amp_slider

freq_slider = Slider(start=0.1, end=10, value=1, step=.1,
                     title="Frequency", callback=callback)
callback.args["freq"] = freq_slider

phase_slider = Slider(start=0, end=6.4, value=0, step=.1,
                      title="Phase", callback=callback)
callback.args["phase"] = phase_slider
phase_slider.visible=False

offset_slider = Slider(start=-5, end=5, value=0, step=.1,
                       title="Offset", callback=callback)
callback.args["offset"] = offset_slider

layout = row(
    plot,
    column(amp_slider, freq_slider, phase_slider, offset_slider),
)

output_file("slider.html", title="slider.py example")

show(layout)

```

bryevdv:
Confirmed working with latest push

PR #8757, opened 2019-03-17
Specialize visuals 'doit' getters

### LINKED ISSUE #8746: When line_width is set to 0 the glyph boundaries don't disappear

bokeh 1.0.4, python 3.6.6, notebook Kaggle, OS Windows 10, browser Chrome

I am trying to remove the glyph boundaries by using `line_width=0` but they don't disappear. As a temporary fix I use `line_alpha=0` instead.

```

output_file("patch.html")

p.patch([1, 2, 3, 4, 5], [6, 7, 8, 7, 3], fill_color='white', line_width=0)

show(p)
```

Linewidth bug

bryevdv:
- [x] issues: fixes #8746
- [x] tests added / passed

This PR generalizes the notion of whether a "visual" should be drawn or not. Currentlty only a single attribute (typically color, e.g. `fill_color`) is checked against `null` to determine whether canvas operations should be skipped or not. This removes the generic code that checks the single configured attribute, with an abstract `doit` property getter on the base, that is specialized for each visual subclass. New behavior is summarized below:

* fill: skip if color is null, or alpha is 0
* line: skip if color is null, or alpha or width is 0
* text: skip if color is null, or alpha is 0

The new code is more explicit and seems to be an improvement from a maintenance POV as well. 

Also it was noticed that the visual classes have almost no tests, despite being very low level and in use in many places. I have added tests to cover the new functionality, as well as some other basic tests.

New MRE output with this PR:

Screen Shot 2019-03-16 at 8 02 42 PM

### DIFF
--- a/bokehjs/src/lib/core/visuals.ts
+++ b/bokehjs/src/lib/core/visuals.ts
@@ -12,18 +12,15 @@ export abstract class ContextProperties {
 
   // prototype {
   attrs: string[]
-  do_attr: string
   // }
 
   readonly cache: {[key: string]: any} = {}
-  readonly doit: boolean
+
+  abstract get doit(): boolean
 
   all_indices: number[]
 
   constructor(readonly obj: HasProps, readonly prefix: string = "") {
-    const do_spec = obj.properties[prefix + this.do_attr].spec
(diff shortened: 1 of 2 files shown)


bryevdv:
There are more tests that could be added but this is an improvement for now.

PR #8758, opened 2019-03-17
Added workaround for amd load in notebook

### LINKED ISSUE #5227: External resources belonging to custom models fail to load inside of Jupyter/IPython notebooks

philippjfr:
This is hacky but very simple fix for https://github.com/bokeh/bokeh/issues/5227. It works well both in the classic notebook and when exporting a notebook using nbconvert. Setting ``window.define = undefined`` is definitely a hack but it is considerably simpler than it would be for us to leverage ``requirejs`` correctly in the notebook and works in all scenarios I have tested. That said I will defer to @mattpap's judgement on this since this does have the potential to break code that does rely on define being defined.

- [x] issues: fixes #5227
- [ ] tests added / passed

### DIFF
--- a/bokeh/core/_templates/autoload_js.js
+++ b/bokeh/core/_templates/autoload_js.js
@@ -61,16 +61,24 @@ calls it with the rendered model.
     console.debug("Bokeh: BokehJS not loaded, scheduling load and callback at", now());
     root._bokeh_is_loading = css_urls.length + js_urls.length;
 
+    var _define = window.define;
+
     function on_load() {
       root._bokeh_is_loading--;
       if (root._bokeh_is_loading === 0) {
         console.debug("Bokeh: all BokehJS libraries/stylesheets loaded");
+        if (window.requirejs) {
+          window.define = _define;
+        }
         run_callbacks()
       }
     }
 
     function on_error() {
       console.error("failed to load " + url);
+      if (window.requirejs) {
+        window.define = _define;
+      }
     }
 
     for (var i = 0; i < css_urls.length; i++) {
@@ -85,6 +93,9 @@ calls it with the rendered model.
       document.body.appendChild(element);
     }
 
+    if (window.requirejs && js_urls.length) {
+      window.define = undefined
+    }
     for (var i = 0; i < js_urls.length; i++) {
       var url = js_urls[i];
       var element = document.createElement('script');


philippjfr:
I'm slightly disinclined to merge this since it can interfere with the notebooks load, a slightly better but still hacky approach might be to temporarily set window define like this:

```javascript
function _define_stub(name, deps, callback) {
    window._define(name, deps, callback)
 }
 window.define = _define_stub;
``` 

This ensures UMD loaders do not find ``window.define.amd`` and therefore execute as normal but does allow the notebook's webpack bundle to be loaded since that does not check for ``window.define.amd``.

mattpap:
If anything, I would proceed with `_define_stub`, assuming the logic behind it holds. However, whatever happens, any changes should be made to autoload_nb_js, not autoload_js, because we don't want to modify thirdparty environments even if external libraries break.

philippjfr:
>If anything, I would proceed with _define_stub, assuming the logic behind it holds.

By putting a console.log inside the define function I can confirm that everything works correctly for the notebook js bundle, but I obviously can't check or guarantee compatibility with every library or notebook extension a user might be loading. I do know that all UMD modules should have a section like this at the top ``if (typeof define === "function" &&  define.amd) { define([], e) }`` which is what is causing this issue in the first place.

So it's currently not clear to me how sure we can be that this is actually working as desired in all cases. Maybe the best way forward is for me to trial the code inside panel for a release and if that works out well we can decide to add it to bokeh itself.  

> However, whatever happens, any changes should be made to autoload_nb_js, not autoload_js, because we don't want to modify thirdparty environments even if external libraries break.

That's a good point, I'll probably have to add a few more templating blocks but will make that change.

philippjfr:
Instead of the stub it's probably cleaner to temporarily override ``window.define.amd``:

```javascript
window._define_amd = window.define.amd
window.define.amd = undefined;
```

philippjfr:
The only alternative I can see here is to allow defining requirejs config on the custom model, here is what that might look like:

```
    __js_require__ = {
        'baseUrl': 'https://cdnjs.cloudflare.com/ajax/libs/',
        'paths': {
            "vega-embed":  "vega-embed/3.30.0/vega-embed.min",
            "vega-lite": "vega-lite/2.6.0/vega-lite.min",
            "vg": "vega/5.2.0/vega.min"
        },
        'exports': {'vega-embed': 'vegaEmbed', 'vega-lib': 'vega', 'vega-lite': 'vl'}
    }
```

This would then get translated into the following code:

```
if (window.requirejs) {
  require.config({"baseUrl": "https://cdnjs.cloudflare.com/ajax/libs/", "paths": {"vega-embed": "vega-embed/3.30.0/vega-embed.min", "vega-lite": "vega-lite/2.6.0/vega-lite.min", "vega-lib": "vega/5.2.0/vega.min"}});;
  require(["vega-embed", "vega-lite", "vega-lib"], function(vegaEmbed, vl, vega) {
    window.vegaEmbed = vegaEmbed;
    window.vl = vl;
    window.vega = vega;
    run_callbacks()
  })
}
```

Basically it would use requirejs if it's available and fall back to the current code path otherwise (e.g. for JupyterLab). I've just tested this and it works well. It's a bit annoying to have to specify such a spec separately but at least it's not a hack.

philippjfr:
I'm going to close this PR while I rethink the approach, for the time being I may play around with different approaches in panel before I commit to any specific approach to contribute to bokeh.

PR #8760, opened 2019-03-17
Allocate fair amount of space for tabs' headers

### LINKED ISSUE #8726: Tabs width set on width of contents - truncating tabs
```
Python version      :  3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
IPython version     :  7.1.1
Tornado version     :  5.1.1
Bokeh version       :  1.1.0dev8
BokehJS static path :  /Users/jsignell/conda/envs/panel-dev/lib/python3.6/site-packages/bokeh/server/static
node.js version     :  v7.10.0
npm version         :  4.2.0
```
The width of the tabs seems to be set based on the width of the item inside the tab without taking the width of the title into account. The result is a truncated tabs object where you can't click on the second tab ( because it isn't visible). 

I originally raised this as https://github.com/pyviz/panel/issues/284

```
from bokeh.models.widgets import Panel, Tabs, Button

output_notebook()

tab1 = Panel(child=Button(label='+', width=30), title="This is a long title")
tab2 = Panel(child=Button(label='-', width=30), title="This is another long title")
tabs = Tabs(tabs=[ tab1, tab2 ])

show(tabs)

```
screen shot 2019-03-07 at 9 20 29 am

mattpap:
I think this should be reasonable middle ground, at least for now. Tabs will have space allocated for at least three headers (I suppose this could be made configurable) and scroll buttons. As an additional change, I made setting `width` and `height` to prefer `fixed` sizing policy if the policy is `auto`. Previously the go-to policy was `fit` for non-plot components, which resulted in confusing behavior (which would be visible in the MRE of #8726 after tabs' fix).

fixes #8726


bryevdv:
checked MRE and a selection of updated baseline examples

Issue #4847, opened 2016-07-20
For bokeh.models.glyphs.Image, images are colormapped individually and not as a set


canavandl:
Referring to this chunk of code: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/models/glyphs/image.coffee#L23-L35

Currently, the only glyph that uses a ColorMapper is the `Image` glyph. In the implementation of `Image`, each image 2d array is individually color mapped instead of the set of images (as `Image` is vectorized).

This seems incorrect, as a visualization that consists of a bunch of tiled images would non-intuitively be colormapped by the min/min of each individual tile.

bryevdv:
Closing in favor of #12355

PR #8765, opened 2019-03-18
Added EqHistColorMapper

jlstevens:
This PR is still very WIP but it demonstrates a prototype of what can be achieved using the refactor in https://github.com/bokeh/bokeh/pull/8743 with the introduction of the `scan` and `cmap` methods. As in https://github.com/bokeh/bokeh/pull/8743, this will help address the issue described in https://github.com/pyviz/datashader/issues/609.

Here is a demonstration:

image

This is the simple test image with an outlier added which helps illustrates the difference (linear on the left and eqhist on the right).

At this stage I am *not* looking for a detailed review, but I would like to discuss the following:

1. `BinnedColorMapper` is a very general concept and could be used to implement *all* continuous colormappers including linear, log, cubic etc. Other than in the simplest linear case, this is probably a more efficient approach in general as you only need to compute 257 bin edges for a typical palette and most images will have a lot more pixels than that. E.g a linear colormapper can easily be implemented with `scan` returning `linspace` from low to high as long as the palette+1.

2. If we switch to using `BinnedColorMapper` throughout, the return type of `scan` can always be an array of bin edges (i.e numbers) addressing an issue raised in https://github.com/bokeh/bokeh/pull/8743 . It would also be an opportunity to address https://github.com/bokeh/bokeh/issues/8724 in the log colormapper.

3. `EqHistColorMapper` is an example of a non-trivial `BinnedColorMapper` which unlike the linear, log and cubic cases is very hard to implement without the binning approach. The code I used for reference is implemented in [datashader](https://github.com/pyviz/datashader/blob/master/datashader/transfer_functions.py#L123).

Note that if we liked the `BinnedColorMapper` approach in general, I would update the implementations of the linear and log mappers (and maybe add a cubic mapper) in a separate PR.

## TODO

- [x] Make the bin number a property (currently hard coded to 256x256).
- [ ] Handle explicit low and high values.
- [ ] Colorbar support. I have a working prototype that I can demonstrate but some discussion will be needed to find the best approach.
- [ ] Unit tests - make sure low and high colors are handled properly etc.
- [ ] Docstrings and docs


jlstevens:
In 7658f3796 I renamed `BinnedColorMapper` to `ScanningColorMapper` as suggested by @bryevdv in #8743.

Commit 9f514703b shows a very rough cut at colorbar support that at least shows some labelled ticks on the screen (linear on the left and eq_hist on the right):

image

Obviously there are a lot of problems with this right now so I want to point out what I think are the fundamental issues before drilling down into more minor details:

- The basic problem as far as I can tell, is that colorbars till now have not had access to any information about the data being colormapped other than 1) the high value (explicitly supplied) 2) the low value (explicitly supplied) 3) the simple monotonic function used (i.e linear, log). To support the `ScanningColorMapper` classes, the data needs to be able to affect the colorbar ticking and arguably it would be useful for the high and low values to be derived from the data even in the simple cases of linear and log colormapping.

- As `color_bar.ts` has access to `this.model.color_mapper` when generating an appropriate `Scale`, I have *assumed* that it is safe for me to access the data generated by `scan` here. This *seems* to work as `scan` step for doing the colormapping does appear to run before the colorbar is generated. In this prototype, `scan_result` is the result of the `scan` step and is declared as state owned by the `ScanningColorMapper` classes.

- To work with the rest of how colorbars are generated, I've added a `LerpScale` model that does linear interpolation on `scan_result` which defines the relationship between color and data space (with one data value for each entry in the palette). At this stage only `v_compute` is needed and I've added the simplest (wrong!) implementations of  `invert`, `v_invert` and `_compute_state` for now, just to satisfy the `Scale` interface.

- Although there are lots of issues to work out, I think something like `LerpScale` could be used to generate labels for all `ScanningColorMapper` classes. Right now `LerpScale` is used in `color_bar.ts` for `case "EqHistColorMapper"` but this should generalizable to all `ScanningColorMapper`s.

I'm hoping 9f514703b shows the outline of the problem here but also hints that it is solvable for `ScanningColorMapper`s. Note that even if we decide to make all mappers subclasses of `ScanningColorMapper`, we can still use `LinearScale` and `LogScale` to leave the color bar behavior unchanged in those cases.

At this point, I would like to ask @bryevdv and @mattpap whether this approach is worth pursuing further and whether they may have some better ideas as to how colorbars could be made to reflect the distribution of the data being colormapped.

bokehjs/src/lib/core/util/arrayable.ts
```diff
@@ -201,3 +201,52 @@ export function sorted_index<T>(array: Arrayable<T>, value: T): number {
...
+    , leftEdgeIndex = 0
+    , rightEdgeIndex = intervals.length - 1
+  while (rightEdgeIndex - leftEdgeIndex !== 1) {
+    indexOfNumberToCompare = leftEdgeIndex + Math.floor((rightEdgeIndex - leftEdgeIndex)/2)
+    point >= intervals[indexOfNumberToCompare]
+      ? leftEdgeIndex = indexOfNumberToCompare
+      : rightEdgeIndex = indexOfNumberToCompare
+  }
+  return leftEdgeIndex
+}
```
bryevdv: A few comments: 

* "lerp" is a terrible name :) We should be explicit e.g `linear_interpolate` and also there is at least one other place that does linear interpolation and we should share code between: https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/transforms/linear_interpolator.ts#L47

* I don't think this is the right module for these. Maybe the math module, maybe a new util module, open to suggestion

* Simple functions like these should get tests

bokeh/models/scales.py
```diff
@@ -85,6 +87,12 @@ class CategoricalScale(LinearScale):
     '''
     pass
 
+class LerpScale(Scale):
```
bryevdv: `LinearInterpolationScale`

bokehjs/src/lib/models/annotations/color_bar.ts
```diff
@@ -473,11 +474,15 @@ export class ColorBarView extends AnnotationView {
...
       }),
+        scan_result: this.model.color_mapper.scan_result,
     }
 
     switch (this.model.color_mapper.type) {
       case "LinearColorMapper": return new LinearScale(ranges)
       case "LogColorMapper":    return new LogScale(ranges)
+      case "EqHistColorMapper": {
+        return new LerpScale(ranges)
+      }
```
bryevdv: `scan_result` is not a range, so it shouldn't go in an object called "ranges" Since it is specific to the linear interp scale, I would just pass it as an additional parameter after ranges.

bokehjs/src/lib/models/mappers/binned_color_mapper.ts
```diff
@@ -0,0 +1,46 @@
...
+        }
+        if (d > edges[edges.length-1]) {
+            return high_color
+        }
+        for (let i = 0, end = (edges.length-1); i < end; i++) {
+            const low_edge =  edges[i]
+            const high_edge =  edges[i+1]
+            key = i
+            if (d < high_edge && d >=low_edge) {
+                break
```
bryevdv: just `return palette[i]` here and avoid the `let` var up top

bokehjs/src/lib/models/mappers/binned_color_mapper.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(attrs?: Partial<ScanningColorMapper.Attrs>) {
+    super(attrs)
+  }
+
+  static initClass(): void {
+    this.prototype.type = "ScanningColorMapper"
+  }
+
+    protected cmap<T>(d : number,  palette: Arrayable<T>,  low_color: T,
+                      high_color: T, edges: any) : any {
```
bryevdv: think the return type can be tightened up

bokehjs/src/lib/models/mappers/eqhist_color_mapper.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export namespace EqHistColorMapper {
+  export type Attrs = p.AttrsOf<Props>
+
+  export type Props = ScanningColorMapper.Props & {
+      bins: p.Property<number>
+    }
+}
+
+export interface EqHistColorMapper extends EqHistColorMapper.Attrs {
```
bryevdv: I think we should be explicit here too, esp since these are most likely not going ot be used by average users. `EqualHistogramColorMapper`

bokehjs/src/lib/models/scales/lerp_scale.ts
```diff
@@ -0,0 +1,60 @@
...
+    return xprime
+  }
+
+  v_invert(xprimes: Arrayable<number>): Arrayable<number> {
+    return xprimes
+  }
+
+  _compute_state(): [number, number] {
+    return [0, 0]
+  }
```
bryevdv: Is this needed?

bokehjs/src/lib/models/annotations/color_bar.ts
bryevdv: Also would prefer identical formatting to the two lines above

bokehjs/src/lib/models/mappers/continuous_color_mapper.ts
```diff
@@ -13,7 +13,9 @@ export namespace ContinuousColorMapper {
   }
 }
 
-export interface ContinuousColorMapper extends ContinuousColorMapper.Attrs {}
+export interface ContinuousColorMapper extends ContinuousColorMapper.Attrs {
+  scan_result: Arrayable<number>
```
bryevdv: This should go on `ScanningColorMapper`, not `ContinuousColorMapper` ?

bokehjs/src/lib/models/mappers/eqhist_color_mapper.ts
```diff
@@ -0,0 +1,63 @@
...
+    scan_result: Arrayable<number>
```
bryevdv: This is redundant since its on the base class?

bryevdv:
@jlstevens I've left some preliminary comments. It occurs to me, since you are proposing having more rich data about the data inform colormapping, that you might want to be aware of this issue:

https://github.com/bokeh/bokeh/issues/4847

bryevdv:
@jlstevens checking in for plans/status

mattpap:
Superseded by PR #8922.

PR #8773, opened 2019-03-20
compute image bounds correctly

### LINKED ISSUE #8668: Displaying custom data on hover for RGBA image
It would be nice to be possible to show some sort of information on RGBA images using Bokeh (https://stackoverflow.com/questions/54753032/hover-image-rgba-bokeh/54753614#54753614).
For instance, I'm working on a project in which we provide an RGBA image where the alpha represents some sort of confidence. We would like, when hovering an image pixel, to be able to show this confidence (in this case, the `rgba[:,:,3]` value).

### LINKED ISSUE #8770: Using range padding with image plots leave lots of empty space

Bokeh: 1.0.4
Python: 3.6.8
Conda: 4.6.8  (used a conda environment here)
Browser: Firefox 65.0.1

When using a value for the `range_padding` attribute of `DataRange1d`,
I expect that this padding is applied on the left side of the range (lower values)
as well as on the right side (higher values).

Instead, the padding seems to be only applied for lower values.
See picture below.

```
from bokeh.models.ranges import DataRange1d

data = [[1,2,3],[4,5,6]
x_range = DataRange1d(range_padding=0.01)
y_range = DataRange1d(range_padding=0.01)

plot = figure(x_range=x_range, y_range=y_range)
plot.image([arr], x=0, y=0, dw=3, dh=2)
show(plot)
```

A warning on the console, in German, which roughly translates to:

"mozImageSmoothingEnabled should not be used any more.
Instead, please use the property imageSmoothingEnabled without prefix."

bokeh_plot(8)

The `x` axis ranges up to 6 and the `y` axis up to 4, twice the values that I would expect.

bryevdv:
- [x] issues: fixes #8770 fixes #8668
- [x] tests added / passed

As far as I can tell, this has been broken forever, possibly since the image glyphs were created. This PR removes the code that modifies the default bounds based on the spatial index, since these are already correct as-is. 

It might be nice to add some image or integration tests to verify this change. We could also update some of the gallery examples to not set ranges explicitly (but they would need to set `range_padding`  to 0 instead). 

Finally, I don't think things will behave well with `dw_units` or `dh_units` set to "screen" but things are clearly currently already broken in that case, by inspection. I would support deprecation being able to set units on these for 2.0 cc @bokeh/dev

### DIFF
--- a/bokehjs/src/lib/core/util/templating.ts
+++ b/bokehjs/src/lib/core/util/templating.ts
@@ -7,7 +7,7 @@ import {escape} from "./string"
 import {isNumber, isString, isArray, isTypedArray} from "./types"
 
 import {ColumnarDataSource} from "models/sources/columnar_data_source"
-import {ImageIndex} from "../../models/glyphs/image"
+import {ImageIndex} from "../../models/glyphs/image_base"
 import {CustomJSHover} from 'models/tools/inspectors/customjs_hover'
 
 export function sprintf(format: string, ...args: unknown[]): string {
--- a/bokehjs/src/lib/models/glyphs/image.ts
+++ b/bokehjs/src/lib/models/glyphs/image.ts
@@ -1,4 +1,4 @@
-import {XYGlyph, XYGlyphView, XYGlyphData} from "./xy_glyph"
+import {ImageBase, ImageBaseView, ImageDataBase} from "./image_base"
 import {ColorMapper} from "../mappers/color_mapper"
 import {LinearColorMapper} from "../mappers/linear_color_mapper"
 import {Class} from "core/class"
@@ -6,26 +6,12 @@ import {Arrayable} from "core/types"
 import * as p from "core/properties"
 import {concat} from "core/util/array"
 import {Context2d} from "core/util/canvas"
-import {SpatialIndex} from "core/util/spatial"
 import * as hittest from "core/hittest"
 import {Selection} from "../selections/selection"
 import {PointGeometry} from "core/geometry"
+import { ImageRGBAData } from './image_rgba'
 
-// XXX: because ImageData is a global
-export interface _ImageData extends XYGlyphData {
-  image_data: Arrayable<HTMLCanvasElement>
-
-  _image: Arrayable<Arrayable<number> | number[][]>
-  _dw: Arrayable<number>
-  _dh: Arrayable<number>
-
-  _image_shape?: Arrayable<[number, number]>
-
-  sw: Arrayable<number>
-  sh: Arrayable<number>
-}
-
-export interface ImageView extends _ImageData {}
+export interface ImageView extends ImageDataBase {}
 
 export interface ImageIndex {
   index: number
@@ -34,7 +20,7 @@ export interface ImageIndex {
   flat_index: number
 }
 
-export class ImageView extends XYGlyphView {
+export class ImageView extends ImageBaseView {
   model: Image
   visuals: Image.Visuals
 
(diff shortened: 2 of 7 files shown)


bryevdv:
OK this expanded slightly. I decided that updating the examples would nicely exercise the change in image diff tests, and discovered that things were not working for `ImageRGBA` because it did not yet do spatial indexing. 

https://github.com/bokeh/bokeh/pull/8773/commits/3e0593819b5ae5406890133406993a1d7e98407a splits off `ImageBase`, etc. which performs spatial indexing and also allows `Image` and `ImageRGBA` to duplicate considerably less code. 

@jlstevens as a side effect this means it is now even closer/simpler to implement hover for `ImageRGBA`

@philippjfr @jlstevens it would be helpful if you could test this branch specifically with HV/DS etc. 

@mattpap would appreciate your review of TS class changes.

bryevdv:
Ok latest push moves hit testing machinery to the base and now hover "just works" for image rgba:

Screen Shot 2019-03-20 at 5 32 52 PM

We talked about maybe adding some new auxiliary fields like `$red` to extra the individiual RGBA channels. That seems like a great idea but I will leave it up to others to do if they are interested.

bryevdv:
Actually rather than further complicating the hover templating codepath, I think I would do this with a new configurable formatter for RGBA int values. For now this could be done already by users with `CustomJSHover`.

mattpap:
> would appreciate your review of TS class changes.

Looks good after a cursory review.

philippjfr:
@bryevdv Could you explain what exactly this PR actually does? It's hard to tell from the description and the code, does it only touch the auto-ranging? If so then holoviews will not likely be affected since it never uses auto-ranging. I think removing support for screen units seems fine, especially since it has always been broken.

This issue is probably unrelated but one thing that has always confused me about the image glyph is its handling of inverted ranges. When inverting the range you currently have to adjust the x/y properties of the image when I'd expect it to correctly infer that the image should be drawn in the other direction. The problem with fixing that is that it would probably be backwards incompatible. I also just realized that negative ``dh`` and ``dw`` values are not allowed:

>ValueError: Distances must be positive or None!

But if you supply them as part of a CDS and wrap them in a list this is not enforced. So holoviews has been handling inverted axes all wrong for a long time.

bryevdv:
@philippjfr The fix for bounds computation depends on the spatial index, which was only previously added to `Image` but not `ImageRGBA`. This PR refactors common code, including the spatial indexing code, to a base class for `Image` and `ImageRGBA`. As a side effect, hit-testing/hover now also works for `ImageRGBA` 

The other thing you describe is definitely a different, separate issue.

jlstevens:
@bryevdv New RGB hover looks great, thanks!

jmontoyam:
> Actually rather than further complicating the hover templating codepath, I think I would do this with a new configurable formatter for RGBA int values. For now this could be done already by users with `CustomJSHover`.

@bryevdv  I am new to bokeh, I have been taking a look to your suggestion regarding using `CustomJSHover` to format the default 32bit RGBA value and show it as a tuple containing (R, G, B, A), but unfortunately I have not been able to find how can I get the 32bit RGBA int value within javascript. The example provided in the docs shows how to access x and y from javascript using special_vars.x and special_vars.y but it does not mention how to access the default 32bit RGBA from javascript. Could you please give me some suggestions?...Thanks!

bryevdv:
@jmontoyam for support/usage questions please use our Project Discourse: https://discourse.bokeh.org/

jmontoyam:
Ok, please excuse me, I will go there ;)

PR #8775, opened 2019-03-20
Add support for grid layout spans

### LINKED ISSUE #8710: Add support for spans do grid layout
This is a follow up on issue #8084. Technically speaking this is a regression since the old layout. This is needed to restore functionality of `examples/app/spectrogram`, etc.

mattpap:
image

fixes #8710


mattpap:
I consider this ready, though I will follow up with a PR tomorrow to improve the API (`GridBox.children`) and to test this some more.

bryevdv:
@mattpap  thanks, I'll check/test it tonight. Just for reference, do you expect the original version 

```
plots = column(waterfall_plot, row(column(signal_plot, spectrum_plot), eq))
``` 
to be workable again at some point in the near future?

mattpap:
Not directly, because a nested row/column layout is not a grid. The old layout didn't have this distinction and everything effectively was one big grid (by flatting nested rows and columns). What I intend to do is to allow users to convert between those representations, so the original version of this example would only need addition of `plot.to_grid()` (or `plots.flatten()`, etc.) to get grid axis alignment from a nested layout.

bryevdv:
@mattpap what if we implicitly flatten all roots now, and plan to stop doing that for 2.0 ?

mattpap:
Not my first choice, but I suppose it would add to backwards compatibility. However, this would also remove a feature, so we would need to allow the non-implicit grid behavior grid somehow. I suppose I would add `flatten` option (default `True`) to box layouts. In any case, let's cross that bridge when I actually implement the flattening algorithm.

bryevdv:
> In any case, let's cross that bridge when I actually implement the flattening algorithm.

Sure, are you saying you want this merged as-is now?

mattpap:
Yes, especially given the extent of this change (it went actually much smoother so far than I expected). I will followup with a PR with other changes.

bryevdv:
OK I am fairly confident to merge this given how small the changes are outside `grid.ts` and the lack of changes to the extensive baselines. So I will merge this and start a dev build. @philippjfr @jlstevens I would highly recommend checking this out ASAP just in case there are unknown issues that might affect your HV integration

PR #8776, opened 2019-03-21
Apply bokeh CSS more selectively to avoid conflicts

### LINKED ISSUE #8680: Custom Widget in doc strange display in 1.1.0dev6
I was looking at custom examples in doc for version 1.1.0
https://bokeh.pydata.org/en/dev/docs/user_guide/extensions_gallery/

and I noticed some difference in the Custom Widget example : 
 - 1.0.4 :
 
image

 - 1.1.0dev6 :

image

mattpap:
image

fixes #8680


mattpap:
@philippjfr, if you have a chance, please test this with your code.

philippjfr:
@mattpap I can confirm that HTML inside ``Div`` models now renders as expected, e.g. here's a list:

Screen Shot 2019-03-30 at 6 54 28 PM

mattpap:
@philippjfr, OK. Finally a solution that universally works, though not ideal. CSS' specificity rules (among other things) are just beyond reason.

philippjfr:
Great, there's only so much you can do, and this seems eminently reasonable.

PR #8781, opened 2019-03-23
correcting validation for BAD_EXTRA_RANGE_NAME (#8723)

### LINKED ISSUE #8723: BAD_EXTRA_RANGE_NAME validation can be incorrect
The following will produce an incorrect BAD_EXTRA_RANGE_NAME
```
a = figure()
b = figure()
b.extra_x_range = {'myx': Range1d(start = 0, end = 1)}
b.circle(..., x_range_name = 'myx')

check_integrity(a) #--> ok
check_integrity(b) #--> ok

c = CustomJS(code = "", args = {"b": b})
a.x_range.callback  = c

check_integrity(a) # --> ko
```
The reason is that in the last check_integrity:

1. the circle glyph belonging to "b" is known to "a" through its "a.x_range.callback". 
2. the _check_bad_extra_range_names uses the extra ranges from "a" rather than "b".

It seems that in _check_bad_extra_range_names, one should only iterate over the layout's renderers.

poldavezac:
- [ x] issues: fixes #8723 
- [ x] tests added / passed: added to test_bad_extra_range_name in models/tests/test_plots

Most significant change is a modified version of bokeh.model.collect_models. The new collect_filtered_models gets to reject a object *and* discontinue exploration of its children depending on a test function provided in the tests. The old collect_models now calls on the new function.

### DIFF
--- a/bokeh/model.py
+++ b/bokeh/model.py
@@ -68,8 +68,12 @@ def collect_filtered_models(discard, *input_values):
     collecting all nested ``Models`` on the go.
 
     Args:
-        discard (Callable[[Model], bool])
-            a callable returning whether to continue down the object's references
+        *discard (Callable[[Model], bool])
+            a callable which accepts a *Model* instance as its single argument
+            and returns a boolean stating whether to discard the instance. The
+            latter means that the instance will not be added to collected
+            models nor will its references be explored.
+
         *input_values (Model)
             Bokeh models to collect other models from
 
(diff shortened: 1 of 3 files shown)


bokeh/model.py
```diff
@@ -59,30 +59,40 @@
...
+    if discard is None:
+        def _visitor(obj):
+            if obj.id not in ids:
+                queued.append(obj)
+        queue_one = _visitor
+    else:
+        def _filteredvistor(obj):
+            if obj.id not in ids and not discard(obj):
+                queued.append(obj)
+        queue_one = _filteredvistor
```
bryevdv: @poldavezac how about simplifying this by using a default filter in all cases:
```
if discard is None: 
    discard = lambda x: False

def _visitor(obj):
    if obj.id not in ids and not discard(obj)::
        queued.append(obj)
    queue_one = _visitor
```
Given the relatively small number of total objects, I don't think there is a performance concern to justify specialized code paths in this case (but certainly open to evidence otherwise)

bokeh/model.py
```diff
@@ -59,30 +59,40 @@
...
+    provided callable.
 
     Iterate over ``input_values`` and descend through their structure
-    collecting all nested ``Models`` on the go. The resulting list is
-    duplicate-free based on objects' identifiers.
+    collecting all nested ``Models`` on the go.
 
     Args:
+        discard (Callable[[Model], bool])
+            a callable returning whether to continue down the object's references
```
bryevdv: Would be good to state the expected signature of the callable here

bryevdv:
@poldavezac I left a couple of minor comments, FYI we should be releasing 1.1 next week if you have time to address them in the next few days. Otherwise 1.1.1 will be come out fairly soon afterwards too

bryevdv:
Thanks @poldavezac

Issue #6389, opened 2017-06-06
Test that bokehjs is npm-installable


SimonBiggs:
#### ALL software version info (bokeh, python, notebook, OS, browser, any other relevant packages)
bokehjs 0.12.5

#### Description of expected behavior and the observed behavior
I expect to be able to run npm install bojehjs@0.12.5 without error.

Observe error at kiwi. I have submitted a kiwi pull request that might fix this https://github.com/bokeh/kiwi/pull/2

Will need to update package.json within bokehjs to point to the new commit hash of kiwi after the pull request is accepted.

#### Complete, minimal, self-contained example code that reproduces the issue
npm install bojehjs@0.12.5

bryevdv:
ping @mattpap

mattpap:
bokeh 0.12.5 was completely broken with respect to this and just installing typescript shouldn't help with this. I fixed this for 0.12.6-dev, where kiwi is installed from `kiwi-prebuilt` branch.

SimonBiggs:
Hi Mateusz,

Thanks for letting me know. Sorry for highlighting something you already
knew about. I'll point my npm install to 0.12.6-dev.

Thank you.

bryevdv:
No apologies needed @SimonBiggs thanks for the report. @mattpap can this be closed then?

mattpap:
> can this be closed then?

I would like to test somehow that bokehjs is npm-installable. This've been an ongoing issue, where bokehjs package being broken and fixed in multiple consecutive recent releases.

bryevdv:
OK, I have labeled it accordingly

SimonBiggs:
@mgruesbeck For others who are having issues with the current state of this in npm I made the following changes to my `package.json` as a temporary solution:

```json
"scripts": {
  "preinstall": "mkdir bokeh; cd bokeh; git init; git remote add -f origin https://github.com/bokeh/bokeh.git; git config core.sparseCheckout true; echo \"bokehjs\" >> .git/info/sparse-checkout; git pull --depth=1 origin 0.12.6; cd .."
},
"dependencies": {
   "bokehjs": "file:bokeh/bokehjs"
}
```

bryevdv:
@mattpap is this still relevant? It occurs to me that there is not really anything at all we can do in CI automation about this, since we can't test that BokehJS is npm-installable until it is published to npm, and that only happens as the last step of a release-deploy.  

[...]

PR #8782, opened 2019-03-24
Relativize module paths in the emitted js/d.ts files

### LINKED ISSUE #6513: Add a link to bokehjs package on npmjs.com
Similarly to what we have for conda and pypi.

### LINKED ISSUE #8197: Relativize module paths in generated js/d.ts files
Hi, 

I am trying to use standalone BokehJS library with typescript and AngularJS.

I have installed JS library using `npm i bokehjs`.

and imported in component using `import * as Bokeh from "bokehjs"`

But after starting application getting multiple errors related to `Module not found: Error: Cannot resolve module`.

Here are first few lines of log - 

```
ERROR in ./~/bokehjs/build/js/tree/document.js
Module not found: Error: Cannot resolve module 'core/bokeh_events' in
@ ./~/bokehjs/build/js/tree/document.js 7:21-49

ERROR in ./~/bokehjs/build/js/tree/protocol/message.js
Module not found: Error: Cannot resolve module 'core/util/string' in
@ ./~/bokehjs/build/js/tree/protocol/message.js 3:15-42

ERROR in ./~/bokehjs/build/js/tree/models/layouts/layout_dom.js
Module not found: Error: Cannot resolve module 'core/dom' in 
@ ./~/bokehjs/build/js/tree/models/layouts/layout_dom.js 5:12-31 
```

mattpap:
fixes #8197 
fixes #6513


paddymul:
I built this branch and locally installed bokehJS as a package in a local package.  I can verify that bokehJS now imports successfully without throwing relative path errors.  Thank you

paddymul:
After further experimentation, I think I found some bugs with this. I'm working on an example project that will demonstrate the difficulty that I am having.

paddymul:
I created [bokehjs-npm-example](https://github.com/paddymul/bokehjs-npm-example).  bokehjs-npm-example has build instructions, I'm happy to turn that repo into documentation for bokeh.  I started with the 8197_relative branch and added the following two lines to bokehjs/src/lib/index.ts .   
```
import * as api                from "./api"
export {api}
```
This gets the bokehjs API into the npm module.  I'm happy to create a pull request for this, but I'm not sure how a PR on a PR works.  I'm also not sure if this adds the API to every bokeh artifact.  I would think that you would want the API to be a part of the npm module.  

When running the application (with code pulled from the bokehjs documentation), I get the following error

```
TypeError: FlatBush is not a constructor
/Users/paddy/code/bokeh/bokehjs/build/js/lib/core/util/spatial.js:19
```
This corresponds with 
```
import FlatBush = require("flatbush")
```
in [bokehjs/src/lib/core/util/spatial.ts](https://github.com/bokeh/bokeh/blob/mattpap/8197_relativize/bokehjs/src/lib/core/util/spatial.ts#L1)

I will dig around and see if I can figure anything obvious out.  Let me know if there are any fixes that I can help with.

paddymul:
After more research, this seems to be related to the module type.  BokehJS currently uses commonJS, npm seems to want es6 modules.  

The whole build system for bokehJS is complicated and hard to follow.  Could you give an overview of the intentions of the bokehJS setup.  I'm working to become more familiar with typescript build setups and npm packaging in general.  When looking at this project with an aim toward contributing, I have to work through the following steps for each change

1. What is the standard npm way of doing things?
2. What is the typescript standard way of doing things?
3. What is bokehJS actually doing here?
4. Why did bokehJS make those choices?
5. If I can make the change which enhances end user functionality that I want, does that mesh with 4 and will it be accepted by the maintainers?

In my view better understanding 4 is the biggest missing part from the bokeh team.  The rest (especially 1 and 2) I should be able to figure out myself.

Thoughts?

mattpap:
@bryevdv, I think this PR can be considered to be merged for 1.1. I did some additional testing on Windows to make sure we don't regress there. I also have code for bringing this functionality under test (issue #6389), but due to unreliability of npm cli commands, I will have to put some work into that.

mattpap:
> `TypeError: FlatBush is not a constructor`
> `/Users/paddy/code/bokeh/bokehjs/build/js/lib/core/util/spatial.js:19`
> This corresponds with
> `import FlatBush = require("flatbush")`

That's typescript. This gets translated to `var FlatBush = require("flatbush")`, which is just a standard commonjs import. That line number in the exception is suspicious. On my system this would appear on line 10. As a starter I would upload `spatial.js` (e.g. as a gist), so that I could inspect it. Otherwise it may be a bunch of different reasons, that would be even hard to enumerate right now.

paddymul:
My mistake in reporting the error.  The error doesn't come from the call to `require` but from calling `new FlatBush(points.length)`.   I read somewhere that there is difference between how commonJS exports constructors vs es6, but I can't find the article right now.  

Here is the code as built.
(built_spatial.js)[https://gist.github.com/paddymul/c8cf974d7a9f48c97a1e98a790e17bcd]

paddymul:
Success!!
I replaced the FlatBush require line with the following two lines in the typescript system
```
import FlatBush_mod = require("flatbush")
const FlatBush = FlatBush_mod.default
```

This results in typescript errors on compile but the built npm package does work properly when installed.  I would think you want to implement a different fix for this.  The important part is that, we are really close to having a functioning npm module.

mattpap:
The underlying problem is that flatbush uses an experimental `module` field in its `package.json` to define the ES module entry point, along with the standard `main` field. webpack and similar linkers seem to prefer `module` over `main`, despite its non-standard nature (this is actually the first time I encountered this, but to be fair I don't care that much for ES modules at this point). I will need to do proper research on this subject to decide whether this is a good idea for us to support the `module` field or not (currently I'm leaning towards the later, due to non-standard by default nature). You can fix this issue on your side by setting up your linker (assuming webpack) with:
```
{ resolve: { mainFields: ['main', 'module'] } }
```
For rollup etc. similar configuration option should be hopefully possible.

mattpap:
I created issue #8801 for the ESM support discussion (which is unrelated to the substance of this PR).

PR #8796, opened 2019-03-27
simplify spinner logic

### LINKED ISSUE #8786: Spinner jumps to low after one click, regardless of step
1.1dev11,  OSX Safari and FF

clicking the down button on the spinner jumps to the low bound after at most 1 click, regardless of the step value

```py
import numpy as np
from bokeh.models import ColumnDataSource, CustomJS, ColorPicker, Spinner
from bokeh.plotting import Figure, show

data = np.random.rand(20, 2)
cds = ColumnDataSource(data=dict(x=data[:, 0], y=data[:, 1]))

p = Figure(x_range=(0, 1), y_range=(0, 1), plot_width=350, plot_height=350)
r = p.circle(x='x', y='y', source=cds, color="red", size=8)

w = ColorPicker(title="Gylp Color", color="red", height=20, width=100)
cb = CustomJS(args={'r': r}, code="""
r.glyph.fill_color = cb_obj.color
r.glyph.line_color = cb_obj.color
""")
w.js_on_change('color', cb)


w2 = Spinner(title="Glyph size", low=1, high=20, step=0.1, value=8, width=100)
cb = CustomJS(args={'r': r}, code="""
r.glyph.size = cb_obj.value
""")
w2.js_on_change('value', cb)

show(row(widgetbox(w2, w), p))
```

bryevdv:
- [x] issues: fixes #8786
- [x] tests added / passed

@xavArtley @mattpap  please comment. I understand what the logic in the older version was for, but I am not sure why it is necessary. Setting min/max/step on the input will already keep the spinner from going outside min/max. We just need to set the value carefully by counting significant digits of step, which I have now done explicitly with a function (an ugly one, but it is clear what it is for). 

We can certainly consider revisiting the logic later, but with this code, the example in the issue and several variants all perform exactly as I would expect them to. This is explicitly a big improvement, so I would like to merge this for 1.1 (I did realize that the snapping of 8 to 8.2 with (min, max, step) == (1, 20, 0.3) is instrinsic to the HTML element operation, not due to our code. So that's fine/is what it is)

### DIFF
--- a/bokehjs/src/lib/models/widgets/spinner.ts
+++ b/bokehjs/src/lib/models/widgets/spinner.ts
@@ -2,7 +2,15 @@ import {InputWidget, InputWidgetView} from "models/widgets/input_widget"
 import {input} from "core/dom"
 import * as p from "core/properties"
 
-const {log10, round, min, max} = Math
+const {abs, floor, log10} = Math
+
+function _get_sig_dig(num: number) : number {
+  let x = abs(Number(String(num).replace(".", ""))) // remove decimal and make positive
+  if (x == 0) return 0
+  while (x != 0 && (x % 10 == 0)) x /= 10 // kill the 0s at the end of n
+
+  return floor(log10(x)) + 1 // get number of digits
+}
 
 export class SpinnerView extends InputWidgetView {
   model: Spinner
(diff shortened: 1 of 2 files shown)


bryevdv:
I've opted to update the two failing integration tests. As best I can tell, the behavior of bare HTML inputs with spinners is to allow any value to be explicitly entered by the user by typing it, and by default I think we ought to hew to that. We could add extended validation policy as options later.

xavArtley:
Ok for me,
I haven't thought to parse directly the step to find the number of signicative digits

bryevdv:
@mattpap I will merge this tonight unless you have comments or want time

PR #8800, opened 2019-03-28
Fixes trying to index dict_keys

przemub:
This pull request fixes a small bug in application.py, causing the application to crash with a cryptic error, by trying to index `kwargs.keys()`.

It is probably a leftover from Python 2 times.


mattpap:
Thanks @przemub.

PR #8803, opened 2019-03-30
Make test_python_execution_with_OO work on Windows

### LINKED ISSUE #8789: codebase tests fail on Windows
There's two issues with ```\tests\codebase\test_python_execution_with_OO.py```: 

* It generates a long command that exceeds the 8191 Cmd.exe character limit
* It generates backslashes that need to be replaced by dot

mattpap:
fixes #8789


mattpap:
I verified that this actually works on Windows (in cmd and git-bash).

PR #8805, opened 2019-03-31
Make GridBox.spacing adhere to CSS argument order

mattpap:
Previously `GridBox.spacing = [horizontal, vertical]` now it's `[vertical, horizontal]` to match CSS, where pairs of values always indicate vertical and horizontal dimensions respectively. This is necessary, because other APIs (e.g. `LayoutBox.margin`) follow CSS' order.


mattpap:
@philippjfr, I don't know if you use this, but if so, changes to your code may be necessary.

philippjfr:
>@philippjfr, I don't know if you use this, but if so, changes to your code may be necessary.

Thanks for the update, I wasn't using it but having discovered it exists I will start doing so.

PR #8808, opened 2019-04-01
Allow for simpler creation of complex grids

mattpap:
Example:
```py
In [1]: from bokeh.layouts import row, column, grid

In [2]: from bokeh.models import Spacer

In [3]: grid(column(Spacer(), row(column(Spacer(), Spacer()), Spacer()))).children
Out[3]: 
[(Spacer(id='1001', ...), 0, 0, 1, 2),
 (Spacer(id='1002', ...), 1, 0, 1, 1),
 (Spacer(id='1003', ...), 2, 0, 1, 1),
 (Spacer(id='1005', ...), 1, 1, 2, 1)]
```


mattpap:
@bryevdv, I implemented the flattening algorithm, but I didn't proceed with implementing implicit grids. There are too many assumptions to make this work and it will be better for the user to do the conversion explicitly. This way we will also not sacrifice layout's new functionality by default.

bryevdv:
@mattpap I'll have to play with this tonight.

@bokeh/core Given that user code  change is required to maintain the nicest possible layout, e.g. with axes aligned, should we just bite the bullet can call this Bokeh 2.0 then?

jbednar:
Bokeh 2.0 would make it simpler for people to specify that they don't want the layout changes that will inevitably accompany this release, and it makes it easier to make a big splash of all the new behavior.  Plus 2.0 lets people expect that there might be a few rough spots that will get ironed out in subsequent minor releases.  But I don't have a strong preference; 1.1 or 2.0 are both reasonable names for this release.

mattpap:
This breaks behavior, but it technically isn't an API change. After all, old code will still work, just in a bit different way. I suppose a lot depends on how we define API changes. Looking at this from user's perspective, version bump to 2.0 gives indication that there may be significant API and/or other breakage, but it's not that the user has an option here, assuming he wants to take advantage of new developments. If we release 2.0, there won't be 1.1.1 or 1.2, because we can't afford supporting multiple branches of development. So the user has to upgrade bokeh and update his code anyway. To me, a major version change would be actually off-putting, because it would likely mean major changes to my code, and this isn't the case here. I think that as long as we use `x.y.z` versioning scheme, major changes to users' code must increase `x` part, minor changes must increase `y` part and no changes may increase `z` or other parts. I suppose both 1.1 and 2.0 are fine choices depending on where we place ourselves on the software versioning spectrum. My only issue with 2.0 version is that, if this was decided earlier, I could have actually improved the APIs instead of trying to fit new layout into old, inadequate APIs.

philippjfr:
I'll test this PR now, merge it and then cut a 1.1.0rc1. If we decide 2.0 is more appropriate we can still decide that at a later stage.

philippjfr:
Tested the shown example and some other grid based examples which all seemed to work well. I'll merge now.

PR #8818, opened 2019-04-06
Unify API of npm package and browser bundle

### LINKED ISSUE #8596: BokehJS examples to use in node js or angular
Hi,

I have installed bokehjs using `npm install bokehjs --save` in Angular application.(means  package.json contains `    "bokehjs": "^1.0.4"` as dependency).

after build I see 

```
No name was provided for external module 'bokehjs' in output.globals – guessing 'Bokehjs'
```

In application : 

```
import * as Bokehjs from 'bokehjs';

```

```
   const plt = Bokehjs.Plotting;
    console.log(plt);   // undefined

```

```
    console.log(Bokehjs);
```
output : 
```
{version: "1.0.4", embed: {…}, protocol: {…}, _testing: {…}, logger: Logger, …}
Models: ƒ (name)
documents: []
embed: {add_document_standalone: ƒ, add_document_from_session: ƒ, embed_items_notebook: ƒ, kernels: {…}, BOKEH_ROOT: "bk-root", …}
index: {}
logger: Logger {_name: "bokeh", _log_level: LogLevel, trace: ƒ, debug: ƒ, info: ƒ, …}
protocol: {Message: ƒ, Receiver: ƒ, __esModule: true}
safely: ƒ safely(fn, silent)
set_log_level: ƒ set_log_level(level)
settings: Settings {_dev: false}
version: "1.0.4"
_testing: {results: {…}, init: ƒ, record: ƒ, count: ƒ, clear: ƒ, …}
__esModule: true
__proto__: Object}
```

```
    console.log(`Bokehjs ${Bokehjs.version}`); // 1.0.4
```

I want to run [chart example](https://github.com/bokeh/bokeh/blob/master/bokehjs/examples/charts/charts.ts) in my angular application

mattpap:
There is a WIP webpack bundling [example](https://github.com/bokeh/bokehjs-examples/tree/master/webpack-example) to test this.

fixes #8596

### DIFF
--- a/bokehjs/make/tasks/scripts.ts
+++ b/bokehjs/make/tasks/scripts.ts
@@ -29,9 +29,8 @@ task("scripts:bundle", ["scripts:compile"], async () => {
     paths.lib.gl.main,
   ]
   const bases = [paths.build_dir.lib, './node_modules']
-  const excludes = ["node_modules/moment/moment.js"]
 
-  const linker = new Linker({entries, bases, excludes})
+  const linker = new Linker({entries, bases})
   const bundles = linker.link()
 
   const [bokehjs, api, widgets, tables, gl] = bundles
--- a/bokehjs/package-lock.json
+++ b/bokehjs/package-lock.json
@@ -82,15 +82,6 @@
         "@types/wnumb": "*"
       }
     },
-    "@types/pikaday": {
-      "version": "1.7.4",
-      "resolved": "https://registry.npmjs.org/@types/pikaday/-/pikaday-1.7.4.tgz",
-      "integrity": "sha512-0KsHVyw5pTG829nqG4IRu7m+BFQlFEBdbE/1i3S5182HeKUKv1uEW0gyEmkJVp5i4IV+9pyh23O83+KpRkSQbw==",
-      "dev": true,
-      "requires": {
-        "moment": ">=2.14.0"
-      }
-    },
     "@types/proj4": {
       "version": "2.3.4",
       "resolved": "https://registry.npmjs.org/@types/proj4/-/proj4-2.3.4.tgz",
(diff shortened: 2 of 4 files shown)


mattpap:
Note that currently this introduces a regression making bokehjs not node-importable, due to pikaday referencing `window` during initialization.

bryevdv:
> I've been battling issues like this for a while and I think in the long term we will have to split models and views to fix this properly. 

I agree, that would be a good task to demarcate 1.2

I'd like to include something about this in the blog announcement, if you can write up a paragraph(that maybe links to the webpack example) that would be helpful.

bryevdv:
I'll merge this can cut (hopefully) the last rc

paddymul:
Does this commit add the Bokeh API into the NPM module?

mattpap:
> 'd like to include something about this in the blog announcement, if you can write up a paragraph(that maybe links to the webpack example) that would be helpful.

@bryevdv, I'm not sure if I want to advertise this at this point. 1.1 is already packed and this would be something nice for 1.2 when finished. I say when finished, because although bokehjs is now a proper npm package and is usable in webpack, it is just barely usable. For example it's not possible to optimize bokehjs' webpack bundle (by using tree shaking), because we don't persist ESM imports/exports in compiled modules, which makes repacking bokehjs a pointless exercise. I'm going to fix this and other related issues in not too distant future, so we could make a splash in the next release.

mattpap:
> I agree, that would be a good task to demarcate 1.2.

Assuming we don't provide backwards compatibility for bokehjs (which I think is true), because this will break imports in all custom models.

paddymul:
I'm not sure if this is the proper place to make note of this, but I filed a bug report on the bokehjs-examples repo https://github.com/bokeh/bokehjs-examples/issues/1 .  I cannot get the webpack example to build properly by following the instructions.

bryevdv:
> Assuming we don't provide backwards compatibility for bokehjs (which I think is true), because this will break imports in all custom models.

OK Well maybe 2.0 then (which I'm fine with in Q3 btw) I did make a claim that *existing* BokehJS demonstrated usage should remain stable. I think we've already bent that slightly, but if we are going t break *everything*, that seems like 2.0 territory.

PR #8822, opened 2019-04-08
Bryanv/1 1 docs

bryevdv:
Small changes, edits, etc or 1.1

Notes:

* had to set a few widget widths (cc @mattpap for comment)

* very tiny code change to BAD_COLUMN check (skip starting at new WebSource base class)


examples/howto/layouts/manual_grid.py
```diff
@@ -27,12 +27,14 @@
 paragraph = Paragraph(text="We build up a grid plot manually. Try changing the mode of the plots yourself.")
 
 select = Select(title="Sizing mode", value="fixed", options=list(SizingMode), width=300)
+
+grid = GridBox(children=[(p1, 0, 1), (p2, 1, 0), (p3, 1, 1), (p4, 1, 2)])
```
mattpap: I would also use `grid([[None, p1, None], [p2, p3, p4]])` instead of the low-level API.

examples/howto/js_events.py
```diff
@@ -53,7 +53,7 @@ def display_event(div, attributes=[]):
 # Add a div to display events and a button to trigger button click events
 
 div = Div(width=1000)
-button = Button(label="Button", button_type="success")
+button = Button(label="Button", button_type="success", width=300)
 layout = column(button, row(p, div))
```
mattpap: I think all the cases are the same issue, so I will answer them here. The layout works correctly by fitting `button` into the available space, which is ~1600px total. If you want to have a button of specific width, then setting `width` will fix it at the given value. If the intent is to align the button with the plot, then I would change this example altogether:
```py
div = Div()
button = Button(label="Button", button_type="success")
layout = row(column(button, p), div)
```
Note that `div` will be initially of zero width, but adding content will trigger relayout.

examples/app/movies/main.py
```diff
@@ -34,7 +34,7 @@
     "Year": "Year",
 }
 
-desc = Div(text=open(join(dirname(__file__), "description.html")).read(), width=800)
+desc = Div(text=open(join(dirname(__file__), "description.html")).read(), width=800, sizing_mode="stretch_width")
```
mattpap: `width` is unnecessary with `stretch_width`.

tests/baselines/examples/app/movies
```diff
@@ -1,19 +1,19 @@
...
+      Slider bbox=[5, 212, 310, 39]
+      Slider bbox=[5, 261, 310, 39]
+      TextInput bbox=[5, 310, 310, 50]
+      TextInput bbox=[5, 370, 310, 50]
+      Select bbox=[5, 430, 310, 50]
+      Select bbox=[5, 490, 310, 50]
+    Plot bbox=[320, 0, 664, 569]
+      LinearAxis bbox=[52, 524, 607, 45]
+      LinearAxis bbox=[0, 29, 52, 495]
+      Title bbox=[52, 0, 607, 29]
```
mattpap: Missing newline character.

mattpap:
@bryevdv, I fixed app/surface3d. server tests don't fail on python exceptions, so this will have to be tested manually until that is resolved.

bryevdv:
@mattpap I think I hav addressed/fixed everything. If there are any last test issues can you take a look and merge? I will not be able to look again until tonight. I'd like to cut a dev build after this is merged, though. Then after a little testing tonight, start the full build machinery.

bryevdv:
cc @philippjfr you could also merge and cut rc4 if you happen to see things go green

bryevdv:
I founf a few minutes and started rc4

PR #8826, opened 2019-04-08
Improve aspect management with scale_height mode

### LINKED ISSUE #8821: Make layout respect aspect ratio with scale_height sizing mode
@mattpap 

* All sizing modes seem to work well except possibly height scaling. Check `sizing_mode.py` to confirm. I'm fine with stating a limitation or known issue I just want your comment. 

* I've had to set explicit widths on several widgets to prevent them from stretching across the entire screen (see the forthcoming docs PR for the commits) Maybe this is fine/expected but perhaps we are missing some guidance.

I tested every other non-reference example on OSX/Safari and everything looks fantastic (any small issues were previously known). I will try to do some spot windows testing soon.

mattpap:
I'm not particularly happy how aspect management works right now, but at least this fixes `scale_height` sizing mode. I will need to revisit this code later on.

fixes #8821

### DIFF
--- a/bokehjs/src/lib/core/layout/layoutable.ts
+++ b/bokehjs/src/lib/core/layout/layoutable.ts
@@ -1,4 +1,4 @@
-import {Size, Sizeable, SizeHint, BoxSizing} from "./types"
+import {Size, Sizeable, SizeHint, BoxSizing, SizingPolicy} from "./types"
 import {BBox, CoordinateTransform} from "../util/bbox"
 
 const {min, max, round} = Math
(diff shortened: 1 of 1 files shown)


philippjfr:
@bryevdv asked me to review and merge this PR but I feel like I don't have a solid enough understanding to do that. Specifically I'm not quite sure I understand the height stretching and scaling behaviors in the sizing_mode.py example. In the example the width of the plot stretches to fill the screen but the height only fills about half the screen in the scale_height, stretch_both and stretch_height modes. 

That being said I agree that the ``scale_height`` behavior is more consistent with this PR applied since it now scales to the same height as the ``stretch_heigh`` case, I just would have expected that to match the height of the screen.

bryevdv:
OK so I checked with the examples I know about, and I would say that things are not perfect, but better. That said, I did not see what you described @philippjfr can you be more specific about what you ran? Also I should have stated explicitly: I am really interested that this PR does not have any adverse affects elsehwere.

philippjfr:
I ran the ``sizing_mode.py`` example. As an example here is the output for stretch_height which I would have expected to fill the whole screen browser window height:

Screen Shot 2019-04-09 at 12 32 31 AM

Overall this PR is definitely an improvement for the scale_height case though, and I could not produce any adverse effects.

bryevdv:
@philippjfr Weird I get this:

Screen Shot 2019-04-08 at 4 36 12 PM


I will merge and cut rc5 for final final testing.

bryevdv:
oh, I do see that FF cc @mattpap

philippjfr:
Might be something to do with my multi-monitor setup then.

PR #8829, opened 2019-04-10
add keys support to autocomplete

### LINKED ISSUE #8828: [BUG] AutocompleteInput widget value attribute does not reflect selection
bokeh --version
1.1.0

python --version
Python 3.6.6

Expected:  Upon selecting a value in the autocomplete box the value attribute should reflect what has been selected.
Actual:  Incorrect value is displayed.  Displayed value corresponds with what was typed into the box and not what has been selected and actually showing in the box.

Possibly related to: Issue #2259
Repro steps:

Run the following app in a terminal with `bokeh serve --show input_box.py

In the autocomplete box begin typing some of the completions and try to select
one in the drop down list.  Only the characters typed will be reflected by "value".  The correct value shows up only if all characters are typed even if one of the values in the drop down is selected.

For example if the user type "100" in the box and the "100001" is selected from the drop down box--the value attribute will still be "100" even though the full  "100001" is showing:

```
from bokeh.models.widgets.inputs import AutocompleteInput

input_box = AutocompleteInput()
input_box.title = "Foo Bar"
input_box.value = "400"
input_box.completions = ["100001", "12344556", "3194567289", "209374209374"]


def update():
    print(input_box.value)


input_box.on_change('value', lambda attr, old, new: update())

curdoc().add_root(input_box)


```

bryevdv:
cc @danielrmeyer @mattpap 

- [x] issues: fixes #8828
- [x] tests added / passed

This is a WIP PR to start discussion. This PR implements the following behaviour: when an autocomplete menu is opened, the first item starts off highlighted. The arrows keys can be used to move the highlighted element up and down. Hitting enter with an open menu sets the value to the highlighted element. 

As a consequence, it is not possible to enter values outside the completions list. Is this desirable? What about the fact that the initial value is not validated, Should we issue a warning or even an exception if it is not? Or should users be able to explicitly enter non-completion values (possibly as a mode?)

Assuming we like this behvaivour (or get to a point where we do) I will add integration tests. 

ScreenFlow

### DIFF
--- a/bokeh/_testing/util/selenium.py
+++ b/bokeh/_testing/util/selenium.py
@@ -51,6 +51,7 @@ __all__ = (
     'get_table_column_cells',
     'get_table_row',
     'get_table_selected_rows',
+    'hover_element',
     'INIT',
     'paste_values',
     'RECORD',
@@ -129,6 +130,10 @@ class element_to_finish_resizing(object):
         else:
             self.previous_width = current_width
             return False
+
+def hover_element(driver, element):
+    hover = ActionChains(driver).move_to_element(element)
+    hover.perform()
 
 def enter_text_in_element(driver, element, text, click=1, enter=True):
     actions = ActionChains(driver)
--- a/bokehjs/src/less/menu.less
+++ b/bokehjs/src/less/menu.less
@@ -28,7 +28,7 @@
   & > :not(.bk-divider) {
     padding: @padding-vertical @padding-horizontal;
 
-    &:hover {
+    &:hover, &.bk-active {
       background-color: darken(white, 10%);
     }
   }
--- a/bokehjs/src/lib/models/widgets/autocomplete_input.ts
+++ b/bokehjs/src/lib/models/widgets/autocomplete_input.ts
@@ -2,11 +2,16 @@ import {TextInput, TextInputView} from "./text_input"
 
 import {empty, display, undisplay, div, Keys} from "core/dom"
 import * as p from "core/properties"
+import {clamp} from "core/util/math"
 
 export class AutocompleteInputView extends TextInputView {
   model: AutocompleteInput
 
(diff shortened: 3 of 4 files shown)


danielrmeyer:
Thank you for this pull request. I have installed your branch and the autocomplete widget works great now. My preference is to not allow non-completion values as I think the non completion input box would be better suited for that use case.  

In any case I have tested the fix and it looks good to me.

bryevdv:
@danielrmeyer do you have thoughts about what a "bad" initial value should do? We could:

* print a warning python side
* raise an exception python side
* do nothing

or possibly other things (open to ideas)

danielrmeyer:
@bryevdv I like the idea of just throwing an exception for a bad initial value; however, its probably best to follow the spirit of similar widgets.  I experimented with the multi-select widget and it seems to "do nothing" when presented with a bad initial value.  Here is my example code:
```
from bokeh.models.widgets.inputs import AutocompleteInput
from bokeh.models.widgets.inputs import MultiSelect
from bokeh.models.layouts import Column

input_box = AutocompleteInput()
input_box.title = "Foo Bar"
input_box.value = "400"
input_box.completions = ["100001", "12344556", "3194567289", "209374209374"]

multi_select = MultiSelect()
multi_select.options = ["foo", "bar", "foobar"]
multi_select.value = ["invalid"]

def update():
    print(input_box.value)


input_box.on_change('value', lambda attr, old, new: update())

curdoc().add_root(Column(input_box, multi_select))
```

bryevdv:
As an aside, `AutocompleteInput` evidently supports the `placeholder` attribute (or at least the python class has the property) but I don't think that makes much sense

bryevdv:
@mattpap I'll plan to merge this tomorrow, unless you request more time to review

PR #8838, opened 2019-04-14
add reset_policy to plots

### LINKED ISSUE #6927: Allow reset tool to only emit reset event, and skip built-in reset code
So far, `Range1d._initial_start` & `Range1d._initial_end` are private JS attributes. Because the content of a plot might change - e.g. an image which size evolves - it's necessary to be able to change the initial range limits so that the proper scale is displayed  when the user clicks on the `ResetTool` button.

bryevdv:
- [x] issues: fixes #6927
- [x] tests added / passed

This PR adds a `reset_policy` property to plots. When set to `"standard"` (default) the current behavior is maintained. When set to `"event_only"` then only the `Reset` event is emitted, i.e. no range or selection resets occur.


bryevdv:
Fairly straightforward, will merge when green

PR #8844, opened 2019-04-16
hook up slider bar color handling

### LINKED ISSUE #8843: bar_color Slider not updating
bokeh=1.1.0
python=3.6.2
notebook=Jupyter Notebook
OS= Win 10
Browser=Chrome

In the new version of Bokeh 1.1.0, the bar color does not change anymore when updated, which was working in version 1.0.4. I've added an example below where you can see that the bar_color attribute is being updated but the color changes are not visible. Initiation with another color (change startColor) does work.

```
from bokeh.models.widgets import Slider
output_notebook()

colors = ['#000008','#00001F','#0084FF','#00CAFF','#32FF4D']
startColor = 4

def plotter(doc):    
    def processChange(attr, old, new):
        print('old '+str(groupSlider.bar_color))
        groupSlider.bar_color = colors[new-1]
        print('new '+str(groupSlider.bar_color))
        
    groupSlider = Slider(start = 1, end =5, step = 1, value = startColor,title = 'Group',width=100,bar_color=colors[startColor-1])           
    groupSlider.on_change('value', processChange)
       
     
            
    doc.add_root(groupSlider)
handler = FunctionHandler(plotter)
app = Application(handler)
show(app)

```

bryevdv:
- [x] issues: fixes #8843
- [ ] tests added / passed

ScreenFlow

### DIFF
--- a/bokehjs/src/lib/models/widgets/abstract_slider.ts
+++ b/bokehjs/src/lib/models/widgets/abstract_slider.ts
@@ -10,6 +10,8 @@ import {SliderCallbackPolicy} from "core/enums"
 import {Control, ControlView} from "./control"
 import {CallbackLike0} from "../callbacks/callback"
 
+const prefix = 'bk-noUi-'
+
 export interface SliderSpec {
   start: number
   end: number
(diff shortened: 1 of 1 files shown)


bryevdv:
OK I would like to add tests, but we are missing any tests at all for slider and range slider, so it's a bit more of a life since those need to be set up completely. Will make a separate PR to work up several tests for sliders.

Issue #8176, opened 2018-08-20
Switch the default driver for export_png to chromedriver


p-himik:
Started at #5714. All relevant details are repeated here.

PhantomJS is no longer maintained, and at this point using it produces PNGs like these:
(a fragment with y-axis)
![image](https://user-images.githubusercontent.com/4410314/44352070-8d428d00-a4cd-11e8-80be-11f5ea09208c.png)
(a fragment with x-axis with rotated labels)
![image](https://user-images.githubusercontent.com/4410314/44352074-916eaa80-a4cd-11e8-9cf1-a4259bfd5a09.png)
Using normal font instead of the italic one does not mitigate the issue.

Currently, I'm using this code to create a driver instance and pass it down to `export_png`:
```python
import os
import shutil

from chromedriver_binary.utils import get_chromedriver_path
from selenium.webdriver import Chrome, ChromeOptions

options = ChromeOptions()
# TODO: Remove when https://bugs.chromium.org/p/chromium/issues/detail?id=875241 is fixed
options.binary_location = shutil.which('chrome')

options.add_argument('--headless')
options.add_argument('--disable-gpu')
options.add_argument("--no-sandbox")
options.add_argument("--window-size=2000x2000")

web_driver = Chrome(executable_path=os.path.join(get_chromedriver_path(), 'chromedriver'),
                    chrome_options=options)
```

`chromedriver_binary` is available by installing `python-chromedriver-binary` from the `conda-forge` channel. I tested my code with version 2.41.0.

There's a few of issues:
1. Google Chrome can be available under names other than `chrome`
2. There's no check for Google Chrome version
3. The fixed window size, 2000x2000. It probably could be somehow computed in advance
4. Need to check whether this could be used with Chromium - many people use it instead of Google Chrome

Also, maybe at some point there could be a simple option that would use Firefox, but right now it appears to be unusable as it complains about missing `log` command.

bryevdv:
Here is a complete example that sets the device pixel ratio to 1.0 to get a correctly scaled image on OSX:
```
from selenium.webdriver import Chrome, ChromeOptions

options = ChromeOptions()

options.add_argument('--headless')
options.add_argument("--window-size=2000x2000")
metrics = { "deviceMetrics": { "pixelRatio": 1.0 } }
options.add_experimental_option("mobileEmulation", metrics)
web_driver = Chrome(chrome_options=options)

import numpy as np

from bokeh.plotting import figure, show, output_file

N = 4000
[...]

PR #8854, opened 2019-04-19
Fix the exports when using Firefox as a webdriver.

### LINKED ISSUE #8853: [BUG] exports do not work with Firefox webdriver
Firefox webdriver does not currently support logs: https://github.com/SeleniumHQ/selenium/issues/2972

It makes it impossible to export figures with Firefox as a webdriver.
Fixing this issue may allow to fix this one: https://github.com/bokeh/bokeh/issues/8176

rmelotte:
- [x] issues: fixes #8853
- [ ] tests added / passed

This PR enables Firefox to be used as a webdriver by disabling logs. 

It can then be used like this:
```
from selenium.webdriver import Firefox, FirefoxOptions
options = FirefoxOptions()
options.headless = True
webdriver = Firefox(options=options)
export_png(item, path, webdriver=webdriver)
```

It _may_ be a starting point to fix this issue: https://github.com/bokeh/bokeh/issues/8176

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -306,7 +306,7 @@ def wait_until_render_complete(driver):
             if len(messages) > 0:
                 log.warning("There were browser warnings and/or errors that may have affected your export")
                 for message in messages:
-                log.warning(message)
+                    log.warning(message)
 
 #-----------------------------------------------------------------------------
 # Private API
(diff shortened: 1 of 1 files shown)


bryevdv:
Thanks for the PR @rmelotte

PR #8859, opened 2019-04-22
Add hatching fill patterns

### LINKED ISSUE #6135: Feature request: Textures to fill properties?
Hey All,

I found [this message in the google group](https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/8Q7DnVRlBtM) asking if adding hatching to graphs like quad/circle, etc. was possible.  Looks like there is probably an unexposed way to do this in a workaround, but just curious if anything higher level is in the works?  Apologies if this is a duplicate request, couldn't find anything with a search.  Thanks!

bryevdv:
Scratching an old itch. 

- [x] issues: fixes #6135
- [x] tests added / passed

TODO:

* [x] verify legends work
* [x] categorical hatch mapper
* [x] custom hatch support?
* [x] tests
* [x] examples
* [ ] user guide docs (incl webgl note)
* [x] ref docs

Needs tests and docs and work to add hatching to patch/patches, as well as hopefully a path for custom patterns. But I wanted to get some preliminary feedback. 

The basic API is to add:

* hatch_pattern - code for which built in pattern
* hatch_color - color to use for fill/stroke on pattern canvas
* hatch_alpha - alpha for pattern
* hatch_weight - basically line width for strokes on the pattern canvas
* hatch_number - pattern canvas size

Here is current output:

Screen Shot 2019-04-21 at 9 09 41 PM

In particular I would like feedback on "/" and "\\". One of them does "dashed" diagonal lines and the other does fully connected diagonal lines, at least the best that I could make from tiling a pattern. Now I'd kind of like to have both forms. What are good new symbols or names to distinguish them? Or is there some other good way? Are there other default patterns we can/should add? 

cc @bokeh/core  @bokeh/dev for comments


bryevdv:
Here's a test script:
```
from bokeh.core.enums import HatchPattern
from bokeh.palettes import Spectral11

output_file("bar_basic.html")

pats = list(HatchPattern)
tops = [4,6,5,7,3,4,5,4,7,6,4]
numbers = [12]* 11
numbers[0] = 5
numbers[1] = 17
numbers[3] = 6
numbers[4] = 7
numbers[5] = 20
numbers[-2] = 8
numbers[-1] = 27

p = figure(x_range=pats, plot_height=350, plot_width=800, title="Hatch Patterns",
           toolbar_location=None, tools="")

p.vbar(x=pats, top=tops, width=0.9, alpha=0.6, fill_color=Spectral11, line_color="grey",
       hatch_pattern=pats, hatch_color="black", hatch_number=numbers, hatch_weight=0.5, hatch_alpha=0.5)

p.y_range.start = 0
p.xaxis.major_label_text_font_size = "18pt"

show(p)
```

bryevdv:
OK I added them with (backquote) and `,` for now which is not a good plan, but just so they can all be viewed:

Screen Shot 2019-04-21 at 9 33 35 PM

LightnerAndrew:
@bryevdv Thanks! This looks great. I may suggest using `hatch_scale` or `hatch_size` instead of `hatch_number`. At first glance, that seems to be more intuitive, but let me know if there is a good reason to keep the description more general.  

I don't think I have a better solution for the markers without introducing two characters. But I can let you know if I come up with something.

bryevdv:
@LightnerAndrew Thanks I will change to `hover_scale`. I wanted something that generally conveyed that the density decreased as the number increased, but wanted to avoid referencing the pattern tile size directly as it is a bit over-specific to be helpful in all cases. 

I don't mind two letters, and in fact we might want to follow the pattern for markers where there is a "real" long name and the letters are just a convenience. Maybe:

* `.` dot
* `o` ring
* `+` cross
* `-` horizontal-line
* `|` vertical-line
* `/` right-diagonal-line
* `\` left-diagonal-line
* `X` diagonal-cross
* ` ` right-diagonal-dash
* `,` left-diagonal-dash
* `v` horizontal-wave
* `>` vertical-wave
* `*` criss-cross

open to suggestions for other names, other abbreviations, or additional patterns to add

LightnerAndrew:
@bryevdv great. that makes sense. (I'm assuming you mean `hatch_scale`).  I probably prefer the longer text option. At least in my work with matplotlib, I always end up documenting what the symbols stand for anyway. But I can see how having both options would be useful.

birdsarah:
This is fabulous!!! I edited you're last post to show the `. Please confirm this is correct. These seem reasonable to me. Is it possible to have a system where a new feature like this doesn't get it's api firmed until the next big release which gives us some time to play?

bryevdv:
>  Is it possible to have a system where a new feature like this doesn't get it's api firmed until the next big release which gives us some time to play? 

Seeing the engineering processes at bigger companies recently has been eye-opening wrt to feature flags and gates and preview APIs! I wish there was bandwidth/resources to support neat things like that, but really there isn't, at least right now, IMO. The best mechanism for that is just our existing dev builds, and publicizing them as much as possible. I'll make sure theres an immediate dev build and at least a week after this is merged before another release, and tweet about asking people to try it out as well. 

> I edited you're last post to show the <code>`</code>

I never knew how this was possible until now! 💯

bryevdv:
New script and output, with vertical and horizontal dash added
```
from bokeh.core.enums import HatchPatternAbbreviation

output_file("bar_basic.html")

pats = list(HatchPatternAbbreviation)
tops = [4,6,5,3,7,4,5,4,7,5,6,4,5,6,4]
numbers = [12]* 15
numbers[0] = 5
numbers[1] = 11
numbers[3] = 6
numbers[4] = 7
numbers[5] = 14
numbers[-2] = 8
numbers[-1] = 21

p = figure(x_range=pats, plot_height=350, plot_width=900, title="Hatch Patterns",
           toolbar_location=None, tools="")

p.vbar(x=pats, top=tops, width=0.9, alpha=0.9, fill_color="lightgrey", line_color="grey",
       hatch_pattern=pats, hatch_color="black", hatch_scale=numbers, hatch_weight=0.5, hatch_alpha=0.5)

p.y_range.start = 0
p.xaxis.major_label_text_font_size = "18pt"

show(p)
```

Screen Shot 2019-04-22 at 9 03 26 PM

bryevdv:
Here is the mapping of abbreviations to long names:
```
_hatch_patterns = (
    (".",  "dot"),
    ("o",  "ring"),
    ("-",  "horizontal-line"),
    ("|",  "vertical-line"),
    ("+",  "cross"),
    ('"',  "horizontal-dash"),
    (":",  "vertical-dash"),
    ("/",  "right-diagonal-line"),
    ("\\", "left-diagonal-line"),
    ("x",  "diagonal-cross"),
    (",",  "right-diagonal-dash"),
    ("`",  "left-diagonal-dash"),
    ("v",  "horizontal-wave"),
    (">",  "vertical-wave"),
    ("*",  "criss-cross"),
)
```
I currently wish we had made all enum values use dashes rather than underscores, and I'd like to work towards that over the long term (even if we always accept old versions)

bryevdv:
OK added support for patches:
```
import numpy as np
import pandas as pd

from bokeh.palettes import brewer
from bokeh.core.enums import HatchPatternAbbreviation

N = 20
cats = 10
df = pd.DataFrame(np.random.randint(10, 100, size=(N, cats))).add_prefix('y')

def stacked(df):
    df_top = df.cumsum(axis=1)
    df_bottom = df_top.shift(axis=1).fillna({'y0': 0})[::-1]
    df_stack = pd.concat([df_bottom, df_top], ignore_index=True)
    return df_stack

areas = stacked(df)
colors = brewer['Spectral'][areas.shape[1]]
x2 = np.hstack((df.index[::-1], df.index))

p = figure(x_range=(0, N-1), y_range=(0, 800))
p.grid.minor_grid_line_color = '#eeeeee'

p.patches([x2] * areas.shape[1], [areas[c].values for c in areas],
          color=colors, alpha=0.8, line_color=None, hatch_alpha=0.6,
          hatch_pattern=list(HatchPatternAbbreviation)[:areas.shape[1]])

show(p)
```
Screen Shot 2019-04-22 at 9 49 38 PM

and patch (just an image here)

Screen Shot 2019-04-22 at 10 01 27 PM



@birdsarah @bokeh/dev I think I can finish this up tomorrow night (tests and docs left), and merge wednesday (hopefully after some input from @mattpap) Any and all comments before then are helpful. Will have dev build for ~1 wk to play with as well.

LightnerAndrew:
@birdsarah @bryevdv Thanks for the effort on this! I'm really excited to add this to the list of options available when I'm generating graphs, especially choropleth maps for demarking disputed territories and such.

jbednar:
> I can finish this up tomorrow night (tests and docs left),

Does that mean that legends work already? I don't see examples above and it's still unclicked in the original message.  Having a legend is definitely important for this to be useful for choropleths and similar cases.

bryevdv:
> Does that mean that legends work already?

I just have not tested at all yet. I would expect them to, since legend rendering goes through the same render function as normal rendering. But I wanted to remind myself to check for certain. If someone else can try it quickly before I can, and post the results, that would certainly be helpful.

bryevdv:
Custom hatching was pretty easy to add by hijaking `CustomJS`:

Screen Shot 2019-04-23 at 8 39 28 PM

I don't love it but I don't hate it and it can be done without adding anything new. It's a pretty obscure use case in any case so I think it will be fine as long as there are docs. But we an discuss once a dev build and examples are out.

bokeh/models/glyphs.py
```diff
@@ -884,6 +896,10 @@ class Quad(Glyph):
     The %s values for the quads.
     """)
 
+    hatch_props = Include(HatchProps, use_prefix=False, help="""
+    The %s values for the horizontal bars.
```
bryevdv: should be for the quads

jbednar:
Nice! Custom hatching seems useful for choropleths, allowing the use of a custom image patch with semantic meaning (e.g. an icon indicating the major industry or the winning political party in a given region).

bryevdv:
Also reminder that none of this will currently work with WebGL, which I will note in the docs

bryevdv:
Ok I got it to load images by (ab?)using promises. If anyone @birdsarah @mattpap wants to tell me why the `logo` CustomJS is terrible and should not be demonstrated, please do. 

Otherewise, I have the examples I want written. I will put them in narrative docs tomorrow night. Plan on 1.1.1 a week form monday or tuesday. 

Screen Shot 2019-04-24 at 10 40 17 PM

```
from bokeh.io import show
from bokeh.core.enums import HatchPattern
from bokeh.models import CustomJS

spiral = CustomJS(code="""
const h = cb_data.hatch_scale
const h2 = h/2

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')

ctx.moveTo(h2, h2);
ctx.beginPath()
for (i = 0; i < 720; i++) {
  angle = 0.1 * i;
  x = h2 + (0.9 * angle) * Math.cos(angle);
  y = h2 + (0.9 * angle) * Math.sin(angle);
  ctx.lineTo(x, y);
}
ctx.strokeStyle = "#000";
ctx.stroke()

return canvas
""")

B = CustomJS(code="""
const h = cb_data.hatch_scale
const h2 = h/2

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')
ctx.font = "12px bold";
ctx.fillText("b", 12, 12)
ctx.fillText("B", 15, 24)
ctx.fillText("β", 25, 10)

return canvas
""")

logo = CustomJS(code="""
const h = cb_data.hatch_scale

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')

const img = new Image()
img.src = 'https://bokeh.pydata.org/en/latest/_static/images/logo.png'

function onload2promise(obj) {
    return new Promise(resolve => {
        obj.onload = () => resolve(obj);
    });
}

img.src = 'https://bokeh.pydata.org/en/latest/_static/images/logo.png'
ctx.drawImage(img, 0, 0, h, h)
return canvas
""")

p = figure(title="Built-in Hatch Patterns",
           toolbar_location=None, tools="", y_axis_location="right")

r = p.vbar(x=0, top=5, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='@'), hatch_scale=80, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"@": spiral})

r = p.patches(xs=[[2, 3, 5, 6.2, 6, 5.3, 4, 2.5]], ys=[[1.2, 1, 2.7, 3.6, 5, 4, 3.2, 2.8]], fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='B'), hatch_scale=34, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"B": B})

r = p.vbar(x=8, top=5, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='logo'), hatch_scale=40, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"logo": logo})

show(p)
```

bryevdv:
Image support will definitely have to come later. We really will need to back async in for handling images  to make it work. Here's what's possible with `hatch_extra` now (comments on usage/api welcome)

Screen Shot 2019-04-24 at 11 14 28 PM

```
from bokeh.io import show
from bokeh.core.enums import HatchPattern
from bokeh.models import CustomJS

spiral = CustomJS(code="""
const h = cb_data.hatch_scale
const h2 = h/2

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')

ctx.moveTo(h2, h2);
ctx.beginPath()
for (i = 0; i < 720; i++) {
  angle = 0.1 * i;
  x = h2 + (0.9 * angle) * Math.cos(angle);
  y = h2 + (0.9 * angle) * Math.sin(angle);
  ctx.lineTo(x, y);
}
ctx.strokeStyle = "#000";
ctx.stroke()

return canvas
""")

B = CustomJS(code="""
const h = cb_data.hatch_scale
const h2 = h/2

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')
ctx.font = "12px bold";
ctx.fillText("b", 12, 12)
ctx.fillText("B", 15, 24)
ctx.fillText("β", 25, 10)

return canvas
""")

p = figure(title="Built-in Hatch Patterns",
           toolbar_location=None, tools="", y_axis_location="right")

r = p.vbar(x=0, top=5, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='@'), hatch_scale=80, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"@": spiral})

r = p.patches(xs=[[2, 3, 5, 6.2, 6, 5.3, 4, 2.5]], ys=[[1.2, 1, 2.7, 3.6, 5, 4, 3.2, 2.8]], fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='B'), hatch_scale=34, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"B": B})

show(p)
```

mattpap:
> I currently wish we had made all enum values use dashes rather than underscores, and I'd like to work towards that over the long term (even if we always accept old versions e.g. "bottom_left").

The reason for using underscores is that it allows for mapping enums to identifiers, which in turn allows for better auto-completion, easier utilization of static type systems, etc. I don't mind having both, with underscores being the default, but I really mind inconsistency. If you want to proceed this path, let's do it in a separate PR and do it at properties' level, covering all enums at the same time.

bokehjs/src/lib/core/property_mixins.ts
```diff
@@ -18,6 +18,15 @@ export type Fill = {
   fill_alpha: p.Property<number>
 }
 
+export type Hatch = {
+  hatch_color: p.Property<Color | null>
+  hatch_alpha: p.Property<number>
+  hatch_scale: p.Property<number>
+  hatch_pattern: p.Property<string>
+  hatch_weight: p.Property<number>
+  hatch_extra: p.Any
```
mattpap: Property implementations should be left for `define()`. In types like this one, you should preferably (there are exceptions related to specs) use abstract property types, so `p.Property<unknown>`. However, given that `Dict(String, Instance("bokeh.models.callbacks.CustomJS"))`  can be expressed in the type system, the complete type is `p.Property<{[key: string]: CustomJS}>`. Then you can use `p.Any` in `define()`, because we don't have a specific implementation for this kind of property.

bokehjs/src/lib/core/visuals.ts
```diff
@@ -8,6 +8,166 @@ import {LineJoin, LineCap, FontStyle, TextAlign, TextBaseline} from "./enums"
 import {HasProps} from "./has_props"
 import {ColumnarDataSource} from "models/sources/columnar_data_source"
 
+function _horz(ctx: CanvasRenderingContext2D, h: number, h2: number): void {
```
mattpap: `Context2d` from `core/util/canvas` would be a better choice, especially if you would like to use any features we extend the native context with.

bokehjs/src/lib/core/property_mixins.ts
```diff
@@ -18,6 +18,15 @@ export type Fill = {
   fill_alpha: p.Property<number>
 }
 
+export type Hatch = {
+  hatch_color: p.Property<Color | null>
+  hatch_alpha: p.Property<number>
+  hatch_scale: p.Property<number>
+  hatch_pattern: p.Property<string>
```
mattpap: Should be `HatchPatternType` instead of `string`.

bokehjs/src/lib/core/visuals.ts
```diff
@@ -8,6 +8,166 @@ import {LineJoin, LineCap, FontStyle, TextAlign, TextBaseline} from "./enums"
...
+}
+
+function _get_canvas(size: number): HTMLCanvasElement {
+  const canvas = document.createElement('canvas')
+  canvas.width = size
+  canvas.height = size
+  return canvas
+}
+
+function create_hatch_canvas(hatch_pattern: string, hatch_color: string, hatch_scale: number, hatch_weight:number ): HTMLCanvasElement {
```
mattpap: `hatch_pattern` should be of type `HatchPatternType`. `hatch_color` should be of `Color` type (event if that directly aliases to `string`). Note that at use site you will have to use `!`, to reduce `Color | null` to `Color` (this is a side effect of having `doit` and retrieving color value independently).

bokehjs/src/lib/core/visuals.ts
```diff
@@ -154,6 +314,65 @@ export class Fill extends ContextProperties {
...
+    ctx.fillStyle = ctx.createPattern(pattern, 'repeat')!
+    ctx.globalAlpha = this.hatch_alpha.value()
+  }
+
+  get doit(): boolean {
+    return !(this.hatch_color.spec.value === null ||
+             this.hatch_alpha.spec.value == 0 ||
+             this.hatch_pattern.spec.value == " " ||
+             this.hatch_pattern.spec.value == "blank" ||
+             this.hatch_pattern.spec.value === null)
```
mattpap: I would remove `" "` and `"blank"`, because having three values indicating no action is an overkill. Note that comparing a value of a type that doesn't include `null` to `null` is unfortunately allowed in TypeScript (I'm working to disallow this on tslint level), so you have to add `null` to the respective type of `hatch_pattern` property, even if the compiler doesn't complain about lack thereof. It will, however, complain at use site (examples/tests), disallowing you to pass `null` as `hatch_pattern`'s value.

bokehjs/src/lib/models/glyphs/patch.ts
```diff
@@ -14,47 +14,37 @@ export class PatchView extends XYGlyphView {
   model: Patch
   visuals: Patch.Visuals
 
+  protected _inner_loop(ctx: Context2d, indices: number[], sx: Arrayable<number>, sy: Arrayable<number>, func: any): void {
```
mattpap: `func` should have type `(this: Context2d) => void`.

bokehjs/src/lib/core/visuals.ts
bryevdv: Without them you can't have a blank pattern included with a hatch mapper

bryevdv:
> The reason for using underscores is that it allows for mapping enums to identifiers, which in turn allows for better auto-completion, easier utilization of static type systems, etc.

Those are reasonable benefits, I will change them back

bryevdv:
Thanks for the notes @mattpap I will get to those tonight. One thing I noted last night: with a strict scalar enum property for `Patch` it's impossible to set any custom hatch pattern, except by overriding the name of an existing pattern, because Enum validation kicks in. Could make `Either(HatchPatternType, String)` but thought I'd ask if you had any other ideas to.  I suppose it  could be argued that the vector property is failing to validate properly, as well. But they both need to accept more than just the enum if custom patterns are allowed.

mattpap:
> (...) with a strict scalar enum property for Patch it's impossible to set any custom hatch pattern (...)

OK, so I overlooked this use case:

> `hatch_pattern=dict(value='B'), hatch_extra={"B": B})`

Unless you really want to have custom hatch pattern enumerations, I would coalesce this into a single property (`hatch_pattern`) of type `Either(HatchPattern, CustomJS)`, at least for now, with `CustomJS` appropriately type parameterized in TypeScript. In the long term `CustomJS` would be replaced with a texture-like type, allowing generated textures, images, etc. The current implementation doesn't support this, but given the discussion, I suppose this could be implemented in future.

bryevdv:
> Either(HatchPattern, CustomJS)

This was my initial thought but then its not workable to have hatch pattens specified in a CDS. E.g. this is not compatible with several bars all having different custom patterns, or even compatible with one bar having a custom pattern and the rest having built in patterns.

bryevdv:
We could also punt on custom patterns for now, though I'd prefer not to. I may add that spiral pattern as a built in regardless, though, it's pretty nice and visually distinctive

birdsarah:
>  I think custom patterns are very useful, but not if hobbled like that. I'm definitely interested in any other ideas that afford the possibility of using custom patterns selectively alongside other patterns.

I'm afraid I do not have an implementation idea. But I just wanted to second your opinion about the value of custom patterns alongside regular. The most obvious example is @LightnerAndrew's point about maps - being able to show forests with unicode trees for example. Combined with the fact that we now have patches with holes, this is huge for geo.

bryevdv:
OK @mattpap I went ahead and formalized a new `Texture` model, with `CanvasTexture` and `ImageURL` texture. This allowed me to make a currently fairly ugly code path to defer a re-render in the image case. I will see about cleaning things up more, but mostly I consider this an exploratory stab that points toward some imporovements later (I like the idea of passing the render code to the visuals to get all the `set_vectorize` etc out of the glyphs) 

Here is a script with @birdsarah's unicode trees, as well as three images. This loads correctly even with caches emptied (you can see the re-render happen when the image becomes available)

```
from bokeh.io import show
from bokeh.core.enums import HatchPattern
from bokeh.models import CanvasTexture, ImageURLTexture


image1 = ImageURLTexture(url='https://mir-s3-cdn-cf.behance.net/project_modules/max_1200/d3616210009205.5727a90633bf2.jpg')
image2 = ImageURLTexture(url='https://cdn.tutsplus.com/vector/uploads/legacy/qt/2011_QT/qt_130_floral_pattern/final.jpg')
image3 = ImageURLTexture(url='https://mdn.mozillademos.org/files/222/Canvas_createpattern.png')

forest = CanvasTexture(code="""
const h = scale
const h2 = h/2

const canvas = document.createElement('canvas')
canvas.width = h
canvas.height = h
const ctx = canvas.getContext('2d')
ctx.font = "12px bold";
ctx.fillText("🌲", 12, 12)
ctx.fillText("🌲", 15, 24)
ctx.fillText("🌲", 25, 10)
ctx.fillText("🌲", 28, 26)

return canvas
""")

p = figure(title="Built-in Hatch Patterns",
           toolbar_location=None, tools="", y_axis_location="right")

r = p.vbar(x=0, top=5, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='forest'), hatch_scale=40, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"forest": image1})

r = p.vbar(x=0, bottom=6, top=10, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='forest'), hatch_scale=40, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"forest": forest})

r = p.vbar(x=0, bottom=11, top=15, width=3, fill_color="#fafafa", line_color="grey",
           hatch_pattern="spiral", hatch_scale=40, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5)

r = p.patch(x=[2, 3, 5, 6.2, 6, 5.3, 4, 2.5], y=[1.2, 1, 2.7, 3.6, 5, 8, 9.2, 4.8], fill_color="#fafafa", line_color="grey",
           hatch_pattern='image', hatch_scale=34, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"image": image2})

r = p.patches(xs=[[2, 3, 5, 6.2, 6, 5.3, 4, 2.5]], ys=[[10.2, 10, 12.7, 13.6, 15, 18, 19.2, 14.8]], fill_color="#fafafa", line_color="grey",
           hatch_pattern=dict(value='image'), hatch_scale=34, hatch_color="black", hatch_weight=0.5, hatch_alpha=0.5,
           hatch_extra={"image": image3})

show(p)
```

Screen Shot 2019-04-27 at 12 29 51 PM

bryevdv:
@birdsarah If I merge this and get it in a dev build by Monday, is there any chance you could provide a nice GIS example that uses the custom patterns?

@canavandl @philippjfr @jlstevens @jbednar anyone else with interesting examples would also welcome the help to round this out (it turned out to be a fair bit more work to get to this point than I expected)

bokehjs/src/lib/core/property_mixins.ts
bryevdv: let it string for now, given requirements for custom textures. Open to other ideas to tighten things up.

bryevdv: I also used `p.Instance` in `hatch_extra` it seems like this module should not depend on models imports

birdsarah:
Unlikely. I have a very tight two weeks coming up.

bokehjs/src/lib/core/property_mixins.ts
mattpap: At least for now, I would add `export type HatchPattern = HatchPatternType | string`. This doesn't increase type safety, but at least indicates more clearly the intent. I would also mark `hatch_extra` and texture support in `hatch_pattern` as experimental (similarly to `width_policy` or `height_policy` in `LayoutDOM`), so that there are no issues with API breakage if a refined API pops up in near future.

bokehjs/src/lib/core/property_mixins.ts
```diff
@@ -18,6 +18,15 @@ export type Fill = {
...
+  hatch_weight: p.Property<number>
+  hatch_extra: p.Property<{[key: string]: p.Instance}>
```
mattpap: `p.Instance` should be replaced with `Texture`. `T` in `p.Property<T>` should never contain any property types.

bokehjs/src/lib/models/textures/canvas_texture.ts
```diff
@@ -0,0 +1,48 @@
...
+    return new Function("ctx", "color", "scale", "weight", "require", "exports", code)
+  }
+
+  get_pattern(color: any, scale: number, weight: number): (ctx: Context2d) => CanvasPattern | null {
+    return (ctx: Context2d): CanvasPattern | null => {
+      const canvas = document.createElement('canvas')
+      canvas.width = scale
+      canvas.height = scale
+      const pattern_ctx = canvas.getContext('2d')
+      this.func.apply(this, [pattern_ctx, color, scale, weight, require, {}])
```
mattpap: More JS idiomatic would be `this.func.call(this, pattern_ctx, color, ...)`. `apply()` indicates variability in the number of arguments. In fact you could `bind(this)` in `get func()` and then just use `this.func(pattern_ctx, color, ...)`. In the end all of those are equivalent at run time, but to me the meaning they convey is different.

bokehjs/src/lib/models/textures/image_url_texture.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (this.image.complete) {
+      defer_func()
+    } else {
+      this.image.onload = () => {
+        defer_func()
+      }
+    }
+    if (this.image.complete) {
+      defer_func()
+    }
```
mattpap: Why checking `image.complete` twice?

bokehjs/src/lib/core/visuals.ts
```diff
@@ -8,6 +8,179 @@ import {LineJoin, LineCap, FontStyle, TextAlign, TextBaseline} from "./enums"
...
+}
+
+function create_hatch_canvas(hatch_pattern: string, hatch_color: string, hatch_scale: number, hatch_weight:number ): HTMLCanvasElement {
+    const h = hatch_scale
+    const h2 = h / 2
+    const h4 = h2 / 2
+
+    const canvas = _get_canvas(hatch_scale)
+
+    const ctx = canvas.getContext("2d")! as Context2d
```
mattpap: So my suggestion to use `Context2d` wasn't entirely correct, because this and what we use primarily for rendering are different types of contexts. I'm a bit surprised this type checks at all. Currently to assert that `ctx` is actually `Context2d`, is to call `fixup_ctx(ctx)` after obtaining the context. In future work I will abstract canvas rendering, so that we won't be dealing with native APIs directly, so there will be no confusion here.

bokehjs/src/lib/core/visuals.ts
```diff
@@ -154,6 +327,86 @@ export class Fill extends ContextProperties {
...
+          return ctx.createPattern(canvas, 'repeat')!
+        }
+      }
+    } else
+      value = super.cache_select(name, i)
+
+    return value
+  }
+
+  _try_defer(defer_func: () => void):void {
```
mattpap: I think this could be made `private`.

bryevdv:
@mattpap sorry for the rebase/squash, things did not go as cleanly as I hoped after the stacked area PR. I believe I have implemented all of your suggestions. I plan to merge this late tomorrow, in case you have time for a last review. The only additions I plan from here on is to add are some user and reference examples.

bokehjs/src/lib/core/visuals.ts
bryevdv: Just to be clear, you think these should be reverted to `CanvasRenderingContext2D` ?

mattpap: Not really, because I would like to have consistency here and I would prefer to use our refined APIs over native ones. I wouldn't be too concerned with this, as I will be cleaning up rendering anyway in the upcoming work.

bryevdv: Ok, any objections if I merge this now as-is to get in a dev build?

mattpap:
Looks good, though I think that `hatch_extra` (and the related functionality) should be marked as experimental to allow refinement in near future.

bokehjs/src/lib/core/visuals.ts
bryevdv: Actually it looks like there are some very slight image diffs for bars, I will look at this tonight

https://ci.bokeh.org/travis/525789333/examples.html

mattpap: Yeah, let's merge.

bryevdv: OK, I will make a note to check out the diffs. It may actually be fine

gdementen:
@bryevdv I am pretty late to the party, but since the release is not out yet, maybe this not too late. Wouldn't it be more consistent with the rest of styling options to have hatch_line_width instead of hatch_weight? Even hatch_line_weight seems more discoverable to me than hatch_weight.

PR #8861, opened 2019-04-22
Create README to explain server_embed usage

JAK0723:
The examples in server_embed have requirements that may not be apparent to new users. This README should help avoid confusion.


bryevdv:
@JAK0723 can you expand on the rationale for suggeting renaming the files? I always just run e.g. `python flask_embed.py` without issue, and that seems simpler

JAK0723:
@bryevdv Sorry, I didn't realise how `if __name__ == '__main__':` worked, I just removed the renaming parts.

Where should the static note be moved to, under the comment about num_procs?

bryevdv:
Thanks for the PR @JAK0723

PR #8864, opened 2019-04-23
Allow export timeouts to be configurable.

### LINKED ISSUE #8863: [FEATURE] Allow the webdriver timeout to be customized
Currently, there is no way to customize the timeouts.
In some cases, this causes exports to fail when extending BokehJS models for example.

rmelotte:
- [x] issues: fixes #8863
- [ ] tests added / passed

This PR is a first attempt to allow for configurable timeouts when exporting figures.

### DIFF
--- a/bokeh/io/export.py
+++ b/bokeh/io/export.py
@@ -56,7 +56,7 @@ __all__ = (
 # General API
 #-----------------------------------------------------------------------------
 
-def export_png(obj, filename=None, height=None, width=None, webdriver=None):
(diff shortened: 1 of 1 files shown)


bryevdv:
@rmelotte Thanks for the PR! Would you mind putting "Added in 1.1.1" somewhere at the end of the docs for the new parameter. We have not historically been good about noting when new API is added but I would like the project to become better about it.

rmelotte:
Sure! Thanks for considering this. 
I just pushed the changes.

PR #8873, opened 2019-04-29
Add stacked lines and directed areas

### LINKED ISSUE #8848: [FEATURE] Stacked Areas and Lines
Strangely, I can't find a suitable issue for this. Anyway, I will submit a PR later this week to add support for this:

Screen Shot 2019-04-17 at 10 21 36 PM

Implementation is basically done, needs docs and tests. Would not mind some feedback on a couple of questions:

* The `harea` and `varea` only have *fill* props currently. The rationale is that AFAICT these kinds of charts never outline the sides/bottom. Does that seem reasonable? 

* x1 and x2? or x0 and x1? 

cc @bokeh/core @bokeh/dev @pzwang for any comments before submitting

bryevdv:
- [x] issues: fixes #8848
- [ ] tests added / passed

@mattpap @birdsarah  I plan to merge this when green so that the hatching PR can be rebased on top (to add hatching support for these new glyphs as well). I'll add narrative documentation in a follow-on PR later this week. 

But please make comments, corrections, or suggestions here. I will make changes in one of the other PRs.


bryevdv:
@birdsarah particularly interested what you think of `line_stack` API. Currently have
```
line_stack(x, y, **kw)
``` 
which expects *one* of x or y to be sequence to "stack", but not both (will error). Alternatively could do
```
line_stack(stackers, **kw)
```
which would explicitly have one arg for the stackers, and whether it is x or y depends on the the other one being provided as a kw arg, e.g `line_stack([...], x=...)` will stack y coords, because `x=` was provided. 

Other option is to explicitly have `hline_stack` and `vline_stack` but that's getting a but verbose even for me.

birdsarah:
Sorry I didn't see your message in time. Here's my thoughts anyway.....don't hate me....but I actually like hline_stack, vline_stack. My reasoning is that it is in keeping with an existing mental model for hbar, vbar, harea, varea... I think it's good to follow patterns whenever they're available. My take is that verbose is annoying, but not as annoying as having to look up syntax.

bryevdv:
FYI I have a WIP branch that makes that change

birdsarah:
Sweet!

PR #8896, opened 2019-05-08
Ensure that js_property_callbacks can be updated

### LINKED ISSUE #8895: [BUG] CustomJS callbacks added after initialization do not work
Using latest bokeh any CustomJS callbacks added after an app is initialized do not seem to ever be called. In the example below for instance we add a button which when clicked should add a JS callback triggered when the selection changes (a contrived example I know). However a change in the selection does not trigger the ``console.log`` after the button has been clicked indicating that the JS callback is not registered correctly:

```python
from bokeh.models import ColumnDataSource, CustomJS, Button
from bokeh.layouts import row

p = figure(tools=['box_select'])
cds = ColumnDataSource(data={'x': [1], 'y': [2]})
p.scatter(x='x', y='y', source=cds)

def add_customjs():
    cds.selected.js_on_change('indices', CustomJS(code='console.log("Test")'))

b = Button()
b.on_click(add_customjs)
r = row(p, b)
curdoc().add_root(r)
```

I have checked that the event is sent but haven't yet dug into what happens to the callback on the BokehJS end.

## Versions

Bokeh: 1.1.0
Python: 3.7
OS: OSX 10.13.4

philippjfr:
This PR ensures that when ``js_on_change`` is called a ``ModelChangedEvent`` is triggered and the callbacks are updated on the frontend. The main problem I'm currently facing is that I haven't yet figured out how to disconnect previous callbacks since at least in theory callbacks could be deleted as well as added. If we don't expect them ever to be deleted we could also simply keep track of callbacks that have been connected and skip them the second.

- [x] issues: fixes #8895
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/model.ts
+++ b/bokehjs/src/lib/model.ts
@@ -43,15 +43,8 @@ export class Model extends HasProps {
   connect_signals(): void {
     super.connect_signals()
 
-    for (const base_evt in this.js_property_callbacks) {
-      const callbacks = this.js_property_callbacks[base_evt]
-      const [evt, attr=null] = base_evt.split(':')
-      for (const cb of callbacks) {
-        const signal = attr != null ? (this.properties as any)[attr][evt] : (this as any)[evt]
-        this.connect(signal, () => cb.execute(this))
-      }
-    }
-
+    this._update_property_callbacks()
+    this.connect(this.properties.js_property_callbacks.change, () => this._update_property_callbacks())
     this.connect(this.properties.js_event_callbacks.change, () => this._update_event_callbacks())
     this.connect(this.properties.subscribed_events.change, () => this._update_event_callbacks())
   }
@@ -78,6 +71,18 @@ export class Model extends HasProps {
       return
     }
     this.document.event_manager.subscribed_models.add(this.id)
+  }
+
+  protected _update_property_callbacks(): void {
+    for (const base_evt in this.js_property_callbacks) {
+      const callbacks = this.js_property_callbacks[base_evt]
+      const [evt, attr=null] = base_evt.split(':')
+      for (const cb of callbacks) {
+        console.log(cb)
+        const signal = attr != null ? (this.properties as any)[attr][evt] : (this as any)[evt]
+        this.connect(signal, () => cb.execute(this))
+      }
+    }
   }
 
   protected _doc_attached(): void {
--- a/bokehjs/src/lib/model.ts
+++ b/bokehjs/src/lib/model.ts
@@ -23,6 +23,7 @@ export interface Model extends Model.Attrs {}
 
 export class Model extends HasProps {
   properties: Model.Props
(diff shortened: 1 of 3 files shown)


bryevdv:
@philippjfr Offhand this seems reasonable to me. Questions:

* do event callbacks need a similar treatment?
* What is the cost of disconnecting/reconnecting everything ever time, i.e. would it be better to try and diff the old/new callbacks dict more carefully?

bryevdv:
GitHub having a normal one with timezones, I see

Screen Shot 2019-05-08 at 7 40 43 AM

philippjfr:
>do event callbacks need a similar treatment?

Definitely worth checking, I think they are already handled on the frontend but I'm not sure `js_on_event` in Python triggers a `ModelChangedEvent` so may need similar treatment with an explicit trigger.

>What is the cost of disconnecting/reconnecting everything ever time, i.e. would it be better to try and diff the old/new callbacks dict more carefully?

The cost question is probably a good one for @mattpap, I can't imagine that's very expensive but I don't really know. Diffing them wouldn't be hard but I'd have to keep track of some more state.

mattpap:
I will need to play with this PR a bit to give my advice, because, as I recall, last time I worked in this area, there were issues with rapidly disconnecting and reconnecting signals, though I don't remember the details right now.

philippjfr:
@mattpap Any chance you could take a look soon? This is pretty high priority for us.

mattpap:
@philippjfr, I've been testing this in various ways and I can't reproduce issue I had before, so I suppose this  is good to go. As to performance issues, I would be too concerned right now. Let's have correct expected behavior first, and then we can get into performance optimizations (signaling API is pretty slow, which is nicely visible in UIEvents and internally in the properties system, so there's a lot to improve in this area).

bokehjs/src/lib/model.ts
```diff
@@ -23,6 +23,7 @@ export interface Model extends Model.Attrs {}
 
 export class Model extends HasProps {
   properties: Model.Props
+  private _js_callbacks: {[key: string]: any[]}
```
mattpap: Replacing `any[]` with `undefined[]` would be an improvement, and replacing it with `(() => void)[]` would be a greater improvement. In general, we are going away from using `any`, which is unsafe, and replace it with `undefined`. At some point referencing `any` will be disallowed completely.

philippjfr:
Okay I'll try to come up with some tests for this. Let me know if you want to see any changes to the fundamental approach here.

bryevdv:
Was there a decision about applying the same fix to event callbacks? Without looking, I would be surprised if they did not also suffer the same original issue

mattpap:
Event callbacks don't need this treatment, because `js_event_callbacks` and `subscribed_events` are processed every time an event is triggered.

philippjfr:
Thanks @mattpap, I appreciate it.

PR #8897, opened 2019-05-08
Reference link updated

chesucr:
`requirements.txt` file updated


bryevdv:
Thanks for the catch @chesucr !

PR #8905, opened 2019-05-10
Add Title.text_line_height property

### LINKED ISSUE #8904: Allow to offset plot frame's side panel annotations
I would like to be able to offset a plot title away from the plot area by an arbitrary amount. Currently the plot border area is dynamically minimized to fit the Title and the Title height appears to be the text height plus a hardcoded offset (https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/annotations/title.ts#L109)

I would like to implement a Title.text_line_height (similar to Text.text_line_height), so that the title annotation could have an arbitrary height. We may have to reconsider the Title.vertical_align property (which sets the title inside of the height + hardcoded offset box). This may be better replaced with a Title.text_baseline prop.

None

**Additional context**
None

canavandl:
This PR:
* Adds a new Title.text_line_height property to support offsetting the plot title away from the plot area an arbitrary distance.

- [x] issues: fixes #8904 
- [x] tests added / passed

Question: should we rename Title.vertical_align to Title.text_baseline for parity with Text.text_baseline (which only supports top/middle/bottom enum values also)?

Example code:
```python

p.title.text_line_height = 2
p.title.vertical_align = 'top'

p.circle(flowers["petal_length"], flowers["petal_width"],

output_file("iris.html", title="iris.py example")

show(p)
```

Screenshot (note the title is vertically offset from the plot area):

Screen Shot 2019-05-09 at 10 34 00 PM


bryevdv:
i don't have a strong feeling but if you do want to do that, with 2.0 not *too* far away we should wait until then to actually make a break-in change, but do a proper deprecation (i.e. with a python property setter/getter for the deprecated version) in the mean time

bokehjs/src/lib/models/annotations/title.ts
```diff
@@ -106,7 +106,7 @@ export class TitleView extends TextAnnotationView {
     else {
       this.visuals.text.set_value(this.ctx)
       const {width, ascent} = this.ctx.measureText(text)
-      return {width, height: ascent + 10}
+      return {width, height: ascent * this.visuals.text.text_line_height.value() + 10}
```
bryevdv: Wouldn't `this.model.text_line_height` work here?

mattpap:
Line height isn't the right tool for the job. Some sort of padding or margin would be much better suited, depending on what's the actual goal (consider that title is the text and the panel below it). Line height was introduced to allow rendering of multi-line text, but it's fully defined yet (PR #8445 supposed to fix that), so I would prefer not to attach any additional meaning to it in the meantime.

bryevdv:
> Line height isn't the right tool for the job. Some sort of padding or margin would be much better suited, depending on what's the actual goal (consider that title is the text and the panel below it). 

The thing is, unless you make an explicit way to modify line height, the baseline setting is never really going to be useful in this setting, because the text space allocated always tightly hugs the text

mattpap:
> should we rename Title.vertical_align to Title.text_baseline for parity with Text.text_baseline 

> (...) the baseline setting is never really going to be useful in this setting,

If anything, baseline is going to be removed (presumably, it depends on the outcome of the multi-line text work), because of its low level nature and incompatibility with multi-line text. Exposing it and some other canvas API directly was a mistake early on. Usually baseline is mistakenly used as vertical align across bokehjs.

bryevdv:
Ok, well can you commit to making a concrete proposal in the next week so we can start putting in deprecations?

canavandl:
I understand about preferring ``vertical_align`` over ``text_baseline`` and won't make that change. In terms of this PR, would it be better to replace the proposed ``Title.text_line_height`` property with something like ``Title.min_height``, so that Title annotations can have an arbitrary height? I think that wouldn't conflict with a future ``text_line_height`` for multiline titles.

bryevdv:
@mattpap do you have a proposal for API to add here?

mattpap:
I don't have a good proposal right now. I will be working on improving positioning (among other things) within the canvas, but that will happen in a month or so. For now I implemented PR #8980, which is a bit less convenient, but much more generic and doesn't force us to figure out any new properties and behavior of side panel annotations right now.

bryevdv:
I'm swayed by the utility of being able to theme this. Since it also does not preclude adding other approaches later in more comprehensive layout/sublayout work, will go with this for now.

PR #8906, opened 2019-05-10
pass layout kwargs to underlying objects

### LINKED ISSUE #8903: [BUG] layout accepts name as keyword arg but doesn't pass to underlying objects
#### software version info
bokeh 1.0.4
python 3.7.1

`bokeh.layouts.layout()` should be capable of taking a `name` argument for use in embedding as described [here](https://bokeh.pydata.org/en/latest/docs/user_guide/embed.html#standard-template).  Currently you can pass a `name` but it fizzles silently.

```

if __name__ == "__main__":
    fig = figure()
    fig2 = figure()
    lay = layout([fig, fig2], name='test')
    print(lay.name)

```

output:

```
(base) $ python test_layout.py
None
```

Looking at the [underlying code](https://github.com/bokeh/bokeh/blob/master/bokeh/layouts.py#L182), this could be fixed by having `_create_grid()` in `bokeh.layouts` take the remaining `kwargs` and pass to the `column()` or `row()` calls.  I'd be happy to contribute a PR for this provided a fix doesn't already exist that I missed.

zdgriffith:
This PR passes the `kwargs` given to `layout()` down to the underlying `row()` or `column()` calls in `bokeh.layouts`.  This fixes the behavior described in #8903.

- [x] issues: fixes #8903
- [x] tests added / passed

### DIFF
--- a/bokeh/layouts.py
+++ b/bokeh/layouts.py
@@ -219,7 +219,7 @@ def layout(*args, **kwargs):
     children = _handle_children(*args, children=children)
 
     # Make the grid
-    return _create_grid(children, sizing_mode)
+    return _create_grid(children, sizing_mode, **kwargs)
 
 def gridplot(children, sizing_mode=None, toolbar_location='above', ncols=None,
              plot_width=None, plot_height=None, toolbar_options=None, merge_tools=True):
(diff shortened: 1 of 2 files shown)


bryevdv:
Thanks for the fix @zdgriffith !

PR #8907, opened 2019-05-10
Render mixed canvas and webgl glyphs in correct z-order

### LINKED ISSUE #7699: Mixed canvas and webgl glyphs are painted in wrong z-order
See PR #7698 for details.

mattpap:
fixes #7699


examples/app/dash/main.py
```diff
@@ -49,11 +49,13 @@
...
 # Donut chart
 
-x = Counter({ 'United States': 157, 'United Kingdom': 93, 'Japan': 89, 'China': 63,
-              'Germany': 44, 'India': 42, 'Italy': 40, 'Australia': 35, 'Brazil': 32,
-              'France': 31, 'Taiwan': 31  })
+x = OrderedDict({
+    'United States': 157, 'United Kingdom': 93, 'Japan': 89, 'China': 63,
+    'Germany': 44, 'India': 42, 'Italy': 40, 'Australia': 35, 'Brazil': 32,
+    'France': 31, 'Taiwan': 31,
+})
```
bryevdv: i think I can count on one hand the number of times I've seen `OrderedDict` in the wild. I don' t think we should contort examples and demonstration for the sake of our tooling. As we are dropping 2.7 support soon we could consider moving image diff tests to python 3.7 or else not use a dict structure here at all

mattpap: When working with an example like this, I have an implicit assumption that the order of things should be maintained as specified in the source code, which unfortunately haven't been the case for long while. So change like this actually fixes the example by making it platform independent. Anyway, this change is still wrong (`OrderedDict({...})`), so I will remove it. I'm happy to switch to 3.7 even today.

PR #8910, opened 2019-05-12
Ensure change in Slider widget title updates view

### LINKED ISSUE #8909: [BUG] Changing Slider widget title property does not update view
In the AbstractSlider model the title property is not connected to any change events so changing the slider title after initialization does not actually update the title. This code:

```javascript
    this.on_change(value, () => this._update_title())
```

should be changed to:

```javascript
    this.on_change([value, title], () => this._update_title())
```

Version: Bokeh 1.1.0

philippjfr:
- [x] issues: fixes #8909
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/widgets/abstract_slider.ts
+++ b/bokehjs/src/lib/models/widgets/abstract_slider.ts
@@ -42,7 +42,7 @@ export abstract class AbstractSliderView extends ControlView {
     const {callback, callback_policy, callback_throttle} = this.model.properties
     this.on_change([callback, callback_policy, callback_throttle], () => this._init_callback())
 
-    const {start, end, value, step} = this.model.properties
+    const {start, end, value, step, title} = this.model.properties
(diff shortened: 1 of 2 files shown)


bryevdv:
@philippjfr please add an assertion in this test:

https://github.com/bokeh/bokeh/blob/7d236624456cee35f764ce60ee9a011faba75350/tests/integration/widgets/test_slider.py#L90-L108

should only be a line or two. 

There's a separate range slider test file that could also be updated similarly

bryevdv:
Actually I guess it will need to be a new server test, but it should be almost a direct copy/paste of this test:

https://github.com/bokeh/bokeh/blob/7d236624456cee35f764ce60ee9a011faba75350/tests/integration/widgets/test_slider.py#L174-L192

except updating/checking the title instead of the bar color.

philippjfr:
Thanks, I keep forgetting where these integration tests live, was looking for something like it. Should hopefully pass.

PR #8912, opened 2019-05-13
Use Python 3.7 in examples' tests on Travis CI

mattpap:


mattpap:
@bryevdv, should I add a 3.5 job with `--no-js`?

bryevdv:
@mattpap I would say this is fine for now, then when we make 2.0 switch the 2.7 job to 3.5

PR #8914, opened 2019-05-14
add value_throttled property to sliders

### LINKED ISSUE #4540: Slider callback_policy should work for apps
Currently `callback_policy` only affects the triggering of JS callbacks, but it would be extremely useful to be able to throttle callbacks triggered on the server as well. It's possible that `throttle` mode is not very consistent with this, we might have to interpret `throttle` as `mouseup` for purposes of rate-limiting server updates.

bryevdv:
- [x] issues: fixes #4540
- [x] tests added / passed

This adds `value_throttled` to sliders, which obeys the callback policy for both JS and Python callbacks. Examples are updated to use this where appropriate, and additionally all slider examples are updated to use `value` or `value_throttled` for callbacks instead of the deprecated `callback` property.

### DIFF
--- a/bokeh/models/widgets/sliders.py
+++ b/bokeh/models/widgets/sliders.py
@@ -80,6 +80,8 @@ class AbstractSlider(Widget):
 
     callback = Instance(Callback, help="""
     A callback to run in the browser whenever the current Slider value changes.
+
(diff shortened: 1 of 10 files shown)


bryevdv:
@mattpap I'd like to merge this tomorrow unless you want more time

PR #8922, opened 2019-05-20
Color mapping improvements

mattpap:
This is a cleaned up version of PR #8765. @jlstevens, I started implementing some of the changes we discussed on Friday.


bryevdv:
@mattpap @jlstevens can #8743 and #8765 be closed in favor of this, then?

jlstevens:
@bryevdv I think so, although afaik there are still some fundamental issues with how data can be made accessible to colorbars (unless @mattpap has figured out a suitable approach?)

bryevdv:
@jlstevens @mattpap can you two confer and figure out which if any of these three PRs should still be open?

bryevdv:
@jlstevens @mattpap what is the plan here? I do think it would be remiss not to fix some o the log mapper issues before 2.0

bryevdv:
closing old PR (left branch intact)

PR #8941, opened 2019-05-29
Make Tornado 5 the minimum version

### LINKED ISSUE #8935: Update tornado minimum version to 5.0

bryevdv:
- [x] issues: fixes #8935
- [x] tests added / passed

cc @bokeh/core I'd like to check with folks before just merging this one. By the time the 2.0 landing branch is merge, Tornado 5.0 will be ~18 months old. I am definitely strongly +1 on making 5.x the new minimum version as it integrates very naturally with `asyncio` which I suspect will be valuable to more users over a fairly short time span.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ jobs:
       env:
       - GROUP=unit
       - PYTHON=3.5
-      - TORNADO=4
+      - TORNADO=5
       - MINIMAL=1
 
     - install: scripts/ci/install.test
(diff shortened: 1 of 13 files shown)


bryevdv:
@mattpap has indicated a +1 as well. @philippjfr @canavandl @birdsarah any thoughts or objections? If not I will plan to merge this in a few days without comment.

birdsarah:
No opinions. Seems sensible.

bryevdv:
OK I I think there is enough agreement to proceed. As a side effect we should be able to get rid of all the ugly `gen.coroutine` decorators and use `async/await` throughout

PR #8955, opened 2019-05-30
Clean up Rect and its validation in spatial indexing

mattpap:
Previously `Rect` was defined as `{minX, maxX, minY, maxY}`, which was what rbush/flatbush uses. Now it's defined as `{x0, x1, y0, y1}`, which better aligns with the rest of bokehjs. Besides this, I removed unnecessary calls to `validate_bbox_coords()`, because `SpatialIndex` does normalization anyway.

### DIFF
--- a/bokehjs/src/lib/core/types.ts
+++ b/bokehjs/src/lib/core/types.ts
@@ -29,20 +29,13 @@ export type Size = {
 }
 
 export type Box = {
-  left: number
-  top: number
+  x: number
+  y: number
   width: number
   height: number
 }
 
 export type Rect = {
-  minX: number
-  minY: number
-  maxX: number
-  maxY: number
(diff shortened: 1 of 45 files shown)


bokehjs/src/lib/core/types.ts
```diff
@@ -29,20 +29,13 @@ export type Size = {
 }
 
 export type Box = {
-  left: number
-  top: number
+  x: number
+  y: number
```
bryevdv: whats the impetus for changing the convention here? `top` and `left` are much more descriptive to me, and match css names

mattpap: Because it's not given that `x,y` is actually `top,left`, it depends on the usage.

PR #8960, opened 2019-05-31
Use superscript exponents

### LINKED ISSUE #8954: Adding support for superscript exponents

micrenda:

```
```

- [x] issues: fixes #8954
- [ ] tests added / passed


bokehjs/src/lib/models/formatters/log_tick_formatter.ts
```diff
@@ -43,11 +45,19 @@ export class LogTickFormatter extends TickFormatter {
...
     const labels: string[] = new Array(ticks.length)
     for (let i = 0, end = ticks.length; i < end; i++) {
-      labels[i] = `${base}^${ Math.round(Math.log(ticks[i]) / Math.log(base)) }`
+	  
+      if (!this.use_superscript_exponents) {
+		labels[i] = `${base}^${Math.round(Math.log(ticks[i]) / Math.log(base))}`
+	  } else {
+	    labels[i] = `${base}${(Math.round(Math.log(ticks[i]) / Math.log(base))).toString().replace(/[0-9\+\-]/g,  function(match){return translate[match]})}`
+	  }
+
```
mattpap: Whitespace should be cleaned up and redundant curly brackets removed. `Math.round(Math.log(ticks[i]) / Math.log(base))` should be extracted into a variable. Use arrow function instead of `function() {}`.

mattpap:
This needs to be showed in an example and/or documentation, so that we can visually inspect workings of this code and allow it to be tested under our image diff CI.

bokehjs/src/lib/models/formatters/log_tick_formatter.ts
micrenda: Thank you for your suggestions. I am not an experienced JavaScript developer so your suggestions were very useful to me. I applied the suggested changes.

micrenda:
Unfortunately, my modifications seem to break the testing chain (in particular, the export to svg function). I think the superscript chars can not to be converted to 'latin-1` and this may cause problems to svg export.

Feel free to reject my pull request :(

bryevdv:
@micrenda I'm not quite ready to give up :) It's entirely possible there is a simple solution, or possibly an unreasonable assumption in those tests that can be relaxed. I'd like to take a close look, but FYI it will be next week sometime.

bryevdv:
The proximate cause of these test failures would seem to be these two places in the codebase that specify a `latin-1` encoding:
>(base) ❯ grip "latin-1"
./bokeh/_testing/plugins/file_server.py:
   65 :             html = open(os.path.join(HTML_ROOT, path), 'r', encoding='latin-1')
./bokeh/io/export.py:
  235 :             file.write(html.encode('latin-1'))

I will have to try to look at the commit history and see if there is any immediate justification for why that encoding needs to be set.

micrenda:
Hello, I tried to replace latin-1 with utf-8 but the problem persisted. I will study better the test log file to understand which files launches the exception.

micrenda:
I think I found the issue.

In file bokeh/io/export.py we have this code:
```
from six import b
...
with _tmp_html() as tmp:
        html = get_layout_html(obj, **kwargs)
        with io.open(tmp.path, mode="wb") as file:
            file.write(b(html))
```
The problem is that the [six.b](https://six.readthedocs.io/#six.b) function encode the html with latin-1 charset. I replaced this occurence with:
```
with _tmp_html() as tmp:
        html = get_layout_html(obj, **kwargs)
        with io.open(tmp.path, mode="wb") as file:
            file.write(html.encode('utf-8'))
```
and now it seems fine.

My only concern is if this maintains the python2/3 compatibility.

Regards

bryevdv:
> My only concern is if this maintains the python2/3 compatibility. 

@micrenda We are dropping Python 2 support in a few months in Bokeh 2.0 We already have a long-running `landing-2.0` branch that has removed `six` and string operations for Python 2. I don't want to confuse the process by having to re-do separate part of that. Why don't you re-submit the PR against the `landing-2.0` branch with only Python 3 support and we will plan to include it when Bokeh 2.0 is released.

micrenda:
@bryevdv I think it is a good idea. My workaround was not very elegant and it makes more sense to do as you suggest.

micrenda:
Hello,

As suggested, I created a [new pull request](https://github.com/bokeh/bokeh/pull/9003) based on `Landing 2.0`.

You can reject this pull request!

PR #8965, opened 2019-06-02
Deprecations for 2.0

### LINKED ISSUE #8964: remove 2.0 deprecations
Everything except tile provider stuff, which will be in a separate PR.

bryevdv:
- [x] issues: fixes #8964

This does not quite removed everything: 

* tile providers will go on their own PR
* remove `.callback` throughout will take some work, so separate PR there too

@mattpap I would appreciate a review of this since it attempts to remove all the coffee script machinery, also I want to confirm that the `Dropdown` properties really should be removed.


bokeh/util/compiler.py
```diff
@@ -278,8 +241,6 @@ def implementation(self):
         if isinstance(impl, str):
             if "\n" not in impl and impl.endswith(exts):
                 impl = FromFile(impl if isabs(impl) else join(self.path, impl))
-            else:
-                impl = CoffeeScript(impl)
```
bryevdv: Should this default to `TypeScript` now?

bryevdv:
@mattpap please take a look tomorrow. I would like to merge tomorrow night (keeping it rebased on `landing-2.0` is a slight chore)

bokeh/util/compiler.py
mattpap: Yes.

bokehjs/src/compiler/compile.ts
```diff
@@ -240,7 +203,7 @@ const compile_and_resolve_deps = (input: {code: string, lang: string, file: stri
 if (argv.file != null) {
   const input = {
     code: fs.readFileSync(argv.file as string, "utf-8"),
-    lang: (argv.lang as string | undefined) || "coffeescript",
+    lang: argv.lang as string,
```
mattpap: The original needs to be restored with `coffeescript` replaced with `typescript`.

PR #8968, opened 2019-06-02
Restrict CDS property changes

bryevdv:
`s.selected` is now read only

`s1.data = s2.data` now throws an error

### DIFF
--- a/bokeh/core/property/bases.py
+++ b/bokeh/core/property/bases.py
@@ -90,9 +90,12 @@ class Property(PropertyDescriptorFactory):
     # This class attribute is controlled by external helper API for validation
     _should_validate = True
 
-    def __init__(self, default=None, help=None, serialized=True, readonly=False):
+    def __init__(self, default=None, help=None, serialized=None, readonly=False):
         # This is how the descriptor is created in the class declaration.
-        self._serialized = False if readonly else serialized
+        if serialized is None:
+            self._serialized = False if readonly else True
+        else:
+            self._serialized = serialized
         self._readonly = readonly
         self._default = default
         self.__doc__ = help
--- a/bokeh/core/property/descriptors.py
+++ b/bokeh/core/property/descriptors.py
@@ -93,7 +93,7 @@ from copy import copy
 # External imports
 
 # Bokeh imports
-from .wrappers import PropertyValueContainer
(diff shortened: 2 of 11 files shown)


bryevdv:
The integration test failures seem to stem from applying `readonly` to `Instance`. Things seems to work as expected in a local `ipython` session, but not syncing selection indices correctly in a bokeh server context. 

I believe what is going on is that although in a non-server case the default `Selection` shows up, this default is not sent to the client, so then when the client updates it, it is not actually around on the Python side? 
```
ERROR:bokeh.server.protocol_handler:error handling message Message 'PATCH-DOC' (revision 1) content: {'events': [{'kind': 'ModelChanged', 'model': {'type': 'Selection', 'id': '4EDD878799824B1D8C67C2F850DEF343'}, 'attr': 'indices', 'new': [2]}], 'references': []}: RuntimeError('Cannot apply patch to 4EDD878799824B1D8C67C2F850DEF343 which is not in the document')
```

mattpap:
If I recall correctly, any model instances created in defaults on bokehjs side not synced with the server, because models' ability to sync changes is not available at the time defaults are resolved. So far this was worked around by making sure that all models are instantiated on the Python side.

bryevdv:
Yah so `_get` does this:
```
if self.name not in obj._property_values:
    return self._get_default(obj)
```
i.e. if a value is not set, it returns any default for that value. That works fine enough for basic types, but does not for Bokeh models. I'll have to think about this some. I definitely would like to make `.selected` be readonly.

bryevdv:
Ok, this is because `Property.__init__` explicitly sets `serialized` to `False` when `readonly` is `True`. An immediate quick-fix is to override the override:

```
diff --git a/bokeh/core/property/instance.py b/bokeh/core/property/instance.py
index 869241d41..8f7b0e941 100644
--- a/bokeh/core/property/instance.py
+++ b/bokeh/core/property/instance.py
@@ -62,6 +62,8 @@ class Instance(Property):

         super().__init__(default=default, help=help, readonly=readonly)

+        self._serialized = True
+
     def __str__(self):
         return "%s(%s)" % (self.__class__.__name__, self.instance_type.__name__)
```

But I will make something better, e.g. only override `serialized` it it is actually not explicitly set at all, but leave it alone if it is set explicitly.

PR #8974, opened 2019-06-03
Use native coroutines

bryevdv:


bryevdv:
I will re-open this later (perhaps even after landing-2.0 has been merged)

PR #8975, opened 2019-06-03
remove selection compat dict

bryevdv:


bryevdv:
@mattpap one BokehJS baseline is now failing:
```
diff --git a/bokehjs/test/baselines/Widgets__should_allow_Select b/bokehjs/test/baselines/Widgets__should_allow_Select
index 23ee7fd..1b4f24b 100644
--- a/bokehjs/test/baselines/Widgets__should_allow_Select
+++ b/bokehjs/test/baselines/Widgets__should_allow_Select
@@ -1 +1 @@
-Select bbox=[0, 0, 300, 20]
+Select bbox=[0, 0, 300, 31]
```
I don't see offhand how the changes in this PR would affect that. Any ideas?

PR #8980, opened 2019-06-05
Allow to offset plot frame's side panels

### LINKED ISSUE #8904: Allow to offset plot frame's side panel annotations

mattpap:
This is a counter-proposal to PR #8905. Here it's possible to offset any kind of side panel. It would be nice to just reuse layout's `Spacer` model, but our models are not reusable enough at this point to allow this. Example:

image

fixes #8904


bryevdv:
@canavandl do you have any comments? I'd say this seems like a reasonable feature on its own and would buy some time to figure out a specific spelling for titles after the current positioning work is done.

canavandl:
I'm ambivalent about this PR. My use case in the parent issue (#8904) is to be able to configure the title offset via theme, but this solution requires adding a Gap annotation to any plots.

Maybe I do too much web now, but I'm unconvinced that a combination of line_height + text_baseline/_vertical align isn't a better solution.

bryevdv:
I'm swayed by the utility of being able to theme a new property. Which is not to say something like this should not also be added, but let's do it in the context of making `Spacer` more useful rather than adding a new model.

PR #8986, opened 2019-06-09
simplify slider callback polic

bryevdv:
This PR removes the following properties from all Sliders:

* `callback`
* `callback_throttle`
* `callback_policy`

Instead users should use standard `on_change` and `js_on_change` with `value` and `value_throtted`


p-himik:
Just noticed this. Is there any discussion available? I was using the default throttle of 200ms to avoid costly computations on changes of a slider with many steps.

bryevdv:
@p-himik you can switch to `value_throttled` then? I can't did the discussion but the gist was

* Complicated timed mode can't be made reliably consistent for the server case, and
* Not having standalone/server case be consistent was a large source of support questions

bryevdv:
That said, at this point we could consider adding a new property that modifies only the `value_throttled` property behavior. That would not be a breaking change and would keep  default usage working simply out of the box in all scenarios.

p-himik:
Yeah, `value_throttled` by itself is not as useful because of the obligatory `mouseup` policy.

There's another consideration - it's useful to have separate throttle policies for JS callbacks and for Python callbacks. In my case, I'd like to have `mouseup` for Python callbacks and debounce for JS callbacks.

This makes me think - maybe it makes sense to just have the `value` property and provide throttle policy as an argument for the `on_change` and `js_on_change` functions? I can definitely see situations where different JS callbacks have different debounce values.

bryevdv:
That's also fine with me as long as `mouseup` is the default policy for `value_throttled`, and I'd say it's worth considering if it should only be on the JS version. There are lots of things (server load and network latency, e.g.) that could cause a time guarantee like that not to be honorable. Maybe better to just no include it on `on_change` at all.

p-himik:
Not sure about switching only to the JS side. It's fine for my projects in particular, but in general people also use notebooks where a debounce value of e.g. 200ms can be perfectly fine.

Given a questionable nature of this feature, a completely different way to approach this is to leave it as is and maybe write a Discourse post about ways to debounce callback calls with a particular delay. And the writing of such a post can be delayed up to a point when someone actually asks about it. :)

PR #8988, opened 2019-06-10
rev nwmatcher version

bryevdv:
This PR stipulates `nwmatcher >= 1.4.4` to address GH reported vulnerability. 

Note that this package is only used for development, and does not impact normal user usage.


bryevdv:
@mattpap see also #8987 and choose which is better.

mattpap:
Merged PR #8987 instead.

PR #8989, opened 2019-06-10
Modularize CSS

### LINKED ISSUE #8984: Modularize CSS
The idea is to drop `*.css` bundles in favor of including individual `*.css` files as proper TypeScript modules (with CSS class names exported for improved static safety).

mattpap:
In this PR each CSS "module" contains of a `*.less` file under `src/less` and an accompanying `*.ts` module under `src/lib/styles`. Such `*.ts` include `import "module_name.css"` to reference a generated CSS file for inclusion. It also may include exported CSS class names (to aid static safety). This import is then rewritten by an AST preprocessor to include the generated CSS and a bit of code for adding it to the primary style sheet. `bokeh*.css` are no longer included in any generated HTML, but they still generated empty and tested. This will have to be removed for 2.0.

fixes #8984


bryevdv:
windows test failures seem to stem from
```
    def _inline(self, path):
        begin = "/* BEGIN %s */" % basename(path)
>       with open(path, 'rb') as f:
E       FileNotFoundError: [Errno 2] No such file or directory: 'C:\\projects\\bokeh\\bokeh\\server\\static\\js\\bokeh.min.js'
```

bryevdv:
which ultimately must be due to this build failure:
```
uilding BokehJS... Failed.
ERROR: 'node make build' returned the following
---- on stdout:
    [16:53:03] Starting 'styles:compile'...
    [16:53:04] Finished 'styles:compile' after 187 ms
    [16:53:04] Starting 'scripts:compile'...
    [16:53:21] Finished 'scripts:compile' after 16.83 s
    [16:53:21] Starting 'scripts:bundle'...
    [16:53:21] failed: can't resolve 'buttons.css' from 'C:\projects\bokeh\bokehjs\build\js\lib\styles\buttons.js'
    ```

**Comment by bryevdv** [comment]
@mattpap folks that want to tinker with CSS to tweak things are unaffected by this (e.g. someone today asked about changing the font color for slider titles, which does not have a property)

**Comment by mattpap** [comment]
> folks that want to tinker with CSS to tweak things are unaffected by this?

The structure of the CSS didn't change in any significant manner. The only think that may matter (possibly in some obscure scenarios), is having `<style>` element  (of the primary style sheet) auto-generated. However, this element is always added as the first element of the `<head>`, so wherever people added their styles, the order of CSS rules should be the correct.

**Comment by mattpap** [comment]
I added one more semi-related change given the churn of this PR. Setting `type` on models isn't necessary anymore. Previously:
```ts
class SomeModel extends Model {
  static initClass(): void {
    this.prototype.type = "SomeModel"
  }
}
SomeModel.initClass()
```
Now one can simply write:
```ts
class SomeModel extends Model {}
```
and `type` (or actually `__name__`) will be inserted during compilation. If you have to include explicit type for some reason (`Figure` is a good example), then the new syntax is:
```ts
class SomeModel extends Model {
  static __name__ = "SomeAlternativeModel"
}
```
The old approach is still allowed to aid backwards compatibility. Note that anonymous model classes (e.g. markers) and any models written in JavaScript will require manual definition of `__name__`.

If one wonders why we can't just use `constructor.name`, then it doesn't work in IE (this can be hacked around to some extent) and requires that any minification procedure applied to such code doesn't rewrite class/function names, which is tedious to assert and would result in much larger code than necessary.

bryevdv:
Ok, well I don't have the wherewithal to make more than a cursory review of 350 file PR. LGTM at a quick glance, examples report looks good. Will merge and get a dev build out so others can test with it directly.

PR #8990, opened 2019-06-12
Use scalar props where applicable

bryevdv:
@mattpap I want to keep looking for more places, but are you aware of any others besides these offhand that could be updated?


bryevdv:
I think just `arrow_heads.py`, `axes.py`, and `plots.py`

PR #8993, opened 2019-06-12
Fix keyboard events on range sliders

### LINKED ISSUE #8911: RangeSlider stuck when modified by pressing the keyboard arrows
When changing the lower value of a RangeSlider by pressing the keyboard arrows the RangeSlider gets stuck on NaNs.

image

Bokeh version: 1.1.0
Platform: Windows
Command line: bokeh serve app

xavArtley:
- [x] issues: fixes #8911
- [x] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/widgets/abstract_slider.ts
+++ b/bokehjs/src/lib/models/widgets/abstract_slider.ts
@@ -173,12 +173,11 @@ export abstract class AbstractSliderView extends ControlView {
         }
         handle[0].setAttribute('tabindex', '0')
         handle[0].addEventListener('keydown', keypress)
-      }
-      else if (handle.length==2) {
+      } else if (handle.length==2) {
         // Add range cursors keyboard events
         const keypress_lower = (e: KeyboardEvent): void => {
           const current = this._calc_to()
-          let value = current.value
+          const value = current.value
           switch (e.which) {
             case 37: {
               value[0] = Math.max(value[0] - step, start)
@@ -191,7 +190,7 @@ export abstract class AbstractSliderView extends ControlView {
             default:
               return
           }
-        
+
           this.model.value = value
           this.noUiSlider.set(value)
           this.model.properties.value.change.emit()
@@ -200,7 +199,7 @@ export abstract class AbstractSliderView extends ControlView {
         }
         const keypress_upper = (e: KeyboardEvent): void => {
           const current = this._calc_to()
-          let value = current.value
+          const value = current.value
           switch (e.which) {
             case 37: {
               value[1] = Math.max(value[1] - step, value[0])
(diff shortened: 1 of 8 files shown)


bryevdv:
@xavArtley Thanks for looking in to this. If I can impose on you slightly, I'd appreciate taking the opportunity to improve things a bit:

* the render methods is 100+ lines long. It should really be split up at least according to some helper methods for n=1 vs n=1 but possibly other helpers can be extracted as well. 

* another option is to move the n=1 vs n=2 keypress handling code into appropriate subclasses under the same name `_handle_keypress` or whatever, then `AbstractSlider` just calls this and it delegates appropriately based on the concrete type. This would be my preference but it is a little more work. 

* The code in the switch statements is basically repeated three times, it could also be pulled into a little helper.

xavArtley:
I added some integration tests, it works locally for me but it seems to fail on the CI build.

bryevdv:
I can take a look soon if you don't figure it out first :)

xavArtley:
I think I know what's going on. When both handle of the range_slider are supperposed the selenium webdriver select the wrong element

xavArtley:
I change tests to avoid handle overlapping however it continues to fail

bryevdv:
I'll try to take a look tonight cc @mattpap in case you have any ideas at a glance

xavArtley:
I finally found the right method

bryevdv:
Thank you for the hard work on this @xavArtley I will get it merged tonight

bryevdv:
@xavArtley When I try this on chrome with the `export_csv` example app, the keyboard arrows move the slider, and show updated values in the UI, but python callbacks in the server are not getting called until I actually scrub the slider with a mouse. Do you see similar behavior?

bokehjs/src/lib/models/widgets/abstract_slider.ts
```diff
@@ -114,6 +114,39 @@ export abstract class AbstractSliderView extends ControlView {
...
+      }
+      case 39: {
+        value[idx] = Math.min(value[idx] + step, high)
+        break
+      }
+      default:
+        return
+    }
+    if (is_range)
+      this.model.properties.value.change.emit()
```
bryevdv: In order to trigger a sync with Python there will need to be an actual, real assignment to `this.model.value`

xavArtley:
I do not really have any perspective on the conflicts that will be introduced with version 2.0. In any case it will be necessary to solve them. Here is a bug fix and it does not break the api. I think it may be interesting for people who are going to stay on version 1. of bokeh even after the release of version 2. to have the fix

bryevdv:
OK let's go ahead and merge. I will figure the conflicts out.

bryevdv:
@xavArtley I resolved the conflicts on the `landing-2.0` branch. But if you could look over the relevant files you changed, in the `landing-2.0` branch, just to have another pair of eyes look over things, it would be reassuring

PR #9000, opened 2019-06-14
Move tests out of tree

bryevdv:
Moving the test out of tree has some advantages:

* easier to ensure tests run against an actual installed package on CI (we definitely don't do that sometimes)
* smaller package size
* could probably get rid of "implicit mark" on pytest and possibly others (can select with `-k` against test directory names)

There are a few warts:

* pytest pukes if filenames are repeated so had to make a few ugly renames


bryevdv:
In fact now you can see the tests are trying to test the checkout:

>    with open(join(__path__[0], 'LICENSE.txt')) as lic:
> E       FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/bokeh/bokeh/bokeh/LICENSE.tx

I'll fix this tomorrow

mattpap:
Hopefully this isn't going to cause too big headache when merging this branch into `master` (assuming there will be changes to tests in the mean time). I will argue that this kind of changes are premature at this point and should happen after `landing-2.0` is merged.

bryevdv:
I've been keeping `landing-2.0` rebased up to date. I will hold off on merging this for now and start merging `landing-2.0` into it if rebasing proves to difficult. That said I'd like to get 2.0 out sooner rather than later. I will send out an email to the whole core team to discuss.

bryevdv:
This will be impossible. Saving for after the `landing-2.0` branch is merged but before 2.0 is released.

PR #9003, opened 2019-06-14
Landing 2.0 - Use superscript exponents

### LINKED ISSUE #8954: Adding support for superscript exponents

micrenda:

```
```

- [x] issues: fixes #8954
- [ ] tests added / passed

Based on Landing 2.0


bryevdv:
superseded by #9004

PR #9005, opened 2019-06-14
Added use_superscript_exponents

### LINKED ISSUE #8954: Adding support for superscript exponents

micrenda:

```
```

- [x] issues: fixes #8954
- [x] tests added / passed

Based on Landing 2.0

Be aware: I had to modify export.py to output html as utf-8


micrenda:
Sorry for the multiple pull request, but I had some problems with the previous one.

bokeh/models/formatters.py
```diff
@@ -278,6 +278,20 @@ class LogTickFormatter(TickFormatter):
     base to use. If unset, the formatter will use base 10 as a default.
     """)
 
+    use_superscript_exponents = Bool(False, help='''
+    Whether to ever display scientific values using exponent superscript.
+
+    Superscript fonts may not be rendered with the same size. Changing the major label
+    font may solve this issue.
```
bryevdv: I'd suggest:
> Some fonts may render superscripts at different sizes. Changing the ...

bokehjs/src/lib/models/formatters/log_tick_formatter.ts
```diff
@@ -24,6 +25,7 @@ export class LogTickFormatter extends TickFormatter {
   static initClass(): void {
     this.define<LogTickFormatter.Props>({
       ticker: [ p.Instance, null ],
+      use_superscript_exponents:   [ p.Boolean, false   ],
```
bryevdv: Since this is going in to 2.0 we could consider enabling this by default. Thoughts @micrenda @mattpap ?

bokeh/io/export.py
```diff
@@ -232,7 +232,7 @@ def get_svgs(obj, driver=None, timeout=5, **kwargs):
     with _tmp_html() as tmp:
         html = get_layout_html(obj, **kwargs)
         with io.open(tmp.path, mode="wb") as file:
-            file.write(html.encode('latin-1'))
+            file.write(html.encode('utf-8'))
```
bryevdv: I do want to still check on this a bit more. Setting to `latin-1` is atypical so we should make sure there was not some reason for it.

bokeh/models/formatters.py
bryevdv: If there is a built-in font that seems to work well across browsers we could also consider making it the default

bryevdv:
@micrenda Thanks for the new PR. I am excited about this, it's a nice upgrade that's been asked for for a long time. I left a few small comments. One other thing that would be helpful if yo have time: making a set of plots that show a few different built-in browser font options side by side, then getting screenshots of the comparison page on different browsers/platforms. It would be good to verify that there aren't any glaring holes in supported platforms (and if there aren't maybe it should just be turned on by default)

bokeh/models/formatters.py
micrenda: In my specific case (Debian 9 / XFCE)  I had very good results with this font:
`p.axis.major_label_text_font = "sans"`

But I don't know if it is a good idea to making it default. In next days I will try on my colleagues computers (macOS and, I hope, Windows).

bokehjs/src/lib/models/formatters/log_tick_formatter.ts
micrenda: I am not very sure if it will be appreciated by all. But, if we should do think correct, we may have to add also to implement it on 'basic_tick_formatter.ts' too, showing something like `2 ✕ 10⁻²⁷`

If you want I can implement it, and we will test it on this branch, to see if it looks good.

bokeh/io/export.py
micrenda: The changes from `latin-1` to 'utf-8' is not a light hearth one, for me. Usually there is a valid reason for this.

micrenda:
I think I can do it. We can have an idea how it is handled in most common scenarios.

bokehjs/src/lib/models/formatters/log_tick_formatter.ts
canavandl: I'm +1 on enabling it by default. I imagine the heavy majority of users prefer the superscript to the "^2" exponents

bryevdv: I'm tentatively +1 as well, I actually think almost all users would prefer it. But I would like to see a survey of how things look with different fonts on different browsers and platforms, so we can make an actual informed decision about what a good default font would be. 

Re: basic tick formatter, I would say let's hold off for now. Certainly if you want to work on it that's great but definitely it should be a separate PR from this one.

bokeh/io/export.py
bryevdv: Possibly :) It might also just be some historical accident. I can't find any documentation for the choice, unfortunately. All the current tests pass either way, I would just want to experiment locally with SVG export some in addition.

micrenda:
Hello, I added the support for the `basic_tick_formatter` and I performed some tests to see if the superscript are rendered well. Unfortunately, only the `sans` font seems to run smoothly.

Attached there is the script and the results.
[test1.zip](https://github.com/bokeh/bokeh/files/3302203/test1.zip)

I think it is not a good idea to have it on by defaults. It is not a [bokeh fault](https://graphicdesign.stackexchange.com/questions/66878/unicode-superscript-numbers-1-to-3-vs-4-and-above) but, if we use the current implementation, bokeh will surely receive a lot of complaints.

Another solution would be to use the '<sup></sup>' tag, but I don't think it is a good idea (we should add support to HTML axis labels and it could penalize performances).

bokehjs/src/lib/models/formatters/log_tick_formatter.ts
```diff
@@ -42,10 +44,19 @@ export class LogTickFormatter extends TickFormatter {
 
     const base = this.ticker != null ? this.ticker.base : 10
 
+    const translate: { [id: string]: string; } = { 0:"\u2070", 1:"\u00b9", 2:"\u00b2", 3:"\u00b3", 4:"\u2074", 5:"\u2075", 6:"\u2076", 7:"\u2077", 8:"\u2078", 9:"\u2079", "+":"\u207a", "-":"\u207b" }
```
bryevdv: @micrenda if we are going to do both tick formatters in this PR, the `translate` value should be moved to a single place and imported as necessary rather than duplicated

bryevdv:
@micrenda I am going to go ahead and close this for now. I do hope we will be able to pursue adding proper MathText support in the near-ish future, in which case this specific capability will fall out naturally. Thank you for the PR and the productive discussions!

micrenda:
Indeed, I agree. The right way to do it is via MathTex or adding the support for the <sup> tag in axis labels.

Unfortunately the current implementation work only with a single font.

PR #9010, opened 2019-06-16
Add support for django channels

### LINKED ISSUE #9009: Add support django channels
The idea is to allow using bokeh applications in django without a need for embedding a tornado server.

mattpap:
This is a very preliminary approach. You can try this out in `examples/howto/server_embed/django_embed` example.

fixes #9009


bryevdv:
Every file in here is new. Given that, I think we should discuss making a separate project for this, i.e. `bokeh/bokeh-django` for this. A separate project could have real dependencies on Django and channels. 

cc @philippjfr @jlstevens @jbednar

swainn:
@mattpap @philippjfr @sdc50 @msouff Here's the pseudo-code example illustrating panel/bokeh use in a Django app. This is by no means set in stone, but meant to stimulate the discussion to make sure everyone is on the same page and aiming for the same target: https://github.com/Aquaveo/django_bokeh_app

bryevdv:
@mattpap did you say this is `ready`?

bokeh/server/django/__init__.py
```diff
@@ -0,0 +1,14 @@
+import six
```
bryevdv: Needs boilerplate header

examples/howto/server_embed/django_embed/django_embed/templates/embed.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Embedding a Bokeh Apps In Django via autoload.js</title>
```
bryevdv: Apps -> App

examples/howto/server_embed/django_embed/manage.py
```diff
@@ -0,0 +1,21 @@
...
+
+def main():
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'django_embed.settings')
+    try:
+        from django.core.management import execute_from_command_line
+    except ImportError as exc:
+        raise ImportError(
+            "Couldn't import Django. Are you sure it's installed and "
+            "available on your PYTHONPATH environment variable? Did you "
+            "forget to activate a virtual environment?"
```
bryevdv: "conda or virtual" ?

examples/howto/server_embed/django_embed/django_embed/views.py
```diff
@@ -0,0 +1,69 @@
...
+from bokeh.themes import Theme
+from bokeh.embed import server_document
+
+from bokeh.sampledata.sea_surface_temperature import sea_surface_temperature
+
+from .shape_viewer import shape_viewer
+
+theme = Theme(filename=join(settings.THEMES_DIR, "theme.yaml"))
+
+def shape_viewer_handler(doc: Document) -> None:
```
swainn: Is the Channels scope and/or Django request object available through this `doc` object? One of the big motivations for integrating Bokeh with Django would be to leverage users and permissions. Access to the currently logged-in user is provided in the Channels scope and/or Django request object. Even if it is available through the `doc` object, it would be better to pass it in as a separate argument.

mattpap: The `doc` is just bokeh's document, but I will add additional arguments to allow accessing the `request` and URL arguments (if provided).

examples/howto/server_embed/django_embed/manage.py
mattpap: That was auto-generated by django. In any case, conda environment is just a implementation of virtual environment concept, so this language seems sufficient to me.

bryevdv:
@mattpap is this `ready`?

swainn:
@mattpap, we are going to be taking a closer look at this next week and try to get our example working. Can we hold off merging until then?

mattpap:
@swainn, sure we can hold off. However, this PR is good enough for merging at this point. New PRs can be started to get the remainder of the features and possibly a refined API in place.

swainn:
> However, this PR is good enough for merging at this point.

I'd say without the `request` being passed into the handler functions, this falls short of full integration with Django. And from my experience, once merged, there will be much less motivation to get it done. @sdc50 what do you think?

examples/howto/server_embed/django_embed/django_embed/views.py
```diff
@@ -0,0 +1,76 @@
...
+    doc.theme = theme
+    doc.add_root(column(slider, plot))
+
+def with_request(f):
+    def wrapper(doc):
+        return f(doc, doc.session_context.request)
+    return wrapper
+
+@with_request
+def sea_surface_handler_with_template(doc: Document, request: Any) -> None:
```
mattpap: @swainn, after all, the only reasonable way of accessing the request (or scope in channels' terms) was to make it available as a part of the document (`doc.session_context.request`), at least without bigger changes to bokeh server. With a simple decorator this can be nicely hidden away, like in this example. I can make this decorator a part of django server API if we settle on this approach.

bryevdv:
Does Django provide or enforce any kind of session affinity? Just noting that if not, storing the request seems likely to fall over as soon as there are multiple processes behind a load balancer (for all the same reasons it falls over with tornado: there's no guarantee the WS request ends up at the same process as the original request)

bryevdv:
I'm actually in the process of trying to improve the auth/security story around Bokeh, e.g. with exportable session ids, sending session id's in cookies, and (in the tornado case) affording a user hook for `get_current_user` so that that an `@authenticated` decorator can just be added on all the relevant request handlers. 

 FWIW I think having Bokeh app code directly handle auto concerns is a Bad Idea. The request handlers should shut down anything immediately, before things ever get to the point of creating a session at all. Which is why I am dubious of an approach that requires making a request (or channel or whatever) available to the actual Bokeh app in order to function. I really wish the requirements for this work had been worked up and published some where public ahead of time so they could have been discussed more broadly.

examples/howto/server_embed/django_embed/django_embed/views.py
swainn: @mattpap this will work for us. The decorator is great for convenience, but I don't mind accessing it on the `doc` object directly either.

msouff:
Hi gentlemen, we have reviewed this and, plus the inclusion of the request object access, it covers everything we needed. Great work.

swainn:
> Hi gentlemen, we have reviewed this and, plus the inclusion of the request object access, it covers everything we needed. Great work.

@mattpap @bryevdv @sdc50 this looks good on our end. Thank you for your patience in allowing us to review before merging.

swainn:
@mattpap @bryevdv any idea when this will be released/what version?

mattpap:
@swainn, bokeh 1.4, but we don't have a date set yet. Though we want to push it out sooner than later, so I suppose maybe in two weeks.

bryevdv:
I would say wk1 Sept is more likely. I very much want to get in all the things I put in the milestone but my bandwidth is limited this month. We can easily make a dev build and provide instructions for installing it, on the explicit understanding that *BokehJS resources for dev builds are not permanently hosted and will disappear in the future (dev builds are for **testing only**)*

swainn:
@bryevdv a dev build would be great.

Issue #8981, opened 2019-06-05
[FEATURE] implement hover anchor on more glyphs


kylrth:
#### Software version info

```
Bokeh 1.0.4
Python 3.7.3
jupyter_client 5.2.4
Windows 8 (64-bit)
Google Chrome 74.0.3729.169 (64-bit)
```

#### Description of expected behavior and the observed behavior

By specifying the keyword argument `anchor='center_right'` to `bokeh.models.HoverTool`, I expect to see the tooltips pointing to the right edge of each bar. Instead, the tooltips still point to the center of each bar.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import bokeh
from bokeh.plotting import figure

bokeh.io.output_notebook()

source = bokeh.models.ColumnDataSource(dict(
    lengths=[1, 2, 3, 4, 5],
    pos=[1, 2, 3, 4, 5]
))

hover = bokeh.models.HoverTool(
    tooltips=[
        ('pos', '@pos'),
        ('length', '@lengths')
    ],
    anchor='center_right'
)

fig = figure()
fig.hbar(y='pos', right='lengths', height=0.7, source=source)
fig.add_tools(hover)

bokeh.io.show(fig)

# bokeh.io.curdoc().add_root(fig)  # this also happens on the Bokeh server
```

#### Browser JavaScript console output when running `bokeh serve`

```
bokeh-1.0.4.min.js:31 [bokeh] setting log level to: 'info'
bokeh-1.0.4.min.js:31 [Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952
t @ bokeh-1.0.4.min.js:31
t.initialize @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
i.build_views @ bokeh-1.0.4.min.js:31
t.build_child_views @ bokeh-1.0.4.min.js:31
t.initialize @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
t @ bokeh-1.0.4.min.js:31
e @ bokeh-1.0.4.min.js:31
i @ bokeh-1.0.4.min.js:31
i.add_document_standalone @ bokeh-1.0.4.min.js:31
[...]

PR #9013, opened 2019-06-19
implement hover anchor for quad glyphs

kylrth:
- [x] issues: fixes [#8981](https://github.com/bokeh/bokeh/issues/8981)
- [ ] tests added / passed

I am unsure of the kinds of tests I should include here. I'm thinking a [screenshot integration test](https://bokeh.pydata.org/en/latest/docs/dev_guide/testing.html#id1)? I haven't written it yet but I wanted to post this PR first so I don't get too far in the wrong direction.

### DIFF
--- a/bokehjs/src/lib/models/glyphs/box.ts
+++ b/bokehjs/src/lib/models/glyphs/box.ts
@@ -1,5 +1,6 @@
 import {LineVector, FillVector, HatchVector} from "core/property_mixins"
 import {Arrayable, Rect} from "core/types"
+import {Anchor} from "core/enums"
 import {Line, Fill, Hatch} from "core/visuals"
 import {SpatialIndex} from "core/util/spatial"
 import {Context2d} from "core/util/canvas"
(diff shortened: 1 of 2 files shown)


bryevdv:
@kylrth I think this code is already on `Quad` correct? I think moving it to `Box` is perfectly sensible but we should remove the now-superfluous code from `Quad` at the same time. 

Regarding tests, it's unfortunately not the easiest thing. Depending on how much effort you can/want to put in there is a few possibilities: 

* I would accept the PR without tests as long as you add screenshots of the different modes functioning in a comment in the PR as documentation that it functions. 

* You could factor all of the logic out to a utility function that accepts all the necessary args, that the view `get_anchor_point` method delegates to, then write a simple unit test for the function.

* A real integration (selenium) test would look like this: Loop over the anchor point values, create a plot with one big quad and that anchor value set. Add a over tool with the `$rx` and `$ry` special values to display where the over tooltip is anchored. move the mouse over the anchor over the quad. Do a DOM inspection to verify the screen coordinates in the tooltip are as expected. As you can see, there's unfortunately not any existing hover tool integration tests yet: https://github.com/bokeh/bokeh/tree/master/tests/integration/tools

kylrth:
I wish I had more time to help, but I think my barrier to entry with TypeScript and Selenium tests is too high for the time I have. Here is my test file:

```python
"""Demonstrate functionality of HoverTool anchors.

Kyle Roth. 2019-06-18.
"""


import bokeh
from bokeh.io import output_file, save, curdoc

source = bokeh.models.ColumnDataSource(dict(
    lengths=[1, 2, 3, 4, 5],
    pos=[1, 2, 3, 4, 5]
))

anchors = (
    'top_left', 'top_center', 'top_right',
    'bottom_left', 'bottom_center', 'bottom_right',
    'center_left', 'center', 'center_right'
)

figs = []
for anchor in anchors:
    hover = bokeh.models.HoverTool(
        tooltips=[
            ('pos', '@pos'),
            ('length', '@lengths')
        ],
        anchor=anchor
    )

    fig = figure(title='Anchor: {}'.format(anchor))
    fig.hbar(y='pos', right='lengths', height=0.7, source=source)
    fig.add_tools(hover)

    figs.append(fig)

if __name__.startswith('bk_script'):
    curdoc().add_root(column(figs))
else:
    output_file("test.html")
    save(column(figs))
```

Curiously, `bokeh serve test.py` renders the tooltips correctly but `python test.py && google-chrome test.py` does not. Any thoughts?

bryevdv:
> Any thoughts?
 Bokeh server apps always use BokehJS resources from the server itself, so local dev version if you are running from there. Bokeh standalone output always uses BokehJS resources from CDN by default, so, the last released version. Would need to explicitly set `BOKEH_RESOURCES=inline` (or other methods) to use locally built BokehJS

kylrth:
2f153f7aa794c0ad5ca2ef6fe8583247a6fceda0 is optional

kylrth:
bokeh1
bokeh2
bokeh3
bokeh4
bokeh5
bokeh6
bokeh7
bokeh8
bokeh9

bryevdv:
Thanks @kylrth !

gargumang411:
What if we wanna add these at the top left of the chart rather than top left of the bars? can we do that?

bryevdv:
No, the hover tooltips currently always follow the cursor (position relative to cursor depends on a few factors). I believe some people may have have tried tinkering with the CSS directly to "anchor" the tooltips in other locations but that's all I know about that.

PR #9030, opened 2019-06-23
replace remote datasource with web datasource

### LINKED ISSUE #9029: Simplify remote data sources
When creating the server sent events data source, we noticed that we had made an assumption made that all remote data sources should have polling interval defined. This was not the case with server sent events and we decided to create an new layer for web events that do not have polling interval embedded in them. This caused us to have a bit too many abstract classes with little use.

This is a proposition to remove unneeded web data source, since the code base will be a little bit simpler at the moment when defining the polling interval in the ajax data source.

alhoo:
- [x] issues: fixes #9029


bryevdv:
@alhoo sorry I should have mentioned more specifically: we can't merge this in to master, it has to be merged in to the `landing-2.0` branch. I tried changing the base here but there are some conflicts. It would be cleaner and safer if you re-do this PR against the `landing-2.0` branch yourself. Basically, these steps:
```
git checkout landing-2.0

git checkout -b new_pr_branch

# make/commit changes

# push to GH and select `landing-2.0` when the PR is opened
```

bryevdv:
Thanks @alhoo!

PR #9037, opened 2019-06-29
Added code to handle document generation using Thread Executor

sridk:
Modified code in contexts.py to handle document generation for new session using ThreadExecutor from concurrent.futures library


bokeh/server/contexts.py
```diff
@@ -212,7 +213,12 @@ def create_session_if_needed(self, session_id, request=None):
             except Exception as e:
                 log.error("Failed to run session creation hooks %r", e, exc_info=True)
 
-            self._application.initialize_document(doc)
+            # using ThreadPoolExecutor create a thread that initializes new session document
+            try:
+                doc_initializer = concurrent.futures.ThreadPoolExecutor(max_workers=1)
+                doc_initializer.submit(self._application.initialize_document, doc)
+            except concurrent.futures.BrokenExecutor as e:
+                log.error("ThreadExecutor failed to create document for session with ID", session_id, e)
```
p-himik: Few issues with the code:
- The executor could be reused and initialized with a larger pool. Otherwise, you could just use `Thread` instead
- Now any exceptions thrown from `initialize_document` are ignored and not even logged (I think) since the future is resolved nowhere
- The immediate use of `ServerSession` below modifies the document without any logs - it can lead to some hard to catch concurrency issues
- `ServerSession` assumes that the document has already been initialized - it processes the document's callbacks

Maybe there are more potential issues somewhere down the line, there the result of `create_session_if_needed` is processed.

bryevdv:
@sridk thanks for getting this started! I agree with @p-himik comments. A few more details:

* Tornado has [`IOLoop.set_default_executor`](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.set_default_executor) which could be set  to a `ThreadPoolExecutor ` when the server starts up. We could potentially even make the `max_workers` configurable, though I would not insist on that in this PR. 

* Then, I think in this code you would then want to use [`IOLoop.run_in_executor`](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.run_in_executor)
   ```
   yield IOLoop.current().run_in_executor(None, 
                                          self._application.initialize_document, 
                                          doc)
   ```
   which will use the default `ThreadPoolExecutor ` configured earlier. The `yield` inside this coroutine means that the following code will not pick up until the the Future is actually resolved (but the rest of the app will not block in the mean time)

* Agree that some more `log.debug` statements around here would be very helpful

@p-himik does this all sound about right?

p-himik:
@bryevdv Sure! Right now I can't think of any potential problems.

sridk:
@bryevdv @p-himik that feedback was very helpful! I will push modified code to this pull request soon

bryevdv:
I believe the build failure on windows is due to a new major version of pytest

bryevdv:
@sridk I just realized `run_in_exectuor` is only in Tornado > 5 so this will need to go into the `landing-2.0` branch for the upcoming  Bokeh 2.0. I will attempt to edit base branch for the PR later today

sridk:
@bryevdv okay.
If you want me to make a PR for `landing-2.0`, I can do that.

bryevdv:
@sridk OK please do that, I tried to change the base on the GH web UI but the results did not fill me with confidence. Thanks!

PR #9039, opened 2019-07-01
fixed analysis

0xflotus:
I only fixed a typo


bryevdv:
Thanks for the catch and the fix @0xflotus !

bryevdv:
I believe the build failure on windows is due to a new major version of pytest

PR #9041, opened 2019-07-02
Added code to set IOLOOP default executor to ThreadPoolExecutor and u…

sridk:
Document creation for new session using ThreadPoolExecutor inside tornado IOLOOP


bryevdv:
@sridk OK, well it seems I was too optimistic/aspirational about how simple this would be. After testing out this change with some real streaming examples (ohlc, fourier_animated, etc), problems definitely turned up. I will try to look at this later this week to see if there is a short path to fixing things up. (Or if you figure things out first, certainly push new changes)

sridk:
@bryevdv okay.
I ran the code against fourier_animated, but I did not know what kind of problems to look for. 
Could you please elaborate on the problems so I can try to fix them.
Thanks!

bryevdv:
> I ran the code against fourier_animated, but I did not know what kind of problems to look for.
Could you please elaborate on the problems so I can try to fix them.

In order to simulate real heavy usage I started the code in one browser (works fine) then opened in another browser and hit reload very fast repeatedly to simulate opening sessions while another is running. And this is where you will see errors crop up. The main issue is that the code after the `yield` does not condition on the session id (it never needed to) but depending on which thread finishes first, the session and document may not match up. There is also an intermittent issue where threads do not automatically start their own IOLoop anymore in Tornado >=5 which can cause another error.  

Here is a diff that *almost* works:

```
diff --git a/bokeh/server/contexts.py b/bokeh/server/contexts.py
index 2a619e6b1..a26223b1e 100644
--- a/bokeh/server/contexts.py
+++ b/bokeh/server/contexts.py
@@ -181,6 +181,34 @@ class ApplicationContext(object):

         self.server_context._remove_all_callbacks()

+    @gen.coroutine
+    def _create_session(self, session_id, request=None):
+        doc = Document()
+
+        session_context = BokehSessionContext(session_id,
+                                              self.server_context,
+                                              doc)
+        # using private attr so users only have access to a read-only property
+        session_context._request = _RequestProxy(request)
+
+        # expose the session context to the document
+        # use the _attribute to set the public property .session_context
+        doc._session_context = session_context
+
+        try:
+            yield yield_for_all_futures(self._application.on_session_created(session_context))
+        except Exception as e:
+            log.error("Failed to run session creation hooks %r", e, exc_info=True)
+
+        self._application.initialize_document(doc)
+
+        session = ServerSession(session_id, doc, io_loop=self._loop)
+        self._sessions[session_id] = session
+        session_context._set_session(session)
+        self._session_contexts[session_id] = session_context
+
+        raise gen.Return(session)
+
     @gen.coroutine
     def create_session_if_needed(self, session_id, request=None):
         # this is because empty session_ids would be "falsey" and
@@ -190,48 +218,19 @@ class ApplicationContext(object):

         if session_id not in self._sessions and \
            session_id not in self._pending_sessions:
-            future = self._pending_sessions[session_id] = gen.Future()
-
-            doc = Document()
-
-            session_context = BokehSessionContext(session_id,
-                                                  self.server_context,
-                                                  doc)
-            # using private attr so users only have access to a read-only property
-            session_context._request = _RequestProxy(request)
-
-            # expose the session context to the document
-            # use the _attribute to set the public property .session_context
-            doc._session_context = session_context
-
-
             try:
-                yield yield_for_all_futures(self._application.on_session_created(session_context))
-            except Exception as e:
-                log.error("Failed to run session creation hooks %r", e, exc_info=True)
-
-            # using IOLOOP's default Thread executor create a thread to initialize new session document
-            try:
-                yield self._loop.run_in_executor(None, self._application.initialize_document, doc)
+                self._pending_sessions[session_id] = yield self._loop.run_in_executor(None, self._create_session, session_id, request)
             except Exception as e:
                 log.error("ThreadExecutor failed to create document for session with ID", session_id, e)

-            session = ServerSession(session_id, doc, io_loop=self._loop)
-            del self._pending_sessions[session_id]
-            self._sessions[session_id] = session
-            session_context._set_session(session)
-            self._session_contexts[session_id] = session_context
-
-            # notify anyone waiting on the pending session
-            future.set_result(session)
-
         if session_id in self._pending_sessions:
-            # another create_session_if_needed is working on
-            # creating this session
+            # another create_session_if_needed is working on creating this session
             session = yield self._pending_sessions[session_id]
+            del self._pending_sessions[session_id]
         else:
             session = self._sessions[session_id]

+        print(session)
         raise gen.Return(session)

     def get_session(self, session_id):
diff --git a/bokeh/server/tornado.py b/bokeh/server/tornado.py
index 443cd3d8a..385889a1c 100644
--- a/bokeh/server/tornado.py
+++ b/bokeh/server/tornado.py
@@ -19,13 +19,15 @@ log = logging.getLogger(__name__)
```

The issue here is that the the initial document renders, but then no periodic callbacks actually work. I think this is because the callbacks are started on the wrong IOLoop (the threads) but have not verified this.

bryevdv:
@sridk I am going to go ahead and close this, as it is more work than I had anticipated, and given the upcoming major internal changes for Bokeh 2.0 to use native coroutines, a larger change now will result in difficult conflicts in the near future. Apologies for misjudging the effort required, but thank you for the PR and discussion. Hopefully we can revisit this after 2.0

PR #9043, opened 2019-07-03
New compiler infrastructure

### LINKED ISSUE #9002: Unify bokehjs build system with extensions' compiler
The idea is to have bring fully featured build system to extensions' compiler, which currently is an ad-hoc approach, e.g. it doesn't allow to use npm packages. This work will also allow for generating custom bokeh bundles at runtime. This issue focuses only on technical aspects. User oriented development workflow will be established in future work.

mattpap:
This:

- [x] unifies bokehjs build with extensions' compiler (build is now a thin wrapper on top of the compiler)
- [x] adds preliminary `bokeh build` command (doesn't do much at this point; further PRs will fix this)
- [x] adds support for artifact caching (this not only improves build speed significantly, but also opens the path towards customizable bundles)
- [x] switches from full program to individual module minification (doesn't result in visible size increase, but allows for much quicker builds and is a requirement for caching)

Backwards compatibility is preserved as much as I can tell.

fixes #9002


mattpap:
I'm going to merge this when green to get PR #9138 finished, which is blocked on this PR.

philippjfr:
Are there any examples or docs so I can start playing around with this?

mattpap:
@philippjfr, not yet, because there isn't much to play with in this PR. Here I'm laying the foundation for further work, which I have in WIP state right now. I added `bokeh build` command, but primarily for testing, and you won't be able to build any extensions with it without settings a few things up exactly right. Next week I should have a follow up PR.

PR #9045, opened 2019-07-03
Add Pandas-Bokeh to downstream tests

### LINKED ISSUE #9044: [TASK] Add Downstream Tests for Pandas-Bokeh
@bryevdv asked me to add downstream tests for Pandas-Bokeh (https://github.com/PatrikHlobil/Pandas-Bokeh/issues/26). I tried to add it in the corresponding pull request.

Thanks Patrik

PatrikHlobil:
@bryevdv asked me to add downstream tests for Pandas-Bokeh.

- [ ] issues: fixes #9044
- [ ] tests added / passed

### DIFF
--- a/scripts/ci/install.downstream
+++ b/scripts/ci/install.downstream
@@ -5,3 +5,7 @@ set -x # echo commands
 
 conda install -n testenv --yes --quiet distributed
 conda install -n testenv --yes --quiet -c pyviz/label/dev holoviews nose
+conda activate testenv
+pip install pandas_bokeh
+pip install geopandas
+conda deactivate 
--- a/scripts/ci/install.downstream
+++ b/scripts/ci/install.downstream
@@ -5,7 +5,5 @@ set -x # echo commands
 
 conda install -n testenv --yes --quiet distributed
 conda install -n testenv --yes --quiet -c pyviz/label/dev holoviews nose
-conda activate testenv
 pip install pandas_bokeh
 pip install geopandas
-conda deactivate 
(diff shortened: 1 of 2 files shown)


bryevdv:
@PatrikHlobil Thanks for adding this. FYI these tests are advisory-only, i.e. they will never fail the build (that would create too-strong a coupling between build success, and things we have no direct control over). Looking at the actual results page on Travis, it looks like several of your tests are all currently failing withe the same error:

> E       AttributeError: module 'pandas_bokeh' has no attribute 'save'

PatrikHlobil:
@bryevdv 

I have seen that the tests failed. The reason is that I recently merged some new test cases of the next release on the master. So, the PyPI release and the github master are out of sync in terms of tests. This will be fixed with the next release.

Can I get notified when the Boleh downstream tests fail? Is there a subscription or anything on Travis CI?

Good luck with your fundraiser,

Best Patrik

bryevdv:
> Can I get notified when the Boleh downstream tests fail? Is there a subscription or anything on Travis CI?

There is nothing currently, unfortunately, and I am not aware of anything on Travis or GH that would help. I think we would have to set up an AWS Lambda or something with a REST api that the downstream job could hit, so that failures there could be reported without affecting the overall build status.

bryevdv:
Actually it's also possible that a GitHub Action could be used to notify. It's a closed beta but I have joined the waiting list. I will go ahead and merge this since you indicate the tests will be in sync soon.

bryevdv:
Another option, of course, is to install from the GH checkout and use that for testing. I don't have a strong opinion.

PR #9046, opened 2019-07-04
New event listener for TextInput

### LINKED ISSUE #8676: make a bokeh TextInput callback responsive as text is typed
This feature is described originally in a Stack Overflow [question,](https://stackoverflow.com/questions/54756780/make-a-bokeh-textinput-callback-responsive-as-text-is-typed) where it was recommended to create a GitHub issue.

Current functionality of the on_change method of a bokeh TextInput object is to run the callback after all text is typed and then some other key is pressed. I want to know if a TextInput object can be configured to run a callback as text is being typed / deleted. The following code is a minimal example.  The application created by the code demonstrates the issue.

    from bokeh.io import curdoc
    from bokeh.layouts import widgetbox, column
    from bokeh.models import TextInput, Button

    button = Button(button_type='success')
    ti = TextInput(title='enter text to enable button')
    layout = column(
        widgetbox(ti),
        widgetbox(button))

    button.disabled = True

    def callback(attr, old, new):
        if ti.value != '':
            button.disabled = False
        else:
            button.disabled = True

    ti.on_change('value', callback)

    curdoc().add_root(layout)

cbouy:
Adds the possibility to use the "input" event for the TextInput event listener. The "change" event is still default, and the "input" event is accessible by setting the new variable "wait_commit" as False.

- [ ] issues: fixes #8676
- [ ] tests added / passed

### DIFF
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -97,7 +97,13 @@ class TextInput(InputWidget):
     """)
 
     placeholder = String(default="", help="""
-    Placeholder for empty input field
+    Placeholder for empty input field.
+    """)
+    
+    wait_commit = Bool(default=True, help="""
+    Run the callback once the user commits the value in the ``TextInput``
+    (i.e. by pressing Enter or clicking outside the text box area). If False, 
+    the callback is executed everytime the value changes (i.e. as text is being typed).
     """)
 
 
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -99,10 +99,10 @@ class TextInput(InputWidget):
     placeholder = String(default="", help="""
     Placeholder for empty input field.
     """)
-    
+
     wait_commit = Bool(default=True, help="""
     Run the callback once the user commits the value in the ``TextInput``
-    (i.e. by pressing Enter or clicking outside the text box area). If False, 
+    (i.e. by pressing Enter or clicking outside the text box area). If False,
     the callback is executed everytime the value changes (i.e. as text is being typed).
     """)
 
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -90,20 +90,20 @@ class TextInput(InputWidget):
     Initial or entered text value.
     """)
 
+    value_input = String(default="", help="""
+    Initial or entered text value that triggers a callback whenever the value changes.
+    """)
+
     callback = Instance(Callback, help="""
     A callback to run in the browser whenever the user unfocuses the
     ``TextInput`` widget by hitting Enter or clicking outside of the text box
     area.
+
+    DEPRECATED: use .js_on_change or .on_change with "value" or "value_input"
     """)
 
     placeholder = String(default="", help="""
     Placeholder for empty input field.
-    """)
-
-    wait_commit = Bool(default=True, help="""
-    Run the callback once the user commits the value in the ``TextInput``
-    (i.e. by pressing Enter or clicking outside the text box area). If False,
-    the callback is executed everytime the value changes (i.e. as text is being typed).
     """)
 
 
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -26,7 +26,7 @@ log = logging.getLogger(__name__)
 
 # Bokeh imports
 from ...core.has_props import abstract
-from ...core.properties import Date, Either, Float, Instance, Int, List, String, Tuple, Dict, ColorHex, Bool
+from ...core.properties import Date, Either, Float, Instance, Int, List, String, Tuple, Dict, ColorHex
 
 from ..callbacks import Callback
 
--- a/bokeh/models/widgets/inputs.py
+++ b/bokeh/models/widgets/inputs.py
@@ -26,7 +26,7 @@ log = logging.getLogger(__name__)
 
 # Bokeh imports
 from ...core.has_props import abstract
-from ...core.properties import Date, Either, Float, Instance, Int, List, String, Tuple, Dict, ColorHex
+from ...core.properties import Date, Either, Float, Instance, Int, List, String, Tuple, Dict, ColorHex, Bool
 
 from ..callbacks import Callback
 
--- a/bokehjs/src/lib/models/widgets/text_input.ts
+++ b/bokehjs/src/lib/models/widgets/text_input.ts
@@ -16,7 +16,6 @@ export class TextInputView extends InputWidgetView {
     this.connect(this.model.properties.value.change, () => this.input_el.value = this.model.value)
     this.connect(this.model.properties.disabled.change, () => this.input_el.disabled = this.model.disabled)
     this.connect(this.model.properties.placeholder.change, () => this.input_el.placeholder = this.model.placeholder)
-    this.connect(this.model.properties.wait_commit.change, () => this.input_el.wait_commit = this.model.wait_commit)
   }
 
   render(): void {
@@ -29,7 +28,6 @@ export class TextInputView extends InputWidgetView {
       value: this.model.value,
       disabled: this.model.disabled,
       placeholder: this.model.placeholder,
-      wait_commit: this.model.wait_commit
     })
     if (this.model.wait_commit) {
         this.input_el.addEventListener("change", () => this.change_input())
@@ -70,7 +68,7 @@ export class TextInput extends InputWidget {
     this.define<TextInput.Props>({
       value:       [ p.String, "" ],
       placeholder: [ p.String, "" ],
-      wait_commit:  [ p.Boolean, true ],
+      wait_commit: [ p.Boolean, true ],
     })
   }
 }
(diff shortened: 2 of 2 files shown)


cbouy:
Regarding the test for this in [tests/integration/widgets/test_text_input.py](https://github.com/cbouy/bokeh/blob/master/tests/integration/widgets/test_text_input.py)
Apparently there's no function to retrieve the event type in JavaScript, but there's a workaround. 
Some example on how to do this in JavaScript: https://developer.mozilla.org/fr/docs/Web/API/Event/type#Example
The only automated way would be to store a variable with the event type in the [change_input](https://github.com/cbouy/bokeh/blob/2bc26190369dc9d33173324c9df84d4daf939c17/bokehjs/src/lib/models/widgets/text_input.ts#L40) function, but I'm not sure how to do this, I'm a complete noob in TypeScript, so I'd appreciate any help.

bryevdv:
@cbouy I'd like to propose an alternate solution and see what you you think. Recently for `Slider` we added a separate `value_throttled` property. The idea is that that `value` updates every slider move and `value_throttled` only updates at the end, on `mouseup`. This is "opposite" in some sense, but I think a similar approach might be good here too? I.e, instead of adding a `Bool` to set a configurable  "mode", just add a new `value_input` property that always updates unconditionally on the input event. Then users can register callbacks on either one, or even both. What do you think?

sbrice:
@bryevdv could you link to the relevant PR/code in `Slider` - If you can recall off the top of your head

sbrice:
@cbouy RE: passing the Event type -

You have the right idea. 

When `addEventListener` triggers `change_input` it automatically passes the `Event` object  to the callback (as shown in Line 10 of the [example](https://developer.mozilla.org/fr/docs/Web/API/Event/type#Example) you linked).

To make use of the event type you'd effectively need something like...

```
change_input(event): void {
  const eventType = event.type;
  // ...
}
```
Keep in mind that `Event` may be `null` or `undefined` - more options here: https://rjzaworski.com/2018/10/typescript-event-handlers

bryevdv:
@sbrice for sure, it is here:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/widgets/abstract_slider.ts

I hadn't linked it earlier because the slider case is actually made somewhat more complicated by having to handle its own throttling. I think this situation is simpler, i.e. we can just hook up one callback to `change` that updates `value` and another callback to `input` that updates `value_input` (or whatever we want to call it).

sbrice:
Ah ok, if that's the case you may be able to generalize this by adding a string attribute e.g. `change_input_event` to `class TextInput` and then passing that attribute to the `addEventListener` function as opposed to using a flag and chaining `if-else` statements. 

The benefit of such an implementation is that you could easily refactor it into `class InputWidget` to give the other widgets within that component similar functionality - or you can only implement it for `class TextInput`. It also reduces complexity and lines-of-code.

For example...

```
class TextInput(InputWidget):
    ''' Single-line input widget.
    '''

    value = String(default="", help="""
    Initial or entered text value.
    """)

    callback = Instance(Callback, help="""
    A callback to run in the browser whenever the user unfocuses the
    ``TextInput`` widget by hitting Enter or clicking outside of the text box
    area.
    """)

    placeholder = String(default="", help="""
    Placeholder for empty input field.
    """)

    change_input_event = String(default="change", help="""
    Run the callback by listening for a specific event type.
    """)
```

...and in the JavaScript side, it would simply be...

```
render(): void {
   // ...
   this.input_el.addEventListener(this.model.change_input_event,  () => this.change_input())
   this.group_el.appendChild(this.input_el)
}
```

bryevdv:
I still think I'd prefer this:
```
class TextInput(InputWidget):
    ''' Single-line input widget.
    '''

    value = String(default="", help="""
    This updates when the "change" event fires    
    """)

    value_input = String(default="", help="""
    This updates when the "input" event fires    
    """)

This is (I think) even simpler on the JS side. E.g. with the versions above, we'd really want things to work if `change_input_event` was changed after an app started. But that would take some code to re-plumb things. With the version above, the event handlers are set once on the JS side, unconditionally, and never have to be updated. Also, this version affords users a way to watch *both* e.g. if they want updates every keystroke, but they also want to distinguish when "enter" has been pressed. The other versions miss this capability I think.

**Comment by cbouy** [comment]
@bryevdv Looks simpler and more complete indeed. Listening to both events could definitely be useful and works like a charm:
```python
from bokeh.io import curdoc
from bokeh.models import TextInput, Div
import os, glob

ti = TextInput(title='Path to your directory', width=500)
div = Div(text="")

def callback_input(attr, old, new):
    path = os.path.join(new + "*", "")
    opts = glob.glob(path)
    completions = "<br/>".join(opts)
    div.text = completions
ti.on_change('value_input', callback_input)

def callback_change(attr, old, new):
    div.text = "You selected %s" % new
ti.on_change('value', callback_change)

layout = column(ti, div)
curdoc().add_root(layout)
```


> @cbouy RE: passing the Event type -
> 
> You have the right idea.
> 
> When `addEventListener` triggers `change_input` it automatically passes the `Event` object to the callback (as shown in Line 10 of the [example](https://developer.mozilla.org/fr/docs/Web/API/Event/type#Example) you linked).
> 
> To make use of the event type you'd effectively need something like...
> 
> ```
> change_input(event): void {
>   const eventType = event.type;
>   // ...
> }
> ```
> 
> Keep in mind that `Event` may be `null` or `undefined` - more options here: https://rjzaworski.com/2018/10/typescript-event-handlers

@sbrice Do we even need to capture the event for testing now with this `value_input` approach ? I tried it and got the following errors:
```
src/lib/models/widgets/text_input.ts:38:3 - error TS2416: Property 'change_input' in type 'TextInputView' is not assignable to the same property in base type 'InputWidgetView'.
      Type '(event: any) => void' is not assignable to type '() => void'.
    
    38   change_input(event): void {
         ~~~~~~~~~~~~
    src/lib/models/widgets/text_input.ts:38:16 - error TS7006: Parameter 'event' implicitly has an 'any' type.
    
    38   change_input(event): void {
                      ~~~~~
```
Using the following code: 
```
  render(): void {
    // [...]
    this.input_el.addEventListener("change", (e: Event) => this.change_input(e))
    this.input_el.addEventListener("input",  (e: Event) => this.change_input(e))
    this.group_el.appendChild(this.input_el)
  }

  change_input(event): void {
    const eventType = event.type;
    if (eventType == "change") {
      this.model.value = this.input_el.value
    } else if (eventType == "input") {
      this.model.value_input = this.input_el.value
    }
    super.change_input()
  }
```

sbrice:
No. You no longer need to capture the event for testing unless you just want to have an assertion test.

The error you're seeing `error TS7006` is from the TypeScript compiler - the compiler settings are such that you must explicitly define the type of function parameters even if the type is `any` e.g. `change_input(event: any): void` or `change_input(event: Event): void`

bryevdv:
There was an unused import that tripped the linter. I removed it using the GH web UI, but there is possibly one selenium test failure to look into (will see how next test run goes)

bryevdv:
This change has impacted the `AutoComplete` widget:
```
____ Test_AutocompleteInput.test_server_on_change_round_trip_partial_entry _____
tests/integration/widgets/test_autocomplete_input.py:263: in test_server_on_change_round_trip_partial_entry
    assert results['data']['val'] == ["400", "100001"]
E   AssertionError: assert ['100', '100001'] == ['400', '100001']
E     At index 0 diff: '100' != '400'
E     Full diff:
E     - ['100', '100001']
E     ?   ^
E     + ['400', '100001']
E     ?   ^
______ Test_AutocompleteInput.test_server_on_change_round_trip_menu_entry ______
tests/integration/widgets/test_autocomplete_input.py:295: in test_server_on_change_round_trip_menu_entry
    assert results['data']['val'] == ["400", "12344557"]
E   AssertionError: assert ['123', '12344557'] == ['400', '12344557']
E     At index 0 diff: '123' != '400'
E     Full diff:
E     - ['123', '12344557']
E     ?   ^^^
E     + ['400', '12344557']
E     ?   ^^^
```
Will need to resolved before merging, I can try to look at it this weekend if no one else gets there first.

bryevdv:
Tho my first guess is that the name change for the `change_input` method is the culprit, since I believe Autocomplete was overriding it and expecting its version to be added to the event handler:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/lib/models/widgets/autocomplete_input.ts#L34

so either `Autocomplete` would need a corresponding change, or (more conservatively) we don't rename `change_input`

bryevdv:
Oh, I had just imagined renaming the one function back and leaving the new one as it was, but if this works I am fine with it

cbouy:
I was playing with the `AutocompleteInput` now that you mentioned it, and it's not behaving as expected: no callback are being triggered, wether using a `callback` EventListener or an `input` one. Not sure if it was the case before with the `on_change("value",callback)`

```python
from bokeh.io import curdoc
from bokeh.models import AutocompleteInput, Div
import glob

ti = AutocompleteInput(
    title="Path to your directory", width=500, completions=[],
)
div = Div(text="")
layout = column(ti, div)

def callback_input(attr, old, new):
    path = os.path.join(new + "*", "")
    opts = glob.glob(path)
    div.text = "<br/>".join(opts)
    ti.completions = opts

def callback_change(attr, old, new):
    div.text = "You selected %s" % new

ti.on_change('value', callback_change)
ti.on_change('value_input', callback_input)

curdoc().add_root(layout)
```

bryevdv:
@cbouy at least at present, an `AutoComplete` will *only* accept completions, and you have not specified any in that code.

bryevdv:
@cbouy I pushed a change to fix the linter error

bryevdv:
Thanks @cbouy !

cbouy:
Glad I could help

PR #9048, opened 2019-07-09
upload to s3 CDN in parallel

bryevdv:


bryevdv:
Will merge this and test with a dev build 

@mattpap you will want to rebase. I deleted the unused SauceLabs secret env vars (scripts now check secret AWS vars to upload or not)

PR #9053, opened 2019-07-11
allow source.data set from pandas types

### LINKED ISSUE #9052: Let source.data accept DataFrame
This has come up a few times in support questions recently:
```
source.data = some_dataframe
```
It does not work, AFAIK, but also does not seem to error either. We should add a `.accepts` to make this work.

bryevdv:
- [x] issues: fixes #9052
- [x] tests added / passed

@mattpap it occurred to me that we could possibly tighten up the pandas "properties" more by making them specialized to raise errors if they were actually added to models. But I think the documentation is sufficient, unless you have a strenuous objection.

### DIFF
--- a/bokeh/core/properties.py
+++ b/bokeh/core/properties.py
@@ -140,6 +140,12 @@ Special Properties
 .. autoclass:: Include
 .. autoclass:: Override
 
+Validation-only Properties
+--------------------------
+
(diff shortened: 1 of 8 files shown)


bryevdv:
I'll merge this in a few hours, unless anyone comments otherwise.

PR #9058, opened 2019-07-12
Add support for mypy

### LINKED ISSUE #9057: Add support for mypy and start statically typing bokeh
Given that we are dropping Python 2 support, enabling mypy is finally possible (without going through hoops). We will take the same gradual typing approach as we did with TypeScript, so every new statically typed module will have to be enabled in `setup.cfg`, at least until we have enough coverage to reverse this logic and explicitly disable untyped modules.

mattpap:
Type type check bokeh issue `mypy bokeh`.

fixes #9057

### DIFF
--- a/bokeh/palettes.py
+++ b/bokeh/palettes.py
@@ -334,6 +334,7 @@ log = logging.getLogger(__name__)
 # Standard library imports
 import sys as _sys
 import types as _types
+from typing import List, Dict
 
 # External imports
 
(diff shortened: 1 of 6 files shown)


bokeh/palettes.py
```diff
@@ -1381,4 +2227,4 @@ def gray(self, n):
 _mod.__doc__ = __doc__
 _mod.__all__ = dir(_mod)
 _sys.modules['bokeh.palettes'] = _mod
-del _autoprop, _mod, _sys, _types
+del _mod, _sys, _types
```
bryevdv: I was about to say could remove `Dict` and `List` here but I'm not actually sure we need to remove anything, since the actual `sys.modules` entry is not this module but the `_PalettesModule` instance

bryevdv:
Thanks for striating this it had been on my mind but I had not made an issue for it yet

mattpap:
I'm preparing for tomorrow's EuroPython sprints. I will be making this the main subject, so hopefully will get some decent coverage.

bokeh/palettes.py
mattpap: That make's sense. I didn't consider this issue at all when adding `typing`'s imports.

PR #9063, opened 2019-07-13
fix table selection issues

### LINKED ISSUE #8921: [BUG] DataTable copy-paste doesn't work with 0s present in a row
ubuntu/16.04
firefox/66.0.5 (64-bit)
chrome/74.0.3729.157 (64-bit)
bokeh/1.1.0

Copy-pasting data from a bokeh gallery example:
https://demo.bokeh.org/export_csv
doesn't work if a row contains 0 (except for the index value).

As an example, copying the first row and/or third works,
```
0	Veronica Okuneva	47298.0146627	3
2	Ashlynn Barrows	45591.9070251	7
```
but copying the second, where `Experience (years)` is 0, doesn't.

### LINKED ISSUE #8923: [BUG] DataTable copy-paste doesn't keep the order of rows after sorting
ubuntu/16.04
firefox/66.0.5 (64-bit)
chrome/74.0.3729.157 (64-bit)
bokeh/1.1.0

Copy-pasting rows from a bokeh gallery example:
https://demo.bokeh.org/export_csv
after sorting that DataTable by `Employee Name`, results in out of order data:
```
97	Abdul Crooks	38581.7791509	3
88	Akeem Pfannerstill	26974.3379311	2
12	Alayna Goyette	44427.462095099996	5
4	Alford Rau	37929.3731943	6
56	Abbey Ziemann	42864.5023151	2
```

The expected row order would be the one a user sees after sorting,
```
56	Abbey Ziemann	42864.5023151	2
97	Abdul Crooks	38581.7791509	3
88	Akeem Pfannerstill	26974.3379311	2
12	Alayna Goyette	44427.462095099996	5
4	Alford Rau	37929.3731943	6
```

bryevdv:
- [x] issues: fixes #8921
- [x] issues: fixes #8923
- [x] tests added / passed

Tests coming tomorrow. The issues here:

* permuted indices can end up re-ordered from row order, they just need to be sorted so that the rows that are selected are copied out in order

* `container` is a jQ object, `appendChild` does not exist (use `append`)

### DIFF
--- a/bokeh/_testing/util/selenium.py
+++ b/bokeh/_testing/util/selenium.py
@@ -144,14 +144,18 @@ def hover_element(driver, element):
     hover = ActionChains(driver).move_to_element(element)
     hover.perform()
 
-def enter_text_in_element(driver, element, text, click=1, enter=True):
+def enter_text_in_element(driver, element, text, click=1, enter=True, mod=None):
     actions = ActionChains(driver)
     actions.move_to_element(element)
     if click == 1: actions.click()
     elif click == 2: actions.double_click()
     if enter:
         text += Keys.ENTER
+    if mod:
+        actions.key_down(mod)
     actions.send_keys(text)
+    if mod:
+        actions.key_up(mod)
     actions.perform()
 
 def enter_text_in_cell(driver, cell, text):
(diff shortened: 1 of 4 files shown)


bryevdv:
OK, having issues getting a copy/paste selenium test working. It may be an OSX issue. If the test I just pushed runs correctly I will continue and add a few more, otherwise I will open a separate issue about copy paste tests for table and go ahead and merge this.

PR #9068, opened 2019-07-13
Small fix for windows dev guide and dev env setup

### LINKED ISSUE #9065: [BUG] Development guide missing `test` argument for conda install and pytest install failure on windows
### Missing `test` argument

The current [developement guide](https://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html#conda-packages) is missing the `test` argument for windows setups. 

As for OSX / Linux (bash / sh), it is: 
- ```conda install `python scripts/deps.py build run test```.

As for windows, the `test` argument is missing for the `deps.py`:
- ```conda install $(python scripts/deps.py build run).split() | where {$_}```
- ```for /F "delims=" %i in ('python scripts\deps.py build run') do (conda install %i)```

Instead, it should be:
- ```conda install $(python scripts/deps.py build run test).split() | where {$_}```
- ```for /F "delims=" %i in ('python scripts\deps.py build run test') do (conda install %i)```

### `pytest<5.0.0` fails
In addition, running ```for /F "delims=" %i in ('python scripts\deps.py build run test') do (conda install %i)``` fails with error `System can't find given file.` which is due to `pytest<5.0.0`. Providing double quotes actually fixes the issue -> `conda install "pytest<5.0.0"`.

mansenfranzen:
Fixes documentation and conda packages for windows dev setup.

- [x] issues: fixes #9065
- [x] tests passed

### DIFF
--- a/scripts/deps.py
+++ b/scripts/deps.py
@@ -1,4 +1,5 @@
 import sys
+import platform
 import jinja2
 import yaml
 
@@ -30,6 +31,10 @@ for name in sys.argv[1:]:
 
 # bare python unpins python version causing upgrade to latest
 if 'python' in spec: spec.remove('python')
+
+# add double quotes to specs for windows, fixes #9065
+if "windows" in platform.platform().lower():
+    spec = ['"{}"'.format(s) for s in spec]
 
 deps = ""
 deps += " ".join(s for s in spec)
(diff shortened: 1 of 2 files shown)


bryevdv:
This will need to be rebased on master to pass. There was a brief build problem merged then fixed and this PR happened to catch the time in between

bryevdv:
Tho actually, given the content and the tests that did pass, I don't think that's necessary. I will just merge now, and things will be green on master.

bryevdv:
Thanks @mansenfranzen !

Issue #9064, opened 2019-07-13
Add static typing to bokeh


mattpap:
This is a meta issue for adding static typing to bokeh:

- [x] `bokeh.palettes`
- [ ] `bokeh.embed.standalone`
- [ ] `bokeh.util.string` @mborus

mansenfranzen:
[ ] bokeh.embed.wrappers

PacoPacov:
- [] `bokeh.embed.notebook`

mborus:
finished bokeh.util.string, will do more in .util

enricapq:
[] bokeh.util.options

mansenfranzen:
- [ ] bokeh.util.serialization.py

enricapq:
- [ ] bokeh.util.datatypes

aCampello:
- [ ] bokeh.util.sampledata

w0rdsm1th:
- [] bokeh.util.hex
- [] bokeh.ext

mjarosie:
[ ] bokeh.core.query

w0rdsm1th:
hi guys, just submitted a PR against landing-2.0 for bokeh.command. there were a few functions where I was unable to resolve typing without changing logic, called these out in the PR notes and commit message. mypy checking fails on some modules because of these difficulties

PR #9069, opened 2019-07-13
Add static types to bokeh.util.options

enricapq:
addresses #9064

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,3 +62,6 @@ ignore_errors = True
 
 [mypy-bokeh.palettes]
 ignore_errors = False
+
+[mypy-bokeh.util.options]
+ignore_errors = False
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -65,3 +65,6 @@ ignore_errors = False
 
 [mypy-bokeh.util.paths]
 ignore_errors = False
+
+[mypy-bokeh.util.options]
+ignore_errors = False
```
bryevdv: @mattpap @enricapq is there way to condense these into one section under a single mypy config block? Guessing not, but it's going to result in a huge even messier list in this file before we get to the end.

mattpap: @bryevdv, I will conflate those as soon as we have enough coverage. For now this results in simpler merge conflicts to resolve.

PR #9071, opened 2019-07-13
Add static types to bokeh.util.paths #9066

PacoPacov:
- [X] `bokeh.embed.elements` (didn't annotate html_page_for_render_items method)
- [X] `bokeh.embed.notebook`


bokeh/embed/notebook.py
```diff
@@ -44,8 +47,10 @@
 #-----------------------------------------------------------------------------
 # Dev API
 #-----------------------------------------------------------------------------
+Theme = Union[Theme, FromCurdoc, None]
```
bryevdv: I think `ThemeSource` might be a better name here

bryevdv: I think `ThemeSource` might be a better name here, in any case I don't think it can be `Theme` since that shadows the `Theme` import above.

bokeh/embed/elements.py
```diff
@@ -22,13 +22,15 @@
...
 # External imports
+from typing import List, Optional
 
 # Bokeh imports
 from ..core.json_encoder import serialize_json
 from ..core.templates import _env, DOC_JS, FILE, MACROS, PLOT_DIV
 from ..document.document import DEFAULT_TITLE
 from ..settings import settings
 from ..util.serialization import make_id
+from .embed.util import RenderItem
```
bryevdv: this should just be `.util` the module is already in `bokeh.embed`

bokeh/embed/notebook.py
mattpap: Including `None` is unnecessary, because the default is always (as far as I can tell) `FromCurdoc`. Even if `None` was used, it would be usually preferred to use `Optional[ThemeSource]`. Another issue is passed around as a class `FromCurdoc` and the above type indicates an instance of `FromCurdoc`. The correct type is:
```
Union[Theme, Type[FromCurdoc]]
```
(`Type` comes from `typing` module). There's an orthogonal question whether this is the right way of working with singleton objects altogether, but that's something we will need to answer later on.

mattpap:
Superseded by PR #9090.

PR #9077, opened 2019-07-14
Add type annotations to `bokeh.util.serialization`

mansenfranzen:
Add type annotations to `bokeh.util.serialization`.

- [x] issues: part of #9064
- [x] tests passed locally

### ToDo
This PR is not complete. Following points need to be resolved:

- [ ] To allow mypy to correctly check types, it was necessary to remove `import_optional` and move its content into the module for the optional `pandas` import. There should be a better solution for this in general.
- [ ]  `convert_datetime_type` returns `None` if the type of the input parameter does not match. I'm not sure if this is the intended behaviour, though. Perhaps raising an error or returning `np.nan` by default is more appropriate.

### For future type annotations
As a side note for future type annotations: mypy seems to work best when there is an explicit `import` statement. This is evident from the above example of removing `import_optional` and also from this [example](https://github.com/bokeh/bokeh/compare/landing-2.0...mansenfranzen:9064_types_bokeh_util?expand=1#diff-07271ee774c39325189271f9b13c7a77L70):
  - ` _pd_timestamp = pd.Timestamp` does not work with mypy
  - `from pandas import Timestamp as _pd_timestamp` works with mypy


mattpap:
Superseded by PR #9090.

PR #9088, opened 2019-07-15
Purge dead references on tools change

### LINKED ISSUE #9032: [BUG] TypeError after replacing tools on a toolbar
My plot should have different sets of tools depending on what is displayed.
setting `plot.toolbar` doesn't help since the toolbar model change not not handled correctly.
Executing something like `plot.toolbar.tools = new_tools_list` seems to work but results in this exception in JS console:
```
bokeh.js:30582 Uncaught TypeError: Cannot read property 'el' of undefined
    at el (bokeh.js:30582)
    at Array.map (<anonymous>)
    at ToolbarBaseView.render (bokeh.js:30586)
    at ToolbarPanelView.render (bokeh.js:11764)
    at PlotView._paint_levels (bokeh.js:22064)
    at PlotView.paint (bokeh.js:22044)
    at PlotView.repaint (bokeh.js:21983)
    at PlotView.<anonymous> (bokeh.js:21892)
    at Signal0.Signal.emit (bokeh.js:4167)
    at Signal0.emit (bokeh.js:4180)
```

Both `Toolbar` and `ToolbarBaseView` attach a signal on the `tools` property change. However, the `Toolbar.gestures` property change does not remove the old references:
```
// Toolbar._init_tools()
if (!some(this.gestures[et].tools, (t) => t.id == tool.id))
  this.gestures[et].tools = this.gestures[et].tools.concat([tool])
```
Because of this, only the addition of tools works without errors. And while `gestures` is a property in BokehJS, it is not in the Python counterpart.

Similar code can be seen in `ProxyToolbar` but it doesn't handle the changes to the list of tools at all, so I guess it was never on the plate.

While it's not reflected in the error above, other attributes have the same issue of not being pruned on tools change: `inspectors`, `help`, `actions`.

p-himik:

- [x] issues: fixes #9032
- [x] tests ~~added~~ / passed

### DIFF
--- a/bokehjs/src/lib/models/tools/toolbar.ts
+++ b/bokehjs/src/lib/models/tools/toolbar.ts
@@ -1,11 +1,8 @@
 import * as p from "core/properties"
-import {logger} from "core/logging"
-import {isString, isArray} from "core/util/types"
-import {some, sort_by, includes} from "core/util/array"
+import {isArray} from "core/util/types"
+import {sort_by, includes, intersection} from "core/util/array"
 
 import {Tool} from "./tool"
-import {ActionTool} from "./actions/action_tool"
-import {HelpTool} from "./actions/help_tool"
 import {GestureTool} from "./gestures/gesture_tool"
 import {InspectTool} from "./inspectors/inspect_tool"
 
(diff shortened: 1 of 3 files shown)


bryevdv:
It looks like there is one linter error:

> [23:37:22] /home/travis/build/bokeh/bokeh/bokehjs/src/lib/models/tools/toolbar.ts:133:9
[23:37:22] ERROR: 133:9   one-line   misplaced 'else'
[23:37:22] 
[23:37:22] /home/travis/build/bokeh/bokeh/bokehjs/src/lib/models/tools/toolbar_base.ts:268:51
[23:37:22] ERROR: 268:51  ban-types  Don't use 'Array<GestureType>' as a type. Use 'T[]' instead.
The command "scripts/ci/test" exited with 1.

I will look at the content tomorrow night

bryevdv:
@p-himik it's been pretty much ages since I have been in this code, can you give me a high-level overview of the approach you took and I will look more closely and test with a checkout

p-himik:
@bryevdv 
- Extracted common parts of `Toolbar` and `ProxyToolbar` into `ToolbarBase
- Rearranged types a bit to make TS happy without having to duplicate stuff or write multiple `X as Y`
- For every reference and set of references to `Tool` that I know of, synchronized them with the set of current tools

bryevdv:
Thanks @p-himik !

PR #9091, opened 2019-07-16
Store msgs and handle them when session is created

### LINKED ISSUE #8403: Bokeh charts load very slow with Uncaught TypeError: in browser console
Bokeh charts render data points very slow with below error in the browser console. 

```
Uncaught TypeError: Cannot read property 'handle' of null
    at t._steady_state_handler (connection.js:262)
    at t.ACK.t.msgtype._current_handler (connection.js:242)
    at t._on_message (connection.js:204)
    at WebSocket.i.socket.onmessage (connection.js:61)
```

screen shot 2018-11-06 at 8 58 44 am

p-himik:

- [x] issues: fixes #8403
- [x] tests ~~added~~ / passed

---

Because of the heavy use of promises, we still need to accumulate messages and handle them when the session is ready. Another reason to accumulate them is that there are kinds of messages other than PATCH - e.g. we don't want to miss any errors.


bryevdv:
Thanks @p-himik !

PR #9093, opened 2019-07-16
fix double slash in autoload.js

### LINKED ISSUE #9092: [BUG] double slash before prefix in autoreload.js
Python 2.7
Bokeh 1.2.0

Resource urls in autoreload.js have double slashes before prefix if server has a prefix

```python
from bokeh.server.tests.utils import ManagedServerLoop, url, http_get

def test_urls_in_autoreloadjs():
    with ManagedServerLoop(Application(), prefix='PREFIX') as server:
        absolute_url = '{server}{prefix}/'.format(server=url(server),  prefix=server.prefix)
        resp = http_get(
            server.io_loop,
            absolute_url + 'autoload.js?bokeh-autoload-element=foo&bokeh-absolute-url=' + absolute_url
        )
        js_code = resp.buffer.read()
        assert not re.findall('\/\/PREFIX', js_code)
```

So autoreload.js can't load files
image

EtsuNDmA:
issues: fixes #9092

### DIFF
--- a/bokeh/server/views/autoload_js_handler.py
+++ b/bokeh/server/views/autoload_js_handler.py
@@ -68,7 +68,7 @@ class AutoloadJsHandler(SessionHandler):
         absolute_url = self.get_argument("bokeh-absolute-url", default=None)
 
         if absolute_url:
-            server_url = '{uri.scheme}://{uri.netloc}/'.format(uri=urlparse(absolute_url))
+            server_url = '{uri.scheme}://{uri.netloc}'.format(uri=urlparse(absolute_url))
         else:
             server_url = None
         resources = self.application.resources(server_url)


bryevdv:
This is failing due to a build issue that only affected PRs from external forks. But a fix was just merged, this will be green if rebased/merged with master. Though I would merge as-is. @EtsuNDmA can you first confirm this does not break the case where there is *not* a prefix?

EtsuNDmA:
Yes, I can.
```python
# bokeh/server/tornado.py:195
if prefix is None:
    prefix = ""
prefix = prefix.strip("/")
if prefix:
     prefix = "/" + prefix
self._prefix = prefix
```
Then empty string concatenates with absolute_url, that does not end with '/'
```python
# bokeh/server/tornado.py:411
def resources(self, absolute_url=None):
    if absolute_url:
        return Resources(mode="server", root_url=absolute_url + self._prefix, path_versioner=StaticHandler.append_version)
return Resources(mode="server", root_url=self._prefix, path_versioner=StaticHandler.append_version)
```
And later "/" will be added to url
```python
# bokeh/resources.py:88
if root_url and not root_url.endswith("/"):
    root_url = root_url + "/"
```

bryevdv:
Thanks @EtsuNDmA

Issue #9094, opened 2019-07-16
[FEATURE] Autobuild docker images


Karel-van-de-Plassche:
Maybe we should add tests for the Dockerfiles and/or auto-update and push to our docker repository https://cloud.docker.com/u/bokeh/repository/docker/bokeh/bokeh. It seems travis can run docker-in-docker https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile. Presumably we'd build a Docker container from source every minor version of bokeh, and maybe test the Dockerfile generation every minor version? It takes quite a while to run, so let's not test every commit. Thoughts?

bryevdv:
I don't use these regularly, the only docker file I interact with much is the one in the demo repo for the Elastic Beanstalk deploy for the demo site. I am generally for automating things, but I don't have strong opinions about how in this case due to my inexperience. This seems like it might be  good task for a nightly build, but I don't know if Travis ever added support for that. I think Azure has that option. Otherwise there's not many good options besides per-commit. We would 100% not want to test on a full release tag, so that only leaves dev/rc tags as candidates, and those get pushed pretty randomly.

Karel-van-de-Plassche:
It looks like travis supports cron jobs: https://docs.travis-ci.com/user/cron-jobs/ So it seems we could run test nightly/weekly. 

Looks we can also trigger a Docker image build for dockerhub every tag: https://stackoverflow.com/questions/28590638/can-i-trigger-travis-ci-to-build-my-app-by-creating-git-tag

bryevdv:
@ianthomas23 is this issue useful to you? If not, we can close

ianthomas23:
@bryevdv Yes, we may as well close it. We have the Docker discussion https://github.com/bokeh/bokeh/discussions/12235 and when we have the full set of CI runs passing I will finish off the PR (https://github.com/bokeh/bokeh/pull/12351) so that we can use it as part of our usual CI.

PR #9095, opened 2019-07-16
Update docker to work with new conda and bokeh

Karel-van-de-Plassche:
I've updated the Dockerfile for the three styles of containers we have.  Opened a related issue for autobuilding/testing here: https://github.com/bokeh/bokeh/issues/9094

### DIFF
--- a/docker-tools/Dockerfile-from-conda
+++ b/docker-tools/Dockerfile-from-conda
@@ -25,6 +25,7 @@ RUN python -c "import tornado; print('tornado version=' + tornado.version)"
 # Workaround, just calling `bokeh info` crashes
 RUN env BOKEH_RESOURCES=cdn bokeh info
 
-#ENV BOKEH_EXAMPLE_DIR bokeh_examples
-#COPY docker_tools/hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
-#COPY docker_tools/hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
+# Add some bokeh examples to be run
+ENV BOKEH_EXAMPLE_DIR /root/bokeh_examples
+COPY examples/models $BOKEH_EXAMPLE_DIR/models
+COPY examples/app $BOKEH_EXAMPLE_DIR/app
--- a/docker-tools/Dockerfile-from-conda
+++ b/docker-tools/Dockerfile-from-conda
@@ -21,9 +21,6 @@ RUN conda install --quiet --yes \
 ARG TORNADO_VERSION
 RUN sh -c 'if [[ ! -z "$TORNADO_VERSION" ]]; then echo Installing old tornado $TORNADO_VERSION; conda install --quiet --yes tornado=$TORNADO_VERSION; conda clean -ay; fi'
 
-# Clean the conda environment
-#RUN conda clean -ay
-
 RUN python -c "import tornado; print('tornado version=' + tornado.version)"
 # Workaround, just calling `bokeh info` crashes
 RUN env BOKEH_RESOURCES=cdn bokeh info
--- a/docker-tools/Dockerfile-from-conda
+++ b/docker-tools/Dockerfile-from-conda
@@ -2,6 +2,9 @@
 FROM frolvlad/alpine-miniconda3
 LABEL org.pydata.bokeh.maintainer="Karel-van-de-Plassche <karelvandeplassche@gmail.com>"
 
+RUN apk add --no-cache \
+    bash
+
 # set default conda channels
 RUN conda config --set auto_update_conda off
 RUN conda config --append channels bokeh
@@ -9,9 +12,6 @@ RUN conda config --append channels conda-forge
 RUN conda config --get channels
 
 ARG BOKEH_VERSION
-# Do not update conda for now
-#RUN conda install --quiet --yes \
-#    conda
 
 RUN conda install --quiet --yes \
     bokeh=$BOKEH_VERSION \
@@ -28,6 +28,6 @@ RUN python -c "import tornado; print('tornado version=' + tornado.version)"
 # Workaround, just calling `bokeh info` crashes
 RUN env BOKEH_RESOURCES=cdn bokeh info
 
-ENV BOKEH_EXAMPLE_DIR bokeh_examples
-COPY hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
-COPY hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
+#ENV BOKEH_EXAMPLE_DIR bokeh_examples
+#COPY docker_tools/hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
+#COPY docker_tools/hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
--- a/docker-tools/Dockerfile-from-source
+++ b/docker-tools/Dockerfile-from-source
@@ -17,7 +17,6 @@ RUN git clone --depth=50 https://github.com/bokeh/bokeh.git $BOKEH_SOURCE_DIR
 #Increase build number to force rebuild
 ENV build 0
 WORKDIR $BOKEH_SOURCE_DIR
-#RUN git fetch origin $BOKEH_VERSION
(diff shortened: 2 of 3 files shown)


docker-tools/Dockerfile-from-conda
```diff
@@ -2,16 +2,16 @@
 FROM frolvlad/alpine-miniconda3
 LABEL org.pydata.bokeh.maintainer="Karel-van-de-Plassche <karelvandeplassche@gmail.com>"
 
+RUN apk add --no-cache \
+    bash
```
bryevdv: why the line continuation?

docker-tools/Dockerfile-from-conda
```diff
@@ -28,6 +28,6 @@ RUN python -c "import tornado; print('tornado version=' + tornado.version)"
 # Workaround, just calling `bokeh info` crashes
 RUN env BOKEH_RESOURCES=cdn bokeh info
 
-ENV BOKEH_EXAMPLE_DIR bokeh_examples
-COPY hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
-COPY hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
+#ENV BOKEH_EXAMPLE_DIR bokeh_examples
+#COPY docker_tools/hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
+#COPY docker_tools/hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
```
bryevdv: should not leave in commented code

docker-tools/Dockerfile-from-source
```diff
@@ -50,9 +50,12 @@ RUN conda build conda.recipe --quiet --no-test --no-anaconda-upload --dirty
...
 FROM frolvlad/alpine-miniconda3 AS install
 COPY --from=build /opt/conda/conda-bld/noarch/ /opt/conda/conda-bld/noarch/
+
+RUN apk add --no-cache \
+    bash
 ## We need a recent conda version for --use-local
-RUN conda install --quiet --yes \
-    conda
+#RUN conda install --quiet --yes \
+    #    conda
```
bryevdv: commented code

docker-tools/Dockerfile-from-travis
```diff
@@ -1,14 +1,15 @@
...
 ADD . /tmp/bokeh
 RUN chown -R travis /tmp/bokeh
 
 USER travis
 
 ENV NO_GIT_FETCH=1
-ENV PATH="/home/travis/miniconda/bin:${PATH}"
-ENV CONDA_REQS="conda=4.4.11 conda-env=2.6.0 conda-build=3.0.27"
+ENV PATH="/home/travis/miniconda/bin/:${PATH}"
+ENV PATH="/home/travis/miniconda/envs/testenv/bin/:${PATH}"
```
bryevdv: It avoids an extra image layer to combine these in to one statement, IIUC

docker-tools/Dockerfile-from-conda
Karel-van-de-Plassche: Ah yeah, actually I should discuss this. Do we want these 'hello_world' scripts in the containers? They just plot something so you can test bokeh in the docker container. If so, should they maybe live in the bokeh examples or somewhere else maybe?

docker-tools/Dockerfile-from-conda
Karel-van-de-Plassche: It should be `RUN apk add --no-cache bash` so I need a line continuation, correct? I got this from "Sort multi-line arguments" from https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

docker-tools/Dockerfile-from-travis
Karel-van-de-Plassche: Indeed

docker-tools/Dockerfile-from-conda
bryevdv: I see why they do that when there is a list I just question whether we will ever have a list. I am fine with it either way

docker-tools/Dockerfile-from-conda
bryevdv: I don't have a strong preference. They seem a little simplistic so I only wonder if it's possible to make something that is a nice intro example (so, fairly simple) but also somewhat compelling and interesting.

Karel-van-de-Plassche: I was trying to have an example that had the least chance of breaking between bokeh versions, as it would be nice to be able to generate containers for arbitrary bokeh versions with more-or-less the same scripts, as well as a quick way to test if bokeh 'works'.

docker-tools/Dockerfile-from-conda
Karel-van-de-Plassche: Ah yeah, we will probably not have a list per se here, but like this I could easily copy from other scripts/templates ;) I'll keep it for this reason.

docker-tools/Dockerfile-from-conda
bryevdv: So the "iris" example in the examples dir has run unchanged for 5+ years. A lot of the bar examples there are unchanged for 2+ years. At this point any basic plot should safe at this point. Even "plot plus a wideget and CusomJS" should be fine now.

bryevdv: We can also add whatever scripts get copied in to our CI testing to make sure they don't ever break

Karel-van-de-Plassche: I ended up just copying the app and models example folders, that'll be enough

docker-tools/Dockerfile-from-conda
```diff
@@ -21,13 +21,11 @@ RUN conda install --quiet --yes \
...
 RUN python -c "import tornado; print('tornado version=' + tornado.version)"
 # Workaround, just calling `bokeh info` crashes
 RUN env BOKEH_RESOURCES=cdn bokeh info
 
-ENV BOKEH_EXAMPLE_DIR bokeh_examples
-COPY hello_world.py $BOKEH_EXAMPLE_DIR/hello_world.py
-COPY hello_nodejs.py $BOKEH_EXAMPLE_DIR/hello_nodejs.py
+# Add some bokeh examples to be run
+ENV BOKEH_EXAMPLE_DIR /root/bokeh_examples
+COPY examples/models $BOKEH_EXAMPLE_DIR/models
```
bryevdv: As long as we are going to just copy directories wholesale, I would say lets either copy `plotting` instead or in addition to. The `models` examples are all "low level" and I would not want that to be anyone's introdction

docker-tools/Dockerfile-from-travis
```diff
@@ -1,16 +1,21 @@
...
 ADD . /tmp/bokeh
 RUN chown -R travis /tmp/bokeh
 
 USER travis
 
 ENV NO_GIT_FETCH=1
-ENV PATH="/home/travis/miniconda/bin:${PATH}"
-ENV CONDA_REQS="conda=4.4.11 conda-env=2.6.0 conda-build=3.0.27"
+ENV PATH="/home/travis/miniconda/envs/testenv/bin/:/home/travis/miniconda/bin/:${PATH}"
+ENV CONDA_REQS="conda=4.6.2 conda-build=3.17.8"
```
bryevdv: we should see if we can bump to 4.7 versions. The latest condas use a different compression scheme that is substantially faster

docker-tools/Dockerfile-from-conda
Karel-van-de-Plassche: I mostly want the examples just to quickly check that everything in the container works, I don't think these dockerfiles are used by 'beginners', although they might not be used at all ;) The reason I didn't put `plotting` is that it is quite big (55MB)

docker-tools/Dockerfile-from-travis
Karel-van-de-Plassche: I got this from the travis settings, and do not dare to touch them. Feel free though!

bryevdv: OK, I am going to try and rev those values on TravisCI after the upcoming release

docker-tools/Dockerfile-from-conda
bryevdv: In that case I'd say we could just hand pick a few examples, but we can revisit this later

PR #9100, opened 2019-07-19
extend FileInput with more attributes.

### LINKED ISSUE #9099: [FEATURE] Extend FileInput to return filename and make output clearer
I have been playing with the new FileInput widget in the 1.3.0rc2 release and have 2 suggestions/needs I found I had:

- Return the filename of the file being uploaded, since I need to display the names of imported files back to the users.
- Return the actual base64 encoded string of the file content separately from the mime type, rather than combined as they are at the moment. (something like this `data:application/json;base64,ewogICAgInBlcnNvbiI6IHsKICAgICAgICAiYWdlIjogMjIKICAgIH0KfQo=`)

I have a working version that does both of these so if its wanted I can create a pull request to maybe get into the 1.3.0 main release.

#### Background:
Following #9083 / #6096 I saw that the new widget is based on the Panels one, but it seems they have some sort of logic to split up the file string in the python part of the code after the ts bit. The new bokehjs code just does this logic in the ts instead.

jbeanland:
Replaces the `file` attribute in the FileInput widget with separate `value` and `mime_type` attributes. Also adds a `filename` attribute.

- [x] issues: fixes #9099
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/widgets/file_input.ts
+++ b/bokehjs/src/lib/models/widgets/file_input.ts
@@ -34,11 +34,11 @@ export class FileInputView extends WidgetView {
   }
 
   file(e: any): void {
-    var file = e.target.result
-    var file_arr = file.split(",")
+    const file = e.target.result
+    const file_arr = file.split(",")
 
-    var content = file_arr[1]
-    var header = file_arr[0].split(":")[1].split(";")[0]
+    const content = file_arr[1]
+    const header = file_arr[0].split(":")[1].split(";")[0]
 
     this.model.value = content
     this.model.mime_type = header
--- a/bokehjs/src/lib/models/widgets/file_input.ts
+++ b/bokehjs/src/lib/models/widgets/file_input.ts
@@ -28,19 +28,30 @@ export class FileInputView extends WidgetView {
 
   load_file(e: any): void {
     const reader = new FileReader()
+    this.model.filename = e.target.files[0].name
     reader.onload = (e) => this.file(e)
     reader.readAsDataURL(e.target.files[0])
   }
 
   file(e: any): void {
-    this.model.file = e.target.result
+    var file = e.target.result
+    var file_arr = file.split(",")
+
+    var content = file_arr[1]
+    var header = file_arr[0].split(":")[1].split(";")[0]
+
+    this.model.value = content
+    this.model.mime_type = header
+
   }
 }
 
 export namespace FileInput {
   export type Attrs = p.AttrsOf<Props>
   export type Props = Widget.Props & {
-    file: p.Property<string>
+    value: p.Property<string>
+    mime_type: p.Property<string>
+    filename: p.Property<string>
     accept: p.Property<string>
   }
 }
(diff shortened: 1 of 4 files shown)


bryevdv:
This looks good, if you have a little more bandwidth a simple example in `examples/plotting/file` would cap this off. No worries if not, though.

jbeanland:
Thanks, I should be able to manage that, I'll fix these failed checks too, looks like my introduction to Typescript didn't go perfectly.

bokehjs/src/lib/models/widgets/file_input.ts
```diff
@@ -28,19 +28,30 @@ export class FileInputView extends WidgetView {
 
   load_file(e: any): void {
     const reader = new FileReader()
+    this.model.filename = e.target.files[0].name
     reader.onload = (e) => this.file(e)
     reader.readAsDataURL(e.target.files[0])
   }
 
   file(e: any): void {
```
mattpap: If you are at it, you could also improve those `any` types. This one should be replaced with `ProgressEvent`. You can learn precise types from https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts (interface `FileReader`).

jbeanland:
I've started hitting walls. There's a filing integration test, with tests/codebase/test_code_quality.py failing for a trailing whitespace check but locally that one passes.

On the example front I hit a weird thing, maybe some sort of bug in CustomJS or maybe I just missed some sort of build step:

If I try to setup a CustomJS callback on a file I'm outputting with `show()` then the callback itself acts as if the new attributes don't exist but the old one (`file`) does. The widget itself is correct because if I try to initiate it with something for `file` it will fail on AttributeError.

But if I run the same thing using `bokeh serve` then the callback will work properly.

bryevdv:
@jbeanland are you setting `BOKEH_RESOURCES=inline`? By default Bokeh uses BokehJS from CDN for standalone content, so if you want to use a locally built BokehJS you have to tell it to

bryevdv:
@jbeanland For the linter error, merge/rebase on master I was editing the README in the web UI and left a trailing space on one commit evidently (but its fixed now)

bryevdv:
@jbeanland one last thing the new example needs a baseline. Can you please check in an empty file at

> tests/baselines/examples/plotting/file/file_input

Then after the baseline tests fails again, it will print a diff and the file can be updated with the actual proper value

jbeanland:
@bryevdv Beauty, I've just been trying to figure out how to make those work. I'll check that in and hopefully(...) we'll be there.

jbeanland:
@bryevdv so to confirm I guess, I should wait for the inevitable failure, go get the diff, put that diff in that new empty file, then recommit with the new, expected test data in there?

bryevdv:
@jbeanland that is 100% correct!

bryevdv:
I paused a few other jobs to get this one up and running sooner: https://travis-ci.org/bokeh/bokeh/jobs/561216250

bryevdv:
@jbeanland The diff from the faillure is here:

https://travis-ci.org/bokeh/bokeh/jobs/561216250#L2884

jbeanland:
Awesome! Done.

bryevdv:
Thanks @jbeanland !

jbeanland:
@bryevdv No worries, happy to be able to help with the project! Thanks for guiding me through it.

PR #9109, opened 2019-07-24
Build on Read the Docs

humitos:
Hi all!

This PR includes all the work that Read the Docs folks have been doing in the last weeks coordinated by @bryevdv.

I'm happy to finally make this PR in the official repository and have Bokeh folks onboarding on Read the Docs!

### Changes

- Create a `docs.txt` requirement file
- Create a `readthedocs.yml` file
- Use `BOKEH_BUILD_JS` on `setup.py` to avoid user input when installing under Read the Docs
- Upgrade `npm` from `conf.py` when building since Read the Docs Docker image has an old version currently
- Create symlink for examples and Sphinx to be picked and rendered
- Set `BOKEH_DOCS_CDN` properly
  - with `test:` prefix to use Resources (JS/CSS) from Read the Docs when building non-tagged versions
  - to use CDN when tagged versions
- Set `BOKEH_DOCS_VERSION` from `READTHEDOCS_VERSION`
- Download all sample data when running on Read the Docs
- Remove Bokeh's custom version selector and use Read the Docs flyout menu instead
- Remove version warning (`version.txt` file and AJAX call)
  - Read the Docs already comes with this feature and can be enabled if needed
- Compile CSS and JS under a specific directory structure
  - Bokeh currently have these URLs hardcoded, so the default `_static` can't be used
- Ignore Google API key for now
- Set `bokeh_plot_use_relative_path = True` to deal with JS files properly
- Enable Read the Docs' Search engine
  - Add `Search` navbar link
  - Add `search.html` template to render the search page
- Modify the `layout.html` from Bokeh customized theme to add common scripts/css blocks
- Make some CSS changes to fix `z-index` property interfering with the flyout menu

### Onboarding

This branch can be activated as a Version on Read the Docs and should build properly latest changes from today's `master` (BokehJS _and_ documentation).

### Online documentation on Read the Docs (temporal)

I created a private project on https://readthedocs.com to show how the docs will look like once imported on Read the Docs. You can check out them at https://bokeh-bokeh.readthedocs-hosted.com/en/latest/


docs.txt
```diff
@@ -0,0 +1,16 @@
...
+pillow
+pscript
+python-dateutil
+pyyaml
+scipy
+setuptools
+six
+tornado
+
+sphinx>1.8
```
bryevdv: Is a text file the only option for specifying all these? Right now we have a single source of truth for all of our build/test/run dependencies in the conda recipe `meta.yaml`. I'd really hate to have to go back to duplicating information in different places it is always a maintenance hassle

readthedocs.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+build:
+  image: latest
+
+formats: []
+
+python:
+  version: 3.7
+  install:
+    - requirements: docs.txt
```
bryevdv: Could we just list the requriements here in this file rather than in a separate docs.txt? That at least would be an improvement since then depedencies are always in yaml files in all cases, not plain text files.  Also, can this be `.readthedocs.yml`? Trying to reduce clutter a the top level for new devs

setup.py
```diff
@@ -72,6 +72,18 @@
...
 )
 
+
+import os
+if os.environ.get('READTHEDOCS') == 'True':
+    # Workaround to upgrade ``npm`` since Read the Docs build environment have
+    # ``3.5.2`` and we need at least ``npm>=6.0``
+    os.system('cd ~ ; mkdir bin ; npm install npm')
+    os.environ['PATH'] = '/home/docs/node_modules/.bin/:{}'.format(os.environ.get('PATH'))
+    print('PATH: {}'.format(os.environ.get('PATH')))
```
bryevdv: Let's put all this in a function in `_setup_helpers.py`, e.g. `fixup_npm_for_rtd()`

setup.py
```diff
@@ -72,6 +72,18 @@
...
+
+    # Tell Bokeh install script to always build the Javascript files
+    os.environ['BOKEH_BUILD_JS'] = '1'
```
bryevdv: Let's put all this in a function in `_setup_helpers.py`, e.g. `fixup_for_rtd()`

sphinx/source/bokeh_theme/layout.html
```diff
@@ -28,32 +28,55 @@
...
+  {%- for js in script_files %}
+  {{ js_tag(js) }}
+  {%- endfor %}
+{%- endmacro %}
+
+{%- block scripts %}
   <script type="text/javascript">
-    var _BOKEH_CURRENT_VERSION = "{{ VERSION }}";
+   var _BOKEH_CURRENT_VERSION = "{{ VERSION }}";
   </script>
```
bryevdv: This could just be removed I think? It only exists for our old version chooser which is being removed

sphinx/source/bokeh_theme/layout.html
```diff
@@ -28,32 +28,55 @@
...
   <script src="{{ pathto('_static/' + theme_js, 1) }}"></script>
+{%- endblock %}
 
   <!-- Favicon
   –––––––––––––––––––––––––––––––––––––––––––––––––– -->
   <link rel="icon" type="image/png" href="{{ pathto('_static/' + theme_favicon, 1) }}">
 
-  <!-- Canonical
-  –––––––––––––––––––––––––––––––––––––––––––––––––– -->
-  <link rel="canonical" href="https://bokeh.pydata.org/en/latest/{{ pagename }}{{ file_suffix }}" />
```
bryevdv: Should there still be cannonical links, but pointing to `docs.python.org` ?

sphinx/sphinx
```diff
@@ -0,0 +1 @@
+../sphinx
```
bryevdv: What is this for (and also the `../examples` one above)?

sphinx/source/conf.py
```diff
@@ -275,3 +270,70 @@
...
+    readthedocs_version = os.environ.get('READTHEDOCS_VERSION')
+    if is_tagged_version():
+        # Use Bokeh Resources from CDN
+        os.environ['BOKEH_DOCS_CDN'] = readthedocs_version
+    else:
+        # Use Bokeh Resources from Read the Docs
+        os.environ['BOKEH_DOCS_CDN'] = f'test:{readthedocs_version}'
+
+    bokeh_docs_cdn = os.environ.get('BOKEH_DOCS_CDN')
+    print(f'BOKEH_DOCS_CDN: {bokeh_docs_cdn}')
```
bryevdv: We should avoid f-strings for now. We still support Python 2.7 until Bokeh 2.0 and even after that we will still support Python 3.5 and we should make it so anyone can contribute to the docs using a supported version

sphinx/sphinx
bryevdv: These are actually breaking our custom linter script at the moment, whatever they are

bryevdv:
```
           if path.startswith(exclude_dirs):
                continue
    
>           with io.open(path, 'r', encoding='utf-8') as file:
E           IsADirectoryError: [Errno 21] Is a directory: 'sphinx/examples'

```

bryevdv:
@humitos thanks for the PR I have left some small comments questions. Apart from those I'd only note that something in this PR is hangin `pytest` on our Python 3.5 "minimal" build:

https://travis-ci.org/bokeh/bokeh/jobs/563005462

docs.txt
humitos: Using a `requirements.txt` file is the standard way of how Python packages specify their dependencies. Read the Docs follows that pattern and uses `pip` to install them.

Dependencies for the documentation, tests, lint, build, etc are usually different. That's why I created a new file for docs requirements.

I don't have too much knowledge about conda. Is this the file that you mention? https://github.com/bokeh/bokeh/blob/36db0337d0f459bd44c3642e4f06d7f284246b7f/conda.recipe/meta.yaml

Read the Docs supports `conda` on its building process, but we don't usually recommend to use it unless it's strictly required. See https://docs.readthedocs.io/en/stable/guides/conda.html

readthedocs.yml
humitos: > Could we just list the actual requirements here in this file rather than in a separate docs.txt?

No, yet. We do have an issue open to support specifying dependencies in this file: https://github.com/readthedocs/readthedocs.org/issues/5572

> Also, can this be `.readthedocs.yml`?

Yes.

sphinx/source/bokeh_theme/layout.html
humitos: Gotcha! I wasn't sure if it was required or not.

sphinx/source/bokeh_theme/layout.html
humitos: No.

The canonical link is removed from here because it automatically added by Read the Docs.

In the current online (temporal) version, Read the Docs adds:

```html
<link rel="canonical" href="https://bokeh-bokeh.readthedocs-hosted.com/en/latest/" />
```

Once the real domain is setup on Read the Docs, it will point to the domain that you setup. There more docs on this at https://docs.readthedocs.io/en/latest/guides/canonical.html

sphinx/sphinx
humitos: When I started working on Bokeh some time ago, without these symlinks, Sphinx was not able to find the example and some rst source files. That's why I had to add them.

Maybe the latest changes on `master` branch have fixed this and it's not needed anymore. I will remove them and trigger another build on Read the Docs.

docs.txt
bryevdv: Right, we specify *runtime* dependencies in `setup.py` an then both `meta.yaml` and `requirements.txt` load from that single source of true. 

For test, dev dependencies, we're definitely 100% conda-based for recommended local dev usage. After a long effort we finally have a process that most new users can attempt to do and have a high chance of success. I don't care so much if RTD uses pip or conda but I am concerned about finding ways to avoid duplicating information in multiple places.

bryevdv: FYI the way we normally install test/dev deps is:
```
conda install `python scripts/deps.py test`
```
from the top level

humitos:
> Apart from those I'd only note that something in this PR is hangin `pytest` on our Python 3.5 "minimal" build:

I'm not sure why this is happening. I tried to run the pytest command locally but couldn't make it work. I suppose that it's blocked on the `Choice` input, maybe?

sphinx/source/conf.py
```diff
@@ -275,3 +270,76 @@
...
+    print('BOKEH_DOCS_VERSION: {readthedocs_version}'.format(
+        readthedocs_version=readthedocs_version,
+    ))
+
+    build_js = os.environ.get('BOKEH_BUILD_JS')
+    print('BOKEH_BUILD_JS: {build_js}'.format(build_js=build_js))
+
+    # Download all the sample data to plot
+    import bokeh
+    bokeh.sampledata.download(progress=False)
```
bryevdv: i don't think this should happen in `conf.py` In general I think `conf.py` should try to avoid executing code much beyond simple assignments. But this this is also wasteful in the local usage scenario, both of our CDN bandwidth as well as of users time having to re-download data even time they do a build. The docs build should assume sample data is installed (and it should be done as part of the build process up front)

sphinx/source/conf.py
```diff
@@ -275,3 +270,76 @@
...
+
+    # Create the proper structure to serve CSS and JS files compiled by
+    # Bokeh. We can't use ``html_static_path`` here because the URLs are
+    # hardcoded to ``static/js`` instead of ``_static/js``, etc
+    os.system('mkdir -p bokehJSbuildstatic')
+    os.system('cd bokehJSbuildstatic; ln -s ../../../bokehjs/build static')
```
bryevdv: I think this would break on Windows, e.g so perhaps anything like this should be conditioned on the "is this RTD?" environment var

sphinx/source/conf.py
humitos: This is under a `if` clause, and will be executed _only_ when the build is ran under Read the Docs (`READTHEDOCS` environment variable will be `True`).

See https://docs.readthedocs.io/en/stable/builds.html#build-environment for more details on environment variables defined by Read the Docs.

bryevdv: I missed above that this was in a conditioned branch. I suppose this OK tho it might be nice to move some of this out to a support module. I like to keep `setup.py` and `conf.py` etc as short and simple as possible

bryevdv:
> I suppose that it's blocked on the Choice input, maybe?

That thought occurred to me tho I don't know why `pytest` would be calling `setup.py` at all, so I am not certain it is that

_setup_helpers.py
```diff
@@ -0,0 +1,15 @@
...
+    os.system('cd ~ ; mkdir bin ; npm install npm')
+    os.environ['PATH'] = '/home/docs/node_modules/.bin/:{}'.format(os.environ.get('PATH'))
+    print('PATH: {}'.format(os.environ.get('PATH')))
+
+def set_env_to_build_bokehjs()
+    # TODO: this can be defined using Environment Variables on Read the Docs
+    # https://docs.readthedocs.io/en/latest/guides/environment-variables.html
+
+    # Tell Bokeh install script to always build the Javascript files
+    os.environ['BOKEH_BUILD_JS'] = '1'
```
bryevdv: @humitos sorry, I meant add to the existing file file, which is

https://github.com/bokeh/bokeh/blob/master/_setup_support.py

I misquoted the name as `_setup_helpers`

bryevdv:
The only differences in that one build is that

* it uninstalls some packages to make the build against a minimal install:

> conda remove -n testenv --yes --quiet --no-pin pandas scipy notebook scikit-learn sympy flask

* it excludes `sampledata`:

> py.test -m 'unit and not sampledata' --cov=bokeh --cov-config=bokeh/.coveragerc

humitos:
@bryevdv it seems that all the tests passed on Travis and AppVeyor. The only one that was not triggered on Travis is the task for deploy, which makes sense to me.

bryevdv:
> The only one that was not triggered on Travis is the task for deploy, which makes sense to me.

It always runs, but does not do anything unless there is a release tag (that's just how travis works, AFAIK) It had not run because of a spurious test failure ealier that I restarted

docs.txt
humitos: Bokeh dependency system is currently too custom for any solution to work on Read the Docs. Unfortunately, we didn't find a solution that does not involves a `requirements.txt` file (`docs.txt` in this case).

In the future, Read the Docs will support to install these docs dependencies directly from the `.readthedocs.yml` file (See https://github.com/readthedocs/readthedocs.org/issues/5572), which still won't support how Bokeh manages its dependencies, but at least it will be better for this case.

Let's use a `docs.txt` file for now, and we should revisit this when Read the Docs has support for arbitrary package name installation.

bryevdv:
@humitos I can't resolve this conflict on the UI, can you push he update? The deleted `all_versions.txt` was updated for the 1.3.1 hotfix release

bryevdv:
Going to go ahead and merge this to start using things to see what further changes are needed on our end or on RTD side.

PR #9116, opened 2019-07-28
Make min_characters configurable on AutoComplete

### LINKED ISSUE #4870: Autocomplete should start with a single character
The `AutocompleteInput` widget looks and works great in a Jupyter notebook (with `output_notebook`), but has styling and functional bugs when used with bokeh's server. For example:

```
from bokeh import plotting as bk
from bokeh.models import AutocompleteInput

ac = AutocompleteInput(title="Some countries", value='Albania', \
completions=['Albania','Austria','Australia','Angola','Argentina'])
bk.curdoc().add_root(ac)
```

Yields this when run with `bokeh serve`

screen shot 2016-07-23 at 6 44 20 pm

Notice how the matched completions show up as an unordered list instead of a dropdown. Any bokeh widgets or figures below the autocomplete text box will cover up this list of completions. Also, when you enter characters, the text "[N] results are available..." appears, which does not happen in Jupyter notebooks.

Finally, `on_change('value', myfunc)` does not seem to get triggered when you try to use autocompletion options. If you type what you want directly without using the arrow keys and then press enter, `on_change` triggers successfully.

I noticed that when using `output_notebook`, bokeh gets its styling and javascript from a CDN. I tried replacing bokeh's static JS and CSS with the files from the CDN, which had no effect.

```
$ bokeh info
Python version      :  2.7.12 (default, Jul 22 2016, 17:42:15) 
IPython version     :  5.0.0
Bokeh version       :  0.12.0
BokehJS static path :  /opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/bokeh/server/static
```

bryevdv:
- [x] issues: fixes #4870
- [x] tests added / passed

This PR adds a configurable `min_characters` property to `AutoComplete` that controls how many characters are needed to be input before completions will be presented. The existing default of 2 is retained. 

This PR also adds a `PositiveInt` property, and fixes missing `NonNegativeInt` documentation.


bryevdv:
@bokeh/core I think the only question is should the default `min_characters` be kept at 2 (what is has been previously) or changed to 1?

bryevdv:
OK there has not been any comment so I will go ahead and merge this. If folks really want we can change `min_characters` to 1 for Bokeh 2.0

PR #9121, opened 2019-07-29
Changes faulty type declaration in docstrings

### LINKED ISSUE #9101: [BUG] Wrong option names in sphinxext documentation
Hi guys, 
I am currently using the sphinxextension to build the documentation of my project. The module bokeh-plot is particularly usefull. While working with bokeh plot I noticed the following Issue.

Issue:
------

On the hompage: https://bokeh.pydata.org/en/latest/docs/reference/sphinxext.html#module-bokeh.sphinxext.bokeh_plot 
The two options for bokeh_plot are described:

grafik

When using this options in an bokeh-plot directive this options can not be found
(see Console Output)

Actually I already found the problem:
In the implementation of the Directory (`bokeh.sphinxext.bokeh_plot.py`) the names of the options are simply different:

-----------------------------------------------------------------------------
    class BokehPlotDirective(Directive):

        has_content = True
        optional_arguments = 2

        option_spec = {
            'source-position': lambda x: choice(x, ('below', 'above', 'none')),
            'linenos': lambda x: True if flag(x) is None else False,
            ...
        }
--------------------------------------------------------------------------
Expected Behaviour/Solution:
-------------------------------

In my opinion the documentation should be updated from:
* source-positionenum(‘above’, ‘below’, ‘none’)
* linenosflag

to:
* source-position(‘above’, ‘below’, ‘none’)
* linenos

Like in the source code.

Environment:
-------------
python=3.6.8
bokeh=1.2.0
Sphinx=2.1.2

Console Output:
-----------------
C:\HGroot\spam-visualizers\GeoViewsion\doc\how_to.rst:112: WARNING: Error in "bokeh-plot" directive:
unknown option: "linenosflag".

.. bokeh-plot:: ../examples/minimum_map/main.py
        :linenosflag:
        :source-positionenum: above

Miiike89:
fixes #9101

### DIFF
--- a/bokeh/sphinxext/bokeh_plot.py
+++ b/bokeh/sphinxext/bokeh_plot.py
@@ -42,11 +42,11 @@ used in docstrings.
 
 The ``bokeh-plot`` directive accepts the following options:
 
-source-position: enum('above', 'below', 'none')
+source-position (enum('above', 'below', 'none')):
     Where to locate the the block of formatted source
     code (if anywhere).
 
-linenos: bool
+linenos (bool):
     Whether to display line numbers along with the source.
 
 Examples
--- a/bokeh/sphinxext/bokeh_plot.py
+++ b/bokeh/sphinxext/bokeh_plot.py
@@ -42,11 +42,11 @@ used in docstrings.
 
 The ``bokeh-plot`` directive accepts the following options:
 
-source-position : enum('above', 'below', 'none')
+source-position: enum('above', 'below', 'none')
     Where to locate the the block of formatted source
     code (if anywhere).
 
-linenos : flag
+linenos: bool
     Whether to display line numbers along with the source.
 
 Examples
--- a/bokeh/sphinxext/collapsible_code_block.py
+++ b/bokeh/sphinxext/collapsible_code_block.py
@@ -20,9 +20,11 @@ This directive takes a heading to use for the collapsible code block:
 This directive is identical to the standard ``code-block`` directive
 that Sphinx supplies, with the addition of one new option:
 
-heading: string
+heading (string):
     A heading to put for the collapsible block. Clicking the heading
-    expands or collapses the block
+    expands or collapses the block.
+
+
 
 Examples
 --------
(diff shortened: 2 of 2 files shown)


bryevdv:
So i realized I stated that the docstring format was the issue, but did not state what the solution would look like. Bokeh docs build expects ["Google Style Docstring"](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) which basically means the type needs to go in parens before the colon, e.g. 

```
heading (string):
```

Miiike89:
@bryevdv i tried both alternatives. With enums it loogs terrible, ecpecially with th e parens. In my opinion the alternative looks better.

mattpap:
> Alternatively, we could also just leave the type off entirely, and just describe acceptable values in the description block below:

Having an explicit type is much better than having some vague explanation. It will make transition to mypy easier, and when that's done the discussion regarding style will be irrelevant.

Miiike89:
I can agree with that. How should I proceed from that?
Do I have to make another PR or can I make the cahnges within this one.

bryevdv:
@Miiike89  You don't need to make a new PR. If you commit new changes to your local branch and push again, any changes will automatically show up here. Can you post a screen shot of what the two alternatives look like when rendered by Sphinx?

Miiike89:
@bryevdv Here the screenshot. Actually it is just a styling thing and it doesn´t look so bad. I think more important is how to use it, which wasn´t clear before.

grafik

What do you say?

bryevdv:
@Miiike89 I think that looks great, let's go with that. It's late here, I will get things merged tomorrow when the build is green. Thanks!

Miiike89:
Okay. So you mean the top alternative in Google Docstring Style?

bryevdv:
Yes, the top one, sorry :)

Miiike89:
Perfekt. Iw ill push it right away.
Good night! :)

bryevdv:
Thanks @Miiike89 !

Miiike89:
No problem. Hopefully i can contribute in the future too.

PR #9138, opened 2019-08-03
Allow to integrate bokeh models with ipywidgets

### LINKED ISSUE #9137: Allow to integrate bokeh models with ipywidgets

mattpap:
Preliminary support for bokeh models and ipywidgets integration, based on [ipybokeh](https://github.com/pyviz/ipybokeh). This allows to use bokeh models (`LayoutDOM`) as a part of ipywidgets' layouts and allows for bidirectional communication between both kinds of widgets. ~~Mind failing `test:size`, due to inclusion of some jupyter libraries, but that will have to change significantly~~.

fixes #9137


mattpap:
I resolved the size issue. `python -OO` tests will fail due to https://github.com/jupyter-widgets/ipywidgets/issues/2524. The fix was already merged, but not released yet. Perhaps we could just skip testing `bokeh.notebook.widgets` in the meantime?

bryevdv:
Fine w me

bokeh/embed/notebook/widgets.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        return render_bundle, document
+
+    def update_from_model(self, model, document=None):
+        self._model = model
+        self.render_bundle, self._document = self._model_to_traits(model, document)
+        self._document.on_change_dispatch_to(self)
+
+    def _document_patched(self, event):
+        msg = Protocol("1.0").create("PATCH-DOC", [event])
```
bryevdv: noting this will have to be updated for 2.0 (no more proto versioning)

bokeh/embed/notebook/widgets.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self.send({"msg": "patch", "payload": msg.header_json})
+        self.send({"msg": "patch", "payload": msg.metadata_json})
+        self.send({"msg": "patch", "payload": msg.content_json})
+        for header, buff in msg.buffers:
+            self.send({"msg": "patch", "payload": json.dumps(header)})
+            self.send({"msg": "patch", "payload": buff})
```
bryevdv: This near-duplication is unfortunate. I suppose we could abstract out a preparation step but maybe its not worth it.

tests/codebase/test_python_execution_with_OO.py
```diff
@@ -31,6 +31,8 @@
 # Tests
 #-----------------------------------------------------------------------------
 
+blacklist = {"bokeh.embed.notebook.widgets"}
```
bryevdv: suggest making an issue to conditionally remove from blacklist if the ipywigets version is high enough and leaving a comment here pointing at the issue

bryevdv:
Are there any implications for the jlab plugin, or corresponding updates to make there?

bryevdv:
So I guess what people want to see is a Bokeh plat that is updated by manipulating ipywidgets. The example notebook should really demo this primary use case.

mattpap:
> Are there any implications for the jlab plugin, or corresponding updates to make there?

This change should be notebook/jlab agnostic. However, I can't test this, because I can't get ipywidgets working in jlab altogether (and it looks like I'm not the only one having those issues).

philippjfr:
>However, I can't test this, because I can't get ipywidgets working in jlab altogether (and it looks like I'm not the only one having those issues).

I'll try this today.

philippjfr:
Getting the following error in JupyterLab:

```
ncaught (in promise) Error: Module bokehjs, semver range 1.3.4-15-gfedd4e030-dirty is not registered as a widget module
    at WidgetManager.loadClass (vendors~main.e633818700579609537d.js:188524)
    at WidgetManager.<anonymous> (vendors~main.e633818700579609537d.js:225103)
    at step (vendors~main.e633818700579609537d.js:224884)
    at Object.next (vendors~main.e633818700579609537d.js:224865)
    at vendors~main.e633818700579609537d.js:224859
    at new Promise (<anonymous>)
    at push.hs2k.__awaiter (vendors~main.e633818700579609537d.js:224855)
    at WidgetManager.push.hs2k.ManagerBase._make_model (vendors~main.e633818700579609537d.js:225097)
    at WidgetManager.<anonymous> (vendors~main.e633818700579609537d.js:225086)
    at step (vendors~main.e633818700579609537d.js:224884)
```

Guessing JupyterLab requires you to register the widget somehow, so we will probably have to register it in the JLab extension somehow.

mattpap:
> Guessing JupyterLab requires you to register the widget somehow, so we will probably have to register it in the JLab extension somehow.

Yeah, I was way too naive to think this will just work. I will have to figure out how to run their own widgets in the first place.

mattpap:
I will go ahead and merge this, and I will add jlab support and more/better examples in future PR(s).

philippjfr:
I think you just need:

```
jupyter labextension install @jupyter-widgets/jupyterlab-manager
```

mattpap:
The problem was with incompatible versions of jupyter libraries installed by conda (for whatever reason).

bryevdv:
@mattpap sorry for the screen shot, I am running late for something. FYI this does not work:

Screen Shot 2019-08-14 at 4 30 04 PM

I confirmed the callback is executing. 

Also, re-executing just the last block immediately throws up

> RuntimeError: Models must be owned by only a single document, LinearScale(id='1787', ...) is already in a doc

which is the main reason I never really wanted pursue something like this (arbitrary notebook cell execution makes this a nightmare) I expect this will be a huge source of support questions if we heavily promote it.

bryevdv:
With that code I just noticed, there is a console error

Screen Shot 2019-08-14 at 4 34 02 PM

Issue #4820, opened 2016-07-16
Hover line width


aernlund:
Add hover_line_width option for hover tool.

bryevdv:
@aernlund Can you elaborate on precisely what you are asking for? There is not enough context / information for me to know for certain.

aernlund:
Sorry, I'm completely new bokeh as well as to contributing to projects on github.  Here is example code below:

p.rect('xname', 'yname', 0.9, 0.9, source=source,
       color='colors', alpha='alphas', line_color=None,
       hover_line_color='black', hover_color='colors', **hover_line_width=2**)

When I create a hover feature for rectangles, I can choose the color of the rectangle as well as the color of the line; hover line width (in bold) does not appear to be an option currently.  Hopefully, I'm not mistaken and wasting your time with this request.

bryevdv:
@aernlund no apologies necessary, everyone starts out new and help new comers is an important task we are very happy to do!

I would actually expect that to work, so if it is not, I would characterize that as a bug. Definitely not a waste of time! Let me try to reproduce on the most current dev build.

masoninman:
This would be extremely helpful. I think it's also related to #2367.

bryevdv:
FYI this works:
```
r = p.rect('xname', 'yname', 0.9, 0.9, source=source,
       color='colors', alpha='alphas', line_color=None,
       hover_line_color='black', hover_color='colors')
r.hover_glyph.line_width=4
```

It's only the convenience API in `figure` that is limited. Perhaps it can be improved, though. Will investigate for future release.

mjarosie:
Just double checked, PR #9153 fixes this issue.

PR #9153, opened 2019-08-10
Convert _pop_colors_and_alpha to _pop_visuals

### LINKED ISSUE #8893: provide selection_line_width and nonselection_line_width
I would like to be able to set a line width for selections. In my case these are "rect", so if implemented this would allow a nice effect of a "frame" around the selection.

I woud like for `selection_line_width` to be usable just like i.e. `selection_line_color`.
I would be willing to try to implement this if I get some guidance on how to proceed.

**Additional context**
In my particular application I have adjacent rectangles which are densely packed in some areas and spread out in other areas, so setting line_width to a larger value and playing with selection_line_alpha is not an option.

solstag:
Ni! This PR replaces `_pop_colors_and_alpha` with a new function `_pop_visuals` that works on all visuals in a more generic fashion. Details are documented in the code. As discussed in issue #8893, changes in property inheritance behavior should be noted in the release. In particular, visuals for a non prefixed glyph now serve as base for their prefixed counterparts (`hover_line_color` defaults to `line_color` if that is set), and aliases now work for all property traits ('selection_{y}' is an alias for 'selection_{x}_{y}' for any (x, y) where x is a visual feature).

Implemented and tested in `/bokeh/plotting/helpers.py`.
Implemented and not tested in `bokehjs/src/lib/api/plotting.ts`.

- [x] issues: fixes #8893, #4820
- [ ] tests added / passed

@bryevdv

### DIFF
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -292,10 +292,10 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
             Keys in `props` matching `glyphclass` properties with added `prefix`
               will get popped, other keys will be ignored.
             Keys take the form '[{prefix}][{feature}_]{trait}'. Only {feature}
-              must not containt underscores.
+              must not contain underscores.
             Keys of the form '{prefix}{trait}' work as lower precedence aliases
-              for all {features}.
-            Ex: {'fill_color': 'blue', 'selection_line_width': 0.5}
+              for {trait} for all {features}.
+            Ex: {'fill_color': 'blue', 'selection_line_width': 4}
 
         prefix (str) :
             Prefix used when accessing `props`. Ex: 'selection_'
@@ -326,13 +326,15 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
         feature, trait = split_feature_trait(ft)
         return feature in ('line', 'fill', 'text') and trait is not None
 
-    result = dict()
-    defaults = dict( [('text_color', 'black')] + list(defaults.items()) )
-    trait_defaults = dict( [('color', get_default_color()), ('alpha', 1.0)]
-                          + list(trait_defaults.items()) )
+    defaults = defaults.copy()
+    defaults.setdefault('text_color', 'black')
+    trait_defaults = trait_defaults.copy()
+    trait_defaults.setdefault('color', get_default_color())
+    trait_defaults.setdefault('alpha', 1.0)
 
+    result = dict()
     for pname in filter(is_visual, glyphclass.properties()):
-        trait = split_feature_trait(pname)[1]
+        _, trait = split_feature_trait(pname)
         if prefix+pname in props:
             result[pname] = props.pop(prefix+pname)
         elif prefix+trait in props:
@@ -341,13 +343,6 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
             result[pname] = defaults[pname]
         elif trait in trait_defaults:
             result[pname] = trait_defaults[trait]
-
-    # TODO: The need to do this and the complexity of managing this kind of
-    # thing throughout the codebase really suggests that we need to have
-    # a real stylesheet class, where defaults and Types can declaratively
-    # substitute for this kind of imperative logic.
-    # NOTE: text fill color should really always default to black, hard coding
-    # this here now until the stylesheet solution exists
 
     return result
 
@@ -930,27 +925,24 @@ def _glyph_function(glyphclass, extra_docs=None):
         if incompatible_literal_spec_values:
             raise RuntimeError(_GLYPH_SOURCE_MSG % nice_join(incompatible_literal_spec_values, conjuction="and"))
 
-        # use glyph_ca as default for other glyphs
-        default_ca = glyph_ca
-
         # handle the nonselection glyph, we always set one
-        nsglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='nonselection_', defaults=default_ca, trait_defaults={'alpha':0.1})
+        nsglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='nonselection_', defaults=glyph_ca, trait_defaults={'alpha':0.1})
 
         # handle the selection glyph, if any properties were given
         if any(x.startswith('selection_') for x in kwargs):
-            sglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='selection_', defaults=default_ca)
+            sglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='selection_', defaults=glyph_ca)
         else:
             sglyph_ca = None
 
         # handle the hover glyph, if any properties were given
         if any(x.startswith('hover_') for x in kwargs):
-            hglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='hover_', defaults=default_ca)
+            hglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='hover_', defaults=glyph_ca)
         else:
             hglyph_ca = None
 
         # handle the mute glyph, if any properties were given
         if any(x.startswith('muted_') for x in kwargs):
-            mglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='muted_', defaults=default_ca)
+            mglyph_ca = _pop_visuals(glyphclass, kwargs, prefix='muted_', defaults=glyph_ca)
         else:
             mglyph_ca = None
 
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -332,7 +332,7 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
     trait_defaults.setdefault('color', get_default_color())
     trait_defaults.setdefault('alpha', 1.0)
 
-    result = dict()
+    result, traits = dict(), set()
     for pname in filter(is_visual, glyphclass.properties()):
         _, trait = split_feature_trait(pname)
         if prefix+pname in props:
@@ -343,6 +343,9 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
             result[pname] = defaults[pname]
         elif trait in trait_defaults:
             result[pname] = trait_defaults[trait]
+        traits.add(trait)
+    for trait in traits:
+        props.pop(prefix+trait, None)
 
     return result
 
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -279,34 +279,43 @@ def _pop_renderer_args(kwargs):
     return result
 
 
-def _pop_colors_and_alpha(glyphclass, kwargs, prefix="", default_alpha=1.0):
+def _pop_colors_and_alpha(glyphclass, kwargs, prefix="", defaults={},
+                          trait_defaults={}):
     """
     Given a kwargs dict, a prefix, and a default value, looks for different
     color and alpha fields of the given prefix, and fills in the default value
     if it doesn't exist.
     """
     result = dict()
+    defaults, trait_defaults = defaults.copy(), trait_defaults.copy()
 
     # TODO: The need to do this and the complexity of managing this kind of
     # thing throughout the codebase really suggests that we need to have
     # a real stylesheet class, where defaults and Types can declaratively
     # substitute for this kind of imperative logic.
-    color = kwargs.pop(prefix + "color", get_default_color())
-    for argname in ("fill_color", "line_color"):
-        if argname not in glyphclass.properties():
-            continue
-        result[argname] = kwargs.pop(prefix + argname, color)
+    def is_visual(argname):
+        feature = argname.split('_', 1)[0]
+        return feature in ('line', 'fill', 'text')
 
     # NOTE: text fill color should really always default to black, hard coding
     # this here now until the stylesheet solution exists
-    if "text_color" in glyphclass.properties():
-        result["text_color"] = kwargs.pop(prefix + "text_color", "black")
+    if 'text_color' not in defaults:
+        defaults['text_color'] = 'black'
 
-    alpha = kwargs.pop(prefix + "alpha", default_alpha)
-    for argname in ("fill_alpha", "line_alpha", "text_alpha"):
-        if argname not in glyphclass.properties():
-            continue
-        result[argname] = kwargs.pop(prefix + argname, alpha)
+    if 'color' not in trait_defaults:
+        trait_defaults['color'] = kwargs.pop(prefix+'color', get_default_color())
+    if 'alpha' not in trait_defaults:
+        trait_defaults['alpha'] = kwargs.pop(prefix+'alpha', 1.0)
+
+    for argname in filter(is_visual, glyphclass.properties()):
+        prefixed = prefix + argname
+        trait = argname.split('_', 1)[1] if '_' in argname else None
+        if argname in defaults:
+            result[argname] = kwargs.pop(prefixed, defaults[argname])
+        elif trait in trait_defaults:
+            result[argname] = kwargs.pop(prefixed, trait_defaults[trait])
+        elif prefixed in kwargs:
+            result[argname] = kwargs.pop(prefixed)
 
     return result
 
@@ -890,7 +899,7 @@ def _glyph_function(glyphclass, extra_docs=None):
             raise RuntimeError(_GLYPH_SOURCE_MSG % nice_join(incompatible_literal_spec_values, conjuction="and"))
 
         # handle the nonselection glyph, we always set one
-        nsglyph_ca = _pop_colors_and_alpha(glyphclass, kwargs, prefix='nonselection_', default_alpha=0.1)
+        nsglyph_ca = _pop_colors_and_alpha(glyphclass, kwargs, prefix='nonselection_', trait_defaults={'alpha':0.1})
 
         # handle the selection glyph, if any properties were given
         if any(x.startswith('selection_') for x in kwargs):
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -187,42 +187,44 @@ def _graph(node_source, edge_source, **kwargs):
             raise ValueError(msg).with_traceback(sys.exc_info()[2])
 
     ## node stuff
+    node_ca = _pop_visuals(Circle, kwargs, prefix="node_")
+
     if any(x.startswith('node_selection_') for x in kwargs):
-        snode_ca = _pop_visuals(Circle, kwargs, prefix="node_selection_")
+        snode_ca = _pop_visuals(Circle, kwargs, prefix="node_selection_", defaults=node_ca)
     else:
         snode_ca = None
 
     if any(x.startswith('node_hover_') for x in kwargs):
-        hnode_ca = _pop_visuals(Circle, kwargs, prefix="node_hover_")
+        hnode_ca = _pop_visuals(Circle, kwargs, prefix="node_hover_", defaults=node_ca)
     else:
         hnode_ca = None
 
     if any(x.startswith('node_muted_') for x in kwargs):
-        mnode_ca = _pop_visuals(Circle, kwargs, prefix="node_muted_")
+        mnode_ca = _pop_visuals(Circle, kwargs, prefix="node_muted_", defaults=node_ca)
     else:
         mnode_ca = None
 
-    nsnode_ca = _pop_visuals(Circle, kwargs, prefix="node_nonselection_")
-    node_ca = _pop_visuals(Circle, kwargs, prefix="node_")
+    nsnode_ca = _pop_visuals(Circle, kwargs, prefix="node_nonselection_", defaults=node_ca)
 
     ## edge stuff
+    edge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_")
+
     if any(x.startswith('edge_selection_') for x in kwargs):
-        sedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_selection_")
+        sedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_selection_", defaults=edge_ca)
     else:
         sedge_ca = None
 
     if any(x.startswith('edge_hover_') for x in kwargs):
-        hedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_hover_")
+        hedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_hover_", defaults=edge_ca)
     else:
         hedge_ca = None
 
     if any(x.startswith('edge_muted_') for x in kwargs):
-        medge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_muted_")
+        medge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_muted_", defaults=edge_ca)
     else:
         medge_ca = None
 
-    nsedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_nonselection_")
-    edge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_")
+    nsedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_nonselection_", defaults=edge_ca)
 
     ## node stuff
     node_kwargs = {k.lstrip('node_'): v for k, v in kwargs.copy().items() if k.lstrip('node_') in Circle.properties()}
@@ -327,7 +329,7 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
     def is_visual(ft):
         """Whether a feature trait name is visual"""
         feature, trait = split_feature_trait(ft)
-        return feature in ('line', 'fill', 'text') and trait is not None
+        return feature in ('line', 'fill', 'text', 'global') and trait is not None
 
     defaults = defaults.copy()
     defaults.setdefault('text_color', 'black')
@@ -920,10 +922,6 @@ def _glyph_function(glyphclass, extra_docs=None):
         # Need to check if user source is present before _pop_renderer_args
         renderer_kws = _pop_renderer_args(kwargs)
         source = renderer_kws['data_source']
-
-        # Assign global_alpha from alpha if glyph type is an image
-        if 'alpha' in kwargs and glyphclass.__name__ in ('Image', 'ImageRGBA', 'ImageURL'):
-            kwargs['global_alpha'] = kwargs['alpha']
 
         # handle the main glyph, need to process literals
         glyph_ca = _pop_visuals(glyphclass, kwargs)
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -188,41 +188,41 @@ def _graph(node_source, edge_source, **kwargs):
 
     ## node stuff
     if any(x.startswith('node_selection_') for x in kwargs):
-        snode_ca = _pop_colors_and_alpha(Circle, kwargs, prefix="node_selection_")
+        snode_ca = _pop_visuals(Circle, kwargs, prefix="node_selection_")
     else:
         snode_ca = None
 
     if any(x.startswith('node_hover_') for x in kwargs):
-        hnode_ca = _pop_colors_and_alpha(Circle, kwargs, prefix="node_hover_")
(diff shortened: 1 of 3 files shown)


bryevdv:
@solstag please accept my apologies, it seems I actually never hit this submit button for the review comments.

bokeh/plotting/helpers.py
```diff
@@ -821,31 +862,34 @@ def func(self, **kwargs):
...
-        glyph_ca = _pop_colors_and_alpha(glyphclass, kwargs)
+        glyph_ca = _pop_visuals(glyphclass, kwargs)
         incompatible_literal_spec_values = []
         incompatible_literal_spec_values += _process_sequence_literals(glyphclass, kwargs, source, is_user_source)
         incompatible_literal_spec_values += _process_sequence_literals(glyphclass, glyph_ca, source, is_user_source)
         if incompatible_literal_spec_values:
             raise RuntimeError(_GLYPH_SOURCE_MSG % nice_join(incompatible_literal_spec_values, conjuction="and"))
 
+        # use glyph_ca as default for other glyphs
+        default_ca = glyph_ca
```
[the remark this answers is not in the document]
solstag: It is logically innocuous, but it made it locally obvious what is happening in case someone doesn't know the arguments of `_pop_visuals`. It can be taken down, no prob.

bokeh/plotting/helpers.py
```diff
@@ -278,34 +278,75 @@ def _pop_renderer_args(kwargs):
...
+    Args:
+        glyphclass :
+            the type of glyph being handled
+
+        props (dict) :
+            Maps properties and prefixed properties to their values.
+            Keys in `props` matching `glyphclass` properties with added `prefix`
+              will get popped, other keys will be ignored.
+            Keys take the form '[{prefix}][{feature}_]{trait}'. Only {feature}
+              must not containt underscores.
```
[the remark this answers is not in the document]
solstag: That's not a typo... it's Frenglish! (contains + contient = containt) ;-P

bokeh/plotting/helpers.py
```diff
@@ -278,34 +278,75 @@ def _pop_renderer_args(kwargs):
...
+
+    def is_visual(ft):
+        """Whether a feature trait name is visual"""
+        feature, trait = split_feature_trait(ft)
+        return feature in ('line', 'fill', 'text') and trait is not None
+
     result = dict()
+    defaults = dict( [('text_color', 'black')] + list(defaults.items()) )
+    trait_defaults = dict( [('color', get_default_color()), ('alpha', 1.0)]
+                          + list(trait_defaults.items()) )
```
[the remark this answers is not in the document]
solstag: I'm not sure I understand your comment. The current code is the exact equivalent of:
```
defaults = defaults.copy()
defaults.setdefault('text_color', 'black')
trait_defaults = trait_defaults.copy()
trait_defaults.setdefault('color', get_default_color())
trait_defaults.setdefault('alpha', 1.0)
```
My feeling was that, since we're required to create new dictionaries because we don't want to modify those passed as defaults, we might as well state their contents in the constructor.

I'll be happy to use the `setdefault` version if you still prefer it, let me know!

bokeh/plotting/helpers.py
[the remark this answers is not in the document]
solstag: Actually, I probably set that `default_ca` before choosing/noting that the keyword argument is explicit. Since we're always called `_pop_visuals( ..., defaults=glyph_ca, ...)` this is really redundant. Gone!

bokeh/plotting/helpers.py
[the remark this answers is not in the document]
bryevdv: Ah it's not obvious (or known before) how the `dict` constructor responds in the presence of repeated keys in tuples. I'd personally prefer the `setdefault` version as clearer (to me)

mattpap:
This will need a rebase, because the base branch `landing-2.0` is being rebased as well. I will assume that `bokeh/bokeh` repository in your fork is referred to as `bokeh` remote. If that's not the case, use the correct name of your choosing. If you didn't add a remote earlier, you will have to do it now with:
```
git remote add bokeh git@github.com:bokeh/bokeh.git
```
In any case, issue:
```
git fetch bokeh
```
and to rebase issue:
```
git rebase --onto bokeh/landing-2.0 145a50dd8c^
```
and finally:
```
git push -f origin style_visuals
```
The commit in the rebase command is the first commit that you introduced. If you need to rebase this PR in the future, which is very likely as `landing-2.0` is continuously rebased, you will have to figure out the SHA of first commit on your own, as those change with each rebase, and update the rebase command.

bryevdv:
@solstag aplogies, but I think this brach has gotten in a bad state. There were some other changes that were needed the last few days because of an upstream problem breaking out windows builds. The commit history already looks messed up so I don't think it's advisable to try and pull in those changes here. At this point, I would ask that you submit a fresh PR on a current clean up to date master, and we close (supercede) this one.

mattpap:
The commit history is messed up because this branch is based off `landing-2.0`. It will need another `rebase --onto`, unless it's intended to switch the base branch.

bryevdv:
@mattpap This is intended to be merged against `landing-2.0`

mattpap:
btw. One can perform the rebase in one command (here split for readability), assuming that `git fetch` was run but local `landing-2.0` branch wasn't updated with `origin/landing-2.0`:
```
FIRST=$(git rev-list style_visuals ^landing-2.0 | tail -n 1)
git rebase --onto origin/landing-2.0 $FIRST~1
```
The magic happens in `git rev-list`, which is a low-level equivalent of `git log`. We are taking the difference (in commits) between this branch and old `landing-2.0`, which should give us four commits, and the we take the initial one. From there it's a usual `rebase --onto`, if that kind of rebase can be called usual at all.

solstag:
I've rebased the branch once more. No worries, in the future just ask me to rebase again when needed. Cheers!

bryevdv:
@solstag Thank you for your infinite patience! 1.4 is taking a bit longer to get out that we had hoped but 2.0 merges will proceed quickly once it does. I think there are currently some real test failures on this branch. Am seeing messages like this in the test logs:

> AttributeError: unexpected attribute 'color' to VBar, similar attributes are line_color, fill_color or hatch_color

solstag:
Hi @bryevdv !
Yeah, there was a silly bug, I have fixed it and rebased the branch one more as it had diverged.
Let's see if the tests give away any other issues.
.~´

solstag:
Ni! So, I took the time today to study some typescript and the workings of bokehjs, and I believe I managed to convert it to _pop_visuals as well. However, this is the first time I try typescript and **I have not tested it at all**, so some review is granted. But from what I could get it should be easy to finish in the likely case I missed something. Abraços .~´

bokehjs/src/lib/api/plotting.ts
```diff
@@ -630,30 +630,47 @@ export class Figure extends Plot {
...
-    const color = _with_default(attrs[prefix + "color"] as any, default_color)
-    const alpha = _with_default(attrs[prefix + "alpha"] as any, default_alpha)
+    const _split_feature_trait = function(ft: string): string {
+      ft = ft.split('_', 2)
+      if (ft.length==2) { return ft }
+      else { return ft.concat([null]) }
+    }
+    const _is_visual = function(ft: string): boolean {
+      [feature, trait] = _split_feature_trait(ft)
+      return (feature in ['line', 'fill', 'text']) && (trait != null)
```
mattpap: > `feature in ['line', 'fill', 'text']`

Unfortunately this isn't Python, so the meaning of this code is pretty surprising. Let me explain. An array in JS can be viewed as a key-value map (or an object), where keys are integer indices and values are actual values in an array. `in` operator works on objects and checks if a key exists in an object. So you can ask if an index exists in an array (not particularly useful for dense arrays, but sparse arrays are perfectly possible), so you can ask if e.g. `0 in ["a value"]`, which is true, but asking if `"a value" in ["a value"]` doesn't make sense, be cause Array object doesn't have `a value` property. The correct way is to write `["a value"].includes("a value")`. TypeScript doesn't complain about this, because despite the code not making sense in the way we would expect it to work, is still perfectly valid JavaScript and has merit in some contexts. At some point we will have ESLint rules that would warn about this kind of nonsense code. This also makes efficient arrays sometimes tricky, because it's easy to turn a dense array into a sparse one, which disables any JS engine's optimizations, making operations such arrays slow.

solstag: Thanks, should be fixed now. Been a while since I last used JS, that one escaped me. :)

solstag:
I've pushed a fix for the case where a glyph has a property that has the same name as a trait.

For example, `rect` has "width" in addition to "line_width". In this case we don't want to use "width" as an alias for "{feature}_width".

The patched code only considers trait aliases if the glyph does not have a property by the same name as the trait.

bryevdv:
@solstag sorry, I screwed up your branch. There are lots of problems with `plotting.ts` so I just wanted to punt on that and see if the python portion worked (and get it merged alone). But unsurprisingly "delete file" actually deletes the file, not the changes to the file. (I probably should not try stuff like this when I am tired and under the weather...) Can you restore the file (without any changes) and let's see if we can't get the python changes merged.

solstag:
Hi @bryevdv ,

I'm not sure I understand, you want to merge the python changes but not (yet) the TS changes? I'm asking because I've always assumed they had to be in sync in order to be merged.

Once you confirm, I will branch out the TS changes like you said.

Also, just to clarify, do you mean there are problems with `plotting.ts` in general, or problems with my changes to `plotting.ts` ?

On a note, I've ported my last fix to the TS version as well.

bryevdv:
> problems with my changes to plotting.ts ?

There were problems with your changes causing TypeScript compile errors and failing the builds before tests even got started

> I'm not sure I understand, you want to merge the python changes but not (yet) the TS changes? I'm asking because I've always assumed they had to be in sync in order to be merged.

It's certainly the ideal but they are technically separate APIs. In this case I would like to prioritize getting this long-running PR merged. We can make a new issue to revisit the BokehJS side in a separate PR.

solstag:
> There were problems with your changes causing TypeScript compile errors and failing the builds before tests even got started

I saw those now. I had never opened the Travis results before, since it was giving out errors long before I touched anything TS related.

I improved the changes to a point where the tests did get started, but since some of them don't pass, I am doing as you requested and I have now pushed a version with the original plotting.ts from landing-2.0.

Let me know what to do next ! Cheers and thanks

bryevdv:
@solstag ok there are two real failures to figure out. This image test example fails to run now:

> bokeh/examples/plotting/file/image_alpha.py

It now complains that `alpha` is not vlaid to pass to `image`. 

Also this one unit test:

```
_______________ test__graph_properly_handle_edge_property_mixins _______________
    def test__graph_properly_handle_edge_property_mixins():
        kwargs = dict(edge_line_color="purple", edge_selection_line_color="blue",
                      edge_nonselection_line_color="yellow", edge_hover_line_color="red",
                      edge_muted_line_color="orange", edge_line_width=23)
    
        kw = bph._graph({}, {}, **kwargs)
    
        r = kw['edge_renderer']
        assert r.glyph.line_color == "purple"
        assert r.selection_glyph.line_color == "blue"
        assert r.nonselection_glyph.line_color == "yellow"
        assert r.hover_glyph.line_color == "red"
        assert r.muted_glyph.line_color == "orange"
    
        assert r.glyph.line_width == 23
>       assert r.selection_glyph.line_width == 23
E       AssertionError: assert 1 == 23
E        +  where 1 = MultiLine(id='19874', ...).line_width
E        +    where MultiLine(id='19874', ...) = GlyphRenderer(id='19877', ...).selection_glyph
```

solstag:
> > bokeh/examples/plotting/file/image_alpha.py
> It now complains that `alpha` is not valid to pass to `image`.

I see the error is:
> [PY] AttributeError: unexpected attribute 'alpha' to Image, possible attributes are color_mapper, dh, dh_units, dilate, dw, dw_units, global_alpha, image, js_event_callbacks, js_property_callbacks, name, subscribed_events, tags, x or y

From what I get, `Image` does not have a `{visual}_alpha` property, so my implementation is considering "alpha" to be none of its business and passing it forward. (And it would do the same in case the glyph had an `alpha` property as well as `{visual}_alpha` ones).

What was previously happening with `_pop_colors_and_alpha` in this case is that `alpha` was being popped but actually ignored, leading to the script to silently fail to deliver what the user expected. I'm guessing the expectation was for `global_alpha` to be affected, but `global` is not a visual feature.

I see a few possible fixes:
- remove this test.
- change the test replacing `alpha` with `global_alpha`.
- change the glyph renaming `global_alpha` as simply `alpha`
- consider "global" a visual feature, so "global_alpha" would be recognized.

(I'll check the other error once we're done with this one.)

bryevdv:
> What was previously happening with _pop_colors_and_alpha in this case is that alpha was being popped but actually ignored, leading to the script to silently fail to deliver what the user expected. I'm guessing the expectation was for global_alpha to be affected, but global is not a visual feature.

I'm not sure about this&mdash;if I run that actual example, the images do show up with an alpha applied. And if I look in the console:
```
In [1]: import numpy as np
   ...:
   ...: from bokeh.plotting import figure, show, output_file
   ...: from bokeh.layouts import Column
   ...:
   ...: # 1. Adding alpha for the image method
   ...:
   ...: N = 500
   ...: x = np.linspace(0, 10, N)
   ...: y = np.linspace(0, 10, N)
   ...: xx, yy = np.meshgrid(x, y)
   ...: d = np.sin(xx)*np.cos(yy)
   ...: d[:125, :125] = np.nan  # Set bottom left quadrant to NaNs
   ...:
   ...: p1 = figure(x_range=(0, 10), y_range=(0, 10))
   ...:
   ...: # Solid line to show effect of alpha
   ...: p1.line([0, 10], [0, 10], color='red', line_width=2)
   ...: # Use alpha kwarg to set alpha value
   ...: img = p1.image(image=[d], x=0, y=0, dw=10, dh=10, palette="Spectral11", alpha=0.1)

In [2]: img.glyph.global_alpha
Out[2]: 0.1
```
it is the case that setting `alpha` forwards on to `global_alpha` I have not had a chance to look in to what is making that happen, though I think it should be preserved going forward.

bokeh/plotting/helpers.py
```diff
@@ -882,31 +926,31 @@ def func(self, **kwargs):
             kwargs['global_alpha'] = kwargs['alpha']
```
bryevdv: @solstag it was happening here, presumably

bryevdv: To be clear I'm not sure why this isn't still working since it seems `_pop_visuals` is called later... all that said this is pretty hacky so I'd be open to any changes that keep the behavior in a less hacky way

solstag: I see.

Well, it doesn't work with _pop_visuals for the reason I mentioned. The parameter that is having an effect on the glyph is not `alpha`, but `global_alpha`.

The parameter `alpha` is being treated the way I mentioned, with `_pop_colors_and_alpha` incorrectly consuming it but doing nothing, and `_pop_visuals` not consuming it since it has nothing to do. However since the glyph doesn't have `alpha`, the correct behaviour will yield an error later on.

At the same time, the parameter `global_alpha` is ignored by both, and treated downstream. Thus it is only circumstantial that `_pop_colors_and_alpha` yields the expected result, since as as you noticed the code hacks the value of "alpha" into "global_alpha" (I missed that one).

So in my opinion we have two reasonable choices:
- treat "global" as a visual feature so that trait aliases will work as usual: not only for Image{,RGBA,URL} and not only for `_alpha`.
- choose between "global_alpha" or "alpha" and adapt the Image glyphs accordingly.

Not knowing the general usage of `global_` in bokeh, if I had to choose it by myself I would go for the second choice, keeping the most used variation. If designing from scratch I would have chosen something more meaningful like "image_alpha", as "image" would naturally go a "visual" feature.

Does that seem right? What do you think?

bryevdv: I think we have to keep both, unfortunately. The reason the model property is called `global_alpha` is because the image data itself an contain an alpha channel. This is to be clear that the alpha applies globally and additively on top of any alpha in the data. The `alpha` for the glyph *functions* is a convenience but at this point, a well-established one and changing it would break many people I suspect. I think this just has to be handled as a special case for the relevant glyphs. If we want to be fancy we can implement some general way to specify per-glyph aliases like this and then configure it appropriately. But I am also OK at this point with just putting a few lines of special case code `pop_visuals` to handle images.

bryevdv: I don't like making "global" a visual feature because as you say:

> not only for Image{,RGBA,URL} and not only for _alpha.

But those are truly the only places they apply (and should apply)

solstag: Hm... can't this be handled by the glyph itself? Exposing two properties that have the same semantics is not really a bad thing modulo documentation, and it would make something that is glyph specific consistent when handling the glyph with or without the plotting API.

Otherwise we'll end up with a "plotting API specific alpha" that does not respect the behaviour of "alpha in the plotting API". Let alone if `Image*` glyphs ever get a "[line|text]_alpha" or a "border_alpha", either in the future or if someone subclasses those in their application.

It doesn't seem nice to expose users to behaviour dependent on the product of glyphs and APIs. «If something is glyph specific it works with every API, if something is API specific it works with every glyph.» is a nice UX principle, and it seems to me (may be wrong as I am new to bokeh devel) that we're pretty close to that as far as the plotting API is concerned.

bryevdv: >  if Image* glyphs ever get a [line|text]_alpha or a border_alpha

They won't

> or if someone subclasses those in their application.

Just being honest: I don't care about this case (which also is very unlikely to actually happen)

Here is the thing to note: None of the glyphs have `alpha`. Not one. `alpha` is purely a convenience layer added at the `bokeh.plotting` layer. Making sure that `alpha` works reasonably (i.e. does the thing users will expect) for *all* the glyphs is exactly the most user-friendly thing to do. In the case of other glyphs, setting `alpha` is a convenience for setting both `line_alpha` and `fill_alpha`. That's no different from having it set `global_alpha` for images (because, for images, that is the reasonable thing users will expect to work). 

This snot a documentation problem, because making it work for everything means:

> Setting `alpha` in a glyph method is a shortcut for setting all the relevant "alpha"-like properties of a glyph

Contrast if we *don't* do that:

> Setting `alpha` in a glyph method is a shortcut for setting all the relevant "alpha"-like properties of a glyph, except for all these glyphs where that doesn't apply: ..., ...

That's worse for documentation.

bryevdv: I will say, if you are only willing to handle this by adding `global` as another "visual" than that's your call and I would accept the PR. But we need to have a *property* "global_alpha" that is aliased by "alpha" in `bokeh.plotting` glyph methods, however that is accomplished.

bryevdv:
@solstag  think my earlier reply was more crisp that I intended, please accept my apology. I have confirmed that this diff will get things working the way they should be:
```
diff --git a/bokeh/models/glyphs.py b/bokeh/models/glyphs.py
index bb7251c4d..b19b5b722 100644
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -485,7 +485,7 @@ class Image(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('image', 'x', 'y', 'dw', 'dh', 'global_alpha', 'dilate')
+    _args = ('image', 'x', 'y', 'dw', 'dh', 'dilate')

     # a hook to specify any additional kwargs handled by an initializer
     _extra_kws = {
@@ -554,7 +554,7 @@ class ImageRGBA(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('image', 'x', 'y', 'dw', 'dh', 'global_alpha', 'dilate')
+    _args = ('image', 'x', 'y', 'dw', 'dh', 'dilate')

     image = NumberSpec(help="""
     The arrays of RGBA data for the images.
@@ -609,7 +609,7 @@ class ImageURL(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('url', 'x', 'y', 'w', 'h', 'angle', 'global_alpha', 'dilate')
+    _args = ('url', 'x', 'y', 'w', 'h', 'angle', 'dilate')

     url = StringSpec(default=None, help="""
     The URLs to retrieve images from.
diff --git a/bokeh/plotting/helpers.py b/bokeh/plotting/helpers.py
index 428f1d125..327acc4d4 100644
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -327,7 +327,7 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
     def is_visual(ft):
         """Whether a feature trait name is visual"""
         feature, trait = split_feature_trait(ft)
-        return feature in ('line', 'fill', 'text') and trait is not None
+        return feature in ('line', 'fill', 'text', 'global') and trait is not None

     defaults = defaults.copy()
     defaults.setdefault('text_color', 'black')
@@ -921,10 +921,6 @@ def _glyph_function(glyphclass, extra_docs=None):

-        # Assign global_alpha from alpha if glyph type is an image
-        if 'alpha' in kwargs and glyphclass.__name__ in ('Image', 'ImageRGBA', 'ImageURL'):
-            kwargs['global_alpha'] = kwargs['alpha']
-
         # handle the main glyph, need to process literals
         glyph_ca = _pop_visuals(glyphclass, kwargs)
         incompatible_literal_spec_values = []
```
If you agree I think we should proceed on  this basis to the last remaining issue.

bryevdv:
And speaking of the last remaining issue, it appears to be that `_graph` does not apply the "_ca" defaults to all the other glyphs the way the glyph function does. A very old bug I think. This full diff gets everything passing I believe:
```
diff --git a/bokeh/models/glyphs.py b/bokeh/models/glyphs.py
index bb7251c4d..b19b5b722 100644
--- a/bokeh/models/glyphs.py
+++ b/bokeh/models/glyphs.py
@@ -485,7 +485,7 @@ class Image(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('image', 'x', 'y', 'dw', 'dh', 'global_alpha', 'dilate')
+    _args = ('image', 'x', 'y', 'dw', 'dh', 'dilate')

     # a hook to specify any additional kwargs handled by an initializer
     _extra_kws = {
@@ -554,7 +554,7 @@ class ImageRGBA(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('image', 'x', 'y', 'dw', 'dh', 'global_alpha', 'dilate')
+    _args = ('image', 'x', 'y', 'dw', 'dh', 'dilate')

     image = NumberSpec(help="""
     The arrays of RGBA data for the images.
@@ -609,7 +609,7 @@ class ImageURL(XYGlyph):

     # a canonical order for positional args that can be used for any
     # functions derived from this class
-    _args = ('url', 'x', 'y', 'w', 'h', 'angle', 'global_alpha', 'dilate')
+    _args = ('url', 'x', 'y', 'w', 'h', 'angle', 'dilate')

     url = StringSpec(default=None, help="""
     The URLs to retrieve images from.
diff --git a/bokeh/plotting/helpers.py b/bokeh/plotting/helpers.py
index 428f1d125..23c590534 100644
--- a/bokeh/plotting/helpers.py
+++ b/bokeh/plotting/helpers.py
@@ -187,42 +187,44 @@ def _graph(node_source, edge_source, **kwargs):
             raise ValueError(msg).with_traceback(sys.exc_info()[2])

     ## node stuff
+    node_ca = _pop_visuals(Circle, kwargs, prefix="node_")
+
     if any(x.startswith('node_selection_') for x in kwargs):
-        snode_ca = _pop_visuals(Circle, kwargs, prefix="node_selection_")
+        snode_ca = _pop_visuals(Circle, kwargs, prefix="node_selection_", defaults=node_ca)
     else:
         snode_ca = None

     if any(x.startswith('node_hover_') for x in kwargs):
-        hnode_ca = _pop_visuals(Circle, kwargs, prefix="node_hover_")
+        hnode_ca = _pop_visuals(Circle, kwargs, prefix="node_hover_", defaults=node_ca)
     else:
         hnode_ca = None

     if any(x.startswith('node_muted_') for x in kwargs):
-        mnode_ca = _pop_visuals(Circle, kwargs, prefix="node_muted_")
+        mnode_ca = _pop_visuals(Circle, kwargs, prefix="node_muted_", defaults=node_ca)
     else:
         mnode_ca = None

-    nsnode_ca = _pop_visuals(Circle, kwargs, prefix="node_nonselection_")
-    node_ca = _pop_visuals(Circle, kwargs, prefix="node_")
+    nsnode_ca = _pop_visuals(Circle, kwargs, prefix="node_nonselection_", defaults=node_ca)

     ## edge stuff
+    edge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_")
+
     if any(x.startswith('edge_selection_') for x in kwargs):
-        sedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_selection_")
+        sedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_selection_", defaults=edge_ca)
     else:
         sedge_ca = None

     if any(x.startswith('edge_hover_') for x in kwargs):
-        hedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_hover_")
+        hedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_hover_", defaults=edge_ca)
     else:
         hedge_ca = None

     if any(x.startswith('edge_muted_') for x in kwargs):
-        medge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_muted_")
+        medge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_muted_", defaults=edge_ca)
     else:
         medge_ca = None

-    nsedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_nonselection_")
-    edge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_")
+    nsedge_ca = _pop_visuals(MultiLine, kwargs, prefix="edge_nonselection_", defaults=edge_ca)

     ## node stuff
     node_kwargs = {k.lstrip('node_'): v for k, v in kwargs.copy().items() if k.lstrip('node_') in Circle.properties()}
@@ -327,7 +329,7 @@ def _pop_visuals(glyphclass, props, prefix="", defaults={}, trait_defaults={}):
     def is_visual(ft):
         """Whether a feature trait name is visual"""
         feature, trait = split_feature_trait(ft)
-        return feature in ('line', 'fill', 'text') and trait is not None
+        return feature in ('line', 'fill', 'text', 'global') and trait is not None

     defaults = defaults.copy()
     defaults.setdefault('text_color', 'black')
@@ -921,10 +923,6 @@ def _glyph_function(glyphclass, extra_docs=None):

-        # Assign global_alpha from alpha if glyph type is an image
-        if 'alpha' in kwargs and glyphclass.__name__ in ('Image', 'ImageRGBA', 'ImageURL'):
-            kwargs['global_alpha'] = kwargs['alpha']
-
         # handle the main glyph, need to process literals
         glyph_ca = _pop_visuals(glyphclass, kwargs)
         incompatible_literal_spec_values = []
```

bryevdv:
I also forgot to mention that there was a recent upstream issue on CI, so this branch will also need a merge/rebase on master to get passing.

bryevdv:
@solstag just an FYI, but there's a few issues that are blocked until this is merged (would be bad conflicts). So I plan to apply the diff above tonight or tomorrow (in case you don't get to it first) and then merge this. Thanks for all your patience this certainly took longer than is typical!

bryevdv:
Failure is unrelated issue with numpy version on the "minimal" build that would need a rebase to fix. But it will go away after the merge, so merging now! Thanks so much @solstag !

solstag:
Hi @bryevdv !

Sorry, I was travelling (to a free software conference here in France!) for the last few days and didn't see your messages.

Thanks a lot for carrying this in its final lap to its merge :)

I do have one question, though. I don't really understand your change to glyphs.py, mostly because I don't understand the code there, but what's the intention of removing '_global_alpha' from `_args`? And why doesn't it then go into `_extra_kws`?

Abraço e obrigado !

bryevdv:
@solstag It had ti be removed from `_args` because those get added to the generated function signature as positional args with defaults, which does not work when the name is also a visual handled by `_pop_visuals`. And `_extra_kws` is for things that are *not* properties, i.e. a `palette` kw that is adapted into an actual `color_mapper` in an initializer.

solstag:
That makes sense, thanks for the answer.

Another thing, is there interest in going forward with the bokehjs conversion as well? It ain't far from working, but you did mention there were general issues with the bokehjs plotting api.

bryevdv:
@solstag That would definitely be reasonable/good. Can you start by making the issue for it?

PR #9158, opened 2019-08-12
Add SSL termination and revamped settings system

### LINKED ISSUE #9139: Support SSL Termination
Apparently it's fairly trivial to support SSL termination in Tornado, e.g. this works with a self-signed cert:
```
diff --git a/bokeh/server/server.py b/bokeh/server/server.py
index 04ce230e1..fb8a6e129 100644
--- a/bokeh/server/server.py
+++ b/bokeh/server/server.py
@@ -394,6 +394,10 @@ class Server(BaseServer):
                                        websocket_max_message_size_bytes=opts.websocket_max_message_size,
                                        **kwargs)

+            import ssl
+            context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+            context.load_cert_chain(certfile="/Users/bryan/tmp/ssl/cert.pem")
+            http_server_kwargs['ssl_options'] = context
             http_server = HTTPServer(tornado_app, **http_server_kwargs)

             http_server.start(opts.num_procs)
```

Propose to add server opts and corresponding command line args for `certfile` and (optionally) `keyfile`.

bryevdv:
This one went a bit afield. I finally got fed up with the existing `bokeh.settings` and inconsistencies in settings preference order, that I redid the whole thing so that:

* the order for checking settings is documented and consistent 
* settings can be auto-documented 
* support for settings config files

There's still work to do (for 2.0 to make sure this new settings system is applied consistently and that there are conventions about who has responsibility for what. 

- [x] issues: fixes #9139
- [x] tests added / passed

### DIFF
--- a/bokeh/settings.py
+++ b/bokeh/settings.py
@@ -413,7 +413,7 @@ class Settings(object):
     will build docs that use BokehJS version ``1.4.0rc1`` from CDN.
     """)
 
-    docs_version = PrioritizedSetting("docs_version", "BOKEH_DOCS_CDN", help="""
+    docs_version = PrioritizedSetting("docs_version", "BOKEH_DOCS_CDN", default=None, help="""
     The Bokeh version to stipulate when building the docs.
 
     This setting is necessary to re-deploy existing versions of docs with new
--- a/bokeh/settings.py
+++ b/bokeh/settings.py
@@ -502,7 +502,7 @@ class Settings(object):
     Whether JSON strings should be pretty-printed.
     """)
 
-    py_log_level = PrioritizedSetting("py_log_level", "BOKEH_PY_LOG_LEVEL", default="info", dev_default="debug", convert=convert_logging, help="""
+    py_log_level = PrioritizedSetting("py_log_level", "BOKEH_PY_LOG_LEVEL", default="none", dev_default="debug", convert=convert_logging, help="""
     The log level for Python Bokeh code.
 
     Valid values are, in order of increasing severity:
--- a/bokeh/settings.py
+++ b/bokeh/settings.py
@@ -7,118 +7,68 @@
 ''' Control global configuration options with environment variables.
 A global settings object that other parts of Bokeh can refer to.
 
-``BOKEH_ALLOW_WS_ORIGIN`` --- List of websocket origins allowed to access bokeh.
+immediately supplied values
+    These are values that are passed to the setting:
 
-  Comma separated list of domains that need to access bokeh websocket interface.
-  This can also be provided using the --allow-websocket-origin parameter.
+    .. code-block:: python
 
-  Note: This option overrides the --allow-websocket-origin flag
+        settings.minified(minified_val)
 
-``BOKEH_BROWSER`` --- What browser to use when opening plots.
+    If ``minified_val`` is not None, then it will be returned, as-is.
+    Otherwise, if None is passed, then the setting will continute to look
+    down the search order for a value. This is useful for passing optional
+    function paramters that are None by default. If the parameter is passed
+    to the function, then it will take precedence.
 
-  Valid values are any of the browser names understood by the python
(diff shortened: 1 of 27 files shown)


bryevdv:
What docs look like currently:

Screen Shot 2019-08-11 at 5 10 49 PM

bryevdv:
@mattpap There is a single error failing the examples tests. You should cherry pick the Python pin commits to your PR or a new PR in the mean time.

Do you have any idea what might cause this:
```
[INFO] Example rendered in 0.645s
[JS] Failed to load resource: the server responded with a status of 403 ()
[JS] ReferenceError: Bokeh is not defined
[JS]     at file:///home/travis/build/bokeh/bokeh/examples/models/file/panels.html:21:13
[JS] ReferenceError: Bokeh is not defined
[JS]     at HTMLDocument.fn (file:///home/travis/build/bokeh/bokeh/examples/models/file/panels.html:49:15)
FAILED
```

bokeh/settings.py
```diff
@@ -157,250 +128,516 @@
...
+
+def is_dev():
+    return convert_bool(os.environ.get("BOKEH_DEV", False))
+
+class PrioritizedSetting(object):
+    ''' Return a value for a global setting according to configuration precedence.
+
+    The following methods are searched in order for the setting:
+
+    6. immediately supplied values
```
mattpap: So there is no way to override the default? E.g.:

https://github.com/bokeh/bokeh/blob/a20f671f1e297c45ca8d3acc3c7d897cc3c1ec00/bokeh/resources.py#L81

This breaks `BOKEH_RESOURCES` and the entire development workflow, including `test_examples`.

mattpap: In fact I can't make `test_examples` to use anything by `inline`.

bryevdv: I figured this out last night. The issue is that there is a cannonically accepted order to search or settings which puts *explicitly passed*  values above env vars, but we have never made a distinction between values explicitly supplied, and default parameter values. The right solution is to do that, by making the default values `None` so that e.g. `settings.resources(mode)` will defer to the the other settings sources. [1] I will do that tonight. That makes the settings the single ground truth for these defaults everywhere and ensures that they are always deduced in the same way everywhere. 


[1] We could also use an explicit `Unset` default value, but at least right now we don't have any cases where `None` is a valid value that would be explicitly passed in, so it works to denote "missing optional"

bryevdv:
@mattpap please review, I'd like to merge this tomorrow. Some comments:

* default mode for `Resources()` was "inline" which is inconsistent with all default modes everywhere else, which is "cdn". I changed it to match (needed to... I am willing to consider this was a bug being inconsistent, in any case I am will to take the risk that no-one relies literally on `Resources()` exactly)

* fun fact: `BOKEH_RESOURCES=inline` worked in some cases by altering how the `CDN` object was initialized! that's pretty ugly. Now new resources are created with the right mode on demand, instead of relying on a pre-fab resource object getting unintentionally modified.

mattpap:
Everything seems to work correctly now. One unrelated, but extremely annoying thing, that possibly exacerbated my yesterday's experience, is that (I presume) pytest does some sort of caching, that prevents new changes from being immediately picked up by tests, e.g. running examples' tests just after checkout resulted in wrong resources being used (`inline` instead of `relative`). I've been seeing this a lot with working on extensions, but now I see this is a more wide spread problem. `git clean -dfx examples/` helps, but slowdowns development.

mattpap:
> fun fact: BOKEH_RESOURCES=inline worked in some cases by altering how the CDN object was initialized! that's pretty ugly. Now new resources are created with the right mode on demand, instead of relying on a pre-fab resource object getting unintentionally modified.

In fact for 2.0, as a part of extensions work, I will split resources into resources and bundles, where resources will be immutable specifications and bundle will contain actual artifacts.

PR #9171, opened 2019-08-19
Automatically insert class init code

### LINKED ISSUE #9170: Automatically insert class initialization code
Currently models are typically defined as follows:
```ts
export class SomeModel extends Model {
  static initClass(): void {
    this.define(...)
  }
}
SomeModel.initClass()
```
It's easy to forget `initClass()` call and not initialize a class, or remove `static initClass()`, but leave the call, effectively running parent class' init code twice. Initialization call can be inserted automatically by AST transformers, which we already use to make other modifications (e.g. insert `__name__` member). Additionally `initClass` will be renamed to `init_SomeModel`, to avoid confusion between inherited init methods.

mattpap:
This change is backwards compatible. Nothing changes from users' perspective and one can still use the old syntax.

fixes #9170

### DIFF
--- a/bokehjs/src/compiler/compiler.ts
+++ b/bokehjs/src/compiler/compiler.ts
@@ -107,6 +107,9 @@ export function default_transformers(options: ts.CompilerOptions, css_dir?: Path
   const insert_class_name = transforms.insert_class_name()
   transformers.before.push(insert_class_name)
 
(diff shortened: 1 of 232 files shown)


mattpap:
I'm going ahead and merge this as it conflicts with my other work.

Issue #9187, opened 2019-08-23
[BUG] Pinch Zoom and Panning not working correctly


DarkCyTic:
I am using Bokeh standalone **1.3.4** on the back end (Django) and bokehjs on the front end. While scrolling to zoom and using the mouse to pan works well on desktop, the experience on mobile devices is ambiguous. In Chrome (**76.0.3809.111**) on mobile devices, nor panning, nor pinching to zoom work. Panning works great in Safari and Firefox on mobile devices, but pinching to zoom doesn't work there as well. 

This is the back end code:
```python
p = figure()

color_palette = Spectral11[0:num_lines]

source = ColumnDataSource(power_data)

p.add_layout(Legend(items=legend_items))
p.legend.click_policy = "hide"

p.add_tools(CrosshairTool(
        dimensions='height',
        line_color='red'
))

item_text = json.dumps(json_item(p, "myplot"))

return Response(item_text)
```

This is the front end code:
```javascript
declare var Bokeh: any

// some lines later, inside a request callback
Bokeh.embed.embed_item(JSON.parse(response))
```

I have found a website where the pinch to zoom and pan work well and they are using an older version of Bokeh(**1.0.4**), so I am not sure if it's something with my code or it no longer works.

Are these features still supported on mobile devices?

mattpap:
They are supposed to work, but we have limited resource to test on mobile, so regressions are inevitable.

mattpap:
The problem with pinch motion is that it often ends up as rotation, and rotation is half-implemented in bokeh, resulting in an exception (which isn't easily accessible on mobile). Panning issue in Chrome Mobile is a bit more severe. I will have to investigate this.

bryevdv:
@mattpap should this be moved out of `1.4` milestone? It would need to be fixed before the end of next week.

mattpap:
I will have this fixed during PyCon.DE sprints this weekend.

bryevdv:
@mattpap should this be moved out?

do1tgood:
[...]

PR #9188, opened 2019-08-23
Finalize support for rotate events

mattpap:
addresses #9187


bryevdv:
What does "addresses" mean, specifically?

mattpap:
> What does "addresses" mean, specifically?

This PR doesn't fix panning issue in Chrome Mobile.

PR #9190, opened 2019-08-25
Auth features for server

### LINKED ISSUE #9140: Support tornado get_current_user
Allow users to specify a function to use for Tornado's `get_current_user`:

https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.current_user

And use the `authenticated` decorator when applicable on endpoint handlers

bryevdv:
- [x] issues: fixes #9140
- [x] tests added / passed

This is WIP, would like to get some early feed back on how the usage feels.

You can now supply an auth modue, e.g. 
```

import tornado
from tornado.web import RequestHandler

# could also define async_get_user
def get_user(request_handler):
    return request_handler.get_secure_cookie("user")

# could also define get_login_url func (but give up LoginHandler)
login_url = "/login"

# optional login page at login_url
class LoginHandler(RequestHandler):

    def get(self):
        try:
            errormessage = self.get_argument("error")
        except:
            errormessage = ""
        self.render("login.html", errormessage=errormessage)

    def check_permission(self, password, username):
        if username == "bokeh" and password == "bokeh":
            return True
        return False

    def post(self):
        username = self.get_argument("username", "")
        password = self.get_argument("password", "")
        auth = self.check_permission(password, username)
        if auth:
            self.set_current_user(username)
            self.redirect(self.get_argument("next", u"/"))
        else:
            error_msg = u"?error=" + tornado.escape.url_escape("Login incorrect")
            self.redirect(login_url + error_msg)

    def set_current_user(self, user):
        if user:
            self.set_secure_cookie("user", tornado.escape.json_encode(user))
        else:
            self.clear_cookie("user")
```
and a `login.html` template, and you can only get to the app if you pass login. 

Some notes: 

* always have to supply `--auth-module` separately, i.e. can't make `auth.py` in a directoy app work because there could be multiple directory apps and that makes no sense (also other reasons)
* `LoginHandler` is optional, "serious" deploys would redirect to some external login probably
* Since any cookie get/set happens in the auth module, users can control max age and expiry there themselves when they call those funcs
* Do we need to support a `LogoutHandler`? Presumably apps could clear the cookie themselved from a Button push, etc. Maybe later?

Initial thoughts, @bokeh/dev ?

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -299,7 +299,7 @@ vary based on the request, or cookies, etc. It is not possible to specify a
 ``LoginHandler`` when ``get_url_function`` is defined.
 
 Analogous to the login options, optional ``logout_url`` and ``LogoutHandler``
-values may be define and endopoint for logging users out.
+values may be define an endpoint for logging users out.
 
 If no auth module is provided, then a default user will be assumed, and no
 authentication will be required to access Bokeh server endpoints.
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -294,9 +294,12 @@ When a relative ``login_url`` is given, an optional ``LoginHandler`` class may
 also be provided, and it will be installed as a route on the Bokeh server
 automatically.
 
-The ``get_url_function`` function is useful in cases where the login URL must
+The ``get_login_url`` function is useful in cases where the login URL must
 vary based on the request, or cookies, etc. It is not possible to specify a
 ``LoginHandler`` when ``get_url_function`` is defined.
+
+Analogous to the login options, option ``logout_url``,  ``get_logout_url``,
+and ``LogoutHandler`` values may be define and endopoint for logging users out.
 
 If no auth module is provided, then a default user will be assumed, and no
 authentication will be required to access Bokeh server endpoints.
@@ -810,11 +813,13 @@ class Serve(Subcommand):
                 m = imp.load_source('module.name', '/path/to/file.py')
 
         result = {}
-        for name in ('get_user', 'get_user_async', 'login_url', 'get_login_url'):
+        for name in ('get_user', 'get_user_async', 'login_url', 'get_login_url', 'logout_url', 'get_logout_url'):
             if hasattr(m, name):
                 result[name] = getattr(m, name)
         if hasattr(m, 'LoginHandler'):
             result['login_handler'] = getattr(m, 'LoginHandler')
+        if hasattr(m, 'LogoutHandler'):
+            result['logout_handler'] = getattr(m, 'LogoutHandler')
         return result
 
 #-----------------------------------------------------------------------------
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -307,6 +307,15 @@ authentication will be required to access Bokeh server endpoints.
 .. warning:
     The contents of the auth module will be executed!
 
+Bokeh can also enable the use of Tornado's XFRF cookie protection. To turn this
+feature on, use the ``--enable-xsrf-cookies`` option, or set the environment
+variable ``BOKEH_XSRF_COOKIES=yes``. If this setting is enabled, any PUT, POST,
+or DELETE operations on custom or login handlers must be instrumented properly
+in order to function. Typically, this means adding the ``xsrf_form_html()``
+module to HTML form submissions templates. For full details, see:
+
+    https://www.tornadoweb.org/en/stable/guide/security.html#cross-site-request-forgery-protection
+
 Session Expiration Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -599,6 +608,14 @@ class Serve(Subcommand):
             help    = 'Absolute path to a Python modules that implements auth hooks',
         )),
 
+        ('--enable-xsrf-cookies', dict(
+            action  = 'store_true',
+            default = False,
+            help    = 'Whether to enable Tornado support for XSRF cookies. All '
+                      'PUT, POST, or DELETE handlers must be properly instrumented '
+                      'when this setting is enabled.'
+        )),
+
         ('--cookie-secret', dict(
             metavar = 'COOKIE_SECRET',
             action  = 'store',
@@ -733,6 +750,7 @@ class Serve(Subcommand):
         else:
             server_kwargs['auth_provider'] = NullAuth()
 
+        server_kwargs['xsrf_cookies'] = settings.xsrf_cookies(getattr(args, 'enable_xsrf_cookies', False))
         server_kwargs['cookie_secret'] = settings.cookie_secret(getattr(args, 'cookie_secret', None))
         server_kwargs['use_index'] = not args.disable_index
         server_kwargs['redirect_root'] = not args.disable_index_redirect
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -450,10 +450,10 @@ base_serve_args = (
     )),
 
     ('--use-config', dict(
-        metavar='CONFIG',
-        type=str,
-        help="Use a YAML config file for settings",
-        default=None,
+        metavar = 'CONFIG',
+        type    = str,
+        help    = "Use a YAML config file for settings",
+        default = None,
     )),
 
 )
@@ -478,18 +478,18 @@ class Serve(Subcommand):
 
     args = base_serve_args + (
         ('files', dict(
-            metavar='DIRECTORY-OR-SCRIPT',
-            nargs='*',
-            help="The app directories or scripts to serve (serve empty document if not specified)",
-            default=None,
+            metavar = 'DIRECTORY-OR-SCRIPT',
+            nargs   = '*',
+            help    = "The app directories or scripts to serve (serve empty document if not specified)",
+            default = None,
         )),
 
         ('--args', dict(
-            metavar='COMMAND-LINE-ARGS',
-            nargs=argparse.REMAINDER,
-            help="Command line arguments remaining to passed on to the application handler. "
-                 "NOTE: if this argument precedes DIRECTORY-OR-SCRIPT then some other argument, e.g. "
-                 "--show, must be placed before the directory or script. ",
+            metavar = 'COMMAND-LINE-ARGS',
+            nargs   = argparse.REMAINDER,
+            help    = "Command line arguments remaining to passed on to the application handler. "
+                      "NOTE: if this argument precedes DIRECTORY-OR-SCRIPT then some other argument, e.g. "
+                      "--show, must be placed before the directory or script. ",
         )),
 
         ('--dev', dict(
@@ -501,7 +501,7 @@ class Serve(Subcommand):
             help    = "Enable live reloading during app development. "
                       "By default it watches all *.py *.html *.css *.yaml files "
                       "in the app directory tree. Additional files can be passed "
-                      "as arguments."
+                      "as arguments. "
                       "NOTE: if this argument precedes DIRECTORY-OR-SCRIPT then some other argument, e.g "
                       "--show, must be placed before the directory or script. "
                       "NOTE: This setting only works with a single app. "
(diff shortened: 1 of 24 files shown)


bryevdv:
Here is a simple example:

ScreenFlow

birdsarah:
Sorry I'm being slow, can you spell out the rest of the incantation that would mean this is all hooked up?

bryevdv:
Sorry, run like this:
```
BOKEH_COOKIE_SECRET='muy super secret' bokeh serve --auth-module=/Users/bryan/tmp/auth.py --show sliders.py
```

I used this template:
```
<div id="main-container">
    <div id="main">
        <div id="login-form">
            <form action="/login" method="post" id="login_form">
                <fieldset>
                    <label for="username">Username</label>
                    <input autocapitalize="off" autocorrect="off" class="text-input" id="username" name="username" tabindex="1" type="text" value="">
                </fieldset>

                <fieldset>
                    <label for="password">Password</label>
                    <input class="text-input" id="password" name="password" tabindex="2" type="password" value="">
                </fieldset>
                <fieldset>
                    <span class="errormessage">{{errormessage}}</span>
                </fieldset>

                <div id="form_btn">
                    <input id="signin-btn" class="btn btn-blue" type="submit" value="Sign In" tabindex="3">
                </div>
            </form>
        </div>
    </div>
</div>
```

bokeh/server/tornado.py
```diff
@@ -249,6 +280,34 @@ def __init__(self,
...
+        if (get_user or get_user_async) and not (login_url or get_login_url):
+            raise ValueError("If a get_user function is provided, login URL must also be provided")
+        if login_url and get_login_url:
+            raise ValueError("At most one of login_url or get_login_url should be supplied")
+        if login_handler and get_login_url:
+            raise ValueError("LoginHandler cannot be used with get_login_url()")
+        if login_handler and not issubclass(login_handler, RequestHandler):
+            raise ValueError("LoginHandler must be a Tornado RequestHandler")
+        # This just catches some common cases up front, let tornado barf on any others
+        if login_handler and login_url.startswith("http") and login_url.startswith("//"):
```
p-himik: Always `False`.

bryevdv: Thank you!

p-himik: Also technically a legitimate relative URL could start with `http` but I guess it's not a big deal.

bryevdv: No it's a fair observation, I will tighten it up. I'm not worried about false negatives here since Tornado will catch the issue if we don't, but should try to avoid any false positives.

birdsarah:
This looms great. Just some clear docs like the other stuff you've done for server. The only bit that I'm left wondering is which class is bokeh looking for - anything that overrides the requesthandler? Or is it the loginhandler class specifically which needs to override the requesthandler.

The enterprise I imagine is the bokeh server being behind the corporate ldap/oauth where you go to the bokeh page, get redirected to oauth, and then get let in if valid, in that case no template would be required its just routing and checking.

The other scenario I imagine is where your bokeh app is embedded in site that has login. In that case the bokeh plot would be an embedded frame that would gate until author handling / checking was complete.

In both of these scenarios the login template itself wouldn't be used. But I think its great to include your scenario too wherethe only server is bokeh / tornado and it serves up the login.

birdsarah:
Maybe this leads to a more general 'hookig into tornado', I'm not sure.

bryevdv:
@p-himik I believe I have addressed your change in 851f00c

@birdsarah @hhuuggoo I think something needs to be done re: logging out. I am just not sure what exactly the best workflow is. I think it would be simple to add a `logout_url` or `LogoutHandler` analagous to the login one. That handler could clear cookies or whatever, etc. Then a  `js_on_click` or session timeout callback could redirect to that URL. Does that sound reasonable?

birdsarah:
> I think something needs to be done re: logging out. I am just not sure what exactly the best workflow is. I think it would be simple to add a logout_url and optional LogoutHandler analagous to the login case. That handler could clear cookies or whatever, etc. 

Good point. Yes. I don't remember ever worrying about logging people out specifically. Typically you just set the cookie expiry to be what you want. I don't know what would happen to bokeh once the cookie expired. Regardless the analagous logout hook would be great, and presumably it's easy.

bryevdv:
Ok I added the logout url/handler support:

ScreenFlow

Imagine if we had added this three years ago 🤦‍♂ 

I've been in a bit of a slump the last two weeks but I am going to do my best to get these things done this weekend:

* Factor all the auth things from the auth module into an `AuthProvider` class. There are lots of bits and they are all currently smeared all over the tornado handler
* Add tests (first task will help with this)
* User guide docs

Hoping to get some final thoughts/review and merge early next week.

bryevdv:
OK one more small feature, `--enable-xsrf-cookies` flag to turn to Tornado's XSRF protection.

bryevdv:
@bokeh/dev I am only going to add test from here out, marking this `ready` for review

bokeh/command/subcommands/serve.py
```diff
@@ -388,10 +459,10 @@
...
 
     ('--use-config', dict(
-        metavar='CONFIG',
-        type=str,
-        help="Use a YAML config file for settings",
-        default=None,
+        metavar = 'CONFIG',
+        type    = str,
+        help    = "Use a YAML config file for settings",
+        default = None,
```
p-himik: Judging by this PR, the style for keyword arguments' values alignment is different throughout Bokeh codebase. And I've never seen such alignment in any Python project before, although that might be because of my limited exposure to other Python projects. Is there any prettifier config or documentation that makes it clear that this style is preferred?

bokeh/server/auth_provider.py
```diff
@@ -0,0 +1,318 @@
...
+    module, or None otherwise.
+
+    The ``logout_handler`` property will return a ``LogoutHandler`` class from
+    the module, or None otherwise.
+
+    '''
+
+    def __init__(self, module_path):
+        if not isfile(module_path):
+            raise ValueError()
```
p-himik: Maybe a message would make sense.

bokeh/server/auth_provider.py
```diff
@@ -0,0 +1,318 @@
...
+    ''' Return True if a URL is not one of the common absolute URL formats.
+
+    Arguments:
+        url (str): a URL string
+
+    Returns
+        bool
+
+    '''
+    return not (url.startswith("http:") or url.startswith("https:") or url.startswith("//"))
```
p-himik: Just realized - could be simplified to `return not url.startswith(("http://", "https://", "//"))`. Even on Python 2.7.

bokeh/server/auth_provider.py
bryevdv: I meant to go back and addd one but forgot, thank you

bokeh/command/subcommands/serve.py
bryevdv: No presently, ideally we can just switch to black maybe at some point and forget about debates. I would never write kwargs in an actual function call this way, but although these are ultimately used that way, here they are more like data literals and that is the case where this is my personal pref (see also: lining up dict literal values)

bokeh/server/auth_provider.py
bryevdv: https://github.com/bokeh/bokeh/pull/9190/commits/5a76573f27f177d24981570445bee59451c8f43b

bokeh/server/auth_provider.py
bryevdv: https://github.com/bokeh/bokeh/pull/9190/commits/d06f2e2def0a503e11a51f6e372d8856ca5410b9

bokeh/command/subcommands/serve.py
bryevdv: In this case I mostly just wanted to make things consistent even within the same file

bryevdv:
#bokeh/dev Barring serious issues raised, I intend to merge this tomorrow.

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1106,6 +1111,91 @@ here:
...
+.. _userguide_server_deployment_auth:
+
+Authentication
+~~~~~~~~~~~~~~
+
+The Bokeh server itself does not have any facilities for authentication or
+authorization. However, the Bokeh server can be configured with and "Auth
+Provider" that hooks in to Tornado's underlying capabilities. For background
+information, see the Tornado docs for `Authentication and security`_. The rest
+of this section assume some familiarity with that material.
```
philippjfr:
```suggestion
of this section assumes some familiarity with that material.
```

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1106,6 +1111,91 @@ here:
...
+
+The module must contain *one* of the following two functions that will return
+the current user (or None):
+
+.. code-block:: python
+
+    def get_user(request_handler):
+        pass
+
+    async def get_user_aync(request_handler):
```
philippjfr:
```suggestion
    async def get_user_async(request_handler):
```

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1106,6 +1111,91 @@ here:
...
+When a relative ``login_url`` is given, an optional ``LoginHandler`` class may
+also be provided, and it will be installed as a route on the Bokeh server
+automatically.
+
+The ``get_login_url`` function is useful in cases where the login URL must
+vary based on the request, or cookies, etc. It is not possible to specify a
+``LoginHandler`` when ``get_url_function`` is defined.
+
+Analogous to the login options, optional ``logout_url`` and ``LogoutHandler``
+values may be define and endopoint for logging users out.
```
philippjfr:
```suggestion
values may be define an endpoint for logging users out.
```

bokeh/settings.py
```diff
@@ -568,6 +583,17 @@ def dev(self):
     See the :class:`~bokeh.resources.Resources` class reference for full details.
     """)
 
+    xsrf_cookies = PrioritizedSetting("xsrf_cookies", "BOKEH_XSRF_COOKIES", default=False, convert=convert_bool, help="""
+    Whether to enable Tornado XSRF cookie protetion on the Bokeh server. This
```
philippjfr:
```suggestion
    Whether to enable Tornado XSRF cookie protection on the Bokeh server. This
```

bokeh/command/subcommands/serve.py
```diff
@@ -246,6 +246,76 @@
...
+
+The module must contain *one* of the following two functions that will return
+the current user (or None):
+
+.. code-block:: python
+
+    def get_user(request_handler):
+        pass
+
+    async def get_user_aync(request_handler):
```
philippjfr:
```suggestion
    async def get_user_async(request_handler):
```

bokeh/command/subcommands/serve.py
```diff
@@ -246,6 +246,76 @@
...
+When a relative ``login_url`` is given, an optional ``LoginHandler`` class may
+also be provided, and it will be installed as a route on the Bokeh server
+automatically.
+
+The ``get_login_url`` function is useful in cases where the login URL must
+vary based on the request, or cookies, etc. It is not possible to specify a
+``LoginHandler`` when ``get_url_function`` is defined.
+
+Analogous to the login options, optional ``logout_url`` and ``LogoutHandler``
+values may be define and endopoint for logging users out.
```
philippjfr:
```suggestion
values may be define an endpoint for logging users out.
```

bokeh/server/auth_provider.py
```diff
@@ -0,0 +1,318 @@
...
+        if self.logout_handler:
+            endpoints.append((self.logout_url, self.logout_handler))
+        return endpoints
+
+    @property
+    def get_login_url(self):
+        ''' A function that computes a URL to redirect unathenticated users
+        to for login.
+
+        This proprty may return None, if a ``login_url`` is supplied
```
philippjfr:
```suggestion
        This property may return None, if a ``login_url`` is supplied
```

bokeh/server/auth_provider.py
```diff
@@ -0,0 +1,318 @@
...
+
+        If a function is returned, it should accept a ``RequestHandler``
+        and return the current authenticated user.
+
+        '''
+        pass
+
+    @property
+    def get_user_async(self):
+        ''' An aysnc function to get the current authenticated user.
```
philippjfr:
```suggestion
        ''' An async function to get the current authenticated user.
```

bokeh/server/auth_provider.py
```diff
@@ -0,0 +1,318 @@
...
+
+        If a class is returned, it must be a subclass of RequestHandler,
+        which will used for the endpoint specified by ``logout_url``
+
+        '''
+        pass
+
+    @property
+    def login_url(self):
+        ''' A URL to redirect unathenticated users to for login.
```
philippjfr:
```suggestion
        ''' A URL to redirect unauthenticated users to for login.
```

bokeh/command/subcommands/serve.py
```diff
@@ -246,6 +246,76 @@
 should be a cryptographically random string with at least 256 bits (32 bytes)
 of entropy. The ``bokeh secret`` command can generate new secret keys.
 
+Authentication Options
+~~~~~~~~~~~~~~~~~~~~~~
+
+The Bokeh server can be configured to only allow connections in case there is
+a properly authenticed user. This is accomplished by providing the path to
```
jbednar:
```suggestion
a properly authenticated user. This is accomplished by providing the path to
```

bokeh/command/subcommands/serve.py
```diff
@@ -246,6 +246,76 @@
...
+``LoginHandler`` when ``get_url_function`` is defined.
+
+Analogous to the login options, optional ``logout_url`` and ``LogoutHandler``
+values may be define and endopoint for logging users out.
+
+If no auth module is provided, then a default user will be assumed, and no
+authentication will be required to access Bokeh server endpoints.
+
+.. warning::
+    The contents of the auth module will be executed!
```
jbednar: Can you explain the implications here, i.e. why that's a warning? Maybe:

```suggestion
    The contents of the auth module will be executed, so you should ensure that this module is trustworthy and only includes code you are comfortable with.
```

bokeh/command/subcommands/serve.py
```diff
@@ -450,119 +521,141 @@ class Serve(Subcommand):
...
             default = None,
             choices = SESSION_ID_MODES,
             help    = "One of: %s" % nice_join(SESSION_ID_MODES),
         )),
 
+        ('--auth-module', dict(
+            metavar = 'AUTH_MODULE',
+            action  = 'store',
+            default = None,
+            help    = 'Absolute path to a Python modules that implements auth hooks',
```
jbednar:
```suggestion
            help    = 'Absolute path to a Python module that implements auth hooks',
```

bokeh/command/subcommands/tests/test_serve.py
```diff
@@ -144,119 +146,141 @@ def test_args():
...
             default = None,
             choices = scserve.SESSION_ID_MODES,
             help    = "One of: %s" % nice_join(scserve.SESSION_ID_MODES),
         )),
 
+        ('--auth-module', dict(
+            metavar = 'AUTH_MODULE',
+            action  = 'store',
+            default = None,
+            help    = 'Absolute path to a Python modules that implements auth hooks',
```
jbednar:
```suggestion
            help    = 'Absolute path to a Python module that implements auth hooks',
```

bokeh/settings.py
```diff
@@ -413,6 +413,15 @@ def dev(self):
     A comma-separated list of allowed websocket origins for Bokeh server applications.
     """)
 
+    auth_module = PrioritizedSetting("auth_module", "BOKEH_AUTH_MODULE", default=None, help="""
+    A path to a Python modules that implements user authentication functions for
+    the Bokeh server.
+
+    .. warning::
+        The contents of this module will be executed!
```
jbednar: (Same as above comment)

sphinx/source/docs/user_guide/server.rst
```diff
@@ -8,6 +8,10 @@ Running a Bokeh Server
 Purpose
 -------
 
+The purpose of the Bokeh server is to make it easy for python users to create
```
jbednar:
```suggestion
The purpose of the Bokeh server is to make it easy for Python users to create
```

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1106,6 +1111,91 @@ here:
 
     }
 
+.. _userguide_server_deployment_auth:
+
+Authentication
+~~~~~~~~~~~~~~
+
+The Bokeh server itself does not have any facilities for authentication or
+authorization. However, the Bokeh server can be configured with and "Auth
```
jbednar:
```suggestion
authorization. However, the Bokeh server can be configured with an "Auth
```

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1106,6 +1111,91 @@ here:
...
+Provider" that hooks in to Tornado's underlying capabilities. For background
+information, see the Tornado docs for `Authentication and security`_. The rest
+of this section assume some familiarity with that material.
+
+Auth Module
+'''''''''''
+
+The Bokeh server can be configured to only allow connections in case there is
+a properly authenticed user. This is accomplished by providing the path to
```
jbednar:
```suggestion
a properly authenticated user. This is accomplished by providing the path to
```

sphinx/source/docs/user_guide/server.rst
```diff
@@ -1232,6 +1322,22 @@ Django or whatever tool is in use).
...
+or set the environment variable ``BOKEH_XSRF_COOKIES=yes``. If this setting is
+enabled, any PUT, POST, or DELETE operations on custom or login handlers must be
+instrumented properly in order to function. Typically, this means adding the
+code:
+
+.. code-block:: html
+
+    {% module xsrf_form_html() %}
+
+to all HTML form submissions templates. For full details, see the Tornado
```
jbednar:
```suggestion
to all HTML form submission templates. For full details, see the Tornado
```

bryevdv:
Ok I am going to go ahead and merge this. Release is still at least a few weeks out so there is certainly time for any follow on work.

PR #9198, opened 2019-08-28
Add support for 'bokeh build' command

### LINKED ISSUE #9196: Add support for building bokehjs extensions

### LINKED ISSUE #9240: [BUG] Building custom extension breaks in notebook
When trying to include a custom bokeh model in a notebook the compilation code currently falls over on bokeh 1.3.5devX and master. This is almost certainly related to the recent changes to the build infrastructure.

```python
import numpy as np

from bokeh.util.compiler import TypeScript

TS_CODE = """
import * as p from "core/properties"
import {Label, LabelView} from "models/annotations/label"
declare const katex: any

export class LatexLabelView extends LabelView {
  model: LatexLabel

  render(): void {
    //--- Start of copied section from ``Label.render`` implementation

    // Here because AngleSpec does units tranform and label doesn't support specs
    let angle: number
    switch (this.model.angle_units) {
      case "rad": {
        angle = -this.model.angle
        break
      }
      case "deg": {
        angle = (-this.model.angle * Math.PI) / 180.0
        break
      }
      default:
        throw new Error("unreachable code")
    }

    const panel = this.panel != null ? this.panel : this.plot_view.frame

    const xscale = this.plot_view.frame.xscales[this.model.x_range_name]
    const yscale = this.plot_view.frame.yscales[this.model.y_range_name]

    let sx = this.model.x_units == "data" ? xscale.compute(this.model.x) : panel.xview.compute(this.model.x)
    let sy = this.model.y_units == "data" ? yscale.compute(this.model.y) : panel.yview.compute(this.model.y)

    sx += this.model.x_offset
    sy -= this.model.y_offset

    //--- End of copied section from ``Label.render`` implementation
    // Must render as superpositioned div (not on canvas) so that KaTex
    // css can properly style the text
    this._css_text(this.plot_view.canvas_view.ctx, "", sx, sy, angle)

    // ``katex`` is loaded into the global window at runtime
    // katex.renderToString returns a html ``span`` element
    katex.render(this.model.text, this.el, {displayMode: true})
  }
}

export namespace LatexLabel {
  export type Attrs = p.AttrsOf<Props>

  export type Props = Label.Props
}

export interface LatexLabel extends LatexLabel.Attrs {}

export class LatexLabel extends Label {
  properties: LatexLabel.Props

  constructor(attrs?: Partial<LatexLabel.Attrs>) {
    super(attrs)
  }

  static initClass() {
    this.prototype.default_view = LatexLabelView
  }
}
LatexLabel.initClass()
"""


class LatexLabel(Label):
    """A subclass of the Bokeh built-in `Label` that supports rendering
    LaTex using the KaTex typesetting library.

    Only the render method of LabelView is overloaded to perform the
    text -> latex (via katex) conversion. Note: ``render_mode="canvas``
    isn't supported and certain DOM manipulation happens in the Label
    superclass implementation that requires explicitly setting
    `render_mode='css'`).
    """
    __javascript__ = ["https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/katex.min.js"]
    __css__ = ["https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/katex.min.css"]
    __implementation__ = TypeScript(TS_CODE)


x = np.arange(0.0, 1.0 + 0.01, 0.01)
y = np.cos(2 * 2 * np.pi * x) + 2

p = figure(title="LaTex Demonstration", plot_width=500, plot_height=500)
p.line(x, y)

# Note: must set ``render_mode="css"``
latex = LatexLabel(text="f = \sum_{n=1}^\infty\\frac{-e^{i\pi}}{2^n}!",
                   x=40, y=420, x_units='screen', y_units='screen',
                   render_mode='css', text_font_size='16pt',
                   background_fill_alpha=0)

p.add_layout(latex)

output_notebook()

show(p)
```

```
Compilation failed:

error TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["./**/*.ts"]' and 'exclude' paths were '["./dist"]'.
```

mattpap:
This adds `bokeh build <base_path>` command. A test case is available in https://github.com/pyviz/panel/pull/626.

fixes #9196 
fixes #9240


xavArtley:
Just for information to build bokehjs after checkout in this branch I had to delete previous build directory else I had this error:
```
 bokehjs@1.3.5-dev2 build D:\PythonWorkspace\modules\bokeh\bokehjs
> node make build

[10:12:11] Starting 'styles:compile'...
[10:12:11] Finished 'styles:compile' after 187 ms
[10:12:11] Starting 'scripts:compile'...
[10:12:19] Finished 'scripts:compile' after 8.03 s
[10:12:19] Starting 'scripts:bundle-es6'...
[10:12:19] failed: Error: both D:\PythonWorkspace\modules\bokeh\bokehjs\build\js\lib\embed\notebook.js and D:\PythonWorkspace\modules\bokeh\bokehjs\build\js\lib\embed\notebook\index.js exist
[10:12:19]     at Linker.resolve_relative (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:449:17)
[10:12:19]     at Linker.resolve_file (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:498:19)
[10:12:19]     at filtered.map (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:559:67)
[10:12:19]     at Array.map (<anonymous>)
[10:12:19]     at Linker.new_module (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:559:43)
[10:12:19]     at Linker.resolve (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:597:27)
[10:12:19]     at Linker.link (D:\PythonWorkspace\modules\bokeh\bokehjs\src\compiler\linker.ts:233:28)
[10:12:19]     at Object.<anonymous> (D:\PythonWorkspace\modules\bokeh\bokehjs\make\tasks\scripts.ts:37:26)
[10:12:19]     at Generator.next (<anonymous>)
[10:12:19]     at D:\PythonWorkspace\modules\bokeh\bokehjs\make\tasks\scripts.ts:7:71
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! bokehjs@1.3.5-dev2 build: `node make build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the bokehjs@1.3.5-dev2 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

xavArtley:
I have an other question but it's more related to my inexperience with javascript/typescript
I'd like to get rid of the cdn for vtk panel and use directly the npm package
So I installed the package:
`npm install vtk.js' under panel directory
Then I created a declaration file with 
vtk.js.d.ts
```
declare module 'vtk.js/Sources/Rendering/Core/AxesActor'
```
and tried to import with:
`import {vtkAxesActor} from "vtk.js/Sources/Rendering/Core/AxesActor"
However when I compile I get the error:
```
Linking modules
Error: D:\PythonWorkspace\modules\panel\panel\node_modules\vtk.js\Sources\Rendering\Core\AxesActor\index.js:1:4: Import statement may only appear at top level
    at minify (d:\pythonworkspace\modules\bokeh\bokeh\server\static\js\compiler.js:139139:19)
```
I suppose I need to tweek some json files for compiling vtkjs sources first (https://kitware.github.io/vtk-js/docs/intro_vtk_as_es6_dependency.html). But I have no Idea how to do it in the context of typescript

mattpap:
@xavArtley, the simplest way to clear that error is to remove `bokehjs/build` directory and rebuild. This happens when a file is split into a submodule (or the the other way). Recently this happened in PR #9138. This isn't an often situation, so this issue has been neglected for a long while.

mattpap:
> `Import statement may only appear at top level`

This look like a bug, but I will have to inspect this closer. I'm currently traveling, so I will get back to this work and the panel PR in a few days.

bokeh/command/bootstrap.py
```diff
@@ -110,9 +112,17 @@ def main(argv):
...
-        die("ERROR: " + str(e))
+        if settings.dev:
+            raise
+        else:
+            die("ERROR: " + str(e))
+
+    if ret is False:
+        sys.exit(1)
+    elif ret is not True and isinstance(ret, int) and ret != 0:
+        sys.exit(ret)
```
bryevdv: Let's just make it conventional that all subcommand `invoke` functions return integer status so this awkwardness can be avoided.

bokeh/embed/standalone.py
```diff
@@ -88,18 +87,12 @@ def autoload_static(model, resources, script_path):
...
+    (_, elementid) = list(render_item.roots.to_json().items())[0]
 
-    js = wrap_in_onload(AUTOLOAD_JS.render(
-        js_urls = resources.js_files,
-        css_urls = resources.css_files,
-        js_raw = resources.js_raw + [bundle, script],
-        css_raw = resources.css_raw_str,
-        elementid = elementid,
-    ))
+    js = wrap_in_onload(AUTOLOAD_JS.render(bundle=bundle, elementid=elementid))
```
bryevdv: So, I hate the name "autoload_whatever". Instead of complicating the existing template to handle this new format, can we just make a new thing with a new name and deprecate the autoload stuff?

bokeh/command/bootstrap.py
mattpap: A bit of the complexity comes from this non-sense:
```
In [1]: isinstance(True, int)                                                                                                                
Out[1]: True
```
Even if we don't proceed with this, we still need differentiate `None` from integers for backwards compatibility. I also didn't want to modify every `invoke()` implementation.

bokeh/embed/standalone.py
mattpap: This is what I wanted to avoid and make minimal, backwards compatible changes, and leave deeper changes for 2.0 or later. At this point I intent to redesign embed and resources from scratch at some point.

xavArtley:
> > `Import statement may only appear at top level`
> 
> This look like a bug, but I will have to inspect this closer. I'm currently traveling, so I will get back to this work and the panel PR in a few days.

Don't you think it's because I need to specify some rules for module export of vtk.js?
Like specified in the webpack.config.js?
```javascript
var path = require('path');
var webpack = require('webpack');
var vtkRules = require('vtk.js/Utilities/config/dependency.js').webpack.core.rules;

// Optional if you want to load *.css and *.module.css files
// var cssRules = require('vtk.js/Utilities/config/dependency.js').webpack.css.rules;

var entry = path.join(__dirname, './src/index.js');
const sourcePath = path.join(__dirname, './src');
const outputPath = path.join(__dirname, './dist');

module.exports = {
  entry,
  output: {
    path: outputPath,
    filename: 'MyWebApp.js',
  },
  module: {
    rules: [
        { test: /\.html$/, loader: 'html-loader' },
    ].concat(vtkRules),
  },
  resolve: {
    modules: [
      path.resolve(__dirname, 'node_modules'),
      sourcePath,
    ],
  },
};



```

philippjfr:
I know this is probably still in progress but for backward compatibility can we re-add `Resources.css_raw_str`.

mattpap:
@philippjfr, I will restore it, in fact I should have just left it in the first place. This PR is essentially ready and I would like it merged as soon as possible. I will address @xavArtley's issues separately, because those are linker-related.

philippjfr:
This branch works and fixes #9240 for me. Can we cut a dev release once this is merged?

bryevdv:
@philippjfr yes, I've been meaning to, we actually need to test that making a second dev build overwrites the deployed dev docs correctly anyway

bokeh/ext.py
```diff
@@ -0,0 +1,66 @@
...
+# Globals and constants
+#-----------------------------------------------------------------------------
+
+__all__ = ["build"]
+
+#-----------------------------------------------------------------------------
+# General API
+#-----------------------------------------------------------------------------
+
+def build(base_dir, debug=False, rebuild=False): # TODO: at 2.0, keyword only
```
bryevdv: If this is top-level public API it needs a docstring

examples/app/surface3d/tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+    "outDir": "./dist/lib",
+    "paths": {
+      "@bokehjs/*": [
+        "./node_modules/bokehjs/build/js/lib/*",
+        "./node_modules/bokehjs/build/js/types/*"
+      ]
+    }
+  },
+  "include": ["./**/*.ts"]
+}
```
bryevdv: I'm really concerned about this. Creating extensions is already a hurdle for almost all users. Adding all this much more will just basically make it impossible. There's no way most users can be expected to be successful with this, especially without documentation support.

bryevdv:
I'm personally -1 on merging this without tests and documentation.

examples/app/surface3d/tsconfig.json
mattpap: Not six but two. The minimum setup is an `index.ts` and a `bokeh.ext.json`, which is an indicator that a directory is a bokeh extension and in future will provide extension's configuration. Having `index.ts` can be made configurable in future. Other files, including `package.json` and `tsconfig.json` are optional. If you don't need dependencies, there is no need for `package.json`. `package-lock.json` is a side effect of how npm works, and if you accept using dependencies, you have to accept this (or one can disable generating this file by npm, though this would require allowing to pass options to npm). If `tsconfig.json` isn't provided, a default one will be used provided by the compiler, but this will make any IDE refuse to work with your `*.ts` files (no auto-completion, no error checking, etc.).

This work is about allowing to create a workflow for developing extensions, especially large extensions like e.g. `panel`. Complexity is inevitable and exposing everything necessary it the right way to do. From here we can look into how we can improve this setup for small, possibly inline extensions.

mattpap:
Documentation and tests will come soon, but for now we need this merged to properly test panel.

bryevdv:
Here's what I would like to see:

* a docstring for `bokeh.ext.build`
* bare minimal unit test files for the new python modules
* examples pared down to the least amount of work for users, not the most. (There should be a "full" example somewhere later)
* An issue(s) for the follow-on work in the 1.4 milestone

PR #9200, opened 2019-08-29
Fix typos in docs, comments, etc

minho42:
Fixed some typos in docs, comments, string literals, etc.


bryevdv:
Wow, thank you @minho42 ! We have some unrelate issues causing failures in our CI pipelines right now, but we will get this merged as soon as those are sorted!

minho42:
Thanks.

PR #9203, opened 2019-08-29
Print full stacktrace to console when an error occurs

### LINKED ISSUE #9202: [FEATURE] Print full stacktrace on error
See discussion on [discussion forum](https://discourse.bokeh.org/t/debugging-recommendations/3934)

My underlying application is getting very big and when running into problems the debugging takes me a lot of time. 

For example I now just get the line:
2019-08-26 13:57:29,620 error handling message Message ‘EVENT’ (revision 1) content: ‘{“event_name”:“button_click”,“event_values”:{“model_id”:“1027”}}’: ValueError(‘Wrong number of items passed 2, placement implies 1’)

My desired solution is to see the stacktrace which includes the file and line of the source of the error.

@bryevdv pointed me to the protocol_handler.py script

The alternative is to set debug level, but I only want error handling improved, not have a full log. Also, I want the error to show directly, because I may not know how to reproduce it on a running server.

PR created

**Additional context**
I am not worried about the log getting too long. Errors should in general not occur, they should be fixed or at least be caught in case they are acceptable

harmbuisman:

- [ ] issues: fixes #9202 
- [ ] tests added / passed

### DIFF
--- a/bokeh/server/protocol_handler.py
+++ b/bokeh/server/protocol_handler.py
@@ -99,10 +99,8 @@ class ProtocolHandler(object):
         try:
             work = yield handler(message, connection)
         except Exception as e:
-            log.error("error handling message, run server with --log-level=debug"
-                      " to see more information \n message: %r \n error: %r",
+            log.error("error handling message\n message: %r \n error: %r",
                       message, e, exc_info=True)
-            log.debug("  message header %r content %r", message.header, message.content)
             work = connection.error(message, repr(e))
         raise gen.Return(work)
 
--- a/bokeh/server/protocol_handler.py
+++ b/bokeh/server/protocol_handler.py
@@ -99,7 +99,7 @@ class ProtocolHandler(object):
         try:
             work = yield handler(message, connection)
         except Exception as e:
-            log.error("error handling message %r: %r", message, e)
+            log.error("error handling message %r: %r", message, e, exc_info=1)
             log.debug("  message header %r content %r", message.header, message.content, exc_info=1)
             work = connection.error(message, repr(e))
         raise gen.Return(work)
--- a/bokeh/server/protocol_handler.py
+++ b/bokeh/server/protocol_handler.py
@@ -99,8 +99,10 @@ class ProtocolHandler(object):
         try:
             work = yield handler(message, connection)
         except Exception as e:
-            log.error("error handling message %r: %r", message, e, exc_info=1)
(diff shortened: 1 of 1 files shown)


bokeh/server/protocol_handler.py
```diff
@@ -99,7 +99,7 @@ def handle(self, message, connection):
         try:
             work = yield handler(message, connection)
         except Exception as e:
-            log.error("error handling message %r: %r", message, e)
+            log.error("error handling message %r: %r", message, e, exc_info=1)
             log.debug("  message header %r content %r", message.header, message.content, exc_info=1)
```
bryevdv: I think the `exc_info=1` should now be removed from here, so it is not duplicated on the error and debug lines. Also, I think it would be good to stat in the error message that turning on debug output will show the entire message (then anyone who wants it has all the information they need right there)

bryevdv:
Thanks for the PR @harmbuisman I had a small request above. Also the CI is currently having issues due to an unrelated problem, we are working resolving it.

harmbuisman:
> Thanks for the PR @harmbuisman I had a small request above. Also the CI is currently having issues due to an unrelated problem, we are working resolving it.

@bryevdv good suggestion, I removed the redundant stacktrace, added a suggestion to run in debug modus and changed 1 to True so that the code is more readable.

I do wonder though if the printing of the message in the error log is not the same as the message.contents in the debug log. The only difference in my repro is that debug prints the header now, but the contents seem to be the same.

bryevdv:
@harmbuisman I think the repo must have changed at some point. The `content` is potentially very large, which was the idea behind separating things out. But if that's the way things now I'm not inclined to change it. Maybe just delete the `debug` statement altogether, and revert the log message to remove the note about `debug` What do you think?

harmbuisman:
> @harmbuisman I think the repo must have changed at some point. The `content` is potentially very large, which was the idea behind separating things out. But if that's the way things now I'm not inclined to change it. Maybe just delete the `debug` statement altogether, and revert the log message to remove the note about `debug` What do you think?

@bryevdv, yes, let's do that, that would be the cleanest solution. I adapted the code.

bryevdv:
Thanks @harmbuisman !

harmbuisman:
> Thanks @harmbuisman !

@bryevdv, you're welcome, it is nice to be able to contribute. Thanks for thinking along

PR #9205, opened 2019-08-30
Don't depend on ipywidgets

mattpap:


mattpap:
There are more errors:
```
CondaUpgradeError: This environment has previously been operated on by a conda version that's newer
than the conda currently being used. A newer version of conda is required.
  target environment location: C:\Miniconda36-x64
  current conda version: 4.5.11
  minimum conda version: 4.7
```
I can safely presume that problems with widgetsnbextension were caused by conda and not by the package itself.

scripts/ci/appveyor/install.ps1
```diff
@@ -1,5 +1,5 @@
 function install() {
-    conda config --set auto_update_conda off
+    conda config --set auto_update_conda on
```
mattpap: @bryevdv, this "fixed" issues on Appveyor, but I'm not sure if this the way to go, and if so, should we change this on travis ci as well.

bryevdv: I would prefer to lock down the version of conda used. We have upgraded to >=4.7 on TravisCI. I don't think I did anything to update the Appveyor config though (not sure where it is specified offhand) I would say we could first try updating Appveyor to use the latest conda version, and keep auto-update off

PR #9206, opened 2019-08-30
Outdated sentence removed from docs

chesucr:
The sentence seems outdated because the new notation is `output_backend = 'webgl'` and saving plots as PNG with `output_backend = 'canvas'` works well.


bryevdv:
Thanks @chesucr we are fixing up some unrelated issues with out CI build pipelines, we will get this merged as soon as things are all back to green!

PR #9214, opened 2019-09-05
Adding the ability to parse globs in quotes in serve

### LINKED ISSUE #9209: Add support for globs to `bokeh serve`
I have been working on a way to make it easier to use `panel serve` to launch deployments at a proxied port on `jupyter notebook` calls, so that your deployment shows up at `localhost:8888/panel`. I've got this working for cases where you have a dir with one notebook in it called main.ipynb, but I'd like to have it look at all the notebooks and generate an index. This would be equivalent to calling `panel serve *.ipynb` but using a flag instead. The reason for the flag is that it isn't possible to pass the * in the [proxy code](https://github.com/pyviz/jupyter-panel-proxy/blob/b0d483d2774d159a065eae3432854d1196c8b5e0/panel_server/__init__.py#L3):

```
'command': ["panel", "serve", ".", "--allow-websocket-origin=*", "--port", "{port}", "--prefix", "{base_url}panel"],
```

For context this functionality would make it very easy to deploy apps on binder. The user would just need to include jupyter-panel-proxy in their env and stick `?urlpath=panel` at the end of the binder url.

These are the relevant PRs:
  - https://github.com/pyviz/jupyter-panel-proxy/pull/1
  - https://github.com/pyviz-demos/glaciers/pull/11

If we were able to add the `--all-notebooks` flag, then I'd be happy to write a jupyter-bokeh-proxy as well :)

jsignell:
- [x] issues: fixes #9209 
- [x] tests added / passed - still working on these

### DIFF
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -90,14 +90,14 @@ that all the files matching a particular pattern should be served.
 
     bokeh serve *.py
 
-Command line shells will normally expand the "*.py" automatically. However,
+Command line shells will normally expand the ``*.py`` automatically. However,
 if you are starting a Bokeh server programmatically, then filename arguments
 with globs may not be expanded by the shell. In situations like this, the
 ``--glob`` flag may be used to make the Bokeh server explicitly perform globbing:
 
 .. code-block:: python
 
-    subprocess.call(["bokeh", "serve", "--glob", "*py"])
+    subprocess.call(["bokeh", "serve", "--glob", "*.py"])
 
 Application Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/bokeh/command/subcommands/serve.py
+++ b/bokeh/command/subcommands/serve.py
@@ -90,12 +90,14 @@ that all the files matching a particular pattern should be served.
 
     bokeh serve *.py
 
-Note that when the filename is wrapped in quotes, the  ``--glob`` flag turns on
-glob parsing.
+Command line shells will normally expand the "*.py" automatically. However,
+if you are starting a Bokeh server programmatically, then filename arguments
+with globs may not be expanded by the shell. In situations like this, the
+``--glob`` flag may be used to make the Bokeh server explicitly perform globbing:
 
-.. code-block:: sh
+.. code-block:: python
 
-    bokeh serve '*.py' --glob
+    subprocess.call(["bokeh", "serve", "--glob", "*py"])
 
 Application Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


bokeh/command/subcommands/serve.py
```diff
@@ -83,6 +83,20 @@
...
 
+Another way to run multiple applications is using glob notation to indicate
+that all the files matching a particular pattern should be served.
+
+.. code-block:: sh
+
+    bokeh serve *.py
+
+Note that when the filename is wrapped in quotes, the  ``--glob`` flag turns on
+glob parsing.
```
bryevdv: I think we should be explicit about the most common time this is actually needed (and that it is probably not needed otherwise) to avoid making anyone think this is something they might normally need:

> If you are starting a Bokeh server programmatically, then filename arguments with globs may not be expanded by the system shell. In situation like this, the `--glob` flag may be used to make the Bokeh server explicitly perform globbing:

    subprocess.call(["bokeh", "serve", "--glob", "*py"])

bokeh/command/subcommands/serve.py
```diff
@@ -83,6 +83,20 @@
...
+
+.. code-block:: sh
+
+    bokeh serve '*.py' --glob
+
```
bryevdv: I think we should be explicit about the most common time this is actually needed (and that it is probably not needed otherwise) to avoid making anyone think this is something they might normally need on the command line:

> .. code-block:: sh
>
>     bokeh serve *.py
>
> Command line shells will normally expand the "*.py" automatically. However, if you are starting a Bokeh server programmatically, then filename arguments with globs may not be expanded by the shell. In situations like this, the `--glob` flag may be used to make the Bokeh server explicitly perform globbing:
> 
>     subprocess.call(["bokeh", "serve", "--glob", "*py"])

bryevdv:
@jsignell I only had one docs request otherwise LGTM

bokeh/command/subcommands/serve.py
```diff
@@ -588,8 +608,15 @@ def invoke(self, args):
         # even if Tornado is not installed
         from bokeh.server.server import Server
 
-        argvs = { f : args.args for f in args.files}
-        applications = build_single_handler_applications(args.files, argvs)
+        files = []
+        for f in args.files:
+            if args.glob and '*' in f:
```
mattpap: The second clause is unnecessary, because by definition `--glob` tells us that the following arguments are globs and, more importantly, `*` is not the only special symbol in globs (e.g. `?`, `[0-9]`).

jsignell: Thanks!

jsignell:
Thanks for you reviews!

bryevdv:
@jsignell Sphinx is persnickety, it was interpreting the `*` as the italic/emphasis marker. I think the small change I pushed should fix.

PR #9229, opened 2019-09-19
Bryanv/update docs automation

bryevdv:
This PR updates the build an deploy automation to upload to the the new S3 location for the docs. Old scripts and configurations that are no longer needed are also removed. Details:

* secure environment variables (except the google api key) are now configured from the TravisCI UI
* dependecies on fabric ~~and artifact~~ are removed
* old encrypted tarball is removed 
* old draw.io sources are removed


bryevdv:
@mattpap any idea about the image differences showing up in this PR report? 

https://ci.bokeh.org/travis/586817673/examples.html

I will have to investigate and see if it's new in this PR or ongoing

bryevdv:
@bokeh/dev OK I forgot where `artifact` was used, I don't want to deal with that in this PR though we probably should try to reduce the number of s3 tools at some point. This PR is green so I'd like to merge it at see whether a dev build works or not. Will merge some time tomorrow, absent objection.

bryevdv:
I can't find any other recent builds with image differences, just seemed to start with this one. I am going to restart the examples test build to make sure it was not some weird fluke. 

As an aside I just realized why the `ci.bokeh.org` bucket costs are higher than I had imagined they would be: besides people looking at the reports, there is egress for all the images on every examples test run.

bryevdv:
OK so the differences persist. I will restart a recent earlier success that did not have any diffs to see if diffs show up (i.e. some dependency change is causing)

bryevdv:
OK I restarted the lasts green merge to master, which previously had no image differences, and now it has the same differences after re-running

https://ci.bokeh.org/travis/584447805/examples.html

cc @mattpap

bryevdv:
I guess I should note that I only restarted the test, and not the BokehJS build step. So the changes in the report are not due to any change in BokehJS.  Given that, this seems like an infrastructure issue. I will merge this tomorrow at the latest to try and get dev/release builds working again.

PR #9232, opened 2019-09-22
Fixed issue of Datepicker displaying the wrong date for users in UTC+…

### LINKED ISSUE #9129: [BUG]Datepicker displayed value is not updating correctly
Hi ! First of all thanks for the amazing work.

I just updated bokeh to 1.3.0 and noticed that the displayed DatePicker value is not updating correctly after changing it manually. Basically, the underlying value responds well but the displayed value corresponds to the selected day - 1 day. So if you click on 30th July 2019, the displayed value is 29th July 2019. The displayed value updates correctly only if you select the 30th July a 2nd time.

I just checked with 1.2.0 and we don't seem to have this issue (tested on chrome, Edge and firefox).

Minimal Example code:
```
from bokeh.models import DatePicker
from bokeh.io import curdoc
import datetime

def callback(attr, old, new):
    print('old:', old, 'new:', new)

date_picker = DatePicker(value=datetime.date.today())

date_picker.on_change('value', callback)

curdoc().add_root(date_picker)
```

Thanks and sorry if it's not the right way to submit an issue, it's my first time doing this.

Software version info:
Bokeh - 1.3.0
Python - 3.7.3
Chrome - 75.0.3770.142

madkopp:
… timezones.

- [x] issues: fixes #9129
- [ ] tests added / passed

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
--- a/bokehjs/src/lib/models/widgets/date_picker.ts
+++ b/bokehjs/src/lib/models/widgets/date_picker.ts
@@ -76,10 +76,10 @@ export class DatePickerView extends InputWidgetView {
   }
 
   _unlocal_date(date: Date): Date {
-    //Get the UTC offset (in minutes) of Date (will be based on the timezone of the user's system).
+    //Get the UTC offset (in minutes) of date (will be based on the timezone of the user's system).
     //Then multiply to get the offset in ms.
-    //This way it can be used to recreate the user specified Date, agnostic to their local systems's timezone.
-    var timeOffsetInMS = date.getTimezoneOffset() * 60000
+    //This way it can be used to recreate the user specified date, agnostic to their local systems's timezone.
+    const timeOffsetInMS = date.getTimezoneOffset() * 60000
     date.setTime(date.getTime() - timeOffsetInMS)
 
     const datestr = date.toISOString().substr(0, 10)
--- a/bokehjs/src/lib/models/widgets/date_picker.ts
+++ b/bokehjs/src/lib/models/widgets/date_picker.ts
@@ -76,8 +76,12 @@ export class DatePickerView extends InputWidgetView {
   }
 
   _unlocal_date(date: Date): Date {
-    // this sucks but the date comes in as a UTC timestamp and pikaday uses Date's local
-    // timezone-converted representation. We want the date to be as given by the user
+    //Get the UTC offset (in minutes) of Date (will be based on the timezone of the user's system).
+    //Then multiply to get the offset in ms.
+    //This way it can be used to recreate the user specified Date, agnostic to their local systems's timezone.
+    var timeOffsetInMS = date.getTimezoneOffset() * 60000
+    date.setTime(date.getTime() - timeOffsetInMS)
+
     const datestr = date.toISOString().substr(0, 10)
     const tup = datestr.split('-')
     return new Date(Number(tup[0]), Number(tup[1])-1, Number(tup[2]))
--- a/tests/integration/widgets/test_datepicker.py
+++ b/tests/integration/widgets/test_datepicker.py
@@ -1,0 +1,71 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2012 - 2017, Anaconda, Inc. All rights reserved.
+#
+# Powered by the Bokeh Development Team.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Boilerplate
+#-----------------------------------------------------------------------------
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import pytest ; pytest
+
+#-----------------------------------------------------------------------------
+# Imports
+#-----------------------------------------------------------------------------
+
+# Standard library imports
+from datetime import datetime
+
+# External imports
+
+# Bokeh imports
+from bokeh.models.widgets import DatePicker
+from bokeh.models import ColumnDataSource
+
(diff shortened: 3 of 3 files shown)


madkopp:
@bryevdv I'm afraid I don't know enough about unit tests to understand the failures that occurred, hopefully it's nothing too complicated to fix.

bryevdv:
@madkopp that you for the PR. There is nothing to fix, we have a couple of flaky interrogation tests that have not been addressed yet. I've restarted that job

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
```
bryevdv: This file should be removed.

madkopp: Got it.  Removed and pushed changes.

bryevdv:
@madkopp OK this seems to be working great for me in PST

@mattpap @philippjfr  I'd like to merge this tomorrow if one or both of you can you test this out locally as well?

madkopp:
@bryevdv Glad to hear it.  Let me know if there's more testing that I can help with; I'm happy to do it, but to be honest I'm out of ideas of what to test.

bryevdv:
@madkopp Ideally we would add some integration tests for this. There aren't any at all yet for this widget. If you are interested in looking in to that, I think this would be a somewhat similar case to emulate:

https://github.com/bokeh/bokeh/blob/master/tests/integration/widgets/test_radio_button_group.py

Perhaps there might even be a way we could futz with the time zone on the test system to run things a few times in different time zones? 

Otherwise, it would be good just to make a separate issue about this.

madkopp:
@bryevdv I made a shell of what the integration test(s) might look like.  Unfortunately, though, I have very little experience with unit/integration testing.  Case and point, I forgot that by pushing it would automatically add the commit to this pull request, so I'm afraid I may have messed things up some.  Let me know if you want me to revert anything, or if you want to just charge on and fix the datepicker_test in place.

bryevdv:
@madkopp I can't seem to push to your branch easily, here is a new full integration test file that I would suggest:
```
# Copyright (c) 2012 - 2017, Anaconda, Inc. All rights reserved.
#
# Powered by the Bokeh Development Team.
#
# The full license is in the file LICENSE.txt, distributed with this software.

# Boilerplate
from __future__ import absolute_import, division, print_function, unicode_literals

import pytest ; pytest

# Imports

# Standard library imports
from datetime import datetime

# External imports

# Bokeh imports
from bokeh.models import Circle, ColumnDataSource, CustomAction, CustomJS, DatePicker, Plot, Range1d
from bokeh._testing.util.selenium import RECORD

# Tests

pytest_plugins = (
    "bokeh._testing.plugins.bokeh",
)

@pytest.mark.integration
@pytest.mark.selenium
class Test_DatePicker(object):

    def test_basic(self, bokeh_model_page):
        dp = DatePicker(title='Select date', value=datetime(2019, 9, 20), min_date=datetime(2019, 9, 1), max_date=datetime.utcnow(), css_classes=["foo"])

        page = bokeh_model_page(dp)

        el = page.driver.find_element_by_css_selector('.foo label')
        assert el.text == "Select date"

        assert page.has_no_console_errors()

    def test_js_on_change_executes(self, bokeh_model_page):
        dp = DatePicker(title='Select date', value=datetime(2019, 9, 20), min_date=datetime(2019, 9, 1), max_date=datetime.utcnow(), css_classes=["foo"])
        dp.js_on_change('value', CustomJS(code=RECORD("value", "cb_obj.value")))

        page = bokeh_model_page(dp)

        el = page.driver.find_element_by_css_selector('.foo input')
        el.click()

        el = page.driver.find_element_by_css_selector('button[data-pika-day="16"]')
        el.click()

        results = page.results
        assert results['value'] == 'Mon Sep 16 2019'

        el = page.driver.find_element_by_css_selector('.bk-input')
        assert el.get_attribute('value') == 'Mon Sep 16 2019'

        assert page.has_no_console_errors()

    def test_server_on_change_round_trip(self, bokeh_server_page):
        def modify_doc(doc):
            source = ColumnDataSource(dict(x=[1, 2], y=[1, 1], val=["a", "b"]))
            plot = Plot(plot_height=400, plot_width=400, x_range=Range1d(0, 1), y_range=Range1d(0, 1), min_border=0)
            plot.add_tools(CustomAction(callback=CustomJS(args=dict(s=source), code=RECORD("data", "s.data"))))
            plot.add_glyph(source, Circle(x='x', y='y', size=20))
            dp = DatePicker(title='Select date', value=datetime(2019, 9, 20), min_date=datetime(2019, 9, 1), max_date=datetime.utcnow(), css_classes=["foo"])
            def cb(attr, old, new):
                source.data['val'] = [old, new]
            dp.on_change('value', cb)
            doc.add_root(column(dp, plot))

        page = bokeh_server_page(modify_doc)

        el = page.driver.find_element_by_css_selector('.foo input')
        el.click()

        el = page.driver.find_element_by_css_selector('button[data-pika-day="16"]')
        el.click()

        page.click_custom_action()

        results = page.results
        d0 = datetime.utcfromtimestamp(results['data']['val'][0]/1000)
        assert d0.timetuple()[:3] == (2019, 9, 20)
        d1 = datetime.utcfromtimestamp(results['data']['val'][1]/1000)
        assert d1.timetuple()[:3] == (2019, 9, 16)

```

bryevdv:
realized I could edit the file from the web UI

bryevdv:
OK I can't get the flaky integration test to pass on the branch. Not sure why this happens sometimes. I am going to go ahead and merge to see if it resolves on master.

bryevdv:
Thanks @madkopp !

madkopp:
@bryevdv Awesome, glad this worked out!

Sorry, I was slammed all week, but to be perfectly honest, I wouldn't have been able to put together these tests, anyways (I'm working on understanding all that - if you know of any good resources, I'd be more than appreciative).

A couple of questions about the tests, though:

1) It looks like for the tests we use the datetime.utcnow() function to define the max date.  Might it be better to just pick an arbitrary, but fixed date range?  It's always possible that someone's system is set to the wrong date/time, and that utcnow() returns a date before the min date.  It's also possible that their system is set to a date/time WAY in the future, thereby creating such a huge date range that it takes up memory - I may be overthinking things there, though.  Either way, using a fixed date range would alleviate these concerns, and allow you to expand the tests by including dates before/after daylight savings time changes.  Just a thought, I'd be interested in hearing your thoughts on whether these are actually valid concerns.

2) We haven't done anything to change the system's timezone to test if that functionality holds up.  I looked into this some.  In general it seems like that's possible to do with Python, but you have to start using winAPI calls, and probably something similar in UNIX environments (see https://stackoverflow.com/questions/20256946/how-to-change-system-timezone-using-python-in-windows).  It might not be a good idea to start changing system level things like that, especially not on the TravisCI workers.  Again, though, I'd be interested in hearing your thoughts.

PR #9247, opened 2019-09-27
Check for empty object

### LINKED ISSUE #8402: No clean way to update vbar_stack
Hi there, 

There is no exemple of how to update a vbar_stack correctly. I tried this but the delete and the update of the plot take a long time and is done consequently for each vertical bar (each bar is displayed one by one). 
This method is not very clean because it throws JS error  : "Uncaught TypeError: Cannot read property 'draw_legend' of undefined". 
The update of the plot is not like any source update for any other kind of plot... need to be fixed...

```
from bokeh.core.properties import value
from bokeh.io import curdoc
from bokeh.palettes import Category20b,Category20c
import random
from bokeh.layouts import row
from bokeh.models import GlyphRenderer,Legend, HoverTool


fruits = ['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries']
years = ["2015", "2016", "2017"]
colors = ["#c9d9d3", "#718dbf", "#e84d60"]

data = {'fruits' : fruits,
        '2015'   : [2, 1, 4, 3, 2, 4],
        '2016'   : [5, 3, 4, 2, 4, 6],
        '2017'   : [3, 2, 4, 4, 5, 3]}

p = figure(x_range=fruits, plot_height=500, plot_width=500, 
           title="Fruit Counts by Year",
           toolbar_location=None, tools="hover", 
           tooltips="$name @fruits: @$name")

p.vbar_stack(years, x='fruits', width=0.9, color=colors, source=data,
             legend=[value(x) for x in years])

p.y_range.start = 0
p.x_range.range_padding = 0.1
p.axis.minor_tick_line_color = None
p.outline_line_color = None
p.legend.orientation = "horizontal"

button_update = Button(label="Update")

def update_data_bar_chart():

    fruits_updated = ['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries']
    years_updated = ["2000", "2001", "2002", "2003", "2004",  "2005",  "2006", 
                     "2007", "2008", "2009", "2010", "2011", "2012", "2013", 
                     "2014", "2015", "2016", "2017", "2018"]
    my_palette_updated=Category20b[20]+Category20c[20]
    colors_updated = random.sample(my_palette_updated, len(years_updated))

    data_updated = {'fruits' : fruits_updated,
        '2000' : random.sample(range(1,7), 6), 
        '2001' : random.sample(range(1,7), 6),   
        '2002' : random.sample(range(1,7), 6),   
        '2003' : random.sample(range(1,7), 6),  
        '2004' : random.sample(range(1,7), 6),   
        '2005' : random.sample(range(1,7), 6),   
        '2006' : random.sample(range(1,7), 6),   
        '2007' : random.sample(range(1,7), 6),  
        '2008' : random.sample(range(1,7), 6),   
        '2009' : random.sample(range(1,7), 6),   
        '2010' : random.sample(range(1,7), 6),   
        '2011' : random.sample(range(1,7), 6),  
        '2012' : random.sample(range(1,7), 6),   
        '2013' : random.sample(range(1,7), 6),   
        '2014' : random.sample(range(1,7), 6),   
        '2015' : random.sample(range(1,7), 6),  
        '2016' : random.sample(range(1,7), 6),  
        '2017' : random.sample(range(1,7), 6),   
        '2018' : random.sample(range(1,7), 6)
            }
    
    for i in range(len(p.renderers), 1, -1):
        if type(p.renderers[i-1]) == GlyphRenderer:
            print("glyph")
            p.renderers.pop(i-1)
        elif type(p.renderers[i-1]) == Legend:
            print("legend")
            p.renderers.pop(i-1)
   
    p.vbar_stack(years_updated, x='fruits', width=0.9, color=colors_updated, 
                  source=data_updated,
                  legend=[value(x) for x in years_updated])
    
    p.x_range.factors=fruits_updated
    p.y_range.start = 0
    p.x_range.range_padding = 0.1
    p.axis.minor_tick_line_color = None
    p.outline_line_color = None
    p.legend.orientation = "horizontal"
    p.add_tools(HoverTool(tooltips="$name @fruits: @$name"))

button_update.on_click(update_data_bar_chart)

curdoc().add_root(row(button_update,p))

```

Versions: 
bokeh==1.0.0
Python 3.7.0

OS: Windows

Thank you in advance and thanks for Bokeh !

kennedydane:

- [x] issues: fixes #8402
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/expressions/stack.ts
+++ b/bokehjs/src/lib/models/expressions/stack.ts
@@ -29,11 +29,12 @@ export class Stack extends Expression {
   protected _v_compute(source: ColumnarDataSource): Arrayable<number> {
     const result = new Float64Array(source.get_length() || 0)
     for (const f of this.fields) {
-      if (Object.keys(source.data).length != 0) {
-        for (let i = 0; i < source.data[f].length; i++) {
-          const x = source.data[f][i]
-          result[i] += x
-        }
+      if (Object.keys(source.data).length == 0) {
+        break;
+      }
+      for (let i = 0; i < source.data[f].length; i++) {
+        const x = source.data[f][i]
+        result[i] += x
       }
     }
     return result
--- a/bokehjs/src/lib/models/expressions/stack.ts
+++ b/bokehjs/src/lib/models/expressions/stack.ts
@@ -27,14 +27,14 @@ export class Stack extends Expression {
   }
 
   protected _v_compute(source: ColumnarDataSource): Arrayable<number> {
-    const result = new Float64Array(source.get_length() || 0)
+    const n = source.get_length() || 0 // TODO: use ?? in TS 3.7
+    const result = new Float64Array(n)
     for (const f of this.fields) {
-      if (Object.keys(source.data).length == 0) {
-        break;
-      }
-      for (let i = 0; i < source.data[f].length; i++) {
-        const x = source.data[f][i]
-        result[i] += x
+      const column = source.data[f]
+      if (column != null) {
+        for (let i = 0, k = Math.min(n, column.length); i < k; i++) {
+          result[i] += column[i]
+        }
       }
     }
     return result
--- a/bokehjs/src/lib/models/expressions/stack.ts
+++ b/bokehjs/src/lib/models/expressions/stack.ts
@@ -29,9 +29,11 @@ export class Stack extends Expression {
   protected _v_compute(source: ColumnarDataSource): Arrayable<number> {
     const result = new Float64Array(source.get_length() || 0)
     for (const f of this.fields) {
-      for (let i = 0; i < source.data[f].length; i++) {
-        const x = source.data[f][i]
-        result[i] += x
+      if (Object.keys(source.data).length != 0) {
+        for (let i = 0; i < source.data[f].length; i++) {
+          const x = source.data[f][i]
+          result[i] += x
+        }
       }
     }
     return result


bokehjs/src/lib/models/expressions/stack.ts
```diff
@@ -29,9 +29,11 @@ export class Stack extends Expression {
...
     const result = new Float64Array(source.get_length() || 0)
     for (const f of this.fields) {
-      for (let i = 0; i < source.data[f].length; i++) {
-        const x = source.data[f][i]
-        result[i] += x
+      if (Object.keys(source.data).length != 0) {
+        for (let i = 0; i < source.data[f].length; i++) {
+          const x = source.data[f][i]
+          result[i] += x
+        }
```
bryevdv: @kennedydane columns in a single CDS are required to all be the same length, so if the first column is empty it is safe to bail at that point immediately. I would propose:
```
if (Object.keys(source.data).length == 0 {
  break
}
for (let i = 0; i < source.data[f].length; i++) {
  const x = source.data[f][i]
  result[i] += x
}
```
which also reduces the overall nesting level.

bryevdv:
@kennedydane Can you comment on this some, or provide a tangible use case that demonstrates when this situation would come up? I guess offhand I would consider applying a transform to a CDS with no columns to be an error, in which case this would make the error silently ignored and I am not sure that's s good thing.

bryevdv:
OK I missed that this was for #8402 I'm traveling this weekend but let me think about this a little, it might be that we only want to loosen the checks for `AjaxDataSource` and not for a regular `ColumnDataSource` e.g.

kennedydane:
Hi @bryevdv,

Sorry I didn't describe what this solves adequately, but I think you since found my reasoning (possibly in the wrong issue :(). But yeah, basically I've been battling to do stacked vbar charts using an Ajaxdatasource. (I imagine the same applies to other stacked charts?) This code fixes it. I'm a complete hack when it comes to javascript through. Your update still fixes it...

bokehjs/src/lib/models/expressions/stack.ts
```diff
@@ -29,6 +29,9 @@ export class Stack extends Expression {
   protected _v_compute(source: ColumnarDataSource): Arrayable<number> {
     const result = new Float64Array(source.get_length() || 0)
     for (const f of this.fields) {
+      if (Object.keys(source.data).length == 0) {
+        break;
+      }
```
mattpap: This should be moved out one level. However, a better approach would be to check if a field `f` exists on `source.data`. This way partially initialized data sources would still be usable (if that makes sense), or at least the wouldn't fail.

```ts
const n = source.get_length() || 0
const result = new Float64Array(n)
for (const f of this.fiels) {
  const column = source.data[f]
  if (column != null) {
    for (let i = 0, k = Math.min(n, column.length); i < k; i++) {
     result[i] += column[i]
    }
  }
  return result
}
```

bryevdv: @kennedydane will you be able to make the change suggested above?

bryevdv:
I will test this tomorrow and try to get it merged for 1.4

bryevdv:
Being optimistic, I'll go ahead and merge this and test in the RC

kennedydane:
Hi sorry, haven't been watching this. Thanks so much for the change :-).

PR #9250, opened 2019-09-28
allow auto-bounds to consider glyph visibility

### LINKED ISSUE #9144: Axis rescaled when legend item visibility changed [FEATURE]
Currently, if legend click_policy = "hide" the legend item is hidden but the axes are not rescaled. This is fine most of the time, but there are circumstances where you'd like the axes to be rescaled, for example, if the hidden item (e.g. line) is much greater than the other items.

Here's a view of the current behavior. Chart 1 shows the full chart, chart 2 shows what's shown when Y3 is hidden.

Chart 1
bokeh_plot (1)

Chart 2
bokeh_plot (2)

Here's what I would like to happen if the appropriate setting is made:
Chart 3
bokeh_plot (3)

I don't think this should be the default behavior. I suggest a new attribute something like: hide_rescale which is False by default (current behavior). If hide_rescale is set to True, the chart axes would be rescaled.

This has been discussed on the Discourse group: https://discourse.bokeh.org/t/bokeh-update-zoom-plot-or-axis-rescaling-when-hide-series-on-legend/2708/4

bryevdv:
- [x] issues: fixes #9144
- [x] tests added / passed

@mattpap I did about the simplest thing possible. If you have thoughts on a different or better way to trigger an `update_dataranges`, please let me know. I may try to add integration tests for interactive legends in this PR, since there are none at all presently.


bryevdv:
OK I don't have the wherewithal to make a test starting from an interactive legend right now, but this integration test exercises `only_visible` in the presence of visible and invisible glyphs and also when visibility is toggled.

@mattpap if I don't hear anything by tomorrow, even just to ask for more time if necessary, then I will assume you are either OK or don't want to comment, and will merge.

PR #9253, opened 2019-10-01
Fixed image hover value on inverted axes

### LINKED ISSUE #8778: Hover over image is showing wrong @image tooltip on flipped axis
This is continuation of this: 
https://github.com/bokeh/bokeh/pull/8031#issuecomment-475488255
I second the comment above.
```python
from matplotlib import pyplot as plt
from skimage.data import astronaut
c = astronaut().mean(axis=-1)[::2,:]
plt.imshow(c)
plt.show()
```
Gives me this
Screen Shot 2019-03-21 at 10 52 08 PM

While this bokeh (1.0.4) code
```python
scale = 1.0
p = figure(x_range=(1, c.shape[1]), y_range=(c.shape[0],1),
           plot_width=int(c.shape[1]*scale), plot_height=int(c.shape[0]*scale),
           tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")],
           tools=["hover"])
print(c[89:91,399:401])
p.image(image=[np.flipud(c)],
        x=1, y=c.shape[0],
        dw=c.shape[1], dh=c.shape[0])
show(p)
```
gives me this
Screen Shot 2019-03-21 at 10 59 39 PM

As you can see - image value is not correct - it seems to be reading it from different region of the image.
Hope this reproducer is clear enough to help dig into it.
Please note that it crucially important to keep Y axis increasing from top to bottom.

philippjfr:
This ensures that the `ImageIndex` generated by hover correctly inverts the indexes when the axis range is inverted. The initial attempt at fixing this (https://github.com/bokeh/bokeh/pull/8031/) did not invert the index.

Screen Shot 2019-10-01 at 1 00 33 PM

- [x] issues: fixes #8778
- [ ] tests added / passed

### DIFF
--- a/bokehjs/src/lib/models/glyphs/image_base.ts
+++ b/bokehjs/src/lib/models/glyphs/image_base.ts
@@ -120,8 +120,12 @@ export class ImageBaseView extends XYGlyphView {
     const height = this._height[index]
     const dx = (r - l) / width
     const dy = (t - b) / height
-    const dim1 = Math.floor((x - l) / dx)
-    const dim2 = Math.floor((y - b) / dy)
+    let dim1 = Math.floor((x - l) / dx)
+    let dim2 = Math.floor((y - b) / dy)
+    if (this.renderer.xscale.source_range.is_reversed)
+      dim1 = width-dim1-1
+    if (this.renderer.yscale.source_range.is_reversed)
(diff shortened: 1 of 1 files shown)


bryevdv:
Thank you @philippjfr !

PR #9270, opened 2019-10-10
Invalidate assets in folder and subfolders

### LINKED ISSUE #9264: CDN invalidations too narrow
Need a wildcard, e.g. 
```
aws cloudfront create-invalidation --distribution-id <<distid>> --paths '/en/dev/*'
```

osule:
All assets with the path prefix in Amazon Cloudfront
CDN are invalidated in its cache.

- [ ] issues: fixes #9264 
- [ ] ~tests added / passed~
- [ ] ~release document entry (if new feature or API change)~

### DIFF
--- a/scripts/build_upload.py
+++ b/scripts/build_upload.py
@@ -387,11 +387,11 @@ def upload_docs():
 
     if V(CONFIG.version).is_prerelease:
         run(sync_cmd % "dev")
-        run(invalidate_cmd % "/en/dev")
+        run(invalidate_cmd % '"/en/dev*"')
     else:
         run(sync_cmd % CONFIG.version)
         run(sync_cmd % "latest")
-        paths =  "/en/latest /versions.json" % CONFIG.version
+        paths = '"/en/latest*" /versions.json'
         run(invalidate_cmd % paths)
     cd("..")
 


scripts/build_upload.py
```diff
@@ -387,11 +387,11 @@ def upload_docs():
 
     if V(CONFIG.version).is_prerelease:
         run(sync_cmd % "dev")
-        run(invalidate_cmd % "/en/dev")
+        run(invalidate_cmd % '"/en/dev*"')
     else:
         run(sync_cmd % CONFIG.version)
         run(sync_cmd % "latest")
-        paths =  "/en/latest /versions.json" % CONFIG.version
+        paths = '"/en/latest*" /versions.json'
```
bryevdv: I *think* we might need a `/` in there before the `*` the CDN docs were unclear to me but all their examples looked like that, e.g., /en/latest/*` with the `/` 

Thanks for catching the string formatting error there, that would not have shown up otherwise until we tried to release and it blew up!

osule: @bryevdv In [Cloudfront example docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html), the /en/dev/* pattern invalidates all objects in the directory whereas /en/dev* invalidates all objects prefixed with the pattern effectively invalidating all objects in the directory and further subdirectories.

The latter should be fitting in this case.

bryevdv:
Thanks @osule !!

PR #9275, opened 2019-10-12
Improve error message if import of channels fails.

### LINKED ISSUE #9274: If import of channels fails, improve error message to install it via pip
Running `pytest tests/codebase/test_python_execution_with_OO.py`fails if channels is not installed:
```
tests/codebase/test_python_execution_with_OO.py:77: AssertionError
------------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------------
  File "/Users/mav/dev/python/bokeh/bokeh/util/dependencies.py", line 87, in import_required
    return import_module(mod_name)
  File "/Users/mav/anaconda3/envs/bokeh-dev/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'channels'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 163, in <module>
  File "/Users/mav/dev/python/bokeh/bokeh/server/django/__init__.py", line 9, in <module>
    import_required("channels", "channels is required by bokeh.server.django")
  File "/Users/mav/dev/python/bokeh/bokeh/util/dependencies.py", line 89, in import_required
    raise RuntimeError(error_msg) from e
RuntimeError: channels is required by bokeh.server.django
```

channels cannot be installed via conda at the moment, but should be installed via pip.
The error message should indicate that

steffenschroeder:
Currently channels in only availeble on pip, but not conda.
Thus the conda install mentioned in the developer guide will not install it.
To run pytest -m codebase (mentioned in the developer guide) to be added as
git pre-commit hook will fail, giving first-time-contributers a bad experience

Point the users to install it via pip instead

- [x] issues: fixes #9274
- [x] tests added / passed

### DIFF
--- a/bokeh/server/django/__init__.py
+++ b/bokeh/server/django/__init__.py
@@ -6,7 +6,8 @@ if six.PY2:
 from bokeh.util.dependencies import import_required
 
 import_required("django", "django is required by bokeh.server.django")
-import_required("channels", "channels is required by bokeh.server.django")
+import_required("channels", "The package channels is required by bokeh.server.django. "
+                            "and must be installed'")
 
 from .apps import DjangoBokehConfig
 default_app_config = "bokeh.server.django.DjangoBokehConfig"


bryevdv:
@steffenschroeder Thank you for the PR.

Personally, I am not inclined to be so specific. It's entirely possible `channels` could get added to conda-forge, if not defaults, for instance, and then this message would be relaying bad information (with no way to ever update in old versions). In general I think our obligation is to inform users *what* packages are necessary, but not stipulate the *how*. 

However, I will defer if other @bokeh/dev folks feel differently.

p-himik:
I think such information belongs in the developer documentation rather than in the codebase.

bryevdv:
I am inclined to agree.  I do think the current message could slighly improved by 

> The package "channels" must be installed...

just to emphasize that it is a package ("channels" is not quite as obviously a package name as .e.g "scikit-learn")

steffenschroeder:
Changed the error message to `The package channels is required by bokeh.server.django. and must be installed`
I can add this to the dev guide as well

bryevdv:
@steffenschroeder awesome that would be perfect, thank you!

PR #9278, opened 2019-10-12
Consistently use https protcol for cdn.pydata.org URLs

slokhorst:
Most browsers will not load http scripts from https pages. A few links were already using https, but some were not. http://cdn.pydata.org redirects to https://cdn.pydata.org so the loaded scripts will be the same.


bryevdv:
Thanks or the PR @slokhorst !

Issue #9280, opened 2019-10-12
Simplify dependency installation of developer environment setup


mansenfranzen:
Setting up bokeh's developer environment consists of several different steps. The section [Installing dependencies](https://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html#installing-dependencies) could be simplified into a single command like `python scripts/install_dev_env.py`.

This should work across all operating systems (and shells) while moving the corresponding steps into a separate python file which coordinates the subsequent steps.

bryevdv:
This was recently done, see new shorter instructions at https://docs.bokeh.org/en/latest/docs/dev_guide/setup.html#devguide-setup

PR #9281, opened 2019-10-12
Simplify package dependency installation for development environment setup

mansenfranzen:
This PR addresses a simplified setup of the python/conda [dependency installation](https://bokeh.pydata.org/en/latest/docs/dev_guide/setup.html#installing-dependencies). Instead of needing several distinct steps (add conda channels, install jinja and pyyaml, select os dependant command line), this PR proposes a simplified approach moving the mentioned steps into a separate python module. 

### Motivation
With this PR, it would be sufficient to run `python scripts/install_dev_env.py`. It should work regardless of operating system and shell. 

### Modification
This PR modifies `scripts/dep.py` to expose its functionality to other modules without changing its actual behavior. Running `scripts/dep.py` still yields the same result.
In addition, `scripts/install_dev_env.py` handles the simplified conda packages installation while relying on `scripts/dep.py`.

### Limitations/considerations:
- requires an existing python interpreter accessible via PATH
- likely works with python 3.5+ only
- only tested under windows 10
- running `install_dev_env.py` does not ask the user for confirmation of package installation in the currently activated conda environment (which is similar to before)

### ToDo
- [x] issues: #9280
- [ ] check correct installation behavior on different operation systems
- [ ] if accepted, the setup documentation needs to modified accordingly


scripts/deps.py
```diff
@@ -1,59 +1,106 @@
...
 import platform
 import sys
-
 import jinja2
 import setuptools
 import yaml
+from unittest import mock
+from functools import reduce
+from operator import add
+from pathlib import Path
```
mattpap: This is available only on Python 3. I think instead of making this compatible with Python 2, this branch could be rebased on `landing-2.0`, which is 3 only (especially we will be merging it soon).

mansenfranzen: @mattpap I rebased this branch on `landing-2.0`.

bryevdv:
I changed the base to `master` but that did not make the confict/diff situation any better. I am going to close and decline this PR. @mansenfranzen please feel free to submit a new PR if you are still interested in this (now against master since `landing-2.0` has been merged.)

PR #9282, opened 2019-10-12
Fix density parameter of streamlines function to accept tuples.

mborus:
small fix found during pyconde19 sprints.

### DIFF
--- a/examples/plotting/file/vector.py
+++ b/examples/plotting/file/vector.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import numpy as np
+import collections.abc
 
 from bokeh.layouts import gridplot
 from bokeh.plotting import figure, show, output_file
@@ -11,7 +12,7 @@ def streamlines(x, y, u, v, density=1):
     * x and y are 1d arrays defining an *evenly spaced* grid.
     * u and v are 2d arrays (shape [y,x]) giving velocities.
     * density controls the closeness of the streamlines. For different
-      densities in each direction, use a tuple or list [densityx, densityy].
+      densities in each direction, use a tuple [densityx, densityy].
 
     '''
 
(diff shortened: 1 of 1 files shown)


bryevdv:
Though to be perfectly honest, I would also not mind just leaving the example the way it is. If `streamlines` were some built-in facility of Bokeh then it certainly make sense for it to be as general as possible, but it's not that. It's just a function that only exists to be called inside one example&mdash;accordingly it only really needs to support just the usage in the example.

bryevdv:
Right, so the actual only usage in the example is with a homogenous density:
```
xs, ys = streamlines(xx, yy, U.T, V.T, density=2)
```
The purpose of the example is to demosntrate bokeh multiline  glyphs, not how to streamline in the most general way. I would much prefer to update the docstring to the function instead, and state that it only takes a single number as a `density` (no sequences at all)

bryevdv:
Thinking even more out loud: I'd be happy to take in an optional dependency e,g. to `scipy` if we could shave 150 lines off this example. Not doing e.g. an rk4 integration by hand in the example, would be the best sort of improvement.

mborus:
> 
> 
> `collections.abc` does not exist on python 2.7, which is still supported for one more upcoming Bokeh 1.x release. The best solution at this point is to update the PR to be against the `landing-2.0` branch, instead of `master`. (Please don't update the PR to use `six`, that would just make more work to undo on the `landing-2.0` branch which does not use `six` anymore)

Applying this to "landing-2.0 " makes sense. This came up when I was adding type hints to for the landing-2.0 branch, when after adding the hints mypy suggested that this is wrong. To the point of the request mainly is to allow type hinting the landing-2.0 branch.

mborus:
> 
> 
> Right, so the actual only usage in the example is with a homogenous density:
> 
> ```
> xs, ys = streamlines(xx, yy, U.T, V.T, density=2)
> ```
> 
> The purpose of the example is to demosntrate bokeh multiline glyphs, not how to do streamlining in the most general way. Making the example longer only obscures the pedagogical intent around using mulit-lines. I would much prefer to update the docstring to the function instead, and state that it only takes a single number as a `density` (no mention or handling of sequences at all).

OK, that is probably the easiet fix.

PR #9090, opened 2019-07-16
Add more static type declarations to bokeh

mattpap:
Given that `landing-2.0` branch is being rebased, this merges together PRs:

- https://github.com/bokeh/bokeh/pull/9079
- https://github.com/bokeh/bokeh/pull/9077
- https://github.com/bokeh/bokeh/pull/9072
- https://github.com/bokeh/bokeh/pull/9071

to speed up development and unify changes introduced there when they intersect.

addresses #9064 

fixes #9070

PR #9285, opened 2019-10-13
Landing 2.0b

mborus:
Removing tuple/list from streamlines/density parameter comment to assist in later type hinting.


examples/plotting/file/vector.py
```diff
@@ -1,15 +1,16 @@
+from __future__ import division
```
bryevdv: No need for this in Python 3

mattpap:
I merged this in PR #9090.

